PR #2, opened 2017-01-23
example: hello: ask odp for available core

muvarov:
Core 0 might be not available and app should
ask for available core before pinning to it.
https://bugs.linaro.org/show_bug.cgi?id=2806


muvarov:
Review-by: Test <tester@test.com>

PR #10, opened 2017-04-03
[PATCH API-NEXT v1] support for SHA-1/-512

lumag:
- major consolidation of cryptographic code
- support for SHA-1/-512

PR #11, opened 2017-04-03
[PATCH v1] crypto rework

lumag:
Port of crypto rework patchset (#10) to the master branch.


lumag:
Nice enough. Build failures are due to checkpatch.pl false positives.

Bill-Fischofer-Linaro:
We already have issues with checkpatch.pl being overly aggressive with
volatile and extern warnings. The camelCase warnings fall into the same
category. The prohibition should be that you may not define camelCase names
/ functions, not that you may not use them, as the use is typically
required by external packages like OpenSSL.

Not sure if checkpatch can make that distinction, but if it cannot, I'd
vote to nuke that check and make that a manual review criteria.

lumag:
Please ignore patch 4 for now, I should rething it (and original crypto implementation) wrt thread safety.

PR #14, opened 2017-04-19
linux-generic: crypto: properly handle errors in packet copy

lumag:
Add proper handling for errors returned by odp_packet_copy_from_pkt().

### DIFF
--- a/platform/linux-generic/odp_crypto.c
+++ b/platform/linux-generic/odp_crypto.c
@@ -871,14 +871,17 @@ odp_crypto_operation(odp_crypto_op_param_t *param,
 	odp_crypto_alg_err_t rc_auth = ODP_CRYPTO_ALG_ERR_NONE;
 	odp_crypto_generic_session_t *session;
 	odp_crypto_op_result_t local_result;
+	odp_bool_t allocated = false;
 
 	session = (odp_crypto_generic_session_t *)(intptr_t)param->session;
 
 	/* Resolve output buffer */
 	if (ODP_PACKET_INVALID == param->out_pkt &&
-	    ODP_POOL_INVALID != session->p.output_pool)
+	    ODP_POOL_INVALID != session->p.output_pool) {
 		param->out_pkt = odp_packet_alloc(session->p.output_pool,
 				odp_packet_len(param->pkt));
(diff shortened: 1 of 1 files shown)


lumag:
@Bill-Fischofer-Linaro again this is a question of having checks. I have stumbled upon this, because during example development I ended up with ODP_PACKET_INVALID in out_pkt and strange crash which took me some time to debug. So I'd prefer to have a check, rather than not to.

Bill-Fischofer-Linaro:
I'm not sure how that would happen given the immediately preceding code:

	/* Resolve output buffer */
	if (ODP_PACKET_INVALID == param->out_pkt &&
	    ODP_POOL_INVALID != session->p.output_pool)
		param->out_pkt = odp_packet_alloc(session->p.output_pool,
				odp_packet_len(param->pkt));

	if (odp_unlikely(ODP_PACKET_INVALID == param->out_pkt)) {
		ODP_DBG("Alloc failed.\n");
		return -1;
	}

If out_pkt is invalid the call should simply fail with a -1 return.

lumag:
@Bill-Fischofer-Linaro it will, if `session->p.output_pool` is not ODP_POOL_INVALID.

Bill-Fischofer-Linaro:
If session->p.output_pool is not ODP_POOL_INVALID then the attempt to allocate out_pkt from that pool will be made. If that allocation fails then you'd still get a -1 return and out_pkt would still be ODP_PACKET_INVALID.

lumag:
But if out_pool and out_pkt are bot _INVALID, you won't get a packet. You will get a nice crash.

Bill-Fischofer-Linaro:
If out_pkt and pool are both invalid, the allocate call is skipped and then the check to return -1 if out_pkt is invalid should trigger, no?

lumag:
I'd just prefer to check the return of underlying API.

muvarov:
@Bill-Fischofer-Linaro  can you please email Review-by to ML. Transfer from email to this page work. Transfer back not yet.

Bill-Fischofer-Linaro:
I've added my Review via the mailing list as well. Thanks.

lumag:
Merged

PR #19, opened 2017-04-26
[PATCH v1] linux-generic: rwlock: fix odp_rwlock_read_trylock()

lumag:
odp_rwlock_read_trylock() currently works only if there are no readers
(and writers) as it compares counter with 0. Make it actually work in
case there are other active readers.

### DIFF
--- a/platform/linux-generic/odp_rwlock.c
+++ b/platform/linux-generic/odp_rwlock.c
@@ -33,9 +33,14 @@ void odp_rwlock_read_lock(odp_rwlock_t *rwlock)
 
 int odp_rwlock_read_trylock(odp_rwlock_t *rwlock)
 {
-	uint32_t zero = 0;
+	uint32_t cnt = odp_atomic_load_u32(&rwlock->cnt);
 
-	return odp_atomic_cas_acq_u32(&rwlock->cnt, &zero, (uint32_t)1);
+	while (cnt != (uint32_t)-1) {
+		if (odp_atomic_cas_acq_u32(&rwlock->cnt, &cnt, cnt + 1))
+			return 1;
+	}
+
+	return 0;
 }
 
 void odp_rwlock_read_unlock(odp_rwlock_t *rwlock)
--- a/test/common_plat/validation/api/lock/lock.c
+++ b/test/common_plat/validation/api/lock/lock.c
@@ -257,7 +257,7 @@ static int ticketlock_api_tests(void *arg UNUSED)
 
 static void rwlock_api_test(odp_rwlock_t *rw_lock)
 {
-	int rc;
+	int rc = 0;
 
 	odp_rwlock_init(rw_lock);
 	/* CU_ASSERT(odp_rwlock_is_locked(rw_lock) == 0); */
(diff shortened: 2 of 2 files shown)


lumag:
@Bill-Fischofer-Linaro updated commit message

muvarov:
<pre>From: "Peltonen, Janne (Nokia - FI/Espoo)" <janne.peltonen@nokia.com>
 
That does not work. Since the value of cnt is not checked, the code
would happily take the lock even when a writer already has it.

	Janne

> -----Original Message-----
> From: lng-odp [mailto:lng-odp-bounces@lists.linaro.org] On Behalf Of Github ODP bot
> Sent: Wednesday, April 26, 2017 8:00 PM
> To: lng-odp@lists.linaro.org
> Subject: [lng-odp] [PATCH] linux-generic: rwlock: fix odp_rwlock_read_trylock()
> 
> From: Dmitry Eremin-Solenikov <dmitry.ereminsolenikov@linaro.org>
> 
> odp_rwlock_read_trylock() currently works only if there are no readers
> (and writers) as it compares counter with 0. Make it actually work in
> case there are other active readers.
> 
> Fixes: https://bugs.linaro.org/show_bug.cgi?id=2974
> 
> Signed-off-by: Dmitry Eremin-Solenikov <dmitry.ereminsolenikov@linaro.org>
> ---
> /** Email created from pull request 19 (lumag:fix-rwlock)
>  ** https://github.com/Linaro/odp/pull/19
>  ** Patch: https://github.com/Linaro/odp/pull/19.patch
>  ** Base sha: 9b993a1531c94782b48292adff54a95de9d2be5c
>  ** Merge commit sha: 3af3b11b695d157ce1649955b52e8a165cc82937
>  **/
>  platform/linux-generic/odp_rwlock.c | 4 ++--
>  1 file changed, 2 insertions(+), 2 deletions(-)
> 
> diff --git a/platform/linux-generic/odp_rwlock.c b/platform/linux-generic/odp_rwlock.c
> index 13c17a2..e0c46ce 100644
> --- a/platform/linux-generic/odp_rwlock.c
> +++ b/platform/linux-generic/odp_rwlock.c
> @@ -33,9 +33,9 @@ void odp_rwlock_read_lock(odp_rwlock_t *rwlock)
> 
>  int odp_rwlock_read_trylock(odp_rwlock_t *rwlock)
>  {
> -	uint32_t zero = 0;
> +	uint32_t cnt = odp_atomic_load_u32(&rwlock->cnt);
> 
> -	return odp_atomic_cas_acq_u32(&rwlock->cnt, &zero, (uint32_t)1);
> +	return odp_atomic_cas_acq_u32(&rwlock->cnt, &cnt, cnt + 1);
>  }
> 
>  void odp_rwlock_read_unlock(odp_rwlock_t *rwlock)

 </pre>

muvarov:
<pre>From: Dmitry Eremin-Solenikov <dmitry.ereminsolenikov@linaro.org>
 On 26.04.2017 20:54, Peltonen, Janne (Nokia - FI/Espoo) wrote:
> 
> That does not work. Since the value of cnt is not checked, the code
> would happily take the lock even when a writer already has it.

Nice catch. Updated to check for writers.

-- 
With best wishes
Dmitry
 </pre>

platform/linux-generic/odp_rwlock.c
```diff
@@ -33,9 +33,14 @@ void odp_rwlock_read_lock(odp_rwlock_t *rwlock)
 
 int odp_rwlock_read_trylock(odp_rwlock_t *rwlock)
 {
-	uint32_t zero = 0;
+	uint32_t cnt = odp_atomic_load_u32(&rwlock->cnt);
+
+	while (cnt != (uint32_t)-1) {
```
Bill-Fischofer-Linaro: This is still not correct. Trylock should not repeatedly try the operation. Change while to if here.

lumag: Well. It is an interesting point. This `while` waits for other _read_ locks to finish locking in parallel. It does not wait for write lock to be lifted, if there is one. Is it OK from `_trylock()` semantics?

lumag: @Bill-Fischofer-Linaro 
E.g. check [MUSL sources](https://git.musl-libc.org/cgit/musl/tree/src/thread/pthread_rwlock_tryrdlock.c).

Bill-Fischofer-Linaro: Good point, and a reasonable one. Trylock says the caller doesn't want to wait for the lock to be released, but given that we're now allowing multiple readers theres the possibility that CAS operations may need to be retried as they all try to increment in parallel. So I withdraw this comment.

lumag:
@Bill-Fischofer-Linaro patch got mangled w/o code escapes. I've added a patch updating testsuite for rwlocks.

Bill-Fischofer-Linaro:
@lumag I saw that. Need to research how to escape things so quoted patches don't get misinterpreted as markdown.

lumag:
You can use the following form:

> ```diff
> patch goes here
> and here
> ```

muvarov:
Meged.

PR #23, opened 2017-05-05
[PATCH v1] Fix AES-GCM tag check

lumag:

PR #24, opened 2017-05-05
[PATCH API-NEXT v4] Update Linux crypto implementation to support recent API additions

lumag:
Support recently added auth_digest_length and aad parameters.

This depends on #23 (included as first two patches) and should be rebased after #23 gets merged into main and further into api-next.


muvarov:
@lumag  I re-run this pull request in Travis. But it again hang. Something wrong with that patches due to api-next works.

lumag:
@muvarov It seems I found the trouble. It is a bad interaction of #25, #24 and odp_crypt performance test. Related question: do we require that `odp_session_create()` (and other such functions) set returned `session` to `ODP_SESSION_INVALID` if it is going to return an error?

lumag:
@psavol could you please review this PR

PR #25, opened 2017-05-09
[PATCH v1] linux-generic: crypto: don't leak sessions if creation fails

lumag:
We should free allocated session in odp_crypto_session_create() error
paths, so that the session is not leaked.


muvarov:
Meged.

PR #32, opened 2017-05-17
[PATCH API-NEXT v1] api: crypto: deprecate per-session IV

lumag:
Per-session IV (in it's current form) are static IVs used (possibly) for
several packets, and so they degradate security of genrated packets.
Instead of modifying semantics of IV wrt processed packets, etc. let's
deprecate (and in future remove) this IV altogether and enforce
applications to always specify correct IV in the per-packet operation
params.

### DIFF
--- a/example/ipsec/odp_ipsec.c
+++ b/example/ipsec/odp_ipsec.c
@@ -694,7 +694,7 @@ pkt_disposition_e do_ipsec_in_classify(odp_packet_t pkt,
 	if (esp) {
 		params.cipher_range.offset = ipv4_data_p(ip) + hdr_len - buf;
 		params.cipher_range.length = ipv4_data_len(ip) - hdr_len;
-		params.override_iv_ptr = esp->iv;
+		params.iv_ptr = esp->iv;
 	}
 
 	/* Issue crypto request */
@@ -884,7 +884,15 @@ pkt_disposition_e do_ipsec_out_classify(odp_packet_t pkt,
 		trl_len = encrypt_len - ip_data_len;
 
 		esp->spi = odp_cpu_to_be_32(entry->esp.spi);
-		memcpy(esp + 1, entry->state.iv, entry->esp.iv_len);
+
+		/* Generate an IV */
+		if (entry->esp.iv_len) {
+			int32_t size = entry->esp.iv_len;
+			int32_t ret = odp_random_data(esp->iv, size, 1);
+
+			if (ret != size)
+				abort();
+		}
 
 		esp_t = (odph_esptrl_t *)(ip_data + encrypt_len) - 1;
 		esp_t->pad_len     = trl_len - sizeof(*esp_t);
@@ -894,6 +902,7 @@ pkt_disposition_e do_ipsec_out_classify(odp_packet_t pkt,
 			esp_t->next_header = ip->proto;
 		ip->proto = ODPH_IPPROTO_ESP;
 
+		params.iv_ptr = esp->iv;
 		params.cipher_range.offset = ip_data - buf;
 		params.cipher_range.length = encrypt_len;
 	}
--- a/include/odp/api/spec/crypto.h
+++ b/include/odp/api/spec/crypto.h
@@ -234,14 +234,14 @@ typedef struct odp_crypto_key {
 /**
  * Crypto API IV structure
  */
-typedef struct odp_crypto_iv {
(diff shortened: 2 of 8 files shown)


example/ipsec/odp_ipsec.c
```diff
@@ -884,7 +884,15 @@ pkt_disposition_e do_ipsec_out_classify(odp_packet_t pkt,
 		trl_len = encrypt_len - ip_data_len;
 
 		esp->spi = odp_cpu_to_be_32(entry->esp.spi);
-		memcpy(esp + 1, entry->state.iv, entry->esp.iv_len);
+
+		/* Generate an IV */
+		if (entry->esp.iv_len) {
+			int32_t size = entry->esp.iv_len;
+			int32_t ret = odp_random_data(esp->iv, size, 1);
```
Bill-Fischofer-Linaro: Should use enum here: ODP_RANDOM_CRYPTO

include/odp/api/spec/crypto.h
```diff
@@ -285,8 +285,11 @@ typedef struct odp_crypto_session_param_t {
 	 */
 	odp_crypto_key_t cipher_key;
 
-	/** Cipher Initialization Vector (IV) */
-	odp_crypto_iv_t iv;
+	/** @deprecated use iv_length and per-packet IV instead */
+	ODP_DEPRECATE(odp_crypto_iv_t) ODP_DEPRECATE(iv);
```
Bill-Fischofer-Linaro: Not clear why checkpatch gets confused here. It seems to think this is a cast of some sort. Another reason perhaps to simply correct this issue without deprecation.

include/odp/api/spec/crypto.h
```diff
@@ -234,14 +234,14 @@ typedef struct odp_crypto_key {
...
-typedef struct odp_crypto_iv {
+typedef struct ODP_DEPRECATE(odp_crypto_iv) {
 	/** IV data */
 	uint8_t *data;
 
 	/** IV length in bytes */
 	uint32_t length;
 
-} odp_crypto_iv_t;
+} ODP_DEPRECATE(odp_crypto_iv_t);
```
Bill-Fischofer-Linaro: Since this is really a bug correction, perhaps this should just be removed without deprecation? Why would anyone want to retain the incorrect behavior "for compatibility"?

platform/linux-generic/include/odp_crypto_internal.h
```diff
@@ -40,8 +42,10 @@ struct odp_crypto_generic_session {
 	odp_bool_t do_cipher_first;
 
 	struct {
+#if ODP_DEPRECATED_API
 		/* Copy of session IV data */
 		uint8_t iv_data[MAX_IV_LEN];
+#endif
```
Bill-Fischofer-Linaro: Needing an #ifdef looks ugly here and elsewhere in this series. Better to correct the code than retain "compatibility" here.

platform/linux-generic/odp_crypto.c
```diff
@@ -241,9 +249,14 @@ odp_crypto_alg_err_t aes_decrypt(odp_crypto_op_param_t *param,
...
 static int process_aes_param(odp_crypto_generic_session_t *session)
 {
-	/* Verify IV len is either 0 or 16 */
-	if (!((0 == session->p.iv.length) || (16 == session->p.iv.length)))
+	/* Verify IV len is 16 */
+#if ODP_DEPRECATED_API
+	if (!((16 == session->p.iv_length) || (16 == session->p.iv.length)))
+		return -1;
+#else
+	if (16 != session->p.iv_length)
```
Bill-Fischofer-Linaro: The comment above got deleted. Magic numbers should either be enums or at least retain a comment explaining their use/significance.

include/odp/api/spec/crypto.h
lumag: @Bill-Fischofer-Linaro I would like to hear a word from other parties. This bug might be local to linux-generic. And there might be use cases for the IV specified at session creation time (e.g. if packets should be encrypted in one continuous stream).

Bill-Fischofer-Linaro: During today's ARCH call this was discussed briefly and Nikhil pointed out that per-session IVs are useful for many cases, so deprecation may be premature, especially as we already permit per-packet IV overrides. We can continue this discussion on Monday's ARCH call if the mailing list can't reach resolution before then.

PR #33, opened 2017-05-17
[PATCH v2] travis: check compile and run from install

muvarov:
Verify that odp app can be compiled and run from installed
odp library and includes.

### DIFF
--- a/.travis.yml
+++ b/.travis.yml
@@ -124,9 +124,21 @@ script:
         - make distcheck
 
         - ./bootstrap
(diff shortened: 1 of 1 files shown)


muvarov:
@Bill-Fischofer-Linaro yes, that is strange. Maybe gcov breaks some files. I will change to compile it in different directory which more like has to help.

muvarov:
@Bill-Fischofer-Linaro this version is good right? Compile and run it from $HOME instead of odp dir.

muvarov:
Ok, approved was clicked. So merging.

PR #36, opened 2017-05-23
[PATCH v1] travis: expand testing with varios configure options

muvarov:
Improve testing with enabling different configure options.

PR #40, opened 2017-05-25
[PATCH v1] travis: split build into separate stages

lumag:
Introduce separate parallel builds for checkpatch, doxygen, distcheck
and plain checks.

### DIFF
--- a/.travis.yml
+++ b/.travis.yml
@@ -33,65 +33,50 @@ env:
     # you need generated new one at https://codecov.io specific for your repo.
     - CODECOV_TOKEN=8e1c0fd8-62ff-411e-a79f-5839f6662c11
 
-matrix:
-  include:
-    - compiler: gcc
-      addons:
+addons:
         apt:
-          sources:
-            - ubuntu-toolchain-r-test
-          packages:
-            - gcc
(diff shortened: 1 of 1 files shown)


lumag:
Got timeout in pktio_run_pcap.sh test when built for coverage analysis. Any idea?

lumag:
Moreover, the test works correctly if running with serial automake driver. It looks like some kind of interference of parallel ODP applications.

lumag:
Should be fixed by #41

muvarov:
@lumag looks nice. In the Travis test table we need some readable note about what is the current stage is. I found that we can only set environment variable. Even if we do not use it. Like STAGE=DISTCHECK

muvarov:
@lumag yes, looks like if we run test in parallel then we can not delete shm, ipc passed but other test might hang.

lumag:
@muvarov If we really need to delete IPC, this should be done from whithin the program, when we know the PID. Or we should stop hardcoding /tmp.

lumag:
I'll update adding 'environment' names.

muvarov:
@lumag maybe combine it with the same time with #36 ? I think pr will overlap.

PR #41, opened 2017-05-25
[PATCH v1] test: linux-generic: stop dropping other ODP files in pktio_ips test

lumag:
Prevent pktio_ipc test from removing sockets and SHM files used by other applications.


lumag:
Merged, closing now.

PR #42, opened 2017-05-26
[PATCH v2] Rework Travis configuration

lumag:
This patchset merges #40 and #36:
- enable tests with different configure options [check all compiler vs options combinations]
- use build stages for distcheck/checkpatch/etc
- dump test logs in case of failure

### DIFF
--- a/.travis.yml
+++ b/.travis.yml
@@ -33,65 +33,50 @@ env:
     # you need generated new one at https://codecov.io specific for your repo.
     - CODECOV_TOKEN=8e1c0fd8-62ff-411e-a79f-5839f6662c11
 
-matrix:
-  include:
-    - compiler: gcc
-      addons:
+addons:
         apt:
-          sources:
-            - ubuntu-toolchain-r-test
-          packages:
-            - gcc
-      env: MY_CF="-O0 -coverage" MY_LDF="--coverage" DOCOV=1
-    - compiler: clang
-      addons:
-        apt:
-          sources:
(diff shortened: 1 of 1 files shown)


muvarov:
Test results looks like ok. With test coverage only timer test failed:

Test: time_test_wait_ns ...Exceed upper limit: diff is 3023969520, upper_limit 2980000000
FAILED

Maybe to wrap this files under bash scripts also like tm? Not related to this serries but some how it's was reproduced now.

lumag:
@muvarov I'd prefer to enlarge boundaries, rather than marking another test as failable under Travis.

lumag:
For example the same commit [succeeded](https://travis-ci.org/lumag/odp/builds/236279060).

muvarov:
Recorded fail to bug: https://bugs.linaro.org/show_bug.cgi?id=3017

muvarov:
@lumag - make check -j $(nproc) looks like a reason why this test starting fail. If performance tests are enabled than running tests in parallel is can make them really slow and results might not pass even accepted boundaries. I think we need to change it back to single run.

muvarov:
ok it was -n 4, looks like -n $proc is good change for that patch. Then we need to check that performance tests do not run in parallel with other tests.

muvarov:
@lumag  please skip 2 above comments. Monday morning :) Run is not parallel only compilation.

muvarov:
Meged.

PR #44, opened 2017-05-31
[PATCH v1] test: time: do not fail under gcov

muvarov:
code coverage gcov make test very slow and it does not pass
accepted boundaries. Test if env TEST=coverage variable is set
and do no generate fail result if some boundaries are missing.
https://bugs.linaro.org/show_bug.cgi?id=3017


muvarov:
Merged.

PR #46, opened 2017-06-07
Add odp_feature_t type and upgrades to odp_init_global()

Bill-Fischofer-Linaro:
Add the odp_feature_t type to control selectable ODP features and use this to enable
applications to indicate that specific ODP features will not be used at odp_init_global() time.

As part of this series, the odp_init_param_init() API to initialize odp_init_t to its default values.


.checkpatch.conf
```diff
@@ -8,5 +8,7 @@
 --ignore=BIT_MACRO
 --ignore=PREFER_PRINTF
 --ignore=PREFER_SCANF
+--ignore=VOLATILE
```
brbrooks: Is a previous commit related to checkpatch is showing up in this pull request?

include/odp/api/spec/feature.h
```diff
@@ -0,0 +1,60 @@
...
+	/** All features */
+	uint32_t all_feat;
+
+	/** Individual feature bits */
+	struct {
+		uint32_t classification:1;
+		uint32_t crypto:1;
+		uint32_t ipsec:1;
+		uint32_t sched:1;
+		uint32_t time:1;
```
brbrooks: I recommend time be omitted as well since it is not a data plane object or component.  The rest all look like they belong here - nice.

platform/linux-generic/odp_init.c
```diff
@@ -147,6 +147,11 @@ static int read_configfile(void)
 	return 0;
 }
 
+void odp_init_param_init(odp_init_t *param)
```
brbrooks: I suspect this was added for consistency.. okay.

include/odp/api/spec/feature.h
Bill-Fischofer-Linaro: The rationale for including time is that we've had a lot of discussion regarding how time should be implemented / represented so if an application knows it isn't going to be using those APIs that may useful for implementations to know. The idea here is for the application to be able to hint the implementation to enable more streamlined implementation logic. The implementation, of course, is free to ignore this info.

platform/linux-generic/odp_init.c
Bill-Fischofer-Linaro: More than just consistency. Most applications today are initializing ODP with the call:

odp_init_global(&instance, NULL, NULL);

If they now want to specify a not_used feature, then they need to provide an odp_init_t as the 2nd argument. This API is used to initialize that to its default values in an implementation-independent manner.

.checkpatch.conf
Bill-Fischofer-Linaro: I'm not sure why these commits got attached to this pull request. I'll need to sort that out with Maxim when he gets back. The pull should only be for the feature.h and related init changes.

include/odp/api/spec/feature.h
```diff
@@ -0,0 +1,60 @@
...
+typedef union odp_feature_t {
+	/** All features */
+	uint32_t all_feat;
+
+	/** Individual feature bits */
+	struct {
+		uint32_t classification:1;
+		uint32_t crypto:1;
+		uint32_t ipsec:1;
+		uint32_t sched:1;
```
nagarahalli: The name **sched** is not inline with the way the spec header file is named. The spec header file is named **schedule.h**. This should be changed to **schedule**.

test/common_plat/validation/api/init/init.c
```diff
@@ -24,10 +24,10 @@ static int odp_init_log(odp_log_level_t level, const char *fmt, ...);
 void init_test_odp_init_global_replace_abort(void)
 {
 	int status;
-	struct odp_init_t init_data;
+	odp_init_t init_data;
 	odp_instance_t instance;
 
-	memset(&init_data, 0, sizeof(init_data));
+	odp_init_param_init(&init_data);
```
Bill-Fischofer-Linaro: Since this is a void function there's no return code to check. We just want to be sure it's included in the code coverage report, hence these changes. It's basically doing the same memset() that's being replaced here, but in an implementation-independent manner. 

The other odp_xxx_param_init() routines follow the same practice in the validation tests, so there's nothing new here.

include/odp/api/spec/feature.h
Bill-Fischofer-Linaro: I'm OK with that. These names were Petri's suggestions so I went with them as a first cut.

.checkpatch.conf
nagarahalli: I was in a similar situation. Maxim and I tried to figure out what happened. I don't think we understood. Ultimately, we did interactive rebase (git rebase -i HEAD~x) and dropped the unrelated commits and force pushed the changes.

Bill-Fischofer-Linaro: I tried that and I don't think it did anything useful. The problem I've always had with git is that eventually clones get too tangled to do anything useful with them as you try to keep them in sync with the master repo with fetches and merges. I usually just use a fresh clone to start over. I may need to do that here. Will see what Maxim suggests when he gets back.

Bill-Fischofer-Linaro: To avoid these problems I've moved this patch series to the mailing list, so please refer to patch series http://patches.opendataplane.org/project/lng-odp/list/?series=240 for further reviews/comments on this request.

Bill-Fischofer-Linaro:
Withdrawing this pull request as it has been replaced by mailing list patch series http://patches.opendataplane.org/project/lng-odp/list/?series=243

PR #47, opened 2017-06-13
[PATCH CLOUD-DEV v1] drv: driver: fix devio registration race condition

joseppc:
Before registering a new devio, we loop through all currently registered
devios to make sure none with the same name has already been
registered. The loop is protected with a read lock, later this lock is
released and a new write lock is taken to insert the new element in the
list. Between the release of the read lock and the acquisition of the
write lock, a new devio with the same name could be successfully
registered. To prevent this we only take the write lock once to enclose
both operations.

### DIFF
--- a/platform/linux-generic/drv_driver.c
+++ b/platform/linux-generic/drv_driver.c
@@ -529,7 +529,7 @@ odpdrv_devio_t odpdrv_devio_register(odpdrv_devio_param_t *param)
 	 * sure no devio providing the same interface using th esame enumerator
 	 * already exists:
 	 */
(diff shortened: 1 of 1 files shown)


heyi-linaro:
For the commit title and message:

title: suggest to follow pattern "drv: driver: <short description>"
message: very clear explanation but a bit longer, better make it more short? Not describe the detailed problem in code only need to mention that devio_register() has race condition could be sufficient.

heyi-linaro:
Reviewed-by: Yi He <yi.he@linaro.org>

PR #49, opened 2017-06-13
[PATCH API-NEXT v4] travis: add sudo for missing tests

muvarov:
Add sudo to make tests run which require root. Also do not run
tests in parallel for coverage case. It's already very slow and
performance tests eat all cpu time. So the is situation where tests
can not fit in any reasonable limit frames.


muvarov:
Merged. Billl said that review stands for that version.

PR #52, opened 2017-06-15
[PATCH v1] configure.ac: do not trap if libatomic is not found

muvarov:
Idea if check was to detect if libatomic is needed or not,
not trap configure on not case
Fixes Linaro CI compilation for:
gcc-linaro-5.3-2016.02-x86_64_arm-linux-gnueabih


muvarov:
Merged.

PR #54, opened 2017-06-19
[PATCH API-NEXT v1] RFC: linux-generic: queue: modular queue interface

nagarahalli:
RFC for modular queue interface using odp_queue_t towards internal ODP
components. There is no internal abstract queue type.

Signed-off-by: Honnappa Nagarahalli <honnappa.nagarahalli@arm.com>

### DIFF
--- a/log
+++ b/log
@@ -1,0 +1,1269 @@
+diff --git a/platform/linux-generic/include/odp_buffer_internal.h b/platform/linux-generic/include/odp_buffer_internal.h
+index 076abe9..c51708b 100644
+--- a/platform/linux-generic/include/odp_buffer_internal.h
++++ b/platform/linux-generic/include/odp_buffer_internal.h
+@@ -30,7 +30,6 @@ extern "C" {
+ #include <odp/api/thread.h>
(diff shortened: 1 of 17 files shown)


Bill-Fischofer-Linaro:
Patch has whitespace issues that should be fixed. Scalable scheduler patch applies but is broken with this RFC, so that would need a v9.

nagarahalli:
@Bill-Fischofer-Linaro It looks to be a GitHub issue. For ex: it shows the whitespace issues in odp_queue_if.h file, but in my local copy, those issues are not present. Checkpatch also does not report any errors (I have checked-in a temp file which the checkpatch is complaining about, I will remove that in the next version).

Bill-Fischofer-Linaro:
If I download the patch checkpatch shows the same whitespace errors for me. You might want to double check the files before pushing them.

muvarov:
 On 06/19/17 08:00, Github ODP bot wrote:
>  log                                                | 1269 ++++++++++++++++++++

log has not be a part of patch.

Maxim.
 </pre>

PR #57, opened 2017-06-23
[PATCH v3] configure: remove dependency of autoconf-archive

muvarov:
autoconf-archive is not available if you build with cross
toolchains. Number of projects simple have local copy
of this file.

---
Linaro CI found this issue while compiling with cross tools.

### DIFF
--- a/.travis.yml
+++ b/.travis.yml
@@ -42,7 +42,7 @@ addons:
                 packages:
                         - gcc
                         - clang-3.8
-                        - automake autoconf autoconf-archive libtool libssl-dev graphviz mscgen doxygen
+                        - automake autoconf libtool libssl-dev graphviz mscgen doxygen
                         - libpcap-dev
 #        coverity_scan:
 #                project:
--- a/DEPENDENCIES
+++ b/DEPENDENCIES
@@ -8,14 +8,13 @@ Prerequisites for building the OpenDataPlane (ODP) API
 
    automake
    autoconf
-   autoconf-archive
    libtool
 
    On Debian/Ubuntu systems:
-   $ sudo apt-get install automake autoconf autoconf-archive libtool
+   $ sudo apt-get install automake autoconf libtool
 
    On CentOS/RedHat/Fedora systems:
-   $ sudo yum install automake autoconf autoconf-archive libtool
+   $ sudo yum install automake autoconf libtool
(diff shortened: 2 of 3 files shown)


configure.ac
```diff
@@ -26,6 +26,7 @@ AM_CONFIG_HEADER([helper/config.h])
 AC_USE_SYSTEM_EXTENSIONS
 AC_SYS_LARGEFILE
 AC_CONFIG_MACRO_DIR([m4])
+m4_include([m4/autoconf-archive/ax_check_compile_flag.m4])
```
lumag: No need to `m4_include` it. Don't do `aclocal`'s job

m4/autoconf-archive/ax_check_compile_flag.m4
```diff
@@ -0,0 +1,72 @@
+# ===========================================================================
+#   http://www.gnu.org/software/autoconf-archive/ax_check_compile_flag.html
+# ===========================================================================
```
lumag: Just put the file to `m4/`, like the rest of ax files.

muvarov: @lumag yes, maybe message is not really clear. autoconf is not depends on cross tools and depends on host env only. Idea was that for some env this package exit. For some other it is not. I see that Linaro CI fails. Or I need to update all docker images if it's possible or just put this file here.

If we already have autoconf-archive in m4/ folder than it's really good to be there. I my patch I wanted to put it to separate folder to be more clear that it's related to autoconf-archive package.

lumag: Then I'd say that ac-archive is not available in Linaro CI setup, so let's bring the file in. Current message is just misleading.

Makefile.am
```diff
@@ -1,4 +1,4 @@
-ACLOCAL_AMFLAGS=-I m4
+ACLOCAL_AMFLAGS=-I m4 -I m4/autoconf-archive
```
lumag: This is not necessary anymore.

lumag:
Reviewed-by: Dmitry Eremin-Solenikov <dmitry.ereminsolenikov@linaro.org>

muvarov:
Merged.

PR #69, opened 2017-07-06
[PATCH API-NEXT v1] api: ipsec: pass OUT_INLINE outer headers as const

lumag:
Pass outer headers for outbound inline packets as const pointers.


lumag:
Replaced by Petri's patch.

PR #72, opened 2017-07-10
[PATCH v1]  ipc pkt free fix

muvarov:
fix packets free for pktio ipc. This patchset makes also Petri's patches workable.


muvarov:
Merged.

PR #76, opened 2017-07-14
[PATCH v1] example: ipsec: fix build with OpenSSL located in non-standard path

lumag:
IPsec example uses OpenSSL directly to generate packages stream. Thus
use $(OPENSSL_CPPFLAGS) and $(OPENSSL_LIBS) to let compiler find OpenSSL
if it is present in non-standard location.


muvarov:
Merged.

PR #79, opened 2017-07-14
[PATCH v2] test: ipc pktio: bind tasks to cores

muvarov:
tasks which use odp time api have to be bind to cores.
https://bugs.linaro.org/show_bug.cgi?id=3126

### DIFF
--- a/test/linux-generic/pktio_ipc/pktio_ipc1.c
+++ b/test/linux-generic/pktio_ipc/pktio_ipc1.c
@@ -42,6 +42,7 @@ static int pktio_run_loop(odp_pool_t pool)
 	int ret;
 	odp_pktin_queue_t pktin;
 	char name[30];
+	int sync_cnt = 0;
 
 	if (master_pid)
 		sprintf(name, TEST_IPC_PKTIO_PID_NAME, master_pid);
@@ -178,7 +179,7 @@ static int pktio_run_loop(odp_pool_t pool)
 
 				cnt_recv++;
 
-				if (head.seq != cnt_recv) {
+				if (head.seq != cnt_recv && sync_cnt) {
 					stat_errors++;
 					odp_packet_free(pkt);
 					EXAMPLE_DBG("head.seq %d - "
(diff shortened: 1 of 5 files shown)


test/linux-generic/pktio_ipc/pktio_ipc1.c
```diff
@@ -309,8 +313,24 @@ int main(int argc, char *argv[])
...
+	odp_cpumask_default_worker(&mask, 0);
+	cpu = odp_cpumask_first(&mask);
+
+
+	CPU_ZERO(&cpu_set);
+	CPU_SET(cpu, &cpu_set);
+
+	pid = getpid();
+
+	if (sched_setaffinity(pid, sizeof(cpu_set_t), &cpu_set)) {
```
Bill-Fischofer-Linaro: Since these are example/test applications, shouldn't they use the odph_odpthread_getaffinity() and opdh_odpthread_setaffinity() helper functions rather than the linux calls directly?

test/linux-generic/pktio_ipc/pktio_ipc2.c
```diff
@@ -217,8 +221,24 @@ int main(int argc, char *argv[])
...
+	odp_cpumask_default_worker(&mask, 0);
+	cpu = odp_cpumask_first(&mask);
+	cpu = odp_cpumask_next(&mask, cpu);
+
+	CPU_ZERO(&cpu_set);
+	CPU_SET(cpu, &cpu_set);
+
+	pid = getpid();
+
+	if (sched_setaffinity(pid, sizeof(cpu_set_t), &cpu_set)) {
```
Bill-Fischofer-Linaro: Same comment about using the helper function here.

test/linux-generic/pktio_ipc/pktio_ipc1.c
muvarov: odph_odpthread_getaffinity and  opdh_odpthread_setaffinity can not be used here because it now multi threaded example, i.e. there is no any thread.  And that is test/linux-generic/ example. I will send v2 with updated current code as it more simple.

muvarov:
Merged.

PR #82, opened 2017-07-20
[PATCH v4] Pktrefs

Bill-Fischofer-Linaro:
Implement zero-copy packet references. This replaces the temporary scaffolding for these APIs with a performant implementation that enables references to share packet data.

### DIFF
--- a/m4/ax_valgrind_check.m4
+++ b/m4/ax_valgrind_check.m4
@@ -4,19 +4,24 @@
 #
 # SYNOPSIS
 #
+#   AX_VALGRIND_DFLT(memcheck|helgrind|drd|sgcheck, on|off)
 #   AX_VALGRIND_CHECK()
 #
 # DESCRIPTION
 #
-#   Checks whether Valgrind is present and, if so, allows running `make
-#   check` under a variety of Valgrind tools to check for memory and
-#   threading errors.
+#   AX_VALGRIND_CHECK checks whether Valgrind is present and, if so, allows
+#   running `make check` under a variety of Valgrind tools to check for
+#   memory and threading errors.
 #
 #   Defines VALGRIND_CHECK_RULES which should be substituted in your
 #   Makefile; and $enable_valgrind which can be used in subsequent configure
 #   output. VALGRIND_ENABLED is defined and substituted, and corresponds to
 #   the value of the --enable-valgrind option, which defaults to being
-#   enabled if Valgrind is installed and disabled otherwise.
+#   enabled if Valgrind is installed and disabled otherwise. Individual
+#   Valgrind tools can be disabled via --disable-valgrind-<tool>, the
+#   default is configurable via the AX_VALGRIND_DFLT command or is to use
+#   all commands not disabled via AX_VALGRIND_DFLT. All AX_VALGRIND_DFLT
+#   calls must be made before the call to AX_VALGRIND_CHECK.
 #
 #   If unit tests are written using a shell script and automake's
 #   LOG_COMPILER system, the $(VALGRIND) variable can be used within the
@@ -28,6 +33,7 @@
 #
 #   configure.ac:
 #
+#     AX_VALGRIND_DFLT([sgcheck], [off])
 #     AX_VALGRIND_CHECK
 #
 #   Makefile.am:
@@ -40,22 +46,38 @@
 #   which includes "@VALGRIND_CHECK_RULES@" (assuming the module has been
 #   configured with --enable-valgrind). Running `make check-valgrind` in
 #   that directory will run the module's test suite (`make check`) once for
-#   each of the available Valgrind tools (out of memcheck, helgrind, drd and
-#   sgcheck), and will output results to test-suite-$toolname.log for each.
-#   The target will succeed if there are zero errors and fail otherwise.
+#   each of the available Valgrind tools (out of memcheck, helgrind and drd)
+#   while the sgcheck will be skipped unless enabled again on the
+#   commandline with --enable-valgrind-sgcheck. The results for each check
+#   will be output to test-suite-$toolname.log. The target will succeed if
+#   there are zero errors and fail otherwise.
+#
+#   Alternatively, a "check-valgrind-$TOOL" rule will be added, for $TOOL in
+#   memcheck, helgrind, drd and sgcheck. These are useful because often only
+#   some of those tools can be ran cleanly on a codebase.
 #
 #   The macro supports running with and without libtool.
 #
 # LICENSE
 #
-#   Copyright (c) 2014, 2015 Philip Withnall <philip.withnall@collabora.co.uk>
+#   Copyright (c) 2014, 2015, 2016 Philip Withnall <philip.withnall@collabora.co.uk>
 #
 #   Copying and distribution of this file, with or without modification, are
 #   permitted in any medium without royalty provided the copyright notice
 #   and this notice are preserved.  This file is offered as-is, without any
 #   warranty.
 
-#serial 3
+#serial 13
+
+dnl Configured tools
+m4_define([valgrind_tool_list], [[memcheck], [helgrind], [drd], [sgcheck]])
+m4_set_add_all([valgrind_exp_tool_set], [sgcheck])
+m4_foreach([vgtool], [valgrind_tool_list],
+           [m4_define([en_dflt_valgrind_]vgtool, [on])])
+
+AC_DEFUN([AX_VALGRIND_DFLT],[
+	m4_define([en_dflt_valgrind_$1], [$2])
+])dnl
 
 AC_DEFUN([AX_VALGRIND_CHECK],[
 	dnl Check for --enable-valgrind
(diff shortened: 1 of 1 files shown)


muvarov:
 On 07/20/17 21:00, Github ODP bot wrote:
> From: Bill Fischofer <bill.fischofer@linaro.org>
> 
> When attempting to send a reference via IPC, make a copy of the packet as
> reference sharing is only supported within a single ODP instance.
> 
> Signed-off-by: Bill Fischofer <bill.fischofer@linaro.org>
> ---
> /** Email created from pull request 82 (Bill-Fischofer-Linaro:pktrefs)
>  ** https://github.com/Linaro/odp/pull/82
>  ** Patch: https://github.com/Linaro/odp/pull/82.patch
>  ** Base sha: 95ba4b394009d92c29c2e22f0776e90bb4c6edec
>  ** Merge commit sha: 01d093d362af5a52e079c622a99ae3495f68b64c
>  **/
>  platform/linux-generic/pktio/ipc.c | 7 +++++--
>  1 file changed, 5 insertions(+), 2 deletions(-)
> 
> diff --git a/platform/linux-generic/pktio/ipc.c b/platform/linux-generic/pktio/ipc.c
> index bc7d7564..8c7db84e 100644
> --- a/platform/linux-generic/pktio/ipc.c
> +++ b/platform/linux-generic/pktio/ipc.c
> @@ -592,7 +592,9 @@ static int ipc_pktio_send_lockless(pktio_entry_t *pktio_entry,
>  
>  	_ipc_free_ring_packets(pktio_entry, pktio_entry->s.ipc.tx.free);
>  
> -	/* Copy packets to shm shared pool if they are in different */
> +	/* Copy packets to shm shared pool if they are in different
> +	 * pool, or if they are references (we can't share across IPC).
> +	 */

In my understanding references have to work in following way. Lets say
two processes A and B.

A: send pkt: queue it to ring.
B: recv: dequeue pkt from ring, inc pkt ref, create pkt handle.
B: free or send: dec pkt ref, queue to 'free ring' to be freeing by A.
A: dequeue 'free ring': finally destroys packet if no more references.

Logic might be complex if B will also modify packet. But because they
share in memory the same packet pool, phdrs also are common. So there is
good chance that references will work in that case.

Maxim.

>  	for (i = 0; i < len; i++) {
>  		odp_packet_t pkt =  pkt_table[i];
>  		pool_t *ipc_pool = pool_entry_from_hdl(pktio_entry->s.ipc.pool);
> @@ -602,7 +604,8 @@ static int ipc_pktio_send_lockless(pktio_entry_t *pktio_entry,
>  		pkt_hdr = odp_packet_hdr(pkt);
>  		pool = pkt_hdr->buf_hdr.pool_ptr;
>  
> -		if (pool->pool_idx != ipc_pool->pool_idx) {
> +		if (pool->pool_idx != ipc_pool->pool_idx ||
> +		    odp_packet_has_ref(pkt)) {
>  			odp_packet_t newpkt;
>  
>  			newpkt = odp_packet_copy(pkt, pktio_entry->s.ipc.pool);
> 

 </pre>

muvarov:
<pre>From: Bill Fischofer <bill.fischofer@linaro.org>
 On Thu, Jul 20, 2017 at 2:32 PM, Maxim Uvarov <maxim.uvarov@linaro.org>
wrote:

> On 07/20/17 21:00, Github ODP bot wrote:
> > From: Bill Fischofer <bill.fischofer@linaro.org>
> >
> > When attempting to send a reference via IPC, make a copy of the packet as
> > reference sharing is only supported within a single ODP instance.
> >
> > Signed-off-by: Bill Fischofer <bill.fischofer@linaro.org>
> > ---
> > /** Email created from pull request 82 (Bill-Fischofer-Linaro:pktrefs)
> >  ** https://github.com/Linaro/odp/pull/82
> >  ** Patch: https://github.com/Linaro/odp/pull/82.patch
> >  ** Base sha: 95ba4b394009d92c29c2e22f0776e90bb4c6edec
> >  ** Merge commit sha: 01d093d362af5a52e079c622a99ae3495f68b64c
> >  **/
> >  platform/linux-generic/pktio/ipc.c | 7 +++++--
> >  1 file changed, 5 insertions(+), 2 deletions(-)
> >
> > diff --git a/platform/linux-generic/pktio/ipc.c
> b/platform/linux-generic/pktio/ipc.c
> > index bc7d7564..8c7db84e 100644
> > --- a/platform/linux-generic/pktio/ipc.c
> > +++ b/platform/linux-generic/pktio/ipc.c
> > @@ -592,7 +592,9 @@ static int ipc_pktio_send_lockless(pktio_entry_t
> *pktio_entry,
> >
> >       _ipc_free_ring_packets(pktio_entry, pktio_entry->s.ipc.tx.free);
> >
> > -     /* Copy packets to shm shared pool if they are in different */
> > +     /* Copy packets to shm shared pool if they are in different
> > +      * pool, or if they are references (we can't share across IPC).
> > +      */
>
> In my understanding references have to work in following way. Lets say
> two processes A and B.
>
> A: send pkt: queue it to ring.
> B: recv: dequeue pkt from ring, inc pkt ref, create pkt handle.
> B: free or send: dec pkt ref, queue to 'free ring' to be freeing by A.
> A: dequeue 'free ring': finally destroys packet if no more references.
>
> Logic might be complex if B will also modify packet. But because they
> share in memory the same packet pool, phdrs also are common. So there is
> good chance that references will work in that case.
>

The problem is that when a packet is referenced, it consists of a shared
part, which must be treated as read/only, and an unshared part, which is
private to that particular reference and hence may be freely extended or
modified. These semantics cannot be maintained if it is sent across an IPC
channel, so to be safe we make a copy of the whole packet. You're already
doing this if the packet resides in another pool, so this seemed the most
natural place to put this safeguard.

The other issue is that the reference count associated with packet segments
is shared metadata and hence must be modifiable by all holders of a
reference (it uses a odp_atomic_u32_t variable). Again, it's not clear if
we can maintain these semantics across IPC since the various sections of
referenced packets are chained together with pointers, which are not
sharable across address space boundaries.

Since there is no clear use case for passing references across IPC
interfaces, the simplest solution is to just do an implied copy whenever a
reference is sent. That ensures that there is no confusion between the two
parties.


>
> Maxim.
>
> >       for (i = 0; i < len; i++) {
> >               odp_packet_t pkt =  pkt_table[i];
> >               pool_t *ipc_pool = pool_entry_from_hdl(pktio_
> entry->s.ipc.pool);
> > @@ -602,7 +604,8 @@ static int ipc_pktio_send_lockless(pktio_entry_t
> *pktio_entry,
> >               pkt_hdr = odp_packet_hdr(pkt);
> >               pool = pkt_hdr->buf_hdr.pool_ptr;
> >
> > -             if (pool->pool_idx != ipc_pool->pool_idx) {
> > +             if (pool->pool_idx != ipc_pool->pool_idx ||
> > +                 odp_packet_has_ref(pkt)) {
> >                       odp_packet_t newpkt;
> >
> >                       newpkt = odp_packet_copy(pkt,
> pktio_entry->s.ipc.pool);
> >
>
>
 </pre>

heyi-linaro:
Hi, Bill

Going through the code and I found myself do not have enough knowledge to understand the implementation :), I learnt that the packet reference APIs not copy the packet anymore.

I have some basic questions about the relationships between odp_packet_t, segments, reference headers (added in PR), and maybe you can elaborate here or point to document to help understand the impl:

1) application uses odp_packet_t to refer to a packet
2) internally a packet can have multiple data segments
3) internally odp_packet_t is actually a packet header data structure,
    is this packet header independent to the data segments?
    create reference is like create new packet headers and share the data segments (or partially in offset)
4) It seems that odp_packet_alloc() allocates header together with segment(s), so create a dynamic reference also allocate extra data segments but leave that unused?
5) can you help elaborate on the calculation and means of fields packet->data_len, ref_offset, ref_len and ref_hdr?  especially in chained reference headers cases, for example, in calculate the packet length.

thanks and best regards, Yi

Bill-Fischofer-Linaro:
Thanks Yi,

With Petri's latest implementation of packets, an odp_packet_hdr_t behaves much like an rte_mbuf in that the same structure represents a packet and a packet segment. So pkt_hdr->buf_hdr.seg[0].hdr == pkt_hdr and seg[1]...seg[segcount-1].hdr are pointers to the odp_packet_hdr_t's that represent the other segments of the packet.

This means that each odp_packet_hdr_t represents a packet segment with the 0th such segment being used to represent the packet as a whole. The frame_len field of the 0th segment is the total packet length over all segments contained in it. The packet length contained in each individual segment, in turn, is held in pkt_hdr->buf_hdr.seg[n].len. As an aside, pkt_hdr->buf_hdr.seg[n].data is a pointer to the start of the packet data within each segment. 

Note that within each individual packet segment, pkt_hdr->buf_hdr.seg[0] == pkt_hdr, so while from the perspective of each individual segment it is capable of being the start of its own packet. If a packet consists of three segments (A, B, C) we'll have the following:
```
A->buf_hdr.seg[0].hdr == A
A->buf_hdr.seg[1].hdr == B
A->buf_hdr.seg[2].hdr == C

B->buf_hdr.seg[0] == B

C->buf_hdr.seg[0] == C
```
So only A "sees" the fact that B and C are connected to it. This is because the seg array is part of the packet metadata, and is not maintained separately for each segment within the packet. The initialization of seg[0] is done during pool creation so there is no overhead to maintain these fields within the individual segments. It's only when segments are put together and manipulated by the various odp_packet_xxx() APIs that they can changed, and then again only in the seg[0] odp_packet_hdr_t for the packet.

With this PR, the new ref_count field in each odp_packet_hdr_t is a reference count for each packet segment. This is set to 1 when the segment is initially allocated, is incremented each time it is referenced and decremented each time it is freed. Segments are only returned to the pool when their ref_count decrements to 0.

The other key new field is ref_hdr. In response to the call:
```
R = odp_packet_ref(P, offset);
```
A new pkt_hdr R is allocated and R->ref_hdr is set to point to P and P->ref_count is incremented to 2. P->ref_hdr == NULL because by default packets are not references. So if the application frees P, it's ref_count will just decrement back to 1 but not be freed because R is still referencing it.

Note that R->frame_len covers the sum of all packet data contained within segments anchored from R. Normally when odp_packet_offset() is called, the supplied offset must be in the range of 0..frame_len-1, but with references that only covers the date in this "section" of the packet. If R->ref_hdr != NULL, this means that offsets beyond R->frame_len are continued in P.  

There are two complications that arise. First, the 0th byte of P from R's perspective starts at the offset that was used when R was originally created, not the 0th byte from P's perspective. Since P may have change it's 0th byte since that time (e.g., via an odp_packet_push_head() call), we need to know where the original 0th byte is from R's perspective. Getting that info efficiently is the reason for the ref_offset and ref_len fields held in R.

When the call
```
R = odp_packet_ref(P, offset);
```
is made, we set the following:
```
R->ref_hdr = P;
R->ref_offset = offset;
R->ref_len = P->frame_len;
```
This "captures" the state of P at the time the reference was created so that P can continue to add data via odp_packet_push_head() or odp_packet_extend_head() and still have R see the original shared section of P. 

With this background, the key function packet_map() that accesses the nth byte of a packet becomes clearer:
```
static inline void *packet_map(odp_packet_hdr_t *pkt_hdr,
			       uint32_t offset, uint32_t *seg_len, int *seg_idx)
{
	void *addr;
	uint32_t len;
	int seg = 0;
	int seg_count = pkt_hdr->buf_hdr.segcount;

	/* Special processing for references */
	while (offset >= pkt_hdr->frame_len && pkt_hdr->ref_hdr) {
		offset   -= (pkt_hdr->frame_len - pkt_hdr->ref_offset);
		offset   += (pkt_hdr->ref_hdr->frame_len - pkt_hdr->ref_len);
		pkt_hdr   = pkt_hdr->ref_hdr;
		seg_count = pkt_hdr->buf_hdr.segcount;
	}

	if (odp_unlikely(offset > pkt_hdr->frame_len))
		return NULL;
```
The offset within pkt_hdr->ref_hdr we need is normalized back to the original offset used when the reference was created and we continue from there. 

Hope that helps.

heyi-linaro:
Thanks very much Bill for the detailed explanation, I'll continue review.

platform/linux-generic/odp_packet.c
```diff
@@ -1444,6 +1710,109 @@ int odp_packet_split(odp_packet_t *pkt, uint32_t len, odp_packet_t *tail)
...
+	ODP_ASSERT(packet_ref_count(pkt_hdr) > 0);
+
+	/* Splicing is from the last section of src pkt */
+	while (ref_hdr->ref_hdr)
+		ref_hdr = ref_hdr->ref_hdr;
+
+	/* Find section where splice begins */
+	while (offset >= pkt_hdr->frame_len && pkt_hdr->ref_hdr) {
+		offset   -= (pkt_hdr->frame_len - pkt_hdr->ref_offset);
+		offset   += (pkt_hdr->ref_hdr->frame_len - pkt_hdr->ref_len);
```
heyi-linaro: Here can suggest to use macros to write code in self explanation, like below (which seems not equal, but just express the opinion of writing code in self explanation):

reference_length = packet->frame_len;
while (offset > reference_length && has_next_reference(packet)) {
    offset -= reference_length;
    reference_length = next_reference_length(packet);
    packet = next_reference(packet);
}

#define next_reference(packet) packet->ref_hdr
#define has_next_reference(packet) (packet->ref_hdr != NULL)

#define next_reference_length(packet) \
    captured_reference_length(packet) + reference_length_compensation(packet)

/* the reference length captured when the reference was created */
#define captured_reference_length(packet) (packet->ref_len - packet->ref_offset)

/* the reference length compensation due to frame manipulation after reference creation */
#define reference_length_compensation(packet) (packet->ref_hdr->frame_len - packet->ref_len)

platform/linux-generic/include/odp/api/plat/packet_flag_inlines.h
```diff
@@ -16,14 +16,17 @@
 #include <odp/api/plat/packet_types.h>
 #include <odp/api/hints.h>
 
+/** @internal Packet field accessor */
+#define _odp_pkt_get(pkt, cast, field) \
```
Bill-Fischofer-Linaro: Good suggestion. I'll look to do that.

platform/linux-generic/odp_packet.c
Bill-Fischofer-Linaro: I'm reluctant to add macros that obscure the design, but I assume this is for future modularization flexibility? I'd rather do that as part of adding such modularity if it's really needed. Too many macros means knowing C is not sufficient to understand the code--you have to also learn a separate macro language. We're only talking 4 new fields here. 

There are inline functions for ref_count manipulation because the mechanics of doing so are not relevant to understanding their use in context. Let me look at these and see if inlining might be able to simplify some of these as well.

PR #84, opened 2017-07-21
[PATCH v1] travis: cache dpdk directory

matiaselo:
Cache DPDK directory so each individual test doesn't have to download and
build the code. Speeds up each tests which require DPDK by ~3 minutes.


muvarov:
merged

PR #86, opened 2017-07-25
[PATCH v1] example:use dpdk pktio with examples

GBalakrishna:
flag was not added to example/ during the commit:
 89b7caa

Signed-off-by: Balakrishna Garapati <balakrishna.garapati@linaro.org>

### DIFF
--- a/example/Makefile.inc
+++ b/example/Makefile.inc
@@ -1,6 +1,6 @@
 include $(top_srcdir)/platform/@with_platform@/Makefile.inc
 LIB   = $(top_builddir)/lib
-LDADD = $(LIB)/libodp-linux.la $(LIB)/libodphelper.la
+LDADD = $(LIB)/libodp-linux.la $(LIB)/libodphelper.la $(DPDK_PMDS)
 AM_CFLAGS += \
 	-I$(srcdir) \
 	-I$(top_srcdir)/example \


example/Makefile.inc
```diff
@@ -1,6 +1,6 @@
 include $(top_srcdir)/platform/@with_platform@/Makefile.inc
 LIB   = $(top_builddir)/lib
-LDADD = $(LIB)/libodp-linux.la $(LIB)/libodphelper.la
+LDADD = $(LIB)/libodp-linux.la $(LIB)/libodphelper.la $(DPDK_PMDS)
```
heyi-linaro: checked with make V=1 and saw it worked, although in automake manual it outlines that:

> https://www.gnu.org/software/automake/manual/automake.html#Libtool-Convenience-Libraries
> LDADD to do so. This variable is used to specify additional objects or libraries to link with; it is inappropriate for specifying specific linker flags, you should use AM_LDFLAGS for this purpose.

the DPDK_PMDS expands to -Wl,--whole-archives,....., which seems belong to the LDFLAGS? but since it already works it is OK to add in LDADD.

GBalakrishna: Can you add your review if it looks oke ?

/Krishna

muvarov:
Merged.

PR #87, opened 2017-07-25
[PATCH v1] Two fixes to make configure work with strict POSIX shells

lumag:
If /bin/sh points to shell stricter than Bash (e.g. dash), configure will error out due to usage of several constructs which are not available.


muvarov:
Merged.

PR #88, opened 2017-07-26
[PATCH v2] fix odp shm dir

muvarov:
fix odp shm dir. That also I hope will make Linaro CI happy with using ODP_SHM_DIR under docker.


lumag:
@muvarov Can we fold this test into any other test script? Maybe we can always set the ODP_SHM_DIR when running testsuite from Travis?

muvarov:
@lumag  I wanted to test ) if it's not set 2) if it's set. to capture. If put it to existence script there are only 2 places: make check and make distcheck. 'make check' looks better place for me. Do you agree with putting there?

lumag:
@muvarov
Yes, `make check` seems fine to me. I'm just trying to limit amount of
additional test scripts, as each one will take some time to complete.

muvarov:
Merge

PR #89, opened 2017-07-26
[PATCH v1] travis: enable ccache and pip cache

lumag:

### DIFF
--- a/.travis.yml
+++ b/.travis.yml
@@ -55,6 +55,10 @@ compiler:
         - gcc
         - clang-3.8
 
+cache:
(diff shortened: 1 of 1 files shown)


muvarov:
ccache is ok, but why we need pip?

lumag:
On 26/07/17 22:35, muvarov wrote:
> ccache is ok, but why we need pip?

Because we install coverage through pip. Trying to speed up builds as
much as we can.

-- 
With best wishes
Dmitry

muvarov:
Merged.

PR #91, opened 2017-07-31
[PATCH v1] linux-gen: crypto: add support for AES-192 and AES-256

lumag:
Add support for AES with keys of 192 and 256 bits.


lumag:
To be replaced by updated PR basing on api-next. I will need some API changes.

PR #95, opened 2017-07-31
[PATCH API-NEXT v2] api: crypto: clarify for odp_crypto_session_create

lumag:

### DIFF
--- a/include/odp/api/spec/crypto.h
+++ b/include/odp/api/spec/crypto.h
@@ -647,11 +647,12 @@ int odp_crypto_auth_capability(odp_auth_alg_t auth,
  *
  * Create a crypto session according to the session parameters. Use
  * odp_crypto_session_param_init() to initialize parameters into their
- * default values.
+ * default values. If call ends up with an error no new session will be
+ * created.
  *
- * @param param             Session parameters
- * @param session           Created session else ODP_CRYPTO_SESSION_INVALID
- * @param status            Failure code if unsuccessful
+ * @param      param        Session parameters
+ * @param[out] session      Created session else ODP_CRYPTO_SESSION_INVALID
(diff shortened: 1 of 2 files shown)


include/odp/api/spec/crypto.h
```diff
@@ -647,7 +647,8 @@ int odp_crypto_auth_capability(odp_auth_alg_t auth,
  *
  * Create a crypto session according to the session parameters. Use
  * odp_crypto_session_param_init() to initialize parameters into their
- * default values.
+ * default values. If call ends up with an error, 'session' argument won't be
+ * updated and no new session will be created.
```
Bill-Fischofer-Linaro: I'm not sure how this is a clarification. The documentation for the session parameter below clearly states it is set to the created session or ODP_CRYPTO_SESSION_INVALID, so in either case it's an output parameter and is set by this API.

include/odp/api/spec/crypto.h
```diff
@@ -647,7 +647,8 @@ int odp_crypto_auth_capability(odp_auth_alg_t auth,
...
  *
  * @param param             Session parameters
  * @param session           Created session else ODP_CRYPTO_SESSION_INVALID
```
Bill-Fischofer-Linaro: The documentation change needed here is that both session and status should be marked @param[out]

Bill-Fischofer-Linaro:
Looks like one of the Travis checks has failed, but I'm not sure why. Needs to be investigated. Also @psavol should review API spec wording change.

Bill-Fischofer-Linaro:
And once again, asking Travis to re-run that "failing" test causes it to pass. Looks like we definitely have some intermittent failure issues.

muvarov:
@psavol  is that ok?

muvarov:
Merged.

PR #96, opened 2017-07-31
[PATCH v1] test: fix cunit include path

muvarov:
If CUnit is places in non standard directory and path
to it specified with configure option following error
occurs:
fatal error: CUnit/Basic.h: No such file or directory
Patch corrects CFLAGS to make compilation pass.


lumag:

muvarov:
Merged

PR #97, opened 2017-07-31
[PATCH API-NEXT v1] linux-gen: timer: set sigev_value in timer_res_init

lumag:
Even if signal is never to be delivered, it is still required to set
sigev_value in sigevent structure.

### DIFF
--- a/platform/linux-generic/odp_timer.c
+++ b/platform/linux-generic/odp_timer.c
@@ -844,6 +844,7 @@ static int timer_res_init(void)
 
 	sigev.sigev_notify = SIGEV_THREAD_ID;
 	sigev._sigev_un._tid = (pid_t)syscall(SYS_gettid);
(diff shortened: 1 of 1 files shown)


platform/linux-generic/odp_timer.c
```diff
@@ -844,6 +844,7 @@ static int timer_res_init(void)
 
 	sigev.sigev_notify = SIGEV_THREAD_ID;
 	sigev._sigev_un._tid = (pid_t)syscall(SYS_gettid);
+	sigev.sigev_value.sival_ptr = NULL;
```
Bill-Fischofer-Linaro: Is this a bug fix? If yes, it should have a Bug created for it with this being the fix patch.

lumag: Created: https://bugs.linaro.org/show_bug.cgi?id=3178

muvarov:
Merged.

PR #107, opened 2017-08-04
[PATCH API-NEXT v3] linux-gen: timer: control the timer pool polling frequency dynamically

JoyceKong-arm:
The frequency of timer pool polling needs to be adjusted according to
the duration of the timer started in the timer module. And there need
to be 0 timer pool polling when no timer pools/timers created.

At the same time, enable inline timers when passing NULL for odp_init_global.

Signed-off-by: Joyce Kong <joyce.kong@arm.com>


muvarov:
Merged.

PR #114, opened 2017-08-04
[PATCH CLOUD-DEV v2] platform: introduce linux-dpdk from odp-dpd.git

GBalakrishna:
For the movement linux-dpdk can only be built with setting "PKG_CONFIG" manually during configure.


GBalakrishna:
I believe many of the chaeckpatch issues are false positive. And about making it smaller chunks I will have to find a way.

PR #119, opened 2017-08-09
[PATCH v1] linux-gen: dpdk: make sure mbuf memory is allocated from huge pages

matiaselo:
DPDK requires that mbuf memory is allocated from huge pages to ensure page
locking. Modify zero-copy dpdk pktio to fall back to packet copy if
transmitted ODP packet is not allocated from huge page memory.

### DIFF
--- a/platform/linux-generic/pktio/dpdk.c
+++ b/platform/linux-generic/pktio/dpdk.c
@@ -129,14 +129,13 @@ static void mbuf_init(struct rte_mempool *mp, struct rte_mbuf *mbuf,
 {
 	void *buf_addr = pkt_hdr->buf_hdr.base_data - RTE_PKTMBUF_HEADROOM;
 
-	rte_mem_lock_page(buf_addr);
-
 	memset(mbuf, 0, sizeof(struct rte_mbuf));
 
 	mbuf->priv_size = 0;
 	mbuf->buf_addr = buf_addr;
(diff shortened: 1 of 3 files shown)


muvarov:
Looks good for me.

platform/linux-generic/pktio/dpdk.c
```diff
@@ -163,6 +162,11 @@ static struct rte_mempool *mbuf_pool_create(const char *name,
 	unsigned num;
 	uint16_t data_room_size;
 
+	if (!(pool_entry->mem_from_huge_pages)) {
+		ODP_ERR("DPDK requires memory is allocated from huge pages\n");
+		return NULL;
```
Bill-Fischofer-Linaro: Is this true? This code was passing prior to this test being added, so was that just luck? I can understand that huge pages are preferred, but are they required?

muvarov: @Bill-Fischofer-Linaro  different pools can be allocated from different memory.

matiaselo: @Bill-Fischofer-Linaro If the memory is not from huge pages the physical address may not be available since the memory is not locked. DPDK NIC drivers require physical addresses to do DMA.

The tests have been passing because the packet pools are always allocated from huge pages. I noticed this problem when I happened to run out of free huge pages on my test system.

muvarov:
Merged with my review.

PR #120, opened 2017-08-09
[PATCH CLOUD-DEV v4] Move pool to modular framework

kevinwangsk:
Separate with two commits.
First one is to add the pool subsystem in modular framework.
Second one is to add the generic pool implementation to the pool subsystem.

### DIFF
--- a/platform/linux-generic/include/odp_pool_subsystem.h
+++ b/platform/linux-generic/include/odp_pool_subsystem.h
@@ -1,0 +1,50 @@
+/* Copyright (c) 2017, ARM Limited. All rights reserved.
+ *
+ * Copyright (c) 2017, Linaro Limited
+ * All rights reserved.
+ *
+ * SPDX-License-Identifier: BSD-3-Clause
+ */
+
+#ifndef ODP_POOL_SUBSYSTEM_H_
+#define ODP_POOL_SUBSYSTEM_H_
+
+#ifdef __cplusplus
+extern "C" {
+#endif
+
+#include <odp_module.h>
+#include <odp/api/pool.h>
+
+/* ODP pool public APIs subsystem */
+ODP_SUBSYSTEM_DECLARE(pool);
+
+/* Subsystem APIs declarations */
+ODP_SUBSYSTEM_API(pool, int, capability, odp_pool_capability_t *capa);
+ODP_SUBSYSTEM_API(pool, odp_pool_t, create, const char *name,
+		  odp_pool_param_t *params);
+ODP_SUBSYSTEM_API(pool, int, destroy, odp_pool_t pool);
+ODP_SUBSYSTEM_API(pool, odp_pool_t, lookup, const char *name);
+ODP_SUBSYSTEM_API(pool, int, info, odp_pool_t pool, odp_pool_info_t *info);
+ODP_SUBSYSTEM_API(pool, void, print, odp_pool_t pool);
+ODP_SUBSYSTEM_API(pool, uint64_t, to_u64, odp_pool_t pool);
+ODP_SUBSYSTEM_API(pool, void, param_init, odp_pool_param_t *params);
+
+typedef ODP_MODULE_CLASS(pool) {
+	odp_module_base_t base;
+
+	odp_api_proto(pool, capability) capability;
+	odp_api_proto(pool, create) create;
+	odp_api_proto(pool, destroy) destroy;
+	odp_api_proto(pool, lookup) lookup;
+	odp_api_proto(pool, info) info;
+	odp_api_proto(pool, print) print;
+	odp_api_proto(pool, to_u64) to_u64;
+	odp_api_proto(pool, param_init) param_init;
+} pool_module_t;
+
+#ifdef __cplusplus
+}
+#endif
+
+#endif
--- a/platform/linux-generic/pool/subsystem.c
+++ b/platform/linux-generic/pool/subsystem.c
@@ -5,6 +5,10 @@
  *
  * SPDX-License-Identifier: BSD-3-Clause
  */
+#include <odp/api/pool.h>
+#include <odp_internal.h>
+#include <odp_pool_subsystem.h>
+#include <odp_debug_internal.h>
 #include <odp_module.h>
 
 #define SUBSYSTEM_VERSION 0x00010000UL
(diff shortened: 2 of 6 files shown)


heyi-linaro:
It failed again at the same place, Kevin please have a look into the Travis CI log and see if you can replicate the problem.

kevinwangsk:
No,  I am not able to reproduce it on my system. From my previous experiences, the helper test case failures are always caused by no enough memory. Does anybody runs those test cases on the test bed?

kevinwangsk:
Seems "pool/generic.lo: CFLAGS += -DIM_ACTIVE_MODULE" doesn't take affect in the 20th test case. So there is not active module in pool subsystem. Need take some more time to figure out the reason.

heyi-linaro:
@Bill-Fischofer-Linaro Hi, Bill this PR has fixed and passed the Travis CI, I've reviewed and if it looks OK to you I'll merge it.

Bill-Fischofer-Linaro:
I'll review this PR today. Thanks.

platform/linux-generic/pool/subsystem.c
```diff
@@ -15,3 +19,219 @@ ODP_SUBSYSTEM_CONSTRUCTOR(pool)
...
+	mod = odp_subsystem_active_module(pool, mod);
+	if (mod == NULL) {
+		ODP_ERR("No active module in pool subsystem\n");
+		return -1;
+	}
+	if (mod->base.init_global == NULL) {
+		ODP_ERR("No defined init_global function "
+			"in module %s of pool subsystem\n", mod->base.name);
+		return -1;
+	}
```
Bill-Fischofer-Linaro: This pattern is highly repetitive, which suggests it should be factored out as part of the modular framework itself rather than having to be repeated for every API. I fact it seems that the entire subsystem.c file for each module could be auto-generated since there's no real variant code here.

platform/linux-generic/include/odp_pool_subsystem.h
```diff
@@ -0,0 +1,50 @@
...
+	odp_module_base_t base;
+
+	odp_api_proto(pool, capability) capability;
+	odp_api_proto(pool, create) create;
+	odp_api_proto(pool, destroy) destroy;
+	odp_api_proto(pool, lookup) lookup;
+	odp_api_proto(pool, info) info;
+	odp_api_proto(pool, print) print;
+	odp_api_proto(pool, to_u64) to_u64;
+	odp_api_proto(pool, param_init) param_init;
```
Bill-Fischofer-Linaro: Any reason why these shouldn't be alphabetical? That's been a standard we've used in the various .am and related files.

platform/linux-generic/pool/subsystem.c
```diff
@@ -0,0 +1,17 @@
+/* Copyright (c) 2017, ARM Limited. All rights reserved.
+ *
+ * Copyright (c) 2017, Linaro Limited
+ * All rights reserved.
+ *
+ * SPDX-License-Identifier: BSD-3-Clause
+ */
+#include <odp_module.h>
+
+#define SUBSYSTEM_VERSION 0x00010000UL
```
Bill-Fischofer-Linaro: @muvarov I'm not sure why GitHub is flagging this comment as "outdated" as it relates to the current (v4) PR.

heyi-linaro: This version info does mean per each subsystem, in future the usage could be that modules who impl this subsystem can check the version compatibility, like, for example:

NXP module impls "pktio" subsystem version 1.0, but the library contains "pktio" subsystem version 2.0.

For versioning there can be a large feature design, currently only this early support to be able to specify a version number for each subsystem in definition.

platform/linux-generic/pool/subsystem.c
heyi-linaro: Yes, I leave it to user programmer, so different policies can be impled, like:

1) IF a subsystem assumes all its api pointers should not be NULL, it can bypass some checks.
2) IF a subsystem assumes at least one active module exist, it can bypass some checks.
3) a subsystem can write these checks in subsystem->init_global() to assure active module exists and/or api pointers are not NULL.
4) in least case, user programmer can write inline/macro to check per their subsystem's requirements.

platform/linux-generic/include/odp_pool_subsystem.h
kevinwangsk: This is to define the function pointer for the public APIs. So I am just following the sequence in ODP API doc.

platform/linux-generic/pool/subsystem.c
Bill-Fischofer-Linaro: OK, so you're saying that this is unique to each module and could just as easily be specified directly in the `ODP_SUBSYSTEM_DEFINE()` macro below?  So how is this used to detect version mismatches? Or is it purely for documentation purposes at this stage? If we're supporting separate loadable modules then it would seem some type of load-time version compatibility check is needed.

platform/linux-generic/pool/subsystem.c
Bill-Fischofer-Linaro: It makes sense to do these checks at init time when dynamic loading is involved, but once a module is loaded and bound these just add overhead, which should be unnecessary. The `odp_pool_xxx()` APIs are not executed with great frequency so this overhead may not matter here, but for things like pktio operations such extra overhead would be very undesirable.

platform/linux-generic/pool/subsystem.c
heyi-linaro: Yes, current only ODP_SUBSYSTEM_DEFINE() specifies this version info, the usage can come late with a design (maybe its too earlier to have it here).

platform/linux-generic/pool/subsystem.c
heyi-linaro: Yes, we saw old code also does not check NULL pointer errors in API wrappers, here can do the same and assume all modules understand this requirement.

kevinwangsk: Agree that we just need to check the active module at the init function. After that, there should be an active module before invoke any other pool operations. This should be the same with other subsystem.

heyi-linaro:
Merged and closed.

PR #124, opened 2017-08-14
[PATCH v4] dpdk pktio: enable hardware checksum support

bogdanPricope:
Add HW checksum calculation/validation support for dpdk pktio.

Note: No UDP/TCP HW checksum calculation/validation support for:
 - IPv4 fragments
 - IPv6 packets with extension headers (including fragments)

### DIFF
--- a/example/generator/odp_generator.c
+++ b/example/generator/odp_generator.c
@@ -46,6 +46,7 @@
 
 typedef struct {
 	odp_pktio_t pktio;
+	odp_pktio_config_t config;
 	odp_pktout_queue_t pktout[MAX_WORKERS];
 	unsigned pktout_count;
 } interface_t;
@@ -91,7 +92,16 @@ static struct {
 /** * Thread specific arguments
  */
 typedef struct {
-	odp_pktout_queue_t pktout; /**< Packet output queue to use*/
+	union {
+		struct {
+			odp_pktout_queue_t pktout; /**< Packet output queue */
+			odp_pktout_config_opt_t *pktout_cfg; /**< Packet output config*/
+		} tx;
+		struct {
+			interface_t *ifs; /**< Interfaces array */
+			int ifs_count; /**< Interfaces array size */
+		} rx;
+	};
 	odp_pool_t pool;	/**< Pool for packet IO */
 	odp_timer_pool_t tp;	/**< Timer pool handle */
 	odp_queue_t tq;		/**< Queue for timeouts */
@@ -115,6 +125,11 @@ static args_t *args;
 
 /** Barrier to sync threads execution */
 static odp_barrier_t barrier;
+
+/** Packet processing function types */
+typedef odp_packet_t (*setup_pkt_ref_fn_t)(odp_pool_t,
+					   odp_pktout_config_opt_t *);
+typedef int (*setup_pkt_fn_t)(odp_packet_t, odp_pktout_config_opt_t *);
 
 /* helper funcs */
 static void parse_args(int argc, char *argv[], appl_args_t *appl_args);
@@ -190,20 +205,22 @@ static int scan_ip(char *buf, unsigned int *paddr)
  * Setup array of reference packets
  *
  * @param pool Packet pool
+ * @param pktout_cfg Interface output configuration
  * @param pkt_ref_array Packet array
  * @param pkt_ref_array_size Packet array size
  * @param setup_ref Packet setup function
  * @return 0 success, -1 failed
 */
 static int setup_pkt_ref_array(odp_pool_t pool,
+			       odp_pktout_config_opt_t *pktout_cfg,
 			       odp_packet_t *pkt_ref_array,
 			       int pkt_ref_array_size,
-			       odp_packet_t (*setup_ref)(odp_pool_t))
+			       setup_pkt_ref_fn_t setup_ref)
 {
 	int i;
 
 	for (i = 0; i < pkt_ref_array_size; i++) {
-		pkt_ref_array[i] = (*setup_ref)(pool);
+		pkt_ref_array[i] = (*setup_ref)(pool, pktout_cfg);
 		if (pkt_ref_array[i] == ODP_PACKET_INVALID)
 			break;
 	}
@@ -218,21 +235,23 @@ static int setup_pkt_ref_array(odp_pool_t pool,
 /**
  * Setup array of packets
  *
+ * @param pktout_cfg Interface output configuration
  * @param pkt_ref_array Reference packet array
  * @param pkt_array Packet array
  * @param pkt_array_size Packet array size
  * @param setup_pkt Packet setup function
  * @return 0 success, -1 failed
 */
-static int setup_pkt_array(odp_packet_t *pkt_ref_array,
+static int setup_pkt_array(odp_pktout_config_opt_t *pktout_cfg,
+			   odp_packet_t *pkt_ref_array,
 			   odp_packet_t  *pkt_array,
 			   int pkt_array_size,
-			   int (*setup_pkt)(odp_packet_t))
+			   setup_pkt_fn_t setup_pkt)
 {
 	int i;
 
 	for (i = 0; i < pkt_array_size; i++) {
-		if ((*setup_pkt)(pkt_ref_array[i]))
+		if ((*setup_pkt)(pkt_ref_array[i], pktout_cfg))
 			break;
 
 		pkt_array[i] = odp_packet_ref_static(pkt_ref_array[i]);
@@ -252,13 +271,15 @@ static int setup_pkt_array(odp_packet_t *pkt_ref_array,
  * set up an udp packet reference
  *
  * @param pool Buffer pool to create packet in
+ * @param pktout_cfg Interface output configuration
  *
  *
  * @retval Handle of created packet
  * @retval ODP_PACKET_INVALID  Packet could not be created
  *
  */
-static odp_packet_t setup_udp_pkt_ref(odp_pool_t pool)
+static odp_packet_t setup_udp_pkt_ref(odp_pool_t pool,
+				      odp_pktout_config_opt_t *pktout_cfg)
 {
 	odp_packet_t pkt;
 	char *buf;
@@ -302,8 +323,10 @@ static odp_packet_t setup_udp_pkt_ref(odp_pool_t pool)
 	udp->src_port = odp_cpu_to_be_16(args->appl.srcport);
 	udp->dst_port = odp_cpu_to_be_16(args->appl.dstport);
 	udp->length = odp_cpu_to_be_16(args->appl.payload + ODPH_UDPHDR_LEN);
-	udp->chksum = 0;
-	udp->chksum = odph_ipv4_udp_chksum(pkt);
+	if (!pktout_cfg->bit.udp_chksum) {
+		udp->chksum = 0;
+		udp->chksum = odph_ipv4_udp_chksum(pkt);
+	}
 
 	return pkt;
 }
@@ -312,11 +335,12 @@ static odp_packet_t setup_udp_pkt_ref(odp_pool_t pool)
  * set up an udp packet
  *
  * @param pkt Reference UDP packet
+ * @param pktout_cfg Interface output configuration
  *
  * @return Success/Failed
  * @retval 0 on success, -1 on fail
  */
-static int setup_udp_pkt(odp_packet_t pkt)
+static int setup_udp_pkt(odp_packet_t pkt, odp_pktout_config_opt_t *pktout_cfg)
 {
 	char *buf;
 	odph_ipv4hdr_t *ip;
@@ -328,9 +352,17 @@ static int setup_udp_pkt(odp_packet_t pkt)
 	ip = (odph_ipv4hdr_t *)(buf + ODPH_ETHHDR_LEN);
 	seq = odp_atomic_fetch_add_u64(&counters.seq, 1) % 0xFFFF;
 	ip->id = odp_cpu_to_be_16(seq);
-	ip->chksum = 0;
-	ip->chksum = odph_chksum(ip, ODPH_IPV4HDR_LEN);
+	if (!pktout_cfg->bit.ipv4_chksum) {
+		ip->chksum = 0;
+		ip->chksum = odph_chksum(ip, ODPH_IPV4HDR_LEN);
+	}
 
+	if (pktout_cfg->bit.ipv4_chksum || pktout_cfg->bit.udp_chksum) {
+		odp_packet_l2_offset_set(pkt, 0);
+		odp_packet_l3_offset_set(pkt, ODPH_ETHHDR_LEN);
+		odp_packet_l4_offset_set(pkt, ODPH_ETHHDR_LEN +
+					 ODPH_IPV4HDR_LEN);
+	}
 	return 0;
 }
 
@@ -338,17 +370,21 @@ static int setup_udp_pkt(odp_packet_t pkt)
  * Set up an icmp packet reference
  *
  * @param pool Buffer pool to create packet in
+ * @param pktout_cfg Interface output configuration
  *
  * @return Handle of created packet
  * @retval ODP_PACKET_INVALID  Packet could not be created
  */
-static odp_packet_t setup_icmp_pkt_ref(odp_pool_t pool)
+static odp_packet_t setup_icmp_pkt_ref(odp_pool_t pool,
+				       odp_pktout_config_opt_t *pktout_cfg)
 {
 	odp_packet_t pkt;
 	char *buf;
 	odph_ethhdr_t *eth;
 	odph_ipv4hdr_t *ip;
 	odph_icmphdr_t *icmp;
+
+	(void)pktout_cfg;
 
 	args->appl.payload = 56;
 	pkt = odp_packet_alloc(pool, args->appl.payload + ODPH_ICMPHDR_LEN +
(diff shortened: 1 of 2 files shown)


muvarov:
Commits short message have to be:
linux-gen: pktio: dpdk: <some name>

take a look at:
git log --oneline ./platform/linux-generic/pktio/dpdk.c

muvarov:
 this patch has to be:

linux-gen: pktio:
 </pre>

platform/linux-generic/odp_packet_io.c
```diff
@@ -150,6 +150,8 @@ static void init_pktio_entry(pktio_entry_t *entry)
 	init_out_queues(entry);
 
 	pktio_classifier_init(entry);
+
+	odp_pktio_config_init(&entry->s.config);
```
matiaselo: It seem odp_pktio_config_init() is now called twice for the same entry as it's already called in setup_pktio_entry().

platform/linux-generic/pktio/dpdk.c
```diff
@@ -301,6 +301,40 @@ static struct rte_mempool_ops ops_stack = {
...
+#define HAS_L4_PROTO(m, proto) ((m->packet_type & RTE_PTYPE_L4_MASK) == proto)
+
+static inline int pkt_set_ol_rx(odp_pktin_config_opt_t *pktin_cfg,
+				odp_packet_hdr_t *pkt_hdr,
+				struct rte_mbuf *mbuf)
+{
+	if (pktin_cfg->bit.ipv4_chksum &&
+	    (mbuf->packet_type & RTE_PTYPE_L3_IPV4) &&
+	    mbuf->ol_flags & PKT_RX_IP_CKSUM_BAD) {
+		if (pktin_cfg->bit.drop_ipv4_err)
```
matiaselo: Based on the DPDK test-pmd code it seems that checking packet_type is not actually needed. Have you tried this?

Checking PKT_RX_IP_CKSUM_BAD alone is deprecated and one should use PKT_RX_IP_CKSUM_MASK.
 E.g. (ol_flags & PKT_RX_IP_CKSUM_MASK) == PKT_RX_IP_CKSUM_BAD

platform/linux-generic/pktio/dpdk.c
```diff
@@ -301,6 +301,40 @@ static struct rte_mempool_ops ops_stack = {
...
+			return -1;
+
+		pkt_hdr->p.error_flags.ip_err = 1;
+	}
+
+	if (pktin_cfg->bit.udp_chksum &&
+	    HAS_L4_PROTO(mbuf, RTE_PTYPE_L4_UDP) &&
+	    mbuf->ol_flags & PKT_RX_L4_CKSUM_BAD) {
```
matiaselo: Same as above.

platform/linux-generic/pktio/dpdk.c
```diff
@@ -301,6 +301,40 @@ static struct rte_mempool_ops ops_stack = {
...
+		if (pktin_cfg->bit.drop_udp_err)
+			return -1;
+
+		pkt_hdr->p.error_flags.udp_err = 1;
+	} else if (pktin_cfg->bit.tcp_chksum &&
+		   HAS_L4_PROTO(mbuf, RTE_PTYPE_L4_TCP)  &&
+		   mbuf->ol_flags & PKT_RX_L4_CKSUM_BAD) {
```
matiaselo: Same as above.

platform/linux-generic/pktio/dpdk.c
```diff
@@ -369,6 +404,12 @@ static inline int mbuf_to_pkt(pktio_entry_t *pktio_entry,
 
 		packet_set_ts(pkt_hdr, ts);
 
+		if (pkt_set_ol_rx(pktin_cfg, pkt_hdr, mbuf)) {
```
matiaselo: You could create a mask from ipv4_chksum, udp_chksum and tcp_chksum, and compare it to pktin_cfg->all_bits before calling pkt_set_ol_rx() to reduce the number of if branches.

platform/linux-generic/pktio/dpdk.c
```diff
@@ -480,6 +522,11 @@ static inline int mbuf_to_pkt_zero(pktio_entry_t *pktio_entry,
...
```
matiaselo: Same as above.

platform/linux-generic/pktio/dpdk.c
```diff
@@ -426,6 +430,96 @@ static inline int mbuf_to_pkt(pktio_entry_t *pktio_entry,
...
+				odp_packet_hdr_t *pkt_hdr,
+				struct rte_mbuf *mbuf,
+				char *mbuf_data)
+{
+	void *l3_hdr, *l4_hdr;
+	uint8_t l3_proto_v4, l4_proto;
+	odp_bool_t ipv4_chksum_pkt, udp_chksum_pkt, tcp_chksum_pkt;
+	packet_parser_t *pkt_p = &pkt_hdr->p;
+
+	l3_hdr = (void *)(mbuf_data + pkt_p->l3_offset);
```
matiaselo: It would be good to check that l3_offset is valid.

platform/linux-generic/pktio/dpdk.c
```diff
@@ -426,6 +430,96 @@ static inline int mbuf_to_pkt(pktio_entry_t *pktio_entry,
...
+		return rte_ipv6_phdr_cksum(l3_hdr, ol_flags);
+}
+
+static inline int pkt_set_ol_tx(odp_pktout_config_opt_t *pktout_cfg,
+				odp_packet_hdr_t *pkt_hdr,
+				struct rte_mbuf *mbuf,
+				char *mbuf_data)
+{
+	void *l3_hdr, *l4_hdr;
+	uint8_t l3_proto_v4, l4_proto;
```
GBalakrishna: l3_proto_v4 is only set to either 1 or 0. It would be good to make this as odp_bool_t, so it would be clear when reading.

platform/linux-generic/pktio/dpdk.c
```diff
@@ -426,6 +430,96 @@ static inline int mbuf_to_pkt(pktio_entry_t *pktio_entry,
...
+
+		*l3_proto_v4 = 0;
+		*l4_proto = ipv6->proto;
+		return 0;
+	}
+
+	return -1;
+}
+
+static inline uint16_t phdr_csum(int ipv4, void *l3_hdr,
```
GBalakrishna: l3_proto_v4 was declared as unit8_t but it's passed as "int" here. Better to replace this with odp_bool_t as it's mainly true or false

platform/linux-generic/pktio/dpdk.c
```diff
@@ -908,9 +913,56 @@ static void dpdk_init_capability(pktio_entry_t *pktio_entry,
...
+
+	capa->config.pktin.bit.udp_chksum = ptype_l4_udp &&
+		(dev_info->rx_offload_capa & DEV_RX_OFFLOAD_UDP_CKSUM) ? 1 : 0;
+	if (capa->config.pktin.bit.udp_chksum)
+		capa->config.pktin.bit.drop_udp_err = 1;
+
+	capa->config.pktin.bit.tcp_chksum = ptype_l4_tcp &&
+		(dev_info->rx_offload_capa & DEV_RX_OFFLOAD_TCP_CKSUM) ? 1 : 0;
+	if (capa->config.pktin.bit.tcp_chksum)
+		capa->config.pktin.bit.drop_tcp_err = 1;
```
GBalakrishna: do we need to check the "DEV_RX_OFFLOAD_TCP_CKSUM" again even after knowing "ptype_l4_tcp" ?

platform/linux-generic/odp_packet_io.c
bogdanPricope: Indeed. It was fixed in the meantime by you in „linux-gen: packet: remove lazy parsing” ce8bebad45c590ad5b06d31d8be9fae8182f944a

platform/linux-generic/pktio/dpdk.c
```diff
@@ -908,9 +913,56 @@ static void dpdk_init_capability(pktio_entry_t *pktio_entry,
...
+		uint32_t ptypes[ptype_cnt];
+		int i;
+
+		ptype_cnt = rte_eth_dev_get_supported_ptypes(pkt_dpdk->port_id,
+							     ptype_mask, ptypes,
+							     ptype_cnt);
+		for (i = 0; i < ptype_cnt; i++)
+			switch (ptypes[i]) {
+			case RTE_PTYPE_L3_IPV4:
+			case RTE_PTYPE_L3_IPV4_EXT_UNKNOWN:
```
bogdanPricope: Added '/* Fall through */'

platform/linux-generic/pktio/dpdk.c
bogdanPricope: ptype_l4_tcp says if dpdk (NIC) can recognize TCP packet type while DEV_RX_OFFLOAD_TCP_CKSUM says if it can do the offload.
We need to use both reported ptype and reported offload flags to determine result of HW csum validation.

platform/linux-generic/pktio/dpdk.c
bogdanPricope: OK

platform/linux-generic/pktio/dpdk.c
bogdanPricope: OK

platform/linux-generic/pktio/dpdk.c
bogdanPricope: OK

platform/linux-generic/pktio/dpdk.c
bogdanPricope: OK

platform/linux-generic/pktio/dpdk.c
bogdanPricope: OK

platform/linux-generic/pktio/dpdk.c
```diff
@@ -301,6 +301,48 @@ static struct rte_mempool_ops ops_stack = {
 
 MEMPOOL_REGISTER_OPS(ops_stack);
 
+#define HAS_IP4_PROTO(m, proto) ((m->packet_type & RTE_PTYPE_L3_MASK) == proto)
+#define HAS_IP4_PROTOS(m) (HAS_IP4_PROTO(m, RTE_PTYPE_L3_IPV4) ||\
+			   HAS_IP4_PROTO(m, RTE_PTYPE_L3_IPV4_EXT) ||\
+			   HAS_IP4_PROTO(m, RTE_PTYPE_L3_IPV4_EXT_UNKNOWN))
```
matiaselo: This looks unnecessarily complex. You can simply do e.g.
`#define PTYPE_IPV4 (RTE_PTYPE_L3_IPV4 | RTE_PTYPE_L3_IPV4_EXT | RTE_PTYPE_L3_IPV4_EXT_UNKNOWN)`
and OR that with `(m->packet_type & RTE_PTYPE_L3_MASK)`.

platform/linux-generic/pktio/dpdk.c
```diff
@@ -436,6 +440,97 @@ static inline int mbuf_to_pkt(pktio_entry_t *pktio_entry,
 	return (i > 0 ? i : -1);
 }
 
+static inline int packet_parse(void *l3_hdr, odp_bool_t *l3_proto_v4,
+			       uint8_t *l4_proto)
```
matiaselo: Confusing function name. We already have packet_parse_*() functions in odp_packet.c which do the input packet parsing. In my opinion e.g. check_proto() would be a better name.

platform/linux-generic/pktio/dpdk.c
```diff
@@ -436,6 +440,97 @@ static inline int mbuf_to_pkt(pktio_entry_t *pktio_entry,
...
+	if (ipv4)
+		return rte_ipv4_phdr_cksum(l3_hdr, ol_flags);
+	else /*ipv6*/
+		return rte_ipv6_phdr_cksum(l3_hdr, ol_flags);
+}
+
+static inline int pkt_set_ol_tx(odp_pktout_config_opt_t *pktout_cfg,
+				odp_packet_hdr_t *pkt_hdr,
+				struct rte_mbuf *mbuf,
+				char *mbuf_data)
```
matiaselo: Function return value is never checked so it can be void.

example/generator/odp_generator.c
```diff
@@ -116,6 +126,10 @@ static args_t *args;
 /** Barrier to sync threads execution */
 static odp_barrier_t barrier;
 
+/** Packet processing function types */
+typedef odp_packet_t (*setup_pkt_ref_t)(odp_pool_t, odp_pktout_config_opt_t *);
+typedef int (*setup_pkt_t)(odp_packet_t, odp_pktout_config_opt_t *);
+
```
matiaselo: Minor thing but since these are function pointer types I would  prefer _fn_t postfix.

example/generator/odp_generator.c
```diff
@@ -742,6 +809,23 @@ static int gen_recv_thread(void *arg)
...
 		for (i = 0, pkt_cnt = 0; i < ev_cnt; i++) {
 			pkt = odp_packet_from_event(events[i]);
+			itf = &itfs[odp_pktio_index(odp_packet_input(pkt))];
+
+			if (odp_packet_has_ipv4(pkt)) {
+				if (itf->config.pktin.bit.ipv4_chksum) {
+					/* HW validation */
+					if (odp_packet_has_l3_error(pkt))
+						printf("HW detected L3 error\n");
+				} /* else SW validation */
```
matiaselo: SW code path missing?

example/generator/odp_generator.c
```diff
@@ -742,6 +809,23 @@ static int gen_recv_thread(void *arg)
...
+			}
+
+			if (odp_packet_has_udp(pkt)) {
+				if (itf->config.pktin.bit.udp_chksum) {
+					/* HW validation*/
+					if (odp_packet_has_l4_error(pkt))
+						printf("HW detected L4 error\n");
+				} /* else SW validation */
```
matiaselo: SW code path missing?

matiaselo:
All commits are missing rationale in commit log.

platform/linux-generic/pktio/dpdk.c
bogdanPricope: I am not sure what you mean:

PTYPE_IPV4 = (RTE_PTYPE_L3_IPV4 | RTE_PTYPE_L3_IPV4_EXT | RTE_PTYPE_L3_IPV4_EXT_UNKNOWN) = 0001 0000 | 0011 0000 | 1001 0000 = 1011 0000

If packet is IPv6 with extensions:

m->packet_type & RTE_PTYPE_L3_MASK = RTE_PTYPE_L3_IPV6_EXT = 1100 0000

1011 0000 | 1100 0000 is not helping us.

matiaselo: True, my mistake. Looking at the binary presentations it looks like you can simply use RTE_PTYPE_L3_IPV4 and RTE_PTYPE_L3_IPV6 values and do OR operation. Those bits won't overlap.

bogdanPricope: True, yet you cannot really trust that those defines will keep their values in the next dpdk versions. This is how APIs are extended.
We should not design this implementation based on the values but based on defines (RTE_PTYPE_L3_XXX and RTE_PTYPE_L3_MASK).

platform/linux-generic/pktio/dpdk.c
bogdanPricope: OK

platform/linux-generic/pktio/dpdk.c
bogdanPricope: OK

platform/linux-generic/pktio/dpdk.c
bogdanPricope: /**
 * Packet output configuration options bit field
 *
 * Packet output configuration options listed in a bit field structure. Packet
 * output checksum insertion may be enabled or disabled. When it is enabled,
 * implementation will calculate and insert checksum into every outgoing packet
 * by default. Application may use a packet metadata flag to disable checksum
 * insertion per packet bases. For correct operation, packet metadata must
 * provide valid offsets for the appropriate protocols. For example, UDP
 * checksum calculation needs both L3 and L4 offsets (to access IP and UDP
 * headers). When application (e.g. a switch) does not modify L3/L4 data and
 * thus checksum does not need to be updated, output checksum insertion should
 * be disabled for optimal performance.
 */

Petri's comment from May 10th:
"API already says that outgoing packet must have (L3/L4) offsets set correctly if (e.g. L4) checksum calculation is offloaded."

Also, is not possible to determine if l3 and l4 offsets are incorrect (wrong values).

example/generator/odp_generator.c
bogdanPricope: ok

example/generator/odp_generator.c
bogdanPricope: SW checking code was deliberately missed: this is a ‘load’ application and SW operations are avoided.

I added this code mainly to highlight how the csum errors detected by ODP implementation can be interpreted by application. We can keep it as ‘code example’ or remove this part since has no other benefit (packets with errors are freed immediately after this code sample).

example/generator/odp_generator.c
bogdanPricope: Same as above.

muvarov:
@matiaselo is everything good now? Please provide your review-by if it is.

muvarov:
Merged.

PR #127, opened 2017-08-15
[PATCH v6] Continue working on improving build system

lumag:
This PR focuses on replacing hand-coded defines with AC_DEFINE/AC_DEFINE_UNQUOTED macros.

### DIFF
--- a/configure.ac
+++ b/configure.ac
@@ -20,8 +20,8 @@ AC_CONFIG_FILES([include/odp/api/spec/version.h
                  include/odp/api/spec/deprecated.h])
 
 AM_INIT_AUTOMAKE([1.9 tar-pax subdir-objects])
-AC_CONFIG_SRCDIR([helper/config.h.in])
-AM_CONFIG_HEADER([helper/config.h])
+AC_CONFIG_SRCDIR([include/odp/api/spec/init.h])
+AM_CONFIG_HEADER([include/config.h])
 
 AC_USE_SYSTEM_EXTENSIONS
 AC_SYS_LARGEFILE
--- a/configure.ac
+++ b/configure.ac
@@ -63,9 +63,10 @@ AC_SUBST([LIBTOOL_DEPS])
 AM_PROG_LIBTOOL
 
 # Checks for library functions.
-AC_FUNC_MALLOC
+dnl breaks cross-compilation and malloc(0) behaviour is not that important
+dnl AC_FUNC_MALLOC
 AC_FUNC_MMAP
-AC_CHECK_FUNCS([bzero clock_gettime gethostbyname getpagesize gettimeofday memset munmap socket strchr strerror strrchr strstr strtoull])
+AC_CHECK_FUNCS([bzero clock_gettime gethostbyname getpagesize gettimeofday malloc memset munmap socket strchr strerror strrchr strstr strtoull])
 
 # Checks for header files.
 AC_HEADER_RESOLV
--- a/configure.ac
+++ b/configure.ac
@@ -261,10 +261,9 @@ AC_ARG_ENABLE([helper-debug-print],
     [  --enable-helper-debug-print    display helper debugging information],
     [if test "x$enableval" = "xyes"; then
 	ODPH_DEBUG_PRINT=1
-     else
-	ODPH_DEBUG_PRINT=0
     fi])
-ODP_CFLAGS="$ODP_CFLAGS -DODPH_DEBUG_PRINT=$ODPH_DEBUG_PRINT"
+AC_DEFINE_UNQUOTED([ODPH_DEBUG_PRINT], [$ODPH_DEBUG_PRINT],
+		   [Define to 1 to display helper debug information])
 
 ##########################################################################
 # Enable/disable ODP_DEBUG
--- a/configure.ac
+++ b/configure.ac
@@ -252,10 +252,9 @@ AC_ARG_ENABLE([debug-print],
     [  --enable-debug-print    display debugging information],
     [if test "x$enableval" = "xyes"; then
 	ODP_DEBUG_PRINT=1
-     else
-	ODP_DEBUG_PRINT=0
     fi])
-ODP_CFLAGS="$ODP_CFLAGS -DODP_DEBUG_PRINT=$ODP_DEBUG_PRINT"
+AC_DEFINE_UNQUOTED([ODP_DEBUG_PRINT], [$ODP_DEBUG_PRINT],
+		   [Define to 1 to display debug information])
 
 ODPH_DEBUG_PRINT=0
 AC_ARG_ENABLE([helper-debug-print],
--- a/configure.ac
+++ b/configure.ac
@@ -174,7 +174,8 @@ else
     exit 1
 fi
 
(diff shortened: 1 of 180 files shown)


platform/linux-generic/_fdserver.c
```diff
@@ -4,6 +4,10 @@
  * SPDX-License-Identifier:     BSD-3-Clause
  */
 
+#ifdef HAVE_CONFIG_H
```
muvarov: how about move that ifdef inside config.h?

configure.ac
```diff
@@ -273,10 +273,9 @@ AC_ARG_ENABLE([debug],
     [  --enable-debug          include additional code],
     [if test "x$enableval" = "xyes"; then
 	ODP_DEBUG=1
-     else
-	ODP_DEBUG=0
     fi])
-ODP_CFLAGS="$ODP_CFLAGS -DODP_DEBUG=$ODP_DEBUG"
+AC_DEFINE_UNQUOTED([ODP_DEBUG], [$ODP_DEBUG],
+		   [Define to 1 to include additional debug code])
```
muvarov: I think it should be configure option, that define in configure.ac.

platform/linux-generic/m4/odp_pcap.m4
```diff
@@ -9,7 +9,7 @@ AC_CHECK_HEADER(pcap/pcap.h,
 [])
 
 if test "$have_pcap" = "yes"; then
-    ODP_CFLAGS="$AM_CFLAGS -DHAVE_PCAP"
+    AC_DEFINE([HAVE_PCAP], 1, [Define to 1 if you have pcap library])
```
muvarov: "Define to 1 if you have pcap library" has to be "Defined to 1 if pcap library found"

configure.ac
lumag: Defined value depends on configure option.

platform/linux-generic/m4/odp_pcap.m4
lumag: Compare this with other descriptions in config.h: 
- `Define to 1 if you have the ANSI C header files`
- `Define to 'int' if <sys/types.h> does not define.`
I think mine version is more in style with them.

platform/linux-generic/_fdserver.c
lumag: Hmm. How can we move define that guards config.h presence into config.h?

muvarov: at the beginning of config.h @ifdef HAVE_CONFIG_H and #endif in the end.

lumag: @muvarov. Ghm. How would source file check if config.h is available then? The usual story is that `AC_CONFIG_HEADER`/`AM_CONFIG_HEADER` would define `HAVE_CONFIG_H` and then application source file should include that file. If `HAVE_CONFIG_H` is not defined, `config.h` is nout available and thus application should not even try including it.

configure.ac
```diff
@@ -20,8 +20,8 @@ AC_CONFIG_FILES([include/odp/api/spec/version.h
                  include/odp/api/spec/deprecated.h])
 
 AM_INIT_AUTOMAKE([1.9 tar-pax subdir-objects])
-AC_CONFIG_SRCDIR([helper/config.h.in])
-AM_CONFIG_HEADER([helper/config.h])
+AC_CONFIG_SRCDIR([include/odp/api/spec/init.h])
```
muvarov: Why is that needed?

muvarov:
why all files have to include that config.h? I think tests and helpers should not know anything how system was configured.

platform/linux-generic/_fdserver.c
muvarov: HAVE_CONFIG_H is always enabled now. Or I do not see a case where it is disabled.

muvarov: https://www.gnu.org/software/autoconf/manual/autoconf-2.66/html_node/acconfig-Header.html as I understand with that link we can add ifdef and endif on top and bottom of generated file.

lumag: @muvarov. Maybe I shoud just drop that `#ifdef` alltogether, leaving just `#include` in place. I've always just used that as an idiom to allow the same source file to be used no matter what. One can enable or disable header file generation, but the source file would still comple. Nevertheless it does not make sense to put `#ifdef` guarding `config.h` inclusion inside config.h. If there is no config.h (because e.g. it was disabled in configure.ac or because Autoconf/Automake semantics change, for whatever reason), such code will compile. Code with #ifndef inside config.h won't/

configure.ac
lumag: Because now there won't be `helper/config.h.in` file and I wanterd to depend on anything less generic than `include/config.h.in`.

lumag:
@muvarov tests and helpers should not depend on main ODP configuration. They should however depend on the system config (presence of libraries and/or header files). Helpers are dependent on `ODPH_HELPER_DEBUG` define. Currently such defines will be passed to all source files via `AM_CFLAGS`. I'm just moving them to more convenient place. Ideally it would be possible to have per-library config file. However Autoconf doesn't support such config. Maybe we should rework ODP into a set of subprojects, each having it's own `configure.ac` and main configure script calling `AC_CONFIG_SUBDIR` for each of subdirs.

lumag:
Updated to drop that `#ifdef`/`#endif` and include `config.h` unconditionally.

test/common_plat/common/mask_common.c
```diff
@@ -4,6 +4,8 @@
  * SPDX-License-Identifier:     BSD-3-Clause
  */
 
+#include "config.h"
```
heyi-linaro: I suppose this config.h will also be included in the distribution package, right?
is that possible to include "config.h" in global odp_api.h?
So applications only need to include the "odp_api.h" and implicitly include the "config.h"

I check the approach of DPDK, which has a similar rte_config.h, included by rte_eal.h, so applications would need to include "rte_config.h" explicitly.

lumag: @heyi-linaro no, this config.h is a part of build system and won't be included into distribution package or installed onto target system.

heyi-linaro: Yes, I see, thanks.

muvarov:
@lumag  going to merge this, but squash similar patches.

lumag:
@muvarov sure, feel free to squash them as you see fit.

muvarov:
@lumag @Bill-Fischofer-Linaro  patches are merged. Merging them also to api-next and ./test/common_plat/performance/odp_crypto fails on pkt_ref. Looks like again double free and enabling debug captures that. It's too late to fix it today.  Merge work is here: https://github.com/muvarov/odp/tree/api-next for that specific issue will look on Monday if there will be no patch in mailing list for that.

Bill-Fischofer-Linaro:
@muvarov Dmirty's PR #143 fixes the duplicate packet free issue. This should probably be merged first.

muvarov:
Yes #143 fixes this. Merged to master and api-next also.

PR #128, opened 2017-08-15
[PATCH CLOUD-DEV v11] linux-dpdk: Enable build & sync with cloud-dev/api-next updates

GBalakrishna:


configure.ac
```diff
@@ -111,6 +111,7 @@ AS_CASE([$host],
   [ARCH_DIR=undefined]
 )
 AC_SUBST([ARCH_DIR])
+AM_CONDITIONAL([ARCH_X86], [test "x${ARCH_DIR}" = "xx86"])
```
brbrooks: Please remove this line and use ARCH_IS_X86 instead.  Maybe move ARCH_IS_XXX lines right here if you think that is better.

configure.ac
```diff
@@ -195,6 +180,27 @@ AC_HELP_STRING([--with-sdk-install-path=DIR path to external libs and headers],
 AC_SUBST(SDK_INSTALL_PATH)
 
 ##########################################################################
+# Run platform specific checks and settings
+##########################################################################
+IMPLEMENTATION_NAME=""
```
brbrooks: Please add a TODO comment about needing to compile with c99 instead of gnu99

platform/linux-dpdk/include/odp/api/feature.h
```diff
@@ -0,0 +1 @@
+../../../../linux-generic/include/odp/api/feature.h
```
Bill-Fischofer-Linaro: These sort of absolute relative path expressions will cause trouble as we start to reorganize the directory structure. In addition to causing checkpatch false positives, an m4 solution to these sort of path specifications would seem to be more flexible in this regard.

configure.ac
GBalakrishna: will fix it in the next version.

configure.ac
GBalakrishna: I thought it was discussed in the cloud call saying we don't need to worry about this flag as it only affects dpdk wrapper.

platform/linux-dpdk/include/odp/api/feature.h
GBalakrishna: These changes can go into separate PR as there are many other files that needs to changes as well.

Bill-Fischofer-Linaro: Agreed. @brbrooks mentioned that he can help in this area. We can discuss this during tomorrow's call. Perhaps open an Issue to track this? In any event, this would be a good starting point for work prior to any directory reorganization.

configure.ac
brbrooks: OK

platform/linux-dpdk/include/odp/api/feature.h
Bill-Fischofer-Linaro: Opened Issue https://github.com/Linaro/odp/issues/130 to track this "to do".

example/ddf_app/Makefile.am
```diff
@@ -1,7 +1,7 @@
 include $(top_srcdir)/example/Makefile.inc
 
 bin_PROGRAMS = odp_ddf_app$(EXEEXT)
-odp_ddf_app_LDFLAGS = $(AM_LDFLAGS) -static -Wl,--export-dynamic
+odp_ddf_app_LDFLAGS = $(AM_LDFLAGS) $(DPDK_LIBS) -static -Wl,--export-dynamic
```
GBalakrishna: pkgconfig won't be included into the test validation, they are two separate things.

Bill-Fischofer-Linaro:
The linux-dpdk test is still failing in Travis. This needs to be resolved before this PR can be merged.

GBalakrishna:
For some reason, travis not running on the latest updates. I am going to close this PR and open a new PR with the same commit message.

GBalakrishna:
This is the new PR create: https://github.com/Linaro/odp/pull/134. And all the validation checks passes except checkpatch.

PR #129, opened 2017-08-17
[PATCH API-NEXT v2] api: scheduler: add odp_schedule_order_unlock_lock function

bala-manoharan:
Adds odp_schedule_order_unlock_lock() API.
This function combines schedule order unlock and lock into a single function for performance optimization in the HW


include/odp/api/spec/schedule.h
```diff
@@ -368,6 +368,26 @@ void odp_schedule_order_lock(unsigned lock_index);
 void odp_schedule_order_unlock(unsigned lock_index);
 
 /**
+ * Release ordered context lock and acquires new lock
+ *
+ * This call is valid only when holding an ordered synchronization context.
+ * Release a previously locked ordered context lock and acquires
+ * a new ordered context lock.
+ * This call is valid only when there is a single ordered context lock and
+ * is invalid if there are nested ordered context lock.
```
Bill-Fischofer-Linaro: Typo: locks, not lock.

Bill-Fischofer-Linaro: You should also clarify what is meant by nested ordered context lock here. As I recall, the restriction is that the caller may only hold a single ordered lock because you didn't want to include the index of the lock to be released as part of this call. So that restriction should be stated explicitly with cautions that results are undefined if it is not observed.

Is the reason for this restriction that your HW doesn't support holding multiple ordered locks concurrently or is there some restriction in the lock switching mechanism that prevents this?

include/odp/api/spec/schedule.h
```diff
@@ -368,6 +368,26 @@ void odp_schedule_order_lock(unsigned lock_index);
...
+ *
+ * @param lock_index	Index of the ordered lock in the current context to be
+ *			acquired. Previously acquired ordered lock is released.
+ *			Results are undefined if there is a nested ordered lock
+ *			in the context. Must be in the range
+ *			0...odp_queue_lock_count() - 1
```
Bill-Fischofer-Linaro: We also have a restriction that the lock being acquired must not have been previously acquired in this ordered context. odp_schedule_order_lock() already includes this caution and it should be echoed here.

include/odp/api/spec/schedule.h
```diff
@@ -368,6 +368,26 @@ void odp_schedule_order_lock(unsigned lock_index);
...
+ *
+ * @see odp_schedule_order_lock(), odp_schedule_order_unlock()
+ *
+ */
+void odp_schedule_order_unlock_lock(uint32_t lock_index);
```
Bill-Fischofer-Linaro: Given that this doesn't take the index value of the lock to be unlocked, would a better name for this be `odp_schedule_order_switch_lock()`?

platform/linux-generic/include/odp_schedule_if.h
```diff
@@ -81,6 +81,7 @@ typedef struct {
 	void (*schedule_resume)(void);
 	void (*schedule_release_atomic)(void);
 	void (*schedule_release_ordered)(void);
+	void (*schedule_release_context)(void);
```
Bill-Fischofer-Linaro: Why release context? You're switching locks here, not releasing the ordered context, no? That's a very different semantic. If you release the ordered context you can no longer acquire any ordered lock associated with that context.

include/odp/api/spec/schedule.h
bala-manoharan: Yes, the restriction is exactly as you have explained.
I will provide more documentation in the next version. The common use-case we have seen involves using ordered lock to regulate the packet flow and it usually does not involve nested atomic locks and also there is a HW limitation in holding multiple ordered locks.

These changes are mainly for performance optimisation since the same behaviour exposed using this API could be achieved by using two different ODP APIs but by having a single API implementation can better optimise this use-case.

include/odp/api/spec/schedule.h
bala-manoharan: Agreed.

include/odp/api/spec/schedule.h
bala-manoharan: I am fine with this naming if there is a consensus.

Bill-Fischofer-Linaro: We can discuss this at Monday's ARCH call.

Bill-Fischofer-Linaro: We also need to update the scheduler validation tests to cover this new API as well as the User Guide documentation covering ordered locks. Once we agree on the API that should be added to this patch or opened as a todo issue.

PR #132, opened 2017-08-18
[PATCH v5] helper: ip: correct ipv4 header checksum calculation

lumag:
Current code for IPv4 header checksum calculation assumes that packet
data is aligned on 2-byte boundary, that there are no optional headers,
etc. Rewrite checksumming code to properly copy & process headers.

### DIFF
--- a/helper/include/odp/helper/chksum.h
+++ b/helper/include/odp/helper/chksum.h
@@ -40,7 +40,7 @@ typedef enum {
  * @param buffer calculate chksum for buffer
  * @param len    buffer length
  *
- * @return checksum value in host cpu order
+ * @return checksum value in network order
  */
 static inline odp_u16sum_t odph_chksum(void *buffer, int len)
 {
--- a/helper/include/odp/helper/ip.h
+++ b/helper/include/odp/helper/ip.h
@@ -74,6 +74,9 @@ extern "C" {
 /** @internal Returns true if IPv4 packet is a fragment */
 #define ODPH_IPV4HDR_IS_FRAGMENT(frag_offset) ((frag_offset) & 0x3fff)
 
+/** @internal Checksum offset in IPv4 header */
+#define ODPH_IPV4HDR_CSUM_OFFSET	10
+
 /** IPv4 header */
 typedef struct ODP_PACKED {
 	uint8_t    ver_ihl;     /**< Version / Header length */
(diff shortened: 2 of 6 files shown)


lumag:
This is a fixup for https://bugs.linaro.org/show_bug.cgi?id=2976

helper/include/odp/helper/ip.h
```diff
@@ -92,6 +95,28 @@ typedef struct ODP_PACKED {
...
+{
+	int nleft = ODPH_IPV4HDR_IHL(ip->ver_ihl) * 4;
+	uint16_t buf[nleft / 2];
+	int res;
+
+	ip->chksum = 0;
+	memcpy(buf, ip, sizeof(*ip));
+	res = odp_packet_copy_to_mem(pkt, offset + sizeof(*ip),
+				     nleft - sizeof(*ip),
+				     buf + sizeof(*ip)/2);
```
Bill-Fischofer-Linaro: Checkpatch flags this. It's a legit issue that should be fixed.

lumag: @Bill-Fischofer-Linaro missed that. Should be fixed now.

helper/include/odp/helper/ip.h
```diff
@@ -132,19 +156,28 @@ static inline int odph_ipv4_csum_valid(odp_packet_t pkt)
  *
  * @return IPv4 checksum in host cpu order, or 0 on failure
  */
-static inline odp_u16sum_t odph_ipv4_csum_update(odp_packet_t pkt)
+static inline void odph_ipv4_csum_update(odp_packet_t pkt)
```
Bill-Fischofer-Linaro: Good catch. The @return is also not correct if this becomes a void function.

lumag:
@muvarov @Bill-Fischofer-Linaro updated PR to include proposed fixes plus few more changes.

helper/include/odp/helper/chksum.h
```diff
@@ -40,16 +40,16 @@ typedef enum {
...
 static inline odp_u16sum_t odph_chksum(void *buffer, int len)
 {
-	uint16_t *buf = (uint16_t *)buffer;
+	odp_u16be_t *buf = (odp_u16be_t *)buffer;
 	uint32_t sum = 0;
 	uint16_t result;
 
 	for (sum = 0; len > 1; len -= 2)
-		sum += *buf++;
+		sum += odp_be_to_cpu_16(*buf++);
```
Bill-Fischofer-Linaro: I'm not sure why this is necessary. Isn't it more efficient to calculate the checksum directly and then return the result in network byte order? The value is going to be the same, it's just how it's being returned.

helper/include/odp/helper/ip.h
```diff
@@ -102,49 +127,55 @@ ODP_STATIC_ASSERT(sizeof(odph_ipv4hdr_t) == ODPH_IPV4HDR_LEN,
...
-		return 0;
+	offset = odp_packet_l3_offset(pkt);
+	if (offset == ODP_PACKET_OFFSET_INVALID)
+		return -1;
+
+	res = odp_packet_copy_to_mem(pkt, offset, sizeof(ip), &ip);
+	if (odp_unlikely(res < 0))
+		return res;
+
+	res = odph_ipv4_csum(pkt, offset, &ip, &chksum);
```
Bill-Fischofer-Linaro: Since you're deleting setting the chksum to 0 elsewhere before calling odph_ipv4_csum_update(), don't you need to set it to zero here before calling odph_ipv4_csum()? This field is supposed to be initialized to zero to calculate the checksum.

helper/include/odp/helper/chksum.h
lumag: Hmm, true. I updated this commit to just change documentation.

helper/include/odp/helper/ip.h
lumag: No. `odph_ipv4_csum()` initialises checksum to 0 in the header copy. This simplifies code for header verification, etc.

Bill-Fischofer-Linaro: OK, I missed that on first read.

muvarov:
Merged.

PR #135, opened 2017-08-18
[PATCH v1] helper: test: fix running process/thread tests on single core systems

matiaselo:
This patch enables running process and thread validation tests on single
core systems.

Fixes bug: https://bugs.linaro.org/show_bug.cgi?id=2812


lumag:

muvarov:
Merged.

PR #138, opened 2017-08-22
[PATCH CLOUD-DEV v1] linux-dpdk: introduce pool modular framework

GBalakrishna:
corrected some checkpatch issues as well

### DIFF
--- a/platform/linux-dpdk/Makefile.am
+++ b/platform/linux-dpdk/Makefile.am
@@ -17,6 +17,7 @@ endif
 AM_CFLAGS +=  $(PLAT_CFLAGS)
 AM_CFLAGS +=  -I$(srcdir)/include
 AM_CFLAGS +=  -I$(top_srcdir)/platform/linux-generic/include
+AM_CFLAGS +=  -I$(top_srcdir)/frameworks/modular
 AM_CFLAGS +=  -I$(top_srcdir)/include/odp/arch/@ARCH_ABI@
 AM_CFLAGS +=  -I$(top_srcdir)/include
 AM_CFLAGS +=  -I$(top_builddir)/include
@@ -196,6 +197,7 @@ noinst_HEADERS = \
 		  ${top_srcdir}/platform/linux-generic/include/odp_packet_socket.h \
 		  ${top_srcdir}/platform/linux-generic/include/odp_pkt_queue_internal.h \
 		  ${srcdir}/include/odp_pool_internal.h \
+		  ${top_srcdir}/platform/linux-generic/include/odp_pool_subsystem.h \
 		  ${srcdir}/include/odp_posix_extensions.h \
 		  ${top_srcdir}/platform/linux-generic/include/odp_queue_internal.h \
 		  ${top_srcdir}/platform/linux-generic/include/odp_queue_if.h \
@@ -243,11 +245,12 @@ __LIB__libodp_dpdk_la_SOURCES = \
 			   ../linux-generic/odp_packet_io.c \
 			   ../linux-generic/pktio/loop.c \
 			   ../linux-generic/odp_pkt_queue.c \
-			   odp_pool.c \
+			   pool/dpdk.c \
 			   ../linux-generic/odp_queue.c \
 			   ../linux-generic/odp_queue_if.c \
 			   ../linux-generic/odp_rwlock.c \
 			   ../linux-generic/odp_rwlock_recursive.c \
+			   ../linux-generic/pool/subsystem.c \
 			   ../linux-generic/odp_schedule.c \
 			   ../linux-generic/odp_schedule_if.c \
 			   ../linux-generic/odp_schedule_iquery.c \
@@ -290,6 +293,8 @@ __LIB__libodp_dpdk_la_SOURCES += arch/x86/cpu_flags.c \
 				 arch/x86/odp_cpu_arch.c \
 				 arch/x86/odp_sysinfo_parse.c
 endif
+
+pool/dpdk.lo: AM_CFLAGS += -DIM_ACTIVE_MODULE
 
 # Build modular framework into odp-linux library
 modularframeworkdir = $(top_srcdir)/frameworks/modular
--- a/platform/linux-dpdk/pool/dpdk.c
+++ b/platform/linux-dpdk/pool/dpdk.c
@@ -48,15 +48,13 @@ typedef struct pool_table_t {
 	odp_shm_t shm;
 } pool_table_t;
 
-
 /* The pool table ptr - resides in shared memory */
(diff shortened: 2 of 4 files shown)


platform/linux-dpdk/Makefile.am
```diff
@@ -196,6 +197,7 @@ noinst_HEADERS = \
 		  ${top_srcdir}/platform/linux-generic/include/odp_packet_socket.h \
 		  ${top_srcdir}/platform/linux-generic/include/odp_pkt_queue_internal.h \
 		  ${srcdir}/include/odp_pool_internal.h \
+		  ${top_srcdir}/platform/linux-generic/include/odp_pool_subsystem.h \
```
brbrooks: Do we need to move the pool subsystem from linux-generic to linux-dpdk?

nagarahalli: Yes, it should be moved out. The current decision is that we will make the directory changes after SFO17. But, the files should be split correctly already. If keeping them where they are currently is resulting in problems, we can move them now.

platform/linux-dpdk/pool/dpdk.c
```diff
@@ -91,12 +90,12 @@ int odp_pool_init_global(void)
 	return 0;
 }
 
-int odp_pool_init_local(void)
+static int dpdk_pool_init_local(void)
```
heyi-linaro: Minor issue: if it does nothing we can remove it and just set the function pointer to NULL.

platform/linux-dpdk/pool/dpdk.c
```diff
@@ -637,18 +638,36 @@ int odp_pool_destroy(odp_pool_t pool_hdl)
...
 	memset(params, 0, sizeof(odp_pool_param_t));
 }
 
-uint64_t odp_pool_to_u64(odp_pool_t hdl)
+static uint64_t dpdk_pool_to_u64(odp_pool_t hdl)
 {
 	return _odp_pri(hdl);
 }
 
+pool_module_t dpdk_pool = {
```
heyi-linaro: can be static pool_module_t dpdk_pool

platform/linux-dpdk/Makefile.am
GBalakrishna: @nagarahalli what do you mean by "splitting already and moving them now", copy the odp_pool_subsystem into linux-dpdk ?

nagarahalli: I meant, the files Kevin ccreated (for ex: odp_pool_subsystem.h) already have the correct code. After the directory structure is created, we do not need to change the code in these files, they just need to be moved to their respective directories.

platform/linux-dpdk/pool/dpdk.c
nagarahalli: I think it is better to keep the empty functions, especially for the data plane functions, so that we do not have to have NULL checks at call sites.

platform/linux-dpdk/Makefile.am
```diff
@@ -196,6 +197,7 @@ noinst_HEADERS = \
 		  ${top_srcdir}/platform/linux-generic/include/odp_packet_socket.h \
 		  ${top_srcdir}/platform/linux-generic/include/odp_pkt_queue_internal.h \
 		  ${srcdir}/include/odp_pool_internal.h \
```
nagarahalli: I am thinking if the file name needs to be changed to indicate that this is a pool implemented using DPDK. Currently, linux-generic mempool also has odp_pool_internal.h (this also should to be renamed to indicate 'generic' pool)

platform/linux-dpdk/pool/dpdk.c
```diff
@@ -595,16 +594,18 @@ void odp_buffer_free_multi(const odp_buffer_t buf[], int num)
 	}
 }
 
-void odp_pool_print(odp_pool_t pool_hdl)
+static void dpdk_pool_print(odp_pool_t pool_hdl)
 {
 	uint32_t pool_id = pool_handle_to_index(pool_hdl);
 	pool_entry_t *pool = get_pool_entry(pool_id);
+
 	rte_mempool_dump(stdout, pool->s.rte_mempool);
```
nagarahalli: The API documentation says the print needs to go to ODP log. Is there a way to correct this?

platform/linux-dpdk/pool/dpdk.c
```diff
@@ -91,12 +90,12 @@ int odp_pool_init_global(void)
 	return 0;
 }
 
-int odp_pool_init_local(void)
+static int dpdk_pool_init_local(void)
 {
```
nagarahalli: I am assuming this function is empty as DPDK already implements per core pool cache and we do not need to implement it ourselves.

GBalakrishna: This is the empty function that you have commented above about making it NULL or just return true all ways. This is needed here to skip compiler issue.

platform/linux-dpdk/Makefile.am
GBalakrishna: currently they are differentiated with platform/linux-*. Unless if we do any directory re-structuring, I assume this is sufficient.

platform/linux-dpdk/pool/dpdk.c
GBalakrishna: I think the only way seems to be is to copy the functionality of rte_mempool_dump to here. In other places we use ODP_DBG when we wanted to print the parts of mempool.

platform/linux-dpdk/pool/dpdk.c
GBalakrishna: @heyi-linaro will fix it in the next version

Bill-Fischofer-Linaro:
As discussed during today's ODP Cloud call, this PR should be good to merge, however we also need to be able to do performance testing to see how this affects performance to help tune the modular framework. Should we apply some tags to the cloud-dev branch to permit easy comparisons of these changes as we make them?

heyi-linaro:
Yes, @Bill-Fischofer-Linaro , @nagarahalli @GBalakrishna I'll merge this PR today.

For tagging can we define the moment as after PR #139 and Makefile re-organization (I'll send today to internal and reach PR next week), so finished the pktio-ops part.

The Makefile re-orgnization also will provide an generic Makefile infrastructure so that new subsystems/modules addition only requires several lines of description instead of writing your own Makefile.am.

proposal tag name: 1st-modular-practices-pool-pktio-ops

After then we continue, there is several works in progress modular buffer @Kevin, queue @Joyce, scheduler @Brian, and linux-dpdk pktio(s) @nagarahalli @GBalakrishna

heyi-linaro:
Merged and closed.

PR #140, opened 2017-08-23
[PATCH CLOUD-DEV v2] linux-generic: discard dpdk pktio

GBalakrishna:
As cloud-dev now has support for dpdk pktio from linux-dpdk,
no need of another dpdk pktio support.


heyi-linaro:
I suggest an approach that add workable solutions first and after then remove obsolete code.

With PR#140 it looks likes that we removed linux-generic dpdk pktio and lost dpdk pktio at all, now wait for a new design which seems still in discussion.

GBalakrishna:
we have the working solution in linux-dpdk.

Bill-Fischofer-Linaro:
Based on discussions at today's ODP Cloud meeting, we should defer this until after the pktio modularization work has reached the point where a dpdk pktio module can be plugged in to replace the integrated dpdk pktio support currently in linux-generic.  Removing this code would otherwise regress functionality until we had the replacement support ready.

heyi-linaro:
@GBalakrishna shall we close this PR and re-create after then?

Bill-Fischofer-Linaro:
That might be easiest since I suspect it will need rework after the pktio modularization is done anyway.

GBalakrishna:
Ya sure will close it.

PR #141, opened 2017-08-23
[PATCH v2] linux-gen: system: correctly set CPU frequency

kelvincheung:
When CPU frequency info under /sys is available,
use the real value instead of default value.

Signed-off-by: Kelvin Cheung <keguang.zhang@gmail.com>

### DIFF
--- a/helper/test/linux/process.c
+++ b/helper/test/linux/process.c
@@ -56,9 +56,10 @@ int main(int argc ODPH_UNUSED, char *argv[] ODPH_UNUSED)
 	cpu = odp_cpumask_first(&cpu_mask);
 	printf("the first CPU:              %i\n", cpu);
 
-	/* reserve cpu 0 for the control plane so remove it from
-	 * the default mask */
-	odp_cpumask_clr(&cpu_mask, 0);
+	/* If possible, remove CPU 0 from the default mask to reserve it for the
+	 * control plane. */
+	if (num_workers > 1)
+		odp_cpumask_clr(&cpu_mask, 0);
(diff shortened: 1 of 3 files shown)


apalos:
Hello Kevin, 
Thanks for the patch.
The "general" approach so far is to read '/proc/cpuinfo', for various info regarding the CPU(model, current speed, max speed estimation etc).

The sysfs approach seems more unified since we can use it for every CPU.
Can you fix the patch in order to read the cpuinfo_max_freq and cpuinfo_cur_freq from sysfs for all available architectures instead of just arm?
Make a generic function, that reads these values and use them on all available architectures seems more consistent. 

Thanks,
Ilias

Bill-Fischofer-Linaro:
Also, Kevin, please be sure to read and sign the ODP License Agreement (linked from this page). This is a standard contribution agreement for open source projects and is needed to be able to accept any patch for inclusion into the ODP code base. Thanks.

kelvincheung:
OK, got it.
PATCH v2 is on the way.

kelvincheung:
Sorry, please ignore PATCH v2.
There will be PATCH v3.
Thanks!

muvarov:
apalos, please review. I would merge buffer[] and path[] to one var to same stack space, but it's not sufficient here.

muvarov:
@kelvincheung hello Kelvin, will you provide v3?

kelvincheung:
@muvarov Sorry for the slow response. I've been busy these days.
Of course I will send the v3 by this weekend, maybe.
Thanks!

Bill-Fischofer-Linaro:
@apalos please review this update.

PR #142, opened 2017-08-23
[MONARCH_LTS PATCH v1] configure.ac: inc library version

muvarov:
I think that the latest thing for monarch is inc of that version.

-----
only source code was changed to inc only middle number.


muvarov:
Merged.

PR #148, opened 2017-08-28
[PATCH API-NEXT v7] api: pool: add packet headroom configuration

bala-manoharan:
Adds packet headroom configuration to pool parameters

### DIFF
--- a/platform/linux-generic/odp_crypto.c
+++ b/platform/linux-generic/odp_crypto.c
@@ -701,14 +701,11 @@ odp_crypto_session_create(odp_crypto_session_param_t *param,
 	odp_crypto_generic_session_t *session;
 	int aes_gcm = 0;
 
-	/* Default to successful result */
-	*status = ODP_CRYPTO_SES_CREATE_ERR_NONE;
-
 	/* Allocate memory for this session */
 	session = alloc_session();
 	if (NULL == session) {
 		*status = ODP_CRYPTO_SES_CREATE_ERR_ENOMEM;
-		return -1;
+		goto err;
 	}
 
 	/* Copy parameters */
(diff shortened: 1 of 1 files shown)


platform/linux-generic/odp_pool.c
```diff
@@ -346,7 +346,10 @@ static odp_pool_t pool_create(const char *name, odp_pool_param_t *params,
 		break;
 
 	case ODP_POOL_PACKET:
-		headroom    = CONFIG_PACKET_HEADROOM;
+		if (params->pkt.headroom < CONFIG_PACKET_HEADROOM)
+			headroom = params->pkt.headroom;
+		else
+			headroom    = CONFIG_PACKET_HEADROOM;
```
Bill-Fischofer-Linaro: Since `CONFIG_PACKET_HEADROOM` is what's currently returned as `min_headroom`, this logic is incorrect.

bala-manoharan: Yes. This logic was based on previous assumption to make headroom parameter as a defined value rather than min. Will change it accordingly.

include/odp/api/spec/pool.h
```diff
@@ -214,6 +214,13 @@ typedef struct odp_pool_param_t {
 			    defined by pool capability pkt.max_uarea_size.
 			    Specify as 0 if no user area is needed. */
 			uint32_t uarea_size;
+
+			/** Headroom size in bytes. Number of bytes of
+			    space preceding packet data to reserve for
+			    use as headroom.
+			    Must not exceed the implementation defined
+			    ODP_PACKET_MAX_HEADROOM. */
```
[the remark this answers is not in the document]
bala-manoharan: Will include these changes in the next version. Maybe we should rename it to default_headroom instead of min_headroom since technically this is value of headroom which the implementation will have by default and min_headroom can be zero.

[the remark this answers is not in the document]
Bill-Fischofer-Linaro: That sounds like a reasonable change.

platform/linux-generic/odp_pool.c
```diff
@@ -346,7 +346,10 @@ static odp_pool_t pool_create(const char *name, odp_pool_param_t *params,
 		break;
 
 	case ODP_POOL_PACKET:
-		headroom    = CONFIG_PACKET_HEADROOM;
+		if (params->pkt.headroom < CONFIG_PACKET_HEADROOM)
+			headroom = params->pkt.headroom;
```
Bill-Fischofer-Linaro: This line is failing many of the Travis checks. Need to recheck your syntax.

PR #152, opened 2017-08-29
[PATCH API-NEXT v11] api: pool: add min and max headroom configuration in pool

bala-manoharan:
api: pool: add min and max headroom configuration in pool

Adds min and max headroom configuration

### DIFF
--- a/include/odp/api/spec/pool.h
+++ b/include/odp/api/spec/pool.h
@@ -97,6 +97,13 @@ typedef struct odp_pool_capability_t {
 		 * packet input and user allocated packets.*/
 		uint32_t min_headroom;
 
+		/** Maximum packet level headroom length in bytes
+		 *
+		 * The maximum value of packet pool headroom parameter
+		 * that can be configured. This value applies to both ODP
+		 * packet input and user allocated packets.*/
+		uint32_t max_headroom;
+
 		/** Minimum packet level tailroom length in bytes
 		 *
 		 * The minimum number of tailroom bytes that newly created
@@ -214,6 +221,13 @@ typedef struct odp_pool_param_t {
 			    defined by pool capability pkt.max_uarea_size.
 			    Specify as 0 if no user area is needed. */
 			uint32_t uarea_size;
+
+			/** Minimum Headroom size in bytes. Each newly allocated
+			    packet from the pool must have at least this much
+			    headroom. The maximum value is defined by pool
+			    capability pkt.max_headroom.
+			    Use zero if headroom is not needed. */
+			uint32_t headroom;
 		} pkt;
 
 		/** Parameters for timeout pools */
--- a/platform/linux-generic/odp_pool.c
+++ b/platform/linux-generic/odp_pool.c
@@ -348,6 +348,10 @@ static odp_pool_t pool_create(const char *name, odp_pool_param_t *params,
 		break;
 
 	case ODP_POOL_PACKET:
+		if (params->pkt.headroom > CONFIG_PACKET_HEADROOM) {
+			ODP_ERR("Packet headroom size not supported.");
+			return ODP_POOL_INVALID;
+		}
 		headroom    = CONFIG_PACKET_HEADROOM;
 		tailroom    = CONFIG_PACKET_TAILROOM;
 		num         = params->pkt.num;
@@ -854,6 +858,7 @@ int odp_pool_capability(odp_pool_capability_t *capa)
 	capa->pkt.max_len          = CONFIG_PACKET_MAX_SEGS * max_seg_len;
 	capa->pkt.max_num	   = CONFIG_POOL_MAX_NUM;
 	capa->pkt.min_headroom     = CONFIG_PACKET_HEADROOM;
+	capa->pkt.max_headroom     = CONFIG_PACKET_HEADROOM;
 	capa->pkt.min_tailroom     = CONFIG_PACKET_TAILROOM;
 	capa->pkt.max_segs_per_pkt = CONFIG_PACKET_MAX_SEGS;
 	capa->pkt.min_seg_len      = max_seg_len;
@@ -913,6 +918,7 @@ odp_pool_t odp_buffer_pool(odp_buffer_t buf)
 void odp_pool_param_init(odp_pool_param_t *params)
 {
 	memset(params, 0, sizeof(odp_pool_param_t));
+	params->pkt.headroom = CONFIG_PACKET_HEADROOM;
 }
 
 uint64_t odp_pool_to_u64(odp_pool_t hdl)


include/odp/api/spec/pool.h
```diff
@@ -214,6 +221,17 @@ typedef struct odp_pool_param_t {
...
+
+			/** Minimum Headroom size in bytes. Minimum number of
+			    bytes of space preceding packet data to reserve for
+			    use as headroom. This value will be used as a hint
+			    and implementation might configure a value of
+			    headroom greater than this value.
+			    Must not exceed the implementation defined
+			    'max_headroom' in odp_pool_capability().
+			    If the value is less than 'min_headroom' defined in
+			    'odp_pool_capability' then it will be ignored.*/
```
Bill-Fischofer-Linaro: If we have defined min and max values, then values outside this range should result in the `odp_pool_create()` call failing, not being silently ignored. Such is the case for other limits so headroom should be no different.

However, as noted previously, I don't think a `min_headroom` is particularly meaningful/useful so this should be reworded. We should definitely include the use case that a `min_headroom` of 0 indicates that the application does not require headroom so the implementation may omit this to conserve storage.

It may also be appropriate to revisit the question of headroom and packet alignment. The current linux-generic pool allocation structure ensures that the IP header will be misaligned due to the presence of a 14-byte Ethernet header, meaning `odp_packet_l3_ptr()` will return an address that is 2-byte rather than 4-byte aligned because we're now aligning `data[offset]` rather than `data[0]`.  I realize this is not directly related to this PR, but since we're making changes in this area something we should make clear.

platform/linux-generic/include/odp_config_internal.h
```diff
@@ -65,6 +65,15 @@
 #define CONFIG_PACKET_HEADROOM 128
 
 /*
+ * Maximum packet headroom
+ *
+ * This defines the maximum number of headroom bytes that newly created packets
+ * can be configured. This value applies to both ODP packet input and user
+ * allocated packets.
+ */
+#define CONFIG_PACKET_MAX_HEADROOM 258
```
Bill-Fischofer-Linaro: Any particular reason for this "magic number"?

platform/linux-generic/odp_pool.c
```diff
@@ -348,7 +348,10 @@ static odp_pool_t pool_create(const char *name, odp_pool_param_t *params,
 		break;
 
 	case ODP_POOL_PACKET:
-		headroom    = CONFIG_PACKET_HEADROOM;
+		if (params->pkt.min_headroom < CONFIG_PACKET_MAX_HEADROOM)
+			headroom = params->pkt.min_headroom;
+		else
+			headroom    = CONFIG_PACKET_MAX_HEADROOM;
```
Bill-Fischofer-Linaro: Trying to specify a headroom > `CONFIG_PACKET_MAX_HEADROOM` should fail the call, as noted earlier.

include/odp/api/spec/pool.h
```diff
@@ -93,10 +93,17 @@ typedef struct odp_pool_capability_t {
...
+		 * packets can be configured. This value applies to both ODP
 		 * packet input and user allocated packets.*/
 		uint32_t min_headroom;
 
+		/** Maximum packet level headroom length in bytes
+		 *
+		 * The maximum number of headroom bytes that newly created
+		 * packets can be configured. This value applies to both ODP
+		 * packet input and user allocated packets.*/
+		uint32_t max_headroom;
```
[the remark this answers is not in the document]
bala-manoharan: The maximum headroom supported will be equal to the maximum segment size supported by the platform and we can discuss in the ARCH call and if required we can indicate either zero or UINT32_MAX as value when the implementation does not have any limitation.

platform/linux-generic/include/odp_config_internal.h
bala-manoharan: We had a requirement from a customer for 256B headroom and I am not aware of any request bigger than this hence this magic number.

Bill-Fischofer-Linaro: Ok, but odp-linux really has no built-in maximum for this so we shouldn't impose one artificially. The effective max_headroom would be something like `CONFIG_PACKET_MAX_SEG_LEN - 1`.

include/odp/api/spec/pool.h
```diff
@@ -93,10 +93,17 @@ typedef struct odp_pool_capability_t {
 		/** Minimum packet level headroom length in bytes
 		 *
 		 * The minimum number of headroom bytes that newly created
-		 * packets have by default. The default apply to both ODP
+		 * packets can be configured. This value applies to both ODP
 		 * packet input and user allocated packets.*/
 		uint32_t min_headroom;
```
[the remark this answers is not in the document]
Bill-Fischofer-Linaro: Just to elaborate, it's always valid for an application to request 0 headroom. Whether that results in packets having 0 headroom is up to the implementation since implementations are free to round up application requests to something larger. So I agree with Petri that this particular wording change probably isn't needed. It simply says that in the absence of any specific requests from the application, this is the minimum amount of headroom that applications can expect to find in every newly received or allocated packet.

platform/linux-generic/include/odp_config_internal.h
bala-manoharan: Okay

include/odp/api/spec/pool.h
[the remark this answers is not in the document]
bala-manoharan: In that case the naming "default_headroom" makes more sense than "min_headroom". Maybe we can leave it as "min_headroom" in this release.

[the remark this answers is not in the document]
psavol: We should change this field since that's not required for the new feature to be functional.

include/odp/api/spec/pool.h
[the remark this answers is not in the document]
psavol: This should refer to the new headroom pool parameter: "This is the maximum value for packet pool headroom param ..." So, it's the maximum application may ask, implementation may still round up from there.

include/odp/api/spec/pool.h
psavol: This is not hint, but minimum. Implementation may round up, but must not round down. When implementation has fixed headroom, it should set max capability to that.

Simply:
"Minimum headroom size in bytes. Every newly allocated packet must have at least this much headroom. Must not exceed max_headroom packet pool capability."

platform/linux-generic/odp_pool.c
psavol: It's better not to change headroom implementation in this set but just implement the capability == return the current headroom as the max. Other parts of the code assume that init time headroom is constant:

return CONFIG_PACKET_HEADROOM + (head - base);

include/odp/api/spec/pool.h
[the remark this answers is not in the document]
bala-manoharan: Okay.

include/odp/api/spec/pool.h
bala-manoharan: Okay

platform/linux-generic/odp_pool.c
Bill-Fischofer-Linaro: @psavol I'm not sure what you're saying here. This PR has two commits: One for the API changes and one to change the odp-linux pool implementation to reflect these new API changes. Are you suggesting that we ignore the API changes?

bala-manoharan: @psavol CONFIG_PACKET_HEADROOM is set as default in odp_pool_params_init() function. So it should not break any existing implementation.

Also do you mean to set CONFIG_PACKET_HEADROOM as the maximum headroom in capability.

bala-manoharan:
@Bill-Fischofer-Linaro @psavol This version v3 incorporates all the previous review comments

include/odp/api/spec/pool.h
```diff
@@ -214,6 +222,15 @@ typedef struct odp_pool_param_t {
 			    defined by pool capability pkt.max_uarea_size.
 			    Specify as 0 if no user area is needed. */
 			uint32_t uarea_size;
+
+			/** Minimum Headroom size in bytes. Minimum number of
+			    bytes of space preceding packet data to reserve for
+			    use as headroom. Each newly allocated packet from
+			    the pool must have atleast this much headroom.
```
Bill-Fischofer-Linaro: Typo: at least

platform/linux-generic/odp_pool.c
```diff
@@ -348,6 +348,11 @@ static odp_pool_t pool_create(const char *name, odp_pool_param_t *params,
 		break;
 
 	case ODP_POOL_PACKET:
+		if (params->pkt.min_headroom > CONFIG_PACKET_HEADROOM) {
+			ODP_ERR("Headroom size not supported");
+			return ODP_POOL_INVALID;
+		}
+
 		headroom    = CONFIG_PACKET_HEADROOM;
```
Bill-Fischofer-Linaro: It's not clear why we're making this change if we don't intend to do anything with it. Shouldn't this be:
`headroom = params->pkt.min_headroom;` ? 

To make this convention work correctly, `odp_pool_param_init()` needs to be updated to set `pkt.min_headroom` to `CONFIG_PACKET_HEADROOM` so that existing applications behave unchanged, which you've already done. So this looks like an omission?

include/odp/api/spec/pool.h
```diff
@@ -214,6 +222,15 @@ typedef struct odp_pool_param_t {
...
+			    Must not exceed the implementation defined
+			    'max_headroom' in odp_pool_capability().
+			    Implementation may round up to a higher value. */
+			uint32_t min_headroom;
```
Bill-Fischofer-Linaro: I'd still like to see a usage note somewhere indicating that setting `min_headroom` to 0 means that the application does not require any packet headroom so the implementation may optimize pool storage usage accordingly.

platform/linux-generic/odp_pool.c
bala-manoharan: Since we are now exposing max_headroom we need to fail the pool creating if the headroom size requested could not be supported.

pkt.min_headroom is odp_pool_param_init() is set to CONFIG_PACKET_HEADROOM already. Its is the second part of this patch.

include/odp/api/spec/pool.h
bala-manoharan: Okay

platform/linux-generic/odp_pool.c
bala-manoharan: setting headroom = CONFIG_PACKET_HEADROOM does not violate the API requirement and it also fits well with our linux-generic implementation.

JannePeltonen:
How is this patch set related to odp_pktio_headroom_set()? I suppose that the function may fail if the requested headroom is larger than the min_headroom of the associated pool? Does the API need some comment about it or is it obvious?

It seems that in the current implementation odp_pktio_headroom_set() always succeeds, even if the requested headroom is larger than the min_headroom of the relevant pool. Is that ok?

Btw, is odp_pktio_headroom_set() supposed to ensure that packets received from that pktio have exactly that much headroom, no more, no less, or can there be more? If the latter, an API update might be needed.

bala-manoharan:
@JannePeltonen 
Yes. we need to update odp_pktio_set_headroom() function accordingly. I believe the current description of odp_pktio_set_headroom specifies that the function will fail is the requested value is greater than ODP_PACKET_MAX_HEADROOM which should be change to reflect the capability.
I was planning to send a patch to update odp_pktio_set_headroom(), I was hoping it could be a different patch from this one.

And yes the the value of headroom reserved by the implementation can be same or greater than the value requested by the application.

platform/linux-generic/odp_pool.c
Bill-Fischofer-Linaro: True, but I don't see why we wouldn't want to honor the application's requested pool size. Setting this unconditionally this way is just ignoring the request and if we're going to do that there's no reason to have this PR at all.

Bill-Fischofer-Linaro:
@bala-manoharan @JannePeltonen In odp-linux today `odp_pktio_set_headroom()` seems to be a no-op. The code is in `odp_classification.c`:
```
int odp_pktio_headroom_set(odp_pktio_t pktio_in, uint32_t headroom)
{
	pktio_entry_t *entry = get_pktio_entry(pktio_in);

	if (entry == NULL) {
		ODP_ERR("Invalid odp_pktio_t handle");
		return -1;
	}
	entry->s.cls.headroom = headroom;
	return 0;
}
```
However I don't see anyplace that references `entry->s.cls.headroom` so it's not clear anyone does anything with this value.

bala-manoharan:
@Bill-Fischofer-Linaro Yes. the headroom setting in pktio has to be modified. This feature is useful for HW which can optimize the memory usage based on the headroom requirement from the pool.
The current api spec definition does not contradict the linux-generic implementation.

I would propose to change the linux-generic after API finalisation.

platform/linux-generic/odp_pool.c
bala-manoharan: This PR helps implementation which have the ability to maintain zero headroom size and our linux-generic implementation does not violate the API spec. We can change linux-generic in future also.

Bill-Fischofer-Linaro: I see no reason not to change it now, since linux-generic can support zero headroom if that's what the application requests.  Besides, this is supposed to be a reference implementation which shows how we intend APIs to behave. We intend lower headroom requests to be honored, so we should do so here.

include/odp/api/spec/pool.h
```diff
@@ -214,6 +222,18 @@ typedef struct odp_pool_param_t {
...
 			    Specify as 0 if no user area is needed. */
 			uint32_t uarea_size;
+
+			/** Minimum Headroom size in bytes. Minimum number of
+			    bytes of space preceding packet data to reserve for
+			    use as headroom. Each newly allocated packet from
+			    the pool must have at least this much headroom.
+			    Must not exceed the implementation defined
+			    'max_headroom' in odp_pool_capability().
+			    @note if application does not required headroom it
```
Bill-Fischofer-Linaro: Typo: If (should be capitalized) and require, not required.

platform/linux-generic/odp_pool.c
```diff
@@ -348,6 +348,10 @@ static odp_pool_t pool_create(const char *name, odp_pool_param_t *params,
 		break;
 
 	case ODP_POOL_PACKET:
+		if (params->pkt.min_headroom > CONFIG_PACKET_HEADROOM) {
+			ODP_ERR("Packet headroom size not supported.");
+			return ODP_POOL_INVALID;
+		}
 		headroom    = CONFIG_PACKET_HEADROOM;
```
Bill-Fischofer-Linaro: To change this to:

```
headroom = params->pkt.min_headroom;
```

To reflect the API change being proposed in this PR all that's needed is a one line update to the `packet_init()` routine in `odp_packet_internal.h`:

Change 

```
	pkt_hdr->headroom  = CONFIG_PACKET_HEADROOM;
```
to 

```
	pkt_hdr->headroom  = pkt_hdr->buf_hdr.pool_ptr->headroom;
```

However since `pool_ptr` is currently defined as `void *` a few more simple infrastructure changes are needed to undo this "shortcut".

1. `pool_ptr` should be of type `pool_t *`, not `void *` in `odp_buffer_internal.h`

2. `odp_buffer_internal.h` needs to `#include <odp_pool_internal.h>` to get the definition of `pool_t`

3.  Since there is now a circular dependency between `odp_buffer_internal.h` and `odp_pool_internal.h` the `odp_forward_typedefs_internal.h` file has to have the added line: `typedef struct pool_t pool_t;` added to it.

With these changes pools will properly use the headroom requested in by the new `min_headroom` field, as intended.

However, as @psavol is in the process or reworking the packet internals again I'll leave it to him as to whether he'd rather see this change before or after he makes his other changes. An alternate approach would be for pools to define an internal inline API that returns the configured headroom for this pool so `packet_init()` doesn't have to rummage for it.

My point remains is that if we're creating a reference implementation that reference implementation should reflect how we intend the APIs to be used. If we're now going to allow applications to specify headroom size on `odp_pool_create()` then odp-linux should honor those requests.

include/odp/api/spec/pool.h
```diff
@@ -214,6 +221,17 @@ typedef struct odp_pool_param_t {
...
+			/** Minimum Headroom size in bytes. Minimum number of
+			    bytes of space preceding packet data to reserve for
+			    use as headroom. Each newly allocated packet from
+			    the pool must have at least this much headroom.
+			    Must not exceed the implementation defined
+			    'max_headroom' in odp_pool_capability().
+			    @note If application does not require headroom it
+			    can configure 'min_headroom' as zero and
+			    implementation may or may not allocate headroom. */
+			uint32_t min_headroom;
```
psavol: The parameter name can be just "headroom".

Text can be compressed a bit and made inline with documentation of the other params:

"Minimum headroom size in bytes. Each newly allocated 
packet from the pool will have at least this much headroom. 
The maximum value is defined by pool capability pkt.max_headroom.
Use zero if headroom is not needed."

platform/linux-generic/odp_pool.c
psavol: Bala's implementation is correct. Currently, it's much more simpler to always use the default headroom. I'd rather optimize odp-linux for simplicity and performance than for saving memory. When memory usage is a worry, user can rebuild with smaller CONFIG_PACKET_HEADROOM. Also due to data align / cacheline round ups there will be always some headroom - so the potential memory saving in question would be 1 or 2 cachelines per packet, which is not much.

platform/linux-generic/odp_pool.c
```diff
@@ -853,7 +857,8 @@ int odp_pool_capability(odp_pool_capability_t *capa)
 	capa->pkt.max_pools        = ODP_CONFIG_POOLS;
 	capa->pkt.max_len          = CONFIG_PACKET_MAX_SEGS * max_seg_len;
 	capa->pkt.max_num	   = CONFIG_POOL_MAX_NUM;
-	capa->pkt.min_headroom     = CONFIG_PACKET_HEADROOM;
+	capa->pkt.min_headroom     = 0;
```
psavol: Actually, min_headroom is still CONFIG_PACKET_HEADROOM since we always round up to that.

include/odp/api/spec/pool.h
bala-manoharan: I prefer naming it as min_headroom is coz the value given by application is not directly configured and implementation may roundup the value and hence the name "min_headroom"

platform/linux-generic/odp_pool.c
Bill-Fischofer-Linaro: Given Petri's other comments, I agree here. Sounds like we just want to make these additional params available in the spec but ignore them in odp-linux. We can revisit this in the cloud-dev branch.

include/odp/api/spec/pool.h
psavol: Sorry, forgot to mention the rationale for "headroom" vs "min_headroom". Plain "headroom" is inline with the other parameter names in the struct (num/len/max_len/seg_len/uarea_size). Those describe in documentation text how implementation may round up things, but the variable names are not min_len, min_seg_len, min_uarea_size, etc 

For example, seg_len spec is very close to headroom spec. Application requires at least 'seg_len' bytes but implementation is free to round up:
"Minimum number of packet data bytes that are stored
in the first segment of a packet. The maximum value
is defined by pool capability pkt.max_seg_len.
Use 0 for default. "
uint32_t seg_len;

So, please rename to "headroom".

bala-manoharan:
@psavol V10 incorporates all your review comments

muvarov:
@bala-manoharan Trvis failed for v10

bala-manoharan:
@muvarov @Bill-Fischofer-Linaro This V11 fixes the Travis issue.
Can this version be merged?

muvarov:
Merged.

PR #154, opened 2017-08-30
[PATCH v1] helper: chksum: add few guarding conditions

lumag:
Add few guarding conditions, otherwise bad packet can hang up ODP code.
Noted by one of the corner case tests for IPsec.


muvarov:
Merged.

PR #153, opened 2017-08-30
[PATCH v1] Travis test all

Bill-Fischofer-Linaro:
Restructure the .travis.yml file and add additional tests for --enable-debug-print and
--enable-helper-linux to ensure all code paths are tested.

PR #155, opened 2017-08-30
[PATCH v4] Rework configure options handling

lumag:
Rework defaults for configure options:
 - Enable all tests by default
 - Provide `--enable-debug=full` shorcut to enable all debugging options.

Also drop now-default options from `.travis.yml`. This incorporates @Bill-Fischofer-Linaro 's changes from #153


muvarov:
Merged.

PR #158, opened 2017-09-01
[PATCH v2] linux-gen: socket: use segment data pointers in _tx_pkt_to_iovec()

matiaselo:
A packet may not have valid L2 offset set, so odp_packet_l2_offset() cannot
be used when transmitting packets.

### DIFF
--- a/platform/linux-generic/pktio/socket.c
+++ b/platform/linux-generic/pktio/socket.c
@@ -697,7 +697,7 @@ static uint32_t _tx_pkt_to_iovec(odp_packet_t pkt,
 				 struct iovec iovecs[MAX_SEGS])
 {
 	uint32_t pkt_len = odp_packet_len(pkt);
-	uint32_t offset = odp_packet_l2_offset(pkt);
+	uint32_t offset = 0;
 	uint32_t iov_count = 0;
 
 	while (offset < pkt_len) {


platform/linux-generic/pktio/socket.c
```diff
@@ -696,19 +696,23 @@ static int sock_mmsg_recv(pktio_entry_t *pktio_entry, int index ODP_UNUSED,
 static uint32_t _tx_pkt_to_iovec(odp_packet_t pkt,
 				 struct iovec iovecs[MAX_SEGS])
 {
-	uint32_t pkt_len = odp_packet_len(pkt);
-	uint32_t offset = odp_packet_l2_offset(pkt);
```
Bill-Fischofer-Linaro: A simpler fix would seem to be to just initialize `offset` to 0 here, which is by definition the start of the packet regardless of its structure. It's also more efficient than the segment-based APIs, at least in linux-generic.

matiaselo: Very true, thanks. Will send V2.

platform/linux-generic/pktio/socket.c
```diff
@@ -697,7 +697,7 @@ static uint32_t _tx_pkt_to_iovec(odp_packet_t pkt,
 				 struct iovec iovecs[MAX_SEGS])
 {
 	uint32_t pkt_len = odp_packet_len(pkt);
-	uint32_t offset = odp_packet_l2_offset(pkt);
+	uint32_t offset = 0;
 	uint32_t iov_count = 0;
 
 	while (offset < pkt_len) {
```
muvarov: that "if" looks like not needed if you set offset to 0.

Bill-Fischofer-Linaro: @muvarov Not sure what you're referring to here. There is no if statement is this level of the PR. It's a one-line change to the existing code.

matiaselo: Me neither, there is no if statements in this whole function.

muvarov: please skip my comment, it's not relevant to here.

muvarov:
Merged.

PR #160, opened 2017-09-05
[PATCH API-NEXT v14] api: add schedule order unlock lock 

bala-manoharan:
Adds odp_schedule_order_unlock_lock() api.
This API combines schedule order unlock and lock into a single api for HW performance optimisation.

### DIFF
--- a/include/odp/api/spec/queue.h
+++ b/include/odp/api/spec/queue.h
@@ -106,7 +106,7 @@ typedef struct odp_queue_capability_t {
 	uint32_t max_queues;
 
 	/** Maximum number of ordered locks per queue */
-	unsigned max_ordered_locks;
+	uint32_t max_ordered_locks;
 
 	/** Maximum number of scheduling groups */
 	unsigned max_sched_groups;
@@ -393,10 +393,11 @@ odp_schedule_group_t odp_queue_sched_group(odp_queue_t queue);
  *
  * @param queue   Queue handle
  *
- * @return Number of ordered locks associated with this ordered queue
- * @retval <0 Specified queue is not ordered
+ * @return	Number of ordered locks associated with this ordered queue
+ * @retval 0	Specified queue is not ordered or no ordered lock associated
+ *		with the ordered queue.
  */
-int odp_queue_lock_count(odp_queue_t queue);
+uint32_t odp_queue_lock_count(odp_queue_t queue);
 
 /**
  * Get printable value for an odp_queue_t
--- a/include/odp/api/spec/schedule.h
+++ b/include/odp/api/spec/schedule.h
@@ -371,6 +371,27 @@ void odp_schedule_order_lock(uint32_t lock_index);
 void odp_schedule_order_unlock(uint32_t lock_index);
 
 /**
+ * Release existing ordered context lock and acquire a new lock
+ *
+ * This call is valid only when holding an ordered synchronization context.
+ * Release a previously locked ordered context lock and acquire a new ordered
+ * context lock. The operation is equivalent to application calling first
+ * odp_schedule_order_unlock(unlock_index) and then
+ * odp_schedule_order_lock(lock_index). The same constraints apply with this
+ * call as with those two.
+ *
+ * @param unlock_index	Index of the acquired ordered lock in the current
+ *			context to be released.
+ * @param lock_index	Index of the ordered lock in the current context to be
+ *			acquired. Must be in the range
+ *			0...odp_queue_lock_count() - 1.
+ *
+ * @see odp_schedule_order_lock(), odp_schedule_order_unlock()
+ *
+ */
+void odp_schedule_order_unlock_lock(uint32_t unlock_index, uint32_t lock_index);
+
+/**
  * @}
  */
 
--- a/include/odp/api/spec/schedule.h
+++ b/include/odp/api/spec/schedule.h
@@ -347,12 +347,15 @@ int odp_schedule_group_info(odp_schedule_group_t group,
  * be protected by its own ordered lock. This promotes maximum parallelism by
  * allowing order to maintained on a more granular basis. If an ordered lock
  * is used multiple times in the same ordered context results are undefined.
+ * Only one ordered lock can be active in an ordered context at any given time.
+ * Results are undefined when multiple ordered locks are acquired in nested
+ * fashion within the same ordered context.
  *
  * @param lock_index Index of the ordered lock in the current context to be
  *                   acquired. Must be in the range 0..odp_queue_lock_count()
  *                   - 1
  */
-void odp_schedule_order_lock(unsigned lock_index);
+void odp_schedule_order_lock(uint32_t lock_index);
 
 /**
  * Release ordered context lock
@@ -365,7 +368,7 @@ void odp_schedule_order_lock(unsigned lock_index);
  *                   hold this lock. Must be in the range
  *                   0..odp_queue_lock_count() - 1
  */
-void odp_schedule_order_unlock(unsigned lock_index);
+void odp_schedule_order_unlock(uint32_t lock_index);
 
 /**
  * @}
--- a/platform/linux-generic/include/odp_schedule_if.h
+++ b/platform/linux-generic/include/odp_schedule_if.h
@@ -34,7 +34,7 @@ typedef int (*schedule_init_local_fn_t)(void);
 typedef int (*schedule_term_local_fn_t)(void);
 typedef void (*schedule_order_lock_fn_t)(void);
 typedef void (*schedule_order_unlock_fn_t)(void);
-typedef unsigned (*schedule_max_ordered_locks_fn_t)(void);
+typedef uint32_t (*schedule_max_ordered_locks_fn_t)(void);
 typedef void (*schedule_save_context_fn_t)(uint32_t queue_index);
 
 typedef struct schedule_fn_t {
@@ -93,8 +93,8 @@ typedef struct {
 	int (*schedule_group_thrmask)(odp_schedule_group_t, odp_thrmask_t *);
 	int (*schedule_group_info)(odp_schedule_group_t,
 				   odp_schedule_group_info_t *);
-	void (*schedule_order_lock)(unsigned);
-	void (*schedule_order_unlock)(unsigned);
+	void (*schedule_order_lock)(uint32_t);
+	void (*schedule_order_unlock)(uint32_t);
 
 } schedule_api_t;
 
--- a/platform/linux-generic/include/odp_schedule_if.h
+++ b/platform/linux-generic/include/odp_schedule_if.h
@@ -34,6 +34,7 @@ typedef int (*schedule_init_local_fn_t)(void);
 typedef int (*schedule_term_local_fn_t)(void);
 typedef void (*schedule_order_lock_fn_t)(void);
 typedef void (*schedule_order_unlock_fn_t)(void);
+typedef void (*schedule_order_unlock_lock_fn_t)(void);
 typedef uint32_t (*schedule_max_ordered_locks_fn_t)(void);
 typedef void (*schedule_save_context_fn_t)(uint32_t queue_index);
 
@@ -53,6 +54,7 @@ typedef struct schedule_fn_t {
 	schedule_term_local_fn_t    term_local;
 	schedule_order_lock_fn_t    order_lock;
 	schedule_order_unlock_fn_t  order_unlock;
+	schedule_order_unlock_lock_fn_t  order_unlock_lock;
 	schedule_max_ordered_locks_fn_t max_ordered_locks;
 
 	/* Called only when status_sync is set */
@@ -95,6 +97,7 @@ typedef struct {
 				   odp_schedule_group_info_t *);
 	void (*schedule_order_lock)(uint32_t);
 	void (*schedule_order_unlock)(uint32_t);
+	void (*schedule_order_unlock_lock)(uint32_t, uint32_t);
 
 } schedule_api_t;
 
--- a/platform/linux-generic/include/odp_schedule_scalable_ordered.h
+++ b/platform/linux-generic/include/odp_schedule_scalable_ordered.h
@@ -79,7 +79,7 @@ typedef struct reorder_window {
 	uint32_t tail;
 	uint32_t turn;
 	uint32_t olock[CONFIG_QUEUE_MAX_ORD_LOCKS];
-	uint16_t lock_count;
+	uint32_t lock_count;
 	/* Reorder contexts in this window */
 	reorder_context_t *ring[RWIN_SIZE];
 } reorder_window_t;
--- a/platform/linux-generic/odp_queue.c
+++ b/platform/linux-generic/odp_queue.c
@@ -175,12 +175,12 @@ static odp_schedule_group_t queue_sched_group(odp_queue_t handle)
 	return handle_to_qentry(handle)->s.param.sched.group;
 }
 
-static int queue_lock_count(odp_queue_t handle)
+static uint32_t queue_lock_count(odp_queue_t handle)
 {
 	queue_entry_t *queue = handle_to_qentry(handle);
 
 	return queue->s.param.sched.sync == ODP_SCHED_SYNC_ORDERED ?
-		(int)queue->s.param.sched.lock_count : -1;
+		queue->s.param.sched.lock_count : 0;
 }
 
 static odp_queue_t queue_create(const char *name,
--- a/platform/linux-generic/odp_queue_scalable.c
+++ b/platform/linux-generic/odp_queue_scalable.c
@@ -333,12 +333,12 @@ static odp_schedule_group_t queue_sched_group(odp_queue_t handle)
 	return qentry_from_int(queue_from_ext(handle))->s.param.sched.group;
 }
 
-static int queue_lock_count(odp_queue_t handle)
+static uint32_t queue_lock_count(odp_queue_t handle)
 {
 	queue_entry_t *queue = qentry_from_int(queue_from_ext(handle));
 
 	return queue->s.param.sched.sync == ODP_SCHED_SYNC_ORDERED ?
-		(int)queue->s.param.sched.lock_count : -1;
+		queue->s.param.sched.lock_count : 0;
 }
 
 static odp_queue_t queue_create(const char *name,
--- a/platform/linux-generic/odp_schedule.c
+++ b/platform/linux-generic/odp_schedule.c
@@ -250,7 +250,7 @@ typedef struct {
 		int         prio;
 		int         queue_per_prio;
 		int         sync;
-		unsigned    order_lock_count;
+		uint32_t    order_lock_count;
 	} queue[ODP_CONFIG_QUEUES];
 
 	struct {
@@ -465,7 +465,7 @@ static inline int grp_update_tbl(void)
 	return num;
 }
 
-static unsigned schedule_max_ordered_locks(void)
+static uint32_t schedule_max_ordered_locks(void)
 {
 	return CONFIG_QUEUE_MAX_ORD_LOCKS;
 }
@@ -699,7 +699,7 @@ static inline void ordered_stash_release(void)
 static inline void release_ordered(void)
 {
 	uint32_t qi;
-	unsigned i;
+	uint32_t i;
 
 	qi = sched_local.ordered.src_queue;
 
@@ -1100,7 +1100,7 @@ static void order_unlock(void)
 {
 }
 
-static void schedule_order_lock(unsigned lock_index)
+static void schedule_order_lock(uint32_t lock_index)
 {
 	odp_atomic_u64_t *ord_lock;
 	uint32_t queue_index;
@@ -1127,7 +1127,7 @@ static void schedule_order_lock(unsigned lock_index)
 	}
 }
 
-static void schedule_order_unlock(unsigned lock_index)
+static void schedule_order_unlock(uint32_t lock_index)
 {
 	odp_atomic_u64_t *ord_lock;
 	uint32_t queue_index;
--- a/platform/linux-generic/odp_schedule.c
+++ b/platform/linux-generic/odp_schedule.c
@@ -1144,6 +1144,13 @@ static void schedule_order_unlock(uint32_t lock_index)
 	odp_atomic_store_rel_u64(ord_lock, sched_local.ordered.ctx + 1);
 }
 
+static void schedule_order_unlock_lock(uint32_t unlock_index,
+				       uint32_t lock_index)
+{
+	schedule_order_unlock(unlock_index);
+	schedule_order_lock(lock_index);
+}
+
 static void schedule_pause(void)
 {
 	sched_local.pause = 1;
@@ -1429,5 +1436,6 @@ const schedule_api_t schedule_default_api = {
 	.schedule_group_thrmask   = schedule_group_thrmask,
 	.schedule_group_info      = schedule_group_info,
 	.schedule_order_lock      = schedule_order_lock,
-	.schedule_order_unlock    = schedule_order_unlock
+	.schedule_order_unlock    = schedule_order_unlock,
+	.schedule_order_unlock_lock    = schedule_order_unlock_lock
 };
--- a/platform/linux-generic/odp_schedule_iquery.c
+++ b/platform/linux-generic/odp_schedule_iquery.c
@@ -1296,6 +1296,13 @@ static void schedule_order_unlock(uint32_t lock_index)
 	odp_atomic_store_rel_u64(ord_lock, thread_local.ordered.ctx + 1);
 }
 
+static void schedule_order_unlock_lock(uint32_t unlock_index,
+				       uint32_t lock_index)
+{
+	schedule_order_unlock(unlock_index);
+	schedule_order_lock(lock_index);
+}
+
 static uint32_t schedule_max_ordered_locks(void)
 {
 	return CONFIG_QUEUE_MAX_ORD_LOCKS;
@@ -1368,7 +1375,8 @@ const schedule_api_t schedule_iquery_api = {
 	.schedule_group_thrmask   = schedule_group_thrmask,
 	.schedule_group_info      = schedule_group_info,
 	.schedule_order_lock      = schedule_order_lock,
-	.schedule_order_unlock    = schedule_order_unlock
+	.schedule_order_unlock    = schedule_order_unlock,
+	.schedule_order_unlock_lock    = schedule_order_unlock_lock
 };
 
 static void thread_set_interest(sched_thread_local_t *thread,
--- a/platform/linux-generic/odp_schedule_iquery.c
+++ b/platform/linux-generic/odp_schedule_iquery.c
@@ -1150,7 +1150,7 @@ static inline void ordered_stash_release(void)
 static inline void release_ordered(void)
 {
 	uint32_t qi;
-	unsigned i;
+	uint32_t i;
 
 	qi = thread_local.ordered.src_queue;
 
@@ -1252,7 +1252,7 @@ static void order_unlock(void)
 {
 }
 
-static void schedule_order_lock(unsigned lock_index)
+static void schedule_order_lock(uint32_t lock_index)
 {
 	odp_atomic_u64_t *ord_lock;
 	uint32_t queue_index;
@@ -1279,7 +1279,7 @@ static void schedule_order_lock(unsigned lock_index)
 	}
 }
 
-static void schedule_order_unlock(unsigned lock_index)
+static void schedule_order_unlock(uint32_t lock_index)
 {
 	odp_atomic_u64_t *ord_lock;
 	uint32_t queue_index;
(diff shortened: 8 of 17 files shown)


platform/linux-generic/odp_schedule.c
```diff
@@ -1138,12 +1144,28 @@ static void schedule_order_unlock(unsigned lock_index)
 		   lock_index <= sched->queue[queue_index].order_lock_count);
 
 	ord_lock = &sched->order[queue_index].lock[lock_index];
+	lock_acquired = &sched->order[queue_index].lock_acquired;
```
Bill-Fischofer-Linaro: Unnecessary

platform/linux-generic/odp_schedule.c
```diff
@@ -1138,12 +1144,28 @@ static void schedule_order_unlock(unsigned lock_index)
...
 
 	ODP_ASSERT(sched_local.ordered.ctx == odp_atomic_load_u64(ord_lock));
 
+	odp_atomic_store_rel_u64(lock_acquired,
+				 sched->queue[queue_index].
+				 order_lock_count + 1);
```
Bill-Fischofer-Linaro: Unnecessary

platform/linux-generic/odp_schedule.c
```diff
@@ -1138,12 +1144,28 @@ static void schedule_order_unlock(unsigned lock_index)
...
 	odp_atomic_store_rel_u64(ord_lock, sched_local.ordered.ctx + 1);
 }
 
+static void schedule_order_unlock_lock(unsigned lock_index)
+{
+	odp_atomic_u64_t *lock_acquired;
+	uint32_t queue_index;
+
+	queue_index = sched_local.ordered.src_queue;
+
```
Bill-Fischofer-Linaro: Add here:
```
 	ODP_ASSERT(queue_index != NULL_INDEX &&
 		   lock_index <= sched->queue[queue_index].order_lock_count &&
  		   !sched_local.ordered.lock_called.u8[lock_index] &&
                   sched_local.ordered.lock_called.u8[sched_local.ordered.lock_index]);
```

platform/linux-generic/odp_schedule.c
```diff
@@ -1138,12 +1144,28 @@ static void schedule_order_unlock(unsigned lock_index)
...
+	lock_acquired = &sched->order[queue_index].lock_acquired;
+	schedule_order_unlock(odp_atomic_load_u64(lock_acquired));
```
Bill-Fischofer-Linaro: Replace above two lines with:
```
        schedule_order_unlock(sched_local.ordered.lock_index);
```

platform/linux-generic/odp_schedule_iquery.c
```diff
@@ -1255,6 +1257,7 @@ static void order_unlock(void)
 static void schedule_order_lock(unsigned lock_index)
 {
 	odp_atomic_u64_t *ord_lock;
+	odp_atomic_u64_t *lock_acquired;
```
Bill-Fischofer-Linaro: I suspect same comments for this patch apply equally to all of the scheduler variants. Not repeating individual review comments here.

platform/linux-generic/odp_schedule_scalable.c
```diff
@@ -1007,6 +1007,8 @@ static void schedule_order_lock(unsigned lock_index)
 		       monitor32(&rctx->rwin->olock[lock_index],
 				 __ATOMIC_ACQUIRE) != rctx->sn)
 			doze();
+	atomic_store_release(&rctx->rwin->lock_acquired, lock_index, false);
```
Bill-Fischofer-Linaro: Again, no need for fancy atomics here since you're already protected by the ordered lock itself. You're just remembering the `lock_index` you locked.

platform/linux-generic/odp_schedule_scalable.c
```diff
@@ -1025,9 +1027,25 @@ static void schedule_order_unlock(unsigned lock_index)
 	atomic_store_release(&rctx->rwin->olock[lock_index],
 			     rctx->sn + 1,
 			     /*readonly=*/false);
+	atomic_store_release(&rctx->rwin->lock_acquired,
+			     rctx->rwin->lock_count + 1, false);
```
Bill-Fischofer-Linaro: Same comment here. A simple `lock_index` tracker is all that's needed.

platform/linux-generic/odp_schedule_scalable.c
```diff
@@ -1025,9 +1027,25 @@ static void schedule_order_unlock(unsigned lock_index)
...
 
+static void schedule_order_unlock_lock(unsigned lock_index)
+{
+	struct reorder_context *rctx;
+
+	rctx = sched_ts->rctx;
+	if (odp_unlikely(rctx == NULL || rctx->rwin == NULL)) {
+		ODP_ERR("Invalid call to odp_schedule_order_unlock_lock\n");
+		return;
+	}
```
Bill-Fischofer-Linaro: This can be optimized into an `ODP_ASSERT()` to be consistent with the other schedulers. Behavior is undefined if these routines are not called within an ordered context, so no need to take a runtime hit to do these type of checks.

platform/linux-generic/odp_schedule_scalable.c
```diff
@@ -1025,9 +1027,25 @@ static void schedule_order_unlock(unsigned lock_index)
...
+	schedule_order_unlock(__atomic_load_n(&rctx->rwin->lock_acquired,
+					      __ATOMIC_ACQUIRE));
```
Bill-Fischofer-Linaro: A simple reference to the stored `lock_index` is sufficient here. No need for atomics.

platform/linux-generic/odp_schedule_sp.c
```diff
@@ -819,6 +819,11 @@ static void schedule_order_unlock(unsigned lock_index)
 	(void)lock_index;
 }
 
+static void schedule_order_unlock_lock(unsigned lock_index)
+{
+	(void)lock_index;
```
Bill-Fischofer-Linaro: I didn't realize the schedule_sp didn't implement ordered locks. @psavol is this a "to do" item?

platform/linux-generic/odp_schedule.c
```diff
@@ -163,6 +163,7 @@ typedef struct {
 		int stash_num; /**< Number of stashed enqueue operations */
 		uint8_t in_order; /**< Order status */
 		lock_called_t lock_called; /**< States of ordered locks */
+		uint32_t lock_index;
```
Bill-Fischofer-Linaro: This should have a doxygen description for consistency with surrounding vars.

platform/linux-generic/odp_schedule.c
```diff
@@ -1141,9 +1143,17 @@ static void schedule_order_unlock(unsigned lock_index)
 
 	ODP_ASSERT(sched_local.ordered.ctx == odp_atomic_load_u64(ord_lock));
 
+	sched_local.ordered.lock_index = sched->queue[queue_index].
+					order_lock_count + 1;
```
Bill-Fischofer-Linaro: Nice way to avoid the extraneous checking in `schedule_order_unlock_lock()`, but actually this isn't needed at all due to the prior `ODP_ASSERT()`. That assert requires that the `lock_index` lock be held at entry to `schedule_order_unlock` so if `odp_schedule_order_unlock_unlock()` is called without an ordered lock being held, the previous `ODP_ASSERT()` will detect that and abort. 

Interestingly, this also means that (in odp-linux, at least) you can call `odp_schedule_order_unlock_lock()` while holding nested ordered locks. The semantics is always to release the most recently acquired ordered lock and replace that with a new ordered lock regardless of how many other ordered locks the caller holds in this ordered context.

So the net is you can delete this statement without any loss of correctness.

platform/linux-generic/odp_schedule_iquery.c
```diff
@@ -223,6 +223,7 @@ struct sched_thread_local {
 		int stash_num; /**< Number of stashed enqueue operations */
 		uint8_t in_order; /**< Order status */
 		lock_called_t lock_called; /**< States of ordered locks */
+		uint32_t lock_index;
```
Bill-Fischofer-Linaro: Again, this should have a doxygen comment for consistency.

platform/linux-generic/odp_schedule_iquery.c
```diff
@@ -1293,9 +1295,17 @@ static void schedule_order_unlock(unsigned lock_index)
 
 	ODP_ASSERT(thread_local.ordered.ctx == odp_atomic_load_u64(ord_lock));
 
+	thread_local.ordered.lock_index = sched->queues[queue_index].
+						lock_count + 1;
```
Bill-Fischofer-Linaro: Delete, as per above explanation.

platform/linux-generic/odp_schedule_scalable.c
```diff
@@ -1007,6 +1007,8 @@ static void schedule_order_lock(unsigned lock_index)
 		       monitor32(&rctx->rwin->olock[lock_index],
 				 __ATOMIC_ACQUIRE) != rctx->sn)
 			doze();
+	rctx->rwin->lock_index = lock_index;
+
```
Bill-Fischofer-Linaro: This needs to go after the `if` statement, not inside it, since the whole statement is skipped if the lock is immediately available--it's only there to wait for the lock.

platform/linux-generic/odp_schedule_scalable.c
```diff
@@ -1025,9 +1027,23 @@ static void schedule_order_unlock(unsigned lock_index)
 	atomic_store_release(&rctx->rwin->olock[lock_index],
 			     rctx->sn + 1,
 			     /*readonly=*/false);
+	rctx->rwin->lock_index = rctx->rwin->lock_count + 1;
```
Bill-Fischofer-Linaro: Delete, as per previous explanation. The scalable scheduler could be further optimized by changing its runtime checks to ODP_ASSERT()s, but that's beyond the scope of this PR.

include/odp/api/spec/schedule.h
```diff
@@ -368,6 +368,27 @@ void odp_schedule_order_lock(unsigned lock_index);
...
+ * context lock. This call is valid only when there is a single ordered context
+ * lock active. Results are undefined if this call is made when multiple ordered
+ * context locks are acquired in nested fashion.
+ * This call does not release the existing ordered context and only switches
+ * from one ordered lock to another ordered lock within the same ordered
+ * context.
+ *
+ * @param lock_index	Index of the ordered lock in the current context to be
+ *			acquired. Previously acquired ordered lock is released.
+ *			Must be in the range 0...odp_queue_lock_count() - 1.
```
WonderfulVoid: Shouldn't lock_index be in the range 1..odp_queue_lock_count()-1? We are supposedly already holding one lock (lock_index >=0), this lock will be released and the next lock will be acquired.

platform/linux-generic/include/odp_schedule_scalable_ordered.h
```diff
@@ -79,6 +79,7 @@ typedef struct reorder_window {
 	uint32_t tail;
 	uint32_t turn;
 	uint32_t olock[CONFIG_QUEUE_MAX_ORD_LOCKS];
+	uint32_t lock_index;
```
WonderfulVoid: Why is this uint32_t? The lock_count (defined below) is only an uint16_t. We don't support more than 65535 locks (per ordered queue). The ordered locks themselves use uint32_t (for the "ticket") but also that is overkill, there just won't be that many threads.

platform/linux-generic/odp_schedule_scalable.c
```diff
@@ -1007,6 +1007,8 @@ static void schedule_order_lock(unsigned lock_index)
 		       monitor32(&rctx->rwin->olock[lock_index],
 				 __ATOMIC_ACQUIRE) != rctx->sn)
 			doze();
+	rctx->rwin->lock_index = lock_index;
```
WonderfulVoid: The reorder window (rwin) is shared by all threads, you can't store per-thread data in it, you would use the reorder context (rctx) for that.

But why are we saving the lock_index? Is it used for making any correctness checks (e.g. verifying that the application isn't unlocking some lock it hasn't acquired)?

platform/linux-generic/odp_schedule_sp.c
WonderfulVoid: @psavol The SP scheduler only exists to provide strict priority scheduling? The scalable scheduler does that as well (and doesn't the default scheduler also do basic SP scheduling?). Are there any specific characteristics of the SP scheduler (beyond the basic SP scheduling) that are of interest? In benchmarking, the SP scheduler seemed to be the slowest by far.

include/odp/api/spec/schedule.h
Bill-Fischofer-Linaro: ODP does not have any architectural requirement that ordered locks be acquired in any particular sequence, or that they be acquired at all. The only requirement is that they only can be acquired once per ordered context. So this wording is correct. It's entirely possible that a thread acquires lock 1 first and then switches to lock 0. It's entirely up to the application. 

The beauty of ordered locks is that they are deadlock-free by design even in the face of random acquisition order among the various threads that use them.

platform/linux-generic/include/odp_schedule_scalable_ordered.h
Bill-Fischofer-Linaro: The lock_index is an implied parameter to `odp_schedule_order_unlock_lock()`. It's the index of the lock that the caller currently holds, so this is a thread-local variable. It's an error to call this API without already holding an ordered lock.

platform/linux-generic/odp_schedule_scalable.c
Bill-Fischofer-Linaro: As noted above, since `odp_schedule_order_unlock_lock()` only takes one parameter (the index of the new lock to be acquired), we need a place to store the index of the lock that the caller now holds to be used as an implied parameter to this API. So this is supposed to be a thread-local variable. If that's not what's happening here then that needs to be fixed, so thanks for pointing that out.

WonderfulVoid:
On Fri, Sep 8, 2017 at 11:38 PM, Bill Fischofer <notifications@github.com>
wrote:

> *@Bill-Fischofer-Linaro* commented on this pull request.
> ------------------------------
>
> In include/odp/api/spec/schedule.h
> <https://github.com/Linaro/odp/pull/160#discussion_r137895347>:
>
> > @@ -368,6 +368,27 @@ void odp_schedule_order_lock(unsigned lock_index);
>  void odp_schedule_order_unlock(unsigned lock_index);
>
>  /**
> + * Release existing ordered context lock and acquire a new lock
> + *
> + * This call is valid only when holding an ordered synchronization context.
> + * Release a previously locked ordered context lock and acquire a new ordered
> + * context lock. This call is valid only when there is a single ordered context
> + * lock active. Results are undefined if this call is made when multiple ordered
> + * context locks are acquired in nested fashion.
> + * This call does not release the existing ordered context and only switches
> + * from one ordered lock to another ordered lock within the same ordered
> + * context.
> + *
> + * @param lock_index	Index of the ordered lock in the current context to be
> + *			acquired. Previously acquired ordered lock is released.
> + *			Must be in the range 0...odp_queue_lock_count() - 1.
>
> ODP does not have any architectural requirement that ordered locks be
> acquired in any particular sequence, or that they be acquired at all. The
> only requirement is that they only can be acquired once per ordered
> context. So this wording is correct. It's entirely possible that a thread
> acquires lock 1 first and then switches to lock 0. It's entirely up to the
> application.
>
I hadn't realised that. I guess I was lucky then when implementing the
ordered lock code in the scalable scheduler (which I think/hope supports
this). Basically a consequence of doing as little as possible and with as
little state as possible.

include/odp/api/spec/schedule.h
bala-manoharan: Just to reiterate, ordered locks are used to make sure packet belonging to a particular flow acquire the lock in order. There is no need to acquire the lock in ascending order. Even though our linux-generic implementation is using them as index number for performance reasons it will usually be the last few bits of the flow id calculated by RSS or any other mechanism decided by the application.

platform/linux-generic/include/odp_schedule_scalable_ordered.h
bala-manoharan: Not true. Linux-generic implementation is using an uint16_t but as explained before implementations are free to use any number of bits as required. The number of bits of lock_index also depends upon the application use-case.
Ordered locks is not related to number of threads in the system but to the total number of flows in the system. Depending upon the implementation it could support a huge number of flows ( 2^20 ) and not all flows will be active.

platform/linux-generic/odp_schedule_scalable.c
bala-manoharan: The initial draft of ordered lock considered API for locking and unlocking ordered locks within a context. We found an use-case from the customer where the application was splitting a single ordered flow to two different flows, which is possible with our current API architecture but the drawback is that it requires the application to make two different calls odp_schedule_unlock() and odp_schedule_lock() which will have performance issues since the implementation has to internally switch out of the lock and then acquire the new lock. The main objective of this new API is to optimize this use-case by using a single API which switches the ordered context from one ordered lock to another.

bala-manoharan: I will be moving lock_index from rctx to sched_scalable_thread_state_t in the next version and thanks for pointing it out.

platform/linux-generic/odp_schedule.c
bala-manoharan: Done

platform/linux-generic/odp_schedule.c
bala-manoharan: I added this change for scenario where odp_schedule_order_unlock_lock() is called after a lock has been released using odp_schedule_order_unlock() but since we are defining in the documentation this is an undefined behaviour I believe your suggestion makes sense.
I will update in the next version.

platform/linux-generic/odp_schedule_sp.c
psavol: It implements ordered lock API by handling all queues as atomic.

It demonstrates a very simple implementation of the API, and thus is an easy starting for new scheduler development.

include/odp/api/spec/schedule.h
```diff
@@ -368,6 +368,27 @@ void odp_schedule_order_lock(unsigned lock_index);
...
+ *
+ * @see odp_schedule_order_lock(), odp_schedule_order_unlock()
+ *
+ */
+void odp_schedule_order_unlock_lock(uint32_t lock_index);
```
psavol: To match other function prototypes, should be **unsigned** instead of uint32_t.

psavol: +*  This call is valid only when there is a single ordered context
+ * lock active. Results are undefined if this call is made when multiple ordered
+ * context locks are acquired in nested fashion.

Why nested locks are banned? Thinks it should not make difference. Unlock_lock(C) is equivalent to unlock(B) + lock(C) under.

lock(A)
lock(B)
unlock(B)
lock(C)
...

psavol: "This call does not release the existing ordered context and only switches
from one ordered lock to another ordered lock within the same ordered
context."

I'd remove this since this call is not different to lock/unlock calls in this regard

platform/linux-generic/include/odp_schedule_if.h
```diff
@@ -95,6 +95,7 @@ typedef struct {
 				   odp_schedule_group_info_t *);
```
psavol: Don't know how to comment on COMMIT LOG on Github !

Anyway, commit log needs correct format "linux-gen: sched: ...." and some more text (it's empty now)

include/odp/api/spec/schedule.h
psavol: Was there some benefit of not having index of the unlocked lock as a parameter ? Can't remember if I suggested it myself.  Anyway, implementation would simplify and it would still be as robust as unlock() (which application may pass bad lock index).

platform/linux-generic/include/odp_schedule_if.h
muvarov: @psavol only comment like you did.

include/odp/api/spec/schedule.h
```diff
@@ -371,6 +371,31 @@ void odp_schedule_order_lock(uint32_t lock_index);
...
+ * This call does not release the existing ordered context and only switches
+ * from one ordered lock to another ordered lock within the same ordered
+ * context.
+ *
+ * @param unlock_index	Index of the acquired ordered lock in the current
+ *			context to be released. Results are undefined if the
+ *			ordered lock with the given unlock_index is not held by
+ *			this ordered context.
+ * @param lock_index	Index of the ordered lock in the current context to be
+ *			acquired. Previously acquired ordered lock is released.
```
psavol: Remove this outdated sentence: " Previously acquired ordered lock is released." We have now unlock_index to identify which lock is unlocked.

include/odp/api/spec/schedule.h
```diff
@@ -371,6 +371,31 @@ void odp_schedule_order_lock(uint32_t lock_index);
...
+ *
+ * This call is valid only when holding an ordered synchronization context.
+ * Release a previously locked ordered context lock and acquire a new ordered
+ * context lock. This call is valid only when there is a single ordered context
+ * lock active. Results are undefined if this call is made when multiple ordered
+ * context locks are acquired in nested fashion.
+ * This call does not release the existing ordered context and only switches
+ * from one ordered lock to another ordered lock within the same ordered
+ * context.
+ *
```
psavol: It's better to remove this duplicate spec ...
"This call is valid only when there is a single ordered context
lock active. Results are undefined if this call is made when multiple ordered
context locks are acquired in nested fashion.
This call does not release the existing ordered context and only switches
from one ordered lock to another ordered lock within the same ordered
context."

... and refer to lock()/unlock() spec instead: "The operation is equivalent to application calling first odp_schedule_order_unlock(unlock_index) and then odp_schedule_order_lock(lock_index). The same constrains apply to this call than those two."

include/odp/api/spec/schedule.h
```diff
@@ -371,6 +371,31 @@ void odp_schedule_order_lock(uint32_t lock_index);
...
+ * @param unlock_index	Index of the acquired ordered lock in the current
+ *			context to be released. Results are undefined if the
+ *			ordered lock with the given unlock_index is not held by
+ *			this ordered context.
```
psavol: "Results are undefined if the
ordered lock with the given unlock_index is not held by
this ordered context."

Better to move this to documentation of unlock() function (not the param, but function text).

include/odp/api/spec/schedule.h
```diff
@@ -365,7 +368,28 @@ void odp_schedule_order_lock(unsigned lock_index);
...
+ * @param unlock_index	Index of the acquired ordered lock in the current
+ *			context to be released.
+ * @param lock_index	Index of the ordered lock in the current context to be
+ *			acquired. Must be in the range
+ *			0...odp_queue_lock_count() - 1.
+ *
+ * @see odp_schedule_order_lock(), odp_schedule_order_unlock()
+ *
+ */
+void odp_schedule_order_unlock_lock(uint32_t unlock_index, uint32_t lock_index);
```
psavol: Maybe this "The same constraints apply to this call similar to calling these two calls in sequence" needs to be re-worded to be more fluent. E.g. "The same constraints apply with this call as with those two."

platform/linux-generic/include/odp_schedule_if.h
```diff
@@ -95,6 +95,7 @@ typedef struct {
...
 	void (*schedule_order_lock)(unsigned);
 	void (*schedule_order_unlock)(unsigned);
+	void (*schedule_order_unlock_lock)(uint32_t, uint32_t);
```
psavol: Also lock / unlock function pointer prototypes need to be updated to uint32_t.

include/odp/api/spec/schedule.h
```diff
@@ -347,12 +347,15 @@ int odp_schedule_group_info(odp_schedule_group_t group,
...
+ * Only one ordered lock can be active in an ordered context at any given time.
+ * Results are undefined when multiple ordered locks are acquired in nested
+ * fashion within the same ordered context.
  *
  * @param lock_index Index of the ordered lock in the current context to be
  *                   acquired. Must be in the range 0..odp_queue_lock_count()
  *                   - 1
  */
-void odp_schedule_order_lock(unsigned lock_index);
+void odp_schedule_order_lock(uint32_t lock_index);
```
psavol: Also lock_count should be updated to uint32_t.

unsigned -> uint32_t
typedef	struct odp_schedule_param_t {
   ....
   unsigned lock_count;
}

int -> int32_t
int odp_queue_lock_count(odp_queue_t queue);

Check also validation code which uses int/unsigned variables today.

bala-manoharan: Okay

platform/linux-generic/include/odp_schedule_if.h
bala-manoharan: Okay

include/odp/api/spec/schedule.h
bala-manoharan: changed in next verion.

Bill-Fischofer-Linaro:
@bala-manoharan Please rebase in the next version as this now has conflicts that cannot be auto-merged.

bala-manoharan:
Yes. Some issue happened after I did a rebase.
Let me send new version.

Regards,
Bala

bala-manoharan:
@Bill-Fischofer-Linaro @psavol This version V12 incorporates all the review comments and I have rebased it with latest changes from api-next.

platform/linux-generic/odp_queue.c
```diff
@@ -175,12 +175,12 @@ static odp_schedule_group_t queue_sched_group(odp_queue_t handle)
...
 }
 
-static int queue_lock_count(odp_queue_t handle)
+static uint32_t queue_lock_count(odp_queue_t handle)
 {
 	queue_entry_t *queue = handle_to_qentry(handle);
 
 	return queue->s.param.sched.sync == ODP_SCHED_SYNC_ORDERED ?
-		(int)queue->s.param.sched.lock_count : -1;
+		(int)queue->s.param.sched.lock_count : 0;
```
psavol: Cast to (int) should not be done anymore.

platform/linux-generic/odp_queue_scalable.c
```diff
@@ -333,12 +333,12 @@ static odp_schedule_group_t queue_sched_group(odp_queue_t handle)
...
 }
 
-static int queue_lock_count(odp_queue_t handle)
+static uint32_t queue_lock_count(odp_queue_t handle)
 {
 	queue_entry_t *queue = qentry_from_int(queue_from_ext(handle));
 
 	return queue->s.param.sched.sync == ODP_SCHED_SYNC_ORDERED ?
-		(int)queue->s.param.sched.lock_count : -1;
+		(int)queue->s.param.sched.lock_count : 0;
```
psavol: Remove cast (int)

include/odp/api/spec/queue.h
```diff
@@ -393,10 +393,11 @@ odp_schedule_group_t odp_queue_sched_group(odp_queue_t queue);
  *
  * @param queue   Queue handle
  *
- * @return Number of ordered locks associated with this ordered queue
- * @retval <0 Specified queue is not ordered
+ * @return	Number of ordered locks associated with this ordered queue
+ * @retval 0	Specified queue is not ordered or No ordered lock associated
+ *		with the ordered queue.
```
psavol: Typo: upper case "No" in the middle of sentence -> change to lower case "no"

platform/linux-generic/odp_queue_scalable.c
bala-manoharan: This parameter is uint32_t hence I moved from error from -1 to 0.

include/odp/api/spec/queue.h
bala-manoharan: Okay

platform/linux-generic/odp_schedule_iquery.c
```diff
@@ -1296,7 +1296,14 @@ static void schedule_order_unlock(unsigned lock_index)
 	odp_atomic_store_rel_u64(ord_lock, thread_local.ordered.ctx + 1);
 }
 
-static unsigned schedule_max_ordered_locks(void)
+static void schedule_order_unlock_lock(uint32_t unlock_index,
+				       uint32_t lock_index)
+{
```
psavol: Is this here due to a rebase error? Implementation of the new function should not be in this patch.

platform/linux-generic/odp_queue_scalable.c
psavol: That's OK, but now cast to "int" should not be done anymore. The function returns now uint32_t and not int.

static uint32_t queue_lock_count(odp_queue_t handle)

bala-manoharan:
@psavol  V13 incorporates your review comments and fixes the rebase error in previous version.

muvarov:
Hm, interesting - "psavol approved these changes 3 hours ago" but he is missing in the list of approvers.

muvarov:
Merged.

PR #162, opened 2017-09-05
[PATCH v1] Update CUnit detection and document behaviour changes

lumag:

### DIFF
--- a/DEPENDENCIES
+++ b/DEPENDENCIES
@@ -9,12 +9,13 @@ Prerequisites for building the OpenDataPlane (ODP) API
    automake
    autoconf
    libtool
+   pkg-config
 
    On Debian/Ubuntu systems:
-   $ sudo apt-get install automake autoconf libtool
+   $ sudo apt-get install automake autoconf libtool pkg-config
 
    On CentOS/RedHat/Fedora systems:
-   $ sudo yum install automake autoconf libtool
+   $ sudo yum install automake autoconf libtool pkgconfig
 
 3. Required libraries
 
@@ -199,11 +200,11 @@ Prerequisites for building the OpenDataPlane (ODP) API
 
 4.0 Packages needed to build API tests
 
-   Cunit test framework version 2.1-3 is required
-   Cunit prvodes a framework to run the API test suite that proves conformance to the
+   CUnit test framework version 2.1-3 is required
+   CUnit provides a framework to run the API test suite that proves conformance to the
    ODP API. The home page http://cunit.sourceforge.net/doc/introduction.html
 
-4.1 Native Cunit install
+4.1 Native CUnit install
 
    # Debian/Ubuntu
    $ apt-get install libcunit1-dev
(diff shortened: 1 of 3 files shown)


DEPENDENCIES
```diff
@@ -232,20 +233,27 @@ Prerequisites for building the OpenDataPlane (ODP) API
...
+4.4 Using CUnit with ODP
 
-   # ... OR the user defined path.
-   ./configure  --with-cunit-path=/home/<my_cunit_path>
+   Configure will automatically look for CUnit if validation testsuite is
+   enabled. By default it uses pkg-config to locate CUnit. Usually no
+   additional configuration will be required. Few corner cases:
+
+   # Systemwide CUnit if configure errors out with 'missing Version' message
+   ./configure CUNIT_CFLAGS="-I/usr/include" CUNIT_LIBS="-lcunit"
```
muvarov: can configure take care about this step?

DEPENDENCIES
```diff
@@ -199,11 +200,11 @@ Prerequisites for building the OpenDataPlane (ODP) API
 
 4.0 Packages needed to build API tests
 
-   Cunit test framework version 2.1-3 is required
-   Cunit prvodes a framework to run the API test suite that proves conformance to the
+   CUnit test framework version 2.1-3 is required
+   CUnit prvodes a framework to run the API test suite that proves conformance to the
```
Bill-Fischofer-Linaro: Typo: provides

DEPENDENCIES
```diff
@@ -9,12 +9,13 @@ Prerequisites for building the OpenDataPlane (ODP) API
...
    libtool
+   pkg-config
 
    On Debian/Ubuntu systems:
-   $ sudo apt-get install automake autoconf libtool
+   $ sudo apt-get install automake autoconf libtool pkg-config
 
    On CentOS/RedHat/Fedora systems:
-   $ sudo yum install automake autoconf libtool
+   $ sudo yum install automake autoconf libtool pkgconfig
```
Bill-Fischofer-Linaro: Shouldn't this be pkg-config here as well?

lumag: no. Check packages on CentOS/any other rpm distro.

DEPENDENCIES
lumag: Not really, that is a problem. Technically it should be a bug opened against distro to fix cunt.pc. I can resurrect `--with-cunit-path` option, if that makes sense from your POV.

muvarov: I prefer better verbose message from ./configure to avoid reply to the same questions :) I.e. or resurrect  --with-cunit-path or move that text to ./configure

DEPENDENCIES
lumag: fixed

lumag:
@muvarov @Bill-Fischofer-Linaro updated to autodetect CUnit in case of pkg-config error.

muvarov:
looks good for me

muvarov:
Merged.

PR #163, opened 2017-09-05
[PATCH v5] linux-gen: deduplicate arch-specific files

lumag:
Rewrite ARCH-specific files so that there is no significant code
duplication between different arch files and default files set.


muvarov:
According to Travis log this pr brakes build.

lumag:
@muvarov fixed

platform/linux-generic/arch/aarch64/odp/api/cpu_arch.h
```diff
@@ -0,0 +1,30 @@
...
+#define _ODP_CACHE_LINE_SIZE 64
+
+static inline void odp_cpu_pause(void)
+{
+	/* YIELD hints the CPU to switch to another thread if possible
+	 * and executes as a NOP otherwise.
+	 * ISB flushes the pipeline, then restarts. This is guaranteed to
+	 * stall the CPU a number of cycles.
+	 */
+	__asm volatile("isb" ::: "memory");
```
muvarov: that looks like has to be __asm__ __volatile__ to be consistent to other calls. That doc https://gcc.gnu.org/onlinedocs/gcc/Extended-Asm.html says about __asm__ but I did not see doc which talked about __volatile__ instead of volatile. Do you know where to read details?

lumag: This is just a copy of arm's file. We can change both of them, but it's a separate doing.

lumag:
@muvarov Interesting. The patch is marked as failed checkpatch check, but I see no checkpatch issues. Could you please check what went wrong?

lumag:
@Bill-Fischofer-Linaro I probably should stop doing last-minute changes before submitting PR. Fixed now.

lumag:
checkpatch seems to be false positive, because checkpatch did not recognize `FILE` as typename.

lumag:
@Bill-Fischofer-Linaro @muvarov any updates on this PR?

Bill-Fischofer-Linaro:
@lumag I concur that the checkpatch issue is a false positive. I've added my approval.

muvarov:
Merged.

PR #164, opened 2017-09-05
[PATCH v1] test: fixup CUNIT_CFLAGS variable reference

lumag:
Fix type in previous refactoring, which used CUNIT_PFLAGS instead of
CUNIT_CFLAGS.


lumag:
@Bill-Fischofer-Linaro do we need a bugzilla entry to fix typo in Makefile?

muvarov:
Merged. I think there is no need for bug entry for this patch.

PR #169, opened 2017-09-08
[PATCH CLOUD-DEV v4] linux-dpdk:introduce buffer submodule

GBalakrishna:
Made it into two patch series to keep the history of renaming files.


platform/linux-dpdk/buffer/dpdk.c
```diff
@@ -0,0 +1,207 @@
...
+#include <stdio.h>
+#include <inttypes.h>
+
+static odp_buffer_t buffer_alloc(odp_pool_t pool_hdl)
+{
+	odp_buffer_t buffer;
+	pool_entry_cp_t *pool_cp;
+	pool_entry_dp_t *pool_dp;
+
+	pool_cp = odp_pool_to_entry_cp(pool_hdl);
```
nagarahalli: Control part of the pool data structure should not be accessed in this function. The 'pool_cp' is accessed only in ODP_ASSERT. ODP_ASSERT has "if (ODP_DEBUG == 1)" statement. When I was writing this code, I was wondering what will happen to 'pool_cp' variable and line 24 when ODP_DEBUG is set. Since, ODP_DEBUG is a constant, I hope the compiler removes line 24.

heyi-linaro:
merged and closed.

PR #170, opened 2017-09-08
[PATCH v3] Packet references re-implementation

psavol:
Current zero copy reference implementation introduced multiple issues:
  * segment pointers and length were not set correctly, but were
    left to NULL
  * reference counting has a race condition which causes random
    assert failures also with non-reference packets
  * there's larger than expected performance drop with
    non-reference packets
  * the first segment of a reference was zero bytes long, which
    is possible by the API but is unusual for users

This patch set re-implements segmentation, so that zero-copy references may be implemented correctly, efficiently and so that non-reference packets do not see much difference in performance or complexity.

Non-reference packet performance (l2fwd packet rate vs current master) is increased 5-15% depending on CPU count, which brings performance close to what it was before.

### DIFF
--- a/platform/linux-generic/odp_packet.c
+++ b/platform/linux-generic/odp_packet.c
@@ -494,6 +494,76 @@ static inline void copy_buf_hdr(odp_packet_hdr_t *pkt_hdr, int first, int num,
 	}
 }
 
+static inline void buffer_ref_inc(odp_buffer_hdr_t *buf_hdr)
+{
+	uint32_t ref_cnt = odp_atomic_load_u32(&buf_hdr->ref_cnt);
+
+	/* First count increment after alloc */
+	if (odp_likely(ref_cnt) == 0)
+		odp_atomic_store_u32(&buf_hdr->ref_cnt, 2);
+	else
+		odp_atomic_inc_u32(&buf_hdr->ref_cnt);
+}
+
+static inline uint32_t buffer_ref_dec(odp_buffer_hdr_t *buf_hdr)
+{
+	return odp_atomic_fetch_dec_u32(&buf_hdr->ref_cnt);
+}
+
+static inline uint32_t buffer_ref(odp_buffer_hdr_t *buf_hdr)
+{
+	return odp_atomic_load_u32(&buf_hdr->ref_cnt);
+}
+
+static inline int is_multi_ref(uint32_t ref_cnt)
+{
+	return (ref_cnt > 1);
+}
+
+static inline void packet_ref_inc(odp_packet_hdr_t *pkt_hdr)
+{
+	seg_entry_t *seg;
+	int i;
+	int seg_count = pkt_hdr->buf_hdr.segcount;
+	odp_packet_hdr_t *hdr = pkt_hdr;
+	uint8_t idx = 0;
+
+	for (i = 0; i < seg_count; i++) {
+		seg = seg_entry_next(&hdr, &idx);
+		buffer_ref_inc(seg->hdr);
+	}
+}
+
+static inline void packet_free_multi(odp_buffer_hdr_t *hdr[], int num)
+{
+	int i;
+	uint32_t ref_cnt;
+	int num_ref = 0;
+
+	for (i = 0; i < num; i++) {
+		/* Zero when reference API has not been used */
+		ref_cnt = buffer_ref(hdr[i]);
+
+		if (odp_unlikely(ref_cnt)) {
+			ref_cnt = buffer_ref_dec(hdr[i]);
+
+			if (is_multi_ref(ref_cnt)) {
+				num_ref++;
+				continue;
+			}
+		}
+
+		/* Skip references and pack to be freed headers to array head */
+		if (odp_unlikely(num_ref))
+			hdr[i - num_ref] = hdr[i];
+	}
+
+	num -= num_ref;
+
+	if (odp_likely(num))
+		buffer_free_multi(hdr, num);
+}
+
 static inline void free_all_segments(odp_packet_hdr_t *pkt_hdr, int num)
 {
 	seg_entry_t *seg;
@@ -506,7 +576,7 @@ static inline void free_all_segments(odp_packet_hdr_t *pkt_hdr, int num)
 		buf_hdr[i] = seg->hdr;
 	}
 
-	buffer_free_multi(buf_hdr, num);
+	packet_free_multi(buf_hdr, num);
 }
 
 static inline odp_packet_hdr_t *free_segments(odp_packet_hdr_t *pkt_hdr,
@@ -563,7 +633,7 @@ static inline odp_packet_hdr_t *free_segments(odp_packet_hdr_t *pkt_hdr,
 
 		pkt_hdr = new_hdr;
 
-		buffer_free_multi(buf_hdr, num);
+		packet_free_multi(buf_hdr, num);
 	} else {
 		/* Free last 'num' bufs.
 		 * First, find the last remaining header. */
@@ -578,7 +648,7 @@ static inline odp_packet_hdr_t *free_segments(odp_packet_hdr_t *pkt_hdr,
 			buf_hdr[i] = seg->hdr;
 		}
 
-		buffer_free_multi(buf_hdr, num);
+		packet_free_multi(buf_hdr, num);
 
 		/* Head segment remains, no need to copy or update majority
 		 * of the metadata. */
@@ -702,7 +772,7 @@ void odp_packet_free(odp_packet_t pkt)
 	int num_seg = pkt_hdr->buf_hdr.segcount;
 
 	if (odp_likely(CONFIG_PACKET_MAX_SEGS == 1 || num_seg == 1))
-		buffer_free_multi((odp_buffer_hdr_t **)&hdl, 1);
+		packet_free_multi((odp_buffer_hdr_t **)&hdl, 1);
 	else
 		free_all_segments(pkt_hdr, num_seg);
 }
@@ -710,7 +780,7 @@ void odp_packet_free(odp_packet_t pkt)
 void odp_packet_free_multi(const odp_packet_t pkt[], int num)
 {
 	if (CONFIG_PACKET_MAX_SEGS == 1) {
-		buffer_free_multi((odp_buffer_hdr_t **)(uintptr_t)pkt, num);
+		packet_free_multi((odp_buffer_hdr_t **)(uintptr_t)pkt, num);
 	} else {
 		odp_buffer_hdr_t *buf_hdr[num * CONFIG_PACKET_MAX_SEGS];
 		int i;
@@ -731,7 +801,7 @@ void odp_packet_free_multi(const odp_packet_t pkt[], int num)
 			bufs += num_seg - 1;
 		}
 
-		buffer_free_multi(buf_hdr, bufs);
+		packet_free_multi(buf_hdr, bufs);
 	}
 }
 
@@ -1942,7 +2012,12 @@ uint64_t odp_packet_seg_to_u64(odp_packet_seg_t hdl)
 
 odp_packet_t odp_packet_ref_static(odp_packet_t pkt)
 {
-	return odp_packet_copy(pkt, odp_packet_pool(pkt));
+	odp_packet_hdr_t *pkt_hdr = packet_hdr(pkt);
+
+	packet_ref_inc(pkt_hdr);
+	pkt_hdr->shared_len = pkt_hdr->frame_len;
+
+	return pkt;
 }
 
 odp_packet_t odp_packet_ref(odp_packet_t pkt, uint32_t offset)
@@ -2004,14 +2079,32 @@ odp_packet_t odp_packet_ref_pkt(odp_packet_t pkt, uint32_t offset,
 
 int odp_packet_has_ref(odp_packet_t pkt)
 {
-	(void)pkt;
+	odp_buffer_hdr_t *buf_hdr;
+	seg_entry_t *seg;
+	int i;
+	uint32_t ref_cnt;
+	odp_packet_hdr_t *pkt_hdr = packet_hdr(pkt);
+	int seg_count = pkt_hdr->buf_hdr.segcount;
+	odp_packet_hdr_t *hdr = pkt_hdr;
+	uint8_t idx = 0;
+
+	for (i = 0; i < seg_count; i++) {
+		seg = seg_entry_next(&hdr, &idx);
+		buf_hdr = seg->hdr;
+		ref_cnt = buffer_ref(buf_hdr);
+
+		if (is_multi_ref(ref_cnt))
+			return 1;
+	}
 
 	return 0;
 }
 
 uint32_t odp_packet_unshared_len(odp_packet_t pkt)
 {
-	return odp_packet_len(pkt);
+	odp_packet_hdr_t *pkt_hdr = packet_hdr(pkt);
+
+	return packet_len(pkt_hdr) - pkt_hdr->shared_len;
 }
 
 /* Include non-inlined versions of API functions */
--- a/platform/linux-generic/odp_packet.c
+++ b/platform/linux-generic/odp_packet.c
@@ -35,10 +35,6 @@ const _odp_packet_inline_offset_t _odp_packet_inline ODP_ALIGNED_CACHE = {
 	.frame_len      = offsetof(odp_packet_hdr_t, frame_len),
 	.headroom       = offsetof(odp_packet_hdr_t, headroom),
 	.tailroom       = offsetof(odp_packet_hdr_t, tailroom),
-	.unshared_len   = offsetof(odp_packet_hdr_t, unshared_len),
-	.ref_hdr        = offsetof(odp_packet_hdr_t, ref_hdr),
-	.ref_offset     = offsetof(odp_packet_hdr_t, ref_offset),
-	.ref_len        = offsetof(odp_packet_hdr_t, ref_len),
 	.pool           = offsetof(odp_packet_hdr_t, buf_hdr.pool_hdl),
 	.input          = offsetof(odp_packet_hdr_t, input),
 	.segcount       = offsetof(odp_packet_hdr_t, buf_hdr.segcount),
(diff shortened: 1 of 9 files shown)


platform/linux-generic/odp_packet.c
```diff
@@ -611,80 +807,59 @@ int odp_packet_alloc_multi(odp_pool_t pool_hdl, uint32_t len,
...
-					ODP_ASSERT
-						((packet_ref_count_set(hdr, 0),
-						  1));
-					buf_hdr[nfree++] = &hdr->buf_hdr;
-				}
-			}
+		if (odp_unlikely(seg_is_link(pkt_hdr))) {
+			buf_hdr2[links] = pkt_hdr->buf_hdr.seg[0].hdr;
+			links++;
+		}
```
Bill-Fischofer-Linaro:
```
if (odp_unlikely(seg_is_link(pkt_hdr)))
        buf_hdr2[links++] = pkt_hdr->buf_hdr.seg[0].hdr;
```

platform/linux-generic/odp_packet.c
```diff
@@ -414,97 +490,217 @@ static inline odp_packet_hdr_t *add_segments(odp_packet_hdr_t *pkt_hdr,
...
+
+static inline void packet_free_multi(odp_buffer_hdr_t *hdr[], int num)
+{
+	int i;
+	uint32_t ref_cnt;
+	int num_ref = 0;
+
+	for (i = 0; i < num; i++) {
+		/* Zero when reference API has not been used */
+		ref_cnt = odp_atomic_load_u32(&hdr[i]->ref_cnt);
```
Bill-Fischofer-Linaro: The previous code wrapped the `ref_cnt` manipulation routines  in `static inline` functions to give more flexibility in the implementation. Not a major point, but I think it would make sense to continue that in this version. For example, in cases when references are not being used it's not really necessary to use an `odp_atomic_load_u32()` to inspect the value of `ref_cnt`.  If you read this as zero, you know there's no reference. It's only when `ref_cnt` is non-zero do you need the atomics.

platform/linux-generic/odp_packet.c
```diff
@@ -414,97 +490,217 @@ static inline odp_packet_hdr_t *add_segments(odp_packet_hdr_t *pkt_hdr,
...
+
+		if (odp_unlikely(ref_cnt)) {
+			ref_cnt = odp_atomic_fetch_dec_u32(&hdr[i]->ref_cnt);
```
Bill-Fischofer-Linaro: As above, I'd prefer to see the `odp_atomic_fetch_dec_u32()` in a wrapper function. The mainline code shouldn't need to be aware of  how `ref_cnt` is maintained, only to make the necessary increment/decrement calls.

platform/linux-generic/odp_packet.c
```diff
@@ -414,97 +490,217 @@ static inline odp_packet_hdr_t *add_segments(odp_packet_hdr_t *pkt_hdr,
...
-			ODP_ASSERT((packet_ref_count_set(hdr, 0), 1));
-			buf_hdr[nfree++] = &hdr->buf_hdr;
+	if (odp_likely(pkt_hdr->buf_hdr.num_seg == num)) {
+		for (i = 0; i < num; i++)
+			buf_hdr[i] = pkt_hdr->buf_hdr.seg[i].hdr;
+
+		if (odp_unlikely(seg_is_link(pkt_hdr))) {
+			buf_hdr[num] = &pkt_hdr->buf_hdr;
+			num++;
 		}
```
Bill-Fischofer-Linaro:
```
if (odp_unlikely(seg_is_link(pkt_hdr)))
        buf_hdr[num++] = &pkt_hdr->buf_hdr;
```

platform/linux-generic/odp_packet.c
```diff
@@ -414,97 +490,217 @@ static inline odp_packet_hdr_t *add_segments(odp_packet_hdr_t *pkt_hdr,
...
+		odp_buffer_hdr_t *link_hdr[num];
+		uint8_t idx = 0;
+		int links = 0;
+
+		for (i = 0; i < num; i++) {
+			/* Free also link headers */
+			if (odp_unlikely(idx == 0 && seg_is_link(pkt_hdr))) {
+				link_hdr[links] = &pkt_hdr->buf_hdr;
+				links++;
+			}
```
Bill-Fischofer-Linaro:
```
if (odp_unlikely(idx == 0 && seg_is_link(pkt_hdr)))
        link_hdr[links++] = &pkt_hdr->buf_hdr;
```

platform/linux-generic/odp_packet.c
```diff
@@ -414,97 +490,217 @@ static inline odp_packet_hdr_t *add_segments(odp_packet_hdr_t *pkt_hdr,
...
+
+			seg = seg_entry_next(&pkt_hdr, &idx);
+			buf_hdr[i] = seg->hdr;
+		}
+
+		if (odp_unlikely(links))
+			packet_free_multi(link_hdr, links);
```
Bill-Fischofer-Linaro: Given what's necessary to enter this arm, shouldn't this be `odp_likely()`? Isn't the only way `segcount` != `num_seg` is if we have a link segment?

platform/linux-generic/odp_packet.c
```diff
@@ -414,97 +490,217 @@ static inline odp_packet_hdr_t *add_segments(odp_packet_hdr_t *pkt_hdr,
...
-			if (packet_ref_count(new_hdr) == 1 ||
-			    packet_ref_dec(new_hdr) == 1) {
-				ODP_ASSERT((packet_ref_count_set(new_hdr, 0),
-					    1));
-				buf_hdr[nfree++] = &new_hdr->buf_hdr;
+			/* Free link headers, if those become empty */
+			if (odp_unlikely(idx == 0 && seg_is_link(tmp_hdr))) {
+				link_hdr[links] = &tmp_hdr->buf_hdr;
+				links++;
 			}
```
Bill-Fischofer-Linaro:
```
if (odp_unlikely(idx == 0 && seg_is_link(tmp_hdr)))
        link_hdr[links++] = &tmp_hdr->buf_hdr;
```

platform/linux-generic/odp_packet.c
```diff
@@ -414,97 +490,217 @@ static inline odp_packet_hdr_t *add_segments(odp_packet_hdr_t *pkt_hdr,
...
+		for (i = 0; i < num; i++) {
+			tmp_hdr    = hdr;
+			seg        = seg_entry_next(&hdr, &idx);
+			buf_hdr[i] = seg->hdr;
+
+			/* Free link headers, if those become empty */
+			if (odp_unlikely(idx == 0 && seg_is_link(tmp_hdr))) {
+				link_hdr[links] = &tmp_hdr->buf_hdr;
+				links++;
+			}
```
Bill-Fischofer-Linaro:
```
if (odp_unlikely(idx == 0 && seg_is_link(tmp_hdr)))
        link_hdr[links++] = &tmp_hdr->buf_hdr;
```

test/common_plat/validation/api/packet/packet.c
```diff
@@ -2226,6 +2319,14 @@ void packet_test_ref(void)
 	if (odp_packet_has_ref(base_pkt) == 1) {
 		/* CU_ASSERT needs braces */
```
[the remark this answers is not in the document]
psavol: OK. Corrected in v3.

platform/linux-generic/odp_packet.c
```diff
@@ -88,19 +74,112 @@ static inline odp_buffer_t packet_to_buffer(odp_packet_t pkt)
...
+static inline void *packet_seg_data(odp_packet_hdr_t *pkt_hdr, uint32_t seg_idx)
 {
-	return pkt_hdr->buf_hdr.seg[seg_idx].data;
+	seg_entry_t *seg = seg_entry(pkt_hdr, seg_idx);
+
+	return seg->data;
 }
 
-static inline uint32_t packet_last_seg(odp_packet_hdr_t *pkt_hdr)
+static inline int packet_last_seg(odp_packet_hdr_t *pkt_hdr)
```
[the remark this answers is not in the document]
psavol: OK. Changed to uint16_t in v3.

platform/linux-generic/odp_packet.c
```diff
@@ -88,19 +74,112 @@ static inline odp_buffer_t packet_to_buffer(odp_packet_t pkt)
...
+
+	if (idx == num_seg - 1) {
+		*cur_hdr = hdr->buf_hdr.next_seg;
+		*cur_idx = 0;
+	} else {
+		*cur_idx = idx + 1;
+	}
+
+	return &hdr->buf_hdr.seg[idx];
+}
```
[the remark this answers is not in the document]
psavol: OK. Done in v3.

Bill-Fischofer-Linaro:
This is still showing as v2. Is v3 still expected?

platform/linux-generic/odp_packet.c
```diff
@@ -271,52 +336,83 @@ void packet_parse_reset(odp_packet_hdr_t *pkt_hdr)
...
-				odp_buffer_hdr_t *buf_hdr;
-
-				packet_ref_count_set(pkt_hdr[i], 1);
-				buf_hdr = &pkt_hdr[i]->buf_hdr;
-				hdr->buf_hdr.seg[i].hdr  = buf_hdr;
-				hdr->buf_hdr.seg[i].data = buf_hdr->base_data;
-				hdr->buf_hdr.seg[i].len  = BASE_LEN;
-			}
+			link_segments(pkt_hdr, num);
+
```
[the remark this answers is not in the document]
psavol: No, since the plan is to optimize link_segment() so that it does not touch the first segment again, but starts from seg[1] instead. I didn't do it yet so that the first version is more robust/readable. Also it will not be very significant improvement, as most packets should be single segment (small packet -> high packet rate).

platform/linux-generic/odp_packet.c
psavol: This line is not from this commit (but from a later one).

platform/linux-generic/odp_packet.c
psavol: Atomic variables must be accessed through atomics API, also when just loading. Atomic_load() is in most cases just a normal load, but still we should use the proper function...

Wrapping load would not help much (load is a load). Increment and store have already been wrapped into buffer_ref_inc(). I'll wrap fetch_dec (into buffer_ref_dec()) and ref_cnt > 1 (into ref_cnt_not_last().

platform/linux-generic/odp_packet.c
psavol: I'll add two new wrapper functions. The main line code here is packet.c, which need to be quite well aware how references are counted ... so it's merely code readability improvement.

platform/linux-generic/odp_packet.c
```diff
@@ -2458,6 +2089,184 @@ uint64_t odp_packet_seg_to_u64(odp_packet_seg_t hdl)
 	return _odp_pri(hdl);
 }
 
+static inline void buffer_ref_inc(odp_buffer_hdr_t *buf_hdr)
+{
+	/* First count increment after alloc */
+	if (odp_atomic_load_u32(&buf_hdr->ref_cnt) == 0)
+		odp_atomic_store_u32(&buf_hdr->ref_cnt, 2);
```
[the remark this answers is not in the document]
psavol: Added likely there. Retry would be dual ref always, multicast would be N refs always. So, it depends on use case which is more likely.

platform/linux-generic/odp_packet.c
```diff
@@ -88,19 +74,112 @@ static inline odp_buffer_t packet_to_buffer(odp_packet_t pkt)
...
+	uint8_t idx, cur_idx;
+	seg_entry_t *seg = NULL;
+	uint32_t seg_start = 0, seg_end = 0;
+	int seg_count;
+
+	hdr     = *p_hdr;
+	cur_hdr = hdr;
+	idx     = 0;
+	cur_idx = 0;
+	seg_count = hdr->buf_hdr.segcount;
```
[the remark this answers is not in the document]
psavol: No, but GCC does not know that. So, those are needed to avoid "may be used uninitialized" warning.

platform/linux-generic/odp_packet.c
psavol: I prefer the more explicit format, where num++ is on a separate line. It's easier to me to track when and where num is incremented.

platform/linux-generic/odp_packet.c
psavol: Same thing for all link++ cases. Links increment catches eye better on a separate line.

platform/linux-generic/odp_packet.c
psavol: No, this arm can be entered also without references (links). The "if arm" (num_seg == num) is for the likely case that the pkt_hdr holds all segments. The "else arm" is for the more complex case where we need to follow next pointer(s) to find all segments. E.g. concat just links two pkt_hdrs together and after that we could e.g. have two segments - one in the pkt_hdr, one behind the next pointer.

platform/linux-generic/odp_packet.c
psavol: Prefer this way.

platform/linux-generic/odp_packet.c
psavol: Prefer separate lines.

platform/linux-generic/odp_packet.c
```diff
@@ -611,80 +807,59 @@ int odp_packet_alloc_multi(odp_pool_t pool_hdl, uint32_t len,
...
 
-void odp_packet_free(odp_packet_t pkt)
-{
-	packet_free(packet_hdr(pkt));
+		buf_hdr[0] = &pkt_hdr->buf_hdr;
+
+		if (odp_unlikely(seg_is_link(pkt_hdr))) {
+			buf_hdr[1] = pkt_hdr->buf_hdr.seg[0].hdr;
+			num++;
+		}
```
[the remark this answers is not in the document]
psavol: I'll update it to num = 2. Since here we only two option, either 1 or 2 buf_hdrs are freed.

platform/linux-generic/odp_packet.c
psavol: Prefer separate lines.

platform/linux-generic/odp_packet.c
```diff
@@ -2458,6 +2089,184 @@ uint64_t odp_packet_seg_to_u64(odp_packet_seg_t hdl)
...
+	odp_packet_hdr_t *hdr = pkt_hdr;
+	seg_entry_t *seg;
+	uint32_t seg_idx = 0;
+	uint8_t idx = 0;
+	uint32_t seg_offset = 0;
+	int i, num_copy, segcount;
+	uint32_t len;
+
+	if (offset >= packet_len(pkt_hdr)) {
+		ODP_ERR("offset too large\n");
```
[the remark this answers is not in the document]
psavol: Isn't ODP_ABORT() for that? We use more often ERR (320 times) than DBG (205 times). I think the difference is that ERR should almost never happen, since it cannot be muted. Where as DBG can be used on more dynamic errors that may end up spamming the screen. Anyway, will change to DBG.

platform/linux-generic/odp_packet.c
```diff
@@ -2458,6 +2089,184 @@ uint64_t odp_packet_seg_to_u64(odp_packet_seg_t hdl)
...
+
+	if (seg_is_link(hdr))
+		buffer_ref_inc((odp_buffer_hdr_t *)hdr);
+
+	seg = seg_entry_next(&hdr, &idx);
+	link_hdr->buf_hdr.num_seg = 1;
+	link_hdr->buf_hdr.seg[0].hdr  = seg->hdr;
+	link_hdr->buf_hdr.seg[0].data = seg->data + seg_offset;
+	link_hdr->buf_hdr.seg[0].len  = seg->len  - seg_offset;
+	buffer_ref_inc(seg->hdr);
```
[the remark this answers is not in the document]
psavol: Link hdr does not point into itself as a segment, but we need to ref count also those. So, hdr != seg->hdr for links. I'll add comments.

platform/linux-generic/odp_packet.c
```diff
@@ -2458,6 +2089,184 @@ uint64_t odp_packet_seg_to_u64(odp_packet_seg_t hdl)
...
+		if (idx == 0 && seg_is_link(hdr))
+			buffer_ref_inc((odp_buffer_hdr_t *)hdr);
+
+		seg = seg_entry_next(&hdr, &idx);
+
+		link_hdr->buf_hdr.num_seg++;
+		link_hdr->buf_hdr.seg[i].hdr  = seg->hdr;
+		link_hdr->buf_hdr.seg[i].data = seg->data;
+		link_hdr->buf_hdr.seg[i].len  = seg->len;
+		buffer_ref_inc(seg->hdr);
```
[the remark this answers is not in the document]
psavol: Same here. Ref count of link header itself needs update also. Added comment.

platform/linux-generic/odp_packet.c
```diff
@@ -2458,6 +2089,184 @@ uint64_t odp_packet_seg_to_u64(odp_packet_seg_t hdl)
...
+
+	next_hdr = hdr;
+
+	/* Increment ref count for remaining segments */
+	for (i = seg_idx + num_copy; i < segcount; i++) {
+		if (idx == 0 && seg_is_link(hdr))
+			buffer_ref_inc((odp_buffer_hdr_t *)hdr);
+
+		seg = seg_entry_next(&hdr, &idx);
+		buffer_ref_inc(seg->hdr);
```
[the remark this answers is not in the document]
psavol: Same answer.

platform/linux-generic/odp_packet.c
```diff
@@ -2458,6 +2089,184 @@ uint64_t odp_packet_seg_to_u64(odp_packet_seg_t hdl)
...
+	len = pkt_hdr->frame_len - offset;
+	link_hdr->buf_hdr.next_seg  = next_hdr;
+	link_hdr->buf_hdr.last_seg  = pkt_hdr->buf_hdr.last_seg;
+	link_hdr->buf_hdr.segcount  = segcount - seg_idx;
+	link_hdr->frame_len         = len;
+	link_hdr->tailroom          = pkt_hdr->tailroom;
+	link_hdr->shared_len        = len;
+
+	/* Link header does not have headroom */
+	link_hdr->headroom          = 0;
```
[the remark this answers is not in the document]
psavol: OK. Added explanation.

platform/linux-generic/odp_packet.c
```diff
@@ -2458,6 +2089,184 @@ uint64_t odp_packet_seg_to_u64(odp_packet_seg_t hdl)
...
+
+	return 0;
+}
+
+uint32_t odp_packet_unshared_len(odp_packet_t pkt)
+{
+	odp_packet_hdr_t *pkt_hdr = packet_hdr(pkt);
+
+	if (odp_packet_has_ref(pkt))
+		return packet_len(pkt_hdr) - pkt_hdr->shared_len;
```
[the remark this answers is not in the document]
psavol: Majority of frame_len references in this file are direct (->frame_len). So, I'll change it to direct also for reference code, then packet_len() is used outside of packet.c (which is how it should be).

muvarov:
Merged.

PR #172, opened 2017-09-11
[PATCH API-NEXT v12] api: classification: add random early detection

bala-manoharan:
Adds random early discard and back pressure feature configurable per CoS

### DIFF
--- a/include/odp/api/spec/classification.h
+++ b/include/odp/api/spec/classification.h
@@ -135,6 +135,13 @@ typedef struct odp_cls_capability_t {
 
 	/** A Boolean to denote support of PMR range */
 	odp_bool_t pmr_range_supported;
+
+	/** Support for Random Early Detection */
+	odp_support_t random_early_detection;
+
+	/** Support for Back Pressure to the remote peer */
+	odp_support_t back_pressure;
+
 } odp_cls_capability_t;
 
 /**
@@ -206,6 +213,53 @@ typedef struct odp_cls_cos_param {
 
 	/** Drop policy associated with CoS */
 	odp_cls_drop_t drop_policy;
+
+	/** Random Early Detection (RED)
+	 * Random Early Detection is enabled to initiate a drop probability
+	 * for the incoming packet when the packets in the queue/pool reaches
+	 * a specified threshold.
+	 * When RED is enabled for a particular flow then further incoming
+	 * packets are assigned a drop probability based on the size of the
+	 * pool/queue and the drop probability becomes 100% when the queue/pool
+	 * is full.
+	 * RED is logically configured in the CoS and could be implemented
+	 * in either pool or queue linked to the CoS depending on
+	 * platform capabilities. Application should make sure not to link
+	 * multiple CoS with different RED or BP configuration to the same queue
+	 * or pool.
+	 * RED is enabled when the resource limit is equal to or greater than
+	 * the maximum threshold value and is disabled when resource limit
+	 * is less than or equal to minimum threshold value. */
+
+	/** A boolean to enable RED
+	 * When true, RED is enabled and configured with RED parameters.
+	 * Otherwise, RED parameters are ignored. */
+	odp_bool_t red_enable;
+
+	/** RED parameters */
+	struct {
+		/** Maximum threshold value for RED */
+		uint32_t max_threshold;
+
+		/** Minimum threshold value for RED */
+		uint32_t min_threshold;
+	} red;
+
+	/** A boolean to enable Back pressure
+	 * When true, back pressure is enabled and configured with the BP
+	 * parameters. Otherwise BP parameters are ignored. When back pressure
+	 * is enabled for a particular flow, the HW can send back pressure
+	 * information to the remote peer indicating a network congestion */
+	odp_bool_t bp_enable;
+
+	/** BP parameters */
+	struct {
+		/** Threshold value for back pressure.
+		 * BP is enabled when queue or pool size is equal to or greater
+		 * than the backpressure threshold. BP threshold is expressed
+		 * as number packets in either queue or pool. */
+		uint32_t bp_threshold;
+	} bp;
 } odp_cls_cos_param_t;
 
 /**
--- a/platform/linux-generic/odp_classification.c
+++ b/platform/linux-generic/odp_classification.c
@@ -190,6 +190,8 @@ int odp_cls_capability(odp_cls_capability_t *capability)
 	capability->supported_terms.bit.tcp_sport = 1;
 	capability->supported_terms.bit.sip_addr = 1;
 	capability->supported_terms.bit.dip_addr = 1;
+	capability->random_early_detection = ODP_SUPPORT_NO;
+	capability->back_pressure = ODP_SUPPORT_NO;
 	return 0;
 }
 


include/odp/api/spec/classification.h
```diff
@@ -135,6 +135,13 @@ typedef struct odp_cls_capability_t {
 
 	/** A Boolean to denote support of PMR range */
 	odp_bool_t pmr_range_supported;
+
+	/** Support for Random Early Discard */
+	odp_support_t random_early_discard;
```
Bill-Fischofer-Linaro: The correct term is Random Early Detection, not Discard. See [RFC 7567](https://tools.ietf.org/html/rfc7567).

include/odp/api/spec/classification.h
```diff
@@ -206,6 +213,43 @@ typedef struct odp_cls_cos_param {
 
 	/** Drop policy associated with CoS */
 	odp_cls_drop_t drop_policy;
+
+	/* Random Early Discard (RED)
+	 * Random Early discard is enabled to initiate a drop probability
+	 * for the incoming packet when the packets in the queue/pool reaches
+	 * a specified threshold.
```
Bill-Fischofer-Linaro: As noted above, Detection.

include/odp/api/spec/classification.h
```diff
@@ -206,6 +213,43 @@ typedef struct odp_cls_cos_param {
...
 	/** Drop policy associated with CoS */
 	odp_cls_drop_t drop_policy;
+
+	/* Random Early Detection (RED)
+	 * Random Early Detection is enabled to initiate a drop probability
+	 * for the incoming packet when the packets in the queue/pool reaches
+	 * a specified threshold.
+	 * When RED is enabled for a particular flow then further incoming
+	 * packets are assigned a drop probability based on the size of the
+	 * pool/queue and the drop probability becomes 1 when the queue/pool
```
psavol: For consistency, it's better to define probability as 100% here, as we use percents below.

include/odp/api/spec/classification.h
```diff
@@ -206,6 +213,43 @@ typedef struct odp_cls_cos_param {
 
 	/** Drop policy associated with CoS */
 	odp_cls_drop_t drop_policy;
+
+	/* Random Early Detection (RED)
+	 * Random Early Detection is enabled to initiate a drop probability
```
psavol: Missing Doxygen tag /**

Run 'make doxygen-doc' locally before pushing and correct all doxygen warnings.

include/odp/api/spec/classification.h
```diff
@@ -206,6 +213,43 @@ typedef struct odp_cls_cos_param {
...
+	 * in either pool or queue linked to the CoS depending on
+	 * platform capabilities.
+	 * RED is controlled using maximum and minimum threshold values
+	 * which are defined as percentage of the system resource.
+	 * RED is enabled when the resource limit is equal to or greater than
+	 * the maximum threshold value and is disabled when resource limit
+	 * is less than or equal to minimum threshold value. */
+
+	/* A boolean to enable RED parameters */
+	odp_bool_t red_enable;
```
psavol: Missing /**

"Enable RED

When true, RED is enabled and configured with the red parameters below.
Otherwise, RED parameters are ignored."

include/odp/api/spec/classification.h
```diff
@@ -206,6 +213,43 @@ typedef struct odp_cls_cos_param {
...
+
+	/* Maximum threshold percentage for RED */
+	uint16_t max_red_threshold;
+
+	/* Minimum threshold percentage for RED */
+	uint16_t min_red_threshold;
```
psavol: Wrap these two into "struct red". Add missing doxygen tagging. We should use 100th parts of a percent. So, that API is not bottle neck if HW can offer higher resolution than 1 percent. TM API uses the same unit.

/** RED parameters
struct {
  Maximum threshold percentage for RED in one-hundredths of a percent. 
  Hence 100% is represented as the integer value 10000.
  uint16_t max_threshold;

  uint16_t min_threshold;
} red;

include/odp/api/spec/classification.h
```diff
@@ -206,6 +213,43 @@ typedef struct odp_cls_cos_param {
...
+
+	/* Back pressure
+	 * When back pressure is enabled for a particular flow, the HW can send
+	 * back pressure information to the remote peer indicating a network
+	 * congestion */
+	odp_bool_t bp_enable;
```
psavol: Missing /**

"Enable back pressure

When true, back pressure is enabled and configured with the parameters below.
Otherwise, back pressure parameters are ignored. When back pressure is enabled 
for a particular flow, the HW can send back pressure information to the remote 
peer indicating a network congestion."

include/odp/api/spec/classification.h
```diff
@@ -206,6 +213,43 @@ typedef struct odp_cls_cos_param {
...
+
+	/* Threshold for enabling back pressure. BP is enabled when pool/queue
+	 * size is equal to or greater than this backpressure threshold.
+	 * BP threshold is expressed as a percentage of the resource size. */
+	uint16_t bp_threshold;
```
psavol: Although this is single param now, it's better to wrap it into a struct, so that it's easy to add more params later (if needed). Also the 100th percent units should be used.

struct {
  uint16_t ena_threshold;
} bp;

Could there be other thresholds params later? ena_threshold would leave room for other (high, low, ...etc) threshold params.

bala-manoharan: BP in HW is implemented as edge triggered and the interface is configured to send Back pressure to the peer when the level increases the threshold. Current our HW only supports only threshold value and is not monitored as a range.
If any HW later implements this as a range we can add in the future.

include/odp/api/spec/classification.h
bala-manoharan: Okay. 
@muvarov Travis does not check doxygen?

bala-manoharan:
@psavol V6 incorporates your review comments

bala-manoharan:
@muvarov @Bill-Fischofer-Linaro This spec have reviewed-by from Bill and Petri API comments have been merged in V6. Can we Merge this to API-NEXT.

NikhilA-Linaro:
In case user has associated same Queue/Pool to two different CoS, and enabled red only on one CoS. How do we handle that situation.
Since the configuration of RED is percentage of Q/Pool len, It creates ambiguity for user on what basis to configure it. Better to have configurations in absolute numbers.

bala-manoharan:
The configuration on CoS is just logical and HW implement this RED on pool or queue hence it will not be possible to have a single pool with two different RED configurations.
I can add limitation saying that the application should take care of this in the configuration wherein if a multiple CoS links to same pool they should have similar configuration.

muvarov:
So is it ok to be merged?

Bill-Fischofer-Linaro:
@bala-manoharan Please rebase this to pick up the latest Travis fixes so regressions can be properly done on this PR.

muvarov:
@bala-manoharan @psavol 
>@muvarov @Bill-Fischofer-Linaro This spec have reviewed-by from Bill and Petri API comments have been merged in V6. Can we Merge this to API-NEXT.

I do not see Petris review. At least this web page does not show that.

bala-manoharan:
@Bill-Fischofer-Linaro @muvarov V11 has been rebased to api-next HEAD.
I am still waiting on comments from Nikhil.

platform/linux-generic/include/odp/api/std_types.h
```diff
@@ -29,6 +29,8 @@ extern "C" {
 
 typedef int odp_bool_t;
 
+typedef uint16_t odp_percent_t;
```
lumag: In my opinion, let's just push it into `odp/api/spec/std_types.h`.

platform/linux-generic/odp_classification.c
```diff
@@ -190,6 +190,8 @@ int odp_cls_capability(odp_cls_capability_t *capability)
 	capability->supported_terms.bit.tcp_sport = 1;
```
lumag: Commit subject is too long. Also it would be good to mention in commit message that this is a dummy (null) implementation

platform/linux-generic/include/odp/api/std_types.h
muvarov: +1 moving will also easy to support ABI compat mode.

include/odp/api/spec/classification.h
```diff
@@ -206,6 +213,53 @@ typedef struct odp_cls_cos_param {
...
+	odp_bool_t red_enable;
+
+	/** RED parameters */
+	struct {
+		/** Maximum threshold value for RED */
+		uint32_t max_threshold;
+
+		/** Minimum threshold value for RED */
+		uint32_t min_threshold;
+	} red;
```
psavol: Threshold value meaning and units need to be specified:  packets in queue, packets allocated from pool. What are the maximum and minimum values of these? And accuracy is implementation specific?

E.g. pool size 10 000 packet, usually values for min 8000 and max 9 500 could be. Are min 1 and max 100 allowed?

include/odp/api/spec/classification.h
```diff
@@ -206,6 +213,53 @@ typedef struct odp_cls_cos_param {
...
+	odp_bool_t bp_enable;
+
+	/** BP parameters */
+	struct {
+		/** Threshold value for back pressure.
+		 * BP is enabled when queue or pool size is equal to or greater
+		 * than the backpressure threshold. BP threshold is expressed
+		 * as number packets in either queue or pool. */
+		uint32_t bp_threshold;
+	} bp;
```
psavol: "number of packets in pool" == number of free packets or allocated packets?

include/odp/api/spec/classification.h
muvarov: @bala-manoharan there was error in check script, I fixed that.

platform/linux-generic/odp_classification.c
bala-manoharan: Will do in next version.

include/odp/api/spec/classification.h
bala-manoharan: On second thought, this number of packets threshold will not work well in our case since the packet count will depend on the size of the packets and will cause an issue. It has to be a number of segments in our case and I believe NXP has an issue with this approach since they implement RED on actual packet length ignoring packet and segment meta-data.
Considering the complexity in merging this RED across the pool and queue, I would propose to revamp the API to strip this configuration from CoS and move it to pool and queue with capability exposing whether the platform supports in Queue or in Pool.
Any thought?

Bill-Fischofer-Linaro: @bala-manoharan I'll add this to Monday's ARCH call agenda. Also note that Travis is still reporting numerous failures on this PR.

bala-manoharan: @Bill-Fischofer-Linaro The build issue seems to be with pktio/dpdk.c. Anyhow for the next version of this PR I will update to the head of api-next

Travis failure log
============
pktio/dpdk.c: In function ‘dpdk_pktio_init’:
pktio/dpdk.c:1013:2: error: implicit declaration of function ‘rte_log_set_global_level’ [-Werror=implicit-function-declaration]
  rte_log_set_global_level(RTE_LOG_WARNING);
  ^
pktio/dpdk.c:1013:2: error: nested extern declaration of ‘rte_log_set_global_level’ [-Werror=nested-externs]
pktio/dpdk.c: In function ‘dpdk_pktio_term’:
pktio/dpdk.c:1068:2: error: implicit declaration of function ‘RTE_ETH_FOREACH_DEV’ [-Werror=implicit-function-declaration]
  RTE_ETH_FOREACH_DEV(port_id) {
  ^
pktio/dpdk.c:1068:2: error: nested extern declaration of ‘RTE_ETH_FOREACH_DEV’ [-Werror=nested-externs]
pktio/dpdk.c:1068:31: error: expected ‘;’ before ‘{’ token
  RTE_ETH_FOREACH_DEV(port_id) {
                               ^
pktio/dpdk.c: At top level:
pktio/dpdk.c:1056:13: error: ‘dpdk_mempool_free’ defined but not used [-Werror=unused-function]
 static void dpdk_mempool_free(struct rte_mempool *mp, void *arg ODP_UNUSED)
             ^
cc1: all warnings being treated as errors
make[1]: *** [pktio/dpdk.lo] Error 1
make[1]: *** Waiting for unfinished jobs....
make[1]: Leaving directory `/home/travis/build/Linaro/odp/platform/linux-generic'
make: *** [all-recursive] Error 1

NikhilA-Linaro:
I agree with bala here to seggregate the functionality across pool and queues as introducing to CoS is adding unnecessary complexity and still the solution will not be inter-operable.

bala-manoharan:
This PR is closed. This requirement is now handled by PR 277
https://github.com/Linaro/odp/pull/277

PR #180, opened 2017-09-15
[PATCH v6] Pkt debug

Bill-Fischofer-Linaro:
Restore ODP_ASSERT() calls in packet processing lost in the recent restructure. Also correct a syntax glitch in the previous merge that somehow got past checkpatch.

### DIFF
--- a/platform/linux-generic/odp_packet.c
+++ b/platform/linux-generic/odp_packet.c
@@ -352,6 +352,16 @@ static inline void link_segments(odp_packet_hdr_t *pkt_hdr[], int num)
 			hdr->buf_hdr.seg[i].hdr  = buf_hdr;
 			hdr->buf_hdr.seg[i].data = buf_hdr->base_data;
 			hdr->buf_hdr.seg[i].len  = seg_len;
+
+			/* init_segments() handles first seg ref_cnt init */
+			if (ODP_DEBUG == 1 && cur > 0) {
+				uint32_t prev_ref =
+					odp_atomic_fetch_inc_u32(
+						&pkt_hdr[cur]->buf_hdr.ref_cnt);
+
+				ODP_ASSERT(prev_ref == 0);
+			}
+
 			cur++;
 
 			if (cur == num) {
@@ -380,6 +390,13 @@ static inline void init_segments(odp_packet_hdr_t *pkt_hdr[], int num)
 	/* Defaults for single segment packet */
 	hdr->buf_hdr.seg[0].data = hdr->buf_hdr.base_data;
 	hdr->buf_hdr.seg[0].len  = seg_len;
+
+	if (ODP_DEBUG == 1) {
+		uint32_t prev_ref =
+			odp_atomic_fetch_inc_u32(&hdr->buf_hdr.ref_cnt);
+
+		ODP_ASSERT(prev_ref == 0);
+	}
 
 	if (!CONFIG_PACKET_SEG_DISABLED) {
 		hdr->buf_hdr.segcount = num;
@@ -846,6 +863,8 @@ void odp_packet_free(odp_packet_t pkt)
 	odp_packet_hdr_t *pkt_hdr = packet_hdr(pkt);
 	int num_seg = pkt_hdr->buf_hdr.segcount;
 
+	ODP_ASSERT(buffer_ref(&pkt_hdr->buf_hdr) > 0);
+
 	if (odp_likely(CONFIG_PACKET_SEG_DISABLED || num_seg == 1)) {
 		odp_buffer_hdr_t *buf_hdr[2];
 		int num = 1;
@@ -874,6 +893,8 @@ void odp_packet_free_multi(const odp_packet_t pkt[], int num)
 	for (i = 0; i < num; i++) {
 		odp_packet_hdr_t *pkt_hdr = packet_hdr(pkt[i]);
 		int num_seg = pkt_hdr->buf_hdr.segcount;
+
+		ODP_ASSERT(buffer_ref(&pkt_hdr->buf_hdr) > 0);
 
 		if (odp_unlikely(num_seg > 1)) {
 			free_all_segments(pkt_hdr, num_seg);
@@ -1055,6 +1076,8 @@ void *odp_packet_push_tail(odp_packet_t pkt, uint32_t len)
 	if (len > pkt_hdr->tailroom)
 		return NULL;
 
+	ODP_ASSERT(odp_packet_has_ref(pkt) == 0);
+
 	old_tail = packet_tail(pkt_hdr);
 	push_tail(pkt_hdr, len);
 
@@ -1069,6 +1092,8 @@ int odp_packet_extend_tail(odp_packet_t *pkt, uint32_t len,
 	uint32_t tailroom  = pkt_hdr->tailroom;
 	uint32_t tail_off  = frame_len;
 	int ret = 0;
+
+	ODP_ASSERT(odp_packet_has_ref(*pkt) == 0);
 
 	if (len > tailroom) {
 		pool_t *pool = pkt_hdr->buf_hdr.pool_ptr;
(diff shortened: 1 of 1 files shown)


muvarov:
subject has to be "correct syntax". Interesting that separate checkpatch run passed but inside travis it failed.

Bill-Fischofer-Linaro:
Changes for v2:

* Change commit subject to avoid checkpatch flagging its own name.

platform/linux-generic/odp_packet.c
```diff
@@ -354,6 +354,8 @@ static inline void link_segments(odp_packet_hdr_t *pkt_hdr[], int num)
 			hdr->buf_hdr.seg[i].hdr  = buf_hdr;
 			hdr->buf_hdr.seg[i].data = buf_hdr->base_data;
 			hdr->buf_hdr.seg[i].len  = BASE_LEN;
+			ODP_ASSERT((odp_atomic_store_u32(&hdr->buf_hdr.ref_cnt,
```
muvarov: Bill, odp_atomic_store_u32() is void. How assert() works here?

Bill-Fischofer-Linaro: It's a compound statement. One of the clever bits of Petri's rework is he doesn't set the initial `ref_cnt` to 1 but rather leaves it as 0 and when the first reference happens bumps it directly to 2. For debugging purposes we want to ensure that the `ref_cnt >= 1` on any free so we need the initial `ref_cnt` to be set to 1, but only for debug builds. The compound statement takes advantage of the fact that in C the "value" of a compound statement is the value of the last component, in this case 1. So the `ODP_ASSERT()` always succeeds and the effect is that the `ref_cnt` is set to 1 only if we're running with `--enable-debug`.

lumag: This is bad, bad, bad idea. Contents of ODP_ASSERT can disappear. So `odp_atomic_store_u32` won't be called.

lumag: @muvarov No, it unrolls to ` ... && (void_func(), 1) ...`. Comma operator.
@Bill-Fischofer-Linaro such asserts need to be reworked.

muvarov: @lumag yep, but what is (void_func(), 1) ? Why gcc compiles that?

Bill-Fischofer-Linaro:
C is an expression language. What's being used here is the comma operator. See (https://stackoverflow.com/questions/52550/what-does-the-comma-operator-do)[here] for a discussion of this operator and its uses.

So this expands to `if (... && (void_func(), 1))` which always evaluates to 1, so the `ODP_ASSERT` never fails. But the expression is evaluated (hence the void function is called) if we are in `--enable-debug` mode. 

The compiler must always evaluate all of the comma-separated parts of the expression, which is why you can write things like:

`for (i = 0, j = 0; i < n; i++) ...`

And be confident that both `i` and `j` will be initialized.

The reason for doing this is to avoid having `#ifdefs` in the code to handle debug-only code since `ODP_ASSERT()` does that automatically. It's a common C idiom in preprocessor macros for the same reason.

muvarov:
@lumag what did you mean by "@Bill-Fischofer-Linaro such asserts need to be reworked." ? Can I merge it now?

@Bill-Fischofer-Linaro, thanks for explanation. It's really expression and does not matter void or not void it is.

lumag:
@muvarov @Bill-Fischofer-Linaro I meant that in my opinion ODP_ASSERT() should contain only conditional code, not functional code (like atomic stores). In fact it would be preferable from my POV to have them under `#ifdef`, if they are really needed only for debugging, with proper comments why. Currently such statements can be easily misunderstood by people reading the code or doing further refactoring.

Bill-Fischofer-Linaro:
@lumag I can add some comments to explain what's going on with those "unusual" `ODP_ASSERT()` calls, but we don't like `#ifdef`s in open code. `ODP_ASSERT()` is used to encapsulate code that is only executed if `--enable-debug` is in force.

Bill-Fischofer-Linaro:
Changes for v3:

* Rebased to latest master
* Added comments to make it clear what's going on with "unusual" `ODP_ASSERT()` calls and why.

platform/linux-generic/odp_packet.c
```diff
@@ -354,6 +354,15 @@ static inline void link_segments(odp_packet_hdr_t *pkt_hdr[], int num)
...
 			hdr->buf_hdr.seg[i].len  = BASE_LEN;
+
+			/* Ensure that ref_cnt is set to 1 for debug builds.
+			 * This makes use of the comma operator to force
+			 * the store only for debug builds since the expression
+			 * (void_func(), 1) always evaluates to 1 so the
+			 * ODP_ASSERT always succeeds.
+			 */
+			ODP_ASSERT((odp_atomic_store_u32(&hdr->buf_hdr.ref_cnt,
+							 1), 1));
```
psavol: Maybe it's better not to change ref_cnt book keeping for debug build. Change of behavior may hide the problem being debugged, etc.

Bill-Fischofer-Linaro: If we want to catch erroneous duplicate packet free calls I don't see how we can avoid that. That's a main source of application errors, especially when references are involved.

The lazy-increment shortcut is a clever efficiency trick for normal processing, but it's counter to what we want in a debug build.

Bill-Fischofer-Linaro:
This should be held until we decide if we're going to merge PR #221. If we do I'll resubmit this since it currently references `odp_packet_unshared_len()` in many of the `ODP_ASSERT()` calls.

platform/linux-generic/odp_packet.c
psavol: I'd expect that a debug build adds checks and printf, but does not change e.g. how an algorithm works.

Also store is not enough for double frees, you'd need to inc/dec and check that the value never gets <0.

Bill-Fischofer-Linaro: @psavol The double free guard is the `ODP_ASSERT(ref_cnt > 0)`checks  in `odp_packet_free()` and `odp_packet_free_multi()`. Incrementing and decrementing is already handled by the mainline code. This isn't changing how reference counts are used as the mainline code works the same. It's only removing the lazy increment optimization.

Bill-Fischofer-Linaro:
Changes for v3:

- Rebased to current master

- Removed references to `odp_packet_unshared_len()`

- Switched to use `odp_atomic_fetch_inc_u32()` for `ref_cnt` initialization. This also eliminates the obscure use of the comma operator in `ODP_ASSERT()` processing.

muvarov:
I'm ok with v3. Does somebody else wants to add their review? @lumag ?

lumag:
@muvarov I stated explicitly, that I do not like functional code inside ODP_ASSERT() arguments. In my opinion this PR should rewritten with `#ifdef's`.

muvarov:
@lumag I think Bill removed "," operator. https://github.com/Linaro/odp/pull/180/commits/0768f9527a52768305af75aba4d2ccf300b63705 So it looks like only check for values which is very common to other code.

lumag:
@muvarov It still results in code doing refcounting increment under `ODP_ASSERT()` clause. In my opinion this should be better written as
```c
/* Increment counter only in debug case, because ........ */
if (ODP_DEBUG == 1) {
  uint32_t old_cnt = odp_atomic_fetch_inc_u32(&hdr->buf_hdr.ref_cnt);
  ODP_ASSERT(old_cnt == 0); 
}
```

muvarov:
@lumag he can not do that. ODP_DEBUG is implementation thing. We can not use it inside tests.

lumag:
@muvarov which tests? Commit is changing the `platform/linux-generic/odp_packet.c` file.

Bill-Fischofer-Linaro:
Changes in v5:

* Rebased to latest master
* Changed `ODP_ASSERT()` to reflect review comments by @lumag

lumag:
@Bill-Fischofer-Linaro Hmm, I don't see any significant changes.

Bill-Fischofer-Linaro:
@lumag You asked that the `== 0` not be split across lines. That's the change in v5.

Bill-Fischofer-Linaro:
Changes for v6:

* Rebased to current master
* Restructured ref_cnt initialization to avoid side effects in `ODP_ASSERT()` calls.

lumag:

muvarov:
Merged.

PR #181, opened 2017-09-19
[PATCH v2] travis: doxygen 1.8.13

muvarov:
Update doxygen to 1.8.13

After we fix warnings we can threat warnings as errors. That option is available only in later doxygen releases.

### DIFF
--- a/.travis.yml
+++ b/.travis.yml
@@ -247,8 +247,8 @@ jobs:
                           # doxygen does not trap on warnings, check for them here.
                           - ./bootstrap
                           - ./configure
-                          - make doxygen-doc |tee doxygen.log
-                          - fgrep -rvq warning ./doxygen.log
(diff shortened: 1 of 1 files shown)


lumag:
Could you please add it to cache?

muvarov:
@lumag to add it to cache we need to install doxygen to some directory and add to odp config option to specify where is doxygen. Cache will save about 1 minute of execution. Probably there is some m4 macros already exist to point to installed doxygen directory.

muvarov:
@lumag @Bill-Fischofer-Linaro  doxygen m4 macros does not have option to specify directory of where is it installed. I will try to play with it, maybe simple path will be enough.

lumag:
Looks good now, thank you!
Reviewed-by: Dmitry Eremin-Solenikov <dmitry.ereminsolenikov@linaro.org>

muvarov:
Merged.

PR #182, opened 2017-09-19
[PATCH API-NEXT v1] linux-generic: fix GCC7 build error in sched_cb_pktin_poll_one

WonderfulVoid:
Resolve Bug https://bugs.linaro.org/show_bug.cgi?id=3246 by
adding missing curly braces around if-clause.

For builds which don't flag this as an error, the resulting
build would be functional but with degraded performance.

Signed-off-by: Ola Liljedahl <ola.liljedahl@arm.com>
Reviewed-by: Honnappa Nagarahalli <honnappa.nagarahalli@arm.com>


muvarov:
Merged.

PR #183, opened 2017-09-19
[PATCH v1] configure: "best effort" approach for CUnit and validation tests

lumag:
Allow configure to succeed (by disabling validation testsuite), if CUnit
library was not found or if user did not specify --enable-test-vald
explicitly).


muvarov:
Merged.

PR #185, opened 2017-09-20
[PATCH API-NEXT v13] Ipsec doc

Bill-Fischofer-Linaro:
Add User Guide updates for IPsec support.


Bill-Fischofer-Linaro:
Changes for v2:

* Miscellaneous typo and spilling corrections in the User Guide documentation update.

Bill-Fischofer-Linaro:
Checkpatch failure on Commit 1 seems to be a false positive as the .svg file is encoded in UTF-8 format.

doc/users-guide/users-guide-ipsec.adoc
```diff
@@ -0,0 +1,410 @@
...
+
+* The IPsec operation modes supported by this implementation. Different
+operation modes may be _not supported_, _supported_, or _preferred_. A
+preferred form means that this mode takes advantage of hardware
+acceleration features to achieve best performance.
+* Whether IPsec AH processing is supported. All ODP platforms must provide
+support for IPsec ESP processing, however since AH is relatively rare, it
+may not be supported, or supported only via software emulation (_e.g.,_ be
+non-preferred).
+* Whether IPsec headers may be retained on decrypt.
```
lumag: ... in inbound inline operations.

doc/users-guide/users-guide-ipsec.adoc
```diff
@@ -0,0 +1,410 @@
...
+Synchronous processing may be preferred if the application has a large number
+of worker threads so that blocking any individual worker while IPsec processing
+is performed represents a reasonable design. The alternative is to use
+_asynchronous_ forms of these APIs:
+
+* `odp_ipsec_in_enq()` for decrypt
+* `odp_ipsec_out_enq()` for encrypt
+
+These simply pass packets to IPsec for processing. When this processing
+is complete, an IPsec completion event is sent back to the completion queue
```
lumag: ... resulting packet is sent back serving as IPsec completion event ...

doc/users-guide/users-guide-ipsec.adoc
```diff
@@ -0,0 +1,410 @@
...
+associated with the SA used by the operation as shown here:
+
+image::ipsec-lookaside.svg[align="center"]
+
```
lumag: If SA was not determined (because SA lookup failed for inbound packet), event will be sent to the default queue.

doc/users-guide/users-guide-ipsec.adoc
```diff
@@ -0,0 +1,410 @@
...
+image::ipsec-inline.svg[align="center"]
+
+On the receive side, once configured for inline processing, arriving IPsec
+packets are recognized at the PktIO interface and decrypted automatically
+before the application ever sees them. On the transmit side, the application
+calls `odp_ipsec_out_inline()` and the packet is encrypted and queued for
+transmission as a single operation without further application involvement.
+
+On the receive side, after an IPsec packet is decrypted, it may be
```
lumag: TBD: describe that some IPsec packets still might be reported via plain PktIO interface (e.g. because of SA lookup failure). They can be resubmitted to IPsec in lookaside mode.

doc/users-guide/users-guide-ipsec.adoc
```diff
@@ -0,0 +1,410 @@
...
+Security Associations it will use, and sets inbound and outbound
+processing options.
+
+==== IPsec Inbound Configuration
+Inbound configuration options for IPsec specify the default `odp_queue_t` to
+be used for processing global events like SA lookup failures, how Security
+Parameter Index (SPI) lookup is to be performed, and whether the application
+requires ODP to retain outer headers for decrypted IPsec packets.
+
+Parsing options specify how "deep" decrypted packets are to be parsed by
```
lumag: ... parsed after IPsec processing ...

doc/users-guide/users-guide-ipsec.adoc
```diff
@@ -0,0 +1,410 @@
...
+specifying the packet layers of interest to the application (None, L2, L3, L4,
```
lumag: L2 does not make sense here, does it?

doc/users-guide/users-guide-ipsec.adoc
```diff
@@ -0,0 +1,410 @@
...
+Association (SA)_. Similar to a crypto session, the SA encapsulates the keying
+material and context needed to perform IPsec protocol processing for inbound
+or outbound packets on a given flow, as well as additional processing options
+that control how IPsec is to be used for packets processed under this
+SA. Security Associations are unidirectional (RX or TX) so a flow that
+requires both inbound (decrypt) and outbound (encrypt) IPsec functions will
+have two SAs associated with it.
+
+After ODP initialization, IPsec support is dormant until it is configured
+by a call to `odp_ipsec_config()` as described earlier. Once configured,
```
lumag: TBD: mention that it MUST be called at most once per IPsec application.

doc/users-guide/users-guide-ipsec.adoc
```diff
@@ -0,0 +1,410 @@
...
+`odp_queue_deq()` if the completion queue was created as a plain queue, or
+processed via the ODP scheduler if the completion queue was created as a
+scheduled queue.
+
+==== Inline Processing
+While lookaside processing offers flexibility, it still requires extra
+processing steps that modern hardware is able to avoid. To avoid this overhead
+ODP also offers _inline_ processing support for IPsec. In this mode the
+processing of IPsec packets on the RX and TX paths is fully offloaded as
+shown here:
```
lumag: It is worth mentioning that depending on the implementation and application needs, inline processing might be enabled either in both directions or in just one direction.

doc/users-guide/users-guide-ipsec.adoc
```diff
@@ -0,0 +1,410 @@
...
+This struct allows hard and/or soft limits to be specified in terms of total
+bytes encrypted/decrypted, total packet count, or both. A limit specification
+of 0 indicates no limit for that metric. If either is specified, the limit
+is triggered on whichever occurs first. Given the defined behavior of hard vs.
+soft limits, the soft limits, if used, should always be specified as lower
+than the hard limits. These should be sufficiently lower to enable adequate
+time to switch over to a replacement SA before the hard limit is reached.
+
+As noted, when an SA hard limit is reached the SA immediately enters the
+disabled state and attempts to use it further are failed with an
```
lumag: Ahem. It does not enter disabled state per se:
- It is an undefined behaviour (iow, an application error) to submit packets to disabled SA
- It is a perfectly valid to submit packets to SA after hard limit overrun (e.g. because other packets might be already queued at this moment).

doc/users-guide/users-guide-ipsec.adoc
```diff
@@ -0,0 +1,410 @@
...
+`odp_ipsec_result_t` that indicates a hard expiration limit.  When the soft
+limit is reached, this results in an `ODP_EVENT_IPSEC_STATUS` event being sent
+to the application on the default queue associated with the SA that has
+reached the soft limit. This status event has an `odp_ipsec_status_id_t` of
```
lumag: Only in OUT INLINE mode, if I remember the outcome of discussions correctly.

doc/users-guide/users-guide-ipsec.adoc
Bill-Fischofer-Linaro: It's part of the `enum`. In this case L2 would effectively be None.

doc/users-guide/users-guide-ipsec.adoc
Bill-Fischofer-Linaro: I thought the restriction is that you can call this repeatedly as long as an SA hasn't yet been created. I can change this (and the state diagram) if that's not the case.

doc/users-guide/users-guide-ipsec.adoc
Bill-Fischofer-Linaro: If you let yourself run out of gas, the car can stop at inconvenient times, which is why one pays attention to that low fuel warning light. A hard limit is a hard limit. That's what makes it hard. Any other definition seems confusingly fuzzy and unnecessary.

doc/users-guide/users-guide-ipsec.adoc
Bill-Fischofer-Linaro: So how is this indicated in lookaside mode? The whole point of ODP providing the limit support is so the application doesn't have to track byte/packet counts itself, so it's expected that soft limit overruns will happen as part of lookaside processing as well.

Bill-Fischofer-Linaro:
Changes in v3:

* Rebased
* Incorporate review feedback from @lumag except for those items for which there is ongoing discussion as noted above.

doc/users-guide/users-guide-ipsec.adoc
lumag: IIRC, one can call it only once in NXP case. @NikhilA-Linaro, could you please comment?

doc/users-guide/users-guide-ipsec.adoc
lumag: Yep. It is just not a 'disabled' state, because we have separate definition for 'disabled SA'.

Bill-Fischofer-Linaro: OK, in v4 I've added a new terminal state `SA_Expired` to the FSM and have updated the doc to say "expired" rather than "disabled". From the expired state the only valid operation is `odp_ipsec_sa_destroy()`.

lumag:
> From the expired state the only valid operation is |odp_ipsec_sa_destroy()|.

I don't think so. We never stated that in API. Expired SAs should be
destroyed via normal `odp_ipsec_sa_disable()`/`odp_ipsec_sa_destroy()` pair.

doc/users-guide/users-guide-ipsec.adoc
lumag: In lookaside mode soft limit expiration is reported as `warn` part of `ipsec_op_status` packet metadata.

Bill-Fischofer-Linaro: OK, corrected in v5.

Bill-Fischofer-Linaro:
Changes in v5:

* Reworked FSM diagram to reflect correct state transitions for SA_Expired state
* Clarified soft expiration handling per @lumag comments

Bill-Fischofer-Linaro:
Note that the checkpach failure seem due to upgrading doxygen which caught a recent omission having nothing to do with this PR. I'll submit a separate PR to address that.

doc/users-guide/users-guide-ipsec.adoc
```diff
@@ -0,0 +1,428 @@
...
+is triggered on whichever occurs first. Given the defined behavior of hard vs.
+soft limits, the soft limits, if used, should always be specified as lower
+than the hard limits. These should be sufficiently lower to enable adequate
+time to switch over to a replacement SA before the hard limit is reached.
+
+As noted, when an SA hard limit is reached the SA immediately enters the
+expired state and attempts to use it further are failed with an
+`odp_ipsec_result_t` that indicates a hard expiration limit.  When the soft
+limit is reached in inline mode, this results in an `ODP_EVENT_IPSEC_STATUS`
+event being sent to the application on the default queue associated with the
```
lumag: s/default //

lumag:
@Bill-Fischofer-Linaro looks great now, thank you. Last tiny comment from my side.

doc/users-guide/users-guide-ipsec.adoc
Bill-Fischofer-Linaro: OK, changed in v6.

Bill-Fischofer-Linaro:
Changes for v7:

* Updated IPsec FSM to remove the configured->configured link to reflect that `odp_ipsec_config()` may only be called once per ODP instance.

Bill-Fischofer-Linaro:
Changes for v8:

* Rebased to latest api-next

* Updated FSM diagram and text to reflect the latest IPsec API changes proposed in PR #197.

Bill-Fischofer-Linaro:
Changes for v9

* Rebased to pick up Travis fixes.

muvarov:
@NikhilA-Linaro @lumag are you ok to merge this?

doc/users-guide/users-guide-ipsec.adoc
```diff
@@ -0,0 +1,437 @@
...
+to the `odp_ipsec_sa_destroy()` API. However, to better support graceful SA
+termination processing, ODP also provides the `odp_ipsec_sa_disable()` API.
+
+This call initiates termination processing for an SA by stopping use of the SA
+for new operations while still allowing those that are "in flight" to complete
+processing. All newly arriving packets following this call will carry a
+disabled warning in the `odp_op_status_t` section of the `odp_ipsec_result_t`
+struct. So after disabling the SA, the application can process pending packets
+normally until it sees packets marked with the disabled warning. At this point
+it knows that all pending packets that arrived before the disable have been
```
lumag: Well, we do not have #197 merged, so it is too early to depend on that.
Also, frankly speaking, this `sa_disabled` warning inside `odp_ipsec_result_t` is a backup plan. It is expected that most of the implementations will report this as a proper `ODP_IPSEC_STATUS` event, carrying this warning bit inside.

doc/users-guide/users-guide-ipsec.adoc
lumag: Any update here? @NikhilA-Linaro, @bala-manoharan  can you please comment wrt your implementations?

doc/users-guide/users-guide-ipsec.adoc
Bill-Fischofer-Linaro: I'd like to see us all make it a priority to close on #197 next week. I'll post a doc update once we're all agreed on the precise semantics and operation flow. So I'm fine with holding off merging this PR until then.

Bill-Fischofer-Linaro:
Changes for v10:

* Rebased to latest api-next
* Document updates to reflect use of dummy packets to indicate odp_ipsec_sa_disable() completion.

Bill-Fischofer-Linaro:
Checkpatch error is spurious (checkpatch trips over .svg files even though they've been passed through xmllint).

doc/users-guide/users-guide-ipsec.adoc
```diff
@@ -0,0 +1,443 @@
...
+
+* The IPsec operation modes supported by this implementation. Different
+operation modes may be _not supported_, _supported_, or _preferred_. A
+preferred form means that this mode takes advantage of hardware
+acceleration features to achieve best performance.
+* Whether IPsec AH processing is supported. All ODP platforms must provide
+support for IPsec ESP processing, however since AH is relatively rare, it
+may not be supported, or supported only via software emulation (_e.g.,_ be
+non-preferred).
+* Whether IPsec headers should be retained on decrypt for inbound inline
```
JannePeltonen: This should be: "...can be retained..." instead of "...should be retained..." as this is a capability and application chooses whether it wants to retain the headers if the ODP is capable.

doc/users-guide/users-guide-ipsec.adoc
```diff
@@ -0,0 +1,443 @@
...
+
+* Parameters controlling the SA lifetime.
+
+* The Security Parameter Index (SPI) that packets will use to indicate that
+they belong to this SA.
+
+* The pipeline mode used by this SA.
+
+* The destination `odp_queue_t` to be used for status events associated with
+this SA.
```
JannePeltonen: The queue is not only for status events but for ipsec packet events too. Maybe just remove the word "status"?

doc/users-guide/users-guide-ipsec.adoc
```diff
@@ -0,0 +1,443 @@
...
+
+* The user context pointer (and length) associated with this SA for
+application use.
+
+In addition, there are specific direction-specific parameters that vary
+based on whether the SA is for inbound or outbound use. For inbound SAs:
+
+* Controls for how this SA is to be looked up.
+
+* The size of the anti-replay window to be used.
```
JannePeltonen: Would "minimum size" be better than "size" here?

doc/users-guide/users-guide-ipsec.adoc
```diff
@@ -0,0 +1,443 @@
...
+
+==== SA Lifecycle Management
+In discussing the lifecycle of an SA, it is useful to refer to the following
+state diagram:
+
+image::ipsec_fsm.svg[align="center"]
+
+After creation, IPsec services are active for this security association. How
+this is done depends on the IPsec operating mode being used
+
```
JannePeltonen: "How this is done" sounds a bit awkward as it is not clear what "this" refers to (the way the IPsec services are used). Maybe the sentence should be reworded.

doc/users-guide/users-guide-ipsec.adoc
```diff
@@ -0,0 +1,443 @@
...
+===== IPsec Lookaside Processing
+If the SA is operating in lookaside mode (the `odp_ipsec_mode_t` is
+`ODP_IPSEC_OP_MODE_SYNC` or `ODP_IPSEC_OP_MODE_ASYNC`), then inbound or
+outbound lookaside operations may be performed.
```
JannePeltonen: The operating mode is a global setting, not per-SA setting as the text implies. And the relevant type is odp_ipsec_op_mode_t, not odp_ipsec_mode_t which is for selecting between tunnel and transport mode.

doc/users-guide/users-guide-ipsec.adoc
```diff
@@ -0,0 +1,443 @@
...
+The synchronous forms of these APIs are:
+
+* `odp_ipsec_in()`
+* `odp_ipsec_out()`
+
+Upon return from these calls, the return code tells the application whether
+the application was successful and if not the reason is contained in the
+`odp_ipsec_result_t` struct that is attached to the failing packet. This
+struct is retrieved via the `odp_ipsec_result()` API.
+
```
JannePeltonen: The return code does not tell whether the application of the SA was successful but how many input packets were consumed by the operation. The per-packet operation status is retrieved via the odp_ipsec_result() api for the outputted packets, which will be available only when the function succeeds.

doc/users-guide/users-guide-ipsec.adoc
```diff
@@ -0,0 +1,443 @@
...
+The asynchronous forms of these APIs are:
+
+* `odp_ipsec_in_enq()`
+* `odp_ipsec_out_enq()`
+
+Here, a return code of 0 simply indicates that the requested IPsec operation
+was successfully initiated.
```
JannePeltonen: Same comment as for odp_ipsec_{in,out}(), the description of the return code is not correct.

doc/users-guide/users-guide-ipsec.adoc
```diff
@@ -0,0 +1,443 @@
...
+packet(s) are non-IPsec packets. Again the option is to use a single SA for
+all input packets or one per input packet.
+
+For outbound operations, an associated array of `odp_ipsec_out_opt_t` structs
+is also used to control the fragmentation mode to be used as part of the
+outbound processing. Options here are to not fragment, to fragment before
+IPsec processing, after IPsec processing, or to only check whether IP
+fragmentation is needed but not to perform it. The `mtu` status error bit
+in the `odp_ipsec_packet_result_t` is set if check processing detects that
+the resulting packet will not fit into the configured MTU. Note that the MTU
```
JannePeltonen: Is it clear enough here that the mtu error bit is set only when MTU checking was requested and not in case fragmentation offload was requested?

doc/users-guide/users-guide-ipsec.adoc
```diff
@@ -0,0 +1,443 @@
...
+limit is reached for packets sent via `odp_ipsec_out_inline()`, this results
+in an `ODP_EVENT_IPSEC_STATUS` event being sent to the application on the
+queue associated with the SA that has reached the soft limit. This status
+event has an `odp_ipsec_status_id_t` of `ODP_IPSEC_STATUS_WARN` with a
+`odp_ipsec_warn_t` bits set to indicate the type of soft expiration reached.
+Receipt of this event alerts the application that the SA is nearing the end of
+its useful life and that it should be replaced. It is the application's
+responsibility to heed this warning. Soft limit alerts are edge triggered and
+are only issued once rather than repeatedly as the SA runs beyond the soft
+limit.
```
JannePeltonen: According to the API spec soft limit alerts do not need to be edge triggered. The spec says: "It's implementation defined how many times soft lifetime expiration is reported: only once, first N or all packets following the limit crossing."

doc/users-guide/users-guide-ipsec.adoc
```diff
@@ -0,0 +1,443 @@
...
+
+When operating in lookaside mode, expiration limits are carried as a warning
+in the `odp_op_status_t` section of the `odp_ipsec_result_t` struct. The same
+is true for inline inbound packets. When the soft limit is reached, these
+packets will carry a warning flag indicating this condition.
+
+==== SA Disablement and Destruction
+When it is time to retire an SA, the application does so by issuing a call
+to the `odp_ipsec_sa_destroy()` API. However, to better support graceful SA
+termination processing, ODP also provides the `odp_ipsec_sa_disable()` API.
```
JannePeltonen: I think this may give the impression that calling odp_ipsec_sa_disable() is optional, when in fact it is mandatory before odp_ipsec_sa_destroy().

doc/users-guide/users-guide-ipsec.adoc
```diff
@@ -0,0 +1,443 @@
...
+
+This call initiates termination processing for an SA by stopping use of the SA
```
JannePeltonen: "This call" could be replaced with odp_ipsec_sa_disable() to make the text more clear.

doc/users-guide/users-guide-ipsec.adoc
```diff
@@ -0,0 +1,443 @@
...
+for new operations while still allowing those that are "in flight" to complete
+processing. A special "dummy" packet with an `odp_ipsec_result_t` that says
+this SA is now disabled indicates that the packet pipeline for this SA is now
+clear. Note that these dummy packets will have an `odp_packet_len()` of zero
+and should not be used as actual packet data. Upon receipt of such dummy
+packets the application should simply take appropriate action and then free
+the event.
```
JannePeltonen: This dummy packet stuff is not yet (if ever?) part of the API spec. I think this patch should be in sync with what is the current state of api-next, which still uses the status event for indicating SA disable completion.

doc/users-guide/users-guide-ipsec.adoc
```diff
@@ -0,0 +1,443 @@
...
+
+This process may be performed _synchronously_ with the APIs `odp_ipsec_in()`
+(to decrypt) and `odp_ipsec_out()` (to encrypt). Upon return from these calls
+the requested packet transformation is complete, or an error return code
+indicates that it could not be performed (_e.g.,_ packet decryption failed).
+
+Synchronous processing may be preferred if the application has a large number
+of worker threads so that blocking any individual worker while IPsec processing
+is performed represents a reasonable design. The alternative is to use
+_asynchronous_ forms of these APIs:
```
JannePeltonen: Maybe it should be made more clear somewhere that the synchronous processing calls are allowed only in the synchronous operating mode and asynchronous calls in asynchronous and inline operation mode (and inline calls only in the inline mode).

doc/users-guide/users-guide-ipsec.adoc
```diff
@@ -0,0 +1,443 @@
...
+
+* The Security Parameter Index (SPI) that packets will use to indicate that
+they belong to this SA.
+
+* The pipeline mode used by this SA.
+
+* The destination `odp_queue_t` to be used for status events associated with
+this SA.
+
+* The user context pointer (and length) associated with this SA for
```
JannePeltonen: Maybe it is not useful to mention length here since it is only a prefetching hint.

doc/users-guide/users-guide-ipsec.adoc
Bill-Fischofer-Linaro: This version of the PR was intended to match the changes in PR #256. I'll revert these in the next version since it seems we're now content with the original semantics.

Bill-Fischofer-Linaro:
Changes in v11:

- Rebased to latest api-next
- Revert changes that were associated with PR #256 
- Incorporate review comments from @JannePeltonen (thanks, Janne!)

doc/users-guide/users-guide-ipsec.adoc
Bill-Fischofer-Linaro: Clarified in v12. Thanks.

Bill-Fischofer-Linaro:
Changes in v12:

- Inadvertently omitted some changes in response to @JannePeltonen review in v11, so all are now reflected in v12.

doc/users-guide/users-guide-ipsec.adoc
```diff
@@ -0,0 +1,459 @@
...
+will be discussed further later.
+
+==== IPsec Operation Modes
+IPsec operates in one of three modes: Synchronous, Asynchronous, and Inline.
+
+==== Lookaside Processing
+Synchronous and Asynchronous are types of _lookaside_ processing. Which of
+these forms may be used depends on the IPsec operation mode. So synchronous
+APIs may only be used when operating in synchronous mode, and asynchronous
+APIs may only be used when operating in asynchronous mode.
```
JannePeltonen: Asynchronous look-a-side APIs may also be used in the inline mode. It is needed for cases where inline processing of certain packets is not possible, e.g. due to them being encapsulated with some other tunneling protocol in addition to IPsec.

doc/users-guide/users-guide-ipsec.adoc
```diff
@@ -0,0 +1,459 @@
...
+
+image::ipsec_fsm.svg[align="center"]
+
+After creation, IPsec services are active for this Security Association.  The
+specific APIs that can be used on this SA depends on the IPsec operating mode
+that has been configured.
+
+===== IPsec Lookaside Processing
+If IPsec is operating in lookaside mode for the SA's direction (the
+`odp_ipsec_mode_t` is `ODP_IPSEC_OP_MODE_SYNC` or `ODP_IPSEC_OP_MODE_ASYNC`),
```
JannePeltonen: The type of the inbound_mode and outbound_mode config settings is odp_ipsec_op_mode_t.

doc/users-guide/users-guide-ipsec.adoc
```diff
@@ -0,0 +1,459 @@
...
+}
+-----
+
+===== IPsec Inline Processing
+When IPsec is configured to operate in `ODP_IPSEC_OP_MODE_INLINE` mode,
+inbound processing is implicit. The application never sees these packets until
+after IPsec has already decrypted them. As shown in the code sketch above,
+such packets appear as events of subtype `ODP_EVENT_PACKET_IPSEC` and the
+`flag` field in the associated `odp_ipsec_packet_result_t` indicates
+`inline_mode`.
```
JannePeltonen: May it should be mentioned somewhere here for clarity that also asynchronous look-a-side operations are allowed in the inline mode.

doc/users-guide/users-guide-ipsec.adoc
Bill-Fischofer-Linaro: Good catch. Fixed in v13.

Bill-Fischofer-Linaro:
Changes in v13:

- Incorporate additional review comments from @JannePeltonen

muvarov:
@JannePeltonen @lumag are you ok with the latest version?

JannePeltonen:
The text looks ok to me now, but the SA state machine picture does not. I'll add a separate comment on it.

doc/images/ipsec_fsm.gv
```diff
@@ -0,0 +1,32 @@
+digraph ipsec_state_machine {
```
JannePeltonen: I do not quite understand the SA state machine picture, maybe because I think it tries to include different kind of states (global IPsec subsystem state, IPsec SA state, IPsec operation state) in the same picture.

The operation related states (Processing, Op_Complete) are not SA states but states of individual packet operations. There can be multiple packet operations ongoing and in different states at the same time for the same SA.

I can identify only the following four SA states:

1) Nonexistent: The SA does not exist
- odp_ipsec_sa_create() changes the SA state to SA_Ready.

2) SA_Ready: The SA can be used for processing packets
- odp_ipsec_sa_disable() changes the SA state to Disable_Pending

3) Disable_Pending: The SA is processing in-flight packets but cannot anymore be passed as a parameter to any new IPsec packet operation.
- The ODP implementation removes the SA from inbound SA lookup (if the SA is an inbound SA that offloads SA lookup) and completes processing the in-flight packets for the SA and enqueues the operation completion IPsec packet events. After that the ODP implementation enqueues a disable complete SA status event and then the SA state changes to SA_Disabled.

4) SA_Disabled: The SA has been disabled.
- No packets are processed by the SA and no new IPsec packet or status events for the SA appear.
- odp_ipsec_sa_destroy() destroys the SA fully and changes the SA state to Nonexistent.

Then one could define SA_expired and SA_not_expired sub-states for some of the states (SA_Ready and Disable_Pending), but the lifetime related states could also be seen as a separate FSM, mostly independent from the main SA FSM.

Bill-Fischofer-Linaro: You're right, the diagram is trying to represent the IPsec subsystem as well as individual SAs. I can separate it out into multiple FSMs if that would be clearer. The point is to show which ODP APIs are valid at various stages of processing. So, for example, once an SA is expired you cannot initiate new IPsec operations with it, etc.

muvarov:
I merged this PR. As agreed diagram will be updated by @Bill-Fischofer-Linaro in separete PR.

PR #191, opened 2017-09-25
[PATCH v1] travis: correct doxygen warning capture

muvarov:
use if else bash syntax for grep warnings from
log.


muvarov:
Merged

PR #192, opened 2017-09-25
[PATCH v2] linux-gen: _fdserver return code cleanup

muvarov:
somewhere we have return (-1) somewhere -1, let's
make it common.


muvarov:
Merged

PR #193, opened 2017-09-25
[PATCH API-NEXT v3] Ones complement sum API

psavol:
This series adds ones complement sum function for IP/UDP/TCP checksum generation, updating and checking. Application may use it e.g. when packet IO is not able to calculate the sum (e.g. for fragments). Initial implementation is not optimized for performance (needed for validation suite development).


platform/linux-generic/odp_chksum.c
```diff
@@ -0,0 +1,36 @@
...
+
+		*(uint8_t *)&left_over = *(const uint8_t *)data;
+		sum += left_over;
+	}
+
+	/* Fold 32-bit sum to 16 bits */
+	while (sum >> 16)
+		sum = (sum & 0xffff) + (sum >> 16);
+
+	return sum;
```
Bill-Fischofer-Linaro: This should be `return ~sum;`. See [RFC 1071](https://tools.ietf.org/html/rfc1071).

Bill-Fischofer-Linaro: You should also rebase this to pick up Maxim's latest doxygen fix.

muvarov: For record put here link here which Illias found for optimization algorithm: https://locklessinc.com/articles/tcp_checksum/

Bill-Fischofer-Linaro: @muvarov The correct code is also in the Lockless example, which is taken directly from the RFC. This looks to be a simple transcription error.

psavol: This is not IP checksum function, but ones complement sum. Application does the final bit flip e.g. after summing multiple of these sums together first.

psavol: v2 is the rebase

psavol:
This has been merged.

PR #194, opened 2017-09-25
[PATCH v2] Fix check-valgrind invocation

lumag:
Fixes https://bugs.linaro.org/show_bug.cgi?id=2254

### DIFF
--- a/configure.ac
+++ b/configure.ac
@@ -99,6 +99,9 @@ CC_VERSION_PATCH=$(echo $CC_VERSION | cut -d'.' -f3)
 ##########################################################################
 # Allow valgrind suite to run against the defined tests
 ##########################################################################
+AX_VALGRIND_DFLT([sgcheck], [off])
+AX_VALGRIND_DFLT([helgrind], [off])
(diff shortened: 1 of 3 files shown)


muvarov:
make check-valgrind does not execute anything for me.

muvarov:
that change is ok,  but I think we need a little bit clean up before merging. At least exclude bash files from execution.

lumag:
@muvarov no, there is no point in removing shell scripts from execution. valgrind will check both shell and underlying programs.

muvarov:
@lumag we do not need to debug memory leaks inside bash or any other test case. I think we only take care about our test code.

lumag:
What would you prefer from:
 - Being able to valgrind tests (maybe having false positives thanks to bash)
 - Not being able to valgrind tests in automatic way?

muvarov:
@lumag yea, looks like it's trade off. I'm going to merge this and when  I will have some ideas how to put this to Travis - it will be separate patch.

muvarov:
Merged.

PR #196, opened 2017-09-26
[PATCH CLOUD-DEV v3] Cloud dev pktio ops data pr

bogdanPricope:
Add generic pktio_ops data storage for linux-generic and linux-dpdk pktios


platform/linux-dpdk/include/odp_packet_io_internal.h
```diff
@@ -42,7 +42,10 @@ typedef union pktio_entry_u pktio_entry_t;
 
 struct pktio_entry {
 	const pktio_ops_module_t *ops;	/**< Implementation specific methods */
-	pktio_ops_data_t ops_data;
```
heyi-linaro: Thanks, @bogdanPricope, previously we thought about using void * to hide implementation private data, and dpdk pktio module implemented that, what's the consideration here to unify the implementation private data as a big data array?

@nagarahalli @Bill-Fischofer-Linaro what do you think about this approach, we can have a talk in Connect?

bogdanPricope: struct pktio_entry should be independent of currently existing pktios and also to support "custom" pktios. The solution I see is to have a buffer of configurable size that can be cast to whatever a specific pktio needs.
The big size is due to pktio_ops data size needed by netmap pktio.
I prefer to have pktio_ops data in shared memory and not allocated with malloc.

Bill-Fischofer-Linaro:
@bogdanPricope @heyi-linaro Need to investigate and resolve Travis failures before merging.

bogdanPricope:
Patch re-posted against 2.0 branch as https://github.com/Linaro/odp/pull/241

PR #17, opened 2017-04-26
Fixes for  bug2969:  traffic_mngr binary was not found from shell script

muvarov:

PR #197, opened 2017-09-26
[PATCH API-NEXT v7] api: ipsec: rework ODP_IPSEC_SA_DISABLE into warning bit

lumag:
It is expected that platforms that are not able to support
odp_ipsec_sa_disable() status result in a form of separate event will
set SA hard expiry time to 0, submit a dummy packet to that SA. Then
after receiving this packet after IPsec processing (which should result
in hard expiry breach) odp_ipsec_result() will detect this packet
through the combination of hard_expiry, size, contents, etc and will
report it as a packet with odp_ipsec_warn_t->sa_disabled bit set.

Cc: Nikhil Agarwal <nikhil.agarwal@linaro.org>
Cc: Balasubramanian Manoharan <bala.manoharan@linaro.org>

### DIFF
--- a/include/odp/api/spec/ipsec.h
+++ b/include/odp/api/spec/ipsec.h
@@ -839,10 +839,16 @@ odp_ipsec_sa_t odp_ipsec_sa_create(const odp_ipsec_sa_param_t *param);
  *
  * When in synchronous operation mode, the call will return when it's possible
  * to destroy the SA. In asynchronous mode, the same is indicated by an
- * ODP_EVENT_IPSEC_STATUS event sent to the queue specified for the SA. The
- * status event is guaranteed to be the last event for the SA, i.e. all
- * in-progress operations have completed and resulting events (including status
- * events) have been enqueued before it.
+ * sa_disabled bit in odp_ipsec_warn_t. Warning can be delivered either by an
+ * ODP_EVENT_IPSEC_STATUS event sent to the queue specified for the SA or by a
+ * packet having this bit set in corresponding odp_ipsec_result_t instance,
+ * which is also sent to the SA queue. Which of the event types is used to
+ * deliver completion depends on the exact implementation details. This warning
+ * is guaranteed to be the last event for the SA, i.e. all in-progress
+ * operations have completed and resulting events (including status events)
+ * have been enqueued before it. The will be no more packets coming from SA
+ * queue. If the packet is used to deliver sa_disabled warning, there will be
+ * only one packet with this bit set.
  *
  * @param sa      IPSEC SA to be disabled
  *
@@ -923,7 +929,12 @@ typedef struct odp_ipsec_error_t {
 
 } odp_ipsec_error_t;
 
-/** IPSEC warnings */
+/** IPSEC warnings
+ *
+ * For outbound SAs in ODP_IPSEC_OP_MODE_INLINE mode warnings can be reported
+ * only as status events. In all other cases warnings can be reported either as
+ * a part of packet result or via separate ODP status event.
+ */
 typedef struct odp_ipsec_warn_t {
 	/** IPSEC warnings */
 	union {
@@ -934,6 +945,9 @@ typedef struct odp_ipsec_warn_t {
 
 			/** Soft lifetime expired: packets */
 			uint32_t soft_exp_packets : 1;
+
+			/** SA was disabled */
+			uint32_t sa_disabled : 1;
 		};
 
 		/** All warnings bits */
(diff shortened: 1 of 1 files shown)


Bill-Fischofer-Linaro:
@bala-manoharan @NikhilA-Linaro Please review this proposal, which came from discussions today. Can your HW work with this revised disable semantics?

Bill-Fischofer-Linaro:
@lumag Travis reports numerous breakages that need to be addressed.

lumag:
@Bill-Fischofer-Linaro sure, it is a proposal, not mergeable thing.

include/odp/api/spec/ipsec.h
```diff
@@ -1146,7 +1142,7 @@ typedef enum odp_ipsec_status_id_t {
 	 *  This status event is generated only for outbound SAs in
 	 *  ODP_IPSEC_OP_MODE_INLINE mode.
 	 */
-	ODP_IPSEC_STATUS_WARN
+	ODP_IPSEC_STATUS_WARN = 0,
```
JannePeltonen: The preceding comment says that this status event is generated only for outbound SAs in the inline mode, but I think that is not the intention anymore, at least not for the sa_disabled warning. Maybe the sentence should continue, e.g.: ...except in the case of sa_disabled warnings.

lumag: We probably should just move this phrase to exact warnings description.

NikhilA-Linaro:
1. Commit message needs update, i am not clear how it is related to SA hard expiry.
2. Dummy packet meant for SA disabled warning, may have some error falgs set as well, eg. ICV checksum failed.

lumag:
@NikhilA-Linaro 
> 1. Commit message needs update, i am not clear how it is related to SA hard expiry.

It is just one of possible ways on how to signal that this SA is in progress of being disabled. If the platform can pass such info in a different way, it is free to do so.

> 2. Dummy packet meant for SA disabled warning, may have some error falgs set as well, eg. ICV checksum failed.

It is then up to platform implementation of `odp_ipsec_result()` to detect such flags combination and set `warn.sa_disable` bit. Application then will be able to check it and determine that this is the last event generated for this SA.

Bill-Fischofer-Linaro:
With this design for inbound inline packets once `odp_ipsec_sa_disable()` is called, all new packets arriving after the call will be marked with the disabled warning bit, so the application may see more than one. The key is that it will see those arriving before the disable (that aren't marked) first, so once it sees a packet marked disabled it knows it's safe to call `odp_ipsec_sa_destroy()`.

If the flow is idle at the time of the disable call, then the implementation may inject a dummy packet so the application can see that the pipeline is clear.

bala-manoharan:
@NikhilA-Linaro The idea behind this proposal is that the implementation is able to send a dummy packet into the SA queue once the SA has been disabled. Is this design implementable in your platform?

Bill-Fischofer-Linaro:
@lumag Please rebase this to pick up the latest Travis fixes so that regressions can be properly done against this PR. Thanks.

lumag:
@Bill-Fischofer-Linaro done

Bill-Fischofer-Linaro:
@lumag Looks like Travis is still reporting across-the board failures.

psavol:
Checkpatch fails

WARNING: Duplicate signature
#17: 

total: 0 errors, 1 warnings, 0 checks, 61 lines checked

lumag:
@psavol should be fixed now

include/odp/api/spec/ipsec.h
```diff
@@ -839,8 +839,10 @@ odp_ipsec_sa_t odp_ipsec_sa_create(const odp_ipsec_sa_param_t *param);
  *
  * When in synchronous operation mode, the call will return when it's possible
  * to destroy the SA. In asynchronous mode, the same is indicated by an
- * ODP_EVENT_IPSEC_STATUS event sent to the queue specified for the SA. The
- * status event is guaranteed to be the last event for the SA, i.e. all
+ * sa_disabled bit in odp_ipsec_warn_t. Warning can be delivered either by an
+ * ODP_EVENT_IPSEC_STATUS event sent to the queue specified for the SA or by
+ * a packet having this bit set in corresponding odp_ipsec_result_t instance.
+ * This warning is guaranteed to be the last event for the SA, i.e. all
```
psavol: 1) It's specified that status event is sent to the SA queue, but it's left open which queue the packet event is sent. There should not be a difference, so both should go to the SA queue.
2) do we need two different event types ?
3) can there be multiple packet events which carry the warning bit, or only one? I think it says only one.

include/odp/api/spec/ipsec.h
```diff
@@ -1129,26 +1139,14 @@ typedef struct odp_ipsec_packet_result_t {
  * IPSEC status ID
  */
 typedef enum odp_ipsec_status_id_t {
-	/** Response to SA disable command
-	 *
-	 *  Following status event (odp_ipsec_status_t) fields have valid
-	 *  content, other fields must be ignored:
-	 *  - sa:       The SA that was requested to be disabled
-	 *  - result:   Operation result
```
psavol: int result would not be needed any more in ipsec_status_t

lumag:
>  1. It's specified that status event is sent to the SA queue, but it's
>     left open which queue the packet event is sent. There should not be
>     a difference, so both should go to the SA queue.

Added necessary text

>  2. do we need two different event types ?

I think so. There are platforms that support original SA_DISABLED
paradigm, so no need to pack that event into the packet.

>  3. can there be multiple packet events which carry the warning bit, or
>     only one? I think it says only one.

Updated text.

> int result would not be needed any more in ipsec_status_t

Dropped

lumag:
Replaced with #256

PR #199, opened 2017-09-27
[PATCH v1] test: l2fwd script: run generator on a single core

matiaselo:
Run packet generator on a single core to minimize the penalty from running
generator and l2fwd applications on overlapping cores (both use
odp_cpumask_default_worker()). On a generic server this change increases
packet rate by ~40X.

Fixes https://bugs.linaro.org/show_bug.cgi?id=2407

### DIFF
--- a/test/common_plat/performance/odp_l2fwd_run.sh
+++ b/test/common_plat/performance/odp_l2fwd_run.sh
@@ -66,12 +66,10 @@ run_l2fwd()
 		exit 1
 	fi
 
-	# Max 4 workers
-	# @todo: ensure that generator and l2fwd workers are not allocated to
-	# the same CPUs
+	# Run generator with one worker
 	(odp_generator${EXEEXT} --interval $FLOOD_MODE -I $IF0 \
 			--srcip 192.168.0.1 --dstip 192.168.0.2 \
-			-m u -w 4 2>&1 > /dev/null) \
+			-m u -w 1 2>&1 > /dev/null) \
 			2>&1 > /dev/null &
 
 	GEN_PID=$!


muvarov:
@matiaselo so it runs generator with 1 worker. And it runs l2fwd with 2 workers. In case of 1 cpu all 3 worker threads are on single cpu. In case of 2 and more cpus: - 1 cpu has worker from l2fwd and generator and 2 cpu has l2fwd worker. I think more clean fix will be to provide core mask to both l2fwd and generator. Then if cores >= 3 do assignment to cores with taskset. If  < 3 then core are overlapped and results are going to be more unpredictable so both apps need to instrument with shed_yeld() and might be  SCHED_RR policy to switch context between cores. But that looks like more complicated. What are your thoughts?

matiaselo:
> On 3 Oct 2017, at 11:12, muvarov <notifications@github.com> wrote:
> 
> @matiaselo so it runs generator with 1 worker. And it runs l2fwd with 2 workers. In case of 1 cpu all 3 worker threads are on single cpu. In case of 2 and more cpus: - 1 cpu has worker from l2fwd and generator and 2 cpu has l2fwd worker. I think more clean fix will be to provide core mask to both l2fwd and generator. Then if cores >= 3 do assignment to cores with taskset. If < 3 then core are overlapped and results are going to be more unpredictable so both apps need to instrument with shed_yeld() and might be SCHED_RR policy to switch context between cores. But that looks like more complicated. What are your thoughts?
> 
> —
> You are receiving this because you were mentioned.
> Reply to this email directly, view it on GitHub, or mute the thread.
> 

True, adding core mask arguments to l2fwd and generator applications would be a better solution. However, I won't have time to implement this until after next week the earliest (if there was a rush to get the bug report closed before next release).

-Matias

muvarov:
@matiaselo  than I will left this PR in opened state, if you fix will be ready it can replace this. If not we can go with this quick hack :)

muvarov:
Merged to master. Will go with that hack. I think I will need core mask for performance measuments then will fix it later.

PR #200, opened 2017-09-28
[PATCH API-NEXT v7] API: Pool: Return address range for pool objects

sachin-saxena:
Signed-off-by: Sachin Saxena <sachin.saxena@linaro.org>

### DIFF
--- a/include/odp/api/spec/pool.h
+++ b/include/odp/api/spec/pool.h
@@ -294,6 +294,22 @@ odp_pool_t odp_pool_lookup(const char *name);
 typedef struct odp_pool_info_t {
 	const char *name;          /**< pool name */
 	odp_pool_param_t params;   /**< pool parameters */
+
+	/** Minimum data address.
+	 * This is the minimum address that application accessible
+	 * data of any object (event) allocated from the pool may
+	 * locate. When there's no application accessible data
+	 * (e.g. ODP_POOL_TIMEOUT pools), the value may be zero.
+	 */
+	uintptr_t min_data_addr;
+
+	/** Maximum data address.
+	 * This is the maximum address that application accessible
+	 * data of any object (event) allocated from the pool may
+	 * locate. When there's no application accessible data
+	 * (e.g. ODP_POOL_TIMEOUT pools), the value may be zero.
+	 */
+	uintptr_t max_data_addr;
 } odp_pool_info_t;
 
 /**


include/odp/api/spec/pool.h
```diff
@@ -172,7 +172,20 @@ int odp_pool_capability(odp_pool_capability_t *capa);
...
 	/** Pool type */
 	int type;
-
+	/** Flag to indicate whether user is requesting for contiguous
+	    memory for buffers or packets in pool.
+	    If TRUE, then address range for buffers or packets
+	    must be returned by API as addr_start & addr_len.
+	    If FALSE, then filling these parameters is optional.
+	 */
+	odp_bool_t contig_mem;
```
lumag: Isn't setting `seg_len` enough? Maybe we should provide an updated definition for `seg_len`  (e.g. in case `seg_len > max_len`, always use single segment).

include/odp/api/spec/pool.h
```diff
@@ -172,7 +172,20 @@ int odp_pool_capability(odp_pool_capability_t *capa);
...
+	/** Start of virtual address */
+	union {
+		void *addr_start;
+		uint64_t addr_start_64;
+	};
```
lumag: What will happen for platforms where `odp_packet_t` is not mapped to the virtual memory?

Bill-Fischofer-Linaro: VPP will not run on such platforms without modification. If the goal is to not modify VPP then an `odp_pool_capability()` output that says whether pools can present a linear interface. During Connect @psavol also mentioned that a separate API to get this info would be more appropriate than bending the input to `odp_pool_create()` to give output parameters.

My personal preference would be something like:

* `odp_pool_capability()` indicates whether pools can be created that provide linear addressing.

* An option on `odp_pool_create()` to request a linear pool.

* A new API (or an extension to the existing `odp_pool_info()` API) to get the mapping info that VPP needs.

That would seem to be more consistent with overall ODP structure.

include/odp/api/spec/pool.h
Bill-Fischofer-Linaro: Yes the existing `seg_len` parameter is set to the minimum size segment that the application requires. This can be set to be equal to `max_len` to require single-segment packets.

However @sachin-saxena mentioned at SFO17 that VPP can deal with multi-segment packets, though I'm not sure how that works.

lumag: Got confused between single-segment packet and linear space for packets, sorry.

include/odp/api/spec/pool.h
lumag: @Bill-Fischofer-Linaro yes, this looks like a good approach.
@sachin-saxena could you please reimplement your PR following @Bill-Fischofer-Linaro 's suggestion?

muvarov: 1) is it needed for tmo and bufs also? why it's on only inside packets? 2) How memory for *start_addr is allocated in VPP? In previous version of odp_pool_create() was odp_shm_t parameter which said to reuse already created shm.  Maybe that needs to be reconsidered. 3) why union is needed? why it's not with uintptr_t ?

Bill-Fischofer-Linaro: @muvarov To your points:

1. No, VPP only deals with packets so there's no need to generalize this since it is an accommodation for VPP, not something we want to encourage other applications to use.

2. We abandoned that approach because it required the application to know how much memory the ODP implementation needed for its internal use, which is not something it can reasonably know. So `odp_pool_create()` is responsible for allocating any shm used by the pool based on input provided by the caller. 

3. Agree, the union seems strange here. Since having an output parameter in the `odp_pool_param_t` is not something we want this will have to be reworked anyway.

psavol: How sparse a "linear pool" may be? All implementations can claim linear pool support by returning info.first_addr = 0, info.last_addr = (uintptr_t) -1, where address size may be 64bits.

Addresses could be used for debugging, but not much more than that. What VPP is going to do with this information ?

Bill-Fischofer-Linaro: As we discussed earlier today. We should only need an extension to `odp_pool_info()` to return the `min_addr` and `max_addr` of any packet contained in the pool. This can simply be added to the end of the `odp_pool_info_t` struct.

include/odp/api/spec/pool.h
```diff
@@ -290,10 +290,16 @@ odp_pool_t odp_pool_lookup(const char *name);
 /**
  * Pool information struct
  * Used to get information about a pool.
+ * @note The difference between end_addr & start_addr
+ * will result in buffers address range belong to this pool.
```
Bill-Fischofer-Linaro: I'd delete this note as it's not necessary and not complete for the ODP API spec.

include/odp/api/spec/pool.h
```diff
@@ -290,10 +290,16 @@ odp_pool_t odp_pool_lookup(const char *name);
...
  */
 typedef struct odp_pool_info_t {
 	const char *name;          /**< pool name */
 	odp_pool_param_t params;   /**< pool parameters */
+	/** Minimum address of any packet contained in this pool */
+	uintptr_t start_addr;
+	/** Maximum address of any packet contained in this pool */
+	uintptr_t end_addr;
```
Bill-Fischofer-Linaro: I'd change these to "visible address" rather than "address". The point is that the spec says nothing about how packets are represented within an implementation. It only states what applications may see by using other ODP APIs that manipulate odp_packet_t object.

Bill-Fischofer-Linaro: The only other ambiguity is whether the address that applications see for a packet is constant. VPP clearly assumes this but this is not implied by the ODP spec. The "lifetime" of packet visibility is from the time a call like `odp_packet_data()` is made until the thread releases the `odp_packet_t` that was used to obtain this address. If some other thread receives that handle and calls `odp_packet_data()` there's no requirement or guarantee that it will get the same address as the previous owner.

include/odp/api/spec/pool.h
sachin-saxena: OK. agreed

include/odp/api/spec/pool.h
sachin-saxena: 1.  Agreed for first suggestion.
2.  For Second, As per my understanding, the VPP only needs to know the range of addresses in order to calculate OFFSET within. VPP doesn't assume/store packet address beyond its lifetime (Rx to Tx)

sachin-saxena: I will wait for other's comments, if any, before sending V3 patchset

psavol:
Git log entry text is missing.  Log entry should give a rationale why this change is needed. E.g. "Added pool memory range info fields. These may be used e.g. while debugging."

include/odp/api/spec/pool.h
psavol: This spec should apply for all pool types (also other than packet). 0 address may be used when there's no data in an event (e.g. timeout).

/** Minimum data address.
   * This is the minimum address that application accessible data of any 
   * object (event) allocated from the pool may locate. When there's no 
   * application accessible data (e.g. ODP_POOL_TIMEOUT pools), the 
   * value maybe zero. */
uintptr_t min_data_addr;

/** Maximum data address.
   * This is the maximum address that application accessible data of any 
   * object (event) allocated from the pool may locate. When there's no 
   * application accessible data (e.g. ODP_POOL_TIMEOUT pools), the 
   * value maybe zero.  */
uintptr_t max_data_addr;

Bill-Fischofer-Linaro:
@muvarov Not sure why ODP License Agreement check is still reporting as failing here.

include/odp/api/spec/pool.h
sachin-saxena: Looks much better. Thank you @psavol for your suggestions.

include/odp/api/spec/pool.h
```diff
@@ -294,6 +294,19 @@ odp_pool_t odp_pool_lookup(const char *name);
...
+	application accessible data (e.g. ODP_POOL_TIMEOUT pools), the
+	value maybe zero. */
+	uintptr_t min_data_addr;
+
+	/** Maximum data address.
+	This is the maximum address that application accessible data of any
+	object (event) allocated from the pool may locate. When there's no
+	application accessible data (e.g. ODP_POOL_TIMEOUT pools), the
+	value maybe zero. */
+	uintptr_t max_data_addr;
```
psavol: While adding commit log text, modify the style in these comments to begin each row with a *.

'/**
  *
  *
  */'

sachin-saxena: Sure.

psavol:
Still missing the commit log entry text.

include/odp/api/spec/pool.h
```diff
@@ -294,6 +294,22 @@ odp_pool_t odp_pool_lookup(const char *name);
 typedef struct odp_pool_info_t {
 	const char *name;          /**< pool name */
 	odp_pool_param_t params;   /**< pool parameters */
+
+	/** Minimum data address.
+	 * This is the minimum address that application accessible data of any
+	 * object (event) allocated from the pool may locate. When there's no
+	 * application accessible data (e.g. ODP_POOL_TIMEOUT pools), the
+	 * value maybe zero.
```
brbrooks: "maybe" -> "may be"

include/odp/api/spec/pool.h
```diff
@@ -294,6 +294,22 @@ odp_pool_t odp_pool_lookup(const char *name);
...
+	 */
+	uintptr_t min_data_addr;
+
+	/** Maximum data address.
+	 * This is the maximum address that application accessible data of any
+	 * object (event) allocated from the pool may locate. When there's no
+	 * application accessible data (e.g. ODP_POOL_TIMEOUT pools), the
+	 * value maybe zero.
```
brbrooks: "maybe" -> "may be"

Bill-Fischofer-Linaro:
Checkpatch flags excessive line length in commit msg. Please reflow to a max line length of 75 chars to keep it happy.

sachin-saxena:
FYI.. I have electronically signed the Individual Contributor License Agreement but still ODP License Check is failing.

Bill-Fischofer-Linaro:
@muvarov Any ideas as to why the license check is failing? @sachin-saxena your linaro.org e-mail address should mean there's no need to sign anything.

muvarov:
@Bill-Fischofer-Linaro  @nxp.com was not in the list. Added it.

muvarov:
https://github.com/Linaro/odp/pull/200.patch
From: Sachin Saxena <sachin.saxena@nxp.com>
Signed-off-by: Sachin Saxena <sachin.saxena@linaro.org>

include/odp/api/spec/pool.h
```diff
@@ -294,6 +294,22 @@ odp_pool_t odp_pool_lookup(const char *name);
 typedef struct odp_pool_info_t {
 	const char *name;          /**< pool name */
 	odp_pool_param_t params;   /**< pool parameters */
+
+	/** Minimum data address.
+	 * This is the minimum address that application accessible
+	 * data of any * object (event) allocated from the pool may
```
psavol: There's an extra * on this line.

muvarov:
Merged.

PR #202, opened 2017-10-02
[PATCH v1] tests: fix validation tests being skipped by default

lumag:
After [d091f2176a28 configure: "best effort" approach for CUnit and
validation tests] by default validation tests will get skipped by
default because test_vald variable will remain set to check instead of
yes. Update it to yes, if it was not set and CUnit was found.


lumag:
@muvarov @Bill-Fischofer-Linaro this should be merged as soon as possible.

muvarov:
Merged.

PR #203, opened 2017-10-02
[PATCH API-NEXT v5] linux-gen: pktio: add MAC address set function

bogdanPricope:

### DIFF
--- a/include/odp/api/spec/packet_io.h
+++ b/include/odp/api/spec/packet_io.h
@@ -482,6 +482,8 @@ typedef union odp_pktio_set_op_t {
 	struct {
 		/** Promiscuous mode */
 		uint32_t promisc_mode : 1;
+		/** MAC address  */
+		uint32_t mac_addr : 1;
 	} op;
 	/** All bits of the bit field structure.
 	  * This field can be used to set/clear all flags, or bitwise
@@ -977,6 +979,22 @@ int odp_pktio_promisc_mode(odp_pktio_t pktio);
  * @retval <0 on failure
  */
 int odp_pktio_mac_addr(odp_pktio_t pktio, void *mac_addr, int size);
+
+/**
+ * Set the default MAC address of a packet IO interface.
+ *
+ * Support of this operation on a packet IO interface is reported
+ * through ‘mac_addr’ set operation capability.
+ *
+ * @param	pktio     Packet IO handle
+ * @param	mac_addr  MAC address to be set as default address
+ * @param	size      Size of the MAC address
+ *
+ * @return 0 on success
+ * @retval <0 on failure
+ */
+int odp_pktio_mac_addr_set(odp_pktio_t pktio, const void *mac_addr,
+			   int size);
 
 /**
  * Setup per-port default class-of-service.
--- a/test/common_plat/validation/api/pktio/pktio.c
+++ b/test/common_plat/validation/api/pktio/pktio.c
@@ -1065,9 +1065,12 @@ void pktio_test_promisc(void)
 void pktio_test_mac(void)
 {
 	unsigned char mac_addr[ODP_PKTIO_MACADDR_MAXSIZE];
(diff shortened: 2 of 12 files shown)


Bill-Fischofer-Linaro:
The previous [patch](http://patches.opendataplane.org/patch/9930/) that included this function allowed the setting of multiple MAC addrs. Is this no longer needed or is the intent to add this in stages?

Also, is this needed as part of Tiger Moth?

bogdanPricope:
Hi Bill,

I know that patch and this one is an attempt to push discussions forward and have at least a basic functionality.

Today we have a simple “get” function for a single MAC address:

int odp_pktio_mac_addr(odp_pktio_t pktio, void *mac_addr, int size);

Minimum addition will be a matching “set” function:

int odp_pktio_mac_addr_set(odp_pktio_t pktio, const void *mac_addr, int size);

If I remember correctly, multiple MAC address API resulted from idea of completeness. Yet, I did not notice an interest in this usecase right now. May be added in another stage, if needed.

It is “nice-to-have” in Tiger Moth (at least in this basic form).

Bill-Fischofer-Linaro:
@muvarov @lumag Not clear why this Travis is reporting failures for certain tests on this PR. Looks like the hello example is hiccuping?

Bill-Fischofer-Linaro:
@bogdanPricope Please rebase this to pick up the lates Travis fixes so regressions can be properly done on this PR.

include/odp/api/spec/packet_io.h
```diff
@@ -979,6 +981,20 @@ int odp_pktio_promisc_mode(odp_pktio_t pktio);
 int odp_pktio_mac_addr(odp_pktio_t pktio, void *mac_addr, int size);
 
 /**
+ * Set the default MAC address of a packet IO interface.
+ *
+ * @param	pktio     Packet IO handle
+ * @param[in]	mac_addr  Input buffer containing MAC address
+ * @param[in]	size      Size of the MAC address contained in
+ *			  input buffer
+ *
```
psavol: There's no need to highlight that these params are input, so remove [in] from param tags.

Also, no need to use term "input buffer". Simply, "mac_addr  MAC address to be set as the default"

include/odp/api/spec/packet_io.h
```diff
@@ -979,6 +981,20 @@ int odp_pktio_promisc_mode(odp_pktio_t pktio);
 int odp_pktio_mac_addr(odp_pktio_t pktio, void *mac_addr, int size);
 
 /**
+ * Set the default MAC address of a packet IO interface.
+ *
```
psavol: It needs to be mentioned here that an interface supports this function only when capa.set_op.op.mac_addr == 1

test/common_plat/validation/api/pktio/pktio.c
```diff
@@ -1088,6 +1091,27 @@ void pktio_test_mac(void)
 	mac_len = odp_pktio_mac_addr(pktio, mac_addr, 2);
 	CU_ASSERT(mac_len < 0);
 
+	CU_ASSERT_FATAL(odp_pktio_capability(pktio, &capa) == 0);
+	if (!capa.set_op.op.mac_addr) {
+		printf("\n mac address set not supported for %s!\n",
+		       iface_name[0]);
+		ret = odp_pktio_close(pktio);
+		CU_ASSERT(ret == 0);
```
psavol: I'd organize this the other way - move MAC addr set test code inside the if-clause, so that the same pktio_close() path would be used at the end of the test.

psavol:
All 3 patches need git commit log text to document the rationale of the changes.

include/odp/api/spec/packet_io.h
bogdanPricope: ok

muvarov:
Merged.

PR #205, opened 2017-10-03
[PATCH v2] timer cleanup

muvarov:
No functional changes just clean up naming. This file also needs some rework, which can be done as separate patches.

### DIFF
--- a/platform/linux-generic/include/odp/api/plat/timer_types.h
+++ b/platform/linux-generic/include/odp/api/plat/timer_types.h
@@ -24,9 +24,9 @@ extern "C" {
  *  @{
  **/
 
-struct odp_timer_pool_s; /**< Forward declaration */
+struct timer_pool_s; /**< Forward declaration */
 
-typedef struct odp_timer_pool_s *odp_timer_pool_t;
+typedef struct timer_pool_s *odp_timer_pool_t;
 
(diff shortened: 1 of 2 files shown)


brbrooks:
```timer_pool_t``` is preferred to ```_timer_pool_t```

Let's stop using the underscore ```_``` prefix.

File: platform/linux-generic/include/odp/api/plat/timer_types.h [hunk 1]
```diff
@@ -24,9 +24,9 @@ extern "C" {
  *  @{
  **/
 
-struct odp_timer_pool_s; /**< Forward declaration */
+struct _timer_pool_s; /**< Forward declaration */
```
Bill-Fischofer-Linaro: The only _ prefix we've been using is _odp, so I'd either change this to use _odp or else just remove odp_ altogether than leave this as timer_pool_s, etc.

brbrooks: I recommend we modify the Coding Style to say ```odp_``` prefix is reserved for the API, and internal data structures should just drop the ```odp_``` prefix.

API data type: ```odp_timer_pool_t```
Internal data type: ```timer_pool_t```

muvarov:
Merged.

muvarov:
Merged.

Issue #206, opened 2017-10-03
./bootstrap throws warnings


nagarahalli:
honnag01@ubuntu:~/api-next/odp$ ./bootstrap
+ aclocal -I config -I m4
./test/linux-dpdk/m4/configure.m4:1: warning: file `test/linux-generic/m4/performance.m4' included several times
../../lib/m4sugar/m4sh.m4:639: AS_IF is expanded from...
./test/linux-dpdk/m4/configure.m4:1: the top level
./test/linux-dpdk/m4/configure.m4:1: warning: file `test/linux-generic/m4/performance.m4' included several times
../../lib/m4sugar/m4sh.m4:639: AS_IF is expanded from...
./test/linux-dpdk/m4/configure.m4:1: the top level
+ libtoolize --copy
+ autoheader
./test/linux-dpdk/m4/configure.m4:1: warning: file `test/linux-generic/m4/performance.m4' included several times
../../lib/m4sugar/m4sh.m4:639: AS_IF is expanded from...
./test/linux-dpdk/m4/configure.m4:1: the top level
+ automake --add-missing --copy --warnings=all
./test/linux-dpdk/m4/configure.m4:1: warning: file `test/linux-generic/m4/performance.m4' included several times
../../lib/m4sugar/m4sh.m4:639: AS_IF is expanded from...
./test/linux-dpdk/m4/configure.m4:1: the top level
+ autoconf --force
./test/linux-dpdk/m4/configure.m4:1: warning: file `test/linux-generic/m4/performance.m4' included several times
../../lib/m4sugar/m4sh.m4:639: AS_IF is expanded from...
./test/linux-dpdk/m4/configure.m4:1: the top level

Bill-Fischofer-Linaro:
I don't see this on the current api-next. This appears to be an artifact introduced by the attempted merge? The base cloud-dev branch doesn't see this either.

```
bill@Ubuntu64:~/linaro/api-next$ ./bootstrap
+ aclocal -I config -I m4
+ libtoolize --copy
libtoolize: putting auxiliary files in '.'.
libtoolize: copying file './ltmain.sh'
libtoolize: putting macros in AC_CONFIG_MACRO_DIRS, 'm4'.
libtoolize: copying file 'm4/libtool.m4'
libtoolize: copying file 'm4/ltoptions.m4'
libtoolize: copying file 'm4/ltsugar.m4'
libtoolize: copying file 'm4/ltversion.m4'
[...]

PR #207, opened 2017-10-03
[PATCH CLOUD-DEV v1] api-next merge to cloud-dev

nagarahalli:
Merge api-next to cloud-dev branch. Last merge was done on Aug 8th 2017.
./bootstrap throws warnings. @lumag looked at it and said, they should not cause any issues. Issue #206 is created to track the issue.
Doxygen build target in Travis is failing, but the Doxygen compilation is working fine. @Bill-Fischofer-Linaro  mentioned that this should be fixed in the mainline.


Bill-Fischofer-Linaro:
@muvarov Would appreciate you looking this over to be sure we're doing it in a manner consistent with how you do it.

PR #209, opened 2017-10-04
[PATCH v1] travis: typo to call dynamic executable

muvarov:
we changed name for dynamic executable on compile
line but did not change it on execute line.


lumag:

Two questions:
1) What is the reason why we did not see this as Travis CI failures?
2) What is wrong now with DPDK tests?

muvarov:
@lumag 
1. Magic Travis cache, I think.
2. Zero copy only test fail. It began after Krishna updated PR to update dpdk version.

muvarov:
Merged.

PR #212, opened 2017-10-05
[PATCH v1] tests: fix disabling of validation tests

lumag:
Fix for the fix...
[b4d17b1f6807 tests: fix validation tests being skipped by default]
attempted to fix tests being disabled by default, but in attempt to do
so just forced them to be enabled even if they are forcibly disabled by
configure option. Update conditions to really enable testsuite if it was
not disabled.


lumag:
This is a fix for https://bugs.linaro.org/show_bug.cgi?id=3300

muvarov:
Merged.

PR #216, opened 2017-10-05
[PATCH v1] linux-gen: Cleanup on printing system information

apalos:
Use unified odp_sys_info_print() instead of copying the same code
on every example.


muvarov:
Merged.

PR #218, opened 2017-10-05
[PATCH API-NEXT v1] travis: build dpdk for general cpu

muvarov:
as described here build dpdk for Sandybridge
https://stackoverflow.com/questions/37386738/error-when-running-dpdk-app-on-valgrind
to make it work in Travis

### DIFF
--- a/.travis.yml
+++ b/.travis.yml
@@ -161,6 +161,8 @@ install:
             make config T=${TARGET} O=${TARGET}
             pushd ${TARGET}
             sed -ri 's,(CONFIG_RTE_LIBRTE_PMD_PCAP=).*,\1y,' .config
+            cat .config |grep RTE_MACHINE
+            sed -ri 's,(CONFIG_RTE_MACHINE=).*,\1"snb",' .config
             popd
             make install T=${TARGET} EXTRA_CFLAGS="-fPIC"
             rm -r ./doc ./${TARGET}/app ./${TARGET}/build


Bill-Fischofer-Linaro:
@muvarov Travis reports all builds were cancelled after 9 hours. Is there something that hangs Travis in this PR?

muvarov:
This is not needed, default RTE_MACHINE as says log is "native". So I think issue is somewhere else..

muvarov:
https://travis-ci.org/Linaro/odp/jobs/284072086 that log says:
ERROR: This system does not support "RDRAND".

Please check that RTE_MACHINE is set correctly.

!!! FAILED !!!

Because of this build was ok than now I think that it is reasonable change.

muvarov:
uh, dpdk does not have "generic" machine by default:
./mk/machine/snb/rte.vars.mk
MACHINE_CFLAGS = -march=corei7-avx

./mk/machine/default/rte.vars.mk
MACHINE_CFLAGS += -march=core2

./mk/machine/native/rte.vars.mk
MACHINE_CFLAGS = -march=native

Default is "native". Travis correctly builds "snb".  I locally checked that if native changed to "snb" RDRAND is not provided to makefile arguments. Also Travis builds and runs "snb" correctly.  I would prefer to compile code for something generic, not specifying cpu arch.

make config T=${TARGET} O=${TARGET} MACHINE_CFLAGS=""
looks like work also really well and do not pass any -march= option.

Going to update it with that change. Also we will need to clean Travis cache to make it work.

muvarov:
Merged commit to master. Did a lot of tests runs and now I hope everything will be stable.

PR #220, opened 2017-10-05
[PATCH v3] travis: fix netmap module loading

lumag:
Travis script will insmod netmap.ko only it was rebuilt during this
session, which is wrong. Split the ifs, so that module loading does not
depend on the cache contents.


muvarov:
Merged.

PR #222, opened 2017-10-06
[PATCH v1] travis: purge dpdk cache on version change

muvarov:
changing dpdk version and not clearing build can lead
to reference to old cached not supported dpdk version.
Needed to remove cache if version was changed.

### DIFF
--- a/.travis.yml
+++ b/.travis.yml
@@ -149,11 +149,17 @@ install:
           fi
         - gem install asciidoctor
 
-#	DPDK pktio. Note that cache must be purged if dpdk version changes.
+        # DPDK pktio. Note that cache must be purged if dpdk version changes.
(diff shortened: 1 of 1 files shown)


lumag:

I wonder if we should instead publish our 'good' dpdk/cunit/netmap/doxygen as some kind of PPA (or repo somewhere else), then just install packages from that PPA to build station?
In theory this PPA can also include e.g. MIPS64 toolchain (which is not available ATM), etc.

muvarov:
Meged

muvarov:
@lumag in the steps we just do what we have in README file and it is distro agnostic steps. Cunit can be for now from  sources until we can jump to more fresh image.

muvarov:
Closed.

PR #223, opened 2017-10-06
[PATCH v3] linux-gen: pool: avoid allocating packets which cross page boundaries

matiaselo:
When packet pool memory page size is >= FIRST_HP_SIZE, leave those memory buffers unused which cross page boundary. This is required by zero-copy dpdk pktio (NIC drivers don't support buffers which cross page boundaries).

V2:
- Rebase (Bill)
- Clarified terminology (Bill)
- Minimized the number of allocated extra packet buffers (Bill)
V3:
- Abort in case of odp_shm_info() failure (Maxim)

### DIFF
--- a/platform/linux-generic/odp_pool.c
+++ b/platform/linux-generic/odp_pool.c
@@ -32,6 +32,10 @@
 #define CACHE_BURST    32
 #define RING_SIZE_MIN  (2 * CACHE_BURST)
 
+/* Make sure packet buffers don't cross huge page boundaries starting from this
+ * page size. 2MB is typically the smallest used huge page size. */
+#define FIRST_HP_SIZE (2 * 1024 * 1024)
+
 /* Define a practical limit for contiguous memory allocations */
 #define MAX_SIZE   (10 * 1024 * 1024)
 
@@ -224,6 +228,7 @@ static void init_buffers(pool_t *pool)
 	uint32_t i;
 	odp_buffer_hdr_t *buf_hdr;
 	odp_packet_hdr_t *pkt_hdr;
+	odp_shm_info_t shm_info;
 	void *addr;
 	void *uarea = NULL;
 	uint8_t *data;
@@ -232,19 +237,41 @@ static void init_buffers(pool_t *pool)
 	uint32_t mask;
 	int type;
 	uint32_t seg_size;
+	uint64_t page_size;
+	int skipped_blocks = 0;
 
+	if (odp_shm_info(pool->shm, &shm_info))
+		ODP_ABORT("Shm info failed\n");
+
+	page_size = shm_info.page_size;
 	ring = &pool->ring->hdr;
 	mask = pool->ring_mask;
 	type = pool->params.type;
 
-	for (i = 0; i < pool->num; i++) {
+	for (i = 0; i < pool->num + skipped_blocks ; i++) {
 		addr    = &pool->base_addr[i * pool->block_size];
 		buf_hdr = addr;
 		pkt_hdr = addr;
+		/* Skip packet buffers which cross huge page boundaries. Some
+		 * NICs cannot handle buffers which cross page boundaries. */
+		if (pool->params.type == ODP_POOL_PACKET &&
+		    page_size >= FIRST_HP_SIZE) {
+			uint64_t first_page;
+			uint64_t last_page;
 
+			first_page = ((uint64_t)(uintptr_t)addr &
+					~(page_size - 1));
+			last_page = (((uint64_t)(uintptr_t)addr +
+					pool->block_size - 1) &
+					~(page_size - 1));
+			if (last_page != first_page) {
+				skipped_blocks++;
+				continue;
+			}
+		}
 		if (pool->uarea_size)
-			uarea = &pool->uarea_base_addr[i * pool->uarea_size];
-
+			uarea = &pool->uarea_base_addr[(i - skipped_blocks) *
+						       pool->uarea_size];
 		data = buf_hdr->data;
 
 		if (type == ODP_POOL_PACKET)
@@ -317,6 +344,7 @@ static odp_pool_t pool_create(const char *name, odp_pool_param_t *params,
 	uint32_t data_size, align, num, hdr_size, block_size;
 	uint32_t max_len, max_seg_len;
 	uint32_t ring_size;
+	uint32_t num_extra = 0;
 	int name_len;
 	const char *postfix = "_uarea";
 	char uarea_name[ODP_POOL_NAME_LEN + sizeof(postfix)];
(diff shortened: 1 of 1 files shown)


muvarov:
@matiaselo should it fix current fails with dpdk pktio?

matiaselo:
> On 6 Oct 2017, at 12:08, muvarov <notifications@github.com> wrote:
> 
> @matiaselo should it fix current fails with dpdk pktio?
> 
> —
> You are receiving this because you were mentioned.
> Reply to this email directly, view it on GitHub, or mute the thread.
> 

No, this only fixes intermittent packet drops with dpdk zero-copy.

platform/linux-generic/odp_pool.c
```diff
@@ -403,6 +430,12 @@ static odp_pool_t pool_create(const char *name, odp_pool_param_t *params,
 	block_size = ROUNDUP_CACHE_LINE(hdr_size + align + headroom +
 					data_size + tailroom);
 
+	/* Allocate extra memory for skipping packet buffers which cross page
+	 * boundaries. */
+	if (params->type == ODP_POOL_PACKET)
+		num_extra = 2 * (((uint64_t)(num * block_size) +
+				CONT_PAGE_SIZE - 1) / CONT_PAGE_SIZE);
```
Bill-Fischofer-Linaro: Why the `2 *` factor here? The number of block size "page" crossings is going to be limited to the rounded up number of pages:

```
<--------------------------> "n" pages in size
---+--------+--- ... ---+---- There can be at most n crossings,
                              (n - 1) if block happily starts on a page boundary.

```

matiaselo: Also the extra buffers can cross page boundaries.

platform/linux-generic/odp_pool.c
```diff
@@ -232,19 +237,40 @@ static void init_buffers(pool_t *pool)
...
 	type = pool->params.type;
 
-	for (i = 0; i < pool->num; i++) {
+	for (i = 0; i < pool->num + skipped_blocks ; i++) {
 		addr    = &pool->base_addr[i * pool->block_size];
 		buf_hdr = addr;
 		pkt_hdr = addr;
-
+		/* Skip packet buffers which cross page boundaries. Some NICs
+		 * cannot handle buffers which cross page boundaries. */
```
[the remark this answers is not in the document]
matiaselo: Currently this is required for the dpdk pmd drivers only, but if/when we are adding support for more drivers this issue is likely to re-emerge. IMHO the applications shouldn't have to know about this, so I wouldn't add a new parameter.

platform/linux-generic/odp_pool.c
```diff
@@ -32,6 +32,10 @@
 #define CACHE_BURST    32
 #define RING_SIZE_MIN  (2 * CACHE_BURST)
 
+/* Make sure packet buffers don't cross page boundaries starting from this page
+ * size.  */
+#define CONT_PAGE_SIZE (2 * 1024 * 1024)
```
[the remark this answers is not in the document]
matiaselo: Yes, I'm referring to all page sizes >= 2MB, so the same operation is also done with 1GB pages. I was struggling to figure out a good name for this define, so all ideas are welcome.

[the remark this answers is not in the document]
Bill-Fischofer-Linaro: I'd call it huge pages if that's the problem you're trying to address. Any jumbo packet is going to cross a 4K boundary if it isn't segmented. I assume that isn't a problem?

platform/linux-generic/odp_pool.c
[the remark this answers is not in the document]
Bill-Fischofer-Linaro: Fair enough. I agree these sort of considerations should be transparent to the application.

platform/linux-generic/odp_pool.c
Bill-Fischofer-Linaro: The blocks are contiguous so you're computing how many of them can fit in a 2M huge page. Since blocks aren't themselves larger than a huge page I don't see what's wrong with my analysis.

Bill-Fischofer-Linaro:
Please rebase this on the next revision to pick up the Travis fixes so the regressions can run properly. Thanks.

platform/linux-generic/odp_pool.c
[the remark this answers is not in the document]
matiaselo: The dpdk drivers only support huge page memory, so packets crossing 4K page boundaries isn't a problem.

platform/linux-generic/odp_pool.c
matiaselo: As you noted there can be at most N crossings with the original pool size ('num'). Now, to compensate for the skipped buffers the pool size is in practice increased by 'num_extra'. When the original pool size is large enough the extra buffers may again cross page boundary. 

Currently, a single buffer is 8KB, so a single 2MB page can hold max 256 buffers. With a pool size of 64K or larger the extra buffers will cross page boundary again.

Using 2x multiplication here is overkill. I'll update the patch and calculate the exact number of required extra buffers.

platform/linux-generic/odp_pool.c
```diff
@@ -232,19 +237,40 @@ static void init_buffers(pool_t *pool)
 	uint32_t mask;
 	int type;
 	uint32_t seg_size;
+	uint64_t page_size = 0;
+	int skipped_blocks = 0;
+
+	if (odp_shm_info(pool->shm, &shm_info) == 0)
+		page_size = shm_info.page_size;
```
muvarov: if call fails it has to be an error? result in that case is not predictable?

Bill-Fischofer-Linaro: @muvarov In the unlikely (impossible in the case of the current linux-generic implementation) event of this call failing, `page_size` will be left at the initialized default 0 value, which means we won't skip any buffers since 0 is less than `FIRST_HP_SIZE`. No no error here.

matiaselo: As Bill noted, if odp_shm_info() could somehow fail, nothing is still broken done.

muvarov: actually it can.  it can return -1 for some reasons. And pool_create() and pool_destroy() from different threads for the same time, will make shm invalid before init_buffers(). But that is corner case and programmable error :)

But it works. My point was a little bit different. In my understanding assuming that odp api function will fail is not correct in places like that. That works like work around - what we will do if call will fail. Instead of that you should force platform to implement this api call in correct way. Especially if it's reference implementation. I would prefer ODP_ABORT() if call fails as that is not expected here.

Bill-Fischofer-Linaro: This is happening within `odp_pool_create()`, which is creating the `odp_pool_t`, so other threads don't have that handle with which to call `odp_pool_destroy()`. 

The only way `odp_shm_info()` can fail is if it's passed an unknown `odp_shm_t` handle, which is not the case here. 

I have no objection to adding an `else ODP_ABORT()` here, however.

matiaselo: OK, I'll add an abort branch.

muvarov: @Bill-Fischofer-Linaro  odp_pool_lookup() works in any thread. shm look up also can get that memory. Destroy path also has look up and destroy. So I think simple ABORT() here is good.

muvarov:
Merged.

PR #225, opened 2017-10-06
[PATCH v3] travis: build dpdk for general cpu

muvarov:
v2: define RTE_MACHINE in install stage, not on configure. There is really works.
---
1)
    disable cpu optimisations for code run in Travis env.
    By default dpdk uses -march=corei7-avx with generates
    specific instructions and execution fails with:
    Please check that RTE_MACHINE is set correctly.
    More info here:
    https://github.com/Linaro/odp/pull/218

2) disabling dpdk caching in Travis for now.

### DIFF
--- a/.travis.yml
+++ b/.travis.yml
@@ -159,6 +159,8 @@ install:
             make config T=${TARGET} O=${TARGET}
             pushd ${TARGET}
             sed -ri 's,(CONFIG_RTE_LIBRTE_PMD_PCAP=).*,\1y,' .config
+            cat .config |grep RTE_MACHINE
+            sed -ri 's,(CONFIG_RTE_MACHINE=).*,\1"snb",' .config
             popd
             make install T=${TARGET} EXTRA_CFLAGS="-fPIC"
             rm -r ./doc ./${TARGET}/app ./${TARGET}/build
--- a/.travis.yml
+++ b/.travis.yml
@@ -57,7 +57,7 @@ cache:
         ccache: true
         pip: true
         directories:
-                - dpdk
+                #- dpdk
                 - netmap
                 - $HOME/cunit-install
                 - $HOME/doxygen-install


.travis.yml
```diff
@@ -57,7 +57,7 @@ cache:
         ccache: true
         pip: true
         directories:
-                - dpdk
+                #- dpdk
```
Bill-Fischofer-Linaro: I assume we'll turn this back on before merging if this fix works.

.travis.yml
```diff
@@ -156,7 +156,7 @@ install:
             git -c advice.detachedHead=false clone -q --depth=1 --single-branch --branch=v17.02 http://dpdk.org/git/dpdk dpdk
             pushd dpdk
             git log --oneline --decorate
-            make config T=${TARGET} O=${TARGET}
+            make config T=${TARGET} O=${TARGET} MACHINE_CFLAGS=""
```
Bill-Fischofer-Linaro: Seems a simple fix. This behavior of Travis should also be a good test of the adaptability of ODP to the target microarchitecture going forward. ODP shouldn't need such compile-time flags since the goal is to have it determine this sort of low-level detail dynamically at runtime.

muvarov:
uh it failed :( one time passed second time failed. 

https://travis-ci.org/muvarov/odp/builds/284239805?utm_source=email&utm_medium=notification

but all dpdkp_pktio tests passed. And there is no RTE_MACHINE error now. So it looks like that is good to apply. odp_hello fails but that might be some other issue.

.travis.yml
lumag: +1

lumag:
Again, illegal instruction at `odp_hello`. Ugh.

Bill-Fischofer-Linaro:
@lumag Where are you seeing that? `odp_hello` seems to be working fine for me. It's such a trivial app that I don't understand why it would have problems in any environment.

PR #226, opened 2017-10-06
[PATCH v1] test: pktio_perf: increase maximum number of workers threads

matiaselo:
Increase maximum number of workers to 128. Also, fixes a buffer overflow
which happened when system's maximum thread count  was larger than the
value of MAX_WORKERS.


muvarov:
Merged.

PR #230, opened 2017-10-15
[PATCH API-NEXT v2] api: ipsec: return maximum antireplay window size via capability

lumag:

### DIFF
--- a/include/odp/api/spec/ipsec.h
+++ b/include/odp/api/spec/ipsec.h
@@ -282,6 +282,9 @@ typedef struct odp_ipsec_capability_t {
 	 *  be used for many SAs. */
 	uint32_t max_queues;
 
+	/** Maximum anti-replay window size. */
+	uint32_t max_antireplay_ws;
+
 	/** Supported cipher algorithms */
 	odp_crypto_cipher_algos_t ciphers;
 


lumag:
@psavol updated

include/odp/api/spec/ipsec.h
```diff
@@ -282,6 +282,9 @@ typedef struct odp_ipsec_capability_t {
 	 *  be used for many SAs. */
 	uint32_t max_queues;
 
+	/** Maximum anti-replay window size. */
+	uint32_t max_antireplay_ws;
```
muvarov: I think spec has to be more clear here. Measurement units has to be described (bits).  0 - means not supported? 0xffffffff -means unlimited?  Usually values are from 64 to 4096 did not find in the doc which is the maximum (uint32_t  or 16).

lumag: Measurement unit is packets, not bits. And 0 is not possible, because RFC explicitly states that `A minimum window size of 32 packets MUST be supported when 32-bit sequence numbers are employed`. Also I would like to see an implementation tracking all 2^32 packets for an unlimited case.

muvarov: ok, thanks.

muvarov:
Merged.

PR #232, opened 2017-10-16
[PATCH API-NEXT v2] Implement start/stop/link status/MAC set operations of TAP pktio

bogdanPricope:
Implement start/stop/link status/MAC address set operations of TAP interface..

### DIFF
--- a/platform/linux-generic/pktio/tap.c
+++ b/platform/linux-generic/pktio/tap.c
@@ -406,6 +406,12 @@ static int tap_mac_addr_get(pktio_entry_t *pktio_entry, void *mac_addr)
 	return ETH_ALEN;
 }
 
+static int tap_link_status(pktio_entry_t *pktio_entry)
+{
+	return link_status_fd(pktio_entry->s.pkt_tap.skfd,
+			      pktio_entry->s.name + 4);
+}
+
 static int tap_capability(pktio_entry_t *pktio_entry ODP_UNUSED,
 			  odp_pktio_capability_t *capa)
 {
@@ -438,6 +444,7 @@ const pktio_if_ops_t tap_pktio_ops = {
 	.promisc_mode_get = tap_promisc_mode_get,
 	.mac_get = tap_mac_addr_get,
 	.mac_set = NULL,
+	.link_status = tap_link_status,
 	.capability = tap_capability,
 	.pktin_ts_res = NULL,
 	.pktin_ts_from_ns = NULL,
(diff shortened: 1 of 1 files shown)


muvarov:
is there any reason why that for api-next?

bogdanPricope:
Yes. Goal was to implement 'set MAC address' for TAP pktio but this feature is available only in api-next.
Start/stop were needed to be able to set 'things' before interface is Up&Running.
Link status was needed to pass validation tests (a.k.a ./test/linux-generic/validation/api/pktio/pktio_run_tap.sh).

There are NO public API changes, if this is what you are asking.

platform/linux-generic/pktio/tap.c
```diff
@@ -412,6 +412,16 @@ static int tap_link_status(pktio_entry_t *pktio_entry)
 			      pktio_entry->s.name + 4);
 }
 
+static int tap_mac_addr_set(pktio_entry_t *pktio_entry, const void *mac_addr)
+{
+	pkt_tap_t *tap = &pktio_entry->s.pkt_tap;
+
+	memcpy(tap->if_mac, mac_addr, ETH_ALEN);
+
+	return mac_addr_set_fd(tap->fd, (char *)pktio_entry->s.name + 4,
```
muvarov: mac_addr_set_fd() defined in socket.c and used in tap.c,  that is confising...

bogdanPricope: mac_addr_get_fd() is also there and should be used for implementation of .mac_set callback of all the related pktios (socket, socket_mmap, etc.).
It is confusing and for 2.0 I moved all those function to common.h/c.

bogdanPricope: Cleaning up the code (moving all common functions to a common file) is not the purpose of this PR.
Should be done in different PR and is under review in cloud-dev/2.0.

lumag:
Looks good,

muvarov:
I do not agree. All tap.c socket.c pcap.c fileas are stand alone isolated modules. There is no reason to put function to socket.c and call it from tap.c. Put this function to tap.c or create common file. I'm speaking about mac_addr_set_fd() not get_fd.  Patch 99e473a626d7 needs restructure.

lumag:
@muvarov This function can (and maybe should) be used from socket.c also, as `SIOCSIFHWADDR` should be also valid for other hw interfaces.

muvarov:
Merged.

PR #233, opened 2017-10-16
[PATCH v3] linux-gen: pool: select packet pool segment size at runtime

matiaselo:
This patch set modifies packet pool implementation to select segment size at runtime based on the pool parameters. After this patch set an application is able to create multiple packet pools with varying segment sizes.

V2:
- Added CONFIG_PACKET_SEG_DISABLED checks (Bill)

V3:
- Document gcc bug fix (Bill)

### DIFF
--- a/platform/linux-generic/include/odp_packet_internal.h
+++ b/platform/linux-generic/include/odp_packet_internal.h
@@ -194,6 +194,7 @@ static inline seg_entry_t *seg_entry_last(odp_packet_hdr_t *hdr)
  */
 static inline void packet_init(odp_packet_hdr_t *pkt_hdr, uint32_t len)
 {
+	pool_t *pool = pool_entry_from_hdl(pkt_hdr->buf_hdr.pool_hdl);
 	uint32_t seg_len;
 	int num = pkt_hdr->buf_hdr.segcount;
 
@@ -203,7 +204,7 @@ static inline void packet_init(odp_packet_hdr_t *pkt_hdr, uint32_t len)
 	} else {
 		seg_entry_t *last;
 
-		seg_len = len - ((num - 1) * CONFIG_PACKET_MAX_SEG_LEN);
+		seg_len = len - ((num - 1) * pool->seg_len);
 
 		/* Last segment data length */
 		last      = seg_entry_last(pkt_hdr);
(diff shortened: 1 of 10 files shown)


platform/linux-generic/odp_packet.c
```diff
@@ -2146,7 +2146,7 @@ odp_packet_t odp_packet_ref(odp_packet_t pkt, uint32_t offset)
 	link_hdr->buf_hdr.seg[0].len  = seg->len  - seg_offset;
 	buffer_ref_inc(seg->hdr);
 
-	for (i = 1; i < num_copy; i++) {
+	for (i = 1; CONFIG_PACKET_SEGS_PER_HDR > 1 && i < num_copy; i++) {
```
Bill-Fischofer-Linaro: Not clear why this is necessary. If `CONFIG_PACKET_SEGS_PER_HDR == 1` then `num_copy` should be 1 here so the original `for` loop will also not execute. Or am I missing something here?

platform/linux-generic/include/odp_packet_internal.h
```diff
@@ -197,7 +197,7 @@ static inline void packet_init(odp_packet_hdr_t *pkt_hdr, uint32_t len)
 	uint32_t seg_len;
 	int num = pkt_hdr->buf_hdr.segcount;
 
-	if (odp_likely(CONFIG_PACKET_MAX_SEGS == 1 || num == 1)) {
+	if (odp_likely(num == 1)) {
```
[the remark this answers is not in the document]
matiaselo: OK, I'll restore the CONFIG_PACKET_SEG_DISABLED checks.

platform/linux-generic/odp_packet.c
matiaselo: Gcc throws an invalid 'array subscript is above array bounds' error if this extra clause is not added when CONFIG_PACKET_SEGS_PER_HDR=1.

matiaselo:
I've synced with Petri and this patch set should not clash with PR #234.

platform/linux-generic/odp_packet.c
Bill-Fischofer-Linaro: Which level of GCC? How about clang? I hate to mangle our code to accommodate a very specific compiler bug. If we need to so something like that at minimum it should be commented so that it can be cleaned up when the bug is eventually fixed.

matiaselo: gcc 5.4.0 20160609 (Ubuntu 5.4.0-6ubuntu1~16.04.5). Clang doesn't complain about this line. I'll add a comment about the gcc bug.

muvarov:
Merged.

PR #238, opened 2017-10-18
[PATCH API-NEXT v1] api: ipsec: move pipeline SA param to inbound substructrure

lumag:
SA param pipeline is related only to inbound configuration. If at some
point it will be possible to configure destination for outbound
operation, it will be done through a separate param. Move pipeline into
inbound params substructure.


muvarov:
Merged.

PR #239, opened 2017-10-18
[PATCH v1] linux-gen: fix missing unlock and make pktio more readable

muvarov:


platform/linux-generic/odp_packet_io.c
```diff
@@ -59,11 +59,10 @@ int odp_pktio_init_global(void)
 	shm = odp_shm_reserve("odp_pktio_entries",
 			      sizeof(pktio_table_t),
 			      sizeof(pktio_entry_t), 0);
-	pktio_tbl = odp_shm_addr(shm);
-
-	if (pktio_tbl == NULL)
+	if (shm == ODP_SHM_INVALID)
```
Bill-Fischofer-Linaro: This technically isn't needed since `odp_shm_addr(ODP_SHM_INVALID)` will return `NULL` in linux-generic, but I agree this is cleaner and more straightforward.

muvarov:
Merged. Patch number 3 subject changed  (forgot to change it for pushed version).

PR #245, opened 2017-10-20
[PATCH API-NEXT v3] Implement checksum insertion override for dpdk pktio

bogdanPricope:
Implement checksum insertion override generic functions and processing for dpdk pktio.

### DIFF
--- a/platform/linux-generic/odp_packet.c
+++ b/platform/linux-generic/odp_packet.c
@@ -1292,6 +1292,22 @@ uint16_t odp_packet_ones_comp(odp_packet_t pkt, odp_packet_data_range_t *range)
 	return 0;
 }
 
+void odp_packet_l3_chksum_insert(odp_packet_t pkt, int insert)
+{
+	odp_packet_hdr_t *pkt_hdr = packet_hdr(pkt);
+
+	pkt_hdr->p.output_flags.l3_chksum_set = 1;
+	pkt_hdr->p.output_flags.l3_chksum = insert;
+}
+
+void odp_packet_l4_chksum_insert(odp_packet_t pkt, int insert)
+{
+	odp_packet_hdr_t *pkt_hdr = packet_hdr(pkt);
+
+	pkt_hdr->p.output_flags.l4_chksum_set = 1;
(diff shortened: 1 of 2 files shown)


platform/linux-generic/odp_packet.c
```diff
@@ -1259,6 +1259,22 @@ int odp_packet_l4_offset_set(odp_packet_t pkt, uint32_t offset)
...
 
+void odp_packet_l3_chksum_insert(odp_packet_t pkt, int l3)
+{
+	odp_packet_hdr_t *pkt_hdr = packet_hdr(pkt);
+
+	pkt_hdr->p.output_flags.l3_chksum_set = 1;
+	pkt_hdr->p.output_flags.l3_chksum = l3;
+}
+
+void odp_packet_l4_chksum_insert(odp_packet_t pkt, int l4)
```
muvarov: maybe it's reasonable to change api at the same time to:
void odp_packet_l3_chksum_insert(odp_packet_t pkt,  odp_bool_t enable)
?
this int l3, int l4 is confusion.
@psavol ?

lumag:
This should be reworked so that linux-generic provides default support for checksumming, if pktio doesn't have optimized implementation.

bogdanPricope:
@lumag  "Calling this function is always allowed but the checksum will not be
inserted if the packet is output through a pktio that does not have
the relevant pktout chksum bit set in the pktio capability."

Let’s start from the assumption that API was discussed and agreed on at the time was proposed and merged.

lumag:
@bogdanPricope sorry, I probably was confused between different API calls.
Yes, this looks good then.

platform/linux-generic/odp_packet.c
Bill-Fischofer-Linaro: Agree with @muvarov 

Perhaps rework this in light of PR #264?

muvarov:
@lumag do you want to add your Review-by ?

lumag:

Bill-Fischofer-Linaro:
@bogdanPricope Please rebase to resolve conflicts in `odp_packet.c`

muvarov:
Merged.

PR #250, opened 2017-10-22
WIP: restructure headers for ABI-compat/platform-optimized modes

lumag:
This is a preview of headers restructuring.
Major changes:
 - select proper headers basing on include path during ODP compilation
 - install only required headers, not a mix of ABI and platform-specific headers
 - no more static_inline.h header (mostly accomplished)

I plan to finish it soon, but wanted to share accomplishments.


Bill-Fischofer-Linaro:
Checkpatch warnings for Part 27 are spurious.

Bill-Fischofer-Linaro:
This looks very good. Review needs input from several people:

* @psavol for changes to API spec files
* @WonderfulVoid for changes to atomics
* @heyi-linaro for sanity check to be sure this doesn't conflict with modular framework
* @bala-manoharan and @NikhilA-Linaro to assess impact to their platform native implementations.

lumag:
ODP specification
=================

ODP specification consists of several types of files, which together provide
full list of types, values and functions that ODP implemention MUST provide.

ODP API specification
---------------------

These are the files from `include/odp/api/spec` directory. They specify a set
of functions that MUST be provided by ODP implementation. Doxygen comments
inside these files document requirements for ABI interface provided by an
implementation. An implementation MUST use these headers AS IS, without any
modifications to be compatible with ODP specification.

ODP ABI compatibility specification
-----------------------------------

These are the files from `include/odp/arch/@ARCH_ABI@/odp/api/abi/` directory.
They specify a set of types and values that MUST be used AS IS without any
modifications by an implementation if it supports and is compiled for
ABI-compatibility mode.

ODP default ABI headers
-----------------------

These are the files from `include/odp/api/abi-default` directory. They provide
default specification for ODP types and values. ODP ABI compatibility files
heavily depend on these headers. These files MUST NOT be changed by an
implementation. If an implementation is compiled for ABI-compatibility mode,
these headers should be used AS IS.

Additional ODP headers
======================

These are the headers provided by an ODP to supplement ODP specification.

ODP API headers
---------------

These are the files from `include/odp/api` directory. They glue together API
and ABI specification headers. Although they are not part of ODP specification
itself, they provide an easy way for an application to use ODP. An
implementation SHOULD use these headers AS IS unless it has strong reason not
to do so.

odp_api.h header
----------------

This header found at `include/odp_api.h` is a main entry point for an
application. It includes all ODP API headers, so that an application doesn't 
have to include them one by one. It is not a part of ODP specification

Platform-specific headers
=========================

Platform ABI headers
--------------------

These are the headers found at
`platform/@with_platform@/include-abi/odp/api/abi` directory. They are used by
the reset of ODP code if and implementation is compiled with ABI compatibility
disabled. They should implement at least a set of types and values documented
in ODP API specification headers. They are permitted to provide any
platform-specific optimizations (i.e. they might provide types and/or values
that map directly onto the hardware details, they might provide inline
functions to speed up execution of the application, etc). These headers MAY use
ODP default ABI headers if they do fit.

Rest of platform-specific headers
---------------------------------

Platform MAY provide additional headers at `platform/@with_platform/include`.
However these headers SHOULD NOT be used directly by an application, because
this will tie it to the exact implementation details. Platform ABI headers MAY
use these headers to implement platform-specific optimizations.

lumag:
@psavol I have updated PR with formal text description. For simplicity it is also provided as GitHub comment, so it should hit your mailbox.

muvarov:
I created doc to more easy see files  layout before and after changes:

https://docs.google.com/document/d/1qwRNbpTvPGBQEptx2YxEgEo5Ekgi80lO6prc7yXFu_s/edit?usp=sharing

muvarov:
Path /include/odp/arch/arm32-linux/odp/api/abi will have 2 empty directories. And it's clear that under arch/ can be only ABI files. I think we need to simplify this.  But that we have before changes.  Other changes just do copy and include default abi. It's more clean place holders to write ABI arch specific things. Needed to check how it will fit to packaging...

Bill-Fischofer-Linaro:
@muvarov Your doc link is non-sharable. Please post a sharable link for it.

muvarov:
@Bill-Fischofer-Linaro  can you try to click on link from github page? I can open it even if I'm not logged to gmail.

include/README
```diff
@@ -0,0 +1,76 @@
...
+
+ODP API specification
+---------------------
+
+These are the files from `include/odp/api/spec` directory. They specify a set
+of functions that MUST be provided by ODP implementation. Doxygen comments
+inside these files document requirements for ABI interface provided by an
+implementation. An implementation MUST use these headers AS IS, without any
+modifications to be compatible with ODP specification.
+
```
psavol: ... a set of function prototypes, types, type names, enumerations, etc that MUST be provided ...

... document requirements for API ... (== typo: not ABI)

psavol: Git commit log text missing for this patch

include/README
```diff
@@ -0,0 +1,76 @@
...
+ODP default ABI headers
+-----------------------
+
+These are the files from `include/odp/api/abi-default` directory. They provide
+default specification for ODP types and values. ODP ABI compatibility files
+heavily depend on these headers. These files MUST NOT be changed by an
+implementation. If an implementation is compiled for ABI-compatibility mode,
+these headers should be used AS IS.
+
```
psavol: .... They provide default specification for ODP types and values for ABI compatibility. CPU architecture specific ODP ABI compatibility files depend heavily ...

The last sentence can be left out. MUST NOT be changed == MUST be used AS IS.

include/README
```diff
@@ -0,0 +1,76 @@
...
+ODP API headers
+---------------
+
+These are the files from `include/odp/api` directory. They glue together API
+and ABI specification headers. Although they are not part of ODP specification
+itself, they provide an easy way for an application to use ODP. An
+implementation SHOULD use these headers AS IS unless it has strong reason not
+to do so.
+
```
psavol: ... Although they are not part of ODP specification itself, they provide an easy way for an implementation to re-use ODP API/ABI specification header files. ...

So, these all are invisible to application which just include odp_api.h. It's helpeful for implementation to use those as is.

include/README
```diff
@@ -0,0 +1,76 @@
...
+odp_api.h header
+----------------
+
+This header found at `include/odp_api.h` is a main entry point for an
+application. It includes all ODP API headers, so that an application doesn't 
+have to include them one by one. It is not a part of ODP specification
+
```
psavol: Application MUST include only odp_api.h, nothing else. So, the file existence is part of ODP spec, the content is not other than the entire API spec is visible through it.

include/README
```diff
@@ -0,0 +1,76 @@
...
+These are the headers found at
+`platform/@with_platform@/include-abi/odp/api/abi` directory. They are used by
+the reset of ODP code if and implementation is compiled with ABI compatibility
+disabled. They should implement at least a set of types and values documented
+in ODP API specification headers. They are permitted to provide any
+platform-specific optimizations (i.e. they might provide types and/or values
+that map directly onto the hardware details, they might provide inline
+functions to speed up execution of the application, etc). These headers MAY use
+ODP default ABI headers if they do fit.
+
```
psavol: typo: "the reset of ODP code if and implementation is compiled"
"if and" -> "if"

include/README
```diff
@@ -0,0 +1,76 @@
...
+Rest of platform-specific headers
+---------------------------------
+
+Platform MAY provide additional headers at `platform/@with_platform/include`.
+However these headers SHOULD NOT be used directly by an application, because
+this will tie it to the exact implementation details. Platform ABI headers MAY
+use these headers to implement platform-specific optimizations.
```
psavol: Application must use only odp_api.h

include/README
lumag: @psavol 
Ack for the first change. For the second change I meant that spec files document requirements for ABI without providing exact declaration, Maybe you can suggest better wording?

psavol:
It would good to get patches to the mailing list also. v1 is there, but other versions are missing.

lumag:
@psavol updated README according to your comments.

include/README
psavol: OK then something like this: "Content of some types and value of some enumerations are left undefined in API spec. These are defined either in ABI spec or implementation specific (non-ABI compatible) headers ."

lumag:
@psavol updated basing on your comments

Bill-Fischofer-Linaro:
@muvarov This is Google's new "security improvements" at work. You no longer can simply copy URLs you have to "get sharable link" which generates this for your doc: https://drive.google.com/open?id=1qwRNbpTvPGBQEptx2YxEgEo5Ekgi80lO6prc7yXFu_s

Bill-Fischofer-Linaro:
Some general comments based on @muvarov's [tree diagrams](https://drive.google.com/open?id=1qwRNbpTvPGBQEptx2YxEgEo5Ekgi80lO6prc7yXFu_s) that help show the effect of this PR. I agree with @psavol that we should agree on the final directory layout and then the particular patches to realize these changes becomes more straightforward. 

1. Having `abi-default` under `include/odp/api` seems inconsistent since all other ABI definitions that are under `include/odp/arch/<name>/odp/api/abi`.  It would seem to be more consistent to have an `include/odp/arch/default` for this.

2. Having "abi" inside the "api" directories also doesn't seem completely clean since they really are two separate things. Wouldn't it make more sense to have `odp/api` and `odp/abi` directories at the same level?

lumag:
On 27/10/17 23:20, Bill Fischofer wrote:
>     Having |abi-default| under |include/odp/api| seems inconsistent
>     since all other ABI definitions that are under
>     |include/odp/arch/<name>/odp/api/abi|. It would seem to be more
>     consistent to have an |include/odp/arch/default| for this.

No. `include/odp/arch` gets into include path only if ABI-compat was
selected. `include/odp/[api/]abi-default` should always be in include path.

>     Having "abi" inside the "api" directories also doesn't seem
>     completely clean since they really are two separate things. Wouldn't
>     it make more sense to have |odp/api| and |odp/abi| directories at
>     the same level?

We can discuss that, it is an easy change. I just followed previous design.

lumag:
>     Having |abi-default| under |include/odp/api| seems inconsistent
>     since all other ABI definitions that are under
>     |include/odp/arch/<name>/odp/api/abi|. It would seem to be more
>     consistent to have an |include/odp/arch/default| for this.

No. `include/odp/arch` gets into include path only if ABI-compat was
selected. `include/odp/[api/]abi-default` should always be in include path.

>     Having "abi" inside the "api" directories also doesn't seem
>     completely clean since they really are two separate things. Wouldn't
>     it make more sense to have |odp/api| and |odp/abi| directories at
>     the same level?

We can discuss that, it is an easy change. I just followed previous design.

Bill-Fischofer-Linaro:
Good point, however couldn't we codify that as odp/nonabi to sit alongside odp/abi?

Agree we should discuss this more. I'll add it to the agenda for Monday's ARCH call.

lumag:
We can reshuffle directories later, if there is fit for that.

lumag:
@Bill-Fischofer-Linaro @muvarov @psavol Any additional work to be done on this PR?

include/odp/api/ipsec.h
```diff
@@ -10,22 +10,14 @@
  * ODP IPSEC API - platform specific header
  */
 
-#ifndef ODP_PLAT_IPSEC_H_
-#define ODP_PLAT_IPSEC_H_
+#ifndef ODP_API_IPSEC_H_
+#define ODP_API_IPSEC_H_
```
muvarov: it will be good to name defines without ODP_ to keep this naming only for API defines. But we can do that clean up later.

lumag: Hmm. They are part of `<odp/api/FOO.h> space, so is define name.

muvarov:
@lumag I'm ok with changed. Petris comments for README only needed to be accounted.

lumag:
@muvarov they were fixed long ago. I have rebased the branch to remove conflicts.

lumag:
@Bill-Fischofer-Linaro @muvarov @psavol  This PR received another set of file conflicts. I will update it, but maybe we can merge it now, rather than later during TM development?

psavol:
I think this would be optimally done when the delta between master and api-next is in minimum. So, that it can be merged to both branches almost instantly. It would not be optimal to merge it e.g. first to api-next, and then continue api-next and master development for couple of months, since that would maximize the number of merge conflicts between master and api-next later on. 

Also I would not want to create dependency between this and e.g. ipsec. So, I'd merge all api changes for  tigermoth first into master, and after that merge this on both branches. So, that if for some reason this would need to be reverted (even temporarily) that would not require to revert ipsec from master as a side effect.

lumag:
@psavol Thanks for the explanation. I will try reworking this PR, so that further rebases will be less painful

lumag:
Split/reworked patches, rebased to remove conflicts

lumag:
Rebased to remove conflict

lumag:
Finished cpu.h and align.h. Now  each api header is combined from abi + spec files.
Dropped compiler.h, as it was used only by byteorder_inlines.h

lumag:
@muvarov As we are now after TM-RC1, this PR should be merged. Should I resubmit it against master or will you pick patches from this PR?

Bill-Fischofer-Linaro:
@lumag Please rebase to resolve recent forced update changes to api-next

lumag:
@Bill-Fischofer-Linaro done

lumag:
Fixed two checkpatch issues. Other seem to be wrong.

lumag:
@muvarov @psavol can we merge this PR?

muvarov:
I think it can go in to master I think. @psavol  as I remember you were ok with most of the changes.

lumag:
@muvarov I'll create a separate PR, rebasing this on top of master

lumag:
@muvarov created #401

PR #253, opened 2017-10-23
[PATCH v1] travis: fix doxygen error capture

muvarov:
grep in doxygen log, not in some test.log


muvarov:
Merged.

PR #255, opened 2017-10-24
[PATCH 2.0 v2] Remove duplicate pktio_ops subsystem implementation

bogdanPricope:
Remove duplicate pktio_ops subsystem implementation


platform/linux-generic/pktio/subsystem.c
```diff
@@ -58,7 +60,9 @@ ODP_SUBSYSTEM_CONSTRUCTOR(pktio_ops)
 #ifdef HAVE_PCAP
 	enable_link_pcap_pktio_ops = 1;
 #endif
+#ifdef ODP_PKTIO_SOCKET
 	enable_link_socket_pktio_ops = 1;
```
heyi-linaro: These are temporary flags, which I was planned to remove after Makefiles update to build all pktio_ops into real individual modules and use --whole-archive or --as-no-needed to link them in.

bogdanPricope: True... and should be removed at that point. This is an intermediary step so we can remove  duplicate pktio_ops subsystem implementation

platform/linux-generic/pktio/subsystem.c
```diff
@@ -44,15 +54,23 @@ ODP_SUBSYSTEM_CONSTRUCTOR(pktio_ops)
 #ifdef ODP_PKTIO_DPDK
 	enable_link_dpdk_pktio_ops = 1;
 #endif
+#ifdef ODP_PKTIO_IPC
 	enable_link_ipc_pktio_ops = 1;
+#endif
 	enable_link_loopback_pktio_ops = 1;
```
nagarahalli: Any reason to keep loopback pkt IO enabled all the time?

nagarahalli:
I think we need to get rid of platform/linux-dpdk/include/odp_packet_io_internal.h as well. The linux-dpdk's dpdk pktio (tongue twister) should be contained within platform/linux-dpdk/pktio/dpdk.c and platform/linux-dpdk/pktio/dpdk.h files.

bogdanPricope:
odp_packet_io_internal.h  is next on my ToDo list.

platform/linux-generic/pktio/subsystem.c
bogdanPricope: Right now, loopback pktio is built for both linux-generic and linux-dpdk (so I did not needed to disable it for one of the platforms). In the next step (see Yi's comment above), a pktio will be enable/disable at build time and all those flags will be removed.

nagarahalli:
@Bill-Fischofer-Linaro I am fine with the changes (assuming the future/promised changes will happen :)). Do you have any comments?

heyi-linaro:
Merged and closed.

PR #61, opened 2017-06-28
[PATCH v2] Documentation build fixes

lumag:
Several misc fixes for the documentation building resulting in major change: documentation can now be built in out-of-tree mode and so user guides building can be enabled in `AM_DISTCHECK_OPTIONS`

PR #258, opened 2017-10-25
[PATCH API-NEXT v1] api: packet: print data

psavol:
Added debug print function which includes packet payload bytes.
Printing starts from an offset, so it's easy to e.g. limit print to IPv4 header bytes only:
odp_packet_print_data(pkt, l3_offset(pkt), 20)

### DIFF
--- a/platform/linux-generic/odp_packet.c
+++ b/platform/linux-generic/odp_packet.c
@@ -1739,9 +1739,68 @@ void odp_packet_print(odp_packet_t pkt)
 		seg = odp_packet_next_seg(pkt, seg);
 	}
 
-	str[len] = '\0';
+	ODP_PRINT("%s\n", str);
+}
 
-	ODP_PRINT("\n%s\n", str);
+void odp_packet_print_data(odp_packet_t pkt, uint32_t offset,
+			   uint32_t byte_len)
+{
+	odp_packet_hdr_t *hdr = packet_hdr(pkt);
+	uint32_t bytes_per_row = 16;
+	int num_rows = (byte_len + bytes_per_row - 1) / bytes_per_row;
+	int max_len = 256 + (3 * byte_len) + (3 * num_rows);
+	char str[max_len];
+	int len = 0;
+	int n = max_len - 1;
+	uint32_t data_len = odp_packet_len(pkt);
+	pool_t *pool = hdr->buf_hdr.pool_ptr;
+
+	len += snprintf(&str[len], n - len, "Packet\n------\n");
+	len += snprintf(&str[len], n - len,
+			"  pool index    %" PRIu32 "\n", pool->pool_idx);
+	len += snprintf(&str[len], n - len,
+			"  buf index     %" PRIu32 "\n", hdr->buf_hdr.index);
+	len += snprintf(&str[len], n - len,
+			"  segcount      %" PRIu16 "\n", hdr->buf_hdr.segcount);
+	len += snprintf(&str[len], n - len,
+			"  data len      %" PRIu32 "\n", data_len);
+	len += snprintf(&str[len], n - len,
+			"  data ptr      %p\n", odp_packet_data(pkt));
+	len += snprintf(&str[len], n - len,
+			"  print offset  %" PRIu32 "\n", offset);
+	len += snprintf(&str[len], n - len,
+			"  print length  %" PRIu32 "\n", byte_len);
+
+	if (offset + byte_len > data_len) {
+		len += snprintf(&str[len], n - len, " BAD OFFSET OR LEN\n");
+		ODP_PRINT("%s\n", str);
+		return;
+	}
+
+	while (byte_len) {
+		uint32_t copy_len;
+		uint8_t data[bytes_per_row];
+		uint32_t i;
+
+		if (byte_len > bytes_per_row)
+			copy_len = bytes_per_row;
+		else
+			copy_len = byte_len;
+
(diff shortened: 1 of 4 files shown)


psavol:
The checkpatch error can be ignored since it's CamelCase on PRIu16

    CHECK: Avoid CamelCase: <PRIu16>
    #61: FILE: platform/linux-generic/odp_packet.c:1764:
    +			"  segcount      %" PRIu16 "\n", hdr->buf_hdr.segcount);

    total: 0 errors, 0 warnings, 1 checks, 78 lines checked

apalos:
Hi Petri there's no check on snprintf regarding the returned len. Is there any possibility in this code that the final buffer will be bigger than max_len?

muvarov:
that has to be odp helper, not need for that in odp api

platform/linux-generic/odp_packet.c
```diff
@@ -1739,9 +1739,68 @@ void odp_packet_print(odp_packet_t pkt)
...
+	uint32_t bytes_per_row = 16;
+	int num_rows = (byte_len + bytes_per_row - 1) / bytes_per_row;
+	int max_len = 256 + (3 * byte_len) + (3 * num_rows);
+	char str[max_len];
+	int len = 0;
+	int n = max_len - 1;
+	uint32_t data_len = odp_packet_len(pkt);
+	pool_t *pool = hdr->buf_hdr.pool_ptr;
+
+	len += snprintf(&str[len], n - len, "Packet\n------\n");
```
lumag: Is there any reason for printing into the string buffer, rather than calling `ODP_PRINT` directly?

platform/linux-generic/odp_packet.c
```diff
@@ -1739,9 +1739,68 @@ void odp_packet_print(odp_packet_t pkt)
...
+	len += snprintf(&str[len], n - len,
+			"  pool index    %" PRIu32 "\n", pool->pool_idx);
```
lumag: My understanding is that `PRIu32/PRIu16` are not required, you can always use `%u/%hu` instead. Only `PRIu64/PRIx64` are required, because exact `uint64_t` definition depends on platform ABI.

lumag:
Missed such function, when developing ODP IPsec implementation. It would be good to have it inside ODP, rather than helper, so that the ODP itself might use it.

platform/linux-generic/odp_packet.c
muvarov: we had some problems with that prints in the past. Arm build or -m32 do not remember exactly what it was.

lumag: @muvarov could you please be more specific? What kind of problems?

muvarov: @lumag  compilation errors. So we have comits:
 27a05a1d61f3b7e4d5

42d0b99c453f269f

d25d276c649f57021f

Bill-Fischofer-Linaro: It's never wrong to use the PRIxxx macros so that's the safest and most portable way to go. So I'm happy to see them used here and we should try to use them in new development that calls for these sort of conversions.

lumag: @muvarov they do not mention any compiler errors, so they are useless in trying to understand, which errors were there.

platform/linux-generic/odp_packet.c
psavol: Yes there is. This avoids interleaved output when two (or more) threads print simultaneously. Output is pretty useless if lines are interleaved. Also, the current packet and buffer print functions do the same.

platform/linux-generic/odp_packet.c
```diff
@@ -1739,9 +1739,68 @@ void odp_packet_print(odp_packet_t pkt)
...
+	len += snprintf(&str[len], n - len,
```
apalos: This is a standard pattern for buffer overflows. This code assumes snprintf returns the bytes it wrote which is not the case. If someone adds extra snprintf statements without increasing max_len this would lead to overflows.

psavol: This patter is used in many places already. Including the odp_packet_print() function which is just above this function in the same file. I think "someone" modifying a line also need to know what he's doing and update pre-calculated string length. Someone may insert a bug in many ways... with or without a check on these lines.

Anyway, the point in this patch series is to add a new API function, and provide a simple implementation and test for that. Once the new API is merged, the implementation maybe modified / improved many times. If this function needs those checks, then should all other similar functions in our code base. That code base clean up could be done as a separate patch set (once the correct snprintf pattern has been agreed in the community).

apalos: My only concern is that it is relatively easy to miss that on a future review(a commit that just adds snprintf lines without increasing the size), but we mostly agree that's why i added this only as a comment. 
I think we'll need a wrapper over snprintf to be safe in the future.

psavol: I understand the concern, but I don't want to invent a new snprintf pattern in this patch set (since the purpose is API change). If / when a new pattern is introduced, it should be then used all over, otherwise we'll have multiple patterns ...

platform/linux-generic/odp_packet.c
muvarov: @lumag  like here: https://answers.launchpad.net/gcc-arm-embedded/+question/293823 old gcc  looks like failed. But int usually  = uint32_t, so it can work on both cases. I prefer see exact type.

muvarov:
Merged.

PR #260, opened 2017-10-26
[PATCH API-NEXT v1] validation: shmem: compare info page size to all supported huge page sizes

matiaselo:
Compare odp_shm_info_t.page_size to all supported page sizes on the system.
This enables passing the test when shmem is not allocated using the
system's default huge page size.


muvarov:
Merged.

PR #261, opened 2017-10-26
[PATCH 2.0 v3] Make public modular framework and subsystem headers

bogdanPricope:
Make public modular framework and subsystem headers


lumag:
I tried to stand away from 2.0 as much as possible, but this looks like a wrong decision from my POV. MF headers should not be visible in the `include/odp/` space. They should not be a part of public API. They are related to one single implementation of ODP, are they not?

bogdanPricope:
MF and subsystems are public APIs and visible from outside. Idea of modular framework is to be flexible enough to permit interwork with customer defined (and not know at core library build time) subsystem implementations: MF APIs to register a subsystem implementation or to add a new subsystem + definition of each subsystem should be public.

bogdanPricope:
P.S> Don't stand away from 2.0! This is the future and your contribution is welcomed ;) !!

lumag:
@bogdanPricope this is what I expected and what I warned against during SFO17 discussion. They should not be a part of application visible space, because an application must not depend and/or care about exact details of ODP implementation. These headers are internal details of one particular ODP implementation. The rest of implementations should not be required to use and/or care about MF, if they do not want to do so.

bogdanPricope:
Yes and NO:
-	Application interface (include/odp/api) is unchanged and unaware of DDF and MF. This is a ‘north’ interface and is where you are right
-	DDF interface (include/odp/drv) is the interface towards DDF drivers/devices/devios/enumerator implementation (loaded as shared libs) and is ALREADY public. It is a ‘south’ interface and where you are wrong.
-	MF + subsystems interface (include/odp/mf, include/odp/subsystem) is the interface towards MF/custom pktio/custom pool/custom scheduler/etc implementations. It is also a ‘south’ interface and it also needs to be public.

lumag:
@bogdanPricope it should be public from the implementation point of view, but it should be separate from ODP headers. `include/odp` should contain only generic headers. Is there still a plan for unified repo with several implementations inside? It doesn't look logical, to have headers in global namespace, which are unused by the rest of implementations.

Regarding drv interface. Currently master branch contains only two headers in this public space, which is obviously not enough for drivers to work. Api-next branch contains more, but I failed to see any real use of the drivers API. Maybe these headers should be moved to linux-generic/linux-smth. platforms, where they depend. Or be moved to a separate helper library/implementation aside of ODP.

bogdanPricope:
@lumag  Include/odp/api/*, include/odp/drv/*, include/odp/mf/* and include/odp/subsystem/* are headers that relate to symbols exposed by ODP library.

As high-level design, there should be:
-	A core library, platform agnostic (only architecture dependent) compose from MF, DDF, (default) subsystems implementations and other platforms independent functionality (locks, list, logs, stats, etc)
-	A set of subsystem implementations (grouped by platform) e.g. dpdk pktio, dpdk pool, dpdk queues, ThX pktio, ThX pool, ThX Scheduler, NXP pktio, NXP pool, etc.


All those subsystem implementations need to be able to register into framework (plug in) and to implement corresponding interfaces: this is why you need those public headers.

lumag:
>   * A set of subsystem implementations (grouped by platform) e.g. dpdk
>     pktio, dpdk pool, dpdk queues, ThX pktio, ThX pool, ThX Scheduler,
>     NXP pktio, NXP pool, etc.

I never heard that there will be ThX foorbar or NXP foobar submodules.
Moreover during SFO17 discussions concerns were expressed about
performance and maintainability of such submodules. linux-thunderx is an
implementation that does not use DDF, MF etc. It is fine to have an
implementation using MF, but we should not force other implementations
into using MF. These headers should be private to the linux-modular
implementation.

-- 
With best wishes
Dmitry

bala-manoharan:
"A set of subsystem implementations (grouped by platform) e.g. dpdk pktio, dpdk pool, dpdk queues, ThX pktio, ThX pool, ThX Scheduler, NXP pktio, NXP pool, etc."

>> We did not conclude on these sub-systems during SFO17. I don't think this was agreed upon.
This will have serious performance issues.

bala-manoharan:
As Dmitry pointed out this framework should not be visible in the public API header. This will create confusion for ODP application users.

Bill-Fischofer-Linaro:
@lumag @bala-manoharan @bogdanPricope Excellent discussion. From my viewpoint:

- ODP to date consists of two main parts: An API specification, and a validation test suite against that specification.

- ODP also provided a reference implementation (linux-generic) that served as a starting point for other ODP implementations. It was never intended for production use.

What's changing in ODP 2.0 is that we're adding a third visible part of ODP:

- A production-grade architecture-agnostic implementation that supports a modular component structure that is intended to be the single distribution for ODP in the NFV space. 

This does not detract from what's gone before, but rather brings it into maturity. Others are still free to write their own ODP implementations from scratch, using only the ODP API specification and the validation test suite to verify conformance. But the hope is that others will choose to participate in the single ODP distribution by writing modules that can provide access to platform-specific acceleration capabilities within this general framework. Note that this single distribution does not imply that all such modular components need to be part of the ODP git repo. Rather, the ODP git repo forms the "ODP core" that is targeted to be distributed via RHEL, SUSE, etc. This core will dynamically seek out and load components based on platform discovery whether they are packaged with the core or distributed separately by their respective platform vendors. 

It's for this reason that the modular framework and associated infrastructure is now part of the "public face" of ODP 2.0. We just need to be careful in how the directory structure is organized so that these different aspects of the LNG ODP deliverable are not confused.

lumag:
@Bill-Fischofer-Linaro yes, it is a public interface of ODP 2.0 modular implementation. It is not a public interface of ODP. So it has to be kept outside of `include/odp` namespace and should be installed into namespace other than `<odp/something>`.

Also I still find ODP 2.0 name confusing. It seems even people working on it are mistreating it as the only ODP implementation. So we should reconsider such naming.

Bill-Fischofer-Linaro:
@lumag "ODP 2.0" is a working designation for what was formerly referred to as "ODP Cloud". At SFO17 the LNG-SC requested that that name be discontinued as confusing. 

What's new with ODP 2.0 is that we're building more then a simple "reference implementation" of ODP. This is intended to be the "standard" production-grade distribution of ODP that can be part of RHEL, SUSE, etc., which again is what the LNG-SC has requested. While this does not preclude the existence of other implementations of ODP, it will be the focus of ODP marketing efforts going forward.

The evolutionary strategy is to move from a collection of individual ODP implementations that are tailored to specific platforms and that are developed and distributed independently to a universal modular framework that can be a single distribution supporting platform-specific modules to achieve the same result. That's why the "ODP 2.0" moniker seems appropriate. 

Since this framework is integral to the ODP 2.0 structure, having it in the include/odp namespace seems reasonable, because those creating plug-in modules for this framework will be making use of it the same way they make use of other ODP interfaces.

Today ODP has two primary "customers": The application writer, and the ODP implementer. ODP 2.0 retains these but introduces a third customer: the module writer that is looking to extend this universal framework with platform-optimized components.

lumag:
@Bill-Fischofer-Linaro does that mean that all implementations not using modular framework will become second-grade citizens?
From your description I see nothing wrong with keeping MF inside ODP 2.0 implementation, not inside top-level `include/odp` dir. Top-level dir should be limited to files common to all ODP implementations (be that 2.0 or any other one). To which location it should be installed is a separate question.

Bill-Fischofer-Linaro:
@lumag The precise directory structure is less important than the principle of what's included in ODP. So if we want to put the component infrastructure elsewhere that's fine. 

I agree that what's included and where things go upon `make install` for ODP is a separate one from where they reside in the git source tree. We can have that discussion as well. The ODP 2.0 git repo will have two primary users:

- Embedded developers who want to customize ODP and create applications targeted to specific platforms.

- Platform distributors (_e.g.,_ Red Hat) who will be building the distribution package for ODP.

There may also still be ODP implementation writers who are using the git repo as merely a starting point for their own ODP implementations.

Most application developers will never use the git repo directly, but will simply write (and run) ODP applications using the installed copy of ODP 2.0 included in their Linux distribution

heyi-linaro:
Hi @bogdanPricope @Bill-Fischofer-Linaro @lumag 

I feel modular framework is for implementors and keep it in frameworks/modular, and of course it is not mandatory for implementors only optional. 

So for this PR I think it is no need to expose it in include/api, application writers should not aware of what implementations it is using? and applications do not require to decide which modules to load?

For how to organize ODP distros into odp-core package and variant implementation/plugins packages, we can have a more detailed design, but this should not require modular framework APIs to be visible for application writers, right?

Please suggest, best regards, Yi

bogdanPricope:
The paths, as proposed by this PR:
./include/odp/api
./include/odp/drv
./include/odp/mf
./include/odp/subsystem
Mf, subsystem and ddf headers are NOT exposed under ./include/odp/api as application writers don’t need to care about mf/ddf, indeed.
However, all headers need to be exposed under ./include/odp/ as are referring visible symbols of ODP library.

lumag:
@bogdanPricope they can go to `platform/linux-modular/include/odp-modular`.

bogdanPricope:
ddf/mf are part of the core library and not part of a "platform". Anyway, think of the "odp/platform/*" more like a "odp/modules/*"

lumag:
@bogdanPricope there is no 'core' library. We should not enforce internal design of vendor-specific implementations. If I remember our discussions, none of vendors agreed to provide only modules for modular ODP implementation and abandon completely their embedded implementation.

Bill-Fischofer-Linaro:
@lumag Actually, there is a core library in ODP 2.0. That's one of the main points of this development effort. RHEL and other distributions will not support multiple distributions of ODP, one for each vendor. They will support one for x86 and one for ARMv8. The ODP core is thus the supported component that is how others will see and access ODP on these distributions. The core, in turn, will allow itself to be extended to cover microarchitectural differences among platform vendors that share the AArch64 architecture. Doing so will require those vendors to supply modules using this framework.

This doesn't mean that others can't create their own ODP implementations, just that they shouldn't expect those implementations to be supported by the major Linux distributions. They would remain private distributions.

psavol:
1) This is a prime example why patch review should happen (in the open) on the mailing list. A review of rather trivial patch set turned into discussion about what ODP is: one API + one implementation, or one API + multiple implementations.

2) I don't think SC has decided that there will be only one ODP implementation for Linux distributions (or cloud in general). Also, I think none of the SoC vendors have requested just an implementation. Who requested the single ODP and when ?

3) 2.0 should not be used in this context. I'd reserve 2.0 for a major API change, which we (hopefully) won't have for a long time. Applications/users are interested on ODP API features and compatibility, not in implementation detail of one SW based implementation. Better name for the branch / new implementation would be e.g. odp-modular (as Dmitry suggests). 

4) the odp-modular branch needs to still demonstrate that:
   * it has customers: which users or SoC vendors need this ?
   * it is better for distros / packaging: e.g. why this technique is better than multiple ODP libs? Distros are built from many dynamically linked libraries already...
   * it has performance: how many CPU cycles does it cost to have additional layer of abstraction, no inline functions and SW based trade-offs (vs native implementations) ?
   * it can be maintained: if everything is dynamic, who ensures that a composition actually works? When I get ODP from a SoC vendor I know who to blame. When there's a mix and match of modules who's in charge ...

lumag:
@Bill-Fischofer-Linaro I thought that RHEL and others will not tolerate multiple source packages.
So it was proposed to provide single source package, incorporating several platforms, then during compilation provide several libraries, from which one can be selected during setup or runtime, as all libraries will use single ABI. My understanding was that so-called-2.0 is an implementation targeting installations with different hardware modules, thus the need for modular framework.

Bill-Fischofer-Linaro:
@lumag Correct. The single source package is the ODP git repo. They build the distribution from that. No, there is no compilation at install time. 

@psavol We are having this discussion on the mailing list. It's nicely organized under this PR.

nagarahalli:
MF allows us to define new subsystems. But the new subsystem cannot be defined by application writers as a subsystem requires corresponding ODP APIs as well. So, I do not think MF needs to be in the public space.

However, I think subsystems need to be in the public as one might want to write their own module for a given subsystem.

The subsystems can also be viewed as drivers. From this point of view, I suggest moving the subsystems under driver directory and not keeping MF in public.
./include/odp/drv/subsystem

nagarahalli:
@bala When you talk about performance degradation, which case are you talking about? If it is the NFV deployement use case, the degradation needs to be understood. If it is the embedded use case, we should not see any degradation, the MF is capable of removing any function pointer indirection during compilation.

MF presents a single interface (to write the modules for) for different use cases. The required tweaks/optimizations happen in the MF, so the modules do not have to change.

muvarov:
we need to write test similar to ./test/performance/odp_bench_packet.c for functions which were moved to  mf to see the difference.
In general PR looks good. Don't  like only idea that there is big code diff between current implementation and implementation with MF. Support both of them will be trickily.
Better place for MF it's subsystem includes can be defined later.
Discussion can be continued on mailing list. PR can be merged to not stop future development.
Changes touch interests of various groups of people. I think we can get agreement before we merge this to master.
Reviewed-by: Maxim Uvarov <maxim.uvarov@linaro.org>

Bill-Fischofer-Linaro:
What seems needed to address lingering performance concerns is validation that `--enable-abi-compat=no` performance is the same with or without the MF changes and that the differences with `--enable-abi-compat=yes` are negligible or improved.

It seems we need to address these questions sooner rather than later. Once that's done the question of where these directories reside become secondary as that's always subject to reorganization.

bogdanPricope:
@nagarahalli  Subsystems headers need access to MF symbols/defines (so to MF headers). MF headers need to be public as well.
Also, there is the case where a subsystem implementation want to use a custom subsystem (some offload functionality) that can have HW or SW implementation. Is that a valid usecase?

heyi-linaro:
Hi, @bogdanPricope 
I feel under odp/frameworks/modular it was already opened for implementors like driver developrs, just wondering why it must be under odp/include/ folder? 

In my understanding implementors like driver developers will work with odp repo, not odp distro packages, so they will see odp/frameworks/modular and later odp/frameworks/driver folders and can use them, but ODP application developers will not see them in distro packages at all?

Best Regards, Yi

heyi-linaro:
Yes, @bogdanPricope 
Your summarize are very clear, and I think this is the key point which leads to this discussion, in my understand, driver developers will use odp repo as their environment not the odp distro packages, which is tend to be used by odp application developers.

In my opinion at least in initial odp distro packaging we'd better only targeting odp application developers, and suggest driver or module developers to work under odp repo, otherwise it will cause confusion for application developers.

nagarahalli:
@bogdanPricope @heyi-linaro I suggest that we do not consider a lot of use cases. I think this change needs to be done in stages. Let us look at minimum common agreement and do that and then improve as the use cases become more clear.

My minimum agreement with this PR is: the subsystem header files need to move out of Linux-generic directory. When we discussed the directory structure for 2.0, we had talked about having a 'subsystem' directory at the MF level. I am fine with moving the subsystem related files into that directory.

Bill-Fischofer-Linaro:
As we discussed in today's call, I can see a case to be made for wanting an odp-dev distribution package to go along with the odp distribution package. The dev package would include the additional headers and components needed to write applications, drivers, and ODP subsystem modules while the "vanilla" ODP package would just be for running precompiled ODP applications. 

This would be consistent with how a lot of other projects are distributed.

heyi-linaro:
@Bill-Fischofer-Linaro @bogdanPricope 
Even for odp and odp-dev distro packages shall we keep them simply targeting ODP application developers? And for odp implementors, driver and module developers, they should use odp source repo as their working environment?

This will reduce complexity and possible confusions, in the future it can be added into odp-dev packaging on real requirement and full understanding from odp developers?

So for modular headers frameworks/modular/ subdirectory should be sufficient?

For subsystem headers the best way are auto-generated from api spec headers, but not installed into distro packages (dist_noinstall_header?), for now in case no time to develop this header autogen @bogdanPricope I agree on your approach here to move it out from platform/linux-generic/include and put them into odp/include/subsystem/ to be shared by both platforms (linux-generic and linux-dpdk).

thanks and best regards, Yi

lumag:
Packages can be split even further: into odp-dev with application headers, odphelper-dev with helper stuff and something like libodp-modules-dev providing headers for submodules build.

Bill-Fischofer-Linaro:
@lumag I don't see any advantage of trying to split the dev side into lots of tiny sub-packages. That's likely to be more annoying than useful to potential developers. We'd either have a single ODP package that includes everything or a main package that's used to provide runtime support for ODP applications plus an odp-dev package that's used for ODP development (of any flavor).

lumag:
@Bill-Fischofer-Linaro mostly to split application API, not-API  and non-application parts. Each implementation will provide equivalent of libodp-dev, some implementations (or even a separate helper instance) will provide libodphelper-dev and modular ODP implementations will provide libodp-modular-dev. This will be less confusing to application writers, who will know exactly what do they need and why.

Bill-Fischofer-Linaro:
@lumag There is no need for implementation-specific build files because of the ABI. I should be able to compile my application on Platform A and have that binary run unchanged on any other platform that conforms to that ABI.

You'd only need implementation-specific files if you're building specifically for that target platform using `--enable-abi-compat=no`. Whether you'd expect to do that using a common ODP distribution as opposed to working with the Git repo directly is something that's TBD. My guess is embedded users are more likely to be using the Git repo directly, especially if they want to customize the configuration limits, etc.

lumag:
@Bill-Fischofer-Linaro Interesting point. Is odp-thunderx coming from Ubuntu package an embedded implementation or not? On the other hand DRV/MF headers **are** implementation-specific files. I would assume that every application writer will install libodp-dev/libodphelper-dev, libodplinuxFOO and libodphelperBAR. With libodplinuxFOO being a virtual package provided by libodp-linux-genericFOO, libodp-linux-modularFOO, libodp-linux-thunderxFOO or other vendor implementation (maybe coming from the same source tree).

lumag:
Submodule developers will install libodp-dev, libodp-modules-dev (and required dependency libodp-linux-modularFOO).

Application writers/users should expect any ODP implementation (as long as it is built in ABI-compat mode) and submodule developers will know that they have to use libodp-modularFOO, because their code will not work with any other implementation.

Bill-Fischofer-Linaro:
@lumag The MF is just an API "switchboard". In itself it imposes no implementation model. If your implementation provides replacements for all of the APIs then you're effectively using your own self-contained ODP implementation (just as you do today). Today ODP requires every implementation to do this, which many do by copying and pasting modules from linux-generic, and then incurring the support costs associated with this fork.

What's different in 2.0 is that you don't have to do this. So implementations are free to supply replacement implementations only for specific APIs s if your implementation is OK with adopting the internals of the ODP core SW components that are relevant to those APIs. We expect many server-type platforms to fall into that category where many of the ODP APIs are implemented in SW with perhaps a few bits subject to HW acceleration. That way you're only supporting those components that are unique to your platform and leveraging the ODP upstream for everything else.

Drivers are similarly expected to be portable. I shouldn't have to have a separate driver for the same device for each platform that supports ODP. That quickly becomes unmanageable. Yes, that means there will need to be some standard interfaces for drivers that all platforms commit to support. Of course you're still free to "go it alone" and have custom drivers for all I/O devices, but while that may work in some embedded use cases I don't think that's a sustainable strategy as we move forward into the NFV world.

lumag:
@Bill-Fischofer-Linaro There is no c&p code in odp-thunderx. I think the same holds for odp-dpaa/odp-dpaa2.

include/Makefile.am
```diff
@@ -68,6 +69,11 @@ odpdrvspecinclude_HEADERS = \
 		  odp/drv/spec/std_types.h \
 		  odp/drv/spec/sync.h
 
+frameworksmodularspecincludedir= $(includedir)/frameworks/modular/spec
+frameworksmodularspecinclude_HEADERS = \
+		  frameworks/modular/spec/module.h \
+		  frameworks/modular/spec/list.h
```
nagarahalli: list.h should not be part of the 'spec' directory. Should be part of framework source directory.

Ideally, we should split module.h to contain only the macros that need to be exposed to subsystems and move the rest to framework source directory.

heyi-live:
Hi, @bogdanPricope for several northbound API subsystems declaration, the movement looks good, yes, they are interfaces, and they are really specs:

odp/include/subsystem/spec/pool_subsystem.h
odp/include/subsystem/spec/queue_subsystem.h
odp/include/subsystem/spec/schedule_subsystem.h

For modular framework, I feel not well to see a simple thing now scattered and complicated ^_^!
- they are not interface specs, so spec is not suitable here

odp/include/framework**~~s~~**/modular/**~~spec~~**/module.h
odp/include/framework**~~s~~**/modular/**~~spec~~**/list.h

Why not keep this simple thing in one place? odp/framework/modular/.c, .h

As we talked, in case we assume driver or module implementer uses odp repo as their development environment there is no need to expose these modular and driver frameworks. And as Bill mentioned we can write guide/document later.

For pktio_ops_subsystem.h it is internal interface of linux-generic, agree with @nagarahalli, better keep it there?

Actually I'm thinking "subsystem" could be a misleading word in modular framework, actually it is "interface", we can change that.

Thanks and best regards, Yi

nagarahalli:
I am not sure about the role of "driver and module developer". But, I would say, we would like to encourage "driver and module developer" to contribute to ODP. So, may be not exposing them into odp/include directory is ok.

The pktio_ops_subsystem.h file needs to be exposed to the drivers. Does it mean, the modular framework header files also need to be exposed to the drivers?

bogdanPricope:
I started this series of patches because I intended to provide a POC of custom pktio_type implementation, registered by a ddf driver.
https://github.com/bogdanPricope/odp/tree/2_0_custom_pktio_type_pr
I understand that Ilias and Mykyta are following similar model.

The ddf module is hosted as an ODP example so it needs to access only public headers & symbols: that is, replacing include of ‘odp_packet_io_internal.h’ with include of ‘odp_pktio_ops_subsystem.h’ (and implicitly ‘odp_module.h’), making relevant ODP symbols visible and cleanup ‘odp_pktio_ops_subsystem.h’ of all ‘odp internal’ stuff.

For me is clear that ‘odp_pktio_ops_subsystem.h’ needs to be public header.

include/Makefile.am
bogdanPricope: true, but this cannot be done in this patch.

bogdanPricope:
Let's postpone this PR until high level design is clear and we decide to start the repository restructure.

heyi-linaro:
Yes, I agree @bogdanPricope, we can wait for at least one driver in shape in platform/linux-generic/pktio/virtio or 82599 or netmdev etc, and that time I suppose we have frameworks/driver/* with more full-fledged DDF.

thanks and best regards, Yi

nagarahalli:
@heyi-linaro In the 2.0 meeting today, we decided that we will make the changes for subsystem header files since we have an agreement on that. We decided to postpone the framework and driver header file changes. Bogdan will change this PR to just include the subsystem header file changes.

heyi-linaro:
Great, thanks!

doc/application-api-guide/Doxyfile
```diff
@@ -5,6 +5,7 @@ PROJECT_NUMBER = $(VERSION)
 PROJECT_LOGO = $(SRCDIR)/doc/images/ODP-Logo-HQ.svg
 INPUT = $(SRCDIR)/doc/application-api-guide \
 	$(SRCDIR)/include
+EXCLUDE = $(SRCDIR)/include/subsystem
```
nagarahalli: Can you please create a github issue to track the documentation?

heyi-linaro:
Merged and closed.

PR #264, opened 2017-10-27
[PATCH API-NEXT v2] api: packet: change argument to enable for l3 and l4 chsums

muvarov:
function name already has l3 and l4 layer, argument enable
is more clean here.

### DIFF
--- a/include/odp/api/spec/packet.h
+++ b/include/odp/api/spec/packet.h
@@ -1414,10 +1414,10 @@ odp_packet_chksum_status_t odp_packet_l4_chksum_status(odp_packet_t pkt);
  * the relevant pktout chksum bit set in the pktio capability.
  *
  * @param pkt     Packet handle
- * @param l3      0: do not insert L3 checksum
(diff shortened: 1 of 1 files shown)


bogdanPricope:
This API is already confusing as it is:
-	Operation name: ‘checksum insert override’
-	Operation performed: overrides ‘higher level configuration’: once you call it, you are overriding pktio level settings (regardless of the value of the argument)
-	Argument meaning: 
o	“enable” = 0 : do not insert checksum
o	“enable” = 1: do insert checksum

I suspect that “enable” is also confusing as may be interpreted as “enable override” / “disable override”

I suggest:

void odp_packet_l3_chksum_override(odp_packet_t pkt, int insert);

psavol:
Agree. "Insert" is better choice than "enable".

muvarov:
@bogdanPricope @psavol  enabled because "override is enabled".  I.e. from function name  odp_packet_l3_chksum_override enabled.  That looks like more logic then insert for me.

muvarov:
Merged.

PR #269, opened 2017-11-02
[PATCH API-NEXT v2] Implement checksum validation status API

bogdanPricope:
Implement checksum validation status API: add flags, implement API, implement functionality for dpdk pktio and update odp_generator example application.

### DIFF
--- a/example/generator/odp_generator.c
+++ b/example/generator/odp_generator.c
@@ -784,15 +784,13 @@ static void print_pkts(int thr, odp_packet_t pkt_tbl[], unsigned len)
 static int gen_recv_thread(void *arg)
 {
 	int thr;
-	thread_args_t *thr_args;
 	odp_packet_t pkts[MAX_RX_BURST], pkt;
 	odp_event_t events[MAX_RX_BURST];
 	int pkt_cnt, ev_cnt, i;
-	interface_t *itfs, *itf;
+	odp_packet_chksum_status_t csum_status;
 
+	(void)arg;
 	thr = odp_thread_id();
-	thr_args = (thread_args_t *)arg;
-	itfs = thr_args->rx.ifs;
 
 	printf("  [%02i] created mode: RECEIVE\n", thr);
 	odp_barrier_wait(&barrier);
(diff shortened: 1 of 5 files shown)


example/generator/odp_generator.c
```diff
@@ -811,21 +809,14 @@ static int gen_recv_thread(void *arg)
...
 
-			if (odp_packet_has_ipv4(pkt)) {
-				if (itf->config.pktin.bit.ipv4_chksum) {
-					if (odp_packet_has_l3_error(pkt))
-						printf("HW detected L3 error\n");
-				}
-			}
+			csum_status = odp_packet_l3_chksum_status(pkt);
+			if (csum_status == ODP_PACKET_CHKSUM_BAD)
+				printf("L3 checksum error detected.\n");
```
muvarov: prints for fast path functions is bad thing. Bellow is print_pkts() which  does all printing. Or we need to move it here at least. Or better to move out all printing from fast path.

platform/linux-generic/pktio/dpdk.c
```diff
@@ -337,39 +337,62 @@ static struct rte_mempool_ops ops_stack = {
...
+			pkt_hdr->p.error_flags.l3_chksum = 1;
+		}
 	}
 
 	if (pktin_cfg->bit.udp_chksum &&
-	    HAS_L4_PROTO(mbuf, RTE_PTYPE_L4_UDP) &&
-	    HAS_L4_CSUM_FLAG(mbuf, PKT_RX_L4_CKSUM_BAD)) {
-		if (pktin_cfg->bit.drop_udp_err)
-			return -1;
+	    HAS_L4_PROTO(mbuf, RTE_PTYPE_L4_UDP)) {
```
muvarov: too many ifs overcompicate function and makes it hard to read. Tcp packet and pktin_cfg->bit.udp_chksum=0 in settings will not to tcp 'if' bellow.

platform/linux-generic/pktio/dpdk.c
```diff
@@ -337,39 +337,62 @@ static struct rte_mempool_ops ops_stack = {
...
-		pkt_hdr->p.error_flags.tcp_err = 1;
+		if (packet_csum_result == PKT_RX_L4_CKSUM_GOOD) {
+			pkt_hdr->p.input_flags.l4_chksum_done = 1;
+		} else if (packet_csum_result != PKT_RX_L4_CKSUM_UNKNOWN) {
+			if (pktin_cfg->bit.drop_tcp_err)
+				return -1;
+
+			pkt_hdr->p.input_flags.l4_chksum_done = 1;
+			pkt_hdr->p.error_flags.tcp_err = 1;
+			pkt_hdr->p.error_flags.l4_chksum = 1;
```
muvarov: common errors set for tcp and udp here.

example/generator/odp_generator.c
bogdanPricope: Prints were already there. We only changed the conditions for printing to highlight the new API.

platform/linux-generic/pktio/dpdk.c
bogdanPricope: Cannot avoid the 'ifs'. Final code (not diffs) looks better and easier to understand.

platform/linux-generic/pktio/dpdk.c
bogdanPricope: error_flags.tcp_err and error_flags.udp_err are part of existing API.

platform/linux-generic/pktio/dpdk.c
muvarov: @bogdanPricope hm, please skip that comment, logic looks like correct.

muvarov: @bogdanPricope will that be the same and more clear https://github.com/muvarov/odp/commit/c56c3663725c73e7cf1a39fb6e37341dd2f3649f ? Also flags are uint16_t in my version of dpdk, not uint32_t.

bogdanPricope: This function should set csum related input/error flags for L3 (IPv4 packets) and for L4 UDP/TCP (IPv4/IPv6 packets); all flags should be set if requested at configuration time (pktin_cfg->bit.*). I don’t see how your function is doing this.
And yes, flags are uint64_t in dpdk 17.08/17.02. I will generate a new version for this.

struct rte_mbuf {
…………
uint64_t ol_flags;        /**< Offload features. */
…………
}

muvarov: @bogdanPricope  yes, 64 bit.  On than only change for 64 is needed.

bogdanPricope: 64 bit change was put in v2.

muvarov:
Merged.

PR #270, opened 2017-11-02
[PATCH v3] Doxygen fix and hooking into make distcheck

lumag:
- Fix one leftover from drv API removal.
- Hook doxygen-doc into distcheck to check that distribution contains all necessary files.


muvarov:
Merged with reordering patch 3 and 4.

PR #271, opened 2017-11-02
[PATCH API-NEXT v3] AES-CTR support

lumag:


lumag:
@Bill-Fischofer-Linaro Actually no. I implement tests for optional feature, making sure that they do not break the build if implementation provides no support for this feature, then I implement the feature.

AES-CTR provides an example of IPsec cipher, where length(salt) + length(IPSEC_IV) != length(cipher_IV).

Bill-Fischofer-Linaro:
@lumag OK, thanks. Still needs @psavol review for API change, but I'm Ok with this.

include/odp/api/spec/crypto.h
```diff
@@ -90,7 +90,10 @@ typedef enum {
 	ODP_DEPRECATE(ODP_CIPHER_ALG_AES128_CBC),
 
 	/** @deprecated  Use ODP_CIPHER_ALG_AES_GCM instead */
-	ODP_DEPRECATE(ODP_CIPHER_ALG_AES128_GCM)
+	ODP_DEPRECATE(ODP_CIPHER_ALG_AES128_GCM),
+
+	/** AES with counter mode */
+	ODP_CIPHER_ALG_AES_CTR,
```
psavol: I'd move this after AES_CBC or AES_GCM. It's easier to see what's deprecated when those are on the bottom.

include/odp/api/spec/crypto.h
```diff
@@ -169,6 +172,9 @@ typedef union odp_crypto_cipher_algos_t {
 		/** @deprecated  Use aes_gcm instead */
 		uint32_t ODP_DEPRECATE(aes128_gcm) : 1;
 
+		/** ODP_CIPHER_ALG_AES_CTR */
+		uint32_t aes_ctr     : 1;
+
```
psavol: Same here, move upwards. On top of deprecated.

include/odp/api/spec/crypto.h
lumag: Doing so would break binary compatibility. Is it ok?

psavol: Yes, since ABI compatibility is promised to be maintained within an API version, but not between API versions.

lumag:
@psavol updated to follow your feedback.

lumag:
@muvarov could you please merge it, I'd like to submit depending patches.

muvarov:
Merged.

PR #273, opened 2017-11-03
[PATCH API-NEXT v6] api: packet parse functions

psavol:
Application may need help on inner packet parsing e.g. after decrypt
or IP reassembly of a packet. Application specifies from where
to start and where to stop parsing. Implementation may be
vectorized (more efficient) when multiple packets are parsed
together with the same parse requirements.


include/odp/api/spec/packet.h
```diff
@@ -1140,6 +1185,48 @@ int odp_packet_move_data(odp_packet_t pkt, uint32_t dst_offset,
...
+ * @param proto   Protocol of the header starting at 'offset'
+ * @param layer   Continue parsing until this layer. Must be the same or higher
+ *                layer than the layer of 'proto'.
+ *
+ * @retval 0 on success
+ * @retval <0 on failure
+ */
+int odp_packet_parse(odp_packet_t pkt, uint32_t offset, odp_proto_t proto,
+		     odp_proto_layer_t layer);
+
```
Bill-Fischofer-Linaro: We had considered an `odp_packet_parse()` function some time back but it was rejected as something that would not fit well with hardware parsers. What's changed?

include/odp/api/spec/packet.h
```diff
@@ -1140,6 +1185,48 @@ int odp_packet_move_data(odp_packet_t pkt, uint32_t dst_offset,
...
+ * @param proto   Protocol of the header starting at 'offset'
+ * @param layer   Continue parsing until this layer. Must be the same or higher
+ *                layer than the layer of 'proto'.
+ *
+ * @retval 0 on success
+ * @retval <0 on failure
+ */
+int odp_packet_parse_multi(const odp_packet_t pkt[], const uint32_t offset[],
+			   int num, odp_proto_t proto, odp_proto_layer_t layer);
+
```
Bill-Fischofer-Linaro: What's the use case for a multi() form of this API? Might VPP use it? Perhaps Sachin can comment?

include/odp/api/spec/packet.h
psavol: The use case is mentioned in log message: parse after decrypt/IP reassembly. Application has recreated an inner packet and needs to parse it before continue. This is inherently SW parse which may be accelerated with CPU vector instructions, etc.

include/odp/api/spec/packet.h
psavol: A vector of packets is CPU vector instructions friendly.

psavol:
I'm not investing into the implementation before the API is approved.

nagarahalli:
I think, this should be a helper function rather than an API if there is no support for it from SoCs. This API is more specific to the application rather than the SoC. For ex: depending on the application, the API can be extended to add any number of higher layer protocols like RTP/RTCP/GTP-U etc. So, why not make it as part of the helper (similar to various protocol related things which are part of the helper currently)?

Bill-Fischofer-Linaro:
Longer term, we've talked about exploring P4 integration as a means of providing flexible parsing solutions for ODP applications that can also exploit HW acceleration. 

Not every ODP API maps directly to some HW feature, and certainly not on every platform on which ODP may be implemented. For example, even within the packet APIs today, most of the higher-level packet manipulation routines (add_data, copy_part, etc.) are most likely implemented in SW. The point is that there are platform-specific SW optimizations that the ODP implementation is better positioned to do than either the application or a generic helper routine, which is why having these higher-level and support functions as ODP APIs is the right place for them, if we're going to have them.

As long as an implementation and validation tests are provided I have no issue with including this in Tiger Moth as I believe Petri already indicated the expected use case for post-decryption, tunnel decapsulation, etc.

nagarahalli:
The packet manipulation APIs, especially the ones that involve memcpy, are better suited to be implemented by the application. The reason being, the application is aware of the size and alignment of the data to be copied which can be used to write more optimized memory copy function. Otherwise, a generic memcpy will have to cater for many more cases resulting in performance degradation.

I am not saying there are no use cases, I am questioning

1) Why this API needs to be part of ODP? This API is not providing any abstraction over a bunch of SoCs. It makes use of CPU core feature which is common to multiple SoCs and is already abstracted by the compiler.
2) What is the line between ODP's responsibility and application's responsibility?

Bill-Fischofer-Linaro:
APIs don't require any specific implementation. So you can't say that any API "requires memcpy" or any other implementation. That's the whole point of leaving how to best achieve the API semantics to the implementation. 

I think you also have it exactly backwards. The ODP application knows nothing of how an odp_packet_t is actually implemented on any given platform. The only thing it "sees" is what we expose through ODP APIs. That's the whole point of having abstract types. So the only thing an application can do is make use of sequences of other ODP API calls, which is likely less optimal than simply specifying what it wants directly and let the ODP implementation do that optimally.

ODP already has a parsing function that is logically a stage of classification. All this is doing is exposing that stage via a separate API. A logical follow-on would be an odp_packet_classify() API, but that's not being proposed now. 

As for your specific points:

1. What compiler abstraction/intrinsic currently knows how to deal with an odp_packet_t?  The answer is none because odp_packet_t is an abstract type. 

2. The application is responsible for deciding what to do with packets, ODP APIs are concerned with how to do that in a manner that is optimal for a given platform. The criteria is (a) is this function likely to be used in a fast path? and (b) would a given platform be able to do better than the application in providing an implementation? 

For these proposed APIs the answers to (a) and (b) are clearly "yes". Any function that may be called for every packet (or for every packet on a given flow type) can be considered a fast path function. And individual implementations may well be able to provide the function in a more optimal manner than the application. The examples given include the use of vector instructions, but it's also possible that a given platform's HW parse engine may be used. The point is the implementation now has the flexibility to decide how best to realize this fast path function. At worst it will be no slower than if the application did the work "by hand" but is more likely to be faster because the implementation can take advantage of packet internals in a manner not available to an ODP application.

include/odp/api/spec/packet.h
```diff
@@ -71,6 +71,51 @@ extern "C" {
...
+ * Protocol
+ */
+typedef enum odp_proto_t {
+	/** No protocol defined */
+	ODP_PROTO_NONE = 0,
+
+	/** Ethernet (including VLAN) */
+	ODP_PROTO_ETH,
+
+	/** IP (including IPv4 and IPv6) */
```
lumag: Maybe it would be better to split this into separate IPv4 and IPv6 packets. It would be an error to pass IPv6 packet with ethtype (or IP tunnel type) being set to IPv4.

include/odp/api/spec/packet.h
```diff
@@ -71,6 +71,51 @@ extern "C" {
...
+	ODP_PROTO_IP,
+
+	/** UDP */
```
lumag: What is the usecase for parsing a packet starting from L4 header? Also there are several (lots) of other L4 protocols. Do we want to support them all here?

Bill-Fischofer-Linaro: @lumag IPsec operating in transport mode is, I'd imagine, the main use case here.

include/odp/api/spec/packet.h
Bill-Fischofer-Linaro: That might be overly complicated since until a decrypted tunnel mode IPsec packet is parsed you don't know whether it's IPv4 or IPv6. It's parsing that makes that determination.

lumag: @Bill-Fischofer-Linaro In IPsec case Next Header field will contain 4 for IPv4 and 41 for IPv6.

include/odp/api/spec/packet.h
lumag: I felt easier to reparse both L3 and L4 headers in IPsec case, especially since transport mode ESP can en/decrypt some of L3 headers in IPv6 case.

include/odp/api/spec/packet.h
psavol: First bits of an IP header marks the version. So, it would be trivial for both app and implementation to read the version from the data. Common IP define is easier for application when a burst of packets may contain both v4 and v6 mixed. Application does not need to sort  packets into two arrays (one for v4 and other for v6) but just pass the entire array for parsing.

There are three ways to define the enumeration: IP, IPv4+IPv6, IP+IPv4+IPv6. I'm OK with any of those. IPv4+IPv6 would be a bit more robust since version information comes from two sources.

include/odp/api/spec/packet.h
psavol: It's there to enable application to call parsing in parts - e.g. first up to IP and then continue from L4. But since IP and transport protocols are tied together with pseudo headers, it's cleaner to remove L4 as a starting point.

include/odp/api/spec/packet.h
lumag: @psavol yes. I just wanted to focus on cases, when passing packet with wrong protocol is an error. E.g. IPv6 packet inside IPsec packet with NH = 4. So it is not a question of selecting proper L3 parser, but rather a question of nailing down error/malicious packets.

lumag: @psavol Also for multi-packet parsing, we can change `proto` to be an array, allowing applications to easily intermix IPv4 and IPv6 packets.

include/odp/api/spec/packet.h
lumag: Failure needs to be defined in a more precise way (and maybe for a single packet case). I assume that it means internal ODP error, rather than just packet with wrong headers. What happens in multi-packet case if failure occurs in the middle of parsing?

include/odp/api/spec/packet.h
lumag: @Bill-Fischofer-Linaro Because it is a parsing error, if lower layer provides us IP version that does not correspond to the in-packet version. We should detect that, rather than silently parsing this header.

lumag:
@psavol API now looks good, thank you. It looks like checksum flags won't be honoured by the implementation though.

include/odp/api/spec/packet.h
```diff
@@ -1140,6 +1179,82 @@ int odp_packet_move_data(odp_packet_t pkt, uint32_t dst_offset,
...
+			uint32_t tcp_chksum    : 1;
+
+		} check;
+
+		/** All check bits. This can be used to set/clear all flags. */
+		uint32_t all_check;
+	};
+
+} odp_packet_parse_param_t;
+
```
Bill-Fischofer-Linaro: Now that we have a `param` struct we should have an `odp_packet_parse_param_init()` API as well for completeness.

include/odp/api/spec/packet.h
```diff
@@ -1140,6 +1179,82 @@ int odp_packet_move_data(odp_packet_t pkt, uint32_t dst_offset,
...
+	union {
+		struct {
+			/** Check IPv4 header checksum */
+			uint32_t ipv4_chksum   : 1;
+
+			/** Check UDP checksum */
+			uint32_t udp_chksum    : 1;
+
+			/** Check TCP checksum */
+			uint32_t tcp_chksum    : 1;
```
lumag: What about just `l3_chksum`, and `l4_chksum`?

Bill-Fischofer-Linaro: +1 since other checksum APIs refer to L3 and L4 rather than specific protocols.

include/odp/api/spec/packet.h
muvarov: maybe return number of correctly parsed packets? And say that parsing always starts from be beginning of array.

include/odp/api/spec/packet.h
```diff
@@ -1140,6 +1179,82 @@ int odp_packet_move_data(odp_packet_t pkt, uint32_t dst_offset,
...
+ * Packet parse parameters
+ */
+typedef struct odp_packet_parse_param_t {
+	/** Protocol header at parse starting point. Valid values for this
+	 *  field are: ODP_PROTO_ETH, ODP_PROTO_IPV4, ODP_PROTO_IPV6. */
+	odp_proto_t proto;
+
+	/** Continue parsing until this layer. Must be the same or higher
+	 *  layer than the layer of 'proto'. */
+	odp_proto_layer_t layer;
```
muvarov: it looks like 2 enumbs are not needed here. odp_proto_layet_t  layer_start; and  odp_proto_layet_t  layer_end;  and one enum for layers.

nagarahalli:
@Bill-Fischofer-Linaro You have mentioned **most of the higher-level packet manipulation routines (add_data, copy_part, etc.) are most likely implemented in SW**.  You have also mentioned **APIs don't require any specific implementation. So you can't say that any API requires memcpy or any other function to implement. That's the whole point of leaving how to best achieve the API semantics to the implementation.**

In these cases, even though the APIs do not require any specific implementation, it is known that mostly it has to be implemented in software. If the SoC vendors agree for this, why to provide these APIs? The application does not know what odp_packet_t is, but we have APIs to get offsets into the packet data. That is enough for the application to manipulate the data. The application is better suited to do it as it is aware of what that memory holds and how much it needs to manipulate.  The requirements of packet manipulation will differ from application to application. So, if any implementation has to implement these APIs in software(without hardware acceleraion), then it has to cater to a large spectrum of applications. That will not be an optimized implementation for any specific application.

If I heard it right from @bala-manoharan and @NikhilA-Linaro, they mentioned they do not have any accelerators that this API can make use of. This means, it is a pure software implementation. ODP already provides APIs to get access to the packet data, why can't the application make use of those APIs and implement its own parsing function. The instructions to implement this parsing function are already abstracted by the compiler. There is no need for ODP to provide an abstraction for what compiler already provides.

In the future, I can imagine that this API will be extended to include other higher level protocol header fields (GTP-U in OFP, RTP, RTCP). Would it be performant enough to every application in that case?

Bill-Fischofer-Linaro:
@nagarahalli What an application does in SW will always be done in SW because the application would need to be redesigned to change that behavior. The point of having ODP APIs is to preserve the "innovation boundary" that allows for implementation evolution without requiring application redesign. So if today an implementation does packet parsing in SW that doesn't mean that tomorrow it can't migrate some or all of that functionality into HW. The application, by contrast, is always constrained to doing things entirely in SW.

An applications "understanding" of packets is at a high-level. For example, an application is better positioned to understand that because two flows are related a state change one one makes the information coming on another one no longer important. The implementation, by contrast, is better positioned to deal with the structure and internals of individual packets to perform packet processing and transforms as requested by the application. 

For example, `odp_packet_concat()` could clearly be done by the application by allocating packets and doing copies, but the point of having this as an ODP API is that the implementation may well be able to perform this function more efficiently. This is true even if the implementation is done in SW because it can manipulate packet internals in a way not available to ODP applications, as is the case in the linux-generic implementation of this function.

include/odp/api/spec/packet.h
```diff
@@ -1140,6 +1179,82 @@ int odp_packet_move_data(odp_packet_t pkt, uint32_t dst_offset,
...
+ * @param num     Number of packets and offsets
+ * @param param   Parse parameters. Proto and layer fields must be set. Clear
+ *                all check bits that are not used.
+ *
+ * @retval 0 on success
+ * @retval <0 on failure
+ */
+int odp_packet_parse_multi(const odp_packet_t pkt[], const uint32_t offset[],
+			   int num, const odp_packet_parse_param_t *param);
+
```
Bill-Fischofer-Linaro: For consistency with other `odp_xxx_multi()` APIs, the RC should indicate the number of input packets that were successfully processed.

platform/linux-generic/odp_packet.c
```diff
@@ -2252,6 +2252,63 @@ int packet_parse_l3_l4(odp_packet_hdr_t *pkt_hdr,
...
+}
+
+int odp_packet_parse_multi(const odp_packet_t pkt[], const uint32_t offset[],
+			   int num, const odp_packet_parse_param_t *param)
+{
+	int i;
+
+	for (i = 0; i < num; i++)
+		if (odp_packet_parse(pkt[i], offset[i], param))
+			return -1;
```
Bill-Fischofer-Linaro: `return i;` if we adopt the RC == number of successful packets parsed convention.

include/odp/api/spec/packet.h
psavol: Param init on fast path is a bit overkill. However, application could call it once and store the result. So, it could avoid extra function call for every packet. So, I'll add it.

include/odp/api/spec/packet.h
psavol: No, other APIs order checksum checking the same way: per protocol. Application may be e.g. terminating only UDP and forwarding all others - so it would be waste to check e.g. TCP checksum.

typedef union odp_pktin_config_opt_t {
	/** Option flags */
	struct {
		/** Timestamp all packets on packet input */
		uint64_t ts_all        : 1;

		/** Timestamp (at least) IEEE1588 / PTP packets
		  * on packet input */
		uint64_t ts_ptp        : 1;

		/** Check IPv4 header checksum on packet input */
		uint64_t ipv4_chksum   : 1;

		/** Check UDP checksum on packet input */
		uint64_t udp_chksum    : 1;

		/** Check TCP checksum on packet input */
		uint64_t tcp_chksum    : 1;

include/odp/api/spec/packet.h
lumag: @psavol `_param_init` seems like an overkill in this case. From my point of view, `param_init` are good for object creation params (where we might not know platform-optimized defaults) but are an overkill for operation functions (we do not have `param_init` for ipsec, crypto, etc operational params).

include/odp/api/spec/packet.h
psavol: Layer is not enough information for the parser. It needs to know the first protocol (e.g. MPLS vs IP) to be able to start the parsing.

include/odp/api/spec/packet.h
```diff
@@ -1140,6 +1179,82 @@ int odp_packet_move_data(odp_packet_t pkt, uint32_t dst_offset,
...
+/**
+ * Parse packet
+ *
+ * Parse protocol headers in packet data. Parsing starts at 'offset', which
+ * is the first header byte of protocol 'param.proto'. Parameter 'param.layer'
+ * defines the last layer application is interested about.
+ * Use ODP_PROTO_LAYER_ALL for all layers. The operation sets or resets packet
+ * metadata for all layers from the layer of 'param.proto' to the application
+ * defined last layer. Metadata of other layers have undefined values.
```
bogdanPricope: What about: "Metadata of other layers remains unmodified."?

include/odp/api/spec/packet.h
```diff
@@ -1140,6 +1179,82 @@ int odp_packet_move_data(odp_packet_t pkt, uint32_t dst_offset,
...
+ * When operation fails, metadata of all protocol layers have undefined values.
```
bogdanPricope: What about: "When operation fails, metadata of checked layers indicates the error condition."?

include/odp/api/spec/packet.h
psavol: This function is different from e.g. enqueue multi, where the destination queue may be close to full and that's why only part of events were enqueued. Here operation fails due to bad params or internal error.

Anyway, I'll change it to return num processed. It's more flexible.

include/odp/api/spec/packet.h
lumag: @psavol ok, agreed

include/odp/api/spec/packet.h
psavol: Application calls parse because metadata is not valid to begin with. There would not be much benefit for application to require that implementation saves old metadata values before the operation and then restores those after if e.g. an internal error occurred. Implementation performance would be hurt by making a copy of the old metadata.

Since it's just (L2/L3/L4) metadata (not data) that may have changed, application can continue processing the packet data also after a failed parse.

include/odp/api/spec/packet.h
psavol: Fail means an operation error (bad param, internal error), not e.g. error in packet header formats.

include/odp/api/spec/packet.h
psavol: Then I don't add it now. We can add param init also later if param struct is extended with other things than check flags. Todays spec defines already that all flags should be set to zero.

 * @param param   Parse parameters. Proto and layer fields must be set. Clear
 *                all check bits that are not used.

psavol:
Parse implementation is as good as packet input parse, which also does not check checksums - except when DPDK has done it. Checksum check implementation should be updated for the entire code base, but can be done in master rather than in this patch set (in api-next) which introduces the new APIs.

lumag:
@psavol ok for checksumming. I think that parsing (especially in software pktio) needs to be partially rethought anyway.

include/odp/api/spec/packet.h
bogdanPricope: Processing of UDP/TCP (in OFP and BSD like) depends on IPv4/IPv6: there are different input functions for UDP/TCP processing for IPv4 and IPv6 (https://github.com/OpenFastPath/ofp/blob/master/src/ofp_in_proto.c, https://github.com/OpenFastPath/ofp/blob/master/src/ofp_in6_proto.c)

include/odp/api/spec/packet.h
bogdanPricope: OK

psavol:
v5 == fixed clang build

include/odp/api/spec/packet.h
bogdanPricope: Then: "When operation fails, metadata of all protocol layers remains unchanged." ?

psavol:
v6 == fixed doxygen warning

include/odp/api/spec/packet.h
Bill-Fischofer-Linaro: Parsing can fail because the packet is malformed in some way. So parsing is also an important integrity check.

muvarov:
Merged.

PR #274, opened 2017-11-03
[PATCH v2] linux-gen: crypto: fix openssl_lock pointer type

psavol:
Wrong pointer type (pointer to pointer) to openssl_lock array
caused data overlapping when running on other than 64-bit
machines.

Fixed pointer type and simplified global data structure to
contain only one dynamically sized array (openssl locks).
Session array was already fixed size but was not defined as
such.

Signed-off-by: Viktor Tikkanen <viktor.tikkanen@nokia.com>

### DIFF
--- a/platform/linux-generic/odp_crypto.c
+++ b/platform/linux-generic/odp_crypto.c
@@ -83,9 +83,9 @@ typedef struct odp_crypto_global_s odp_crypto_global_t;
 
 struct odp_crypto_global_s {
 	odp_spinlock_t                lock;
-	odp_ticketlock_t **openssl_lock;
 	odp_crypto_generic_session_t *free;
-	odp_crypto_generic_session_t  sessions[0];
+	odp_crypto_generic_session_t  sessions[MAX_SESSIONS];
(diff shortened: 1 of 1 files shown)


Bill-Fischofer-Linaro:
As this is a bug fix, please open a bug and refer to it in the commit log as the fix for it. Thanks.

psavol:
https://bugs.linaro.org/show_bug.cgi?id=3411

Not sure why bug reports need to be done after bug has been fixed ...

Maxim, could you copy the bug report link during merge.

Bill-Fischofer-Linaro:
@psavol The purpose of the bug report is to make it easier to track which changes are enhancements / new features vs. bug fixes. A bug fix is a reason why someone might want to upgrade to a newer ODP level and it may not be relevant to other ODP implementations. It also makes putting the CHANGELOG together a lot easier. 

Easiest way to handle this after the fact is to simply update your commit log. That will generate a v2 and the existing review will convey to that. That will then be part of the commit log when the PR is merged without Maxim having to do anything special.

muvarov:
Merged.

PR #276, opened 2017-11-05
[PATCH v3] port features commits from api-next to next

muvarov:


muvarov:
hm, v3 did not go to Travis check and it shows status only for v2. But muvarov/odp/next passes tests. Will close this PR and reopen.

PR #277, opened 2017-11-06
[PATCH API-NEXT v8] api: random early detection and back pressure

bala-manoharan:
adds RED and BP configuration to both pool and queue parameters

### DIFF
--- a/include/odp/api/spec/classification.h
+++ b/include/odp/api/spec/classification.h
@@ -20,6 +20,7 @@ extern "C" {
 
 #include <odp/api/packet_io.h>
 #include <odp/api/support.h>
+#include <odp/api/threshold.h>
 /** @defgroup odp_classification ODP CLASSIFICATION
  *  Classification operations.
  *  @{
@@ -107,6 +108,61 @@ typedef union odp_cls_pmr_terms_t {
 	uint64_t all_bits;
 } odp_cls_pmr_terms_t;
 
+/** Random Early Detection (RED)
+ * Random Early Detection is enabled to initiate a drop probability for the
+ * incoming packet when the packets in the queue/pool cross the specified
+ * threshold values. RED is enabled when 'red_enable' boolean is true and
+ * the resource usage is equal to or greater than the minimum threshold value.
+ * Resource usage could be defined either as the percentage of pool being full
+ * or the number of packets/bytes occupied in the queue depening on the platform
+ * capabilities.
+ * When RED is enabled for a particular flow then further incoming packets are
+ * assigned a drop probability based on the size of the pool/queue.
+ *
+ * Drop probability is configured as follows
+ * * Drop probability is 100%, when resource usage >= threshold.max
+ * * Drop probability is 0%, when resource usage <= threshold.min
+ * * Drop probability is between 0...100 % when resource usage is between
+ *	threshold.min and threshold.max
+ *
+ * RED is logically configured in the CoS and could be implemented in either
+ * pool or queue linked to the CoS depending on platform capabilities.
+ * Application should make sure not to link multiple CoS with different RED or
+ * BP configuration to the same queue or pool.
+ */
+typedef struct odp_red_param_t {
+	/** A boolean to enable RED
+	 * When true, RED is enabled and configured with RED parameters.
+	 * Otherwise, RED parameters are ignored. */
+	odp_bool_t enable;
+
+	/** Threshold parameters for RED
+	 * RED is enabled when the resource usage is equal to or greater than
+	 * the minimum threshold value and is disabled otherwise
+	 */
+	odp_threshold_t threshold;
+} odp_red_param_t;
+
+/** Back pressure (BP)
+ * When back pressure is enabled for a particular flow, the HW can send
+ * back pressure information to the remote peer indicating a network congestion.
+ */
+typedef struct odp_bp_param_t {
+	/** A boolean to enable Back pressure
+	 * When true, back pressure is enabled and configured with the BP
+	 * parameters. Otherwise BP parameters are ignored.
+	 */
+	odp_bool_t enable;
+
+	/** Threshold value for back pressure.
+	 * BP is enabled when the resource usage is equal to or greater than the
+	 * max backpressure threshold. Min threshold parameters are ignored for
+	 * BP configuration.
+	 * @see odp_red_param_t for 'resource usage' documentation.
+	 */
+	odp_threshold_t threshold;
+} odp_bp_param_t;
+
 /**
  * Classification capabilities
  * This capability structure defines system level classification capability
@@ -135,6 +191,18 @@ typedef struct odp_cls_capability_t {
 
 	/** A Boolean to denote support of PMR range */
 	odp_bool_t pmr_range_supported;
+
+	/** Support for Random Early Detection */
+	odp_support_t random_early_detection;
+
+	/** Supported threshold type for RED */
+	odp_threshold_types_t threshold_red;
+
+	/** Support for Back Pressure to the remote peer */
+	odp_support_t back_pressure;
+
+	/** Supported threshold type for BP */
+	odp_threshold_types_t threshold_bp;
 } odp_cls_capability_t;
 
 /**
@@ -206,6 +274,12 @@ typedef struct odp_cls_cos_param {
 
 	/** Drop policy associated with CoS */
 	odp_cls_drop_t drop_policy;
+
+	/** Random Early Detection configuration */
+	odp_red_param_t red;
+
+	/** Back Pressure configuration */
+	odp_bp_param_t bp;
 } odp_cls_cos_param_t;
 
 /**
--- a/include/odp/api/spec/std_types.h
+++ b/include/odp/api/spec/std_types.h
@@ -15,6 +15,8 @@
 #ifndef ODP_API_STD_TYPES_H_
 #define ODP_API_STD_TYPES_H_
 #include <odp/visibility_begin.h>
+/* uint64_t, uint32_t, etc */
+#include <stdint.h>
 
 #ifdef __cplusplus
 extern "C" {
@@ -30,6 +32,14 @@ extern "C" {
  * regardless which compiler is used as this facilities interoperability
  * between e.g. different compilers.
  */
+
+/**
+ * Percentage type
+ * Use odp_percent_t for specifying fields that are percentages. It is a fixed
+ * point integer whose units are expressed as one-hundredth of a percent.
+ * Hence 100% is represented as integer value 10000.
+ */
+typedef uint32_t odp_percent_t;
 
 /**
  * @}
--- a/include/odp/api/spec/threshold.h
+++ b/include/odp/api/spec/threshold.h
@@ -1,0 +1,105 @@
+/* Copyright (c) 2017, Linaro Limited
+ * All rights reserved.
+ *
+ * SPDX-License-Identifier:     BSD-3-Clause
+ */
+
+/**
+ * @file
+ *
+ * ODP threshold descriptor
+ */
+
+#ifndef ODP_API_THRESHOLD_H_
+#define ODP_API_THRESHOLD_H_
+#include <odp/visibility_begin.h>
+#include <odp/api/std_types.h>
+
+#ifdef __cplusplus
+extern "C" {
+#endif
+
+/**
+ * Supported threshold types
+ *
+ * Supported threshold types in a bit field structure.
+ */
+typedef union odp_threshold_types_t {
+	/** bitfields for different threshold types */
+	struct {
+		/** Percentage of the total size of pool or queue */
+		uint8_t percent:1;
+
+		/** Total number of all transient packets */
+		uint8_t packet:1;
+
+		/** Total size of all transient packets in bytes */
+		uint8_t bytes:1;
+	};
+
+	/** All bits of the bit field structure */
+	uint8_t all_bits;
+} odp_threshold_types_t;
+
+/**
+ * ODP Threshold types
+ *
+ * Different types of threshold measurements
+ */
+typedef enum odp_threshold_type_t {
+	/** Percentage of the total size of pool or queue */
+	ODP_THRESHOLD_PERCENT,
+
+	/** Total number of all transient packets */
+	ODP_THRESHOLD_PACKET,
+
+	/** Total size of all transient packets in bytes */
+	ODP_THRESHOLD_BYTE
+} odp_threshold_type_t;
+
+/**
+ * ODP Threshold
+ *
+ * Threshold configuration
+ */
+typedef struct odp_threshold_t {
+	/** Type of threshold */
+	odp_threshold_type_t type;
+
+	/** Different threshold types */
+	union {
+		/** Percentage */
+		struct {
+			/** Max percentage value */
+			odp_percent_t max;
+
+			/** Min percentage value */
+			odp_percent_t min;
+		} percent;
+
+		/** Packet count */
+		struct {
+			/** Max packet count */
+			uint64_t max;
+
+			/** Min packet count */
+			uint64_t min;
+		} packet;
+
+		/** Sum of all data bytes of all packets */
+		struct {
+			/** Max byte count */
+			uint64_t max;
+
+			/** Min byte count */
+			uint64_t min;
+		} byte;
+	};
+} odp_threshold_t;
+
+#ifdef __cplusplus
+}
+#endif
+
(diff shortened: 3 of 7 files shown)


include/odp/api/spec/pool.h
```diff
@@ -228,6 +243,12 @@ typedef struct odp_pool_param_t {
 			    capability pkt.max_headroom.
 			    Use zero if headroom is not needed. */
 			uint32_t headroom;
+
+			/* Random Early Detection configuration */
+			odp_red_param_t red;
+
+			/* Back Pressure configuration */
+			odp_bp_param_t bp;
 		} pkt;
```
Bill-Fischofer-Linaro: Need to clarify under what conditions RED (and BP) apply to pools. If an application calls `odp_packet_alloc()` that should either succeed or fail. "Drop probability" doesn't make any obvious sense here. Tying these concepts to `odp_pktio_t` configurations seems more in keeping with what's needed, but I understand the difficulties that that caused in the previous PR.

Perhaps tying this to a `odp_cos_t` might work? An `odp_cos_t` has an associated queue and pool and these can then be viewed as admission controls to packets being added to a CoS. That would mean that these features would only be applicable if the classifier is being used, but is that a significant drawback since we want to encourage its use anyway?

include/odp/api/spec/threshold.h
```diff
@@ -0,0 +1,89 @@
...
+		} pkt_size;
+	};
+} odp_threshold_t;
+
+#ifdef __cplusplus
+}
+#endif
+
+#include <odp/visibility_end.h>
+#endif
```
Bill-Fischofer-Linaro: Adding new files requires updates to the various Makefile.am and related autotools. That's why Travis is reporting failures since these new files aren't able to be found / used.

include/odp/api/spec/pool.h
```diff
@@ -134,6 +136,19 @@ typedef struct odp_pool_capability_t {
 		 * The value of zero means that limited only by the available
 		 * memory size for the pool. */
 		uint32_t max_uarea_size;
+
+		/** Support for Random Early Detection */
+		odp_support_t random_early_detection;
+
+		/** Supported threshold type for RED */
+		odp_threshold_type_t threshold_red;
+
```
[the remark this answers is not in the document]
bala-manoharan: The implementation provides preference whether the RED and BP are configured either in pool or queue and provide support for which threshold method is supported using capability.
I believe this is sufficient from the implementation point of view.

include/odp/api/spec/pool.h
bala-manoharan: The previous PR was handling this RED and BP at cos level. There were some concerns from NXP since CoS is a logical entity and application configuration should make sure that multiple CoS with different RED configuration should not direct to the same pool or queue.

HW implements the RED and BP effectively either at the pool or the queue.
Since a pool represents the real bottleneck for resource exhausion and packet drop.
 We can discuss further in todays Public call.

include/odp/api/spec/queue.h
```diff
@@ -202,6 +216,11 @@ typedef struct odp_queue_param_t {
 	  * default size. */
 	uint32_t size;
 
+	/* Random early detection configuration */
+	odp_red_param_t red;
+
```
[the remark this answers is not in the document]
bala-manoharan: NXP is implementing RED at queue level. Hence I had added this based on the feedback from Nikhil.

include/odp/api/spec/queue.h
```diff
@@ -202,6 +216,11 @@ typedef struct odp_queue_param_t {
...
+	/* Back Pressure configuration */
+	odp_bp_param_t bp;
 } odp_queue_param_t;
```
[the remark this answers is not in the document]
bala-manoharan: Back pressure is to indicate the remote peer that there is a network congestion on this particular flow. Whether the remote peer takes respective action is beyond the scope of this RFC.

include/odp/api/spec/threshold.h
```diff
@@ -0,0 +1,89 @@
...
+ * Different types of threshold measurements
+ */
+typedef	enum {
+	/** Percentage of the total size of pool or queue */
+	odp_percent_e,
+	/** Number of transient packets */
+	odp_pkt_count_e,
+	/** Total size of all transient packets in bytes */
+	odp_pkt_size_e
+} odp_threshold_type_e;
```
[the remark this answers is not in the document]
bala-manoharan: I felt using a bitfield is more useful in exposing the platform capability whereas using an enum makes more sense during configuration. I am open for suggestions along these lines.

include/odp/api/spec/threshold.h
bala-manoharan: Oops. Will take care of that in next version.

platform/linux-generic/include/odp/api/std_types.h
```diff
@@ -29,6 +29,8 @@ extern "C" {
 
 typedef int odp_bool_t;
 
+typedef uint16_t odp_percent_t;
+
 /**
```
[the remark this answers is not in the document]
lumag: @bala-manoharan  @Bill-Fischofer-Linaro In my opinion, let's just push it into `odp/api/spec/std_types.h`.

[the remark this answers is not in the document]
bala-manoharan: They typedef for odp_bool_t is still under platform/linux-generic directory hence I had this here. I am fine pushing this to spec/std_types.h but is odp_bool_t going to be moved later?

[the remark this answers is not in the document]
lumag: No, `odp_bool_t` exact implementation is not part of the spec, as it is part of platform ABI. Percent type is just data, so from my point of view it should not be a part of ABI, but rather be a part of API spec.

muvarov:
@bala-manoharan Travis failed on Doxygen, please fix it.

platform/linux-generic/odp_classification.c
```diff
@@ -190,6 +190,8 @@ int odp_cls_capability(odp_cls_capability_t *capability)
 	capability->supported_terms.bit.tcp_sport = 1;
 	capability->supported_terms.bit.sip_addr = 1;
 	capability->supported_terms.bit.dip_addr = 1;
+	capability->random_early_detection = ODP_SUPPORT_NO;
+	capability->back_pressure = ODP_SUPPORT_NO;
```
Bill-Fischofer-Linaro: We shouldn't leave fields undefined in the returned `odp_cls_capability_t` struct. Either set `threshold_red` and `threshold_bp` explicitly or just clear the struct to zeros as the start of this routine.

platform/linux-generic/include/odp/api/std_types.h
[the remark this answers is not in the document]
bala-manoharan: Currently, I have followed the syntax existing in api-next. If ABI changes are merged first I will change my patch to match the specifications. If my patch gets merged first, change this as part of your ABI spec.

platform/linux-generic/odp_classification.c
bala-manoharan: Done.

muvarov:
@psavol  Petri, please review API patches.

bala-manoharan:
@psavol Petri, Could you please review this PR.

include/odp/api/spec/std_types.h
```diff
@@ -32,6 +32,13 @@ extern "C" {
  */
 
 /**
+ * @typedef odp_percent_t
+ * Use odp_percent_t for specifying fields that are percentages. It is a fixed
+ * point integer whose units are expressed as one-hundredth of a percent.
+ * Hence 100% is represented as integer value 10000.
+ */
+
```
psavol: Since API specifies already what odp_percent_t is, it's better do also the typedef here. Also the documentation should say that it's _unsigned_ integer. May be uint32_t is safer choice than uint16_t, so that percent calculations do not easily wrap around.

include/odp/api/spec/threshold.h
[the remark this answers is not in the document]
psavol: the enum: odp_threshold_type_t
the bitfield: odp_threshold_types_t

The enum values need to be UPPERCASE and contain a common prefix: ODP_THLD_PERCENT, ODP_THLD_PACKET, ODP_THLD_BYTE

If bitfield is only needed in one place (one capability struct) it could be defined there only (no typedef needed).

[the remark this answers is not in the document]
psavol: Another option for enums: ODP_THRESHOLD_PCT, ODP_THRESHOLD_PKT, ODP_THRESHOLD_BYTE

include/odp/api/spec/classification.h
```diff
@@ -107,6 +108,55 @@ typedef union odp_cls_pmr_terms_t {
...
+	/** A boolean to enable RED
+	 * When true, RED is enabled and configured with RED parameters.
+	 * Otherwise, RED parameters are ignored. */
+	odp_bool_t red_enable;
+
+	/** Threshold parameters for RED
+	 * RED is enabled when the resource limit is equal to or greater than
+	 * the maximum threshold value and is disabled when resource limit
+	 * is less than or equal to minimum threshold value. */
+	odp_threshold_t red_threshold;
```
psavol: I think it should be:
    * drop probability == 100%, when resource usage > threshold.max
    * drop probability == 0%, when resource usage < threshold.min

Now the text describe min/max as hysteresis: > max enables, < min disables. Which is not the intention, I guess.

bala-manoharan: The drop probability is 100% only when the pool is completely full i.e there is no further buffer to allocate packet. The intention of this description is that when resource usage is greater than threshold.max then the drop probability is enabled and packets will get dropped on an increasing drop probability and when it is less than min threshold then drop probability is disabled.

include/odp/api/spec/std_types.h
bala-manoharan: I can move the typedef here.
Not sure if I understand the need to move to uint32_t, I have only defined 100% as 10,000 and we should be fine with uin16_t. Any other specific reason for using uint32_t?

include/odp/api/spec/classification.h
Bill-Fischofer-Linaro: As we discussed during today's ARCH call. `max` and `min` can be renamed `start` and `stop` to more accurately reflect the actions. RED / backpressure begins when utilization hits the `start` threshold and is deasserted when utilization drops back to the `stop` threshold. So `start` >= `stop`.  If the two are equal it means there is no hysteresis.

The question arises whether a third `max` threshold value should exist which drops / backpressure is held at 100%. The idea here is to reserve a portion of the pool/queue resource for application use independent of use by PktIOs. Since this may not be feasible in all implementations this should probably be advertised with additional capability info.r

bala-manoharan:
@psavol Pls review V5 and provide your comments

include/odp/api/spec/threshold.h
```diff
@@ -0,0 +1,105 @@
...
+ * ODP Threshold types
+ *
+ * Different types of threshold measurements
+ */
+typedef	enum {
+	/** Percentage of the total size of pool or queue */
+	ODP_THRESHOLD_PERCENT,
+
+	/** Total number of all transient packets */
+	ODP_THRESHOLD_PKT,
```
psavol: When percent is not abbreviated, then this should not be either. So, _PERCENT and _PACKET, or PCT and PKT.

include/odp/api/spec/threshold.h
```diff
@@ -0,0 +1,105 @@
...
+		} percent;
+
+		/** Packet count */
+		struct {
+			/** Max packet count */
+			uint64_t max;
+
+			/** Min packet count */
+			uint64_t min;
+		} pkt_count;
```
psavol: This should match the type enum: packet

include/odp/api/spec/threshold.h
```diff
@@ -0,0 +1,105 @@
...
+
+		/** Packet size */
+		struct {
+			/** Max size of all packets */
+			uint64_t max;
+
+			/** Min size of all packets */
+			uint64_t min;
+		} pkt_size;
```
psavol: This should match the type enum: byte. Also better description is needed: e.g. "Sum of all data bytes of all packets". Packet size does not tell if it's the size of one or many packets.

include/odp/api/spec/classification.h
psavol: "Reaches between" is not valid since drop probability is defined also when <min and >max. "... when the packets in the queue/pool reaches between the specified threshold." -> "... when the packets in the queue/pool cross specified threshold values."

The "resource limit" needs to be specified: is it free or used space? Or is it different for pools and queues: free space in a pool, used space in a queue? May be the text is easier to understand with a bullet list:
* drop probability is 100%, when resource usage > threshold.max
* drop probability is 0%, when resource usage < threshold.min
* drop probability is between 0 ... 100%, when resource usage is between threshold.min and threshold.max

... and then define what "resource usage" means. Pools: space used, queues packet/bytes in queue

NikhilA-Linaro:
Reviewed-by: Nikhil Agarwal <nikhil.agarwal@linaro.org>

bala-manoharan:
@Bill-Fischofer-Linaro @muvarov This V7 incorporates review comments from Petri and has reviewed-by from Nikhil.
@psavol Pls review this version and provide your feedback

include/odp/api/spec/classification.h
psavol: red_ prefix can be dropped from the struct field names. When red param is used, the name of the variable is likely red or red_param, etc. So, after removing the prefix it looks like this

red.enable = 1;
red.threshold.packets.min = 1000;

vs 

red.red_enable = 1;
red.red_threshold....
...

psavol: "... RED is enabled when the resource limit is equal to ..."

This refers still to "resource limit", it should be "resource usage" which is a term that has been define above.

include/odp/api/spec/classification.h
```diff
@@ -107,6 +108,61 @@ typedef union odp_cls_pmr_terms_t {
...
+	 * the minimum threshold value and is disabled otherwise
+	 */
+	odp_threshold_t red_threshold;
+} odp_red_param_t;
+
+/** Back pressure (BP)
+ * When back pressure is enabled for a particular flow, the HW can send
+ * back pressure information to the remote peer indicating a network congestion.
+ */
+
```
psavol: Remove this extra line feed.

include/odp/api/spec/classification.h
```diff
@@ -107,6 +108,61 @@ typedef union odp_cls_pmr_terms_t {
...
+	 */
+	odp_bool_t bp_enable;
+
+	/** Threshold value for back pressure.
+	 * BP is enabled when queue or pool value is equal to or greater
+	 * than the max backpressure threshold.
+	 * Min threshold parameters are ignored for BP configuration.
+	 */
+	odp_threshold_t bp_threshold;
+} odp_bp_param_t;
```
psavol: bp_ prefixes can be removed from "bp_enable" and "bp_threshold".

"BP is enabled when queue or pool value is equal to or greater than the max backpressure threshold."
This should refer to "resource usage" (which has been defined on RED documentation). "queue and pool value" are  undefined terms.

include/odp/api/spec/std_types.h
psavol: Replace "@typedef odp_percent_t" doxygen tag with the short comment line ("Percent type"). The tag is not needed anymore as the typedef follows the comment.

Also, remove the extra line feed.

include/odp/api/spec/threshold.h
```diff
@@ -0,0 +1,105 @@
...
+	uint8_t all_bits;
+} odp_threshold_types_t;
+
+/**
+ * ODP Threshold types
+ *
+ * Different types of threshold measurements
+ */
+typedef	enum {
+	/** Percentage of the total size of pool or queue */
```
psavol: Change <tab> to <space> between typedef and enum. Also typically the typedef'ed enum/struct/union has also name, so that a debugger does not refer to it as "enum <no name> does not ...".

typedef enum odp_threshold_type_t {
  ...
} odp_threshold_type_t;

bala-manoharan:
@psavol V8 incorporates your review comments. Pls check and provide your feedback.

psavol:

bala-manoharan:
@muvarov could you pls merge this PR.

muvarov:
@bala-manoharan merging master to api-next then this can go.

muvarov:
Merged.

PR #278, opened 2017-11-06
[PATCH v3] port features commits from api-next to next  v2

muvarov:


muvarov:
hm actually it passes https://travis-ci.org/muvarov/odp/builds/298165062 interesting why here it's not :(

muvarov:
after restart everything is good. Looks like Travis internal bug where ./configure could not find openssl headers while packages were installed.

muvarov:
Merged.

PR #279, opened 2017-11-07
[PATCH API-NEXT v3] api: session level AAD length

psavol:
Added AAD length to crypto session level since it's likely
to be constant per session. Also some implementations (such
as DPDK) expect a constant AAD length at session create time.

### DIFF
--- a/include/odp/api/spec/crypto.h
+++ b/include/odp/api/spec/crypto.h
@@ -333,6 +333,14 @@ typedef struct odp_crypto_session_param_t {
 	 */
 	uint32_t auth_digest_len;
 
+	/** Additional Authenticated Data (AAD) length in bytes
+	 *
+	 *  AAD length is constant for all operations (packets) of the session.
+	 *  Set to zero when AAD is not used. Use odp_crypto_auth_capability()
+	 *  for supported AAD lengths. The default value is zero.
+	 */
+	uint32_t auth_aad_len;
+
 	/** Async mode completion event queue
 	 *
 	 *  The completion queue is used to return completions from
@@ -401,12 +409,11 @@ typedef struct odp_crypto_op_param_t {
 
 	/** Additional Authenticated Data (AAD) */
 	struct {
-		/** Pointer to ADD */
+		/** Pointer to AAD. AAD length is defined by 'auth_aad_len'
+		 *  session parameter.
+		 */
 		uint8_t *ptr;
 
-		/** AAD length in bytes. Use odp_crypto_auth_capability() for
-		 *  supported AAD lengths. */
-		uint32_t length;
 	} aad;
 
 	/** Data range to apply cipher */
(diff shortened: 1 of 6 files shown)


Bill-Fischofer-Linaro:
This seems reasonable but should get review input from @bala-manoharan and @NikhilA-Linaro for their respective platforms.

Bill-Fischofer-Linaro:
Also @GBalakrishna please comment from an odp-dpdk perspective.

include/odp/api/spec/crypto.h
```diff
@@ -310,6 +310,19 @@ typedef struct odp_crypto_session_param_t {
...
+	 *  operations (packets) of the session. When AAD length varies, this
+	 *  must be set to zero and operation level AAD length parameter is
+	 *  used instead. Both session and operation level parameters must not
+	 *  be used simultaneously.
+	 *
+	 *  Use odp_crypto_auth_capability() for supported AAD lengths.
+	 *  The default value is zero.
+	 */
+	uint32_t auth_aad_len;
+
```
Bill-Fischofer-Linaro: Why this restriction? Wouldn't it be cleaner to say this is simply a session-level default and if individual ops have a non-zero AAD they serve as overrides?

lumag:
@psavol maybe we can just drop individual AAD length?

psavol:
We'd need to check first that there's no common security protocol (IPSec AH, TLS, Wifi, Kasumi/Snow3g, ...) which would need per operation aad length. E.g. Figure 2 in RFC 4543 shows a variable length AAD ... which may be actually payload length. Also RFC 5246 defines AAD as

   The additional authenticated data, which we denote as
   additional_data, is defined as follows:

      additional_data = seq_num + TLSCompressed.type +
                        TLSCompressed.version + TLSCompressed.length;

... may be those types are fixed length per session.

lumag:
@psavol for TLS AAD is fixed length. I should check TLS 1.3, but I also think that it is fixed length.

RFC 4543 is an ugly beast. I have not decided yet how it should be implemented in ODP. My current plan is to use ENC_NULL + AUTH_GMAC settings plus use AAD pointer to pass pseudoheader data.

include/odp/api/spec/crypto.h
```diff
@@ -381,8 +394,15 @@ typedef struct odp_crypto_op_param_t {
...
+		/** AAD length in bytes.
+		 *
+		 *  Used only when AAD length is variable within the session.
+		 *  When AAD length is constant, this must be set to zero and
+		 *  session level AAD length parameter is used instead.
+		 *
+		 *  Use odp_crypto_auth_capability() for supported AAD lengths.
+		 *  The default value is zero.
+		 */
 		uint32_t length;
```
GBalakrishna: @psavol in dpdk 17.08 there is no support of variable aad length but only session specific aad length. ex: If application tries to set variable length instead of constant length, dpdk crypto operation will fail. Because once the crypto session is created with auth_aad_len=0, dpdk driver will use zero length in encryption_gcm & decryption_gcm.

Bill-Fischofer-Linaro: The spec says that if you are using a constant AAD length you must set that in the session AAD length, so the only reason why this still exists in the op parameters is to cover the (unlikely?) case of needing varying AAD lengths, which is a case that DPDK does not support.

lumag:
@psavol provided RFC4543 implementation as #288.

Bill-Fischofer-Linaro:
I'm OK with these changes as-is, but if @GBalakrishna feels the per-packet AAD lengths should be removed entirely for consistency with DPDK I'm OK with that too.

lumag:
@let's drop per-packet AAD length for now. We can reintroduce it later, if necessary.

psavol:
v2: removed per operation aad_len
v3: rebased

lumag:

muvarov:
Merged.

PR #280, opened 2017-11-07
[PATCH API-NEXT v3] Use chksum API

lumag:


helper/include/odp/helper/chksum.h
```diff
@@ -44,21 +44,7 @@ typedef enum {
...
-	if (len == 1)
-		sum += *(unsigned char *)buf;
-
-	sum = (sum >> 16) + (sum & 0xFFFF);
-	sum += (sum >> 16);
-	result = ~sum;
-
-	return  (__odp_force odp_u16sum_t) result;
+	return ~odp_chksum_ones_comp16(buffer, len);
 }
```
muvarov: for this patch I think more clean fix is to drop odph_chksum() completely. There is no need to implement helper function if it's simple follows odp function.

muvarov:
patch 2 is ok.

helper/include/odp/helper/chksum.h
lumag: Well... It is a part of ODP helper API, so we can not drop it easily.

muvarov: so v2 for this?

lumag: @muvarov this requires consent, as it will be an API change for header library. @psavol any thoughts?

muvarov: we do not hold api for helpers. They are free to modify.

muvarov: @lumag  we discussed this on todays meeting. It looks like nobody uses this helper in their app. So dessision is to drop it and we will take a not in changelog.

lumag:
@muvarov updated

Bill-Fischofer-Linaro:
Looks like Travis is failing most of the tests for v2. Needs rework.

lumag:
@muvarov should be fixed now

muvarov:
Merged.

PR #281, opened 2017-11-07
[PATCH v2] Misc fast path optimizations

matiaselo:
Improve fast path performance by reorganizing and cleaning buffer headers.

V2:
- Add pktio capability to pktio_entry_t (Maxim)

### DIFF
--- a/platform/linux-generic/include/odp/api/plat/packet_inlines.h
+++ b/platform/linux-generic/include/odp/api/plat/packet_inlines.h
@@ -21,6 +21,9 @@
 /** @internal Inline function offsets */
 extern const _odp_packet_inline_offset_t _odp_packet_inline;
 
+/** @internal Pool inline function offsets */
+extern const _odp_pool_inline_offset_t _odp_pool_inline;
+
 #if ODP_ABI_COMPAT == 1
 /** @internal Inline function @param seg @return */
 static inline uint32_t _odp_packet_seg_to_ndx(odp_packet_seg_t seg)
@@ -68,7 +71,9 @@ static inline uint32_t _odp_packet_tailroom(odp_packet_t pkt)
 /** @internal Inline function @param pkt @return */
 static inline odp_pool_t _odp_packet_pool(odp_packet_t pkt)
 {
-	return _odp_pkt_get(pkt, odp_pool_t, pool);
+	void *pool = _odp_pkt_get(pkt, void *, pool);
+
+	return _odp_pool_get(pool, odp_pool_t, pool_hdl);
 }
 
 /** @internal Inline function @param pkt @return */
@@ -98,7 +103,9 @@ static inline void *_odp_packet_user_area(odp_packet_t pkt)
 /** @internal Inline function @param pkt @return */
 static inline uint32_t _odp_packet_user_area_size(odp_packet_t pkt)
 {
-	return _odp_pkt_get(pkt, uint32_t, user_area_size);
+	void *pool = _odp_pkt_get(pkt, void *, pool);
+
+	return _odp_pool_get(pool, uint32_t, uarea_size);
 }
 
 /** @internal Inline function @param pkt @return */
--- a/platform/linux-generic/include/odp/api/plat/pool_types.h
+++ b/platform/linux-generic/include/odp/api/plat/pool_types.h
@@ -48,6 +48,19 @@ typedef enum odp_pool_type_t {
 
 #endif
 
+/** @internal Pool field accessor */
+#define _odp_pool_get(pool, cast, field) \
+	(*(cast *)(uintptr_t)((uint8_t *)pool + _odp_pool_inline.field))
+
+/** @internal Pool header field offsets for inline functions */
+typedef struct _odp_pool_inline_offset_t {
+	/** @internal field offset */
+	uint16_t pool_hdl;
+	/** @internal field offset */
+	uint16_t uarea_size;
+
+} _odp_pool_inline_offset_t;
+
 #ifdef __cplusplus
 }
 #endif
--- a/platform/linux-generic/include/odp_buffer_internal.h
+++ b/platform/linux-generic/include/odp_buffer_internal.h
@@ -107,9 +107,6 @@ struct odp_buffer_hdr_t {
 	/* User area size */
 	uint32_t uarea_size;
 
-	/* Max data size */
-	uint32_t size;
-
 	/* ipc mapped process can not walk over pointers,
 	 * offset has to be used */
 	uint64_t ipc_data_offset;
--- a/platform/linux-generic/include/odp_buffer_internal.h
+++ b/platform/linux-generic/include/odp_buffer_internal.h
@@ -104,16 +104,9 @@ struct odp_buffer_hdr_t {
 	/* User area pointer */
 	void    *uarea_addr;
 
-	/* User area size */
-	uint32_t uarea_size;
-
 	/* ipc mapped process can not walk over pointers,
 	 * offset has to be used */
 	uint64_t ipc_data_offset;
-
-	/* Pool handle: will be removed, used only for odp_packet_pool()
-	 * inlining */
-	odp_pool_t pool_hdl;
 
 	/* Data or next header */
 	uint8_t data[0];
--- a/platform/linux-generic/include/odp_packet_dpdk.h
+++ b/platform/linux-generic/include/odp_packet_dpdk.h
@@ -50,20 +50,21 @@ typedef union {
 typedef struct {
 	odp_pool_t pool;		  /**< pool to alloc packets from */
 	struct rte_mempool *pkt_pool;	  /**< DPDK packet pool */
-	odp_pktio_capability_t	capa;	  /**< interface capabilities */
 	uint32_t data_room;		  /**< maximum packet length */
-	uint16_t mtu;			  /**< maximum transmission unit */
-	/** Use system call to get/set vdev promisc mode */
-	odp_bool_t vdev_sysc_promisc;
-	uint8_t port_id;		  /**< DPDK port identifier */
 	unsigned min_rx_burst;		  /**< minimum RX burst size */
 	odp_pktin_hash_proto_t hash;	  /**< Packet input hash protocol */
-	odp_bool_t lockless_rx;		  /**< no locking for rx */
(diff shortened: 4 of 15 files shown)


platform/linux-generic/include/odp/api/plat/packet_inlines.h
```diff
@@ -21,6 +21,9 @@
 /** @internal Inline function offsets */
 extern const _odp_packet_inline_offset_t _odp_packet_inline;
 
+/** @internal Pool inline function offsets */
+extern const _odp_pool_inline_offset_t _odp_pool_inline;
```
lumag: This should go to pool_inlines.h

platform/linux-generic/include/odp/api/plat/pool_types.h
```diff
@@ -48,6 +48,19 @@ typedef enum odp_pool_type_t {
...
+
+/** @internal Pool header field offsets for inline functions */
+typedef struct _odp_pool_inline_offset_t {
+	/** @internal field offset */
+	uint16_t pool_hdl;
+	/** @internal field offset */
+	uint16_t uarea_size;
+
+} _odp_pool_inline_offset_t;
+
```
lumag: And this to pool_inline_types.h.

lumag:
@matiaselo first patches look good, Reviewed-by: Dmitry Eremin-Solenikov

Last one might clash with #250

platform/linux-generic/include/odp/api/plat/packet_inlines.h
matiaselo: pool_inlines.h would only include these two lines and looking at pool API there aren't any additional functions which should be inlined. Do you have a particular use case in mind for this header?

platform/linux-generic/include/odp/api/plat/pool_types.h
matiaselo: I followed the same pattern as the packet inlines. _odp_packet_inline_offset_t is defined in packet_types.h.

platform/linux-generic/include/odp/api/plat/packet_inlines.h
lumag: Just wanted to have pool functions in pool header, etc.

matiaselo: Since there are currently no functions in sight to put into pool_inlines.h I would suggest not to add a new header file. If/when some pool inline function are added later on the header can be created them. Is this OK for you?

lumag: @matiaselo ok.

platform/linux-generic/include/odp_buffer_internal.h
```diff
@@ -104,17 +104,10 @@ struct odp_buffer_hdr_t {
 	/* User area pointer */
 	void    *uarea_addr;
 
-	/* User area size */
```
nagarahalli: This change is done already in 2.0.

lumag: @nagarahalli Any reason why we shouldn't have it in master?

nagarahalli: It is about 2.0 merge to master, it will happen sometime in the future.

lumag: This is not planned untill TigerMoth.

platform/linux-generic/pktio/dpdk.c
```diff
@@ -1079,7 +1079,7 @@ static int dpdk_input_queues_config(pktio_entry_t *pktio_entry,
 				    const odp_pktin_queue_param_t *p)
 {
 	odp_pktin_mode_t mode = pktio_entry->s.param.in_mode;
-	odp_bool_t lockless;
+	uint8_t lockless;
```
muvarov: why is that needed?

matiaselo: While packing the pkt_dpdk_t struct I changed pkt_dpdk_t.lockless_rx and pkt_dpdk_t.lockless_tx to uint8_t. I'm simply using the same type here.

muvarov: @matiaselo  yes, I see. While reading patches in that web interface sometime I miss some pieces of code.  Just above you changed that.

matiaselo:
@nagarahalli In l2fwd this patch set provides ~2-4% improvement (mainly in scheduled mode).

platform/linux-generic/include/odp_buffer_internal.h
Bill-Fischofer-Linaro: 2.0 is based on api-next, not master. This will be one of several conflicts that will need to be resolved at merge time.

platform/linux-generic/include/odp/api/plat/pool_types.h
Bill-Fischofer-Linaro: @matiaselo I agree this is correctly placed here.

platform/linux-generic/include/odp/api/plat/packet_inlines.h
Bill-Fischofer-Linaro: The API being inlined is `odp_packet_pool()` so this is an artifact of the implementation of that API. So I'm OK with it being here since this is supposed to be fastpath code.

muvarov: extern is not needed in .h file, also extern 2 lines above.

platform/linux-generic/odp_packet.c
```diff
@@ -1780,26 +1778,28 @@ int _odp_packet_copy_md_to_packet(odp_packet_t srcpkt, odp_packet_t dstpkt)
...
 	if (dsthdr->buf_hdr.uarea_addr != NULL &&
-	    srchdr->buf_hdr.uarea_addr != NULL)
-		memcpy(dsthdr->buf_hdr.uarea_addr,
-		       srchdr->buf_hdr.uarea_addr,
-		       dsthdr->buf_hdr.uarea_size <=
-		       srchdr->buf_hdr.uarea_size ?
-		       dsthdr->buf_hdr.uarea_size :
-		       srchdr->buf_hdr.uarea_size);
+	    srchdr->buf_hdr.uarea_addr != NULL) {
+		memcpy(dsthdr->buf_hdr.uarea_addr, srchdr->buf_hdr.uarea_addr,
```
muvarov: odp_memcpy() might be good fit here.

platform/linux-generic/include/odp_packet_dpdk.h
```diff
@@ -49,16 +49,17 @@ typedef union {
...
 	odp_pktin_hash_proto_t hash;	  /**< Packet input hash protocol */
-	odp_bool_t lockless_rx;		  /**< no locking for rx */
-	odp_bool_t lockless_tx;		  /**< no locking for tx */
+	uint16_t mtu;			  /**< maximum transmission unit */
+	/** Use system call to get/set vdev promisc mode */
+	uint8_t vdev_sysc_promisc;
+	uint8_t lockless_rx;		  /**< no locking for rx */
+	uint8_t lockless_tx;		  /**< no locking for tx */
+	uint8_t port_id;			  /**< DPDK port identifier */
+	/* --- 34 bytes --- */
```
muvarov: what is propose of sayng 34 bytes here? Cache line is 64 so maybe comments for cache lines borders are more interesting here? vdev_sysc_promisc, lockless_rx, lockless_tx can be packet to bit field. That will also reduce structure size. Is there any reason to do that?

platform/linux-generic/include/odp_packet_dpdk.h
```diff
@@ -49,16 +49,17 @@ typedef union {
...
+	odp_pktio_capability_t	capa;	 /**< interface capabilities */
```
muvarov: pktio capability is big struct and it's same for all packets.  Is it really needs to be here? as well as uint16_t mtu? why not in pktio_entry_t ?

muvarov:
@matiaselo few questions in that PR.

platform/linux-generic/include/odp/api/plat/packet_inlines.h
matiaselo: In this case it is needed. _odp_pool_inline is defined in odp_pool.c and _odp_packet_inline in odp_packet.c. ODP build with '--disable-abi-compat' will fail without extern declarations.

platform/linux-generic/odp_packet.c
matiaselo: odp_memcpy() is just a wrapper for memcpy() and I would avoid using api functions in the implementation unless there is a clear benefit.

platform/linux-generic/include/odp_packet_dpdk.h
matiaselo: The next member in the struct named 'capa' is 40 bytes long, so it will cross a cache line boundary. I was thinking this comment was more useful than an ambiguous cache line comment.

There is still space in the first cache line for additional fast path variables, so I wouldn't use a bitmask before it is actually needed.

platform/linux-generic/include/odp_packet_dpdk.h
matiaselo: A valid point, at least I don't see any reason why capability struct is stored inside each pktio implementation. MTU is used in the fast path so I wouldn't move that. Should I move the capability in this PR or create a new one?

muvarov: I think if it will be in this request it will make easy to review all changes at once.

platform/linux-generic/odp_packet.c
muvarov: ok. I just thought maybe some other implementations already speedup this call and we can get this benefits for free. But I don't might if it will be simple memcpy().

platform/linux-generic/include/odp/api/plat/packet_inlines.h
muvarov: ok, then.

lumag: @muvarov extern is not needed for function prototypes.

muvarov:
Merged.

PR #282, opened 2017-11-07
[PATCH 2.0 v1] linux-gen: pktio: remove extra indirection for accessing pktio_entry fields

bogdanPricope:
Remove extra indirection for accessing pktio_entry fields by modifying interface to use inner structure.
Inner pktio_entry structure was renamed as a public odp symbol since is used by pktio_ops subsystem.

### DIFF
--- a/platform/linux-dpdk/pktio/dpdk.c
+++ b/platform/linux-dpdk/pktio/dpdk.c
@@ -41,7 +41,7 @@ pktio_table_t *pktio_tbl;
 /* Forward declaration */
 static pktio_ops_module_t dpdk_pktio_ops;
 
-static uint32_t mtu_get_pkt_dpdk(pktio_entry_t *pktio_entry);
+static uint32_t mtu_get_pkt_dpdk(odp_pktio_entry_t *pktio_entry);
 
 /* Test if s has only digits or not. Dpdk pktio uses only digits.*/
 static int _dpdk_netdev_is_valid(const char *s)
@@ -96,12 +96,12 @@ static void _dpdk_print_port_mac(uint8_t portid)
 		eth_addr.addr_bytes[5]);
 }
 
-static int input_queues_config_pkt_dpdk(pktio_entry_t *pktio_entry,
+static int input_queues_config_pkt_dpdk(odp_pktio_entry_t *pktio_entry,
 					const odp_pktin_queue_param_t *p)
 {
 	pktio_ops_dpdk_data_t *pkt_dpdk =
 		odp_ops_data(pktio_entry, dpdk);
-	odp_pktin_mode_t mode = pktio_entry->s.param.in_mode;
+	odp_pktin_mode_t mode = pktio_entry->param.in_mode;
 
 	/**
 	 * Scheduler synchronizes input queue polls. Only single thread
@@ -126,7 +126,7 @@ static int input_queues_config_pkt_dpdk(pktio_entry_t *pktio_entry,
 	return 0;
 }
 
-static int output_queues_config_pkt_dpdk(pktio_entry_t *pktio_entry,
+static int output_queues_config_pkt_dpdk(odp_pktio_entry_t *pktio_entry,
 					 const odp_pktout_queue_param_t *p)
 {
 	pktio_ops_dpdk_data_t *pkt_dpdk =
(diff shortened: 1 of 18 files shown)


platform/linux-generic/include/odp_packet_io_internal.h
```diff
@@ -33,7 +32,7 @@ extern "C" {
 
 #define PKTIO_MAX_QUEUES 64
 /* Forward declaration */
-typedef union pktio_entry_u pktio_entry_t;
+typedef struct pktio_entry odp_pktio_entry_t;
```
heyi-linaro: this data structure belongs to one specific pktio implementation, so should not be public?

I understand that native drivers depend on this symbol because of pktio_ops interface is operating on this data structure.

So in order to make future native drivers independent of specific pktio implementation, instead of opening this internal data structure definition, instead we should refactory the pktio_ops interface definition.

And this task we can do later, after we have 1 or 2 native drivers in shape, what do you think?

platform/linux-dpdk/pktio/dpdk.c
```diff
@@ -96,12 +96,12 @@ static void _dpdk_print_port_mac(uint8_t portid)
...
 }
 
-static int input_queues_config_pkt_dpdk(pktio_entry_t *pktio_entry,
+static int input_queues_config_pkt_dpdk(odp_pktio_entry_t *pktio_entry,
 					const odp_pktin_queue_param_t *p)
 {
 	pktio_ops_dpdk_data_t *pkt_dpdk =
 		odp_ops_data(pktio_entry, dpdk);
-	odp_pktin_mode_t mode = pktio_entry->s.param.in_mode;
+	odp_pktin_mode_t mode = pktio_entry->param.in_mode;
```
heyi-linaro: code 'union->s.member' compared to code 'struct->member':

they compiled into same assemble code, right?

and many odp data structure follow the union->s.member pattern, we may better keep it for consistency.

bogdanPricope:
I don’t know how much benefit, in terms of performance, will be due to this patch but the main gain is to have pktio_ops implementations to depend to only a portion of pktio_entry_t: odp_pktio_entry_t will be moved to a different header file that at some point in future will become ‘public’; some fields of odp_pktio_entry_t  (struct pktio_entry) may need to be moved to a ‘private’ part of pktio_entry_t. With all those planned changes in mind, removing extra indirection makes sense as a first step.

platform/linux-dpdk/pktio/dpdk.c
bogdanPricope: You should not access pktio_entry_t from odp_pktio_ops_subsystem.h (since odp_pktio_ops_subsystem.h will become 'kind of' public API) but only a subset of it. This patch Is not necessary for performance.

platform/linux-generic/include/odp_packet_io_internal.h
bogdanPricope: It is working right now. Why wait?

heyi-linaro:
Yes, @bogdanPricope understand your thoughts as:

native drivers --> pktio_ops interface --> pktio_entry_t (which is one implementation specific)

You plan to make part of pktio_entry_t public interface in the future, here we can also consider:

native drivers --> pktio_ops interface (refine the interface to be independent of impls) --> odp_pktio_t ?? 

So that's why I suggest to postpone this PR until we have 1 or 2 native drivers in shape and rethink about the pktio_ops interface refactory.

nagarahalli:
I have been making some experiments on this front. Few things I am thinking are:

1) Changing ops_data to a pointer. Right now we have a 80000B array there and it is big. I understand this change was done to allocate that memory from shared memory. We need to use the small memory allocater from shared memory (this is done in scalable queues). This will allow only the amount of memory used by a particular driver to be allocated.

2) Logically, the Pkt I/O APIs (odp_packet_io_.c) and the drivers(or ops) do not need to be aware of each other's structures. Pkt I/O APIs need to be aware of pointer to driver's data. So, is there a need to expose pktio_entry_t to the drivers? All the config data (which are part of ODP API definition) is being passed to the driver through pkt I/O ops. The drivers can construct their own data structure and return a pointer to it to the Pkt I/O APIs. I think this way, there is no need to make new structures public.

bogdanPricope:
Let's postpone this PR until pktio_entry structure is clear, replace ops_data array with pool memory and rework pktio_ops API. Can be done latter on a more stable code.

PR #285, opened 2017-11-08
[PATCH v2] next update version

muvarov:

### DIFF
--- a/configure.ac
+++ b/configure.ac
@@ -3,7 +3,7 @@ AC_PREREQ([2.5])
 # Set correct API version
 ##########################################################################
 m4_define([odpapi_generation_version], [1])
-m4_define([odpapi_major_version], [15])
+m4_define([odpapi_major_version], [16])
 m4_define([odpapi_minor_version], [0])
 m4_define([odpapi_point_version], [0])
 m4_define([odpapi_version],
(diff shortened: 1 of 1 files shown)


muvarov:
@lumag are you good in .so version increasing? Please review that patch.

configure.ac
```diff
@@ -36,10 +36,10 @@ AM_SILENT_RULES([yes])
 # 3. if interfaces were removed, then use C+1:0:0
 ##########################################################################
 
-ODP_LIBSO_VERSION=115:0:2
+ODP_LIBSO_VERSION=116:0:3
```
lumag: As we are not providing ABI-level compatibility between releases (and we are not ATM), this always should be XXX:0:0.
Especially since we removed `odp_packet_unshare_len()`. So proper version should be 116:0:0

lumag:
@Bill-Fischofer-Linaro  this should not be approved.

configure.ac
muvarov: @lumag what is ATM stands for?

lumag: @muvarov "at this moment"

muvarov: @lumag  I like that idea to keep LIBSO version always the same as API version. But when I said that first time than Anders said that it's not correct and we have to follow the same rules as libcurl. So for some last releases we followed that magic scheme. Which might be a not good fit for mainline version.

lumag:

muvarov:
merged to next.

PR #286, opened 2017-11-08
[PATCH v1] example: fix pointers to odp_l2fwd.c source

lumag:
During mass-move of tests I forgot to update odp_l2fwd example to point
to new locattion. Fix that now.


muvarov:
ok, we run tests from other dir. That needs to be applied.

muvarov:
Merged.

PR #287, opened 2017-11-08
[PATCH v1] DPDK pktio fixes

matiaselo:


platform/linux-generic/pktio/dpdk.c
```diff
@@ -1372,10 +1382,10 @@ static int dpdk_recv(pktio_entry_t *pktio_entry, int index,
 	if (!pkt_dpdk->lockless_rx)
 		odp_ticketlock_lock(&pkt_dpdk->rx_lock[index]);
 	/**
-	 * ixgbe_pmd has a minimum supported RX burst size ('min_rx_burst'). If
-	 * 'num' < 'min_rx_burst', 'min_rx_burst' is used as rte_eth_rx_burst()
-	 * argument and the possibly received extra packets are cached for the
-	 * next dpdk_recv_queue() call to use.
+	 * ixgbe and i40e drivers have a minimum supported RX burst size
```
nagarahalli: 'min_rx_burst' should be added to the capability as it is a restriction from the device. The application should adjust the 'num' according to the capability it reads from the pkt I/O.

platform/linux-generic/pktio/dpdk.c
```diff
@@ -1249,10 +1253,16 @@ static int dpdk_open(odp_pktio_t id ODP_UNUSED,
...
+	/* Drivers requiring minimum burst size. Supports also *_vf versions
+	 * of the drivers. */
+	if (!strncmp(dev_info.driver_name, IXGBE_DRV_NAME,
+		     strlen(IXGBE_DRV_NAME)) ||
+	    !strncmp(dev_info.driver_name, I40E_DRV_NAME,
+		     strlen(I40E_DRV_NAME)))
+		pkt_dpdk->min_rx_burst = DPDK_MIN_RX_BURST;
 	else
 		pkt_dpdk->min_rx_burst = 0;
+
```
Bill-Fischofer-Linaro: Wouldn't it be simpler (and more extensible) to have `min_rx_burst` as a field in the `dev_info` struct?

platform/linux-generic/pktio/dpdk.c
Bill-Fischofer-Linaro: Presumably these drivers have a way of flushing short bursts? If the `min_rx_burst` is 4 and the other side sends an odd number of packets, presumably the receiver application isn't left hanging forever waiting for the rest of a burst that will never arrive?

platform/linux-generic/pktio/dpdk.c
```diff
@@ -1135,6 +1135,13 @@ static void dpdk_init_capability(pktio_entry_t *pktio_entry,
 	rte_eth_dev_info_get(pkt_dpdk->port_id, dev_info);
 	capa->max_input_queues = RTE_MIN(dev_info->max_rx_queues,
 					 PKTIO_MAX_QUEUES);
+
+	/* ixgbe devices support only 16 rx queues in RSS mode */
+	if (!strncmp(dev_info->driver_name, IXGBE_DRV_NAME,
+		     strlen(IXGBE_DRV_NAME)))
+		capa->max_input_queues = RTE_MIN((unsigned)16,
+						 capa->max_input_queues);
+
```
lumag: It looks like this should be fixed inside DPDK rather than here.

Bill-Fischofer-Linaro: Again, shouldn't this sort of configuration be a field in the `dev_info` struct rather than having to be special-cased for each driver? Also, must 16 be a "magic number" here?

platform/linux-generic/pktio/dpdk.c
matiaselo: @nagarahalli This dpdk driver "feature" can be completely hidden from odp applications (and it already is), so I don't support adding a new device capability parameter. This would make writing simple applications more complex. E.g. an application wouldn't be able to receive one packet at a time anymore.

matiaselo: @Bill-Fischofer-Linaro Yes, the application has to request at least 'min_rx_burst' packets but it may receive less.

platform/linux-generic/pktio/dpdk.c
matiaselo: I'm not sure I got your point but anyway dev_info is type 'struct rte_eth_dev_info' and comes from dpdk, so nothing can be added there.

As @lumag noted this fix is required because of a dpdk defect. With ixgbe drivers rte_eth_dev_info_get() returns a max_rx_queues value which is not valid (at least when using RSS). This is the only driver a have observer this issue with.

The number 16 is explained in the comment line (1139) just above.

matiaselo: @lumag That would be the best solution but I don't see it happening anytime soon. The number of rx queues returned by rte_eth_dev_info_get() is in principle correct. Here the limiting factor is RSS capability and as far as I can see there is no way to query this information.

platform/linux-generic/pktio/dpdk.c
nagarahalli: @matiaselo Agree on the simple applications. But for real applications or for performance benchmarking, this adds additional cycles. If this is made as part of capability, application can decide the burst size upfront.

matiaselo: @nagarahalli I don't see a new capability parameter being very useful as these two dpdk drivers are likely the only devices which would use it. At least we haven't come across any other network devices which have this kind of limitation.

I ran a quick test and the two additional if statements caused by rx caching have a minuscule  performance impact.

nagarahalli: @matiaselo I see these kind of validations in data plane functions at multiple places. If we combine all of them, it is significant improvement (I see 1% improvement in Linux-dpdk/l2fwd application by removing all the validation checks in pkt I/O module alone). But on an individual basis, you will not see much of a difference.

It might be few device drivers, but from ODP perspective, it is a capability of the device that should be exposed through capability API. We could do it outside of this PR as well.

matiaselo: @nagarahalli Where do you see these kinds of checks? As far as I can see this is only relevant to dpdk pktio and if there are similar checks elsewhere they should probably be removed.

I agree that this issue should be discussed outside of this PR as these patches are only simple bug fixes to the current implementation.

Bill-Fischofer-Linaro:
@nagarahalli To return, briefly, to the discussion from today's ARCH call. The issue here is how ODP drivers attach devices to `odp_pktio_t` objects. The ODP application doesn't care about the specific device behind an `odp_pktio_t`, or what's needed to operate that device in an optimal manner--that's the responsibility of the ODP implementation and/or device driver. 

For example, earlier @matiaselo discovered that DPDK PMDs don't support buffers that cross hugepage boundaries, so a PR was posted that accounted for that in how DPDK PktIO support worked. This is the sort of artificial restriction that an ODP application doesn't care about because it doesn't control or need to be aware of the physical layout of pools. For the same reason if a particular device expects packet I/O to be done in bursts rather than individually, then it is the responsibility of the ODP implementation/driver to do so and provide whatever I/O caching is necessary in a manner transparent to the application. 

The ODP PktIO APIs support burst-mode operation either for direct or scheduled I/O, but that is for application design convenience, not to accommodate the oddities of particular devices, which may vary from device to device, release to release, etc. So if an application wishes to burst I/O for its own reasons (_e.g.,_ VPP) ODP supports that, but it does not require that. Conversely, if a device requires I/O to be done as bursts for best performance, again that is something that the ODP implementation/driver handles without requiring application awareness or involvement. That's because application I/O is done at an abstract level rather than a device-specific level, just as other ODP APIs operate using ODP-defined high-level semantics rather than platform-specific semantics.

muvarov:
Merged.

PR #243, opened 2017-10-20
[PATCH API-NEXT v16] IPsec implementation updates

lumag:
- security fix for GCM mode: proper IV generation
- implementation of 32-bit replay window support
- implementation of PIPELINE_CLS support

PR #288, opened 2017-11-08
[PATCH API-NEXT v8] AES-GMAC implementation

lumag:
This is an example of AES-GMAC(-ESP) implementation. Plaintext is passed as normal packet, IV/salt are passed according to the rest of API. This incorporates one commit from #243.


lumag:
@Bill-Fischofer-Linaro no, these are checkpatch errors. Especially the second one.

test/validation/api/crypto/test_vectors.h
```diff
@@ -361,6 +361,39 @@ static crypto_test_reference_t aes_gcm_reference[] = {
...
+				0x45, 0x00, 0x00, 0x30, 0xda, 0x3a, 0x00, 0x00,
+				0x80, 0x01, 0xdf, 0x3b, 0xc0, 0xa8, 0x00, 0x05,
+				0xc0, 0xa8, 0x00, 0x01, 0x08, 0x00, 0xc6, 0xcd,
+				0x02, 0x00, 0x07, 0x00, 0x61, 0x62, 0x63, 0x64,
+				0x65, 0x66, 0x67, 0x68, 0x69, 0x6a, 0x6b, 0x6c,
+				0x6d, 0x6e, 0x6f, 0x70, 0x71, 0x72, 0x73, 0x74,
+				0x01, 0x02, 0x02, 0x01 },
+		.digest_length = AES_GCM_DIGEST_LEN,
+		.digest = { 0xf2, 0xa9, 0xa8, 0x36, 0xe1, 0x55, 0x10,
+			    0x6a, 0xa8, 0xdc, 0xd6, 0x18, 0xe4, 0x09, 0x9a, 0xaa }
```
Bill-Fischofer-Linaro: Checkpatch flags this line as > 80 chars. Easy to reflow, e.g., move the `0x6a` to the earlier line.

platform/linux-generic/include/odp_crypto_internal.h
```diff
@@ -51,7 +51,10 @@ struct odp_crypto_generic_session {
 		uint8_t  key[EVP_MAX_KEY_LENGTH];
 		uint32_t key_length;
 		uint32_t bytes;
-		const EVP_MD *evp_md;
+		union {
+			const EVP_MD *evp_md;
+			const EVP_CIPHER *evp_cipher;
+		};
 		crypto_func_t func;
```
Bill-Fischofer-Linaro: You're right. This one is a false positive on checkpatch's part.

include/odp/api/spec/crypto.h
```diff
@@ -131,6 +131,12 @@ typedef enum {
 	 */
 	ODP_AUTH_ALG_AES_GCM,
 
+	/** AES in Galois/Counter MAC Mode
+	 *
+	 *  @note Must be paired with cipher ODP_CIPHER_ALG_NULL
+	 */
+	ODP_AUTH_ALG_AES_GMAC,
```
psavol: This should specify that GMAC authentication is done for packet payload (instead of AAD). NIST/RFC documents use term AAD for the data. So, add here (or somewhere in the file) a reference to e.g. 
https://csrc.nist.gov/publications/detail/sp/800-38d/final, and document that AAD there means packet payload in ODP with GMAC.

For other algos AAD is the same as NIST/RFC specs.

lumag:
@psavol updated AES-GMAC description

include/odp/api/spec/crypto.h
psavol: I think this needs to be still a bit more specific. For example:

"NIST and RFC specifications about GCM/GMAC refer to all data to be authenticated as AAD. In constrast to that, ODP API specifies the bulk of authenticated data to be located in packet payload for all authentication algorithms - including GMAC."

The first '@note' doxygen tag can be removed.

Also do we support still AAD in GMAC case? I think we should if there's no reason not to support. So, bulk of data in packet, additional data possible through AAD (like any other algo).

lumag: There is no AAD for other MAC algorithms, so AAD is not supported for GMAC.

lumag: @psavol updated description

muvarov:
Merged.

PR #289, opened 2017-11-08
[PATCH v1] linux-gen: crypto: drop DES-CBC support

lumag:
DES-CBC (not 3DES-CBC) support is invalid (DES should use 8 bytes key,
not 24 bytes), it is not covered by testsuite and DES is
cracable/deprecated/etc since long ago. Stop providing single-key DES
support. In case one really needs it, it can be emulated through
supplying same key triple times to 3DES-CBC.


muvarov:
Merged.

PR #292, opened 2017-11-10
[PATCH v1] travis: add build only test stage

matiaselo:
Add basic build only test stage to the beginning of test run. This way
potential build errors are caught early on without having to wait for the
longer test jobs to finish.

Checkpatch is not moved to the build only stage as potential false
positives would prevent the next test stage being run.

### DIFF
--- a/.travis.yml
+++ b/.travis.yml
@@ -11,6 +11,9 @@ language: c
 sudo: required
 dist: trusty
 group: deprecated-2017Q2
+stages:
+  - "build only"
+  - test
 env:
   global:
     # COVERITY_SCAN_TOKEN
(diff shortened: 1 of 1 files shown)


lumag:
Nice idea. Could you please move all cross-compiler tests to build stage?

matiaselo:
@lumag This would be nice but it seems currently Travis supports matrix expansion only on the top level ([link](https://github.com/travis-ci/travis-ci/issues/8293)) E.g. in our case one cannot add a second build matrix to the build only stage. Because of this, moving all cross compiler tests to build only stage would mean a lot of copy-and-paste scripting. In essence, the top level 'script' section would have to be repeated for each platform and compiler combination.

I've selected the current build targets in build only stage so that they should catch most of the build failures. Since Travis offers only five parallel jobs, adding more platforms to the build only stage would increase test duration.

muvarov:
@matiaselo  checkpatch also can go on top also arm and powerpc (i.e. all 2 minutes tests).

muvarov:
and rename it  from build only to "Quick tests" because doxygen and checkpatch are not really "Build only"

matiaselo:
@muvarov False positives are a problem with checkpatch. If checkpatch is moved to the build only / quick tests stage the potential false positives would prevent the next test stage being run.

lumag:
@matiaselo ok then.

muvarov:
Merged.

PR #295, opened 2017-11-15
[PATCH 2.0 v2] driver.h: add parameter when registering enumerators

joseppc:
If a class registers more than once the same enumerator, as it could
happen for different PCI platforms, we need to provide information for
the enumerator probe function to know where to act.

This information shall be provided in the odpdrv_enumr_param_t structure
as a void pointer, which will be passed to the probe() function of the
enumerator.


bogdanPricope:
You may add the initialization of the new member 'enumr_data' for dpdk and generic enumerators (ddf_ifs_enumr_dpdk.c, ddf_ifs_enumr_generic.c). Is not used (for the moment), but it helps to better understand this patch.

The change 'fit in' since we have something similar for ddf_devices but DDF API as a hall is odd since callbacks don't have access to object's private data.

joseppc:
Yes, I will provide another patch with that showing its usage if this is accepted.
I agree about DDF being a bit odd...

heyi-linaro:
merged and closed.

PR #296, opened 2017-11-16
[PATCH API-NEXT v2] doc: userguide: ipsec state machine changes

Bill-Fischofer-Linaro:
Split the single IPsec FSM into separate FSM diagrams showing
SA state transitions and IPsec packet operations on SAs.
Update User Guide to reflect these changes as well.


JannePeltonen:
In the SA state machine the transition from Disable_Pending to SA_Disabled state is not triggered by odp_ipsec_status() call. Therefore I think it would be more clear to show something like "disable complete status event" instead of odp_ipsec_status() as the label of the transition.

JannePeltonen:
In the SA state machine the transition from Disable_Pending to SA_Disabled state is not triggered by odp_ipsec_status() call. Therefore I think it would be more clear to show something like "disable complete status event" instead of odp_ipsec_status() as the label of the transition.

I think the diagram that shows IPsec related packet operations has the following issues:
- It is not clear to which object the states belong to. The name of the SA_Ready state may give the incorrect impression that the states are SA states. Since the states are operation states per packet, they are really packet states, imho, can could be called such.
- In my view, odp_ipsec_result() and odp_ipsec_status() do not cause any state transition.
- I am not sure it is good to label the transition from Processing to Op_Complete with odp_queue_deq() amd odp_schedule() because those calls do not cause the processing to complete and generally do not wait for it.

I tried to redraw the diagram in several ways but I am not sure I was able to come up with anything better. Here is my attempt at showing "packet states" related to IPsec and packet I/O (I cheated a little and have Packet free state twice):

digraph ipsec_op_state_machine {
        rankdir=LR;
        node [shape=circle];

        free [label = "Packet free"]
        free2 [label = "Packet free"]
        ready [label = "Packet ready"]
        processing [label = "Processing"]
        input [label = "In input queue"]
        output [label = "In pktout"]
        done [label = "In completion queue"]

        ready -> ready [label = "odp_ipsec_in()"];
        ready -> ready [label = "odp_ipsec_out()"]
        ready -> processing [label = "odp_ipsec_in_enq()"];
        ready -> processing [label = "odp_ipsec_out_enq()"];
        ready -> output [label = "odp_ipsec_out_inline()"];
        ready -> output [label = "send packet"];

        processing -> done [label = "processing completes"]
        done -> ready [label = "odp_queue_deq() / odp_schedule()"];
        output -> free2 [label = "sending completes"]
        free -> input [label = "packet arrives"]
        input -> ready [label = "receive packet"]
}

Bill-Fischofer-Linaro:
@JannePeltonen The purpose here is to show which APIs are valid at various points in processing, so although transitions may not be *caused* by certain API calls, that's how they are discovered. So the only way a program discovers that it's now in the `SA_Disabled` state is to call `odp_ipsec_status()` and observe a disable complete event.

The reason for using `SA_Ready` in the second FSM to to relate it to the first. You can only launch IPsec operations when the SA they are to be performed against is in this state. So the second FSM is simply an expansion of the "ODP IPsec packet operations" arrow in the first FSM. This results from splitting the original combined FSM into separate FSMs, one for SA state transitions, and one for packet processing.

I'm not sure expanding the second diagram from three to seven states clarifies things. 

There was a typo in the second diagram that is corrected in v2.

Bill-Fischofer-Linaro:
Changes in v2:

* Rebased to latest api-next
* Typo correction in second FSM.

psavol:
A message or sequence diagram would suite better this purpose than a state diagram. It's very easy to understand each use case when drawn into separate sequence diagrams. Application on left, implementation on right. Messages in between are the API calls. You could also draw a bit more complex  examples: e.g. application starts multiple IPSEC requests, some are completed normally while others completed after SA disable sequence.

https://en.wikipedia.org/wiki/Sequence_diagram

Bill-Fischofer-Linaro:
@psavol I've posted PR #320 as an alternative approach following your suggestion. I'm not convinced it's clearer than the FSMs, but something for us to discuss.

Bill-Fischofer-Linaro:
Closing this PR in preference to PR #320.

PR #297, opened 2017-11-16
[PATCH 2.0 v3] Replace pktio ops_data array implementation with memory pool implementation

bogdanPricope:
Replace pktio ops_data array implementation with memory pool implementation.

### DIFF
--- a/platform/linux-generic/Makefile.am
+++ b/platform/linux-generic/Makefile.am
@@ -175,6 +175,8 @@ noinst_HEADERS = \
 		  include/odp_name_table_internal.h \
 		  include/odp_packet_internal.h \
 		  include/odp_packet_io_internal.h \
+		  include/odp_packet_io_pool.h \
+		  include/odp_packet_io_pool_access.h \
 		  include/odp_packet_io_ring_internal.h \
 		  pktio/ethtool.h \
 		  pktio/common.h \
@@ -244,6 +246,7 @@ __LIB__libodp_linux_la_SOURCES = \
 			   odp_packet.c \
 			   odp_packet_flags.c \
 			   odp_packet_io.c \
+			   odp_packet_io_pool.c \
 			   pktio/ethtool.c \
 			   pktio/subsystem.c \
 			   pktio/ipc.c \
--- a/platform/linux-generic/include/odp_packet_io_internal.h
+++ b/platform/linux-generic/include/odp_packet_io_internal.h
@@ -43,8 +43,8 @@ typedef union pktio_entry_u pktio_entry_t;
 
 struct pktio_entry {
 	const pktio_ops_module_t *ops;	/**< Implementation specific methods */
-	uint8_t ops_data[ODP_PKTIO_ODPS_DATA_MAX_SIZE]; /**< IO operation
-							specific data */
+	void *ops_data;			/**< IO operation specific data */
+
 	/* These two locks together lock the whole pktio device */
 	odp_ticketlock_t rxl;		/**< RX ticketlock */
 	odp_ticketlock_t txl;		/**< TX ticketlock */
--- a/platform/linux-generic/include/odp_pktio_ops_subsystem.h
+++ b/platform/linux-generic/include/odp_pktio_ops_subsystem.h
@@ -16,6 +16,7 @@ extern "C" {
 
 #include <odp_module.h>
 #include <odp/api/packet_io.h>
+#include <odp_packet_io_pool_access.h>
 
 /* ODP packet IO operations subsystem declaration */
 ODP_SUBSYSTEM_DECLARE(pktio_ops);
@@ -84,13 +85,6 @@ typedef ODP_MODULE_CLASS(pktio_ops) {
 	odp_api_proto(pktio_ops, output_queues_config) output_queues_config;
 	odp_api_proto(pktio_ops, print) print;
 } pktio_ops_module_t;
-
-/* Maximum size of pktio specific ops data.*/
-#define ODP_PKTIO_ODPS_DATA_MAX_SIZE 80000
-
-/* Extract pktio ops data from pktio entry structure */
-#define odp_ops_data(_p, _mod) \
-	((pktio_ops_ ## _mod ## _data_t *)(uintptr_t)_p->s.ops_data)
 
 #ifdef __cplusplus
 }
--- a/platform/linux-generic/odp_packet_io_pool.c
+++ b/platform/linux-generic/odp_packet_io_pool.c
@@ -1,0 +1,55 @@
+/* Copyright (c) 2017, Linaro Limited
+ * All rights reserved.
+ *
+ * SPDX-License-Identifier:     BSD-3-Clause
+ */
+
+#include "config.h"
+
+#include <odp_posix_extensions.h>
+#include <odp/api/packet_io.h>
+#include <odp/api/std_clib.h>
+#include <odp/drv/shm.h>
+#include "odp_packet_io_pool.h"
+#include "odp_packet_io_pool_access.h"
+#include "odp_debug_internal.h"
+
+#define ODP_PKTIO_OPS_DATA_POOL_SIZE 160000
+
+int _odp_packet_io_pool_create(void)
+{
+	odpdrv_shm_pool_t pktio_ops_pool;
+	odpdrv_shm_pool_param_t pktio_ops_param;
+
+	odp_memset(&pktio_ops_param, 0, sizeof(pktio_ops_param));
+	pktio_ops_param.pool_size = ODP_PKTIO_OPS_DATA_POOL_SIZE;
+	pktio_ops_param.min_alloc = 1;
+	pktio_ops_param.max_alloc = ODP_PKTIO_OPS_DATA_POOL_SIZE;
+
+	pktio_ops_pool = odpdrv_shm_pool_create(ODP_PKTIO_OPS_DATA_POOL_NAME,
+						&pktio_ops_param);
+	if (pktio_ops_pool == ODPDRV_SHM_POOL_INVALID) {
+		ODP_ERR("error to pool_create pktio_ops pool\n");
+		return -1;
+	}
+
+	return 0;
+}
+
+int _odp_packet_io_pool_destroy(void)
+{
+	odpdrv_shm_pool_t pktio_ops_pool;
+
+	pktio_ops_pool = odpdrv_shm_pool_lookup(ODP_PKTIO_OPS_DATA_POOL_NAME);
+	if (pktio_ops_pool == ODPDRV_SHM_POOL_INVALID) {
+		ODP_ERR("error pool_lookup pktio_ops pool\n");
+		return -1;
+	}
+
+	if (odpdrv_shm_pool_destroy(pktio_ops_pool)) {
+		ODP_ERR("error pool_destroy pktio_ops pool\n");
+		return -1;
+	}
+
+	return 0;
+}
--- a/platform/linux-generic/pktio/tap.c
+++ b/platform/linux-generic/pktio/tap.c
@@ -92,23 +92,31 @@ static int tap_pktio_open(odp_pktio_t id ODP_UNUSED,
 	int fd, skfd, flags;
 	uint32_t mtu;
 	struct ifreq ifr;
-	pktio_ops_tap_data_t *tap = odp_ops_data(pktio_entry, tap);
+	pktio_ops_tap_data_t *tap = NULL;
 
 	if (strncmp(devname, "tap:", 4) != 0)
 		return -1;
+
+	if (pool == ODP_POOL_INVALID)
+		return -1;
+
+	pktio_entry->s.ops_data = ODP_OPS_DATA_ALLOC(sizeof(*tap));
+	if (odp_unlikely(pktio_entry->s.ops_data == NULL)) {
+		ODP_ERR("Failed to allocate pktio_ops_tap_data_t struct");
+		return -1;
+	}
+	tap = pktio_entry->s.ops_data;
 
 	/* Init pktio entry */
 	memset(tap, 0, sizeof(*tap));
 	tap->fd = -1;
 	tap->skfd = -1;
 
-	if (pool == ODP_POOL_INVALID)
-		return -1;
-
 	fd = open("/dev/net/tun", O_RDWR);
 	if (fd < 0) {
 		__odp_errno = errno;
 		ODP_ERR("failed to open /dev/net/tun: %s\n", strerror(errno));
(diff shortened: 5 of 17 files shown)


platform/linux-generic/include/odp_pktio_ops_subsystem.h
```diff
@@ -85,12 +86,39 @@ typedef ODP_MODULE_CLASS(pktio_ops) {
...
 #define odp_ops_data(_p, _mod) \
-	((pktio_ops_ ## _mod ## _data_t *)(uintptr_t)_p->s.ops_data)
+	((pktio_ops_ ## _mod ## _data_t *)_p->s.ops_data)
+
+#define odp_ops_data_alloc(_p, _mod)					\
+({ \
+	odpdrv_shm_pool_t _pool = ODPDRV_SHM_POOL_INVALID;		\
+	uint64_t _size = sizeof(pktio_ops_ ## _mod ## _data_t);		\
+									\
+	_p->s.ops_data = NULL;			\
```
joseppc: (_p)?  There are a couple more, also in odp_ops_data_free.

platform/linux-generic/pktio/tap.c
```diff
@@ -92,23 +92,30 @@ static int tap_pktio_open(odp_pktio_t id ODP_UNUSED,
 	int fd, skfd, flags;
 	uint32_t mtu;
 	struct ifreq ifr;
-	pktio_ops_tap_data_t *tap = odp_ops_data(pktio_entry, tap);
+	pktio_ops_tap_data_t *tap = NULL;
 
 	if (strncmp(devname, "tap:", 4) != 0)
 		return -1;
 
+	if (pool == ODP_POOL_INVALID)
```
joseppc: It probably belongs to its own patch, but now that you are at it, it could even be moved even further up, as it is probably faster than checking for "tap:" in the device string.

platform/linux-generic/include/odp_pktio_ops_subsystem.h
bogdanPricope: _p comes form (pktio_entry_t *)

platform/linux-generic/include/odp_pktio_ops_subsystem.h
```diff
@@ -85,12 +86,39 @@ typedef ODP_MODULE_CLASS(pktio_ops) {
...
+	_pool = odpdrv_shm_pool_lookup(ODP_PKTIO_OPS_DATA_POOL_NAME);	\
+	if (_pool != ODPDRV_SHM_POOL_INVALID)				\
+		_p->s.ops_data = odpdrv_shm_pool_alloc(_pool, _size);	\
+									\
+	((pktio_ops_ ## _mod ## _data_t *)_p->s.ops_data);		\
```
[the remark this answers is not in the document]
bogdanPricope: It will require a cast when is called. 
A pktio_type may implement another way to allocate memory starting form the name of the pool / is not mandatory to use those macros but are helpful for existing pktio types.

platform/linux-generic/pktio/tap.c
bogdanPricope: Alloc/free vs. array has this disadvantage: you need to allocate the memory at some point and free it if the operation fails for any reason. It is better to delay the allocation until after some common checks.

stncmp: open calls are not on fast path ... no reason to optimize the performance ... but repeated memory alloc/free may affect some pool implementations

joseppc: Ok, I may be splitting hairs now, but I thought we were just checking whether the pool parameter passed to pktio_open was valid, and bail out if not. We are not actually using this pool to allocate this pktio's private data, right?

bogdanPricope: No, this pool is used to allocate packets (for recv side).

platform/linux-generic/include/odp_pktio_ops_subsystem.h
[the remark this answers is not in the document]
bogdanPricope: @joseppc   Btw, 'odp_ops_data_alloc(_p, _mod)' vs. odp_ops_data_alloc(_p, _size) ?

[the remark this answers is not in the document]
joseppc: ah, yes, true, I didn't think about this detail...

platform/linux-generic/pktio/subsystem.c
```diff
@@ -15,9 +15,55 @@
 ODP_SUBSYSTEM_DEFINE(pktio_ops, "packet IO operations", SUBSYSTEM_VERSION);
 
 /* Instantiate init and term functions */
-ODP_SUBSYSTEM_FOREACH_TEMPLATE(pktio_ops, init_global, ODP_ERR)
+SUBSYSTEM_FOREACH_TEMPLATE(pktio_ops, init_global, ODP_ERR)
+SUBSYSTEM_FOREACH_TEMPLATE(pktio_ops, term_global, ODP_ERR)
 ODP_SUBSYSTEM_FOREACH_TEMPLATE(pktio_ops, init_local, ODP_ERR)
-ODP_SUBSYSTEM_FOREACH_TEMPLATE(pktio_ops, term_global, ODP_ABORT)
+
+int odp_pktio_ops_init_global(bool continue_on_errors)
```
nagarahalli: This functionality can be done in odp_pktio_global_init function. This will avoid the changes to modular framework as well.

When we have done additional enhancements to shared memory, this code will be deleted. So, can be part of odp_pktio_global_init without affecting the modular framework.

platform/linux-generic/pktio/subsystem.c
```diff
@@ -15,9 +15,55 @@
...
+	if (pktio_ops_pool == ODPDRV_SHM_POOL_INVALID) {
+		ODP_ERR("error to pool_create pktio_ops pool\n");
+		if (!continue_on_errors)
+			return -1;
+	}
+
+	return pktio_ops_init_global(continue_on_errors);
+}
+
+int odp_pktio_ops_term_global(bool continue_on_errors)
```
nagarahalli: Same here, this can be part of odp_pktio_term_global

platform/linux-generic/include/odp_pktio_ops_subsystem.h
```diff
@@ -88,10 +89,40 @@ typedef ODP_MODULE_CLASS(pktio_ops) {
...
 
+/* Pktio ops data pool */
+#define ODP_PKTIO_OPS_DATA_POOL_SIZE 160000
+#define ODP_PKTIO_OPS_DATA_POOL_NAME "ODP_PKTIO_OPS_DATA"
+
 /* Extract pktio ops data from pktio entry structure */
 #define odp_ops_data(_p, _mod) \
 	((pktio_ops_ ## _mod ## _data_t *)(uintptr_t)_p->s.ops_data)
 
+#define odp_ops_data_alloc(_p, _mod)					\
```
nagarahalli: Should be odp_ops_data_alloc(_p, size).

As per the pkt I/O changes document, _p (pktio_entry_t) is not required to be exposed to drivers. Do you plan to do it as part of this PR?

platform/linux-generic/include/odp_pktio_ops_subsystem.h
```diff
@@ -88,10 +89,40 @@ typedef ODP_MODULE_CLASS(pktio_ops) {
...
+({ \
+	odpdrv_shm_pool_t _pool = ODPDRV_SHM_POOL_INVALID;		\
+	uint64_t _size = sizeof(pktio_ops_ ## _mod ## _data_t);		\
```
nagarahalli: Temporarily, roundup the size to cache line size. This way all the memory allocations will be cache line aligned.

frameworks/modular/odp_module.h
```diff
@@ -237,6 +228,32 @@ done:								\
...
+/* SUBSYSTEM_FOREACH_TEMPLATE() generates static walk through
+ * functions to invoke subsystem's modules initialization or
+ * termination methods, examples:
+ *
+ * SUBSYSTEM_FOREACH_TEMPLATE(subsystem, init_global, DBG)
+ * will generate a function walk through all the modules of the
+ * subsystem and invoke init_global method for each.
+ */
+
+#define SUBSYSTEM_FOREACH_TEMPLATE(subs, method, print)	\
```
heyi-linaro: seems no need to add an extra macro here?

ODP_SUBSYSTEM_FOREACH_TEMPLATE(...) is extern
we can just use to generate a static function use the same macro:
static ODP_SUBSYSTEM_FOREACH_TEMPLATE(...)

platform/linux-generic/include/odp_pktio_ops_subsystem.h
```diff
@@ -88,10 +89,40 @@ typedef ODP_MODULE_CLASS(pktio_ops) {
 /* Maximum size of pktio specific ops data.*/
 #define ODP_PKTIO_ODPS_DATA_MAX_SIZE 80000
 
+/* Pktio ops data pool */
+#define ODP_PKTIO_OPS_DATA_POOL_SIZE 160000
+#define ODP_PKTIO_OPS_DATA_POOL_NAME "ODP_PKTIO_OPS_DATA"
+
 /* Extract pktio ops data from pktio entry structure */
 #define odp_ops_data(_p, _mod) \
```
heyi-linaro: Look good no more comments from me to this commit after Honnappa and Josep's, this is a step forward for the pktio ops data

This commit also reveals how complex in ODP to allocate an arbitrary sized piece of memory, need to prepare a pool (and guess the largest usage), lookup this pool in every allocation/free by name, and do the allocation/free after then.

platform/linux-generic/include/odp_packet_io_internal.h
```diff
@@ -43,8 +43,8 @@ typedef union pktio_entry_u pktio_entry_t;
 
 struct pktio_entry {
 	const pktio_ops_module_t *ops;	/**< Implementation specific methods */
-	uint8_t ops_data[ODP_PKTIO_ODPS_DATA_MAX_SIZE]; /**< IO operation
-							specific data */
+	void *ops_data;			/**< IO operation specific data */
```
heyi-linaro: Like this!

platform/linux-generic/include/odp_pktio_ops_subsystem.h
heyi-linaro: In future, since each pktio_ops module will not expose their private data type, this macro can be changed to
>`#define odp_ops_data(_entry, _pdata) \
        pdata = (typeof(_pdata))(_entry->s.ops_data)`

So the odp_pktio_ops_subsystem.h header won't need to know all pktio_ops' private data structure declaration. Can be considered next time.

platform/linux-generic/include/odp_pktio_ops_subsystem.h
```diff
@@ -87,12 +87,8 @@ typedef ODP_MODULE_CLASS(pktio_ops) {
...
 
-/* Maximum size of pktio specific ops data.*/
-#define ODP_PKTIO_ODPS_DATA_MAX_SIZE 80000
-
 /* Extract pktio ops data from pktio entry structure */
-#define odp_ops_data(_p, _mod) \
-	((pktio_ops_ ## _mod ## _data_t *)(uintptr_t)_p->s.ops_data)
+#define odp_ops_data(_p, _mod) _p->s.ops_data
 
 #define odp_ops_data_alloc(_p, _size)					\
```
nagarahalli: '_p' is not required as the macro is returning 'ops_data'. Makes the macro simple as well.

Similarly for odp_ops_data_free can just take 'ops_data' as input.

This will be inline with future plans to not expose 'pktio_entry_t' to the drivers.

platform/linux-generic/Makefile.am
```diff
@@ -175,6 +175,7 @@ noinst_HEADERS = \
 		  include/odp_name_table_internal.h \
 		  include/odp_packet_internal.h \
 		  include/odp_packet_io_internal.h \
+		  include/odp_packet_io_pool.h \
```
nagarahalli: Should we rename it to odp_packet_io_shm.h and odp_packet_io_shm.c?

include/odp/api/spec/packet_io.h
```diff
@@ -1188,6 +1188,11 @@ uint64_t odp_pktin_ts_res(odp_pktio_t pktio);
  */
 odp_time_t odp_pktin_ts_from_ns(odp_pktio_t pktio, uint64_t ns);
 
+/**
+ * Packet IO operations data pool name
+ */
+#define ODP_PKTIO_OPS_DATA_POOL_NAME "ODP_PKTIO_OPS_DATA"
```
nagarahalli: This should be part of odp_pktio_ops_subsystem.h file.

platform/linux-generic/pktio/tap.c
Bill-Fischofer-Linaro: This is a parameter passed through from `odp_pktio_open()`. The spec simply says that this must be of type `ODP_POOL_PACKET`, so results are undefined if the caller doesn't abide by the spec. A courtesy validation check doesn't seem unreasonable, but any validation checking should be done in the main API, not in each individual driver. As such drivers should assume that `pool` is valid at entry since these entry points cannot be invoked directly by applications.

platform/linux-generic/Makefile.am
nagarahalli: This file is for allocating the shmem for storing the driver's global data.

The name 'pool' seems to be creating confusion with the packet/buffer pool. May be we should change the name of '_odp_ishm_pool_create/odpdrv_shm_pool_create'.

platform/linux-generic/pktio/tap.c
bogdanPricope: I agree, this check should  be part of odp_pktio_open() or setup_pktio_entry() but this should be done in another PR. Here, I only move this check upper in the function (is not a new code).

platform/linux-generic/odp_packet_io_pool.c
```diff
@@ -0,0 +1,54 @@
...
+#include "config.h"
+
+#include <odp_posix_extensions.h>
+#include <odp/api/packet_io.h>
+#include <odp/api/std_clib.h>
+#include <odp/drv/shm.h>
+#include "odp_packet_io_pool.h"
+#include "odp_debug_internal.h"
+
+#define ODP_PKTIO_OPS_DATA_POOL_SIZE 160000
```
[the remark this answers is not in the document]
bogdanPricope: Actual default size of this pool and the mechanism to automatically extend it will be part of another PR. This size covers requirements for two netmap interfaces  (to pass netmap tests).

heyi-linaro:
merged and closed

PR #298, opened 2017-11-16
[PATCH API-NEXT v1] api: pktio: max frame length

psavol:
MTU is not a well defined term for link layer maximum receive and transmit frame sizes. Added functions to request maximum packet input and output frame lengths. Deprecated odp_pktio_mtu() function.

### DIFF
--- a/include/odp/api/spec/packet_io.h
+++ b/include/odp/api/spec/packet_io.h
@@ -970,6 +970,36 @@ int odp_pktio_promisc_mode_set(odp_pktio_t pktio, odp_bool_t enable);
 int odp_pktio_promisc_mode(odp_pktio_t pktio);
 
 /**
+ * Maximum frame length at packet input
+ *
+ * Maximum frame length in bytes that the packet IO interface can receive.
+ * For Ethernet, the frame length bytes start with MAC addresses and continue
+ * to the end of the payload. So, Ethernet checksum, interpacket gap
+ * and preamble bytes are excluded from the length.
+ *
+ * @param pktio  Packet IO handle.
+ *
+ * @return Maximum frame length at packet input
+ * @retval 0 on failure
+ */
+uint32_t odp_pktin_maxlen(odp_pktio_t pktio);
+
+/**
+ * Maximum frame length at packet output
+ *
+ * Maximum frame length in bytes that the packet IO interface can transmit.
+ * For Ethernet, the frame length bytes start with MAC addresses and continue
+ * to the end of the payload. So, Ethernet checksum, interpacket gap
+ * and preamble bytes are excluded from the length.
+ *
+ * @param pktio  Packet IO handle.
+ *
+ * @return Maximum frame length at packet output
+ * @retval 0 on failure
+ */
+uint32_t odp_pktout_maxlen(odp_pktio_t pktio);
+
+/**
  * Get the default MAC address of a packet IO interface.
  *
  * @param	pktio     Packet IO handle
--- a/include/odp/api/spec/packet_io.h
+++ b/include/odp/api/spec/packet_io.h
@@ -954,8 +954,8 @@ uint32_t ODP_DEPRECATE(odp_pktio_mtu)(odp_pktio_t pktio);
 /**
  * Enable/Disable promiscuous mode on a packet IO interface.
  *
- * @param[in] pktio	Packet IO handle.
- * @param[in] enable	1 to enable, 0 to disable.
+ * @param pktio   Packet IO handle.
+ * @param enable  1 to enable, 0 to disable.
  *
  * @retval 0 on success
  * @retval <0 on failure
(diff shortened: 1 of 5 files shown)


include/odp/api/spec/packet_io.h
```diff
@@ -970,6 +970,36 @@ int odp_pktio_promisc_mode_set(odp_pktio_t pktio, odp_bool_t enable);
...
+ * For Ethernet, the frame length bytes start with MAC addresses and continue
+ * to the end of the payload. So, Ethernet checksum, interpacket gap
+ * and preamble bytes are excluded from the length.
+ *
+ * @param pktio  Packet IO handle.
+ *
+ * @return Maximum frame length at packet input
+ * @retval 0 on failure
+ */
+uint32_t odp_pktin_maxlen(odp_pktio_t pktio);
```
muvarov: bad naming. odp_pktin_xxx functions take argument odp_pktin_queue_t not odp_pktio_t.  If you want to keep odp_pktio_t as argument for function, then odp_pktio_capability_t is more logical place for it,

psavol: int odp_pktin_event_queue(odp_pktio_t pktio, odp_queue_t queues[], int num);
int odp_pktin_queue(odp_pktio_t pktio, odp_pktin_queue_t queues[], int num);
uint64_t odp_pktin_wait_time(uint64_t nsec);
uint64_t odp_pktin_ts_res(odp_pktio_t pktio);

These are not the first pktin/pkout functions that have pktio as param. Previously, I considered  odp_pktio_rx_xxx() and odp_pktio_tx_xxx() prefixes for these, but odp_pktin_xxx() and odp_pktout_xxx() was more compact and as intuitive.

Also these are not capabilities but the current configured frame lengths. Later on, we'll add configuration options for changing these and capabilities to indicate min/max values. This is now a  simple replacement of MTU function.

muvarov: @psavol then I'm ok.

include/odp/api/spec/packet_io.h
```diff
@@ -954,8 +954,8 @@ uint32_t ODP_DEPRECATE(odp_pktio_mtu)(odp_pktio_t pktio);
 /**
  * Enable/Disable promiscuous mode on a packet IO interface.
  *
- * @param[in] pktio	Packet IO handle.
- * @param[in] enable	1 to enable, 0 to disable.
+ * @param pktio   Packet IO handle.
+ * @param enable  1 to enable, 0 to disable.
  *
```
Bill-Fischofer-Linaro: This seems a worthwhile cleanup but seems orthogonal to the rest of this PR.

test/performance/odp_l2fwd.c
```diff
@@ -108,6 +108,7 @@ typedef struct {
 	int error_check;        /**< Check packet errors */
 	int sched_mode;         /**< Scheduler mode */
 	int num_groups;         /**< Number of scheduling groups */
+	int verbose;		/**< Verbose output */
 } appl_args_t;
```
Bill-Fischofer-Linaro: It's not obvious what these changes have to do with the stated purpose of this PR. This looks like a separate PR altogether.

psavol: I used this option to check the frame sizes. E.g. dpdk pktio returned IP MTU size (1500), not max ethernet frame size (1514). This an easy way to check it on various pktios as l2fwd takes interface as a command line parameter.

include/odp/api/spec/packet_io.h
psavol: This changes the same API header file. Doing it as part of the patch set avoids potential merge conflicts.

psavol:
Actually, these RFCs refer to various different MTUs and highlight why the change is needed. MTU may refer to maximum IP packet length that can fit into an ethernet frame (e.g. 1500 bytes), or what IP end points have agreed (PMTU) for a link (e.g. 1420 bytes when there's an IPSEC tunnel on the way), or IP packet size that fits into ethernet with PPP headers (1492 bytes), or etc.

Since ODP is not an IP (or any other protocol) stack, it does not maintain IP MTU information. ODP just returns what is currently configured into NIC RX/TX HW registers (pktin_maxlen, pktout_maxlen). Those are the absolute maximums that the HW can handle - no matter what (standard or proprietary) protocols application may be using on top of the link. odp_pktout_maxlen() is the maximum number of bytes that odp_pktout_send() can send out.

Bill-Fischofer-Linaro:
@pasvol The term MTU is also protocol-agnostic as you can find it in both Fibre Channel and Infiniband as well as Ethernet-based systems. The difference is that beyond the device MTU there is also the MTUs associated with various protocols. We just need to clarify in the docs that ODP's MTU support is for device MTUs.

Bill-Fischofer-Linaro:
If @bala-manoharan and @NikhilA-Linaro are OK with this change, I'm OK with it also.

muvarov:
@NikhilA-Linaro  do you have any update for this?

psavol:
At least, old Freescale devices had MAXFRM (max frame length) register. In practice, every NIC has max frame registers, but none has MTU registers. So, let's just merge this and move on.

http://elixir.free-electrons.com/linux/v4.2/source/drivers/net/ethernet/freescale/gianfar.c

/* Initialize the max receive frame/buffer lengths */
	gfar_write(&regs->maxfrm, priv->rx_buffer_size);
	gfar_write(&regs->mrblr, priv->rx_buffer_size);

muvarov:
Merged.

PR #299, opened 2017-11-16
[PATCH 2.0 v1] test: enable dpdk initialization before running validation on linux-dpdk

GBalakrishna:
currently linux-dpdk validation not being run due to missing dpdk environment setup which actually does through wrapper-script. The script is is being called by exporting LOG_COMPILER flag and used in run-test.sh common between linux-generic & linux-dpdk platforms. Currently wrapper-script already exist for linux-dpdk but not in linux-generic. This PR introduces an empty wrapper-script for linux-generic & set the LOG_COMPILER flag in common Makefile.inc

### DIFF
--- a/platform/linux-generic/test/wrapper-script.sh
+++ b/platform/linux-generic/test/wrapper-script.sh
@@ -1,0 +1,3 @@
+#wrapper script for pre setting environment for validation suit.
+#currently this is empty but needs to be created to make it align with
+#linux-dpdk.
--- a/test/Makefile.inc
+++ b/test/Makefile.inc
@@ -20,6 +20,8 @@ AM_CFLAGS = $(CUNIT_CFLAGS)
 
 AM_LDFLAGS = -L$(LIB) -static
 
+LOG_COMPILER = $(top_srcdir)/platform/@with_platform@/test/wrapper-script.sh
+
 @VALGRIND_CHECK_RULES@
 
 TESTS_ENVIRONMENT = ODP_PLATFORM=${with_platform} \
--- a/test/performance/Makefile.am
+++ b/test/performance/Makefile.am
@@ -31,6 +31,6 @@ odp_sched_latency_SOURCES = odp_sched_latency.c
 odp_scheduling_SOURCES = odp_scheduling.c
 odp_pktio_perf_SOURCES = odp_pktio_perf.c
 
-dist_check_SCRIPTS = $(TESTSCRIPTS)
+dist_check_SCRIPTS = $(TESTSCRIPTS) $(LOG_COMPILER)
 
 dist_check_DATA = udp64.pcap


platform/linux-generic/test/wrapper-script.sh
```diff
@@ -0,0 +1,3 @@
+#wrapper script for pre setting environment for validation suit.
```
heyi-linaro: @GBalakrishna I understand this PR will make linux-dpdk platform build really run all the tests, can you help elaborate why this script with only comments can help archive this?

test/Makefile.inc
```diff
@@ -20,6 +20,8 @@ AM_CFLAGS = $(CUNIT_CFLAGS)
 
 AM_LDFLAGS = -L$(LIB) -static
 
+LOG_COMPILER = $(top_srcdir)/platform/@with_platform@/test/wrapper-script.sh
```
heyi-linaro: linux-generic platform does not have this wrapper-script.sh

test/performance/Makefile.am
```diff
@@ -31,6 +31,6 @@ odp_sched_latency_SOURCES = odp_sched_latency.c
 odp_scheduling_SOURCES = odp_scheduling.c
 odp_pktio_perf_SOURCES = odp_pktio_perf.c
 
-dist_check_SCRIPTS = $(TESTSCRIPTS)
+dist_check_SCRIPTS = $(TESTSCRIPTS) $(LOG_COMPILER)
```
heyi-linaro: what happened to linux-generic platform since it does not have this wrapper shell script?

heyi-linaro:
This PR has errors in travis CI while building linux-dpdk, which were the 4 ipsec validation tests.

Does this mean this PR makes linux-dpdk build really run the validation tests, should this PR also include fixes for those ipsec validation failures?

platform/linux-generic/test/wrapper-script.sh
GBalakrishna: @heyi-linaro as I written it in the comment section in the script, it is to be aligned with linux-dpdk wrapper-script. I couldn't find any better way of enabling wrapper-script from linux-dpdk.

test/Makefile.inc
GBalakrishna: That's why I added an empty file that doesn't effect the test run.

test/performance/Makefile.am
GBalakrishna: it has now with this PR.

GBalakrishna:
@heyi-linaro Yes it actually runs now and tells that the api-next merge broke the linux-dpdk. As I mentioned in the cloud call before merging of api-next, whoever doing the merge needs to be aware that we are merging untested code but needs to fix the conflicts once we enable the validation on linux-dpdk.

heyi-linaro:
O, I see @GBalakrishna, you add the counterpart wrapper script in linux-generic platform
OK, I'll look into validation failures and fix them with this PR.

heyi-linaro:
I suggest that we will fix this linux-dpdk build and validation together in next merge window, which expects to be after odp-dpdk repo's upgrade after tiger-moth release, the benefits would be:

- All merge and validation fixes are done with odp-dpdk repo upgrade
- dpdk dependency is also upgraded to 17.08 and we can remove the temporary two dpdk builds.

GBalakrishna:
@heyi-linaro , @nagarahalli seems like not possible to just disable ipsec for linux-dpdk, we need to disable it for both the platforms. Wanted to check if that's what we wanted ?.

heyi-linaro:
@GBalakrishna seems not good to affect linux-generic build, better can do it only for linux-dpdk.

GBalakrishna:
@heyi-linaro I don't think it's possible to disable it only for linux-dpdk as the tests are common between the platforms. It would be better to fix ipsec changes for linux-dpdk instead of disabling the tests.

heyi-linaro:
Hi, @GBalakrishna as we talked by email the ipsec validation failures of linux-dpdk build are caused by its outdated odp_crypto.c impl, (I saw your email describing the difficulties of bringing it up to dpdk 17.08)

On 2.0 branch linux-dpdk build with dpdk 17.02 is a temporary situation and will import you and Matias patch to upgrade to dpdk 17.08 eventually.

I found a way to temporarily disable ipsec validations for linux-dpdk build only (and linux-generic build still evaluate them), do you think this is a good approach and can update it into your PR? Also we can create a GitHub ISSUE to track this temporary solution:

In platform/linux-dpdk/odp_crypto.c, function odp_crypto_capability(), return 0 to disable all capabilities:
```
diff --git a/platform/linux-dpdk/odp_crypto.c b/platform/linux-dpdk/odp_crypto.c
index 8e0f8a9..5257267 100644
--- a/platform/linux-dpdk/odp_crypto.c
+++ b/platform/linux-dpdk/odp_crypto.c
@@ -326,6 +326,9 @@ int odp_crypto_capability(odp_crypto_capability_t *capability)
        /* Initialize crypto capability structure */
        memset(capability, 0, sizeof(odp_crypto_capability_t));

+       /* */
+       return 0;
+
        cdev_count = rte_cryptodev_count();
        if (cdev_count == 0) {
                ODP_ERR("No crypto devices available\n");
```

Bill-Fischofer-Linaro:
This PR is still not passing all Travis tests. This should be fixed before it's ready to merge.

heyi-linaro:
Obsolete by PR #317 close this one

PR #300, opened 2017-11-20
[PATCH v1] [RFC] set a clean base for native drivers development

heyi-linaro:
1, temporarily disable the DDF validations and samples to enable excessive reorganization and cleanup.
2, cleanup DDF to remove duplicated aligns ,byteorders, etc.
3, move DDF into frameworks/driver, lately collect all bus drivers, vfio/uio drivers here
4, create platform/linux-generic/pktio/virtio for virtio-net driver development (copying DPDK PMDs)


heyi-linaro:
Hi, @nagarahalli, do you think I, Jianbo and Josep can collaborate like this on this branch? Once the code reaches stages we can rewrite commits to be more well-formed.

@joseppc @jianbo.liu@arm.com

This is my thoughts about code arrangement, so now we can copy DPDK eal pci and uio/vfio code into frameworks/driver/common and frameworks/driver/linux (linuxapp), and virtio code into pktio/virtio

Try mostly reuse DPDK code at first and after then thinking how to convert them into enumerators/devios.

nagarahalli:
@heyi-linaro 1) Cleanup DDF to remove duplicated byteorder, barriers, locks etc - I do not have any information on the history, why this was done. Can we talk about it in today's 2.0 call to get some information from Bill/Francois?
2) You had moved files from include/odp/drv/spec directory to framework/driver. I think we should not do this right now, otherwise the discussion will un-necessarily go into the packaging topics. If the duplicated files are removed, this directory will contain driver.h alone. If it does not bother much for driver development, we should leave it there for now.

nagarahalli:
@heyi-linaro I am fine with the rest of the changes.

heyi-linaro:
Hi, @nagarahalli I'm just making the driver framework simple and straightforward and collect all code into one place (for later code importing from DPDK also). let's talk in today's call.

joseppc:
I am ok with moving driver headers to the framework directory, unless they should be part of the ODP API.

Something I don't quite like is that some pktios depend on the driver framework, while others don't. Would it be possible to put those pktios dependent on the driver framework inside framework/drivers/pktio directory? And maybe even make it optional to compile them? Maybe for a future improvement...

Unrelated to the above, I assume there can be drivers for other drivers that are not necessarily NICs, and they could fit in its own directory too (I don't know, "classifier", etc).

nagarahalli:
@heyi-linaro Any code re-organization seems to be a touchy topic requiring lot of discussions :) Since this is development branch we should go ahead with experimenting with different directory structures. I am fine  with these changes for the development branch.

nagarahalli:
Once Travis CI passes, we can merge?

joseppc:
I think the build fails already in the original branch :disappointed:

joseppc:
Since #297 was merged, `odp_packet_io_pool_access.h` depends on `drv/shm.h`, so this PR can not be merged at the moment unless functions similar to `odpdrv_shm_pool_{alloc,free}` are provided.

heyi-linaro:
Close and prefer do it in the future.

PR #302, opened 2017-11-21
[PATCH API-NEXT v2] api: queue: lockfree queue

psavol:
Add queue parameter for indicating need for non-blocked operation. Non-blocking guarantee is important for real-time applications. E.g. when RT and non-RT threads uses queues for communication, RT threads must not block if an non-RT thread blocks (e.g. kernel interrupts it). Many HW implementations are inherently non-blocking, SW implementations need special care for non-block support.

### DIFF
--- a/include/odp/api/spec/queue.h
+++ b/include/odp/api/spec/queue.h
@@ -97,6 +97,54 @@ typedef enum odp_queue_op_mode_t {
 } odp_queue_op_mode_t;
 
 /**
+ * Non-blocking level
+ *
+ * A non-blocking level defines implementation guarantees for application
+ * progress when multiple threads operate on the same resource (e.g. a queue)
+ * simultaneously. The first level (ODP_BLOCKING) does not have any block
+ * freedom guarantees, but a suspending thread may block the other threads for
+ * the entire time it remains suspended (infinitely if crashed).
+ * On the contrary, actual non-blocking levels provide guarantees of progress:
+ *
+ * ODP_NONBLOCKING_LF:  A non-blocking and lock-free implementation guarantees
+ *                      that at least one of the threads successfully completes
+ *                      its operations, regardless of what other threads do.
+ *                      Application progress is guaranteed, but individual
+ *                      threads may starve while trying to execute their
+ *                      operations on the shared resource.
+ *
+ * ODP_NONBLOCKING_WF:  A non-blocking and wait-free implementation guarantees
+ *                      application progress with starvation freedom. All
+ *                      threads are guaranteed to complete their operations in
+ *                      a bounded number of steps, regardless of what other
+ *                      threads do.
+ *
+ * Non-blocking levels are listed from the weakest to the strongest guarantee of
+ * block freedom. Performance of a non-blocking implementation may be lower than
+ * the blocking one. Non-blocking guarantees are important e.g. for real-time
+ * applications when real-time and non real-time threads share a resource.
+ */
+typedef enum odp_nonblocking_t {
+	/** Blocking implementation. A suspeding thread may block all other
+	 *  threads, i.e. no block freedom guarantees. This is the lowest level.
+	 */
+	ODP_BLOCKING = 0,
+
+	/** Non-blocking and lock-free implementation. Other threads can make
+	 *  progress while a thread is suspended. Starvation freedom is not
+	 *  guaranteed.
+	 */
+	ODP_NONBLOCKING_LF,
+
+	/** Non-blocking and wait-free implementation. Other threads can make
+	 *  progress while a thread is suspended. Starvation freedom is
+	 *  guaranteed.
+	 */
+	ODP_NONBLOCKING_WF
+
+} odp_nonblocking_t;
+
+/**
  * Queue capabilities
  */
 typedef struct odp_queue_capability_t {
@@ -125,6 +173,10 @@ typedef struct odp_queue_capability_t {
 		  * store all available events. */
 		uint32_t max_size;
 
+		/** The strongest guarantee of block freedom that is supported
+		  * for plain queues. */
+		odp_nonblocking_t nonblocking;
+
 	} plain;
 
 	/** Scheduled queue capabilities */
@@ -137,6 +189,10 @@ typedef struct odp_queue_capability_t {
 		  * queues do not have a size limit, but a single queue can
 		  * store all available events. */
 		uint32_t max_size;
+
+		/** The strongest guarantee of block freedom that is supported
+		  * for scheduled queues. */
+		odp_nonblocking_t nonblocking;
 
 	} sched;
 
@@ -177,6 +233,13 @@ typedef struct odp_queue_param_t {
 	  * These parameters are considered only when queue type is
 	  * ODP_QUEUE_TYPE_SCHED. */
 	odp_schedule_param_t sched;
+
+	/** Non-blocking level
+	  *
+	  * Queue implementation must guarantee at least this level of block
+	  * freedom for queue enqueue and dequeue/schedule operations.
+	  * The default value is ODP_BLOCKING. */
+	odp_nonblocking_t nonblocking;
 
 	/** Queue context pointer
 	  *
--- a/platform/linux-generic/odp_queue.c
+++ b/platform/linux-generic/odp_queue.c
@@ -150,7 +150,9 @@ static int queue_capability(odp_queue_capability_t *capa)
 	capa->max_sched_groups  = sched_fn->num_grps();
 	capa->sched_prios       = odp_schedule_num_prio();
 	capa->plain.max_num     = capa->max_queues;
+	capa->plain.nonblocking = ODP_BLOCKING;
 	capa->sched.max_num     = capa->max_queues;
+	capa->sched.nonblocking = ODP_BLOCKING;
 
 	return 0;
 }
@@ -601,6 +603,7 @@ static void queue_param_init(odp_queue_param_t *params)
 	params->type = ODP_QUEUE_TYPE_PLAIN;
 	params->enq_mode = ODP_QUEUE_OP_MT;
 	params->deq_mode = ODP_QUEUE_OP_MT;
+	params->nonblocking = ODP_BLOCKING;
 	params->sched.prio  = ODP_SCHED_PRIO_DEFAULT;
 	params->sched.sync  = ODP_SCHED_SYNC_PARALLEL;
 	params->sched.group = ODP_SCHED_GROUP_ALL;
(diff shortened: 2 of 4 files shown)


include/odp/api/spec/queue.h
```diff
@@ -96,6 +96,32 @@ typedef enum odp_queue_op_mode_t {
...
+ * using the same resource for the entire time that one of the threads remain
+ * suspended. On the contrary, a lock-free implementation allows other threads
+ * to make progress while a thread is suspended. Blocking levels are listed from
+ * the weakest to the strongest guarantee of block freedom.
+ *
+ * Performance of non-blocking operations may be lower than blocking ones, but
+ * can be used safely e.g. between threads of real-time and non real-time parts
+ * of an application.
+ */
+typedef enum odp_blocking_level_t {
```
Bill-Fischofer-Linaro: Presumably you're talking about thread preemption here. Isn't that why we have dedicated cores? Why restrict this to just queues? `odp_packet_alloc()/free()` and other routines might be subject to the same considerations as well.

psavol:
The cover letter gives a use case: "Add queue parameter for indicating need for non-blocked operation. Non-blocking guarantee is important for real-time applications. E.g. when RT and non-RT threads uses queues for communication, RT threads must not block if an non-RT thread blocks (e.g. kernel interrupts it). Many HW implementations are inherently non-blocking, SW implementations need special care for non-block support."

RT <-> non RT thread message passing through a queue.

include/odp/api/spec/queue.h
psavol: If lockfreeness is not guaranteed, a thread interrupted during an enqueue/dequeue operation can block all other threads (spinning inside enq/deq operation). Dedicated thread may be also interrupted (or call system call) and control (slow path) threads typically share a core.

The type name is generic, so that we can extend this to other APIs later on. Queue is the most important one to fix for thread to thread communication.

muvarov:
@psavol I still think we need be more accurate with naming to not have overlaps in future. odp_blocking_level_t can be named odp_queue_blocking_level_t blvl; Also ODP_BLOCKING  -> ODP_QUEUE_BLOCKING_NONE and ODP_QUEUE_BLOCKING_LOCKFREE.  Also it looks like good to move odp_queue_op_mode_t  to capabilities because it's almost the same thing.

psavol:
The type name (odp_blocking_level_t) is generic, so that we could extend this to other APIs (e.g. pool, pktio, ...) later on. You wouldn't want to define the same thing again and again: odp_pool_blocking_level_t, odp_pktio_blocking_level_t, ...

Also enums values are not so straight forward since:
ODP_BLOCKING == this is a blocking implementation
ODP_LOCKFREE == this is a non-blocking implementation which is lockfree

psavol:
One option is to add "OP" for operation as a prefix: odp_op_blocking_level_t, ODP_OP_BLOCKING, ODP_OP_LOCKFREE, ...

It seemed unnecessary, but could be added to be a bit more future proof.

include/odp/api/spec/queue.h
nagarahalli: Why would the data plane thread get interrupted? The programming model for ODP has been that, the data plane threads are non preemptive.

Are you suggesting to change this programming model?

WonderfulVoid: This execution model might not hold for all types of ODP applications. E.g. control plane threads might run on non-dedicated cores and thus suffer interrupts and preemption. Even on dedicated and isolated cores, isolation could be imperfect (e.g. Linux 1Hz scheduling tick). Finally, HW is not perfect and deterministic and one limitation to scalability is (SW) dependencies between threads.

WonderfulVoid:
LOCK_FREE != NON_BLOCKING. Lock-free means a misbehaving thread (e.g. a thread that crashes in the wrong place) cannot stop progress for other threads so there will be system-wide progress. Wait-free indicates guaranteed per-thread progress. I think the term non-blocking can also give room to some less strict definitions, e.g. threads don't block each other but all threads must eventually complete (i.e not crash-safe) which I think would be good enough in many real-time use cases.

muvarov:
@psavol is it to merge this?

WonderfulVoid:
I think we need an additional level ODP_NONBLOCKING for implementations that are not blocking but also not properly lock-free. E.g. SW ring buffer designs where ring slots are first acquired, then read or written and finally released are not lock-free (a thread that dies in between acquire and release will stop other threads from making progress) but it is possibly to make them non-blocking (threads are not blocked from making progress due to temporary stalls in other threads). So non-blocking but not crash-safe (which a proper lock-free algorithm is).

I am raising this because I do have such a non-blocking ring buffer design which will be useful for latency sensitive applications.

psavol:
I think "temporary non-blocking" == blocking. It would be hard for a RT application to evaluate if a "temporary non-blocking" implementation is non-blocking enough. We are better of with terms that are widely accepted and used.

non-blocking == fast, but risk of blocking
lock-free == potentially slower, but no risk of blocking
wait-free == potentially even slower, but no risk of blocking or starvation

Potentially here means that with HW implementation all would be fast.

What would "temporary" mean? Suspend can last 1sec, 1msec, 1usec, one wrap around of the ring (how soon that would happen), ...?

WonderfulVoid:
I wrote "threads are not blocked from making progress due to temporary stalls in other threads" but if a thread dies in the middle of an operation, there will be problems (eventually the queue will become full/empty and no more enqueues/dequeues can be performed). So it is not really about how long other threads stall.

If a design cannot be lock-free/wait-free, an alternative to blocking is to fail without side effects and return a status code to the caller. This is better than just blocking. What do you want to call this?

Bill-Fischofer-Linaro:
The question of failure recovery is a broad one and one that we've avoided in ODP, so I don't think it makes sense to try to focus on that here. Given the nature of SoC HW and offloads it may not be possible to define recovery semantics that can be robust across all implementations other than a hot standby / failover model, which is likely to be the most useful in production anyway.

My problem with "Lock free" as a designator is that it refers to an implementation technique rather than an observed / desired behavior. What an application cares about is whether a given operation has a bounded execution time, guaranteed completion, etc. Whether that's achieved with or without locks, etc. is an implementation detail that the application really doesn't see or care about.

psavol:
Non-blocking algorithms are divided into lock-free and wait-free ones. See e.g. this page  https://en.wikipedia.org/wiki/Non-blocking_algorithm  and the references on bottom of it. The same terms are used in academic research, so those are well defined in the context of block freedom.

Bill-Fischofer-Linaro:
@WonderfulVoid Are you OK with this version or are their other changes you'd still like to see?

WonderfulVoid:
No I still want an ODP_NONBLOCKING level for queues that are not blocking but still neither lock-free nor wait-free. The explanation text is basically the same as the one currently used for ODP_NONBLOCKING_LF ("Other threads can make progress while a thread is suspended" until the queue is full/empty, then operations will fail not block, let's call this back-pressure) while lock-freedom also guarantees eternal progress for other threads even if one thread e.g. crashes or dies in the wrong place.

I don't care that this behaviour is not described in academic research. It is valid behaviour (back-pressure), possible to implement in software (performance is actually better than the standard blocking ring buffer in many scenarios) and possibly desired by some latency sensitive applications.

WonderfulVoid:
Depending on how we define "operation" (e.g. is a single enqueue or dequeue an operation?) and how we define "completion" (are enqueued events immediately available for dequeue or just eventually - sort of eventual consistency), obstruction freedom sort of defines the behaviour I am thinking of.

https://en.wikipedia.org/wiki/Non-blocking_algorithm
"Obstruction-freedom is the weakest natural non-blocking progress guarantee. An algorithm is obstruction-free if at any point, a single thread executed in isolation (i.e., with all obstructing threads suspended) for a bounded number of steps will complete its operation."

psavol:
Non-blocking means that there is no "bad place" to halt (for some time or inifinitely). If there's a bad place in the algorithm it's blocking.

Obstruction free algorithm is non-blocking, but may live-lock (if contention is not managed). As soon as contention disappears at least one of the thread succeeds to make progress again.

When the queue is full, an enqueue operation is allowed to fail (to enqueue the event) in both blocking and non-blocking implementations.

So, I think you try to specify a blocking algorithm == there's a "bad place" in the algorithm which breaks the queue if one thread suspends infinitely - e.g. all en-/dequeue operations fail (infinitely) for other threads if a thread crashed during dequeue. The queue gets full and stays full as the crashed thread folds some index, etc that other threads wait to increment (which does not happen anymore since it crashed).

brbrooks:
This PR could improve even more...

Simplify it down to adding a NORMAL or REALTIME param for **describing** the queue.

Elaborate about what REALTIME means and potential implementations in the spec. Keep the API simple and stable.

psavol:
You'd need to describe what normal and realtime means:
1) blocking vs non-blocking
2) what kind of block freesom is needed: obstruction freedom, lock freedom or wait freedom

This enumeration answers those questions. "Realtime" does not give enough specification and an implementation would need to expect the worse (realtime == wait freedom).

brbrooks:
The app use case clearly calls out for two types of queues, NORMAL and REALTIME. The user would only ever use two types of queues.

If there are multiple types of REALTIME queues, the app would still only ever use the 'best' one, where 'best' is defined by wait-free > lock-free > obstruction-free. So, you only need one representation of a REALTIME queue at the API layer. Leave it up to the implementations to do the best they can (or copy shamelessly).

psavol:
Lock-free is rather easy to implement with CAS, while wait-free is not. I did originally defined only two levels: blocking and lock-free. People didn't want to wait for later to define other levels, so wait-free was added. HW implementation can easily support all levels, while SW only blocking and lock-free.

These three levels allow e.g. wait-free SW implementation to emerge when it's actually needed. We'll start with lock-free SW implementation.

Hopefully, this starts to be clear after 3 weeks of discussion.

brbrooks:
Your last reply is only talking about queue implementations. Do you understand that we're really only ever talking about two types of queues from the application's perspective?

psavol:
From applcation POV, it makes a difference if a (real-time) thread may starve (lock-free) or not (wait-free). If application is OK with the risk of starving and wants better performance it will choose lock-free. If it really needs starvation freedom, it must select wait-free - no matter how slow that will be.

Lock-free and wait-free algorithms offer trade-off between starvation freedom vs performance.

brbrooks:
Aha.. so a REALTIME queue corresponds to a wait-free queue.

This is starting to get complicated, but is there a real use case for a third queue type: NORMAL, REALTIME_SOFT, REALTIME_HARD? If the user knows that they don't need REALTIME_HARD, how do they (intelligently) choose btwn REALTIME_SOFT and NORMAL? Does it matter at this point?

psavol:
We don't need to invent new terms (realtime, or any flavor of it) as lock-free and wait-free are well defined terms already. Applications vary, also real-time applications may or may not be OK with just lock-free guarantees. As many times explained already, the difference between lock-free and wait-free is the number of thread guaranteed to progress. A real-time app can very well meet it's latency goal also with one thread progressing.

Real-time is property of the entire system, a queue is one piece of the puzzle and cannot alone make an application real-time. So, queue gives lock-free/wait-free guarantee (when application asks for it) - nothing more is promised here.

brbrooks:
My goal was to simplify the API and tailor it towards the use case rather than the implementation. Thanks for the discussion.

muvarov:
Merged.

PR #306, opened 2017-11-22
[PATCH API-NEXT v2] linux-gen: dpdk: check config bits not capa bits

psavol:
Application is required to set L3 offset only when it requests
checksum offload.

This fixes https://bugs.linaro.org/show_bug.cgi?id=3491

It caused segfault to assume that L3 offset is always when
the interface supports checksumming.


bogdanPricope:
void odp_packet_l4_chksum_insert(odp_packet_t pkt, int insert);

* Override checksum insertion configuration per packet. This per packet setting
 * overrides a higher level configuration for checksum insertion into a L4
 * header during packet output processing.

That means, you are overriding pktout_cfg when you are using odp_packet_l4_chksum_insert()

A offset validity check should be enough:

e.g.
 
if (pkt_p->l3_offset == ODP_PACKET_OFFSET_INVALID ||
	pkt_p->l4_offset == ODP_PACKET_OFFSET_INVALID)
	return;

bogdanPricope:

muvarov:
Merged.

PR #307, opened 2017-11-22
[PATCH API-NEXT v1] example: tm: update gitignore

psavol:
Added missing files to gitignore file.


Bill-Fischofer-Linaro:
@muvarov We've had several instances of PRs getting merged that create various temp outputs that should be in `.gitignore` files that then need follow-on PRs to clean up. Not sure if a check for these could be easily added to Travis.

muvarov:
Merged.

PR #308, opened 2017-11-22
[PATCH API-NEXT v1] pktio: dpdk: skip TX csum calculation if L3/L4 offsets are invalid

bogdanPricope:
Skip TX csum calculation if L3/L4 offsets are invalid

L3/L4 offsets may be invalid (ODP_PACKET_OFFSET_INVALID)
for a TX packet. This will prevent from providing the arguments
needed by dpdk for csum calculation.


muvarov:
@psavol  Petri it looks like good bug fix for dpdk issue. Do you want to review that?

psavol:
This is against API spec, since application may only request IP checksum and set only L3 offset (not both).

I did send v2 of my patch.

bogdanPricope:
PR306 is replacing this one.

PR #309, opened 2017-11-24
[PATCH v12] one more try to add nice stats for Shippable

muvarov:


test/common/odp_cunit_common.c
```diff
@@ -7,13 +8,15 @@
...
 #include <string.h>
+#include <unistd.h>
 #include <odp_api.h>
 #include "odp_cunit_common.h"
 #include <odp/helper/odph_api.h>
 /* Globals */
 static odph_odpthread_t thread_tbl[MAX_WORKERS];
 static odp_instance_t instance;
 
+extern const char *__progname;
```
lumag: Please use a copy of `argv[0]` instead of progname.

test/common/odp_cunit_common.c
```diff
@@ -291,13 +294,30 @@ static int cunit_update_suite(odp_suiteinfo_t *updated_sinfo)
...
+		CU_set_output_filename(name);
+	} else {
+		/* forked processes do not have __progname */
+		ret = readlink("/proc/self/exe", name, sizeof(name) - 1);
+		name[ret] = 0;
+		CU_set_output_filename(basename(name));
+	}
+
+	CU_automated_run_tests();
+#else
```
lumag: This should be refactored to a function which is selected basing on `TEST_CUNIT_XML`, rather than having `#ifdef` directly in the function body.

test/m4/validation.m4
```diff
@@ -19,6 +19,14 @@ AS_IF([test "x$cunit_support" != "xno"],
 		 [cunit_support=no])
 ])])
 
+
+##########################################################################
+AC_ARG_ENABLE([cunit_out_xml],
+    [AS_HELP_STRING([--enable-cunit_out_xml],
+		    [include additional debugging code ]
+		    [(set to 'full' to enable all --enable-*-debug-* options)])],
```
lumag: This is just incorrect

test/Makefile.inc
```diff
@@ -17,6 +17,9 @@ INCFLAGS = \
 
 AM_CPPFLAGS = $(INCFLAGS)
 AM_CFLAGS = $(CUNIT_CFLAGS)
+if cunit_out_xml
+AM_CFLAGS  += -DTEST_CUNIT_XML
+endif
```
lumag: This looks like an overkill. I'd suggest to use Autoconf variable rather than Automake conditional here.

test/common/odp_cunit_common.c
muvarov: save argv[0] in some global variable?

test/m4/validation.m4
muvarov: yep, thanks it's copy pase

muvarov:
@ttrahan this PR adds Junit stats and copies it to results directory. But I think parsing script needs to be adjusted to support that output. I do not see that stats are updated.

test/common/odp_cunit_common.c
lumag: @muvarov yes

muvarov:
@lumag  is this version ok?

scripts/Dockerfile
```diff
@@ -22,6 +22,7 @@ RUN sudo apt-get update && sudo apt-get install -yy \
 RUN wget "https://github.com/Shippable/support/raw/master/bin/reports" && \
     chmod +x reports && \
     rm -f /home/shippable/cexec/bin/reports && \
+	mkdir -p /home/shippable/cexec/bin && \
```
ttrahan: @muvarov: this dir should already exist in the base image for the Dockerfile. Are you seeing the need to create the directory in your testing?

test/m4/validation.m4
```diff
@@ -31,3 +31,14 @@ AM_CONDITIONAL([test_vald], [test "x$test_vald" = "xyes"])
 
 AC_SUBST([CUNIT_CFLAGS])
 AC_SUBST([CUNIT_LIBS])
+
+AC_ARG_ENABLE([cunit_out_xml],
+    [AS_HELP_STRING([--enable-cunit_out_xml],
+		    [output to xml]
+		    [(output test results to xml instead of plain text)])],
+		    cunit_out_xml=yes, cunit_out_xml=no)
```
lumag: Ideally this should also be escaped with square brackets.

test/common/odp_cunit_common.c
lumag: @muvarov I still do not like this piece. Maybe you can add a runtime switch through environment variable? Then there will be no requirement for extra AM_CONDITIONAL/AC_DEFINE

test/common/odp_cunit_common.h
```diff
@@ -15,8 +15,11 @@
 
 #include <stdint.h>
 #include <inttypes.h>
-#include "CUnit/Basic.h"
-#include "CUnit/TestDB.h"
+#include <CUnit/Basic.h>
+#include <CUnit/TestDB.h>
+#if TEST_CUNIT_XML
+#include <CUnit/Automated.h>
+#endif
```
lumag: No need to hide it under `#if`

test/m4/validation.m4
```diff
@@ -31,3 +31,14 @@ AM_CONDITIONAL([test_vald], [test "x$test_vald" = "xyes"])
...
+AS_IF([test "x$cunit_out_xml" == "xyes"], [TEST_CUNIT_XML=1],
+      [TEST_CUNIT_XML=0])
+AC_DEFINE_UNQUOTED([TEST_CUNIT_XML], [$TEST_CUNIT_XML],
+		   [Define to 1 to output to xml])
+AM_CONDITIONAL([cunit_out_xml], [test "x$cunit_out_xml" = "xyes"])
```
lumag: Conditional is unused

test/common/odp_cunit_common.h
muvarov: our arm tests use some previous version where function declared as finc(); i.e. not func(void) and gcc fails on it. In task  which run inside Shippable it's corrected.

lumag: @muvarov then it's better to update CUnit version on target, rather than workaround its bugs.

muvarov:
@lumag fixed comments. This version has to show counters. For Cunit I think we only can fix header because it's already the latest release. I added comment to travis. Please take a look.

muvarov:
Travis fails here due to caches for cunit. After merge it has to be ok. Clean run is here:
https://travis-ci.org/muvarov/odp/builds/316450770

muvarov:
@lumag ping.

lumag:
@muvarov still an #ifdef inside function code. I really suggest to move xml/plain output type to test arguments.

muvarov:
v12: switched to getenv() instead of ifdefs as agreed.

lumag:

muvarov:
Merged.

PR #310, opened 2017-11-24
[PATCH API-NEXT v1] api: time: time difference in nsec

psavol:
Added function which returns time difference in nanoseconds.
This short cuts a common pattern...

t1 = odp_time_global();
foo();
t2 = odp_time_global();
printf("foo() takes %u nsec\n", odp_time_diff_ns(t2, t1);

... by combining convert and diff functions.


muvarov:
this looks very reasonable for me. In many cases app just needs time diff in some readable form.

muvarov:
Merged.

PR #311, opened 2017-11-28
[PATCH API-NEXT v2] IPsec API update

lumag:


lumag:
@bala-manoharan could you please be more specific, what is the issue with commit message? Second patch bears the `api, linux-gen: ` prefix

Bill-Fischofer-Linaro:
Normally API patches would be separate, however in this case bisectability requires that the `const` qualifier be applied to both the API and the implementation, so I don't see a problem here.

bala-manoharan:
@lumag The commit message seems little odd to me. Since this is mainly an api change I would suggest naming the commit as "api: ipsec: ..." the description can have the details of linux-generic changes...

muvarov:
Merged.

PR #312, opened 2017-11-28
[PATCH 2.0 v3] allocate ring memory based on pool size

nagarahalli:
The maximum size of the pool ring is 1M entries as it uses an array. However, the application provides the number of buffers in a pool during pool creation. Hence, the maximum pool ring size can be dynamic.

### DIFF
--- a/platform/linux-generic/pool/generic.c
+++ b/platform/linux-generic/pool/generic.c
@@ -156,11 +156,12 @@ static int generic_pool_term_local(void)
 	return 0;
 }
 
-static pool_t *reserve_pool(void)
+static pool_t *reserve_pool(uint32_t ring_size)
 {
 	int i;
 	pool_t *pool;
 	char ring_name[ODP_POOL_NAME_LEN];
+	uint32_t ring_shm_size;
 
 	for (i = 0; i < ODP_CONFIG_POOLS; i++) {
 		pool = pool_entry(i);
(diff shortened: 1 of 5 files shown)


platform/linux-generic/pool/generic.c
```diff
@@ -506,11 +510,6 @@ static int check_params(odp_pool_param_t *params)
 
 	switch (params->type) {
 	case ODP_POOL_BUFFER:
-		if (params->buf.num > capa.buf.max_num) {
```
Bill-Fischofer-Linaro: Same here:
```
if (capa.buf.max_num && params->buf.num > capa.buf.max_num) ...
```

platform/linux-generic/pool/generic.c
```diff
@@ -550,10 +549,6 @@ static int check_params(odp_pool_param_t *params)
 		break;
 
 	case ODP_POOL_TIMEOUT:
-		if (params->tmo.num > capa.tmo.max_num) {
```
Bill-Fischofer-Linaro: And here too.

Bill-Fischofer-Linaro:
We discussed this PR briefly in the Wednesday ARCH call. @psavol asked for the rationale behind it. Presumably it is to conserve memory? Can we quantify the expected savings?

nagarahalli:
@Bill-Fischofer-Linaro odp_pool_create API already takes number of buffers/packets required in the pool as input from the user. For ODP 2.0, only restriction to support any number is available amount of memory. Hence, it is not required to add an artificial limit which is also an additional configuration to worry about.

May be it is ok to have that limit in Linux-Generic to show clarity, but for ODP 2.0, such a limitation is not required.

nagarahalli:
@Bill-Fischofer-Linaro I suggest that we look into how this will go to master when Tiger Moth is done. For now, we should get it into ODP 2.0.

heyi-linaro:
Merged and closed.

PR #314, opened 2017-11-29
[PATCH API-NEXT v10] IPsec checksumming API

lumag:
Separate checksumming flags into separate type to be used by both packet parse and IPsec functions.

### DIFF
--- a/include/odp/api/spec/ipsec.h
+++ b/include/odp/api/spec/ipsec.h
@@ -158,33 +158,9 @@ typedef struct odp_ipsec_inbound_config_t {
 	/** Flags to control IPSEC payload data checks up to the selected parse
 	 *  level. Checksum checking status can be queried for each packet with
 	 *  odp_packet_l3_chksum_status() and odp_packet_l4_chksum_status().
+	 *  Default value for all bits is 0 (skip all checksum checks).
 	 */
-	union {
-		/** Mapping for individual bits */
-		struct {
-			/** Check IPv4 header checksum in IPSEC payload.
-			 *  Default value is 0. */
-			uint32_t ipv4_chksum   : 1;
-
-			/** Check UDP checksum in IPSEC payload.
-			 *  Default value is 0. */
-			uint32_t udp_chksum    : 1;
-
-			/** Check TCP checksum in IPSEC payload.
-			 *  Default value is 0. */
-			uint32_t tcp_chksum    : 1;
-
-			/** Check SCTP checksum in IPSEC payload.
-			 *  Default value is 0. */
-			uint32_t sctp_chksum   : 1;
-		} check;
-
-		/** All bits of the bit field structure
-		  *
-		  * This field can be used to set/clear all flags, or bitwise
-		  * operations over the entire structure. */
-		uint32_t all_check;
-	};
+	odp_proto_chksums_t chksums;
 
 } odp_ipsec_inbound_config_t;
 
--- a/include/odp/api/spec/packet.h
+++ b/include/odp/api/spec/packet.h
@@ -1179,6 +1179,31 @@ int odp_packet_move_data(odp_packet_t pkt, uint32_t dst_offset,
  */
 
 /**
+ * Flags to control packet data checksum checking
+ */
+typedef union odp_proto_chksums_t {
+	/** Individual checksum bits. */
+	struct {
+		/** IPv4 header checksum */
+		uint32_t ipv4   : 1;
+
+		/** UDP checksum */
+		uint32_t udp    : 1;
+
+		/** TCP checksum */
+		uint32_t tcp    : 1;
+
+		/** SCTP checksum */
+		uint32_t sctp   : 1;
+
+	} chksum;
+
+	/** All checksum bits. This can be used to set/clear all flags. */
+	uint32_t all_chksum;
+
+} odp_proto_chksums_t;
+
+/**
  * Packet parse parameters
  */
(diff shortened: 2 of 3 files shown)


include/odp/api/spec/packet.h
```diff
@@ -1178,6 +1178,31 @@ int odp_packet_move_data(odp_packet_t pkt, uint32_t dst_offset,
  *
  */
 
+/**
+ * Flags to control payload checksum checking
```
Bill-Fischofer-Linaro: Why "payload"? Wouldn't "packet checksum checking" be more accurate?

lumag: @Bill-Fischofer-Linaro updated

lumag:
@psavol could you please review.

include/odp/api/spec/packet.h
```diff
@@ -1178,6 +1178,31 @@ int odp_packet_move_data(odp_packet_t pkt, uint32_t dst_offset,
...
+		/** Check SCTP checksum */
+		uint32_t sctp_chksum   : 1;
+
+	} check;
+
+	/** All check bits. This can be used to set/clear all flags. */
+	uint32_t all_check;
+
+} odp_packet_parse_chksum_t;
+
```
psavol: Better name for this type would be odp_proto_chksums_t, just a bit field list of protocol checksums. Similar to other odp_proto_xxx_t types.

typedef union odp_packet_chksums_t {
	/** Individual checksum bits. */
	struct {
		/** IPv4 header checksum */
		uint32_t ipv4   : 1;

		/** UDP checksum */
		uint32_t udp    : 1;

		/** TCP checksum */
		uint32_t tcp    : 1;

		/** SCTP checksum */
		uint32_t sctp   : 1;

	} chksum;

	/** All checksum bits. This can be used to set/clear all flags. */
	uint32_t all_chksum;

} odp_packet_chksums_t;

Then odp_packet_parse_param_t would contain: proto, layer, check.

For example:
parse_param.check.chksum.ip = 1;
parse_param.check.chksum.udp = 1;

lumag: @psavol updated

lumag:
@psavol Do you think we should extract flags from `odp_pktout_config_opt_t` and `odp_ipsec_outbound_config_t` to common structure?

lumag:
@psavol ping

include/odp/api/spec/packet.h
psavol: There was actually multiple suggestions:
1) odp_proto_chksums_t == bit field of all checksums, chksum_t would be an enum of one checksum
2) struct == chksum (not check)
3) bits: ipv4, udp, tcp, sctp (not udp_chksum)
4) parser function API is new and still in api-next. So, parser_param_t could re-use this type. For example, parse_param.check.chksum.udp = 1

lumag: @psavol sorry, it looks like I misviewed original comment and missed these suggestions. PR is updated now.

Number 4 is already in place, but we do not have API checks for packet checksum parsing (and it is not implemented anyway). I'm looking onto that part now.

include/odp/api/spec/ipsec.h
```diff
@@ -159,32 +159,7 @@ typedef struct odp_ipsec_inbound_config_t {
...
-			uint32_t sctp_chksum   : 1;
-		} check;
-
-		/** All bits of the bit field structure
-		  *
-		  * This field can be used to set/clear all flags, or bitwise
-		  * operations over the entire structure. */
-		uint32_t all_check;
-	};
+	odp_proto_chksums_t check;
```
psavol: Default value == 0 for all chksum bits must be documented still here. This struct is initialized by odp_ipsec_config_init().

lumag: @psavol fixed

include/odp/api/spec/packet.h
```diff
@@ -1194,26 +1219,7 @@ typedef struct odp_packet_parse_param_t {
...
-
-			/** Check SCTP checksum */
-			uint32_t sctp_chksum   : 1;
-
-		} check;
-
-		/** All check bits. This can be used to set/clear all flags. */
-		uint32_t all_check;
-	};
+	odp_proto_chksums_t check;
```
psavol: Sorry that I'm changing my mind here. But just realized that the original "check" bit-field was meant to be extendable with potential other checks, which are not only checksums but e.g. "check.ipv6.extension_xyz". So, this definition (check == chksums only) is unnecessary limiting. Could you change it to e.g.

odp_proto_chksums_t sums;

I think this is still readable:
parse.sums.all_chksum = 0;
parse.sums.chksum.udp = 1;
parse.sums.chksum.tcp = 1;

And then we are more future proof again:
parse.ipv6.all_ext = 0;
parse.ipv6.ext.mobile_ip = 1;

include/odp/api/spec/ipsec.h
psavol: This would be also "sums" then.

include/odp/api/spec/packet.h
lumag: @psavol changing to `checksums` (rather than just `sums`) name.

psavol: Then it would need to be "chksums" per the type and chksum used in all other APIs (instead "checksum").

Note that the IPSEC param name needs the same update.

lumag: done

muvarov:
Merged.

PR #316, opened 2017-11-30
[PATCH API-NEXT v1] optimize pktio performance

matiaselo:
Optimize pktio performance by rearranging odp_buffer_hdr_t and odp_packet_hdr_t struct members.

Event subtype is only used by packets, so move subtype to packet header to optimize cache usage. This change fixes the ~30% performance penalty in l2fwd (zero-copy dpdk pktio) caused by the initialization of event subtype.

### DIFF
--- a/platform/linux-generic/include/odp_packet_internal.h
+++ b/platform/linux-generic/include/odp_packet_internal.h
@@ -132,6 +132,9 @@ typedef struct {
 	uint16_t headroom;
 	uint16_t tailroom;
 
+	/* Event subtype */
+	int8_t subtype;
+
 	/*
 	 * Members below are not initialized by packet_init()
 	 */
@@ -195,6 +198,16 @@ static inline seg_entry_t *seg_entry_last(odp_packet_hdr_t *hdr)
 	return &last->buf_hdr.seg[last_seg];
 }
 
+static inline odp_event_subtype_t packet_subtype(odp_packet_t pkt)
+{
+	return odp_packet_hdr(pkt)->subtype;
+}
+
+static inline void packet_subtype_set(odp_packet_t pkt, int ev)
+{
+	odp_packet_hdr(pkt)->subtype = ev;
+}
+
 /**
  * Initialize packet
  */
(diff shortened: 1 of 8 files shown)


platform/linux-generic/include/odp_packet_internal.h
```diff
@@ -234,9 +247,10 @@ static inline void packet_init(odp_packet_hdr_t *pkt_hdr, uint32_t len)
 	pkt_hdr->headroom  = CONFIG_PACKET_HEADROOM;
 	pkt_hdr->tailroom  = pool->seg_len - seg_len + CONFIG_PACKET_TAILROOM;
 
-	pkt_hdr->input = ODP_PKTIO_INVALID;
-	pkt_hdr->buf_hdr.event_subtype = ODP_EVENT_PACKET_BASIC;
+	if (odp_unlikely(pkt_hdr->subtype != ODP_EVENT_PACKET_BASIC))
+		pkt_hdr->subtype = ODP_EVENT_PACKET_BASIC;
```
lumag: Is this faster than just always setting the subtype?

platform/linux-generic/odp_event.c
```diff
@@ -24,17 +25,24 @@ odp_event_type_t odp_event_type(odp_event_t event)
 
 odp_event_subtype_t odp_event_subtype(odp_event_t event)
 {
-	return _odp_buffer_event_subtype(odp_buffer_from_event(event));
+	if (_odp_buffer_event_type(odp_buffer_from_event(event)) !=
+			ODP_EVENT_PACKET)
+		return ODP_EVENT_NO_SUBTYPE;
+
+	return packet_subtype(odp_packet_from_event(event));
 }
```
lumag: Looking on this code. Maybe we can change this part so that subtypes do exist only for packets from the API point of view? So instead of `odp_event_subtype()` we will have `odp_packet_subtype()`.

platform/linux-generic/include/odp_packet_internal.h
matiaselo: Yes, there's about 1% difference in l2fwd throughput.

platform/linux-generic/odp_event.c
matiaselo: That's the situation today, but in the future we may add subtypes for also other event types. With odp_event_subtype() the application has to do only a single API call compared to e.g. odp_packet_subtype() where an application would have to always call odp_event_type() first.

lumag: ok, fine

lumag:

matiaselo:
@lumag Thanks!

muvarov:
Merged.

PR #317, opened 2017-11-30
[PATCH 2.0 v1] linux-dpdk: enable validation & crypto updates

GBalakrishna:
This patch series has enabled the linux-dpdk validation & fixed the crypto that caused validation failure


heyi-linaro:
Updated by #323 and close this one.

PR #318, opened 2017-12-01
[PATCH API-NEXT v1] api: multi event functions

psavol:
Extended common event / packet functions with multi versions. It's more efficient to process multiple events / packets with a single call than with separate calls. Event / packet free implementation can be more efficient when all events / packets are from the same pool (xxx_free_sp()). Event conversion to packets one of the most used ODP functions, added a functions to convert and filter (check event type + convert) multiple packets at a time.

All implementation may be optimized later. Multi APIs for buffers could be added also, but event / packet APIs are higher priority.

### DIFF
--- a/include/odp/api/spec/event.h
+++ b/include/odp/api/spec/event.h
@@ -126,6 +126,22 @@ odp_event_type_t odp_event_types(odp_event_t event,
 				 odp_event_subtype_t *subtype);
 
 /**
+ * Event type of multiple events
+ *
+ * Returns the number of first events in the array which have the same event
+ * type. Outputs the event type of those events.
+ *
+ * @param      event    Array of event handles
+ * @param      num      Number of events (> 0)
+ * @param[out] type     Event type pointer for output
+ *
+ * @return Number of first events (1 ... num) with the same event type
+ *         (includes event[0])
+ */
+int odp_event_type_multi(const odp_event_t event[], int num,
+			 odp_event_type_t *type);
+
+/**
  * Get printable value for an odp_event_t
  *
  * @param hdl  odp_event_t handle to be printed
--- a/include/odp/api/spec/event.h
+++ b/include/odp/api/spec/event.h
@@ -150,6 +150,28 @@ uint64_t odp_event_to_u64(odp_event_t hdl);
 void odp_event_free(odp_event_t event);
 
 /**
+ * Free multiple events
+ *
+ * Otherwise like odp_event_free(), but frees multiple events to their
+ * originating pools.
+ *
+ * @param event    Array of event handles
+ * @param num      Number of events to free
+ */
+void odp_event_free_multi(const odp_event_t event[], int num);
+
+/**
+ * Free multiple events to the same pool
+ *
+ * Otherwise like odp_event_free_multi(), but all events must be from the
+ * same originating pool.
+ *
+ * @param event    Array of event handles
+ * @param num      Number of events to free
+ */
+void odp_event_free_sp(const odp_event_t event[], int num);
+
+/**
  * @}
  */
 
--- a/include/odp/api/spec/event.h
+++ b/include/odp/api/spec/event.h
@@ -19,11 +19,12 @@
 extern "C" {
 #endif
 
+#include <odp/api/packet.h>
+
 /** @defgroup odp_event ODP EVENT
  *  Operations on an event.
  *  @{
  */
-
 
 /**
  * @typedef odp_event_t
@@ -140,6 +141,27 @@ odp_event_type_t odp_event_types(odp_event_t event,
  */
 int odp_event_type_multi(const odp_event_t event[], int num,
 			 odp_event_type_t *type);
+
+/**
+ * Filter and convert packet events
+ *
+ * Checks event type of all input events, converts all packet events and outputs
+ * packet handles. Returns the number packet handles outputted. Outputs the
+ * remaining, non-packet event handles to 'remain' array. Handles are outputted
+ * to both arrays in the same order those are stored in 'event' array. Both
+ * output arrays must fit 'num' elements.
+ *
+ * @param      event    Array of event handles
+ * @param[out] packet   Packet handle array for output
+ * @param[out] remain   Event handle array for output of remaining, non-packet
+ *                      events
+ * @param      num      Number of events (> 0)
+ *
+ * @return Number of packets outputted (0 ... num)
+ */
+int odp_event_filter_packet(const odp_event_t event[],
+			    odp_packet_t packet[],
+			    odp_event_t remain[], int num);
 
 /**
  * Get printable value for an odp_event_t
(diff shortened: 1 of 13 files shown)


include/odp/api/spec/event.h
```diff
@@ -149,6 +149,28 @@ uint64_t odp_event_to_u64(odp_event_t hdl);
...
+/**
+ * Free multiple events to the same pool
+ *
+ * Otherwise like odp_event_free_multi(), but all events must be from the
+ * same originating pool.
+ *
+ * @param event    Array of event handles
+ * @param num      Number of events to free
+ */
+void odp_event_free_sp(const odp_event_t event[], int num);
```
Bill-Fischofer-Linaro: Not clear why this is needed. How does an application determine this more efficiently than the implementation would?

include/odp/api/spec/packet.h
```diff
@@ -204,6 +204,17 @@ void odp_packet_free(odp_packet_t pkt);
...
+/**
+ * Free multiple packets to the same pool
+ *
+ * Otherwise like odp_packet_free_multi(), but all packets must be from the
+ * same originating pool.
+ *
+ * @param pkt           Array of packet handles
+ * @param num           Number of packets to free
+ */
+void odp_packet_free_sp(const odp_packet_t pkt[], int num);
```
Bill-Fischofer-Linaro: Again, it's not clear why this is needed.

include/odp/api/spec/packet.h
```diff
@@ -244,6 +267,16 @@ odp_packet_t odp_packet_from_event(odp_event_t ev);
...
 
+/**
+ * Convert multiple packet handles to events
+ *
+ * @param      pkt  Array of packet handles to convert
+ * @param[out] ev   Event handle array for output
+ * @param      num  Number of packets and events
+ */
+void odp_packet_to_event_multi(const odp_packet_t pkt[], odp_event_t ev[],
+			       int num);
```
Bill-Fischofer-Linaro: Again since `odp_packet_to_event()` is likely to be just a cast that can be inlined away it's not clear why a `multi` version is needed.

include/odp/api/spec/packet.h
```diff
@@ -235,6 +246,18 @@ int odp_packet_reset(odp_packet_t pkt, uint32_t len);
...
+ * Convert multiple packet events to packet handles
+ *
+ * All events must be of type ODP_EVENT_PACKET.
+ *
+ * @param[out] pkt  Packet handle array for output
+ * @param      ev   Array of event handles to convert
+ * @param      num  Number of packets and events
+ */
+void odp_packet_from_event_multi(odp_packet_t pkt[], const odp_event_t ev[],
+				 int num);
```
Bill-Fischofer-Linaro: Even with the `odp_event_type_multi()` API semantics, it's still not clear why this is needed. Given that `odp_packet_from_event()` is likely just a cast there doesn't seem to be a great deal to be gained by having a `multi` version of this which can't be easily inlined away.

include/odp/api/spec/event.h
```diff
@@ -141,6 +142,27 @@ odp_event_type_t odp_event_types(odp_event_t event,
...
+ * @param[out] packet   Packet handle array for output
+ * @param[out] remain   Event handle array for output of remaining, non-packet
+ *                      events
+ * @param      num      Number of events (> 0)
+ *
+ * @return Number of packets outputted (0 ... num)
+ */
+int odp_event_filter_packet(const odp_event_t event[],
+			    odp_packet_t packet[],
+			    odp_event_t remain[], int num);
```
Bill-Fischofer-Linaro: Doesn't this reorder events? As described consider an array of events that consist of packets (p) and non-packets (np).

With an input array of events: np1, p1, np2, p2, p3 as described `odp_event_filter_packet()` would have RC = 3, the `packet[]` array would contain p1, p2, p3, and the `remain[]` event would contain np1, np2.  Is that behavior what we want?

include/odp/api/spec/event.h
psavol: Application knows the configuration, so it knows when all packets (e.g. from certain queue/pkt input/etc) are allocated from the same pool. When that's the case, free implementation can be more efficient (than today) as it does not have to access/check/sort all packets of the burst. It can just free all into the pool of the first packet.

platform/linux-generic/odp_packet.c
```diff
@@ -944,6 +949,45 @@ odp_event_t odp_packet_to_event(odp_packet_t pkt)
...
+void odp_packet_to_event_multi(const odp_packet_t pkt[], odp_event_t ev[],
+			       int num)
+{
+	int i;
+
+	for (i = 0; i < num; i++)
+		ev[i] = odp_packet_to_event(pkt[i]);
+}
+
+int odp_event_filter_packet(const odp_event_t event[],
```
muvarov: is it valid case when odp_schedule() returns mixed events types? I.e. some packets, some timeouts, some packets from one pool and packets from other pool.

include/odp/api/spec/event.h
muvarov: why not odp_schedule_type(ODP_PACKET,  void  *array) ? I.e. if you need only packets then do scheduling only for patches. That has to be more fast than get all and then filer them.

psavol: " Handles are outputted to both arrays in the same order those are stored in 'event' array. "

Events are not reordered. Packets are copied into packet[] and non-packets into remain[]. Both arrays maintain p and np order of the original array. Also original array is not modified.

packet[] = p1, p2, p3
remain[] = np1, np2

include/odp/api/spec/packet.h
psavol: The same answer. Improved implementation efficiency.

include/odp/api/spec/packet.h
psavol: pkts = odp_schedule_multi(NULL, ODP_SCHED_NO_WAIT, ev_tbl, MAX_PKT_BURST);

if (pkts <= 0)
	continue;

for (i = 0; i < pkts; i++)
	pkt_tbl[i] = odp_packet_from_event(ev_tbl[i]);

VS.

odp_packet_from_event_multi(pkt_tlb, ev_tbl, num_pkts);

1) Single call vs. many calls makes difference when not inlined (ABI compat).
2) This is not only cast, it's also a copy of the handle. Implementation may be vectorized, for loop unrolled, etc.
3) Depending on implementation the conversion may include additional things to the handle copy: extra checks, post processing of the event, etc HW/implementation specific tricks. Which would open more optimization opportunity in addition to load+store.
4) Application code is cleaner

include/odp/api/spec/packet.h
psavol: The same answer. Improved implementation efficiency and cleaner application code.

include/odp/api/spec/event.h
psavol: Schedulers do not work that way. You cannot peak into head of schedule queue and ignore non-packets and just select packets. Scheduler returns N events.

Also storing non-packets locally (inside scheduler SW) would not help, since application could easily starve non-packets by polling those too seldom.

platform/linux-generic/odp_packet.c
psavol: Yes.

muvarov:
Merged.

PR #13, opened 2017-04-10
[PATCH API-NEXT v2] IPsec api updates

lumag:
Few small IPsec API updates living in my repo.

PR #321, opened 2017-12-03
[PATCH v4] enhance DPDK integration

lumag:
- support DPDK 16.11
- support DPDK on non-x86 platforms
- support building using system DPDK

### DIFF
--- a/platform/linux-generic/m4/odp_dpdk.m4
+++ b/platform/linux-generic/m4/odp_dpdk.m4
@@ -3,9 +3,10 @@
 ##########################################################################
 pktio_dpdk_support=no
 AC_ARG_WITH([dpdk-path],
-AS_HELP_STRING([--with-dpdk-path=DIR   path to dpdk build directory]),
+[AS_HELP_STRING([--with-dpdk-path=DIR], [path to dpdk build directory])],
     [DPDK_PATH="$withval"
     DPDK_CPPFLAGS="-msse4.2 -isystem $DPDK_PATH/include"
+    DPDK_LDFLAGS="-L$DPDK_PATH/lib"
     pktio_dpdk_support=yes],[])
 
 ##########################################################################
@@ -13,16 +14,10 @@ AS_HELP_STRING([--with-dpdk-path=DIR   path to dpdk build directory]),
 ##########################################################################
 zero_copy=0
 AC_ARG_ENABLE([dpdk-zero-copy],
-    [  --enable-dpdk-zero-copy  enable experimental zero-copy DPDK pktio mode],
+    [AS_HELP_STRING([--enable-dpdk-zero-copy], [enable experimental zero-copy DPDK pktio mode])],
     [if test x$enableval = xyes; then
         zero_copy=1
     fi])
-
-##########################################################################
-# Save and set temporary compilation flags
-##########################################################################
-OLD_CPPFLAGS="$CPPFLAGS"
-CPPFLAGS="$DPDK_CPPFLAGS $CPPFLAGS"
 
 ##########################################################################
 # Check for DPDK availability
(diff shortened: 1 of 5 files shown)


matiaselo:
Several patches are missing proper commit logs.

platform/linux-generic/m4/odp_dpdk.m4
```diff
@@ -5,9 +5,21 @@ pktio_dpdk_support=no
 AC_ARG_WITH([dpdk-path],
 [AS_HELP_STRING([--with-dpdk-path=DIR], [path to dpdk build directory])],
     [DPDK_PATH="$withval"
+     pktio_dpdk_support=yes],[])
+
+if test "x$DPDK_PATH" = "xsystem" ; then
```
matiaselo: If I understood this correctly, a user would have to use '--with-dpdk-path=system' configure option for this to work. Wouldn't a more standard way of doing this be to add e.g. '--enable-dpdk-support' flag in the same manner as with netmap pktio.

matiaselo:
The commit subject line 'linux-gen: pktio: support using DPDK lt 17.08' could be more clear. At least I understood it first that the patch adds support for long term support dpdk 17.08 (which is not true).

platform/linux-generic/pktio/dpdk.c
```diff
@@ -123,8 +123,8 @@ static unsigned cache_size(uint32_t num)
 static inline uint16_t mbuf_data_off(struct rte_mbuf *mbuf,
 				     odp_packet_hdr_t *pkt_hdr)
 {
-	return (uint64_t)pkt_hdr->buf_hdr.seg[0].data -
-			(uint64_t)mbuf->buf_addr;
+	return (unsigned long)pkt_hdr->buf_hdr.seg[0].data -
+			(unsigned long)mbuf->buf_addr;
 }
```
matiaselo: uintptr_t would be a more suitable type here.

matiaselo:
Which distros and versions did you use to validate that the support for system-wide dpdk is working?

platform/linux-generic/m4/odp_dpdk.m4
lumag: Good question. For more or less typical case (non-system DPDK) this would require adding two flags (`--enable-dpdk-support --with-dpdk-path=PATH_TO_DPDK`).

platform/linux-generic/pktio/dpdk.c
lumag: ack

lumag:
@matiaselo I targeted DPDK 16.11 from Debian testing. Only lightly tested at this moment, sorry.

platform/linux-generic/m4/odp_dpdk.m4
matiaselo: I would suggest going forward with this approach (two separate flags).

If the user only gives '--with-dpdk-path= PATH_TO_DPDK' flag, you can presumably handle this in the configure script (so that only one flag has to be used).

lumag:
@matiaselo updated PR:
- provide detailed commit messages
- add `--enable-pktio-dpdk` switch
- added `checking ...` message when trying to compile test code with `-ldpdk`

platform/linux-generic/pktio/dpdk.c
```diff
@@ -29,14 +29,27 @@
...
 #include <rte_ip_frag.h>
+#include <rte_log.h>
 #include <rte_udp.h>
 #include <rte_tcp.h>
 #include <rte_string_fns.h>
+#include <rte_version.h>
+
+#if RTE_VERSION < RTE_VERSION_NUM(17, 8, 0, 0)
+#define rte_log_set_global_level rte_set_log_level
+#endif
```
matiaselo: This breaks build on Ubuntu 17.10, which uses DPDK 17.05: 'error: ‘rte_set_log_level’ is deprecated' 

So the correct version to use here is RTE_VERSION_NUM(17, 5, 0, 0)

platform/linux-generic/m4/odp_dpdk.m4
```diff
@@ -2,12 +2,27 @@
 # Enable DPDK support
 ##########################################################################
 pktio_dpdk_support=no
+
+AC_ARG_ENABLE([pktio-dpdk],
+	      [AS_HELP_STRING([--enable-pktio-dpdk], [enable DPDK support for Packet I/O])],
```
matiaselo: I would prefer making naming uniform with netmap pktio (e.g. '--enable-netmap' & '--enable-dpdk'). The 'pktio-' prefix can be dropped. This way if we choose to use other DPDK components later on (e.g. crypto) we don't have to rename the flag.

lumag:
@matiaselo updated

matiaselo:
Checkpatch still throws one warning:

`WARNING: Possible unwrapped commit description (prefer a maximum 75 chars per line)
#13: 
pktio/dpdk.c:126:9: error: cast from pointer to integer of different size [-Werror=pointer-to-int-cast]

total: 0 errors, 1 warnings, 0 checks, 10 lines checked

../outgoing/0005-linux-gen-dpdk-cast-addresses-to-uintptr_t-rather-th.patch has style problems, please review.`

Otherwise, Reviewed-and-tested-by: Matias Elo <matias.elo@nokia.com>

lumag:
@matiaselo did not want to change origin commit message. Should be fixed now.

muvarov:
Merged.

PR #322, opened 2017-12-04
[PATCH API-NEXT v1] api: remove doxygen param in tags

psavol:
Replace param[in] tags with param tag. All parameters are input only by default. Output 
parameters are better highlighted, when only those are tagged.


muvarov:
Merged.

PR #328, opened 2017-12-08
[PATCH API-NEXT v1] api: ipsec: require IPv4 or IPv6 flag to be set

lumag:
Require IPv4/IPv6 flag to be set for IPsec processing. Hardware usually
requires this, so require application to set flag that it knows.

### DIFF
--- a/include/odp/api/spec/ipsec.h
+++ b/include/odp/api/spec/ipsec.h
@@ -1207,6 +1207,7 @@ typedef struct odp_ipsec_status_t {
  *
  * Each input packet must have a valid value for these metadata (other metadata
  * is ignored):
+ * - IPv4 or IPv6 flag: Set packet flag according to packet contents.
  * - L3 offset: Offset to the first byte of the (outmost) IP header
  * - L4 offset: When udp_encap is enabled, offset to the first byte of the
  *              encapsulating UDP header
@@ -1231,6 +1232,7 @@ typedef struct odp_ipsec_status_t {
  * Each successfully transformed packet has a valid value for these metadata
  * regardless of the inner packet parse configuration
  * (odp_ipsec_inbound_config_t):
+ * - IPv4 or IPv6 flag: Set packet flag according to packet contents.
  * - L3 offset: Offset to the first byte of the (outmost) IP header
  * - pktio:     For inline IPSEC processed packets, original packet input
  *              interface
@@ -1277,6 +1279,7 @@ int odp_ipsec_in(const odp_packet_t pkt_in[], int num_in,
  *
  * Each input packet must have a valid value for these metadata (other metadata
  * is ignored):
+ * - IPv4 or IPv6 flag: Set packet flag according to packet contents.
  * - L3 offset: Offset to the first byte of the (outmost) IP header
  * - L4 offset: Offset to the L4 header if L4 checksum offload is requested
  *
@@ -1294,6 +1297,7 @@ int odp_ipsec_in(const odp_packet_t pkt_in[], int num_in,
  * and content of packet data before the IP header is undefined.
  *
  * Each successfully transformed packet has a valid value for these metadata:
+ * - IPv4 or IPv6 flag: Set packet flag according to packet contents.
  * - L3 offset: Offset to the first byte of the (outmost) IP header
  *
  * @param          pkt_in   Packets to be processed


Bill-Fischofer-Linaro:
The IP version is part of the SA configuration so I'm not sure why this would be needed, especially given the potential performance implications of requiring an extra `odp_packet_has_ipvx_set()` call.

lumag:
@Bill-Fischofer-Linaro it is not a mandatory part of SA configuration. IP version is a part of inbound lookup params if DST_ADDR lookup is used. Also it is a part of outbound tunnel params for outer headers.

Moreover, in optimized cases such call would be inlined and will cost nothing.

include/odp/api/spec/ipsec.h
```diff
@@ -1207,6 +1207,7 @@ typedef struct odp_ipsec_status_t {
  *
  * Each input packet must have a valid value for these metadata (other metadata
  * is ignored):
+ * - IPv4 or IPv6 flag: Set packet flag according to packet contents.
```
lumag: @psavol @JannePeltonen @muvarov @NikhilA-Linaro 
In most cases such flag will be set by inbound packet parser (or known by the application). Otherwise such packet will not be received by inbound API. Well, unless an application submits all inbound packets to IPsec, which seems quite strange to me.

include/odp/api/spec/ipsec.h
```diff
@@ -1231,6 +1232,7 @@ typedef struct odp_ipsec_status_t {
  * Each successfully transformed packet has a valid value for these metadata
  * regardless of the inner packet parse configuration
  * (odp_ipsec_inbound_config_t):
+ * - IPv4 or IPv6 flag: Set packet flag according to packet contents.
```
lumag: @psavol @JannePeltonen @muvarov @NikhilA-Linaro 
In most cases application will already have knowledge about payload protocol. So it is easier to provide such knowledge to IPsec machinery. So this is basically a question to @NikhilA-Linaro, If I got it right from SEC manual, your hardware also requires providing IP version, does it not?

include/odp/api/spec/ipsec.h
```diff
@@ -1277,6 +1279,7 @@ int odp_ipsec_in(const odp_packet_t pkt_in[], int num_in,
  *
  * Each input packet must have a valid value for these metadata (other metadata
  * is ignored):
+ * - IPv4 or IPv6 flag: Set packet flag according to packet contents.
```
lumag: @psavol @JannePeltonen @muvarov @NikhilA-Linaro
In most cases application will already have knowledge about payload protocol. So it is easier to provide such knowledge to IPsec machinery. So this is basically a question to @NikhilA-Linaro, If I got it right from SEC manual, your hardware also requires providing IP version, does it not?

include/odp/api/spec/ipsec.h
psavol: There would be significant performance difference of using this flag vs reading the version from packet data, if any SW (application or implementation) would not need otherwise to read/write the IP header at all. If SW reads/writes it for any other reason during packet processing/output - e.g. application reads/writes IP addresses, implementation insert pseudo header checksum to L4 (as Intel NICs require), etc ... the benefit would diminish as the data is already in CPU cache.

I'd expect that in common case an IPSEC application reads/writes the IP header, instead of blindly forwarding the packet.

Also, IPSEC and pktout (chksum offload, etc) spec should have the same requirements for the same functionality. So, if IP flag would be added here, it would be needed to be added to pktout as well.

lumag: @psavol ... or if the packet is not stored in main memory. Should we still consider this case?

Back to your example. Performance difference would be less in a case when IP header is in main memory. However it is just easier to check for a flag in implementation code, rather than refetching again IP header again and checking first byte.

As a side question, I was thinking about several `odp_packet_*_ptr()` functions that return pointer to packet data without returning segment length. It is easy to misuse them with segmented packets. We should consider rethinking them after TigerMoth. 

I'm not quite sure about pktout. pktout hardware receives explicit IP version in L2 header (unline IPsec, for which there is no L2 header) and we do not support generating checksums in software, so there is no point in adding requirement for IP flags to pktout. Well, unless there will be support for hardware which does automatic L2 header generation.

include/odp/api/spec/ipsec.h
NikhilA-Linaro: I think it is overhead for the the implementations(NXP) where hardware itself retrieves IP version from the packet. if the implementation needs it, it should read from packet data itself.

lumag:
Closing per the discussion during public call. Implementations needing IP version should examine packet contents on their own.

PR #329, opened 2017-12-08
[PATCH API-NEXT v10] IPsec TFC packets/padding support

lumag:
Support TFC features:
 - TFC padding TX and RX via adding additonal data after packet payload
 - TFC packets RX via `ODP_IPSEC_STATUS` event. TX and RX SYNC is TBD.

### DIFF
--- a/include/odp/api/spec/ipsec.h
+++ b/include/odp/api/spec/ipsec.h
@@ -1206,6 +1206,13 @@ typedef struct odp_ipsec_status_t {
  * restored. The amount and content of packet data before the IP header is
  * undefined.
  *
+ * Additional TFC padding might be present after packet contents. ODP
+ * packet will extend beyond the end of the IP packet it contains or the
+ * resulting IP packet will extend beyond the end of the IP payload protocol
+ * (for tunnel or transport cases respectively). An ODP application can detect
+ * and remove such padding by inspecting the length fields of the relevant
+ * protocol headers in the result packet.
+ *
  * Each successfully transformed packet has a valid value for these metadata
  * regardless of the inner packet parse configuration
  * (odp_ipsec_inbound_config_t):
@@ -1258,9 +1265,11 @@ int odp_ipsec_in(const odp_packet_t pkt_in[], int num_in,
  * - L3 offset: Offset to the first byte of the (outmost) IP header
  * - L4 offset: Offset to the L4 header if L4 checksum offload is requested
  *
- * Additionally, input IP packet length (odp_packet_len() minus
- * odp_packet_l3_offset()) must match values in protocol headers. Otherwise
- * results are undefined.
+ * Additionally, input L3 packet length (odp_packet_len() minus
+ * odp_packet_l3_offset()) must not be smaller than the IP packet lenght
+ * indicated by the IP header. Otherwise results are undefined. If the input L3
+ * packet length is bigger than the IP packet length, the additional packet
+ * data is used as TFC padding.
  *
  * Packets are processed in the input order. Packet order is maintained from
  * input 'pkt' array to output 'pkt' array. Packet order is not guaranteed
@@ -1270,6 +1279,9 @@ int odp_ipsec_in(const odp_packet_t pkt_in[], int num_in,
  * e.g. RFC 4302 and 4303). Resulting packets are well formed IP packets
  * with IPSEC, etc headers constructed according to the standards. The amount
  * and content of packet data before the IP header is undefined.
+ *
+ * Additional TFC padding might be present after packet payload. Such padding
+ * should be filled by application before submitting packet to ODP.
  *
  * Each successfully transformed packet has a valid value for these metadata:
  * - L3 offset: Offset to the first byte of the (outmost) IP header
--- a/include/odp/api/spec/ipsec.h
+++ b/include/odp/api/spec/ipsec.h
@@ -1216,9 +1216,12 @@ typedef struct odp_ipsec_status_t {
  * Each successfully transformed packet has a valid value for these metadata
  * regardless of the inner packet parse configuration
  * (odp_ipsec_inbound_config_t):
- * - L3 offset: Offset to the first byte of the (outmost) IP header
- * - pktio:     For inline IPSEC processed packets, original packet input
- *              interface
+ * - L3 offset:       Offset to the first byte of the (outmost) IP header
+ * - IPv4/IPv6 flags: These flags can be used to determine packet contents.
+ *                    If neither IPv4 nor IPv6 flag are set, received packet is
+ *                    TFC dummy packet and should be dropped by an application
+ * - pktio:           For inline IPSEC processed packets, original packet input
+ *                    interface
  *
  * Other metadata for parse results and error checks depend on configuration
  * (selected parse and error check levels).


include/odp/api/spec/ipsec.h
```diff
@@ -1148,6 +1148,14 @@ typedef enum odp_ipsec_status_id_t {
...
 	ODP_IPSEC_STATUS_SA_DISABLE = 0,
 
+	/** Received dummy packet with NextHeader set to NoNextHeader
+	 *
+	 *  Following status event (odp_ipsec_status_t) fields have valid
+	 *  content, other fields must be ignored:
+	 *  - sa:       The SA that caused the warning
+	 */
+	ODP_IPSEC_STATUS_NONH,
+
```
Bill-Fischofer-Linaro: If this were an `odp_ipsec_packet_result_t` bit rather than a status event all input packets (synchronous, asynchronous, and inline) cases could be handled the same without the TBD. Why use status events for these packets?

include/odp/api/spec/ipsec.h
```diff
@@ -1293,6 +1297,10 @@ int odp_ipsec_in(const odp_packet_t pkt_in[], int num_in,
  * with IPSEC, etc headers constructed according to the standards. The amount
  * and content of packet data before the IP header is undefined.
  *
+ * Additional TFC padding might be present after packet payload for ESP-tunnel
+ * mode. It will be included into encrypted packet. Receiver side will skip
+ * this padding automatically.
+ *
```
Bill-Fischofer-Linaro: What controls the number of such padding bytes? This should be documented.

include/odp/api/spec/ipsec.h
lumag: Because there might be no packet associated with such event. Just information that there was TFC packet

include/odp/api/spec/ipsec.h
lumag: Nothing controls number of padding bytes. It is up to an application to decide.

lumag:
@JannePeltonen sorry, I got disconnected, when we were discussing this PR. It looks like NXP hardware does not support stripping such padding automatically. Should we require implementations to actually clip decrypted  packet according to length from IP header?
Also, regarding statistics. Adding proper stats requires per-thread variables, adding them when asked, etc. I'm thinking about easier approach:
 - Report dummy packets as events in ASYNC/INLINE mode
 - Silently skip them in SYNC mode
 - Add a flag for SA which makes ASYNC/INLINE modes always skip such packets

include/odp/api/spec/ipsec.h
Bill-Fischofer-Linaro: OK, then I would make it clear that these are supplied by the application. For example:
```
The application may choose to insert additional padding bytes after the packet payload (bytes in the packet beyond IP length). These will be included in the encrypted output packet, to be skipped by the receiver.
```
I wouldn't make representations about automatic skipping since  there's no assurance that the receiver is an ODP application so we can't say what some other recipient will do.

Bill-Fischofer-Linaro:
@lumag I think we want to be careful about automatic stripping or packet dropping in the API spec. There are four distinct cases to consider. 

1. For TX packets containing TFC padding, this padding is under application control as documented in this PR. If the application adds padding bytes then they'll be encrypted along with the rest of the payload. 

2. For RX packets containing TFC padding, by symmetry these should also be under application control and be ignored / discarded by the application.

3. For RX dummy packets (NH = 59) it's sufficient for IPsec to simply flag these and let the application deal with them as it chooses. That's why I think simply treating these as `ODP_EVENT_PACKET_IPSEC` subtypes with a result flag makes sense, though introducing a new subtype (e.g., `ODP_EVENT_PACKET_IPSEC_DUMMY`) would be another possibility. This handles all cases (sync, async, and inline) in a uniform manner. 

4. For TX dummy packets, this would seem to require a new API, e.g., `odp_ipsec_out_dummy()` or a modifier in the `odp_ipsec_out_opt_t` bits that control the `out` operation. But this case isn't part of this PR.

In doing some research I also came across [this proposal](http://www.ieee-icnp.org/2006/posters/icnp06-posters-teofili.pdf) that introduces an explicit TFC header to handle padding in a uniform and protocol-independent manner that seems to make a lot of sense. I'm not sure where this stands in terms of platform support, however.

lumag:
@Bill-Fischofer-Linaro 
1. OK
2. OK
3. Not always possible. E.g. some hardware could just silently drop NH=59 packets
4. Yes, but such API is not a part of this PR, as I did not want to add another call at this moment (especially lacking any feedback from @NikhilA-Linaro ).

Bill-Fischofer-Linaro:
@lumag Automatic dropping of inbound dummy packets might be acceptable in inline mode (a stat counter would still be needed) but not in other cases. What does it mean to "drop" a packet passed to `odp_ipsec_in()`?  This is a synchronous call that obviously has to complete with some sort of RC. Similarly if `odp_ipsec_in_enq()` is called the application is owed a completion response (this was Janne's point yesterday).

JannePeltonen:
I pretty much agree with Bill's comments above.

Silently dropping an inbound dummy packet in the async case could be considered (e.g. after Tiger Moth) as an optimization that an application could optionally turn on if it does not care about the completion event in that case. But the default behavior should IMHO be that an event gets generated just like for normal packets. A new bit (or maybe the protocol number) in the op_status structure would identify dummy packets to the application.

Another possible optimization would be to not deliver the payload data of the dummy packet to the application (which would anyway have to drop it without further processing to be RFC compliant) but just the packet event.  That could perhaps allow saving some bus bandwidth in some implementations.

bala-manoharan:
3.  We can have a mechanism in which the Dummy packet is NOT dropped by the implementation but in that specific case, there will not be any authentication or selector checks done on this packet.
This needs to be mentioned specifically. We need to understand the handling in the NXP platform as well

@Bill-Fischofer-Linaro regarding the proposal where TFC padding is added by the implementation I think we can move this discussion after Tigermoth.

lumag:
Dropping packet means there will be no packet-like output for such
inbound packets. ASYNC calls can return completion via IPSEC_STATUS
event. For SYNC calls we might want to add special return code, meaning
'processed N packets, encountered dummy packet now'.

lumag:
@NikhilA-Linaro any confirmation/details/etc on handling of dummy packets on NXP hardware?

bala-manoharan:
I would suggest to change the current proposal so that the implementation sends IPSEC status for dummy packet along with the packet to application and application can free the packet. This simplifies handling of inbound dummy packets in both inline and look aside mode.

lumag:
@bala-manoharan this seems logical.
So the outcome of current discussion:
1. For TigerMoth
   - Application provides TFC padding for outgoing packets
   - Implementation strips TFC padding for inbound packets
   - TFC packets are reported as error packets
2. Future changes/optimizations:
   - TFC packet generation
   - Capability/option to drop TFC packets w/o any event
   - Capability to support automatic generation of TFC padding for outgoing packets

@bala-manoharan @NikhilA-Linaro @psavol @JannePeltonen does this look good from your point of view?

Bill-Fischofer-Linaro:
@lumag Based on @NikhilA-Linaro's description of how NXP HW operates (TFC dummy packets are auto-dropped in inline mode) that isn't sufficient. I had suggested we have some additional `odp_ipsec_capability()` flags to indicate:

1. Whether inline TFC dummy packets are discarded.
2. Whether RX TFC padding bytes are stripped

We're still waiting for confirmation from @NikhilA-Linaro as to how TFC dummy packets are handled when presented to `odp_ipsec_in()` or `odp_ipsec_in_enq()` since these APIs need to present some sort of completion indication to the caller.

bala-manoharan:
@lumag I think this there is some discrepancy we need to be split the handling of TFC dummy packets and support for TFC arbitrary padding. My comments were with respect to TFC dummy packets alone.

Let me summarize,
For Tigermoth TFC Dummy packet handling,
1. Inbound direction --> Implementation sends status as IPSEC_STATUS_DUMMY with packet
2. Outbound direction --> Application gives dummy packet length, Implementation creates and sends dummy packet.

For Tigermoth TFC arbitrary padding support,
1. Inbound direction -- Implementation does NOT strip the TFC arbitrary padding. It just notifies the application of TFC
2. Outbound direction -- Application fills TFC padding bytes and sends information to implementation saying TFC arbitrary bytes have been added in this packet.

IMO, we can discuss optimizations after TigerMoth once we implement the feature.

Bill-Fischofer-Linaro:
@bala-manoharan The behavior you describe is not compatible with how NXP HW behaves, per @NikhilA-Linaro. That's why I suggest we need an `odp_ipsec_capability()` flag to let the application know how these will be handled. Per Nikhil. NXP auto-drops inbound TFC dummy packets and auto-strips TFC padding. There are no controls for this so we can't force alternate behavior.

bala-manoharan:
@Bill-Fischofer-Linaro  I understand and I am perfectly fine with updating the proposal. I have just explained the behaviour from Cavium's point of view and once we have a requirement from Nikhil we can modify and update the proposal.

NikhilA-Linaro:
Let me summarize NXP HW implementation

Dummy packet handling:
    Inbound direction --> Implementation sends status as IPSEC_STATUS_DUMMY and packet already dropped. So the event should be IPSEC_STATUS not IPSEC_PACKET.
    Outbound direction --> Application gives dummy packet, with content and length, implementations encrypts and sends dummy packet with NH=59.

For padding support,

    Inbound direction -- Implementation strips the TFC padding, and sends stripped packet to application.
    Outbound direction -- Application fills TFC padding bytes and sends packet with length field updated (including TFC padding), implementation will encrypt the packet.

P.S. Apologies for the delay. was traveling. Hope it helps.

JannePeltonen:
I do not think NXP HW implementation sends IPSEC_STATUS_DUMMY or any other event defined in the ODP SW API but I suppose the HW implementation sends some HW event, which the NXP ODP SW implementation could map to an IPsec packet event. IOW, I do not see why status event would be ok and packet event not from NXP HW perspective. A packet event would, IMHO, be better for the dummy packets than the status event in the inline case and especially in the async case when the application expects to get a packet event back.

One other reason why packet events for dummy packets look better to me is than an application might want to handle status events in some special manner (prioritized, control plane path, ...) compared to normal packet events. Reporting dummy packets through status events instead of packet events could break this.

Bill-Fischofer-Linaro:
@NikhilA-Linaro I understand the case for inline RX dummy packets, but agree with @JannePeltonen  that the implementation should be able to present these as packet events (same as we handle disable completion notification as status events) as both of these are constructs of the ODP implementation.

For sync and async RX calls, what does the HW do in these cases? Presumably in the case of sync calls this has to be translated into an RC of some sort, so again we have a packet. Async should be no different.

On the outbound side, you say that "application gives dummy packet". How is that indicated? Currently the `odp_ipsec_out()`, `odp_ipsec_out_enq()` and `odp_ipsec_out_inline()` APIs have no mechanism for indicating whether the outbound packet is to be  treated as a dummy or not. We may not include such capability in Tiger Moth, but I'm curious as to how this is indicated in your HW. Within the scope of the current ODP API definition it seems that this would be an `odp_ipsec_out_opt_t` flag.

lumag:
@bala-manoharan @NikhilA-Linaro @psavol @Bill-Fischofer-Linaro 
Updated PR basing on previous discussions. We need feedback from @NikhilA-Linaro and maybe other inetersted parties (Barry(?)).

include/odp/api/spec/ipsec.h
```diff
@@ -259,6 +259,10 @@ typedef struct odp_ipsec_capability_t {
 	 */
 	odp_support_t retain_header;
 
+	/** Implementation will automatically truncate TFC padding in received
+	 *  packets */
+	odp_bool tfc_padding_truncate;
```
Bill-Fischofer-Linaro: Why not `odp_support_t` same as other capabilities here?

include/odp/api/spec/ipsec.h
```diff
@@ -167,6 +167,13 @@ typedef struct odp_ipsec_inbound_config_t {
 	 */
 	odp_proto_chksums_t chksums;
 
+	/** Drop dummy packets (NH=59) instead of reporting it via
+	 *  @dummy_packet error bit.
+	 *
+	 *  @see dummy_packets_drop
+	 */
+	odp_bool_t dummy_packets_drop;
```
Bill-Fischofer-Linaro: Since you're referencing `tfc_padding_truncate`, the TFC prefix would seem consistent here. So `tfc_dummy_packets_drop` and comments should similarly reference TFC.

include/odp/api/spec/ipsec.h
```diff
@@ -263,6 +270,11 @@ typedef struct odp_ipsec_capability_t {
 	 *  packets */
 	odp_bool tfc_padding_truncate;
 
+	/**
+	 * Support dropping dummy packets
+	 */
+	odp_support_t dummy_packets_drop;
```
Bill-Fischofer-Linaro: `tfc_dummy_packets_drop` would be clearer.

include/odp/api/spec/ipsec.h
lumag: Because it's not a declaration of supporting a feature but rather a statement how implementation handles this case. It is not possible to switch it. So I preferred to have odp_bool_t here.

include/odp/api/spec/ipsec.h
lumag: ok, updated.

include/odp/api/spec/ipsec.h
Bill-Fischofer-Linaro: Some implementations may be able to configure this behavior in `odp_ipsec_config()`, in which case `odp_support_t` would be useful. We can make that change later, but it just looks more consistent here. Not a big deal, however.

Bill-Fischofer-Linaro:
Ping to @psavol and @NikhilA-Linaro. Need review input on this.

include/odp/api/spec/ipsec.h
```diff
@@ -146,6 +146,13 @@ typedef struct odp_ipsec_inbound_config_t {
 	 */
 	odp_proto_chksums_t chksums;
 
+	/** Drop TFC dummy packets (NH=59) instead of reporting it via
+	 *  @tfc_dummy_packet error bit.
+	 *
+	 *  @see tfc_dummy_packets_drop
+	 */
+	odp_bool_t tfc_dummy_packets_drop;
```
NikhilA-Linaro: This configuration option is a bit confusing. We need to specifically mention inline and look aside cases. It may not be configurable in many hardwares.

lumag: Hmm. What about `tfc_dummy_packet_drop_sync/async/inline_configurable` capabilities and three flags `tfc_dummy_packet_drop_sync/async/inline`?

psavol: We should minimize the number of configuration options at this time. I think it's not important for application to control dummy packet dropping. So, it can be implementation defined in inline. For sync/async we must defined how many packets can be expected back from each IPSEC operation. The simplest would be that each input packet produces an output packet, and it's implementation specific what the output packet contains (NONH packet or just a warning flag telling that the packet was dropped since it was NONH). Application would not need to control dropping but notice if that happened. Async interface would be problematic for application if it's possible that nothing comes back from an operation (all packets silently dropped).

JannePeltonen: I understood that we wanted the automatic dropping only in inline processing mode and not in other modes.

JannePeltonen: The documentation of odp_ipsec_in() says: "Outputted packets with error status have not been transformed but the original packet is returned."

Maybe it is too much promised for TFC packets (and the application should not need the original packet in that case anyway). It would anyway be good to clarify in the comments how exactly odp_ipsec_in() behaves with TFC packets.

include/odp/api/spec/ipsec.h
```diff
@@ -238,6 +238,10 @@ typedef struct odp_ipsec_capability_t {
 	 */
 	odp_support_t retain_header;
 
+	/** Implementation will automatically truncate TFC padding in received
+	 *  packets */
+	odp_bool_t tfc_padding_truncate;
```
JannePeltonen: Is the tfc_padding_truncate capability even needed? It is an optimization for applications that would benefit from knowing in advance that the padding is always truncated, but is there a good use case for that?

include/odp/api/spec/ipsec.h
```diff
@@ -1206,6 +1210,12 @@ typedef struct odp_ipsec_status_t {
  * restored. The amount and content of packet data before the IP header is
  * undefined.
  *
+ * Additional TFC padding might be present after packet contents for ESP tunnel
```
JannePeltonen: I interpret the "might be present" so that if there was TFC padding in a received packet, it might or might not be present after ODP IPsec processing. Is that the intended interpretation and does it need to be made clearer?

include/odp/api/spec/ipsec.h
```diff
@@ -1206,6 +1210,12 @@ typedef struct odp_ipsec_status_t {
...
+ * mode. Received side can use total (IPv4) or payload (IPv6) length from
```
JannePeltonen: The term "TFC padding" in this API proposal seems to differ from that in the RFC. It appears that here TFC padding refers only to padding after IP payload in tunnel mode.

TFC padding, as defined in the RFC, is not restricted to tunnel mode. Transport mode which carries e.g. UDP or IP or other suitable protocol can have TFC padding but that can not be recognized by ODP in the general case (e.g. with proprietary IP protocols).

JannePeltonen: ... my point being, the description should be improved to make it more accurate and unambiguous.

include/odp/api/spec/ipsec.h
lumag: Updating to specifically mention "inline" mode. We can add drop/etc. settings to sync/async modes later.

include/odp/api/spec/ipsec.h
lumag: @JannePeltonen 
Basically it is there to describe hardware behaviour. NXP SEC truncates packets according to L4 length, our hardware does not.

include/odp/api/spec/ipsec.h
lumag: Well. RX side has no way to know in advance if there will TFC padding or not in the packet. That is why I used "might" here. Could you please suggest better wording?

include/odp/api/spec/ipsec.h
JannePeltonen: Yes, but is it needed? With this, an application could check if the capability was set instead of checking if packet length is bigger than the header indicates for deciding whether to trim padding. I just wonder if that is beneficial enough to justify yet another capability flag.

include/odp/api/spec/ipsec.h
lumag: @psavol @JannePeltonen for async the major issue is the packet metadata/user pointer here. I think we might need to give some thought to this area anyway because of fragmentation handling. We might need additional copy_udata/free_udata function pointers to sa config.

include/odp/api/spec/ipsec.h
JannePeltonen: So you are saying that my interpretation was not the intended one. I think it is important to make it clear whether we mean that there might of might not be padding in the received packet in the first place or whether the implementation might or might not strip said padding when present.

I was thinking it would be good to not make any promises on whether the padding is removed or not, i.e. specify that there may or may not be padding left in the result packet that the application gets. This would work with any implementation, even those that do not always do the same thing for every packet (e.g. depending on the input port).

Not promising anything about the padding would be similar to how the API currently leaves it up to the implementation whether there is any L2 or other bytes in front of the IP header in the ODP packet data.

Anyway, I will not attempt better wording before it is clear what the intended semantics are.

include/odp/api/spec/ipsec.h
lumag: I will update wording to match RFC. Noted that ODP does not touch TFC padding in transport mode.

lumag:
@bala-manoharan @Bill-Fischofer-Linaro   @JannePeltonen @NikhilA-Linaro  @psavol
I've updated PR, please review.

include/odp/api/spec/ipsec.h
lumag: This will allow fast-path forwarding for some applications. If this flag is set, an application can skip header checks/lookups and forward ready packet.

include/odp/api/spec/ipsec.h
lumag: I have to think about it.

include/odp/api/spec/ipsec.h
psavol: I still vote for minimal TFC specific API changes at this point. This capa is not strictly needed. A portable application would need to implement the length checks anyway.

include/odp/api/spec/ipsec.h
```diff
@@ -242,6 +242,11 @@ typedef struct odp_ipsec_capability_t {
 	 *  packets in ESP tunnel mode */
 	odp_bool_t tfc_padding_truncate;
 
+	/** In INLINE mode drop TFC dummy packets (NH=59) instead of reporting
+	 *  them via tfc_dummy_packet error bit.
+	 */
+	odp_bool_t tfc_dummy_packets_drop;
```
psavol: Also this capa is not strictly needed. A portable application will need to handle also the case where inline input does not drop NH packets.

include/odp/api/spec/ipsec.h
```diff
@@ -902,6 +907,9 @@ typedef struct odp_ipsec_error_t {
 
 			/** Hard lifetime expired: packets */
 			uint32_t hard_exp_packets : 1;
+
+			/** TFC dummy packet (NextHeader = 59) */
+			uint32_t tfc_dummy_packet : 1;
```
psavol: I suggest that when implementation does not drop "no next header" packets, those come through as any other successfully transformed packets: no error/warning bits set. However, when HW has dropped a no NH packet it generates a dummy packet (length zero, etc) with a warning or operation flag bit on telling that this is not a normal packet, but an indication that a no NH packet was dropped.

Errors/warnings indicate a failure and may require some specific action (SA negotiation). This instead could be could be just a odp_ipsec_op_flag_t bit, as the operation was successful and does not need specific action (other than drop packet after finding out that it does not contain any data). So, application could ignore the bit and drop packet as part of normal processing of too short packets.

lumag: They can not come as "normal" packets. An application will have no way to determine that the package was dummy after receiving it. NH field will be gone in case of tunnel packets.

include/odp/api/spec/ipsec.h
lumag: Fine, I'll drop this and add a phrase that an implementation might have removed the padding.

include/odp/api/spec/ipsec.h
lumag: True.

include/odp/api/spec/ipsec.h
lumag: I prefer to have this bit as an error to keep very simple constraint. No error => app may forward packet (if it is not interested in details). Error => packet must be threated separately and must not be forwarded.

include/odp/api/spec/ipsec.h
```diff
@@ -1206,6 +1210,14 @@ typedef struct odp_ipsec_status_t {
...
  * undefined.
  *
+ * Additional TFC padding might be present after packet contents. For ESP
+ * transport mode ODP does not truncate such padding, it up to an application
+ * to detect and drop it. For ESP tunnel mode, received side can use total
+ * (IPv4) or payload (IPv6) length from internal headers to drop such TFC
+ * padding. If tfc_padding_truncate capability is set, implementation will
+ * truncate received packets automatically. Otherwise ODP application has to
+ * truncate packets manually.
+ *
```
JannePeltonen: I suggest rewording like this (after dropping tfc_padding_truncate capa):

An inbound ESP packet may contain TFC padding. odp_ipsec_in() might or might not remove the padding. When TFC padding is not removed, the resulting ODP packet will extend beyond the end of the IP packet it contains or the resulting IP packet will extend beyond the end of the IP payload protocol. An ODP application can detect and remove such padding by inspecting the length fields of the relevant protocol headers in the result packet.

include/odp/api/spec/ipsec.h
JannePeltonen: This is in conflict with this existing statement a little earlier:

"Additionally, input IP packet length (odp_packet_len() minus odp_packet_l3_offset()) must match values in protocol headers. Otherwise esults are undefined."

It could be reworded like this:

"Additionally, input L3 packet length (odp_packet_len() minus odp_packet_l3_offset()) must not be smaller than the IP packet lenght indicated by the IP header. Otherwise results are undefined. If the input L3 packet length is bigger than the IP packet length, the additional packet data is used as TFC padding."

And I think that would be sufficient about TFC here.

include/odp/api/spec/ipsec.h
JannePeltonen: A dummy packet in transport mode could be delivered to an application in the normal way. The result packet would be an IP packet with protocol 59. But it is true that in tunnel mode there is no way to deliver the packet to the application in the current API. In my interpretation of the current API a tunnel mode dummy packet would result in a proto error (since the decapsulated packet is neither IPv4 nor IPv6).

So some warning/error/flag bit is indeed needed if one does not want to confuse dummy packets with protocol errors.

JannePeltonen: Currently the API promises to give the original packet back to the application in case of any error. This would apply to dummy packets too. Maybe the API should be so that such preservation of the original packet is not required in the case of dummy packets to reduce the overhead of handling dummy packets.

Reporting dummy packet through an error is a bit misleading since no error happened. An application cannot e.g. have a global IPsec error counter without first filtering out dummy packets.

psavol: Maybe a generic solution would be to add the IPSEC next header field value into odp_ipsec_packet_result_t. I guess all implementations should be able to return that after decrypt as RFCs mention also TCP as one possible value. So, payload could be at least IPv4, IPv6, TCP or NONH.

" The value of this
   field is chosen from the set of IP Protocol Numbers defined on the
   web page of Internet Assigned Numbers Authority (IANA).  For example,
   a value of 4 indicates IPv4, a value of 41 indicates IPv6, and a
   value of 6 indicates TCP."

JannePeltonen: I do not think it is necessary. The next header field can be anything (like TCP) in transport mode, but in transport mode this API delivers the application and IP packet and inserts the protocol value from the next header field in the protocol field of the IP header. For tunnel mode SAs other next header values than those for IPv4, IPv6 and dummy packets should not appear.

lumag: The issue with transport mode ESP and NONH is that we should never forward NOHN packets even in transport mode. So both transport and tunnel NONH packets should raise same exceptions.

JannePeltonen: I was just pointing out that from functional completeness perspective API changes are not strictly needed. But as was discussed in the Wednesday meeting, having the next header field in the packet result could be useful and convenient anyway.

bala-manoharan: In case if we update the warning bit when the dummy packet is found then there will not be much impact on the application side since odp_ipsec_status_t consists both error and warning bits. Updating next proto field in ipsec_op_packet_result_t requires implementation to process and update this field for every packet.

The only API change required is to specify clearly that in-case of a dummy packet the implementation will not do any selector checks on the received packets.

include/odp/api/spec/ipsec.h
bala-manoharan: We added this capability mainly coz NXP were removing the padding and Cavium was not removing the padding. If from the application point of view you are fine with the wording you have suggested above then this capability can be removed.

include/odp/api/spec/ipsec.h
JannePeltonen: What selector checks are you referring to? Traffic selectors are handled fully in the application side. The ODP implementation does not know the traffic selectors and cannot do any checks anyway.

bala-manoharan: I was referring to ODP_IPSEC_LOOKUP_DSTADDR_SPI. This cannot be matched in case of NH   = 59.  Actually, this raises another question. Do you really need the SA from which this Dummy packet was received in the application? Like do we have to fill the SA value in the odp_ipsec_packet_result_t field?

JannePeltonen: ODP_IPSEC_LOOKUP_DSTADDR_SPI means that for an SA to match a received packet, both the SPI and the destination IP address of the received ESP/AH packet must match those configured in the SA. This lookup must happen before IPsec processing (decryption), i.e. before it is even possible to know that the packet is actually a dummy TFC packet. You have to find the correct SA and perform decryption before you can tell whether a received packet is a genuine data packet or a dummy TFC packet.

An application expects the used SA to be set in the packet result and, IMHO, dummy packets should not be any different since they got processed using the SA.

bala-manoharan: Hmm... I misread the specification so LOOKUP_DSTADDR_SPI only uses ip addr and proto field for lookup and does not validate the src/dst ip address or proto filed after decryption. I believe this validation of IP addr and version can be done by the HW and could be something to add as an enhancement to the spec in the future.

Yes, we could add this SA to packet result but since the RFC specifies in the implementation note to discard the dummy packet after the reception I was inclined towards not updating these structures for dummy packets.

What is your opinion on updating the next protocol field in odp_packet_result vs adding a warning bit for dummy packet?

lumag: We might add next_proto field to packet_result. @NikhilA-Linaro how hard would it be to get this field on your platform?

I'm against using warning bit for dummy packets. From ODP packet flow point of view, dummy packet is clearly an errorneous packet.

psavol: From IPSEC operation point of view a successfully decrypted dummy packet is a successful operation (SA matched, hash was OK, decrypt was OK, protocol headers were OK, no replay issues, etc). Just the payload format is different than usual. If TFC is really used to hide traffic pattern, all spare capacity of the connection could be filled by dummy packets. So, it would be more common for application to receive these vs. the normal packet (as long as HW is not dropping dummies).

So, I'd like to reserve error/warning bits for the cases when there's really a problem (SA configuration miss-match, attack, broken sender implementation, etc).

lumag: @psavol I agree with your arguments. However I'm worried about fast-forwarding use case. Do you think it will make sense to make tfc-packet warning (or even just flag) but still require that `odp_packet_has_error()` returns true for such packets?

psavol: Application needs to anyway check if
 * packet/ipsec operation result has errors
 * IP version of packet (== what's in the tunnel)

If we add result.next_header field, application can use that for both checking IP version and NONH in one switch case. There's no extra check for NONH.

    if (result.status != 0)
        goto error;

    switch (result.next_header) {
        case IPV4:
            ....
        case IPV6:
            ....
        case NONH:
            odp_packet_free(pkt);
            break;
        default:
            goto error;

lumag: @psavol Yes, but it requires to call `odp_ipsec_result()` (which can be quite heavyweight).  Currently it is possible to forward packets right after calling `odp_packet_has_error()` and `odp_packet_has_ipv4()`, which can be inlined to just bit checks.

Anyway, this is now a question to @bala-manoharan and @NikhilA-Linaro if it would be easy to add next_header field to `odp_ipsec_packet_result_t`

include/odp/api/spec/ipsec.h
Bill-Fischofer-Linaro: This is useful information for the application to know about how the implementation will behave. Applications are free to ignore this if they wish, but I see no harm in keeping it.

include/odp/api/spec/ipsec.h
Bill-Fischofer-Linaro: @bala-manoharan @NikhilA-Linaro Please provide input here.

include/odp/api/spec/ipsec.h
JannePeltonen: > I see no harm in keeping it

I understand you refer to the tfc_padding_truncate capability. Note that the proposal to get rid of it and discussion on that (with comments from Petri too) are located elsewhere in this tool.

That said, the harm of keeping the capability (in addition to growing the API) is that the capability makes a promise that may (depending on what value zero means) unnecessarily restrict implementations where the fate of the padding is not the same globally in the whole system and at all times (e.g. NIC based IPsec offload where padding is removed or not depending no through which NIC the packet came in, or a system where removal of the padding depends somehow on the packet).

If the capability is kept, its meaning needs to be documented better, including specifying what kinds of TFC padding it applies to (see other comments regarding that).

Bill-Fischofer-Linaro:
@lumag Please rebase to resolve recent forced update conflicts with api-next

lumag:
@psavol @JannePeltonen 
I have updated PR:
 - documented truncating TFC padding in application
 - kept tfc_padding_truncate capability as a hint to application whether to use ESP_TFC_PADDING_NOT_SUPPORTED during key exchange.
 - make ipv4/ipv6 flags mandatory for packets after IPsec transformation. Document TFC packets as packets having both ipv4 and ipv6 flags clear.

include/odp/api/spec/ipsec.h
Bill-Fischofer-Linaro: OK. Agree with this approach for Tiger Moth.

include/odp/api/spec/ipsec.h
Bill-Fischofer-Linaro: @JannePeltonen I think @lumag's latest wording is concise. Any application using TFC should be familiar with it's requirements and we shouldn't have to repeat the RFC here.

include/odp/api/spec/ipsec.h
```diff
@@ -232,6 +232,12 @@ typedef struct odp_ipsec_capability_t {
 	 */
 	odp_support_t pipeline_cls;
 
+	/** Implementation will automatically truncate TFC padding in received
+	 *  packets in ESP tunnel mode. Application can use this capability to
+	 *  determine necessity for ESP_TFC_PADDING_NOT_SUPPORTED notification.
+	 */
+	odp_support_t tfc_padding_truncate;
```
Bill-Fischofer-Linaro: I thought we discussed removing these capabilities from Tiger Moth?

include/odp/api/spec/ipsec.h
```diff
@@ -1223,9 +1223,12 @@ typedef struct odp_ipsec_status_t {
...
  * (odp_ipsec_inbound_config_t):
- * - L3 offset: Offset to the first byte of the (outmost) IP header
- * - pktio:     For inline IPSEC processed packets, original packet input
- *              interface
+ * - L3 offset:       Offset to the first byte of the (outmost) IP header
+ * - IPv4/IPv6 flags: These flags can be used to determine packet contents.
+ *                    If neither IPv4 nor IPv6 flag are set, received packet is
+ *                    TFC dummy packet and should be dropped by an application
+ * - pktio:           For inline IPSEC processed packets, original packet input
+ *                    interface
```
Bill-Fischofer-Linaro: Is this not equally applicable to `odp_ipsec_in_enq()`? And what about packets received inline? Are we assuming that TFC dummy packets are dropped by the inline RX path? That should also be documented if yes. If not then application must be prepared to receive them that way as well.

include/odp/api/spec/ipsec.h
lumag: I moved it to a separate comment, because of `ESP_TFC_PADDING_NOT_SUPPORTED` notification. I can still drop it.

include/odp/api/spec/ipsec.h
lumag: `odp_ipsec_in_enq()` just refers to `odp_ipsec_in()` function. As for inline processing, an application should be prepared to receive them, because some hardware can not drop them.

lumag:
- dropped last capability
- updated wording according to @JannePeltonen's suggestion

include/odp/api/spec/ipsec.h
JannePeltonen: The reference to IKEv2 ESP_TFC_PADDING_NOT_SUPPORTED notification should be removed. IKE protocol details do not belong to this API spec which should just describe the API. One can read the IKE RFC for IKE details or one might not even use IKE but something else.

Even when IKE is used, the remark is inaccurate. Even if we send the notification for some SA, it does not guarantee that the peer does not send us TFC padding but just declares that we do not accept such packets if we receive them.

include/odp/api/spec/ipsec.h
JannePeltonen: @Bill-Fischofer-Linaro I did not add the reference to the RFC.
@lumag Same comment also here about dropping the remark about the ESP_TFC_PADDING_NOT_SUPPORTED notification. Actually, I think the whole last added paragraph could be just removed.

lumag:
Closing in favour of #403

muvarov:
@lumag did you forgot to click "close" ?

lumag:
@muvarov yes

PR #331, opened 2017-12-08
[PATCH API-NEXT v2] api: packet: add API to convert user area ptr to odp_packet_t

semihalf-mazur-michal:
This new API function will take a pointer to user area and calculate
pointer to ODP buffer's header associated with it.
It is needed when application wants to send buffers back to ODP and
has only a pointer to its private type. Some implementation (like
ODP-DPDK) can implement this API using subtraction of offset to user
area relative to the beginning of the ODP header. Storing an ODP
pointer for every packet in application is not efficient.

Signed-off-by: Michal Mazur <mkm@semihalf.com>


include/odp/api/spec/packet.h
```diff
@@ -1333,6 +1333,17 @@ void odp_packet_user_ptr_set(odp_packet_t pkt, const void *ctx);
  */
 void *odp_packet_user_area(odp_packet_t pkt);
 
+/**
+ * Packet header from user area pointer
```
Bill-Fischofer-Linaro: Packet handle, not header

include/odp/api/spec/packet.h
```diff
@@ -1333,6 +1333,17 @@ void odp_packet_user_ptr_set(odp_packet_t pkt, const void *ctx);
...
+ *
+ * Return address of packet header associated with user area.
```
Bill-Fischofer-Linaro: Return handle of packet...

include/odp/api/spec/packet.h
```diff
@@ -1333,6 +1333,17 @@ void odp_packet_user_ptr_set(odp_packet_t pkt, const void *ctx);
...
+ *
+ * @param uarea  User area address
+ *
+ * @return       Packet header address
```
Bill-Fischofer-Linaro: Packet handle

psavol:
This function cannot be added to API, since it easily breaks portability / implementation freedom.

odp_packet_t is a handle to a packet. When application receives a packet handle, it has ownership of the packet. It can request user area pointer of the packet using the handle. When application gives up the ownership of the packet (e.g. enqueue the packet), it also must stop using the user area pointer. When application next time receives ownership (e.g. dequeue) for the same packet, the handle may be different and also the user area pointer may be different. It's implementation defined where packet and user area data are stored, and how a handle is composed. The packet and user area contents are the same - handle and pointer may be different. So, in general application must not store user area pointer and expect that the same pointer can be used again as a synonym for the packet (after certain API calls).

However, when application does not use API calls that change ownership it can store packet handle to user context data and use it later on.

The problem to be solved is data structure layout optimization in VPP and/or odp-dpdk implementation. API change is not needed for that.

semihalf-mazur-michal:
> odp_packet_t is a handle to a packet. When application receives a packet handle, it has ownership of the packet. It can request user area pointer of the packet using the handle. When application gives up the ownership of the packet (e.g. enqueue the packet), it also must stop using the user area pointer. 

I agree, but it is not related to this problem. Application (VPP) gets new packets through input node implemented in ODP4VPP plugin and pointer to user area is obtained using odp_packet_user_area every time.
After packet is received by ODP4VPP input node, pointer to user area is passed to VPP. VPP knows nothing about ODP. After processing of packets is finished they are pushed to output node.
Output node gets only pointer to user area. ODP handle can be stored there but it requires a memory read for every packet and is not efficient. On odp-dpdk it can be done much faster using simple calculation.
With this API each implementation could optimize process of finding ODP handle and application could choose if it is faster to keep pointer in memory or use API.

> It's implementation defined where packet and user area data are stored

Yes, and in my opinion it is bad because it imposes additional requirement on application. 
For performance reasons, VPP uses 4 byte buffer index instead of a pointer which identifies whole packet (header and data located after it). To support every ODP implementation it would have to store two indexes. Because of this, ODP4VPP is not compatible with linux-generic implementation.

> So, in general application must not store user area pointer and expect that the same pointer can be used again as a synonym for the packet (after certain API calls).

ODP4VPP plugin uses receive/dequeue/schedule API only in input nodes and enqueue/send API only in output node. Outside this nodes VPP knows nothing about ODP and never uses user area or ODP handle after packet is sent to ODP.

> However, when application does not use API calls that change ownership it can store packet handle to user context data and use it later on.
> The problem to be solved is data structure layout optimization in VPP and/or odp-dpdk implementation. API change is not needed for that.

VPP keeps all required data in single cacheline, adding an ODP handle to VPP header will require to read another cacheline for every packet. It adds over 10 million cacheline fetches per second at 10G link.

Bill-Fischofer-Linaro:
On Fri, Dec 8, 2017 at 9:32 AM, semihalf-mazur-michal <
notifications@github.com> wrote:

> odp_packet_t is a handle to a packet. When application receives a packet
> handle, it has ownership of the packet. It can request user area pointer of
> the packet using the handle. When application gives up the ownership of the
> packet (e.g. enqueue the packet), it also must stop using the user area
> pointer.
>
> I agree, but it is not related to this problem. Application (VPP) gets new
> packets through input node implemented in ODP4VPP plugin and pointer to
> user area is obtained using odp_packet_user_area every time.
> After packet is received by ODP4VPP input node, pointer to user area is
> passed to VPP. VPP knows nothing about ODP. After processing of packets is
> finished they are pushed to output node.
> Output node gets only pointer to user area. ODP handle can be stored there
> but it requires a memory read for every packet and is not efficient. On
> odp-dpdk it can be done much faster using simple calculation.
>
The reason it can "be done faster" is because VPP is exploiting DPDK
internals. ODP by design has no user-visible internals precisely to allow
HW innovation. When you force HW to mimic SW structures the SW "wins" but
actually loses because HW can't do all it could otherwise do to accelerate
application-relevant functions rather than have SW manipulate data
structures to achieve the same function. So rather than SW dictating how
data should be organized so that SW can do the work based on static HW
assumptions, the ODP philosophy is simply to have SW describe the packet
transforms it wishes to effect and let the ODP implementation realize those
transforms in a platform-optimal manner.


> With this API each implementation could optimize process of finding ODP
> handle and application could choose if it is faster to keep pointer in
> memory or use API.
>
The same could be said for anything else held in the VLIB buffer. By that
reasoning every application that has a data structure it references might
want ODP to provide APIs to access fields in that data structure rather
than simply accessing them itself.


> It's implementation defined where packet and user area data are stored
>
> Yes, and in my opinion it is bad because it imposes additional requirement
> on application.
> For performance reasons, VPP uses 4 byte buffer index instead of a pointer
> which identifies whole packet (header and data located after it). To
> support every ODP implementation it would have to store two indexes.
> Because of this, ODP4VPP is not compatible with linux-generic
> implementation.
>
No, what we're saying here is that applications can't determine HW
organization without imposing constraints that themselves limit performance.

> So, in general application must not store user area pointer and expect
> that the same pointer can be used again as a synonym for the packet (after
> certain API calls).
>
> ODP4VPP plugin uses receive/dequeue/schedule API only in input nodes and
> enqueue/send API only in output node. Outside this nodes VPP knows nothing
> about ODP and never uses user area or ODP handle after packet is sent to
> ODP.
>
So how are packets received via ODP4VPP examined / manipulated by VPP
nodes? The VLIB buffer clearly isn't a copy of the packet so these nodes
need to access the packet itself, presumably. That requires access to the
odp_packet_t because the only way to access packet contents is via its
handle using ODP APIs.


> However, when application does not use API calls that change ownership it
> can store packet handle to user context data and use it later on.
> The problem to be solved is data structure layout optimization in VPP
> and/or odp-dpdk implementation. API change is not needed for that.
>
> VPP keeps all required data in single cacheline, adding an ODP handle to
> VPP header will require to read another cacheline for every packet. It adds
> over 10 million cacheline fetches per second at 10G link.
>
That's what prefetching is for. The packet itself likely consists of more
than a single cache line so clearly you're referencing more than one cache
line anyway. Also, has this difference been measured on Arm systems or only
x86? High-end x86 systems can be unbalanced to the point where they become
very brittle in this area. This is not a virtue and is the reason Intel
keeps adding application cache-management instructions as a mitigation.
Presumably VPP is already using these on x86?

lumag:
I'd suggest to just add necessary data to user area for the particular application (VPP), rather than clobber API with one application usecase.

include/odp/api/spec/packet.h
semihalf-mazur-michal: corrected in v2

include/odp/api/spec/packet.h
semihalf-mazur-michal: corrected in v2

include/odp/api/spec/packet.h
semihalf-mazur-michal: corrected in v2

nagarahalli:
This is about performance. So far the tests indicate, performance is better with the API.

sachin-saxena:
For a given following ODP usecase:

       -------------------------------
         Upper Layer (ODP unaware)
       -------------------------------
        ODP aware layer (odp API user)
       -------------------------------
         ODP Implementation
       -------------------------------

A given ODP API implementation can support following buffer format:
        --------------------------------------------------    
            ODP_mbuf | User Area | Headroom | Data
        --------------------------------------------------

Without any API change, It can also support following:
        --------------------------------------------------    
          ODP_mbuf | XXX | User Area | Headroom | Data
        --------------------------------------------------

Scenario:
Now in run to completion mode, If an ODP user (may be Wrapper layer/plugin) further passes this packet to **an upper Layer (Independent of ODP)** via this "User Area" pointer and that layer does all packet processing and return packet to ODP user layer via same User area pointer, then
**How the ODP user layer can retrieve ODP_mbuf handle from given user area pointer**

Solutions:

1. ODP user/user-layer , must maintain an odp_packet_t <-->uarea mapping and waste intensive cpu cycles in searching and retrieving the value per packet

2. The ODP independent upper layer must provide a mechanism to store ODP specific handle in their own User Area. In my opinion, this is not the acceptable case always.

3. We have an ODP API, which can return ODP mbuf handle by taking user area pointer as input. Each platform is aware of the packet format it is going to use so, no issue in calculating odp_packet_t in any of the above packet format scenario.

We are suggesting 3rd option.

Bill-Fischofer-Linaro:
@sachin-saxena Thanks for the detail. Here are some of the concerns.

1. The "maintenance" required is simply to store the `odp_packet_t` in the user area where it can be retrieved with a single load instruction, independent of the ODP implementation. Had it not been for the fact that for this particular case that spilled over into an extra cache line, you clearly would have gone that route since it is obviously the most portable and efficient means of doing this. As @psavol pointed out, data structures expand all the time for various reasons and applications have to deal with the effect of this sort of growth on a business-as-usual basis. So there's nothing special about this growth vs. any other application-driven growth to justify adding a new API.

2. Even with the additional cache line, I don't see why prefetching cannot be used to solve this problem. Can we explore that further? x86 is designed to prefetch sequential cache lines efficiently. Are you seeing the same performance drop on Arm? 

3. You're assuming that `odp_packet_from_user_area()` can be an inline function that uses some simple arithmetic. This is only the case if you are compiling with `--enable-abi-compat=no`. Given the different potential packet implementations (even in your own example shown) I don't see how a general inline functional model can be defined within an ABI for a net gain on all platforms. In an NFV environment this will not be the case, hence this will be an actual function call. Once you introduce a function call, the cost of this vs a simple load instruction is unlikely to provide any performance benefit, regardless of cache breaks.

I suggest we focus on the general problem of tuning ODP4VPP to achieve best performance to show benefit on Arm SoCs that offer meaningful HW acceleration, since that is the use case of most interest. Do we have a detailed performance profile at this point to use as a starting point?

bala-manoharan:
@sachin-saxena The solution 3 can not be effectively implementable across different platforms. This will reduce the performance for all different use-case where this API is not used by the application.
Modifying the API for one specific use-case is not effective.

PR #332, opened 2017-12-08
[PATCH API-NEXT v2] api: refine parser metadata specifications

psavol:
Offset functions just point to the start of an layer. Offsets
may be set also when a layer does not have a known header or
a known header has an error. For example, L4 offset may be
set always after a successful parse of L3 (IP) layer. There
are other API calls to check if L4 is a known protocol
(e.g. packet_has_l4, packet_has_udp), or if it contain errors
(e.g. packet_has_l4_error).

### DIFF
--- a/include/odp/api/spec/packet.h
+++ b/include/odp/api/spec/packet.h
@@ -1347,15 +1347,16 @@ uint32_t odp_packet_user_area_size(odp_packet_t pkt);
 /**
  * Layer 2 start pointer
  *
- * Returns pointer to the start of the layer 2 header. Optionally, outputs
- * number of data bytes in the segment following the pointer.
+ * Returns pointer to the start of layer 2. Optionally, outputs number of data
+ * bytes in the segment following the pointer. The pointer value is generated
+ * from the current layer 2 offset.
  *
  * @param      pkt      Packet handle
  * @param[out] len      Number of data bytes remaining in the segment (output).
  *                      Ignored when NULL.
  *
- * @return  Layer 2 start pointer
- * @retval  NULL packet does not contain a valid L2 header
+ * @return Layer 2 start pointer
+ * @retval NULL  Layer 2 offset has not been set
  *
  * @see odp_packet_l2_offset(), odp_packet_l2_offset_set(), odp_packet_has_l2()
  */
@@ -1364,16 +1365,16 @@ void *odp_packet_l2_ptr(odp_packet_t pkt, uint32_t *len);
 /**
  * Layer 2 start offset
  *
- * Returns offset to the start of the layer 2 header. The offset is calculated
- * from the current odp_packet_data() position in bytes.
- *
- * User is responsible to update the offset when modifying the packet data
- * pointer position.
+ * Returns offset to the start of layer 2. The offset is calculated from the
+ * current odp_packet_data() position in bytes. Packet parsing sets the offset
+ * according to parse configuration and layers recognized in the packet. Data
+ * start position updating functions (e.g. odp_packet_push_head()) do not modify
+ * the offset, but user sets a new value when needed.
  *
  * @param pkt  Packet handle
  *
- * @return  Layer 2 start offset
- * @retval ODP_PACKET_OFFSET_INVALID packet does not contain a valid L2 header
+ * @return Layer 2 start offset
+ * @retval ODP_PACKET_OFFSET_INVALID  Layer 2 offset has not been set
  *
  * @see odp_packet_l2_offset_set(), odp_packet_has_l2()
  */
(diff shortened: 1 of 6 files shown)


include/odp/api/spec/packet_flags.h
```diff
@@ -23,87 +23,114 @@ extern "C" {
...
+ *  Operations on packet metadata flags.
  *  @{
  */
 
 /**
- * Check for packet errors
+ * Check for all errors in packet
  *
- * Checks all error flags at once.
+ * Check if packet parsing have found any errors in the packet. The level of
```
Bill-Fischofer-Linaro: ...if packet parsing has found...

include/odp/api/spec/packet_flags.h
```diff
@@ -23,87 +23,114 @@ extern "C" {
...
+ * error checking depends on parse configuration (e.g. included layers and
```
Bill-Fischofer-Linaro: "the parse configuration" would be better here

include/odp/api/spec/packet_flags.h
```diff
@@ -23,87 +23,114 @@ extern "C" {
...
+ * @retval 0           No errors were found
  */
 int odp_packet_has_error(odp_packet_t pkt);
 
 /**
- * Check for packet L2 errors
+ * Check for errors in layer 2
  *
- * check for all L2 errors
+ * Check if packet parsing have found any errors in layer 2 of the packet.
```
Bill-Fischofer-Linaro: has found

include/odp/api/spec/packet_flags.h
```diff
@@ -23,87 +23,114 @@ extern "C" {
...
+ * Make sure that the layer is included into parse configuration.
```
Bill-Fischofer-Linaro: By context  this seems to be an input requirement usage note, so a clearer statement might be "If layer 2 is included in the parse configuration, check if packet parsing has found any errors in layer 2 of the packet".

include/odp/api/spec/packet_flags.h
```diff
@@ -23,87 +23,114 @@ extern "C" {
...
  *
- * @param pkt Packet handle
- * @retval non-zero packet has L2 errors
- * @retval 0 packet has no L2 error
+ * @param pkt          Packet handle
+ *
+ * @retval non-zero    Packet has errors in layer 2
+ * @retval 0           No errors were found in layer 2
```
Bill-Fischofer-Linaro: Worth noting this may be because layer 2 parsing was not configured? Or are results undefined in this case?

include/odp/api/spec/packet_flags.h
```diff
@@ -23,87 +23,114 @@ extern "C" {
...
+ * Check if packet parsing have found any errors in layer 3 of the packet.
+ * Make sure that the layer is included into parse configuration.
+ *
+ * @param pkt          Packet handle
+ *
+ * @retval non-zero    Packet has errors in layer 3
+ * @retval 0           No errors found in layer 3
  */
-int odp_packet_has_l2(odp_packet_t pkt);
+int odp_packet_has_l3_error(odp_packet_t pkt);
```
Bill-Fischofer-Linaro: Similar phrasing / grammar comments apply to all routines included in this patch as for `odp_packet_has_l2_error()` above.

include/odp/api/spec/packet.h
```diff
@@ -1347,15 +1347,16 @@ uint32_t odp_packet_user_area_size(odp_packet_t pkt);
 /**
  * Layer 2 start pointer
  *
- * Returns pointer to the start of the layer 2 header. Optionally, outputs
- * number of data bytes in the segment following the pointer.
+ * Returns pointer to the start of the layer 2. Optionally, outputs
```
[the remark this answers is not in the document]
psavol: Changed to "... layer X"

include/odp/api/spec/packet_flags.h
psavol: OK.

include/odp/api/spec/packet_flags.h
psavol: OK.

include/odp/api/spec/packet_flags.h
psavol: OK.

include/odp/api/spec/packet_flags.h
psavol: OK. Changed to: "When layer 2 is included in the parse configuration, check if any errors were found in layer 2 of the packet."

include/odp/api/spec/packet_flags.h
psavol: I didn't change this, since patch 4 gives the generic rule "... Other layer/protocol specific metadata have undefined values.". If application first says it does not need e.g. L4 parsing and then still asks for L4 metadata, those values are  undefined (implementation may leave it uninitialized).

include/odp/api/spec/packet_flags.h
psavol: OK.

lumag:

muvarov:
Merged.

PR #333, opened 2017-12-11
[PATCH API-NEXT v1] api: shm name is optional

psavol:
Align shared memory name parameter specification with other
APIs. Name is optional and does not need to be unique, except
when it's used for lookups.


muvarov:
Merged.

PR #337, opened 2017-12-12
[PATCH v1] tests: compile and execute validation tests before the rest of tests

lumag:
Validation tests examine API thoroughly. Compile and execute them before
performance and misc tests.

### DIFF
--- a/test/Makefile.am
+++ b/test/Makefile.am
@@ -1,1 +1,1 @@
-SUBDIRS = common performance miscellaneous validation
+SUBDIRS = common validation performance miscellaneous


test/Makefile.am
```diff
@@ -1 +1 @@
-SUBDIRS = common performance miscellaneous validation
+SUBDIRS = common validation performance miscellaneous
```
muvarov: For parallel make case you need ".". Might be for the same time it's good to isolate performance tests. I.e.:
SUBDIRS = common validation . performance . miscellaneous
https://www.gnu.org/software/automake/manual/html_node/Subdirectories.html

lumag: @muvarov hmm, you do not need "." here, as there is nothing executed in the current (test/) directory.

muvarov: @lumag --enable-test-perf enables performance tests. On parallel with 'make check' it's better to not run them with validation tests at the same time.

lumag: Dot just allows current directory to be built in parallel with the subdirs.

muvarov: than it has to be ok.

muvarov:
Merged.

PR #338, opened 2017-12-13
[PATCH API-NEXT v3] Deprecate odp_pktio_parser_layer_t

lumag:
`odp_pktio_parser_layer_t` duplicates `odp_proto_layer_t`. So let's deprecate the former type in favour of later one.


lumag:
As per agreement dropped actual `ODP_DEPRECATE` of old type/values. Doxygen tags mention that this type/values are deprecated

muvarov:
Merged.

PR #341, opened 2017-12-13
[PATCH v20] linux-generic: pktio: genuinely sleep in timeout receive

jmtilli:
ODP has two functions, odp_pktin_recv_tmo and odp_pktin_recv_mq_tmo that
allow sleeping for a certain time. However, the sleep was not a genuine
interruptible sleep (select/poll/epoll) but rather a busy loop. This
resulted in high CPU use when this call was used. This functionality
adds to two pktio mechanisms a way to allow sleeping using select()
system call. The way is used if all used pktio mechanisms support it.

### DIFF
--- a/example/l2fwd_simple/odp_l2fwd_simple.c
+++ b/example/l2fwd_simple/odp_l2fwd_simple.c
@@ -88,6 +88,7 @@ static int run_worker(void *arg ODP_UNUSED)
 	odp_packet_t pkt_tbl[MAX_PKT_BURST];
 	int pkts, sent, tx_drops, i;
 	int total_pkts = 0;
+	uint64_t wait_time = odp_pktin_wait_time(ODP_TIME_SEC_IN_NS);
 
 	if (odp_pktio_start(global.if0)) {
 		printf("unable to start input interface\n");
@@ -103,7 +104,7 @@ static int run_worker(void *arg ODP_UNUSED)
 
 	while (!exit_thr) {
 		pkts = odp_pktin_recv_tmo(global.if0in, pkt_tbl, MAX_PKT_BURST,
-					  ODP_PKTIN_NO_WAIT);
+					  wait_time);
 
 		if (odp_unlikely(pkts <= 0))
 			continue;
--- a/platform/linux-generic/odp_packet_io.c
+++ b/platform/linux-generic/odp_packet_io.c
@@ -32,10 +32,11 @@
 #include <errno.h>
 #include <time.h>
 
-/* Sleep this many nanoseconds between pktin receive calls */
-#define SLEEP_NSEC  1000
+/* Sleep this many microseconds between pktin receive calls. Must be smaller
+ * than 1000000 (a million), i.e. smaller than a second. */
+#define SLEEP_USEC  1
 
-/* Check total sleep time about every SLEEP_CHECK * SLEEP_NSEC nanoseconds.
+/* Check total sleep time about every SLEEP_CHECK * SLEEP_USEC microseconds.
  * Must be power of two. */
 #define SLEEP_CHECK 32
 
@@ -1679,9 +1680,10 @@ int odp_pktin_recv_tmo(odp_pktin_queue_t queue, odp_packet_t packets[], int num,
 	odp_time_t t1, t2;
 	struct timespec ts;
 	int started = 0;
+	uint64_t sleep_round = 0;
 
 	ts.tv_sec  = 0;
-	ts.tv_nsec = SLEEP_NSEC;
+	ts.tv_nsec = 1000 * SLEEP_USEC;
 
 	while (1) {
 		ret = odp_pktin_recv(queue, packets, num);
@@ -1698,21 +1700,20 @@ int odp_pktin_recv_tmo(odp_pktin_queue_t queue, odp_packet_t packets[], int num,
 			if (odp_unlikely(!started)) {
 				odp_time_t t;
 
-				t = odp_time_local_from_ns(wait * SLEEP_NSEC);
+				t = odp_time_local_from_ns(wait * 1000);
 				started = 1;
 				t1 = odp_time_sum(odp_time_local(), t);
 			}
 
 			/* Check every SLEEP_CHECK rounds if total wait time
 			 * has been exceeded. */
-			if ((wait & (SLEEP_CHECK - 1)) == 0) {
+			if ((++sleep_round & (SLEEP_CHECK - 1)) == 0) {
 				t2 = odp_time_local();
 
 				if (odp_time_cmp(t2, t1) > 0)
 					return 0;
 			}
-
-			wait--;
+			wait = wait > SLEEP_USEC ? wait - SLEEP_USEC : 0;
 		}
 
 		nanosleep(&ts, NULL);
@@ -1728,9 +1729,10 @@ int odp_pktin_recv_mq_tmo(const odp_pktin_queue_t queues[], unsigned num_q,
 	odp_time_t t1, t2;
 	struct timespec ts;
 	int started = 0;
+	uint64_t sleep_round = 0;
 
 	ts.tv_sec  = 0;
-	ts.tv_nsec = SLEEP_NSEC;
+	ts.tv_nsec = 1000 * SLEEP_USEC;
 
 	while (1) {
 		for (i = 0; i < num_q; i++) {
@@ -1750,19 +1752,20 @@ int odp_pktin_recv_mq_tmo(const odp_pktin_queue_t queues[], unsigned num_q,
 			if (odp_unlikely(!started)) {
 				odp_time_t t;
 
-				t = odp_time_local_from_ns(wait * SLEEP_NSEC);
+				t = odp_time_local_from_ns(wait * 1000);
 				started = 1;
 				t1 = odp_time_sum(odp_time_local(), t);
 			}
 
-			if ((wait & (SLEEP_CHECK - 1)) == 0) {
+			/* Check every SLEEP_CHECK rounds if total wait time
+			 * has been exceeded. */
+			if ((++sleep_round & (SLEEP_CHECK - 1)) == 0) {
 				t2 = odp_time_local();
 
 				if (odp_time_cmp(t2, t1) > 0)
 					return 0;
 			}
-
-			wait--;
+			wait = wait > SLEEP_USEC ? wait - SLEEP_USEC : 0;
 		}
 
 		nanosleep(&ts, NULL);
@@ -1774,9 +1777,9 @@ uint64_t odp_pktin_wait_time(uint64_t nsec)
 	if (nsec == 0)
 		return 0;
 
-	/* number of nanosleep calls rounded up by one, so that
+	/* number of microseconds rounded up by one, so that
 	 * recv_mq_tmo call waits at least 'nsec' nanoseconds. */
-	return (nsec / SLEEP_NSEC) + 1;
+	return (nsec / (1000)) + 1;
 }
 
 int odp_pktout_send(odp_pktout_queue_t queue, const odp_packet_t packets[],
--- a/platform/linux-generic/odp_packet_io.c
+++ b/platform/linux-generic/odp_packet_io.c
@@ -1681,12 +1681,23 @@ int odp_pktin_recv_tmo(odp_pktin_queue_t queue, odp_packet_t packets[], int num,
 	struct timespec ts;
 	int started = 0;
 	uint64_t sleep_round = 0;
+	pktio_entry_t *entry;
 
 	ts.tv_sec  = 0;
 	ts.tv_nsec = 1000 * SLEEP_USEC;
 
+	entry = get_pktio_entry(queue.pktio);
+	if (entry == NULL) {
+		ODP_DBG("pktio entry %d does not exist\n", queue.pktio);
+		return -1;
+	}
+
+	if (entry->s.ops->recv_tmo && wait != ODP_PKTIN_NO_WAIT)
+		return entry->s.ops->recv_tmo(entry, queue.index, packets, num,
+					      wait);
+
 	while (1) {
-		ret = odp_pktin_recv(queue, packets, num);
+		ret = entry->s.ops->recv(entry, queue.index, packets, num);
 
 		if (ret != 0)
 			return ret;
@@ -1730,6 +1741,26 @@ int odp_pktin_recv_mq_tmo(const odp_pktin_queue_t queues[], unsigned num_q,
 	struct timespec ts;
 	int started = 0;
 	uint64_t sleep_round = 0;
+	int trial_successful = 0;
+
+	for (i = 0; i < num_q; i++) {
+		ret = odp_pktin_recv(queues[i], packets, num);
+
+		if (ret > 0 && from)
+			*from = i;
+
+		if (ret != 0)
+			return ret;
+	}
+
+	if (wait == 0)
+		return 0;
+
+	ret = sock_recv_mq_tmo_try_int_driven(queues, num_q, from,
+					      packets, num, wait,
+					      &trial_successful);
+	if (trial_successful)
+		return ret;
 
 	ts.tv_sec  = 0;
 	ts.tv_nsec = 1000 * SLEEP_USEC;
--- a/platform/linux-generic/pktio/netmap.c
+++ b/platform/linux-generic/pktio/netmap.c
@@ -732,6 +732,44 @@ static inline int netmap_recv_desc(pktio_entry_t *pktio_entry,
 	return 0;
 }
 
+static int netmap_fd_set(pktio_entry_t *pktio_entry, int index, fd_set *readfds)
+{
+	struct nm_desc *desc;
+	pkt_netmap_t *pkt_nm = &pktio_entry->s.pkt_nm;
+	unsigned first_desc_id = pkt_nm->rx_desc_ring[index].s.first;
+	unsigned last_desc_id = pkt_nm->rx_desc_ring[index].s.last;
+	unsigned desc_id;
+	int num_desc = pkt_nm->rx_desc_ring[index].s.num;
+	int i;
+	int max_fd = 0;
+
+	if (odp_unlikely(pktio_entry->s.state != PKTIO_STATE_STARTED))
+		return 0;
+
+	if (!pkt_nm->lockless_rx)
+		odp_ticketlock_lock(&pkt_nm->rx_desc_ring[index].s.lock);
+
+	desc_id = pkt_nm->rx_desc_ring[index].s.cur;
+
+	for (i = 0; i < num_desc; i++) {
+		if (desc_id > last_desc_id)
+			desc_id = first_desc_id;
+
+		desc = pkt_nm->rx_desc_ring[index].s.desc[desc_id];
+
+		FD_SET(desc->fd, readfds);
+		if (desc->fd > max_fd)
+			max_fd = desc->fd;
+		desc_id++;
+	}
+	pkt_nm->rx_desc_ring[index].s.cur = desc_id;
+
+	if (!pkt_nm->lockless_rx)
+		odp_ticketlock_unlock(&pkt_nm->rx_desc_ring[index].s.lock);
+
+	return max_fd;
+}
+
 static int netmap_recv(pktio_entry_t *pktio_entry, int index,
 		       odp_packet_t pkt_table[], int num)
 {
@@ -784,6 +822,78 @@ static int netmap_recv(pktio_entry_t *pktio_entry, int index,
 		odp_ticketlock_unlock(&pkt_nm->rx_desc_ring[index].s.lock);
 
 	return num_rx;
+}
+
+static int netmap_recv_tmo(pktio_entry_t *pktio_entry, int index,
+			   odp_packet_t pkt_table[], int num, uint64_t usecs)
+{
+	struct timeval timeout;
+	int ret;
+	int maxfd;
+	fd_set readfds;
+
+	ret = netmap_recv(pktio_entry, index, pkt_table, num);
+	if (ret != 0)
+		return ret;
+
+	timeout.tv_sec = usecs / (1000 * 1000);
+	timeout.tv_usec = usecs - timeout.tv_sec * (1000ULL * 1000ULL);
+	FD_ZERO(&readfds);
+	maxfd = netmap_fd_set(pktio_entry, index, &readfds);
+
+	if (select(maxfd + 1, &readfds, NULL, NULL,
+		   usecs == ODP_PKTIN_WAIT ? NULL : &timeout) == 0)
+		return 0;
+
+	return netmap_recv(pktio_entry, index, pkt_table, num);
+}
+
+static int netmap_recv_mq_tmo(pktio_entry_t *pktio_entry[], int index[],
+			      int num_q, odp_packet_t pkt_table[], int num,
+			      unsigned *from, uint64_t usecs)
+{
+	struct timeval timeout;
+	int i;
+	int ret;
+	int maxfd = -1, maxfd2;
+	fd_set readfds;
+
+	for (i = 0; i < num_q; i++) {
+		ret = netmap_recv(pktio_entry[i], index[i], pkt_table, num);
+
+		if (ret > 0 && from)
+			*from = i;
+
+		if (ret != 0)
+			return ret;
+	}
+
+	FD_ZERO(&readfds);
+
+	for (i = 0; i < num_q; i++) {
+		maxfd2 = netmap_fd_set(pktio_entry[i], index[i], &readfds);
+		if (maxfd2 > maxfd)
+			maxfd = maxfd2;
+	}
+
+	timeout.tv_sec = usecs / (1000 * 1000);
+	timeout.tv_usec = usecs - timeout.tv_sec * (1000ULL * 1000ULL);
+
+	if (select(maxfd + 1, &readfds, NULL, NULL,
+		   usecs == ODP_PKTIN_WAIT ? NULL : &timeout) == 0)
+		return 0;
+
+	for (i = 0; i < num_q; i++) {
+		ret = netmap_recv(pktio_entry[i], index[i], pkt_table, num);
+
+		if (ret > 0 && from)
+			*from = i;
+
+		if (ret != 0)
+			return ret;
+	}
+
+	return 0;
 }
 
 static int netmap_send(pktio_entry_t *pktio_entry, int index,
@@ -975,7 +1085,10 @@ const pktio_if_ops_t netmap_pktio_ops = {
 	.input_queues_config = netmap_input_queues_config,
 	.output_queues_config = netmap_output_queues_config,
 	.recv = netmap_recv,
-	.send = netmap_send
+	.recv_tmo = netmap_recv_tmo,
+	.recv_mq_tmo = netmap_recv_mq_tmo,
+	.send = netmap_send,
+	.fd_set = netmap_fd_set
 };
 
 #endif /* ODP_NETMAP */
--- a/platform/linux-generic/pktio/socket.c
+++ b/platform/linux-generic/pktio/socket.c
@@ -693,6 +693,89 @@ static int sock_mmsg_recv(pktio_entry_t *pktio_entry, int index ODP_UNUSED,
 	return nb_rx;
 }
 
+static int sock_fd_set(pktio_entry_t *pktio_entry, int index ODP_UNUSED,
+		       fd_set *readfds)
+{
+	pkt_sock_t *pkt_sock = &pktio_entry->s.pkt_sock;
+	const int sockfd = pkt_sock->sockfd;
+
+	FD_SET(sockfd, readfds);
+	return sockfd;
+}
+
+static int sock_recv_tmo(pktio_entry_t *pktio_entry, int index,
+			 odp_packet_t pkt_table[], int num, uint64_t usecs)
+{
+	struct timeval timeout;
+	int ret;
+	int maxfd;
+	fd_set readfds;
+
+	ret = sock_mmsg_recv(pktio_entry, index, pkt_table, num);
+	if (ret != 0)
+		return ret;
+
+	timeout.tv_sec = usecs / (1000 * 1000);
+	timeout.tv_usec = usecs - timeout.tv_sec * (1000ULL * 1000ULL);
+
+	FD_ZERO(&readfds);
+	maxfd = sock_fd_set(pktio_entry, index, &readfds);
+
+	if (select(maxfd + 1, &readfds, NULL, NULL,
+		   usecs == ODP_PKTIN_WAIT ? NULL : &timeout) == 0)
+		return 0;
+
+	return sock_mmsg_recv(pktio_entry, index, pkt_table, num);
+}
+
+static int sock_recv_mq_tmo(pktio_entry_t *pktio_entry[], int index[],
+			    int num_q, odp_packet_t pkt_table[], int num,
+			    unsigned *from, uint64_t usecs)
+{
+	struct timeval timeout;
+	int i;
+	int ret;
+	int maxfd = -1, maxfd2;
+	fd_set readfds;
+
+	for (i = 0; i < num_q; i++) {
+		ret = sock_mmsg_recv(pktio_entry[i], index[i], pkt_table, num);
+
+		if (ret > 0 && from)
+			*from = i;
+
+		if (ret != 0)
+			return ret;
+	}
+
+	timeout.tv_sec = usecs / (1000 * 1000);
+	timeout.tv_usec = usecs - timeout.tv_sec * (1000ULL * 1000ULL);
+
+	FD_ZERO(&readfds);
+
+	for (i = 0; i < num_q; i++) {
+		maxfd2 = sock_fd_set(pktio_entry[i], index[i], &readfds);
+		if (maxfd2 > maxfd)
+			maxfd = maxfd2;
+	}
+
+	if (select(maxfd + 1, &readfds, NULL, NULL,
+		   usecs == ODP_PKTIN_WAIT ? NULL : &timeout) == 0)
+		return 0;
+
+	for (i = 0; i < num_q; i++) {
+		ret = sock_mmsg_recv(pktio_entry[i], index[i], pkt_table, num);
+
+		if (ret > 0 && from)
+			*from = i;
+
+		if (ret != 0)
+			return ret;
+	}
+
+	return 0;
+}
+
 static uint32_t _tx_pkt_to_iovec(odp_packet_t pkt,
 				 struct iovec iovecs[MAX_SEGS])
 {
@@ -868,6 +951,9 @@ const pktio_if_ops_t sock_mmsg_pktio_ops = {
 	.stats = sock_stats,
 	.stats_reset = sock_stats_reset,
 	.recv = sock_mmsg_recv,
+	.recv_tmo = sock_recv_tmo,
+	.recv_mq_tmo = sock_recv_mq_tmo,
+	.fd_set = sock_fd_set,
 	.send = sock_mmsg_send,
 	.mtu_get = sock_mtu_get,
 	.promisc_mode_set = sock_promisc_mode_set,
--- a/platform/linux-generic/pktio/socket_mmap.c
+++ b/platform/linux-generic/pktio/socket_mmap.c
@@ -600,6 +600,20 @@ error:
 	return -1;
 }
 
+static int sock_mmap_fd_set(pktio_entry_t *pktio_entry, int index ODP_UNUSED,
+			    fd_set *readfds)
+{
+	pkt_sock_mmap_t *const pkt_sock = &pktio_entry->s.pkt_sock_mmap;
+	int fd;
+
+	odp_ticketlock_lock(&pktio_entry->s.rxl);
+	fd = pkt_sock->sockfd;
+	FD_SET(fd, readfds);
+	odp_ticketlock_unlock(&pktio_entry->s.rxl);
+
+	return fd;
+}
+
 static int sock_mmap_recv(pktio_entry_t *pktio_entry, int index ODP_UNUSED,
 			  odp_packet_t pkt_table[], int len)
 {
@@ -612,6 +626,79 @@ static int sock_mmap_recv(pktio_entry_t *pktio_entry, int index ODP_UNUSED,
 	odp_ticketlock_unlock(&pktio_entry->s.rxl);
 
 	return ret;
+}
+
+static int sock_mmap_recv_tmo(pktio_entry_t *pktio_entry, int index,
+			      odp_packet_t pkt_table[], int num, uint64_t usecs)
+{
+	struct timeval timeout;
+	int ret;
+	int maxfd;
+	fd_set readfds;
+
+	ret = sock_mmap_recv(pktio_entry, index, pkt_table, num);
+	if (ret != 0)
+		return ret;
+
+	timeout.tv_sec = usecs / (1000 * 1000);
+	timeout.tv_usec = usecs - timeout.tv_sec * (1000ULL * 1000ULL);
+
+	FD_ZERO(&readfds);
+	maxfd = sock_mmap_fd_set(pktio_entry, index, &readfds);
+
+	if (select(maxfd + 1, &readfds, NULL, NULL,
+		   usecs == ODP_PKTIN_WAIT ? NULL : &timeout) == 0)
+		return 0;
+
+	return sock_mmap_recv(pktio_entry, index, pkt_table, num);
+}
+
+static int sock_mmap_recv_mq_tmo(pktio_entry_t *pktio_entry[], int index[],
+				 int num_q, odp_packet_t pkt_table[], int num,
+				 unsigned *from, uint64_t usecs)
+{
+	struct timeval timeout;
+	int i;
+	int ret;
+	int maxfd = -1, maxfd2;
+	fd_set readfds;
+
+	for (i = 0; i < num_q; i++) {
+		ret = sock_mmap_recv(pktio_entry[i], index[i], pkt_table, num);
+
+		if (ret > 0 && from)
+			*from = i;
+
+		if (ret != 0)
+			return ret;
+	}
+
+	FD_ZERO(&readfds);
+
+	for (i = 0; i < num_q; i++) {
+		maxfd2 = sock_mmap_fd_set(pktio_entry[i], index[i], &readfds);
+		if (maxfd2 > maxfd)
+			maxfd = maxfd2;
+	}
+
+	timeout.tv_sec = usecs / (1000 * 1000);
+	timeout.tv_usec = usecs - timeout.tv_sec * (1000ULL * 1000ULL);
+
+	if (select(maxfd + 1, &readfds, NULL, NULL,
+		   usecs == ODP_PKTIN_WAIT ? NULL : &timeout) == 0)
+		return 0;
+
+	for (i = 0; i < num_q; i++) {
+		ret = sock_mmap_recv(pktio_entry[i], index[i], pkt_table, num);
+
+		if (ret > 0 && from)
+			*from = i;
+
+		if (ret != 0)
+			return ret;
+	}
+
+	return 0;
 }
 
 static int sock_mmap_send(pktio_entry_t *pktio_entry, int index ODP_UNUSED,
@@ -725,7 +812,10 @@ const pktio_if_ops_t sock_mmap_pktio_ops = {
 	.stats = sock_mmap_stats,
 	.stats_reset = sock_mmap_stats_reset,
 	.recv = sock_mmap_recv,
+	.recv_tmo = sock_mmap_recv_tmo,
+	.recv_mq_tmo = sock_mmap_recv_mq_tmo,
 	.send = sock_mmap_send,
+	.fd_set = sock_mmap_fd_set,
 	.mtu_get = sock_mmap_mtu_get,
 	.promisc_mode_set = sock_mmap_promisc_mode_set,
 	.promisc_mode_get = sock_mmap_promisc_mode_get,
--- a/platform/linux-generic/include/odp_packet_io_internal.h
+++ b/platform/linux-generic/include/odp_packet_io_internal.h
@@ -207,6 +207,12 @@ typedef struct pktio_if_ops {
 	odp_time_t (*pktin_ts_from_ns)(pktio_entry_t *pktio_entry, uint64_t ns);
 	int (*recv)(pktio_entry_t *entry, int index, odp_packet_t packets[],
 		    int num);
+	int (*recv_tmo)(pktio_entry_t *entry, int index, odp_packet_t packets[],
+			int num, uint64_t wait_usecs);
+	int (*recv_mq_tmo)(pktio_entry_t *entry[], int index[], int num_q,
+			   odp_packet_t packets[], int num, unsigned *from,
+			   uint64_t wait_usecs);
+	int (*fd_set)(pktio_entry_t *entry, int index, fd_set *readfds);
 	int (*send)(pktio_entry_t *entry, int index,
 		    const odp_packet_t packets[], int num);
 	uint32_t (*mtu_get)(pktio_entry_t *pktio_entry);
@@ -274,6 +280,26 @@ int sock_stats_fd(pktio_entry_t *pktio_entry,
 		  odp_pktio_stats_t *stats,
 		  int fd);
 int sock_stats_reset_fd(pktio_entry_t *pktio_entry, int fd);
+
+/**
+ * Try interrupt-driven receive
+ *
+ * @param queues Pktin queues
+ * @param num_q Number of queues
+ * @param packets Output packet slots
+ * @param num Number of output packet slots
+ * @param from Queue from which the call received packets
+ * @param usecs Microseconds to wait
+ * @param trial_successful Will receive information whether trial was successful
+ *
+ * @return >=0 on success, number of packets received
+ * @return <0 on failure
+ */
+int sock_recv_mq_tmo_try_int_driven(const struct odp_pktin_queue_t queues[],
+				    unsigned num_q, unsigned *from,
+				    odp_packet_t packets[], int num,
+				    uint64_t usecs,
+				    int *trial_successful);
 
 #ifdef __cplusplus
 }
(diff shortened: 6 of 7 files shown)


Bill-Fischofer-Linaro:
@jmtilli Thank you for the contribution. Please sign the Contributor's Agreement for ODP to enable this to be considered for inclusion into the code base.

jmtilli:
Hi, I now signed the contributor's agreement electronically. So this part should be okay.

example/l2fwd_simple/odp_l2fwd_simple.c
```diff
@@ -88,6 +88,7 @@ static int run_worker(void *arg ODP_UNUSED)
 	odp_packet_t pkt_tbl[MAX_PKT_BURST];
 	int pkts, sent, tx_drops, i;
 	int total_pkts = 0;
+	uint64_t wait_time = odp_pktin_wait_time(1000 * 1000 * 1000);
```
matiaselo: ODP_TIME_SEC_IN_NS would work here.

matiaselo:
Overall seems like a good idea. I'd suggest doing some refactoring to keep the pktio specific code in odp_packet_io.c in minimum. A better place for the `select()` based code would probably be `socket.c`.

Cleaning `odp_pktin_recv_tmo()` is simple, just add a new function pointer to `pktio_if_ops` (e.g. recv_tmo). Today only the pktios supporting `select()` will set this and the rest will fall back to the current generic sleep based implementation.

E.g.
```
if (entry->s.ops->recv_tmo)
		return entry->s.ops->recv_tmo(entry, param,...);

while (1) {
	ret = odp_pktin_recv(queue, packets, num);
...
```

`odp_pktin_recv_mq_tmo()` is more tricky. At least the select() based code should be split into another function, which will be called from `odp_pktin_recv_mq_tmo()`.

jmtilli:
Would it be a good idea to have recv_mq_tmo function pointer in the supported operations as well? Then for the most common case where all of the pktios are netmap, socket or whatever but nevertheless the same type, the odp_pktin_recv_mq_tmo() function could call the recv_mq_tmo function pointer? Of course, this would not work for mixed socket/netmap pktios, so I can't foresee removing the select() code and the associated .fd_set function completely.

Nevertheless, this recv_mq_tmo might allow doing the same with DPDK. I'm not sure what the status of 0% idle CPU use in DPDK is today, but I'm pretty sure if DPDK reaches wide adoption some kind of solution to the 100% idle CPU use problem is required. We can be certain that the DPDK solution will not use file descriptors, as DPDK is a userspace solution and not a kernelspace solution. File descriptors are part of the interface between the userspace and the kernelspace, and thus not applicable in DPDK.

muvarov:
"Would it be a good idea to have recv_mq_tmo function pointer in the supported operations as well?" I like this idea.  In some cases scheduler don't need to poll recv function, it can provide just timeout and polling will be inside rx function. From other side sheduler can return a lot of other different events and you can not roll inside rx while other events are ready. For direct mode it might work. For scheduler it's  unclear. We have to discuss that.

matiaselo:
> Would it be a good idea to have recv_mq_tmo function pointer in the supported operations as well? Then for the most common case where all of the pktios are netmap, socket or whatever but nevertheless the same type, the odp_pktin_recv_mq_tmo() function could call the recv_mq_tmo function pointer?

It is not enough that all pktin queues are from the same type of pktio device. They would also have to be from the same pktio device. Only the pktio device, which the input queue belongs to, can receive packets from it. For example packet counters or input tags would break if this does not hold.

> 
> Nevertheless, this recv_mq_tmo might allow doing the same with DPDK. I'm not sure what the status of 0% idle CPU use in DPDK is today, but I'm pretty sure if DPDK reaches wide adoption some kind of solution to the 100% idle CPU use problem is required. We can be certain that the DPDK solution will not use file descriptors, as DPDK is a userspace solution and not a kernelspace solution. File descriptors are part of the interface between the userspace and the kernelspace, and thus not applicable in DPDK.
> 

DPDK seems to have support for interrupt mode PMDs (at least for ixgbe and igb), which maybe could be used to implement this.

platform/linux-generic/odp_packet_io.c
```diff
@@ -1577,10 +1577,27 @@ int odp_pktin_recv_tmo(odp_pktin_queue_t queue, odp_packet_t packets[], int num,
...
+	}
+
+	if (entry->s.ops->recv_tmo) {
+		uint64_t usecs = (wait * SLEEP_NSEC + 999) / 1000;
+
+		if (wait == ODP_PKTIN_WAIT)
+			usecs = ODP_PKTIN_WAIT;
+
+		return entry->s.ops->recv_tmo(entry, queue.index, packets, num,
+					      usecs);
```
matiaselo: You should pass the `wait` parameter to `entry->s.ops->recv_tmo()` and not do any conversions here.

platform/linux-generic/odp_packet_io.c
```diff
@@ -1617,6 +1634,46 @@ int odp_pktin_recv_tmo(odp_pktin_queue_t queue, odp_packet_t packets[], int num,
 	}
 }
 
+static int odp_pktin_recv_mq_tmo_select(const odp_pktin_queue_t queues[],
```
matiaselo: Internal functions should not use odp_ prefix.

platform/linux-generic/odp_packet_io.c
```diff
@@ -1626,6 +1683,68 @@ int odp_pktin_recv_mq_tmo(const odp_pktin_queue_t queues[], unsigned num_q,
...
+		if (entry[i] == NULL) {
+			ODP_DBG("pktio entry %d does not exist\n",
+				queues[i].pktio);
+			return -1;
+		}
+		if (!impl_set) {
+			impl = entry[i]->s.ops->recv_mq_tmo;
+			impl_set = 1;
+		} else {
+			if (impl != entry[i]->s.ops->recv_mq_tmo) {
```
matiaselo: As I noted in the previous comment, I think this check isn't enough. The input queues would have to be from the same pktio device for this to work.

platform/linux-generic/odp_packet_io.c
jmtilli: I disagree. SLEEP_NSEC is a private #define in odp_packet_io.c so what the wait parameter actually means cannot be interpreted in the drivers. (Believe me, I tried it!) Thus, the conversion from the sleep wait parameter to microseconds is needed.

platform/linux-generic/odp_packet_io.c
jmtilli: Will fix.

platform/linux-generic/odp_packet_io.c
jmtilli: Several options: (1) I could implement a binary search algorithm that incrementally checks various usec values and uses odp_pktin_wait_time() to check which one of these trial attempts result in the same sleep wait value, but that would be slow; (2) I could implement odp_pktin_wait_time_reverse() that does the conversion in the other direction, but that would be an API change; (3) I could just do the conversion and pass usec values instead of sleep wait values. I carefully considered each and selected (3), but do you think (2) or (1) would be better?

platform/linux-generic/odp_packet_io.c
jmtilli: Are you entirely sure about this? This is a function pointer, and if two pktio devices have the same function pointer for recv_mq_tmo(), the other function pointers including the recv() function pointer should be identical as well. We do pass the device (entry[i]) to the function pointed to by recv_mq_tmo as well, so it should be able to update the correct counters. The ops structure is const, after all...

So, when receiving packets from a particular device using netmap, we do call netmap_recv() and pass to it the pktio_entry[i] and also index[i] for the queue index. This is the same that the odp_pktin_recv_mq_tmo() function already did. It calls odp_pktin_recv() in a loop, and the latter function calls entry->s.ops->recv(entry, queue.index, packets, num); which to me seems to be exactly the same thing -- you pass exactly the same entry and the same queue.index. Same function, same arguments, I can't see how the behavior would be different...

Now, if somebody opposes the .recv_mq_tmo function pointer due to added code line count, I might perhaps agree. But it seems to be the only way the DPDK support for genuine sleep could be added, so I see some value in it.

platform/linux-generic/odp_packet_io.c
muvarov: just pass nsec to drivers or you will have double calculation like in netmap code.

platform/linux-generic/odp_packet_io.c
matiaselo: > Are you entirely sure about this? This is a function pointer, and if two pktio devices have the same function pointer for recv_mq_tmo(), the other function pointers including the recv() function pointer should be identical as well. We do pass the device (entry[i]) to the function pointed to by recv_mq_tmo as well, so it should be able to update the correct counters. The ops structure is const, after all...

I stand corrected, didn't notice the pktio_entry_t array parameter. A bit confusing as `recv_mq_tmo()` behaves differently than pretty much every other pktio op.

platform/linux-generic/odp_packet_io.c
matiaselo: Every pktio type we currenly have uses microseconds for timeout, so you could change `recv_tmo()` and `recv_mq_tmo()` to use microseconds instead of nanoseconds. This removes the need for per operation conversions.

platform/linux-generic/odp_packet_io.c
```diff
@@ -1577,10 +1577,29 @@ int odp_pktin_recv_tmo(odp_pktin_queue_t queue, odp_packet_t packets[], int num,
...
 	ts.tv_nsec = SLEEP_NSEC;
 
+	entry = get_pktio_entry(queue.pktio);
+	if (entry == NULL) {
+		ODP_DBG("pktio entry %d does not exist\n", queue.pktio);
+		return -1;
+	}
+
+	if (entry->s.ops->recv_tmo) {
+		uint64_t usecs = (wait * SLEEP_NSEC + 999) / 1000;
```
matiaselo: The motivation for a separate `odp_pktin_wait_time()` function was to avoid having to do time conversions in the fast path functions. As `odp_pktin_wait_time()` already returns the time in microseconds no additional conversions are needed here. Same thing for `odp_pktin_recv_mq_tmo()`.

Currently, the value returned by `odp_pktin_recv_mq_tmo()` depends on `SLEEP_NSEC` definition. This should be modified so that the returned value is always in microseconds.

platform/linux-generic/pktio/netmap.c
```diff
@@ -786,6 +824,78 @@ static int netmap_recv(pktio_entry_t *pktio_entry, int index,
...
+	FD_ZERO(&readfds);
+
+	for (i = 0; i < num_q; i++) {
+		maxfd2 = netmap_fd_set(pktio_entry[i], index[i], &readfds);
+		if (maxfd2 > maxfd)
+			maxfd = maxfd2;
+	}
+
+	for (i = 0; i < num_q; i++) {
+		ret = netmap_recv(pktio_entry[i], index[i], pkt_table, num);
```
matiaselo: Since this loop is a performance optimization I would move it before the loop for setting the read descriptors. No use to set the descriptors if packets are already available.

platform/linux-generic/pktio/socket.c
```diff
@@ -693,6 +693,88 @@ static int sock_mmsg_recv(pktio_entry_t *pktio_entry, int index ODP_UNUSED,
...
+	FD_ZERO(&readfds);
+
+	for (i = 0; i < num_q; i++) {
+		maxfd2 = sock_fd_set(pktio_entry[i], index[i], &readfds);
+		if (maxfd2 > maxfd)
+			maxfd = maxfd2;
+	}
+
+	for (i = 0; i < num_q; i++) {
+		ret = sock_mmsg_recv(pktio_entry[i], index[i], pkt_table, num);
```
matiaselo: Same thing with the loop order than with netmap.

platform/linux-generic/pktio/socket.c
```diff
@@ -855,6 +937,115 @@ static int sock_init_global(void)
...
+			*from = i;
+
+		if (ret != 0)
+			return ret;
+	}
+
+	return 0;
+}
+
+int sock_recv_mq_tmo_try_int_driven(const struct odp_pktin_queue_t queues[],
```
matiaselo: Couple comment lines describing the different stages of this function wouldn't hurt when someone is reading this code later on.

platform/linux-generic/pktio/socket_mmap.c
```diff
@@ -614,6 +628,78 @@ static int sock_mmap_recv(pktio_entry_t *pktio_entry, int index ODP_UNUSED,
...
+	FD_ZERO(&readfds);
+
+	for (i = 0; i < num_q; i++) {
+		maxfd2 = sock_mmap_fd_set(pktio_entry[i], index[i], &readfds);
+		if (maxfd2 > maxfd)
+			maxfd = maxfd2;
+	}
+
+	for (i = 0; i < num_q; i++) {
+		ret = sock_mmap_recv(pktio_entry[i], index[i], pkt_table, num);
```
matiaselo: Same thing with the loop order than with netmap.

platform/linux-generic/odp_packet_io.c
jmtilli: I'm somewhat worried that somebody could adjust SLEEP_NSEC and then the entire thing fails. What about this:

<pre>
#if SLEEP_NSEC == 1000
  uint64_t usecs = wait;
#else
  uint64_t usecs = (wait * SLEEP_NSEC + 999) / 1000;
#endif
</pre>

For the return value, you mean <code>odp_pktin_wait_time()</code> and not <code>odp_pktin_recv_mq_tmo()</code>, right? Because I don't see how the return value of the latter could depend on <code>SLEEP_NSEC</code> definition.

Also, if we modify <code>odp_pktin_wait_time()</code> to always return microseconds, then <code>odp_pktin_recv_mq_tmo()</code> suddenly starts to work incorrectly with <code>SLEEP_NSEC != 1000</code> as there is a <code>wait--</code> line in there that does this just before sleeping one microsecond, exactly.

matiaselo:
> 
> @jmtilli commented on this pull request.
> 
> In platform/linux-generic/odp_packet_io.c:
> 
> >  
>  	ts.tv_sec  = 0;
>  	ts.tv_nsec = SLEEP_NSEC;
>  
> +	entry = get_pktio_entry(queue.pktio);
> +	if (entry == NULL) {
> +		ODP_DBG("pktio entry %d does not exist\n", queue.pktio);
> +		return -1;
> +	}
> +
> +	if (entry->s.ops->recv_tmo) {
> +		uint64_t usecs = (wait * SLEEP_NSEC + 999) / 1000;
> 
> I'm somewhat worried that somebody could adjust SLEEP_NSEC and then the entire thing fails.
> What about this:
> 
> #if SLEEP_NSEC == 1000
>   uint64_t usecs = wait;
> #else
>   uint64_t usecs = (wait * SLEEP_NSEC + 999) / 1000;
> #endif

Ifdefs are a big no-no.

> 
> For the return value, you mean odp_pktin_wait_time() and not odp_pktin_recv_mq_tmo(), right? Because I don't see how the return value of the latter could depend on SLEEP_NSEC definition.

Yep, odp_pktin_wait_time().

> 
> Also, if we modify odp_pktin_wait_time() to always return microseconds, then odp_pktin_recv_mq_tmo() suddenly starts to work incorrectly with SLEEP_NSEC != 1000 as there is a wait-- line in there that does this just before sleeping one microsecond, exactly.

The code has to be modified a bit to prevent this happening. Maybe the simplest fix is to change SLEEP_NSEC to SLEEP_MSEC. Then 'wait--' just becomes 'wait -= SLEEP_MSEC' (zero has to be handled). The couple other places also have be modified where SLEEP_NSEC is currently used.

Maybe it would be cleaner to do this change in a separate commit.

-Matias

jmtilli:
Ok, let's do the SLEEP_NSEC fix separately. I now removed the #if away. The (wait*SLEEP_NSEC+999)/1000 code is still there but that isn't good to be removed in the same commit.

JannePeltonen:
I am afraid this could perhaps hurt some applications that want high throughput and low latency.

With the current polling implementation the _tmo functions are fast and work well in the case there is very little idle time and/or the timeout is very short. This patch trades off speed (good throughput and low latency) for lower CPU utilization and lower power consumption in the idle and partial load case. Is this ok?

Shouldn't odp_pktin_recv_tmo() and especially odp_pktin_recv_mq_tmo() first try reception without a timeout and dive to the more costly timeout handling code only when sleeping is really needed?

I think the code should be optimized to cause less impact to pktio types (DPDK in particular) that do not support the recv_mq_tmo operation.

Finally, I do not think the sleeping should be extended to the DPDK pktio without a possibility to disable it if needed. The latency of waking up from an OS level sleep can be unacceptable to some applications that instead would happily spend resources in busy polling to get lower packet reception latency.

jmtilli:
@JannePeltonen I could perhaps optimize this more so that the select() call and the interrupt driven mode is called only if wait is different from ODP_PKTIN_NO_WAIT. For high throughput and low latency, ODP_PKTIN_NO_WAIT is what you want to use. I agree that the current patch isn't necessarily as well optimized for ODP_PKTIN_NO_WAIT as it could be, but this is just a feature of the current patch version, not a feature of the grand idea that some applications might benefit from 0% idle CPU use. I actually currently at work observed 10Gbit/s performance with ODP_PKTIN_WAIT and 18Gbit/s performance with ODP_PKTIN_NO_WAIT (although the latter caused 100% CPU use even when at idle). So certainly on some machines and for some applications ODP_PKTIN_NO_WAIT is what you want to use, and the code should be perhaps optimized a bit more for this use case.

matiaselo:
> I am afraid this could perhaps hurt some applications that want high throughput and low latency.
> 
> With the current polling implementation the _tmo functions are fast and work well in the case there is very little idle time and/or the timeout is very short. This patch trades off speed (good throughput and low latency) for lower CPU utilization and lower power consumption in the idle and partial load case. Is this ok?
> 
> I think the code should be optimized to cause less impact to pktio types (DPDK in particular) that do not support the recv_mq_tmo operation.

I agree with the performance concern. The odp_pktin_recv_tmo() seems OK though. It only adds a single if clause compared to the old implementation. The one odp_pktin_recv() call could be replaced with entry->s.ops->recv() as pktio entry is already known.

odp_pktin_recv_mq_tmo() definitely needs more optimization. One method would be to keep track if any pktios supporting recv_mq_tmo() have been opened.

> Shouldn't odp_pktin_recv_tmo() and especially odp_pktin_recv_mq_tmo() first try reception without a timeout and dive to the more costly timeout handling code only when sleeping is really needed?

The pktio specific recv_tmo() implementations already try reception before calling select so they are fine. odp_pktin_recv_mq_tmo() should be modified to do the same. The same receive loop could also check if it is necessary to call sock_recv_mq_tmo_try_int_driven() at  all.

platform/linux-generic/pktio/socket.c
```diff
@@ -855,6 +937,126 @@ static int sock_init_global(void)
...
+
+	for (i = 0; i < num_q; i++) {
+		entry[i] = get_pktio_entry(queues[i].pktio);
+		index[i] = queues[i].index;
+		if (entry[i] == NULL) {
+			ODP_DBG("pktio entry %d does not exist\n",
+				queues[i].pktio);
+			*trial_successful = 0;
+			return -1;
+		}
```
matiaselo: By checking both ops here you can return right away and minimize overhead for pktios not supporting interrupt driven mode.
E.g.
```
		if (!entry[i]->s.ops->recv_mq_tmo && !entry[i]->s.ops->fd_set) {
			*trial_successful = 0;
			return -1;
		}
```

jmtilli: I would prefer return 0; as the trial_successful already tells not to use this interrupt-driven way, but it probably doesn't matter much. I'll do this (with return 0).

matiaselo:
The commit message shouldn't include comments about patch revisions, but otherwise the patch is looking good.

 Reviewed-and-tested-by: Matias Elo <matias.elo@nokia.com>

muvarov:
1. Patch for l2fwd should be separate patch.

muvarov:
2. dpdk changes are missing. was this tested with dpdk pktio?

muvarov:
3. odp_pktin_recv_mq_tmo() calls sock_recv_mq_tmo_try_int_driven() which is specific for sockets. Is that suitable for netmap, dpdk, tap, ipc?

muvarov:
4. odp_pktin_recv_mq_tmo() converts nsec to usec. Then passes to pktio driver. Then each driver converts it to struct timeval. So you have 4 divisions on 64 values in fast path function. PKTIO like ipc might not need timeval. I did not understand the point doing calculations in common code then again in driver code if the same uint64_t is passed.

muvarov:
5. function which is shared between socket and netmap should be in common file. I think here: platform/linux-generic/pktio/pktio_common.c

jmtilli:
Hi,

Answers for the comments:
1. Will do.
2. I don't have now access to any machine with DPDK, but the intention was that this would not break DPDK support due to DPDK having the critical pointers set to the default NULL value (and thus using the old mechanism). The problem is that I'm not a DPDK expert, and thus, I'm not capable of offering the same 0% CPU load for DPDK. Matias Elo said that DPDK now supports interrupt-driven operation nowadays, but how, that is a mystery to me. All I know about interruptible sleeps is file descriptors and select()/poll()/epoll_wait(), and thus this is the only mechanism I was able to implement.
3. Yes, it is suitable for all pktio mechanisms. The keyword here is "try". If it's not suitable, it returns with an indication that the trial failed. Then the old mechanism is used.
4. odp_pktin_recv_mq_tmo() does not convert nsec to usec. It converts the "wait units" into usec, and the conversion in practice is a no-op because "wait units" are already in usecs (SLEEP_NSEC is 1000). It may be the case that due to overflow considerations the compiler cannot entirely optimize it away. I provided an "#if SLEEP_NSEC == 1000" preprocessor conditional to avoid the unnecessary division, leading to faster operation, but was instructed by Matias Elo that preprocessor use should be avoided, and thus there is no such conditional anymore. However, please note that the gcc compiler is quite intelligent and uses the famous trick in the paper <a href="https://gmplib.org/~tege/divcnst-pldi94.pdf">Division by Invariant Integers using Multiplication</a> to do the division. Thus, the division is extremely fast because the divisor is invariant. If you think that fast invariant division is worse than use of preprocessor conditionals, I can add the preprocessor conditional back. What if I also moved the usec calculation later in odp_pktin_recv_mq_tmo(), after the function has tried each pktin queue to see if there are packets? That way, the fast path in odp_pktin_recv_mq_tmo() would not have the division if the queues are full. As for odp_pktin_recv_tmo(), if you care about performance at the expense of idle CPU load, you use ODP_PKTIN_NO_WAIT which causes the division to be avoided completely.
5. I was instructed by Matias Elo to put it into socket.c. But now it seems that pktio_common.c is a better location for it. Will fix.

jmtilli:
About the division performance... Please test this program:

<pre>
#include <sys/time.h>
#include <stdio.h>
#include <stdlib.h>
#include <stdint.h>

int main(int argc, char **argv)
{
  volatile uint64_t x = 123456789;
  volatile uint64_t y = 0;
  struct timeval tv1, tv2;
  int i;
  gettimeofday(&tv1, NULL);
  for (i = 0; i < 1000*1000*1000; i++)
  {
    y = (x + 999) / 1000;
  }
  gettimeofday(&tv2, NULL);
  printf("%g MPPS\n", 1e3 / ( tv2.tv_sec - tv1.tv_sec +
                             (tv2.tv_usec - tv1.tv_usec) / 1e6));
  return 0;
}
</pre>

On my laptop, it gives 1163.2 MPPS. This is two orders of magnitude faster than your typical pktio mechanism. If you take into account that typically you process packets in bursts of at least 10, it means that 0.1% of all time is spent in the invariant division. Do you see this as a problem? If you do, I can use the preprocessor trick or alternatively delay the division until pktin queues are checked for packets available immediately.

The code I provided does an actual invariant division, gcc -O3 -S gives this:
<pre>
        movl    $1000000000, %ecx
        movabsq $2361183241434822607, %rsi
        .p2align 4,,10
        .p2align 3
.L2:
        movq    (%rsp), %rdx
        addq    $999, %rdx
        shrq    $3, %rdx
        movq    %rdx, %rax
        mulq    %rsi
        shrq    $4, %rdx
        subl    $1, %ecx
        movq    %rdx, 8(%rsp)
        jne     .L2
</pre>

See the missing div instruction. If you divide by invariant integer, it doesn't use division; it uses multiplication.

jmtilli:
Also, about the performance... I do believe the "sleep 1 microsecond, try again after the timeout has elapsed" is fundamentally worse approach than "sleep 1 millisecond/second/whatever or until the next packet arrives" approach. I have seen some really poor performance with the approach of sleeping always 1 microsecond (which is the approach currently used in ODP). I genuinely believe the interruptible sleep is going to INCREASE performance, not decrease it, if your traffic load is such that the queues may be occasionally empty. Previously, I had to use ODP_PKTIN_NO_WAIT to obtain best possible performance. With this patch, that may not be needed anymore.

Think about this... Which is the worst offender? The invariant division, which takes less than 1 nanosecond when compiled with a modern C compiler? Or the constant 1 microsecond sleep in the current codebase?

muvarov:
```
I reduced you cycle on 10, i.e. to:

  uint64_t i;
  gettimeofday(&tv1, NULL);
  for (i = 0; i < 100*1000*1000; i++)
  {
    y = (x + 999) / 1000;
  }

time ./test 
61.1395 MPPS <--- this number is wrong due to cycle reduced.

real    0m16.372s
user    0m15.655s
sys     0m0.035s

So It's 16 seconds only for doing divisions for 100M. I think that is big number if we can avoid it.

ARMv7 1Ghz 32bit.


00010428 <main>:
   10428:       e92d 41f0       stmdb   sp!, {r4, r5, r6, r7, r8, lr}
   1042c:       f241 0838       movw    r8, #4152       ; 0x1038
   10430:       f2c0 0802       movt    r8, #2
   10434:       a730            add     r7, pc, #192    ; (adr r7, 104f8 <main+0xd0>)
   10436:       e9d7 6700       ldrd    r6, r7, [r7]
   1043a:       b08a            sub     sp, #40 ; 0x28
   1043c:       2100            movs    r1, #0
   1043e:       f8d8 4000       ldr.w   r4, [r8]
   10442:       a805            add     r0, sp, #20
   10444:       2200            movs    r2, #0
   10446:       2300            movs    r3, #0
   10448:       f44f 4561       mov.w   r5, #57600      ; 0xe100
   1044c:       f2c0 55f5       movt    r5, #1525       ; 0x5f5
   10450:       9409            str     r4, [sp, #36]   ; 0x24
   10452:       460c            mov     r4, r1
   10454:       e9cd 6700       strd    r6, r7, [sp]
   10458:       f240 36e7       movw    r6, #999        ; 0x3e7
   1045c:       e9cd 2302       strd    r2, r3, [sp, #8]
   10460:       f7ff efc4       blx     103ec <gettimeofday@plt>
   10464:       2700            movs    r7, #0
   10466:       e9dd 0100       ldrd    r0, r1, [sp]
   1046a:       f44f 727a       mov.w   r2, #1000       ; 0x3e8
   1046e:       2300            movs    r3, #0
   10470:       3401            adds    r4, #1
   10472:       1980            adds    r0, r0, r6
   10474:       4179            adcs    r1, r7
   10476:       f000 f8b3       bl      105e0 <__aeabi_uldivmod>
   1047a:       42ac            cmp     r4, r5
   1047c:       e9cd 0102       strd    r0, r1, [sp, #8]
   10480:       d1f1            bne.n   10466 <main+0x3e>
   10482:       2100            movs    r1, #0
   10484:       a807            add     r0, sp, #28
   10486:       f7ff efb2       blx     103ec <gettimeofday@plt>
   1048a:       9a06            ldr     r2, [sp, #24]
   1048c:       9b08            ldr     r3, [sp, #32]
   1048e:       f240 717c       movw    r1, #1916       ; 0x77c
   10492:       ed9f 4b15       vldr    d4, [pc, #84]   ; 104e8 <main+0xc0>
10496:       f2c0 0101       movt    r1, #1
   1049a:       1a9b            subs    r3, r3, r2
   1049c:       9a05            ldr     r2, [sp, #20]
   1049e:       2001            movs    r0, #1
   104a0:       ee07 3a10       vmov    s14, r3
   104a4:       9b07            ldr     r3, [sp, #28]
   104a6:       ed9f 5b12       vldr    d5, [pc, #72]   ; 104f0 <main+0xc8>
   104aa:       1a9b            subs    r3, r3, r2
   104ac:       eeb8 7bc7       vcvt.f64.s32    d7, s14
   104b0:       ee87 6b04       vdiv.f64        d6, d7, d4
   104b4:       ee07 3a90       vmov    s15, r3
   104b8:       eeb8 7be7       vcvt.f64.s32    d7, s15
   104bc:       ee36 7b07       vadd.f64        d7, d6, d7
   104c0:       ee85 6b07       vdiv.f64        d6, d5, d7
   104c4:       ec53 2b16       vmov    r2, r3, d6
   104c8:       f7ff efa2       blx     10410 <__printf_chk@plt>
```

jmtilli:
Okay, may be that ARM is slower than AMD64. Did you compile using -O2 because at default compiler settings the division by invariant integer optimization may not be applied? But you are right that we should avoid it, and that's why the newest patch does the division only for mq_tmo() variant after checking all queues for packets. The basic tmo() single-queue variant however cannot use this kind of approach, because we call the implementation function directly if it's there.

Do you think I should add back the "#if SLEEP_NSEC == 1000" preprocessor statement to avoid the division in the most common case, i.e. the user hasn't tinkered with SLEEP_NSEC? Matias Elo instructed me to remove this preprocessor statement, but I can add it back if needed.

muvarov:
I tried both -O2 and -O3, no big speed up on that case. 

nanosleep(&ts, NULL) in odp_pktin_recv_tmo() I think has to be completely removed. In your case select will take care about giving some time to cpu when there is no packets.

In the code:
uint64_t wait_time = odp_pktin_wait_time(ODP_TIME_SEC_IN_NS);
odp_pktin_recv_tmo(global.if0in, pkt_tbl, MAX_PKT_BURST, wait_time);

odp_pktin_wait_time() converts nanoseconds to some uint64_t value which is specific for platform. I.e. return code is not nanoseconds, it's some platform internal representation.  Unfortunately  odp_pktin_wait_time() does not have pktio input parameter were you could do all calculations for specific pktio. 

If we assume that nsecs for timeouts for linux-generic is too exact values. Then we can return from odp_pktin_wait_time() usec value which will be directly be used in pktio.

jmtilli:
I'm really surprised gcc on ARM didn't optimize the division away by using multiplication and the trick I explained.

I don't entirely agree that nanosleep should be removed. It may save some CPU resources when using a pktio that does not support select(), such as DPDK. Also, if it should be removed, it should perhaps be removed in a separate pull request, as removing the nanosleep is distinct from what I propose here in this pull request (although vaguely related).

Actually, odp_pktin_wait_time() in linux-generic returns microseconds because SLEEP_NSEC is 1000. Also, I don't believe a pktio input parameter for odp_pktin_wait_time() would be a good idea. What if somebody has one null pktio, another pcap pktio, a third DPDK pktio, a fourth netmap pktio and a fifth socket_mmsg pktio and uses the recv_mq_tmo() function to receive from all pktios? There is a fundamental conflict there, one that can be solved only by returning a general-purpose value from odp_pktin_wait_time(), and the general purpose value is now in microseconds.

The only thing preventing the removal of the first division is now the fact that SLEEP_NSEC is a #define that the programmer can change.

jmtilli:
You are right; I tested the volatile division sample code on Raspberry Pi 3 model B and it's horribly slow! We can only conclude the implementation of gcc on ARM architectures is total crap. It should optimize it to work in a much faster way. The same, by the way, applies to clang on ARM as well. It also doesn't optimize away the division.

jmtilli:
Hey... I noticed something strange. If I do 32-bit division on ARM, it's very fast (132.216 MPPS) whereas 64-bit division is very slow (5.50628 MPPS). I could eliminate the division by checking whether the timeout is at most about 4 billion nanoseconds meaning 4 seconds, and if so, doing a 32-bit division instead of a 64-bit division. However, the code to do the check would slow down things in the case of AMD64. I'm not sure if some ifdefs checking for 32-bit ARM architecture will make sense.

Also, the same applies for the code inside pktio drivers, but since they use microseconds, the limit is 4000 seconds or a bit more than an hour. So I believe that in most cases, inside the pktio drivers the division could be completely optimized away. Also, here the same caveat applies: it would slow down things in the case of AMD64.

apalos:
Can you please paste the assembler part of the division? Since it's a 64bit
it should be handled by __aeabi_uldivmod when the compiler optimizes that
out.

Can you check it uses that?

Regards
Ilias

jmtilli:
@apalos It has already been done by @muvarov. I have nothing to add to the assembler output pasted by @muvarov, so please take a look at the already pasted assembler output. It seems to be using __aeabi_uldivmod, as you suggested.

apalos:
@jmtilli thanks, i somehow managed to miss Maxim's paste. 
Yes that seems correct. 
Linux kernel has it's own div64_() family functions since it's not linking against gcc.
I have no idea if it's any better/faster than the ABI call, but maybe we can have a quick look.

jmtilli:
Also, this doesn't seem to be an ARM issue only. On 32-bit i386 architecture, the division is slow as well, but 32-bit division would be fast. So perhaps the bittiness of the architecture should be tested, and if bittiness is 32, then it should have a branch to check if the timeout value fits in 32 bits.

With the #if SLEEP_NSEC == 1000 trick and by checking for 32 bit value in drivers if the architecture is 32 bits, all sane use cases (timeout less than an hour) should require no slow divisions.

jmtilli:
Here's a program that is interesting to be compiled with both -m32 and -m64 options:

    #include <stdio.h>
    #include <stdint.h>
    #include <sys/time.h>
 
    int main(int argc, char **argv)
    {
      struct timeval timeout;
      volatile uint64_t usecs2 = 123456789;
      int i;
      for (i = 0; i < 1000*1000*1000; i++)
      {
        uint64_t usecs = usecs2;
        timeout.tv_sec = usecs / (1000 * 1000);
        timeout.tv_usec = usecs % (1000 * 1000);
      }
      printf("%lld %lld\n", (long long)timeout.tv_sec, (long long)timeout.tv_usec);
      return 0;
    }

On my Raspberry Pi 3 model B, it takes 9.274 seconds for billion packets, i.e. 107.83 MPPS. That should be plenty fast for us. And this 107.83 MPPS assumes packets are processed in bursts of one packet. If a burst has 10 packets, the performance is ten times that.

The slower division, (x * SLEEP_NSEC + 999) / 1000, can be optimized away if SLEEP_NSEC is 1000, which it is in linux-generic now.

I'll push new patches soon.

muvarov:
why nanosleep(&ts, NULL); is needed at all after your changes?

matiaselo:
> On 28 Dec 2017, at 10:42, muvarov <notifications@github.com> wrote:
> 
> why nanosleep(&ts, NULL); is needed at all after your changes?
> 

Only socket and netmap pktios use the new select() based code path. The rest still use the same old code path.

platform/linux-generic/odp_packet_io.c
matiaselo: Ifdefs are not allowed.  IMO the cleanest fix would be to add a separate commit (before this one) to this patch set, which modifies odp_pktin_wait_time() to always return microseconds and converts SLEEP_NSEC to SLEEP_USEC (and does necessary changes to make everything work).

Or does someone oppose having a minimum sleep time of 1us between pktin receive calls in the old code path?

platform/linux-generic/pktio/netmap.c
```diff
@@ -786,6 +824,100 @@ static int netmap_recv(pktio_entry_t *pktio_entry, int index,
...
+	int ret;
+	int maxfd;
+	fd_set readfds;
+
+	ret = netmap_recv(pktio_entry, index, pkt_table, num);
+	if (ret != 0)
+		return ret;
+
+	/* 32 bit architectures use 32-bit division if possible */
+#if SIZE_MAX < 18446744073709551615ULL
```
matiaselo: No ifdefs allowed.

platform/linux-generic/odp_packet_io.c
jmtilli: Now that's a great idea! If somebody sees one microsecond as too big sleep time, it is always possible to implement a "divisor": multiply the sleep time by e.g. 4, and sleep e.g. 250 nanoseconds. No slow division operation is needed for this "divisor", only multiplication. But I won't implement this divisor now, as SLEEP_NSEC anyway always used to be 1000.

matiaselo: Agreed, the "divisor" can be added later if needed.

jmtilli:
Hi,

Wondering what the status of this is after 5 days of practical inactivity. Can this be merged? Is merging to Tiger Moth LTS realistic? Note this is an implementation change, not an API change, so this shouldn't break any applications. We have an application where the policy is to use LTS version of ODP, and this patch would benefit our application in a major way.

I've done the following things:
- Change SLEEP_NSEC into SLEEP_USEC, as Matias Elo suggested, making it possible to make odp_pktin_wait_time() return always microseconds
- Optimize division by invariant integer on 32-bit architectures, if the timeout is at most 4 billion microseconds, i.e. less than about an hour. This means inside pktios there is no longer an actual assembly language division instruction or worse, a function call to a division function on architectures lacking assembly language division instruction. Instead, the trick in the famous paper "Division by Invariant Integers using Multiplication" is used by the optimizing C compiler.
- Eliminate all uses of preprocessor conditionals, which was proposed by Matias Elo.
- Rebase the code to newest master.

The following code:

    #include <stdio.h>
    #include <stdint.h>
    #include <sys/time.h>
    
    int main(int argc, char **argv)
    {
      volatile struct timeval timeout;
      volatile uint64_t usecs2 = 123456789;
      int i;
      for (i = 0; i < 1000*1000*1000; i++)
      {
        uint64_t usecs = usecs2;
        if (sizeof(size_t) < 8) {
          if (usecs <= UINT32_MAX) {
            timeout.tv_sec = ((uint32_t)usecs) / (1000 * 1000);
            timeout.tv_usec = ((uint32_t)usecs) % (1000 * 1000);
          } else {
            timeout.tv_sec = usecs / (1000 * 1000);
            timeout.tv_usec = usecs % (1000 * 1000);
          }
        } else {
          timeout.tv_sec = usecs / (1000 * 1000);
          timeout.tv_usec = usecs % (1000 * 1000);
        }
      }
      printf("%lld %lld\n", (long long)timeout.tv_sec, (long long)timeout.tv_usec);
      return 0;
    }

...runs at a rate of billion divisions per 10.926 seconds or 91.5 MPPS on my Raspberry Pi 3 Model B. And note this assumes burst size of 1 packet per burst. At 10 packets per burst, it's 915 MPPS! So certainly, the division should no longer be an issue, assuming the timeout is at most about an hour.

Bill-Fischofer-Linaro:
We just tagged Tiger Moth RC1 (v1.17.0.0). This should be doable for RC2, which will be later this month after Maxim returns from his holiday break.

platform/linux-generic/odp_packet_io.c
```diff
@@ -33,9 +33,9 @@
 #include <time.h>
 
 /* Sleep this many nanoseconds between pktin receive calls */
-#define SLEEP_NSEC  1000
+#define SLEEP_USEC  1
```
matiaselo: Comment still says nanoseconds. Should also state that the maximum value is 1 000 000, so that timespec.tv_nsec is used correctly.

platform/linux-generic/odp_packet_io.c
```diff
@@ -1774,9 +1778,9 @@ uint64_t odp_pktin_wait_time(uint64_t nsec)
 	if (nsec == 0)
 		return 0;
 
-	/* number of nanosleep calls rounded up by one, so that
+	/* number of microseconds rounded up by one, so that
 	 * recv_mq_tmo call waits at least 'nsec' nanoseconds. */
```
matiaselo: Still says nanoseconds.

platform/linux-generic/odp_packet_io.c
```diff
@@ -1698,21 +1699,22 @@ int odp_pktin_recv_tmo(odp_pktin_queue_t queue, odp_packet_t packets[], int num,
...
 				t1 = odp_time_sum(odp_time_local(), t);
 			}
 
 			/* Check every SLEEP_CHECK rounds if total wait time
 			 * has been exceeded. */
-			if ((wait & (SLEEP_CHECK - 1)) == 0) {
-				t2 = odp_time_local();
+			for (i = 0; i < SLEEP_USEC; i++) {
+				if ((wait & (SLEEP_CHECK - 1)) == 0) {
+					t2 = odp_time_local();
```
matiaselo: Seems unnecessarily complex. Something like the following could do the same:
```
uint64_t sleep_round = 0;
...
	/* Check every SLEEP_CHECK rounds if total wait time
	 * has been exceeded. */
	if ((sleep_round++ & (SLEEP_CHECK - 1)) == 0) {
		t2 = odp_time_local();

		if (odp_time_cmp(t2, t1) > 0)
			return 0;
	}
	wait = wait > SLEEP_USEC ? wait - SLEEP_USEC : 0;
```

platform/linux-generic/odp_packet_io.c
```diff
@@ -1698,21 +1699,22 @@ int odp_pktin_recv_tmo(odp_pktin_queue_t queue, odp_packet_t packets[], int num,
...
 
-				if (odp_time_cmp(t2, t1) > 0)
-					return 0;
+					if (odp_time_cmp(t2, t1) > 0)
+						return 0;
+				}
+				wait--;
```
matiaselo: Can roll over if SLEEP_USEC > 1.

platform/linux-generic/odp_packet_io.c
```diff
@@ -1750,19 +1752,21 @@ int odp_pktin_recv_mq_tmo(const odp_pktin_queue_t queues[], unsigned num_q,
...
 
-				t = odp_time_local_from_ns(wait * SLEEP_NSEC);
+				t = odp_time_local_from_ns(wait * 1000);
 				started = 1;
 				t1 = odp_time_sum(odp_time_local(), t);
 			}
 
-			if ((wait & (SLEEP_CHECK - 1)) == 0) {
-				t2 = odp_time_local();
+			for (i = 0; i < SLEEP_USEC; i++) {
```
matiaselo: Same comments than to `odp_pktin_recv_tmo()`.

platform/linux-generic/odp_packet_io.c
```diff
@@ -1680,12 +1680,26 @@ int odp_pktin_recv_tmo(odp_pktin_queue_t queue, odp_packet_t packets[], int num,
...
 	ts.tv_nsec = 1000 * SLEEP_USEC;
 
+	entry = get_pktio_entry(queue.pktio);
+	if (entry == NULL) {
+		ODP_DBG("pktio entry %d does not exist\n", queue.pktio);
+		return -1;
+	}
+
+	if (entry->s.ops->recv_tmo && wait != ODP_PKTIN_NO_WAIT) {
+		uint64_t usecs = wait;
```
matiaselo: Not needed.

platform/linux-generic/odp_packet_io.c
```diff
@@ -1730,6 +1744,29 @@ int odp_pktin_recv_mq_tmo(const odp_pktin_queue_t queues[], unsigned num_q,
 	odp_time_t t1, t2;
 	struct timespec ts;
 	int started = 0;
+	uint64_t usecs;
```
matiaselo: Not needed.

platform/linux-generic/pktio/netmap.c
```diff
@@ -786,6 +824,100 @@ static int netmap_recv(pktio_entry_t *pktio_entry, int index,
...
+	fd_set readfds;
+
+	ret = netmap_recv(pktio_entry, index, pkt_table, num);
+	if (ret != 0)
+		return ret;
+
+	/* 32 bit architectures use 32-bit division if possible */
+	if (sizeof(size_t) < 8) {
+		if (usecs <= UINT32_MAX) {
+			timeout.tv_sec = ((uint32_t)usecs) / (1000 * 1000);
```
matiaselo: To me the value of this optimization seems questionable. System calls are used here anyway, so one division operation (next  comment) shouldn't have that much effect.

platform/linux-generic/pktio/netmap.c
```diff
@@ -786,6 +824,100 @@ static int netmap_recv(pktio_entry_t *pktio_entry, int index,
...
+			timeout.tv_usec = ((uint32_t)usecs) % (1000 * 1000);
+		} else {
+			timeout.tv_sec = usecs / (1000 * 1000);
+			timeout.tv_usec = usecs % (1000 * 1000);
+		}
+	} else {
+		timeout.tv_sec = usecs / (1000 * 1000);
+		timeout.tv_usec = usecs % (1000 * 1000);
```
matiaselo: The second division can be avoided.
```
timeout.tv_usec = usecs - (timeout.tv_sec * 1000000);
```

platform/linux-generic/odp_packet_io.c
jmtilli: Will fix both problems.

platform/linux-generic/odp_packet_io.c
jmtilli: I'll have to think about this a bit more, but your idea seems valid.

platform/linux-generic/odp_packet_io.c
jmtilli: D'oh! That is a serious problem and definitely needs to be fixed.

platform/linux-generic/odp_packet_io.c
jmtilli: Yep, a historical remnant.

platform/linux-generic/odp_packet_io.c
jmtilli: Same here.

platform/linux-generic/pktio/netmap.c
jmtilli: Have you tried 64-bit division on 32-bit ARM? It's really slow! I was surprised how slow it is until I tested it on my Raspberry Pi. ARM does not have a 32-bit division... But perhaps your next comment will make the optimization a bit less necessary and it thus could be removed.

platform/linux-generic/pktio/netmap.c
jmtilli: Strange how I didn't realize this... But this definitely will be in the next patch version!

platform/linux-generic/pktio/netmap.c
jmtilli: Some measurements: empty select() with zero timeout operates at 1.2 M ops / second, 64-bit division operates at 8.8 M ops / second on my Raspberry Pi 3 Model B with the optimization in your next comment. The problem is that in the current code the division is done quite early, but I think I could restructure it so that the division is done only whenever absolutely needed for select(). So yes, I think you may be right.

matiaselo: Looks good now. With more modern/bigger CPUs the division should be faster. Just for curiosity, If you share your benchmark code I can run it on A72.

jmtilli: Sure, https://github.com/jmtilli/divbenchmark

It gives me this on my Raspberry Pi 3 Model B running on 32-bit mode:

    $ make unit
    div
    123 456789
    1.14user 0.00system 0:01.15elapsed 99%CPU (0avgtext+0avgdata 1044maxresident)k
    0inputs+0outputs (0major+73minor)pagefaults 0swaps
    divfast
    123 456789
    0.13user 0.00system 0:00.13elapsed 96%CPU (0avgtext+0avgdata 1092maxresident)k
    0inputs+0outputs (0major+74minor)pagefaults 0swaps
    select
    2.40user 6.60system 0:09.00elapsed 99%CPU (0avgtext+0avgdata 752maxresident)k
    0inputs+0outputs (0major+69minor)pagefaults 0swaps

matiaselo:
> 
> 
> Sure, https://github.com/jmtilli/divbenchmark
> 
> It gives me this on my Raspberry Pi 3 Model B running on 32-bit mode:
> 
> $ make unit
> div
> 123 456789
> 1.14user 0.00system 0:01.15elapsed 99%CPU (0avgtext+0avgdata 1044maxresident)k
> 0inputs+0outputs (0major+73minor)pagefaults 0swaps
> divfast
> 123 456789
> 0.13user 0.00system 0:00.13elapsed 96%CPU (0avgtext+0avgdata 1092maxresident)k
> 0inputs+0outputs (0major+74minor)pagefaults 0swaps
> select
> 2.40user 6.60system 0:09.00elapsed 99%CPU (0avgtext+0avgdata 752maxresident)k
> 0inputs+0outputs (0major+69minor)pagefaults 0swaps
> 


Thanks! This is what I get with MACCHIATObin (1.3GHz A72, gcc (Ubuntu/Linaro 5.4.0-6ubuntu1~16.04.5)). The numbers are pretty much the same for both tests (some variation between test runs).

# taskset 3 make unit
div
123 456789
0.06user 0.00system 0:00.06elapsed 98%CPU (0avgtext+0avgdata 1088maxresident)k
0inputs+0outputs (0major+68minor)pagefaults 0swaps
divfast
123 456789
0.06user 0.00system 0:00.06elapsed 100%CPU (0avgtext+0avgdata 1024maxresident)k
0inputs+0outputs (0major+66minor)pagefaults 0swaps
select
1.83user 3.90system 0:05.73elapsed 99%CPU (0avgtext+0avgdata 896maxresident)k
0inputs+0outputs (0major+63minor)pagefaults 0swaps


A longer test run:
	-  for (i = 0; i < 10*1000*1000; i++)
	+  for (i = 0; i < 10*1000*1000*1000; i++)

# taskset 3 make unit
div
123 456789
8.54user 0.00system 0:08.54elapsed 99%CPU (0avgtext+0avgdata 1016maxresident)k
0inputs+0outputs (0major+66minor)pagefaults 0swaps
divfast
123 456789
8.59user 0.00system 0:08.59elapsed 99%CPU (0avgtext+0avgdata 1104maxresident)k
0inputs+0outputs (0major+67minor)pagefaults 0swaps

jmtilli:
<blockquote>
Thanks! This is what I get with MACCHIATObin (1.3GHz A72, gcc (Ubuntu/Linaro 5.4.0-6ubuntu1~16.04.5)). The numbers are pretty much the same for both tests (some variation between test runs).
</blockquote>

Seems you're using 64-bit ARM operating system, giving it a lot of advantage. The compiler optimizes the 64-bit division away if 64-bit operations are supported.

My RasPi, that has 64-bit CPU, runs a 32-bit operating system.

Bill-Fischofer-Linaro:
@matiaselo Please add your review if you're now OK with this PR.

muvarov:
Merged.

PR #342, opened 2017-12-14
[PATCH v18] OpenSSL ctx rework

lumag:
@JannePeltonen

This is the promised OpenSSL CTX rework, bringing significant crypto speedup


lumag:
@JannePeltonen I partially implemented your suggestion. Now I see more than 60% speed increase for aes-gcm with 1024-byte packets and more than 53% for sha1-96 (also 1024-byte packets).

JannePeltonen:
@lumag That speedup sounds similar to what I saw in my tests (with AES-CBC + SHA1). Maybe the intermediate approach would be faster if there would be a separate ready-made context for every algorithm so that the cipher specific init would not be needed for every packet (but key config would still be needed). OTOH, that makes the code more complicated.

lumag:
@JannePeltonen for AES-CBC+SHA1 I got only 34% speedup.

Another option would be to use `EVP_CIPHER_CTX_copy()`.

lumag:
Hmm. Using `_copy()` would significantly complicate code, as context will need to be allocated from shared memory. I think this version is good enough for now.

lumag:
Fixing HMAC reinit in both PRs.

platform/linux-generic/odp_crypto.c
```diff
@@ -112,8 +112,18 @@ struct odp_crypto_generic_session_t {
...
+	uint32_t cipher_valid[(ODP_THREAD_COUNT_MAX + 32 - 1) / 32];
+	uint32_t hmac_valid[(ODP_THREAD_COUNT_MAX + 32 - 1) / 32];
+	unsigned idx;
 };
 
+#define IS_VALID(session, type, id)	\
+	(session->type ## _valid[id / 32] & (1 << (id % 32)))
+
+#define SET_VALID(session, type, id)	\
+	session->type ## _valid[id / 32] |= (1 << (id % 32))
```
muvarov: is 32 everywhere MAX_SESSIONS?

platform/linux-generic/odp_crypto.c
```diff
@@ -1060,6 +1081,41 @@ int odp_crypto_term_global(void)
 	return rc;
 }
 
+int odp_crypto_init_local(void)
```
muvarov: `_odp_crypto_init_local`

platform/linux-generic/odp_crypto.c
```diff
@@ -1060,6 +1081,41 @@ int odp_crypto_term_global(void)
...
+			return -1;
+		}
+	}
+
+	local.id = odp_thread_id();
+
+	return 0;
+}
+
+int odp_crypto_term_local(void)
```
muvarov: `_odp_crypto_term_local`

platform/linux-generic/odp_crypto.c
lumag: No, `sizeof uint32_t`

platform/linux-generic/odp_crypto.c
Bill-Fischofer-Linaro: I don't see the need to do that at this point as these are internal routines anyway. That type of restructure might be something to consider for Caterpillar.

lumag: Then I would prefer to use names w/o underscore for now and to change all of them later.

Bill-Fischofer-Linaro: That makes sense to me.

muvarov: I prefer the rule that new code goes in in clean way. Even if rest of the code used old rules.

Bill-Fischofer-Linaro: Then I'd follow "Maxim's rule" and do the renames as a separate follow-on PR.

lumag:
Fixed both issues

platform/linux-generic/odp_crypto.c
```diff
@@ -146,6 +171,8 @@ odp_crypto_generic_session_t *alloc_session(void)
 	}
 	odp_spinlock_unlock(&global->lock);
 
+	session->idx = (session - global->sessions) / sizeof(*session);
+
```
JannePeltonen: The cipher_valid and hmac_valid bits are not cleared at session allocation, so they can be left to what they were in an old and freed session that occupied the same memory slot.

platform/linux-generic/odp_crypto.c
```diff
@@ -165,24 +192,51 @@ null_crypto_routine(odp_crypto_op_param_t *param ODP_UNUSED,
...
+		SET_VALID(session, hmac, local.id);
+		HMAC_Init_ex(ctx,
+			     session->auth.key,
+			     session->auth.key_length,
+			     session->auth.evp_md,
+			     NULL);
+	} else {
+		HMAC_Init_ex(ctx,
+			     NULL,
+			     session->auth.key_length,
```
JannePeltonen: Why is this needed here? The key length was already set. If this is because of some OpenSSL weirdness, an explanatory comment would be nice.

JannePeltonen:
Now up to MAX_SESSIONS times thread count OpenSSL sessions may get created and kept until ODP termination. I guess this is not a problem now that MAX_SESSIONS is 32, but what is ODP needs to start supporting e.g. a bigger number of IPsec SAs (like 1000)? Maybe this approach would need to be revisited then.

platform/linux-generic/odp_crypto.c
lumag: Whole `odp_crypto_generic_session_t` data is cleared during `odp_crypto_session_destroy()`

platform/linux-generic/odp_crypto.c
lumag: OK, I will add a comment. This is necessary to restart (reset) HMAC calculation without resetting the key and auth length.

platform/linux-generic/odp_crypto.c
JannePeltonen: ok, I can see it now. I think it would be less fragile to clear the session at alloc() instead of relying it to be cleared at init_global() and destroy().

platform/linux-generic/odp_crypto.c
JannePeltonen: Yes, the hmac reset of course (what was I thinking?). But for that this should be enough (I have used it in my test patch with OpenSSL 1.1.0 and 1.0.2): HMAC_Init_ex(ctx, NULL, 0, NULL, NULL) and be more clear in that key len is not being changed.

Looking at the man page I do not think HMAC_Init_Ex() is going to even look at key_len when key is NULL.

platform/linux-generic/odp_crypto.c
```diff
@@ -112,8 +115,22 @@ struct odp_crypto_generic_session_t {
...
+	crypto_valid hmac_valid[(ODP_THREAD_COUNT_MAX + CV_SIZE - 1) /
+				CV_SIZE];
+	unsigned idx;
 };
 
+#define IS_VALID(session, type, id)	\
+	(session->type ## _valid[id / CV_SIZE] & (1 << (id % CV_SIZE)))
+
+#define SET_VALID(session, type, id)	\
+	session->type ## _valid[id / CV_SIZE] |= (1 << (id % CV_SIZE))
```
Bill-Fischofer-Linaro: Checkpatch wants this enclosed in parentheses. Since `IS_VALID` follows that model, no harm in doing so with `SET_VALID` as well.
```
ERROR: Macros with complex values should be enclosed in parentheses
#63: FILE: platform/linux-generic/odp_crypto.c:131:
+#define SET_VALID(session, type, id)	\
+	session->type ## _valid[id / CV_SIZE] |= (1 << (id % CV_SIZE))
```

lumag:
Updated both PRs:
- rewrote to use atomics
- use 0 for key length in HMAC reinit case.

platform/linux-generic/odp_crypto.c
JannePeltonen: I did not say the keys should not be cleared at destroy() but that it would be better (as in, more difficult to introduce a bug later) if the alloc did not rely on the clearing of the valid bits etc having happened earlier (in destroy or in global init). But this is not a sticking point for me.

Btw, if you are concerned about storing keys in memory for too long, then this patch does just that by keeping the openssl contexts alive until local_term().

platform/linux-generic/odp_crypto.c
```diff
@@ -112,8 +116,32 @@ struct odp_crypto_generic_session_t {
...
 
+/* Use _mm versions, because they are always inlined */
+static int crypto_should_init(crypto_valid *ptr, unsigned int id)
+{
+	unsigned int mask = 1 << (id % CV_BITS);
+	crypto_valid *cv = ptr + (id / CV_BITS);
+	uint32_t cur = _odp_atomic_u32_load_mm(ptr, _ODP_MEMMODEL_ACQ);
+
+
+	while (!_odp_atomic_u32_cmp_xchg_strong_mm(cv, &cur, cur | mask,
```
JannePeltonen: I think the atomic cmp-xchg is too heavyweight here.

AFAICS, it is needed only because of the way the valid bits for different threads are packed in the same memory locations. The need for the syncing atomic op could be avoided by just using a whole uint8_t for one thread. I do not think the increase in memory use would be so big that it would be a problem.

If you really want to keep packing the valid bits, then you could do so also by having per-thread arrays of valid bits for all sessions in the global SHM instead of the current approach of per-session arrays of valid bits for all threads. That would avoid the syncing needs since the valid bits would not share memory locations with other threads.

IOW, you would have something like this in the global SHM:

struct openssl_ctx_status {
	crypto_valid cipher_valid[(MAX_SESSIONS + CV_BITS - 1) / CV_BITS];
	crypto_valid   hmac_valid[(MAX_SESSIONS + CV_BITS - 1) / CV_BITS];
} [ODP_THREAD_COUNT_MAX];

lumag: Hmm. I like this idea. I will implement it, thank you.

platform/linux-generic/odp_crypto.c
lumag: Ok, I will add some cleaning to _create stage. At lease it is not a fast path operation, so performance should not matter that much.

platform/linux-generic/odp_crypto.c
lumag: @JannePeltonen even with your suggestion there can be congestion between set-as-valid session1 and clear-valid session2 (which can happen in completely different thread). In fact I'm balancing between two possible solutions:

- always init all contexts on session init/destroy them on session destroy
- just add a per-thread lock guarding writing to per-thread bits

Any suggestions?

lumag:
@JannePeltonen updated this implementation to use per-thread rather than per-session bitfields. Also replaced bare atomic operations with Ola's bitsets.

lumag:
@JannePeltonen added another patch on top, to simplify init code. Before/when merging it can be squashed into previous commit.

platform/linux-generic/odp_crypto.c
```diff
@@ -137,15 +139,56 @@ struct odp_crypto_global_s {
 	odp_spinlock_t                lock;
 	odp_crypto_generic_session_t *free;
 	odp_crypto_generic_session_t  sessions[MAX_SESSIONS];
+
+	/* These bitfields are cleared at alloc_session()
+	 * together with the rest of data */
+	bitset_t hmac_valid[ODP_THREAD_COUNT_MAX]
```
JannePeltonen: I think the benefits of using a bitset may not outweigh the drawbacks. In addition to making the code slightly more complicated, using a bitset forces us to use atomic operations (since, as you pointed out, we still have to handle the case where thread A allocates session X while thread B is initializing session Y in the same bit set, both updating the same memory location, although different bits).

By using a full byte for the validity flag would simplify the code and avoid the need for atomics since different threads would no longer write and read the same memory location at the same time (there must be some synchronization in the application side to ensure that a session is not destroyed when still in use and that creation of a session happens before its use).

The atomics can slow down ARM because the bitset tries to provide atomicity over a 16 byte range, which causes additional overhead (e.g. reading a valid bit is not just a plain memory load) even though not required in this context. A weaker atomic bitset would suffice, but I think we could do without atomics altogether.

A benefit of the bit set is reduced memory use (and possibly some improvement in memory access locality), but I doubt it is significant enough. The savings are less than MAX_SESSIONS * ODP_THREAD_COUNT_MAX bytes (after combining the crypto/auth valid bits as in the last path). At the same time the thread local crypto_local_t structs are using 12 or 24 times MAX_SESSIONS bytes per thread, depending on pointer size, plus the memory reserved by OpenSSL for all the pre-created contexts.

So, how about just using a full byte for the validity flag?

platform/linux-generic/odp_crypto.c
```diff
@@ -137,15 +139,56 @@ struct odp_crypto_global_s {
...
+
 	odp_ticketlock_t              openssl_lock[0];
 };
 
 static odp_crypto_global_t *global;
 
+typedef struct crypto_local_t {
+	HMAC_CTX *hmac_ctx[MAX_SESSIONS];
+	EVP_CIPHER_CTX *cipher_ctx[MAX_SESSIONS];
+	EVP_CIPHER_CTX *mac_cipher_ctx[MAX_SESSIONS];
```
JannePeltonen: This does not seem to be needed. There can be only cipher and mac in a session.

lumag: Consider a case of encryption + AES-GMAC.

platform/linux-generic/odp_crypto.c
lumag: I feel uneasy for propagation of clear flags to other threads. What about using `atomic_flag` and `atomic_flag_test_and_set`/`atomic_flag_clear` from stdatomics?

JannePeltonen: After some more pondering I still think additional synchronization for the valid flag is not necessary.

When an application creates a crypto session and gets a handle back, the side effects (memory stores) of the creation are not yet necessarily visible in other threads (according to Petri, this is an intended property of the ODP API). If the application passes the handle to another thread it must use some synchronization mechanism to ensure that creation is fully visible in the other thread before it tries to use the handle. The valid flag is just one more item in shared memory that needs to be visible before a crypto operation is started. Right?

Similarly, when a crypto session is destroyed, the crypto operations must have fully completed before session destroy can be called in another thread. This again requires some synchronization by the application between the threads, but that also ensures that setting of the valid bit inside a crypto operation happens-before session destroy which happens-before (due to the locking in session alloc and free) the storage of the session is reused which happens-before clearing the valid bit. So this case should also be ok, right?

And if the valid flags are all in their own memory locations (i.e. not bitfields of the same location), then there are no other inter-thread synchronization concerns regarding the valid flag.

But this requires more than just my word. If you cannot get convinced about not needing additional synchronization for the valid flag, then you could also consider memory barriers after each write to the valid flag instead.

If you want to stick with atomic ops, then I am not sure if using stdatomic is ok as it seems to create an additional dependency to a tool chain that supports it. Maybe ODP's own atomic primitives should be used.

I do not see why test-and-set would be needed.

platform/linux-generic/odp_crypto.c
JannePeltonen: ok, now I can see it. Ignore the previous comment.

platform/linux-generic/odp_crypto.c
lumag: @JannePeltonen this sounds logical, but not 101% convincing, sorry. When we are talking about security, I'll try to be on a safe side. Maybe I have to give it another thought.

The rest of linux-generic code heavily uses atomic operation (even worse than stdatomic, it uses GCC builtins). We do not have (even internal) API for memory barriers, so I assumed that atomic flags are best fit for this case (and each flags uses just one byte).

JannePeltonen: Is there a particular sequence that worries you or is it just general uneasiness?

Maybe you should spin a new version of the patch set, with the last patch squashed into the third one and with the byte-wide atomic flag replacing the bitset_t?

I think the implementation using the atomics may end up doing the same assumptions as the approach I proposed but it would be easier to discuss with a concrete code at hand (= I will not quite yet stop suggesting an approach with minimal atomic and synchronization operations).

lumag: Initially I kept this patch separate, as a way to demonstrate the change. Squashed now into third patch. Is a general uneasiness about cache propagation of flag clear in `alloc_session` and flag set in `crypto_init`. E.g. set-clear-set done in fast enough way. Maybe I'm wrong and flags will propagate in such case.

lumag: Maybe @WonderfulVoid can comment on this issue.

WonderfulVoid:
On Mon, Jan 15, 2018 at 1:57 PM, Dmitry Eremin-Solenikov <
notifications@github.com> wrote:

> *@lumag* commented on this pull request.
> ------------------------------
>
> In platform/linux-generic/odp_crypto.c
> <https://github.com/Linaro/odp/pull/342#discussion_r161510284>:
>
> > @@ -137,15 +139,56 @@ struct odp_crypto_global_s {
>  	odp_spinlock_t                lock;
>  	odp_crypto_generic_session_t *free;
>  	odp_crypto_generic_session_t  sessions[MAX_SESSIONS];
> +
> +	/* These bitfields are cleared at alloc_session()
> +	 * together with the rest of data */
> +	bitset_t hmac_valid[ODP_THREAD_COUNT_MAX]
>
> Maybe @WonderfulVoid <https://github.com/wonderfulvoid> can comment on
> this issue.
>
I haven't been involved with the ODP crypto support so know very little
about this. What's the issue here?

lumag:
@WonderfulVoid We were discussing on the primitives necessary to synchronize context setting/clearing.
There is a per-thread context, which is initialized on demand when crypto function gots first scheduled for the crypto session in this particular thread. `alloc_session()` function marks contexts as not initialized. Main question now is in synchro routines that we have to use. Currently I used your bitsets to save memory. @JannePeltonen suggests using plain uint8_t flags w/o any additional syncrho between crypto function and session allocation. I'm a bit unsettled that this will always be safe and advocate for atomic_flag routines.

JannePeltonen:
@WonderfulVoid My argument for not needing additional sync is that the user of the crypto API must ensure (i.e. use suitable synchronization mechanisms) that odp_crypto_session_create() and thus alloc_session() happens before the returned session handle is passed to odp_crypto_op() (by possibly some other thread). So when odp_crypto_op() reads the valid bit, the store to it by alloc_session() must already be visible. This requires that different valid bits do not share memory locations like they would in a bit set.

For example:

atomic g_handle = NULL

Thread_A()
{
	handle = odp_crypto_session_create()  // clears the valid bit
	atomic_store_rel(&g_handle, handle)
}

Thread_B()
{
	handle = atomic_load_acq(g_handle)
	if (handle)
		odp_crypto_op(handle, ...)  // reads and sets the valid bit
}

Here the release-acquire sequence ensures that by the time odp_crypto_op() is called, everything done in odp_crypto_session_create() is visible. The synchronization is needed even without the added valid bit since all stores to SHM variables in session creation must be complete before thread_B tries to use them.

And similar argument for session destroy (odp_crypto_op() must be fully complete before odp_crypto_session_destroy() can be called).

It would be useful if someone else could sanity check the above.

WonderfulVoid:
@JannePeltonen your example seems correct. If two threads share such a data
structure (e.g. passed through a shared pointer but could also be
transferred in an enqueued/dequeued event etc), Thread_A needs to store the
data pointer using store-release and Thread-B needs to load the data
pointer using load-acquire. If using an event to transfer the pointer, the
enqueue operatio will have release ordering and the dequeue operation will
have acquire ordering.

It is probably not a good idea to put all the valid bits together in a
separate location. You would need an atomic operation (e.g.
__atomic_fetch_and/__atomic_fetch_or) to operate on that location. Atomic
RMW operations are normally much more expensive than normal load+store. And
you are potentially creating write contention for this shared bitset, this
is bad for scalability.

I was worried about this data structure being passed to and accessed by HW
(which is normally controlled using "device accesses"). Then we will
probably need a stricter barrier, e.g. DSB on ARM, because device accesses
(on ARM) are not ordered with the normal memory DMB or release ordering.

lumag:
@WonderfulVoid what about data that we store inside `odp_crypto_session_create()` and reuse during `odp_crypto_op()`. I have a flag that I want to clear during create and set during crypto_op. Should I use atomic_flag from stdatomic, `_odp_atomic_flag_t` from ODP, which enforces ACQ for set and REL for clear or we can just use plain byte variables, which will be propagated to other threads because of external fence?

WonderfulVoid:
If the reference (e.g. a pointer) to the crypto session is passed using
some mechanism that already provides release/acquire ordering (e.g.
enqueuing/dequeueing an event on/from a queue), then you don't need
additional ordering.
If the reference to the crypto session is passed through e.g. a shared
pointer, then "threadA" (the sender) uses store-release (e.g.
__atomic_store_n(&global, session_ptr, __ATOMIC_RELEASE)) and "threadB"
(the receiver) uses load-acquire (e.g. session_ptr =
__atomic_load_n(&global, __ATOMIC_ACQUIRE)).
If the receiver already knows the pointer to the crypto session and just
polls some valid bit (which can be located separate from the crypto session
data structure), then the sender needs set this bit using release ordering
and the receiver needs to poll (read) the bit using acquire ordering.

lumag:
@WonderfulVoid thanks for your feedback.
I have updated code to use `_odp_atomic_flag_t`, so it should be safe in all cases.

JannePeltonen:
What are the cases that would not be safe without using the atomic ops for the valid flag? Why is atomic needed? Why relaxed memory ordering is not enough but a synchronizing atomic is needed? Why TAS as opposed to separate load and store?

I have tried to explain why I do not think the atomic syncing operation is needed here. Maybe you could try to explain what could go wrong with plain normal memory accesses?

I do not like the atomics here not only because they cause overhead, but, more importantly, because I do not think we should add code constructs without fully understanding their purpose. Adding an atomic or syncing op that is not needed and for which the design is not clear can confuse the reader of the code and make future code changes harder.

WonderfulVoid:
I agree with @JannePeltonen
In normal situations (my #1 and #2), the necessary ordering is already
provided outside of the crypto session abstract data type.
If there is any situation similar to #3 (what would that be?),
synchronisation could be done separately, "producer" and "consumer"
application code could synchronise using some user-level variable (accessed
using load-acquire and store-release). I don't think synchronisation has to
be part of the crypto session data type itself. We need a clear description
of why you (Dmitry) think this is needed.

-- Ola

lumag:
@WonderfulVoid mostly because in API we have no additional requirements on `session_create`/`crypto_op`. If you think that our current API enforces a memory fence between 'producer' and 'consumer', I will replace atomic flags with bare bytes (@JannePeltonen sorry, I preferred to be on a safe side).

JannePeltonen:
The current API requires that the application ensures that there is proper memory ordering between session creation and crypto op, otherwise the implementation may fail in mysterious ways. The API or the ODP implementation cannot enforce that the application does the right thing. This is not specific to the new valid bit but also all other writes to shared memory in session creation must be visible and happen-before crypto operation for the same session is started in another thread.

Maybe @psavol could comment on ODP API principles regarding resource creation and concurrency?

psavol:
1) Thread A creates a session. The creation call writes HW registers and memory locations. It's implementation specific if HW register writes needs extra barriers. If extra HW register barriers are needed, those are done in the create call. Thread A gets a handle.
2A) A calls an ODP mechanism to give handle to thread B. API calls guarantee that all memory writes from A (including writes inside session creation) are visible to B. Thread B gets handle.
2B) Application invented a new sync mechanism. A passes handle to B with that. The new thing synchronizes all memory writes from thread A to B (as any multi-thread C program would need to do). Thread B gets handle. 
3) Thread B uses handle. Memory and HW register reads from B (or HW) see valid data written by A.

JannePeltonen:
Thanks petri. That is what I have been trying to say all the time. The application must already guarantee that odp_crypto_session_create() happens-before (in C11 sense) odp_crypto_operation() for the same session. So if odp_crypto_operation() is done in a different thread from the creation, there needs to be proper synchronization (e.g. a release-acquire sequence) in place when passing the session handle between threads.

ODP does not do this synchronization for the application, except when the application passes the handle using an ODP provided mechanism, like a queue, that performs such synchronization. If the application just stores a newly created handle to a memory location and polls it in another thread using relaxed memory order and then uses the handle, undefined behavior will result.

lumag:
@JannePeltonen @psavol @WonderfulVoid ok, thanks for your comments. Updated to use plain uint8_t.

Bill-Fischofer-Linaro:
@JannePeltonen @WonderfulVoid @psavol Please add your review to confirm that all of your comments have been addressed.

muvarov:
@JannePeltonen @WonderfulVoid @psavol ping.

JannePeltonen:
This (v18) looks ok to me now.

muvarov:
@WonderfulVoid @psavol going to merge PR. Last call for objections.

WonderfulVoid:
I don't understand the purpose of ctx_valid and crypto_init(). There are no comments.

Is crypto_init() something we need to do for each thread whenever it first uses a specific crypto session? What sort of crypto setup is specific to a thread? Isn't all configuration and state etc stored in the shared crypto session (in shared memory)?

lumag:
@WonderfulVoid actually no. OpenSSL 1.1.x lacks ability to use preallocated structures as contexts (they switched to completely opaque structures). Thus each thread has its own set of contexts. Context is initialized on first use with data stored in session params.

WonderfulVoid:
So the context (ctx) refers to the OpenSSL contexts (one per thread) which used with the shared ODP crypto sessions.

I think I understand.

Please add comments about this so that the next reader can understand without asking.

lumag:
OK

WonderfulVoid:
The valid flags (shouldn't they be called "init_done" or similar?) are read and written by the same thread (after crypto session initialisation where the flag for all threads is reset)? Then a non-atomic bit set can be used. Possibly (if there is space in the crypto session), the bit set can be located inside the crypto session, avoiding accessing a different cache line. 128 threads => 128 bits == 16 bytes.

lumag:
@WonderfulVoid no, they are written by the thread which scheduled crypto operation. We can not init all contexts during creation because we lack inter-process generic memory allocator (for cases when ODP thread is a separate process).

lumag:
I will extend code with comments, so that it would be easier to understand

WonderfulVoid:
In crypto_init(), a thread is just testing and setting its own flag.
In alloc_session(), the creator thread resets all flags but this is before the session can be accessed by other threads.
Where is a thread modifying these flags for other threads?

lumag:
In `session_create()` only. I initially did an implementation using your bitsets, then plain atomic flags, but in the end rewrote using bytes according to @JannePeltonen request.

JannePeltonen:
Using a bitset would mean that the flags for different sessions would share the same memory location. Since one thread could be allocing a session when another thread is using an adjacent session, an atomic bitset would have to be used.

WonderfulVoid:
+	uint8_t ctx_valid[ODP_THREAD_COUNT_MAX][MAX_SESSIONS];
So there is a separate flag for each thread for each session.
If we combine the flags for all sessions for a specific thread, only that thread will read and write those flags (after session initialisation).
The accesses in crypto_init() only access per-thread flags (local.ctx_valid initialised by local.ctx_valid = global->ctx_valid[id]; where id is odp_thread_id().)

If the access pattern is more complicated than this, we really need some more documentation (comments).

JannePeltonen:
Consider the case when thread A creates crypto session #1 (clearing all the per-thread valid flags for that session) at the same time when thread B is using session #2 (reading and writing the per-thread valid flag). If the valid flags were implemented using a bit set, then the thread A and B in the above scenario might be accessing the same memory location (valid flag bitset (containing bits for session #1 and #2) for thread B) at the same time.

WonderfulVoid:
The set of valid flags (e.g. one bit per thread, perhaps 16 bytes in total) for session #1 should all be stored in session #1, separate from session #2 which has its own set of valid flags.
Thread A:
<initialise crypto session1>
session1->valid_flags = 0;
Thread B:
if ((session2->valid_flags & (1U << tidx_B)) == 0)
{
   <init context for thread B>
   session2->valid_flags |= 1U << tidx_B;
}
session1 != session2

JannePeltonen:
If you do that, then different threads for the same session manipulate different bit fields in the same memory location at the same time. Add thread C, similar to thread B (also handling session 2) to your example.

Valid flags of different sessions cannot share a memory location because there is no inter-thread synchronization across different sessions. But in addition to that, valid flags of different threads cannot share a memory location because after session creation multiple threads are processing the session simultaneously, each updating their own valid flag.

WonderfulVoid:
Aha different threads can use the same crypto session at the same time. Things are getting clearer now. I was not aware of all the legal access patterns to the crypto data structures.

Yes then we would need the atomic bitset again. Possibly this is still better than having an array of one byte per thread (per session) which takes 2 64-byte cache lines for 128 threads. The bitset would only be updated once per thread (the first time it uses the crypto session) so it would quickly become sharable without contention.

WonderfulVoid:
With an atomic bitset we can still used relaxed memory order for the accesses (check bit and set bit). A specific bit is only accessed by a specific thread, the bitset is not used for inter-thread synchronisation.

JannePeltonen:
Yes, I agree that load and stores with relaxed memory order would be sufficient in that case.

The bitset_t type does not seem suitable for this since it does not offer single-bit operations but wants to provide atomicity over up to 16 byte set of bits, which can be expensive and the reason I objected the use of that type in a previous version of this patch.

lumag:
@JannePeltonen @WonderfulVoid So, should I switch to atomic access to 8-bit bitset flags? Or should we prefer 32-bit sets? Also is it correct, that atomic_fetch_or/atomic_store(0) is sufficent in this case?

WonderfulVoid:
I don't know if performance is improved by using a per-session bit set for
the threads. Possibly, if the footprint of the session structure doesn't
increase (in cache lines). Then we should avoid an access to a separate
cache line just to check the "valid" (init) flag. Probably small benchmarks
won't show any improvement, the working set more or less fits into the
cache.

tidx / 8 => byte index, tidx % 8 => bit in byte
Reads and writes only have to access the specific byte. We don't need to
atomically read the whole bit set.
Only setting the bit needs to be done atomically, this can be done using
e.g. __atomic_fetch_or_n(&session->bitset[byte], 1U << bit,
__ATOMIC_RELAXED);

JannePeltonen:
I do not have a strong opinion on how much the memory layout of the valid bits should or should not be optimized. As long as at most relaxed atomic loads and stores (i.e. no actual synchronization operations) get used, I am fine with either way. A big obstacle to further optimization is anyway the OpenSSL API that does not fit this usage pattern very well.

muvarov:
Merged.

PR #343, opened 2017-12-14
[PATCH v3] odp_generator: update RX side

bogdanPricope:
- Add configuration option for RX burst size
- Replace scheduler mode with direct pktin mode

### DIFF
--- a/example/generator/odp_generator.c
+++ b/example/generator/odp_generator.c
@@ -28,7 +28,8 @@
 #define DEFAULT_PKT_INTERVAL   1000  /* Interval between each packet */
 #define DEFAULT_UDP_TX_BURST	16
 #define MAX_UDP_TX_BURST	512
-#define MAX_RX_BURST		32
+#define DEFAULT_RX_BURST	32
+#define MAX_RX_BURST		512
 
 #define APPL_MODE_UDP    0			/**< UDP mode */
 #define APPL_MODE_PING   1			/**< ping mode */
@@ -80,6 +81,8 @@ typedef struct {
 				     each packet */
 	int udp_tx_burst;	/**< number of udp packets to send with one
 				      API call */
+	int rx_burst;	/**< number of packets to receive with one
+				      API call */
 	odp_bool_t csum;	/**< use platform csum support if available */
 } appl_args_t;
 
@@ -130,6 +133,7 @@ typedef struct {
 	/** Global arguments */
 	int thread_cnt;
 	int tx_burst_size;
+	int rx_burst_size;
 } args_t;
 
 /** Global pointer to args */
@@ -828,10 +832,11 @@ static int gen_recv_thread(void *arg)
 	odp_event_t events[MAX_RX_BURST];
 	int pkt_cnt, ev_cnt, i;
 	odp_packet_chksum_status_t csum_status;
+	int burst_size;
 
-	(void)arg;
 	thr = odp_thread_id();
 	thr_args = (thread_args_t *)arg;
+	burst_size = args->rx_burst_size;
 
 	printf("  [%02i] created mode: RECEIVE\n", thr);
 	odp_barrier_wait(&barrier);
@@ -842,7 +847,7 @@ static int gen_recv_thread(void *arg)
 
 		/* Use schedule to get buf from any input queue */
 		ev_cnt = odp_schedule_multi(NULL, ODP_SCHED_NO_WAIT,
-					    events, MAX_RX_BURST);
+					    events, burst_size);
 		if (ev_cnt == 0)
 			continue;
 		for (i = 0, pkt_cnt = 0; i < ev_cnt; i++) {
@@ -1076,12 +1081,16 @@ int main(int argc, char *argv[])
 	args->thread_cnt = num_workers;
 
 	/* Burst size */
-	if (args->appl.mode == APPL_MODE_PING)
+	if (args->appl.mode == APPL_MODE_PING) {
 		args->tx_burst_size = 1;
-	else if (args->appl.mode == APPL_MODE_UDP)
+		args->rx_burst_size = 1;
+	} else if (args->appl.mode == APPL_MODE_UDP) {
 		args->tx_burst_size = args->appl.udp_tx_burst;
-	else
+		args->rx_burst_size = 0;
+	} else {
 		args->tx_burst_size = 0;
+		args->rx_burst_size = args->appl.rx_burst;
+	}
 
 	/* Create packet pool */
 	odp_pool_param_init(&params);
@@ -1383,11 +1392,12 @@ static void parse_args(int argc, char *argv[], appl_args_t *appl_args)
 		{"interval", required_argument, NULL, 'i'},
 		{"help", no_argument, NULL, 'h'},
 		{"udp_tx_burst", required_argument, NULL, 'x'},
+		{"rx_burst", required_argument, NULL, 'r'},
 		{"csum", no_argument, NULL, 'y'},
 		{NULL, 0, NULL, 0}
 	};
 
-	static const char *shortopts = "+I:a:b:s:d:p:i:m:n:t:w:c:x:he:f:y";
+	static const char *shortopts = "+I:a:b:s:d:p:i:m:n:t:w:c:x:he:f:yr:";
 
 	/* let helper collect its own arguments (e.g. --odph_proc) */
 	odph_parse_options(argc, argv, shortopts, longopts);
@@ -1398,6 +1408,7 @@ static void parse_args(int argc, char *argv[], appl_args_t *appl_args)
 	appl_args->timeout = -1;
 	appl_args->interval = DEFAULT_PKT_INTERVAL;
 	appl_args->udp_tx_burst = DEFAULT_UDP_TX_BURST;
+	appl_args->rx_burst = DEFAULT_RX_BURST;
 	appl_args->srcport = 0;
 	appl_args->dstport = 0;
 	appl_args->csum = 0;
@@ -1540,6 +1551,14 @@ static void parse_args(int argc, char *argv[], appl_args_t *appl_args)
 				exit(EXIT_FAILURE);
 			}
 			break;
+		case 'r':
+			appl_args->rx_burst = atoi(optarg);
+			if (appl_args->rx_burst >  MAX_RX_BURST) {
+				EXAMPLE_ERR("wrong Rx burst size (max %d)\n",
+					    MAX_RX_BURST);
+				exit(EXIT_FAILURE);
+			}
+			break;
 
 		case 'y':
 			appl_args->csum = 1;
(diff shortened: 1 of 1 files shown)


bogdanPricope:
https://lng.validation.linaro.org/scheduler/job/23451.0
https://lng.validation.linaro.org/scheduler/job/23451.1

TX 14.8 /  RX  8.8 mpps (one core execution)
It brings 2 mpps.

For better performance on RX, I suspect it needs parser update patches (api-next).

example/generator/odp_generator.c
```diff
@@ -1623,6 +1643,7 @@ static void usage(char *progname)
 	       "  -n, --count the number of packets to be send\n"
 	       "  -c, --cpumask to set on cores\n"
 	       "  -x, --udp_tx_burst size of UDP TX burst\n"
+		   "  -z, --rx_burst size of RX burst\n"
```
muvarov: -r ?

muvarov:
gen_recv_thread() itself looks a little bit buggy. 1) scheduler can return timeout, not only packets. 2) if(err)  {odp_packet_free(); coninue; }  then again odp_packet_free_multi() for all array is not correct. 3) print_pkts() walks to all array one more time. For udp case just needed to increase the counter. Has to be inline.

example/generator/odp_generator.c
```diff
@@ -784,10 +811,33 @@ static void print_pkts(int thr, thread_args_t *thr_args,
...
+
+		if (odp_packet_has_udp(pkt)) {
+			if (itf->config.pktin.bit.udp_chksum) {
+				if (odp_packet_has_l4_error(pkt))
+					printf("HW detected L4 error\n");
+			}
+		}
+
+		/* Drop packets with errors */
+		if (odp_unlikely(odp_packet_has_error(pkt)))
```
muvarov: that has to be the first check.

muvarov:
Ah, ok in patch 2 you completely removed recv with scheduler and comments above are no longer valid. But I think it's better to keep both scheduler and direct recv inside generator. This will be good test tool to check scheduler performance.

example/generator/odp_generator.c
```diff
@@ -517,10 +520,13 @@ static int create_pktio(const char *dev, odp_pool_t pool,
...
 	odp_pktout_queue_param_t pktout_param;
-	odp_pktio_op_mode_t pktout_mode;
+	odp_pktio_op_mode_t pktout_mode, pktin_mode;
 
 	odp_pktio_param_init(&pktio_param);
-	pktio_param.in_mode = ODP_PKTIN_MODE_SCHED;
+	pktio_param.in_mode = num_rx_queues ? ODP_PKTIN_MODE_DIRECT :
+		ODP_PKTIN_MODE_DISABLED;
+	pktio_param.out_mode = num_tx_queues ? ODP_PKTOUT_MODE_DIRECT :
+		ODP_PKTOUT_MODE_DISABLED;
```
Bill-Fischofer-Linaro: Having an option to use direct mode seems reasonable, but shouldn't we retain schedule mode (perhaps as a command line switch). This would provide an easy means of testing scheduler efficiency as it is tuned. At least in some environments we'd like schedule mode to show better performance than direct.

example/generator/odp_generator.c
bogdanPricope: '-r' may work.

example/generator/odp_generator.c
bogdanPricope: No -  we need to print csum errors first.
This part was significantly changed in api-next (csum checks use different/ new API) and it makes no sense to optimize it for the old (master) code. After integration in api-next, this part will be  reworked  to use less parser flags (reduce parsing level).
For example, removing L4 parsing and locating interface is bringing an extra 1 mpps.

example/generator/odp_generator.c
bogdanPricope: This will complicate this already over-complicated code: we may need to decide between ultimate performance and feature richness.

muvarov: just separate rx function for scheduler and on thread start you just select scheduler or direct.

example/generator/odp_generator.c
```diff
@@ -838,39 +892,55 @@ static int gen_recv_thread(void *arg)
...
+	odp_barrier_wait(&barrier);
+
+	for (;;) {
+		if (thr_args->stop)
+			break;
+
+		pkt_cnt = odp_schedule_multi(NULL, ODP_SCHED_NO_WAIT,
+					     events, burst_size);
+
+		if (pkt_cnt > 0) {
```
muvarov:
```
 * @return Next highest priority event
 * @retval ODP_EVENT_INVALID on timeout and no events available
```

example/generator/odp_generator.c
```diff
@@ -838,39 +892,55 @@ static int gen_recv_thread(void *arg)
...
+			for (i = 0; i < pkt_cnt; i++)
+				pkts[i] = odp_packet_from_event(events[i]);
```
muvarov: not all events are packets.

example/generator/odp_generator.c
```diff
@@ -838,39 +892,55 @@ static int gen_recv_thread(void *arg)
...
-			itf = &itfs[odp_pktio_index(odp_packet_input(pkt))];
-
-			if (odp_packet_has_ipv4(pkt)) {
-				if (itf->config.pktin.bit.ipv4_chksum) {
-					if (odp_packet_has_l3_error(pkt))
-						printf("HW detected L3 error\n");
-				}
-			}
+		pkt_cnt = odp_pktin_recv_tmo(pktin, pkts, burst_size,
+					     ODP_PKTIN_NO_WAIT);
```
muvarov: why not ODP_PKTIN_WAIT?

muvarov: and why odp_pktin_recv_tmo() and not odp_pktin_recv() ?

example/generator/odp_generator.c
muvarov: ok

example/generator/odp_generator.c
```diff
@@ -563,44 +572,68 @@ static int create_pktio(const char *dev, odp_pool_t pool,
...
 
-	odp_pktin_queue_param_init(&pktin_param);
-	pktin_param.num_queues = num_rx_queues;
-	pktin_param.queue_param.sched.sync = ODP_SCHED_SYNC_ATOMIC;
+		odp_pktin_queue_param_init(&pktin_param);
+		pktin_param.num_queues = num_rx_queues;
+		pktin_param.op_mode = pktin_mode;
+		if (sched)
+			pktin_param.queue_param.sched.sync =
+				ODP_SCHED_SYNC_ATOMIC;
```
Bill-Fischofer-Linaro: Might be good to have options for controlling the queue sync type here as `ODP_SCHED_SYNC_PARALLEL` should result in highest throughput, and `ODP_SCHED_SYNC_ORDERED` would be useful in testing performance of scheduler implementations (in theory should be better than `ODP_SCHED_SYNC_ATOMIC`).

Something to explore in another PR

example/generator/odp_generator.c
Bill-Fischofer-Linaro: Checksum errors will result in `odp_packet_has_error()` being set as well, so these checks can be done only if the summary packet error predicate is set, avoiding unnecessary checks for known good packets.

example/generator/odp_generator.c
Bill-Fischofer-Linaro: Agree with @muvarov, this could use some comments to explain why these calls are being used. You'd expect a dedicated RX thread to simply wait for packet input.

example/generator/odp_generator.c
```diff
@@ -838,39 +892,55 @@ static int gen_recv_thread(void *arg)
...
+	burst_size = args->rx_burst_size;
+
+	printf("  [%02i] created mode: RECEIVE SCHEDULER\n", thr);
+	odp_barrier_wait(&barrier);
+
+	for (;;) {
+		if (thr_args->stop)
+			break;
+
+		pkt_cnt = odp_schedule_multi(NULL, ODP_SCHED_NO_WAIT,
```
Bill-Fischofer-Linaro: Why `ODP_SCHED_NO_WAIT` vs. `ODP_SCHED_WAIT` here? You're just spinning if no packets are available so why not let the scheduler do the waiting?

example/generator/odp_generator.c
Bill-Fischofer-Linaro: The new `odp_event_filter_packet()` API would be useful here.

example/generator/odp_generator.c
bogdanPricope: * @return Number of events outputted (0 ... num)

example/generator/odp_generator.c
bogdanPricope: Yes, this can part of another PR.

example/generator/odp_generator.c
bogdanPricope: This csum check is done with newer API in API-NEXT (odp_packet_l3_chksum_status()). No sense to optimize this part for this older implementation

example/generator/odp_generator.c
bogdanPricope: yep. My impression was that odp_pktin_recv() is blocking but it seems is no: in 'ping' mode we need to be able to stop the receive thread after a number of pings.

example/generator/odp_generator.c
bogdanPricope: As mentioned before, we need to be able to stop receive side in 'ping' mode.

example/generator/odp_generator.c
bogdanPricope: In general case is true, but in this particular case is not:  we are not waiting other events / configuring other scheduled queues for ODP_EVENT_BUFFER, ODP_EVENT_TIMEOUT, ODP_EVENT_CRYPTO_COMPL or ODP_EVENT_IPSEC_RESULT. 
I hope ODP is not throwing garbage events in default scheduler group... or?

example/generator/odp_generator.c
muvarov: how about ODP_SCHED_WAIT and timeout event to break the loop and exit application?

example/generator/odp_generator.c
muvarov: it's not blocking, it just polling function.

example/generator/odp_generator.c
muvarov: That is good question. In general there should not be any garbage. But odp app should dispatch event first. I think we can add timeout event to break loop here instead of polling some global variable.

Bill-Fischofer-Linaro:
Ping to @bogdanPricope. Waiting on v3.

bogdanPricope:
@Bill-Fischofer-Linaro : In progress.

bogdanPricope:
https://lng.validation.linaro.org/scheduler/job/23909.0
https://lng.validation.linaro.org/scheduler/job/23909.1

TX 14.8 mpps / RX 11.59 mpps

bogdanPricope:
checkpatch fails due to false negative:

CHECK: Avoid CamelCase: <ODPH_IPPROTO_ICMPv4>

Bill-Fischofer-Linaro:
@bogdanPricope I've created PR #386 to address the issue of the camelcase ICMP defines.

muvarov:
Merged.

PR #346, opened 2017-12-18
[PATCH CATERPILLAR v1] Pad pool cache to cache line

nagarahalli:
1) Pad pool cache such that, the next pool cache in an array always starts at a cache line
2) Convert checks for programming errors to ODP_ASSERT.

### DIFF
--- a/platform/linux-generic/include/odp_pool_internal.h
+++ b/platform/linux-generic/include/odp_pool_internal.h
@@ -29,10 +29,15 @@ extern "C" {
 
 #define CACHE_BURST    32
 
-typedef struct pool_cache_t {
+struct pool_cache_s {
 	uint32_t num;
 	uint32_t buf_index[CONFIG_POOL_CACHE_SIZE];
 
+};
+
+typedef union pool_cache_u {
+	struct pool_cache_s s;
+	uint8_t pad[ROUNDUP_CACHE_LINE(sizeof(struct pool_cache_s))];
 } pool_cache_t ODP_ALIGNED_CACHE;
 
 /* Buffer header ring */
--- a/platform/linux-generic/odp_packet.c
+++ b/platform/linux-generic/odp_packet.c
@@ -816,10 +816,7 @@ odp_packet_t odp_packet_alloc(odp_pool_t pool_hdl, uint32_t len)
 	odp_packet_t pkt;
 	int num, num_seg;
 
-	if (odp_unlikely(pool->params.type != ODP_POOL_PACKET)) {
-		__odp_errno = EINVAL;
-		return ODP_PACKET_INVALID;
-	}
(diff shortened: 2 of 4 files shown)


platform/linux-generic/include/odp_pool_internal.h
```diff
@@ -29,10 +29,15 @@ extern "C" {
...
+struct pool_cache_s {
 	uint32_t num;
 	uint32_t buf_index[CONFIG_POOL_CACHE_SIZE];
 
+};
+
+typedef union pool_cache_u {
+	struct pool_cache_s s;
+	uint8_t pad[ROUNDUP_CACHE_LINE(sizeof(struct pool_cache_s))];
 } pool_cache_t ODP_ALIGNED_CACHE;
```
Bill-Fischofer-Linaro: With Kevin's latest change (PR #326) this is already an integral multiple of the cache size. Not sure what value we're gaining here. Is this to guard against random changes to `CONFIG_POOL_CACHE_SIZE`? That still could leave a lot of wasted space, which was the idea behind Kevin's PR.

nagarahalli: Kevin's PR is to adjust the default value so that by default we are using 1 cache line. However, CONFIG_POOL_CACHE_SIZE is configurable. If the user is not aware of the internal details of the pool implementation, parts of 2 elements of the pool cache array will be in the same cache line. This change will guard against that. If the CONFIG_POOL_CACHE_SIZE is configured correctly, this will not add any additional space.

Bill-Fischofer-Linaro: My point is that we want two things: Cache alignment and no wasted space. Kevin's PR addressed this issue. It seems all we need is document the implications for those who wish to set `CONFIG_POOL_CACHE_SIZE` since otherwise we'll just be inserting a lot of padding that once again will waste space.

platform/linux-generic/odp_packet.c
```diff
@@ -816,10 +816,7 @@ odp_packet_t odp_packet_alloc(odp_pool_t pool_hdl, uint32_t len)
 	odp_packet_t pkt;
 	int num, num_seg;
 
-	if (odp_unlikely(pool->params.type != ODP_POOL_PACKET)) {
-		__odp_errno = EINVAL;
-		return ODP_PACKET_INVALID;
-	}
+	ODP_ASSERT(pool->params.type == ODP_POOL_PACKET);
```
joseppc: I think this is not correct. Before if we detected an error we interrupted the function returning an error code, now we will only interrupt this function (by calling abort_fn() global function) if and only if ODP was compiled with ODP_DEBUG enabled.

heyi-linaro:
Close this PR due to high level decision

PR #350, opened 2017-12-19
[PATCH API-NEXT v3] api, linux-gen, validation: crypto change api.ptr to api_ptr in packe…

lumag:
…t params

In new odp_crypto_packet_op_param_t change AAD from struct with a single
pointer to just data pointer. This API is new, so there is no need to
deprecate and/or handle old pointer in any specific way.


lumag:
@psavol should we apply same change to `odp_crypto_op_param_t` ?

muvarov:
Merged.

PR #351, opened 2017-12-19
[PATCH API-NEXT v5] Rework IPsec crypto capabilities

lumag:
Generic crypto caps provide too much information to be useful for IPsec applications (because some of the data is fixed). Rework respective functions to provide just necessary data.

### DIFF
--- a/platform/linux-generic/odp_ipsec.c
+++ b/platform/linux-generic/odp_ipsec.c
@@ -57,16 +57,75 @@ int odp_ipsec_capability(odp_ipsec_capability_t *capa)
 	return 0;
 }
 
+/* This should be enough for all ciphers and auths. Currently used maximum is 3
+ * capabilities */
+#define MAX_CAPS 10
+
 int odp_ipsec_cipher_capability(odp_cipher_alg_t cipher,
-				odp_crypto_cipher_capability_t capa[], int num)
+				odp_ipsec_cipher_capability_t capa[], int num)
 {
-	return odp_crypto_cipher_capability(cipher, capa, num);
+	odp_crypto_cipher_capability_t crypto_capa[MAX_CAPS];
+	uint32_t req_iv_len;
+	int rc, i, out;
+
+	rc = odp_crypto_cipher_capability(cipher, crypto_capa, MAX_CAPS);
+	if (rc <= 0)
+		return rc;
+
+	ODP_ASSERT(rc <= MAX_CAPS);
+	if (rc > MAX_CAPS)
+		rc = MAX_CAPS;
+
+	req_iv_len = _odp_ipsec_cipher_iv_len(cipher);
+	for (i = 0, out = 0; i < rc; i++) {
+		if (crypto_capa[i].iv_len != req_iv_len)
+			continue;
+
+		if (out < num)
+			capa[out].key_len = crypto_capa[i].key_len;
+		out++;
+	}
+
+	return out;
 }
 
 int odp_ipsec_auth_capability(odp_auth_alg_t auth,
-			      odp_crypto_auth_capability_t capa[], int num)
+			      odp_ipsec_auth_capability_t capa[], int num)
 {
-	return odp_crypto_auth_capability(auth, capa, num);
+	odp_crypto_auth_capability_t crypto_capa[MAX_CAPS];
+	uint32_t req_digest_len;
+	int rc, i, out;
+
+	rc = odp_crypto_auth_capability(auth, crypto_capa, MAX_CAPS);
+	if (rc <= 0)
+		return rc;
+
+	ODP_ASSERT(rc <= MAX_CAPS);
+	if (rc > MAX_CAPS)
+		rc = MAX_CAPS;
+
+	req_digest_len = _odp_ipsec_auth_digest_len(auth);
+	for (i = 0, out = 0; i < rc; i++) {
+		if (crypto_capa[i].digest_len != req_digest_len)
+			continue;
+
+		if (ODP_AUTH_ALG_AES_GCM == auth ||
+		    ODP_DEPRECATE(ODP_AUTH_ALG_AES128_GCM) == auth) {
+			uint8_t aad_len = 12;
+
+			if (aad_len < crypto_capa[i].aad_len.min ||
+			    aad_len > crypto_capa[i].aad_len.max ||
+			    0 != (aad_len - crypto_capa[i].aad_len.min) %
+				  crypto_capa[i].aad_len.inc)
+				continue;
+		}
+
+		if (out < num)
+			capa[out].key_len = crypto_capa[i].key_len;
+		out++;
+	}
+
+	return out;
 }
 
 void odp_ipsec_config_init(odp_ipsec_config_t *config)
--- a/platform/linux-generic/odp_ipsec_sad.c
+++ b/platform/linux-generic/odp_ipsec_sad.c
@@ -190,12 +190,66 @@ void odp_ipsec_sa_param_init(odp_ipsec_sa_param_t *param)
 	param->dest_queue = ODP_QUEUE_INVALID;
 }
 
+/* Return IV length required for the cipher for IPsec use */
+uint32_t _odp_ipsec_cipher_iv_len(odp_cipher_alg_t cipher)
+{
+	switch (cipher) {
+	case ODP_CIPHER_ALG_NULL:
+		return 0;
+	case ODP_CIPHER_ALG_DES:
+	case ODP_CIPHER_ALG_3DES_CBC:
+		return 8;
+#if ODP_DEPRECATED_API
+	case ODP_CIPHER_ALG_AES128_CBC:
+#endif
+	case ODP_CIPHER_ALG_AES_CBC:
+	case ODP_CIPHER_ALG_AES_CTR:
+		return 16;
+#if ODP_DEPRECATED_API
+	case ODP_CIPHER_ALG_AES128_GCM:
+#endif
+	case ODP_CIPHER_ALG_AES_GCM:
+		return 12;
+	default:
+		return (uint32_t)-1;
+	}
+}
+
+/* Return digest length required for the cipher for IPsec use */
+uint32_t _odp_ipsec_auth_digest_len(odp_auth_alg_t auth)
+{
+	switch (auth) {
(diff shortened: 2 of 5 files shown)


platform/linux-generic/odp_ipsec.c
```diff
@@ -58,15 +58,55 @@ int odp_ipsec_capability(odp_ipsec_capability_t *capa)
...
+	rc = odp_crypto_auth_capability(auth, crypto_capa, caps);
+	if (rc <= 0)
+		return rc;
+
+	req_digest_len = _odp_ipsec_auth_digest_len(auth);
+	for (i = 0, out = 0; i < rc; i++) {
+		if (crypto_capa[i].digest_len != req_digest_len)
+			continue;
+
+		/* FIXME: check AAD len for AES_GCM/AES128_GCM */
```
Bill-Fischofer-Linaro: Is this fix anticipated soon or will it be deferred until after Tiger Moth RC1?

lumag: The fix will not chsnge an API, so it can come  up anytime.

lumag: @Bill-Fischofer-Linaro fixed now

platform/linux-generic/odp_ipsec.c
```diff
@@ -58,15 +58,64 @@ int odp_ipsec_capability(odp_ipsec_capability_t *capa)
...
+		return rc;
+
+	req_iv_len = _odp_ipsec_cipher_iv_len(cipher);
+	for (i = 0, out = 0; i < rc; i++) {
+		if (crypto_capa[i].iv_len != req_iv_len)
+			continue;
+
+		if (out < num)
+			capa[out].key_len = crypto_capa[i].key_len;
+		out++;
```
muvarov: "capa[out].key_len =" has to be in the same place where you do "out++". Now it's out of if. I would just add break if  "out" reaches "num"

platform/linux-generic/odp_ipsec.c
```diff
@@ -58,15 +58,64 @@ int odp_ipsec_capability(odp_ipsec_capability_t *capa)
...
 {
-	return odp_crypto_cipher_capability(cipher, capa, num);
+	int caps = odp_crypto_cipher_capability(cipher, NULL, 0);
+	odp_crypto_cipher_capability_t crypto_capa[caps];
+	uint32_t req_iv_len;
+	int rc, i, out;
+
+	rc = odp_crypto_cipher_capability(cipher, crypto_capa, caps);
+	if (rc <= 0)
+		return rc;
```
muvarov: return -1 is more valid here. What is return 0 is not clear. From api it's also not really clear what 0 means. validation test has assert for > 0 and does not expect 0.

platform/linux-generic/odp_ipsec.c
```diff
@@ -58,15 +58,64 @@ int odp_ipsec_capability(odp_ipsec_capability_t *capa)
...
-				odp_crypto_cipher_capability_t capa[], int num)
+				odp_ipsec_cipher_capability_t capa[], int num)
 {
-	return odp_crypto_cipher_capability(cipher, capa, num);
+	int caps = odp_crypto_cipher_capability(cipher, NULL, 0);
+	odp_crypto_cipher_capability_t crypto_capa[caps];
+	uint32_t req_iv_len;
+	int rc, i, out;
+
+	rc = odp_crypto_cipher_capability(cipher, crypto_capa, caps);
```
muvarov: please add check that caps > 0.

platform/linux-generic/odp_ipsec.c
lumag: No, because this function should count and return amount of capabilities.

platform/linux-generic/odp_ipsec.c
```diff
@@ -58,15 +58,64 @@ int odp_ipsec_capability(odp_ipsec_capability_t *capa)
...
+			if (aad_len < crypto_capa[i].aad_len.min ||
+			    aad_len > crypto_capa[i].aad_len.max ||
+			    0 != (aad_len - crypto_capa[i].aad_len.min) %
+				  crypto_capa[i].aad_len.inc)
+				continue;
+		}
+
+		if (out < num)
+			capa[out].key_len = crypto_capa[i].key_len;
+		out++;
```
muvarov: same here. You return more than you fill into capa[]

platform/linux-generic/odp_ipsec.c
lumag: Basically I wanted to keep return from odp_crypto_cipher_capability. `0` here would mean that there are no capabilities to report.

platform/linux-generic/odp_ipsec.c
lumag: This is expected.

platform/linux-generic/odp_ipsec_sad.c
```diff
@@ -190,12 +190,66 @@ void odp_ipsec_sa_param_init(odp_ipsec_sa_param_t *param)
...
+	case ODP_CIPHER_ALG_AES_CBC:
+	case ODP_CIPHER_ALG_AES_CTR:
+		return 16;
+#if ODP_DEPRECATED_API
+	case ODP_CIPHER_ALG_AES128_GCM:
+#endif
+	case ODP_CIPHER_ALG_AES_GCM:
+		return 12;
+	default:
+		return (uint32_t)-1;
```
muvarov: should it be 0 also here?

platform/linux-generic/odp_ipsec_sad.c
```diff
@@ -190,12 +190,66 @@ void odp_ipsec_sa_param_init(odp_ipsec_sa_param_t *param)
...
+	case ODP_AUTH_ALG_SHA512_HMAC:
+		return 32;
+#if ODP_DEPRECATED_API
+	case ODP_AUTH_ALG_AES128_GCM:
+#endif
+	case ODP_AUTH_ALG_AES_GCM:
+	case ODP_AUTH_ALG_AES_GMAC:
+		return 16;
+	default:
+		return (uint32_t)-1;
```
muvarov: should it be 0 also here?

platform/linux-generic/odp_ipsec_sad.c
muvarov: or ODP_ABORT()

lumag: 0, it's "unsupported", so -1 fits better.

platform/linux-generic/odp_ipsec_sad.c
lumag: same here

platform/linux-generic/odp_ipsec_sad.c
lumag: 0 would mean no iv/icv needed

platform/linux-generic/odp_ipsec.c
muvarov:
```
 * @return Number of capability structures for the algorithm. If this is larger
 *         than 'num', only 'num' first structures were output and application
 *         may call the function again with a larger value of 'num'.
```

muvarov: so you did not fill anything to capa[], increased "out" and returned "out". Then application expects "out" elements in capa[] but some of them are empty. Or I'm wrong?

muvarov: i.e. you can not return more then expected.

lumag: Actually no. Please consider `odp_crypto_cipher_capability()`/`odp_crypto_auth_capability()` code. They return amount of elements to be filled.

muvarov:
```
int odp_crypto_auth_capability(odp_auth_alg_t auth,
			       odp_crypto_auth_capability_t dst[], int num_copy)
{
....
	if (num < num_copy)
		num_copy = num;

	memcpy(dst, src, num_copy * size);

	return num;
```
Here logic is ok. Returned requested num_copy or num if it's smaller.

In patch:
```
odp_crypto_auth_capability_t crypto_capa[caps];
for (i = 0, out = 0; i < rc; i++) {
....
		if (out < num)
￼ 			capa[out].key_len = crypto_capa[i].key_len;
￼		out++;
}
return out;
```
You set up key_len for first num elements, leave rest of element uninitialized and return number of possible elements. Or it's too late for today, but I think I'm correct.

odp_crypto_auth_capability() with dst = NULL calls:
memcpy(dst, src, num_copy * size);
Did you tried it with -O0? I guess memcpy(NULL, *ptr, 0) might fail on some conditions.

platform/linux-generic/odp_ipsec_sad.c
lumag: For example 0 is a valid length for `ODP_CIPHER_ALG_NULL` and for `ODP_AUTH_ALG_NULL`. Thus -1. Added error checking in `odp_ipsec_sa_create()`.

platform/linux-generic/odp_ipsec_sad.c
```diff
@@ -297,32 +351,34 @@ odp_ipsec_sa_t odp_ipsec_sa_create(const odp_ipsec_sa_param_t *param)
 	crypto_param.auth_alg = param->crypto.auth_alg;
 	crypto_param.auth_key = param->crypto.auth_key;
 
+	crypto_param.iv.length =
+		_odp_ipsec_cipher_iv_len(crypto_param.cipher_alg);
```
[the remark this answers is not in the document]
lumag: Yes, adding both checks below.

platform/linux-generic/odp_ipsec.c
lumag: Rewrote this piece of code to use fixed length arrays.

platform/linux-generic/odp_ipsec.c
Bill-Fischofer-Linaro: Since `odp_ipsec_xxx_capability()` is just a filter on top of the `odp_crypto_xxx_capability()` base layer, passing the RC through seems reasonable here. The Crypto API will be returning RC = -1 if there was a problem anyway.

muvarov:
Merged.

PR #352, opened 2017-12-19
[PATCH API-NEXT v7] Add separate auth IV

lumag:
I was thinking again about GMAC auth case. Currently GMAC just reuses cipher IV data fields, which is not future proof and not quite obvious. So let's add separate set of IV-related fields for authentication algorithms.


lumag:
Rebased to solve the conflict.

lumag:
@psavol ping?

psavol:
Double IVs are needed if there's a case that both cipher and auth algo need IV (simultaneously and different IVs). The case of null cipher + AES GMAC auth the current spec ...

	/** Cipher Initialization Vector (IV) */
	odp_crypto_iv_t iv;
 
... could be extended in params to say "cipher IV or auth IV (in case of AES GMAC)".

In any case, auth capa needs iv_len added for GMAC. IV length for null cipher is too much of a hack. I'd avoid duplicating iv and iv override pointer, if there's no case where both cipher and auth need it.

lumag:
Initially I used single iv array and separate lengths for cipher and auth. Later I thought about SNOW 3G and ZUC (where both encryption and integrity algorithms use IVs). We do not support them in ODP API but we might need to add them later (ODP-DPAA2 supports them). Or about strange but possible case of using e.g. AES-CBC + AES-GMAC (in fact any non-null cipher + gmac). Thus I ended up with separate iv fields and override pointers.

lumag:
- added GMAC tests with IV override
- small cleanup for crypto tests
- rebase on top of api-next

Bill-Fischofer-Linaro:
@psavol Have your questions been answered in latest version? API changes still need your review.

include/odp/api/spec/crypto.h
```diff
@@ -327,6 +327,9 @@ typedef struct odp_crypto_session_param_t {
 	 */
 	odp_crypto_key_t auth_key;
 
+	/** Authentication Initialization Vector (IV) */
+	odp_crypto_iv_t auth_iv;
+
```
psavol: /** Cipher Initialization Vector (IV) */
odp_crypto_iv_t iv;

Shame that original cipher IV was not named "cipher_iv", but just "iv". I'm wondering if we should create an unnamed union of iv and cipher_iv, move all our code to use "cipher_iv" and then later on deprecate "iv" (some time after TigerMoth).

psavol:
Snow3G use case supports this addition, so I think we could do it. See a comment about old "iv" vs. "cipher_iv" as the new name for it.

lumag:
@psavol Implemented your suggestion plus squashed auth_iv commits together (so there are two meaningful commits: one for cipher_iv, one for auth_iv).

Bill-Fischofer-Linaro:
Ping to @psavol for review.

include/odp/api/spec/crypto.h
```diff
@@ -403,6 +406,9 @@ typedef struct odp_crypto_op_param_t {
 	/** Override session IV pointer */
 	uint8_t *override_iv_ptr;
 
+	/** Override session authentication IV pointer */
+	uint8_t *override_auth_iv_ptr;
```
psavol: The old override_iv_ptr is now ptr for cipher. In minimum, the documentation should say that. Although, while changing IV pointers, we could update these also.
/* Override IV pointer for cipher */
union {
DEPRECATED(override_iv_ptr);
cipher_iv_ptr;
}

/* Override IV pointer for auth */
auth_iv_ptr;

Bill-Fischofer-Linaro: @lumag @psavol Looks like we're very close here. Can this PR be finalized for Monday's ARCH call?

lumag: @psavol @Bill-Fischofer-Linaro updated.

muvarov:
Merged. But after rewriting api-next history all merges look ugly :(

lumag:
Ugly?
Well, `git rebase --onto origin/api-next origin/api-next@{1} BRANCH` did the trick for me.

PR #353, opened 2017-12-20
api: lock-free queues implementation

psavol:
Simple implementation of lockfree queues. Capability API changed to enable totally different queue implementation of non-blocking vs blocking queues.


include/odp/api/spec/queue.h
```diff
@@ -164,35 +164,81 @@ typedef struct odp_queue_capability_t {
...
+		  * implementation. */
+		struct {
+			/** Maximum number of queues. Wait-free queues are not
+			  * supported when zero. */
+			uint32_t max_num;
+
+			/** Maximum queue size */
+			uint32_t max_size;
+
+		} waitfree;
```
Bill-Fischofer-Linaro: This seems to imply that these limits are orthogonal. How do we handle cases where the number of queues is fixed but whether they are non-blocking, etc. is up to the user? In the case of a HW implementation where all queues are WF by default, how does that implementation express that here? If I say `max_num` = 1024 and `waitfree.max_num` = 1024 that might lead users to assume they can create 2048 queues, 1024 of each, when in fact there is only a single limit. 

Just as with the pool limits, it sounds like we should specify that `max_num` and `max_size` are total limits and the others are qualifiers for the subtypes. So `max_num` = 1024 says that I can define a total of up to 1024 queues and `waitfree.max_num` = 1024 says that up to that many of the `max_num` total may be defined as `ODP_NONBLOCKING_WF`, etc.

include/odp/api/spec/queue.h
```diff
@@ -164,35 +164,81 @@ typedef struct odp_queue_capability_t {
...
```
Bill-Fischofer-Linaro: Same comments apply here. Though we have the added consideration of whether the `plain.max_num` and `sched.max_num` also come out of a common limit or if they are independent?

test/validation/api/queue/queue.c
```diff
@@ -326,6 +414,7 @@ void queue_test_info(void)
 odp_testinfo_t queue_suite[] = {
 	ODP_TEST_INFO(queue_test_capa),
 	ODP_TEST_INFO(queue_test_mode),
+	ODP_TEST_INFO(queue_test_lockfree),
```
Bill-Fischofer-Linaro: Should there be a corresponding `queue_test_waitfree`?

include/odp/api/spec/queue.h
psavol: /** Maximum number of event queues of any type (default size). Use
	  * this in addition to queue type specific 'max_num', if both queue
	  * types are used simultaneously. */
	uint32_t max_queues;

Top level max_queues is defined as max queues of any type already. Above, "both" could be changed to "different", if that's clearer.

include/odp/api/spec/queue.h
psavol: Max size is per type: one queue type may store infinite (== linked list) many events while another have a limit (== ring). So, the same does not apply here.

See max_queues definition on top level.

test/validation/api/queue/queue.c
psavol: Not yet, since I don't have an implementation for that yet. There's no code to test the test.

psavol:
I'd rather get this into RC1, since what we have currently in the repo needs this change anyway. If some word in spec text needs an update, that can be changed in RC2. This change needs application code modification, while potential comment change is just matter of improving reader understanding of the spec.

Bill-Fischofer-Linaro:
@psavol Please rebase to resolve forced update conflicts with api-next

PR #355, opened 2017-12-20
[PATCH API-NEXT v1] Fill crypto packet-mode capabilities

lumag:
This PR is not an API change per se, it just implements part of API previously defined (and unfortunately ommited from implementation).


muvarov:
Merged.

PR #358, opened 2017-12-21
[PATCH v1] Arch directory clean up

psavol:
Arch directory implement API functions (CPU cycles and system info) and internal time register read functions. Rename and move time internal header under arch dir, so that it's clear which internal functions must be implemented for every arch. Later on other odp_arch_xxx_internal.h headers may be added. Combined odp_cpu_cycles.c and odp_cpu_arch.c files, since both implemented CPU cycle functions. Removed unused files.


lumag:

platform/linux-generic/arch/powerpc/odp_cpu_arch.c
```diff
@@ -1,41 +0,0 @@
...
-#include <odp/api/cpu.h>
-#include <odp/api/hints.h>
-#include <odp/api/system_info.h>
-#include <odp_debug_internal.h>
-#include <odp_time_internal.h>
-
-#define GIGA 1000000000
-
-uint64_t odp_cpu_cycles(void)
-{
```
Bill-Fischofer-Linaro: Travis is still doing cross-compilation testing for PowerPC, so I don't see why we want to remove these files. There doesn't seem to be any compelling reason for doing so.

psavol: As the commit log text says: these files are not currently used. These are not used even when building for PowerPC (uses default arch files). As you can see, the patch does not touch any Makefiles or Travis files, just removes these C files and all Travis tests still pass - including the PowerPC build.

Bill-Fischofer-Linaro: Ok.

muvarov:
Merged.

PR #361, opened 2017-12-21
[PATCH v1] test: validation: correct comparison in packet_test_ref()

Bill-Fischofer-Linaro:
Fix typo in CU_ASSERT() that used an assignment instead of a
comparison in testing reference length.

This fixes Bug https://bugs.linaro.org/show_bug.cgi?id=3466

### DIFF
--- a/test/validation/api/packet/packet.c
+++ b/test/validation/api/packet/packet.c
@@ -2354,7 +2354,7 @@ void packet_test_ref(void)
 	ref_pkt[0] = odp_packet_ref(base_pkt, offset[0]);
 	CU_ASSERT_FATAL(ref_pkt[0] != ODP_PACKET_INVALID);
 	ref_len[0] = odp_packet_len(ref_pkt[0]);
-	CU_ASSERT(ref_len[0] = odp_packet_len(base_pkt) - offset[0]);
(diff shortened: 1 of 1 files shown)


test/validation/api/packet/packet.c
```diff
@@ -2354,7 +2354,7 @@ void packet_test_ref(void)
 	ref_pkt[0] = odp_packet_ref(base_pkt, offset[0]);
 	CU_ASSERT_FATAL(ref_pkt[0] != ODP_PACKET_INVALID);
 	ref_len[0] = odp_packet_len(ref_pkt[0]);
-	CU_ASSERT(ref_len[0] = odp_packet_len(base_pkt) - offset[0]);
+	CU_ASSERT(ref_len[0] == odp_packet_len(base_pkt) - offset[0]);
```
muvarov: I would add just additional braces for (odp_packet_len(base_pkt) - offset[0])

lumag: @muvarov that would be still an error, wouldn't it?

muvarov: I think now but in other places it's the same style. So changes are not needed.

muvarov:
Merged.

PR #365, opened 2017-12-23
[PATCH v13] linux-generic: pktio: add null pktio support

jmtilli:
Sometimes, you may wish to run an application by using a dummy interface
that never receives any packets and that discards transmitted packets.
This is possible with non-looping pcap pktio with no input and output
pcaps. So, the pktio would be just "pcap:". However, you cannot have two
pktios named "pcap:" because there is code to check if the pktio is
already opened. Also, the pcap pktio requires that ODP is compiled with
the pcap library, which it may not be.

Also, recently a patch was provided to ODP to allow interruptible sleep
when receiving packets. This, however, does not work with mixed pcap and
socket/netmap pktio types.

To fix these issues, a new "null" pktio is introduced. It is used by
specifying the interface name to be "null:0", "null:1", or similar. All
packets sent to the null pktio are discarded and no packets are
received. The null pktio fully supports the interruptible sleep
infrastructure, meaning applications can have 0% idle CPU load when one
interface is a null interface and another interface is socket or netmap.

Note this patch depends on the genuinesleep patch. So, do not merge this
patch yet! The patch can, however, be reviewed.

### DIFF
--- a/example/generator/Makefile.am
+++ b/example/generator/Makefile.am
@@ -3,3 +3,8 @@ include $(top_srcdir)/example/Makefile.inc
 bin_PROGRAMS = odp_generator
 
 odp_generator_SOURCES = odp_generator.c
+
+if test_example
+TESTS = generator_run.sh
+endif
+EXTRA_DIST = generator_run.sh
--- a/example/generator/generator_run.sh
+++ b/example/generator/generator_run.sh
@@ -1,0 +1,17 @@
+#!/bin/bash
+#
+# Copyright (c) 2018, Linaro Limited
+# All rights reserved.
+#
+# SPDX-License-Identifier:     BSD-3-Clause
+#
+
+./odp_generator${EXEEXT} -w 1 -n 1 -I null:0 -m u
+STATUS=$?
+
+if [ "$STATUS" -ne 0 ]; then
+  echo "Error: status was: $STATUS, expected 0"
+  exit 1
+fi
+
+exit 0
--- a/example/l2fwd_simple/l2fwd_simple_run.sh
+++ b/example/l2fwd_simple/l2fwd_simple_run.sh
@@ -29,4 +29,17 @@ fi
 
 rm -f pcapout.pcap
 
+./odp_l2fwd_simple${EXEEXT} null:0 null:1 \
+	02:00:00:00:00:01 02:00:00:00:00:02 &
+
+sleep 1
+kill -s SIGINT $!
+wait $!
+STATUS=$?
+
+if [ "$STATUS" -ne 255 ]; then
+  echo "Error: status was: $STATUS, expected 255"
+  exit 1
+fi
+
 exit 0
--- a/example/l2fwd_simple/odp_l2fwd_simple.c
+++ b/example/l2fwd_simple/odp_l2fwd_simple.c
@@ -145,6 +145,8 @@ int main(int argc, char **argv)
 	odph_odpthread_params_t thr_params;
 	int opt;
 	int long_index;
+	odph_ethaddr_t correct_src;
+	uint32_t mtu1, mtu2;
 
 	static const struct option longopts[] = { {NULL, 0, NULL, 0} };
 	static const char *shortopts = "";
@@ -203,6 +205,23 @@ int main(int argc, char **argv)
 								&global.if0out);
 	global.if1 = create_pktio(argv[optind + 1], pool, &global.if1in,
 								&global.if1out);
+
+	/* Do some operations to increase code coverage in tests */
+	if (odp_pktio_mac_addr(global.if0, &correct_src, sizeof(correct_src))
+	    != sizeof(correct_src))
+		printf("Warning: can't get MAC address\n");
+	else if (memcmp(&correct_src, &global.src, sizeof(correct_src)) != 0)
+		printf("Warning: src MAC invalid\n");
+
+	odp_pktio_promisc_mode_set(global.if0, true);
+	odp_pktio_promisc_mode_set(global.if1, true);
+	(void)odp_pktio_promisc_mode(global.if0);
+	(void)odp_pktio_promisc_mode(global.if1);
+
+	mtu1 = odp_pktin_maxlen(global.if0);
+	mtu2 = odp_pktout_maxlen(global.if1);
+	if (mtu1 && mtu2 && mtu1 > mtu2)
+		printf("Warning: input MTU bigger than output MTU\n");
 
 	odp_cpumask_default_worker(&cpumask, MAX_WORKERS);
 
--- a/platform/linux-generic/include/odp_packet_null.h
+++ b/platform/linux-generic/include/odp_packet_null.h
@@ -1,0 +1,16 @@
+/* Copyright (c) 2018, Linaro Limited
+ * All rights reserved.
+ *
+ * SPDX-License-Identifier:     BSD-3-Clause
+ */
+
+#ifndef ODP_PACKET_NULL_H_
+#define ODP_PACKET_NULL_H_
+
+#include <odp/api/pool.h>
+
+typedef struct {
+	int promisc;			/**< whether promiscuous mode is on */
+} pkt_null_t;
+
+#endif
--- a/platform/linux-generic/pktio/null.c
+++ b/platform/linux-generic/pktio/null.c
@@ -1,0 +1,175 @@
+/* Copyright (c) 2018, Linaro Limited
+ * All rights reserved.
+ *
+ * SPDX-License-Identifier:     BSD-3-Clause
+ */
+
+#include "config.h"
+
+#include <odp_api.h>
+#include <odp_packet_io_internal.h>
+#include <odp_packet_null.h>
+
+static int null_close(pktio_entry_t *pktio_entry ODP_UNUSED)
+{
+	return 0;
+}
+
+static int null_open(odp_pktio_t id ODP_UNUSED,
+		     pktio_entry_t *pktio_entry,
+		     const char *devname, odp_pool_t pool ODP_UNUSED)
+{
+	if (strncmp(devname, "null:", 5) != 0)
+		return -1;
+	pktio_entry->s.pkt_null.promisc = 0;
+	return 0;
+}
+
+static int null_recv(pktio_entry_t *pktio_entry ODP_UNUSED,
+		     int index ODP_UNUSED, odp_packet_t pkt_table[] ODP_UNUSED,
+		     int len ODP_UNUSED)
+{
+	return 0;
+}
+
+static int null_fd_set(pktio_entry_t *pktio_entry ODP_UNUSED,
+		       int index ODP_UNUSED, fd_set *readfds ODP_UNUSED)
+{
+	return 0;
+}
+
+static int null_recv_tmo(pktio_entry_t *pktio_entry ODP_UNUSED,
+			 int index ODP_UNUSED,
+			 odp_packet_t pkt_table[] ODP_UNUSED,
+			 int num ODP_UNUSED, uint64_t usecs)
+{
+	struct timeval timeout;
+	int maxfd = -1;
+	fd_set readfds;
+
+	timeout.tv_sec = usecs / (1000 * 1000);
+	timeout.tv_usec = usecs - timeout.tv_sec * (1000ULL * 1000ULL);
+	FD_ZERO(&readfds);
+
+	select(maxfd + 1, &readfds, NULL, NULL,
+	       usecs == ODP_PKTIN_WAIT ? NULL : &timeout);
+
+	return 0;
+}
+
+static int null_recv_mq_tmo(pktio_entry_t *pktio_entry[] ODP_UNUSED,
+			    int index[] ODP_UNUSED, int num_q ODP_UNUSED,
+			    odp_packet_t pkt_table[] ODP_UNUSED,
+			    int num ODP_UNUSED, unsigned *from ODP_UNUSED,
+			    uint64_t usecs)
+{
+	struct timeval timeout;
+	int maxfd = -1;
+	fd_set readfds;
+
+	timeout.tv_sec = usecs / (1000 * 1000);
+	timeout.tv_usec = usecs - timeout.tv_sec * (1000ULL * 1000ULL);
+
+	FD_ZERO(&readfds);
+
+	select(maxfd + 1, &readfds, NULL, NULL,
+	       usecs == ODP_PKTIN_WAIT ? NULL : &timeout);
+
+	return 0;
+}
+
+static int null_send(pktio_entry_t *pktio_entry ODP_UNUSED,
+		     int index ODP_UNUSED, const odp_packet_t pkt_table[],
+		     int num)
+{
+	odp_packet_free_multi(pkt_table, num);
+
+	return num;
+}
+
+#define PKTIO_NULL_MTU (64 * 1024)
+
+static uint32_t null_mtu_get(pktio_entry_t *pktio_entry ODP_UNUSED)
+{
+	return PKTIO_NULL_MTU;
+}
+
+static const char null_mac[] = {0x02, 0xe9, 0x34, 0x80, 0x73, 0x05};
+
+static int null_mac_addr_get(pktio_entry_t *pktio_entry ODP_UNUSED,
+			     void *mac_addr)
+{
+	memcpy(mac_addr, null_mac, ETH_ALEN);
+	return ETH_ALEN;
+}
+
+static int null_promisc_mode_set(pktio_entry_t *pktio_entry, odp_bool_t enable)
+{
+	pktio_entry->s.pkt_null.promisc = !!enable;
+	return 0;
+}
+
+static int null_promisc_mode_get(pktio_entry_t *pktio_entry)
+{
+	return pktio_entry->s.pkt_null.promisc;
+}
+
+static int null_capability(pktio_entry_t *pktio_entry ODP_UNUSED,
+			   odp_pktio_capability_t *capa)
+{
+	memset(capa, 0, sizeof(odp_pktio_capability_t));
+
+	capa->max_input_queues  = PKTIO_MAX_QUEUES;
+	capa->max_output_queues = PKTIO_MAX_QUEUES;
+	capa->set_op.op.promisc_mode = 1;
+
+	odp_pktio_config_init(&capa->config);
+	capa->config.pktin.bit.ts_all = 1;
+	capa->config.pktin.bit.ts_ptp = 1;
+	return 0;
+}
+
+static int null_inqueues_config(pktio_entry_t *pktio_entry ODP_UNUSED,
+				const odp_pktin_queue_param_t *p ODP_UNUSED)
+{
+	return 0;
+}
+
+static int null_outqueues_config(pktio_entry_t *pktio_entry ODP_UNUSED,
+				 const odp_pktout_queue_param_t *p ODP_UNUSED)
+{
+	return 0;
+}
+
+static int null_init_global(void)
+{
+	ODP_PRINT("PKTIO: initialized null interface.\n");
+	return 0;
+}
+
+const pktio_if_ops_t null_pktio_ops = {
+	.name = "null",
+	.print = NULL,
+	.init_global = null_init_global,
+	.init_local = NULL,
+	.term = NULL,
+	.open = null_open,
+	.close = null_close,
+	.start = NULL,
+	.stop = NULL,
+	.recv = null_recv,
+	.recv_tmo = null_recv_tmo,
+	.recv_mq_tmo = null_recv_mq_tmo,
+	.fd_set = null_fd_set,
+	.send = null_send,
+	.mtu_get = null_mtu_get,
+	.promisc_mode_set = null_promisc_mode_set,
+	.promisc_mode_get = null_promisc_mode_get,
+	.mac_get = null_mac_addr_get,
+	.capability = null_capability,
+	.pktin_ts_res = NULL,
+	.pktin_ts_from_ns = NULL,
+	.config = NULL,
+	.input_queues_config = null_inqueues_config,
+	.output_queues_config = null_outqueues_config,
+};
--- a/platform/linux-generic/Makefile.am
+++ b/platform/linux-generic/Makefile.am
@@ -107,6 +107,7 @@ noinst_HEADERS = \
 		  include/odp_packet_dpdk.h \
 		  include/odp_packet_socket.h \
 		  include/odp_packet_tap.h \
+		  include/odp_packet_null.h \
 		  include/odp_pkt_queue_internal.h \
 		  include/odp_pool_internal.h \
 		  include/odp_posix_extensions.h \
@@ -163,6 +164,7 @@ __LIB__libodp_linux_la_SOURCES = \
 			   pktio/pktio_common.c \
 			   pktio/loop.c \
 			   pktio/netmap.c \
+			   pktio/null.c \
 			   pktio/dpdk.c \
 			   pktio/socket.c \
 			   pktio/socket_mmap.c \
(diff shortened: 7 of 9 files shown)


jmtilli:
Oh, and one forgotten argument for why null pktio is required. The null pktio is multi-queue. Pcap pktio is single-queue. In a modern multithreaded application, you cannot use pcap pktio as a dummy interface alone for this reason.

platform/linux-generic/pktio/null.c
```diff
@@ -0,0 +1,214 @@
...
+}
+
+#define PKTIO_NULL_MTU (64 * 1024)
+
+static uint32_t null_mtu_get(pktio_entry_t *pktio_entry ODP_UNUSED)
+{
+	return PKTIO_NULL_MTU;
+}
+
+static const char null_mac[] = {0x02, 0xe9, 0x34, 0x80, 0x73, 0x05};
```
Bill-Fischofer-Linaro: Where does this "magic number" come from? Why not `00:00:00:00:00:00`?

example/l2fwd_simple/odp_l2fwd_simple.c
```diff
@@ -204,6 +206,23 @@ int main(int argc, char **argv)
...
+	else if (memcmp(&correct_src, &global.src, sizeof(correct_src)) != 0)
+		printf("Warning: src MAC invalid\n");
+
+	odp_pktio_promisc_mode_set(global.if0, true);
+	odp_pktio_promisc_mode_set(global.if1, true);
+	(void)odp_pktio_promisc_mode(global.if0);
+	(void)odp_pktio_promisc_mode(global.if1);
+
+	mtu1 = odp_pktio_mtu(global.if0);
+	mtu2 = odp_pktio_mtu(global.if1);
```
Bill-Fischofer-Linaro: `odp_pktio_mtu()` is a deprecated API and should not be used in new code.

example/l2fwd_simple/odp_l2fwd_simple.c
```diff
@@ -204,6 +206,23 @@ int main(int argc, char **argv)
...
+	if (odp_pktio_mac_addr(global.if0, &correct_src, sizeof(correct_src))
+	    != sizeof(correct_src))
+		printf("Warning: can't get MAC address\n");
+	else if (memcmp(&correct_src, &global.src, sizeof(correct_src)) != 0)
+		printf("Warning: src MAC invalid\n");
+
+	odp_pktio_promisc_mode_set(global.if0, true);
+	odp_pktio_promisc_mode_set(global.if1, true);
+	(void)odp_pktio_promisc_mode(global.if0);
+	(void)odp_pktio_promisc_mode(global.if1);
```
lumag: This should probably go to a separate PR. And ideally examples should not add much to code coverage on top of tests, so you might want also to add code to api testcases.

example/generator/.gitignore
```diff
@@ -1 +1,3 @@
 odp_generator
+*.log
+*.trs
```
lumag: I think this should be now in one of top-level .gitignore files.

example/generator/Makefile.am
```diff
@@ -3,3 +3,6 @@ include $(top_srcdir)/example/Makefile.inc
 bin_PROGRAMS = odp_generator
 
 odp_generator_SOURCES = odp_generator.c
+
+TESTS = generator_run.sh
```
lumag: this should go under `if test_example`

example/l2fwd_simple/l2fwd_simple_run.sh
```diff
@@ -29,4 +29,17 @@ fi
...
+	02:00:00:00:00:01 02:00:00:00:00:02 &
+
+sleep 1
+kill -s SIGINT $!
+wait $!
+STATUS=$?
+
+if [ "$STATUS" -ne 255 ]; then
+  echo "Error: status was: $STATUS, expected 255"
+  exit 1
```
lumag: just exit $STATUS. This would allow underlying program to return 42 to mark the test as skipped

platform/linux-generic/pktio/null.c
```diff
@@ -0,0 +1,214 @@
...
+static int null_fd_set(pktio_entry_t *pktio_entry ODP_UNUSED,
+		       int index ODP_UNUSED, fd_set *readfds)
+{
+	/* Somebody might wonder why readfds was not marked ODP_UNUSED.
+	 * The reason is separately run checkpatch script complains that
+	 * there must be spaces around '*', and if there are spaces, then
+	 * the continuous integration build checkpatch script complains
+	 * there must not be spaces around '*'. */
+	(void)readfds;
+	return 0;
```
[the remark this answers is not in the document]
lumag: Just use ODP_UNUSED and forget about checkpatch being unoptimal here.

platform/linux-generic/pktio/null.c
lumag: Maybe we should add internal API to generate random (or semi-random) local MAC-addresses with L bit being set?

platform/linux-generic/pktio/null.c
```diff
@@ -0,0 +1,214 @@
...
+			     void *mac_addr)
+{
+	memcpy(mac_addr, null_mac, ETH_ALEN);
+	return ETH_ALEN;
+}
+
+static int null_promisc_mode_set(pktio_entry_t *pktio_entry ODP_UNUSED,
+				 odp_bool_t enable ODP_UNUSED)
+{
+	return 0;
```
lumag: This should be supportable also. Just like it is done in other software interfaces.

example/l2fwd_simple/odp_l2fwd_simple.c
jmtilli: Ok, the only reason I used this call was to increase code coverage. But is it deprecated? I grepped through the repository, finding to evidence that any API having the name mtu would be deprecated...

platform/linux-generic/pktio/null.c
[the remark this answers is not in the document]
jmtilli: Will do. This will result in a checkpatch warning then, but if that is not an obstacle to a merge, I can do this.

platform/linux-generic/pktio/null.c
```diff
@@ -0,0 +1,214 @@
...
+	struct timeval timeout;
+	int maxfd = -1;
+	fd_set readfds;
+
+	timeout.tv_sec = usecs / 1000 / 1000;
+	timeout.tv_usec = usecs % (1000 * 1000);
+	FD_ZERO(&readfds);
+
+	select(maxfd + 1, &readfds, NULL, NULL,
+	       usecs == ODP_PKTIN_WAIT ? NULL : &timeout);
```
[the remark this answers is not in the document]
jmtilli: I'm not entirely sure about this... If the programmer wants to wait indefinitely, that's perhaps what we should do, which is indeed what is done now. The odp_pktio_stop() comment should perhaps be made to the PR #341. A good question is do we really need the empty readfds fd_set. Probably we could do without it.

platform/linux-generic/pktio/null.c
```diff
@@ -0,0 +1,214 @@
...
+
+	return 0;
+}
+
+static int null_send(pktio_entry_t *pktio_entry ODP_UNUSED,
+		     int index ODP_UNUSED, const odp_packet_t pkt_table[],
+		     int len)
```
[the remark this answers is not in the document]
jmtilli: I'll fix in the next version.

platform/linux-generic/pktio/null.c
```diff
@@ -0,0 +1,214 @@
...
+{
+	int i;
+
+	for (i = 0; i < len; i++)
+		odp_packet_free(pkt_table[i]);
```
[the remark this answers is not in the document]
jmtilli: This will also be fixed in the next version.

platform/linux-generic/pktio/null.c
jmtilli: This is taken from the .pcap pktio, and incremented by one to make it distinct from the .pcap MAC address. But agree, an internal API to generate the MAC address would be great. It needs to be made sure the multicast bit is not being set, then.

example/generator/.gitignore
jmtilli: It isn't, at least in the master version on which this branch is based.

example/l2fwd_simple/odp_l2fwd_simple.c
jmtilli: The problem is that the pktio validation test case doesn't like an interface where packets are not being received. It fails a number of tests, and then for the pktio_test_recv_tmo it just hangs. With the validation tests, an entirely different test for a quiescent interface would be needed. Also, considering that this is testing a particular driver ("null"), not a particular API, I'm not entirely sure that API validation test would be the right location...

example/generator/Makefile.am
jmtilli: Will fix.

example/l2fwd_simple/l2fwd_simple_run.sh
jmtilli: In theory, a STATUS 0 would be considered an error here because the expected status is -1 which is converted to 255.

platform/linux-generic/pktio/null.c
jmtilli: Shouldn't be too hard to support, just needs some space to store the flag. But will do.

example/l2fwd_simple/odp_l2fwd_simple.c
jmtilli: Ah, it is deprecated in api-next. My patch, however, is to master as it has no API changes.

platform/linux-generic/pktio/null.c
```diff
@@ -0,0 +1,214 @@
...
+
+	return len;
+}
+
+#define PKTIO_NULL_MTU (64 * 1024)
+
+static uint32_t null_mtu_get(pktio_entry_t *pktio_entry ODP_UNUSED)
```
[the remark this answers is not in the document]
jmtilli: Not at least on master; it isn't deprecated there. Because my changes weren't an API change, I based the code on master instead of basing it on api-next. Of course I'm willing to support this patch by implementing the non-deprecated API as well, if this is ever merged to master and then via master to api-next.

example/l2fwd_simple/odp_l2fwd_simple.c
Bill-Fischofer-Linaro: At this point, since we're tagging Tiger Moth RC1 this week, all of api-next is being merged into master, so this will be on that base.

platform/linux-generic/pktio/null.c
```diff
@@ -0,0 +1,207 @@
...
+#include <stdint.h>
+#include <string.h>
+#include <net/if.h>
+#include <inttypes.h>
+#include <poll.h>
+#include <sys/ioctl.h>
+#include <errno.h>
+#include <sys/syscall.h>
+#include <linux/ethtool.h>
+#include <linux/sockios.h>
```
muvarov: please clean up include files almost all of them are not needed for null.

platform/linux-generic/pktio/null.c
muvarov: @lumag  any generation thing is bad when you need to reproduce something. I think all zeroes is good option here.

platform/linux-generic/pktio/null.c
```diff
@@ -0,0 +1,207 @@
...
+static int null_outqueues_config(pktio_entry_t *pktio_entry ODP_UNUSED,
+				 const odp_pktout_queue_param_t *p ODP_UNUSED)
+{
+	return 0;
+}
+
+const pktio_if_ops_t null_pktio_ops = {
+	.name = "null",
+	.print = NULL,
+	.init_global = NULL,
```
muvarov: I think other pktio have debug print that pktio is available on early init. Please keep the same style.

platform/linux-generic/pktio/null.c
lumag: @muvarov 
At least it should have local bit set on. I was thinking about using part of pktio name to generate mac.

platform/linux-generic/pktio/null.c
jmtilli: Cleaned up now.

platform/linux-generic/pktio/null.c
jmtilli: Done, the same style is used now in null pktio as well.

matiaselo:
In my opinion this interface would be more useful if you could also receive packets from it. `null_open()` has a packet pool argument anyway, so the device could allocate new packets from the pool during receive.

Packet length could be read from the pool argument `odp_pool_param_t.pkt.len` and the contents could be hardcoded to something reasonable. Valid Ethernet/IP/UDP headers?

jmtilli:
If you can receive packets from the null interface, it does not fulfill its purpose, unless the reception option can be turned off by the application. The purpose of null interface is to be just like /dev/null.

The idea here is that sometimes a user might want to run e.g. a mobile network gateway without Internet connection, for testing purposes. Then all packet going to Internet will be dropped, and no packets arrive from the Internet.

Or a firewall with two Internet connections: default and backup. You may want to use the null interface for the backup.

There are two options: (1) code the mobile network gateway / firewall / whatever to work without Internet interface, or (2) modify ODP to allow null interfaces, and use said null interface for the Internet interface in the mobile network application. I think option (2) is better, as there may be many users that want a null interface, and (2) needs to be done only once for ODP.

Now, if you want to read packets from some interface, why not use a pcap interface instead of null interface? Of course it requires ODP is compiled with libpcap... Of course, you could use an empty pcap as null interface replacement anyway, as long as two interfaces won't share the same .pcap file name.

Also, packet reception would be tricky with multi-queue. Which queue should receive the packets? Presumably there should be some kind of hash... pcap pktio solves this issue by not supporting multi-queue, but then again it's not applicable for modern high-performance applications alone for that reason.

matiaselo:
> On 8 Jan 2018, at 19:53, Juha-Matti Tilli <notifications@github.com> wrote:
> 
> If you can receive packets from the null interface, it does not fulfill its purpose, unless the reception option can be turned off by the application. The purpose of null interface is to be just like /dev/null.

This would still be possible by emptying the packet pool.

> 
> Now, if you want to read packets from some interface, why not use a pcap interface instead of null interface? Of course it requires ODP is compiled with libpcap... Of course, you could use an empty pcap as null interface replacement anyway, as long as two interfaces won't share the same .pcap file name.

Null pktio would be more simple than pcap and remove the external dependency. What I'm looking for here is to eventually use null pktio devices to implement validation tests. This requires adding some new API definitions (e.g. how to define packet contents) and would be done in a different patch set.

> 
> Also, packet reception would be tricky with multi-queue. Which queue should receive the packets? Presumably there should be some kind of hash... pcap pktio solves this issue by not supporting multi-queue, but then again it's not applicable for modern high-performance applications alone for that reason.
> 

In the initial version MQ reception doesn't have to be supported. Transmit side is trivial since it's only a packet free operation.

jmtilli:
<blockquote>
This would still be possible by emptying the packet pool.
</blockquote>

Can you please explain this a bit more? I have used ODP only for packet I/O, so I don't know how to empty a packet pool. Is emptying the packet pool an operation that only affects the null pktio? I thought that the packet pool is usually global, and there's one pool per application, so if it is emptied, it would presumably affect all interfaces the application uses.

Now, if it's possible to somehow use multiple pools, one per interface, what if some application is coded in such a manner that it uses just one pool? My intention of the null interface was that it works as a drop-in null interface in all ODP-using applications, with no application-specific code modifications required. It sounds to me like the application-specific code modifications would be required for emptying the packet pool.

<blockquote>
Null pktio would be more simple than pcap and remove the external dependency. What I'm looking for here is to eventually use null pktio devices to implement validation tests. This requires adding some new API definitions (e.g. how to define packet contents) and would be done in a different patch set.
</blockquote>

The intention was to not be an API modification.

matiaselo:
> This would still be possible by emptying the packet pool.
> Can you please explain this a bit more? I have used ODP only for packet I/O, so I don't know how to empty a packet pool. Is emptying the packet pool an operation that only affects the null pktio? I thought that the packet pool is usually global, and there's one pool per application, so if it is emptied, it would presumably affect all interfaces the application uses.
> 

odp_pktio_open() takes a packet pool argument and you can use a separate pool for each pktio device. So an empty pool doesn't effect other pktio devices. You can empty a pool simply by calling odp_packet_alloc() until it returns ODP_PACKET_INVALID (you can't free the packets until the application is stopped).

> Now, if it's possible to somehow use multiple pools, one per interface, what if some application is coded in such a manner that it uses just one pool? My intention of the null interface was that it works as a drop-in null interface in all ODP-using applications, with no application-specific code modifications required. It sounds to me like the application-specific code modifications would be required for emptying the packet pool.

Is it actually a problem that a null device would return packets on RX? It's hard to imagine a use case where you would call odp_pktin_recv() and want it to always return nothing. Special testing scenarios (e.g. for testing odp_pktin_recv_tmo()) could still be implemented using an empty packet pool.

> The intention was to not be an API modification.

Creating this kind of pktio device has been on our to-do list for a while. I'm trying to align this patch with our old idea so we could kill two birds with one stone.

example/generator/generator_run.sh
```diff
@@ -0,0 +1,17 @@
+#!/bin/bash
+#
+# Copyright (c) 2017, Linaro Limited
+# All rights reserved.
```
Bill-Fischofer-Linaro: 2018 copyright at this point, please.

platform/linux-generic/include/odp_packet_null.h
```diff
@@ -0,0 +1,16 @@
+/* Copyright (c) 2017, Linaro Limited
```
Bill-Fischofer-Linaro: 2018

platform/linux-generic/pktio/null.c
```diff
@@ -0,0 +1,175 @@
+/* Copyright (c) 2017, Linaro Limited
```
Bill-Fischofer-Linaro: 2018

platform/linux-generic/pktio/null.c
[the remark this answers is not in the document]
Bill-Fischofer-Linaro: I'd like to get clarification about this. Will discuss in tomorrow's ARCH call.

platform/linux-generic/pktio/null.c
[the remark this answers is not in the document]
Bill-Fischofer-Linaro: master is now at the Tiger Moth RC1 level so this can be updated.

platform/linux-generic/pktio/null.c
Bill-Fischofer-Linaro: Need to decide what's needed here.

jmtilli:
@Bill-Fischofer-Linaro you requested support for the new API. But there is no driver support yet, at least in origin/master:

    uint32_t ODP_DEPRECATE(odp_pktio_mtu)(odp_pktio_t pktio)
    {
            return pktio_mtu(pktio);
    }
    
    uint32_t odp_pktin_maxlen(odp_pktio_t pktio)
    {
            return pktio_mtu(pktio);
    }
    
    uint32_t odp_pktout_maxlen(odp_pktio_t pktio)
    {
            return pktio_mtu(pktio);
    }

I will fix the copyright years and push again. But before I can add support for pktin/pktout maxlen, the driver support infrastructure needs to be added.

Besides, the pktin/pktout maxlen code looks suspicious. Shouldn't it be 14 bytes larger than MTU because it includes Ethernet header?

Bill-Fischofer-Linaro:
@psavol Can you comment on @jmtilli comment above regarding `maxlen`?

psavol:
pktio_mtu() calls pktio driver specific mtu function. All drivers return max frame size, expect socket driver which returns still 1500 instead of 1514. This does not break anything, but leaves 14 bytes not used. Matias will send a patch to correct socket frame sizes.

psavol:
About recv call on null pktio. I agree with Matias that null pktio should return (dummy) packets as long as the pool is not empty. If application does not want to receive packets from a null pktio - it either does not call recv, or first makes sure that the pool is empty (by calling alloc or recv so many times that no packets are returned) and only after that assumes that recv does not return packets (but e.g. timeouts).

We are thinking to use null pktio for some simple pktio testing. Today application needs create and send packets (to loopback) before it can test interface, with null it could just do a receive call (which is simpler than create->send->receive).

Bill-Fischofer-Linaro:
@psavol If we're going to introduce the notion of a "null packet" it seems that should be formalized rather than left as something implementation-dependent (and hence non-portable). But that seems like a fairly major add at this stage in Tiger Moth.

I thought the original idea behind this was to have a a pktio equivalent to `/dev/null`, which discards anything sent to it and returns EOF when read.  So for a null pktio, a send is the equivalent of calling `odp_packet_free()`.  For RX,  presumably this would mean that `odp_pktin_recv()` would immediately return 0 to indicate no packets are available and `odp_pktin_recv_tmo()` would return 0 after the timeout expires. The only question would be what happens if `odp_pktin_recv_tmo()` is invoked with `ODP_PKTIN_WAIT` for a null device? We could either say this parameter is invalid for null devices or define some appropriate semantics (e.g., returns 0 immediately, or only when `odp_pktio_stop()` is called).

jmtilli:
I can indeed confirm that @Bill-Fischofer-Linaro 's description about the idea behind the null pktio patch is correct. The original plan was that it would receive no packets, and thus e.g. a mobile network gateway without Internet connection could use null pktio for the interface facing towards Internet in this case. If null pktio is receiving packets, it would be an entirely different animal then, and should be called something else than null then.

psavol:
OK. We can create another packet IO type for testing.

platform/linux-generic/pktio/null.c
muvarov: From other side it's not clear why VMs do not have this address set https://www.techrepublic.com/blog/data-center/mac-address-scorecard-for-common-virtual-machine-platforms/
Docker has 02:42:ae:7f:27:e9

but linux lo does not have any mac at all. Maybe do null we also don't need any mac?

muvarov:
odp_pktin_recv_tmo()  should return 0 after timeout.  

For ODP_PKTIN_WAIT we have 3 options:
1) Or it has to hang like read() operation. And if queue/pktio was destroyed/stopped with other thread  odp_pktin_recv_tmo() should error. But that has to be reflected in API doc.

2) we need remove ODP_PKTIN_WAIT. But that is worse idea because really huge timeout is about the same as infinite timeout.

3) we are ok that call can hang forever and it's up to application developer how to use this call.

Collecting 1, 2 and 3 I think 1 is the best solution. For example on socket case  odp_pktio_stop() can close socket and blocked recv() will return error. Other pktio roll in loops so it has to be easy for them to check pktio state.

I find it more flexible it app developer wants application spin inside pktio driver or in application loop. So my vote is for 1 solution.

Bill-Fischofer-Linaro:
@jmtilli Please update this to remove references to PR #341, which is now merged, and to add local bit to Mac addr as discussed. Thanks.

jmtilli:
@Bill-Fischofer-Linaro Hi, I have been busy lately, but I will find some time this week to do the changes, ASAP. About the option (1) as proposed by @muvarov, I'm not entirely sure how it should be implemented. Also, if option (1) is possible on Linux, it may prove infeasible on some other platforms. So my opinion would be to leave the behavior of waiting forever and then stopping the pktio during the wait undefined.

platform/linux-generic/pktio/null.c
jmtilli: The local bit is already set in the address. The binary representation shown is in inverse order, i.e. bit0, bit1, bit2, bit3, bit4, bit5, bit6, bit7 whereas the hexadecimal representation has bit7, bit6, bit5, bit4, bit3, bit2, bit1, bit0. So the local bit is (1<<1), i.e. 2. The 0x02 has the local bit and it also does not have the group bit, so it is a non-multicast locally administrated address.

jmtilli:
@Bill-Fischofer-Linaro The local bit is alerady set, as I said in the comments. So I'll just remove the mention not to merge this patch yet before PR #341.

Bill-Fischofer-Linaro:
I've confirmed that the checkpatch issue is a false positive. @lumag are you OK with this PR now?

lumag:
Yes.

muvarov:
Merged.

PR #366, opened 2017-12-23
[PATCH v1] linux-gen: _ishm: proper alignment for current page size

apalos:
fixes https://bugs.linaro.org/show_bug.cgi?id=3478
on systems compiled without hugetlb odp_init_global() failed
on systems with hugetlb alignment was superfluous

### DIFF
--- a/platform/linux-generic/_ishm.c
+++ b/platform/linux-generic/_ishm.c
@@ -864,8 +864,8 @@ int _odp_ishm_reserve(const char *name, uint64_t size, int fd,
 		 * the same address every where, otherwise alignment may be
 		 * be wrong for some process */
 		hp_align = align;
-		if (hp_align <= odp_sys_huge_page_size())
-			hp_align = odp_sys_huge_page_size();
+		if (hp_align <= page_hp_size)
+			hp_align = page_hp_size;
 		else
 			flags |= _ODP_ISHM_SINGLE_VA;
 
@@ -1441,6 +1441,8 @@ int _odp_ishm_init_global(void)
 	void *spce_addr;
 	int i;
 	uid_t uid;
+	char *hp_dir = odp_global_data.hugepage_info.default_huge_page_dir;
+	uint64_t align;
 
 	odp_global_data.main_pid = getpid();
 	odp_global_data.shm_dir = getenv("ODP_SHM_DIR");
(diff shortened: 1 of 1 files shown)


platform/linux-generic/_ishm.c
```diff
@@ -864,8 +864,8 @@ int _odp_ishm_reserve(const char *name, uint64_t size, int fd,
 		 * the same address every where, otherwise alignment may be
 		 * be wrong for some process */
 		hp_align = align;
-		if (hp_align <= odp_sys_huge_page_size())
-			hp_align = odp_sys_huge_page_size();
+		if (hp_align <= page_hp_size)
+			hp_align = page_hp_size;
```
Bill-Fischofer-Linaro: Presumably this has nothing to do with the bug fix and is just an efficiency improvement since on Line 804 above we have `page_hp_size = odp_sys_huge_page_size();`.

apalos: Correct

lumag:
Looks good. BTW: maybe we can change `odp_sys_huge_page_*` to fallback to normal pages if huge pages are not available?

Bill-Fischofer-Linaro:
@lumag Not sure that would be a good idea. One of the purposes of `odp_sys_huge_page_size()` is to determine whether huge pages are defined for this system, and that's done with the RC since it takes no arguments. 

If we're looking for a combined function, an `odp_sys_page_size_all()` API might be useful. The first element returned would be the (regular) page size and the subsequent ones would be the various huge page sizes available. Like `odp_sys_huge_page_size_all()` the RC would indicate the total number of different sizes, with 1 being the case for no huge pages.

muvarov:
@lumag it looks like allocation to specific hp memory does not work now. Only default memory works but shared files can be redefined with char *shm_dir = getenv("ODP_SHM_DIR");. In that case it might be some other directory with other huge page sizes. So for quick look if we want to support various huge pages we need a little bit rework memory allocation functions. I think this PR can go in as bugfix. And later we can correct allocation.

lumag:
@muvarov yes, I'm fine with mering this PR. I just think that API supporting memory allocation should be rethought.

muvarov:
Merged.

PR #368, opened 2017-12-26
[PATCH CATERPILLAR v1] remove some metadata with less effort from odp_buffer

JoyceKong-arm:


platform/linux-dpdk/pool/dpdk.c
```diff
@@ -194,7 +194,7 @@ static void
 odp_dpdk_mbuf_ctor(struct rte_mempool *mp,
 		   void *opaque_arg,
 		   void *raw_mbuf,
-		   unsigned i)
+		   unsigned i ODP_UNUSED)
```
Bill-Fischofer-Linaro: Since this is an internal routine, why not just drop the parameter rather than marking it `ODP_UNUSED` if it's no longer needed?

JoyceKong-arm: This is a combination of rte_pktmbuf_init in rte_mbuf.c . So the style needs to the same as rte_pktmbuf_init in dpdk , otherwise there would be some error.

Bill-Fischofer-Linaro: Ok.

heyi-linaro:
Merged and close.

PR #374, opened 2017-12-27
[PATCH API-NEXT v3] api: crypto: deprecate DES algorithm

lumag:
DES is too weak, 56-bit keys, 64-bit blocks. Can be cracked nearly
bare-handed nowdays. Depreate it now, to be dropped in next release.
Note: Tripple-DES (TDES) remains in place.

### DIFF
--- a/include/odp/api/spec/crypto.h
+++ b/include/odp/api/spec/crypto.h
@@ -71,9 +71,6 @@ typedef enum {
 	/** No cipher algorithm specified */
 	ODP_CIPHER_ALG_NULL,
 
-	/** DES */
-	ODP_CIPHER_ALG_DES,
-
 	/** Triple DES with cipher block chaining */
 	ODP_CIPHER_ALG_3DES_CBC,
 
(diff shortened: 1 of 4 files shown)


lumag:
Well, I would point to [RFC 8221](https://tools.ietf.org/html/rfc8221#section-5).

Updated PR to fix validation tests and IPsec code.

Bill-Fischofer-Linaro:
Not suggesting anything be done with TDES for now, but NIST is limiting a key set to 8MB, which severely limits its utility, unless you like to renegotiate a lot.

lumag:
I think that such limits should be handled by upper layer.

psavol:
ODP API spec does not need to give recommendations which algorithms to use. API enumerates algos which implementations may implement and applications may require/use. Since DES has still SW (openSSL, DPDK) and HW implementations, there's no need to remove the enumeration. Algos may be used for demonstration, etc purposes where strength does not matter. Validation test passes also when DES is not implemented (not in capability).

lumag:
@psavol closing then.

PR #380, opened 2018-01-04
[PATCH CATERPILLAR v6] Caterpillar mdev i40e

MykytaI:
Add mediated i40e driver along with common mdev code.

### DIFF
--- a/platform/linux-generic/pktio/mdev/i40e.c
+++ b/platform/linux-generic/pktio/mdev/i40e.c
@@ -1,0 +1,641 @@
+/* Copyright (c) 2018, Linaro Limited
+ * All rights reserved.
+ *
+ * SPDX-License-Identifier:     BSD-3-Clause
+ */
+
+#include "config.h"
+
+#ifdef ODP_MDEV
+
+#include <linux/types.h>
+#include <protocols/eth.h>
+#include <stdint.h>
+#include <stdio.h>
+#include <string.h>
+#include <sys/mman.h>
+#include <unistd.h>
+
+#include <odp_packet_io_internal.h>
+#include <odp_posix_extensions.h>
+
+#include <odp/api/hints.h>
+#include <odp/api/packet.h>
+#include <odp/api/plat/packet_inlines.h>
+#include <odp/drv/hints.h>
+#include <odp/drv/mmio.h>
+
+#include <pktio/common.h>
+#include <pktio/ethtool.h>
+#include <pktio/mdev.h>
+#include <pktio/sysfs.h>
+#include <pktio/uapi_net_mdev.h>
+
+#define MODULE_NAME "i40e"
+
+#define I40E_TX_BUF_SIZE 2048U
+#define I40E_RX_BUF_SIZE 2048U
+
+#define I40E_TX_PACKET_LEN_MIN 17
+
+/* RX queue definitions */
+#define I40E_RX_QUEUE_NUM_MAX 32
+
+/** RX descriptor */
+typedef struct {
+	odp_u64le_t addr;
+#define I40E_RXD_QW1_LEN_S 38
+#define I40E_RXD_QW1_LEN_M (0x3FFFULL << I40E_RXD_QW1_LEN_S)
+	odp_u64le_t status_error_len;
+	odp_u64le_t reserved[2];
+} i40e_rx_desc_t;
+
+/** RX queue data */
+typedef struct ODP_ALIGNED_CACHE {
+	i40e_rx_desc_t *rx_descs;	/**< RX queue base */
+	odp_u32le_t *doorbell;		/**< RX queue doorbell */
+
+	uint16_t rx_queue_len;		/**< Number of RX desc entries */
+	uint16_t cidx;			/**< Next RX desc to handle */
+
+	mdev_dma_area_t rx_data;	/**< RX packet payload area */
+
+	odp_ticketlock_t lock;		/**< RX queue lock */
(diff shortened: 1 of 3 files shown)


MykytaI:
Checkpatch failure seems to be some glitch, as both checkpatch.pl on my workstation and in Travis are happy about the patches.

Bill-Fischofer-Linaro:
@MykytaI Travis reports checkpatch success. The "checkpatch failed" callout is a glitch in the reporting scripts Maxim is using. It's been a bit flakey for a while and is something Maxim can look at when he returns.

platform/linux-generic/m4/odp_mdev.m4
```diff
@@ -0,0 +1,34 @@
+##########################################################################
+# Enable mdev support
+##########################################################################
+mdev_support=no
```
Bill-Fischofer-Linaro: This is fine as a default, but we need:

1. The `configure` script should include this setting (yes/no) in its summary printout of configuration options

2. Travis needs to include appropriate test variants with `--enable-mdev-support=yes`. Otherwise we're just including uncompiled and untested code, which isn't good.

3. Do we need any updates to the validation tests to exercise this code? Those can be a follow-on PR as long as the basics compile.

When I manually run `./configure --enable-mdev-support=yes` I get the following:

```
  CC       pktio/mdev/i40e.lo
pktio/mdev/i40e.c:31:10: fatal error: uapi/net_mdev.h: No such file or directory
 #include <uapi/net_mdev.h>
          ^~~~~~~~~~~~~~~~~
compilation terminated.
Makefile:1234: recipe for target 'pktio/mdev/i40e.lo' failed
make[1]: *** [pktio/mdev/i40e.lo] Error 1
make[1]: Leaving directory '/home/bill/linaro/hubodp/platform/linux-generic'
Makefile:528: recipe for target 'all-recursive' failed
make: *** [all-recursive] Error 1
```

I realize this PR has external dependencies, but they need to be documented so that they can be provided to permit compilation and testing. We can't include uncompilable code in the repo.

apalos: This is probably a typo since the required header is located in platform/linux-generic/pktio/uapi_net_mdev.h

platform/linux-generic/pktio/uapi_net_mdev.h
```diff
@@ -0,0 +1,33 @@
...
+	VFIO_NET_DESCRIPTORS,
+	VFIO_NET_MMIO,
+};
+
+enum net_mdev_subtypes {
+	VFIO_NET_MDEV_STATS,
+	VFIO_NET_MDEV_RX,
+	VFIO_NET_MDEV_TX,
+	VFIO_NET_MDEV_BARS,
+};
```
lumag: Subtypes of which type? Or are all subtypes applicable to all types?

apalos: At the moment MMIO is using subtype BARS and DESCRIPTORS are using Rx/Tx
Shadow is unused

lumag: I then suggest to rename subtypes, so that they will reference the main type.

lumag: like `VFIO_NET_MDEV_MMIO_BARS`, `V_N_M_DESCRIPTORS_RX`, etc.

Bill-Fischofer-Linaro: Agree with @lumag that having names the reflect the hierarchy will be helpful in avoiding errors.

platform/linux-generic/m4/odp_mdev.m4
MykytaI: Correct, this is indeed typo.

platform/linux-generic/pktio/uapi_net_mdev.h
MykytaI: Ok, will be addressed in the patch revision.

MykytaI: New separate PR for common code (patch #1 in this PR) is created: https://github.com/Linaro/odp/pull/384
This PR and other driver-specific ones will be rebased as soon as #384 gets merged.

platform/linux-generic/pktio/mdev/i40e.c
```diff
@@ -0,0 +1,641 @@
...
+		if (odp_unlikely(pkt_len < I40E_TX_PACKET_LEN_MIN)) {
+			tx_pkts++;
+			continue;
+		}
+
+		/* Skip oversized packets silently */
+		if (odp_unlikely(pkt_len > I40E_TX_BUF_SIZE)) {
+			tx_pkts++;
+			continue;
+		}
```
Bill-Fischofer-Linaro: Do you plan to have different behavior in the future for undersized vs. oversized packets? If not, it would seem better to combine these into a single `if` test.

heyi-linaro:
Merged and closed.

platform/linux-generic/pktio/mdev/i40e.c
MykytaI: We'll need to add statistics counters to all these drivers at some point, so I thought that it's better to keep 'oversized' and 'undersized' separately with an intention to update different counters. If we find out that we don't have separate counters (I guess stats are covered by some RFC, isn't it ?) -- then we definitely can merge those into a single condition check.

PR #384, opened 2018-01-08
[PATCH CATERPILLAR v3] linux-gen: mediated devices common code

MykytaI:
Add APIs to create/destroy mediated devices, manipulate their regions
and allocate/free DMA areas accessible by both ODP and HW.

### DIFF
--- a/platform/linux-generic/Makefile.am
+++ b/platform/linux-generic/Makefile.am
@@ -177,10 +177,12 @@ noinst_HEADERS = \
 		  include/odp_packet_io_pool.h \
 		  include/odp_packet_io_pool_access.h \
 		  include/odp_packet_io_ring_internal.h \
-		  pktio/ethtool.h \
 		  pktio/common.h \
-		  pktio/sysfs.h \
 		  pktio/dpdk.h \
+		  pktio/ethtool.h \
+		  pktio/mdev.h \
+		  pktio/sysfs.h \
+		  pktio/uapi_net_mdev.h \
 		  include/odp_pktio_ops_ipc.h \
 		  include/odp_pktio_ops_loopback.h \
 		  include/odp_pktio_ops_netmap.h \
@@ -243,18 +245,19 @@ __LIB__libodp_linux_la_SOURCES = \
 			   odp_packet_flags.c \
 			   odp_packet_io.c \
 			   odp_packet_io_pool.c \
-			   pktio/ethtool.c \
-			   pktio/subsystem.c \
-			   pktio/ipc.c \
 			   pktio/common.c \
-			   pktio/loopback.c \
-			   pktio/netmap.c \
 			   pktio/dpdk.c \
+			   pktio/ethtool.c \
+			   pktio/ipc.c \
+			   pktio/loopback.c \
+			   pktio/mdev.c \
+			   pktio/netmap.c \
+			   pktio/ring.c \
 			   pktio/socket.c \
 			   pktio/socket_mmap.c \
+			   pktio/subsystem.c \
 			   pktio/sysfs.c \
 			   pktio/tap.c \
-			   pktio/ring.c \
 			   pool/generic.c \
 			   pool/subsystem.c \
 			   odp_pkt_queue.c \
--- a/platform/linux-generic/pktio/mdev.c
+++ b/platform/linux-generic/pktio/mdev.c
@@ -1,0 +1,593 @@
+/* Copyright (c) 2018, Linaro Limited
+ * All rights reserved.
+ *
+ * SPDX-License-Identifier:     BSD-3-Clause
+ */
+
+#include "config.h"
+
+#ifdef ODP_MDEV
+
+#include <dirent.h>
+#include <errno.h>
+#include <fcntl.h>
+#include <stddef.h>
+#include <stdint.h>
+#include <stdio.h>
+#include <stdlib.h>
+#include <string.h>
+#include <sys/ioctl.h>
+#include <sys/mman.h>
+#include <unistd.h>
+
+#include <linux/vfio.h>
+
+#include <odp_posix_extensions.h>
+#include <odp_align_internal.h>
+#include <odp_debug_internal.h>
+
+#include <odp/api/hints.h>
+
+#include <pktio/mdev.h>
+#include <pktio/uapi_net_mdev.h>
+
+/**
+ * returns a valid VFIO container
+ * fd must be closed by caller
+ */
+static int get_container(void)
+{
+	int container;
+	int ret;
+
+	/* Create a new container */
+	container = open("/dev/vfio/vfio", O_RDWR);
+	if (container < 0) {
+		ODP_ERR("Failed to create new VFIO container\n");
+		goto out;
+	}
+
+	ret = ioctl(container, VFIO_GET_API_VERSION);
+	if (ret != VFIO_API_VERSION) {
+		ODP_ERR("VFIO API version mismatch: expected %i, got %i\n",
+			VFIO_API_VERSION, ret);
+		goto out;
+	}
+
+	ret = ioctl(container, VFIO_CHECK_EXTENSION, VFIO_TYPE1_IOMMU);
+	if (!ret) {
+		ODP_ERR("Container doesn't support VFIO_TYPE1_IOMMU\n");
+		goto out;
+	}
+
+	return container;
+
+out:
+	if (container != -1)
+		close(container);
+	return -1;
+}
+
+/**
+ * returns a valid VFIO group
+ * fd must be closed by caller
+ */
+static int get_group(int grp_id)
+{
+	char path[32];
+	int group;
+	int ret;
+	struct vfio_group_status group_status = {
+		.argsz = sizeof(group_status)
+	};
+
+	snprintf(path, sizeof(path), "/dev/vfio/%d", grp_id);
+	group = open(path, O_RDWR);
+	if (group < 0) {
+		ODP_ERR("Failed to open %s: %s\n", path, strerror(errno));
+		goto out;
+	}
+
+	ret = ioctl(group, VFIO_GROUP_GET_STATUS, &group_status);
+	if (ret < 0) {
+		ODP_ERR("Failed to get group status\n");
+		goto out;
+	}
+
+	/* Test the group is viable and available */
+	if (!(group_status.flags & VFIO_GROUP_FLAGS_VIABLE)) {
+		ODP_ERR("Group is not viable\n");
+		goto out;
+	}
+
+	return group;
+
+out:
+	if (group != -1)
+		close(group);
+	return -1;
+}
+
+static void vfio_find_sparse_mmaps(struct vfio_info_cap_header *hdr,
+				   struct vfio_region_info_cap_sparse_mmap
+				   **sparse)
+{
+	*sparse =
+	    odp_container_of(hdr, struct vfio_region_info_cap_sparse_mmap,
+			     header);
+}
+
+static struct vfio_info_cap_header *
+vfio_get_region_info_cap(struct vfio_region_info *info, __u16 id)
+{
+	struct vfio_info_cap_header *hdr;
+	void *ptr = info;
+
+	if (!(info->flags & VFIO_REGION_INFO_FLAG_CAPS))
+		return NULL;
+
+	for (hdr =
+	     (struct vfio_info_cap_header *)(void *)((uint8_t *)ptr
+						     + info->cap_offset);
+	     hdr != ptr;
+	     hdr = (struct vfio_info_cap_header *)(void *)((uint8_t *)ptr
+							   + hdr->next)) {
+		if (hdr->id == id)
+			return hdr;
+	}
+
+	return NULL;
+}
+
+static struct vfio_info_cap_header *
+vfio_get_cap_info(struct vfio_region_info *region_info, __u16 id)
+{
+	struct vfio_info_cap_header *caps = NULL;
+
+	caps = vfio_get_region_info_cap(region_info, id);
+
+	return caps;
+}
+
+int vfio_get_region_sparse_mmaps(struct vfio_region_info *region_info,
+				 struct vfio_region_info_cap_sparse_mmap
+				 **sparse)
+{
+	struct vfio_info_cap_header *caps = NULL;
+	int ret = -ENOENT;
+
+	if (region_info->flags & VFIO_REGION_INFO_FLAG_CAPS &&
+	    region_info->argsz > sizeof(*region_info)) {
+		caps = vfio_get_cap_info(region_info,
+					 VFIO_REGION_INFO_CAP_SPARSE_MMAP);
+		if (!caps)
+			goto out;
+		vfio_find_sparse_mmaps(caps, sparse);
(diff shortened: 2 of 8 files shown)


platform/linux-generic/pktio/mdev.c
```diff
@@ -0,0 +1,585 @@
...
+	region_info->index = region;
+	region_info->argsz = sizeof(*region_info);
+	ret = ioctl(mdev->device, VFIO_DEVICE_GET_REGION_INFO, region_info);
+	if (ret < 0) {
+		ODP_ERR("Failed to get PCI region info\n");
+		goto out;
+	}
+
+	if (!region_info->size) {
+		ODP_DBG("region info %d is empty, skipping\n", region);
```
Bill-Fischofer-Linaro: It's not clear why the region not existing results in an `ODP_ERR()` but it being empty is an `ODP_DBG()`. It seems both should be `ODP_DBG()` calls here, since the caller gets a `NULL` return in either case and presumably continues.

platform/linux-generic/pktio/mdev.c
```diff
@@ -0,0 +1,585 @@
...
+			goto out;
+
+		region_info = tmp;
+
+		ODP_DBG("region info %d with extended capabilities size: %u\n",
+			region, region_info->argsz);
+		ret = ioctl(mdev->device, VFIO_DEVICE_GET_REGION_INFO,
+			    region_info);
+		if (ret < 0 || !region_info->size) {
+			ODP_ERR("Failed to get PCI region info\n");
```
Bill-Fischofer-Linaro: This is inconsistent with the earlier tests above. `ret < 0` above gives `ODP_ERR()` while `!region_info->size` gets an `ODP_DBG()`. It seems `ODP_DBG()` should be used consistently here.

platform/linux-generic/pktio/mdev.c
```diff
@@ -0,0 +1,585 @@
...
+
+	dma_area->iova = req.iova;
+
+	ODP_DBG("dma_area alloc: %llx@%llx -> %llx\n", dma_area->size,
+		dma_area->vaddr, dma_area->iova);
+
+	return 0;
+}
+
+int mdev_dma_area_free(mdev_device_t *mdev, mdev_dma_area_t *dma_area)
```
Bill-Fischofer-Linaro: This routine says `free` but it appears to perform `unmap` functions rather than resource frees. Would `mdev_dma_area_unmap()` be a better name here?

platform/linux-generic/pktio/mdev.c
```diff
@@ -0,0 +1,585 @@
...
+		return addr;
+
+	mdev->mappings[mdev->mappings_count].addr = addr;
+	mdev->mappings[mdev->mappings_count].size = size;
+	mdev->mappings_count++;
+
+	return addr;
+}
+
+int mdev_dma_area_alloc(mdev_device_t *mdev, mdev_dma_area_t *dma_area)
```
Bill-Fischofer-Linaro: This routine says `alloc` but it seems to perform `map` functions here. Would `mdev_dma_area_map()` be a better name?

platform/linux-generic/pktio/mdev.c
```diff
@@ -0,0 +1,585 @@
...
+	}
+
+	snprintf(sysfs_path, sizeof(sysfs_path),
+		 "/sys/class/net/%s/device/mdev_supported_types/%s-netmdev/devices/",
+		 if_name, driver);
+
+	dir = opendir(sysfs_path);
+	if (!dir)
+		return -1;
+	/* FIXME only the last uuid will be returned now */
```
Bill-Fischofer-Linaro: What is needed to resolve/add this? Is this intended to be a follow-on PR?

platform/linux-generic/pktio/mdev.c
```diff
@@ -0,0 +1,585 @@
...
+	if (device < 0) {
+		ODP_ERR("Failed to get device FD\n");
+		goto out;
+	}
+
+	/* Test and setup the device */
+	ret = ioctl(device, VFIO_DEVICE_GET_INFO, dev_info);
+	if (ret < 0) {
+		ODP_ERR("Failed to get device info\n");
+		goto out;
```
Bill-Fischofer-Linaro: `device` is not reset here so it will still be returned if this test fails. Is that intentional?

platform/linux-generic/pktio/mdev.c
```diff
@@ -0,0 +1,585 @@
...
+
+		region_info = vfio_get_region(mdev, region);
+		if (!region_info)
+			continue;
+
+		ret = region_info_cb(mdev, region_info);
+		if (ret < 0) {
+			ODP_ERR("Region info cb fail on region_info[%u]\n",
+				region);
+			return -1;
```
Bill-Fischofer-Linaro: Doesn't this return leak `region_info` here? You free it only after this check.

platform/linux-generic/pktio/mdev.c
```diff
@@ -0,0 +1,585 @@
...
+#include <unistd.h>
+#include <stdio.h>
+#include <stdint.h>
+#include <linux/vfio.h>
+#include <stddef.h>
+#include <stdlib.h>
+#include <dirent.h>
+
+#include <config.h>
+#include <odp_posix_extensions.h>
```
Bill-Fischofer-Linaro: Duplicate `#include`? This is included above on line 11.

platform/linux-generic/pktio/mdev.c
apalos: This effectively maps Rx/Tx descriptors and a failure indicates that the card will not be operational, that's why it's under ODR_ERR

platform/linux-generic/pktio/mdev.c
apalos: This call tries to get the area info that's exported from the linux kernel. We loop through all PCI areas which is PCI BARS + Driver defined areas for Rx/Tx descriptors.
Some(usually all but one) PCI BARS are empty, but it's normal. That's why we only issue a debug message here

platform/linux-generic/pktio/mdev.c
apalos: This will need a new PR for multiple card support. we need to associate the NICs name(eth0/esp4s0 etc), with the appropriate uuid that was created. For now we assume only one card will be plugged in as a mediated device so we just return the last entry of the directory(which is actually the only entry)

platform/linux-generic/pktio/mdev.c
```diff
@@ -0,0 +1,585 @@
...
+	if (!ioctl(container, VFIO_CHECK_EXTENSION, VFIO_TYPE1_IOMMU)) {
+		ODP_ERR("Doesn't support the IOMMU driver we want\n");
+		goto out;
+	}
+
+	return container;
+out:
+	close(container);
+	container = -1;
+	return ret;
```
heyi-linaro: "ret" is arbitrary, zero to indicate success and non-zero (either negative or positive) to indicate failures?

platform/linux-generic/pktio/mdev.c
```diff
@@ -0,0 +1,585 @@
...
+	if (!(group_status.flags & VFIO_GROUP_FLAGS_VIABLE)) {
+		ODP_ERR("Group is not viable\n");
+		goto out;
+	}
+
+	return group;
+out:
+	close(group);
+	group = -1;
+	return ret;
```
heyi-linaro: Negative to indicate error and 0 and positive to indicate success?

platform/linux-generic/Makefile.am
```diff
@@ -255,6 +257,7 @@ __LIB__libodp_linux_la_SOURCES = \
 			   pktio/sysfs.c \
 			   pktio/tap.c \
 			   pktio/ring.c \
+			   pktio/mdev.c \
```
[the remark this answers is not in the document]
MykytaI: I've reordered this and previous one (pktio/*.[ch]) but left the rest as is as sorting everything is too much of a change to be in context of this PR.

platform/linux-generic/pktio/mdev.c
MykytaI: Thanks, removed 2 duplicated headers.

platform/linux-generic/pktio/mdev.c
```diff
@@ -0,0 +1,585 @@
...
+		return container;
+
+	ret = ioctl(container, VFIO_GET_API_VERSION);
+	if (ret != VFIO_API_VERSION) {
+		ODP_ERR("Unknown API version\n");
+		goto out;
+	}
+
+	if (!ioctl(container, VFIO_CHECK_EXTENSION, VFIO_TYPE1_IOMMU)) {
+		ODP_ERR("Doesn't support the IOMMU driver we want\n");
```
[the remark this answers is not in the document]
MykytaI: Thanks, fixed

platform/linux-generic/pktio/mdev.c
```diff
@@ -0,0 +1,585 @@
...
+	int container;
+	/* Create a new container */
+	container = open("/dev/vfio/vfio", O_RDWR);
+
+	if (container < 0)
+		return container;
+
+	ret = ioctl(container, VFIO_GET_API_VERSION);
+	if (ret != VFIO_API_VERSION) {
+		ODP_ERR("Unknown API version\n");
```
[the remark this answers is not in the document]
MykytaI: Thanks, fixed.

platform/linux-generic/pktio/mdev.c
MykytaI: Thanks, fixed

platform/linux-generic/pktio/mdev.c
```diff
@@ -0,0 +1,585 @@
...
+	return container;
+out:
+	close(container);
+	container = -1;
+	return ret;
+}
+
+/**
+ * returns a valid VFIO group
+ * fd must be close by caller
```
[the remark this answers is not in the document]
MykytaI: Thanks, fixed.

platform/linux-generic/pktio/mdev.c
```diff
@@ -0,0 +1,585 @@
...
+ */
+static int get_group(int grp_id)
+{
+	char path[64];
```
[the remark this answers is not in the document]
MykytaI: Thanks, fixed.

platform/linux-generic/pktio/mdev.c
MykytaI: Thanks, fixed.

platform/linux-generic/pktio/mdev.c
MykytaI: The difference here is because 0-sized region is OK and we don't care about it, but on the other hand we would like to spot GET_REGION_INFO failures even with debug disabled. I would prefer to keep it as is.

platform/linux-generic/pktio/mdev.c
MykytaI: I agree with Bill in some sense here. It indeed needs to be consistent, but instead of making it ODP_DBG I'll just reorder region->size check.

platform/linux-generic/pktio/mdev.c
MykytaI: The map/unmap assumes that resource is already allocated and provided among function arguments, but it's not the case here. Actual allocation is done in kernel context, so I would prefer to keep the name as is for now. This is going to be changed anyway drastically once we get zerocopy implemented.

platform/linux-generic/pktio/mdev.c
MykytaI: See above.

platform/linux-generic/pktio/mdev.c
MykytaI: Actually, there shall be exactly one UUID per kernel netdevice as we have it like that on the kernel part right now. I'm in doubt that it will change ... So maybe we can just drop the FIXME.

platform/linux-generic/pktio/mdev.c
MykytaI: Thanks, fixed.

platform/linux-generic/pktio/mdev.c
MykytaI: Thanks, fixed.

platform/linux-generic/pktio/mdev.c
```diff
@@ -0,0 +1,589 @@
...
+	struct vfio_info_cap_header *hdr;
+	void *ptr = info;
+
+	if (!(info->flags & VFIO_REGION_INFO_FLAG_CAPS))
+		return NULL;
+
+	for (hdr =
+	     (struct vfio_info_cap_header *)((char *)ptr + info->cap_offset);
+	     hdr != ptr;
+	     hdr = (struct vfio_info_cap_header *)((char *)ptr + hdr->next)) {
```
Bill-Fischofer-Linaro: Specifying `./configure --enable-mdev-support=yes CC=clang` results in lines 130 and 132 being flagged by clang:
```
  CC       pktio/mdev.lo
pktio/mdev.c:130:7: error: cast from 'char *' to 'struct vfio_info_cap_header *'
      increases required alignment from 1 to 4 [-Werror,-Wcast-align]
             (struct vfio_info_cap_header *)((char *)ptr + info->cap_offset);
             ^~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~
pktio/mdev.c:132:13: error: cast from 'char *' to
      'struct vfio_info_cap_header *' increases required alignment from 1 to 4
      [-Werror,-Wcast-align]
             hdr = (struct vfio_info_cap_header *)((char *)ptr + hdr->next)) {
                   ^~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~
2 errors generated.
Makefile:1222: recipe for target 'pktio/mdev.lo' failed
make[1]: *** [pktio/mdev.lo] Error 1
```

platform/linux-generic/pktio/mdev.c
Bill-Fischofer-Linaro: OK.

platform/linux-generic/pktio/mdev.c
Bill-Fischofer-Linaro: OK

platform/linux-generic/pktio/mdev.c
Bill-Fischofer-Linaro: Either would be good. This can always be enhanced in follow-on PRs, but each PR should stand alone.

MykytaI: Thanks, updated code according to an assumption that we have just one file in the directory which is UUID we're looking for.

platform/linux-generic/pktio/mdev.c
MykytaI: I've fixed clang compilation.

I guess if we insert compilation test into .travis.yml without proper Ubuntu 17.10 docker images -- it will fail on every upcoming PR which doesn't make much sense. I believe it shall be in separate PR accompanied with all prerequisite changes in Travis images. What do you think ?

heyi-linaro:
Merged and closed.

PR #385, opened 2018-01-09
[PATCH CATERPILLAR v5] linux-dpdk: crypto support for cipher & auth only feature

GBalakrishna:
This patch introduces the support for Cipher only & Auth only crypto operation

### DIFF
--- a/platform/linux-dpdk/odp_crypto.c
+++ b/platform/linux-dpdk/odp_crypto.c
@@ -31,6 +31,14 @@
 #define MAX_SESSIONS 2048
 #define NB_MBUF  8192
 
+enum crypto_chain_order {
+	CRYPTO_CHAIN_ONLY_CIPHER,
+	CRYPTO_CHAIN_ONLY_AUTH,
+	CRYPTO_CHAIN_CIPHER_AUTH,
+	CRYPTO_CHAIN_AUTH_CIPHER,
+	CRYPTO_CHAIN_NOT_SUPPORTED
+};
+
 typedef struct crypto_session_entry_s crypto_session_entry_t;
 struct crypto_session_entry_s {
 		struct crypto_session_entry_s *next;
@@ -108,7 +116,8 @@ static int cipher_alg_odp_to_rte(odp_cipher_alg_t cipher_alg,
 }
 
 static int auth_alg_odp_to_rte(odp_auth_alg_t auth_alg,
-			       struct rte_crypto_sym_xform *auth_xform)
+			       struct rte_crypto_sym_xform *auth_xform,
+			       uint32_t auth_digest_len)
 {
 	int rc = 0;
 
@@ -116,35 +125,45 @@ static int auth_alg_odp_to_rte(odp_auth_alg_t auth_alg,
 	switch (auth_alg) {
 	case ODP_AUTH_ALG_NULL:
 		auth_xform->auth.algo = RTE_CRYPTO_AUTH_NULL;
+		auth_xform->auth.digest_length = auth_digest_len;
 		break;
-	case ODP_AUTH_ALG_MD5_HMAC:
 #if ODP_DEPRECATED_API
 	case ODP_AUTH_ALG_MD5_96:
-#endif
 		auth_xform->auth.algo = RTE_CRYPTO_AUTH_MD5_HMAC;
 		auth_xform->auth.digest_length = 12;
 		break;
-	case ODP_AUTH_ALG_SHA256_HMAC:
+#endif
+	case ODP_AUTH_ALG_MD5_HMAC:
+		auth_xform->auth.algo = RTE_CRYPTO_AUTH_MD5_HMAC;
+		auth_xform->auth.digest_length = auth_digest_len;
+		break;
 #if ODP_DEPRECATED_API
 	case ODP_AUTH_ALG_SHA256_128:
-#endif
 		auth_xform->auth.algo = RTE_CRYPTO_AUTH_SHA256_HMAC;
 		auth_xform->auth.digest_length = 16;
 		break;
+#endif
+	case ODP_AUTH_ALG_SHA256_HMAC:
+		auth_xform->auth.algo = RTE_CRYPTO_AUTH_SHA256_HMAC;
+		auth_xform->auth.digest_length = auth_digest_len;
+		break;
 	case ODP_AUTH_ALG_SHA1_HMAC:
 		auth_xform->auth.algo = RTE_CRYPTO_AUTH_SHA1_HMAC;
-		auth_xform->auth.digest_length = 20;
+		auth_xform->auth.digest_length = auth_digest_len;
 		break;
 	case ODP_AUTH_ALG_SHA512_HMAC:
 		auth_xform->auth.algo = RTE_CRYPTO_AUTH_SHA512_HMAC;
-		auth_xform->auth.digest_length = 64;
+		auth_xform->auth.digest_length = auth_digest_len;
 		break;
-	case ODP_AUTH_ALG_AES_GCM:
 #if ODP_DEPRECATED_API
 	case ODP_AUTH_ALG_AES128_GCM:
-#endif
 		auth_xform->auth.algo = RTE_CRYPTO_AUTH_AES_GCM;
 		auth_xform->auth.digest_length = 16;
+		break;
+#endif
+	case ODP_AUTH_ALG_AES_GCM:
+		auth_xform->auth.algo = RTE_CRYPTO_AUTH_AES_GCM;
+		auth_xform->auth.digest_length = auth_digest_len;
 		break;
 	default:
 		rc = -1;
@@ -538,9 +557,9 @@ int odp_crypto_cipher_capability(odp_cipher_alg_t cipher,
 		i = 0;
 		cap = &dev_info.capabilities[i];
 		while (cap->op != RTE_CRYPTO_OP_TYPE_UNDEFINED) {
-			cap_cipher_algo = cap->sym.cipher.algo;
 			if (cap->sym.xform_type ==
 			    RTE_CRYPTO_SYM_XFORM_CIPHER) {
+				cap_cipher_algo = cap->sym.cipher.algo;
 				if (cap_cipher_algo == cipher_xform.cipher.algo)
 						break;
 			}
@@ -594,7 +613,7 @@ int odp_crypto_auth_capability(odp_auth_alg_t auth,
 	enum rte_crypto_auth_algorithm cap_auth_algo;
 	struct rte_crypto_sym_xform auth_xform;
 
-	rc = auth_alg_odp_to_rte(auth, &auth_xform);
+	rc = auth_alg_odp_to_rte(auth, &auth_xform, 0);
 
 	/* Check result */
 	if (rc)
@@ -615,7 +634,7 @@ int odp_crypto_auth_capability(odp_auth_alg_t auth,
 		while (cap->op != RTE_CRYPTO_OP_TYPE_UNDEFINED) {
 			cap_auth_algo = cap->sym.auth.algo;
 			if (cap->sym.xform_type ==
-			    RTE_CRYPTO_SYM_XFORM_CIPHER) {
+			    RTE_CRYPTO_SYM_XFORM_AUTH) {
 				if (cap_auth_algo == auth_xform.auth.algo)
 						break;
 			}
@@ -663,16 +682,37 @@ int odp_crypto_auth_capability(odp_auth_alg_t auth,
 	return idx;
 }
 
-static int get_crypto_dev(struct rte_crypto_sym_xform *cipher_xform,
-			  struct rte_crypto_sym_xform *auth_xform,
+static int get_crypto_dev(struct rte_crypto_sym_xform *first_xform,
+			  enum crypto_chain_order order,
 			  uint16_t iv_length, uint8_t *dev_id)
 {
 	uint8_t cdev_id, id;
 	const struct rte_cryptodev_capabilities *cap;
+	struct rte_crypto_sym_xform *auth_xform = NULL;
+	struct rte_crypto_sym_xform *cipher_xform = NULL;
 	enum rte_crypto_cipher_algorithm cap_cipher_algo;
 	enum rte_crypto_auth_algorithm cap_auth_algo;
 	enum rte_crypto_cipher_algorithm app_cipher_algo;
 	enum rte_crypto_auth_algorithm app_auth_algo;
+
+	switch (order) {
+	case CRYPTO_CHAIN_ONLY_CIPHER:
+		cipher_xform = first_xform;
+		break;
+	case CRYPTO_CHAIN_ONLY_AUTH:
+		auth_xform = first_xform;
+		break;
+	case CRYPTO_CHAIN_CIPHER_AUTH:
+		cipher_xform = first_xform;
+		auth_xform = first_xform->next;
+		break;
+	case CRYPTO_CHAIN_AUTH_CIPHER:
+		auth_xform = first_xform;
+		cipher_xform = first_xform->next;
+		break;
+	default:
+		return -1;
+	}
 
 	for (id = 0; id < global->enabled_crypto_devs; id++) {
 		struct rte_cryptodev_info dev_info;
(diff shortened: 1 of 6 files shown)


platform/linux-dpdk/odp_crypto.c
```diff
@@ -116,13 +124,14 @@ static int auth_alg_odp_to_rte(odp_auth_alg_t auth_alg,
...
 		auth_xform->auth.algo = RTE_CRYPTO_AUTH_NULL;
+		auth_xform->auth.digest_length = 0;
 		break;
 	case ODP_AUTH_ALG_MD5_HMAC:
 #if ODP_DEPRECATED_API
 	case ODP_AUTH_ALG_MD5_96:
 #endif
 		auth_xform->auth.algo = RTE_CRYPTO_AUTH_MD5_HMAC;
-		auth_xform->auth.digest_length = 12;
+		auth_xform->auth.digest_length = 16;
```
lumag: This is wrong. MD5_96 should use 12-byte digest length. MD5_HMAC should use digest length specified by application.

platform/linux-dpdk/odp_crypto.c
```diff
@@ -868,35 +926,36 @@ int odp_crypto_session_create(odp_crypto_session_param_t *param,
...
+			first_xform->next = NULL;
+			res = CRYPTO_CHAIN_ONLY_AUTH;
+		} else if (auth_xform.auth.algo == RTE_CRYPTO_AUTH_NULL) {
+			first_xform = &cipher_xform;
+			first_xform->next = NULL;
+			res = CRYPTO_CHAIN_ONLY_CIPHER;
+		} else {
+			first_xform->next = &cipher_xform;
+			res = CRYPTO_CHAIN_AUTH_CIPHER;
+		}
```
lumag: I'd suggest to use ALG_NULL checks directly, rather than introduce chain order.

test/performance/odp_crypto.c
```diff
@@ -32,17 +32,26 @@
 
 static uint8_t test_iv[8] = "01234567";
 
-static uint8_t test_key16[16] = { 0x01, 0x02, 0x03, 0x04, 0x05,
+static uint8_t test_key24[24] = { 0x01, 0x02, 0x03, 0x04, 0x05,
```
lumag: `non tangere testos meos`

test/validation/api/crypto/odp_crypto_test_inp.c
```diff
@@ -416,7 +416,7 @@ static void alg_test(odp_crypto_op_t op,
 	/* Create a crypto session */
 	odp_crypto_session_param_init(&ses_params);
 	ses_params.op = op;
-	ses_params.auth_cipher_text = false;
+	ses_params.auth_cipher_text = true;
```
lumag: why?

platform/linux-dpdk/odp_crypto.c
GBalakrishna: openssl PMD in 17.02 supports the digest length to be min 16 bytes. But I have now removed the setting digest_lengths from here as it sets it from the application.

test/validation/api/crypto/odp_crypto_test_inp.c
GBalakrishna: my understanding of auth_cipher_text = true means do encode then authenticate. and when it is applied together with if (ODP_CRYPTO_OP_ENCODE == param->op) sets the "entry->do_cipher_first"  to true by default and if the param->op == DECODE it sets the !auth_cipher_text. I felt this is more easier for understanding.

platform/linux-dpdk/odp_crypto.c
GBalakrishna: Not sure if I have fallowed. I introduced Chain order because dpdk PMD's doesn't have support for combination of NULL + Valid algo. So linux-dpdk implementation converts the NULL also to chain order as if it cipher only & auth only.

test/performance/odp_crypto.c
GBalakrishna: not sure what do u mean here.

test/validation/api/crypto/odp_crypto_test_inp.c
lumag: Theoretically we'd better be testing both combinations. The problem with test changes in your commit is that you change tests rather than expand them. Can you import tests from #379?

test/performance/odp_crypto.c
lumag: "Do not change tests". See my previous comment.

platform/linux-dpdk/odp_crypto.c
lumag: For deprecated MD5_96 (and SHA256_128/AES128_GCM) application won't set `auth.digest_len`, so you have to enforce 12 and 16 bytes digest len. For MD5_HMAC/SHA256_HMAC (and others) application will set `auth.digest_len`.

IIRC QAT driver should support truncated digests out-of-box. Also you can try expanding pmd_openssl to support truncated lengths, it should not be very complicated. Also note, that full-length HMAC won't be useful for IPsec.

platform/linux-dpdk/odp_crypto.c
lumag: I suggest to factor out cipher and auth checks as separate functions and apply them only if corresponding algorithm is not NULL. Code would be simpler.

platform/linux-dpdk/test/wrapper-script.sh
```diff
@@ -1,6 +1,6 @@
 #!/bin/bash
 
-export ODP_PLATFORM_PARAMS=${ODP_PLATFORM_PARAMS:--n 4 --vdev "crypto_openssl"}
+export ODP_PLATFORM_PARAMS=${ODP_PLATFORM_PARAMS:--n 4 --vdev "crypto_openssl" --vdev "crypto_null"}
```
Bill-Fischofer-Linaro: Checkpatch flags this line:
```
#363: FILE: platform/linux-dpdk/test/wrapper-script.sh:3:
+export ODP_PLATFORM_PARAMS=${ODP_PLATFORM_PARAMS:--n 4 --vdev "crypto_openssl" --vdev "crypto_null"}
total: 0 errors, 1 warnings, 0 checks, 530 lines checked
NOTE: Ignored message types: AVOID_EXTERNS BIT_MACRO COMPARISON_TO_NULL DEPRECATED_VARIABLE NEW_TYPEDEFS PREFER_PRINTF PREFER_SCANF SPLIT_STRING SSCANF_TO_KSTRTO VOLATILE
0001-linux-dpdk-crypto-support-for-cipher-auth-only-featu.patch has style problems, please review.
```

test/validation/api/crypto/odp_crypto_test_inp.c
GBalakrishna: what do you mean by testing both combinations ?. We are testing both the combinations ENCODE & DECODE. It's the default value we are changing it.

test/performance/odp_crypto.c
GBalakrishna: I have updated the tests so that it works with both the platforms linux-generci & linux-dpdk.

platform/linux-dpdk/odp_crypto.c
GBalakrishna: setting digest_length is moved to get_crypto_dev(), where it reads from PMD and set the min digest length if application doesn't set it. Reagrding the QAT, I need to look at it and understand. may be as a separate PR.

lumag: Implementation should not set digest len for cases other than several deprecated auth algos.

test/performance/odp_crypto.c
lumag: Could you please add tests cases rather than modify existing ones?

test/validation/api/crypto/odp_crypto_test_inp.c
lumag: both = Encrypt-then-MAC and MAC-cleartext

GBalakrishna: oke I see. I will make the update in the next version.

lumag: It's really not a problem at this moment, this can be fixed later.

Bill-Fischofer-Linaro:
@lumag Can you please add your review to this PR?

platform/linux-dpdk/odp_crypto.c
```diff
@@ -122,29 +130,24 @@ static int auth_alg_odp_to_rte(odp_auth_alg_t auth_alg,
 	case ODP_AUTH_ALG_MD5_96:
 #endif
 		auth_xform->auth.algo = RTE_CRYPTO_AUTH_MD5_HMAC;
-		auth_xform->auth.digest_length = 12;
```
lumag: `auth.digest_length` should be set for MD5_96 here, but not for MD5_HMAC.

platform/linux-dpdk/odp_crypto.c
```diff
@@ -122,29 +130,24 @@ static int auth_alg_odp_to_rte(odp_auth_alg_t auth_alg,
...
 		break;
 	case ODP_AUTH_ALG_SHA256_HMAC:
 #if ODP_DEPRECATED_API
 	case ODP_AUTH_ALG_SHA256_128:
 #endif
 		auth_xform->auth.algo = RTE_CRYPTO_AUTH_SHA256_HMAC;
-		auth_xform->auth.digest_length = 16;
```
lumag: Same issue here.

platform/linux-dpdk/odp_crypto.c
```diff
@@ -122,29 +130,24 @@ static int auth_alg_odp_to_rte(odp_auth_alg_t auth_alg,
...
 	case ODP_AUTH_ALG_SHA512_HMAC:
 		auth_xform->auth.algo = RTE_CRYPTO_AUTH_SHA512_HMAC;
-		auth_xform->auth.digest_length = 64;
 		break;
 	case ODP_AUTH_ALG_AES_GCM:
 #if ODP_DEPRECATED_API
 	case ODP_AUTH_ALG_AES128_GCM:
 #endif
 		auth_xform->auth.algo = RTE_CRYPTO_AUTH_AES_GCM;
-		auth_xform->auth.digest_length = 16;
```
lumag: same issue here.

platform/linux-dpdk/odp_crypto.c
```diff
@@ -615,7 +618,7 @@ int odp_crypto_auth_capability(odp_auth_alg_t auth,
 		while (cap->op != RTE_CRYPTO_OP_TYPE_UNDEFINED) {
 			cap_auth_algo = cap->sym.auth.algo;
 			if (cap->sym.xform_type ==
-			    RTE_CRYPTO_SYM_XFORM_CIPHER) {
+			    RTE_CRYPTO_SYM_XFORM_AUTH) {
 				if (cap_auth_algo == auth_xform.auth.algo)
```
lumag: Liked this typo. It shows that the code never worked as expected.

platform/linux-dpdk/odp_crypto.c
```diff
@@ -663,76 +666,105 @@ int odp_crypto_auth_capability(odp_auth_alg_t auth,
...
+					ODP_ERR("Invalid cipher key length\n");
+					return -1;
+				}
+			/* No size provided, use minimum size. */
+			} else
+				cipher_xform->cipher.key.length =
+						cap->sym.cipher.key_size.min;
+
+			/* Check if iv length is supported by the algorithm. */
+			if (iv_length) {
```
lumag: No need to keep this under `if (iv_length)`

platform/linux-dpdk/odp_crypto.c
```diff
@@ -776,6 +808,22 @@ static int get_crypto_dev(struct rte_crypto_sym_xform *cipher_xform,
 	return -1;
 }
 
+static void set_chain_order(struct rte_crypto_sym_xform **first_xform,
+			    struct rte_crypto_sym_xform *auth_xform,
+			    struct rte_crypto_sym_xform *cipher_xform,
+			    enum crypto_chain_order *res)
```
lumag: Make it return `crypto_chain_order`.

platform/linux-dpdk/odp_crypto.c
```diff
@@ -855,48 +924,51 @@ int odp_crypto_session_create(odp_crypto_session_param_t *param,
...
-	/* Check result */
-	if (rc) {
-		*status = ODP_CRYPTO_SES_CREATE_ERR_INV_AUTH;
-		/* remove the crypto_session_entry_t */
-		memset(entry, 0, sizeof(*entry));
-		free_session(entry);
-		return -1;
+	/* Process based on cipher */
+	/* Derive order */
+	if (entry->do_cipher_first) {
```
lumag: Push this logic to `set_chain_order`

platform/linux-dpdk/odp_crypto.c
```diff
@@ -784,8 +832,9 @@ int odp_crypto_session_create(odp_crypto_session_param_t *param,
 	uint8_t cdev_id = 0;
 	struct rte_crypto_sym_xform cipher_xform;
 	struct rte_crypto_sym_xform auth_xform;
-	struct rte_crypto_sym_xform *first_xform;
+	struct rte_crypto_sym_xform *first_xform = NULL;
 	struct rte_cryptodev_sym_session *session;
+	enum crypto_chain_order res = CRYPTO_CHAIN_NOT_SUPPORTED;
```
lumag: `res` does not quite follow `crypto_chain_order`, does it? Please rename to 'order' or smth. like that.

platform/linux-dpdk/test/wrapper-script.sh
```diff
@@ -1,6 +1,7 @@
 #!/bin/bash
 
-export ODP_PLATFORM_PARAMS=${ODP_PLATFORM_PARAMS:--n 4 --vdev "crypto_openssl"}
+export ODP_PLATFORM_PARAMS=${ODP_PLATFORM_PARAMS:--n 4 "--vdev \
+"crypto_openssl" --vdev "crypto_null""}
```
lumag: Will it work w/o crypto_null driver?

platform/linux-generic/odp_ipsec_sad.c
```diff
@@ -303,13 +303,15 @@ odp_ipsec_sa_t odp_ipsec_sa_create(const odp_ipsec_sa_param_t *param)
...
 		ipsec_sa->esp_block_len = 8;
+		crypto_param.iv.length = 8;
 		break;
 #if ODP_DEPRECATED_API
 	case ODP_CIPHER_ALG_AES128_CBC:
 #endif
 	case ODP_CIPHER_ALG_AES_CBC:
 		ipsec_sa->esp_iv_len = 16;
 		ipsec_sa->esp_block_len = 16;
+		crypto_param.iv.length = 16;
```
lumag: and this

test/performance/odp_crypto.c
```diff
@@ -226,6 +241,18 @@ static crypto_alg_config_t algs_config[] = {
...
+	{
+		.name = "null-hmac-md5-96",
+		.session = {
+			.cipher_alg = ODP_CIPHER_ALG_NULL,
+			.auth_alg = ODP_AUTH_ALG_MD5_HMAC,
+			.auth_key = {
+				.data = test_key64,
+				.length = sizeof(test_key64)
+			},
+			.auth_digest_len = 16,
```
lumag: Name of the test suggests, that test is using truncated hmac, but then you specify full auth_digest_len.

test/performance/odp_crypto.c
```diff
@@ -45,6 +45,21 @@ static uint8_t test_key24[24] = { 0x01, 0x02, 0x03, 0x04, 0x05,
...
+				  0x15, 0x16, 0x17, 0x18, 0x19,
+				  0x1a, 0x1b, 0x1c, 0x1d, 0x1e,
+				  0x1f, 0x20, 0x21, 0x22, 0x23,
+				  0x24, 0x25, 0x26, 0x27, 0x28,
+				  0x29, 0x2a, 0x2b, 0x2c, 0x2d,
+				  0x2e, 0x2f, 0x30, 0x31, 0x32,
+				  0x33, 0x34, 0x35, 0x36, 0x37,
+				  0x38, 0x39, 0x3a, 0x3b, 0x3c,
+				  0x3d, 0x3e, 0x3f, 0x40
+};
```
lumag: All test updates should go in separate commit.

platform/linux-dpdk/odp_crypto.c
GBalakrishna: it is set in get_crypto_dev(): auth_xform->auth.digest_length = cap->sym.auth.digest_size.min

platform/linux-dpdk/odp_crypto.c
GBalakrishna: it is set in get_crypto_dev()

platform/linux-dpdk/odp_crypto.c
GBalakrishna: it is set in get_crypto_dev()

platform/linux-dpdk/test/wrapper-script.sh
GBalakrishna: yes it works.

platform/linux-dpdk/odp_crypto.c
lumag: OK. You always use `param->auth_digest_len` if it is set. This logic is incorrect. It should be used for all cases except MD5_96, SHA256_128, AES128_GCM (where an override should be used).

platform/linux-dpdk/odp_crypto.c
GBalakrishna: you mean is_valid_size() ?

platform/linux-generic/odp_ipsec_sad.c
GBalakrishna: you mean to make this in separate commit ?

platform/linux-dpdk/odp_crypto.c
GBalakrishna: you mean to say for MD5_96, we should ignore the digest_length from application even if it is set ?

platform/linux-dpdk/odp_crypto.c
lumag: Yes. 0 is also a valid iv length.

platform/linux-dpdk/odp_crypto.c
lumag: Yes.

platform/linux-dpdk/test/wrapper-script.sh
lumag: Can you please paste the test log w/o crypto_null somewhere (pastebin or e-mail)? Thanks!

GBalakrishna: all the tests in validations will be skipped because the validation tests works with combination of null algo which is not supported by openssl PMD. If you have a test case that works with valid algos (i.e without null_algo), it should work without an issue.

platform/linux-dpdk/odp_crypto.c
GBalakrishna: in the last review version, you mention that application won't set diget_length for MD5_96 etc....  And also I see that from linux-generic odp doesn't not supporting MD5_96 & SHA256_128 and we shouldn't care about them here I suppose. In face those should be removed in the next version.

platform/linux-dpdk/test/wrapper-script.sh
lumag: As in fact you are not using crypto_null (except for the lame case of CIPHER_NULL/AUTH_NULL) I suggest to return static capabilities for NULL algos.

platform/linux-dpdk/odp_crypto.c
lumag: linux-generic supports them.

GBalakrishna: Ys I see that. Will update to support those algos as well.

platform/linux-dpdk/test/wrapper-script.sh
GBalakrishna: why static?.  we have the support from dpdk and you just have to initialize it like other PMD's and it is consistent with other PMD's when reading capabilities etc..

platform/linux-generic/odp_ipsec_sad.c
lumag: yes,

platform/linux-dpdk/test/wrapper-script.sh
lumag: Because you don't use crypto_null. In fact all code will work perfectly w/o crypto_null, if it did not miss NULL capabilities

GBalakrishna: but I suppose it doesn't hurt the current way of initializing null algos. To change this to static needs some rework to be done and I think it can be updated later by anyone.

lumag:
OK

heyi-linaro:
@lumag @Bill-Fischofer-Linaro is this PR looks good to you?

heyi-linaro:
Merged and closed.

PR #388, opened 2018-01-12
[PATCH API-NEXT v2] User pointer init to NULL

psavol:
Require that implementation initializes the pointer value to NULL. Application can use the pointer more robustly when it can tell if the pointer has been set previously or not.


platform/linux-generic/include/odp/api/plat/packet_types.h
```diff
@@ -67,101 +67,130 @@ typedef enum {
...
+	};
+
+	/* Flag groups */
+	struct {
+		uint32_t reserved2:     12;
+		uint32_t other:         12; /* All other flags */
+		uint32_t error:          8; /* All error flags */
+	} all;
+
+} _odp_packet_flags_t;
```
lumag: Hmm. What about the following code:
```c
typedef struct {
  union {
   struct {
    // ...
   };
   uint32_t all_error : 8;
  } error;
  union {
    sruct {
    // ...
    };
    uint32_t all_other;
  } other;
} _odp_packet_flags_t;
```

I have not tried, but I think this should work and still provide cleaner separation between error non-error flags.

platform/linux-generic/include/odp/api/plat/packet_types.h
```diff
@@ -67,101 +67,130 @@ typedef enum {
...
+		uint64_t tcpopt:1;    /* TCP options present */
+		uint64_t sctp:1;      /* SCTP */
+		uint64_t icmp:1;      /* ICMP */
+
+		uint64_t color:2;     /* Packet color for traffic mgmt */
+		uint64_t nodrop:1;    /* Drop eligibility status */
+
+		uint64_t l3_chksum_done:1; /* L3 checksum validation done */
+		uint64_t l4_chksum_done:1; /* L4 checksum validation done */
+		uint64_t ipsec_udp:1; /* UDP-encapsulated IPsec packet */
```
lumag: Could you please split Doxygen-related and flags shuffling parts into separate commits?

platform/linux-generic/include/odp/api/plat/packet_types.h
```diff
@@ -67,101 +67,130 @@ typedef enum {
...
+		uint32_t ipsec_err:      1; /* IPsec error */
+		uint32_t crypto_err:     1; /* Crypto packet operation error */
+	};
+
+	/* Flag groups */
+	struct {
+		uint32_t reserved2:     12;
+		uint32_t other:         12; /* All other flags */
+		uint32_t error:          8; /* All error flags */
+	} all;
```
Bill-Fischofer-Linaro: This should logically go after `all_flags`. You're progressing from coarse to fine-grain field definition. Everything, flag class, then individual flag definitions.

platform/linux-generic/include/odp/api/plat/packet_types.h
Bill-Fischofer-Linaro: To much nesting of structs and unions tends to make things harder to follow. Two deep is clearer than a three deep definition. I think I prefer the original.

platform/linux-generic/include/odp/api/plat/packet_types.h
```diff
@@ -156,7 +157,12 @@ typedef union {
 	uint32_t all_flags;
 
 	struct {
-		uint32_t reserved1:     12;
+		uint32_t reserved1:     11;
+
+	/*
+	 * Init flags
+	 */
+		uint32_t user_ptr_set:   1; /* Ptr has been set vs. NULL */
```
Bill-Fischofer-Linaro: The "vs" comment is confusing. "User ptr has been set" seems sufficient here.

platform/linux-generic/include/odp/api/plat/packet_types.h
psavol: Dmitry, those two unions (error and other) would result two 32 bit words, which would defeat the purpose of combining flags into one word. If the top level would be union, it would be single word but now each group would need to be padded:
```
union foo {
	union {
		struct {
			uint32_t bit1:  1;
			uint32_t bit2:  1;
			uint32_t res1: 30;
		};

		struct {
			uint32_t all:   2;
			uint32_t res2: 30;
		};

	} bar;

	union {
		struct {
			uint32_t res1:  2;
			uint32_t bit1:  1;
			uint32_t bit2:  1;
			uint32_t res2: 28;
		};

		struct {
			uint32_t res3:  2;
			uint32_t all:   2;
			uint32_t res4: 28;
		};

	} bar2;
};

```

Since this is implementation internal, we can relay on developer remembering e.g. that xxx_err is an error flag, use a clean/flat struct. Also flag naming can be updated any time (since it's not API).

platform/linux-generic/include/odp/api/plat/packet_types.h
psavol: OK.

platform/linux-generic/include/odp/api/plat/packet_types.h
psavol: OK.

platform/linux-generic/include/odp/api/plat/packet_types.h
lumag: @psavol Sorry, I though that C compiler would be able to merge those two unions into consecutive bit space. I experimented with it a bit and it seems you are correct. I can not find an easy way to force gcc into using following bits. Only using `uint8_t` helps.

Bill-Fischofer-Linaro:
@psavol Please rebase to resolve conflicts from recent forced update of api-next.

Bill-Fischofer-Linaro:
Superseded by PR #392

PR #389, opened 2018-01-13
[PATCH v19] Validate checksums during packet parsing

lumag:
software checksum validation/insertion.

TODO:
 - SCTP (also is not supported for hardware-based DPDK PktIO).

### DIFF
--- a/platform/linux-generic/include/odp_packet_internal.h
+++ b/platform/linux-generic/include/odp_packet_internal.h
@@ -252,7 +252,8 @@ int packet_alloc_multi(odp_pool_t pool_hdl, uint32_t len,
 
 /* Perform packet parse up to a given protocol layer */
 int packet_parse_layer(odp_packet_hdr_t *pkt_hdr,
-		       odp_proto_layer_t layer);
+		       odp_proto_layer_t layer,
+		       odp_proto_chksums_t chksums);
 
 /* Reset parser metadata for a new parse */
 void packet_parse_reset(odp_packet_hdr_t *pkt_hdr);
@@ -293,7 +294,8 @@ static inline void packet_set_ts(odp_packet_hdr_t *pkt_hdr, odp_time_t *ts)
 }
 
 int packet_parse_common(packet_parser_t *pkt_hdr, const uint8_t *ptr,
-			uint32_t pkt_len, uint32_t seg_len, int layer);
+			uint32_t pkt_len, uint32_t seg_len, int layer,
+			odp_proto_chksums_t chksums);
 
 int _odp_cls_parse(odp_packet_hdr_t *pkt_hdr, const uint8_t *parseptr);
 
--- a/platform/linux-generic/odp_packet.c
+++ b/platform/linux-generic/odp_packet.c
@@ -2084,7 +2084,8 @@ error:
  * Parser helper function for IPv4
  */
 static inline uint8_t parse_ipv4(packet_parser_t *prs, const uint8_t **parseptr,
-				 uint32_t *offset, uint32_t frame_len)
+				 uint32_t *offset, uint32_t frame_len,
+				 odp_proto_chksums_t chksums)
 {
 	const _odp_ipv4hdr_t *ipv4 = (const _odp_ipv4hdr_t *)*parseptr;
 	uint32_t dstaddr = _odp_be_to_cpu_32(ipv4->dst_addr);
@@ -2098,6 +2099,15 @@ static inline uint8_t parse_ipv4(packet_parser_t *prs, const uint8_t **parseptr,
 			 (l3_len > frame_len - *offset))) {
 		prs->flags.ip_err = 1;
 		return 0;
+	}
+
+	if (chksums.chksum.ipv4) {
+		prs->input_flags.l3_chksum_done = 1;
+		if (odp_chksum_ones_comp16(ipv4, ihl * 4) != 0xffff) {
+			prs->flags.ip_err = 1;
+			prs->flags.l3_chksum_err = 1;
+			return 0;
+		}
 	}
 
 	*offset   += ihl * 4;
@@ -2230,7 +2240,8 @@ static inline
 int packet_parse_common_l3_l4(packet_parser_t *prs, const uint8_t *parseptr,
 			      uint32_t offset,
 			      uint32_t frame_len, uint32_t seg_len,
-			      int layer, uint16_t ethtype)
+			      int layer, uint16_t ethtype,
+			      odp_proto_chksums_t chksums)
 {
 	uint8_t  ip_proto;
 
@@ -2246,7 +2257,8 @@ int packet_parse_common_l3_l4(packet_parser_t *prs, const uint8_t *parseptr,
 	switch (ethtype) {
 	case _ODP_ETHTYPE_IPV4:
 		prs->input_flags.ipv4 = 1;
-		ip_proto = parse_ipv4(prs, &parseptr, &offset, frame_len);
+		ip_proto = parse_ipv4(prs, &parseptr, &offset, frame_len,
+				      chksums);
 		prs->l4_offset = offset;
 		break;
 
@@ -2334,7 +2346,7 @@ int packet_parse_common_l3_l4(packet_parser_t *prs, const uint8_t *parseptr,
  */
 int packet_parse_common(packet_parser_t *prs, const uint8_t *ptr,
 			uint32_t frame_len, uint32_t seg_len,
-			int layer)
+			int layer, odp_proto_chksums_t chksums)
 {
 	uint32_t offset;
 	uint16_t ethtype;
@@ -2352,7 +2364,7 @@ int packet_parse_common(packet_parser_t *prs, const uint8_t *ptr,
 	ethtype = parse_eth(prs, &parseptr, &offset, frame_len);
 
 	return packet_parse_common_l3_l4(prs, parseptr, offset, frame_len,
-					 seg_len, layer, ethtype);
+					 seg_len, layer, ethtype, chksums);
 }
 
 static inline int packet_ipv4_chksum(odp_packet_t pkt,
@@ -2512,13 +2524,14 @@ int _odp_packet_udp_chksum_insert(odp_packet_t pkt)
  * Simple packet parser
  */
 int packet_parse_layer(odp_packet_hdr_t *pkt_hdr,
-		       odp_proto_layer_t layer)
+		       odp_proto_layer_t layer,
+		       odp_proto_chksums_t chksums)
 {
 	uint32_t seg_len = packet_first_seg_len(pkt_hdr);
 	void *base = packet_data(pkt_hdr);
 
 	return packet_parse_common(&pkt_hdr->p, base, pkt_hdr->frame_len,
-				   seg_len, layer);
+				   seg_len, layer, chksums);
 }
 
 int odp_packet_parse(odp_packet_t pkt, uint32_t offset,
@@ -2545,7 +2558,8 @@ int odp_packet_parse(odp_packet_t pkt, uint32_t offset,
 
 	if (proto == ODP_PROTO_ETH) {
 		ret = packet_parse_common(&pkt_hdr->p, data, packet_len,
-					  seg_len, layer);
+					  seg_len, layer,
+					  param->chksums);
 
 		if (ret)
 			return -1;
@@ -2557,7 +2571,9 @@ int odp_packet_parse(odp_packet_t pkt, uint32_t offset,
 
 		ret = packet_parse_common_l3_l4(&pkt_hdr->p, data, offset,
 						packet_len, seg_len,
-						layer, ethtype);
+						layer, ethtype,
+						param->chksums);
+
 
 		if (ret)
 			return -1;
--- a/platform/linux-generic/odp_packet.c
+++ b/platform/linux-generic/odp_packet.c
@@ -1999,6 +1999,22 @@ static uint32_t packet_sum16_32(odp_packet_hdr_t *pkt_hdr,
 	return sum;
 }
 
+static uint16_t packet_sum_ones_comp16(odp_packet_hdr_t *pkt_hdr,
+				       uint32_t offset,
+				       uint32_t len,
+				       uint32_t l4_part_sum)
+{
+	uint32_t sum = l4_part_sum;
+
+	sum += packet_sum16_32(pkt_hdr, offset, len);
+
+	/* Not more than two additions */
+	sum = (sum & 0xffff) + (sum >> 16);
+	sum = (sum & 0xffff) + (sum >> 16);
+
+	return sum;
+}
+
 /** Parser helper function for Ethernet packets */
 static inline uint16_t parse_eth(packet_parser_t *prs, const uint8_t **parseptr,
 				 uint32_t *offset, uint32_t frame_len)
@@ -2085,7 +2101,8 @@ error:
  */
 static inline uint8_t parse_ipv4(packet_parser_t *prs, const uint8_t **parseptr,
 				 uint32_t *offset, uint32_t frame_len,
-				 odp_proto_chksums_t chksums)
+				 odp_proto_chksums_t chksums,
+				 uint32_t *l4_part_sum)
 {
 	const _odp_ipv4hdr_t *ipv4 = (const _odp_ipv4hdr_t *)*parseptr;
 	uint32_t dstaddr = _odp_be_to_cpu_32(ipv4->dst_addr);
@@ -2113,6 +2130,10 @@ static inline uint8_t parse_ipv4(packet_parser_t *prs, const uint8_t **parseptr,
 	*offset   += ihl * 4;
 	*parseptr += ihl * 4;
 
+	if (chksums.chksum.udp || chksums.chksum.tcp)
+		*l4_part_sum = segment_sum16_32((const uint8_t *)&ipv4->src_addr,
+						2 * _ODP_IPV4ADDR_LEN, 0);
+
 	if (odp_unlikely(ihl > _ODP_IPV4HDR_IHL_MIN))
 		prs->input_flags.ipopt = 1;
 
@@ -2139,7 +2160,9 @@ static inline uint8_t parse_ipv4(packet_parser_t *prs, const uint8_t **parseptr,
  */
 static inline uint8_t parse_ipv6(packet_parser_t *prs, const uint8_t **parseptr,
 				 uint32_t *offset, uint32_t frame_len,
-				 uint32_t seg_len)
+				 uint32_t seg_len,
+				 odp_proto_chksums_t chksums,
+				 uint32_t *l4_part_sum)
 {
 	const _odp_ipv6hdr_t *ipv6 = (const _odp_ipv6hdr_t *)*parseptr;
 	const _odp_ipv6hdr_ext_t *ipv6ext;
@@ -2161,6 +2184,10 @@ static inline uint8_t parse_ipv6(packet_parser_t *prs, const uint8_t **parseptr,
 	/* Skip past IPv6 header */
 	*offset   += sizeof(_odp_ipv6hdr_t);
 	*parseptr += sizeof(_odp_ipv6hdr_t);
+
+	if (chksums.chksum.udp || chksums.chksum.tcp)
+		*l4_part_sum = segment_sum16_32((const uint8_t *)&ipv6->src_addr,
+						2 * _ODP_IPV6ADDR_LEN, 0);
 
 	/* Skip past any IPv6 extension headers */
 	if (ipv6->next_hdr == _ODP_IPPROTO_HOPOPTS ||
@@ -2200,7 +2227,10 @@ static inline uint8_t parse_ipv6(packet_parser_t *prs, const uint8_t **parseptr,
 /**
  * Parser helper function for TCP
  */
-static inline void parse_tcp(packet_parser_t *prs, const uint8_t **parseptr)
+static inline void parse_tcp(packet_parser_t *prs, const uint8_t **parseptr,
+			     uint16_t tcp_len,
+			     odp_proto_chksums_t chksums,
+			     uint32_t *l4_part_sum)
 {
 	const _odp_tcphdr_t *tcp = (const _odp_tcphdr_t *)*parseptr;
 	uint32_t len = tcp->hl * 4;
@@ -2208,20 +2238,51 @@ static inline void parse_tcp(packet_parser_t *prs, const uint8_t **parseptr)
 	if (odp_unlikely(tcp->hl < sizeof(_odp_tcphdr_t) / sizeof(uint32_t)))
 		prs->flags.tcp_err = 1;
 
+	if (chksums.chksum.tcp &&
+	    !prs->input_flags.ipfrag) {
+		*l4_part_sum += odp_cpu_to_be_16(tcp_len);
+#if ODP_BYTE_ORDER == ODP_BIG_ENDIAN
+		*l4_part_sum += _ODP_IPPROTO_TCP;
+#else
+		*l4_part_sum += _ODP_IPPROTO_TCP << 8;
+#endif
+	}
+
 	*parseptr += len;
 }
 
 /**
  * Parser helper function for UDP
  */
-static inline void parse_udp(packet_parser_t *prs, const uint8_t **parseptr)
+static inline void parse_udp(packet_parser_t *prs, const uint8_t **parseptr,
+			     odp_proto_chksums_t chksums,
+			     uint32_t *l4_part_sum)
 {
 	const _odp_udphdr_t *udp = (const _odp_udphdr_t *)*parseptr;
 	uint32_t udplen = _odp_be_to_cpu_16(udp->length);
 	uint16_t ipsec_port = _odp_cpu_to_be_16(_ODP_UDP_IPSEC_PORT);
 
-	if (odp_unlikely(udplen < sizeof(_odp_udphdr_t)))
+	if (odp_unlikely(udplen < sizeof(_odp_udphdr_t))) {
 		prs->flags.udp_err = 1;
+		return;
+	}
+
+	if (chksums.chksum.udp &&
+	    !prs->input_flags.ipfrag) {
+		if (udp->chksum == 0) {
+			prs->input_flags.l4_chksum_done = 1;
+			prs->flags.l4_chksum_err =
+				(prs->input_flags.ipv4 != 1);
+		} else {
+			*l4_part_sum += udp->length;
+#if ODP_BYTE_ORDER == ODP_BIG_ENDIAN
+			*l4_part_sum += _ODP_IPPROTO_UDP;
+#else
+			*l4_part_sum += _ODP_IPPROTO_UDP << 8;
+#endif
+		}
+		prs->input_flags.udp_chksum_zero = (udp->chksum == 0);
+	}
 
 	if (odp_unlikely(ipsec_port == udp->dst_port && udplen > 4)) {
 		uint32_t val;
@@ -2241,7 +2302,8 @@ int packet_parse_common_l3_l4(packet_parser_t *prs, const uint8_t *parseptr,
 			      uint32_t offset,
 			      uint32_t frame_len, uint32_t seg_len,
 			      int layer, uint16_t ethtype,
-			      odp_proto_chksums_t chksums)
+			      odp_proto_chksums_t chksums,
+			      uint32_t *l4_part_sum)
 {
 	uint8_t  ip_proto;
 
@@ -2258,14 +2320,14 @@ int packet_parse_common_l3_l4(packet_parser_t *prs, const uint8_t *parseptr,
 	case _ODP_ETHTYPE_IPV4:
 		prs->input_flags.ipv4 = 1;
 		ip_proto = parse_ipv4(prs, &parseptr, &offset, frame_len,
-				      chksums);
+				      chksums, l4_part_sum);
 		prs->l4_offset = offset;
 		break;
 
 	case _ODP_ETHTYPE_IPV6:
 		prs->input_flags.ipv6 = 1;
 		ip_proto = parse_ipv6(prs, &parseptr, &offset, frame_len,
-				      seg_len);
+				      seg_len, chksums, l4_part_sum);
 		prs->l4_offset = offset;
 		break;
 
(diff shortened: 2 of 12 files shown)


platform/linux-generic/include/odp_chksum_internal.h
```diff
@@ -0,0 +1,59 @@
+/* Copyright (c) 2013, Linaro Limited
```
Bill-Fischofer-Linaro: 2018

platform/linux-generic/odp_packet.c
```diff
@@ -2191,7 +2201,7 @@ int packet_parse_common_l3_l4(packet_parser_t *prs, const uint8_t *parseptr,
 	switch (ethtype) {
 	case _ODP_ETHTYPE_IPV4:
 		prs->input_flags.ipv4 = 1;
-		ip_proto = parse_ipv4(prs, &parseptr, &offset, frame_len);
+		ip_proto = parse_ipv4(prs, &parseptr, &offset, frame_len, chksums);
```
Bill-Fischofer-Linaro: Checkpatch flag:
```
#102: FILE: platform/linux-generic/odp_packet.c:2204:
+		ip_proto = parse_ipv4(prs, &parseptr, &offset, frame_len, chksums);
total: 0 errors, 1 warnings, 0 checks, 163 lines checked
```

lumag:
@Bill-Fischofer-Linaro I'll fix issues before next iteration. It would be nice to receive feedback on functionality.

JannePeltonen:
The proposed SW implementation of checksumming for incoming packets can hurt performance a lot, depending on how ODP is used.

Where I come from, the same NE is often handling many different flows that require different processing. For example, some packets may be only forwarded (i.e. no L4 checksum checking done) and some are terminated locally (L4 checksum checking needed). To speed up the latter case, L4 checksum checking is done in HW. It does not hurt the former case since checksumming done by the HW is cheap and can just be ignored when not needed.

If the SW based checksum implementation goes in, I think it would be good to be able to disable it at ODP start time (or config flag is the default is off).

I wonder how well the proposed code is prepared to accommodate pktios that do support HW based checksum offload in RX (it would be nice to have that in the DPDK pktio).

bogdanPricope:
Do you intend to have this behavior?
-	Rx csum OFF
-	RX csum ON – do sw or hw csum depending on platform capabilities
In this case, setting should be per protocol layer.

lumag:
@JannePeltonen it is hooked into loop pktio and software IPsec implementation. Basically I had to write it, as IPsec (unlike PktIO) lacks capabilities to specify if checksumming is suported on RX or TX side.

lumag:
... And it is enabled only if configured via config params.

lumag:
@bogdanPricope basically the intention was to have it the other way around. HW pktios provide their own way of implementing checksumming. SW pktios can use these to implement necessary capabilities. I will check DPDK PktIO (which is HW interface, but uses packet parsing from ODP).

JannePeltonen:
@lumag If something is needed for IPsec it should not adversely affect non-IPsec use cases. My point is that is should be possible to use RX checksum offloading with a performance benefit when the checksum needs to be checked only for a subset of received packets. It would be better if an application would not need to decide whether to enable RX checksum offload based on whether the ODP implementation implements it efficiently.

lumag:
Well, application does not know anything about `efficiency` of checksum offloading. It is either supported or not. Maybe I should split `software-feature-extended` version of loop pktio (which will support IPsec inline, checksumming, etc) from plain old loop, which did not support anything. As I said in previous comment, I will make sure that this code does not worsen speed of interfaces that won't need software checksumming.
An alternative approach would be to add IPsec capabilities describing chksumming (like we have for PktIO) and drop this code alltogether.

JannePeltonen:
I am not sure if we understand each other. My use case is somewhat like this: System receives UDP and TCP packets that belong to many streams. Half of the packets are forwarded without further action and half of them processed locally.

Without checksum offload support the system ends up doing SW checksum check for 50 % of traffic. 

With HW based checksum offload, the system requests L4 checksum for all incoming packets but checks the result only for the packets that are processed locally. Performance increases a lot since no SW based checksumming happens any more and the HW based checksumming is virtually free.

When I port the application to ODP, should I use checksum offload or not? If I don't use it, I suffer from SW based checksumming of 50 % of the packets when HW checksumming might be available. If I do use it but the ODP implementation implements checksumming in SW, I suffer because SW based checksumming is now done for 100 % of the packets instead of 50 %. My only resort is to add a configuration flag to tell whether to really use the checksum offload that ODP advertises.

My point is that adding SW based checksum offload support in Linux generic is not good for performance and not very helpful since portable ODP applications must anyway be able to handle checksum checking by themselves.

btw, I have another use case where portion of incoming traffic is switched through at L2 and portion is processed locally, benefitting from L4 checksum offload.

lumag:
@JannePeltonen I understand your concerns
 - SW checksumming will not be enabled for any PktIO except loop (or special full-featured-loop) which is used for testing.
 - SW checksumming is a required fallback for IPsec packet processing, where we do not have capabilities that app can use to determine if checksumming is supported or not.

platform/linux-generic/odp_packet.c
```diff
@@ -1948,6 +1949,35 @@ int _odp_packet_copy_md_to_packet(odp_packet_t srcpkt, odp_packet_t dstpkt)
...
+	while (len > 0) {
+		uint32_t seglen = 0; /* GCC */
+		void *mapaddr = packet_map(pkt_hdr, offset, &seglen, NULL);
+
+		if (seglen < len)
+			seglen = len;
+
+		len -= seglen;
+		sum += _odp_chksum_ones_comp16_32(mapaddr, seglen, odd_offset);
+		odd_offset ^= (seglen % 2);
```
Bill-Fischofer-Linaro: Drop use of `odd_offset` as noted earlier.

platform/linux-generic/include/odp_chksum_internal.h
```diff
@@ -0,0 +1,59 @@
...
+ * Based on RFC1071 and its errata.
+ */
+static uint32_t _odp_chksum_ones_comp16_32(const void *p, uint32_t len,
+					   odp_bool_t odd_offset)
+
+{
+	uint32_t sum = 0;
+	const uint16_t *data = p;
+
+	if (odd_offset) {
```
Bill-Fischofer-Linaro: `if (len > 1 && (uintptr_t)p % 2) ...`

platform/linux-generic/include/odp_chksum_internal.h
```diff
@@ -0,0 +1,59 @@
...
+
+#ifdef __cplusplus
+extern "C" {
+#endif
+
+/* Simple implementation of ones complement sum.
+ * Based on RFC1071 and its errata.
+ */
+static uint32_t _odp_chksum_ones_comp16_32(const void *p, uint32_t len,
+					   odp_bool_t odd_offset)
```
Bill-Fischofer-Linaro: The `odd_offset` parameter is unnecessary here. What matters is whether `p` starts on an even or odd byte boundary since that determines whether you can fetch `data` as `uint16_t` quantities without making unaligned storage references. Each call to `_odp_chksum_ones_comp16_32()` is independent in this regard, so it must always be determined anew for each call.

platform/linux-generic/include/odp_chksum_internal.h
```diff
@@ -0,0 +1,59 @@
...
+{
+	uint32_t sum = 0;
+	const uint16_t *data = p;
+
+	if (odd_offset) {
+		uint16_t left_over = 0;
+
+		*(uint8_t *)&left_over = *(const uint8_t *)data;
+		sum = left_over;
+		data++;
```
Bill-Fischofer-Linaro: This erroneously increments `data` by 2 when you want to increment it by 1 here so that it points to `p + 1`, not `p + 2`. Correct is `data = (const uint16_t *)((uintptr_t)p + 1);`

platform/linux-generic/odp_packet.c
```diff
@@ -2052,6 +2083,18 @@ static inline uint8_t parse_ipv4(packet_parser_t *prs, const uint8_t **parseptr,
 	*offset   += ihl * 4;
 	*parseptr += ihl * 4;
 
+	if (chksums.chksum.udp || chksums.chksum.tcp) {
+		l4_part_sum = _odp_chksum_ones_comp16_32(
+				(const uint16_t *)&ipv4->src_addr,
+				2 * _ODP_IPV4ADDR_LEN, false);
```
Bill-Fischofer-Linaro: Drop last parameter (`odd_offset`)

platform/linux-generic/odp_packet.c
```diff
@@ -2099,6 +2144,18 @@ static inline uint8_t parse_ipv6(packet_parser_t *prs, const uint8_t **parseptr,
 	*offset   += sizeof(_odp_ipv6hdr_t);
 	*parseptr += sizeof(_odp_ipv6hdr_t);
 
+	if (chksums.chksum.udp || chksums.chksum.tcp) {
+		l4_part_sum = _odp_chksum_ones_comp16_32(
+				(const uint16_t *)(uintptr_t)&ipv6->src_addr,
+				2 * _ODP_IPV6ADDR_LEN, false);
```
Bill-Fischofer-Linaro: Drop last parameter (`odd_offset`)

platform/linux-generic/odp_packet.c
```diff
@@ -2052,6 +2083,18 @@ static inline uint8_t parse_ipv4(packet_parser_t *prs, const uint8_t **parseptr,
...
 
+	if (chksums.chksum.udp || chksums.chksum.tcp) {
+		l4_part_sum = _odp_chksum_ones_comp16_32(
+				(const uint16_t *)&ipv4->src_addr,
+				2 * _ODP_IPV4ADDR_LEN, false);
+#if ODP_BYTE_ORDER == ODP_BIG_ENDIAN
+		l4_part_sum += ipv4->proto;
+#else
+		l4_part_sum += ((uint16_t)ipv4->proto) << 8;
+#endif
```
Bill-Fischofer-Linaro: `ipv4->proto` is a `uint8_t` field here, so why is this needed? These are the same independent of endianness.

platform/linux-generic/odp_packet.c
```diff
@@ -2099,6 +2144,18 @@ static inline uint8_t parse_ipv6(packet_parser_t *prs, const uint8_t **parseptr,
...
 
+	if (chksums.chksum.udp || chksums.chksum.tcp) {
+		l4_part_sum = _odp_chksum_ones_comp16_32(
+				(const uint16_t *)(uintptr_t)&ipv6->src_addr,
+				2 * _ODP_IPV6ADDR_LEN, false);
+#if ODP_BYTE_ORDER == ODP_BIG_ENDIAN
+		l4_part_sum += ipv6->next_hdr;
+#else
+		l4_part_sum += ((uint16_t)ipv6->next_hdr) << 8;
+#endif
```
Bill-Fischofer-Linaro: Again `ipv6->next_hdr` is a `uint8_t` field, so is endian-neutral.

platform/linux-generic/odp_packet.c
```diff
@@ -2155,14 +2212,35 @@ static inline void parse_tcp(packet_parser_t *prs,
...
+			prs->error_flags.l4_chksum =
+				(prs->input_flags.ipv4 != 1);
+		} else {
+			prs->input_flags.l4_chksum_done = 1;
+			prs->l4_part_sum += udp->length;
+			/* Do not include checksum into partial sum */
+			prs->l4_part_sum += _odp_chksum_ones_comp16_32(
+					(const void *)udp,
+					_ODP_UDPHDR_LEN - 2,
+					false);
```
Bill-Fischofer-Linaro: Drop last parameter (`odd_offset`)

platform/linux-generic/include/odp_chksum_internal.h
lumag: It is an offset insite packet.

platform/linux-generic/odp_packet.c
lumag: Yes, but the adding word is not (proto is the lat byte in pseudo header).

Bill-Fischofer-Linaro: The protocol field in the pseudo-header is simply the 8-bit IP protocol field considered as a 16-bit value. It's also in network byte order so these two have the same bit value.

platform/linux-generic/include/odp_chksum_internal.h
lumag: sorry. odd_offset references to an offset inside the packet. If you start calculating checksum from odd offset, you have to insert additional 0.

platform/linux-generic/odp_packet.c
lumag: Yes. However depending on the endianness of the host, it should be converted to a different host u16 value. It is a pair `{0, proto}`. On BE it becomes just `proto` u16, on LE it is equal to `proto << 8`.

platform/linux-generic/include/odp_chksum_internal.h
Bill-Fischofer-Linaro: The calculation starts from the address passed in as `p`. Whether that address is odd or even doesn't need a separate parameter to determine. The various ODP packet routines that return packet addresses, with the exception of `odp_packet_align()`, make no guarantees as to the alignment of any address they return.

lumag: Calculation starts at address p, but an offset inside packet might be odd or even. If it is odd, we need to accomodate. Consider the code that uses it.

Bill-Fischofer-Linaro: What offset? The calculation is for `len` bytes starting at address `p`. Whether `p` is the address of an even or odd offset in its containing packet is irrelevant since what matters is whether `p` itself is even or odd, which is an independent question.

lumag: Calculation of checksum starts at `l4_offset`. If for some reason this segment has odd amount of bytes after `l4_offset`, then checksumming will use that byte as `{b, 0}` `uint16_t` value. However calculation of checksum for the next segment (with first bytes `{c, d, e...}`) can not start as usual (`{c ,d}`, `{e, ....}). It should start as `{0, c}`, `{d, e}`, .... This is an odd offset. An offset in checksum calculation bytestream. Feel free to propose better name.

Bill-Fischofer-Linaro: OK, I see what you're saying now. But it looks like we need to consider both. The `odd_offset` parameter is saying whether the (continued) checksum needs to insert a leading zero into the calculation, however it's still true that you have to take into account whether `p` is even or odd since that can also vary independently. You can't fetch a `uint16_t` value from an odd byte address unless your CPU supports unaligned data accesses. Even for those machines that do, this typically results in a significant performance penalty.

But now we have another problem. If at entry `p` is even and `odd_offset` is true, meaning that the previous segment contained an odd number of bytes, then assuming this buffer is `{c, d, e, ...}` and we've added in `{0, c}` to account for the `odd_offset`, the next 16-bit value we want to sum is `{d, e}`, but `d` is on an odd address and `e` is on an even address so we can't fetch them as a `uint16_t` but need to assemble them from two `uint8_t` values. So it sounds like we need to factor out the logic required to fetch the next 16 bits to sum from the main loop:
```
static inline uint16_t data16(void *p) 
{
        if ((uintptr_t)p % 2)
                return (*((uint8_t *)p) << 8) + (*((uint8_t *)p + 1));

        return *((uint16_t *)p);
}
```
This can then be added to the main logic:
```
static uint32_t _odp_chksum_ones_comp_16_32(const void *p, uint32_t len, odp_bool_t odd_offset)
{
        void  *data = p;
        uint32_t sum = 0;
    
        if (odd_offset) {
               sum = *((uint8_t *)data++);
               len--;
        }

        while (len > 1) {
                sum += data16(data);
                data += 2;
                len -= 2;
        }
        
       ...etc.
}
```

lumag: Hmm. Should we create a list of all possible unaligned access locations?

Bill-Fischofer-Linaro: One of the reasons why `odp_packet_copy_to_mem()` exists is to handle this sort of alignment problem since the caller can ensure that the destination memory address is aligned properly for intended use. But whenever you use `odp_packet_offset()` or similar routines there's no guarantee that the address you get back will have any particular alignment associated with it. Packets may start out aligned, but once head push/pull operations are issued all bets are off. 

Applications are responsible for keeping track of this sort of thing themselves, but for ODP internal use we have to be prepared for packets that have been manipulated in such a way that individual byte offsets may not have any particular alignment.

platform/linux-generic/pktio/dpdk.c
```diff
@@ -542,7 +542,8 @@ static inline int mbuf_to_pkt(pktio_entry_t *pktio_entry,
 		if (pktio_cls_enabled(pktio_entry))
 			copy_packet_cls_metadata(&parsed_hdr, pkt_hdr);
 		else if (parse_layer != ODP_PROTO_LAYER_NONE)
-			packet_parse_layer(pkt_hdr, parse_layer);
+			packet_parse_layer(pkt_hdr, parse_layer,
+					   (odp_proto_chksums_t)0U);
```
Bill-Fischofer-Linaro: Useful to have this here and elsewhere as a `#define`?

lumag:
@Bill-Fischofer-Linaro @muvarov @bogdanPricope 
I think this PR is feature-complete by now. It is implemented on par with hardware-based DPDK PktIO: it does not support SCTP checksumming (which requires CRC calculation).

Bill-Fischofer-Linaro:
@lumag Thanks. Regarding SCTP, @psavol and @JannePeltonen may wish to comment on that and how urgently this is needed.

test/validation/api/ipsec/test_vectors.h
```diff
@@ -269,7 +269,7 @@ static const ODP_UNUSED ipsec_test_packet pkt_ipv4_icmp_0_ah_sha256_1_bad1 = {
 
 		/* IP */
 		0x45, 0x00, 0x00, 0x9a, 0x00, 0x00, 0x00, 0x00,
-		0x40, 0x33, 0xab, 0xd9, 0xc0, 0xa8, 0x6f, 0x02,
+		0x40, 0x33, 0xab, 0xdb, 0xc0, 0xa8, 0x6f, 0x02,
```
muvarov: @lumag how did you verify numbers? Did you run wireshark for that packet?

platform/linux-generic/odp_packet.c
```diff
@@ -1910,6 +1911,47 @@ int _odp_packet_copy_md_to_packet(odp_packet_t srcpkt, odp_packet_t dstpkt)
...
+					  uint32_t len,
+					  odp_bool_t odd_offset)
+{
+	uint32_t sum = 0;
+
+	if (offset + len > pkt_hdr->frame_len)
+		return 0;
+
+	while (len > 0) {
+		uint32_t seglen = 0; /* GCC */
```
muvarov: this is not because of gcc. Gcc warns because packet_map() can return seglen unmodified and code will make no sense here.

platform/linux-generic/odp_packet.c
```diff
@@ -1910,6 +1911,47 @@ int _odp_packet_copy_md_to_packet(odp_packet_t srcpkt, odp_packet_t dstpkt)
...
+		void *mapaddr = packet_map(pkt_hdr, offset, &seglen, NULL);
```
muvarov: check return code for packet_map() is missing here

Bill-Fischofer-Linaro: @muvarov This is an internal routine as is `packet_map()`. The only way `packet_map()` can fail is if `offset` is out of range, which we know isn't the case here. So I don't think the RC check is needed.

platform/linux-generic/odp_packet.c
Bill-Fischofer-Linaro: Not sure if this is still the case. The original code had issues with GCC only (not clang). `seglen` is always set as long as the returned addr isn't `NULL` and `seglen` itself isn't `NULL` indicating that the return is not needed.

platform/linux-generic/include/odp_chksum_internal.h
Bill-Fischofer-Linaro: It appears that the comments here still apply and need to be addressed.

platform/linux-generic/include/odp_chksum_internal.h
Bill-Fischofer-Linaro: This still needs fixing.

platform/linux-generic/odp_packet.c
```diff
@@ -2263,20 +2365,227 @@ int packet_parse_common(packet_parser_t *prs, const uint8_t *ptr,
...
+		return -1;
+	if (pkt_hdr->p.l4_offset == ODP_PACKET_OFFSET_INVALID)
+		return -1;
+
+	odp_packet_copy_to_mem(pkt, pkt_hdr->p.l3_offset, 2, &l3_ver);
+
+	if (_ODP_IPV4HDR_VER(l3_ver) == _ODP_IPV4)
+		printf("IPV4\n");
+	else
+		printf("IPV6\n");
```
Bill-Fischofer-Linaro: Residual debug code left here? We shouldn't have `printf()` calls in ODP code.

platform/linux-generic/odp_packet.c
muvarov: ok

platform/linux-generic/pktio/loop.c
```diff
@@ -173,6 +176,41 @@ static int loopback_recv(pktio_entry_t *pktio_entry, int index ODP_UNUSED,
...
+			      odp_bool_t *l3_proto_v4,
+			      uint8_t *l4_proto)
+{
+	uint8_t l3_proto_ver = _ODP_IPV4HDR_VER(*(uint8_t *)l3_hdr);
+
+	if (l3_proto_ver == _ODP_IPV4 && l3_len >= _ODP_IPV4HDR_LEN) {
+		_odp_ipv4hdr_t *ip = l3_hdr;
+
+		if (_ODP_IPV4HDR_IS_FRAGMENT(_odp_be_to_cpu_16(ip->frag_offset)))
+			return -1;
```
bogdanPricope: If it is an IPv4 fragment, we should still be able to compute IPv4 csum (but not UDP/TCP csum).
Use (as for dpdk pktio):
*l3_proto_v4 =1;
*l4_proto = 0;

lumag: ack, thanks

lumag:
@Bill-Fischofer-Linaro I'm not convinced about data16 change at this moment. We had code like this for ages (in checksumming code and in other places). I will put this into my todo list, but I do not feel like we should change this at this moment. Instead we should find time (after TM) and audit all type casts in ODP code.

bogdanPricope:
My feeling is that we should avoid this kind of last minute / big patch set.

My thoughts/worries below:
1.	Too big for a thorough review.
2.	Need performance analysis to determine impact of supplemental operations on packet parsing for DPDK pktio – seems to work as far as I can tell but no idea on performance. 
Doing checksum validation with packet parsing is nice - I would prefer to have a new packet parser for DPDK pktio (using info provided by dpdk) + dpdk csum part  and a generic software parser with SW csum instead of modifying existing parser to have csum on/off switch.

3.	Implementation seems incomplete – if we add checksum validation/calculation for loop pktio than why not adding for all ‘software’ pktios
4.	Computing partial sums (pseudo header sum as you parse) is a nice idea – no idea on performance impact.

platform/linux-generic/include/odp_chksum_internal.h
```diff
@@ -0,0 +1,71 @@
...
+		data = (const uint16_t *)((uintptr_t)p+1);
+		len--;
+	} else {
+		sum = 0;
+		data = (const uint16_t *)p;
+	}
+
+	while (len > 1) {
+		sum += *data++;
+		len -= 2;
```
Bill-Fischofer-Linaro: This routine will still fail if `odd_offset` is true and `p` is even at entry. This will occur if a packet consists of two segments, the first of which has an odd number of bytes to be checksummed since ODP will start each segment on an even boundary. If will similarly fail if `odd_offset` is false and `p` is odd at entry. That would occur if the second segment had been truncated . 

`odd_offset` and the even/odd values of `p` are independent and need to be handled separately.

lumag: Yes, I understand that. I'd like to point to `helper/chksum.c` and current `platform/l-g/odp_chksum.c`.

test/validation/api/ipsec/test_vectors.h
lumag: I hand-calculated them.

lumag:
@bogdanPricope I can try dropping the major part, leaving just IPsec checksum generation, which is required from API point of view.

bogdanPricope:
@lumag If the others are OK with current PR (had time to review it in detail, etc.) then is OK with me, too.

JannePeltonen:
I cannot fully see the usefulness of adding SW based L4 checksum validation in RX. If a significant portion of traffic does not end up in local UDP termination (e.g. it gets forwarded or dropped) enabling SW based RX checksum validation would waste a lot of cycles. The added code also adds a little bit of overhead even when checksumming is not enabled.

platform/linux-generic/odp_packet.c
```diff
@@ -2468,6 +2473,97 @@ int _odp_packet_ipv4_chksum_insert(odp_packet_t pkt)
...
+ * Calculate and fill in UDP checksum
+ *
+ * @param pkt  ODP packet
+ *
+ * @retval 0 on success
+ * @retval <0 on failure
+ */
+int _odp_packet_udp_chksum_insert(odp_packet_t pkt)
+{
+	return _odp_packet_tcp_udp_chksum_insert(pkt, _ODP_IPPROTO_UDP);
```
JannePeltonen: This code does not seem to handle the case when the calculated UDP checksum is 0 in which case it should be transmitted as 0xffff.

platform/linux-generic/include/odp_chksum_internal.h
Bill-Fischofer-Linaro: This can be made robust without any real loss of clarity or performance:
```
static inline uint16_t data16_unaligned(void *p)
{
	return (*(uint8_t *)p << 8) | (*(uint8_t *)(p + 1));
}

static uint32_t _odp_chksum_ones_comp16_32(const void *p, uint32_t len,
					   odp_bool_t odd_offset)

{
	uint32_t sum;
	const uint16_t *data;

	if (odp_unlikely(len > 0 && odd_offset)) {
		swap_buf_t sw;

		sw.b[0] = 0;
		sw.b[1] = *(const uint8_t *)p;
		sum = sw.w;
		data = (const uint16_t *)((uintptr_t)p + 1);
		len--;
	} else {
		sum = 0;
		data = (const uint16_t *)p;
	}

	if (odp_unlikely(data % 2)) {
		while (len > 1) {
			sum += data16_unaligned(data++);
			len -= 2;
		}
	} else {
		while (len > 1) {
			sum += *data++;
			len -= 2;
		}
	}

	/* Add left-over byte, if any */
	if (len > 0) {
		swap_buf_t sw;

		sw.b[0] = *(const uint8_t *)data;
		sw.b[1] = 0;
		sum += sw.w;
	}

	return sum;
}
```

platform/linux-generic/odp_packet.c
lumag: @JannePeltonen good catch, thank you. Should be fixed now.

lumag:
@Bill-Fischofer-Linaro I've changed low-level checksum calculation code to properly support unaligned case. It should be more robust now.

lumag:
Rebased on top of current master

platform/linux-generic/pktio/dpdk.c
psavol: Agree. XXX_NO_CHKSUMS is better than explicit casts.

platform/linux-generic/include/odp_packet_internal.h
```diff
@@ -57,6 +57,12 @@ typedef struct {
 
 	/* offset to L4 hdr (TCP, UDP, SCTP, also ICMP) */
 	uint16_t l4_offset;
+
+	/* Partial sum for L4 checksumming */
+	uint32_t l4_part_sum;
+
+	/* L4 checksum */
+	uint32_t l4_sum;
```
psavol: This is performance sensitive part of packet header. Data before and after this point is needed frequently. And small modifications here have made +-10% performance impact previously.

When L4 checksum is calculated in SW, we are already ruined performance, so these two could be located in the bottom of the struct, e.g. just before crypto results.

psavol:
Not yet Maxim, may be it was some other update (rebase ?).

platform/linux-generic/odp_packet_io.c
```diff
@@ -455,6 +455,12 @@ int odp_pktio_config(odp_pktio_t hdl, const odp_pktio_config_t *config)
 
 	entry->s.config = *config;
 
+	entry->s.in_chksums.all_chksum = 0;
+	entry->s.in_chksums.chksum.ipv4 = config->pktin.bit.ipv4_chksum;
```
bogdanPricope: Any plan to integrate 'drop' flags (drop_ipv4_err, drop_ipv6_err, drop_udp_err, drop_tcp_err) with the parser?
Will it make sense to use odp_pktin_config_opt_t (pktio_entry->s.config.pktin) instead of making a copy of csum flags in the new in_chksums field?

platform/linux-generic/pktio/dpdk.c
psavol: A define should be added for "(odp_proto_chksums_t)0U"  - e.g. PARSE_NO_CHKSUM.

matiaselo:
The performance impact of this patch set is minimal, so it's OK for me.

platform/linux-generic/pktio/dpdk.c
lumag: @psavol rebased PR, so that NO_CHKSUM is unnecessary now.

platform/linux-generic/odp_packet.c
```diff
@@ -2200,28 +2227,64 @@ static inline uint8_t parse_ipv6(packet_parser_t *prs, const uint8_t **parseptr,
...
-	if (odp_unlikely(udplen < sizeof(_odp_udphdr_t)))
+	if (odp_unlikely(udplen < sizeof(_odp_udphdr_t))) {
 		prs->flags.udp_err = 1;
+		return;
+	}
+
+	if (chksums.chksum.udp &&
+	    !prs->input_flags.ipfrag) {
+		if (udp->chksum == 0) {
+			prs->input_flags.l4_chksum_done =
```
bogdanPricope: Is this right? l4_chksum_err?

muvarov:
Merged.

PR #390, opened 2018-01-16
[PATCH v7] Support linking tests/examples dynamically with libodp-linux.so

lumag:
Distributions won't like statically-linked binaries. Provide configure switch to link examples and tests dynamically.

### DIFF
--- a/configure.ac
+++ b/configure.ac
@@ -183,6 +183,16 @@ AC_DEFINE_UNQUOTED([IMPLEMENTATION_NAME], ["$IMPLEMENTATION_NAME"],
 		   [Define to the name of the implementation])
 
 ##########################################################################
+# Build examples/tests dynamically
+##########################################################################
+AC_ARG_ENABLE([static-applications],
+	      [AS_HELP_STRING([--disable-static-applications],
+			      [disable static linking of examples and tests]
+			      [ with ODP])], [],
+	      [enable_static_applications=yes])
+AM_CONDITIONAL([STATIC_APPS], [test "x$enable_static_applications" != "xno"])
+
+##########################################################################
 # Include m4 files
 ##########################################################################
 m4_include([./doc/m4/configure.m4])
@@ -393,6 +403,7 @@ AC_MSG_RESULT([
 	Deprecated APIs:	${deprecated}
 	debug:			${enable_debug}
 	cunit:			${cunit_support}
+	static tests linkage:	${enable_static_applications}
 	test_vald:		${test_vald}
 	test_perf:		${test_perf}
 	test_perf_proc:		${test_perf_proc}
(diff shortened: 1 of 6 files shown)


lumag:
Related questions:
 - Should we make this a default behaviour?
 - Should we drop static-linkage of tests alltogether?

example/Makefile.inc
```diff
@@ -13,4 +13,8 @@ AM_CFLAGS = \
 	-I$(top_srcdir)/platform/@with_platform@/arch/@ARCH_DIR@ \
 	-I$(top_builddir)/include
 
+if STATIC_LINK
 AM_LDFLAGS = -L$(LIB) -static
+else
+AM_LDFLAGS =
```
bogdanPricope: I am using something similar for ODP code instrumentation with PAPI project in Caterpillar. Yet, I am surprised you don't get link errors when building examples with shared libs.

+if CODE_INSTRUM
 +AM_LDFLAGS = -L$(LIB) -lssl -lcrypto -latomic
 +else
  AM_LDFLAGS = -L$(LIB) -static
 +endif

lumag: Hmm. Strange. Those libs should come from -lodp-linux, because they are included into .la file.

configure.ac
```diff
@@ -172,6 +172,16 @@ AS_IF([test "${with_platform}" = "linux-generic"],
 AC_DEFINE_UNQUOTED([IMPLEMENTATION_NAME], ["$IMPLEMENTATION_NAME"],
 		   [Define to the name of the implementation])
 
+##########################################################################
+# Build examples/tests dynamically
+##########################################################################
+AC_ARG_ENABLE([static-tests],
```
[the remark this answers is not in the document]
lumag: I'll change to static-applications

lumag:
@bogdanPricope 
On Debian dpdk is also a shared lib. So no libs are statically linked in such case.

Regarding the wrapper, there is no way I know. Libtool uses it to substitute library paths so that you can run/debug apps whith local library version. Use `./libtool --mode=execute gdb ./examples/generator/odp_generator` to run e.g. gdb.

lumag:
Rebasing to fix conflicts

muvarov:
hm, looks like Travis disappeared from this PR. Clicking Close and then again Open should help.

muvarov:
btw, I'm ok with changes.

bogdanPricope:
Fine with me, yet I am not an autotools expert.
ODP shared lib contains dpdk symbols… I guess this was intended, right?

lumag:
@bogdanPricope it was indended that libodp-linux.so does not require any additional flags to link application. If DPDK is static library, ODP will contain it. If DPDK is shared library, ODP will link to it (of course if PktIO DPDK is enabled). I'm yet to play with versioning/mapping/hiding of symbols (see https://bugs.linaro.org/show_bug.cgi?id=2988).

lumag:
@muvarov there were issues with Travis, now it is back (see the last line in checks list, I had to scroll to it).

Bill-Fischofer-Linaro:
@bogdanPricope Please add your review if you're OK with this.

bogdanPricope:

muvarov:
Merged.

PR #392, opened 2018-01-16
[PATCH API-NEXT v4] User pointer init to NULL - rebased

psavol:
Rebased PR #388 into a new branch due to heavy api-next history rewrite.

### DIFF
--- a/platform/linux-generic/odp_packet.c
+++ b/platform/linux-generic/odp_packet.c
@@ -339,9 +339,8 @@ void *_odp_packet_map(void *pkt_ptr, uint32_t offset, uint32_t *seg_len,
 void packet_parse_reset(odp_packet_hdr_t *pkt_hdr)
 {
 	/* Reset parser metadata before new parse */
-	pkt_hdr->p.error_flags.all  = 0;
 	pkt_hdr->p.input_flags.all  = 0;
-	pkt_hdr->p.output_flags.all = 0;
+	pkt_hdr->p.flags.all.error  = 0;
 	pkt_hdr->p.l2_offset        = ODP_PACKET_OFFSET_INVALID;
 	pkt_hdr->p.l3_offset        = ODP_PACKET_OFFSET_INVALID;
 	pkt_hdr->p.l4_offset        = ODP_PACKET_OFFSET_INVALID;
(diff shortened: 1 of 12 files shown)


platform/linux-generic/odp_packet.c
```diff
@@ -1259,7 +1260,15 @@ int odp_packet_input_index(odp_packet_t pkt)
 
 void odp_packet_user_ptr_set(odp_packet_t pkt, const void *ptr)
 {
-	packet_hdr(pkt)->buf_hdr.user_ptr = ptr;
+	odp_packet_hdr_t *pkt_hdr = packet_hdr(pkt);
+
+	if (odp_unlikely(ptr == NULL)) {
+		pkt_hdr->p.flags.user_ptr_set = 0;
+		return;
+	}
```
Bill-Fischofer-Linaro: This "optimization" is unnecessary. If the user wants to set the user ptr to `NULL` that's as legitimate a value as any and will be retrieved normally. All we care about is whether `odp_packet_user_ptr_set()` has been called. The flag is reset by `odp_packet_reset()`.

psavol: This makes the get side work consistently - code path and performance is the same for NULL values. The flag tells if NULL is returned, or pointer value needs to be read. Reading the pointer value may cause cache miss. The flag needs to be read in any case. If flag is zero, we can return NULL right away (no need to wait for cache to bring NULL value from memory).

     if (flags.user_ptr_set == 0)
         return NULL;

All NULL cases avoid going to the memory:
pkt = alloc()
user_ptr(); <-- NULL
user_ptr_set(pkt, &foo);
user_ptr(); <-- != NULL
user_ptr_set(pkt, NULL);
user_ptr(); <-- NULL
user_ptr(); <-- NULL
user_ptr(); <-- NULL
user_ptr(); <-- NULL

platform/linux-generic/include/odp/api/plat/packet_types.h
```diff
@@ -67,101 +67,89 @@ typedef enum {
 
 #endif
 
-/** @internal Packet field accessor */
+/** @cond _ODP_HIDE_FROM_DOXYGEN_ */
+
```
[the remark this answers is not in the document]
psavol: This a section label for doxygen. We'll never add it to doxygen config or define it otherwise. I think it just need to high light that this portion of a file intentionally left hiden from doxygen documentation. "Internal" does not tell that this is only a doxygen label, it would seem like a  configuration option or a preprocessor macro.

platform/linux-generic/odp_packet.c
Bill-Fischofer-Linaro: OK

platform/linux-generic/include/odp/api/plat/packet_types.h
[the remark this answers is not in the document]
Bill-Fischofer-Linaro: OK

Bill-Fischofer-Linaro:
@psavol Please rebase to resolve conflicts.

psavol:
v2, v3, v4: rebased

muvarov:
Merged.

PR #395, opened 2018-01-16
[PATCH v5] enable linux helpers by default

psavol:
Enable Linux helpers by default as e.g. OFP uses those. Build
and test helper code does not add significant overhead, but
catches potential issues early.

### DIFF
--- a/helper/m4/configure.m4
+++ b/helper/m4/configure.m4
@@ -8,15 +8,12 @@ AC_ARG_ENABLE([test-helper],
 AM_CONDITIONAL([test_helper], [test x$test_helper = xyes ])
 
 ##########################################################################
-# Enable/disable helper-ext
-# platform specific non portable extensions
+# Enable/disable Linux helpers
 ##########################################################################
-helper_linux=no
 AC_ARG_ENABLE([helper-linux],
-	[  --enable-helper-linux	build helper platform extensions (not portable)],
-	[if test "x$enableval" = "xyes"; then
-		helper_linux=yes
-	fi])
+    [AS_HELP_STRING([--disable-helper-linux], [disable Linux helpers])],
+    [helper_linux=$enableval],
+    [helper_linux=yes])
 
 AC_CONFIG_FILES([helper/Makefile
 		 helper/libodphelper.pc


helper/m4/configure.m4
```diff
@@ -8,12 +8,11 @@ AC_ARG_ENABLE([test-helper],
...
-# Enable/disable helper-ext
-# platform specific non portable extensions
+# Enable/disable Linux helpers
 ##########################################################################
-helper_linux=no
+helper_linux=yes
 AC_ARG_ENABLE([helper-linux],
-	[  --enable-helper-linux	build helper platform extensions (not portable)],
+	[  --enable-helper-linux   enable Linux helpers],
 	[if test "x$enableval" = "xyes"; then
```
joseppc: Shouldn't we add a test now checking if $enableval is "no" here?

psavol: corrected in v2

helper/m4/configure.m4
```diff
@@ -8,14 +8,13 @@ AC_ARG_ENABLE([test-helper],
...
-# platform specific non portable extensions
+# Enable/disable Linux helpers
 ##########################################################################
-helper_linux=no
+helper_linux=yes
 AC_ARG_ENABLE([helper-linux],
-	[  --enable-helper-linux	build helper platform extensions (not portable)],
-	[if test "x$enableval" = "xyes"; then
-		helper_linux=yes
+	[  --enable-helper-linux   enable Linux helpers],
```
lumag: As you are touching this line, could you please use AS_HELP_STRING here?

lumag:
@psavol how hard would be porting of OFP to platform-agnostic thread helpers?

helper/m4/configure.m4
muvarov: if you enable it by default ./configure --help should print --disable-helper-linux   disable Linux helpers.

muvarov: @lumag I guess just copy paste code from odp to ofp :)

psavol: I think all config options we have added should be documented as enable-xxx. The help output is harder to read when some options are documented as enable-xxx (do xxx) and others disable-xxx (don't do xxx). The disable option is available in any case.

I trust only on the defaults printed by ./configure, not the help documentation which may be outdated.

psavol: Done in v3

muvarov: I think no. You have some option by default. They might be on or off. ./configure can change default settings. And help should print right information about that. --enable for already enabled option makes no sense.

psavol: Compare e.g. autoconf enable shared/static library configs:

--enable-shared[=PKGS]  build shared libraries [default=yes]
--enable-static[=PKGS]  build static libraries [default=yes]

I did the same in v3
--enable-helper-linux   enable Linux helpers [default=yes]

I think we could do the same for all our configs in help text: --enable-xxx [default=yes/no]
--disable-xxx is always there for those that want to use that instead of --enable-xxx=no

muvarov:
@psavol that is not consistent with other configure options. I think --disable- is better here. @lumag what is you opinion?

lumag:
I've checked several other projects. They use `--disable-SMTH` for options which default to `yes`.

psavol:
This is from our own ./configure --help output, autoconfig default options:

--enable-shared[=PKGS] build shared libraries [default=yes]
--enable-static[=PKGS] build static libraries [default=yes]

lumag:
In this case it makes sense, as `--enable-shared` option takes list of packages as an argument [libtool manual](https://www.gnu.org/software/libtool/manual/html_node/LT_005fINIT.html#LT_005fINIT). In our case all options are simple yes/no.

muvarov:
Merged.

PR #396, opened 2018-01-16
[PATCH v1] validation: packet: minimal chksum status test

psavol:
There was no references into chksum status functions in validation
test suite. Add minimal test to see at least that functions exist.
Parse function should be updated with a checksum implementation,
before actual checksum checking can be tested.


muvarov:
Merged.

PR #398, opened 2018-01-17
[PATCH v2] linux-gen: align: Fix alignment for typedef definitions

apalos:
* typedef struct { } ODP_ALIGN() test_t; is correct
	* typedef struct ODP_ALIGN() {} test_t; is correct and
	preferred from gcc standards
	* typedef struct {} test_t ODP_ALIGN() discards align
	Had segfaults on gcc-7 using that


apalos:
There's a false positive failure on checkpatch on test/validation/api/chksum/chksum.c

platform/linux-generic/include/odp_ipsec_internal.h
```diff
@@ -92,7 +92,7 @@ int _odp_ipsec_status_send(odp_queue_t queue,
 #define ODP_CONFIG_IPSEC_SAS	8
 
 struct ipsec_sa_s {
-	odp_atomic_u32_t state ODP_ALIGNED_CACHE;
+	odp_atomic_u32_t ODP_ALIGNED_CACHE state;
```
psavol: The new attribute placement rule was for structs, union, etc, but how about integers ?

apalos: The real bug was only present on typedef since the attribute was ignored if  placed after the terminating }.
typedef unsigned long long __u64 __attribute__((aligned(16))); 
should be equal to 
typedef unsigned long long  __attribute__((aligned(16))) __u64;

apalos:
V2: Updated checkpatch for the false positive

PR #401, opened 2018-01-17
[PATCH v1] restructure headers for ABI-compat/platform-optimized modes

lumag:
#250 rebased on top of master branch


Bill-Fischofer-Linaro:
@lumag I thought we had suggested to stop patching master directly but only via api-next to avoid the backporting issue.

lumag:
@Bill-Fischofer-Linaro It is up to @muvarov. Anyway we would need to do one final merge/rebase of api-next and master

PR #403, opened 2018-01-18
[PATCH API-NEXT v5] api: ipsec: TFC padding and dummy packets

psavol:
Specify how application can recognize and create TFC padding and dummy packets, both in inbound and outbound directions.

### DIFF
--- a/include/odp/api/spec/ipsec.h
+++ b/include/odp/api/spec/ipsec.h
@@ -991,6 +991,22 @@ typedef struct odp_ipsec_out_opt_t {
 			/** Use fragmentation mode option */
 			uint32_t frag_mode: 1;
 
+			/** Use TFC padding length option */
+			uint32_t tfc_pad:   1;
+
+			/** Tunnel mode TFC dummy packet. This can be used only
+			 *  in tunnel mode. When the flag is set, packet length
+			 *  and content is ignored and instead a TFC dummy
+			 *  packet is created during IPSEC operation. The dummy
+			 *  packet length is defined by 'tfc_pad_len' option.
+			 *  If the SA is configured to copy IP header fields
+			 *  from inner IP packet, those fields must be passed
+			 *  with IP parameters option. */
+			uint32_t tfc_dummy: 1;
+
+			/** Use IP parameters option */
+			uint32_t ip_param:  1;
+
 		} flag;
 
 		/** All flag bits */
@@ -999,6 +1015,26 @@ typedef struct odp_ipsec_out_opt_t {
 
 	/** Fragmentation mode */
 	odp_ipsec_frag_mode_t frag_mode;
+
+	/** TFC padding length
+	 *
+	 *  Number of TFC padding bytes added to the packet during IPSEC
+	 *  processing. Resulting packet should not exceed the maximum packet
+	 *  length of the pool, otherwise IPSEC operation may fail.
+	 *  Implementation guarantees that the padding does not contain any
+	 *  confidential information. */
+	uint32_t tfc_pad_len;
+
+	/** Union of IP parameters */
+	union {
+		/** Override IPv4 parameters in outer header creation.
+		 *  IP addresses are ignored. */
+		odp_ipsec_ipv4_param_t ipv4;
+
+		/** Override IPv6 parameters in outer header creation.
+		 *  IP addresses are ignored. */
+		odp_ipsec_ipv6_param_t ipv6;
+	};
 
 } odp_ipsec_out_opt_t;
 
@@ -1302,7 +1338,13 @@ int odp_ipsec_in(const odp_packet_t pkt_in[], int num_in,
  * The operation does packet transformation according to IPSEC standards (see
  * e.g. RFC 4302 and 4303). Resulting packets are well formed IP packets
  * with IPSEC, etc headers constructed according to the standards. The amount
- * and content of packet data before the IP header is undefined.
+ * and content of packet data before the IP header is undefined. Use outbound
+ * operation parameters to specify the amount of TFC padding appended to
+ * the packet during IPSEC transformation. Options can be used also to create
+ * TFC dummy packets. Packet data content is ignored in tunnel mode TFC dummy
+ * packet creation as tfc_pad_len option defines solely the packet length.
+ * In all other cases, payload length for the IPSEC transformation is specified
+ * by odp_packet_len() minus odp_packet_l3_offset() plus tfc_pad_len option.
  *
  * Each successfully transformed packet has a valid value for these metadata:
  * - L3 offset: Offset to the first byte of the (outmost) IP header
--- a/include/odp/api/spec/ipsec.h
+++ b/include/odp/api/spec/ipsec.h
@@ -1226,12 +1226,23 @@ typedef struct odp_ipsec_status_t {
  * e.g. RFC 4302 and 4303). Resulting packets are well formed, reconstructed
  * original IP packets, with IPSEC headers removed and valid header field values
  * restored. The amount and content of packet data before the IP header is
- * undefined.
+ * undefined. Some amount of TFC padding may follow the IP packet payload,
+ * in which case packet length is larger than protocol headers indicate.
+ * TFC dummy packets have l3_type set to ODP_PROTO_L3_TYPE_NONE in tunnel mode
+ * or l4_type set to ODP_PROTO_L4_TYPE_NO_NEXT in transport mode. Dummy
+ * packets contain implementation specific amount of (dummy) data. Furthermore,
+ * inline IPSEC processing may drop dummy packets.
  *
  * Each successfully transformed packet has a valid value for these metadata
  * regardless of the inner packet parse configuration
  * (odp_ipsec_inbound_config_t):
- * - L3 offset: Offset to the first byte of the (outmost) IP header
+ * - l3_offset: Offset to the first byte of the original IP packet. The value
+ *              is implementation specific for tunnel mode TFC dummy packets.
+ * - l3_type:   Specifies if the original packet is IPv4 or IPv6. For tunnel
+ *              mode TFC dummy packets set to ODP_PROTO_L3_TYPE_NONE.
+ * - l4_type:   Always set to ODP_PROTO_L4_TYPE_NO_NEXT for transport mode dummy
+ *              packets. Otherwise, depends on parse configuration. Default
+ *              value is ODP_PROTO_L4_TYPE_NONE.
  * - pktio:     For inline IPSEC processed packets, original packet input
  *              interface
  *
--- a/include/odp/api/spec/ipsec.h
+++ b/include/odp/api/spec/ipsec.h
@@ -983,6 +983,20 @@ typedef struct odp_ipsec_op_flag_t {
  * These may be used to override some SA level options
  */
 typedef struct odp_ipsec_out_opt_t {
+	/** Union of all flag bits */
+	union {
+		/** Option flags. Set flag for those options that are
+		 *  used, all other options are ignored. */
+		struct {
+			/** Use fragmentation mode option */
+			uint32_t frag_mode: 1;
+
+		} flag;
+
+		/** All flag bits */
+		uint32_t all_flags;
+	};
+
 	/** Fragmentation mode */
 	odp_ipsec_frag_mode_t frag_mode;
 
--- a/include/odp/api/spec/ipsec.h
+++ b/include/odp/api/spec/ipsec.h
@@ -238,6 +238,11 @@ typedef struct odp_ipsec_capability_t {
 	 */
 	odp_support_t retain_header;
 
+	/**
+	 * Inner packet checksum check offload support in inbound direction.
+	 */
+	odp_proto_chksums_t chksums_in;
+
 	/** Maximum number of different destination CoSes in classification
 	 *  pipelining. The same CoS may be used for many SAs. This is equal or
 	 *  less than 'max_cos' capability in classifier API.
--- a/include/odp/api/spec/ipsec.h
+++ b/include/odp/api/spec/ipsec.h
@@ -984,7 +984,7 @@ typedef struct odp_ipsec_op_flag_t {
  */
 typedef struct odp_ipsec_out_opt_t {
 	/** Fragmentation mode */
-	odp_ipsec_frag_mode_t mode;
+	odp_ipsec_frag_mode_t frag_mode;
 
 } odp_ipsec_out_opt_t;
 
--- a/include/odp/api/spec/ipsec.h
+++ b/include/odp/api/spec/ipsec.h
@@ -382,6 +382,44 @@ typedef struct odp_ipsec_crypto_param_t {
 
 } odp_ipsec_crypto_param_t;
 
+/** IPv4 header parameters */
+typedef struct odp_ipsec_ipv4_param_t {
+	/** IPv4 source address (NETWORK ENDIAN) */
+	void *src_addr;
+
+	/** IPv4 destination address (NETWORK ENDIAN) */
+	void *dst_addr;
+
+	/** IPv4 Differentiated Services Code Point */
+	uint8_t dscp;
+
+	/** IPv4 Don't Fragment bit */
+	uint8_t df;
+
+	/** IPv4 Time To Live */
+	uint8_t ttl;
+
+} odp_ipsec_ipv4_param_t;
+
+/** IPv6 header parameters */
+typedef struct odp_ipsec_ipv6_param_t {
+	/** IPv6 source address (NETWORK ENDIAN) */
+	void *src_addr;
+
+	/** IPv6 destination address (NETWORK ENDIAN) */
+	void *dst_addr;
+
+	/** IPv6 flow label */
+	uint32_t flabel;
+
+	/** IPv6 Differentiated Services Code Point */
+	uint8_t dscp;
+
+	/** IPv6 hop limit */
+	uint8_t hlimit;
+
+} odp_ipsec_ipv6_param_t;
+
 /**
  * IPSEC tunnel parameters
  *
@@ -397,40 +435,10 @@ typedef struct odp_ipsec_tunnel_param_t {
 	/** Variant mappings for tunnel parameters */
 	union {
 		/** IPv4 header parameters */
-		struct {
-			/** IPv4 source address (NETWORK ENDIAN) */
-			void *src_addr;
-
-			/** IPv4 destination address (NETWORK ENDIAN) */
-			void *dst_addr;
-
-			/** IPv4 Differentiated Services Code Point */
-			uint8_t dscp;
-
-			/** IPv4 Don't Fragment bit */
-			uint8_t df;
-
-			/** IPv4 Time To Live */
-			uint8_t ttl;
-		} ipv4;
+		odp_ipsec_ipv4_param_t ipv4;
 
 		/** IPv6 header parameters */
-		struct {
-			/** IPv6 source address (NETWORK ENDIAN) */
-			void *src_addr;
-
-			/** IPv6 destination address (NETWORK ENDIAN) */
-			void *dst_addr;
-
-			/** IPv6 Differentiated Services Code Point */
-			uint8_t dscp;
-
-			/** IPv6 flow label */
-			uint32_t flabel;
-
-			/** IPv6 hop limit */
-			uint8_t hlimit;
-		} ipv6;
+		odp_ipsec_ipv6_param_t ipv6;
 	};
 } odp_ipsec_tunnel_param_t;
 
--- a/include/odp/api/abi-default/packet.h
+++ b/include/odp/api/abi-default/packet.h
@@ -30,6 +30,34 @@ typedef _odp_abi_packet_seg_t *odp_packet_seg_t;
 #define ODP_PACKET_SEG_INVALID    ((odp_packet_seg_t)0xffffffff)
 #define ODP_PACKET_OFFSET_INVALID 0xffff
 
+typedef uint8_t odp_proto_l3_type_t;
+
+#define ODP_PROTO_L3_TYPE_NONE   0
+#define ODP_PROTO_L3_TYPE_ARP    1
+#define ODP_PROTO_L3_TYPE_RARP   2
+#define ODP_PROTO_L3_TYPE_MPLS   3
+#define ODP_PROTO_L3_TYPE_IPV4   4
+#define ODP_PROTO_L3_TYPE_IPV6   6
+
+typedef uint8_t odp_proto_l4_type_t;
+
+/* Numbers from IANA Assigned Internet Protocol Numbers list */
+#define ODP_PROTO_L4_TYPE_NONE      0
+#define ODP_PROTO_L4_TYPE_ICMPV4    1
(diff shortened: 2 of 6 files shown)


include/odp/api/spec/ipsec.h
```diff
@@ -382,6 +382,44 @@ typedef struct odp_ipsec_crypto_param_t {
...
+	void *src_addr;
+
+	/** IPv6 destination address (NETWORK ENDIAN) */
+	void *dst_addr;
+
+	/** IPv6 Differentiated Services Code Point */
+	uint8_t dscp;
+
+	/** IPv6 flow label */
+	uint32_t flabel;
```
Bill-Fischofer-Linaro: Can `flabel` be placed after `dst_addr`? This would avoid the pad bytes that would otherwise be inserted between `dspc` and `flabel`.

include/odp/api/spec/ipsec.h
```diff
@@ -1210,16 +1256,22 @@ typedef struct odp_ipsec_status_t {
...
  * Each successfully transformed packet has a valid value for these metadata
  * regardless of the inner packet parse configuration
  * (odp_ipsec_inbound_config_t):
- * - L3 offset: Offset to the first byte of the (outmost) IP header
- * - pktio:     For inline IPSEC processed packets, original packet input
- *              interface
+ * - l3_offset:  Offset to the first byte of the original IP (or TFC dummy)
+ *               packet
+ * - has_ipv4/6: Specifies if the original packet is IPv4 or IPv6. For tunnel
+ *               mode TFC dummy packets neither flag is set.
```
Bill-Fischofer-Linaro: I assume  this is referring to the `odp_packet_has_ipv4()` and `odp_packet_has_ipv6()` accessor functions? Since these bits are only accessible via these functions, this forces applications to play a guessing game with them and their L4 counterparts. Might it be better to consider having `odp_packet_l3_proto()` and `odp_packet_l4_proto()` functions?

include/odp/api/spec/ipsec.h
```diff
@@ -238,6 +238,11 @@ typedef struct odp_ipsec_capability_t {
 	 */
 	odp_support_t retain_header;
 
+	/**
+	 * Inner packet checksum check offload support in inbound direction.
+	 */
+	odp_proto_chksums_t chksums_in;
```
Bill-Fischofer-Linaro: Is there no need for a corresponding `chksums_out` capability?

lumag: There is one indeed.

include/odp/api/spec/ipsec.h
```diff
@@ -983,9 +983,55 @@ typedef struct odp_ipsec_op_flag_t {
...
 
+	/** Union of IP parameters */
+	union {
+		/** Override IPv4 parameters in outer header creation.
+		 *  IP addresses are ignored. */
+		odp_ipsec_ipv4_param_t ipv4;
+
+		/** Override IPv6 parameters in outer header creation.
+		 *  IP addresses are ignored. */
+		odp_ipsec_ipv6_param_t ipv6;
```
lumag: What is the use case for these options?

include/odp/api/spec/ipsec.h
```diff
@@ -983,9 +983,55 @@ typedef struct odp_ipsec_op_flag_t {
...
+			/** Use fragmentation mode option */
+			uint32_t frag_mode: 1;
+
+			/** Use IP parameters option */
+			uint32_t ip_param:  1;
+
+			/** Use TFC padding length option */
+			uint32_t tfc_pad:   1;
+
+			/** Tunnel mode TFC dummy packet. In tunnel mode, set
```
lumag: Hmm. I think RFC 4303 does not limit TFC dummy packets to tunnel mode. One can generate them in transport mode.

lumag:
@psavol I like the approach. If we can postpone ipv4/ipv6_param separation till we are able to define API for tunnels, that would be great.

include/odp/api/spec/ipsec.h
JannePeltonen: Why? The rationale goes that if checksumming is requested in the outbound direction the implementation can always calculate in in SW since that is what the application would have to do otherwise. Inbound direction is different since the need for L4 checksum checking (i.e. is the packet destined to this system of just forwarded) is not yet known at the time of reception (so if an implementation sets the inbound capability, it should mean that it can do checksumming clearly more efficiently than pure SW).

include/odp/api/spec/ipsec.h
JannePeltonen: Two cases:
1) Enabling TFC dummy packet generation for tunnel-mode SAs that have been configured to copy the fields from the inner header. This way the input packet to the IPsec operation does not have to contain valid IP header for the copying to work which would be difficult to specify for dummy packets (e.g. how to tell if the inner packet is IPv4 or IPv6). The fields cannot just be left to some default values in dummy packets because that could allow one to distinguish the dummy packets from normal packets in the wire, rendering TFC dummy packets useless.

2) Making the current API more complete (regardless of TFC support). Currently it is possible to set those fields only to the same SA-specific value or copy from the inner header, but not set the value depending on the packet. I could imagine that for DSCP there could be real use cases where the DSCP cannot just be copied (e.g. since if the inner and outer packet belong to different QoS domains with different DSCP interpretation) but the DSCP cannot also be the same for all packets of an SA (although It would be better to you separate SAs in that case).

include/odp/api/spec/ipsec.h
JannePeltonen: That is right, but there is no limitation in this API.

This bit is just for tunnel mode dummy packets that cannot otherwise be sent. Transport mode dummy TFC packets are sent in the normal way: The input packet to an oubound IPsec operation is a well formed IP packet just like in the normal packet case, but the application just sets the protocol field of the IP header to 59. The ODP implementation needs the IP header to be there since that IP header is used (after some adjustments) in the resulting ESP/AH packet. This is different from tunnel mode where the outer IP header is generated based on the information in the SA.

include/odp/api/spec/ipsec.h
psavol: yes it could. I try to remember that if v2 is needed.

include/odp/api/spec/ipsec.h
psavol: Application needs to anyway check if packet is v4 or v6. Today it's checking that from first byte of the packet. With TFC tunnel mode, first byte is garbage that cannot be used any more. So, application uses these APIs instead in if - else if - else fashion. There's no more guessing than before.

include/odp/api/spec/ipsec.h
psavol: 3. Also regardless of TFC support. If inner packet is IPv4, but outer is IPv6 and e.g. flabel has been configured to be copied from inner to outer - there's no flabel in inner packet to copy. So, application is able to use this option to pass per packet IPv6 parameters when inner is IPv4, and wise versa.

psavol:
As explained above, those IP params are need for TFC dummy creation in tunnel mode and to enabled normal v4 packets in v6, or v6 packets in v4 tunnel (override copy from inner config).

include/odp/api/spec/ipsec.h
Bill-Fischofer-Linaro: OK

include/odp/api/spec/ipsec.h
Bill-Fischofer-Linaro: Please do. While padding is sometimes unavoidable, we should try not to introduce it in new structs.

include/odp/api/spec/ipsec.h
Bill-Fischofer-Linaro: OK

include/odp/api/spec/ipsec.h
```diff
@@ -983,9 +983,55 @@ typedef struct odp_ipsec_op_flag_t {
...
+		/** Override IPv6 parameters in outer header creation.
+		 *  IP addresses are ignored. */
+		odp_ipsec_ipv6_param_t ipv6;
+	};
+
+	/** TFC padding length
+	 *
+	 *  Number of TFC padding bytes added to the packet during IPSEC
+	 *  processing. Implementation guarantees that the padding does not
+	 *  contain any confidential information. */
```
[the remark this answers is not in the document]
Bill-Fischofer-Linaro: The implementation simply needs to ensure that it doesn't reuse application buffers, as they may contain residual confidential information. This can be done by inserting zeros or random bytes as the implementation chooses. Given the importance of security, it's appropriate to put this caveat in the spec as a reminder to implementations.

include/odp/api/spec/ipsec.h
```diff
@@ -983,9 +983,55 @@ typedef struct odp_ipsec_op_flag_t {
...
+			/** Use TFC padding length option */
+			uint32_t tfc_pad:   1;
+
+			/** Tunnel mode TFC dummy packet. In tunnel mode, set
+			 *  this flag to create a TFC dummy packet. The flag
+			 *  indicates packet data (at L3 offset) does not
+			 *  contain an inner packet IP header. If SA is
+			 *  configured to copy IP header fields from inner
+			 *  packet, those fields must be passed with
+			 *  IP parameters option. */
```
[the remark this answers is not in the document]
psavol: It has been specified that number of bytes to IPsec encrypt is always: packet_len - l3_offset + tfc_padding. In all other cases (than TFC dummy), L3 offset point to first byte of IP header. With dummy, application is not required to form an IP header and also packet data content is ignored. So, application can e.g. set L3 offset = 0, packet_len = 0 and tfc_padding = 512, OR L3 offset = 64, packet_len = 320, tfc_padding = 256, OR L3 offset = 0, packet_len = 512, tfc_padding= 0.

All those examples would end up implementation creating a 512 byte dummy packet.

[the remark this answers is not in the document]
psavol: We can also decide to write the spec differently. I.e. if dummy flag is set, only tfc_padding length defines the dummy packet length. May be it's better spec and as easy to implement than the one above.

include/odp/api/spec/ipsec.h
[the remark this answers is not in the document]
psavol: This should be clear for implementers already from RFC, but it's also here since it's easy to forget that data cannot be sent out from any random address (e.g. end of an buffer). Implementation must be sure about the memory content (zeros, rand, counter) before including that as padding.

include/odp/api/spec/ipsec.h
```diff
@@ -1210,16 +1256,22 @@ typedef struct odp_ipsec_status_t {
...
  * e.g. RFC 4302 and 4303). Resulting packets are well formed, reconstructed
- * original IP packets, with IPSEC headers removed and valid header field values
- * restored. The amount and content of packet data before the IP header is
- * undefined.
+ * original IP (or TFC dummy) packets, with IPSEC headers removed and valid
+ * header field values restored. The amount and content of packet data before
+ * the IP header is undefined. TFC padding may follow the IP packet payload,
+ * in which case packet length is larger than protocol headers indicate.
+ * TFC dummy packets have both IPv4 and IPv6 flags cleared, although L3 offset
+ * is set also for those.
```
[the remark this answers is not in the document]
psavol: I think the offset is already known (and possibly set into packet header) before implementation knows a packet is dummy, since that's part of decrypt operation (== start of cipher/clear text data). This spec allows L3 offset to be left at an implementation specific value, e.g. 0 if HW did drop the dummy content, e.g. 28 if HW does not recognize dummy packets (just another decrypted packet from HW point of view).

When dummy content is left there, application can see how much it is and the content. May be that helps in debugging dummy vs non-dummy packet, etc.

Result of an IPSEC decrypt transformation is always the original packet (== what the sender did send). L3 offset points to the first byte of the original data. If dummy data was dropped, then there's no data.

include/odp/api/spec/ipsec.h
psavol: About 2 vs. 3 step capability. Two level should be enough (no, yes). If there are 3 levels (no, yes in SW, yes in HW), throughput would be optimized is application enabled only when done in HW. Application would not request SW checksum for all packets, as it can filter out those that actually need it (e.g. half of the packets) and then do it in SW only for those.

include/odp/api/spec/ipsec.h
psavol: done

include/odp/api/spec/ipsec.h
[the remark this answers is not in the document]
bala-manoharan: We can also decide whether the application has to provide a valid odp_packet_t handle when "tfc_dummy" flag is set.

include/odp/api/spec/ipsec.h
```diff
@@ -1224,16 +1224,22 @@ typedef struct odp_ipsec_status_t {
...
  *
  * Each successfully transformed packet has a valid value for these metadata
  * regardless of the inner packet parse configuration
  * (odp_ipsec_inbound_config_t):
- * - L3 offset: Offset to the first byte of the (outmost) IP header
- * - pktio:     For inline IPSEC processed packets, original packet input
- *              interface
+ * - l3_offset:  Offset to the first byte of the original IP packet. The value
+ *               is implementation specific for tunnel mode TFC dummy packets.
+ * - has_ipv4/6: Specifies if the original packet is IPv4 or IPv6. For tunnel
```
lumag: I'm actually thinking about adding special `is_dummy` flag here. Transport mode TFC packets should have `has_ipv4`/`has_ipv6` flag according to their headers. However having special flag which can be checked by an application will allow one to drop them without going to L3 headers.

include/odp/api/spec/ipsec.h
```diff
@@ -1224,16 +1224,22 @@ typedef struct odp_ipsec_status_t {
...
  * e.g. RFC 4302 and 4303). Resulting packets are well formed, reconstructed
- * original IP packets, with IPSEC headers removed and valid header field values
- * restored. The amount and content of packet data before the IP header is
- * undefined.
+ * original IP (or TFC dummy) packets, with IPSEC headers removed and valid
+ * header field values restored. The amount and content of packet data before
+ * the IP header is undefined. TFC padding may follow the IP packet payload,
+ * in which case packet length is larger than protocol headers indicate.
+ * TFC dummy packets in tunnel mode have both IPv4 and IPv6 flags cleared and
+ * may contain some (dummy) data.
```
lumag: Also it should be documented that in INLINE mode hardware might drop dummy packets on its own, without reporting them to an application.

include/odp/api/spec/ipsec.h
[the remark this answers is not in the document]
lumag: @psavol I don't feel good about the 'if dummy flag is set, only tfc_padding length defines the dummy packet length' proposal. Your current phrase (`packet_len + tfc_padding` seems better to me).
@bala-manoharan if we do not provide a packet, how will implementation allocate a packet for SYNC/ASYNC modes?

include/odp/api/spec/ipsec.h
```diff
@@ -991,6 +991,21 @@ typedef struct odp_ipsec_out_opt_t {
 			/** Use fragmentation mode option */
 			uint32_t frag_mode: 1;
 
+			/** Use IP parameters option */
+			uint32_t ip_param:  1;
+
+			/** Use TFC padding length option */
+			uint32_t tfc_pad:   1;
+
+			/** Tunnel mode TFC dummy packet. This can be used only
```
lumag: Limiting this option to tunnel mode only seems logical but not that effective for an application. Maybe we can allow using this option together with L3 header override to create transport mode dummy packets. Then it would be just easier to inject dummy packets in transport case.

include/odp/api/spec/ipsec.h
```diff
@@ -1000,6 +1015,24 @@ typedef struct odp_ipsec_out_opt_t {
...
+		 *  IP addresses are ignored. */
+		odp_ipsec_ipv6_param_t ipv6;
+	};
+
+	/** TFC padding length
+	 *
+	 *  Number of TFC padding bytes added to the packet during IPSEC
+	 *  processing. Implementation guarantees that the padding does not
+	 *  contain any confidential information. */
+	uint32_t tfc_pad_len;
```
lumag: I have mixed feelings towards this option. Allocating additional buffer space from application-specific pools seems fragile to me. Maybe we can leave this out for now till there is hardware which actually supports generating additional TFC padding?

JannePeltonen: On January 17 pvanleeuwn@insidesecure.com wrote in the lng-odp list:

"However, for the EIP96 - EIP197PP range of solutions our HW /is/ capable of TFC padding insertion, we would just need to expose this through our driver and firmware. So API-wise you might want to support adding n bytes of TFC padding for outbound. "

So apparently such HW exists. And even with pure SW solution, I do not see why buffer space would need to be allocated from application specific pools for the padding data.

include/odp/api/spec/ipsec.h
```diff
@@ -1224,16 +1224,22 @@ typedef struct odp_ipsec_status_t {
  *
  * The operation does packet transformation according to IPSEC standards (see
  * e.g. RFC 4302 and 4303). Resulting packets are well formed, reconstructed
- * original IP packets, with IPSEC headers removed and valid header field values
- * restored. The amount and content of packet data before the IP header is
- * undefined.
+ * original IP (or TFC dummy) packets, with IPSEC headers removed and valid
```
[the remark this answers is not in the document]
lumag: After discussion during the call: this needs to be rephrased to point that TFC dummy packets can be truncated by an implementation. Maybe we should not mention that "restructured data" is returned for TFC dummy packet.

include/odp/api/spec/ipsec.h
```diff
@@ -1224,16 +1224,22 @@ typedef struct odp_ipsec_status_t {
...
  * Each successfully transformed packet has a valid value for these metadata
  * regardless of the inner packet parse configuration
  * (odp_ipsec_inbound_config_t):
- * - L3 offset: Offset to the first byte of the (outmost) IP header
- * - pktio:     For inline IPSEC processed packets, original packet input
- *              interface
+ * - l3_offset:  Offset to the first byte of the original IP packet. The value
+ *               is implementation specific for tunnel mode TFC dummy packets.
+ * - has_ipv4/6: Specifies if the original packet is IPv4 or IPv6. For tunnel
+ *               mode TFC dummy packets neither flag is set.
```
bala-manoharan: In tunnel mode not setting ipv4/v6 flag for dummy packet should be fine. But how do Dummy packets get handled in transport mode? coz if the implementation does inner packet parsing then transport mode dummy packet will have the valid ipv4/v6 flag set.

JannePeltonen: Transport mode dummy packets _are_ IP packets. The protocol field of the IP header (which is constructed based on the IP header of the ESP packet) is 59.

bala-manoharan: Yes. Just to be clear even if the implementation know that its a dummy packet in Transport mode the application/implementation will have to parse the IP header to know the dummy packet and drop the packet.

include/odp/api/spec/ipsec.h
```diff
@@ -1226,14 +1226,21 @@ typedef struct odp_ipsec_status_t {
  * e.g. RFC 4302 and 4303). Resulting packets are well formed, reconstructed
  * original IP packets, with IPSEC headers removed and valid header field values
  * restored. The amount and content of packet data before the IP header is
- * undefined.
+ * undefined. Some amount of TFC padding may follow the IP packet payload,
+ * in which case packet length is larger than protocol headers indicate.
+ * In tunnel mode, a TFC dummy packet have both IPv4 and IPv6 flags cleared, and
+ * the packet may contain some (dummy) data. Furthermore, inline IPSEC
+ * processing may drop dummy packets.
```
lumag: It seems @bala-manoharan and me would still prefer separate is_dummy flag. We don't even have to limit it to TFC dummy packets, but might use it later for all kinds of dummy packets.

Bill-Fischofer-Linaro: @lumag Would exposing he next header byte also be acceptable? That would similarly seem to be future-proof and would also avoid having to introduce an `odp_packet_has_dummy()` predicate function.

lumag: In fact I'm more biased towards `odp_packet_get_l3_proto()`/`odp_packet_get_l4_proto()` functions.

psavol:
v2: protocol types
v3: typo correction ROHC

include/odp/api/spec/ipsec.h
```diff
@@ -1226,12 +1226,23 @@ typedef struct odp_ipsec_status_t {
  * e.g. RFC 4302 and 4303). Resulting packets are well formed, reconstructed
  * original IP packets, with IPSEC headers removed and valid header field values
  * restored. The amount and content of packet data before the IP header is
- * undefined.
+ * undefined. Some amount of TFC padding may follow the IP packet payload,
+ * in which case packet length is larger than protocol headers indicate.
+ * TFC dummy packets have l3_type set to ODP_PROTO_L3_TYPE_NONE in tunnel mode
```
lumag: It also should be NO_NEXT, shouldn't it?

include/odp/api/spec/ipsec.h
```diff
@@ -1226,12 +1226,23 @@ typedef struct odp_ipsec_status_t {
...
+ * inline IPSEC processing may drop dummy packets.
  *
  * Each successfully transformed packet has a valid value for these metadata
  * regardless of the inner packet parse configuration
  * (odp_ipsec_inbound_config_t):
- * - L3 offset: Offset to the first byte of the (outmost) IP header
+ * - l3_offset: Offset to the first byte of the original IP packet. The value
+ *              is implementation specific for tunnel mode TFC dummy packets.
+ * - l3_type:   Specifies if the original packet is IPv4 or IPv6. For tunnel
+ *              mode TFC dummy packets set to ODP_PROTO_L3_TYPE_NONE.
```
lumag: And here.

lumag:
Looks good! I don't like that we now have third enumeration of protocols. Maybe we can later switch helper and internals `_ODP_IPPROTO` to these new constants?

Bill-Fischofer-Linaro:
Sounds like everyone is happy with this version of the API, so we should be good to go with it.

include/odp/api/spec/ipsec.h
psavol: L3 types are basically what you find from L2 type field: ARP, RARP, IPV4, IPV6. There's no "no next header" type in e.g. ethernet types. So, NO_NEXT is defined only on L4 types. But that's OK since, for tunnel mode application just checks agaist l3_type == IPV4, IPV6 or NONE, where NONE == dummy. In transport mode, application just checks l4_type == NO_NEXT for dummy. Application L3 type in transport should always have either IPV4 or IPV6.

include/odp/api/spec/ipsec.h
psavol: L3 NONE is possible only in tunnel mode dummy case. In tunnel case, L4 type is set to NONE by default, only if application has not requested extra parsing L4 type is set to e.g. TCP/UDP.

In transport mode, both L4 is set to NONE or NO_NEXT by default, and extra parsing is needed for other values.

muvarov:
Merged.

PR #405, opened 2018-01-18
[PATCH v2] fix: cls: invalid CoS index

bala-manoharan:
Fixes: https://bugs.linaro.org/show_bug.cgi?id=3576

### DIFF
--- a/platform/linux-generic/odp_classification.c
+++ b/platform/linux-generic/odp_classification.c
@@ -212,6 +212,12 @@ static void _odp_cls_update_hash_proto(cos_t *cos,
 		cos->s.hash_proto.udp = 1;
 }
 
+static inline void _cls_queue_unwind(uint32_t tbl_index, uint32_t j)
+{
+	while (j > 0)
+		odp_queue_destroy(queue_grp_tbl->s.queue[tbl_index + --j]);
+}
+
 odp_cos_t odp_cls_cos_create(const char *name, odp_cls_cos_param_t *param)
 {
 	int i, j;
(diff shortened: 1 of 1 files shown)


platform/linux-generic/odp_classification.c
```diff
@@ -250,7 +250,7 @@ odp_cos_t odp_cls_cos_create(const char *name, odp_cls_cos_param_t *param)
 				cos->s.num_queue = param->num_queue;
 				_odp_cls_update_hash_proto(cos,
 							   param->hash_proto);
-				tbl_index = cos->s.index * CLS_COS_QUEUE_MAX;
+				tbl_index = i * CLS_COS_QUEUE_MAX;
```
Bill-Fischofer-Linaro: Since we're bug fixing here, if the `odp_queue_create()` call below fails, it should unwind the previous successful creates before returning `ODP_COS_INVALID` to avoid leaking queues:
```
for (j = 0; j < CLS_COS_QUEUE_MAX; j++) {
        queue = odp_queue_create(NULL, &cos->s.queue_param);
        if (queue == ODP_QUEUE_INVALID) {
                while (j > 0) 
                    odp_queue_destroy(queue_grp_tbl->s.queue[tbl_index + j--]);
                UNLOCK(&cos->s.lock);
                return ODP_COS_INVALID;
        }
        queue_grp_tbl->s.queue[tbl_index + j] = queue;
}
```
This could be a separate commit or PR as you prefer.

Bill-Fischofer-Linaro:
@muvarov Not clear why checkpatch is failing on this PR. It looks like it's failing to launch properly. I've forced a restart of the test but it's still getting the same result.

muvarov:
@Bill-Fischofer-Linaro  looks like github but to deliver right commit sha to validate. Top level checkpatch check is passed. So it has to be ok.

muvarov:
Merged.

PR #412, opened 2018-01-20
[PATCH v4] Several Travis tweaks

lumag:
- fix DPDK path passing to configure script
- speedup building
- stop early if due to networking errors installation of some packages was impossible


muvarov:
Merged.

PR #416, opened 2018-01-22
[PATCH v4] Build and test ODP with shared DPDK

lumag:
Another PR dedicated to distro support. Build and test ODP with shared version of DPDK.

### DIFF
--- a/.travis.yml
+++ b/.travis.yml
@@ -14,28 +14,6 @@ group: deprecated-2017Q2
 stages:
   - "build only"
   - test
-env:
-  global:
-    # COVERITY_SCAN_TOKEN
-    # ** specific to your project **
-    # Note:
-    # You should have a github account and travis linked travis account.
-    # The secure key to be filled below is the 685 character long encrypted
-    # token you can find as follow from your coverity dashboard
-    # (at https://scan.coverity.com/dashboard):
-    # Click on the github project (<you>/odp)
-    # Click on "submit build"
-    # Click on "Configure Travis CI"
-    # Look at the COVERITY_SCAN_TOKEN in the env: global: section
-    # of the configuration example.
-    # copy the secure:<key> below
-    #
-    - secure: "xxxx"
-    #
-    # By default Linaro CODECOV_TOKEN token is used. It's ok to use it to see
-    # for individual commit validation. But you you want to track tests history
-    # you need generated new one at https://codecov.io specific for your repo.
-    - CODECOV_TOKEN=a733c34c-5f5c-4ff1-af4b-e9f5edb1ab5e
 
 addons:
         apt:
@@ -53,10 +31,6 @@ addons:
 #                        build_command:   "make"
 #                        branch_pattern: coverity_scan
 
-compiler:
-        - gcc
-        - clang-3.8
-
 cache:
         ccache: true
         pip: true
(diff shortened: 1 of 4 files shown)


Bill-Fischofer-Linaro:
@lumag Looks like recent merges now require this PR to be rebased.

lumag:
@Bill-Fischofer-Linaro done

.travis.yml
```diff
@@ -198,7 +199,14 @@ install:
...
         - |
           if [ -n "$TARGET" ] ; then
-           if [ ! -f "dpdk/${TARGET}/lib/libdpdk.a" ] ; then
+           # force DPDK rebuild if it does not match shared/non-shared requirement
+           if [ -n "$DPDK_SHARED" -a ! -f "dpdk/${TARGET}/lib/libdpdk.so" ] ; then
+            rm -rf dpdk/"$TARGET"
+           fi
+           if [ -z "$DPDK_SHARED" -a ! -f "dpdk/${TARGET}/lib/libdpdk.a" ] ; then
+            rm -rf dpdk/"$TARGET"
+           fi
```
muvarov: maybe simple  
`if [  -a ! -f "dpdk/${TARGET}/lib/libdpdk.a" ] ; then`
instead of 2 checks?

muvarov: In general that is not right solution. I think we need to cache 2 dpdk builds. Shared and non shared and force dpdk recompilation only on version change.

.travis.yml
```diff
@@ -258,11 +271,17 @@ script:
...
         - if [ -z "$CROSS_ARCH" ] ; then
-          LD_LIBRARY_PATH="${HOME}/odp-install/lib:$LD_LIBRARY_PATH" ./odp_hello_inst_dynamic ;
-          ./odp_hello_inst_static ;
+            LD_LIBRARY_PATH="${HOME}/odp-install/lib:$LD_LIBRARY_PATH" ./odp_hello_inst_dynamic ;
+          fi
+        - echo "Static link.."
+        - |
+          # it is not possible to do static linking if we only have shared DPDK library. Compiler complains about missing -ldpdk
+          if [ -z "$TARGET" -o -z "$DPDK_SHARED" ] ; then
+            ${CC} ${CFLAGS} ${OLDPWD}/example/hello/odp_hello.c -o odp_hello_inst_static `PKG_CONFIG_PATH=${HOME}/odp-install/lib/pkgconfig:${PKG_CONFIG_PATH} pkg-config --cflags --libs libodp-linux --static` -static || exit 1
```
muvarov: why exit 1 is needed? Travis should stop on error return code on the command. exit 1 should not be needed.

.travis.yml
```diff
@@ -258,11 +271,17 @@ script:
...
+            if [ -z "$CROSS_ARCH" ] ; then
+            ./odp_hello_inst_static || exit 1;
```
muvarov: same here.

muvarov:
"build: another fix for linking with shared libdpdk"  is better to rename to "build: odp_dpdk.m4: another fix for linking with shared libdpdk". It will be good to add some readable description what was a problem with previous version.

.travis.yml
lumag: Hmm. I can try doing that.

.travis.yml
lumag: exit 1 is needed, to actually return the error code out of if.

lumag: Hmm. Maybe it's not that necessary.

.travis.yml
lumag: Generally first `exit 1` is necessary, otherwise `if` will return the status of last command, not the first failed one.

lumag: Try `if true ; then false; if false ; then true ; fi; fi; echo $?`

lumag:
@muvarov updated PR to follow your suggestions.

muvarov:
Merged.

PR #419, opened 2018-01-23
[PATCH v2] linux-gen: test: update gitignore

psavol:
Add missing .env file into gitignore.


muvarov:
env is set is here:

./platform/linux-generic/test/Makefile:TESTENV = tests-$(TESTNAME).env

I think we need to add *.env on top directory:
./platform/linux-generic/test/.gitignore

Also ./test/validation define this env. So I'm thinking that most right place for it is project top directory.

muvarov:
Merged.

PR #422, opened 2018-01-24
[PATCH API-NEXT v2] crypto: support SHA384 and AES-XCBC algorithm in ipsec-odp

SonicwallYhe:
As there're some products support more authentication method in IPSec VPN, I add SHA384 and AES-XCBC as the IPSec Phase 2 authentication algorithm into odp. SHA384 is similar as SHA256 and SHA512, the difference is just the key length. AES-XCBC reference to RFC 3566 "The AES-XCBC-MAC-96 Algorithm and Its Use With IPsec". I tested with SonicWALL product, and both algorithm works well.


lumag:
@SonicwallYhe 
- you need to add respective bits to capabilities
- could you please split into separate API and implementation commits
- please add testsuite for these algorithms
- using AES_* OpenSSL interface is not recommended, please switch to EVP_* interface

SonicwallYhe:
@lumag 
Thanks for your comments.
I will submit another commit according to your comments later.

For the EVP_* interfaces, as I know, for AES algorithm, it has several modes such as ecb, cbc, ofb, cfb, gcm. But I cannot find a related mode for AES-XCBC.  If you could find one, please kindly tell me and it will give me great help.

lumag:
@SonicwallYhe There is a BSD-licensed [XCBC implementation](https://github.com/grivet/XCBC) using EVP_aes128_cbc(). It follows Init/Update/Final semantics, so it should be easy to adapt. Please note, that your code will not work for multi-segment packets. You can use HMAC-related code as boilerplate.

include/odp/api/spec/crypto.h
```diff
@@ -122,12 +122,26 @@ typedef enum {
...
 	 */
 	ODP_AUTH_ALG_SHA512_HMAC,
 
+	/** AES-XCBC
+	 *
+	 *  AES-XCBC algorithm, 
+	 *	reference to RFC 3566 
+	 *  The AES-XCBC-MAC-96 Algorithm and Its Use With IPsec
+	 */	
+	ODP_AUTH_ALG_AES_XCBC,
```
psavol: I'd use AES-XCBC-MAC as the algorithm name and thus enumeration would be ODP_AUTH_ALG_AES_XCBC_MAC. 

Remove reference to IPSEC RFC from here, since this is a crypto API definition, which is more generic than IPSEC use case. Digest size (96) does not need to be mentioned here, as application gets it from capacity call.

In addition to this enum, odp_crypto_auth_algos_t bit field needs to be updated with .aes_xcbc_mac.

include/odp/api/spec/crypto.h
```diff
@@ -122,12 +122,26 @@ typedef enum {
 	 */
 	ODP_AUTH_ALG_SHA256_HMAC,
 
+	 /** HMAC-SHA-384
+	 *
+	 *  SHA-384 algorithm in HMAC mode
+	 */
+	ODP_AUTH_ALG_SHA384_HMAC,
```
psavol: In addition to this enum, odp_crypto_auth_algos_t bit field needs to be updated with .sha384_hmac.

Bill-Fischofer-Linaro:
PR has Travis issues that need to be resolved.

lumag:
@SonicwallYhe it still does not compile:
```
odp_crypto_test_inp.c:1104:13: error: ‘crypto_test_gen_alg_hmac_sha384’ defined but not used [-Werror=unused-function]
 static void crypto_test_gen_alg_hmac_sha384(void)
             ^
odp_crypto_test_inp.c:1161:13: error: ‘crypto_test_gen_alg_aes_xcbc’ defined but not used [-Werror=unused-function]
 static void crypto_test_gen_alg_aes_xcbc(void)
```

Also you have to apply `git reset`/`git add -i`/`git commit` magic to split your two commits into logical pieces:
 - "api: crypto: add ..."
   This patch should change only `include/odp/api/spec/crypto.h` adding constants and bit fields
- "validation: crypto: ..."
   This one should change `tests/validation/api/crypto` adding new tests
- "linux-gen: crypto: ..."
   This one should add linux-generic implementation of your algorithms

It seems you add support for those algorightms into IPsec code. It also has to go in a separate commit.

platform/linux-generic/odp_crypto.c
```diff
@@ -30,6 +30,9 @@
 
 #define MAX_SESSIONS 32
 
+#define AES_BLOCK_SIZE 16
+typedef uint32_t aes_block[4];
```
lumag: Just use uint8_t pointers/arrays instead of adding aes_block.

platform/linux-generic/odp_crypto.c
```diff
@@ -236,6 +247,111 @@ void packet_hmac(odp_packet_t pkt,
...
+	res[3] ^= op[3];
+}
+
+static
+odp_crypto_alg_err_t aesxcbc_gen(odp_packet_t pkt,
+			      const odp_crypto_packet_op_param_t *param,
+			      odp_crypto_generic_session_t *session)
+{
+	aes_block e = {0, 0, 0, 0};
+	uint8_t *data  = odp_packet_data(pkt);
```
lumag: This will not work for segmented packets. In ODP a packet is not always a contiguous memory array. Consider how the rest of crypto code deals with that.

lumag:
@SonicwallYhe could you please rearrange commits and make fix tests compilation issues, so that we can merge API definition and tests. We can work on fixing implementation issues afterwards.

bala-manoharan:
@SonicwallYhe  It might be better to split the patch into three logical patches
"api: crypto: ..."
"linux-gen: crypto: ..."
"vald: crypto: ..."

The "api" patch could be a separate PR request so that we could merge the enum support earlier before API freeze and we can continue working on the implementation and validation as a separate PR .

muvarov:
yes. it has to be 3 patches. please follow previous git commit style. Patch which changes api has to be separate patch and named as "api: ipsec: .....",  patch to linux-generic implementation "linux-gen: ipsec: ..." and the same patch for "test: ...."

lumag:
@SonicwallYhe please rebase on top of current api-next branch and split patches to testsuite/implementation/ipsec

lumag:
@muvarov could you please close this PR, it's  replaced by #470 .

PR #426, opened 2018-01-24
[PATCH v1] linux-gen: ipsec: fix block len for AES-CTR mode

lumag:
AES-CTR is stream cipher, so block length should be 1, rather than 16.


muvarov:
Merged.

PR #427, opened 2018-01-24
[PATCH v2] Fix several issues in linux-generic IPsec SAD implementation

lumag:


lumag:
@JannePeltonen this should fix those two issues you have pointed out and two more issues spotted when fixing surrounding code.

lumag:
@JannePeltonen could you please review.

JannePeltonen:
This patch set looks good to me too.

When reviewing the patches, I noticed that there is another SA leak in odp_ipsec_sa_create() in the cipher algorithm check for AES_GCM, similar to the leak fixed in this patch set.

JannePeltonen:
I meant AES_GMAC, i.e.  this place:
        case ODP_AUTH_ALG_AES_GMAC:
                if (ODP_CIPHER_ALG_NULL != crypto_param.cipher_alg)
                        return ODP_IPSEC_SA_INVALID;

JannePeltonen:
This patch set looks good to me also now after the additional SA leak fix.

muvarov:
Merged.

PR #428, opened 2018-01-24
[PATCH API-NEXT v5] Add ChaCha20-Poly1305 support

lumag:
RFC 7634 defines ChaCha20-Poly1305 algorithm for IPsec. RFC 8221 marks it at level 'SHOULD':

```
   ENCR_CHACHA20_POLY1305 was not ready to be considered at the time of
   [RFC7321].  It has been recommended by the Crypto Forum Research
   Group (CFRG) and others as an alternative to AES-CBC and AES-GCM.  At
   the time of writing, there are not enough ESP implementations of
   ENCR_CHACHA20_POLY1305 to be able to introduce it at the SHOULD+
   level.  Its status has been set to SHOULD and is expected to become
   MUST in the long term.
```


platform/linux-generic/odp_crypto.c
```diff
@@ -28,6 +28,12 @@
 #include <openssl/hmac.h>
 #include <openssl/evp.h>
 
+#if (OPENSSL_VERSION_NUMBER >= 0x10100000L) && !defined(OPENSSL_NO_POLY1305)
+#define ODP_HAVE_CHACHA20_POLY1305 1
+#else
+#define ODP_HAVE_CHACHA20_POLY1305 0
```
muvarov: _ODP_HAVE_

test/validation/api/ipsec/test_vectors.h
```diff
@@ -1730,6 +1735,67 @@ static const ipsec_test_packet pkt_mcgrew_gcm_test_15_esp = {
...
+		0x08, 0x09, 0x0a, 0x0b, 0x0c, 0x0d, 0x0e, 0x0f,
+		0x10, 0x11, 0x12, 0x13, 0x14, 0x15, 0x16, 0x17,
+		0x18, 0x19, 0x1a, 0x1b, 0x1c, 0x1d, 0x1e, 0x1f,
+		0x20, 0x21, 0x22, 0x23, 0x24, 0x25, 0x26, 0x27,
+		0x28, 0x29, 0x2a, 0x2b, 0x2c, 0x2d, 0x2e, 0x2f,
+		0x30, 0x31, 0x32, 0x33, 0x34, 0x35, 0x36, 0x37,
+	},
+};
+
+static const ODP_UNUSED ipsec_test_packet pkt_ipv4_rfc7634_esp = {
```
muvarov: do you need ODP_UNUSED here? structure has reference in the code.

muvarov:
It will be great if somebody can review validation tests for it. Also I see that we need be very careful with rebases.

psavol:
I'd suggest to merge the first patch right away. It does not need to wait impl/test changes to get approved. Couple of other patches touch same lines in API. We can resolve those conflicts in the mean time and are closer to API freeze.

test/validation/api/ipsec/test_vectors.h
lumag: Yes, but not in every code that includes this header file. See other test vectors there, which also have such this attribute.

lumag:
@muvarov test cases come from RFC 7539. I have added comment to them.
Also fixed the guarding define.

lumag:
Let's merge #430 first.

lumag:
Replaced by #434

PR #430, opened 2018-01-25
[PATCH API-NEXT v1] api: crypto: add AES CCM enumerations

psavol:


lumag:
Second patch:

include/odp/api/spec/crypto.h
```diff
@@ -128,23 +131,29 @@ typedef enum {
...
-	 * authenticated as AAD. In constrast to that, ODP API specifies the
-	 * bulk of authenticated data to be located in packet payload for all
-	 * authentication algorithms, including GMAC. Thus for GMAC application
-	 * should also pass all data to be authenticated as packet data. AAD is
-	 * not used for GMAC. GMAC IV should be passed via session IV or
-	 * per-packet IV override.
+	 *  AES Galois Message Authentication Code (GMAC) algorithm. AES-GMAC
+	 *  is based on AES-GCM operation, but provides authentication only.
+	 *  Hence this algorithm can be paired only with ODP_CIPHER_ALG_NULL
+	 *  cipher.
```
lumag: Initially I wrote that it must be used only with ENC_NULL so that there will no conflict over IV. Now nothing stops one from using any cipher with AUTH_GMAC.

lumag:
@psavol will you provide an implementation of CCM?

include/odp/api/spec/crypto.h
psavol: Since GMAC is GCM without cipher, user which needs also ciphering (with the same auth properties as GMAC offers) would naturally choose GCM. This limitation may help some implementations to be more efficient in resource usage. Is there application need for e.g. 3DES + GMAC configuration?

psavol:
No just API enumerations.

lumag:
@muvarov let's merge this first, then I will update/replace #428 to reflect header changes and to provide AES-CCM implementation.

muvarov:
Merged.

PR #431, opened 2018-01-25
[PATCH CATERPILLAR v2] shippable: fix 'make check' CI failures

MykytaI:

### DIFF
--- a/platform/linux-generic/Makefile.am
+++ b/platform/linux-generic/Makefile.am
@@ -248,6 +248,7 @@ __LIB__libodp_linux_la_SOURCES = \
 			   pktio/common.c \
 			   pktio/dpdk.c \
 			   pktio/ethtool.c \
+			   pktio/subsystem.c \
 			   pktio/ipc.c \
 			   pktio/loopback.c \
 			   pktio/mdev.c \
@@ -257,7 +258,6 @@ __LIB__libodp_linux_la_SOURCES = \
 			   pktio/ring.c \
 			   pktio/socket.c \
 			   pktio/socket_mmap.c \
-			   pktio/subsystem.c \
 			   pktio/sysfs.c \
 			   pktio/tap.c \
 			   pool/generic.c \


muvarov:
After some thinking there might be several ways to fix it. 1) somewhere in docker run increase shared memory size with --shm-size argument to docker run. 2) use ODP_SHM_DIR like here. 3) mount hp more correctly inside docker. increase privileges or use host hugetlsbs.

I need to think which solution is more clean. This patch might be good as a quick fix. If we need to test hp on arm then we need step 3.

This patch has to be suitable for master branch.

MykytaI:
Please don't merge this patch yet, further investigation is needed.

When I start docker image with the same options as Shippable does -- I can't reproduce issue by manually checking out, compiling and running 'make check'.

Note on hugetlbfs: looks like '--privileged=true' docker option takes care of permissions.

MykytaI:
After git bisect session and patch reduction it's clear now that
Shippable regression is caused by link order change in
ea7b4281992e3cf772fb84708ed7f3985bce78e7
linux-gen: mediated devices common code

Specifically, moving about 'subsystem.c' causes 'make check' to fail.
I've updated PR and hopefully now Shippable check will pass. I have no
idea so far what causes failure, so maybe we need to investigate
further and find the root cause.

It's not in my engineering habits to do fixes like that without
figuring out the real root cause, but it can turn out that time
invested for further investigation isn't worth the end result. I guess
we need to decide if I need to continue with that or not. Possible
causes (from my experience):
 * Due to link order change __attribute__ constructor can misbehave
 * Known/unknown clang bug
 * Memory corruption which was hidden before due to different
.data/.bss/.text layout

MykytaI:
My recommendation would be to merge this PR as is as soon as possible so that other PRs going into Caterpillar branch can get proper Shippable CI.

If further investigation to figure out root cause is needed -- it can be done separately.

heyi-linaro:
Yes, once this is approved by @Bill-Fischofer-Linaro @lumag I'll merge it.

lumag:
Well, it should be named 'linux-gen: foo bar', rather than 'shippable' with rationality given in commit message.

heyi-linaro:
Merged and closed, accommodated Dmitry's comment by changing the commit message to:

linux-gen: fix shippable 'make check' CI failures

PR #433, opened 2018-01-26
[PATCH v1] Inline thread and cpu id

psavol:
Inline thread and cpu id functions as those may be used frequently in application. E.g. packet rate of OFP packet forward test (fpm_burst) improves 3% (single thread).


Bill-Fischofer-Linaro:
@muvarov is the "required statuses" check new? If yes, we need to exclude checkpatch from the requirement as it still generates false positives.

muvarov:
Merged.

PR #439, opened 2018-01-31
[PATCH v5] travis: build DPDK with clang if it was selected

lumag:
gcc-built DPDK is not fully compatible with clang (see
https://travis-ci.org/lumag/odp/jobs/335324053 for example). Use clang
to compile DPDK if selected compiler is clang.

### DIFF
--- a/.travis.yml
+++ b/.travis.yml
@@ -101,6 +101,7 @@ before_install:
                                 DPDK_CFLAGS="-m32" ;
                         else
                                 DPDK_CROSS="$CROSS_GNU_TYPE-" ;
+                                DPDK_CFLAGS="--target=$CROSS_GNU_TYPE" ;
                         fi
                 fi ;
                 export CXX="${CC/clang/clang++}";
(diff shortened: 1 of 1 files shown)


.travis.yml
```diff
@@ -158,33 +158,40 @@ install:
           if [ "${CACHED_DPDK_VERS}" != "${DPDK_VERS}" ]; then
             rm -rf dpdk
           fi
+        - if [ "${CC#clang}" != "${CC}" ] ;
+          then
+            DPDKCC=clang ;
```
muvarov: why do we need this and not just $CC ?

.travis.yml
```diff
@@ -193,6 +200,7 @@ install:
             sed -i -e 's/40900/40800/g' lib/librte_eal/common/include/arch/arm/rte_vect.h
             sed -i -e 's/!(/!(defined(__arm__) \&\& defined(__clang__) || /g' lib/librte_eal/common/include/arch/arm/rte_byteorder.h
             sed -i -e 's/__GNUC__/defined(__arm__) \&\& defined(__clang__) || __GNUC__/' lib/librte_eal/common/include/generic/rte_byteorder.h
+            sed -i -e 's,\$(CC),\0 $(EXTRA_CFLAGS),g' lib/librte_acl/Makefile
```
muvarov: should this go to dpdk instead of odp?

.travis.yml
lumag: It is used to:
 - set DPDK target directory (`...-clang` vs `...-gcc`)
 - set DPDK configuration items based on the toolchain selected

.travis.yml
lumag: I know nothing about contributing to DPDK. Also what is the chance of fixing old DPDK releases (which we use)?

muvarov: if we push this to dpdk then we will not need to fix it with new releases.

muvarov:
Merged.

PR #441, opened 2018-01-31
[PATCH v13] Shippable: fix tests statistics

lumag:


muvarov:
gcovr: error: no such option: --exclude-dir

lumag:
It seems that enabling code coverage with clang causes build timeouts. I'll update test to enable coverate only for gcc.

muvarov:
where coverage results uploaded and how to see them?

lumag:
@muvarov see screenshot
stats2

muvarov:
@lumag in coverage I see only stats for each file. But I can not click to file and look to functions. Is that expected?

lumag:
@muvarov I don't know, I never used Shippable before. I also see stats for individual files, but no details

lumag:
@muvarov are there any remaining comments/issues with this PR?

muvarov:
@lumag  I'm not sure if we need stats if we can not look into details. It just increase test functionality which we will have to maintain. I would keep coverage tests in Travis until there will be any issues.

lumag:
@muvarov dropped code coverage part

muvarov:
Merged.

PR #442, opened 2018-02-01
[PATCH CATERPILLAR v1] linux-gen: add cxgbe RX slot mode support

MykytaI:
To use RX slot_mode one needs to redefine CXGB4_RX_BUF_SIZE_INDEX=2 either
in cxgbe4.c locally or via CFLAGS in ./configure

One needs supplementary patch on kernel driver applied as well as load
kernel driver with rx_mode=1 parameter.

### DIFF
--- a/platform/linux-generic/pktio/mdev/cxgb4.c
+++ b/platform/linux-generic/pktio/mdev/cxgb4.c
@@ -36,6 +36,25 @@
 /* RX queue definitions */
 #define CXGB4_RX_QUEUE_NUM_MAX 32
 
+/* RX buffer size indices for firmware */
+#define CXGB4_RX_BUF_SIZE_INDEX_4K	0x0UL
+#define CXGB4_RX_BUF_SIZE_INDEX_64K	0x1UL
+#define CXGB4_RX_BUF_SIZE_INDEX_1500	0x2UL
+#define CXGB4_RX_BUF_SIZE_INDEX_9000	0x3UL
+
+/* Default RX buffer size index to use */
+#ifndef CXGB4_RX_BUF_SIZE_INDEX
+#define CXGB4_RX_BUF_SIZE_INDEX CXGB4_RX_BUF_SIZE_INDEX_4K
+#endif /* CXGB4_RX_BUF_SIZE_INDEX */
+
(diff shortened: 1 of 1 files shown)


platform/linux-generic/pktio/mdev/cxgb4.c
```diff
@@ -36,6 +36,25 @@
...
 #define CXGB4_RX_QUEUE_NUM_MAX 32
 
+/* RX buffer size indices for firmware */
+#define CXGB4_RX_BUF_SIZE_INDEX_4K	0x0UL
+#define CXGB4_RX_BUF_SIZE_INDEX_64K	0x1UL
+#define CXGB4_RX_BUF_SIZE_INDEX_1500	0x2UL
+#define CXGB4_RX_BUF_SIZE_INDEX_9000	0x3UL
+
+/* Default RX buffer size index to use */
+#ifndef CXGB4_RX_BUF_SIZE_INDEX
```
MykytaI: The question is if we need to reverse CXGB4_RX_BUF_SIZE_INDEX <-> CXGB4_RX_BUF_SIZE dependency and have CXGB4_RX_BUF_SIZE as configuration option (possible values at the moment are 2048 and 4096). Probably then it will be more intuitively understandable and we'll hide HW specific index stuff hidden.

That is:

#ifndef CXGB4_RX_BUF_SIZE
#define CXGB4_RX_BUF_SIZE 4096
#endif

CXGB4_RX_BUF_SIZE == 2048 => define CXGB4_RX_BUF_SIZE_INDEX CXGB4_RX_BUF_SIZE_INDEX_1500
CXGB4_RX_BUF_SIZE == 4096 => define CXGB4_RX_BUF_SIZE_INDEX CXGB4_RX_BUF_SIZE_INDEX_4K

apalos: I personally prefer the code as-is. Keeping a notifier for different hardware functionality will be a better indicator for someone reading the code than buffer size values.

Bill-Fischofer-Linaro:
@muvarov Not clear why the callout for checkpatch reports failure here but the checkpatch run in Travis reports success.

heyi-linaro:
With Ilias' review approve let's merge this to complete caterpillar mediated device driver works.

PR #444, opened 2018-02-01
[PATCH v2] api: init: global init param for shm size

psavol:
Total shm memory usage vary a lot between applications. Current odp-linux shm implementation fixes max shm memory size to 512 MB as it allocates a fixed size VA address space block. E.g. OFP cannot currently scale to high number of route table entries due to this built time limit.


Bill-Fischofer-Linaro:
@psavol This appears to have been based on a non-current level of master? It looks like it's pulling in a lot of commits that have already been merged, which makes it difficult to review. Please rebase on current master to have this PR be just its own changes.

PR #446, opened 2018-02-01
[PATCH API-NEXT v2] api: init: global shm memory size

psavol:
Memory usage vary a lot between applications. If implementation has to decide a hard limit at init time (as odp-linux does), it's hard to pick a value that is large enough for every application but don't reserve too much memory in average. E.g. OFP already suffers from current fixed limit of 512MB in odp-linux.

### DIFF
--- a/include/odp/api/spec/init.h
+++ b/include/odp/api/spec/init.h
@@ -107,23 +107,29 @@ typedef int (*odp_log_func_t)(odp_log_level_t level, const char *fmt, ...);
 typedef void (*odp_abort_func_t)(void) ODP_NORETURN;
 
 /**
- * ODP initialization data
+ * Global initialization parameters
  *
- * Data that is required to initialize the ODP API with the
- * application specific data such as specifying a logging callback, the log
- * level etc.
+ * These parameters may be used at global initialization time to configure and
+ * optimize ODP implementation to match the intended usage. Application
+ * specifies maximum resource usage. Implementation may round up resource
+ * reservations as needed. Initialization function returns a failure if resource
+ * requirements are too high. Init parameters may be used also to override
+ * logging and abort functions.
  *
- * @note It is expected that all unassigned members are zero
+ * Use odp_init_param_init() to initialize the parameters into their default
+ * values. Unused parameters are left to default values.
  */
 typedef struct odp_init_t {
 	/** Maximum number of worker threads the user will run concurrently.
 	    Valid range is from 0 to platform specific maximum. Set both
 	    num_worker and num_control to zero for default number of threads. */
 	int num_worker;
+
 	/** Maximum number of control threads the user will run concurrently.
 	    Valid range is from 0 to platform specific maximum. Set both
 	    num_worker and num_control to zero for default number of threads. */
 	int num_control;
+
 	/** Pointer to bit mask mapping CPUs available to this ODP instance
 	    for running worker threads.
 	    Initialize to a NULL pointer to use default CPU mapping.
@@ -139,6 +145,7 @@ typedef struct odp_init_t {
 		  worker masks
 	 */
 	const odp_cpumask_t *worker_cpus;
+
 	/** Pointer to bit mask mapping CPUs available to this ODP instance
 	    for running control threads.
 	    Initialize to a NULL pointer to use default CPU mapping.
@@ -150,10 +157,13 @@ typedef struct odp_init_t {
 	    worker and control masks do not overlap.
 	 */
 	const odp_cpumask_t *control_cpus;
+
 	/** Replacement for the default log fn */
 	odp_log_func_t log_fn;
+
 	/** Replacement for the default abort fn */
 	odp_abort_func_t abort_fn;
+
 	/** Unused features. These are hints to the ODP implementation that
 	 * the application will not use any APIs associated with these
 	 * features. Implementations may use this information to provide
@@ -161,6 +171,16 @@ typedef struct odp_init_t {
 	 * that a feature will not be used and it is used anyway.
 	 */
 	odp_feature_t not_used;
+
+	/** Shared memory parameters */
+	struct {
+		/** Maximum memory usage in bytes. This is the maximum
+		 *  amount of shared memory that application will reserve
+		 *  concurrently. Use 0 when not set. Default value is 0.
+		 */
+		uint64_t max_memory;
+	} shm;
+
 } odp_init_t;
 
 /**
@@ -189,7 +209,8 @@ void odp_init_param_init(odp_init_t *param);
  * system and outputs a handle for it. The handle is used in other calls
  * (e.g. odp_init_local()) as a reference to the instance. When user provides
  * configuration parameters, the platform may configure and optimize the
- * instance to match user requirements.
+ * instance to match user requirements. A failure is returned if requirements
+ * cannot be met.
  *
  * Configuration parameters are divided into standard and platform specific
  * parts. Standard parameters are supported by any ODP platform, where as
--- a/platform/linux-generic/_ishm.c
+++ b/platform/linux-generic/_ishm.c
@@ -258,7 +258,7 @@ static void *alloc_fragment(uintptr_t size, int block_index, intptr_t align,
 	ishm_fragment_t *rem_fragmnt;
 	uintptr_t border;/* possible start of new fragment (next alignement)  */
 	intptr_t left;	 /* room remaining after, if the segment is allocated */
-	uintptr_t remainder = ODP_CONFIG_ISHM_VA_PREALLOC_SZ;
+	uintptr_t remainder = odp_global_data.shm_max_memory;
 
 	/*
 	 * search for the best bit, i.e. search for the unallocated fragment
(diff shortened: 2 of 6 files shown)


include/odp/api/spec/init.h
```diff
@@ -150,17 +157,29 @@ typedef struct odp_init_t {
...
 	 * that a feature will not be used and it is used anyway.
 	 */
 	odp_feature_t not_used;
+
+	/** Shared memory parameters */
+	struct {
+		/** Maximum memory usage in bytes. This is the maximum
+		 *  amount of shared memory that application will reserve
+		 *  concurrently. */
+		uint64_t max_memory;
```
muvarov: 0 - means not specified. And what about continuous of memory chunks? Requesting  one big continues shared memory chunk is not general solution.

test/validation/api/shmem/shmem.c
```diff
@@ -212,6 +214,55 @@ void shmem_test_basic(void)
...
+	CU_ASSERT_FATAL(odp_shm_capability(&capa) == 0);
+
+	CU_ASSERT(capa.max_blocks > 0);
+
+	size  = capa.max_size;
+	align = capa.max_align;
+
+	/* Assuming that system has at least MAX_MEMORY_USED bytes available */
+	if (capa.max_size == 0 || capa.max_size > MAX_MEMORY_USED)
+		size = MAX_MEMORY_USED;
```
Bill-Fischofer-Linaro: Why do you want to limit the size in a test that named `shmem_test_max_reserve()`? If `capa.max_size == 0` then you have to pick a specific target, but if it's non-zero why wouldn't you want to try to reserve that much to see if the limit is true?

test/validation/api/shmem/shmem.c
```diff
@@ -212,6 +214,55 @@ void shmem_test_basic(void)
...
+
+	if (capa.max_align == 0 || capa.max_align > MEGA)
+		align = MEGA;
```
Bill-Fischofer-Linaro: Shouldn't `MEGA` be a power of 2 for alignment purposes? I.e., 1024 x 1024 rather than 1000 x 1000? And if the implementation supports an even higher `max_align` why not test that as well?

test/validation/api/shmem/shmem.c
```diff
@@ -212,6 +214,55 @@ void shmem_test_basic(void)
...
+	CU_ASSERT(shm != ODP_SHM_INVALID);
+
+	data = odp_shm_addr(shm);
+	CU_ASSERT(data != NULL);
+
+	if (data) {
+		memset(data, 0xde, size);
+		for (i = 0; i < size; i++) {
+			if (data[i] != 0xde) {
+				CU_FAIL("Data error");
```
Bill-Fischofer-Linaro: An additional `printf()` giving a bit more detail (i.e., `i` value) would be useful here.

include/odp/api/spec/init.h
Bill-Fischofer-Linaro: @muvarov `odp_shm_capability()` already tells the application the largest contiguous size it can reserve (`max_size`) and the maximum number of reserves it can do (`max_blocks`). This is just hinting to the implementation the total size of all reserves the application will do.

test/validation/api/shmem/shmem.c
psavol: Since actual amount of available memory typically depends on system load. SHM implementation may not have a limit (max_size==0), or limit may be due to address space (e.g. 40bit == 1TB). System might not have always the max amount (e.g. 1TB) available. I limit validation test to assume that at least 100MB should be always available.

test/validation/api/shmem/shmem.c
psavol: Very large align could result very large allocation and thus again system run out of memory (e.g. 1TB align => >1TB alloc).

OK. I'll change align max to be a power of two.

test/validation/api/shmem/shmem.c
psavol: OK

include/odp/api/spec/init.h
psavol: I'll add comment about. Although, I already changed documentation to require param_init() call and say that don't change values that you are not going to use (init sets it to zero).

test/validation/api/shmem/shmem.c
Bill-Fischofer-Linaro: If the implementation doesn't have a specific predefined upper limit then it should return `capa.max_size == 0`. If it says it has a non-zero upper limit then if it's unable to provide that limit that's a failure. Otherwise what's the point of having a specified limit?

test/validation/api/shmem/shmem.c
Bill-Fischofer-Linaro: Same story as for `capa.max_size`. I'd expect most implementations to return `capa.max_align` to be either 0 or some reasonable value like 4K or 1M. However, if they specify something else then they should be able to deliver that.

test/validation/api/shmem/shmem.c
psavol: Implementation may have a limitation of e.g. 4GB due to limit of using a 32 bit address space reservation, etc. It would be waste to reserve 4GB of system memory for every ODP instance, as  implementation could not guarantee 4GB otherwise, as other applications allocate memory as well. So, init  phase there could be 4.2GB available, but by the time ODP application starts calling shm_reserve() there would be less than 4GB left and some reserves would fail.

So, implementation may have large upper limit, which is not related to amount of available memory but e.g. due implementation of the address mapping (number of bits, hugepages, etc).

test/validation/api/shmem/shmem.c
psavol: Maybe with 1GB hugepages the max align is 1GB, and 16GB hugepages 16GB, ...

test/validation/api/shmem/shmem.c
Bill-Fischofer-Linaro: The 0 value is intended to say that addressability / available memory is the only limit, so again in this case the implementation should return 0, not 4GB. If the implementation says 0 and the application tries to reserve something huge and that fails that's OK. The application needs to check RCs in any event. But what's the point of having a non-zero limit if there's no reasonable expectation that it means anything? At that point it's useless and might as well be ignored.

test/validation/api/shmem/shmem.c
Bill-Fischofer-Linaro: Again, what does the application gain from having a non-zero `max_align` returned if it's unable to make use of it?

test/validation/api/shmem/shmem.c
psavol: In the example, system is 64 bit, has e.g. 32 GB of memory, ODP SHM implementation is limited e.g. by address space pre-reservation etc (e.g. 32 bits / 4GB). So, ODP implementation limits max SHM size to 4GB, 0 cannot be used as capability. A 4 GB reservation succeeds as long as system has memory free. When other applications or ODP instances have reserved all 32GB memory, yet another 4GB reservation will fail.

So, 0 says that there's no other limit than amount of currently free memory. E.g. odp-linux implementation lied by returning 0 and at the same time limiting max reservation to 512MB (due to fixed size address space allocation).

test/validation/api/shmem/shmem.c
psavol: If max size is 4GB, max align is 1GB. A resevation may consume e.g. 4.5GB memory. Again it depends on other applications and ODP instances if such amount of system memory is still free when application tries to reserve it. Maybe first couple instances succeeded, but then system did run out of memory.

psavol:
Ping. Ready to be merged. Second last API change.

platform/linux-generic/_ishm.c
```diff
@@ -1436,15 +1436,23 @@ int _odp_ishm_cleanup_files(const char *dirpath)
...
+int _odp_ishm_init_global(const odp_init_t *init)
 {
 	void *addr;
 	void *spce_addr;
 	int i;
 	uid_t uid;
 	char *hp_dir = odp_global_data.hugepage_info.default_huge_page_dir;
 	uint64_t align;
+	uint64_t max_memory = ODP_CONFIG_ISHM_VA_PREALLOC_SZ;
+	uint64_t internal   = ODP_CONFIG_ISHM_VA_PREALLOC_SZ / 8;
```
muvarov: 'internal' has to be also adjusted. Or you can get overflow at line 1455.

psavol: if (init && init->shm.max_memory)
		max_memory = init->shm.max_memory + internal;

line 1455 does not underflow even with application requested init->shm.max_memory value:
shm_max_size = init->shm.max_memory + internal - internal;

psavol: Internal is for ODP implementation internal allocations. E.g. if application passes 1MB as init->shm.max_memory, we need to reserve much more as ODP internally needs SHM for various tables, etc. Application's max_memory just says how much it will itself request.

muvarov:
Merged.

PR #449, opened 2018-02-02
[PATCH v6] Symbol visibility handling for libodp-linux

lumag:
First patch is required, so that libodp-linux does not export unnecessary symbols linked in through DPDK.
Second patch (to be discussed) adds version script, binding symbols to library version (for ABI-compat case only).


platform/libodp-linux.map
```diff
@@ -0,0 +1,600 @@
...
+	odp_tm_vlan_marking;
+	odp_tm_wred_create;
+	odp_tm_wred_destroy;
+	odp_tm_wred_lookup;
+	odp_tm_wred_params_init;
+	odp_tm_wred_params_read;
+	odp_tm_wred_params_update;
+	odp_version_api_str;
+	odp_version_impl_name;
+	odp_version_impl_str;
```
Bill-Fischofer-Linaro: Must these be enumerated individually? The ODP API symbols should all match the pattern `odp_*`.

lumag: Yes. However we must be sure to export correct symbols with correct versions.

Bill-Fischofer-Linaro: Not sure I follow. The concern is that maintaining this list would impose a burden as every new API defined (or removed) would have to remember to update this list.

lumag: @Bill-Fischofer-Linaro is that really a "burden"? First commit does exactly that, however it makes impossible to stop library from exporting extra `_odp_*` symbols (which makes ABI definition loose).  Having such list would allow us to have proper symbol versioning, if we decide to have stable ABI (and we are almost there, judging from library ABI tracking).

muvarov: will Travis capture error if some function is missed in this list?

Bill-Fischofer-Linaro: It's a burden because it requires a separate list to be maintained apart from the main API definitions contained in the `.h` files, as well as the associated files lists we maintain in the various `Makefile.am` files.  Whenever you have to keep multiple lists in sync that's always an error-prone process unless these are auto-generated by scripts or other build tools.

The original idea behind the visibility controls was to make all symbols internal by default and only export symbols defined between the `visibility_begin.h` and `visibility_end.h` include files. It seems this has a similar goal, but via a different enumeration. It's not clear what we're gaining here vs. using a pattern since all ODP APIs share the `odp_` prefix.

lumag: @muvarov no, as far as I remember. I will check, when I get back to this PR.
@Bill-Fischofer-Linaro this list should not be autogenerated, it binds symbols to library versions. Consider [libc](https://sourceware.org/git/?p=glibc.git;a=blob;f=math/Versions;h=01ed388ae5dc05115713dbb3c7abfa88b649e260;hb=HEAD), [GnuTLS](https://gitlab.com/gnutls/gnutls/blob/master/lib/libgnutls.map], [libgcrypt](https://git.gnupg.org/cgi-bin/gitweb.cgi?p=libgcrypt.git;a=blob;f=src/libgcrypt.vers;h=edf8a76ae72b9c74431c1b85567f9d33369a45c8;hb=HEAD).

In theory we can go [nettle](https://git.lysator.liu.se/nettle/nettle/blob/master/libnettle.map.in) way. However this limits usefullness of version mapping (versions are limited to the same id).

lumag: @muvarov no, it does not capture an error. It can be implemented though.

@Bill-Fischofer-Linaro as I wrote, second patch is to be discussed.

Bill-Fischofer-Linaro: @lumag OK, we'll cover this on Monday's ARCH call.

muvarov: That is the same solution that DPDK has and we need such changes. But we should clear understand  how to maintain that and use. It's not part of API, it's  implementation linked map script. API will continue use visibility_begin/end.h so that implementation will not export it's functions. So we use it for linux-generic implementation.  Then should we use versions inside .map file? Also I as Bill not sure if we need list of all functions. Instead of export all odp_ and depricated_odp_ functions.

lumag: We might use symbol versions, if we try to provide ABI backwards-compatibility between releases (we do not for now). Otherwise fist patch should be enough.

lumag:
@Bill-Fischofer-Linaro this is post-TigerMoth anyway.

lumag:
Closing for now. We can return to version scripts later

PR #454, opened 2018-02-05
[PATCH v1] build: limit symbol visibility for libodp-linux

lumag:
If libodp-linux is linked with static DPDK libraries, it will re-export
all DPDK symbols. Use libtool's -export-symbols-regex to limit symbol
visibility.


lumag:
Split first patch from #449

muvarov:
Merged.

PR #455, opened 2018-02-06
[PATCH API-NEXT v1] Fix recv timeout issue with socket mmap

psavol:
The issue is on both master and api-next branches. Fixed on api-next branch to avoid merge conflicts as the same lines have been modified on api-next previously.


muvarov:
Merged.

PR #456, opened 2018-02-06
[PATCH v4] Fixes for IPsec inbound inline processing

lumag:


platform/linux-generic/pktio/loop.c
```diff
@@ -155,7 +155,7 @@ static int loopback_recv(pktio_entry_t *pktio_entry, int index ODP_UNUSED,
 
 		/* Try IPsec inline processing */
 		if (pktio_entry->s.config.inbound_ipsec &&
-		    odp_packet_has_ipsec(pkt))
+		    pkt_hdr->p.input_flags.ipsec)
```
Bill-Fischofer-Linaro: Better to use an inline form `_odp_packet_has_ipsec()`, following the model of recent inline enhancements posted by @psavol ?

psavol:
Inside implementation both _odp_packet_has_ipsec(pkt) and pkt_hdr->p.input_flags.ipsec are fine. What implementation should avoid is odp_packet_has_ipsec(pkt) (the full API function) as it's not inlined when ABI compatible.

muvarov:
Merged.

PR #457, opened 2018-02-07
[PATCH API-NEXT v3] Packet L3/L4 type implementation and API extension

lumag:
Implement new packet API. For simplicity add setting functions for L3/L4 packet types.
While we are at it, should we add L2 packet type (only ETH for now) for completeness?


psavol:
I would not add new or deprecate old set functions this close to Tigermoth. Current has_xxx flags map directly HW descriptor flag on some HW. We'll likely revisit the parser metadata fields after Tigermoth as part of tunneling API work.

lumag:
@psavol we can merge first patches, implementing this API, but without deprecation.

lumag:
Replaced by #459

PR #458, opened 2018-02-07
[PATCH v1] linux-gen: dpdk: conf NICs to drop packets if no descriptors are available

matiaselo:
Configure DPDK NIC RX rings to automatically drop packets if no free
descriptors are available. Without this option being set, some NICs (e.g.
82599, X540) will stop receiving packets from all RX queues if any of the
RX queues is not emptied fast enough.

Fixes: https://bugs.linaro.org/show_bug.cgi?id=3618


muvarov:
Merged.

PR #459, opened 2018-02-08
[PATCH API-NEXT v1] Packet L3/L4 type implementation 

lumag:
Implement new packet API.
While we are at it, should we add L2 packet type (only ETH for now) for completeness?

### DIFF
--- a/platform/linux-generic/include/odp/api/plat/packet_inline_types.h
+++ b/platform/linux-generic/include/odp/api/plat/packet_inline_types.h
@@ -80,14 +80,7 @@ typedef union {
 
 		uint64_t ipsec:1;     /* IPSec packet. Required by the
 					   odp_packet_has_ipsec_set() func. */
-		uint64_t ipsec_ah:1;  /* IPSec authentication header */
-		uint64_t ipsec_esp:1; /* IPSec encapsulating security
-					   payload */
-		uint64_t udp:1;       /* UDP */
-		uint64_t tcp:1;       /* TCP */
 		uint64_t tcpopt:1;    /* TCP options present */
-		uint64_t sctp:1;      /* SCTP */
-		uint64_t icmp:1;      /* ICMP */
 
 		uint64_t color:2;     /* Packet color for traffic mgmt */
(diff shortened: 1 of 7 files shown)


psavol:
I suggest to do first implementation the simple way. And optimize after it's in master. This large change conflicts easy other changes and I'd like to compare performance of (current) flag based and the new type field based implementations. I just did optimize parser on master and as usual it's not always obvious which sequence of operations is optimal.

So, let's do first this:

    odp_proto_l3_type_t odp_packet_l3_type(odp_packet_t pkt)
    {
        odp_packet_hdr_t *pkt_hdr = odp_packet_hdr(pkt);

        if (pkt_hdr->p.input_flags.ipv4)
            return ODP_PROTO_L3_TYPE_IPV4;
        else if (pkt_hdr->p.input_flags.ipv6)
            return ODP_PROTO_L3_TYPE_IPV6;
        ....
    }

Since it's a new API, no one is not yet using it and we have time to optimize it. Old APIs are already used, so performance of those (or parser in general) should not lower due this addition.

bala-manoharan:
@psavol I agree its okay to have a simpler implementation in first phase but ipv4 flags will be set only when the parser level is set to L3 and If I am not wrong the odp_packet_l3_type() should return even if parser level is set to L2.

psavol:
odp_packet_parse() API says that e.g. when L2 level parsing is requested, L2 protocol is parsed and L3 offset/ptr is set. Other metadata (incl. l3 type) is not set by current spec.

"In addition, offset (and pointer) to the next layer is set. Other layer/protocol specific metadata have undefined values."

lumag:
@psavol fine, I will redo this PR.

lumag:
@psavol implemeted #463 . I'd suggest to merge master to api-next, compare performance of these two PRs and select one of them for merging. What do you think?

psavol:
I approved 463. So, that should go to api-next and this should wait. Once both that (463) and my parser optimization are in master, we can try to further optimize parser code. Saving the type is one option to implement parser, but it's hard to say in advance if it's faster or slower than the current code (after optimizations).

lumag:
Closing for now

PR #463, opened 2018-02-09
[PATCH API-NEXT v2] Packet L3/L4 type implementation, alternative way

lumag:
This is a simple alternative to #459 implemented on @psavol 's request.

### DIFF
--- a/platform/linux-generic/odp_packet.c
+++ b/platform/linux-generic/odp_packet.c
@@ -2505,3 +2505,41 @@ int odp_packet_has_ref(odp_packet_t pkt)
 
 	return 0;
 }
+
+odp_proto_l3_type_t odp_packet_l3_type(odp_packet_t pkt)
+{
+	odp_packet_hdr_t *pkt_hdr = packet_hdr(pkt);
+
+	if (pkt_hdr->p.input_flags.ipv4)
+		return ODP_PROTO_L3_TYPE_IPV4;
+	else if (pkt_hdr->p.input_flags.ipv6)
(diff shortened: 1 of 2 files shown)


lumag:
Repushed to trigger build

platform/linux-generic/odp_packet.c
```diff
@@ -2544,3 +2544,41 @@ int odp_packet_has_ref(odp_packet_t pkt)
...
+		return ODP_PROTO_L3_TYPE_ARP;
+
+	return ODP_PROTO_L3_TYPE_NONE;
+}
+
+odp_proto_l4_type_t odp_packet_l4_type(odp_packet_t pkt)
+{
+	odp_packet_hdr_t *pkt_hdr = odp_packet_hdr(pkt);
+
+	if (pkt_hdr->p.input_flags.tcp)
```
bala-manoharan: Why don't we use a switch case instead of a nested if-else?

muvarov: @bala-manoharan swithc for set bits and mix of this bils? I don't think that it's possible here.

muvarov:
commits are good with small rebase. There was discussion abot this PR. Waiting for resolution.

lumag:
@muvarov which rebase? I don't see a conflict.

psavol:
This is ready to be merged. Bala suggested to use switch-case but that does not work well with bit fields.

lumag:
@muvarov can we please finally merge this, I have other pending patches.

muvarov:
Merged.

PR #464, opened 2018-02-09
[PATCH API-NEXT v9] crypto: 3GPP algorithms definitions

lumag:
This is post-TigerMoth API change. Proposing standard low level API definitions for KASUMI/SNOW 3G/ZUC algorithms. Protocol-level API (constructing IV data from protocol data) is not defined/implemented.


lumag:
V2: fixing checkpatch for tests

Bill-Fischofer-Linaro:
@psavol Please review for API changes.

Bill-Fischofer-Linaro:
@lumag Please rebase to eliminate conflicts and restart the discussion of this PR.

include/odp/api/spec/crypto.h
```diff
@@ -510,6 +510,15 @@ typedef struct odp_crypto_session_param_t {
...
 	odp_pool_t output_pool;
 
+	/** Bitstring mode
+	 *
+	 *  Cipher and auth will operate in bitstring mode. Fields cipher_range
+	 *  and auth_range will count bits for offset and length, rather than
+	 *  full bytes. Note: not algorithms support this mode of operation, it
+	 *  has to be reported via bitstring field in cipher and auth
+	 *  capability. */
+	odp_bool_t bitstring;
```
Bill-Fischofer-Linaro: The term `bitstring` is a bit generic as there are lots of potential uses of bit strings. Perhaps call this `bitstring_mode`, which also has the advantage of matching the documentation here? Same comment for other two uses of this field name.

psavol: Typo: "Note: not algorithms.." should be "Note: not all algo ..."

psavol: I think it would be simpler API if we just divide algorithms into byte and bit offset/length ones. Kasumi/Snow/ZUC would expect that cipher_ and auth_range fields contain bit offset/length. Others expect bytes. So, all bitstring specification can be removed and replaced with something like this:

	/** Data range to apply cipher
+	 *   Range is specified in bits for Kasumi/Snow/ZUC and in bytes for all other algorithms. 
+	*/
	odp_packet_data_range_t cipher_range;

	/** Data range to authenticate
+	 *   Range is specified in bits for Kasumi/Snow/ZUC and in bytes for all other algorithms. 
+	*/
	odp_packet_data_range_t auth_range;

} odp_crypto_packet_op_param_t;

Bill-Fischofer-Linaro: When bit offsets are used do we have to be concerned about endianness?

lumag: @psavol I thought about this (and it is how DPDK defines those algorithms). However 128-EEA2/128-EIA2 algorithms are plain AES-CTR/AES-CMAC algorithms, but applied to bit strings, rather than bytes.

@Bill-Fischofer-Linaro Hmm. I think no. There are no chips with different endianness of bits inside a byte.

lumag:
Updated PR according to review comments.
TODO:
 - define if there is need to express half-byte start offsets for 3GPP algorithms.

Bill-Fischofer-Linaro:
@lumag Looks like this needs to be rebased to avoid picking up extraneous commits.

lumag:
@Bill-Fischofer-Linaro I've erronously rebased it against master branch, rather than api-next. Should be fixed now.

matiaselo:
I verified that these algorithms are supported by our QAT card, so we can use that for testing. One thing I noticed is that the DPDK QAT driver requires Kasumi F9 IV size to be zero.

qat_crypto_capabilities.h:
```
	{       /* KASUMI (F9) */					\
		.op = RTE_CRYPTO_OP_TYPE_SYMMETRIC,			\
		{.sym = {						\
			.xform_type = RTE_CRYPTO_SYM_XFORM_AUTH,	\
			{.auth = {					\
				.algo = RTE_CRYPTO_AUTH_KASUMI_F9,	\
				.block_size = 8,			\
				.key_size = {				\
					.min = 16,			\
					.max = 16,			\
					.increment = 0			\
				},					\
				.digest_size = {			\
					.min = 4,			\
					.max = 4,			\
					.increment = 0			\
				},					\
				.iv_size = { 0 }			\
			}, }						\
		}, }							\
	},
```

include/odp/api/spec/crypto.h
```diff
@@ -698,6 +755,14 @@ typedef struct odp_crypto_cipher_capability_t {
 	/** IV length in bytes */
 	uint32_t iv_len;
 
+	/** Cipher is operating in bitstring mode
+	 *
+	 * This cipher works on bit strings, rather than sequences of bytes:
+	 * cipher_range's length parameter in odp_crypto_op_param_t and
+	 * odp_crypto_packet_op_param_t will be expressed in bits, rather than
+	 * bytes */
+	odp_bool_t bitstring_mode;
```
psavol: I'd call this bit_mode. We don't call byte length/offset defined data as "byte strings", so I would not introduce "bit string" term either. Also, length and offset should be give in the same unit (normally both are bytes, in bit mode both are bits).

Just define that: "in bit mode cipher_range parameters (length and offset) are given in bits, instead of bytes"

psavol:

lumag:
@psavol 
Basing on tests done with ODP-DPDK by @matiaselo and me:

- QAT driver does not support these algorithms if both length and offset are not bit aligned (and this is not expressed in any way in capabilities, there is no way to know that beforehand)
- Intel's Kasumi/Snow3G/ZUC drivers do not support these algorithms in bit mode in in-place mode (src mbuf = dst mbuf), which is used by ODP-DPDK (again, no capability for that in DPDK).

Viable alternatives for this PR (and ODP-DPDK's one):
- Redefine these algorithms using just bytes instead of bits and forget about bits at all
- Define that although `bit_mode` requires offset/length to be expressed in bits, real support for unaligned values might depend on exact implementation, drop non-byte-aligned test vectors.
- Just drop non-byte-aligned test vectors
- Drop this PR altogether
- ???

Bill-Fischofer-Linaro:
@bala-manoharan since this requirement originated from Cavium, please advise as to your preference here. From earlier discussions it seems that nobody is actually using unaligned bit strings so it seems we should either define these as aligned bit strings of integral length and convert to bytes as needed for DPDK use, or alternately redefine these as standard byte offset/lengths and convert to bits internally if needed by, e.g., Cavium HW.

We can discuss in more detail on Monday's call.

Bill-Fischofer-Linaro:
@lumag Please rebase to pick up changes from sync of master to api-next

lumag:
@Bill-Fischofer-Linaro rebased

muvarov:
Merged.

PR #465, opened 2018-02-09
linux-generic: misc: add _odp_typeval64() internal macro

Bill-Fischofer-Linaro:
Add the internal macro _odp_typeval64() to extract a full 64-bit value
from ODP handles and use this in the _odp_pri() internal macro used
by the various odp_xxx_to_u64() APIs to avoid truncation issues on
64-bit systems.


lumag:
I don't think it is a good idea. I think two better approaches would be:
 - redefine `_odp_typeval` to return `uintptr_t`
OR
 - stop using strong type macros (typeval, pri) with packets/buffers, as they are not handlers, but pointers internally. So we should just return the pointer as 64-bit integer without going into `odp_typeval`/`_odp_pri` calls.

Bill-Fischofer-Linaro:
Dropped in favor of PR #468

PR #466, opened 2018-02-10
test: misc: use C++ I/O instead of C

lumag:
To verify that this test is really compiled using C++ compiler use cout
instead of printf.


muvarov:
Merged.

muvarov:
This patch was skipped. It's merged now.

PR #467, opened 2018-02-10
[PATCH v6] Select scheduler+queue in runtime, basing on environment

lumag:
This would allow us to drop compile-time option in favour of runtime configuration. We do compile all schedulers anyway, so it's a minor step to select them in runtime. This would allow us to do code coverage for non-default schedulers.

### DIFF
--- a/.travis.yml
+++ b/.travis.yml
@@ -53,9 +53,6 @@ env:
         - CONF=""
         - CONF="--disable-abi-compat"
         - CONF="--enable-deprecated"
-        - CONF="--enable-schedule-sp"
-        - CONF="--enable-schedule-iquery"
-        - CONF="--enable-schedule-scalable"
         - CONF="--enable-dpdk-zero-copy"
         - CONF="--disable-static-applications"
         - CONF="--disable-host-optimization"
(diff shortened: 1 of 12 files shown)


lumag:
Interesting. With this PR Clang 3.8/3.9 produces executables which segfault in access to . Upgrading to Clang 4.0 fixes the issue.

```
Program received signal SIGSEGV, Segmentation fault.
0x000000000042c670 in _odp_thread_id () at ../../../odp/platform/linux-generic/include/odp/api/plat/thread_inlines.h:27
27		return _odp_this_thread->thr;
(gdb) bt
#0  0x000000000042c670 in _odp_thread_id () at ../../../odp/platform/linux-generic/include/odp/api/plat/thread_inlines.h:27
#1  odp_thread_id () at ../../../odp/platform/linux-generic/include/odp/api/plat/thread_inlines_api.h:20
#2  0x00000000004353a4 in _odp_crypto_init_local () at ../../../odp/platform/linux-generic/odp_crypto.c:1309
#3  0x0000000000404b54 in odp_init_local (instance=<optimized out>, thr_type=ODP_THREAD_WORKER) at ../../../odp/platform/linux-generic/odp_init.c:330
#4  0x00000000004034dc in main (argc=<optimized out>, argv=<optimized out>) at ../../../odp/helper/test/chksum.c:48
```

.travis.yml
```diff
@@ -53,9 +53,6 @@ env:
         - CONF=""
         - CONF="--disable-abi-compat"
         - CONF="--enable-deprecated"
-        - CONF="--enable-schedule-sp"
-        - CONF="--enable-schedule-iquery"
-        - CONF="--enable-schedule-scalable"
         - CONF="--enable-dpdk-zero-copy"
```
psavol: Run time selection is good addition and can be even the default behavior. Anyway, I'd like to keep the build time selection also, since usually one application uses the same scheduler always. Fixing the scheduler at build time avoids accidentally usage of different schedulers in different servers (or VMs or test environments) and thus avoid problems/debugging caused by that.

Bill-Fischofer-Linaro: Perhaps `--enable-schedule-dynamic` (default) to indicate run time selection that can be overridden by a specific static selection at `./configure` time via the other `--enable-schedule-xxx` options?

lumag: This would mean that we have now 4 options to test, instead of just one. What about renaming 'default' scheduler to some sensible name and adding `--with-scheduler-default=FOO`? Then we will have just one combination to test.

psavol: There can be just one config option: --with-scheduler=foo. When that's defined then "foo" would be used (fixed at build time). If --with-scheduler is not found, the value of ODP_SCHEDULER=foo environment variable is used (run time selection is the default behavior). If the variable is not found then the default scheduler is used.

lumag: @psavol this would also require seprate compile testing of `--with-scheduler` option. I'm thinking about `--with-scheduler-default` option with the following behaviour:
 - If environment option is given, it selects scheduler/queue implementation
 - If no environment option is given, compile-time option selects the scheduler/queue
 - If no compile-time option was given, we use some sensible default (basic for now, maybe scalable in future).

Bill-Fischofer-Linaro: @psavol Is @lumag's latest suggestion OK with you?

Bill-Fischofer-Linaro:
Ping to @psavol. Still waiting for your response to @lumag's latest suggestion.

lumag:
@Bill-Fischofer-Linaro current PR is up to agreement reached during Monday's call.

psavol:

lumag:
@muvarov rebased PR

muvarov:
Merged.

PR #471, opened 2018-02-12
validation: shm: fix use of wrong name length define

matiaselo:
Wrong name length define was used when setting terminating character for
shm name.


muvarov:
Merged.

PR #479, opened 2018-02-14
[PATCH v5] validation: timer: fix test failure

lumag:
timer_main test was regularly failing due to worker code receiving
timeout events for cancelled timers. I'm fixing worker code to handle
these events correctly, however this might signify a loophole in API
definition or in linux-generic implementation. Shoud application be able
to copy with timers that are cancelled after expiration (but before
event reception)? Should implementation mark such timers as non-fresh?

Fixes: https://bugs.linaro.org/show_bug.cgi?id=3517

### DIFF
--- a/test/validation/api/timer/timer.c
+++ b/test/validation/api/timer/timer.c
@@ -282,15 +282,28 @@ static void handle_tmo(odp_event_t ev, bool stale, uint64_t prev_tick)
 		if (ttp && ttp->tick != TICK_INVALID)
 			CU_FAIL("Stale timeout for active timer");
 	} else {
-		if (!odp_timeout_fresh(tmo))
-			CU_FAIL("Wrong status (stale) for fresh timeout");
-		/* Fresh timeout => local timer must have matching tick */
-		if (ttp && ttp->tick != tick) {
-			LOG_DBG("Wrong tick: expected %" PRIu64
-				" actual %" PRIu64 "\n",
-				ttp->tick, tick);
-			CU_FAIL("odp_timeout_tick() wrong tick");
+		if (ttp && ttp->tick == TICK_INVALID) {
+			/* Timer was cancelled when it has already expired */
+
+			if (odp_timeout_fresh(tmo))
+				CU_FAIL("Wrong status (fresh) "
+					"for cancelled timeout");
+		} else {
+			if (!odp_timeout_fresh(tmo))
+				CU_FAIL("Wrong status (stale) "
+					"for fresh timeout");
+
+			/* Fresh timeout => local timer must have matching tick
+			 */
+			if (ttp && ttp->tick != tick) {
+				LOG_DBG("Wrong tick: expected %" PRIu64
+					" actual %" PRIu64 "\n",
+					ttp->tick, tick);
+				CU_FAIL("odp_timeout_tick() wrong tick");
+			}
 		}
+		if (ttp && ttp->ev != ODP_EVENT_INVALID)
+			CU_FAIL("Wrong state for fresh timer (event)");
 		/* Check that timeout was delivered 'timely' */
 		if (tick > odp_timer_current_tick(tp))
 			CU_FAIL("Timeout delivered early");
@@ -401,9 +414,11 @@ static int worker_entrypoint(void *arg TEST_UNUSED)
 		    (rand_r(&seed) % 2 == 0)) {
 			/* Timer active, cancel it */
 			rc = odp_timer_cancel(tt[i].tim, &tt[i].ev);
-			if (rc != 0)
+			if (rc != 0) {
 				/* Cancel failed, timer already expired */
 				ntoolate++;
+				LOG_DBG("Failed to cancel timer, probably already expired\n");
+			}
 			tt[i].tick = TICK_INVALID;
 			ncancel++;
 		} else {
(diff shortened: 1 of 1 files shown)


lumag:
@psavol this needs your review from the API point of view.

psavol:
Should application be
able to copy with timers that are cancelled after expiration (but before
event reception)? Should implementation mark such timers as non-fresh?

Cancel API clearly says that either cancel succeeds and the event is returned by the cancel call, or cancel failed (it was too late to cancel) and event will come eventually to the destination queue.

Yes - application must be prepared for cancel call to return "fail".

Fresh call should return "stale" for those timeouts that were cancelled during tmo is in a queue, but obviously it depends on timing cancel and fresh calls between cpus. A very late cancel (just when another cpu is calling fresh()) will not be noticed by the fresh() call. 

Yes - implementation marks timeouts stale, but there may be latency between CPUs. So, tmo can be still fresh (on another cpu) when cancel returns on another one.

lumag:
@psavol thank you for your comments. Then I assume my fix is correct.

WonderfulVoid:
On Fri, Feb 16, 2018 at 12:11 PM, Petri Savolainen <notifications@github.com
> wrote:

> Should application be
> able to copy with timers that are cancelled after expiration (but before
> event reception)? Should implementation mark such timers as non-fresh?
>
Either the application or the timer bg thread (might not actually be a
separate thread) atomically changes the timer expiration value to make it
invalid. Only one of them can succeed. So if the timer has expired (and
been updated by the bg thread), it cannot be cancelled (odp_timer_cancel()
should return <0)  and the associated timeout event has been (or is going
to be) enqueued to the queue associated with the timer.

Cancel API clearly says that either cancel succeeds and the event is
> returned by the cancel call, or cancel failed (it was too late to cancel)
> and event will come eventually to the destination queue.
>
> Yes - application must be prepared for cancel call to return "fail".
>
> Fresh call should return "stale" for those timeouts that were cancelled
> during tmo is in a queue, but obviously it depends on timing cancel and
> fresh calls between cpus. A very late cancel (just when another cpu is
> calling fresh()) will not be noticed by the fresh() call.
>
If a timer was successfully cancelled, the timeout event will not be
enqueued. If the timer has already expired and the timeout event has been
enqueued, cancel will fail. So I don't understand when "timeouts that were
cancelled during tmo is in a queue" can happen.

Yes - implementation marks timeouts stale, but there may be latency between
> CPUs. So, tmo can be still fresh (on another cpu) when cancel returns on
> another one.
>
The implementation marks expired timers in a different way from cancelled
timers. Expired timers still have the original timeout value available.
This is not true for cancelled timers where the whole timeout value is
"cleared".

lumag:
@WonderfulVoid ok, let me rephrase the question in a different way: is the PR correct from your point of view, or the issue that we see points to the bug in linux-gen implementation?

muvarov:
@lumag  please change first patch description.

lumag:
@muvarov done

WonderfulVoid:
"Should application be
able to copy with timers that are cancelled after expiration (but before
event reception)?"

You cannot cancel an expired timer. odp_timer_cancel() will detect
that the timer has expired and return an error (<0). The timer will be
unchanged.

"Should implementation mark such timers as non-fresh?"

The timer is marked when it expires but the expiration value is still
available. A timeout is considered fresh if the timeout expiration
tick matches the current timer expiration tick (ignoring the
active/inactive status bit which is stored in the same location). This
means the timer has not been reset (to a new timeout tick). If the
timer has been reset, the (old) timeout is considered stale (== not
fresh).

So freshness is a property of the timeout, not the timer. A fresh
timeout can become stale without any action from the thread which has
the timeout.

I think the timer implementation does things the right way. So likely
it is the timer validation suite which doesn't handle things
correctly. As there are a lot of race conditions here, this seems
plausible.


Regarding the code changes:

+			/* Timer was cancelled when it has already expired */
As an expired timer cannot actually be cancelled, this comment is
invalid. But the test program can have attempted to cancel a timer and
failed because the timer had already expired. The test program might
thus consider the timer cancelled but I think that's misuse of the
timer terminology and can only create more confusion.

			/* Timer active, cancel it */

			rc = odp_timer_cancel(tt[i].tim, &tt[i].ev);
			if (rc != 0) {
				/* Cancel failed, timer already expired */
				ntoolate++;
				LOG_DBG("Failed to cancel timer, probably already expired\n");
			}
			tt[i].tick = TICK_INVALID;
			ncancel++;

Possibly setting tt[i].tick to TICK_INVALID also when timer cancel fails
(because timer has expired) is not a good idea. Then we will think the
timer has been cancelled when it actually expired normally.
I think this is the only place where we set tt->tick to TICK_INVALID in the
"main" loop of the test. The initial value of tt->tick is also
TICK_INVALID. tt->tick is written for a successful odp_timer_set_abs/rel()
call. I think we should check if odp_timer_cancel() fails above and if this
is related to the failed tests. Probably we need to handle
odp_timer_cancel() failures differently, i.e. not just setting tt[i].tick =
TICK_INVALID.

lumag:
I see then. `odp_timer_cancel()` should be fixed not to touch timer, if timeout has already fired, correct?

WonderfulVoid:
Yes possibly odp_timer_cancel() should not modify exp_tck if the timer has
already expired. Currently is sets exp_tck to 0x8000000000000000 regardless
of the old state.
How would this change affect the test program? I think failure to cancel a
timer needs to change as well (not setting ttp->ttick).

-- Ola

WonderfulVoid:
With the current implementation of odp_timer_cancel(), even if the timer
had already expired and cancellation thus failed, it will look like the
timer was successfully cancelled because exp_tck was updated regardless.
The test program updates ttp->tick to TICK_INVALID even if the cancellation
failed. But we will still receive that timeout event which refers to the
timer. Inside the timer implementation the timer will look cancelled, in
the test program state the timer will look cancelled but we still get the
timeout event. Chaos ensues.

So start by changing odp_timer_cancel() not to change any internal state if
the timer cannot actually be cancelled (because it is inactive/has already
expired) and we will take it from there.

lumag:
Closing in favour of #491

PR #483, opened 2018-02-15
[PATCH v2] shippable: disable perf-proc tests

lumag:
These tests have tendency of hanging up Shippable occasionally.


muvarov:
Merged.

PR #485, opened 2018-02-15
[PATCH v1] example: generator move to platform tests

muvarov:
run generator example with null pktio from linux-generic test directory.
Null pktio might not be implemented on other odp implementations.


muvarov:
Merged.

PR #488, opened 2018-02-16
[PATCH v5] odp: pktio: add pcapng capture capabilities

apalos:
Changes since RFC:
	- addressed issues reported in RFC
	- rewrote some functions for easier understanding
	- added capability on all input queues

How to test:
sudo mkdir /var/run/odp/
start the ODP application
sudo dd if=/var/run/odp/"odp pid"-"inteface"-flow-"queue number" of=~/test.pcap

### DIFF
--- a/.travis.yml
+++ b/.travis.yml
@@ -58,6 +58,7 @@ env:
         - CONF="--disable-static-applications"
         - CONF="--disable-host-optimization"
         - CONF="--disable-host-optimization --disable-abi-compat"
+        - CONF="--enable-pcapng-support"
         - DPDK_SHARED="y" CONF="--disable-static-applications"
 
 compiler:
--- a/doc/users-guide/users-guide-utilities-examples.adoc
+++ b/doc/users-guide/users-guide-utilities-examples.adoc
@@ -1,0 +1,19 @@
+== Utilities and examples
+
+=== PcapNg capture
+If compiled using `--enable-pcapng-support` ODP will offer packet capturing
+functionality in PcapNg format. If the /var/run/odp directory exists prior to
+launching the application ODP will create a fifo for each NIC queue.
+Queue naming will be of the following format: *<odp global pid>-<NIC
+name>-flow-<queue number>*. Linux dd application can be used for capturing a 
+sample of the live stream from the fifo. Killing ether the application or dd
+will stop the capturing process.
+
+. `./configure --enable-pcapng-support`
+. `sudo mkdir /var/run/odp`
+. `sudo ./example/generator/odp_generator -I enp2s0 -mu --dstmac
+A0:F6:FD:AE:62:6C --dstip 192.168.49.20 --srcmac 2c:56:dc:9a:8f:06 --srcip
+192.168.49.4 -i0 -w1`
+. `sudo dd if=/var/run/odp/26737-enp2s0-flow-0 of=~/test.pcap`
+. `ctrl^c`
+. `wireshark ~/test.pcap`
--- a/doc/users-guide/users-guide.adoc
+++ b/doc/users-guide/users-guide.adoc
@@ -1204,4 +1204,6 @@ include::users-guide-tm.adoc[]
 
 include::users-guide-cls.adoc[]
 
+include::users-guide-utilities-examples.adoc[]
+
 include::../glossary.adoc[]
--- a/platform/linux-generic/m4/odp_pcapng.m4
+++ b/platform/linux-generic/m4/odp_pcapng.m4
@@ -1,0 +1,18 @@
+##########################################################################
+# Enable PCAPNG support
+##########################################################################
+have_pcapng=no
+pcapng_support=0
+
+AC_ARG_ENABLE([pcapng-support],
+	[AS_HELP_STRING([--enable-pcapng-support],
+	[enable experimental tcpdump for pktios])],
+	have_pcapng=$enableval
+    [if test x$enableval = xyes; then
+        pcapng_support=1
+    fi])
+
+AC_DEFINE_UNQUOTED([_ODP_PCAPNG], [$pcapng_support],
+	[Define to 1 to enable pcapng support])
+
+AM_CONDITIONAL([have_pcapng], [test x$have_pcapng = xyes])
--- a/platform/linux-generic/odp_packet_io.c
+++ b/platform/linux-generic/odp_packet_io.c
@@ -25,6 +25,7 @@
 #include <odp_debug_internal.h>
 #include <odp_packet_io_ipc_internal.h>
 #include <odp/api/time.h>
+#include <odp_pcapng.h>
 
 #include <string.h>
 #include <inttypes.h>
@@ -512,6 +513,11 @@ int odp_pktio_start(odp_pktio_t hdl)
 	ODP_DBG("interface: %s, input queues: %u, output queues: %u\n",
 		entry->s.name, entry->s.num_in_queue, entry->s.num_out_queue);
 
+	if (_ODP_PCAPNG) {
+		if (pcapng_prepare(entry))
+			ODP_ERR("pcap init failed, won't capture\n");
+	}
+
 	return res;
 }
 
@@ -535,6 +541,9 @@ static int _pktio_stop(pktio_entry_t *entry)
 		entry->s.state = PKTIO_STATE_STOP_PENDING;
 	else
 		entry->s.state = PKTIO_STATE_STOPPED;
+
+	if (_ODP_PCAPNG)
+		pcapng_destroy(entry);
 
 	return res;
 }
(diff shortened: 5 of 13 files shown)


platform/linux-generic/m4/odp_pcapng.m4
```diff
@@ -0,0 +1,19 @@
...
+	have_pcapng=$enableval)
+
+if test x$have_pcapng = xyes
+then
+    AC_DEFINE([ODP_PCAPNG], [1],
+	      [Define to 1 to enable pcapng support])
+else
+    AC_DEFINE([ODP_PCAPNG], [0],
+	      [Define to 0 to disable pcapng support])
+fi
```
lumag: Just use one AC_DEFINE and check using `#ifdef` inside.

platform/linux-generic/odp_packet_io.c
```diff
@@ -485,6 +486,11 @@ int odp_pktio_start(odp_pktio_t hdl)
 		sched_fn->pktio_start(_odp_pktio_index(hdl), num, index, odpq);
 	}
 
+	if (ODP_PCAPNG) {
```
lumag: Neat idea, but just put pcapng_prepare inside header and define it to empty val if ODP_PCAPNG is undefined.
Also if it returns an error, it should not require one to call pcapng_destroy() afterwards. Just make it clean up after itself in case of an error.

platform/linux-generic/odp_packet_io.c
```diff
@@ -1650,18 +1659,30 @@ int odp_pktout_queue(odp_pktio_t pktio, odp_pktout_queue_t queues[], int num)
...
 	entry = get_pktio_entry(pktio);
 	if (entry == NULL) {
 		ODP_DBG("pktio entry %d does not exist\n", pktio);
 		return -1;
 	}
 
-	return entry->s.ops->recv(entry, queue.index, packets, num);
+	ret = entry->s.ops->recv(entry, queue.index, packets, num);
+	if (ODP_PCAPNG)
+		dump_pcapng_pkts(entry, queue.index, packets, ret);
```
lumag: `_odp_smth...`

platform/linux-generic/odp_packet_io.c
lumag: And also it should be `_odp_pcapng_FOO`

.travis.yml
```diff
@@ -58,6 +58,7 @@ env:
         - CONF="--enable-schedule-scalable"
         - CONF="--enable-dpdk-zero-copy"
         - CONF="--disable-static-applications"
+        - CONF="--enable-pcapng-support"
```
lumag: I'm thinking about limiting variety of build options here. Not in your PR, but there are just way tooo many

lumag:
@apalos also a side note. Latest FHS talks about `/run` as a replacement for `/var/run`. Do you know whether Ubuntu 14.04 has `/run`?

doc/users-guide/users-guide-utilities-examples.adoc
```diff
@@ -0,0 +1,17 @@
+== Utilities and examples
+
+=== PcapNg capture
+If compiled using `--enable-pcapng-support` ODP will offer packet capturing
+functionality in PcapNg format. If /var/run/odp directory exists prior to
+launching the application ODP will create as many fifos as the NIC queues.
+Queue naming will be of the following format: *<odp global pid>-<NIC
+name>-flow-<queue *number>*
```
Bill-Fischofer-Linaro: This formats as __\<odp global pid\>-\<NIC name\>-flow-\<queue *number\>__ 

Is that what you wanted?

doc/users-guide/users-guide.adoc
```diff
@@ -1198,4 +1198,6 @@ include::users-guide-tm.adoc[]
 
 include::users-guide-cls.adoc[]
 
+include::users-guide-utilities-examples.adoc[]
```
Bill-Fischofer-Linaro: Having this as a section of the User Guide is fine, but you need to add this file to the list in `doc/users-guide/Makefile.am` so that changes to it will cause the User Guide to be regenerated.

platform/linux-generic/odp_packet_io.c
Bill-Fischofer-Linaro: @lumag This style ensures that the code is always syntax-checked since it is compiled, but then optimized away if it is not active. I've found this to be a useful idiom.

apalos: This is how dpdk zero-copy was introduced. I think we need to keep that for the same reasons @Bill-Fischofer-Linaro mentioned

platform/linux-generic/m4/odp_pcapng.m4
apalos: We did that in net_mdev and the if was corrected to check the value was equal to 1

.travis.yml
apalos: Cuinit caught actua; bugs in the original implementation, i'd like to keep that

doc/users-guide/users-guide-utilities-examples.adoc
apalos: no the extra * will be removed on v2

apalos:
@lumag no idea about /run directory on older Ubuntu. If you have better suggestion than /var/run directory for fifos, please let me know

platform/linux-generic/m4/odp_pcapng.m4
lumag: Anyway, just call AC_DEFINE once.

platform/linux-generic/odp_packet_io.c
lumag: There is no need to syntax check function call, is it?
Anyway, `_prepare` call should not have side-effects in error case, so cleanup should be done internally.

.travis.yml
lumag: @apalos I'm not arguing against this check. I'm thinking if we can lower the amount of different configurations that we check. Like having just big and little configs.

apalos:
checkpatch failure on V2 is due to typo on the commit message. It also suggest splitting the commit message but that would make it unreadable

platform/linux-generic/odp_packet_io.c
```diff
@@ -485,6 +486,11 @@ int odp_pktio_start(odp_pktio_t hdl)
 		sched_fn->pktio_start(_odp_pktio_index(hdl), num, index, odpq);
 	}
 
+	if (ODP_PCAPNG) {
+		if (pcapng_prepare(entry))
```
lumag: _odp_pcapng

platform/linux-generic/odp_pcapng.c
```diff
@@ -0,0 +1,434 @@
...
+	int iovcnt = 0;
+	ssize_t block_len = 0;
+	int fd = entry->s.pcapng.fd[qidx];
+	ssize_t len = 0, wlen;
+
+	for (i = 0; i < num; i++) {
+		odp_packet_hdr_t *pkt_hdr = packet_hdr(packets[i]);
+		uint32_t seg_len;
+		char *buf = (char *)odp_packet_offset(packets[i], 0, &seg_len,
+						      NULL);
```
lumag: Is writing only first segment enough?

apalos: No but it won't break the capture. This was suggested by @Bill-Fischofer-Linaro as an improvement over the initial commit

doc/users-guide/users-guide-utilities-examples.adoc
```diff
@@ -0,0 +1,17 @@
+== Utilities and examples
+
+=== PcapNg capture
+If compiled using `--enable-pcapng-support` ODP will offer packet capturing
+functionality in PcapNg format. If /var/run/odp directory exists prior to
```
Bill-Fischofer-Linaro: If the `/var/run/odp` directory exists...

doc/users-guide/users-guide-utilities-examples.adoc
```diff
@@ -0,0 +1,17 @@
...
+launching the application ODP will create as many fifos as the NIC queues.
```
Bill-Fischofer-Linaro: ...as many fifos as there are NIC queues. Alternately, "will create a fifo for each NIC queue".

doc/users-guide/users-guide-utilities-examples.adoc
```diff
@@ -0,0 +1,17 @@
...
+Queue naming will be of the following format: *<odp global pid>-<NIC
+name>-flow-<queue number>*
+
+. `./configure --enable-pcapng-support`
+. `sudo mkdir /var/run/odp`
+. `sudo ./example/generator/odp_generator -I enp2s0 -mu --dstmac
+A0:F6:FD:AE:62:6C --dstip 192.168.49.20 --srcmac 2c:56:dc:9a:8f:06 --srcip
+192.168.49.4 -i0 -w1`
+. `sudo dd if=/var/run/odp/26737-enp2s0-flow-0 of=~/test.pcap`
+. `ctrl^c`
```
Bill-Fischofer-Linaro: Why wouldn't you do `ctrl^c` before `dd`? This seems backwards.

apalos: ctrl^c is there to stop the dd process, so it should follow the dd command

apalos: fixed

muvarov:
Merged. I think it's good debugging feature. Some minor things we can clean up later.

Bill-Fischofer-Linaro:
@muvarov Did you correct the checkpatch issues as part of the merge? I noticed them after my review. They should be fixed to avoid latent issues in the base.

PR #489, opened 2018-02-16
[PATCH v1] odp_dpdk.m4: fix detection of system libraries

joseppc:


muvarov:
Merged.

PR #490, opened 2018-02-20
[PATCH v1] Fix creating pools larger than 4.29GB

matiaselo:
Fixes bug: https://bugs.linaro.org/show_bug.cgi?id=3632


muvarov:
Merged.

PR #491, opened 2018-02-20
[PATCH v2] Timer fix, another approach

lumag:
This is a PR to replace #479 . It stops `odp_timer_cancel()`  from marking timeouts as stale for already-fired timeouts. @psavol @WonderfulVoid please review.

### DIFF
--- a/platform/linux-generic/odp_timer.c
+++ b/platform/linux-generic/odp_timer.c
@@ -375,9 +375,8 @@ static inline odp_timer_t timer_alloc(timer_pool_t *tp,
 	return hdl;
 }
 
-static odp_buffer_t timer_cancel(timer_pool_t *tp,
-				 uint32_t idx,
-				 uint64_t new_state);
+static odp_buffer_t timer_set_unused(timer_pool_t *tp,
+				     uint32_t idx);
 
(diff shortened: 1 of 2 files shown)


muvarov:
@lumag it looks like timer_run(); which is called under scheduler has to be void().  Might be good to add to this patch series.

muvarov:
timer_cancel() function is not need and code can go directly to odp_timer_cancel() now.

muvarov:
In general patches is ok. I only have few questions 1) did you test both arms for ifdefs? 2) is there performance reason to do 32 bit exp_tck.v counter on 32 bit systems? bit field operation on 64 bits are very slow for example on arm32.

Bill-Fischofer-Linaro:
@WonderfulVoid Your comments on this version of the fix appreciated.

lumag:
@muvarov For `timer_cancel()` I followed the rest of the code.
I'll recheck them with locked implementation.

lumag:
@muvarov manually tested by changing preprocessor macros to use locks instead of U128 atomics.

lumag:
I plan to rework this piece of code sometime in future, but it is not the top-priority item.

muvarov:
Merged.

PR #492, opened 2018-02-21
Ring implementation of queues

psavol:
Optimize queues and scheduling with a ring based implementation. Ring operations are not atomic and  behind locks. Scheduler throughput improves in L2fwd test about 30%. Buffer header gets smaller and not related to burst size, which has positive impact to direct mode pktio throughput also.

```
2x10GE
            | direct |                  | atomic | 
cpus | before | after | speed up | before | after | speed up
1    | 19.4   | 19.8  | 1.02     | 13.1   | 15.0  | 1.14
2    | 29.7   | 29.7  | 1.00     | 16.9   | 21.5  | 1.27
4    |                           | 24.7   | 28.7  | 1.16

```
```
2x40GE
         | direct |                | atomic |
cpus | before| after | speed up | before | after | speed up
1    | 0.182 | 0.186 | 1.02     | 0.12   | 0.148 | 1.23
2    | 0.367 | 0.37  | 1.01     | 0.16   | 0.205 | 1.28
4    | 0.597 | 0.596 | 1.00     | 0.241  | 0.314 | 1.30
6    | 0.571 | 0.571 | 1.00     | 0.295  | 0.386 | 1.31
8    | 0.556 | 0.556 | 1.00     | 0.356  | 0.498 | 1.40


```


platform/linux-generic/include/odp_buffer_internal.h
```diff
@@ -41,11 +41,19 @@ typedef struct seg_entry_t {
 	uint32_t  len;
 } seg_entry_t;
 
+typedef union buffer_index_t {
+	uint32_t u32;
+
+	struct {
+		uint32_t pool   :8;
+		uint32_t buffer :24;
```
Bill-Fischofer-Linaro: We originally had this index partitioning based on `ODP_CONFIG_POOLS`. Do we want to return to that here?

If not, we at least need an `ODP_STATIC_ASSERT()` to ensure that `ODP_CONFIG_POOLS < 256` or else bad things will happen here.

platform/linux-generic/odp_pool.c
```diff
@@ -296,7 +282,9 @@ static void init_buffers(pool_t *pool)
 		memset(buf_hdr, 0, (uintptr_t)data - (uintptr_t)buf_hdr);
 
 		/* Initialize buffer metadata */
-		buf_hdr->index = i;
+		buf_hdr->index.u32    = 0;
+		buf_hdr->index.pool   = pool->pool_idx;
+		buf_hdr->index.buffer = i;
```
Bill-Fischofer-Linaro: Since you're initializing `index.pool` and `index.buffer` there's no need to set `index.u32` here.

platform/linux-generic/odp_queue.c
```diff
@@ -143,8 +150,10 @@ static int queue_capability(odp_queue_capability_t *capa)
 	capa->max_sched_groups  = sched_fn->num_grps();
 	capa->sched_prios       = odp_schedule_num_prio();
 	capa->plain.max_num     = capa->max_queues;
+	capa->plain.max_size    = CONFIG_QUEUE_SIZE;
```
Bill-Fischofer-Linaro: As noted earlier, due to "losing" one entry to distinguish queue empty/full, this should be returned as `CONFIG_QUEUE_SIZE - 1`, and we also need to ensure that `CONFIG_QUEUE_SIZE` is itself a power of 2.

platform/linux-generic/odp_queue.c
```diff
@@ -143,8 +150,10 @@ static int queue_capability(odp_queue_capability_t *capa)
...
 	capa->plain.nonblocking = ODP_BLOCKING;
 	capa->sched.max_num     = capa->max_queues;
+	capa->sched.max_size    = CONFIG_QUEUE_SIZE;
```
Bill-Fischofer-Linaro: Same comment here as for plain queues.

platform/linux-generic/odp_queue.c
```diff
@@ -192,6 +201,9 @@ static odp_queue_t queue_create(const char *name,
 		param = &default_param;
 	}
 
+	if (param->size > CONFIG_QUEUE_SIZE)
```
Bill-Fischofer-Linaro: Change to `if (param->size >= CONFIG_QUEUE_SIZE)` to handle the effective queue capacity. The user-supplied `size` should then be set to `ROUNDUP_POWER2_U32(size) - 1` for the masking to work properly.

platform/linux-generic/odp_queue.c
```diff
@@ -263,7 +275,7 @@ static int queue_destroy(odp_queue_t handle)
 		ODP_ERR("queue \"%s\" already destroyed\n", queue->s.name);
 		return -1;
 	}
-	if (queue->s.head != NULL) {
+	if (ring_st_is_empty(&queue->s.ring_st) == 0) {
```
Bill-Fischofer-Linaro: `if (!ring_st_is_empty(&queue->s.ring_st))` seems more natural here.

platform/linux-generic/odp_queue.c
```diff
@@ -471,51 +476,18 @@ static inline int deq_multi(queue_entry_t *queue, odp_buffer_hdr_t *buf_hdr[],
...
-	if (hdr == NULL)
-		queue->s.tail = NULL;
-
 	if (status_sync && queue->s.type == ODP_QUEUE_TYPE_SCHED)
 		sched_fn->save_context(queue->s.index);
 
 	UNLOCK(&queue->s.lock);
 
-	return i;
+	buffer_index_to_buf(buf_hdr, buf_idx, num_deq);
```
Bill-Fischofer-Linaro: Given its name, this looks like an extraneous statement that should be deleted. Renaming this to something like `prefetch_dequeued_bufs()` would make the intent clearer here.

platform/linux-generic/odp_queue.c
```diff
@@ -584,8 +556,9 @@ static int queue_init(queue_entry_t *queue, const char *name,
 	queue->s.pktin = PKTIN_INVALID;
 	queue->s.pktout = PKTOUT_INVALID;
 
-	queue->s.head = NULL;
-	queue->s.tail = NULL;
+	ring_st_init(&queue->s.ring_st,
+		     queue_tbl->ring_data[queue->s.index].data,
+		     CONFIG_QUEUE_SIZE);
```
Bill-Fischofer-Linaro: Must use `CONFIG_QUEUE_SIZE - 1` here, as noted earlier, if we're not going to used the user-supplied queue size.

platform/linux-generic/include/odp_config_internal.h
```diff
@@ -144,7 +144,7 @@ extern "C" {
  * This controls the burst size on various enqueue, dequeue, etc calls. Large
  * burst size improves throughput, but may degrade QoS (increase latency).
  */
-#define CONFIG_BURST_SIZE 16
+#define CONFIG_BURST_SIZE 32
```
Bill-Fischofer-Linaro: Does this mean that sizes larger than 32 have no added performance benefit?

platform/linux-generic/include/odp_ring_st_internal.h
```diff
@@ -0,0 +1,118 @@
...
+static inline uint32_t ring_st_deq_multi(ring_st_t *ring, uint32_t data[],
+					 uint32_t max_num)
+{
+	uint32_t head, tail, mask, idx;
+	uint32_t num, i;
+
+	head = ring->head;
+	tail = ring->tail;
+	mask = ring->mask;
+	num  = tail - head;
```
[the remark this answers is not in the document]
psavol: There's no need for abs(). Since it's all uint32_t variables, wrap a round is handled already.
An example in 8bits:
0xff - 0xfd = 0x02
0x00 - 0xfe = 0x02
0x01 - 0xff = 0x02
0x02 - 0x00 = 0x02

This passes both gcc and clang, and is used already in the other ring implementation see ring_deq_multi().

platform/linux-generic/odp_pool.c
psavol: It's there to ensure that all bits are zero also when someone would modify the bitfield from two to three fields later on. Similarly to memset() zero is used for struct inits.

platform/linux-generic/odp_queue.c
psavol: I prefer style where '== 0' is used instead of '!'. Especially, when the if clause is complex enough for reader to miss the '!' sign.

platform/linux-generic/odp_queue.c
psavol: This function converts 32 bit buffer indexes to buffer header pointers. The counter operation is buffer_index_from_buf(). The prefetch is a side effect of the function, which may be changed/moved any time if it's found out that there's a place for prefetching. I actually plan to test if number of prefetches should be limited as e.g. 32 consecutive prefetches may be too much for some CPU architectures.

platform/linux-generic/include/odp_ring_st_internal.h
```diff
@@ -0,0 +1,118 @@
...
+	uint32_t *data;
+
+} ring_st_t;
+
+/* Initialize ring. Ring size must be a power of two. */
+static inline void ring_st_init(ring_st_t *ring, uint32_t *data, uint32_t size)
+{
+	ring->head = 0;
+	ring->tail = 0;
+	ring->mask = size - 1;
```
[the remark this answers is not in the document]
psavol: It's already documented 5 lines above:

/* Initialize ring. Ring size must be a power of two. */
static inline void ring_st_init(ring_st_t *ring, uint32_t *data, uint32_t size)
{

platform/linux-generic/include/odp_ring_st_internal.h
```diff
@@ -0,0 +1,118 @@
...
+					 uint32_t num_data)
+{
+	uint32_t head, tail, mask, size, idx;
+	uint32_t num, i;
+
+	head = ring->head;
+	tail = ring->tail;
+	mask = ring->mask;
+	size = mask + 1;
+	num  = size - (tail - head);
```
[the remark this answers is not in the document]
psavol: Tail and head indexes are (masked from) uint32_t and do not wrap around when the ring is full. I think you assume that the store index is 0...size-1, while it's full uint32_t which is then masked to get the actual index.

For example:
size = 100;

Empty:
head = 100
tail = 100
num = 100 - 100 = 0

Full:
head = 100
tail = 200
num = 200 - 100 = 100

Wrap uint32_t + full:
head = 0xFFFFFF9C
tail = 0
num = 0 - 0xFFFFFF9C = 0x64 = 100

So, no abs() needed. Ring size can be 4096, instead of 4095.

platform/linux-generic/include/odp_ring_st_internal.h
```diff
@@ -0,0 +1,118 @@
...
+	uint32_t head, tail, num;
+
+	head = ring->head;
+	tail = ring->tail;
+	num  = tail - head;
+
+	if (num == 0)
+		return 1;
+
+	return 0;
```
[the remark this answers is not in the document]
psavol: OK. Compiler probably did that already, but changed in v2.

platform/linux-generic/include/odp_buffer_internal.h
psavol: OK, added checks in v2.

platform/linux-generic/odp_queue.c
psavol: One entry is not lost.

platform/linux-generic/odp_queue.c
psavol: One entry is not lost.

platform/linux-generic/odp_queue.c
psavol: One entry is not lost. User provided size if not (currently) used. Queue size is always 4k.

platform/linux-generic/odp_queue.c
psavol: One entry is not lost.

platform/linux-generic/include/odp_config_internal.h
psavol: Didn't try larger than 32. 32 is already quite large from QoS point of view. 

I'm planning to use config file for run time tunning, so this hard coding may change in that phase.

platform/linux-generic/odp_queue.c
Bill-Fischofer-Linaro: Elsewhere you write `if (ring_st_is_empty(...))`, not `if (ring_st_is_empty(...) == 1)` so this is inconsistent.

platform/linux-generic/odp_queue.c
Bill-Fischofer-Linaro: That functionality is not obvious from the name. It either implies that one of the input arguments is written (not true here) or the reader might assume that it is an expression without side-effect and should be deleted (what I originally thought when reading it). You should pick a routine name that makes it clear it's actually doing something real, in this case performing prefetch processing.

platform/linux-generic/include/odp_ring_st_internal.h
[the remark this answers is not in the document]
Bill-Fischofer-Linaro: OK, the `ODP_ASSERT()` would still be useful for debugging.

platform/linux-generic/include/odp_ring_st_internal.h
```diff
@@ -0,0 +1,118 @@
...
+#include <odp_align_internal.h>
+
+/* Basic ring for single thread usage. Operations must be synchronized by using
+ * locks (or other means), when multiple threads use the same ring. */
+typedef struct {
+	uint32_t head;
+	uint32_t tail;
+	uint32_t mask;
+	uint32_t *data;
+
```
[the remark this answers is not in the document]
Bill-Fischofer-Linaro: OK

platform/linux-generic/include/odp_ring_st_internal.h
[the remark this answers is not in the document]
psavol: The point is that ring size will never be close to 4 billion entries. E.g. currently tail is always max 4096 larger than head. Your example above is based assumption of 4 billion entry ring. Overflow is avoided when ring size if <2 billion, as 32 bit indexes can be still used to calculate number of items correctly.

platform/linux-generic/include/odp_ring_st_internal.h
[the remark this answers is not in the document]
psavol: No since ring size will be much smaller than 4 billion.

platform/linux-generic/odp_pool.c
psavol: Plus this is done only once - in pool create phase.

platform/linux-generic/include/odp_ring_st_internal.h
```diff
@@ -0,0 +1,109 @@
...
+		num = max_num;
+
+	idx = head & mask;
+
+	for (i = 0; i < num; i++) {
+		data[i] = ring->data[idx];
+		idx     = (idx + 1) & mask;
+	}
+
+	ring->head = head + num;
```
Bill-Fischofer-Linaro: Note that neither `head` nor `tail` get masked when doing enqueues or dequeues, so they will traverse the entire 32-bit range of the variable over time. While the ring itself will never hold more than `size` entries, `head` and `tail` will wrap around. When you index into the ring you do so via a mask (`idx = head & mask;`), but calculating the number of elements in the ring doesn't do this, which is why `abs()` is needed.

platform/linux-generic/include/odp_ring_st_internal.h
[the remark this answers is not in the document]
Bill-Fischofer-Linaro: Not true. See the comment below where it's clear that `head` and `tail` are free running and thus will eventually wrap as described above. If these were `uint64_t` variables you could argue that the wraps would take centuries and hence are nothing to worry about, but being `uint32_t` variables this should be expected to happen regularly on heavily used queues. 

`abs()` adds no overhead since compilers treat it as an intrinsic.

platform/linux-generic/include/odp_ring_st_internal.h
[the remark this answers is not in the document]
Bill-Fischofer-Linaro: Please see above discussion. The comment stands.

platform/linux-generic/include/odp_ring_st_internal.h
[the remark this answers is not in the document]
psavol: It's unsigned 32 bit arithmetic, not signed integer arithmetic. Also, tail is always ahead of head, but no more than queue size (e.g. 4k), So, e.g. when tail is 0, and head is 0xffffffff the number of items in ring is 0 - 0xffffffff = 1

```
head            tail
hex    [index], hex    [index], num, num_free
0xfffffffe [6], 0xffffffff [7], 1,   7
0xffffffff [7], 0x00000000 [0], 1,   7
0x00000000 [0], 0x00000001 [1], 1,   7
```

platform/linux-generic/odp_queue.c
psavol:
```
if (ring_st_is_empty(&queue->s.ring_st))

if (ring_st_is_empty(&queue->s.ring_st))

if (ring_st_is_empty(&queue->s.ring_st) == 0)

if (ring_st_is_empty(&queue->s.ring_st))

if (ring_st_is_empty(&queue->s.ring_st))

if (!ring_st_is_empty(&queue->s.ring_st))

if (ring_st_is_empty(&queue->s.ring_st))

if (ring_st_is_empty(&queue->s.ring_st))
```

Find if clause for "ring is not empty" from picture above.

platform/linux-generic/include/odp_ring_st_internal.h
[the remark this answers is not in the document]
Bill-Fischofer-Linaro: OK

Bill-Fischofer-Linaro:
@psavol Reminder: Please rebase so this can be merged.

psavol:
v3: Rebased

PR #494, opened 2018-02-21
[PATCH API-NEXT v6] IPsec TFC implementation

lumag:


include/odp/api/spec/ipsec.h
```diff
@@ -1346,9 +1346,7 @@ int odp_ipsec_in(const odp_packet_t pkt_in[], int num_in,
  * and content of packet data before the IP header is undefined. Use outbound
  * operation parameters to specify the amount of TFC padding appended to
  * the packet during IPSEC transformation. Options can be used also to create
- * TFC dummy packets. Packet data content is ignored in tunnel mode TFC dummy
- * packet creation as tfc_pad_len option defines solely the packet length.
- * In all other cases, payload length for the IPSEC transformation is specified
+ * TFC dummy packets. Payload length for the IPSEC transformation is specified
  * by odp_packet_len() minus odp_packet_l3_offset() plus tfc_pad_len option.
```
Bill-Fischofer-Linaro: Unless it has been parsed, `odp_packet_l3_offset()` is initialized to `ODP_PACKET_OFFSET_INVALID`, so this seems an undue burden. The original wording seems cleaner from an application perspective.

platform/linux-generic/odp_packet.c
```diff
@@ -1286,7 +1286,7 @@ int odp_packet_l2_offset_set(odp_packet_t pkt, uint32_t offset)
 {
 	odp_packet_hdr_t *pkt_hdr = packet_hdr(pkt);
 
-	if (offset >= pkt_hdr->frame_len)
+	if (offset > pkt_hdr->frame_len)
```
Bill-Fischofer-Linaro: This change requires an API change as the spec says relevant offsets must be in the range `0..odp_packet_len(pkt) - 1` .  Same comment for the L3 and L4 changes in this patch.

In theory the validation tests should test these bounds, but as with most parts of the API violations simply result in undefined behavior, so this is an "honor system". Still, we can't violate the spec here without changing the spec.

platform/linux-generic/odp_ipsec.c
```diff
@@ -817,20 +819,30 @@ static ipsec_sa_t *ipsec_in_single(odp_packet_t pkt,
...
 		status->error.proto = 1;
 		goto err;
 	}
 
-	parse_param.proto = state.is_ipv4 ? ODP_PROTO_IPV4 :
-			    state.is_ipv6 ? ODP_PROTO_IPV6 :
-					    ODP_PROTO_NONE;
-	parse_param.last_layer = ipsec_config.inbound.parse_level;
-	parse_param.chksums = ipsec_config.inbound.chksums;
+	if (_ODP_IPPROTO_NO_NEXT == state.ip_next_hdr &&
```
Bill-Fischofer-Linaro: Nit: could use `odp_unlikely()` here.

platform/linux-generic/odp_ipsec.c
```diff
@@ -1165,6 +1167,8 @@ static int ipsec_out_esp(odp_packet_t *pkt,
 				 ipsec_offset + _ODP_ESPHDR_LEN,
 				 ipsec_sa->esp_iv_len,
 				 state->iv + ipsec_sa->salt_length);
+	_odp_packet_set_data(*pkt, esptrl_offset - esptrl.pad_len - tfc_len,
+			     0xa5, tfc_len);
```
Bill-Fischofer-Linaro: Should `0xa5` be a `#define` rather than a "magic number"?

include/odp/api/spec/ipsec.h
lumag: We require that an application sets `l3_offset` for all packet it pushes to IPsec. For TFC dummy packets it resulted in `l3_offset` being set but ignored. Thus I proposed this change. Other solution might be to stop requiring `l3_offset` for TFC dummy packets.

platform/linux-generic/odp_packet.c
lumag: Hmm. We've spent several minutes on this, but nobody reminded of API convention. Should we change the spec here? @psavol what is your opinion?

platform/linux-generic/odp_ipsec.c
lumag: Well, it's just a magic number. Same as 0 would be.

include/odp/api/spec/ipsec.h
Bill-Fischofer-Linaro: A dummy packet is defined only by the fact that it is a dummy and the requested length. No further metadata should be needed or required, so I'd be happy to see that requirement dropped. I view this as a building block for a TBD higher-level configuration service that would simply permit IPsec to be configured to provide a requested level of traffic masking and leave it up to the implementation to decide how best and most efficiently to achieve that via a combination of TFC dummy packets and padding.

platform/linux-generic/odp_ipsec.c
Bill-Fischofer-Linaro: True, but non-zero numbers tend to stand out when used like this. Why pick this number vs. some other? Is there something wrong with using zeros?

lumag:
Postponed till #500 is merged

lumag:
@bala-manoharan @Bill-Fischofer-Linaro @muvarov @psavol 
I've updated TFC implementations. Could you please review.

test/validation/api/ipsec/ipsec.c
```diff
@@ -604,10 +604,14 @@ static int ipsec_send_out_one(const ipsec_test_part *part,
 			hdr_len = part->out[0].pkt_out->l3_offset;
 			CU_ASSERT_FATAL(hdr_len <= sizeof(hdr));
 			memcpy(hdr, part->out[0].pkt_out->data, hdr_len);
-		} else {
+		} else if (part->pkt_in->l3_offset != ODP_PACKET_OFFSET_INVALID) {
```
Bill-Fischofer-Linaro: Checkpatch flags this:
```
#22: FILE: test/validation/api/ipsec/ipsec.c:607:
+		} else if (part->pkt_in->l3_offset != ODP_PACKET_OFFSET_INVALID) {
total: 0 errors, 1 warnings, 0 checks, 210 lines checked
NOTE: Ignored message types: AVOID_EXTERNS BIT_MACRO COMPARISON_TO_NULL DEPRECATED_VARIABLE NEW_TYPEDEFS PREFER_PRINTF PREFER_SCANF SPLIT_STRING SSCANF_TO_KSTRTO VOLATILE
0003-validation-ipsec-verify-TFC-dummy-packet-generation.patch has style problems, please review.
```

platform/linux-generic/odp_ipsec.c
```diff
@@ -1327,15 +1365,30 @@ static ipsec_sa_t *ipsec_out_single(odp_packet_t pkt,
...
-	state.ip_offset = odp_packet_l3_offset(pkt);
-	ODP_ASSERT(ODP_PACKET_OFFSET_INVALID != state.ip_offset);
-
-	state.ip = odp_packet_l3_ptr(pkt, NULL);
-	ODP_ASSERT(NULL != state.ip);
-
 	ipsec_sa = _odp_ipsec_sa_use(sa);
 	ODP_ASSERT(NULL != ipsec_sa);
 
+	if (opt->flag.tfc_dummy) {
```
Bill-Fischofer-Linaro: An `odp_unlikely()` would be appropriate for this `if`.

platform/linux-generic/odp_ipsec.c
Bill-Fischofer-Linaro: I'd still prefer to see this at least commented as to why `0xa5` is used as a pad character.

lumag:
@Bill-Fischofer-Linaro updated

muvarov:
this needs one more rebase.

lumag:
@muvarov rebased.

muvarov:
Merged.

PR #503, opened 2018-02-28
[PATCH v1] shippable: pass CI=true to tests

lumag:
sudo will strip most variables from host environment, including precious
CI=true setting, which we use to skip some obscure test results (like
traffic mngr) if tests are running in non-isolated environment. So
enforce CI=true when calling tests.


muvarov:
Merged.

PR #506, opened 2018-03-01
[PATCH v2] configure.ac: update API and .so version to v1.18.0.0

muvarov:

### DIFF
--- a/configure.ac
+++ b/configure.ac
@@ -3,7 +3,7 @@ AC_PREREQ([2.5])
 # Set correct API version
 ##########################################################################
 m4_define([odpapi_generation_version], [1])
-m4_define([odpapi_major_version], [17])
+m4_define([odpapi_major_version], [18])
 m4_define([odpapi_minor_version], [0])
 m4_define([odpapi_point_version], [0])
 m4_define([odpapi_version],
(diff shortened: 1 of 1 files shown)


configure.ac
```diff
@@ -36,10 +36,10 @@ AM_SILENT_RULES([yes])
 # 3. if interfaces were removed, then use C+1:0:0
 ##########################################################################
 
-ODP_LIBSO_VERSION=117:0:0
+ODP_LIBSO_VERSION=118:0:0
 AC_SUBST(ODP_LIBSO_VERSION)
 
-ODPHELPER_LIBSO_VERSION=113:2:1
+ODPHELPER_LIBSO_VERSION=118:0:0
```
lumag: Helper is compatible, as far as I understand, so we should not bump major version here.

Bill-Fischofer-Linaro: Perhaps, but as this is a new LTS release might be best to sync everything? Apps are going to be recompiling for Tiger Moth anyway if they're coming from Monarch.

lumag: @Bill-Fischofer-Linaro soname is not about releases, but about compatibility, so I vote for 114:0:2 for odphelper.

muvarov:
v2: change helper .so name to 14:0:2

lumag:

muvarov:
Merged.

PR #507, opened 2018-03-02
[PATCH v1] validation: pktio: remove pktio_test_send_failure test

matiaselo:
pktio_test_send_failure() tries to transmit a packet which is larger
than odp_pktout_maxlen. ODP API doesn't guarantee what happens when one
tries to send an oversized packet, so remove the test.


muvarov:
Merged.

PR #510, opened 2018-03-05
[PATCH v3] linux-gen: fix crypto merge aes_gmac_gen_init/aes_gmac_check_init

muvarov:


muvarov:
for now we invoke aes_gmac_gen_init() with following parameters:
  Test: crypto_test_gen_alg_aes_gmac ...auth_iv.length 12, session->p.cipher_iv.length 0
  Test: crypto_test_gen_alg_aes_gmac_ovr_iv ...auth_iv.length 12, session->p.cipher_iv.length 0

As I understand we need new test case with ODP_CIPHER_ALG_NULL changed to some alg.

Bill-Fischofer-Linaro:
If @lumag is OK with this, I'm OK with it. But he should be primary reviewer on it.

lumag:
The problem is that 0/unset is also a valid setting for OpenSSL.

muvarov:
Merging.

PR #511, opened 2018-03-05
[PATCH v1] build: make so numbering to be tied to ODP versions

lumag:
We do not make any promises on ABI backwards compatibility, we have been
using -version-number instead of -version-info for ages, so affirm this
approach. Drop misleading comment and tie SO version numbers to ODP
relase numbers.


muvarov:
Merged.

PR #512, opened 2018-03-06
[PATCH v1] Improve debug prints

psavol:
Corrected system info print out and add debug information for packet IO and queue startup.


configure.ac
```diff
@@ -240,7 +240,7 @@ AS_IF([test "${with_platform}" = "linux-generic"],
       [m4_include([./platform/linux-generic/m4/configure.m4])],
       [AC_MSG_ERROR([UNSUPPORTED PLATFORM: ${with_platform}])])
 
-AC_DEFINE_UNQUOTED([IMPLEMENTATION_NAME], ["$IMPLEMENTATION_NAME"],
+AC_DEFINE_UNQUOTED([ODP_IMPLEMENTATION_NAME], ["$ODP_IMPLEMENTATION_NAME"],
```
muvarov: This change is not clear. If we wanted to keep ODP_ prefix for api. Than might be not needed to do such renames.

psavol: #define ODP_VERSION_IMPL_STR \
	ODP_VERSION_TO_STR(IMPLEMENTATION_NAME) " " \
	ODP_VERSION_TO_STR(ODP_VERSION_API_GENERATION) "." \
	ODP_VERSION_TO_STR(ODP_VERSION_API_MAJOR) "." \
	ODP_VERSION_TO_STR(ODP_VERSION_API_MINOR) "-" \
	ODP_VERSION_TO_STR(ODP_VERSION_IMPL) " (v" \
	ODP_VERSION_TO_STR(ODP_VERSION_API_GENERATION) "." \
	ODP_VERSION_TO_STR(ODP_VERSION_API_MAJOR) "." \
	ODP_VERSION_TO_STR(ODP_VERSION_API_MINOR) ") " \
	ODP_VERSION_TO_STR(GIT_HASH)

Everything else used for ODP implementation string generation are ODP_ prefixed, except IMPLEMENTATION_NAME and GIT_HASH. When we define new macros or makefile variables (especially when visible in C code) those should be ODP_ prefixed, so that it's easy to understand/find where those come from (from our configure/makefiles instead of e.g. automatic variables of automake) and we avoid potential name space clash with other SW.

muvarov:
Merged.

PR #514, opened 2018-03-08
[PATCH v1] configure.ac: update version to v1.18.0.1

muvarov:
From now we don't need to update any so numbers manually. We just need update version inside configure.ac


muvarov:
Merged.

PR #515, opened 2018-03-09
[PATCH v2] Slightly change build order

lumag:


Makefile.am
```diff
@@ -12,11 +12,11 @@ SUBDIRS = \
...
 	  $(PLATFORM_DIR) \
 	  helper \
+	  test \
+	  $(PLATFORM_TEST_DIR) \
 	  helper/test \
 	  doc \
-	  example . \
-	  test \
-	  $(PLATFORM_TEST_DIR)
+	  example
```
muvarov: looked more detail in this change. ./test/performance/odp_l2fwd_run.sh needs example/odp_generator to run. If you run make check -j $(npoc) and code was not compiled before, then  odp_l2fwd_run.sh will be executed before odp_generator binary is build. That is why "." is here.

muvarov: now I looked to second path and this comment is not valid :)

muvarov:
Merged.

PR #516, opened 2018-03-09
[PATCH v5] linux-gen: netmap: optimal ring configuration for VALE

jmtilli:
Configure ring optimally for VALE. On my test laptop, this increases
odp_l2fwd performance from 1.9 MPPS to 2.3 MPPS, so it gives over 20%
more performance.

### DIFF
--- a/config/odp-linux-generic.conf
+++ b/config/odp-linux-generic.conf
@@ -31,6 +31,15 @@ pktio_dpdk: {
 	}
 }
 
+# netmap pktio options
+pktio_netmap: {
+	# Interface specific options
+	virt: {
+		nr_rx_slots = 0
+		nr_tx_slots = 0
+	}
+}
+
 queue_basic: {
 	# Maximum queue size. Value must be a power of two.
 	max_queue_size = 8192
--- a/platform/linux-generic/pktio/netmap.c
+++ b/platform/linux-generic/pktio/netmap.c
@@ -27,6 +27,8 @@
 #include <odp_classification_datamodel.h>
 #include <odp_classification_inlines.h>
 #include <odp_classification_internal.h>
+#include <odp_libconfig_internal.h>
+
 
 #include <inttypes.h>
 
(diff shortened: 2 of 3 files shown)


jmtilli:
Thinking about this again, perhaps ODP may require greater nr_tx_slots, as the TX side of ODP is non-blocking whereas the TX side of the programs where I originally optimized the VALE ring configuration is blocking. This may explain the test failures.

platform/linux-generic/pktio/netmap.c
```diff
@@ -526,6 +526,10 @@ static int netmap_start(pktio_entry_t *pktio_entry)
 
 	base_desc.self = &base_desc;
 	base_desc.mem = NULL;
+	if (pktio_entry->s.pkt_nm.is_virtual) {
+		base_desc.req.nr_rx_slots = 256;
+		base_desc.req.nr_tx_slots = 64;
```
muvarov: This is good to be placed to odp config.

jmtilli: Thanks for the review! Can you explain a bit more on how to find information about ODP config and how to use it? I guess making this configurable could be a good addition. It probably won't make it to Tiger Moth LTS, though.

matiaselo: The runtime configuration support was added in commit 534667daec58ce8b2e417709e788934bea6f8f09. You can take a look at e.g. dpdk pktio as it uses libconfig to configure the number of rx/tx descriptors.

According to netmap_user.h, NM_OPEN_RING_CFG flag should be used when passing these options, so I'm wondering if this is really working.

matiaselo: This could be a common option for all netmap device types and not only for virtual devices. In the configuration file the default value should probably be 0 -> use default values defined by netmap.

jmtilli: My understanding is that these affect the operation of virtual devices only, i.e. that the is_virtual check is there only for safety reasons (I didn't have 40GbE NICs at my disposal when writing the patch and didn't want to take the risk of reducing their performance even though I was 99% sure it doesn't affect them).

But thanks for the information! When I have time, I will take a look at the configuration support. The performance results are repeatable, and I could observe genuine performance increase, so I suspect they are working.

muvarov:
Waiting on this patch for more testing.

muvarov:
@matiaselo @jmtilli is there any update for this patch?

jmtilli:
Now the patch has been finalized and tested. Can somebody please review and perhaps merge?

config/odp-linux-generic.conf
```diff
@@ -31,6 +31,19 @@ pktio_dpdk: {
...
+pktio_netmap: {
+	# Default options
+	nr_rx_slots = 0
+	nr_tx_slots = 0
+
+	# Interface specific options
+	eth0: {
+		nr_rx_slots = 0
+		nr_tx_slots = 0
+	}
```
matiaselo: Since these options are only significant for virtual interfaces I would simplify the config and make it clear that these options are only relevant for virtual interfaces. It's probably enough to have a single value for all virtual interfaces, so the interface specific options are not needed.

E.g.
```
pktio_netmap: {
	virt: {
		nr_rx_slots = 0
		nr_tx_slots = 0
	}
}
```

muvarov:
Merged.

PR #517, opened 2018-03-10
[PATCH v1] Fix/odp dpdk.m4

joseppc:


.travis.yml
```diff
@@ -222,10 +222,11 @@ install:
             echo CONFIG_RTE_TOOLCHAIN_${DPDKCC^^}=y >> .config
             popd
             make build O=${TARGET} EXTRA_CFLAGS="-fPIC $DPDK_CFLAGS" CROSS="$DPDK_CROSS" CC="$CC" HOSTCC=gcc -j $(nproc)
+            make install O=${TARGET} DESTDIR=${TARGET}
```
Bill-Fischofer-Linaro: If I understand correctly, if we discover that DPDK has been installed, we use it, but if not we download and build a private copy of it. Why would ODP try to install DPDK? Shouldn't that be up to the sysadmin?  The user building ODP may not have the required privileges to install anything. Or are we just talking about a "private installation" here?

Bill-Fischofer-Linaro: OK, that makes sense. Thanks.

.travis.yml
```diff
@@ -195,7 +195,7 @@ install:
             LIBDPDKEXT=a
            fi
            DPDK_TARGET="${DPDK_TARGET}gcc"
-           if [ ! -f "dpdk/${TARGET}/lib/libdpdk.$LIBDPDKEXT" ]; then
+           if [ ! -f "dpdk/${TARGET}/usr/local/lib/libdpdk.$LIBDPDKEXT" ]; then
```
muvarov: that should not be correct.  dpdk is installed to $HOME/dpdk/ dir. And there lib/ and etc folders are created. Also dpdk directory is cached by travis to not rebuild it each time.

m4/odp_dpdk.m4
```diff
@@ -120,7 +120,7 @@ AS_IF([test "x$1" = "xsystem"], [dnl
     fi
     DPDK_LIB_PATH=`AS_DIRNAME(["$DPDK_LIB_PATH"])`
 ], [dnl
-    DPDK_CPPFLAGS="-isystem $1/include"
+    DPDK_CPPFLAGS="-isystem $1/include/dpdk"
```
lumag: Should we then switch to including `<dpdk/rteFOO.h>` headers instead. This would allow us to simplify `DPDK_CPPFLAGS`.

.travis.yml
joseppc: This shouldn't change anything. DPDK will still be installed in the home directory (DESTDIR is still $HOME/dpdk/$TARGET), the difference is that now you will be testing with an installed version of DPDK rather than the directory with build artifacts.

Maybe we should set "prefix" when installing to something more meaningful like dpdk-17.08 instead of the default /usr/local, and also set DESTDIR to $HOME. This way the check done earlier can just look for the presence of this directory rather than grepping a file inside the build directory. This could also allow us to have several versions of DPDK installed which we could test ODP against. If disk space in Travis is not an issue, that is.

m4/odp_dpdk.m4
joseppc: We could do that too, but we would still have to set the correct include path anyway, so it doesn't simplify `DPDK_CPPFLAGS` much.
DPDK's example code doesn't prefix the headers, though.

lumag: It does not, as it is built with non-installed tree.
Another question, should we support non-installed DPDK tree?

joseppc: Good question. I think building against a non-installed tree is not right, but this change may break someone's workflow, which made me realize `scripts/build-pktio-dpdk` needed to be updated as well.

muvarov: ./scripts/build-pktio-dpdk does make install T=${TARGET} EXTRA_CFLAGS="-fPIC".  Is it not enough for install?

joseppc: According to the DPDK documentation:
> install T=       configure, build and install a target in DESTDIR
I get an error if you don't provide `DESTDIR`, when doing so:
> Installation cannot run with T defined and DESTDIR undefined

`DESTDIR` may not be needed, but since it will try to install kernel modules, it is probably good to provide a staging directory somewhere where the user can write to.

lumag:

muvarov:
Merged.

PR #518, opened 2018-03-12
[PATCH v2] shippable: point shm directory to huge pages

muvarov:
/dev/shm is limited under docker by default to 64M, so
we can not use it for tests which require more memory.
Instead of doing this just point to mounted huge pages
directory.

### DIFF
--- a/.shippable.yml
+++ b/.shippable.yml
@@ -33,7 +33,8 @@ build:
     options:
 
   ci:
-    - echo 1000 | sudo tee /proc/sys/vm/nr_hugepages
+    # 200M of 2048 pages
+    - echo 102400 | sudo tee /proc/sys/vm/nr_hugepages
     - sudo mkdir -p /mnt/huge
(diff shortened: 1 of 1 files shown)


muvarov:
hm. ok. will send v2 for this.

echo 1000 | sudo tee /proc/sys/vm/nr_hugepages 0s  
1000

FAIL cuckootable (exit status: 1)
FAILURE detected at ./helper/test/odpthreads_as_pthreads.trs
_ishm.c:470:create_file():ftruncate failed: fd=3, err=Invalid argument.
_ishm.c:913:_odp_ishm_reserve():_ishm_reserve failed.
_ishm.c:1299:_odp_ishm_address():Request for address on an invalid block
odp_init.c:89:odp_init_global():ODP thread init failed.
HW time counter freq: 100000000 hz

muvarov:
Closing this. Problem is in other place.

PR #523, opened 2018-03-13
[PATCH v1] Queue and scheduler config file options

psavol:
Added first queue and scheduler configuration options into the configuration file. Queue size and scheduler internal queue spreading can be configured. Spread configuration enables tuning scheduler for small thread counts.

### DIFF
--- a/config/odp-linux-generic.conf
+++ b/config/odp-linux-generic.conf
@@ -30,3 +30,8 @@ pktio_dpdk: {
 		rx_drop_en = 1
 	}
 }
+
+queue_basic: {
+	# Default queue size. Value must be a power of two.
+	default_queue_size = 4096
+}
--- a/config/odp-linux-generic.conf
+++ b/config/odp-linux-generic.conf
@@ -38,3 +38,12 @@ queue_basic: {
 	# Default queue size. Value must be a power of two.
 	default_queue_size = 4096
 }
+
+sched_basic: {
+	# Priority level spread. Each priority level is spread into multiple
+	# scheduler internal queues. A higher spread value typically improves
+	# parallelism and thus is better for high thread counts, but causes
+	# uneven service level for low thread counts. Typically, optimal
+	# value is the number of threads using the scheduler.
+	prio_spread = 4
+}
--- a/config/odp-linux-generic.conf
+++ b/config/odp-linux-generic.conf
@@ -32,6 +32,9 @@ pktio_dpdk: {
 }
 
 queue_basic: {
+	# Maximum queue size. Value must be a power of two.
+	max_queue_size = 8192
+
 	# Default queue size. Value must be a power of two.
 	default_queue_size = 4096
 }
--- a/platform/linux-generic/odp_queue_basic.c
+++ b/platform/linux-generic/odp_queue_basic.c
@@ -38,7 +38,7 @@
 #include <inttypes.h>
 
 #define MIN_QUEUE_SIZE 8
-#define MAX_QUEUE_SIZE CONFIG_QUEUE_SIZE
+#define MAX_QUEUE_SIZE (1 * 1024 * 1024)
 
 static int queue_init(queue_entry_t *queue, const char *name,
 		      const odp_queue_param_t *param);
@@ -64,11 +64,11 @@ static int queue_capa(odp_queue_capability_t *capa, int sched)
 	/* Reserve some queues for internal use */
 	capa->max_queues        = ODP_CONFIG_QUEUES - NUM_INTERNAL_QUEUES;
 	capa->plain.max_num     = capa->max_queues;
-	capa->plain.max_size    = MAX_QUEUE_SIZE;
+	capa->plain.max_size    = queue_glb->config.max_queue_size;
 	capa->plain.lockfree.max_num  = queue_glb->queue_lf_num;
 	capa->plain.lockfree.max_size = queue_glb->queue_lf_size;
 	capa->sched.max_num     = capa->max_queues;
-	capa->sched.max_size    = MAX_QUEUE_SIZE;
+	capa->sched.max_size    = queue_glb->config.max_queue_size;
 
 	if (sched) {
 		capa->max_ordered_locks = sched_fn->max_ordered_locks();
@@ -87,7 +87,7 @@ static int read_config_file(queue_global_t *queue_glb)
 
 	ODP_PRINT("Queue config:\n");
 
-	str = "queue_basic.default_queue_size";
+	str = "queue_basic.max_queue_size";
 	if (!_odp_libconfig_lookup_int(str, &val)) {
 		ODP_ERR("Config option '%s' not found.\n", str);
 		return -1;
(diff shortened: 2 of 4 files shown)


lumag:
@psavol just a syntax comment:
Wouldn't it be better to have `sched.basic.FOO` option names. This will allow us to have the following config:
```
sched {
  select = "basic";
  basic { 
   FOO = "BAR";
  }
}
```

psavol:
I would not mix options of different schedulers. So, scheduler selection would be on top level and each scheduler would have their own set of options. E.g. basic scheduler is not interested about how it was selected. Selection is part of another module - schedule_if.c currently, but can be moved around without basic scheduler module needing to care about that.

```
scheduler = basic

sched_basic: {
   zyx = a
   foo = 111
}

sched_sp: {
  sfrf = 7
  anf = 438
}

...

```

lumag:
@psavol so, less hierarchical approach. Fine with me.
Reviewed-by: Dmtiry Eremin-Solenikov

config/odp-linux-generic.conf
```diff
@@ -38,3 +38,12 @@ queue_basic: {
 	# Default queue size. Value must be a power of two.
 	default_queue_size = 4096
 }
+
+sched_basic: {
+	# Priority level spread. Each priority level is spread into multiple
+	# scheduler internal queues. A higher spread value typically improves
+	# parallelism and thus is better for high thread counts, but causes
+	# uneven service level for low thread counts. Typically, optimal
+	# value is the number of threads using the scheduler.
```
Bill-Fischofer-Linaro: Does this mean that the optimal value is the number of worker threads? In that case, why not make this a default? I thought there was an effective limit on `prio_spread` in that setting it too high would be counterproductive. If an application has 128 threads are you really suggesting that `prio_spread = 128` should be used?

platform/linux-generic/odp_schedule_basic.c
```diff
@@ -41,15 +42,18 @@ ODP_STATIC_ASSERT((ODP_SCHED_PRIO_NORMAL > 0) &&
 /* Number of scheduling groups */
 #define NUM_SCHED_GRPS 32
 
-/* Priority queues per priority */
-#define QUEUES_PER_PRIO  4
+/* Maximum priority queue spread */
+#define MAX_SPREAD 4
```
Bill-Fischofer-Linaro: This limit seems to contradict the recommendation mentioned earlier.

platform/linux-generic/odp_schedule_basic.c
```diff
@@ -60,14 +64,10 @@ ODP_STATIC_ASSERT((ODP_SCHED_PRIO_NORMAL > 0) &&
...
- * equal to ODP_CONFIG_QUEUES / QUEUES_PER_PRIO, so that it can hold all
- * queues in the worst case. */
-#define PRIO_QUEUE_RING_SIZE  (ODP_CONFIG_QUEUES / QUEUES_PER_PRIO)
-
-/* Mask for wrapping around priority queue index */
-#define RING_MASK  (PRIO_QUEUE_RING_SIZE - 1)
+/* Maximum priority queue ring size. A ring must be large enough to store all
+ * queues in the worst case (all queues are scheduled, have the same priority
+ * and no spreading). */
+#define MAX_RING_SIZE ODP_CONFIG_QUEUES
```
Bill-Fischofer-Linaro: Do we require `ODP_CONFIG_QUEUES` to be a power of two? If not should this be rounded up here?

platform/linux-generic/odp_schedule_basic.c
```diff
@@ -221,11 +226,41 @@ static __thread sched_local_t sched_local;
...
+static int read_config_file(sched_global_t *sched)
+{
+	const char *str;
+	int val = 0;
+
+	ODP_PRINT("Scheduler config:\n");
+
+	str = "sched_basic.prio_spread";
+	if (!_odp_libconfig_lookup_int(str, &val)) {
+		ODP_ERR("Config option '%s' not found.\n", str);
```
Bill-Fischofer-Linaro: Same comments here about requiring fields to be present. We should use built-in defaults if they are not.

platform/linux-generic/odp_schedule_basic.c
```diff
@@ -221,11 +226,41 @@ static __thread sched_local_t sched_local;
...
+		return -1;
+	}
+
+	if (val > MAX_SPREAD || val < MIN_SPREAD) {
+		ODP_ERR("Bad value %s = %u\n", str, val);
```
Bill-Fischofer-Linaro: Saturation is better than an error here.

platform/linux-generic/odp_queue_basic.c
```diff
@@ -75,6 +79,34 @@ static int queue_capa(odp_queue_capability_t *capa, int sched)
...
+	if (!_odp_libconfig_lookup_int(str, &val)) {
+		ODP_ERR("Config option '%s' not found.\n", str);
+		return -1;
+	}
+
+	val_u32 = val;
+
+	if (val_u32 > MAX_QUEUE_SIZE || val_u32 < MIN_QUEUE_SIZE ||
+	    !CHECK_IS_POWER2(val_u32)) {
+		ODP_ERR("Bad value %s = %u\n", str, val_u32);
```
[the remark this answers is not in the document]
psavol: Out of range or not a power of 2 are bad configuration options. User may have mistakenly given wrong or corrupted file, so it's better to stop right there on the first bad value found, than try to adapt to it. User must understand what a value does and carefully choose a value for it. He would not benefit if we round down e.g. 700 to 8, when he actually meant 7 (but typoed 700).

Config file is part of an implementation. It's content and format is implementation specific. Config file content is not portable. User will start experiments with each implementation from default config and change config only if needed.

platform/linux-generic/odp_queue_basic.c
```diff
@@ -104,6 +136,11 @@ static int queue_init_global(void)
 		queue->s.handle = queue_from_index(i);
 	}
 
+	if (read_config_file(queue_glb)) {
+		odp_shm_free(shm);
+		return -1;
```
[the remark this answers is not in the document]
psavol: The function uses _odp_libconfig_lookup_int() which reads user's config file first (when present) and otherwise the built in default config (header) file. So, there's always a config file.

config/odp-linux-generic.conf
psavol: *Typically*, the number of threads calling scheduler is optimal. It really needs to be measured in each case. Four seems to be good default for most cases. Current max is 8, so 128 would just error due to bad value.

Application decides number of threads calling scheduler, so implementation cannot "default" to that. Currently, it defaults to 4.

platform/linux-generic/odp_schedule_basic.c
psavol: On this patch level, max is still 4. A later patch increases it to 8.

platform/linux-generic/odp_schedule_basic.c
psavol: Ring size needs to be power of 2. It's optimal for memory consumption and performance that ring size is not overly large. So, it's optimal to keep these two defines in sync.

platform/linux-generic/odp_schedule_basic.c
psavol: _odp_libconfig_lookup_int() first checks user provided config file and falls back to the default config file. There's always a config file with this option set.

platform/linux-generic/odp_schedule_basic.c
psavol: Error is better. Saturation would cause e.g. a file with garbage values to set everything silently to max/min - which is not typically what user expects to happen. Exact values from config file are either  used or not - otherwise user can not be sure what config he's actually running.

muvarov:
Merged

PR #525, opened 2018-03-13
[PATCH v1] shippable: do not use huge pages

muvarov:
huge pages under docker are trickly also /dev/shm/ is limited
to 64Mb. Use different folder for shared memory files.


muvarov:
Merged. Now shippable  should be fine.

PR #526, opened 2018-03-14
[PATCH v1] test: mmap_vlan_ins: wait before removing test interfaces

matiaselo:
Removing the test interfaces while the applications are still running may
cause a test application instance to not close properly.


muvarov:
Merged.

PR #527, opened 2018-03-14
[PATCH v6] WIP: work around on aarch64 runtime fails

muvarov:
Quick workaround to make Shippable happy with calling tests on thunder-x. Some more clean patch is needed.

### DIFF
--- a/.shippable.yml
+++ b/.shippable.yml
@@ -34,18 +34,18 @@ build:
 
   ci:
     - mkdir -p $HOME/odp-shmdir
+    - export CI=true ODP_SHM_DIR=$HOME/odp-shmdir ODP_TEST_OUT_XML=yes
     - ./bootstrap
-    - if [ "${CC#clang}" != "${CC}" ] ; then export CXX="${CC/clang/clang++}"; fi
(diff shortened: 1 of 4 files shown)


.shippable.yml
```diff
@@ -36,6 +36,7 @@ build:
     - mkdir -p $HOME/odp-shmdir
     - ./bootstrap
     - if [ "${CC#clang}" != "${CC}" ] ; then export CXX="${CC/clang/clang++}"; fi
+    - if [ "${CC#clang}" != "${CC}" ] ; then export CFLAGS="-fPIC"; fi
```
lumag: NAK, this hides the issue instead of fixing it.

muvarov: yep, that is why patches are WIP.  Will wait for Shippable tests results.

muvarov:
ok, now both tests hang on the same tests. Looks like something else is there. I want to figure out where all hungs are and then think about clean patch.

muvarov:
@lumag with that changes looks like shippable is happy.

semihalf-kardach-stanislaw:
@muvarov If you inspect the generated assembly can you see any stores generated in between `ldxp` and `stxp`? Any such store will reset the monitor and cause the `stxp` to fail.

muvarov:
Close this.

PR #529, opened 2018-03-15
[PATCH v6] tests: add IPsec performance test, improve crypto perf test

lumag:
Use this tool to measure IPsec performance with different algorithms
For now only outbound direction is supported.

### DIFF
--- a/test/performance/odp_ipsec.c
+++ b/test/performance/odp_ipsec.c
@@ -122,6 +122,12 @@ typedef struct {
 	 * Specified through -t argument.
 	 */
 	int tunnel;
+
+	/*
+	 * Use AH transformation.
+	 * Specified through -u argument.
+	 */
+	int ah;
 } ipsec_args_t;
 
 /*
@@ -778,6 +784,7 @@ static void usage(char *progname)
 	       "  -s, --schedule       Use scheduler for completion events.\n"
 	       "  -p, --poll           Poll completion queue for completion events.\n"
 	       "  -t, --tunnel         Use tunnel-mode IPsec transformation.\n"
+	       "  -u, --ah             Use AH transformation instead of ESP.\n"
 	       "  -h, --help	       Display help and exit.\n"
 	       "\n");
 }
@@ -798,10 +805,11 @@ static void parse_args(int argc, char *argv[], ipsec_args_t *cargs)
 		{"poll", no_argument, NULL, 'p'},
 		{"schedule", no_argument, NULL, 's'},
 		{"tunnel", no_argument, NULL, 't'},
+		{"ah", no_argument, NULL, 'u'},
 		{NULL, 0, NULL, 0}
 	};
 
-	static const char *shortopts = "+a:c:df:hi:m:nl:spt";
+	static const char *shortopts = "+a:c:df:hi:m:nl:sptu";
 
 	/* let helper collect its own arguments (e.g. --odph_proc) */
 	odph_parse_options(argc, argv, shortopts, longopts);
@@ -813,6 +821,7 @@ static void parse_args(int argc, char *argv[], ipsec_args_t *cargs)
 	cargs->payload_length = 0;
 	cargs->alg_config = NULL;
 	cargs->schedule = 0;
+	cargs->ah = 0;
 
 	opterr = 0; /* do not issue errors on helper options */
 
@@ -859,6 +868,9 @@ static void parse_args(int argc, char *argv[], ipsec_args_t *cargs)
 			break;
 		case 't':
 			cargs->tunnel = 1;
+			break;
+		case 'u':
+			cargs->ah = 1;
 			break;
 		default:
 			break;
@@ -1011,6 +1023,10 @@ int main(int argc, char *argv[])
 		for (i = 0;
 		     i < (sizeof(algs_config) / sizeof(ipsec_alg_config_t));
 		     i++) {
+			if (cargs.ah &&
+			    algs_config[i].crypto.cipher_alg !=
+			    ODP_CIPHER_ALG_NULL)
+				continue;
 			run_measure_one_config(&cargs, algs_config + i);
 		}
 	}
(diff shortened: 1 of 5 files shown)


test/performance/odp_ipsec.c
```diff
@@ -0,0 +1,1035 @@
+/* Copyright (c) 2015-2018, Linaro Limited
```
muvarov: year

test/performance/odp_ipsec.c
```diff
@@ -0,0 +1,1035 @@
...
+static void
+print_result(ipsec_args_t *cargs,
+	     unsigned int payload_length,
+	     ipsec_alg_config_t *config,
+	     crypto_run_result_t *result)
+{
+	unsigned int throughput;
+
+	throughput = (1000000.0 / result->elapsed) * payload_length / 1024;
+	printf(REPORT_LINE,
```
muvarov: No need to hide formated string there under define. Anyway it's used only once.

test/performance/odp_ipsec.c
```diff
@@ -0,0 +1,1035 @@
...
+
+	/* Initialize parameters block */
+	memset(&param, 0, sizeof(param));
+	param.num_sa = 1;
+	param.num_opt = 0;
+	param.sa = &sa;
+
+	fill_time_record(&start);
+
+	while ((packets_sent < cargs->iteration_count) ||
```
muvarov: this function has to be spitted up to 2 or 3 small functions. with removing declaration variables in the middles. also  while cycles will be more easy to read.

test/performance/odp_ipsec.c
```diff
@@ -0,0 +1,1035 @@
...
+			}
+		}
+
+		if (cargs->schedule || cargs->poll) {
+			odp_event_t ev;
+			odp_crypto_packet_result_t result;
+			odp_packet_t out_pkt;
+
+			if (cargs->schedule)
+				ev = odp_schedule(NULL,
```
muvarov: for performance reason might be odp_schedule_multi() is better.

test/performance/odp_ipsec.c
```diff
@@ -0,0 +1,1035 @@
...
+						  ODP_SCHED_NO_WAIT);
+			else
+				ev = odp_queue_deq(out_queue);
+
+			while (ev != ODP_EVENT_INVALID) {
+				out_pkt = odp_crypto_packet_from_event(ev);
```
muvarov: add does not use any other events.  But might be it's also reasonable to add odp_event_type() check here.

test/performance/odp_ipsec.c
```diff
@@ -0,0 +1,1035 @@
...
+				if (cargs->reuse_packet)
+					pkt = out_pkt;
+				else
+					odp_packet_free(out_pkt);
+			}
+		}
+
+		if (cargs->schedule || cargs->poll) {
+			odp_event_t ev;
+			odp_crypto_packet_result_t result;
```
muvarov: do not name with the same name as function arg parameter. Nothing bad but different names are easy to maintain.

test/performance/odp_ipsec.c
```diff
@@ -0,0 +1,1035 @@
...
+
+			if (cargs->schedule)
+				ev = odp_schedule(NULL,
+						  ODP_SCHED_NO_WAIT);
+			else
+				ev = odp_queue_deq(out_queue);
+
+			while (ev != ODP_EVENT_INVALID) {
+				out_pkt = odp_crypto_packet_from_event(ev);
+				odp_crypto_result(&result, out_pkt);
```
muvarov: check for result is missing.

test/performance/odp_ipsec.c
```diff
@@ -0,0 +1,1035 @@
...
+							  ODP_SCHED_NO_WAIT);
+				else
+					ev = odp_queue_deq(out_queue);
+			};
+		}
+	}
+
+	fill_time_record(&end);
+
+	{
```
muvarov: move this also to separate function

test/performance/odp_ipsec.c
```diff
@@ -0,0 +1,1035 @@
...
+	parse_args(argc, argv, &cargs);
+
+	/* Init ODP before calling anything else */
+	if (odp_init_global(&instance, NULL, NULL)) {
+		app_err("ODP global init failed.\n");
+		exit(EXIT_FAILURE);
+	}
+
+	/* Init this thread */
+	odp_init_local(instance, ODP_THREAD_WORKER);
```
muvarov: return code check is needed.

test/performance/odp_ipsec.c
```diff
@@ -0,0 +1,1035 @@
...
+	config.inbound.chksums.all_chksum = 0;
+	config.outbound.all_chksum = 0;
+
+	odp_queue_param_init(&qparam);
+	if (cargs.schedule) {
+		qparam.type = ODP_QUEUE_TYPE_SCHED;
+		qparam.sched.prio  = ODP_SCHED_PRIO_DEFAULT;
+		qparam.sched.sync  = ODP_SCHED_SYNC_PARALLEL;
+		qparam.sched.group = ODP_SCHED_GROUP_ALL;
+		out_queue = odp_queue_create("ipsec-out", &qparam);
```
muvarov: check creation.

muvarov: see it bellow. skip this comment.

test/performance/odp_ipsec.c
lumag: I've split this into sync and async cases. Splitting it further does not make  real sense, as it will lower performance.

test/performance/odp_ipsec.c
lumag: Yes, I thought about it. I might give it a try, however it will complicate code.

test/performance/odp_ipsec.c
lumag: Not under performance-critical part.

test/performance/odp_ipsec.c
lumag: Moved to calling code.

test/performance/odp_ipsec.c
lumag: ack

lumag:
@muvarov updated PR fixing most of your comments

lumag:
@muvarov @psavol I've updated PR: IPsec out performance test, updated crypto performance test and added small updates to implementation.

muvarov:
Merged.

PR #530, opened 2018-03-15
[PATCH v1] shippable: disable abi compat mode tests on aarch64

muvarov:
disable tests which fails under our docker runs with
native clang 4.8


muvarov:
@lumag @Bill-Fischofer-Linaro are you ok to apply this before we fixed other builds?

muvarov:
Shippable is passed. But marged status to unstable. I will rerun this test one more time to double check that it works.

muvarov:
Merged this.

PR #531, opened 2018-03-16
[PATCH v1] allow checkpatch to fail

muvarov:


muvarov:
ah, this  has to be for odp-dpdk. Closed.

PR #532, opened 2018-03-16
[PATCH v3] New odp_sched_pktio test application

psavol:
New application for scheduler testing with pktio. Simpler code base than l2fwd as this focuses on scheduled mode only.  Performance is the same with l2fwd scheduled mode. The test will be extended with event queue, etc options (e.g. application pipelining through multiple event queues). Also Travis test integration is not yet done in this patch set.

### DIFF
--- a/platform/linux-generic/odp_packet_io.c
+++ b/platform/linux-generic/odp_packet_io.c
@@ -1170,12 +1170,14 @@ void odp_pktio_print(odp_pktio_t hdl)
 	len += snprintf(&str[len], n - len,
 			"pktio\n");
 	len += snprintf(&str[len], n - len,
-			"  handle            %" PRIu64 "\n",
-			odp_pktio_to_u64(hdl));
-	len += snprintf(&str[len], n - len,
 			"  name              %s\n", entry->s.name);
 	len += snprintf(&str[len], n - len,
 			"  type              %s\n", entry->s.ops->name);
+	len += snprintf(&str[len], n - len,
+			"  index             %i\n", _odp_pktio_index(hdl));
+	len += snprintf(&str[len], n - len,
+			"  handle (u64)      %" PRIu64 "\n",
+			odp_pktio_to_u64(hdl));
 	len += snprintf(&str[len], n - len,
 			"  state             %s\n",
 			entry->s.state ==  PKTIO_STATE_STARTED ? "start" :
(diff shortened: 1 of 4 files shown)


test/performance/odp_sched_pktio.c
```diff
@@ -0,0 +1,712 @@
+#include <stdio.h>
```
Bill-Fischofer-Linaro: Needs standard header with copyright, etc.

test/performance/odp_sched_pktio.c
```diff
@@ -0,0 +1,712 @@
...
+		odp_packet_from_event_multi(pkt, ev, num);
+
+		in     = odp_packet_input_index(pkt[0]);
+		out    = test_global->pktio_map[in];
+		pktout = test_global->pktio[out].pktout[worker_id];
+
+		fill_eth_addr(pkt, num, test_global, out);
+
+		sent = odp_pktout_send(pktout, pkt, num);
+		drop = num - sent;
```
Bill-Fischofer-Linaro: Need to account for possibility that `sent < 0` here if `odp_pktout_send()` returned an error. Otherwise you'll calculate `drop > num` and attempt to reference a negative pkt index on the subsequent `odp_packet_free_multi()`.

platform/linux-generic/odp_packet_io.c
```diff
@@ -376,8 +376,10 @@ int odp_pktio_close(odp_pktio_t hdl)
 	int res;
 
 	entry = get_pktio_entry(hdl);
-	if (entry == NULL)
+	if (entry == NULL) {
+		ODP_ERR("Bad handle\n");
```
Bill-Fischofer-Linaro: The commit log says "add debug prints" so shouldn't this be `ODP_DBG()` rather than `ODP_ERR()`? Same comment for the rest of the `ODP_ERR()` calls here.

test/performance/odp_sched_pktio.c
psavol: Done in v2.

test/performance/odp_sched_pktio.c
psavol: Done in v2.

platform/linux-generic/odp_packet_io.c
psavol: ODP_ERR() is used throughout the file to report errors. So, ODP_ERR is used also for these errors. The reader of commit log gets the idea of prints been added to error conditions (no functional changes just prints added).

Bill-Fischofer-Linaro: OK, but then I'd change the commit log to something like "add error prints to log bad handles". That would make the log and the code consistent.

psavol: Changed "debug" to "error" in v3.

muvarov:
Merged

PR #533, opened 2018-03-16
[PATCH v1] shippable: reenable non-ABI-compat build for GCC

lumag:
Non-ABI-compat build is broken only for Clang, so disable it only for
that compiler, rather than disabling it completely.


muvarov:
Merged.

PR #535, opened 2018-03-20
[PATCH v2] performance: fix sched_latency test with huge cpu count

lumag:
odp_sched_latency has off-by-one error in accessing cpu stats: worker
thread id do not start from 0. Instead of fixing just off-by-one, use
ODP_THREAD_COUNT_MAX directly to allocate proper amount of cpu stats
structures.

### DIFF
--- a/test/performance/odp_sched_latency.c
+++ b/test/performance/odp_sched_latency.c
@@ -27,7 +27,6 @@
 /* GNU lib C */
 #include <getopt.h>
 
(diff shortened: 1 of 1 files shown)


muvarov:
@psavol are you ok? I can correct line brakes checkpatch issue on apply.

muvarov:
See Bills review. Merged.

muvarov:
Merged with Bills review.

PR #537, opened 2018-03-22
[PATCH v2] pktio: dpdk: fix IPv4 csum calculation when l4 offset is not set

bogdanPricope:
Fix IPv4 csum calculation when l4 offset is not set

IPv4 header contains length of the header and options in IHL field. It
can be used to replace L3 length calculation based on offsets.

See:  https://bugs.linaro.org/show_bug.cgi?id=3686


muvarov:
will be overwritten with Bogdan's new PR.

muvarov:
has to be good to merge

muvarov:
Merged.

PR #540, opened 2018-03-27
[PATCH v1] pktio: dpdk: accept UDPv4 packets with all-zero csum

bogdanPricope:
Accept IPv4 UDP packets with all-zero checksum field even if
DPDK reported it as bad checksum value.

This patch fixes bug: https://bugs.linaro.org/show_bug.cgi?id=3685.


JannePeltonen:
This patch (v1) fixes the problem in my test setup.

muvarov:
Merged.

PR #541, opened 2018-03-27
[PATCH v3] Run sched pktio test as part of validation suite

psavol:
Added option and script to run the test as part of validation suite.

### DIFF
--- a/test/performance/odp_sched_pktio.c
+++ b/test/performance/odp_sched_pktio.c
@@ -8,6 +8,8 @@
 #include <string.h>
 #include <signal.h>
 #include <stdlib.h>
+#include <stdint.h>
+#include <inttypes.h>
 
 #include <odp_api.h>
 #include <odp/helper/odph_api.h>
@@ -22,11 +24,17 @@
 #define BURST_SIZE        32
 #define CHECK_PERIOD      10000
 #define MAX_PKTIO_INDEXES 256
+#define TEST_PASSED_LIMIT 5000
 
 typedef struct {
 	int  worker_id;
 	void *test_global_ptr;
 } worker_arg_t;
+
+typedef struct ODP_ALIGNED_CACHE {
+	uint64_t rx_pkt;
+	uint64_t tx_pkt;
+} worker_stat_t;
 
 typedef struct {
 	volatile int  stop_workers;
@@ -35,6 +43,7 @@ typedef struct {
 	struct {
 		int num_worker;
 		int num_pktio;
+		uint8_t collect_stat;
 	} opt;
 
 	int max_workers;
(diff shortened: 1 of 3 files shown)


test/performance/odp_sched_pktio.c
```diff
@@ -696,13 +758,28 @@ int main(int argc, char *argv[])
...
 quit:
 	stop_pktios(test_global);
 	empty_queues();
 	close_pktios(test_global);
 
+	if (test_global->opt.collect_stat) {
+		print_stat(test_global, odp_time_diff_ns(t2, t1));
+
+		/* Encode return value for validation test usage. */
+		if (test_global->rx_pkt_sum > 5000)
```
muvarov: 5000 should be in config option or at least in top level define.

test/performance/odp_sched_pktio.c
```diff
@@ -718,5 +795,5 @@ int main(int argc, char *argv[])
 		return -1;
 	}
 
-	return 0;
+	return ret;
```
muvarov: I think it's better to do opposite.  C test pass if it's exit is 0. And bash should just check if it's 0 or not. That will not confuse people who run it manually. Something like return !(ret == TEST_EXPECTED_VAL).

test/performance/odp_sched_pktio_run.sh
```diff
@@ -0,0 +1,101 @@
...
+	kill ${GEN_PID}
+	wait ${GEN_PID}
+
+	# Kill with SIGINT to output statistics
+	kill -2 ${TEST_PID}
+	wait ${TEST_PID}
+
+	ret=$?
+
+	if [ $ret -eq 3 ]; then
```
muvarov: Change C code to expect 0. Maybe  better to have argument how many packets you will feed from generator.

Bill-Fischofer-Linaro:
@psavol Please rebase to resolve merge conflicts due to intervening changes.

test/performance/odp_sched_pktio.c
psavol: OK. Added a define in v3.

test/performance/odp_sched_pktio.c
psavol: By default this program does not check rx/tx packet counts and returns 0. Validation test need to use special option (-s) to enable rx/tx packet statistics counting and checking. When rx and tx packet counts are checked and both are OK, return value is 3. In all other cases, return value is not 3 (but -1 or 0). E.g. if validation test script does not use -s option, we catch the error instead of returning "passed".

test/performance/odp_sched_pktio_run.sh
psavol: Program returns by default 0, but when asked for statistics checking (-s) it returns check status: 0, 1,  2 or 3 (== success on both rx and tx). Return value is limited to max value 255, so cannot return packet count as return value.

muvarov:
Merged.

PR #542, opened 2018-03-27
[PATCH API-NEXT v4] api: packet: UDP checksum value of zero

psavol:
UDP checksum value of zero results OK status when checksum
is checked.

### DIFF
--- a/include/odp/api/spec/packet.h
+++ b/include/odp/api/spec/packet.h
@@ -1711,6 +1711,9 @@ odp_packet_chksum_status_t odp_packet_l3_chksum_status(odp_packet_t pkt);
  * attempt. It depends on packet input (or IPSEC) configuration, packet content
  * and implementation capabilities if checksum check is attempted for a packet.
  *
+ * When a UDP packet does not have a checksum (e.g. checksum field of a UDP/IPv4
+ * packet is zero), checksum check result is ODP_PACKET_CHKSUM_OK.
+ *
  * @param pkt     Packet handle
  *
  * @return L4 checksum check status


include/odp/api/spec/packet.h
```diff
@@ -1711,6 +1711,9 @@ odp_packet_chksum_status_t odp_packet_l3_chksum_status(odp_packet_t pkt);
  * attempt. It depends on packet input (or IPSEC) configuration, packet content
  * and implementation capabilities if checksum check is attempted for a packet.
  *
+ * Checksum check result for an UDP packet with checksum field value of zero
```
Bill-Fischofer-Linaro: Might be worth clarifying that this is for IPv4 UDP packets with zero checksums?

include/odp/api/spec/packet.h
```diff
@@ -1711,6 +1711,9 @@ odp_packet_chksum_status_t odp_packet_l3_chksum_status(odp_packet_t pkt);
  * attempt. It depends on packet input (or IPSEC) configuration, packet content
  * and implementation capabilities if checksum check is attempted for a packet.
  *
+ * When an UDP packet does not have a checksum (checksum field of an UDP/IPv4
+ * packet is zero), checksum check result is ODP_PACKET_CHKSUM_OK.
```
Bill-Fischofer-Linaro: The rules for indefinite articles ("a" vs. "an") in English are somewhat tricky. An is usually used for words beginning with vowels, however there are exceptions.  See [this article](https://owl.english.purdue.edu/owl/resource/591/01/) for an explanation. So "a UDP packet" is preferable here.

An alternative is to rephrase the line to avoid use of indefinite articles altogether. For example: `Since checksums are optional for IPv4 UDP packets, checksum result in this case is ODP_PACKET_CHKSUM_OK.`

muvarov:
Merged to api-next

PR #543, opened 2018-03-28
linux-gen: dpdk: fix runtime config read order

matiaselo:
If used runtime configuration is missing driver specific options, use top
level dpdk pktio options, not the hard coded driver specific options.

Internal _odp_libconfig_lookup_int() function changed to return the found
option type (not found/default/runtime).

### DIFF
--- a/platform/linux-generic/pktio/dpdk.c
+++ b/platform/linux-generic/pktio/dpdk.c
@@ -98,19 +98,36 @@ static int lookup_opt(const char *path, const char *drv_name, int *val)
 {
 	const char *base = "pktio_dpdk";
 	char opt_path[256];
-	int ret = 0;
+	int drv_ret = 0;
+	int def_ret = 0;
+	int drv_val = 0;
+	int def_val = 0;
+
+	/* Driver specific option */
+	snprintf(opt_path, sizeof(opt_path), "%s.%s.%s", base, drv_name, path);
+	drv_ret = _odp_libconfig_lookup_int(opt_path, &drv_val);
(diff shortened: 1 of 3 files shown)


bogdanPricope:
Yes, it is working now (in a ‘scratching your left ear w/ your right hand’ kind of way).

Reviewed-and-tested-by: Bogdan Pricope <bogdan.pricope@linaro.org>

platform/linux-generic/pktio/dpdk.c
```diff
@@ -98,19 +98,36 @@ static int lookup_opt(const char *path, const char *drv_name, int *val)
...
+	int drv_ret = 0;
+	int def_ret = 0;
+	int drv_val = 0;
+	int def_val = 0;
+
+	/* Driver specific option */
+	snprintf(opt_path, sizeof(opt_path), "%s.%s.%s", base, drv_name, path);
+	drv_ret = _odp_libconfig_lookup_int(opt_path, &drv_val);
+	if (drv_ret == LIBCONFIG_OPT_RUNTIME) {
+		*val = drv_val;
```
muvarov: why you modify val is error is returned?

muvarov: same comment for other changes in this file. If val is not needed to modify then def_val and def_ret might be simple ret and rval.

matiaselo: This is not error handling. 1 is returned if a value was found.

bogdanPricope:
We may consider alternative implementation: https://github.com/bogdanPricope/odp/commit/8ed975b11b6f7f205532936f01673813f1098223

PR #544, opened 2018-03-29
[PATCH v1] build: fix autoconf error caused by double-registering config item

lumag:
Autoconf uses provided tag as is, when registering config command.
Require platform code to pass platform argument, so that different
config commands will be registered.


muvarov:
Merged.

PR #547, opened 2018-04-02
[PATCH v1] Devel/master clang aarch64

muvarov:
This patch has to fix clang. I hope that there is no degradation on speed and alloca() will be internal inlinded built-in function. Might be it's better to put this cache inside cache = local.cache[pool->pool_idx], so it will be in the dynamic memory near cache->num value where this code also accesses to.


lumag:
Any reason for using alloca instead of just moving an array allocation?

muvarov:
@lumag  only to keep closer to original code. Moving array definition on top also helps. I think there has to be no difference in performance.

platform/linux-generic/odp_pool.c
```diff
@@ -722,8 +722,9 @@ int buffer_alloc_multi(pool_t *pool, odp_buffer_hdr_t *buf_hdr[], int max_num)
 	if (odp_unlikely(num_deq)) {
 		/* Temporary copy needed since odp_buffer_t is uintptr_t
 		 * and not uint32_t. */
-		uint32_t data[burst];
+		uint32_t *data;
 
+		data = alloca(burst * sizeof(uint32_t));
```
Bill-Fischofer-Linaro: There are some [cautionary comments](https://stackoverflow.com/questions/1018853/why-is-the-use-of-alloca-not-considered-good-practice) regarding the use of `alloca()` at Stackoverflow. I'd prefer to find a better workaround to this issue. Sounds like we should be able to simply bound `num_deq` and size the `data` array to a constant maximum value rather than making it dynamic.

muvarov:
alloca() is bad solution. Closed.

PR #549, opened 2018-04-06
[PATCH v1] clean up ishm file naming and debug prints

muvarov:
I think there is quite few people who understand what ivshm debugs prints actually prints. So I think it's better to hide debug messages which can be enabled in the code.

- travis: use -M option for generated patches
- linux-gen: ishm: remove useless debug print
- linux-gen: hide debug prints from fd server
- linux-gen: clean up ishm file naming


muvarov:
Merged.

PR #553, opened 2018-04-09
example: l2fwd: update readme regarding ordered queue usage

Bill-Fischofer-Linaro:
Update the README file to clarify the effect of using ordered queues
with direct I/O.


muvarov:
good note

Bill-Fischofer-Linaro:
@JannePeltonen Can you please review this to ensure it reflects the changes you wanted?

example/l2fwd/README
```diff
@@ -5,3 +5,7 @@ Source code and Makefiles placed under test/performance/ directory.
 This L2 forwarding application can be used as example reference as well
 as performance test for different odp modes (direct, queue or scheduler
 with parallel, atomic or ordered queues).
+
+Note that this example is tuned for performance. As a result when using
+ordered queues with direct packet I/O output, packet order is not
+guaranteed. To achieve guaranteed order, use queued output (-o option).
```
JannePeltonen: Using queued output is not sufficient in case of l2fwd since the l2fwd may send packets received from the same input queue to different output queues of the same output port. This is because the output queue is selected based on the thread that happens to be handling the packet and not e.g. based on flow hash or the input queue. For the same reason, order is not necessarily preserved with atomic queues either (enqueuing would happen in order with atomic queues, but packets in more than one queue may pass each other before output to the wire).

The only case where l2fwd guarantees that the input order is preserved in output is when there is only one worker thread.

JannePeltonen: Yes, having only one output queue per interface would preserve the order, but that would indeed require some implementation work. Multiple output queues per interface with a flow based or input queue based output queue selection would also preserve the order.

Bill-Fischofer-Linaro:
Changes for v2:
* Rebased to latest master
* Reword based on feedback from @JannePeltonen

example/l2fwd/README
```diff
@@ -5,3 +5,10 @@ Source code and Makefiles placed under test/performance/ directory.
 This L2 forwarding application can be used as example reference as well
 as performance test for different odp modes (direct, queue or scheduler
 with parallel, atomic or ordered queues).
+
+Note that this example is tuned for performance. As a result, when using
+ordered queues with direct packet I/O output or queued output with multiple
```
JannePeltonen: I forgot in my previous comment that l2fwd support also direct pktin mode without the packet order problem. OTOH, in scheduled mode the issue affects atomic queues in addition to ordered queues.

More accurate wording would be something like this: "As a result, when using scheduled mode with direct packet I/O output or queued output with multiple output queues, packet order is not guaranteed."

Or maybe more simply: "As a result, packet order is not guaranteed when using scheduled mode."

Bill-Fischofer-Linaro: OK, used your suggested wording in v3.

example/l2fwd/README
```diff
@@ -5,3 +5,10 @@ Source code and Makefiles placed under test/performance/ directory.
 This L2 forwarding application can be used as example reference as well
 as performance test for different odp modes (direct, queue or scheduler
 with parallel, atomic or ordered queues).
+
+Note that this example is tuned for performance. As a result, when using
+scheduled mode with direct packet I/O output or queued output with multiple
+output queues, packet order is not guaranteed. To achieve guaranteed order,
+use a single worker thread or output interfaces with single output
+queues. Other examples of scalable processing using ordered queues that
+preserve order can be seen in the odp_pktio_ordered performance test.git
```
JannePeltonen: Looks ok to me now.

It seem there is accidental text "git" at the very end.

Bill-Fischofer-Linaro: Hmmm. Not sure how that stray got in there. Fixed in v4.

example/l2fwd/README
```diff
@@ -5,3 +5,10 @@ Source code and Makefiles placed under test/performance/ directory.
 This L2 forwarding application can be used as example reference as well
 as performance test for different odp modes (direct, queue or scheduler
 with parallel, atomic or ordered queues).
+
+Note that this example is tuned for performance. As a result, when using
+scheduled mode with direct packet I/O output or queued output with multiple
+output queues, packet order is not guaranteed. To achieve guaranteed order,
+use a single worker thread or output interfaces with single output
+queues. Other examples of scalable processing using ordered queues that
+preserve order can be seen in the odp_pktio_ordered performance test.
```
JannePeltonen: Looks good to me.

muvarov: @JannePeltonen Reviewed-by: Janne Peltonen <janne.peltonen@...> ?

PR #558, opened 2018-04-13
[PATCH v1] linux-gen: dpdk: allocate huge page memory for all numa nodes

matiaselo:
Allocate DPDK memory for all numa nodes in the system.

Fixes bug: https://bugs.linaro.org/show_bug.cgi?id=3657

### DIFF
--- a/platform/linux-generic/pktio/dpdk.c
+++ b/platform/linux-generic/pktio/dpdk.c
@@ -49,6 +49,13 @@
 #include <rte_string_fns.h>
 #include <rte_version.h>
 
+/* NUMA is not supported on all platforms */
+#ifdef RTE_EAL_NUMA_AWARE_HUGEPAGES
+#include <numa.h>
+#else
+#define numa_num_configured_nodes() 1
+#endif
+
 #if RTE_VERSION < RTE_VERSION_NUM(17, 5, 0, 0)
 #define rte_log_set_global_level rte_set_log_level
 #endif
@@ -1086,6 +1093,7 @@ static int dpdk_pktio_init(void)
 	int32_t masklen;
 	int mem_str_len;
 	int cmd_len;
+	int numa_nodes;
 	cpu_set_t original_cpuset;
 	struct rte_config *cfg;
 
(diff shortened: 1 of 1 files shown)


platform/linux-generic/pktio/dpdk.c
```diff
@@ -14,6 +14,9 @@
 #include <ctype.h>
 #include <unistd.h>
 #include <math.h>
+#if !defined(__arm__) || defined(__aarch64__)
+#include <numa.h>
+#endif
```
Bill-Fischofer-Linaro: How about adding:
```
#else
#define numa_num_configured_nodes() 1
#endif
```
to eliminate the need for the other `#if` below? Also, are we using `!__arm__` as a synonym for x86_64? If I'm running on PPC or MIPS is `<numa.h>` available?

platform/linux-generic/pktio/dpdk.c
```diff
@@ -1120,21 +1124,34 @@ static int dpdk_pktio_init(void)
 		return -1;
 	}
 
-	mem_str_len = snprintf(NULL, 0, "%d", DPDK_MEMORY_MB);
+	mem_str_len = snprintf(NULL, 0, "%d,", DPDK_MEMORY_MB);
+
+#if defined(__arm__) && !defined(__aarch64__)
+	numa_nodes = 1;
+# else
+	numa_nodes = numa_num_configured_nodes();
```
muvarov: That looks like a hack. Instead of cross build libnuma you ifdefed call here. If you want make code working without libnuma let ./configure to test if libnuma exist and set variable which you will use here.

platform/linux-generic/pktio/dpdk.c
lumag: I'd also prefer dummy `numa_num_configured_nodes()` implementation for the platforms where it is not supported.

lumag: Configure test seems perfectly fine here.

platform/linux-generic/pktio/dpdk.c
lumag: The problem here is not cross-building of libnuma, but rather libnuma not supporting 32-bit ARM platforms (well, kernel not supporting `migrate_pages` syscall on that platform).

muvarov: thanks Dmirty. Then maybe to put this call under odp arch specific files?

muvarov: I'm ok with that.

Bill-Fischofer-Linaro: @lumag, @muvarov Isn't the real question whether `#include <numa.h>` is available to avoid compilation failures? As long as `numa_num_configured_nodes()` returns 1 for non-numa environments this should be good.

matiaselo: @Bill-Fischofer-Linaro That was the reason I added the ifdef around the #include. However, at least on Travis the numa.h seems to be available even on arm32 cross compilation.

platform/linux-generic/pktio/dpdk.c
matiaselo: Makes sense, will update this.

matiaselo: Would the following be OK for everyone?

```
#if RTE_EAL_NUMA_AWARE_HUGEPAGES
#include <numa.h>
#else
#define numa_num_configured_nodes() 1
#endif
```

DPDK build system will take care of detecting numa awareness.

muvarov: That is ok for me.

lumag: @matiaselo fine with me also

lumag:

muvarov:
Merged.

PR #561, opened 2018-04-16
[PATCH v2] linux-gen: shm: check return value when registering fds

joseppc:
fdserver functions such as that for registering a new fd may
fail. In this case we need to check the return value and do
a cleanup of any allocated blocks and fragments.
If it fails on deregistering, just return the error code.

### DIFF
--- a/platform/linux-generic/odp_ishm.c
+++ b/platform/linux-generic/odp_ishm.c
@@ -768,6 +768,65 @@ static void procsync(void)
 }
 
 /*
+ * Free a block as described in block_free(), but
+ * considering whether to close the file descriptor or not, and
+ * whether to deregister from the fdserver.
+ */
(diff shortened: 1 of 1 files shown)


platform/linux-generic/odp_ishm.c
```diff
@@ -767,6 +767,65 @@ static void procsync(void)
...
+		return -1;
+	}
+
+	block = &ishm_tbl->block[block_index];
+
+	proc_index = procfind_block(block_index);
+	if (proc_index >= 0) {
+		/* close the related fd */
+		if (close_fd)
+			close(ishm_proctable->entry[proc_index].fd);
```
muvarov: Isn't here close() has to be after unmap(), not before?

muvarov:
Merged.

PR #562, opened 2018-04-16
[PATCH v2] fdserver: prevent process from handling some signals

joseppc:
Make fdserver block all signals, so it won't receive signals sent to the
process group, which may interfere with the normal operation of the
application.
All signals are blocked except those that we want to handle (SIGTERM),
or those that can't be blocked, or those for which ignoring them results
in undefined behavior.

This patch fixes: https://bugs.linaro.org/show_bug.cgi?id=3690

### DIFF
--- a/platform/linux-generic/odp_fdserver.c
+++ b/platform/linux-generic/odp_fdserver.c
@@ -647,6 +647,34 @@ int _odp_fdserver_init_global(void)
 	}
 
 	if (server_pid == 0) { /*child */
+		sigset_t sigset;
+		struct sigaction action;
+
+		sigfillset(&sigset);
+		/* undefined if these are ignored, as per POSIX */
+		sigdelset(&sigset, SIGFPE);
+		sigdelset(&sigset, SIGILL);
+		sigdelset(&sigset, SIGSEGV);
+		/* can not be masked */
+		sigdelset(&sigset, SIGKILL);
+		sigdelset(&sigset, SIGSTOP);
+		/* these we want to handle */
+		sigdelset(&sigset, SIGTERM);
+		if (sigprocmask(SIG_SETMASK, &sigset, NULL) == -1) {
+			ODP_ERR("Could not set signal mask");
+			exit(1);
+		}
+
+		/* set default handlers for those signals we can handle */
+		memset(&action, 0, sizeof(action));
+		action.sa_handler = SIG_DFL;
+		sigemptyset(&action.sa_mask);
+		action.sa_flags = 0;
+		sigaction(SIGFPE, &action, NULL);
+		sigaction(SIGILL, &action, NULL);
+		sigaction(SIGSEGV, &action, NULL);
+		sigaction(SIGTERM, &action, NULL);
+
 		/* TODO: pin the server on appropriate service cpu mask */
 		/* when (if) we can agree on the usage of service mask  */
 


platform/linux-generic/odp_fdserver.c
```diff
@@ -647,6 +647,32 @@ int _odp_fdserver_init_global(void)
...
+			ODP_ERR("Could not mask terminal signals");
+			exit(1);
+		}
+
+		/* set default handlers for those signals we can handle */
+		action.sa_handler = SIG_DFL;
+		sigemptyset(&action.sa_mask);
+		action.sa_flags = 0;
+		sigaction(SIGKILL, &action, NULL);
+		sigaction(SIGSTOP, &action, NULL);
```
JannePeltonen: You cannot set a signal handler for SIGKILL or SIGSTOP (see sigaction(2)).

platform/linux-generic/odp_fdserver.c
```diff
@@ -647,6 +647,32 @@ int _odp_fdserver_init_global(void)
...
+		sigdelset(&sigset, SIGFPE);
+		sigdelset(&sigset, SIGILL);
+		sigdelset(&sigset, SIGSEGV);
+		/* can not be masked */
+		sigdelset(&sigset, SIGKILL);
+		sigdelset(&sigset, SIGSTOP);
+		/* these we want to handle */
+		sigdelset(&sigset, SIGTERM);
+		if (sigprocmask(SIG_SETMASK, &sigset, NULL) == -1) {
+			ODP_ERR("Could not mask terminal signals");
```
JannePeltonen: I would remove the word "terminal" since signals are not specific to terminals.

platform/linux-generic/odp_fdserver.c
```diff
@@ -647,6 +647,32 @@ int _odp_fdserver_init_global(void)
...
 
 	if (server_pid == 0) { /*child */
+		sigset_t sigset;
+		struct sigaction action;
+
+		sigfillset(&sigset);
+		/* undefined if these are ignored, as per POSIX */
+		sigdelset(&sigset, SIGFPE);
+		sigdelset(&sigset, SIGILL);
+		sigdelset(&sigset, SIGSEGV);
```
JannePeltonen: Maybe the commit comment should be reworded since fdserver is no longer in the same process group as the application. Something about setting signal handling to a known good state regardless of what the application may have done?

platform/linux-generic/odp_fdserver.c
```diff
@@ -647,6 +647,32 @@ int _odp_fdserver_init_global(void)
 	}
 
 	if (server_pid == 0) { /*child */
+		sigset_t sigset;
+		struct sigaction action;
```
JannePeltonen: Parts of this struct may be left uninitialized. I would suggest memset(&action, 0, sizeof(action)) at some point before the sigaction call.

platform/linux-generic/odp_fdserver.c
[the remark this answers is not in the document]
joseppc: Agreed, fixed

platform/linux-generic/odp_fdserver.c
joseppc: True, thanks!

platform/linux-generic/odp_fdserver.c
joseppc: I have no idea how this comment came about :smile:, fixed

platform/linux-generic/odp_fdserver.c
joseppc: Yes, thanks

platform/linux-generic/odp_fdserver.c
joseppc: Yes, makes sense.

joseppc:
@muvarov  there is a typo in the commit message, there is one instance of "fdeserver" (extra "e")

muvarov:
Merged.

PR #566, opened 2018-04-18
[PATCH v1] configure.ac: update version to v1.19.0.0

muvarov:
as we discussed we have change in api and there we increase api version. No api function just, just text clarificaion. But because of text can be understood in diferent way with new note and without it api version change has to be increased. So this is 1.19.0.0


Bill-Fischofer-Linaro:
I had to restart a number of the Travis tests. Not sure what happened, but everything seems to be passing now.

lumag:

muvarov:
That was merged.

PR #568, opened 2018-04-20
Timer validation test updates

psavol:
Various fixes on timer validation test. E.g. test assumed low (lower than CPU freq) timer tick frequency.


muvarov:
Merged. @Bill-Fischofer-Linaro  I think we will test patches in master first and after some time will backport to TM-LTS.

PR #569, opened 2018-04-22
[PATCH v1] linux-generic: random: split from crypto module

lumag:
While random and crypto might share some implementation details, in case
of linux-generic and linux-DPDK it might be easier to split them to two
different files.


muvarov:
Merged.

PR #571, opened 2018-04-22
[PATCH v1] linux-gen: dpdk: fix linking to libnuma

lumag:
Linking to dpdk/numa can fail with the following message:
/usr/bin/x86_64-linux-gnu-ld: lib/.libs/libodp-linux.a(dpdk.o): undefined reference to symbol 'numa_num_configured_nodes@@libnuma_1.2'
//usr/lib/x86_64-linux-gnu/libnuma.so.1: error adding symbols: DSO missing from command line

### DIFF
--- a/platform/linux-generic/m4/odp_dpdk.m4
+++ b/platform/linux-generic/m4/odp_dpdk.m4
@@ -40,6 +40,14 @@ then
       ;;
     esac
 
+    OLD_LIBS=$LIBS
+    LIBS="-lnuma"
+    AC_TRY_LINK_FUNC([numa_num_configured_nodes],
+		     [AC_DEFINE([HAVE_NUMA_LIBRARY], [1],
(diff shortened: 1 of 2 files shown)


lumag:
@muvarov no. It lists dependency of libdpdk.

muvarov:
@lumag ah ok, we have m4/odp_dpdk.m4 and platform/lg/m4/odp_dpdk.m4. If dpdk depends on that library why not to put it to common directory m4/ ?

lumag:
@muvarov because it's pktio/dpdk, who also depends on libnuma.

lumag:
It complains, because error message is ineed unwrapped. I'd prefer not to change it.

muvarov:
Merged. @lumag  I agree with @Bill-Fischofer-Linaro. commit message can be splitted across the lines to fit to size.

PR #573, opened 2018-04-23
[PATCH v1] test: crypto perf: check algorithm capabilities

psavol:
Test only those algorithms that implementation supports.


lumag:

muvarov:
Merged.

PR #574, opened 2018-04-23
[PATCH v1] linux-gen: fdserver: remove unnecessary locking

joseppc:
The locks in fdserver's operations don't serve any purpose, there is no
need to acquire a lock when registering a file descriptor to the server,
specially because the only place where the registering function is used
is already protected by another lock, likewise for
deregister and lookup. Also, the fdserver handles requests sequentially.


muvarov:
Merged.

PR #575, opened 2018-04-23
linux-gen: ipsec: parse IPv6 next header in tunnel case

lumag:
IPsec parsing code will fail to update next header field in internal
state in case of outbound tunnel processing of IPv6 packets.

Fixes: https://bugs.linaro.org/show_bug.cgi?id=3764


lumag:
Note to @muvarov. This fix should also go into TigerMoth LTS branch.

PR #576, opened 2018-04-24
[PATCH v1] Use explicit DPDK version when fetching

lumag:
This makes our Travis script use explicit DPDK version, rather than branch name. This allows one to test ODP against different point releases of DPDK at the same time removing need to play drop-the-cache game each time.


lumag:
@matiaselo could you please review

muvarov:
@lumag @matiaselo  what about installation dpdk into dpdk-sha directory? I.e. stay on stable branch but cache inside Travis directory which matches the latest commit in the stable branch?

matiaselo:
I prefer Dmitry's solution as it explicitly states which code revision is being used and tested.

muvarov:
Merged.

PR #578, opened 2018-04-24
[PATCH v2] validation: timer: loosen period accuracy requirement

psavol:
Test passes if measured time period is with +-20%
(+-4x resolution) of the requested period.


psavol:
```
		t2 = odp_time_local();
		diff_test = odp_time_diff_ns(t2, t0);

		if (ev != ODP_EVENT_INVALID) {
			diff_period = odp_time_diff_ns(t2, t1);
			t1 = t2;
```

I think two patches for a three lines changes would be an overkill. If you look couple lines back, it's obvious that it's unnecessary to call odp_time_local() for t1 again as it was called three lines before (for t2).

muvarov:
Merged.

PR #580, opened 2018-04-24
[PATCH API-NEXT v18] api: compression: compression specification

bala-manoharan:
ODP compression specification

Signed-off-by: Shally Verma <shally.verma@cavium.com>
Signed-off-by: Mahipal Challa <mahipal.challa@cavium.com>
Reviewed-by: Bogdan Pricope bogdan.pricope@linaro.org

### DIFF
--- a/include/odp/api/abi-default/comp.h
+++ b/include/odp/api/abi-default/comp.h
@@ -1,0 +1,35 @@
+/* Copyright (c) 2018, Linaro Limited
+ * All rights reserved.
+ *
+ * SPDX-License-Identifier:     BSD-3-Clause
+ */
+
+#ifndef ODP_ABI_COMP_H_
+#define ODP_ABI_COMP_H_
+
+#ifdef __cplusplus
+extern "C" {
+#endif
+
+#include <stdint.h>
+
+/** @internal Dummy type for strong typing */
+typedef struct { char dummy; /**< @internal Dummy */ } _odp_abi_comp_session_t;
+
+/** @ingroup odp_compression
+ *  @{
+ */
+
+typedef _odp_abi_comp_session_t *odp_comp_session_t;
+
+#define ODP_COMP_SESSION_INVALID        ((odp_comp_session_t)0)
+
+/**
+ * @}
+ */
+
+#ifdef __cplusplus
+}
+#endif
+
+#endif
--- a/include/odp/api/spec/comp.h
+++ b/include/odp/api/spec/comp.h
@@ -1,0 +1,613 @@
+/* Copyright (c) 2018, Linaro Limited
+ * All rights reserved.
+ *
+ * SPDX-License-Identifier:     BSD-3-Clause
+ */
+
+/**
+ * @file
+ *
+ * ODP Compression
+ */
+
+#ifndef ODP_API_COMP_H_
+#define ODP_API_COMP_H_
+
+#include <odp/visibility_begin.h>
+#include <odp/api/support.h>
+#include <odp/api/packet.h>
+
+#ifdef __cplusplus
+extern "C" {
+#endif
+
+/** @defgroup odp_compression ODP COMP
+ *  Operations for Compression and Decompression API.
+ *  Hash calculation may be combined with de-/compression operations
+ *
+ *  @{
+ */
+
+/**
+ * @def ODP_COMP_SESSION_INVALID
+ * Invalid session handle
+ */
+
+/**
+ * @typedef odp_comp_session_t
+ * Compression/Decompression session handle
+ */
+
+/**
+ * Compression operation mode
+ */
+typedef enum {
+	/** Synchronous Compression operation
+	 *
+	 * Application uses synchronous operation,
+	 * which outputs all results on function return.
+	 * */
+	ODP_COMP_OP_MODE_SYNC,
+
+	/** Asynchronous Compression operation
+	 *
+	 * Application uses asynchronous operation,
+	 * which return results via events.
+	 * */
+	ODP_COMP_OP_MODE_ASYNC
+} odp_comp_op_mode_t;
+
+/**
+ * Compression operation type.
+ */
+typedef enum {
+	/** Operation type - Compress */
+	ODP_COMP_OP_COMPRESS,
+
+	/** Operation type - Decompress */
+	ODP_COMP_OP_DECOMPRESS
+} odp_comp_op_t;
+
+/**
+ * Compression hash algorithms
+ */
+typedef enum {
+	/** No hash algorithm selected. */
+	ODP_COMP_HASH_ALG_NONE,
+
+	/** SHA-1 hash algorithm. */
+	ODP_COMP_HASH_ALG_SHA1,
+
+	/** SHA-2 hash algorithm 256-bit digest length. */
+	ODP_COMP_HASH_ALG_SHA256
+} odp_comp_hash_alg_t;
+
+/**
+ * Compression algorithms
+ *
+ */
+typedef enum {
+	/** No algorithm specified. Added for testing purpose. */
+	ODP_COMP_ALG_NULL,
+
+	/** DEFLATE - RFC1951 */
+	ODP_COMP_ALG_DEFLATE,
+
+	/** ZLIB - RFC1950 */
+	ODP_COMP_ALG_ZLIB,
+
+	/** LZS */
+	ODP_COMP_ALG_LZS
+} odp_comp_alg_t;
+
+/**
+ * Compression operation status codes
+ */
+typedef enum {
+	/** Operation completed successfully*/
+	ODP_COMP_STATUS_SUCCESS,
+
+	/** Operation terminated due to insufficient output buffer */
+	ODP_COMP_STATUS_OUT_OF_SPACE_TERM,
+
+	/** Operation failure */
+	ODP_COMP_STATUS_FAILURE,
+} odp_comp_status_t;
+
+/**
+ * Hash algorithms in a bit field structure
+ */
+typedef union odp_comp_hash_algos_t {
+	/** hash algorithms */
+	struct {
+		/** ODP_COMP_HASH_ALG_NONE */
+		uint32_t none	: 1,
+
+		/** ODP_COMP_HASH_ALG_SHA1 */
+		uint32_t sha1	: 1;
+
+		/** ODP_COMP_HASH_ALG_SHA256 */
+		uint32_t sha256	: 1;
+
+	} bit;
+
+	/** All bits of the bit field structure
+	 *
+	 * This field can be used to set/clear all flags, or bitwise
+	 * operations over the entire structure.
+	 */
+	uint32_t all_bits;
+} odp_comp_hash_algos_t;
+
+/**
+ * Compression algorithms in a bit field structure
+ */
+typedef union odp_comp_algos_t {
+	/** Compression algorithms */
+	struct {
+		/** ODP_COMP_ALG_NULL */
+		uint32_t null       : 1;
+
+		/** ODP_COMP_ALG_DEFLATE */
+		uint32_t deflate    : 1;
+
+		/** ODP_COMP_ALG_ZLIB */
+		uint32_t zlib       : 1;
+
+		/** ODP_COMP_ALG_LZS */
+		uint32_t lzs        : 1;
+	} bit;
+
+	/** All bits of the bit field structure
+	 * This field can be used to set/clear all flags, or bitwise
+	 * operations over the entire structure.
+	 */
+	uint32_t all_bits;
+} odp_comp_algos_t;
+
+/**
+ * Compression Interface Capabilities
+ */
+typedef struct odp_comp_capability_t {
+	/** Maximum number of  sessions */
+	uint32_t max_sessions;
+
+	/** Supported compression algorithms */
+	odp_comp_algos_t comp_algos;
+
+	/** Supported hash algorithms */
+	odp_comp_hash_algos_t hash_algos;
+
+	/** Synchronous compression mode support (ODP_COMP_OP_MODE_SYNC) */
+	odp_support_t sync;
+
+	/** Aynchronous compression mode support (ODP_COMP_OP_MODE_SSYNC) */
+	odp_support_t async;
+} odp_comp_capability_t;
+
+/**
+ * Hash algorithm capabilities
+ */
+typedef struct odp_comp_hash_alg_capability_t {
+	/** Digest length in bytes */
+	uint32_t digest_len;
+} odp_comp_hash_alg_capability_t;
+
+/**
+ * Compression algorithm capabilities
+ */
+typedef struct odp_comp_alg_capability_t {
+
+	/** Maximum compression level supported by implementation of this
+	 * algorithm. Indicates number of compression levels supported by
+	 * implementation. Valid range from (1 ... max_level)
+	 */
+	uint32_t max_level;
+
+	/** Supported hash algorithms */
+	odp_comp_hash_algos_t hash_algo;
+
+	/** Compression ratio
+	 * Optimal compression operation ratio for this algorithm.
+	 * This is an estimate of maximum compression operation output for this
+	 * algorithm. It is expressed as a percentage of maximum expected
+	 * output data size with respect to input data size.
+	 * i.e a value of 200% denotes the output data is 2x times the input
+	 * data size. This is an optimal/most case estimate and it is possible
+	 * that the percentage of output data produced might be greater
+	 * than this value.
+	 *
+	 * @see odp__percent_t
+	 */
+	odp_percent_t compression_ratio;
+} odp_comp_alg_capability_t;
+
+/**
+ * Compression Huffman type. Used by DEFLATE algorithm
+ */
+typedef enum odp_comp_huffman_code {
+	/** Fixed Huffman code */
+	ODP_COMP_HUFFMAN_FIXED,
+
+	/** Dynamic Huffman code */
+	ODP_COMP_HUFFMAN_DYNAMIC,
+
+	/** Default huffman code selected by implementation */
+	ODP_COMP_HUFFMAN_DEFAULT,
+} odp_comp_huffman_code_t;
+
+/**
+ * Compression DEFLATEe algorithm parameters.
+ * Also initialized by other deflate based algorithms , ex. ZLIB
+ */
+typedef struct odp_comp_deflate_param {
+	/**
+	 * Compression level
+	 *
+	 * Valid range is integer between (0 ... max_level)
+	 * level supported by the implementation.
+	 *
+	 * where,
+	 * 0 - implemention default
+	 *
+ 	 * 1 - fastest compression i.e. output produced at
+ 	 * best possible speed at the expense of compression quality
+	 *
+	 * max_level - High quality compression
+	 *
+	 * @see 'max_level' in odp_comp_alg_capability_t
+	 */
+	uint32_t comp_level;
+
+	/** huffman code to use */
+	odp_comp_huffman_code_t huffman_code;
+} odp_comp_deflate_param_t;
+
+/**
+ * Compression algorithm specific parameters
+ */
+typedef union odp_comp_alg_param_t {
+	/** deflate parameter */
+	odp_comp_deflate_param_t deflate;
+
+	/** Struct for defining zlib algorithm parameters */
+	struct {
+		/** deflate algo params */
+		odp_comp_deflate_param_t deflate;
+	} zlib;
+} odp_comp_alg_param_t;
+
+ /**
+ * Compression session creation parameters
+ */
+typedef struct odp_comp_session_param_t {
+	/** Compression operation type Compress vs Decompress */
+	odp_comp_op_t op;
+
+	/** Compression operation mode
+	 *
+	 * Operation mode Synchronous vs Asynchronous
+	 *
+	 * @see odp_comp_op(), odp_comp_op_enq()
+	 */
+	odp_comp_op_mode_t mode;
+
+	/** Compression algorithm
+	 *
+	 *  @see odp_comp_capability()
+	 */
+	odp_comp_alg_t comp_algo;
+
+	/** Hash algorithm
+	 *
+	 *  @see odp_comp_alg_capability()
+	 */
+	odp_comp_hash_alg_t hash_algo;
+
+	/** parameters specific to compression */
+	odp_comp_alg_param_t alg_param;
+
+	/** Session packet enqueue ordering
+	 * Boolean to indicate if packet enqueue ordering is required per
+	 * session. Valid only for Asynchronous operation mode
+	 * (ODP_COMP_OP_MODE_ASYNC). Packet order is always maintained for
+	 * synchronous operation mode (ODP_COMP_OP_MODE_SYNC)
+	 *
+	 * true: packet session enqueue order maintained
+	 *
+	 * false: packet session enqueue order is not maintained
+	 *
+	 * @note: By disabling packet order requirement, performance oriented
+	 * application can leverage HW offered parallelism to increase operation
+	 * performance.
+	 */
+	odp_bool_t packet_order;
+
+	/** Destination queue for compression operations result.
+	 * Results are enqueued as ODP_EVENT_PACKET with subtype
+	 * ODP_EVENT_PACKET_COMP
+	 */
+	odp_queue_t compl_queue;
+} odp_comp_session_param_t;
+
+/**
+ * Compression packet operation result
+ */
+typedef struct odp_comp_packet_result_t {
+	/** Operation status code */
+	odp_comp_status_t status;
+
+	/** Input packet handle */
+	odp_packet_t pkt_in;
+
+	/** Output packet data range
+	 * Specifies offset and length of data resulting from compression
+	 * operation. When hashing is configured output_data_range.len equals
+	 * length of output data + 'digest+len'
+	 */
+	odp_packet_data_range_t output_data_range;
+} odp_comp_packet_result_t;
+
+/**
+ * Compression per packet operation parameters
+ */
+typedef struct odp_comp_packet_op_param_t {
+	/** Session handle */
+	odp_comp_session_t session;
+
+	/** Input data range to process. where,
+	 *
+	 * offset - starting offset
+	 * length - length of data for compression operation
+	 * */
+	odp_packet_data_range_t in_data_range;
+
+	/** Output packet data range.
+	 * Indicates where processed packet will be written. where,
+	 *
+	 * offset - starting offset
+	 * length - length of buffer available for output
+	 *
+	 * Output packet data is not modified outside of this provided data
+	 * range. If output data length is not sufficient for compression
+	 * operation ODP_COMP_STATUS_OUT_OF_SPACE_TERM error will occur
+	 */
+	odp_packet_data_range_t out_data_range;
+} odp_comp_packet_op_param_t;
+
+/**
+ * Query compression capabilities
+ *
+ * Output compression capabilities on success.
+ *
+ * @param[out] capa   Pointer to capability structure for output
+ *
+ * @retval 0 on success
+ * @retval <0 on failure
+ */
+int odp_comp_capability(odp_comp_capability_t *capa);
+
+/**
+ * Query supported compression algorithm capabilities
+ *
+ * Output algorithm capabilities.
+ *
+ * @param	comp	Compression algorithm
+ * @param[out]	capa	Compression algorithm capability
+ *
+ * @retval 0 on success
+ * @retval <0 on failure
+ */
+int odp_comp_alg_capability(odp_comp_alg_t comp,
+			    odp_comp_alg_capability_t *capa);
+
+/**
+ * Query supported hash algorithm capabilities
+ *
+ * Outputs all supported configuration options for the algorithm.
+ *
+ * @param	hash     Hash algorithm
+ * @param	capa     Hash algorithm capability
+ *
+ * @retval 0 on success
+ * @retval <0 on failure
+ */
+int odp_comp_hash_alg_capability(odp_comp_hash_alg_t hash,
+				 odp_comp_hash_alg_capability_t *capa);
+
+/**
+ * Initialize compression session parameters
+ *
+ * Initialize an odp_comp_session_param_t to its default values for
+ * all fields.
+ *
+ * @param param   Pointer to odp_comp_session_param_t to be initialized
+ */
+void odp_comp_session_param_init(odp_comp_session_param_t *param);
+
+/**
+ * Compression session creation
+ *
+ * Create a comp session according to the session parameters. Use
+ * odp_comp_session_param_init() to initialize parameters into their
+ * default values.
+ *
+ * @param	param	Session parameters
+ *
+ * @retval		Comp session handle
+ * @retval		ODP_COMP_SESSION_INVALID on failure
+ */
+odp_comp_session_t
+odp_comp_session_create(const odp_comp_session_param_t *param);
+
+/**
+ * Compression session destroy
+ *
+ * Destroy an unused session. Result is undefined if session is being used
+ * (i.e. asynchronous operation is in progress).
+ *
+ * @param	session		Session handle
+ *
+ * @retval	0 on success
+ * @retval	<0 on failure
+ */
+int odp_comp_session_destroy(odp_comp_session_t session);
+
+/**
+ * Synchronous packet compression operation
+ *
+ * This operation does packet compression in synchronous mode. A successful operation
+ * returns the number of successfully processed input packets and updates the
+ * results in the corresponding output packets. Outputted packets contain
+ * compression results metadata (odp_comp_packet_result_t), which should be
+ * checked for operation status. Length of outputted data can be got from
+ * output_data_range.len.
+ *
+ * When hashing is configured along with compression operation the
+ * result is appended at the end of the output data, output_data_range.len
+ * equals length of output data + 'digest_len'. Processed data length
+ * can be computed by subtracting 'digest_len' from output_data_range.len where
+ * 'digest_len' can be queried from odp_comp_hash_alg_capability().
+ * Hash is always performed on plain text. Hash validation in decompression is
+ * performed by the application.
+ * For every input packet entry in 'pkt_in' array, application should pass
+ * corresponding valid output packet handle. If any error occurs during
+ * processing of packets, the API returns with number of entries successfully
+ * processed.
+ * Output packet metadatas like length or data pointer will not be updated.
+ *
+ * @param          pkt_in	Packets to be processed
+ * @param          pkt_out	Packet handle array for resulting packets
+ * @param          num_pkt	Number of packets to be processed
+ * @param          param	Operation parameters
+ *
+ * @return Number of input packets consumed (0 ... num_pkt)
+ * @retval <0 on failure
+ *
+ * @note The 'pkt_in','pkt_out'and 'param' arrays should be of same length,
+ * Results are undefined if otherwise.
+
+ * @note Same packet handle cannot be used as input and output parameter.
+ * In-place compression operation is not supported
+ */
+int odp_comp_op(const odp_packet_t pkt_in[], odp_packet_t pkt_out[],
+		int num_pkt, const odp_comp_packet_op_param_t param[]);
+
+/**
+ * Asynchronous packet compression  operation
+ *
+ * This operation does packet compression in asynchronous mode. It processes
+ * packets otherwise identical to odp_comp_op(), but the resulting packets are
+ * enqueued to 'compl_queue' configured during session (odp_comp_session_t)
+ * creation. For every input packet entry in in_pkt array, user should pass
+ * corresponding valid output packet handle. On return, API returns with
+ * number of entries successfully submitted for operation.
+ *
+ * When hashing is configured along with compression operation the
+ * result is appended at the end of the output data, output_data_range.len
+ * equals length of output data + 'digest_len'. Processed data length
+ * can be computed by subtracting 'digest_len' from output_data_range.len where
+ * 'digest_len' can be queried from odp_comp_hash_alg_capability().
+ * Hash is always performed on plain text. Hash validation in decompression is
+ * performed by the application.
+ *
+ * In case of partially accepted array i.e.
+ * when number of packets returned < num_pkt, application may attempt to
+ * resubmit subsequent entries via calling any of the operation API.
+ *
+ * All the packets successfully enqueued will be submitted to 'compl_queue'
+ * after compression operation, Application should check 'status' of the
+ * operation in odp_comp_packet_result_t.
+ * Output packet metadatas like length or data pointer will not be updated.
+ *
+ * Please note it is always recommended that application using async mode,
+ * provide sufficiently large buffer size to avoid
+ * ODP_COMP_STATUS_OUT_OF_SPACE_TERM.
+ *
+ * @param	pkt_in		Packets to be processed
+ * @param	pkt_out		Packet handle array for resulting packets
+ * @param	num_pkt		Number of packets to be processed
+ * @param	param		Operation parameters
+ *
+ * @return	Number of input packets enqueued (0 ... num_pkt)
+ * @retval	<0 on failure
+ *
+ * @note The 'pkt_in','pkt_out'and 'param' arrays should be of same length,
+ * Results are undefined if otherwise.
+
+ * @note Same packet handle cannot be used as input and output parameter.
+ * In-place compression operation is not supported
+
+ * @see odp_comp_op(), odp_comp_packet_result()
+ */
+int odp_comp_op_enq(const odp_packet_t pkt_in[], odp_packet_t pkt_out[],
+		    int num_pkt, const odp_comp_packet_op_param_t param[]);
+
+/**
+ * Get compression operation results from processed packet.
+ *
+ * Successful compression operations of all modes (ODP_COMP_OP_MODE_SYNC and
+ * ODP_COMP_OP_MODE_ASYNC) produce packets which contain compression result
+ * metadata. This function copies operation results from compression processed
+ * packet. Event subtype of this packet is ODP_EVENT_PACKET_COMP. Results are
+ * undefined if non-compression processed packet is passed as input.
+ *
+ * @param[out]    result  pointer to operation result for output
+ * @param	  packet  compression processed packet (ODP_EVENT_PACKET_COMP)
+ *
+ * @retval  0     On success
+ * @retval <0     On failure
+ */
+int odp_comp_result(odp_comp_packet_result_t *result, odp_packet_t packet);
+
+ /**
+  * Convert compression processed packet event to packet handle
+  *
+  * Get packet handle corresponding to processed packet event. Event subtype
+  * must be ODP_EVENT_PACKET_COMP. Compression operation results can be
+  * examined with odp_comp_result().
+  *
+  * @param	event	Event handle
+  *
+  * @return	Valid Packet handle on success,
+  * @retval	ODP_PACKET_INVALID on failure
+  *
+  * @see odp_event_subtype(), odp_comp_result()
+  *
+  */
+odp_packet_t odp_comp_packet_from_event(odp_event_t event);
+
+ /**
+  * Convert processed packet handle to event
+  *
+  * The packet handle must be an output of a compression operation
+  *
+  * @param	pkt	Packet handle from compression operation
+  * @return Event handle
+  */
+odp_event_t odp_comp_packet_to_event(odp_packet_t pkt);
+
+/**
+ * Get printable value for an odp_comp_session_t
+ *
+ * @param hdl  odp_comp_session_t handle to be printed
+ * @return     uint64_t value that can be used to print/display this
+ *             handle
+ *
+ * @note This routine is intended to be used for diagnostic purposes
+ * to enable applications to generate a printable value that represents
+ * an odp_comp_session_t handle.
+ */
+uint64_t odp_comp_session_to_u64(odp_comp_session_t hdl);
+
+/**
+ * @}
+ */
+
+#ifdef __cplusplus
+}
+#endif
+
+#include <odp/visibility_end.h>
+#endif
+


Bill-Fischofer-Linaro:
General comment: DPDK is working on a similar set of enhancements. [Latest DPDK RFC for compressdev changes](http://dpdk.org/dev/patchwork/patch/34821/).

include/odp/api/spec/comp.h
```diff
@@ -0,0 +1,863 @@
...
+	/** Input data range to process */
+	odp_packet_data_range_t in_data_range;
+
+	/** Output packet data range.
+	 * Indicates where processed packet will be written
+	 * where length specifies,
+	 * length of available packet buffer at input, and
+	 * length of data written at output
+	 */
+	odp_packet_data_range_t out_data_range;
```
Bill-Fischofer-Linaro: How does caller know how large a range to specify for decompress operations? Or should this be an output parameter that specifies the range that was written in the output packet corresponding to the `in_data_range`?

include/odp/api/spec/comp.h
```diff
@@ -0,0 +1,863 @@
...
+ * If session is created in ODP_COMP_ASYNC mode, behavior is undefined.
+ *
+ * For compression + hash, call returns with hash appended at the end of
+ * packet which has last field set in corresponding param.
+ *
+ * User should compute processed data len = output_data_range.len - digest_len,
+ * where digest_len queried through odp_comp_hash_alg_capability().
+ *
+ * For every input packet, user should pass corresponding valid output
+ * packet.
```
Bill-Fischofer-Linaro: For crypto, `pkt_out` can be specified as `ODP_PACKET_INVALID` to request that ODP allocate output packets. This would seem to make more sense here since the size of the output packets is going to be unpredictable.

include/odp/api/spec/comp.h
```diff
@@ -0,0 +1,863 @@
...
+ * If session is created in ODP_COMP_SYNC mode, behavior is undefined.
+ *
+ * For compression + hash, call returns with hash appended at the end of
+ * packet which has last field set in corresponding param.
+ *
+ * User should compute processed data len = output_data_range.len - digest_len,
+ * where digest_len queried through odp_comp_hash_alg_capability().
+ *
+ * For every input packet, user should pass corresponding valid output
+ * packet.
```
Bill-Fischofer-Linaro: Same comment as for sync operation. We should support having ODP allocate the output packets.

include/odp/api/spec/comp.h
```diff
@@ -0,0 +1,863 @@
...
+ */
+typedef struct odp_comp_capability_t {
+	/** Maximum number of  sessions */
+	uint32_t max_sessions;
+
+	/** Supported compression algorithms */
+	odp_comp_algos_t comp_algos;
+
+	/** Supported hash algorithms. */
+	odp_comp_hash_algos_t hash_algos;
```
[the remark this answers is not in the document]
lumag: yes

include/odp/api/spec/comp.h
```diff
@@ -0,0 +1,863 @@
...
+ * its default choice of compression level.
+ *
+ */
+typedef enum {
+	/* Use implementation default */
+	ODP_COMP_LEVEL_DEFAULT = 0,
+	/** Minimum compression (fastest in speed) */
+	ODP_COMP_LEVEL_MIN,
+	/** Maximum compression (slowest in speed) */
+	ODP_COMP_LEVEL_MAX,
```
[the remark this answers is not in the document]
lumag: I think we can drop `ODP_COMP_LEVEL_MAX` completely.

include/odp/api/spec/comp.h
```diff
@@ -0,0 +1,863 @@
...
+ *
+ */
+typedef enum {
+	/** use implementation default to choose between compression codes  */
+	ODP_COMP_HUFFMAN_CODE_DEFAULT = 0,
+	/** use fixed huffman codes */
+	ODP_COMP_HUFFMAN_CODE_FIXED,
+	/** use dynamic huffman coding */
+	ODP_COMP_HUFFMAN_CODE_DYNAMIC,
+} odp_comp_huffman_code_t;
```
lumag: Could you please define this closer to `odp_comp_algo_param_t`? (or even completely inside that structure)?

include/odp/api/spec/comp.h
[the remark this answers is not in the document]
bala-manoharan: Agreed. Will drop in the next version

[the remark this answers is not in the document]
muvarov:
```
gzip -h
  -1, --fast        compress faster
  -9, --best        compress better
```
Maximum value might have some reason. I check lzma, gzip, bz2 and etc. All of them use 1 to 9 compression level. Default level in most cases is 6. But might be some other algos use different version.

Bill-Fischofer-Linaro:
Note that DPDK 18.05 was released on May 30th. It includes [compressdev API support](http://dpdk.org/doc/api/rte__compressdev_8h.html). ODP compression APIs should be mappable onto this support for odp-dpdk.

include/odp/api/abi-default/comp.h
```diff
@@ -0,0 +1,35 @@
...
+#ifdef __cplusplus
+extern "C" {
+#endif
+
+#include <stdint.h>
+
+/** @internal Dummy type for strong typing */
+typedef struct { char dummy; /**< @internal Dummy */ } _odp_abi_comp_session_t;
+
+/** @ingroup odp_comp
```
psavol: This group name needs to match the name defined with defgroup tag in the spec file (odp_compression).

include/odp/api/spec/comp.h
```diff
@@ -0,0 +1,863 @@
...
+#include <odp/api/support.h>
+#include <odp/api/packet.h>
+
+#ifdef __cplusplus
+extern "C" {
+#endif
+
+/** @defgroup odp_compression ODP COMP
+ *  ODP Compression is an API set to do compression+hash or decompression+hash
+ *  operations on data. Hash is calculated on plaintext.
```
psavol: Comp/decomp without hash is also supported. This chapter gives an impression that hash is always calculated.

" ... API to do compression or decompression operations on data. Hash calculation may be combined with comp/decomp operations..."

include/odp/api/spec/comp.h
```diff
@@ -0,0 +1,863 @@
...
+	ODP_COMP_SES_CREATE_ERR_NONE,
+	/** Creation failed, no resources */
+	ODP_COMP_SES_CREATE_ERR_ENOMEM,
+	/** Creation failed, bad compression params */
+	ODP_COMP_SES_CREATE_ERR_INV_COMP,
+	/** Creation failed, bad hash params */
+	ODP_COMP_SES_CREATE_ERR_INV_HASH,
+	/** Creation failed,requested configuration not supported*/
+	ODP_COMP_SES_CREATE_ERR_NOT_SUPPORTED
+} odp_comp_ses_create_err_t;
```
psavol: With suffient capability struct application should be able to avoid bad session create requests. So, typically return code 0 (success) vs <0 (fail) is enough. <0 enables implementation specific failure codes. So, this type may not be needed.

include/odp/api/spec/comp.h
```diff
@@ -0,0 +1,863 @@
...
+	 * output buffer but no other alteration to operation params
+	 * (odp_comp_op_param_t).
+	 *
+	 * if using async mode, application should either make sure to
+	 * provide sufficient output buffer size OR maintain relevant
+	 * context (or ordering) information with respect to each input packet
+	 * en-queued for processing.
+	 *
+	 */
+	ODP_COMP_ERR_OUT_OF_SPACE,
```
psavol: This out-of-space feature complicates API definition and usage. I'd have rather capa defitions for max input data size and min output buffer size. So, that normally operation completes successfully (also in async after successful operation start call).

include/odp/api/spec/comp.h
```diff
@@ -0,0 +1,863 @@
...
+	ODP_COMP_ERR_DATA_SIZE,
+	/**  Compression Algo failure */
+	ODP_COMP_ERR_ALGO_FAIL,
+	/** Error if operation has been requested in an invalid state */
+	ODP_COMP_ERR_INV_STATE,
+	/** Error if API does not support any of the operational parameter. */
+	ODP_COMP_ERR_NOT_SUPPORTED,
+	/** Error if session is invalid. */
+	ODP_COMP_ERR_INV_SESS
+} odp_comp_err_t;
```
psavol: If failure is not part of normal operation, API defined error codes are not need (0 == success, <0 == failure).

include/odp/api/spec/comp.h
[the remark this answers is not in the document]
psavol: Since enum is int, range needs to be defined. Other option is not to use enum, but defines and e.g. uint8_t. Anyway, default =0, min = 1, max = 255? makes sense to define. As default level is algo specific, implementation needs to swap 0 to another value per algo anyway (no harm to fix it to 0 in API).

include/odp/api/spec/comp.h
```diff
@@ -0,0 +1,863 @@
...
+typedef union odp_comp_hash_algos_t {
+	/** hash algorithms */
+	struct {
+		/** ODP_COMP_HASH_ALG_SHA1 */
+		uint32_t sha1  : 1;
+
+		/** ODP_COMP_HASH_ALG_SHA256 */
+		uint32_t sha256 : 1;
+
+	} bit;
```
psavol: Add "null" algo for not calculating hash ?

include/odp/api/spec/comp.h
```diff
@@ -0,0 +1,863 @@
...
+ * Comp set dictionary
+ *
+ * Should be called when there is no operation in progress i.e.
+ * before initiating processing of first chunk of data and
+ * after processing of last chunk of data is complete.
+ *
+ * @param		session		Session handle
+ * @param[in, out]	dict		Pointer to dictionary.
+ *					Implementation should update
+ *					length of dictionary at output.
```
psavol: Where dict length is updated ? Why dict pointer is in and out parameter?

include/odp/api/spec/comp.h
```diff
@@ -0,0 +1,863 @@
...
+	/** Optional Maximum length of dictionary supported
+	 *  by implementation of an algorithm.
+	 *
+	 *  Invalid if support_dict == ODP_SUPPORT_NO.
+	 *
+	 *  Implementation use dictionary of length less than or equal to value
+	 *  indicated by dict_len. if set to 0 and if support_dict ==
+	 *  ODP_SUPPORT_YES, then implementation will use dictionary length
+	 *  less than or equal to user input length in odp_comp_set_dict()
+	 *  and update used dictionary length at output of the call.
```
psavol: Wouldn't this capa would just tell what's the max dict len that application can request. If implementation uses some default length to me it's not a capa but an info. Capas are limits for application requests. Didn't quite get how this dict feature is set up.

include/odp/api/spec/comp.h
```diff
@@ -0,0 +1,863 @@
...
+	 *
+	 * Example, if max_level = 4 , it means algorithm supports four levels
+	 * of compression from value 1 up to 4. User can set this value from
+	 * 1 (fastest compression) to 4 (best compression).
+	 * See RFC1950 for an example explanation to level.
+	 *
+	 * Value 0 mean implementation use its default value.
+	 *
+	 */
+	uint32_t max_level;
```
psavol: Just reference to odp_comp_alg_param_t level here and move level specification text to odp_comp_alg_param_t. So, here just say that: "Maximum value for level parameter in compression algorithm parameters (odp_comp_alg_param_t). 0 when algorithm does not have a level parameter."

include/odp/api/spec/comp.h
```diff
@@ -0,0 +1,863 @@
...
+
+	/** Supported hash algorithms */
+	odp_comp_hash_algos_t hash_algo;
+} odp_comp_alg_capability_t;
+
+/**
+ * Comp API dictionary type
+ * Consists of pointer to byte buffer. length of dictionary
+ * indicated by length parameter.
+ */
```
psavol: Any requirements / specification for dictionary data format?

include/odp/api/spec/comp.h
```diff
@@ -0,0 +1,863 @@
...
+ * Comp API algorithm specific parameters
+ *
+ */
+typedef struct odp_comp_algo_param_t {
+	/**
+	 * Struct for defining deflate algorithm parameters.
+	 * Also initialized by other deflate based algorithms , ex. ZLIB
+	 */
+	struct comp_alg_def_param {
+		/**
```
psavol: Better to typedef this when it's used in more than one place.

include/odp/api/spec/comp.h
```diff
@@ -0,0 +1,863 @@
...
+		 * compression level where
+		 * ODP_COMP_LEVEL_MIN <= level <= ODP_COMP_LEVEL_MAX
+		 */
+		odp_comp_level_t level;
```
psavol: Move level specification here. DEFAULT or MIN <= level <= capa.max_level

include/odp/api/spec/comp.h
```diff
@@ -0,0 +1,863 @@
...
+	 *  Use odp_comp_capability() for supported algorithms.
+	 */
+	odp_comp_alg_t comp_algo;
+
+	/** Hash algorithm
+	 *
+	 *  Use odp_comp_alg_capability() for supported hash algo for
+	 *  compression algo given as comp_alg. Implementation should not
+	 *  support hash only operation on data. output should always contain
+	 *  data + hash.
```
psavol: "Implementation should not support hash only operation on data. output should always contain data + hash."

Maybe this can be removed since hash_algos are listed is per comp algo capability. Null comp algo can set all hash algo bits to zero (== hash not supported). If some HW supports null comp + hash algo, then that implementation may support hash only also in comp API.

include/odp/api/spec/comp.h
```diff
@@ -0,0 +1,863 @@
...
+ * @param         param    Operation parameter
+ * @param[out] result   Result parameter, containing after operation result.
+ *
+ * @return 0 on success
+ * @retval <0 on failure
+ */
+int odp_comp_op(const odp_packet_t pkt_in,
+		const odp_packet_t pkt_out,
+		const odp_comp_packet_op_param_t *param,
+		odp_comp_packet_op_result_t *result);
```
psavol: IPSEC operation function model could be copied here:
```
 * @param          pkt_in   Packets to be processed
 * @param          num_in   Number of packets to be processed
 * @param[out]     pkt_out  Packet handle array for resulting packets
 * @param[in, out] num_out  Number of resulting packets. Application sets this
 *                          to 'pkt_out' array size. A successful operation sets
 *                          this to the number of outputted packets
 *                          (1 ... num_out).
 * @param          param    Inbound operation parameters
 *
 * @return Number of input packets consumed (0 ... num_in)
 * @retval <0     On failure
 *
 * @see odp_packet_user_ptr(), odp_packet_user_area(), odp_packet_l3_offset(),
 * odp_packet_l4_offset()
 */
int odp_ipsec_in(const odp_packet_t pkt_in[], int num_in,
		 odp_packet_t pkt_out[], int *num_out,
		 const odp_ipsec_in_param_t *param);
```
Implementation decides the number of output packets and allocates those. Application just gives pointer to output array. If output array is too small, some input packets were not processed.

include/odp/api/spec/comp.h
```diff
@@ -0,0 +1,863 @@
...
+
+/** TODO: This section requires further enhancement mainly,
+ * ODP_COMP_OUT_OF_SPACE error needs to be handled properly */
+/**
+ * Comp API to process a packet synchronously
+ *
+ * If session is created in ODP_COMP_SYNC mode, this call wait for operation
+ * to complete.
+ *
+ * If session is created in ODP_COMP_ASYNC mode, behavior is undefined.
```
psavol: It's enough to specify once that this operation is for SYNC mode.

"This operation does de-/compression processing in synchronous mode (ODP_COMP_MODE_SYNC)"

include/odp/api/spec/comp.h
```diff
@@ -0,0 +1,863 @@
...
+			  const odp_comp_packet_op_param_t param[],
+			  int num_pkt);
+
+/**
+ * Get compression/decompression operation results from an processed packet.
+ *
+ * User should call this API on every asynchronously processed packet
+ * or on every synchronously processed *failed* packet.
+ * For synchronous and successfully processed packets, it is an optional
+ * call. See odp_comp_compress() and odp_comp_decompress() for more details.
```
bogdanPricope: See odp_comp_compress() and odp_comp_decompress()...  where?

include/odp/api/spec/comp.h
bogdanPricope: Is not actually used... or?

include/odp/api/spec/comp.h
Bill-Fischofer-Linaro: Typo: s/output/Output/

Are should and should not as used here intended to be SHOULD and SHOULD NOT in the sense of [RFC 2119](https://tools.ietf.org/html/rfc2119)?

include/odp/api/spec/comp.h
Bill-Fischofer-Linaro: Typo s/length/Length/

include/odp/api/abi-default/comp.h
Bill-Fischofer-Linaro: Has this comment by @psavol been addressed?

include/odp/api/spec/comp.h
Bill-Fischofer-Linaro: Above comment by @psavol still unaddressed?

include/odp/api/spec/comp.h
Bill-Fischofer-Linaro: Still need to discuss this point. "Sorry, guess again" type APIs are not the most user friendly.

include/odp/api/spec/comp.h
Bill-Fischofer-Linaro: Still seems unclear here. Need a precise specification for this behavior.

include/odp/api/spec/comp.h
Bill-Fischofer-Linaro: I believe the point of @psavol's comment is we don't need to specify what happens if you violate the spec. By convention anything done in violation of the spec results in undefined behavior.

include/odp/api/spec/comp.h
Bill-Fischofer-Linaro: Still part of discussion that needs to be had.

include/odp/api/spec/comp.h
Bill-Fischofer-Linaro: These comments still not addressed.

include/odp/api/spec/comp.h
Bill-Fischofer-Linaro: @bogdanPricope's comment still not addressed. These two routines are not found in this PR.

include/odp/api/spec/comp.h
```diff
@@ -0,0 +1,752 @@
...
+	uint8_t *buf;
+	/** length of the dictionary. */
+	uint32_t len;
+} odp_comp_dict_t;
+
+/**
+ * Comp API algorithm specific parameters
+ *
+ */
+typedef struct odp_comp_algo_param_t {
```
bogdanPricope: 'union' instead of 'struct'?

include/odp/api/spec/comp.h
```diff
@@ -0,0 +1,752 @@
...
+ * For every input packet, user should pass corresponding valid output
+ * packet.
+ *
+ * For every successful or partially processed packet, api update result param
+ * with out_data_range indicating length of the data written in output packet
+ * and error code indicating operation status.
+ *
+ * A packet is said to be partially processed if it runs out of buffer during
+ * processing and need more buffer from user to continue. Under such condition,
+ * API fails with error code code in result param set to
```
bogdanPricope: 'code' is present twice.

include/odp/api/spec/comp.h
```diff
@@ -0,0 +1,752 @@
...
+	 *
+	 */
+	odp_queue_t compl_queue;
+} odp_comp_session_param_t;
+
+/**
+ * Comp API per packet operation result
+ *
+ */
+typedef struct odp_comp_packet_op_result_t {
```
bogdanPricope: Should contain an error code or... ?

include/odp/api/spec/comp.h
```diff
@@ -0,0 +1,752 @@
...
+int odp_comp_op_multi(const odp_packet_t pkt_in[],
+		      const odp_packet_t pkt_out[],
+		      odp_comp_packet_op_param_t param[],
+		      odp_comp_packet_op_result_t result[],
+		      int num_pkt);
+
+/**
+ * Comp API to process multiple packets asynchronously
+ *
+ * Performs the ASYNC decompression operations on the packet array.
```
bogdanPricope: I guess this function can do both compression and decompression asynchronously. This line can be removed.

include/odp/api/spec/comp.h
```diff
@@ -0,0 +1,752 @@
...
+ * For every input packet entry in in_pkt array, user should pass
+ * corresponding valid output packet handle.On return, API returns with
+ * number of entries successfully submitted for operation.
+ * In case of partially accepted array i.e. when
+ * number of packets returned < num_pkt, application my attempt to resubmit
+ * subsequent entries via calling any of the operation API.
+ *
+ * A successful submitted entry means it has been queued for processing but
+ * not fully processed. Each submitted entry results will be notified via
+ * event after operation on that entry is complete. On received event w.r.t
```
bogdanPricope: 'w.r.t' ?

include/odp/api/spec/comp.h
```diff
@@ -0,0 +1,752 @@
...
+ * ODP_COMP_ERR_NONE.
+ *
+ * In case of partially processed array i.e. when
+ * number of packets returned < num_pkt, application should check on result
+ * on 1st failed entry (i.e. n+1th entry where
+ * n = number of entried returned) to ensure if error code is
+ * ODP_COMP_ERR_OUT_OF_SPACE.
+ *
+ * If status code on 1st failed entry set to ODP_COMP_ERR_OUT_OF_SPACE,
+ * application should call API again with corresponding output packet
```
bogdanPricope: I guess it should call API again only with remaining (not processed) packets (n + 1 to num_pkt) , right?
The n + 1 packet will be a replacement or continuation of original n +1 packet?

include/odp/api/spec/comp.h
```diff
@@ -0,0 +1,752 @@
...
+ *
+ * A successful submitted entry means it has been queued for processing but
+ * not fully processed. Each submitted entry results will be notified via
+ * event after operation on that entry is complete. On received event w.r.t
+ * each submitted packet, application should call odp_comp_result() to retrieve
+ * operational result.
+ *
+ * If any of the packet has status code set to ODP_COMP_ERR_OUT_OF_SPACE in
+ * its result, application should call operation API again with corresponding
+ *  output packet available with free space.
```
bogdanPricope: Usecase: assume we enqueue 10 packets and 6 are processed OK but 7th has ODP_COMP_ERR_OUT_OF_SPACE.
1. Remaining 3 packets will not be delivered until  7th is 'resolved' (keep the order)? 
2. Application should resubmit only 7th packet, taken from the received event, with larger output packet?
3. What will happen with original 7th output packet?

include/odp/api/spec/comp.h
```diff
@@ -0,0 +1,863 @@
...
+	 * This is not an error condition. On seeing this situation,
+	 * Implementation should maintain context of in-progress operation and
+	 * application should call packet processing API again with valid
+	 * output buffer but no other alteration to operation params
+	 * (odp_comp_op_param_t).
+	 *
+	 * if using async mode, application should either make sure to
+	 * provide sufficient output buffer size OR maintain relevant
+	 * context (or ordering) information with respect to each input packet
+	 * en-queued for processing.
```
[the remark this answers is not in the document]
bala-manoharan: Done

include/odp/api/spec/comp.h
```diff
@@ -0,0 +1,863 @@
...
+	 *
+	 */
+	ODP_COMP_ERR_OUT_OF_SPACE,
+	/** Invalid user data pointers */
+	ODP_COMP_ERR_DATA_PTR,
+	/** Invalid input data size */
+	ODP_COMP_ERR_DATA_SIZE,
```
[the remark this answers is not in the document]
bala-manoharan: These additional error have been removed

include/odp/api/spec/comp.h
bala-manoharan: Done

include/odp/api/abi-default/comp.h
```diff
@@ -0,0 +1,35 @@
+/* Copyright (c) 2017-2018, Linaro Limited
```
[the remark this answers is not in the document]
bala-manoharan: Done

include/odp/api/spec/comp.h
```diff
@@ -0,0 +1,863 @@
+/* Copyright (c) 2014-2018, Linaro Limited
```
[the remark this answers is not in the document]
bala-manoharan: Done

include/odp/api/abi-default/comp.h
bala-manoharan: Done

include/odp/api/spec/comp.h
bala-manoharan: I have updated the documentation.

include/odp/api/spec/comp.h
bala-manoharan: I have removed this error code in next version.

include/odp/api/spec/comp.h
bala-manoharan: null hash type has been added for this use-case

include/odp/api/spec/comp.h
```diff
@@ -0,0 +1,709 @@
...
+ * before making call to this API.
+ */
+int odp_comp_set_dict(odp_comp_session_t session,
+		      odp_comp_dict_t *dict);
+
+/**
+ * Synchronous packet comp
+ *
+ * This operation does packet compression/decompression in synchronous mode.
+ * A successful proceesed entry means input packet has been completely consumed
```
bogdanPricope: Typo: processed

include/odp/api/spec/comp.h
```diff
@@ -0,0 +1,709 @@
...
+		      odp_comp_packet_op_result_t result[],
+		      int num_pkt);
+
+/**
+ * Asynchronous packet comp
+ *
+ * This operation does packet compression/decompression in asynchronous mode.
+ * A successful submitted entry means it has been queued for processing but
+ * not fully processed. Each submitted entry results will be notified via
+ * event after operation on that entry is complete. On received event w.r.t
```
bogdanPricope: 'w.r.t.' - avoid abbreviations, pls.

include/odp/api/spec/comp.h
```diff
@@ -0,0 +1,709 @@
...
+ * and whole output has been written into output packet and corresponding
+ * result entry is updated with offset:length information and error code set to
+ * ODP_COMP_ERR_NONE.
+ *
+ * For every input packet entry in in_pkt array, user should pass
+ * corresponding valid output packet handle. If during processing of packet,
+ * corresponding buffer goes out of space, then API returns with number of
+ * entries successfully processed.
+ *
+ * User should compute processed data len = output_data_range.len - digest_len,
```
bogdanPricope: Depends on odp_comp_packet_op_param_t:last field ?

include/odp/api/spec/comp.h
```diff
@@ -0,0 +1,709 @@
...
+ * number of packets returned < num_pkt, application should check on result
+ * on 1st failed entry (i.e. n+1th entry where
+ * n = number of entried returned) to ensure if error code is
+ * ODP_COMP_ERR_OUT_OF_SPACE.
+ *
+ * If status code on 1st failed entry set to ODP_COMP_ERR_OUT_OF_SPACE,
+ * application should call API again with corresponding output packet
+ * available with free space.
+ * For compression + hash, call returns with hash appended at the end of
+ * packet on which odp_packet_op_param_t:last field is set.
```
bogdanPricope: odp_comp_packet_op_param_t instead of odp_packet_op_param_t.

include/odp/api/spec/comp.h
```diff
@@ -0,0 +1,709 @@
...
+ * number of packets returned < num_pkt, application my attempt to resubmit
+ * subsequent entries via calling any of the operation API.
+ *
+ * User should compute processed data len = output_data_range.len - digest_len,
+ * where digest_len queried through odp_comp_hash_alg_capability().
+ * Application should monitor ODP_EVENT_PACKET
+ * be queued with subtype ODP_EVENT_PACKET_COMP on queue.
+ *
+ * for compression + hash, call returns with hash appended at the end of
+ * packet on which odp_packet_op_param_t:last field is set.
```
bogdanPricope: 'odp_comp_packet_op_param_t' instead of 'odp_packet_op_param_t'.
Also, this should be merged into paragraph above.

include/odp/api/spec/comp.h
```diff
@@ -0,0 +1,709 @@
...
+ *
+ * If any of the packet has status code set to ODP_COMP_ERR_OUT_OF_SPACE in
+ * its result, application should call operation API again with corresponding
+ *  output packet available with free space.
+ *
+ * Please note it is always recommended that application using async mode,
+ * provide sufficiently large buffer size to avoid ODP_COMP_ERR_OUT_OF_SPACE.
+ * Else it is recommended that application set required user context on packet
+ * before calling each odp_comp_op_enq() to correctly retrieve required
```
bogdanPricope: 'odp_comp_op_enq_multi()'

include/odp/api/spec/comp.h
```diff
@@ -0,0 +1,709 @@
...
+ * For every input packet entry in in_pkt array, user should pass
+ * corresponding valid output packet handle.On return, API returns with
+ * number of entries successfully submitted for operation.
+ * In case of partially accepted array i.e. when
+ * number of packets returned < num_pkt, application my attempt to resubmit
+ * subsequent entries via calling any of the operation API.
+ *
+ * User should compute processed data len = output_data_range.len - digest_len,
+ * where digest_len queried through odp_comp_hash_alg_capability().
+ * Application should monitor ODP_EVENT_PACKET
```
bogdanPricope: Assuming odp_comp_packet_from_event() will return output packet, you should have a way to also deliver input packet to the application: either to free input packet after a successful operation or to resubmit it in case of ODP_COMP_ERR_OUT_OF_SPACE. I suggest to deliver packet handler with odp_comp_packet_op_result_t.

bogdanPricope: I guess packet order will be respected while 'ODP_COMP_ERR_OUT_OF_SPACE' situation is solved. We need to state this.

include/odp/api/spec/comp.h
```diff
@@ -0,0 +1,661 @@
...
+	odp_bool_t last;
+
+	/** Input data range to process */
+	odp_packet_data_range_t in_data_range;
+
+	/** Output packet data range.
+	 * Indicates where processed packet will be written
+	 * where length specifies, data available for output.
+	 */
+	odp_packet_data_range_t out_data_range;
```
Bill-Fischofer-Linaro: Unclear how this is expected to be used. Suppose at input `odp_packet_len(pkt)` = 100 and a compression operation specifies an `in_data_range` of 20..39. Since compression (should) result in fewer output bytes than input bytes, what should be used as the `out_data_range`? If this is specified as 20..39 (same as the input), what's going to be the `odp_packet_len()` of the output packet? If those 20 input bytes compress down to 10 bytes, what does the resulting output packet look like? Presumably I'd want the packet to consist of the original bytes 0..19, then bytes 20..29, which represent the compressed `in_data_range`, and then the original bytes 40..99 which now occupy offsets 30..89 in the output packet which now has an `odp_packet_len()` of 90. But is this what happens?

Can the `out_data_range` be disjoint from the `in_data_range`? Can it partially overlap? And exactly what happens in these cases in the output packet?

Similar questions arise for decompression. If I say decompress `in_data_range` 20..39 how is `out_data_range` specified such that these bytes are decompressed and the following packet bytes remain unchanged?

I think we need some practical examples here to illustrate exactly how these parameters should be used to achieve a desired result.

include/odp/api/spec/comp.h
```diff
@@ -0,0 +1,661 @@
...
+ * @retval		ODP_COMP_SESSION_INVALID on failure
+ */
+odp_comp_session_t odp_comp_session_create(const odp_comp_session_param_t *
+					   param);
+
+/**
+ * Comp session destroy
+ *
+ * Destroy an unused session. Result is undefined if session is being used
+ * (i.e. asynchronous operation is in progress).
```
Bill-Fischofer-Linaro: Is it OK to destroy a session that is in a non-complete state (_i.e.,_ last operation specified `last` = 0)?

include/odp/api/spec/comp.h
Bill-Fischofer-Linaro: It's unclear what actual role the application has here. If the specified compression algorithm makes use of a dictionary, why wouldn't that be internal to the implementation and simply part of the `odp_comp_session_t`, allocated on `odp_comp_session_create()` and freed on `odp_comp_session_destroy()`? This seems to imply that `odp_comp_session_create()` is incomplete until this extra API is called. 

If the dictionary is internal to the session, then perhaps an `odp_comp_session_reset()` API might be useful to reset the session back to its "freshly created" state, but it's not clear why the application should need to concern itself directly with dictionaries. The same would go for any huffman tables, which are implied elsewhere but not defined here.

include/odp/api/spec/comp.h
```diff
@@ -0,0 +1,661 @@
...
+ * entries successfully processed.
+ *
+ * User should compute processed data len = output_data_range.len - digest_len,
+ * where digest_len queried through odp_comp_hash_alg_capability().
+ *
+ * In case of partially processed array i.e. when
+ * number of packets returned < num_pkt, application should check on result
+ * on 1st failed entry (i.e. n+1th entry where
+ * n = number of entried returned) to ensure if error code is
+ * ODP_COMP_ERR_OUT_OF_SPACE.
```
Bill-Fischofer-Linaro: `ODP_COMP_STATUS_OUT_OF_SPACE`?

include/odp/api/spec/comp.h
```diff
@@ -0,0 +1,661 @@
...
+ * @param         pkt_out  Output Packet to store output of corresponding entry
+ * @param         param    Operation parameters array
+ * @param[out]    result   Result array corresponding to each entry
+ * @param         num_pkt  Number of packets to be processed
+ *
+ * @return Number of input packets consumed (0 ... num_pkt)
+ * @retval <0 on failure
+ */
+int odp_comp_op(const odp_packet_t pkt_in[],
+		      const odp_packet_t pkt_out[],
```
Bill-Fischofer-Linaro: How is this a `const`? Aren't the `pkt_out[]` packets written here? Also need to fix indentation on these parameters.

include/odp/api/spec/comp.h
```diff
@@ -0,0 +1,661 @@
...
+ * Asynchronous packet comp
+ *
+ * This operation does packet compression/decompression in asynchronous mode.
+ * A successful submitted entry means it has been queued for processing but
+ * not fully processed. Each submitted entry results will be notified via
+ * event after operation on that entry is complete. On received event w.r.t
+ * each submitted packet, application should call odp_comp_result() to retrieve
+ * operational result.
+ * For every input packet entry in in_pkt array, user should pass
+ * corresponding valid output packet handle.On return, API returns with
```
Bill-Fischofer-Linaro: Missing space following .

include/odp/api/spec/comp.h
Bill-Fischofer-Linaro: If I submit N packets does this mean that they are processed independently? That is, if the input is packets 1..4 can 1 and 2 succeed, 3 fail, and 4 succeed?

include/odp/api/spec/comp.h
Bill-Fischofer-Linaro: @bogdanPricope IPsec uses `odp_ipsec_in/out_enq()` to accept multiple input packets, so I think this is consistent here.

include/odp/api/spec/comp.h
```diff
@@ -0,0 +1,661 @@
...
+			  int num_pkt);
+
+/**
+ * Get Comp operation results from processed packet.
+ *
+ * Successful Comp operation on async type produce packet which contain
+ * comp result metadata. This function copies operation results from an Comp
+ * processed packet. Event subtype of this kind of packet is
+ * ODP_EVENT_PACKET_COMP. Results are undefined if non-Comp processed packet is
+ * passed as input.
```
Bill-Fischofer-Linaro: Actually, shouldn't this be called for every `ODP_EVENT_PACKET` of subtype `ODP_EVENT_PACKET_COMP`? You don't know what the actual result is (success or failure) until you get its `odp_comp_packet_result_t` from this call.

include/odp/api/spec/comp.h
```diff
@@ -0,0 +1,661 @@
...
+	 * level supported by the implementation. where,
+ 	 * 1 - fastest compression i.e. output produced at
+ 	 * best possible speed at the expense of compression quality.
+	 * max_value - High quality compression.
+	 * 0 - implemention default
+	 */
+	uint32_t comp_level;
+
+	/** huffman code to use */
+	odp_comp_huffman_code_t comp_code;
```
bogdanPricope: 'odp_comp_huffman_code_t' not defined.

include/odp/api/spec/comp.h
```diff
@@ -0,0 +1,661 @@
...
+ * Also initialized by other deflate based algorithms , ex. ZLIB
+ */
+typedef struct odp_comp_alg_def_param {
+	/**
+	 * Compression level
+	 * Valid range is integer between 0 and max compression
+	 * level supported by the implementation. where,
+ 	 * 1 - fastest compression i.e. output produced at
+ 	 * best possible speed at the expense of compression quality.
+	 * max_value - High quality compression.
```
bogdanPricope: 'max_level' instead of 'max_value'

include/odp/api/spec/comp.h
```diff
@@ -0,0 +1,661 @@
...
+ * not fully processed. Each submitted entry results will be notified via
+ * event after operation on that entry is complete. On received event w.r.t
+ * each submitted packet, application should call odp_comp_result() to retrieve
+ * operational result.
+ * For every input packet entry in in_pkt array, user should pass
+ * corresponding valid output packet handle.On return, API returns with
+ * number of entries successfully submitted for operation.
+ * In case of partially accepted array i.e. when
+ * number of packets returned < num_pkt, application my attempt to resubmit
+ * subsequent entries via calling any of the operation API.
```
bogdanPricope: Misleading: syns/async is a session level option: resubmitting subsequent entries on the same session will have to use the same (async) operation API.

include/odp/api/spec/comp.h
```diff
@@ -0,0 +1,661 @@
...
+
+	/** Session packet enqueue ordering
+	 * Boolean to indicate if packet enqueue ordering is required per session
+	 * If packet_order is false, packet order is maintained on
+	 * operation completion order and not in session enqueue order.
+	 * By disabling packet order requirement, performance oriented
+	 * application can leverage HW offered parallelism to increase operation
+	 * performance.
+	 */
+	odp_bool_t packet_order;
```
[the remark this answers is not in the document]
bala-manoharan: This is application for both sync and ASYNC since some HW might submit the sync operation also through a co-processor.

include/odp/api/spec/comp.h
```diff
@@ -0,0 +1,603 @@
...
+	uint32_t digest_len;
+} odp_comp_hash_alg_capability_t;
+
+/**
+ * Compression dictionary.
+ * The dictionary is a sequence of bytes which are initially fed to the
+ * compressor engine. A preset dictionary is specially useful to compress
+ * short input sequences. The compression engine can take advantage of the
+ * dictionary context to encode the input in a more compact manner. (RFC 1950)
+ */
```
psavol: It would be useful to add a bit more description from RFC 1950
* The compressor and the decompressor must use exactly the same dictionary.
* The dictionary may be fixed or may be chosen among a certain number of predefined dictionaries, according to the kind of input data.
* the optimal dictionaries are application specific.

Also still some specification of dictionary data content/format is needed.

See for example http://www.zlib.net/manual.html 
" The dictionary should consist of strings (byte sequences) that are likely to be encountered later in the data to be compressed, with the most commonly used strings preferably put towards the end of the dictionary. Using a dictionary is most useful when the data to be compressed is short and can be predicted with good accuracy; the data can then be compressed better than with the default empty dictionary"

How are the strings separated? With '0' characters, etc formatting specification. How long/short strings? Which way strings should be ordered? ...

include/odp/api/spec/comp.h
```diff
@@ -0,0 +1,603 @@
...
+ * @param	pkt_in		Packets to be processed
+ * @param	pkt_out		Packet handle array for resulting packets
+ * @param	num_pkt		Number of packets to be processed
+ * @param	param		Operation parameters
+ *
+ * @return	Number of input packets enqueued (0 ... num_pkt)
+ * @retval	<0 on failure
+ *
+ * @note Both 'pkt_in' and 'pkt_out' array should be of same length, Results
+ * are undefined if otherwise.
```
Bill-Fischofer-Linaro: Again, shouldn't `odp_comp_packet_op_param_t` array also be the same length?

include/odp/api/spec/comp.h
```diff
@@ -0,0 +1,603 @@
...
+ * corresponding valid output packet handle. On return, API returns with
+ * number of entries successfully submitted for operation.
+ *
+ * In case of partially accepted array i.e.
+ * when number of packets returned < num_pkt, application may attempt to
+ * resubmit subsequent entries via calling any of the operation API.
+ *
+ * Processed data length can be computed by subtracting 'digest_len' from
+ * output_data_range.len where 'digest_len' can be queried from
+ * odp_comp_hash_alg_capability().
```
Bill-Fischofer-Linaro: Isn't the digest only the output from compress operations? Decompress operations should simply produce the uncompressed data expanded from the compressed input. What is the application expected to do with the hash? Presumably the compressed data + digest should be passed to whoever is doing to do the decompress so that it can serve as an integrity check? Does this mean we should have a defined error return that indicates the decompress operation was successful but the hash didn't compare?

include/odp/api/spec/comp.h
Bill-Fischofer-Linaro: The above comments seem equally applicable to this version of the PR. It's not at all clear how these APIs can be used to compress or decompress partial packets because it seems its undefined whether bytes not in the specified range will be altered as a result of these operations.

Should we have a simplified set of initial APIs that compress or decompress entire input packets?

include/odp/api/spec/comp.h
bala-manoharan: The dictionary support is exported as a capability field. Hence implementation can specify if they can support dictionary or not

include/odp/api/spec/comp.h
```diff
@@ -0,0 +1,603 @@
...
+ *
+ * All the packets successfully enqueue will be submitted to 'compl_queue'
+ * after compression operation, Application should check 'status' of the
+ * operation in odp_comp_packet_result_t.
+ *
+ * Please note it is always recommended that application using async mode,
+ * provide sufficiently large buffer size to avoid
+ * ODP_COMP_ERR_OUT_OF_SPACE_TERM.
```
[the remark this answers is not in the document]
bala-manoharan: This guidance does not seem to be a straightforward derivation. The size of the compression will depend on the packet data and repetition of any specific strings in the input packet.

include/odp/api/spec/comp.h
```diff
@@ -0,0 +1,603 @@
...
+ * Sets the dictionary value to the corresponding compression session.
+ *
+ * @param	session		Session handle
+ * @param	dict		Pointer to dictionary.
+
+ * @retval 0 on success
+ * @retval <0 on failure
+ *
+ * @note: Application should check dictionary support using
+ * odp_comp_alg_capability()
```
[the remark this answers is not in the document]
bala-manoharan: The dictionary value is specific to input data and hence even if we keep it as par,t of the session, we need to update for each and every input data. 
Hence it is not useful to add as part of the session parameters.

include/odp/api/spec/comp.h
```diff
@@ -0,0 +1,603 @@
...
+ * @param          pkt_in	Packets to be processed
+ * @param          pkt_out	Packet handle array for resulting packets
+ * @param          num_pkt	Number of packets to be processed
+ * @param          param	Operation parameters
+ *
+ * @return Number of input packets consumed (1 ... num_pkt)
+ * @retval <0 on failure
+ *
+ * @note Both 'pkt_in' and 'pkt_out' array should be of same length, Results
+ * are undefined if otherwise.
```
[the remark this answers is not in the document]
bala-manoharan: Yes. Thanks for pointing it out.

include/odp/api/spec/comp.h
bala-manoharan: Digest output is available during both compress and decompresses operation.
During decompression, the Hashing validation is done by the application since the compression engine does not have the hash on uncompressed data to validate.

include/odp/api/spec/comp.h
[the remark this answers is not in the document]
bogdanPricope: How common is to use the same session for multiple input data types?  Why not creating multiple sessions, each with its own data type and customized dictionary?

include/odp/api/spec/comp.h
```diff
@@ -0,0 +1,619 @@
...
+ * used strings preferably put towards the end of the dictionary.
+ * Using a dictionary is most useful when the data to be compressed is short
+ * and can be predicted with good accuracy; the data can then be compressed
+ * better than with the default empty dictionary.
+ *
+ * The compression and decompression must use exactly the same dictionary.
+ * The optimal dictionaries are application specific
+ * The dictionary may be fixed or may be chosen among a certain number of
+ * predefined dictionaries, according to the kind of input data
+ *
```
psavol: I suggest this wording for the dictionary chapter:

A dictionary contains a set of ??? null terminated ??? strings (byte sequences
delimited with zero bytes) that are likely to be encountered later in the data to be
compressed. The most commonly used strings are preferably put towards the
??? beginning OR end ??? of the dictionary. Using a dictionary is most useful
when the data to be compressed is short and can be predicted with good accuracy.
The compression and decompression must use exactly the same dictionary. The
optimal dictionaries are application specific.

data format needs decisions about:
 * back to back null terminated strings ?
 * order of strings? most relevant first ? or last ?
 * any limitations to string length: 1 bytes strings OK ?, single capa.dict_len string OK?

include/odp/api/spec/comp.h
```diff
@@ -0,0 +1,619 @@
...
+ 	 * best possible speed at the expense of compression quality
+	 *
+	 * max_level - High quality compression
+	 *
+	 * @see 'max_level' in odp_comp_alg_capability_t
+	 */
+	uint32_t comp_level;
+
+	/** huffman code to use */
+	odp_comp_huffman_code_t comp_code;
```
psavol: Maybe "huffman" or "huffman_code" is better name for the field, which defines huffman code type.

include/odp/api/spec/comp.h
```diff
@@ -0,0 +1,619 @@
...
+} odp_comp_alg_def_param_t;
```
psavol: Better type name is odp_comp_deflate_param_t. odp_comp_alg_def_param_t can be mixed to "algorithm default" parameters...

include/odp/api/spec/comp.h
```diff
@@ -0,0 +1,619 @@
...
+typedef union odp_comp_alg_param_t {
+	/** deflate parameter */
+	odp_comp_alg_def_param_t deflate;
+
+	/**
+	 * Struct for defining zlib algorithm parameters
+	 */
+	struct comp_alg_zlib_param {
+		/** deflate algo params */
+		odp_comp_alg_def_param_t def;
```
psavol: Use the same name for both deflate param fields. Preferably "deflate" for both.

include/odp/api/spec/comp.h
```diff
@@ -0,0 +1,619 @@
...
+	odp_comp_alg_t comp_algo;
+
+	/** Hash algorithm
+	 *
+	 *  @see odp_comp_alg_capability()
+	 */
+	odp_comp_hash_alg_t hash_algo;
+
+	/** parameters specific to compression */
+	odp_comp_alg_param_t algo_param;
```
psavol: Field name should match the type name. Field/type names are easier to remember / search when those match.
algo_param ==> alg_param

bala-manoharan: Will update in next version

include/odp/api/spec/comp.h
bala-manoharan: Done

include/odp/api/spec/comp.h
bala-manoharan: Okay

include/odp/api/spec/comp.h
```diff
@@ -0,0 +1,574 @@
...
+	 * true: packet session enqueue order maintained
+	 *
+	 * false: packet order is maintained on operation completion order
+	 * and not in session enqueue order.
+	 *
+	 * @note: By disabling packet order requirement, performance oriented
+	 * application can leverage HW offered parallelism to increase operation
+	 * performance.
+	 */
+	odp_bool_t packet_order;
```
psavol: It should be documented that this is only for async operation and sync operation always maintain ordered (pkt_in[0] -> pkt_out[0], ...)

It's enough to say that: "false: packet enqueue order is not maintained"

include/odp/api/spec/comp.h
```diff
@@ -0,0 +1,574 @@
...
+ * Output algorithm capabilities.
+ *
+ * @param	comp	Compression algorithm
+ * @param[out]	capa	Compression algorithm capability
+ *
+ * @retval 0 on success
+ * @retval <0 on failure
+ */
+int odp_comp_alg_capability(odp_comp_alg_t comp,
+			    odp_comp_alg_capability_t capa);
```
psavol: Missing pointer type: odp_comp_alg_capability_t *capa

include/odp/api/spec/comp.h
```diff
@@ -0,0 +1,574 @@
...
+ * This operation does packet compression in synchronous mode. A successful operation
+ * returns the number of successfully processed input packets and updates the
+ * results in the corresponding output packets. Outputted packets contain
+ * compression results metadata (odp_comp_packet_result_t), which should be
+ * checked for operation status.
+ *
+ * For every input packet entry in 'pkt_in' array, application should pass
+ * corresponding valid output packet handle. If any error occurs during
+ * processing of packets, the API returns with number of entries successfully
+ * processed.
```
bogdanPricope: Assuming 5 packets are processed, each with a different session, why failing to process 1st packet should trigger the failure for all? I mean, we have a 'result' per packet. Why not returning retval 4 and let application to get result per packet with odp_comp_result()?

include/odp/api/spec/comp.h
```diff
@@ -0,0 +1,574 @@
...
+ * Outputs all supported configuration options for the algorithm.
+ *
+ * @param	hash     Hash algorithm
+ * @param	capa     Hash algorithm capability
+ *
+ * @retval 0 on success
+ * @retval <0 on failure
+ */
+int odp_comp_hash_alg_capability(odp_comp_hash_alg_t hash,
+				 odp_comp_hash_alg_capability_t capa);
```
psavol: Missing pointer type: *capa

include/odp/api/spec/comp.h
```diff
@@ -0,0 +1,574 @@
...
+ * 'digest_len' can be queried from odp_comp_hash_alg_capability().
+ * Hash is always performed on plain text. Hash validation in decompression is
+ * performed by the application.
+ *
+ * @param          pkt_in	Packets to be processed
+ * @param          pkt_out	Packet handle array for resulting packets
+ * @param          num_pkt	Number of packets to be processed
+ * @param          param	Operation parameters
+ *
+ * @return Number of input packets consumed (1 ... num_pkt)
```
psavol: 0 is a valid return value == no packets processed (e.g. HW engine input queues full)

include/odp/api/spec/comp.h
bala-manoharan: There is a possibility of co-processor queue being full and the input packet was not even sent to the compression coprocessor, In that scenario, it might be better to just return the number of input packets processed, else implementation will have to update all the output packets even when they were not processed and the application will have to check the result for each and every output packet.

include/odp/api/spec/comp.h
bala-manoharan: Agreed

include/odp/api/spec/comp.h
psavol: Agree, output range is useful only if bytes outside of the range must not be modified during the operation.

include/odp/api/spec/comp.h
bala-manoharan: Updated

include/odp/api/spec/comp.h
bogdanPricope: OK. Makes sense.

include/odp/api/spec/comp.h
bala-manoharan: The compression engine will not access any other data in the input and output packet apart from the data_range given as input. Since the compression engine works on buffers and files the engine does not understand the concept of a packet.
output_data_range is used to specify the location on the output buffer where the compressed/decompressed output will be written. The compression engine only checks the input_data_range in input packet and compresses the same and writes the output to output packet in the output_data_range. The rest of the bytes of the input packet are not copied to output packet since the format could be completely different from input packet. Neither does the compression engine update the packet length, headroom or tailroom.

'output_data_range' is required since the entire output packet is not accessed by the compression engine and it will only update the result on output packet within the range provided. i.e application could provide output_data_range from offset = 100 and length = 200, for one input packet and can use the same output packet with data range offset = 300 and length = 200 for another input packet.

psavol: Text above needs to specify that "output packet data is not modified outside of the provided out_data_range"

bala-manoharan: I had this documentation in a previous version and somehow lost during the update. Will update in next version

include/odp/api/spec/comp.h
bala-manoharan: These comments are now shown under outdated and I have missed them. Will update in next version.

include/odp/api/spec/comp.h
```diff
@@ -0,0 +1,575 @@
...
+ */
+typedef struct odp_comp_packet_result_t {
+	/** Operation status code */
+	odp_comp_status_t status;
+
+	/** Input packet handle */
+	odp_packet_t pkt_in;
+
+	/** Output packet data range */
+	odp_packet_data_range_t out_data_range;
```
psavol: To be completely clear: "Specifies offset and length of the data that resulted from a successful de-/compression operation."

So, this is not copy-paste of the range of input params, but where the resulting data is. Maybe it needs to be highlighted that packet data pointer, length or other metadata has not been updated.

include/odp/api/spec/comp.h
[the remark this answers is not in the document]
bala-manoharan: I have updated the documentation for only ASYNC operation

include/odp/api/spec/comp.h
```diff
@@ -0,0 +1,603 @@
...
+
+/**
+ * Query supported compression algorithm capabilities
+ *
+ * Output algorithm capabilities.
+ *
+ * @param	comp	Compression algorithm
+ * @param[out]	capa	Compression algorithm capability
+ *
+ * @retval <0 on failure
```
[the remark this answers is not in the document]
bala-manoharan: Updated

include/odp/api/spec/comp.h
```diff
@@ -0,0 +1,603 @@
...
+
+/**
+ * Query supported hash algorithm capabilities
+ *
+ * Outputs all supported configuration options for the algorithm.
+ *
+ * @param	hash     Hash algorithm
+ * @param	capa     Hash algorithm capability
+ *
+ * @retval <0 on failure
```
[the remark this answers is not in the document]
bala-manoharan: Updated

include/odp/api/spec/comp.h
bala-manoharan: Yes. Updated

include/odp/api/spec/comp.h
bala-manoharan: Updated.

include/odp/api/spec/comp.h
bala-manoharan: Updated

include/odp/api/spec/comp.h
bala-manoharan: Updated in the next version.

psavol:
API is in good enough shape for first iteration of implementation / testing. I think there's only two open questions left from me and those will be fixed in v15.
* same packet handle cannot be passed as an input and output
* per algorithm capability needed for application to estimate output packet size (at least for decomp direction)

bogdanPricope:
I lost track of the changes at some point. What was the conclusion on:
- dictionary structure / set API
- partially processed packet ?

bogdanPricope:

include/odp/api/spec/comp.h
```diff
@@ -0,0 +1,608 @@
...
+ *
+ * @param          pkt_in	Packets to be processed
+ * @param          pkt_out	Packet handle array for resulting packets
+ * @param          num_pkt	Number of packets to be processed
+ * @param          param	Operation parameters
+ *
+ * @return Number of input packets consumed (0 ... num_pkt)
+ * @retval <0 on failure
+ *
+ * @note The 'pkt_in','pkt_out'and 'param' array should be of same length,
```
Bill-Fischofer-Linaro: s/array/arrays/

include/odp/api/spec/comp.h
```diff
@@ -0,0 +1,608 @@
...
+ * Results are undefined if otherwise.
+
+ * @note Same packet handle cannot be used as input and output parameter.
+ * In-place compression operation is not supported
```
Bill-Fischofer-Linaro: This means that the caller must copy the input packet to the output packet prior to calling `odp_comp_op()` to operate on a data range within the `pkt_in`.  (I'm assuming that this copying is not performed by this API itself?) Perhaps that's OK for an initial cut but I'd think this again is something that we'd want the implementation to handle behind the scenes to avoid that sort of overhead.

include/odp/api/spec/comp.h
```diff
@@ -0,0 +1,608 @@
...
+ * creation. For every input packet entry in in_pkt array, user should pass
+ * corresponding valid output packet handle. On return, API returns with
+ * number of entries successfully submitted for operation.
+ *
+ * In case of partially accepted array i.e.
+ * when number of packets returned < num_pkt, application may attempt to
+ * resubmit subsequent entries via calling any of the operation API.
+ *
+ * When hashing is configured along with compression operation the
+ * result is appended at the end of the output packet. Processed data length
```
Bill-Fischofer-Linaro: What does "at the end of the output packet" mean, since these APIs do not change packet lengths? Shouldn't this be "at the end of the specified `out_data_range`"?

include/odp/api/spec/comp.h
```diff
@@ -0,0 +1,608 @@
...
+ * can be computed by subtracting 'digest_len' from output_data_range.len where
+ * 'digest_len' can be queried from odp_comp_hash_alg_capability().
+ * Hash is always performed on plain text. Hash validation in decompression is
```
Bill-Fischofer-Linaro: What does "plain text" mean in this context? Did you mean it's performed on the decompressed data?

include/odp/api/spec/comp.h
```diff
@@ -0,0 +1,608 @@
...
+ * performed by the application.
+ *
+ * All the packets successfully enqueue will be submitted to 'compl_queue'
```
Bill-Fischofer-Linaro: s/enqueue/enqueued/

include/odp/api/spec/comp.h
bogdanPricope: "In cryptography, plaintext or cleartext is unencrypted information, as opposed to information encrypted for storage or transmission. Plaintext usually means unencrypted information pending input into cryptographic algorithms, usually encryption algorithms. Cleartext usually refers to data that is transmitted or stored unencrypted ('in the clear')."  See https://en.wikipedia.org/wiki/Plaintext

Bill-Fischofer-Linaro: @bogdanPricope I know, but it's strange to see cryptogrphic terms used in a compression API since compression and encryption are very different operations. Domain-specific terms would seem better here to avoid confusion.

include/odp/api/spec/comp.h
bala-manoharan: Yes. This should refer to end of out_data_range. Will change in the next version.

include/odp/api/spec/comp.h
bala-manoharan: This is not a requirement to copy all the input packet parameters to output packet. It depends on the use-case. If let's say the application is combining several compressed input packets to a single output packet then the meta-data of input and output packet might be different.
If there is an application need we could create an API outside of this comp scope odp_packet_copy_metadata() which will copy input packet metadata to output packet

include/odp/api/spec/comp.h
bala-manoharan: Done.

include/odp/api/spec/comp.h
Bill-Fischofer-Linaro: OK, but we need to be precise here. If the `out_data_range` is specified as length 100 but the decompression consumes 50 bytes of that range that should mean that the returned `out_data_range` indicates a length of 54 bytes where the last 4 bytes are the hash. If the last four bytes of the original 100 length were used for the hash then the application would have no way of knowing how many bytes of decompressed data were returned.

include/odp/api/spec/comp.h
bogdanPricope: True, yet compression/decompression is similar in a way... I don't know if there is a more appropriate  word for this.

bogdanPricope: The other definition: "In computing, plain text is the data (e.g. file contents) that represent only characters of readable material but not its graphical representation nor other objects (images, etc.). It may also include a limited number of characters that control simple arrangement of text, such as line breaks or tabulation characters. Plain text is different from formatted text, where style information is included, and from "binary files" in which some portions must be interpreted as binary objects (encoded integers, real numbers, images, etc.)."  https://en.wikipedia.org/wiki/Plain_text

include/odp/api/spec/comp.h
bala-manoharan: Done

include/odp/api/spec/comp.h
```diff
@@ -0,0 +1,608 @@
...
+	 * Indicates where processed packet will be written. where,
+	 *
+	 * offset - starting offset
+	 * length - length of buffer available for output
+	 *
+	 * Output packet data is not modified outside of this provided data
+	 * range. If output data length is not sufficient for compression
+	 * operation ODP_COMP_STATUS_OUT_OF_SPACE_TERM error will occur
+	 */
+	odp_packet_data_range_t out_data_range;
```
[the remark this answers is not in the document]
bala-manoharan: There is no use-case for in-place compression.

include/odp/api/spec/comp.h
bala-manoharan: "plain text" refers to un-compressed or decompressed data. I believe the context of the plaintext is explanatory in this compression documentation.

include/odp/api/spec/comp.h
bala-manoharan: The hash result is appended at the end of the compressed data and the out_data_range.length is the length of compressed data + digest_length.

include/odp/api/spec/comp.h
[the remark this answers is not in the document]
Bill-Fischofer-Linaro: Ok.

muvarov:
Merged.

PR #581, opened 2018-04-25
[PATCH v1] Fix timer pool index allocation bug

psavol:
Fixes https://bugs.linaro.org/show_bug.cgi?id=3772

### DIFF
--- a/platform/linux-generic/odp_timer.c
+++ b/platform/linux-generic/odp_timer.c
@@ -197,7 +197,9 @@ typedef struct timer_pool_s {
 	int timer_thread_exit; /* request to exit for timer thread */
 } timer_pool_t;
 
-#define MAX_TIMER_POOLS 255 /* Leave one for ODP_TIMER_INVALID */
+/* Timer pool index must fit into 8 bits with one index value reserved to
+ * ODP_TIMER_POOL_INVALID. */
+#define MAX_TIMER_POOLS 32
 #define INDEX_BITS 24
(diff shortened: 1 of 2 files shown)


platform/linux-generic/odp_timer.c
```diff
@@ -197,7 +197,9 @@ typedef struct timer_pool_s {
 	int timer_thread_exit; /* request to exit for timer thread */
 } timer_pool_t;
 
-#define MAX_TIMER_POOLS 255 /* Leave one for ODP_TIMER_INVALID */
+/* Timer pool index must fit into 8 bits with one index value reserved to
+ * ODP_TIMER_POOL_INVALID. */
+#define MAX_TIMER_POOLS 32
```
Bill-Fischofer-Linaro: This patch isn't really part of the Bug fix, but is a separate enhancement. As such it really should be a separate PR.

psavol: The bug fix is the first patch. Patches 2 and 3 are in the same patch set as they are related changes. This patch decreases the size of newly introduced global struct and the last patch adds a new test case. It would be odd to send this as a separate patch as it touches the same feature (timer pool allocation).

Bill-Fischofer-Linaro: OK.

muvarov:
Merged. @psavol  please add bugz url inside git commit. In that case bugzilla will be automatically updated with commit entry.

PR #582, opened 2018-04-25
[PATCH v3] linux-gen: pktio: dpdk: add dpdk packet parser

bogdanPricope:
Add dpdk packet parser.

### DIFF
--- a/platform/linux-generic/include/odp_packet_dpdk.h
+++ b/platform/linux-generic/include/odp_packet_dpdk.h
@@ -72,4 +72,22 @@ typedef struct ODP_ALIGNED_CACHE {
 	dpdk_opt_t opt;
 } pkt_dpdk_t;
 
+/** Packet parser using DPDK interface */
+int dpdk_packet_parse_common(packet_parser_t *pkt_hdr,
+			     const uint8_t *ptr,
+			     uint32_t pkt_len,
+			     uint32_t seg_len,
+			     struct rte_mbuf *mbuf,
+			     int layer);
+
+static inline int dpdk_packet_parse_layer(odp_packet_hdr_t *pkt_hdr,
+					  struct rte_mbuf *mbuf,
+					  odp_pktio_parser_layer_t layer)
+{
+	uint32_t seg_len = pkt_hdr->buf_hdr.seg[0].len;
+	void *base = pkt_hdr->buf_hdr.seg[0].data;
+
+	return dpdk_packet_parse_common(&pkt_hdr->p, base, pkt_hdr->frame_len,
+					seg_len, mbuf, layer);
+}
 #endif
--- a/platform/linux-generic/include/odp_packet_dpdk.h
+++ b/platform/linux-generic/include/odp_packet_dpdk.h
@@ -78,16 +78,18 @@ int dpdk_packet_parse_common(packet_parser_t *pkt_hdr,
 			     uint32_t pkt_len,
 			     uint32_t seg_len,
 			     struct rte_mbuf *mbuf,
-			     int layer);
+			     int layer,
+			     odp_pktin_config_opt_t pktin_cfg);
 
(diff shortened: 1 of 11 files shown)


platform/linux-generic/pktio/dpdk.c
```diff
@@ -549,7 +553,7 @@ static inline int mbuf_to_pkt(pktio_entry_t *pktio_entry,
 		if (pktio_cls_enabled(pktio_entry))
 			copy_packet_cls_metadata(&parsed_hdr, pkt_hdr);
 		else if (parse_layer != ODP_PROTO_LAYER_NONE)
-			packet_parse_layer(pkt_hdr, parse_layer);
+			dpdk_packet_parse_layer(pkt_hdr, mbuf, parse_layer);
```
lumag: I have been thinking about similar change in generic code. However in generic code some PktIOs call parsing function using buffer, which is not a part of the packet (yet). So it is not possible to always pass odp_packet_t/odp_buffer_t to parsing functions.

bogdanPricope:
@lumag  Sorry, I don't get your point. Can you elaborate your comment (what len, what parse flags) ?

bogdanPricope:
@lumag  Yes,I liked it too... yet, was dropped due review (not changing all pktios makes sense, also).

lumag:
@bogdanPricope what kind of response did it receive? I don't see any comment here.

bogdanPricope:
@lumag  It was in an 'ODP Arch Sync' meeting: suggestion from Bill was to avoid adding repetitive code sequence. (packet_parse_reset(&parsed_hdr);packet_set_len(&parsed_hdr, pkt_len);)  but add this flag to parse or not internally.

lumag:
Reviewed-by: Dmtiry Eremin-Solenikov

matiaselo:
I run quick tests with odp_l2fwd (dpdk-zero-copy + disable-abi-compat, odp_l2fwd -a 0 -i 0,1 -m 0 -e 1 -k 1 -c N) and with these patches the performance is on average ~4% lower than before. I'm using Intel XL710 NICs.

platform/linux-generic/include/odp_packet_dpdk.h
```diff
@@ -72,4 +72,15 @@ typedef struct ODP_ALIGNED_CACHE {
 	dpdk_opt_t opt;
 } pkt_dpdk_t;
 
+/** Packet parser using DPDK interface */
+int dpdk_packet_parse_layer(odp_packet_hdr_t *pkt_hdr,
```
matiaselo: Should be inlined.

bogdanPricope:
@matiaselo  I am not really surprised about this 4% drop as it implements more features than current parser: drop packets on ipv4/ipv6/udp/tcp errors is not implemented in current parser (we are not comparing the same things). More, the gain on using rte_mbuf packet_type and ol_flags is not so big, while moving csum tests inside parser is not removing many conditionals. 

Point is, this parser is making code cleaner, cleanly extensible with dpdk calls (if needed) and add the missing feature; it may be useful for odp-dpdk as well.

And yes, the drop can be recuperated with inlines and other tricks.

muvarov:
lets discuss this PR on Monday one more time.

matiaselo:
>Point is, this parser is making code cleaner, cleanly extensible with dpdk calls (if needed) and add the missing feature

This is all good and I'm not against merging this patch set. I'm just pointing out there are still optimizations to be made. I can live with the small performance degradation for now, as it doesn't effect raw throughput when packet parser is disabled.

The commit log for `linux-gen: pktio: dpdk: add specific packet parser` patch is misleading as it only states that parsing performance is increased.

> it may be useful for odp-dpdk as well.

This should definitely be implemented for odp-dpdk as well.

bogdanPricope:
Technically, only third patch (linux-gen: pktio: dpdk: integrate csum with packet parser) is reducing performance… and for that patch I already updated the commit log.

muvarov:
Merged.

PR #583, opened 2018-04-25
[PATCH v1] test: crypto perf: add missing breaks to switch case

psavol:
Breaks are needed for switch-cases. Missing breaks were
breaking build for GCC 7.2.


lumag:

muvarov:
Merged.

PR #584, opened 2018-04-27
sched_pktio test development

psavol:
Developed sched_pktio test further to maintain packet order as well as added -q and -t options. The first selecting packet IO interface queue count and second enabled timer performance testing with packet IO. Also found and corrected a bug in timer: it did not set highest resolution capability in scheduler polled mode (inline_timers=true). Application modifications are tested with the bug correction, hence it's included into the patch set.


platform/linux-generic/odp_timer.c
```diff
@@ -76,7 +76,7 @@ static _odp_atomic_flag_t locks[NUM_LOCKS]; /* Multiple locks per cache line! */
 #endif
 
 /* Max timer resolution in nanoseconds */
-static uint64_t highest_res_ns;
+static uint64_t highest_res_ns = 500;
```
Bill-Fischofer-Linaro: This doesn't do what you say you want since the `timer_res_init()` routine overwrites this. Sounds like what you want is to introduce a `max_res_ns` limit (setting it to 500) and then use that to override the computed `highest_res_ns` if `timer_res_init()` computes a lower value. So timer pools may have a higher `highest_res_ns` but they cannot have one less than `max_res_ns`.

test/performance/odp_sched_pktio.c
```diff
@@ -186,6 +196,121 @@ static int worker_thread(void *arg)
...
+		timer = test_global->timer.timer[src_pktio][src_queue];
+
+		for (i = 0; i < num; i++) {
+			if (odp_unlikely(odp_event_type(ev[i]) ==
+					 ODP_EVENT_TIMEOUT)) {
+				tmos++;
+				ret = odp_timer_set_rel(timer, tick, &ev[i]);
+
+				if (ret != ODP_TIMER_SUCCESS) {
+					/* Such never happen */
```
Bill-Fischofer-Linaro: Typo: Should never happen

test/performance/odp_sched_pktio.c
```diff
@@ -186,6 +196,121 @@ static int worker_thread(void *arg)
...
+			}
+		}
+
+		if (tmos == 0) {
+			/* Reset timer with existing timeout event */
+			ret = odp_timer_set_rel(timer, tick, NULL);
+
+			if (ret != ODP_TIMER_SUCCESS &&
+			    ret != ODP_TIMER_TOOEARLY) {
+				/* Such never happen. Reset should either
```
Bill-Fischofer-Linaro: Typo: Should never happen.

platform/linux-generic/odp_timer.c
psavol: When timer uses posix timer, this value is measured and I'm OK what ever the measurement results (won't be nano- but milliseconds anyway). When timer won't use posix but rely on polls from scheduler, this value is not measured and was left to zero. This fix is for that mode.

test/performance/odp_sched_pktio.c
psavol: OK. Fixed in v2.

test/performance/odp_sched_pktio.c
psavol: Fixed all three in v2.

platform/linux-generic/odp_timer.c
Bill-Fischofer-Linaro: OK, I missed that subtlety in the code. A comment would be useful to make it clear what the intent is here.

muvarov:
Merged

PR #585, opened 2018-04-28
[PATCH v1] Travis: do not run gcov before executing codecov script

lumag:
There is no need to manually run gcov from Travis script, codecov.io
will do that for us.


muvarov:
Merged

PR #586, opened 2018-04-28
[PATCH v1] doc: allow user to provide exta asciidoctor flags

lumag:
Allow user to provide extra asciidoctor flags. E.g. this allows one to
override icons and/or fonts placement.


muvarov:
Don't know what is real use case for that. But if you need to change style I don't mind to merge it. Merged!

lumag:
@muvarov I had to pass additional arguments to build docs suitable for debian (`-a iconfont-cdn=LOCAL_PATH -a webfonts!`), so that it does not reference external resources.

PR #589, opened 2018-05-04
[PATCH v1] linux-gen: pktio: ipc free ring memory on error

muvarov:
ipc pktio can fail on open if second process is not
yet ready to connect. But on this error free of ring
memory is missing which leads to taking all available shm
descriptors.
https://bugs.linaro.org/show_bug.cgi?id=3788


muvarov:
Merged.

PR #590, opened 2018-05-04
[PATCH v1] linux-gen: packet: rearrange odp_packet_hdr_t members

matiaselo:
Save 64 bytes by rearranging odp_packet_hdr_t members.

### DIFF
--- a/platform/linux-generic/include/odp_packet_internal.h
+++ b/platform/linux-generic/include/odp_packet_internal.h
@@ -96,6 +96,9 @@ typedef struct {
 	 * Members below are not initialized by packet_init()
 	 */
 
+	/* Type of extra data */
+	uint8_t extra_type;
+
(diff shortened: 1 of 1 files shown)


platform/linux-generic/include/odp_packet_internal.h
```diff
@@ -96,6 +96,9 @@ typedef struct {
 	 * Members below are not initialized by packet_init()
 	 */
 
+	/* Type of extra data */
+	uint8_t extra_type;
```
Bill-Fischofer-Linaro: We still have 16 bits of padding between this field and `flow_hash`. Should that be explicit? I'm Ok either way, but if we're trying to pack the header having everything mapped explicitly would seem desirable.

muvarov:
Merged.

PR #591, opened 2018-05-04
[PATCH v2] linux-gen: fix compilation warnings

joseppc:
gcc 8 complains when strncpy may truncate the destination string,
possibly not copying the null terminating character.

This solves these two compilation warnings:
odp_fdserver.c:591:2: error: ‘strncpy’ output may be truncated
  copying 108 bytes from a string of length 254 [-Werror=stringop-truncation]
odp_ishm.c:478:3: error: ‘strncpy’ output may be truncated
  copying 191 bytes from a string of length 191 [-Werror=stringop-truncation]

### DIFF
--- a/platform/linux-generic/odp_fdserver.c
+++ b/platform/linux-generic/odp_fdserver.c
@@ -588,7 +588,9 @@ int _odp_fdserver_init_global(void)
 
 	/* bind to new named socket: */
 	local.sun_family = AF_UNIX;
-	strncpy(local.sun_path, sockpath, sizeof(local.sun_path));
+	memcpy(local.sun_path, sockpath, sizeof(local.sun_path));
+	local.sun_path[sizeof(local.sun_path) - 1] = '\0';
+
 	res = bind(sock, (struct sockaddr *)&local, sizeof(struct sockaddr_un));
 	if (res == -1) {
 		ODP_ERR("_odp_fdserver_init_global: %s\n", strerror(errno));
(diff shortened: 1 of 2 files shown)


joseppc:
Yes, I saw the warning, I hope Maxim can fix it before committing the patch, so we don't have to run all the tests again.

I don't know if this is the complete list of gcc8 warnings we are going to get, these are just the ones I stumbled upon.

muvarov:
@joseppc it looks like local.sun_len setting is missing. I think it's better to add memset(&local, 0, ..) and strncpy(local.sun_path, sockpath, sizeof(local.sun_path) - 1);  I think that can keep gcc8 also happy.

joseppc:
@muvarov I don't exactly understand what you mean with the "local.sun_len setting is missing" (SUN_LEN macro?). In any case it doesn't seem to help, but we could use memcpy instead of strncpy to fool gcc8, I'll resend a patch with that.

muvarov:
@joseppc  memset(0 for this is missing. So there are some garbage without it.

joseppc:
@muvarov In both cases the string being copied is null-terminated already, so copying extra bytes should not matter.
In `odp_ishm.c` the strings are the same size, so I don't need to care about the last byte, in `odp_fdserver.c`, for `local.sun_path`, the string copied may be longer than the size of `sun_path`, so I do need to set the last byte to 0 just to be sure.

muvarov:
Merged

PR #593, opened 2018-05-07
[PATCH v1] Event type function inline

psavol:
Use event type function as an example how API functions could be inlined automatically inside implementation (by including xxx_inlines.h).


platform/linux-generic/include/odp/api/plat/event_inlines.h
```diff
@@ -0,0 +1,37 @@
...
+#include <odp/api/plat/buffer_inline_types.h>
+
+/** @cond _ODP_HIDE_FROM_DOXYGEN_ */
+
+extern const _odp_buffer_inline_offset_t _odp_buffer_inline_offset;
+
+#ifndef _ODP_NO_INLINE
+	/* Inline functions by default */
+	#define _ODP_INLINE static inline
+	#define odp_event_type __odp_event_type
```
lumag: It might be better to put this `#define` right before the function.

platform/linux-generic/odp_crypto.c
```diff
@@ -20,6 +20,9 @@
 #include <odp/api/plat/packet_inlines.h>
 #include <odp_packet_internal.h>
 
+/* Inlined API functions */
+#include <odp/api/plat/event_inlines.h>
```
lumag: I had a different approach to this issue. It is quite easy to forget to actually include the `_inlines.h` header file. My suggestion is to create a set of `<odp_FOO.h>` headers used only in platform code, which include both `<odp/api/FOO.h>` and `<odp/api/plat/FOO_inlines.h>`. Platform code will always include the former header, which will pull inlineable functions.

platform/linux-generic/include/odp/api/plat/event_inlines.h
Bill-Fischofer-Linaro: Actually, I'd expect this to be in a common file rather than in each `xxx_inline.h` file. Presumably it's here because that was simpler to illustrate the alternate pattern given that only events are using this structure for now.

platform/linux-generic/odp_crypto.c
Bill-Fischofer-Linaro: @lumag That would also have the advantage that new functions can be inlined without requiring any changes in those modules that make use of the newly-inlined functions. So these can be optimized individually without chasing who uses them.

platform/linux-generic/include/odp/api/plat/event_inlines.h
psavol: I think it's cleaner to have on one ifdef in the file and gather all function renames under that. It's hard to read a file with N functions and N ifdefs scattered around the file (a ifdef before each function).

Bill, these rename defines are file specific (no common rename header). Also it does not make this file simpler or more readable, if _ODP_INLINE would be defined in another header file and then that file would be included here - one line define vs one line include. Downside of an include is that the reader of this file would need to go and check that file also and cross check an ifdef there.

platform/linux-generic/odp_crypto.c
psavol: Bill, in both approaches each c file would need only one change: include odp_foo.h or odp/api/plat/foo_inlines.h.

I noticed odp_foo.h, but I think it's name should tell that it's used for getting access to inline implementation (so odp_foo_inlines.h). When foo module includes already odp/api/foo.h, it unnecessary to create another file that include the same again (plus inlines.h). Also I think it's more readable (and flexible) to include in each c file explicitly include those headers and only those headers it needs. Also we might not want to inline every API in every c file. This explicit include of odp/api/plat/foo_inlines.h documents when a c file uses foo inlines, and when it does not use those.

lumag: @psavol fine, let's see, how this will work.

lumag:

muvarov:
Merged.

PR #596, opened 2018-05-14
[PATCH v1] helper: thread: don't use getopt library

psavol:
Don't use getopt library calls for helper options. Getopt
library reset is not portable in practice. Library interface
includes global variables and different internal state
variables depending on POSIX version. There's no need for helper
to use getopt calls or set getopt global variables at all. It's
much more simple (and portable) to remove helper options from
argv[] and return new value for argc.


lumag:

muvarov:
Merged.

PR #599, opened 2018-05-18
[PATCH v1] codecov: disable check by patch

muvarov:
Inside PRs there is check for code coverage for patch lines.
There is no big reason for that check due to patch also can
only remove lines. Only complete project has to be checked
for code coverage.


muvarov:
Merged.

PR #601, opened 2018-05-21
[PATCH v3] example: generator: add UDP port range support

bogdanPricope:
Add options to generate UDP packets with source and destination ports in a range of values. 
This will simulate multiple packet flows (that can be hashed to different queues on RX side).

### DIFF
--- a/example/generator/odp_generator.c
+++ b/example/generator/odp_generator.c
@@ -74,7 +74,9 @@ typedef struct {
 	unsigned int srcip;	/**< src ip addr */
 	unsigned int dstip;	/**< dest ip addr */
 	uint16_t srcport;	/**< src udp port */
+	uint16_t srcport_end;	/**< src udp end port */
 	uint16_t dstport;	/**< dest udp port */
+	uint16_t dstport_end;	/**< dest udp end port */
 	int mode;		/**< work mode */
 	int number;		/**< packets number to be sent */
 	int payload;		/**< data len */
@@ -102,6 +104,17 @@ typedef struct {
 	uint64_t ctr_icmp_reply_rcv;	/**< icmp reply packets */
 } counters_t;
 
+/** UDP Packet processing function argument */
+typedef struct {
+	odp_bool_t multi_flow;
+	uint16_t srcport_crt;
+	uint16_t srcport_start;
+	uint16_t srcport_end;
+	uint16_t dstport_crt;
+	uint16_t dstport_start;
+	uint16_t dstport_end;
+} udp_args_t;
+
 /** * Thread specific arguments
  */
 typedef struct {
@@ -111,6 +124,7 @@ typedef struct {
 		struct {
 			odp_pktout_queue_t pktout; /**< Packet output queue */
 			odp_pktout_config_opt_t *pktout_cfg; /**< Packet output config*/
+			udp_args_t udp_param;  /**< UDP configuration */
 		} tx;
 		struct {
 			odp_pktin_queue_t pktin; /**< Packet input queue */
@@ -148,7 +162,7 @@ static odp_barrier_t barrier;
 typedef odp_packet_t (*setup_pkt_ref_fn_t)(odp_pool_t,
 					   odp_pktout_config_opt_t *);
 typedef int (*setup_pkt_fn_t)(odp_packet_t, odp_pktout_config_opt_t *,
-			      counters_t *);
+			      counters_t *, void *);
 
 /* helper funcs */
 static void parse_args(int argc, char *argv[], appl_args_t *appl_args);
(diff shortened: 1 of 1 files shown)


example/generator/odp_generator.c
```diff
@@ -458,7 +497,7 @@ static odp_packet_t setup_icmp_pkt_ref(odp_pool_t pool,
  */
 static int setup_icmp_pkt(odp_packet_t pkt,
 			  odp_pktout_config_opt_t *pktout_cfg,
-			  counters_t *counters)
+			  counters_t *counters, void *arg)
```
lumag: I'd suggest to use `ODP_UNUSED` here, rather than just `(void) arg`.

example/generator/odp_generator.c
```diff
@@ -1333,6 +1376,13 @@ int main(int argc, char *argv[])
 
 	} else {
 		int cpu = odp_cpumask_first(&cpumask);
+		udp_args_t *udp_param = NULL;
+		uint16_t sport_range = args->appl.srcport_end -
+			args->appl.srcport + 1;
+		uint16_t dport_range = args->appl.dstport_end -
+			args->appl.dstport + 1;
+		float sport_step = (float)(sport_range) / num_workers;
```
lumag: Is it possible to use integer math here?

example/generator/odp_generator.c
```diff
@@ -1733,7 +1819,9 @@ static void usage(char *progname)
 	       "Optional OPTIONS\n"
 	       "  -h, --help       Display help and exit.\n"
 	       "  -e, --srcport src udp port\n"
+	       "  -j, --srcport_end src udp end port\n"
 	       "  -f, --dstport dst udp port\n"
+	       "  -k, --dstport_end dst udp end port\n"
```
lumag: Note, that they default to corresponding start ports

example/generator/odp_generator.c
```diff
@@ -379,6 +395,29 @@ static int setup_udp_pkt(odp_packet_t pkt, odp_pktout_config_opt_t *pktout_cfg,
...
+	    (udp_arg->dstport_start != udp_arg->dstport_end)) {
+		odph_udphdr_t *udp = (odph_udphdr_t *)(buf + ODPH_ETHHDR_LEN +
+						       ODPH_IPV4HDR_LEN);
+
+		if (udp_arg->srcport_start != udp_arg->srcport_end) {
+			udp->src_port = odp_cpu_to_be_16(udp_arg->srcport_crt);
+			if (udp_arg->srcport_crt >= udp_arg->srcport_end)
+				udp_arg->srcport_crt = udp_arg->srcport_start;
+			else
+				udp_arg->srcport_crt++;
```
muvarov: So do all threads inc the single port counter variable? I think for performance reasons there has to be counter per thread  instead of sharing global variable.

bogdanPricope: It is one per thread already. See udp_param = &args->thread[i].tx.udp_param;

example/generator/odp_generator.c
bogdanPricope: performance -  it is called only once before all the workers start, so is not a performance problem
functionality -  it is useful when you have less ports than workers (when you need  to split 3 ports to 7 workers)

example/generator/odp_generator.c
bogdanPricope: yes, can be mentioned...

example/generator/odp_generator.c
```diff
@@ -361,11 +376,12 @@ static odp_packet_t setup_udp_pkt_ref(odp_pool_t pool,
  * @retval 0 on success, -1 on fail
  */
 static int setup_udp_pkt(odp_packet_t pkt, odp_pktout_config_opt_t *pktout_cfg,
-			 counters_t *counters)
+			 counters_t *counters, void *arg)
 {
 	char *buf;
 	odph_ipv4hdr_t *ip;
 	unsigned short seq;
+	udp_args_t *udp_arg = (udp_args_t *)arg;
```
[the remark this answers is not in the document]
bogdanPricope: setup_udp_pkt() and setup_icmp_pkt() are arguments of setup_pkt_array() function:
int setup_pkt_array(.... setup_pkt_fn_t setup_pkt,, void *setup_pkt_arg)
That argument has 'void *' type to accommodate different kinds of function arguments: udp_args_t for udp function and not used for icmp function.
This design may be further extended with other kind of packet setup functions/arguments

example/generator/odp_generator.c
bogdanPricope: Application is keeping a set of per thread variables:

typedef struct {
.................
	/** Thread specific arguments */
	thread_args_t thread[MAX_WORKERS];
................
} args_t;

typedef struct {
.................
      union {
		struct {
......................
			udp_args_t udp_param;  /**< UDP configuration */
		} tx;
............................
	};
.................
} thread_args_t;

So, port counters are kept per thread.

example/generator/odp_generator.c
bogdanPricope: OK

bogdanPricope:
@muvarov Packet setup functions will 'know' more that they need to... don't like it from design point of view.

example/generator/odp_generator.c
lumag: ack

example/generator/odp_generator.c
```diff
@@ -1733,7 +1822,13 @@ static void usage(char *progname)
 	       "Optional OPTIONS\n"
 	       "  -h, --help       Display help and exit.\n"
 	       "  -e, --srcport src udp port\n"
```
lumag: What about "src udp start port"?

bogdanPricope: I am not a fan of abbreviation. It is not part of this patch but I guess we can make it more verbose in next version.

lumag: @bogdanPricope For me it is a problem of having an "end", but not having "start" corresponding to that "end".

lumag: Yes, this is what I expected. Thank you!

lumag:
Sure.

muvarov:
Merged.

PR #602, opened 2018-05-25
Inline CPU functions

psavol:
Inline CPU functions on x86 (cycle_diff is inlined on all). Add check for x86 rdtsc instruction to global init phase, so that application exits right away instead of crashing, if rdtsc is not supported. Before that, clean up global init header file (split odp_internal.h).


muvarov:
On thunder-x TM test failed with:
Test failed at line  3205  in file  traffic_mngr.c :  (wred_pkt_cnts->min_cnt <= pkts_sent) && (pkts_sent <= wred_pkt_cnts->max_cnt) 

But actually result has to be skipped:
if [ "${CI}" = "true" ] && [ $ret -eq 255 ]; then
	echo "SKIP: skip due to not isolated environment"
	exit ${TEST_SKIPPED}
fi

export CI=true 

I did not find FAILED log and see:
SKIP: traffic_mngr/traffic_mngr.sh

it looks like log parser from cunit to junit interpreted test results as failure because c code trapped on CU_ASSERT().

PR #604, opened 2018-05-29
Inline queue context function

psavol:
Clean up queue handle vs qentry pointer conversions. Change both basic and scalable queues to use qentry pointer as handle. Add common inline code for queue context function.

### DIFF
--- a/platform/linux-generic/odp_queue_scalable.c
+++ b/platform/linux-generic/odp_queue_scalable.c
@@ -52,11 +52,6 @@ typedef struct queue_table_t {
 static queue_table_t *queue_tbl;
 static _odp_ishm_pool_t *queue_shm_pool;
 
-static inline odp_queue_t queue_from_id(uint32_t queue_id)
-{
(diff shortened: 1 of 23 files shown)


platform/linux-generic/odp_queue_scalable.c
```diff
@@ -246,7 +241,7 @@ static int queue_init_global(void)
 		queue = get_qentry(i);
 		LOCK_INIT(&queue->s.lock);
 		queue->s.index  = i;
-		queue->s.handle = queue_from_id(i);
+		queue->s.handle = (odp_queue_t)queue;
```
lumag: `queue_to_ext()` here?

lumag:

platform/linux-generic/odp_queue_scalable.c
psavol: I think it's better to show explicitly that handle is qentry pointer. Actually, this qentry->s.handle is not needed as it's value is qentry itself. But didn't remove it now to limit the number of changes (and it's easy to change back if needed).

PR #605, opened 2018-05-29
[PATCH v2] tm: validate return codes

muvarov:
validate return codes and print more verbose error messages. Also skip test result is CI=true env variable is set. Propose is to capture bug packets values on thunder-x:
Test failed at line 3205 in file traffic_mngr.c : (wred_pkt_cnts->min_cnt <= pkts_sent) && (pkts_sent <= wred_pkt_cnts->max_cnt)


lumag:
@muvarov can we drop shell wrapper script together with this patch?

muvarov:
@lumag yes, shell wrapped now can be dropped. will send v2.

muvarov:
Merged.

PR #606, opened 2018-05-30
[PATCH v1] Inline time API functions

psavol:


platform/linux-generic/arch/x86/odp/api/abi/cpu_rdtsc.h
```diff
@@ -0,0 +1,29 @@
+/* Copyright (c) 2018, Linaro Limited
+ * All rights reserved.
+ *
+ * SPDX-License-Identifier:     BSD-3-Clause
+ */
+
+#ifndef ODP_ARCH_CPU_RDTSC_H_
+#define ODP_ARCH_CPU_RDTSC_H_
```
lumag: I'm quite unsure about this header. It stays in the global `<odp/api/abi/FOO.h>` place. Maybe it can be moved to `<odp/api/abi/plat/....>`?

psavol: The content is arch specific (x86 assembly). We have all arch specific implementation files under linux-generic/arch/xxx dirs. It's under odp/api/abi since it's included in API inline headers and needs to be installed (cannot be under arch/x86 directly). This is included only from other headers in the same arch dir, not from API/ABI spec headers.

lumag:

I will add inlining support for ARMv8 after this PR is merged (if nobody does that before).

muvarov:
Merged.

PR #607, opened 2018-05-30
[PATCH v4] Switch pktio to use private data field

lumag:
Instead of publishing all pkt_FOO_t types in global header, make each of ptkio's hide private data structure.


platform/linux-generic/include/odp_packet_io_internal.h
```diff
@@ -112,6 +112,8 @@ typedef	struct {
 					_ipc_map_remote_pool() */
 } _ipc_pktio_t;
 
+#define PKTIO_PRIVATE_SIZE 384
```
Bill-Fischofer-Linaro: Can this be a computed value. _e.g._, `sizeof` some union? Fixed sizes like this seem error-prone in the face of potential future growth.

platform/linux-generic/pktio/loop.c
```diff
@@ -30,6 +30,21 @@
...
+	odp_bool_t promisc;		/**< promiscuous mode state */
+	uint8_t idx;			/**< index of "loop" device */
+} pkt_loop_t;
+
+ODP_STATIC_ASSERT(PKTIO_PRIVATE_SIZE >= sizeof(pkt_loop_t),
+		  "PKTIO_PRIVATE_SIZE too small");
+
+static inline pkt_loop_t *pkt_priv(pktio_entry_t *pktio_entry)
+{
+	return (pkt_loop_t*)(uintptr_t)(pktio_entry->s.pkt_priv);
```
Bill-Fischofer-Linaro: This could be made generic and polymorphic via a macro rather than having a version of this for each pktio class. See the `_odp_pkt_get()` macro defined in `packet_inline_types.h` as an example.

platform/linux-generic/include/odp_packet_io_internal.h
lumag: Well. The whole idea is to have generic space with all unions hidden in the file implementations. Another approach might be to have `void *` field here and let individual drivers allocate private data from internal pools.

platform/linux-generic/pktio/loop.c
lumag: It would require caller to pass target type to such macro. With my implementation driver just calls `pkt_priv(entry)` and gets access to private data without a need to pass type each time.

platform/linux-generic/include/odp_packet_io_internal.h
Bill-Fischofer-Linaro: Allocation would require indirection, which is undesirable on fast paths. Since you're guarding this with `ODP_STATIC_ASSERT()` in each of the modules it won't cause overruns, so you're OK there. I'm just thinking more about futures when some may not be as careful.

platform/linux-generic/pktio/loop.c
Bill-Fischofer-Linaro: Agreed.

muvarov:
Merged.

PR #608, opened 2018-06-01
[PATCH v1] Plain queue test application

psavol:
Added new test application for plain queues in different modes (including lockfree mode). Also improved lockfree validation test to print out a note when test is skipped (e.g. skipped currently on ARM).


muvarov:
Merged

PR #610, opened 2018-06-04
[PATCH v2] example/l2fw_simple: increase wait time

joseppc:
When using huge pages, the wrapper script that's used to run this test
will send a SIGINT signal too early to the signal, causing it to be
ignored by the application. This in turn causes the application to never
exit its main loop, which depends on handling this signal to set the
variable that terminates the loop.

Increasing the delay between starting the application and sending the
SIGINT signal fixes this issue.

This fixes https://bugs.linaro.org/show_bug.cgi?id=3879

### DIFF
--- a/example/l2fwd_simple/l2fwd_simple_run.sh
+++ b/example/l2fwd_simple/l2fwd_simple_run.sh
@@ -12,7 +12,7 @@ echo "using PCAP_IN = ${PCAP_IN}"
 ./odp_l2fwd_simple${EXEEXT} pcap:in=${PCAP_IN} pcap:out=pcapout.pcap \
 	02:00:00:00:00:01 02:00:00:00:00:02 &
 
-sleep 1
+sleep 2
 kill -s SIGINT $!
 wait $!
 STATUS=$?
@@ -32,7 +32,7 @@ rm -f pcapout.pcap
 ./odp_l2fwd_simple${EXEEXT} null:0 null:1 \
 	02:00:00:00:00:01 02:00:00:00:00:02 &
 
-sleep 1
+sleep 2
 kill -s SIGINT $!
 wait $!
 STATUS=$?


joseppc:
@Bill-Fischofer-Linaro actually, now that you mention it... maybe I should have been more conservative and put 4 to be on the safe side... On the other hand, if 2 seconds is not enough, maybe we have a performance problem and it is good to have a test failing, although we won't know why :thinking:

muvarov:
@joseppc would you like to rewrite this example to have graceful exit? That will be more clear solution than   hard code sleeping time.

joseppc:
@muvarov I think it is good as it is. The example is very simple and straightforward to read, it showcases how easy it is to forward packets from one pktio to another one. For simplicity it stops when ctrl-c is pressed, which is what the test script is simulating, except that we need to make sure it gives "enough" time to the application to finish its job.

muvarov:
@joseppc show about to fix it like that:
https://github.com/muvarov/odp/commit/51ed6354dba1f4a91024b2ef8c583ad4fce833e0

joseppc:
@muvarov test will fail if you only copy one packet, unless the pcap file is made of only one packet

joseppc:
@muvarov and why would hardcoding a packet count be better than putting a timeout?

muvarov:
@joseppc number of packets is already hard coded in original code. Also timeout is also hard coded in odp_recv_tmo(.., 1sec).

joseppc:
@muvarov I only see a requirement to process 10 or more packets in order for the application to return success, which is definitely questionable. The test will fail unless all packets in the original pcap file have been forwarded, so your patch is likely to fail the test suite. The timeout in `odp_pktin_recv_tmo()` has a very different purpose, but it may definitely delay execution.

Let's not forget that this is an example, the sole purpose should be to showcase the ODP API. Any logic around it to make tests pass is, in my opinion, superfluous and should be avoided.

I leave it here, please do as you feel is best, once the cause of the problem was found, I thought this would be a one-liner, I feel I've spent too much time with this already :smile:

muvarov:
@joseppc ok, I will apply thins and create new PR for my changes which people can review.

muvarov:
Merged.

PR #611, opened 2018-06-05
[PATCH v1] Enable lock-free queues on ARMv8

psavol:
GCC built-in atomics does not have lock-free implementation of 16 byte variants. Use ODP internal implementation of 16 byte CAS operations on ARM.

### DIFF
--- a/platform/linux-generic/odp_queue_lf.c
+++ b/platform/linux-generic/odp_queue_lf.c
@@ -24,14 +24,55 @@
 
 typedef unsigned __int128 u128_t;
 
-static inline u128_t atomic_load_u128(u128_t *atomic)
-{
-	return __atomic_load_n(atomic, __ATOMIC_RELAXED);
-}
-
 static inline void atomic_zero_u128(u128_t *atomic)
 {
 	__atomic_store_n(atomic, 0, __ATOMIC_RELAXED);
+}
+
+#if defined(__aarch64__)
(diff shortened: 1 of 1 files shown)


platform/linux-generic/odp_queue_lf.c
```diff
@@ -24,14 +24,55 @@
...
+static inline void atomic_zero_u128(u128_t *atomic)
+{
+	__atomic_store_n(atomic, 0, __ATOMIC_RELAXED);
+}
+
+#if defined(__aarch64__)
+/* ARMv8 has atomic load-acq/store-rel instructions for a pair of
+ * 64bit of data. GCC atomic built-in for 128bits does not utilize these
+ * instructions but uses locks instead. Override GCC built-in for ARMv8.
+ */
```
muvarov: That sounds like not complete gcc support. Which compiler/version did you use? Isn't arch/aarch64/odp_atomic.h  more suitable for this?

psavol: At least GCC 5.4, 7.1 and 8.1 does not support lock-free 16 byte operations.

In it's current form odp_atomic.h cannot be included directly, but through odp_cpu.h:

#ifndef PLATFORM_LINUXGENERIC_ARCH_ARM_ODP_CPU_H
#error This file should not be included directly, please include odp_cpu.h
#endif

muvarov:
Merged.

PR #613, opened 2018-06-06
[PATCH v5] example: l2fwd_simple remove predefined sleep

muvarov:

### DIFF
--- a/example/l2fwd_simple/odp_l2fwd_simple.c
+++ b/example/l2fwd_simple/odp_l2fwd_simple.c
@@ -18,7 +18,7 @@
 #define MAX_WORKERS 1
 
 static int exit_thr;
-static int g_ret;
+static int wait_sec;
 
 struct {
 	odp_pktio_t if0, if1;
@@ -87,7 +87,6 @@ static int run_worker(void *arg ODP_UNUSED)
 {
 	odp_packet_t pkt_tbl[MAX_PKT_BURST];
 	int pkts, sent, tx_drops, i;
-	int total_pkts = 0;
 	uint64_t wait_time = odp_pktin_wait_time(ODP_TIME_SEC_IN_NS);
 
 	if (odp_pktio_start(global.if0)) {
@@ -106,8 +105,13 @@ static int run_worker(void *arg ODP_UNUSED)
 		pkts = odp_pktin_recv_tmo(global.if0in, pkt_tbl, MAX_PKT_BURST,
 					  wait_time);
 
-		if (odp_unlikely(pkts <= 0))
+		if (odp_unlikely(pkts <= 0)) {
+			if (wait_sec > 0)
+				if (!(--wait_sec))
+					break;
 			continue;
+		}
+
 		for (i = 0; i < pkts; i++) {
 			odp_packet_t pkt = pkt_tbl[i];
 			odph_ethhdr_t *eth;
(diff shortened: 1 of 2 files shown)


example/l2fwd_simple/odp_l2fwd_simple.c
```diff
@@ -18,7 +18,8 @@
 #define MAX_WORKERS 1
 
 static int exit_thr;
-static int g_ret;
+static int g_ret = -1;
+static int exit_10pkts;
```
joseppc: the name of this variable is confusing

example/l2fwd_simple/odp_l2fwd_simple.c
```diff
@@ -129,8 +133,8 @@ static int run_worker(void *arg ODP_UNUSED)
 			odp_packet_free_multi(&pkt_tbl[sent], tx_drops);
 	}
 
-	if (total_pkts < 10)
-		g_ret = -1;
+	if (total_pkts >= 10)
```
joseppc: I would take the opportunity to remove this altogether, what are we trying to accomplish here?

example/l2fwd_simple/odp_l2fwd_simple.c
```diff
@@ -160,6 +164,8 @@ int main(int argc, char **argv)
 		       " and source MAC address\n");
 		exit(1);
 	}
+	if (argc == 6 && !strncmp(argv[5], "-t", 2))
```
joseppc: As per previous comments, maybe allow -t to be an option that takes an optional parameter (defaulting to 1). This would mean how many seconds without traffic from pktin to wait before exiting the loop.

example/l2fwd_simple/odp_l2fwd_simple.c
```diff
@@ -106,8 +107,11 @@ static int run_worker(void *arg ODP_UNUSED)
 		pkts = odp_pktin_recv_tmo(global.if0in, pkt_tbl, MAX_PKT_BURST,
 					  wait_time);
 
-		if (odp_unlikely(pkts <= 0))
+		if (odp_unlikely(pkts <= 0)) {
+			if (exit_10pkts)
```
joseppc: `if (exit_10pkts--)`
This could be a count to mean how many seconds of silence in pktin we can accept before exiting the loop (see below)

muvarov: uh, yes,  total packets is bellow. This variable is just break or not break this loop. I will send update.

example/l2fwd_simple/odp_l2fwd_simple.c
muvarov: yep, that also good thing to do.

example/l2fwd_simple/odp_l2fwd_simple.c
muvarov: return 0 return code if some packets were processed and mark test as passed.

joseppc: This is not needed, we should only exit with an error code if some error happened to the application, and it should not be an error that no packets were processed (i.e. we used null pktio as input).

The test has other means to decide whether it passed successfully or not.

muvarov: yep, bash script validates in and out pcap files.

joseppc:
Reviewed-by: Josep Puigdemont <<josep.puigdemont@linaro.org>>

muvarov:
Merged.

PR #615, opened 2018-06-08
[PATCH v1] do not use Travis deprecated image

muvarov:
I updated ubuntu locally and see  bug where tap pktio test fails. Problem with test is that Ubuntu switched default policy for forwarding rule. That leads traffic to not go thought 2 bridged tap devices.

After removing deprecated image tap test passes but I see some conflicts with arm64 toolchain installations with apt-get. I don't know if it's temporary problem with apt servers or not.  Arm64 is commented out for not and this patch is not intended to be merged.

### DIFF
--- a/.travis.yml
+++ b/.travis.yml
@@ -382,14 +382,14 @@ jobs:
                           - ./bootstrap
                           - ./configure --enable-helper-linux $CROSS
                           - make -j $(nproc)
-                - stage: "build only"
-                  env: CROSS_ARCH="arm64"
-                  compiler: gcc
-                  install: true
-                  script:
-                          - ./bootstrap
-                          - ./configure --enable-helper-linux $CROSS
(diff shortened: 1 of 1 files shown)


muvarov:
@lumag take a look at test logs for current builds. arm64 fails. I'm thinking about switching everything inside docker. Might be it will be more stable.

muvarov:
@lumag i.e. apt-get fails

lumag:
It seems that updated image contains updated Apt, which treats 'Index not found' as an error thather than a warning and so does not rebuild package lists. I'll think about replacing sources.list with fixed version, which will point exactly, which arches to download from which source. Also you can blame Ubuntu for silently deleting architectures from the mirror without changing Release file (it might be worth opening an issues against them at Launchpad).

muvarov:
Closing this. Will open new PR with docker.

PR #616, opened 2018-06-11
Use odp_queue_capability_t in various tests

matiaselo:


test/validation/api/queue/queue.c
```diff
@@ -432,10 +431,12 @@ static void queue_test_info(void)
...
 	param.sched.sync = ODP_SCHED_SYNC_ORDERED;
 	param.sched.group = ODP_SCHED_GROUP_ALL;
-	if (capability.max_ordered_locks)
+	if (capability.max_ordered_locks) {
 		param.sched.lock_count = 1;
-	else
+	} else {
+		printf("\n    Ordered locks NOT supported\n");
 		param.sched.lock_count = 0;
+	}
```
Bill-Fischofer-Linaro: Wouldn't it be simpler to say `param.sched.lock_count = capability.max_ordered_locks`?

test/validation/api/queue/queue.c
```diff
@@ -461,7 +462,7 @@ static void queue_test_info(void)
 	CU_ASSERT(info.param.sched.sync == odp_queue_sched_type(q_order));
 	CU_ASSERT(info.param.sched.group == odp_queue_sched_group(q_order));
 	ret = odp_queue_lock_count(q_order);
-	CU_ASSERT(ret > 0);
+	CU_ASSERT(ret == param.sched.lock_count);
```
Bill-Fischofer-Linaro: Won't this `CU_ASSERT()` fail if `odp_queue_lock_count()  > 1` since line 435 sets `param.sched.lock_count = 1`?

matiaselo: True, will fix.

test/validation/api/queue/queue.c
matiaselo: Will fix.

test/validation/api/queue/queue.c
matiaselo: On second thought, this should work since odp_queue_lock_count() should return the lock_count used in odp_queue_create() parameters. So, a case you described should actually fail.

Bill-Fischofer-Linaro: Agreed. Test is good as currently stated, however there's considerable overlap with the similar `CU_ASSERT()` on line 466. Perhaps just combine them? E.g., 
```
CU_ASSERT(odp_queue_lock_count(q_order) == info.param.sched.lock_count);
```

matiaselo: This would decrease code coverage a bit since both `odp_queue_info()` and `odp_queue_lock_count()` could potentially return invalid values. `param.sched.lock_count`is set by the test, so it's always safe to compare to.

Bill-Fischofer-Linaro: Good point.

PR #618, opened 2018-06-12
Implement generic crc hash 

psavol:
Implemented 32, 24 and 16 bit CRCs. Everything works except 24 bit CRCs with reflection. It's marked unsupported for now. It seems that that configuration is not very popular, so implementation can be added/fixed later on.

### DIFF
--- a/platform/linux-generic/odp_hash_crc_gen.c
+++ b/platform/linux-generic/odp_hash_crc_gen.c
@@ -1,0 +1,222 @@
+/* Copyright (c) 2018, Linaro Limited
+ * All rights reserved.
+ *
+ * SPDX-License-Identifier:     BSD-3-Clause
+ */
+
+#include <stdio.h>
+#include <stdint.h>
+#include <inttypes.h>
+#include <string.h>
+
+#include <odp/api/hash.h>
+#include <odp/api/hints.h>
+#include <odp/api/rwlock.h>
+
+#include <odp_debug_internal.h>
+#include <odp_init_internal.h>
+
+typedef struct crc_table_t {
+	uint32_t crc[256];
+	uint32_t width;
+	uint32_t poly;
+	int      reflect;
+	odp_rwlock_t rwlock;
+
+} crc_table_t;
+
+static crc_table_t crc_table;
+
+int _odp_hash_init_global(void)
+{
+	memset(&crc_table, 0, sizeof(crc_table_t));
+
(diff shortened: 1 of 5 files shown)


platform/linux-generic/odp_hash_crc_gen.c
```diff
@@ -0,0 +1,222 @@
...
+
+	odp_rwlock_read_lock(&crc_table.rwlock);
+
+	update_table = (crc_table.width != width) ||
+		       (crc_table.poly != poly) ||
+		       (crc_table.reflect != reflect);
+
+	/* Generate CRC table if not yet generated. */
+	if (odp_unlikely(update_table)) {
+		odp_rwlock_read_unlock(&crc_table.rwlock);
```
lumag: Here somebody can take this lock. This has to be replaced with new API like `odp_rwlock_promote`, which promotes taken read lock to write lock.

platform/linux-generic/odp_hash_crc_gen.c
```diff
@@ -0,0 +1,222 @@
...
+		odp_rwlock_write_lock(&crc_table.rwlock);
+
+		crc_table_gen(poly, reflect, width);
+	}
```
lumag: This adds internal state to the crc calculation.
I think it might be better to:
 - pre-generate several typically used tables
 - use thread local storage for non-typical tables.

platform/linux-generic/odp_hash_crc_gen.c
psavol: It's OK if a new reader comes to use the old table. Write lock waits until all readers exit.

platform/linux-generic/odp_hash_crc_gen.c
psavol: There's a large number of combinations, but a single application uses only one or few. I think it's better to "harden" those ones that are commonly used with API calls (like crc32 and crc32c currently), that enables better optimization for those. This generic function is there for odd cases, and apparently has not been needed frequently as it was not implemented for a long time.

Couple of parallel tables could be added to ease the pressure on the single table. But I'd do that on top of this set (this patch set from 0 to 1 table, later from 1 to N tables).

Bill-Fischofer-Linaro:
@lumag is most qualified to add a review for this PR. Dmitry, have your comments been addressed?

lumag:
Yes, all my comments were addressed.

Reviewed-By: Dmitry Eremin-Solenikov

PR #619, opened 2018-06-13
[PATCH v1] Table based crc32 implementation

psavol:
Remove dependency to zlib and use own table based implementation
of crc32. The table was generated with odp_hash_crc_gen64()
code (printed the table it generates internally).

This is about 2x faster than zlib (with 200 bytes of data).


lumag:

muvarov:
Merged.

PR #620, opened 2018-06-13
delete non bsd m4 macros

muvarov:
Most of marcro are copy from autoconf-archive and are enough stable to be used from distro.


m4/ax_compiler_version.m4
```diff
@@ -1,492 +0,0 @@
...
-#   An empty string is returned otherwise.
-#
-# LICENSE
-#
-#   Copyright (c) 2014 Bastien ROUCARIES <roucaries.bastien+autoconf@gmail.com>
-#
-#   Copying and distribution of this file, with or without modification, are
-#   permitted in any medium without royalty provided the copyright notice
-#   and this notice are preserved. This file is offered as-is, without any
-#   warranty.
```
lumag: This is even more relaxed license compared to BSD. No need to remove it.

lumag:
I see little benefit in removing GPL'ed macros, which do not find their way into produced binaries.

m4/ax_compiler_version.m4
muvarov: @lumag yes,  agree about that file.

muvarov:
@lumag benefit for us is that automatic licence scanners will not capture these files and people don't need spend time on reviewing and understanding what is there.

PR #622, opened 2018-06-15
[PATCH v3] timer_accuracy: convert full nsec time to ticks

psavol:
Calculate timeout values in nsec instead of ticks to avoid
(accumulating) rounding error caused by nsec to tick conversion.

Fixes bug https://bugs.linaro.org/show_bug.cgi?id=3787,
timeouts are received consistently with different resolution
values.

### DIFF
--- a/example/timer/odp_timer_accuracy.c
+++ b/example/timer/odp_timer_accuracy.c
@@ -111,8 +111,8 @@ static int start_timers(test_global_t *test_global)
 	odp_timer_t timer;
 	odp_queue_t queue;
 	odp_queue_param_t queue_param;
-	uint64_t tick, first_tick, period_tick, offset_tick;
-	uint64_t period_ns, res_ns, first_ns, res_capa;
+	uint64_t tick, start_tick;
+	uint64_t period_ns, res_ns, start_ns, time_ns, res_capa;
 	odp_event_t event;
 	odp_timeout_t timeout;
 	odp_timer_set_t ret;
@@ -204,9 +204,6 @@ static int start_timers(test_global_t *test_global)
 	odp_timer_pool_start();
 
 	test_global->timer_pool = timer_pool;
-	first_tick  = 0;
-	offset_tick = odp_timer_ns_to_tick(timer_pool, START_OFFSET_NS);
-	period_tick = odp_timer_ns_to_tick(timer_pool, period_ns);
 
 	for (i = 0; i < num; i++) {
 		timer = odp_timer_alloc(timer_pool, queue, NULL);
@@ -218,6 +215,8 @@ static int start_timers(test_global_t *test_global)
 
 		test_global->timer[i] = timer;
 	}
+
+	start_tick = 0;
 
 	for (i = 0; i < num; i++) {
 		timer = test_global->timer[i];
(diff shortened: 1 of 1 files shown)


example/timer/odp_timer_accuracy.c
```diff
@@ -231,13 +230,14 @@ static int start_timers(test_global_t *test_global)
...
+			start_ns   = odp_time_to_ns(time);
+			test_global->first_ns = start_ns + START_OFFSET_NS;
 		}
 
-		tick = first_tick + offset_tick + (i * period_tick);
+		time_ns = test_global->first_ns + (i * period_ns);
+		tick = start_tick + odp_timer_ns_to_tick(timer_pool, time_ns);
 		ret = odp_timer_set_abs(timer, tick, &event);
 
 		if (ret != ODP_TIMER_SUCCESS) {
```
muvarov: while reviewing this I see that odp_timer_free() is missing. Not related to this issue but this function has returns with allocated timer and timeouts.

example/timer/odp_timer_accuracy.c
```diff
@@ -219,6 +216,8 @@ static int start_timers(test_global_t *test_global)
 		test_global->timer[i] = timer;
 	}
 
+	start_tick = 0;
```
muvarov: not needed.

psavol: needed by some compiler versions (GCC for ARM) which think that it may be used uninitialized otherwise.

example/timer/odp_timer_accuracy.c
psavol: All destroys are handled in destroy_timers() function

muvarov:
@psavol I think this example misses fundamental thing that worker has to be pinned to some core. So that odp local time and ticks will have correlation. Without pinning to core odp_time_local depending on implementation might return time of current cpu which executes this code.

example/timer/odp_timer_accuracy.c
muvarov: ok if compiler warns.

Bill-Fischofer-Linaro: GCC has an unfortunate habit of not being too good at seeing whether variables are really uninitialized. The Linux kernel typically flags these superfluous initializations with a comment so that some future maintenance doesn't delete them without further investigation. Might be a good practice to adopt here.

psavol: Currently, our build fails on Travis if this line is removed (that's why I needed to add it). So, if someone removes it later Travis will tell if it's OK, or not. It's OK to remove it when upgraded compilers do not need it anymore.

psavol:
By the spec, odp_local_time() is local to the calling *thread*, not local to cpu. It would be against the spec to share local time stamp between threads, but this is a single thread application not sharing time stamp with other threads. Sure, global time could be used also. In practice, those are usually the same (as it's with linux-generic).

muvarov:
@psavol agree that specs says only about thread. I thought that odp_time_local() is light weight and fast call to get local time. Platform can use different tricks to make it happen, for example recount from cpu cycles. If single thread application run with cpu mask allowing execution on different cpus that logic will not work. I.e. if single thread can jump between several cpus you can not use any per cpu counters directly.

psavol:
For example, under Linux x86 TSC registers are in sync. It's a CPU local register, but running in global sync. TSC used for both local and global time on odp-linux on x86.

It's implementation problem to sync local time per thread (and not only per CPU), or then implementation documentation problem to state that ODP cannot run without threads pinned to cpu. Application is written against spec, and spec says it's local time per thread.

psavol:
v3: suggested-by Josep

muvarov:
Merged.

Issue #563, opened 2018-04-16
fork() in fdserver is problematic


JannePeltonen:
platform/linux-generic/odp_fdserver.c forks a child process to manage shared memory. This is done to make shared memory work even if ODP threads are actually OS processes. The fork happens inside odp_init_global() and is not followed by exec.

Doing fork in a multithreaded program is somewhat problematic (see fork(2) and http://www.linuxprogrammingblog.com/threads-and-fork-think-twice-before-using-them). In particular, the child process should call only async-signal-safe functions until exec. The fdserver implementation is not adhering to this rule and can (at least theoretically) e.g. get stuck waiting indefinitely a mutex that was held by another thread at fork time and will never be freed in the child.

One possible fix would be to add such a restriction to ODP API that the calling process must be single threaded when it calls odp_init_global(). Another would be to have fdserver exec the actual server after the fork.

Bill-Fischofer-Linaro:
Thanks for capturing this issue. It sounds like `exec` would be the preferred (safest) way to go, but would that require that `fdserver` be built as a separate module? That would clearly have implications for ODP builds/distribution that would need to be worked out.

JannePeltonen:
Yes, the exec approach would seem to require a separate fdserver executable somewhere (since ODP is a library, it cannot e.g. execve("/proc/self/exe", ...)).

Yet another possibility might be to not fork but spawn a thread when ODP threads are OS threads and not OS processes (but how would ODP know?), although in that case I suppose such a separate thread would not even be needed.

JannePeltonen:
Yes, it is still a (potential) issue. It is not related to the signal handling issue but about how at fork() all the other threads magically disappear (from child process point-of-view), regardless of their state.

For example, if another thread (a worker thread or some application thread) is executing a critical section and holds a lock inside some library (like C-library), the lock will never be freed. If the child process attempts to use the library and acquire the lock, it will essentially deadlock.

I have not observed this problem in practice, but just by looking at the code.

joseppc:
I took the fdserver code and made it an executable so it could be `execve()` in odp. It is in its [own repo](https://github.com/joseppc/fdserver) but it could be plugged back into ODP without much effort...

Bill-Fischofer-Linaro:
@JannePeltonen Should this issue be closed for now or is it worthwhile to keep it open?

JannePeltonen:
@Bill-Fischofer-Linaro I leave it to others whether this issue should be kept open or not.

[...]

PR #628, opened 2018-06-21
[PATCH v1] [RFC] fdserver executable

joseppc:
An attempt to solve issue #563 by moving fdserver outside of the ODP library and into an external application.
The `fdserver_*` API has changed a bit and is not yet documented (it wasn't before anyway). For instance, the "contexts" used were previously hardcoded in an enumeration, now they can be created at run-time.

This is just an RFC, tests won't pass because they are actually unable to find the fdserver binary (could be solved by providing a run-time configuration file).


Bill-Fischofer-Linaro:
@JannePeltonen Please review to see if this addresses Issue #563 for you.

JannePeltonen:
@Bill-Fischofer-Linaro I cannot review the patch since it is quite large and I am not familiar with the fdserver and ishm implementations.

This seems like a reimplementation of fdserver and not just something that changes the current implementation to a separate executable. I wonder if both changes are good to do in single patch.

PR #629, opened 2018-06-21
[PATCH v1] Fix SPSC and lock-free queue issues

psavol:
Fix issues in SPSC and LF queues. Add validation test cases which found the issues.


muvarov:
Merged

PR #630, opened 2018-06-21
[PATCH v10] Support building ODP without libcrypto from OpenSSL

### LINKED ISSUE #626: ODP should be able to be built without libcrypto from OpenSSL
Currently, my understanding is that ODP always requires libcrypto from OpenSSL. At least in Tiger Moth LTS, I didn't see any option to turn off linking with libcrypto.

I see this as problematic for many reasons:

* Not all ODP applications require cryptography. Some may want to use ODP only for packet I/O without cryptography.
* OpenSSL has a huge static link size. Dynamic linking, on the other hand, easily leads to version incompatibilities.
* The security history of OpenSSL is very terrible. There are better alternatives such as BearSSL, but dependency on BearSSL would not be much better, since OpenSSL is often installed and BearSSL is rarely installed as default.
* Cryptography often has export/import restrictions. If shipping a software using ODP linked with libcrypto across country borders, it will be exported from one country and imported to another country.
* The license of OpenSSL is extremely problematic, unlike the license of ODP. (But, as a counterexample, the license of libconfig is even more problematic.)

Therefore, it should be considered if it's feasible to allow building ODP without linking to libcrypto. If this is done, there should be a `./configure` option that allows turning off the linking even if OpenSSL is installed (like it usually is).

lumag:
Closes #626.


Bill-Fischofer-Linaro:
@lumag Thanks. Will review in more detail later, but a quick comment this should also include an update to the DEPENDENCIES file to indicate that OpenSSL is now optional depending on configuration.

lumag:
@Bill-Fischofer-Linaro well, it is not really an _optional_ dependency. I'd rather say that there is a use-at-your-own-risk option not to use it.

Bill-Fischofer-Linaro:
@lumag If you build without crypto support it should be as if you called `odp_init_global()` indicating the crypto feature is not used. in fact we should probably update `odp_init_param_init()` to reflect that for such builds.

lumag:
@Bill-Fischofer-Linaro true, but I don't think we should advertise that. I'll probably add a notice that it is possible to build ODP w/o OpenSSL, but I would not advertise it in the first place.

Bill-Fischofer-Linaro:
It's not a feature of ODP, but it sounds like it can be a feature of various ODP implementations, including odp-linux. Presumably odp-dpdk would not have this option since we don't control the DPDK configuration?

Bill-Fischofer-Linaro:
@jmtilli Please review this PR to see if it addresses your concerns expressed in Issue #626.

Bill-Fischofer-Linaro:
Checkpatch also has issues with this PR:
```
0001-Support-building-ODP-without-libcrypto-from-OpenSSL.patch
--------------------------------------------------------------
WARNING: Prefer 'unsigned int' to bare use of 'unsigned'
#150: FILE: platform/linux-generic/odp_crypto_null.c:66:
+	unsigned idx;
WARNING: Prefer 'unsigned int' to bare use of 'unsigned'
#172: FILE: platform/linux-generic/odp_crypto_null.c:88:
+	unsigned i;
CHECK: No space is necessary after a cast
#689: FILE: platform/linux-generic/odp_random_null.c:22:
+	(void) buf;
CHECK: No space is necessary after a cast
#690: FILE: platform/linux-generic/odp_random_null.c:23:
+	(void) len;
CHECK: No space is necessary after a cast
#691: FILE: platform/linux-generic/odp_random_null.c:24:
+	(void) kind;
total: 0 errors, 2 warnings, 3 checks, 665 lines checked
0001-Support-building-ODP-without-libcrypto-from-OpenSSL.patch has style problems, please review.
```

jmtilli:
I believe this PR, once fixed to have the solution to the IPsec problem and the minor checkpatch issues, addresses all of my concerns about cryptography and OpenSSL. Thanks for the rapid implementation!

lumag:
@Bill-Fischofer-Linaro I've updated `DEPENDENCIES`, fixed checkpatch issues and fixed examples compilation.

DEPENDENCIES
```diff
@@ -21,6 +21,11 @@ Prerequisites for building the OpenDataPlane (ODP) API
 
    Libraries currently required to link: openssl, libatomic, libconfig
 
+   It is possible to build ODP without OpenSSL by passing flag
+   --without-openssl to configure script. However this will result in
+   ODP-generic not supporting any cryptography algorithm (except NULL) and all
+   random API will return errors. Use at your own risk!
```
muvarov: ODP_RANDOM_BASIC still will be supported. Right?

lumag: Yes, because we have no way to specify that BASIC is not supported. I'd not dare using it, it depend on the time() during random call, so it is wery basic and wery unsecure.

lumag:
Only TM test failure.

DEPENDENCIES
jmtilli: Might it be possible to use /dev/urandom as the randomness provider? I think the Linux implementation if fact is a cryptographically secure PRNG. This will move e.g. export restrictions of cryptograhy to be the Linux distribution's problem instead of a problem of the application linked to ODP.

lumag: @jmtilli then it might be easier to use `getrandom()`.

lumag: @muvarov I've updated the file
@jmtilli I would not want to use /dev/urandom at this point. First it might be not available (or might have not enough entropy to generate secure random). Using `getrandom()` syscall might bring additional kernel version dependencies. Let's get this in first and then consider using `getrandom` (maybe even as a separate random implementation) in addition to OpenSSL and null ones.

lumag:
@muvarov TM test fails with the following errors. Could you please check, since you've touched that test:

```
Suite: traffic_mngr tests
  Test: traffic_mngr_test_capabilities ...passed
  Test: traffic_mngr_test_tm_create ...passed
  Test: traffic_mngr_test_shaper_profile ...passed
  Test: traffic_mngr_test_sched_profile ...passed
  Test: traffic_mngr_test_threshold_profile ...passed
  Test: traffic_mngr_test_wred_profile ...passed
  Test: traffic_mngr_test_shaper ...passed
  Test: traffic_mngr_test_scheduler ...passed
  Test: traffic_mngr_test_thresholds ...passed
  Test: traffic_mngr_test_byte_wred ...FAILED
    1. traffic_mngr.c:3139  - (wred_pkt_cnts->min_cnt <= pkts_sent) && (pkts_sent <= wred_pkt_cnts->max_cnt)
    2. traffic_mngr.c:3139  - (wred_pkt_cnts->min_cnt <= pkts_sent) && (pkts_sent <= wred_pkt_cnts->max_cnt)
    3. traffic_mngr.c:3139  - (wred_pkt_cnts->min_cnt <= pkts_sent) && (pkts_sent <= wred_pkt_cnts->max_cnt)
    4. traffic_mngr.c:3139  - (wred_pkt_cnts->min_cnt <= pkts_sent) && (pkts_sent <= wred_pkt_cnts->max_cnt)
  Test: traffic_mngr_test_pkt_wred ...traffic_mngr.c:3228:test_pkt_wred():min_cnt 54 <= pkts_sent 0 <= max_cnt 84
traffic_mngr.c:3228:test_pkt_wred():min_cnt 54 <= pkts_sent 0 <= max_cnt 84
traffic_mngr.c:3228:test_pkt_wred():min_cnt 34 <= pkts_sent 0 <= max_cnt 66
traffic_mngr.c:3228:test_pkt_wred():min_cnt 16 <= pkts_sent 0 <= max_cnt 46
passed
  Test: traffic_mngr_test_query ...passed
  Test: traffic_mngr_test_marking ...traffic_mngr.c:3982:traffic_mngr_test_marking():
test_vlan_marking was not run because tm_capabilities indicates no vlan marking support
passed
  Test: traffic_mngr_test_fanin_info ...passed
  Test: traffic_mngr_test_destroy ...passedodp_packet_io.c:574:odp_pktio_stop():interface: loop
odp_packet_io.c:411:odp_pktio_close():interface: loop
```

jmtilli:
WRED sounds like a variant of random early detection (RED), an active queue management (AQM) algorithm, and the keyword "random" here might explain what is going on...

muvarov:
@lumag  that debug now says that 0 packets were transmitted and expected from 54 to 84: 
min_cnt 54 <= pkts_sent 0 <= max_cnt 84

bogdanPricope:
Applying: Support building ODP without libcrypto from OpenSSL
.git/rebase-apply/patch:710: new blank line at EOF.
+
warning: 1 line adds whitespace errors.

bogdanPricope:
@@ -182,7 +182,7 @@ static void tm_init_random_data(tm_random_data_t *tm_random_data)
        byte_cnt = 0;
        while (byte_cnt < 256)
                byte_cnt += odp_random_data(&tm_random_data->buf[byte_cnt],
-                                           256 - byte_cnt, 1);
+                                               256 - byte_cnt, odp_random_max_kind());

This is why is not a good idea to:
- use constants
- mix unsigned and signed operations (uint32_t and int32_t) (especially when (-1) can be returned).

However, this is a partial fix:
traffic_mngr.c:3229:test_pkt_wred():min_cnt 34 <= pkts_sent 26 <= max_cnt 66

muvarov:
@bogdanPricope  thanks. For second issue odp_cunit_ret()  might be used. Will you send PR for both changes or you want me to do that?

bogdanPricope:
@muvarov Sorry, I am in the middle of something. Anyway, it is  better to fix it in one step.

muvarov:
@lumag can you please add Bogdans proposed changed to this PR?

lumag:
@muvarov @bogdanPricope I'm still fighting with the `traffic_mngr.c:3229:test_pkt_wred():min_cnt 34 <= pkts_sent 26 <= max_cnt 66` error

platform/linux-generic/odp_traffic_mngr.c
```diff
@@ -182,7 +182,7 @@ static void tm_init_random_data(tm_random_data_t *tm_random_data)
 	byte_cnt = 0;
 	while (byte_cnt < 256)
 		byte_cnt += odp_random_data(&tm_random_data->buf[byte_cnt],
-					    256 - byte_cnt, 1);
+					    256 - byte_cnt, ODP_RANDOM_BASIC);
```
Bill-Fischofer-Linaro: Why not `odp_random_max_kind()`?

Bill-Fischofer-Linaro:
@lumag Please rebase onto current master as @muvarov has merged TM test fixes. Hopefully that will resolve Travis issues.

lumag:
With this PR I'm observing following side effects:
 - TM tests generate warnings without OpenSSL, but no warnings in normal build
 - Tap PktIO test hangs out without OpenSSL, but works in normal build.

jmtilli:
I can see call to odp_random_data in tap pktio test, and also WRED is fundamentally based on randomness, needing random data for its operation. Might this be related to a poor or missing random number generator or deterministic seed?

bogdanPricope:
Seems that random_null odp_random_data() function (with seed based on time) is not producing random enough data: creates tap interfaces with the same MAC address. It seems we are 'too fast&furious' for this 'random' .... we need better seed and/or algorithm.

lumag:
I've changed random to use per-thread state, initialized once.

lumag:
It looks like this version will finally pass all the tests in Travis.

platform/linux-generic/odp_traffic_mngr.c
muvarov: I think this patch simple change constant to define.

platform/linux-generic/test/pktio_ipc/pktio_ipc1.c
```diff
@@ -200,7 +200,11 @@ static int pktio_run_loop(odp_pool_t pool)
 		}
 
 		/* 3. emulate that pkts packets were received  */
-		odp_random_data((uint8_t *)&pkts, sizeof(pkts), 0);
+		ret = odp_random_data((uint8_t *)&pkts, sizeof(pkts), ODP_RANDOM_BASIC);
+		if (ret != sizeof(pkts)) {
+			LOG_ABORT("random failed");
```
muvarov: I think it's ok if odp_random_data() will write less data then requested in one shot.

platform/linux-generic/test/pktio_ipc/pktio_ipc1.c
```diff
@@ -200,7 +200,11 @@ static int pktio_run_loop(odp_pool_t pool)
...
+			break;
```
muvarov: break after abort() ?

muvarov:
@lumag thead has few comments.

platform/linux-generic/test/pktio_ipc/pktio_ipc1.c
lumag: I do not want to have a loop around `odp_random_data` filling single int value (4 bytes). Feel free to add one, if you'd like to.

platform/linux-generic/test/pktio_ipc/pktio_ipc1.c
lumag: why not?

platform/linux-generic/test/pktio_ipc/pktio_ipc1.c
lumag: Just note that previously there was no check for return value at all.

Bill-Fischofer-Linaro:
Latest PR looks reasonable. Quick note, should we include a shippable test with `--without-openssl` for completeness?

muvarov:
Merged.

PR #632, opened 2018-06-25
[PATCH v1] linux-gen: use common posix extensions header

psavol:
odp_posix_extensions.h should be the only file to define POSIX extension levels


muvarov:
fix to TM is here: https://github.com/Linaro/odp/pull/635

muvarov:
Merged.

PR #634, opened 2018-06-26
[PATCH v4] Arm system info

psavol:
ARM CPU model string returned "unknown". Use /proc/cpuinfo to get aarch64 CPU model. Add also sysinfo  application which was used to test implementation.

### DIFF
--- a/example/Makefile.am
+++ b/example/Makefile.am
@@ -9,6 +9,7 @@ SUBDIRS = classifier \
 	  l3fwd \
 	  packet \
 	  switch \
+	  sysinfo \
 	  time \
 	  timer \
 	  traffic_mgmt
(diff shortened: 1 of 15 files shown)


psavol:
It seems that ThunderX in Shippable does not have max cpu frequency defined in the same file as other ARMs I tested. I'll fix it tomorrow either with correct information or making the validation test to skip when max freq is zero. It has passed before due to hard coded default max freq of 1.4GHz (which is wrong information obviously).

psavol:
v3: use hard coded max freq on aarch64 cpus that does not currently support cpufreq (ThunderX). Added ThunderX model enumerations. Passes all tests (also ThunderX at Shippable).
v4: fixed x86 model string back to original format ("@ XXX GHz" not included in string)

muvarov:
Merged.

PR #636, opened 2018-06-29
[PATCH v1] linux-gen: ring: ensure head and tail load order in dequeue

psavol:
Fixes a ring head/tail index synchronization issue that was found with a new test application with lots of ARM cpus (>64). Missing acquire mem ordering allowed tail index read to happen so much before head index read, that the tail index had a smaller value than the head index (which should not be possible).


Bill-Fischofer-Linaro:
@muvarov This should be in 19.0.2 as it's a bug fix against Tiger Moth.

muvarov:
Merged.

PR #637, opened 2018-07-02
[PATCH v2] Unify example application default/max worker counts

matiaselo:
V2:
- Rebase
- Make sure core count >= 0 (Maxim)
- Simplify odp_generator worker count handling (Bogdan)


bogdanPricope:
Applying: example: use min number of workers by default
Applying: example: use ODP_THREAD_COUNT_MAX as max worker count
error: patch failed: example/ipsec/odp_ipsec.c:43
error: example/ipsec/odp_ipsec.c: patch does not apply
error: patch failed: example/ipsec_api/odp_ipsec.c:43
error: example/ipsec_api/odp_ipsec.c: patch does not apply
Patch failed at 0002 example: use ODP_THREAD_COUNT_MAX as max worker count

example/generator/odp_generator.c
```diff
@@ -1169,12 +1169,10 @@ int main(int argc, char *argv[])
 	/* Print both system and application information */
 	print_info(NO_PATH(argv[0]), &args->appl);
 
-	/* Default to max number of workers, unless user specified number of
-	 * workers or cpumask */
 	num_workers = MAX_WORKERS;
```
bogdanPricope: Set default num_workers to 1.... no need to complicate '-w' or '-c' processing for this.

-       /* Default to max number of workers, unless user specified number of
+       /* Default to 1 worker, unless user specified number of
         * workers or cpumask */
-       num_workers = MAX_WORKERS;
+       num_workers = 1;

bogdanPricope:
Why using minimum core count instead of showcasing max performance?

example/classifier/odp_classifier.c
```diff
@@ -519,9 +519,8 @@ int main(int argc, char *argv[])
 	/* Print both system and application information */
 	print_info(NO_PATH(argv[0]), args);
 
-	/* Default to system CPU count unless user specified */
 	num_workers = MAX_WORKERS;
-	if (args->cpu_count)
+	if (args->cpu_count && args->cpu_count < MAX_WORKERS)
```
muvarov: if (args->cpu_count > 0 &&

example/generator/odp_generator.c
matiaselo: The added complexity is required to handle case `-w 0` -> use all available cores. Setting num_workers to MAX_WORKERS inside parse_args() could perhaps be cleaner.

example/classifier/odp_classifier.c
matiaselo: Will fix.

example/generator/odp_generator.c
bogdanPricope: In my view, I see the setting of number of workers like this:
- one default: MAX_WORKERS (now) or 1 in your case
- two overrides:
         '-w':  override default by specify the number of workers as number
         '-c':  override default by specify the number of workers/thread distribution as core mask
- '-w' has precedence over '-c'

As you said, -w 0 can be converted to 'appl_args->num_workers = MAX_WORKERS'.

Yet, I still believe we should run with maximum performance (worker count) instead of minimum.

matiaselo:
> Why using minimum core count instead of showcasing max performance?

Using all cores on high core count systems causes degraded performance with most of the example applications. This is caused by for example the scheduler or the application design itself not scaling that high. In my opinion it is better to default to a minimum core count, which should work always. A user can add more workers as needed.

bogdanPricope:
@matiaselo Using all cores on a high core count system with odp_generator may tear down someone's network... yet, will understand the odp performance. Else, may say "I have 48/254/whatever cores and cannot fill a 10G interface"

bogdanPricope:
@Bill-Fischofer-Linaro As long as we all agree we should go for a 'safe' (first) run instead of a performance run, I am OK with that. Generally, I agree with @matiaselo argument, yet odp is meant for performance. 200+ cores is not a very common setup...

Bill-Fischofer-Linaro:
@bogdanPricope I think it's easy enough to use the `-c 0` invocation to do a stress test and leave the default as a quick "sanity check" on a single core. We can update the Utilities and Examples section of the User Guide to cover this common convention.

bogdanPricope:
@Bill-Fischofer-Linaro  Fine with me.

muvarov:
Merged.

PR #639, opened 2018-07-03
[PATCH v2] Ring optimization for pool and scheduler

psavol:
Optimize ring performance by removing unnecessary ring fullness check. Pool and scheduler implementation uses the ring. Added new standalone performance tests to demonstrate (pool and scheduler) performance improvement and validate multi-thread functionality. Tested on x86 and ARM up to 127 threads.

Performance improvement vary with thread count and test case. With the new pool and sched tests it improves on all thread counts. E.g. improvement is about 20% with odp_sched_perf on 12 threads on x86

BEFORE:
RESULTS - per thread (Million events per sec):
----------------------------------------------
        1      2      3      4      5      6      7      8      9     10
     42.5   40.6   40.2   41.8   42.4   41.0   40.5   41.8   43.0   40.6 
     40.1   41.6

AFTER:
RESULTS - per thread (Million events per sec):
----------------------------------------------
        1      2      3      4      5      6      7      8      9     10
     51.1   48.0   50.6   49.3   51.1   48.4   51.0   49.7   51.1   48.5 
     50.8   49.4

### DIFF
--- a/platform/linux-generic/include/odp_ring_internal.h
+++ b/platform/linux-generic/include/odp_ring_internal.h
@@ -36,7 +36,6 @@ typedef struct ODP_ALIGNED_CACHE {
 
 	/* Reader head and tail */
 	odp_atomic_u32_t r_head;
-	odp_atomic_u32_t r_tail;
 
 	uint32_t data[0];
 } ring_t;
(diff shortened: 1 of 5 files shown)


muvarov:
Merged.

PR #642, opened 2018-07-04
[PATCH v1] test: performance: add new CPU benchmarking application

matiaselo:
New self-contained test application for CPUs stress testing.

The application launches a number of worker threads, which receive
preallocated test UDP packets from the scheduler and perform the following
per packet processing:

1. Calculate a CRC-32C hash over the packet data and check that it matches
   to the value stored in test header
2. Perform full packet parsing
3. Read a value from dummy lookup table based on packet data
4. Modify packet payload based on the lookup table data and sequence number
5. Update UDP checksum
6. Recalculate CRC-32C hash and store it in test header

Dummy lookup table (default 1M entries) is used to increase memory usage.
The size of the lookup table can be modified with '-l' option.

The application measures and prints the maximum throughput per each worker
thread.

### DIFF
--- a/test/performance/odp_cpu_bench.c
+++ b/test/performance/odp_cpu_bench.c
@@ -1,0 +1,813 @@
+/* Copyright (c) 2018, Linaro Limited
+ * All rights reserved.
+ *
+ * SPDX-License-Identifier:     BSD-3-Clause
+ */
+
+#include "config.h"
+
+#include <odp_api.h>
+#include <odp/helper/odph_api.h>
+
+#include <test_debug.h>
+
+#include <getopt.h>
+#include <inttypes.h>
+#include <signal.h>
+#include <stdlib.h>
+#include <unistd.h>
+
+/* Queues are divided into groups and tests packets are passed only between
+ * queues which belong to the same group. */
+#define MAX_GROUPS       64
+#define QUEUES_PER_GROUP 4
+#define PKTS_PER_QUEUE   256
+
+#define MAX_EVENT_BURST  32
+#define CRC_INIT_VAL     123456789
+#define PASS_PACKETS     10000
+
+/* Default number of entries in the test lookup table */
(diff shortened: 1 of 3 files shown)


test/performance/odp_cpu_bench.c
```diff
@@ -0,0 +1,813 @@
...
+/* Default number of entries in the test lookup table */
+#define DEF_LOOKUP_TBL_SIZE (1024 * 1024)
+
+#define MAX_WORKERS      (ODP_THREAD_COUNT_MAX - 1)
+ODP_STATIC_ASSERT(MAX_WORKERS <= MAX_GROUPS * QUEUES_PER_GROUP,
+		  "Not enough queues for all workers");
+
+/* Get rid of path in filename - only for unix-type paths using '/' */
+#define NO_PATH(file_name) (strrchr((file_name), '/') ? \
+			    strrchr((file_name), '/') + 1 : (file_name))
```
Bill-Fischofer-Linaro: Looks like `file_name` is a reserved name for Checkpatch. Might be best to change?
```
CHECK: Macro argument reuse 'file_name' - possible side-effects?
#109: FILE: test/performance/odp_cpu_bench.c:38:
+#define NO_PATH(file_name) (strrchr((file_name), '/') ? \
+			    strrchr((file_name), '/') + 1 : (file_name))
total: 0 errors, 0 warnings, 1 checks, 834 lines checked
0001-test-performance-add-new-CPU-benchmarking-applicatio.patch has style problems, please review.
```

matiaselo: This warning is not about argument name `file_name` being reserved but for using the same variable multiple times in a macro. The string `file_name` is not modified and we use this same macro in multiple examples, so this warning can be ignored.

bogdanPricope:

muvarov:
Merged.

PR #643, opened 2018-07-05
[PATCH v2] Configurable scheduler burst size

psavol:
Add burst size configuration options for scheduler. This enables tuning application throughput vs real-time responsiveness. Patch set also enables large (larger than burst size) event requests, which improves throughput when application is prepared to receive for large number of events. In this case, event handle copy to stash is avoided, which improves performance. So, an application that calls scheduler with large max_num events can benefit from small burst size config. However, didn't change default burst sizes yet. Patch set has been tested also with burst sizes being 1.

	# Default burst sizes for high and low priority queues. The default
	# and higher priority levels are considered as high. Scheduler
	# rounds up number of requested events up to these values. In general,
	# larger requests are not round down. So, larger bursts than these may
	# received when requested. A large burst size improves throughput,
	# but decreases application responsiveness to high priority events
	# due to head of line blocking cause by a burst of low priority
	# events.
	burst_size_hi  = 32
	burst_size_low = 16

### DIFF
--- a/config/odp-linux-generic.conf
+++ b/config/odp-linux-generic.conf
@@ -55,4 +55,15 @@ sched_basic: {
 	# uneven service level for low thread counts. Typically, optimal
 	# value is the number of threads using the scheduler.
 	prio_spread = 4
+
+	# Default burst sizes for high and low priority queues. The default
+	# and higher priority levels are considered as high. Scheduler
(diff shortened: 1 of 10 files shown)


psavol:
v2: rebased

muvarov:
Merged.

PR #644, opened 2018-07-05
[PATCH v1] linux-gen: odp_pool : Change tab to whitespace

linuxias:
Change the tab to the space for arranging the indentation.


muvarov:
Sorry that is not need. We use tabs a lot.

PR #646, opened 2018-07-09
[PATCH v4] Switch Travis to use current image instead of deprecated one

lumag:


.travis.yml
```diff
@@ -143,6 +143,8 @@ before_install:
           fi
         - export PKG_CONFIG_PATH="$HOME/cunit-install/$CROSS_ARCH/lib/pkgconfig:${PKG_CONFIG_PATH}"
         - find $HOME/cunit-install
+          # workaround for tap driver issues on recent Travis images
+        - sed -e '/tap/d' -i platform/linux-generic/test/Makefile.am
```
muvarov: @lumag take iptables line from here to make tap work https://github.com/Linaro/odp/pull/615/commits/3f0d341faad21b57e3d980f3df5b0b54701c7fbd

muvarov:
@lumag interesting that in your case --distable-static-applications compiles. In my case it fails with clang. Easily reproduced locally.  https://travis-ci.org/muvarov/odp/builds/401795931

muvarov:
ok found. ./configure --disable-static-applications CC=clang LD=clang does not generate dymamic library.  But ./configure --disable-static-applications CC=clang LD=ld compiles odp without issues.

.travis.yml
lumag: @muvarov thanks for the suggestion

lumag:
@muvarov I'm not so sure about using docker. Will it allow using cache/ccache? Also who will be responsible for supporting/updating images? Currently everybody can just pull in additional packages/software/etc. With docker this process is not so transparent.

muvarov:
@lumag docker supports shared directory between host and container. So all caches will work there (I did not ported that feature but it has). Docker container can be build with "docker build <dir>" command where <dir> is directory with Dockerfile. Dockerfile is scenario how to build container. I.e. you need to build it only once then upload. If you need to switch to other base (other Ubuntu version for example) you select that base in Dockerfile and upload new container. The most powerful thing is that with command "docker run <container>  <script>" it downloads container and runs script on your local machine. I.e. tests are not linked to CI system and can be easily run everywhere.

muvarov:
@lumag for x86_64 I put CI script under ./scripts/ci/build_x86_64.sh. If you want you can add additional apt-get install commands there. But I prefer that image is already pre set with all requirements.

For Ubuntu 14.04.5 unfortunately I can not install all cross tools on single machine. Packages depends on each other. Installing powerpc toolchains remove native and arm64. The same problem is that libpcap installation for one arch remove installation for other arch. So CI scripts for arm, arm64 and powerpc are inside docker contatiner, for x86 under ./scripts/ci directory. (it looks like ports for non x86 arches are not ready for new  Ubuntu.).

Plus of course we can use any container what we want (Debian, CentOS).

lumag:
@muvarov ok, let me rephrase the question. How should I behave if I want to have new software piece installed on build system. Currently it is as easy as adding a line to `.travis.yml` file. How will this be handled with Docker?

muvarov:
@lumag looks like old image works so this is not asap but it's good that it's fixed. Patches 3 and 4 can be squashed, I can do it on apply.

About Docker. Yes it adds some complexity but also provides some benefits. What is more convenient for us we have to decide. The idea to use docker is to have pre set environment with everything installed. 3 ways to add new piece to container:

A) Download container, Modify, Push back to sever.
1)  docker pull maximuvarov/odp_ubuntu_14.04.5
2) docker run --privileged -i -t -v `pwd`:/odp -v /lib/modules:/lib/modules -v /usr/src:/usr/src --shm-size 8g -e CC="gcc" -e LD="ld" -e CONF="${CONF}" maximuvarov/odp_ubuntu_14.04.5  /bin/bash
3) (second window) docker ps
4) (second window) docker commit 6027ea569528 maximuvarov/odp_ubuntu_14.04.5
5) (second window) docker tag maximuvarov/odp_ubuntu_14.04.5 maximuvarov/odp_ubuntu_14.04.5
6) (second window) docker push maximuvarov/odp_ubuntu_14.04.5

B) Rebuild Docker container without live modifying it.
1) Add to Dockerfile https://github.com/muvarov/odp/blob/master/scripts/docker/x86_64/Dockerfile
2) docker build .
3) docker tag <outout from docker build> maximuvarov/odp_ubuntu_14.04.5
4) docker push

C) Use apt-get install inside testing script.
Travis shares directory between host and container. Test script is inside odp.git. So just inside testing script you can use apt or what you want.
Example:
docker run --privileged -i -t -v `pwd`:/odp -v /lib/modules:/lib/modules -v /usr/src:/usr/src --shm-size 8g -e CC="gcc" -e LD="ld" -e CONF="${CONF}" maximuvarov/odp_ubuntu_18.04  /odp/scripts/ci/build_x86_64.sh
In build_x86_64.sh you can use apt-get install for example.

Pre compiled things you can put inside container. Like I did for CUnit.

lumag:
@muvarov yes, please squash those two commits.

muvarov:
@lumag I can create opendataplane/  group there and give write permission that group and add there who is interested. maximuvarov/ is temporary name space.

matiaselo:
@lumag Could you please submit this patch also to ODP-DPDK?

muvarov:
Merged.

PR #647, opened 2018-07-10
[PATCH v7] travis: add docker tests

muvarov:
initial Travis docker runs

### DIFF
--- a/.travis.yml
+++ b/.travis.yml
@@ -7,6 +7,12 @@
 # pushing to github/coverity_scan will also launch a static analysis
 # See https://scan.coverity.com/travis_ci
 
+#
+# Travis uses Docker images which mainained here:
+# 	https://github.com/Linaro/odp-docker-images
+# CI scirpts are maintained under ./scripts/ci/ directory
+# which passed into container during the test run.
+
 language: c
 sudo: required
 dist: trusty
@@ -53,7 +59,7 @@ env:
         - CONF=""
         - CONF="--disable-abi-compat"
         - CONF="--enable-deprecated"
-        - CONF="--enable-dpdk-zero-copy"
+        - CONF="--enable-dpdk-zero-copy --disable-static-applications"
         - CONF="--disable-static-applications"
         - CONF="--disable-host-optimization"
         - CONF="--disable-host-optimization --disable-abi-compat"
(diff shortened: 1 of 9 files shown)


muvarov:
v2: rebase on top of current master.

Bill-Fischofer-Linaro:
Needs rebase. Checkpatch still complains about excess line length:
```
0001-travis-add-docker-tests.patch
+perl ./scripts/checkpatch.pl 0001-travis-add-docker-tests.patch
No structs that should be const will be found - file '/home/travis/build/Linaro/odp/scripts/const_structs.checkpatch': No such file or directory
#516: FILE: scripts/ci/build_dpdk.sh:57:
+  sed -i -e 's/!(/!(defined(__arm__) \&\& defined(__clang__) || /g' lib/librte_eal/common/include/arch/arm/rte_byteorder.h
#517: FILE: scripts/ci/build_dpdk.sh:58:
+  sed -i -e 's/__GNUC__/defined(__arm__) \&\& defined(__clang__) || __GNUC__/' lib/librte_eal/common/include/generic/rte_byteorder.h
#540: FILE: scripts/ci/build_dpdk.sh:81:
+  make build O=${TARGET} EXTRA_CFLAGS="-fPIC $DPDK_CFLAGS" CROSS="$DPDK_CROSS" CC="${CC}" HOSTCC="gcc" -j $(nproc)
#561: FILE: scripts/ci/build_x86_64.sh:7:
+export PKG_CONFIG_PATH="$HOME/cunit-install/x86_64/lib/pkgconfig:${PKG_CONFIG_PATH}"
#592: FILE: scripts/ci/build_x86_64.sh:38:
+  find . -name "*.trs" | xargs grep -l '^.test-result. FAIL' | while read trs ; do echo FAILURE detected at $trs; cat ${trs%%.trs}.log ; done
#612: FILE: scripts/ci/coverage.sh:7:
+export PKG_CONFIG_PATH="$HOME/cunit-install/x86_64/lib/pkgconfig:${PKG_CONFIG_PATH}"
#653: FILE: scripts/ci/distcheck.sh:7:
+export PKG_CONFIG_PATH="$HOME/cunit-install/x86_64/lib/pkgconfig:${PKG_CONFIG_PATH}"
total: 0 errors, 7 warnings, 0 checks, 644 lines checked
0001-travis-add-docker-tests.patch has style problems, please review.
```

muvarov:
@lumag  Hi Dmitry, can please review following idea. Project with travis job to build and upload imgas to docker cloud https://github.com/muvarov/odp-docker-images . Travis run is here https://travis-ci.org/muvarov/odp-docker-images/builds/404945899 . If you need to add new software just add it to Dockerfile. If you need to clone full CI just clone this repo and use your own docker namespace.

lumag:
@muvarov yes, this approach looks good to me. Could you please move it to Linaro/ namespace (and maybe add few words of documentation)? Also it might be a good idea to use 16.04, which is more common at this moment.

muvarov:
v3:
- rebase on current master
- move docker images creation under separate project

scripts/ci/distcheck.sh
```diff
@@ -0,0 +1,25 @@
+#!/bin/bash
+set -e
+
+# CC LD AR CXX has to be predifubed
+#
+
+export PKG_CONFIG_PATH="$HOME/cunit-install/x86_64/lib/pkgconfig:${PKG_CONFIG_PATH}"
```
lumag: Ideally this should go to external environment setup. Also if you are building docker images, it might be easier to just install CUnit into system path properly (by installing proper packages).

scripts/ci/distcheck.sh
```diff
@@ -0,0 +1,25 @@
...
+
+CWD=$(dirname "$0")
+TDIR=`mktemp -d -p ~`
+
+cd ${TDIR}
```
lumag: Why do you need a separate temp dir. Are these docker images isolated or not?

scripts/ci/distcheck.sh
```diff
@@ -0,0 +1,25 @@
...
+git clone ${CWD}/../../ odp
+cd ./odp
+./bootstrap
+./configure --enable-user-guides
+
+make clean
```
lumag: Why do you do `make clean` before each build process? It seems like an overkill here.

scripts/ci/distcheck.sh
```diff
@@ -0,0 +1,25 @@
...
+make distcheck
+
+make clean
+make distcheck DISTCHECK__CONFIGURE_FLAGS=--disable-abi-compat
```
lumag: As you are merging two make distcheks here, it might be interesting to do one extra step: let's try doine one cross-build using generated tarball. Thus we can ensure that we have all arch files in the tarball.

scripts/ci/build_x86_64.sh
```diff
@@ -0,0 +1,45 @@
...
+
+git clone ${CWD}/../../ odp
+cd ./odp
+./bootstrap
+./configure ${CONF} \
+	--with-dpdk-path=${DPDKPATH}
+
+make clean
+make -j 8
+# Tell some time sensative ODP test that they can be skipped due to not
```
lumag: sensitive

lumag: Also docker should be pretty much isolated. I'd suggest commenting instead that we ignore possible failures there because these tests depends on measurements and systems might differ in performance.

scripts/ci/build_dpdk.sh
```diff
@@ -0,0 +1,89 @@
+#!/bin/bash -x
+
+set -e
+
+DPDK_VERS="17.11.2"
```
lumag: Ugh. This whole script should go into image building stage.
Also ideally we should be able to install DPDK into system location. Can you check if backported DPDK packages are available? Or maybe we can just backport and build them manually?

.travis.yml
```diff
@@ -364,85 +138,83 @@ jobs:
...
                 - stage: test
-                  compiler: clang-3.8
+                  compiler: clang
                   env: CROSS_ARCH="i386"
+                  script:
+                          - docker run --privileged -i -t
+                              -v `pwd`:/odp -v /lib/modules:/lib/modules -v /usr/src:/usr/src
+                              -e CC="${CC}" -e LD="${LD}" -e CXX="${CXX}"
+                              -e CONF="${CONF}"
+                              maximuvarov/odp_ubuntu_14.04.5_i386 /root/build_${CROSS_ARCH}.sh ;
```
lumag: `maximuvarov/` namespace again. We should not use it.

.travis.yml
```diff
@@ -63,278 +70,45 @@ env:
...
-              pushd netmap/LINUX
-              ./configure
-              make -j $(nproc)
-              popd
-            fi
-            sudo insmod ./netmap/LINUX/netmap.ko
-            EXTRA_CONF="$EXTRA_CONF --with-netmap-path=`pwd`/netmap"
-          fi
-
+        - sudo apt-get install linux-headers-`uname -r`
```
lumag: Again, just install into image.

.travis.yml
```diff
@@ -364,85 +138,83 @@ jobs:
...
-                  env: CROSS_ARCH="i386"
-                  compiler: gcc
-                  install: true
-                  script:
-                          - ./bootstrap
-                          - ./configure --enable-helper-linux $CROSS
-                          - make -j $(nproc)
-                - stage: "build only"
-                  env: CROSS_ARCH="arm64"
+                  env: Ubuntu14.04_arm64
```
lumag: What is this?

scripts/ci/distcheck.sh
muvarov: yes, or that can be in Dockerfile or ubuntu18 already has version which we need. I will re check.

scripts/ci/distcheck.sh
muvarov: I use it do debug build manually. I.e. I can run docker with /bin/bash and than type commands. For me it's more convenient to do it in some directory which I can delete and restart again.

scripts/ci/distcheck.sh
muvarov: yep, that can be good check.

scripts/ci/build_x86_64.sh
muvarov: ok

scripts/ci/build_dpdk.sh
muvarov: apt-get has some dpdk packages. I build dpdk  with kernel headers from host. Docker use hosts kernel. Not sure if we can build dpdk inside Docker and run it on different hosts.

.travis.yml
muvarov: I think this can be deprecated 14.05 ubuntu version. Because it's not supported and we can switch cross compilation to later Ubuntu version. That is why I left this image in current name space. If you still think that it has to be in lngodp/ namespace I will do it (it's one command).

.travis.yml
muvarov: No, kernel inside container is host kernel. You need to build kernel modules for host kernel. Image in container might not be able to install headers for host kernel.

.travis.yml
muvarov: env is set only for better view in results spreadsheet.

scripts/ci/distcheck.sh
muvarov: ok. it's optional.

scripts/ci/build_dpdk.sh
muvarov: dpdk in this version is also 17.11,  looking if odp will work with dpdk from apt..

muvarov:
v4: - @lumag comments were fixed.
I used dpdk installed with apt-get. But in some cases CI shows following error:
```
/root/tmp.BgAkq7yGVb/odp/lib/.libs/libodp-linux.a(dpdk.o): In function `mbuf_pool_create':
dpdk.c:(.text+0x8777): undefined reference to `rte_mempool_ops_alloc'
```
In general odp found dpdk installed with apt-get just with --enable-dpdk. Now I think we should think about level of compatibility of dpdk. Looks like in ubuntu version this was not exported.

.travis.yml
lumag: As you are using provided DPDK, we can stop installing kernel headers.

.travis.yml
```diff
@@ -63,278 +70,45 @@ env:
...
-          fi
-        - make install
-
-        - echo "Checking linking and run from install..."
-        - pushd $HOME
-        - echo "Dynamic link.."
-        - ${CC} ${CFLAGS} ${OLDPWD}/example/hello/odp_hello.c -o odp_hello_inst_dynamic `PKG_CONFIG_PATH=${HOME}/odp-install/lib/pkgconfig:${PKG_CONFIG_PATH} pkg-config --cflags --libs libodp-linux`
-        - if [ -z "$CROSS_ARCH" ] ; then
-            LD_LIBRARY_PATH="${HOME}/odp-install/lib:$LD_LIBRARY_PATH" ./odp_hello_inst_dynamic ;
+        - if [ "${CC#clang}" != "${CC}" ] ; then LD="" CXX=clang++; fi
```
lumag: Any reason for resetting `LD`? Is it set anywhere?

.travis.yml
```diff
@@ -364,85 +138,82 @@ jobs:
...
-                  env: CROSS_ARCH="arm64"
-                  compiler: gcc
-                  install: true
+                  env: Ubuntu14.04_arm64
                   script:
-                          - ./bootstrap
-                          - ./configure --enable-helper-linux $CROSS
-                          - make -j $(nproc)
+                          - docker run  -i -t -v `pwd`:/odp -v /lib/modules:/lib/modules -v /usr/src:/usr/src
+                              lngodp/odp_ubuntu_14.04.5  /root/build_arm64.sh
```
lumag: Can we have this `build_FOO.sh` scripts somewhere in the source tree?

.travis.yml
```diff
@@ -364,85 +138,82 @@ jobs:
...
-                - stage: "build only"
-                  env: CROSS_ARCH="arm64"
-                  compiler: gcc
-                  install: true
+                  env: Ubuntu14.04_arm64
                   script:
-                          - ./bootstrap
-                          - ./configure --enable-helper-linux $CROSS
-                          - make -j $(nproc)
+                          - docker run  -i -t -v `pwd`:/odp -v /lib/modules:/lib/modules -v /usr/src:/usr/src
```
lumag: Also I don't think there is a reason now to pass `/lib/modules` and `/usr/src` inside container.

.travis.yml
```diff
@@ -63,278 +70,45 @@ env:
...
-        - echo "Dynamic link.."
-        - ${CC} ${CFLAGS} ${OLDPWD}/example/hello/odp_hello.c -o odp_hello_inst_dynamic `PKG_CONFIG_PATH=${HOME}/odp-install/lib/pkgconfig:${PKG_CONFIG_PATH} pkg-config --cflags --libs libodp-linux`
-        - if [ -z "$CROSS_ARCH" ] ; then
-            LD_LIBRARY_PATH="${HOME}/odp-install/lib:$LD_LIBRARY_PATH" ./odp_hello_inst_dynamic ;
+        - if [ "${CC#clang}" != "${CC}" ] ; then LD="" CXX=clang++; fi
+        - if [ -n "${CROSS_ARCH}" ] ; then
+               docker run  -i -t -v `pwd`:/odp -v /lib/modules:/lib/modules -v /usr/src:/usr/src
+                 -e CC="${CC}" -e LD="${LD}" -e CXX="${CXX}"
+                 -e CONF="${CONF}" -e DPDK_SHARED="${DPDK_SHARED}"
+                 lngodp/odp_ubuntu_14.04.5  /root/build_${CROSS_ARCH}.sh ;
```
lumag: This image is not built from `odp-docker-images` repo. Please fix that.

.travis.yml
```diff
@@ -364,85 +138,82 @@ jobs:
...
                 - stage: test
-                  compiler: clang-3.8
+                  compiler: clang
                   env: CROSS_ARCH="i386"
+                  script:
+                          - docker run --privileged -i -t
+                              -v `pwd`:/odp -v /lib/modules:/lib/modules -v /usr/src:/usr/src
+                              -e CC="${CC}" -e LD="${LD}" -e CXX="${CXX}"
+                              -e CONF="${CONF}"
+                              lngodp/odp_ubuntu_14.04.5_i386 /root/build_${CROSS_ARCH}.sh ;
```
lumag: This image is not built from `odp-docker-images` repo. Please fix that.

lumag:
Also, according to docs you should have following lines in `.travis.ci`:
```YAML
services:
 - docker
```

.travis.yml
muvarov: netmap modules will require that.

muvarov:
@lumag  ubuntu 14.04.5  is deprecated ubuntu version. I think we have to switch to something new. I specially did not put this image generation to odp-docker-images.git. So plan is to generate cross images for ubuntu 18. But whey I stated to do that ports were not ready. I.e. I think that first step is change to docker but has the same functionality and same ubuntu version and second step will be update images to current. That will be separate PR.

.travis.yml
lumag: No, you will just have to  build netmap modules on host system and libs in guest. Maybe we just have to package netmap.

lumag: Ah, there are no libs. Then it's even simpler. Just build modules on host system and provide headers into guest.

lumag:
Well, you have already switched to 18.04 in the major of tests. I think that we should be able to reproduce/update build environment from the point of merging this PR, not postponing it to later dates (updating to current versions might be postponed, but all images should come from `odp-docker-images`) repo.

BTW: I think that maybe we should stick to 16.04 (which seems to be more common for now) plus PPA for DPDK packages.

.travis.yml
```diff
@@ -82,8 +82,18 @@ script:
...
+               pushd netmap/LINUX;
+               ./configure;
+               make -j $(nproc);
+               popd;
+               sudo insmod ./netmap/LINUX/netmap.ko;
+               CONF="$CONF --with-netmap-path=/odp/netmap";
+               echo "Running test" ;
                docker run --privileged -i -t
-                 -v `pwd`:/odp -v/lib/modules:/lib/modules -v/usr/src:/usr/src --shm-size 8g
+                 -v `pwd`:/odp --shm-size 8g
```
lumag: We can drop additional mounts in the rest of test case invocations now.

lumag:
Squashing commits before merge looks like a good idea.
I will return to the question of cross-tools on Monday.

lumag:
I will try helping with updating images/fixing DPDK/etc later.

muvarov:
Merged.

PR #648, opened 2018-07-12
[PATCH v1] linux-gen: ipsec: fall back to lower odp_random_kind_t if necessary

matiaselo:
Create IV with lower random kind if ODP_RANDOM_CRYPTO is not supported.
This enables passing validation tests using odp_random_null.

### DIFF
--- a/platform/linux-generic/odp_ipsec.c
+++ b/platform/linux-generic/odp_ipsec.c
@@ -13,6 +13,7 @@
 #include <odp/api/byteorder.h>
 #include <odp/api/plat/byteorder_inlines.h>
 
+#include <odp_global_data.h>
 #include <odp_init_internal.h>
 #include <odp_debug_internal.h>
 #include <odp_packet_internal.h>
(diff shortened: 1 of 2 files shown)


platform/linux-generic/odp_ipsec.c
```diff
@@ -1892,6 +1893,10 @@ int _odp_ipsec_init_global(void)
 
 	memset(&default_out_opt, 0, sizeof(default_out_opt));
 
+	odp_global_data.ipsec_rand_kind = ODP_RANDOM_CRYPTO;
+	if (odp_global_data.ipsec_rand_kind > odp_random_max_kind())
+		odp_global_data.ipsec_rand_kind = odp_random_max_kind();
```
Bill-Fischofer-Linaro: If `odp_random_max_kind()` is less than `ODP_RANDOM_CRYPTO`, wouldn't it be better to simply say that IPsec is not available in this implementation? Providing poor crypto quality seems undesirable.

lumag: This is used to generate IV. I would not use BASIC by default, but it is fine to be used if better random source is not available. Remember, before BEAST attack it was quite typical to send 'previous packet' as IV, which isn't even random.

lumag:

platform/linux-generic/odp_ipsec.c
Bill-Fischofer-Linaro: OK, we'll cover this caveat in the documentation.

muvarov:
Merged.

PR #653, opened 2018-07-23
Improve high core count system support

matiaselo:


platform/linux-generic/test/performance/odp_scheduling_run_proc.sh
```diff
@@ -11,7 +11,7 @@
 TEST_DIR="${TEST_DIR:-$(dirname $0)}"
 PERFORMANCE="$TEST_DIR/../../../../test/performance"
 ret=0
-ALL=0
+MAX=32
```
muvarov: Matias, I think that limiting this just hides real problem instead of doing real fix. As I understand the test code fetches events and places it back to queue. With high core you probably see situation that first cores take events and other cores don't.  Might be we need to limit number of event for each core to receive or add synchronization barriers per some number of events.

matiaselo: A valid point. I'll rethink this commit.

PR #657, opened 2018-07-30
[PATCH v2] Fix array out of bounds

linuxias:
Array 'timer_global.timer_pool[32]' accessed at index 32, which is out of
bounds in line 323.

If the i variable in the for statement goes all the way to the end of the condition, the i variable will be 32 and out of the bounds problem will occur on the 323 line.

### DIFF
--- a/platform/linux-generic/odp_timer.c
+++ b/platform/linux-generic/odp_timer.c
@@ -262,7 +262,7 @@ static odp_timer_pool_t timer_pool_new(const char *name,
 		return ODP_TIMER_POOL_INVALID;
 	}
 
-	for (i = 0; i < MAX_TIMER_POOLS; i++) {
+	for (i = 0; i < MAX_TIMER_POOLS - 1; i++) {
 		if (timer_global.timer_pool_used[i] == 0) {
 			timer_global.timer_pool_used[i] = 1;
 			break;


Bill-Fischofer-Linaro:
Hi Seungha,

Please complete a [contributor's agreement](https://www.opendataplane.org/contributor/individual/) so that your patch may be included in ODP and also each commit requires a Signed-off-by: tag:
```
0001-Fix-array-out-of-bounds.patch
+perl ./scripts/checkpatch.pl 0001-Fix-array-out-of-bounds.patch
No structs that should be const will be found - file '/home/travis/build/Linaro/odp/scripts/const_structs.checkpatch': No such file or directory
ERROR: Missing Signed-off-by: line(s)
total: 1 errors, 0 warnings, 0 checks, 8 lines checked
0001-Fix-array-out-of-bounds.patch has style problems, please review.
```

matiaselo:
Are you able to reproduce this condition when running the code? To me it seems this should never happen as the check on line 260: `if (timer_global.num_timer_pools >= MAX_TIMER_POOLS)` should handle this.

muvarov:
@matiaselo yes it should. But anyway it's better to fix array scan.

linuxias:
@matiaselo 
I agree your opinion.  It is not able to reproduce this condition. 
I considered handling exceptions where developers can make mistakes.

So I want to add an error message. What do you think of this modification?

if (timer_global.num_timer_pools >= MAX_TIMER_POOLS) {
        ODP_ABORT("timer pool is full")
	odp_ticketlock_unlock(&timer_global.lock);
	return ODP_TIMER_POOL_INVALID;
}

linuxias:
@muvarov 
Hello, I writed agreements and sent scanned file to legal@linaro.org.
But this mail address seems invalid.
Could you describe how to send agreements to anybody?

Thanks!

muvarov:
@linuxias actually link should just show web form where you type your email and click I agree. But looks like it does not work now. You can send me maxim.uvarov@linaro.org or Bill and I will forward your email. I will also take care about link to be working.

matiaselo:
>  I agree your opinion. It is not able to reproduce this condition.
> I considered handling exceptions where developers can make mistakes.
> 
> So I want to add an error message. What do you think of this modification?
> 
> if (timer_global.num_timer_pools >= MAX_TIMER_POOLS) {
> ODP_ABORT("timer pool is full")
> odp_ticketlock_unlock(&timer_global.lock);
> return ODP_TIMER_POOL_INVALID;
> }

@linuxias This is actually not exactly an error, the implementation just happens to be out of timer pools. So in my opinion ODP_DBG would fit better here. Definitely not ODP_ABORT, since it terminates program execution.

The new check for detecting developer errors would go after the following loop:
```
	for (i = 0; i < MAX_TIMER_POOLS; i++) {
		if (timer_global.timer_pool_used[i] == 0) {
			timer_global.timer_pool_used[i] = 1;
			break;
		}
	}
```

E.g.
```
	if (i == MAX_TIMER_POOLS)
		ODP_ABORT("Unable find free timer pool\n");
```
I would use abort here as there is really no point trying to recover from internal error.

muvarov:
I think ABORT is too much for that. System is out of resources in that case. It should not terminate application. Just return error.

matiaselo:
> I think ABORT is too much for that. System is out of resources in that case. It should not terminate application. Just return error.

@muvarov If you were replying to my previous comment, I used abort in the example because the internal logic would be broken if this ever happens: timer_global.num_timer_pools value is incorrect.

PR #660, opened 2018-08-01
[PATCH API-NEXT v3] api: pktio: extend odp_pktin_queue_param_t to support per queue configuration

bogdanPricope:
Per queue configuration enables advanced usecases where input queues of the same interface may belong to different scheduler groups or have different multithread safe modes.

### DIFF
--- a/include/odp/api/spec/packet_io.h
+++ b/include/odp/api/spec/packet_io.h
@@ -149,6 +149,17 @@ typedef enum odp_pktio_op_mode_t {
 } odp_pktio_op_mode_t;
 
 /**
+ * Packet input queue parameters override
+ */
+typedef struct odp_pktin_queue_param_ovr_t {
+	/** Override for schedule group in odp_schedule_param_t
+	  *
+	  * This parameter is considered only when queue type is
+	  * ODP_QUEUE_TYPE_SCHED. */
+	odp_schedule_group_t group;
+} odp_pktin_queue_param_ovr_t;
+
+/**
  * Packet input queue parameters
  */
 typedef struct odp_pktin_queue_param_t {
@@ -203,6 +214,18 @@ typedef struct odp_pktin_queue_param_t {
 	  * value is ignored. */
 	odp_queue_param_t queue_param;
 
+	/** Queue parameters override
+	  *
+	  * When the override array is defined, the same parameter value
+	  * in 'queue_param' is ignored and these per queue parameter
+	  * values are used instead. Array elements are used in order
+	  * (i.e. the first queue gets parameters from the first array
+	  * element, etc).
+	  * Must point to an array of num_queues elements or NULL to
+	  * disable queue parameters override. The default value is
+	  * NULL.
+	  */
+	odp_pktin_queue_param_ovr_t *queue_param_ovr;
 } odp_pktin_queue_param_t;
 
 /**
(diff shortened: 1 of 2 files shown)


bogdanPricope:
Issue report on limited number of scheduling groups : https://bugs.linaro.org/show_bug.cgi?id=3958

matiaselo:
I'm wondering if this extension is that useful since queue_params would only be used with input flow hashing. A user doesn't know beforehand which packet flows ends up in which input queues, so it doesn't make much sense to use different multithread options, queue types etc. for different input queues.

Using a different scheduling group per input queue could perhaps be useful but this would require that packets are evenly distributed into input queues. Also, one can effectively do this already by using ODP_PKTIN_MODE_QUEUE/ODP_PKTIN_MODE_DIRECT modes.

bogdanPricope:
@matiaselo  Imagine a webserver, where each new connection comes with a different src TCP port. Then , connections can be spread cleanly(in most cases) on different queues/cores.

Indeed, it works fine in "direct" mode already, but scheduler mode may be useful, too.

matiaselo:
> Imagine a webserver, where each new connection comes with a different src TCP port. Then , connections can be spread cleanly(in most cases) on different queues/cores.

Agree, this makes sense.

Bill-Fischofer-Linaro:
@psavol Please review/comment on this.

include/odp/api/spec/packet_io.h
```diff
@@ -201,7 +223,22 @@ typedef struct odp_pktin_queue_param_t {
...
+		 */
+		odp_queue_param_t queue_param;
+
+		/** Per queue parameter set
+		 *
+		 * Coresponds to ODP_PKTIN_QUEUE_PARAM_MODE_MULTI
+		 * queue_param_mode.
+		 * queue_params must point to an array with num_queues
+		 * elements. */
+		odp_queue_param_t *queue_params;
```
psavol: I think this enables too much variability for input queue hashing. Things that are useful to be able set per queue are: schedule group and queue context. It potentially increases implementation complexity if application can request various combinations of other parameters.

include/odp/api/spec/packet_io.h
```diff
@@ -148,6 +148,17 @@ typedef enum odp_pktio_op_mode_t {
...
 } odp_pktio_op_mode_t;
 
+/**
+ * Packet input queue parameters override
+ */
+typedef struct odp_pktin_queue_param_ovr_t {
+	/** Thread group
+	  *
+	  * This parameter is considered only when queue type is
+	  * ODP_QUEUE_TYPE_SCHED. */
```
psavol: It's better to document it as: "Override for schedule group in odp_schedule_param_t". For some reason  odp_schedule_param_t documentation label it thread group, but the type is odp_schedule_group_t so better use that term.

include/odp/api/spec/packet_io.h
```diff
@@ -203,6 +214,13 @@ typedef struct odp_pktin_queue_param_t {
 	  * value is ignored. */
 	odp_queue_param_t queue_param;
 
+	/** Queue parameters override
+	  *
+	  * Overrides queue_param with per queue values. Must point to
+	  * an array of num_queues elements or set to NULL to disable
+	  * queue parameters override. The default value is NULL.
+	  */
+	odp_pktin_queue_param_ovr_t *queue_param_ovr;
```
psavol: A bit more specification is needed how things are overridden: "When the override array is defined, the same parameter value in 'queue_param' is ignored and these per queue parameter values are used instead. Array elements are used in order (i.e. the first queue gets parameters from the first array element, etc)."

include/odp/api/spec/packet_io.h
bogdanPricope: OK

include/odp/api/spec/packet_io.h
bogdanPricope: OK

muvarov:
Merged.

Issue #662, opened 2018-08-01
rte_mempool_ops_alloc() is not dpdk api


muvarov:
rte_mempool_ops_alloc() is used by dpdk --zero-copy but is missing in api file:
./lib/librte_mempool/rte_mempool_version.map
for Ubuntu 18.04  dpdk-17.11.2.
In that case we can not build zero copy odp with recompiled dpdk (apt-get install dpdk).

PR #663, opened 2018-08-02
[PATCH v1] linux-gen: dpdk: fix calling internal dpdk function

matiaselo:
Use pool_alloc() directly instead of using internal DPDK wrapper function.

Fixes: https://github.com/Linaro/odp/issues/662

Reported-by: Maxim Uvarov <maxim.uvarov@linaro.org>


muvarov:
Tested, it works fine. Thanks.

muvarov:
Merged to master and TM.

Issue #659, opened 2018-07-31
Timer API missing capabilities


matiaselo:
ODP timer API doesn't currently provide capabilities for some of the parameters used in odp_timer_pool_param_t.

- Maximum number of timer pools
- Maximum number of timers in a timer pool
- Minimum relative timeout
- Maximum relative timeout

PR #664, opened 2018-08-06
[PATCH API-NEXT v4] api: timer: add timer pool capabilities

matiaselo:
Fixes issue #659

V2:
- Changed `max_pools` from `unsigned int` to `uint32_t` (Barry & Bill)

V3:
- Renamed `common.max_pools` to `max_pools_combined` and improved related documentation

V4:
- Renamed `max_num` to `max_timers` (Petri)

### DIFF
--- a/include/odp/api/spec/timer.h
+++ b/include/odp/api/spec/timer.h
@@ -135,6 +135,22 @@ typedef struct {
  * Timer capability
  */
 typedef struct {
+	/** Maximum number of timer pools over all clock sources
+	 *
+	 * The total number of timer pools that can be created combining
+	 * different clock sources.
+	 */
+	uint32_t max_pools_combined;
+
+	/** Maximum number of timer pools for the requested clock source */
+	uint32_t max_pools;
+
+	/** Maximum number of timers in a pool
+	 *
+	 * The value of zero means that limited only by the available
+	 * memory size for the pool. */
+	uint32_t max_timers;
+
 	/** Highest timer resolution in nanoseconds.
 	 *
 	 *  This defines the highest resolution supported by a timer.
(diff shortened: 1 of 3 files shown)


include/odp/api/spec/timer.h
```diff
@@ -135,6 +135,21 @@ typedef struct {
...
  */
 typedef struct {
+	/** Common for all clock sources */
+	struct {
+		/** Maximum number of timer pools of any clock source */
+		unsigned int max_pools;
+	} common;
+
+	/** Maximum number of timer pools */
+	unsigned int max_pools;
```
Bill-Fischofer-Linaro: Barry pointed out at today's ODP public call that these `unsigned int` fields should be `uint32_t` for consistency with standard usage in ODP APIs.

matiaselo: Makes sense. As a separate change we might want to clean-up couple other structs which use `unsigned` (`odp_pool_capability_t.max_pools`, `odp_pktio_capability_t.max_input_queues`, `odp_queue_capability_t.max_sched_groups` etc.).

include/odp/api/spec/timer.h
```diff
@@ -135,6 +135,21 @@ typedef struct {
...
  */
 typedef struct {
+	/** Common for all clock sources */
+	struct {
+		/** Maximum number of timer pools of any clock source */
+		uint32_t max_pools;
+	} common;
+
+	/** Maximum number of timer pools */
+	uint32_t max_pools;
```
Bill-Fischofer-Linaro: What's the difference between `max_pools` and `common.max_pools` here? This seems confusing.

matiaselo: `max_pools` is for the clock source `odp_timer_capability()` was called with. `common.max_pools` is for all clock sources combined.

Bill-Fischofer-Linaro: Then wouldn't something like `this_clock_max_pools` and `all_clocks_max_pools` be a bit less confusing? If variable needs an explanation to understand it then perhaps a better name is needed.

muvarov:
Travis restarted.

muvarov:
It looks like I wrongly closed this PR without merging. Now it's really merged. I'm sorry about that.

PR #665, opened 2018-08-07
[PATCH API-NEXT v3] api: schedule: add scheduler flow aware mode

bala-manoharan:
ODP scheduler configuration to support flow aware mode


include/odp/api/spec/schedule.h
```diff
@@ -45,6 +46,53 @@ extern "C" {
...
+} odp_schedule_config_t;
+
+typedef struct odp_schedule_capability_t {
+
+	/* Maximum supported flow
+	 * Specifies the maximum number of flows supported by the
+	 * implementation. A value of 1 indicates the schedule does not
+	 * support flow aware mode.
+	 */
+	uint32_t max_flow;
```
muvarov: what is this number for your hardware? should it be big number?

Bill-Fischofer-Linaro:
Doxygen fails, which is preventing mainline Travis tests to be run:
```
$ make doxygen-doc 2>&1 |tee doxygen.log
  DXGEN  doc/application-api-guide/Doxyfile
/home/travis/build/Linaro/odp/include/odp/api/spec/schedule.h:86: warning: Compound odp_schedule_capability_t is not documented.
/home/travis/build/Linaro/odp/include/odp/api/spec/schedule.h:94: warning: Member odp_schedule_capability_t (typedef) of group odp_scheduler is not documented.
/home/travis/build/Linaro/odp/include/odp/api/spec/schedule.h:94: warning: Member odp_schedule_capability_t (typedef) of group odp_scheduler is not documented.
/home/travis/build/Linaro/odp/include/odp/api/spec/schedule.h:93: warning: Member max_flow (variable) of class odp_schedule_capability_t is not documented.
/home/travis/build/Linaro/odp/include/odp/api/spec/schedule.h:60: warning: Member flow_aware (variable) of class odp_schedule_config_t is not documented.
/home/travis/build/Linaro/odp/include/odp/api/spec/schedule.h:76: warning: Member flow_count (variable) of class odp_schedule_config_t is not documented.
/home/travis/build/Linaro/odp/include/odp/api/spec/schedule.h:83: warning: Member queue_count (variable) of class odp_schedule_config_t is not documented.
  DXGEN  doc/helper-guide/Doxyfile
  DXGEN  doc/platform-api-guide/Doxyfile
/home/travis/build/Linaro/odp/include/odp/api/spec/schedule.h:86: warning: Compound odp_schedule_capability_t is not documented.
/home/travis/build/Linaro/odp/include/odp/api/spec/schedule.h:94: warning: Member odp_schedule_capability_t (typedef) of group odp_scheduler is not documented.
/home/travis/build/Linaro/odp/include/odp/api/spec/schedule.h:94: warning: Member odp_schedule_capability_t (typedef) of group odp_scheduler is not documented.
/home/travis/build/Linaro/odp/include/odp/api/spec/schedule.h:93: warning: Member max_flow (variable) of class odp_schedule_capability_t is not documented.
/home/travis/build/Linaro/odp/include/odp/api/spec/schedule.h:60: warning: Member flow_aware (variable) of class odp_schedule_config_t is not documented.
/home/travis/build/Linaro/odp/include/odp/api/spec/schedule.h:76: warning: Member flow_count (variable) of class odp_schedule_config_t is not documented.
/home/travis/build/Linaro/odp/include/odp/api/spec/schedule.h:83: warning: Member queue_count (variable) of class odp_schedule_config_t is not documented.
```

include/odp/api/spec/packet.h
```diff
@@ -1799,15 +1799,23 @@ uint32_t odp_packet_flow_hash(odp_packet_t pkt);
  * Store the packet flow hash for the packet and sets the flow hash flag. This
  * enables (but does not require!) application to reflect packet header
  * changes in the hash.
+ * When scheduler is configured as flow aware, schedule queue synchronization
+ * will be based on flow within each queue.
+ * When scheduler is configured as flow unaware, packet flow hash is ignored by
+ * the implementation.
+ * The value of flow hash should not be greater than the max flow count
+ * supported by the implementation.
```
Bill-Fischofer-Linaro: It would be better to rewrite this description entirely rather than to insert this addendum as the original text refers to behavior in flow unaware mode and then this describes both flow aware and unaware modes again.

This is a good place to provide an overall discussion of what a flow hash represents in the two modes and how ODP will interpret these values.

include/odp/api/spec/schedule.h
```diff
@@ -45,6 +46,53 @@ extern "C" {
...
+
+	/* Scheduler flow aware
+	 * Application can configure the scheduler to be flow aware or unaware.
+	 * When the scheduler is configured as flow aware the ORDERED and
+	 * ATOMIC schedule queue synchronization by the scheduler is performed
+	 * per flow within each queue.
+	 */
+	odp_support_t flow_aware;
+
+	/* Number of parallel flows per queue to be supported
```
Bill-Fischofer-Linaro: Why "parallel" here? Is this a synonym for independent? We already use parallel for a queue type so this may be a confusing reuse of the term.

include/odp/api/spec/schedule.h
```diff
@@ -45,6 +46,53 @@ extern "C" {
...
+	 * Only valid when scheduler is configured as flow_aware
+	 *
+	 * Flows are lightweight entity and packets can be assigned to specific
```
Bill-Fischofer-Linaro: agreement: entities

include/odp/api/spec/schedule.h
```diff
@@ -45,6 +46,53 @@ extern "C" {
...
+	 * flow by the application using odp_packet_flow_hash_set() before
```
Bill-Fischofer-Linaro: agreement: flows

include/odp/api/spec/schedule.h
```diff
@@ -45,6 +46,53 @@ extern "C" {
...
+ */
+typedef struct odp_schedule_config_t {
+
+	/* Scheduler flow aware
+	 * Application can configure the scheduler to be flow aware or unaware.
+	 * When the scheduler is configured as flow aware the ORDERED and
+	 * ATOMIC schedule queue synchronization by the scheduler is performed
+	 * per flow within each queue.
+	 */
+	odp_support_t flow_aware;
```
Bill-Fischofer-Linaro: An `odp_support_t` makes sense in a capability but a configuration sets a specific mode and hence should be a boolean.

include/odp/api/spec/schedule.h
Bill-Fischofer-Linaro: @muvarov since `flow_count` is `uint32_t` it makes sense that `max_flow` would have the same magnitude.

Bill-Fischofer-Linaro: Shouldn't `odp_schedule_capability_t` be where whether flow-aware mode is supported as well as the max flows and queues is returned? Basically this struct should return the options and limits that the application then chooses from in its `odp_schedule_config()` call.

Bill-Fischofer-Linaro: Do we want to move things like number of scheduler groups here? That would seem to make more sense than burying it in the queue config as @bogdanPricope has previously noted.

include/odp/api/spec/schedule_types.h
```diff
@@ -79,6 +79,8 @@ extern "C" {
  * requests another event from the scheduler, which implicitly releases the
  * context. User may allow the scheduler to release the context earlier than
  * that by calling odp_schedule_release_atomic().
+ * When scheduler is enabled as flow-aware they synchronization is mainted per
```
Bill-Fischofer-Linaro: Typos: s/they/then and s/mainted/maintained

include/odp/api/spec/schedule_types.h
```diff
@@ -105,6 +107,8 @@ extern "C" {
  * (e.g. freed or stored) within the context are considered missing from
  * reordering and are skipped at this time (but can be ordered again within
  * another context).
+ * When scheduler is enabled as flow-aware they synchronization is mainted per
```
Bill-Fischofer-Linaro: Same typos here.

include/odp/api/spec/packet.h
Bill-Fischofer-Linaro: The wording here also makes it unclear that (at least in the Cavium implementation) the 32-bit hash value is divided into some number of bits designating a queue with the remaining bits representing the flow within that queue. We certainly need to make this implementation use case explicit. Whether there are other possible implementation-specific interpretations of the hash field is something that still needs to be discussed, preferably with other HW implementations to see if there are other variants the spec really should be addressing.

include/odp/api/spec/schedule.h
matiaselo: `odp_schedule_capability_t.max_flow` should probably be per queue to match `odp_schedule_config_t .flow_count` ? If so, it should be clearly stated.

include/odp/api/spec/schedule.h
matiaselo: Is this option really needed, since setting `odp_schedule_config_t .flow_count` to 1 means practically the same thing?

include/odp/api/spec/schedule.h
```diff
@@ -187,6 +235,42 @@ void odp_schedule_prefetch(int num);
...
+ */
+void odp_schedule_config_init(odp_schedule_config_t *config);
+
+
+/**
+ * Global schedule configuration
+ *
+ * Initialize and configure scheduler with global configuration options.
+ *
+ * @param config   Pointer to scheduler configuration sturcture
```
matiaselo: typo: structure

include/odp/api/spec/schedule.h
```diff
@@ -187,6 +235,42 @@ void odp_schedule_prefetch(int num);
...
+ *
+ * @retval 0 on success
+ * @retval <0 on failure
+ *
+ * @see odp_schedule_capability(), odp_schedule_config_init()
+ */
+int odp_schedule_config(const odp_schedule_config_t *config);
```
matiaselo: It looks like implementing `odp_schedule_config()` requires that we also add `odp_schedule_start()`.

It should be stated when it's OK to call `odp_schedule_config()` and if this is optional. A proper startup sequence should also be defined.
E.g.
```
odp_schedule_config_init()
odp_schedule_config()
odp_schedule_start()
odp_schedule()
```

include/odp/api/spec/packet.h
psavol: Bill, only flow hash part is saved into packet. 32 bit hash value in HW is combined from (bottom) bits of this hash and the queue ID. Here we need to specify just which bits/values of this uint32_t are considered as flow ID. Queue ID is stored as odp_queue_t as before.

include/odp/api/spec/packet.h
```diff
@@ -1799,15 +1799,23 @@ uint32_t odp_packet_flow_hash(odp_packet_t pkt);
...
  *
  * @param      pkt              Packet handle
  * @param      flow_hash        Hash value to set
  *
  * @note If the platform needs to keep the original hash value, it has to
- * maintain it internally. Overwriting the platform provided value doesn't
- * change how the platform handles this packet after it.
+ * maintain it internally.
+ * When scheduler is configured as flow unaware, overwriting the platform
+ * provided value doesn't change how the platform handles this packet after it.
```
psavol: I'd document this the other way (positive vs. negative spec text):
" When scheduler is configured as flow aware, the flow hash value affects scheduling of the event."

psavol: Actually, we'll need this flow ID for all events, not just packets as it affects event synchronization. E.g. application may need to sync timeouts / ctrl plane messages with packets of a flow. So, application would need to set the same flow ID for those events as well.

odp_event_flow_id() / odp_event_flow_id_set() are needed. Maybe only the event level function is needed, as scheduler and queue enq/deq work with events anyway.

psavol: This flow_hash() could keep it's original purpose as input HW calculated 32 bit hash that does not affect scheduling. Only the event level flow ID would affect scheduling. Those could contain the same value for incoming packets, or event flow ID would have only e.g. 20bits but packet flow hash full 32 bits.

include/odp/api/spec/schedule.h
psavol: Agree, single parameter (flow_count) can be used to enable the feature.

include/odp/api/spec/schedule.h
```diff
@@ -45,6 +46,53 @@ extern "C" {
...
+	 * enqueuing the packet into the scheduler.
+	 * This value is valid only when scheduler is configured in flow aware
+	 * mode.
+	 * Depening on the implementation this number might be rounded-off to
+	 * nearest supported value (e.g power of 2)
+	 * This number should be less than maximum flow supported by the
+	 * implementation.
+	 * @see odp_schedule_capability_t
+	 */
+	uint32_t flow_count;
```
psavol: It's not clear if count is always rounded down (could it be rounded up to nearest power of two). When rounded, application needs to know the actual value used, so that it's clear which flow ID values are possible / actually affect scheduling.

Application usually wants to spread flows evenly to flow IDs. For example, it makes a difference if flow ID values 0...6 or 0..7 are actually used within scheduler. If application would use value 7, but scheduler would round down 7 to 0 (without application knowing), there would be double the load on flow ID 0 vs other IDs (1..6). When application knows that 7 is not used, it could (potentially) spread those flows evenly to over all flow IDs (0...6).


It could be actually useful to limit this to powers of two

include/odp/api/spec/schedule.h
```diff
@@ -45,6 +46,53 @@ extern "C" {
...
+
+	/* Maximum number of queues to be supported
+	 * Application configures the maximum number of queues to be supported
+	 * by the implementation. This is a combination of all queue types.
+	 * @see odp_queue_capability_t
+	 */
+	uint32_t queue_count;
```
psavol: " Maximum number of scheduled queues used

 Application will create or request other ODP APIs to create at most this many scheduled queues. See odp_queue_capability_t for scheduled queue capabilities.
"

May be queue size is also needed.
uint32_t queue_count;
uint32_t queue_size;

include/odp/api/spec/schedule.h
psavol: Also when packet input/crypto/ipsec/timer send events to scheduler, the config/start order needs to be defined against those as well.

Maybe simply, odp_schedule_start() must be called before application itself or any ODP module creates any (scheduled)  queues.

bala-manoharan: Yes. We need to add odp_schedule_start() and I believe schedule start should have priority before odp_pktio_start().

include/odp/api/spec/schedule.h
bala-manoharan: queue size is already available in odp_queue_capability_t. We can add it here but it might be redundant.

Bill-Fischofer-Linaro:
Doxygen still has issues with this PR:
```
25.09s$ make doxygen-doc 2>&1 |tee doxygen.log
  DXGEN  doc/application-api-guide/Doxyfile
/home/travis/build/Linaro/odp/include/odp/api/spec/schedule.h:84: warning: Compound odp_schedule_capability_t is not documented.
/home/travis/build/Linaro/odp/include/odp/api/spec/schedule.h:104: warning: Member odp_schedule_capability_t (typedef) of group odp_scheduler is not documented.
/home/travis/build/Linaro/odp/include/odp/api/spec/schedule.h:104: warning: Member odp_schedule_capability_t (typedef) of group odp_scheduler is not documented.
/home/travis/build/Linaro/odp/include/odp/api/spec/schedule.h:120: warning: Found unknown command `\odp_schedule_stop'
/home/travis/build/Linaro/odp/include/odp/api/spec/schedule.h:91: warning: Member max_flow_count (variable) of class odp_schedule_capability_t is not documented.
/home/travis/build/Linaro/odp/include/odp/api/spec/schedule.h:97: warning: Member max_queue_count (variable) of class odp_schedule_capability_t is not documented.
/home/travis/build/Linaro/odp/include/odp/api/spec/schedule.h:103: warning: Member max_queue_size (variable) of class odp_schedule_capability_t is not documented.
/home/travis/build/Linaro/odp/include/odp/api/spec/schedule.h:68: warning: Member flow_count (variable) of class odp_schedule_config_t is not documented.
/home/travis/build/Linaro/odp/include/odp/api/spec/schedule.h:75: warning: Member queue_count (variable) of class odp_schedule_config_t is not documented.
/home/travis/build/Linaro/odp/include/odp/api/spec/schedule.h:81: warning: Member queue_size (variable) of class odp_schedule_config_t is not documented.
  DXGEN  doc/helper-guide/Doxyfile
  DXGEN  doc/platform-api-guide/Doxyfile
/home/travis/build/Linaro/odp/include/odp/api/spec/schedule.h:84: warning: Compound odp_schedule_capability_t is not documented.
/home/travis/build/Linaro/odp/include/odp/api/spec/schedule.h:104: warning: Member odp_schedule_capability_t (typedef) of group odp_scheduler is not documented.
/home/travis/build/Linaro/odp/include/odp/api/spec/schedule.h:104: warning: Member odp_schedule_capability_t (typedef) of group odp_scheduler is not documented.
/home/travis/build/Linaro/odp/include/odp/api/spec/schedule.h:120: warning: Found unknown command `\odp_schedule_stop'
/home/travis/build/Linaro/odp/include/odp/api/spec/schedule.h:91: warning: Member max_flow_count (variable) of class odp_schedule_capability_t is not documented.
/home/travis/build/Linaro/odp/include/odp/api/spec/schedule.h:97: warning: Member max_queue_count (variable) of class odp_schedule_capability_t is not documented.
/home/travis/build/Linaro/odp/include/odp/api/spec/schedule.h:103: warning: Member max_queue_size (variable) of class odp_schedule_capability_t is not documented.
/home/travis/build/Linaro/odp/include/odp/api/spec/schedule.h:68: warning: Member flow_count (variable) of class odp_schedule_config_t is not documented.
/home/travis/build/Linaro/odp/include/odp/api/spec/schedule.h:75: warning: Member queue_count (variable) of class odp_schedule_config_t is not documented.
/home/travis/build/Linaro/odp/include/odp/api/spec/schedule.h:81: warning: Member queue_size (variable) of class odp_schedule_config_t is not documented.
The command "make doxygen-doc 2>&1 |tee doxygen.log" exited with 0.
```

include/odp/api/spec/event.h
```diff
@@ -209,6 +209,45 @@ void odp_event_free_multi(const odp_event_t event[], int num);
...
+ * Store the event flow hash for the event and sets the flow hash flag.
+ * When scheduler is configured as flow aware, schedule queue synchronization
+ * will be based on flow within each queue.
+ * When scheduler is configured as flow unaware, event flow hash is ignored by
+ * the implementation.
+ * The value of flow hash should not be greater than the max flow count
+ * supported by the implementation.
+ *
+ * @param      event		Event handle
+ * @param      flow_hash        Hash value to set
```
Bill-Fischofer-Linaro: Comments seem unaligned here.

include/odp/api/spec/schedule.h
```diff
@@ -45,6 +46,95 @@ extern "C" {
...
+ */
+typedef struct odp_schedule_config_t {
+
+	/* Number of flows per queue to be supported
+	 * Only valid when scheduler is configured as flow_aware mode
+	 *
+	 * Flows are lightweight entities and packets can be assigned to specific
+	 * flows by the application using odp_event_flow_hash_set() before
+	 * enqueuing the packet into the scheduler.
+	 * This value is valid only when scheduler is configured in flow aware
```
Bill-Fischofer-Linaro: Better phrasing might be `This value is ignored unless the schedule is configured in flow aware mode.`

include/odp/api/spec/schedule.h
```diff
@@ -45,6 +46,95 @@ extern "C" {
...
+	 * mode.
+	 * Depening on the implementation this number might be rounded-off to
+	 * nearest supported value (e.g power of 2)
+	 * This number should be less than maximum flow supported by the
+	 * implementation.
```
Bill-Fischofer-Linaro: As we discussed before, if the implementation is going to change the application's input value, then it needs to return to the application the exact number that is being used so the application can properly size its own tables, etc.

include/odp/api/spec/schedule.h
Bill-Fischofer-Linaro: @psavol comment still seems valid here.

include/odp/api/spec/schedule.h
```diff
@@ -45,6 +46,95 @@ extern "C" {
...
+	/* Maximum number of schedule queues to be supported
+	 * Application configures the maximum number of schedule queues to be
+	 * supported by the implementation.
+	 * @see odp_queue_capability_t
+	 */
+	uint32_t queue_count;
+
+	/* Maximum number of events required to be stored simultaneously in
+	 * schedule queue. This number should be less than 'max_queue_size'
+	 * supported by the implementation.
```
Bill-Fischofer-Linaro: Need to clarify case when implementation doesn't impose a fixed queue size (capability returns a queue_size of 0).

include/odp/api/spec/schedule.h
```diff
@@ -45,6 +46,95 @@ extern "C" {
...
+ * odp_schedule_config()
+ * odp_schedule_start()
+ * odp_schedule()
+ *
+ * @retval 0 on success
+ * @retval <0 on failure
+ *
+ * @odp_schedule_stop()
+ */
+int odp_schedule_start();
```
Bill-Fischofer-Linaro: needs to be `int odp_schedule_start(void);` here.

include/odp/api/spec/schedule.h
```diff
@@ -45,6 +46,95 @@ extern "C" {
...
+/**
+ * Stop scheduler operations
+ *
+ * Stop scheduler module. The application should make sure there are no further
+ * events in the scheduler before calling odp_schedule_stop.
+ *
+ * @retval 0 on success
+ * @retval <0 on failure
+ */
+int odp_schedule_stop();
```
Bill-Fischofer-Linaro: `int odp_schedule_stop(void);`

include/odp/api/spec/schedule.h
```diff
@@ -45,6 +46,95 @@ extern "C" {
...
+ * Activate scheduler module to schedule packets across different schedule
+ * queues. The scheduler module should be started before creating any odp
+ * queues. The scheduler module can be stopped usinig odp_schedule_stop().
+ *
+ * The initialization sequeunce should be,
+ * odp_schedule_capability()
+ * odp_schedule_config_init()
+ * odp_schedule_config()
+ * odp_schedule_start()
+ * odp_schedule()
```
Bill-Fischofer-Linaro: As we discussed, need to consider current applications that are happy with the current scheduler behavior. They shouldn't need to be changed to say "I don't need flow-aware scheduling".  We can either add a hint in the params for `odp_init_global()` that the application will use flow-aware mode or else say that the scheduler will operate by default in flow-unaware mode unless `odp_schedule_config()/odp_schedule_start()` is called.

include/odp/api/spec/event.h
bala-manoharan: Done

include/odp/api/spec/schedule.h
bala-manoharan: Updated

include/odp/api/spec/schedule.h
bala-manoharan: updated

include/odp/api/spec/schedule.h
bala-manoharan: The flow_count is always rounded up to the nearest power of 2.
I agree we need a mechanism to indicate the value selected by the implementation.
Will update in the next version

include/odp/api/spec/schedule.h
bala-manoharan: I believe it is easy to update in odp_init_global() saying that the default configuration of scheduler is flow-unware mode unless explicitly configured using odp_schedule_config()

muvarov:
is v3 final version or v4 is expected?

lumag:
Replaced by #734, @muvarov please close.

Bill-Fischofer-Linaro:
Closing this PR, since it's been superseded by PR #734, per @lumag request.

PR #667, opened 2018-08-09
[PATCH v1] abi: unify *_INVALID define values

matiaselo:
Always use zero as invalid value when base data type is defined as a pointer.


muvarov:
Merged.

PR #669, opened 2018-08-09
[PATCH API-NEXT v1] Add packet subtype API call

psavol:
Enables checking packet subtype with packet handle. Useful when event handle is not available any more.


muvarov:
Merged.

PR #672, opened 2018-08-10
[PATCH API-NEXT v3] Packet parse result API

psavol:
Get most commonly used packet parse results with a single function call. May improve performance when multiple metadata fields are read, especially in ABI compat mode.

### DIFF
--- a/include/odp/api/abi-default/packet.h
+++ b/include/odp/api/abi-default/packet.h
@@ -72,6 +72,70 @@ typedef enum {
 
 #define ODP_NUM_PACKET_COLORS 3
 
+/** Parse result flags */
+typedef struct odp_packet_parse_result_flag_t {
+	/** Flags union */
+	union {
+		/** All flags as a 64 bit word */
+		uint64_t all;
+
+		/** Flags as a bitfield struct */
+		struct {
+			/** @see odp_packet_has_error() */
+			uint64_t has_error    : 1;
+			/** @see odp_packet_has_l2_error() */
+			uint64_t has_l2_error : 1;
+			/** @see odp_packet_has_l3_error() */
+			uint64_t has_l3_error : 1;
+			/** @see odp_packet_has_l4_error() */
+			uint64_t has_l4_error : 1;
+			/** @see odp_packet_has_l2() */
+			uint64_t has_l2 : 1;
+			/** @see odp_packet_has_l3() */
+			uint64_t has_l3 : 1;
+			/** @see odp_packet_has_l4() */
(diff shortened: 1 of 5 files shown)


include/odp/api/spec/packet.h
```diff
@@ -1415,6 +1415,126 @@ int odp_packet_parse(odp_packet_t pkt, uint32_t offset,
...
+			/** @see odp_packet_has_tcp() */
+			uint64_t has_tcp : 1;
+			/** @see odp_packet_has_sctp() */
+			uint64_t has_sctp : 1;
+			/** @see odp_packet_has_icmp() */
+			uint64_t has_icmp : 1;
+		};
+	};
+
+} odp_packet_parse_result_flag_t;
```
lumag: I think this should go into ABI files. Platform-optimized ODP versions might have different views on flags layout.

psavol:
v3: moved odp_packet_parse_result_flag_t into ABI files

lumag:
@psavol I have no more issues from my side. @bala-manoharan do you have any?

include/odp/api/abi-default/packet.h
```diff
@@ -72,6 +72,70 @@ typedef enum {
 
 #define ODP_NUM_PACKET_COLORS 3
 
+/** Parse result flags */
+typedef struct odp_packet_parse_result_flag_t {
+	/** Flags union */
+	union {
+		/** All flags as a 64 bit word */
+		uint64_t all;
```
muvarov: 'all' can be used to clear all bit flags. It might be useful to group all errors to u8 or u16 to quick check if there any error.

include/odp/api/spec/packet.h
```diff
@@ -1428,6 +1428,62 @@ int odp_packet_parse(odp_packet_t pkt, uint32_t offset,
...
+ * Read parse results
+ *
+ * Read out the most commonly used packet parse results. The same information is
+ * available through individual function calls, but this call may be more
+ * efficient when reading multiple results from a packet.
+ *
+ * @param      pkt     Packet handle
+ * @param[out] result  Pointer for parse result output
+ */
+void odp_packet_parse_result(odp_packet_t pkt,
```
muvarov: what is expectation if ODP_PACKER_INVALID is passed here?

include/odp/api/abi-default/packet.h
muvarov: however current has_error also works here...

muvarov:
Merged.

PR #673, opened 2018-08-10
[PATCH v1] abi: timer: use strong type for odp_timer_pool_t

matiaselo:


muvarov:
Merged.

PR #675, opened 2018-08-17
[PATCH v1] Separate queue operations for plain and scheduled queues 

psavol:
Queue synchronization needs are different for plain and scheduled queues. Implement separate enq/deq operations for plain and scheduled queues. Once separated these implementations can be further optimized. Separation is done by setting different function pointer to queue entry. Also internal interface needs to use these the same functions, so those were inlined. In this phase, plain queue enqueue performance increased a bit due to simplified function. Scheduled queue implementation was not changed.


psavol:
This is complete clean up / enabler patch set. Following PRs can e.g. change plain / scheduled queues implementations to lock-free.

include/odp/api/spec/queue_types.h
```diff
@@ -0,0 +1,311 @@
...
+ * Non-blocking levels are listed from the weakest to the strongest guarantee of
+ * block freedom. Performance of a non-blocking implementation may be lower than
+ * the blocking one. Non-blocking guarantees are important e.g. for real-time
+ * applications when real-time and non real-time threads share a resource.
+ */
+typedef enum odp_nonblocking_t {
+	/** Blocking implementation. A suspeding thread may block all other
+	 *  threads, i.e. no block freedom guarantees. This is the lowest level.
+	 */
+	ODP_BLOCKING = 0,
```
muvarov: It's better to group all enumbs with the same prefix. I.e. ODP_NONBLOCKING_NONE here.

muvarov: or ODP_BLK_BLOCKED, ODP_BLK_FREE and ODP_BLK_FREE_NO_WAIT

include/odp/api/spec/queue_types.h
```diff
@@ -0,0 +1,311 @@
...
+		/** Wait-free (ODP_NONBLOCKING_WF) implementation capabilities.
+		  * The specification is the same as for the blocking
+		  * implementation. */
+		struct {
+			/** Maximum number of queues. Wait-free queues are not
+			  * supported when zero. */
+			uint32_t max_num;
+
+			/** Maximum queue size */
+			uint32_t max_size;
```
muvarov: this struct looks over complicated.  Shouldn't it be union over all combinations? I.e.  .type and .max_num and .max_size ?

muvarov:
Merged. Please review my comments also.

PR #678, opened 2018-08-22
travis: setup default docker name space

muvarov:
Travis runs docker images which build under
github/Linaro/odp-docker-images.git project. DOCKER_NAMESPACE variable
should be set in Travis settings (in web interface) to match image
which build to images which is used for testing. If that variable is
not set, then switch to mainline images (lngodp/<image>).


.travis.yml
```diff
@@ -55,6 +55,7 @@ env:
         # you need generated new one at https://codecov.io specific for your repo.
         - CODECOV_TOKEN=a733c34c-5f5c-4ff1-af4b-e9f5edb1ab5e
         - DPDK_VERS="17.11.3"
+        - if [ -z "${DOCKER_NAMESPACE} ] ; then export DOCKER_NAMESPACE="lngodp"; fi
```
Bill-Fischofer-Linaro: We should consider another name as LNG is being retired.

muvarov:
some of tests failed due to DOCKER_NAMESPACE is already set and I can not disable it only for one PR.
I tested here that all CI test has to pass: https://travis-ci.org/muvarov/odp/builds/419329249

PR #680, opened 2018-08-23
[PATCH v1] Multi-thread support for queue perf test

psavol:
Modified queue performance test to support multiple worker threads.


muvarov:
Merged with my review.

PR #681, opened 2018-08-24
[PATCH v5] Travis: use Ubuntu 16.04 Docker image

lumag:
This depends on https://github.com/Linaro/odp-docker-images/pull/1


muvarov:
travis-odp-lng-ubuntu_16.04 needs to be added to https://github.com/Linaro/odp-docker-images to make this PR pass.

muvarov:
PR to images is Merged, but looks like you removed 18.04 which is under ./x86_64 directory there. I think it has to be there. Also this PR removes quick tests. That is also not good. What was the reason to drop 18.04?

lumag:
@muvarov build only tests are removed while it is in WIP stage.
Regarding 18.04... Well. I'd prefer to get docker in shape first. To get back build logs, installation testing, scheduler testing, etc. Those little things that you've silently dropped in your PR.

lumag:
@muvarov @Bill-Fischofer-Linaro I've updated this PR.
It restores most of previously available tests and test features, while switching testing docker image to Ubuntu 16.04 LTS

scripts/ci/distcheck.sh
```diff
@@ -1,25 +1,17 @@
 #!/bin/bash
 set -e
 
-# CC LD AR CXX has to be predifubed
-#
+if [ "${CC#clang}" != "${CC}" ] ; then
+	export CXX="clang++"
```
muvarov: interesting will CXX=`echo $CC | sed 's/clang/clang++/'`  work here. If we provide clang-3.8  for example it will use correct version for C++.

muvarov: web hides execution quotes.

scripts/ci/coverage.sh
```diff
@@ -1,29 +1,23 @@
...
-echo 1000 | tee /proc/sys/vm/nr_hugepages
-mkdir -p /mnt/huge
-mount -t hugetlbfs nodev /mnt/huge
+if [ "${CC#clang}" != "${CC}" ] ; then
+	export CXX="clang++"
+fi
 
-git clone ${CWD}/../../ odp
-cd ./odp
+cd "$(dirname "$0")"/../..
```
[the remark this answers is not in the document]
lumag: Because I want to be able to retrieve logs after build fails. With your scripts `set -e` ensures that build script exits as soon as anything fails, leaving us with empty `after_failure` output.

scripts/ci/distcheck.sh
lumag: Ugh. Then `${CC/clang/clang++}` would be easier.

.travis.yml
```diff
@@ -152,12 +152,41 @@ jobs:
...
                              fi
                 - stage: "build only"
-                  env: Ubuntu16.04_arm64
+                  env: ARCH=x86_64
+                  install:
+                          - true
+                  script:
+                          - if [ -z "${DOCKER_NAMESPACE}" ] ; then export DOCKER_NAMESPACE="opendataplane"; fi
+                          - docker run  -i -t -v `pwd`:/odp
+                              -e CC="${CC}"
```
muvarov: -e CI="$CI" also can be here. Might be in next version. Writing to not forget.

scripts/ci/check.sh
```diff
@@ -0,0 +1,17 @@
...
+echo 1000 | tee /proc/sys/vm/nr_hugepages
+mkdir -p /mnt/huge
+mount -t hugetlbfs nodev /mnt/huge
+
+# Ignore possible failures there because these tests depends on measurements
+# and systems might differ in performance.
+export CI="true"
+make check
+
+umount /mnt/huge
```
muvarov: umount will not work on error. Maybe it's not critical for Travis runs. Not sure what happens with mount inside container when container terminates. More likely good unmount will happen.

.travis.yml
```diff
@@ -68,14 +68,7 @@ compiler:
...
         - sudo apt-get install linux-headers-`uname -r`
-script:
-        - if [ -z "${DOCKER_NAMESPACE}" ] ; then export DOCKER_NAMESPACE="opendataplane"; fi
-        - if [ -n "${CROSS_ARCH}" ] ; then
-               docker run  -i -t -v `pwd`:/odp
-                 -e CC="${CC}"
-                 -e CONF="${CONF}"
-                 ${DOCKER_NAMESPACE}/travis-odp-lng-ubuntu_16.04 /odp/scripts/ci/build_${CROSS_ARCH}.sh ;
-          else
+        - if [ -z "${CROSS_ARCH}" ] ; then
```
muvarov: `if [ -z "${CROSS_ARCH}" ] && [ ! -f  /netmap/LINUX/netmap.ko ]`

caching is needed to be enabled.

.travis.yml
```diff
@@ -42,9 +33,7 @@ cache:
         ccache: true
         pip: true
         directories:
-                - dpdk
                 - netmap
-                - $HOME/cunit-install
                 - $HOME/doxygen-install
```
muvarov: doxygen can also go to container image.

.travis.yml
```diff
@@ -108,22 +102,28 @@ jobs:
                           - docker run --privileged -i -t
                               -v `pwd`:/odp --shm-size 8g
                               -e CODECOV_TOKEN="${CODECOV_TOKEN}"
-                              -e CC="${CC}" -e LD="${LD}" -e CXX="${CXX}"
+                              -e CC="${CC}"
                               -e CONF="${CONF}"
-                              ${DOCKER_NAMESPACE}/travis-odp-lng-x86 /odp/scripts/ci/coverage.sh
+                              ${DOCKER_NAMESPACE}/travis-odp-lng-ubuntu_16.04 /odp/scripts/ci/coverage.sh
```
muvarov: just make attention here if we need lng in naming. Looks like better to switch to travis-odp-ubuntu_16.04.

scripts/ci/coverage.sh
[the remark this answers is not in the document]
muvarov: gh... yep. Maybe  how about trap function just to print logs to stdout?
function print_logs()
{
}
trap print_logs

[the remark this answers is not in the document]
lumag: I don't like that. This might easily leave us with no logs in case of any bash error. And having build in a random-name dir leaves us with no easy way to recover them. You can always do git clone locally and run script using new copy.

.travis.yml
lumag: Later

.travis.yml
lumag: Also, we can change that later.

scripts/ci/coverage.sh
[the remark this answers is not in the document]
muvarov: in case if any bash error 1) bash error will be visible on stdout,stderr 2) this function will called and you can do something with logs. Linaro Jenkins for example calls script https://git.linaro.org/ci/publishing-api.git/plain/linaro-cp.py  to upload artifacts to snapshot.linaro.org. But now I do not see that other projects do upload logs. We can just find all logs and print them and Travis will store them. Or else upload them somewhere.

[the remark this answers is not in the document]
lumag: @muvarov For now I'm trying to restore back functionality that existed before merging container-based config. You can submit trap-error-logs-etc later.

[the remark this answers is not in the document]
muvarov: yes, but for now can you leave git clone, because it's already exist in the code. Removing it and pushing back is a little bit wrong workflow.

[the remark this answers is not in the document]
lumag: @muvarov well, leaving us without error logs was quite a wrong workflow.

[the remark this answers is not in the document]
lumag: In other words: `scripts/ci/*.sh` should first serve needs of CI integration rather that one's personal needs wrt his containers (e.g. I'd like to do a out-of-tree build in host-determined directory, because that better fits my development process, but it is not necessary and a bit counterproductive for CI builds).

muvarov:
Merged.

PR #683, opened 2018-08-24
[PATCH v1] Optimize plain queues by using mpmc ring instead of lock

psavol:
Performance improves (measured with queue_perf test) especially when there's contention on queue operations and especially on ARM.


muvarov:
Reviewed. Merged.

PR #685, opened 2018-08-28
[PATCH v5] linux-gen: ishm: implement huge page cache

joseppc:
With this patch, ODP will pre-allocate several huge pages at init
time. When memory is to be mapped into a huge page, one that was
pre-allocated will be used, if available, this way ODP won't have to
trap into the kernel to allocate huge pages.

The idea with this implementation is to trick ishm into thinking that
a file descriptor where to map the memory was provided, this way it
it won't try to allocate one itself. This file descriptor is one of
those previously allocated at init time. When the system is done with
this file descriptor, instead of closing it, it is put back into the
list of available huge pages, ready to be reused.

A collateral effect of this patch is that memory is not zeroed out
when it is reused.

WARNING: This patch will not work when using process mode threads.
For several reasons, this may not work when using ODP_ISHM_SINGLE_VA
either, so for this case the list of pre-allocated files is not used.

This patch should mitigate, if not solve, bug #3774:
https://bugs.linaro.org/show_bug.cgi?id=3774

To pre-allocate huge pages, define the environment variable
ODP_HP_CACHE, and possibly set it to the number of huge pages that
should be pre-allocated, setting it to -1 will reserve up to 32 huge
pages, which is currently a hard-coded limit.

example usage:
ODP_HP_CACHE=-1 ./test/validation/api/shmem/shmem_main

### DIFF
--- a/config/odp-linux-generic.conf
+++ b/config/odp-linux-generic.conf
@@ -18,14 +18,20 @@
 odp_implementation = "linux-generic"
 config_file_version = "0.0.1"
 
-# Internal shared memory allocator
+# Shared memory options
 shm: {
-	# ODP will try to reserve as many huge pages as the number indicated
-	# here, up to 64. A zero value means that no pages should be reserved.
+	# Number of cached default size huge pages. These pages are allocated
+	# during odp_init_global() and freed back to the kernel in
+	# odp_term_global(). A value of zero means no pages are cached.
+	# No negative values should be used here, they are reserved for future
+	# implementations.
+	#
+	# ODP will reserve as many huge pages as possible, which may be less
+	# than requested here if the system does not have enough huge pages
+	# available.
+	#
 	# When using process mode threads, this value should be set to 0
 	# because the current implementation won't work properly otherwise.
-	# These pages will only be freed when the application calls
-	# odp_term_global().
 	num_cached_hp = 0
 }
 
--- a/config/odp-linux-generic.conf
+++ b/config/odp-linux-generic.conf
@@ -18,6 +18,17 @@
 odp_implementation = "linux-generic"
 config_file_version = "0.0.1"
 
+# Internal shared memory allocator
+shm: {
+	# ODP will try to reserve as many huge pages as the number indicated
+	# here, up to 64. A zero value means that no pages should be reserved.
+	# When using process mode threads, this value should be set to 0
+	# because the current implementation won't work properly otherwise.
+	# These pages will only be freed when the application calls
+	# odp_term_global().
+	num_cached_hp = 0
+}
+
 # DPDK pktio options
 pktio_dpdk: {
 	# Default options
--- a/platform/linux-generic/odp_ishm.c
+++ b/platform/linux-generic/odp_ishm.c
@@ -239,15 +239,15 @@ typedef struct {
 } ishm_ftable_t;
 static ishm_ftable_t *ishm_ftbl;
 
-#define HP_CACHE_SIZE 64
 struct huge_page_cache {
 	uint64_t len;
+	int max_fds; /* maximum amount requested of pre-allocated huge pages */
 	int total;   /* amount of actually pre-allocated huge pages */
 	int idx;     /* retrieve fd[idx] to get a free file descriptor */
-	int fd[HP_CACHE_SIZE]; /* list of file descriptors */
+	int fd[];    /* list of file descriptors */
 };
 
-static struct huge_page_cache hpc;
+static struct huge_page_cache *hpc;
 
 #ifndef MAP_ANONYMOUS
 #define MAP_ANONYMOUS MAP_ANON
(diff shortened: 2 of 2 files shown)


muvarov:
this looks like ok, test case will be good to have.

Bill-Fischofer-Linaro:
@matiaselo Can you confirm this PR's affect on Bug 3774?

matiaselo:
> Can you confirm this PR's affect on Bug 3774?

Yes it does.

platform/linux-generic/odp_ishm.c
```diff
@@ -238,13 +238,150 @@ typedef struct {
 } ishm_ftable_t;
 static ishm_ftable_t *ishm_ftbl;
 
+#define HP_CACHE_SIZE 32
+struct huge_page_cache {
+	uint64_t len;
+	int total; /* index in fd that's the highest allocated */
+	int idx;   /* retrieve fd[idx] to get a free file descriptor */
+	unsigned int seq_num;
```
matiaselo: It seems there is no apparent reason to store `seq_num` here.

platform/linux-generic/odp_ishm.c
```diff
@@ -238,13 +238,150 @@ typedef struct {
...
+			ODP_ERR("Failed to creatr dir: %s\n", strerror(errno));
+			return -1;
+		}
+	}
+
+	snprintf(filename, ISHM_FILENAME_MAXLEN,
+		 "%s/odp-%d-ishm_cached-%04x",
+		 dir,
+		 odp_global_data.main_pid,
+		 seq_num++);
```
matiaselo: Why `seq_num` is increased here again?

matiaselo:
Are the cached fds closed properly in global terminate?

platform/linux-generic/odp_ishm.c
joseppc: agreed, it is not necessary

platform/linux-generic/odp_ishm.c
joseppc: Correct. It's actually not necessary, this variable is passed by value, so it has no effect whatsoever.

joseppc:
@matiaselo right, the file descriptors are not properly closed, a terminate function should be implemented.

platform/linux-generic/odp_ishm.c
```diff
@@ -238,13 +238,150 @@ typedef struct {
...
+
+	fd = hpc.fd[hpc.idx];
+	hpc.fd[hpc.idx--] = -1;
+
+	return fd;
+}
+
+static int hp_put_cached(int fd)
+{
+	if (hpc.idx > hpc.total) {
```
joseppc: Should be `>=` or we might write out of bounds.

config/odp-linux-generic.conf
```diff
@@ -18,6 +18,16 @@
 odp_implementation = "linux-generic"
 config_file_version = "0.0.1"
 
+# Internal shared memory allocator
+ishm: {
```
matiaselo: I would prefer calling this simply shm.

config/odp-linux-generic.conf
```diff
@@ -18,6 +18,16 @@
...
+	# ODP will try to reserve as many huge pages as the number indicated
+	# here, up to 64. Zero or a negative value means that no pages should
+	# be reserved.
+	# These pages will only be freed when the application call
+	# odp_term_global().
+	num_reserved_hp = 0
+}
```
matiaselo: The comment should state more clearly that this option affects only the number of cached huge pages. More pages can still be reserved, but not cached. A warning about not supporting process mode would also be helpful.

`num_cached_hp` would in my option be more clear name for this variable.

config/odp-linux-generic.conf
```diff
@@ -18,6 +18,16 @@
 odp_implementation = "linux-generic"
 config_file_version = "0.0.1"
 
+# Internal shared memory allocator
+ishm: {
+	# ODP will try to reserve as many huge pages as the number indicated
+	# here, up to 64. Zero or a negative value means that no pages should
```
matiaselo: Looks like the limit of 64 pages is removed in the next patch but this comment is not updated.

config/odp-linux-generic.conf
joseppc: Done

config/odp-linux-generic.conf
joseppc: Done!

platform/linux-generic/odp_ishm.c
```diff
@@ -238,13 +239,159 @@ typedef struct {
...
+		return -1;
+	}
+	_odp_ishmphy_unmap(addr, len, 0);
+
+	ODP_DBG("Created HP cache file %s, fd: %d\n", filename, fd);
+
+	return fd;
+}
+
+static void hp_init(void)
```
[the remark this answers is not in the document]
joseppc: The idea was to provide a "best effort" cache, so to speak, but let's discuss during the meeting today how this should behave.

joseppc:
For my own convenience, I've left the code to reserve huge pages on a "best effort" basis. However I think the team could perhaps discuss @matiaselo suggestion during connect, and decide whether ODP should behave differently before an official release is done and we have to live with it forever.

I also added the comment that negative numbers are reserved for the future. I thought that maybe a "-1" could mean that ODP should reserve all huge pages available to the system, for instance.

muvarov:
Merged.

PR #686, opened 2018-08-29
[PATCH v2] linux-gen: odp_shm_print_all() refine shm stats print

muvarov:


lumag:
@muvarov Is "lost" really lost (like in Valgrind) or it is just unused

muvarov:
@lumag odp_shm_reserve reserve some memory, then do proper alignment and provides to user pointer to aligned memory. So yes, it's unused memory which user does not have access. odp_shm_free() will give back that memory to system. Lost might be not good word as well as unused. So it's like memory overhead. Not sure which word is more suitable here.

platform/linux-generic/odp_ishm.c
```diff
@@ -1747,10 +1749,10 @@ int _odp_ishm_status(const char *title)
 			max_name_len = str_len;
 	}
 
-	ODP_PRINT("ishm blocks allocated at: %s\n", title);
-
-	ODP_PRINT("    %-*s flag len        user_len seq ref start        fd"
-		  "  file\n", max_name_len, "name");
+	ODP_PRINT("%s\n", title);
+	ODP_PRINT("    %-*s flag %-08s   %-08s   %-08s   %-3s %-3s  %-08s       %-3s file\n",
+		  max_name_len, "name", "len", "user_len", "lost",
```
muvarov: switching to unused.

muvarov:
Merged.

PR #687, opened 2018-08-30
[PATCH v2] Add options to sched_perf test application

psavol:
Add options for better control of queue usage in the test.


psavol:
v2: When event forward option is used, queue size is "total events" instead of "event per queue" (all events may end up into a single queue during a test run).

muvarov:
Merged.

PR #691, opened 2018-09-03
[PATCH v1] Tell Asciidoctor where to find the images for the IPsec documentation.

pcolmer:
The IPsec documentation doesn't specify a value for imagesdir, but the Makefile for the images is putting them into the images folder, which means that there are broken links to the images.

Setting imagesdir fixes that.


pcolmer:
Actually, I'm not sure this is needed … the users-guide.adoc file should be pulling in the IPsec documentation … and users-guide sets imagesdir. It looks like the documentation building may be broken somewhere else. I need to investigate this further …

PR #693, opened 2018-09-04
[PATCH v5] Two fxes for sysinfo

lumag:


platform/linux-generic/odp_system_info.c
```diff
@@ -443,7 +443,7 @@ int odp_sys_huge_page_size_all(uint64_t size[], int num)
 	dir = opendir("/sys/kernel/mm/hugepages");
 	if (!dir) {
 		ODP_ERR("Failed to open huge page directory\n");
```
muvarov: ODP_ERR -> ODP_DBG then.

lumag: Why? It's still an error, but just not critical error.

platform/linux-generic/arch/x86/odp_sysinfo_parse.c
```diff
@@ -15,34 +15,54 @@ int cpuinfo_parser(FILE *file, system_info_t *sysinfo)
...
+				freq_set = true;
+			}
+		} else if (!freq_set &&
+			   strstr(str, "bogomips") != NULL) {
+			pos     = strchr(str, ':');
+			if (pos == NULL)
+				continue;
+
+			if (sscanf(pos + 2, "%lf", &mhz) == 1) {
+				/* On typical x86 BogoMIPS is freq * 2 */
```
muvarov: First 2 patches are ok. But I'm not sure about that statement. My i7-4710HQ CPU @ 2.50GHz has 4988.28 bogomips. Which is about double. But frequency jumps from 1Ghz to 2.5Ghz. If we say odp app rolls in busy loop than it looks like true. At least it's better to set up some dummy number. Other solution might be to calculate frequency with odp_cpu_cycles() (rdtsc) and clock_gettime().

Bill-Fischofer-Linaro:
Typo in the PR name (fixes, not fxes). I assume this can be fixed on merge?

muvarov:
@Bill-Fischofer-Linaro it's not typo in patch itself. Nothing to correct.

platform/linux-generic/arch/x86/odp_sysinfo_parse.c
lumag: It's about max frequency, of course.

muvarov:
Looks like this can be merged now.

muvarov:
Merged.

PR #694, opened 2018-09-04
[PATCH v2] Time-related test fixes

lumag:
An attempt to fix https://bugs.linaro.org/show_bug.cgi?id=3987

### DIFF
--- a/test/validation/api/scheduler/scheduler.c
+++ b/test/validation/api/scheduler/scheduler.c
@@ -185,8 +185,19 @@ static void scheduler_test_wait_time(void)
 	upper_limit = odp_time_local_from_ns(5 * ODP_TIME_SEC_IN_NS +
 							ODP_WAIT_TOLERANCE);
 
-	CU_ASSERT(odp_time_cmp(diff, lower_limit) >= 0);
-	CU_ASSERT(odp_time_cmp(diff, upper_limit) <= 0);
+	if (odp_time_cmp(diff, lower_limit) <= 0) {
+		fprintf(stderr, "Exceed lower limit: "
+			"diff is %" PRIu64 ", lower_limit %" PRIu64 "\n",
+			odp_time_to_ns(diff), odp_time_to_ns(lower_limit));
+		CU_FAIL("Exceed lower limit\n");
+	}
+
+	if (odp_time_cmp(diff, upper_limit) >= 0) {
+		fprintf(stderr, "Exceed upper limit: "
+			"diff is %" PRIu64 ", upper_limit %" PRIu64 "\n",
+			odp_time_to_ns(diff), odp_time_to_ns(upper_limit));
+		CU_FAIL("Exceed upper limit\n");
+	}
 
 	CU_ASSERT_FATAL(odp_queue_destroy(queue) == 0);
 }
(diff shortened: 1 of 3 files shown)


test/validation/api/scheduler/scheduler.c
```diff
@@ -185,8 +185,19 @@ static void scheduler_test_wait_time(void)
 	upper_limit = odp_time_local_from_ns(5 * ODP_TIME_SEC_IN_NS +
 							ODP_WAIT_TOLERANCE);
 
-	CU_ASSERT(odp_time_cmp(diff, lower_limit) >= 0);
-	CU_ASSERT(odp_time_cmp(diff, upper_limit) <= 0);
+	if (odp_time_cmp(diff, lower_limit) <= 0) {
```
Bill-Fischofer-Linaro: Shouldn't this be < rather than <= ?

test/validation/api/scheduler/scheduler.c
```diff
@@ -185,8 +185,19 @@ static void scheduler_test_wait_time(void)
...
+		fprintf(stderr, "Exceed lower limit: "
+			"diff is %" PRIu64 ", lower_limit %" PRIu64 "\n",
+			odp_time_to_ns(diff), odp_time_to_ns(lower_limit));
+		CU_FAIL("Exceed lower limit\n");
+	}
+
+	if (odp_time_cmp(diff, upper_limit) >= 0) {
```
Bill-Fischofer-Linaro: Similarly, should be > rather than >= ?

test/validation/api/timer/timer.c
```diff
@@ -334,8 +334,8 @@ static void timer_test_queue_type(odp_queue_type_t queue_type)
 			tim = odp_timeout_timer(tmo);
 			tick = odp_timeout_tick(tmo);
 
-			CU_ASSERT(diff_period > (period_ns - (4 * res_ns)));
-			CU_ASSERT(diff_period < (period_ns + (4 * res_ns)));
+			CU_ASSERT(diff_period > (period_ns - (5 * res_ns)));
+			CU_ASSERT(diff_period < (period_ns + (5 * res_ns)));
```
Bill-Fischofer-Linaro: Not sure what the best way to deal with this is, but these sort of ad-hoc "fudge factors" seem awkward.

muvarov:
As we discussed on meeting call validation tests should not test exact time precision. Aim of validation tests is to check that time/timer generally operates and can be used for ODP application. Precision question is for measurement and performance tests and they are depend on platform and execution environment.

From other side, after I wrote lines about we might check if we can use platform capabilities and relay on that numbers instead just increasing boundaries on some numbers.

test/validation/api/scheduler/scheduler.c
lumag: It's c&p from timer code. And '<' vs '<=' is 1ns difference.

test/validation/api/timer/timer.c
lumag: Yes. Moreover they still do not guarantee stable behaviour. I'm playing with the code to fix such kinds of tests.

muvarov:
Merged.

PR #695, opened 2018-09-10
[PATCH v1] linux-gen: ishm: don't initialize anonymous mappings

joseppc:
Anonymous mappings are automatically initalized to zero.
From the mmap manual: "MAP_ANONYMOUS: The mapping is not backed by any
file; its contents are initialized to zero."

### DIFF
--- a/platform/linux-generic/odp_ishm.c
+++ b/platform/linux-generic/odp_ishm.c
@@ -1508,7 +1508,6 @@ int _odp_ishm_init_global(const odp_init_t *init)
 		goto init_glob_err1;
 	}
 	ishm_tbl = addr;
-	memset(ishm_tbl, 0, sizeof(ishm_table_t));
 	ishm_tbl->dev_seq = 0;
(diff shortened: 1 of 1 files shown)


Bill-Fischofer-Linaro:
Is this a further fix to [Bug 3774](https://bugs.linaro.org/show_bug.cgi?id=3774)?

joseppc:
@Bill-Fischofer-Linaro: no, this is just something I saw while reading the code, and I thought I would send a patch for it. It's not important, but it may save some milliseconds at start time...

muvarov:
@joseppc If it's just to save few milliseconds I think it's not really important thing. I see that  MAP_ANONYMOUS is not supported on platforms like FreeBSD and might be easy to port there. At least memset(0..) is more known then that  MAP_ANONYMOUS returns zeroed pages. But also if you removed memset you also can remove zeroing other fields. But I would left code as is.

muvarov:
Closing this PR.  Leaving memset(0) for this to stress that we want to see zero'ed memory.

PR #699, opened 2018-09-11
[PATCH v1] Scheduler optimizations and clean ups

psavol:
Incremental optimizations and clean ups. Sched_perf test shows 4-20% improvement on some test cases, 0% on others. Improvement not visible with L2fwd as majority of cycles are on packet IO side.

### DIFF
--- a/platform/linux-generic/include/odp_ring_internal.h
+++ b/platform/linux-generic/include/odp_ring_internal.h
@@ -18,9 +18,6 @@ extern "C" {
 #include <odp/api/plat/atomic_inlines.h>
 #include <odp/api/plat/cpu_inlines.h>
 
-/* Ring empty, not a valid data value. */
-#define RING_EMPTY ((uint32_t)-1)
-
 /* Ring of uint32_t data
  *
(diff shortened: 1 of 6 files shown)


platform/linux-generic/include/odp_ring_internal.h
```diff
@@ -59,7 +56,7 @@ static inline void ring_init(ring_t *ring)
 }
 
 /* Dequeue data from the ring head */
-static inline uint32_t ring_deq(ring_t *ring, uint32_t mask)
+static inline uint32_t ring_deq(ring_t *ring, uint32_t mask, uint32_t *data)
```
muvarov: type might be int here then.

psavol: The return value usage and type matches with ring_deq_multi() - for both calls it tells how many items were dequeued. It's easier to swap between single and multi calls when function prototypes are similar.

muvarov: Ok.

muvarov:
Merged.

PR #700, opened 2018-09-11
[PATCH v1] helper: iplookuptable fix puting values to table

muvarov:
On putting values to table we have to validate input data
and reject unaccepted data.

### DIFF
--- a/helper/iplookuptable.c
+++ b/helper/iplookuptable.c
@@ -585,24 +585,25 @@ prefix_insert_into_lx(
 		odph_iplookup_table_impl *tbl, prefix_entry_t *entry,
 		uint8_t cidr, odp_buffer_t nexthop, uint8_t level)
 {
-	uint8_t ret = 0;
+	int ret = 0;
 	uint32_t i = 0, limit = (1 << (level - cidr));
 	prefix_entry_t *e = entry, *ne = NULL;
 
 	for (i = 0; i < limit; i++, e++) {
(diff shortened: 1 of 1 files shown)


muvarov:
One thing which I do not understand from the code is why netmask (cidr) is coded with 7 bits. And how one bit is used for leaf in the sub tree. If somebody has more clean understanding please write few words about it. 

```
/** @internal entry struct
 *   Structure store an entry of the ip prefix table.
 *   Because of the leaf pushing, each entry of the table must have
 *   either a child entry, or a nexthop info.
 *   If child == 0 and index != ODP_BUFFER_INVALID, this entry has
 *		a nexthop info, index indicates the buffer that stores the
 *		nexthop value, and ptr points to the address of the buffer.
 *   If child == 1, this entry has a subtree, index indicates
 *		the buffer that stores the subtree, and ptr points to the
 *		address of the buffer.
 */
typedef struct {
	union {
		odp_buffer_t nexthop;
		void *ptr;
	};
	union {
		uint8_t u8;
		struct {
#if ODP_BYTE_ORDER == ODP_BIG_ENDIAN
			uint8_t child : 1;
			uint8_t cidr  : 7;
#else
			uint8_t cidr  : 7;
			uint8_t child : 1;
#endif
		};
	};
} prefix_entry_t;

```

muvarov:
wrongly pushed it to main repo :( delete this branch.

PR #701, opened 2018-09-11
[PATCH v2] helper: iplookuptable fix puting values to table

muvarov:
On putting values to table we have to validate input data
and reject unaccepted data.

### DIFF
--- a/helper/iplookuptable.c
+++ b/helper/iplookuptable.c
@@ -585,24 +585,25 @@ prefix_insert_into_lx(
 		odph_iplookup_table_impl *tbl, prefix_entry_t *entry,
 		uint8_t cidr, odp_buffer_t nexthop, uint8_t level)
 {
-	uint8_t ret = 0;
+	int ret = 0;
 	uint32_t i = 0, limit = (1 << (level - cidr));
 	prefix_entry_t *e = entry, *ne = NULL;
 
 	for (i = 0; i < limit; i++, e++) {
-		if (e->child == 1) {
-			if (e->cidr > cidr)
-				continue;
+		if (e->cidr > cidr)
+			continue;
 
+		if (e->child == 1) {
 			e->cidr = cidr;
 			/* push to next level */
 			ne = (prefix_entry_t *)e->ptr;
 			ret = prefix_insert_into_lx(
 					tbl, ne, cidr, nexthop, cidr + 8);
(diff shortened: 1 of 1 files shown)


muvarov:
One thing which I do not understand from the code is why netmask (cidr) is coded with 7 bits. And how one bit is used for leaf in the sub tree. If somebody has more clean understanding please write few words about it.

```
/** @internal entry struct
 *   Structure store an entry of the ip prefix table.
 *   Because of the leaf pushing, each entry of the table must have
 *   either a child entry, or a nexthop info.
 *   If child == 0 and index != ODP_BUFFER_INVALID, this entry has
 *		a nexthop info, index indicates the buffer that stores the
 *		nexthop value, and ptr points to the address of the buffer.
 *   If child == 1, this entry has a subtree, index indicates
 *		the buffer that stores the subtree, and ptr points to the
 *		address of the buffer.
 */
typedef struct {
	union {
		odp_buffer_t nexthop;
		void *ptr;
	};
	union {
		uint8_t u8;
		struct {
#if ODP_BYTE_ORDER == ODP_BIG_ENDIAN
			uint8_t child : 1;
			uint8_t cidr  : 7;
#else
			uint8_t cidr  : 7;
			uint8_t child : 1;
#endif
		};
	};
} prefix_entry_t;

```

helper/iplookuptable.c
```diff
@@ -678,8 +679,9 @@ odph_iplookup_table_put_value(odph_table_t tbl, void *key, void *value)
 
 	nexthop = *((odp_buffer_t *)value);
 
-	if (prefix->cidr == 0)
+	if (prefix->cidr == 0 || prefix->cidr == 255)
 		return -1;
+
 	prefix->ip = prefix->ip & (0xffffffff << (IP_LENGTH - prefix->cidr));
```
Bill-Fischofer-Linaro: Given this statement, isn't the valid range for `prefix->cidr` 0..31? In that case the above check for `prefix->cidr == 255` seems off.

muvarov: looks like yes, and it's also will be good to add to other function put,get,remove validation of passed values.

muvarov:
v2:  CIDR is in rate 0..32.

muvarov:
That is already merged. Forgot to close it.

PR #702, opened 2018-09-13
Scheduler burst size configuration options

psavol:
Changed configuration file options to control burst sizes per queue priority. Max burst size is needed to guarantee that low priority queues cannot cause large head of line blocking, when application uses schedule_multi() with large max burst size.


muvarov:
PR looks ok for me.

muvarov:
Merged.

PR #703, opened 2018-09-13
linux-gen: netmap: remove minimum frame len check

jmtilli:
(Note: this pull request is for monarch_lts)

Checking the minimum frame length is unnecessary as netmap drops truncated frames internally.

This also helps when using netmap+ODP with veth interfaces. They don't enlarge frames to 60 bytes + CRC like real Ethernet devices do. Such veth interfaces have smaller than 60 byte packets for ARP, for example. Checking the minimum frame length unnecessarily drops such packets, causing ARP to not work.

The change was taken from tigermoth_lts branch. My opinion is that although monarch_lts is starting to be a bit outdated, the change could be backported to monarch_lts as there still may be people using monarch_lts.

### DIFF
--- a/platform/linux-generic/pktio/netmap.c
+++ b/platform/linux-generic/pktio/netmap.c
@@ -606,11 +606,6 @@ static inline int netmap_pkt_to_odp(pktio_entry_t *pktio_entry,
 		return -1;
 	}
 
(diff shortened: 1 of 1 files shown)


muvarov:
Thank you. Patch added to monarch_lts branch. Not sure if we need new tag special for this. But it will be good to have in lts branch.

PR #705, opened 2018-09-14
Use enqueue multi for pktin dest queue

psavol:
Use enqueue_multi call to allow improved performance when pktio has sorted incoming packets per dest_queue.


muvarov:
Merged.

PR #706, opened 2018-09-24
linux-gen: shm: use single VA space for internal allocations in process mode

matiaselo:
V2:
- Made ODP_CONFIG_ISHM_VA_PREALLOC_SZ mode readable (Bill)

### DIFF
--- a/config/odp-linux-generic.conf
+++ b/config/odp-linux-generic.conf
@@ -33,6 +33,10 @@ shm: {
 	# When using process mode threads, this value should be set to 0
 	# because the current implementation won't work properly otherwise.
 	num_cached_hp = 0
+
+	# Allocate internal shared memory using a single virtual address space.
+	# Set to 1 to enable using process mode.
+	single_va = 0
 }
 
 # DPDK pktio options
--- a/platform/linux-generic/include/odp_config_internal.h
+++ b/platform/linux-generic/include/odp_config_internal.h
@@ -141,7 +141,7 @@ extern "C" {
  * all ODP threads (when the _ODP_ISHM_SINGLE_VA flag is used).
  * In bytes.
  */
-#define ODP_CONFIG_ISHM_VA_PREALLOC_SZ (536870912L)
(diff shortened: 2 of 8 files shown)


platform/linux-generic/include/odp_config_internal.h
```diff
@@ -141,7 +141,7 @@ extern "C" {
  * all ODP threads (when the _ODP_ISHM_SINGLE_VA flag is used).
  * In bytes.
  */
-#define ODP_CONFIG_ISHM_VA_PREALLOC_SZ (536870912L)
+#define ODP_CONFIG_ISHM_VA_PREALLOC_SZ (1073741824L)
```
Bill-Fischofer-Linaro: Would it be better to change this to `#define ODP_CONFIG_ISHM_VA_PREALLOC_SZ (1024 * 1024 * 1024L)` to make it clearer what this number is?

config/odp-linux-generic.conf
```diff
@@ -33,6 +33,10 @@ shm: {
 	# When using process mode threads, this value should be set to 0
 	# because the current implementation won't work properly otherwise.
 	num_cached_hp = 0
+
+	# Allocate internal shared memory using a single virtual address space.
+	# Set to 1 to enable using process mode.
+	single_va = 0
```
Bill-Fischofer-Linaro: Need to add a validation test that sets `single_va = 1` so that we know this actually works. Is this something planned for a follow-on PR?

platform/linux-generic/include/odp_config_internal.h
matiaselo: Makes sense, will update.

config/odp-linux-generic.conf
matiaselo: Yes, process mode is not yet working but I'll add a test when I have all the fixes done.

muvarov:
Merged. Now I have to really update my PR for this changes.

PR #707, opened 2018-09-25
[PATCH v2] linux-gen: fix process mode issues

matiaselo:
V2:
- Rebase

V3:
- Added testing and Travis support

V4:
- Put only the modified option in `process-mode.conf` (Bill)

### DIFF
--- a/.travis.yml
+++ b/.travis.yml
@@ -99,6 +99,21 @@ script:
 jobs:
         include:
                 - stage: test
+                  env: TEST=process_mode
+                  install:
+                          - true
+                  compiler: gcc
+                  script:
+                          - if [ -z "${DOCKER_NAMESPACE}" ] ; then export DOCKER_NAMESPACE="opendataplane"; fi
+                          - docker run --privileged -i -t
+                              -v `pwd`:/odp --shm-size 8g
+                              -e CODECOV_TOKEN="${CODECOV_TOKEN}"
+                              -e CC="${CC}"
+                              -e CONF="${CONF}"
+                              -e ODP_CONFIG_FILE=/odp/platform/linux-generic/test/process-mode.conf
+                              -e ODPH_PROC_MODE=1
+                              ${DOCKER_NAMESPACE}/travis-odp-lng-ubuntu_${UBUNTU_VERS} /odp/scripts/ci/check.sh ;
+                - stage: test
                   env: TEST=coverage
                   compiler: gcc
                   script:
--- a/example/classifier/odp_classifier.c
+++ b/example/classifier/odp_classifier.c
@@ -73,14 +73,13 @@ typedef struct {
 	unsigned int cpu_count; /**< Number of CPUs to use */
 	uint32_t time;		/**< Number of seconds to run */
 	char *if_name;		/**< pointer to interface names */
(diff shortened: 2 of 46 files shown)


config/odp-linux-generic.conf
```diff
@@ -33,6 +33,10 @@ shm: {
 	# When using process mode threads, this value should be set to 0
 	# because the current implementation won't work properly otherwise.
 	num_cached_hp = 0
+
+	# Allocate internal shared memory using a single virtual address space.
+	# Set to 1 to enable using process mode.
+	single_va = 0
```
Bill-Fischofer-Linaro: This seems to be duplicate of PR #706. Just temporary?

Given the scope of these changes, we really need to test with `single_va = 1` as well as the default 0 here.

Bill-Fischofer-Linaro:
Travis has issues with IPC test. Needs investigation.

config/odp-linux-generic.conf
matiaselo: >This seems to be duplicate of PR #706. Just temporary?

Yes, the first three commits are temporarily copied from PR #706 as this patch set builds on top of that PR.

matiaselo:
>Travis has issues with IPC test. Needs investigation.

This issue seems to happen quite randomly as my private Travis run passed. I suspect that the problem may be in here:
```
+		uint32_t flags = ODP_SHM_EXPORT;
+
+		if (odp_global_data.shm_single_va)
+			flags |= ODP_SHM_SINGLE_VA;
+
 		pkt_priv(pktio_entry)->type = PKTIO_TYPE_IPC_MASTER;
 		snprintf(name, sizeof(name), "%s_info", dev);
 		shm = odp_shm_reserve(name, sizeof(struct pktio_info),
-				      ODP_CACHE_LINE_SIZE,
-				      _ODP_ISHM_EXPORT | _ODP_ISHM_LOCK);
+				      ODP_CACHE_LINE_SIZE, flags);

```

The original code was passing internal defines to odp_shm_reserve() and it looked like _ODP_ISHM_LOCK was ignored internally. Taking a more closer look it seems _ODP_ISHM_LOCK maps to ODP_SHM_SINGLE_VA. @muvarov is this a correct interpretation?

muvarov:
Interesting that normal exit was printed but exit status is not 0 for ipc under coverage. I will re run this test.
```
odp_packet_io.c:521:odp_pktio_start():interface: ipc:23935:ipktio, input queues: 1, output queues: 1
odp_packet_io.c:573:odp_pktio_stop():interface: ipc:23935:ipktio
odp_packet_io.c:410:odp_pktio_close():interface: ipc:23935:ipktio
normal exit of 2 application
!!!First stage  FAILED 255!!!
FAIL pktio_ipc/pktio_ipc_run.sh (exit status: 255)
```

matiaselo:
> Interesting that normal exit was printed but exit status is not 0 for ipc under coverage. I will re run this test.

@muvarov That's odd, I can't repeat this on my system and the coverage test seems to be now passing. This PR doesn't really change anything in IPC, so this failure was probably unrelated (unless the failure happens in process mode).

muvarov:
Merged.

PR #708, opened 2018-09-26
New example application: packet dump

psavol:
Receives a packet from an interface, prints it on the screen and drops it. Options can be used to filter output. Maybe useful when debugging connection problems.


psavol:
v2: Increased max pktio name to 256 from 32 bytes. Make distcheck failed due to long path in pcap interface name.

Bill-Fischofer-Linaro:
@muvarov It seems that version number incrementing has stopped working. Please look into this when you get back.

PR #709, opened 2018-09-27
Fix queue error print format

psavol:
Add missing '%' from printf formats. Also use hex instead of int
format as queue handles are pointers.


muvarov:
Merged.

PR #710, opened 2018-09-27
test: ipc: reduce interface start poll frequency

psavol:
This test fails frequently in Travis since log get filled with
debug prints from odp_pktio_start(). Reduce polling frequency to
once per 50ms. This gives both process time to start up, before
log is filled with unnecessary error messages.


muvarov:
Merged.

PR #713, opened 2018-10-01
linux-gen: dpdk: prefetch pkt_hdr before packet init

psavol:
These prefetches improve performance with L2fwd 1 - 7.5% and with OFP fpm_burst about 9.5%.


psavol:
V2: prefetch also data at pkt[i + 2]. Performance is improved when parsing is done (e.g. with OFP).

muvarov:
Merged.

PR #714, opened 2018-10-03
[PATCH v3] IPsec perfomance improvements

JannePeltonen:
This set of patches improve the performance of outbound IPsec processing by making it scale better to multiple threads.

The patches reduce the amount of shared read-write data accessed for every packet. Many of the patches introduce some form of thread-local buffering. One of the patches simply removes per-packet SA refcount update and relies on the application for synchronization as required by the ODP API. See the individual commit messages for more details.

Biggest speedup can be seen with small packets and outbound processing only. The SA lookup in inbound IPsec processing still has a scalability bottleneck that this patch series does not address.

A few test results about forwarding to an IPsec tunnel using OFP in x86:

kpps before / kpps after / case
417 / 941 / AES-CBC, SHA-1, 2 threads
597 / 4495 / AES-CBC, SHA-1, 10 threads
1558 / 1864 / AES-GCM, 2 threads
2444 / 9020 / AES-GCM, 10 threads

One of the patches contains a fix for an IP ID bug (PR 4013).

### DIFF
--- a/platform/linux-generic/include/odp_ipsec_internal.h
+++ b/platform/linux-generic/include/odp_ipsec_internal.h
@@ -181,9 +181,6 @@ struct ipsec_sa_s {
 				odp_ipsec_ipv4_param_t param;
 				odp_u32be_t	src_ip;
 				odp_u32be_t	dst_ip;
-
-				/* 32-bit from which low 16 are used */
-				odp_atomic_u32_t hdr_id;
 			} tun_ipv4;
 			struct {
 				odp_ipsec_ipv6_param_t param;
@@ -273,6 +270,12 @@ int _odp_ipsec_sa_replay_precheck(ipsec_sa_t *ipsec_sa, uint32_t seq,
  */
 int _odp_ipsec_sa_replay_update(ipsec_sa_t *ipsec_sa, uint32_t seq,
 				odp_ipsec_op_status_t *status);
+
+/**
+  * Allocate an IPv4 ID for an outgoing packet.
+  */
+uint16_t _odp_ipsec_sa_alloc_ipv4_id(ipsec_sa_t *ipsec_sa);
+
 /**
  * Try inline IPsec processing of provided packet.
  *
--- a/platform/linux-generic/include/odp_ipsec_internal.h
+++ b/platform/linux-generic/include/odp_ipsec_internal.h
@@ -109,8 +109,11 @@ struct ipsec_sa_s {
 			} in;
 
 			struct {
-				odp_atomic_u64_t counter; /* for CTR/GCM */
-				odp_atomic_u32_t seq;
+				/*
+				 * 64-bit sequence number that is also used as
+				 * CTR/GCM IV
+				 */
+				odp_atomic_u64_t seq;
 			} out;
 		};
 	} hot;
--- a/platform/linux-generic/include/odp_ipsec_internal.h
+++ b/platform/linux-generic/include/odp_ipsec_internal.h
@@ -207,6 +207,11 @@ uint32_t _odp_ipsec_cipher_iv_len(odp_cipher_alg_t cipher);
 /* Return digest length required for the cipher for IPsec use */
 uint32_t _odp_ipsec_auth_digest_len(odp_auth_alg_t auth);
 
+/*
+ * Get SA entry from handle without obtaining a reference
+ */
+ipsec_sa_t *_odp_ipsec_sa_entry_from_hdl(odp_ipsec_sa_t sa);
+
 /**
  * Obtain SA reference
  */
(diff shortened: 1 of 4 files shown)


platform/linux-generic/include/odp_ipsec_internal.h
```diff
@@ -207,6 +207,11 @@ uint32_t _odp_ipsec_cipher_iv_len(odp_cipher_alg_t cipher);
 /* Return digest length required for the cipher for IPsec use */
 uint32_t _odp_ipsec_auth_digest_len(odp_auth_alg_t auth);
 
+/*
+ * Get SA entry from handle without obtaining a reference
+ */
+ipsec_sa_t *odp_ipsec_sa_entry_from_hdl(odp_ipsec_sa_t sa);
```
lumag: Please use `_odp` prefix for internal functions

platform/linux-generic/odp_ipsec.c
```diff
@@ -1884,9 +1885,6 @@ int odp_ipsec_out_inline(const odp_packet_t pkt_in[], int num_in,
 		in_pkt++;
 		sa_idx += sa_inc;
 		opt_idx += opt_inc;
-
-		/* Last thing */
-		_odp_ipsec_sa_unuse(ipsec_sa);
```
lumag: I'm unsure about this patch. Application can not know when out inline packets were actually sent, so an application will send several packets using `odp_ipsec_out_inline()`, then it will call `odp_ipsec_sa_disable()` and wait for SA to be actually disabled. Thus refcounting is necessary here.

lumag: After discussion, this code will be safe with the current code. Could you please add huuuge warning, that if one would like to add async crypto as IPsec backing code or proper async IPsec implementation, he has to re-implement some kind of SA safety here.

platform/linux-generic/odp_ipsec_sad.c
```diff
@@ -24,8 +24,47 @@
...
+ * SA for every packet, we increment the global counters once for several
+ * packets. We decrement a preallocated thread-local quota for every
+ * packet. When the quota runs out, we get a new quota by incementing the
+ * global counter.
+ *
+ * This improves performance but the looser synchronization between
+ * threads makes life time warnings and errors somewhat inaccurate.
+ * The warnings and errors may get triggered a bit too early since
+ * some threads may still have unused quota when the first thread
+ * hits the limit.
```
lumag: I'd like to ask you to split this patch to a separate pull request:
 - threads can report failure early (e.g. in worst case on 96-core box warning can come up around 6000 patches early)
 - threads can disagree on counters state: one thread can start returning 'overrun' errors, while other threads (still having quota) can process packets freely.

JannePeltonen:
I updated the patches. Changes in v2:

- Renamed odp_ipsec_sa_entry_from_hdl() to _odp_ipsec_sa_entry_from_hdl()
- Added a comment that warns about the need to rethink concurrency if the current synchronous design is changed to asynchronous.
- Added a new patch for fixing PR 4017 that caused the validation failures
- Removed SA lifetime check improvement from this pull request (moved to a separate pull request)
- Rebased on top of latest master

platform/linux-generic/include/odp_ipsec_internal.h
```diff
@@ -94,6 +94,27 @@ int _odp_ipsec_status_send(odp_queue_t queue,
...
+			struct {
+				odp_atomic_u64_t antireplay;
+			} in;
+
+			struct {
+				odp_atomic_u64_t counter; /* for CTR/GCM */
+				odp_atomic_u32_t seq;
+			} out;
+		};
+	} hot;
```
Bill-Fischofer-Linaro: Should `ROUNDUP_CACHE_LINE()` be used here to guarantee that the `hot` struct is a integral number of cache lines? Otherwise data following hot may reside in the same cache line, leading to the line sharing issues you're trying to avoid here.

platform/linux-generic/odp_ipsec.c
JannePeltonen: @Bill-Fischofer-Linaro The speedup of a single patch in this patch set depends on the order of the patches. In the current patch order, the refcount removal patch speeds up my unidirectional IPsec tunneling test about 45 % with 10 worker threads.

platform/linux-generic/include/odp_ipsec_internal.h
JannePeltonen: I am not totally sure what is the correct way to use the alignment attribute but to my limited understanding the alignment attribute in this patch is a type attribute (as opposed to a variable attribute) that also makes sizeof(hot) a multiple of cache line size. At least in my GCC 7.1 & x86_64 environment that is what happens and struct hot is not sharing cache lines with what follows it.

muvarov:
<pre>v3 checks:
checkpatch.pl - OK
</pre>

muvarov:
Merged.

muvarov:
Forgot to git push. Now merged.

PR #715, opened 2018-10-04
[PATCH v1] travis: export CI for first distcheck

muvarov:
scripts runs 2 distchecks but only for second one CI
was exported.


muvarov:
Merged as simple fix and Travis will be more stable.

PR #716, opened 2018-10-04
[PATCH v1] linux-gen: do not use huge pages for internal allocations

muvarov:
Some linux-generic internal shared memory allocations have
to be in normal pages due to small required data. Relaying
on odp_sys_huge_page_size() is not really correct, because
call returns default huge page size. But default huge page
size is definned in kernel boot parameter as:
default_hugepagesz=1G hugepagesz=1G hugepages=2
So in that case for small allocation linux-gen will create 1GB
huge page. This patch introduces internal flag to shm funtion
to allocate hp.
This patch remains allowing changes on top of it:
For now pools are in huge page. And for apps with small pool it's
big overhead of unused memory. We should take into account
odp_sys_huge_page_size_all() call and found best sized huge pages.

---
v2: - rebase
       - use internal shm flags then ishm flags


muvarov:
Merged.

PR #717, opened 2018-10-05
Sched pktio pipeline options

psavol:
Added options to configure sched_pktio test to run packets through multiple pipeline steps (queues) between packet input and output. This emulates better real application usage of scheduler.

While testing run into performance problems due to many, long queues can hold lots of packets (==> even a large pool may get empty). So, changed queue implementation to follow queue size parameter more closely (also on small queue sizes). Also added a validation test case to test it.


psavol:
Maxim, this can be merged as is. The performance degradation (3-5%) happens only for single core test cases in L2fwd. Matias is updating DPDK zero copy integration, which will anyway change packet alignment after this.

platform/linux-generic/include/odp_config_internal.h
```diff
@@ -125,7 +125,9 @@ extern "C" {
  * defined segment length (seg_len in odp_pool_param_t) will be rounded up into
  * this value.
  */
-#define CONFIG_PACKET_SEG_LEN_MIN CONFIG_PACKET_MAX_SEG_LEN
+#define CONFIG_PACKET_SEG_LEN_MIN ((2 * 1024) - \
+				   CONFIG_PACKET_HEADROOM - \
+				   CONFIG_PACKET_TAILROOM)
```
Bill-Fischofer-Linaro: Have we quantified the effect of this change on applications that use jumbo frames?

psavol: Those applications should set pool parameters pkt.len, pkt.seg_len, etc. This config rounds up small len/seg_len parameters (e.g. 1.5k => 2k), but does not round down large parameters. E.g. 8k jumbo frame support remains the same if application has set pkt.len to 8k.

PR #718, opened 2018-10-08
[PATCH v3] IPsec:  Make SA lifetime checking scale better to multiple cores

JannePeltonen:
The last patch in this pull request improves IPsec performance by making SA lifetime checking scale better to multiple cores. The patch was separated from pull request 714.

All the patches except the last one in this pull request are from PR 714 and discussed there. They are included here only to provide the dependencies for the last patch and will be removed from here when PR 714 gets resolved. IOW, the last patch is what this PR is about.

### DIFF
--- a/platform/linux-generic/odp_ipsec_sad.c
+++ b/platform/linux-generic/odp_ipsec_sad.c
@@ -42,7 +42,42 @@
 #warning IPV4_ID_RING_SIZE is too small for the maximum number of threads.
 #endif
 
+/*
+ * To avoid checking and updating the packet and byte counters in the
+ * SA for every packet, we increment the global counters once for several
+ * packets. We decrement a preallocated thread-local quota for every
+ * packet. When the quota runs out, we get a new quota by incementing the
+ * global counter.
+ *
+ * This improves performance but the looser synchronization between
+ * threads makes life time warnings and errors somewhat inaccurate.
+ * The warnings and errors may get triggered a bit too early since
+ * some threads may still have unused quota when the first thread
+ * hits the limit.
+ */
+#define SA_LIFE_PACKETS_PREALLOC  64
+#define SA_LIFE_BYTES_PREALLOC    4000
+
+typedef struct sa_thread_local_s {
+	/*
+	 * Packets that can be processed in this thread before looking at
+	 * the SA-global packet counter and checking hard and soft limits.
+	 */
+	uint32_t packet_quota;
+	/*
+	 * Bytes that can be processed in this thread before looking at
+	 * at the SA-global byte counter and checking hard and soft limits.
+	 */
+	uint32_t byte_quota;
+	/*
+	 * Life time status when this thread last checked the global
+	 * counter(s).
+	 */
+	odp_ipsec_op_status_t lifetime_status;
+} sa_thread_local_t;
+
 typedef struct ODP_ALIGNED_CACHE ipsec_thread_local_s {
+	sa_thread_local_t sa[ODP_CONFIG_IPSEC_SAS];
 	uint16_t first_ipv4_id; /* first ID of current block of IDs */
 	uint16_t next_ipv4_id;  /* next ID to be used */
 } ipsec_thread_local_t;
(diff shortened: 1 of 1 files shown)


muvarov:
is it replacement of #714 or it's new version of #714. What to merge?

JannePeltonen:
This is not a replacement of 714 but basically one patch (linux-gen: ipsec: make SA lifetime checking more scalable to multiple threads) separated from 714. Since this one patch does not apply directly to master but depends on the patches of 714, this PR also includes the prerequisite patches that are discussed and reviewed in PR 714 and not here. This is what I was suggested to do in the architecture meeting.

PR 714 should be merged once reviewed. After that is done, I will update this PR to have only the additional patch that is no longer part of 714. After that and proper review this PR is ok for merging.

Bill-Fischofer-Linaro:
@JannePeltonen Please rebase since PR #714 has now been merged.

muvarov:
<pre>v1 checks:
checkpatch.pl - OK
</pre>

lumag:
@JannePeltonen could you please rebase this PR?

JannePeltonen:
v2: Rebased on the latest master.

lumag:
My major concern is that after this patch cores/threads might behave in the inconsistent manner: some cores have acquired quota that they are still willing to spend, while other cores might miss the quota and start rejecting packets in advance. An application might use 'overcommit' event as a notification from IPsec subsystem that it is now a safe time to disable SA, as it will not accept packet (and start deleting/renegotiating), while it's own threads will still use this SA.
I think that proper way would be to use this as fast path and switch back to old method when remaining amounts drop below `num_cores * quota`.

JannePeltonen:
> An application might use 'overcommit' event as a notification from IPsec subsystem that it is now a safe time to disable

This would be against the API specification, which says: "Application must stop providing the SA as parameter to new IPSEC input/output operations before calling disable."

I am not sure how switching to the old (more synchronized) method can be done on the fly without slowing down the normal operation. And I think it may be problematic to switch to use a "slow path" at some point during operation (i.e. after reaching soft limit but before reaching hard limit) since then the slow path performance dictates the maximum capacity of the system. Maybe the approach could be ok if the switch to the slow path was done way later than reaching the soft limit so that the slow path should not happen normally. But setting the thresholds would be difficult.

Maybe a better way to proceed would be to make the parameters of the current patch configurable as someone (Petri?) proposed in one call?

platform/linux-generic/odp_ipsec_sad.c
```diff
@@ -701,30 +750,47 @@ int _odp_ipsec_sa_stats_precheck(ipsec_sa_t *ipsec_sa,
...
-	uint64_t bytes = odp_atomic_fetch_add_u64(&ipsec_sa->hot.bytes, len) + len;
-	uint64_t packets = odp_atomic_fetch_add_u64(&ipsec_sa->hot.packets, 1) + 1;
-	int rc = 0;
+	sa_thread_local_t *sa_tl = ipsec_sa_thread_local(ipsec_sa);
+	uint64_t packets, bytes;
+
+	if (odp_unlikely(sa_tl->packet_quota == 0)) {
+		packets = odp_atomic_fetch_add_u64(&ipsec_sa->hot.packets,
+						   SA_LIFE_PACKETS_PREALLOC);
+		packets++;
```
psavol: Shouldn't this be:

packets += SA_LIFE_PACKETS_PREALLOC

platform/linux-generic/odp_ipsec_sad.c
```diff
@@ -701,30 +750,47 @@ int _odp_ipsec_sa_stats_precheck(ipsec_sa_t *ipsec_sa,
...
+	}
+	sa_tl->packet_quota--;
 
-	if (ipsec_sa->soft_limit_bytes > 0 &&
-	    bytes > ipsec_sa->soft_limit_bytes)
-		status->warn.soft_exp_bytes = 1;
+	if (odp_unlikely(sa_tl->byte_quota < len)) {
+		bytes = odp_atomic_fetch_add_u64(&ipsec_sa->hot.bytes,
+						 len + SA_LIFE_BYTES_PREALLOC);
+		bytes++;
```
psavol: += len + SA_LIFE_BYTES_PREALLOC ??

JannePeltonen:
Updated with a fix to the bug found by Petri. Limit reaching indication no longer comes too late.

muvarov:
Merged.

PR #721, opened 2018-10-10
[PATCH v2] Fix packets to different destination queues interleaved 

psavol:
Packets to different destination queues interleaved resulted
bad indexes when pktin_recv_buf() was recording which packets
belong to the same dest_queue.


muvarov:
<pre>v1 checks:
checkpatch.pl - OK
</pre>

psavol:
V2: added dummy initialization to make a compiler happy in Shippable

muvarov:
Merged.

PR #722, opened 2018-10-10
[PATCH v1] validation: packet: add packet reset test

psavol:
Added test for odp_packet_reset(). Tried to reproduce bug https://bugs.linaro.org/show_bug.cgi?id=3971 , but everything seems to work fine.


muvarov:
Merged.

PR #724, opened 2018-10-15
[PATCH v1] Remove iquery scheduler

psavol:
Iquery scheduler is clearly slower and scales worse than basic and scalable scheduler. It has not been developed for a long time. Remove it to reduce maintenance overhead.


muvarov:
Merged to master and api-next.

PR #726, opened 2018-10-17
[PATCH v3] Travis tuning

psavol:
Tune, fix and (try to) optimize travis build/test sequence. Many small patches are used for easier validation that each fix works as it should.

Build/test time optimizations:
- Use make -j
- Prefer many equal length tests vs one long test => enables parallel execution
- Start long test early => enables parallel execution of other jobs while the long one is running
- Avoid duplicate tests, and duplicate 'make check' when build only is enough
- Fail fast. Run first all tests with GCC, if all those pass it's very likely that the same passes on clang (first phase include already generic clang build)
- Don't install and test Netmap on every test case. Many test cases do not affect netmap operation, but netmap build/test takes time

Fixes:
- Build/test optional schedulers (sp and scalable) with normal compiler flags (== optimizations on)

Additionally, it seems that netmap tests has been (and are still) skipped. So, there's no difference in netmap vs non-netmap test times, but there should once netmap tests are fixed to run again. 

Travis run time improvement:1 hr 36 min 36 sec -  1 hr 21 min 33 sec = 15min
V3: 1 hr 37 min 23 sec - 1 hr 22 min 12 sec = 15min

### DIFF
--- a/.travis.yml
+++ b/.travis.yml
@@ -154,7 +154,17 @@ jobs:
                           - docker run --privileged -i -t
                               -v `pwd`:/odp --shm-size 8g
                               -e CC="${CC}"
-                              -e CONF="${CONF}"
+                              -e CONF="--enable-user-guides"
+                              ${DOCKER_NAMESPACE}/travis-odp-lng-ubuntu_16.04 /odp/scripts/ci/distcheck.sh
+                - stage: test
+                  env: TEST=distcheck_nonabi
+                  compiler: gcc
+                  script:
+                          - if [ -z "${DOCKER_NAMESPACE}" ] ; then export DOCKER_NAMESPACE="opendataplane"; fi
+                          - docker run --privileged -i -t
+                              -v `pwd`:/odp --shm-size 8g
+                              -e CC="${CC}"
+                              -e CONF="--enable-user-guides --disable-abi-compat"
                               ${DOCKER_NAMESPACE}/travis-odp-lng-ubuntu_16.04 /odp/scripts/ci/distcheck.sh
                 - stage: "build only"
                   env: TEST=doxygen
--- a/.travis.yml
+++ b/.travis.yml
@@ -99,6 +99,28 @@ script:
 jobs:
         include:
                 - stage: test
+                  env: TEST=scheduler_sp
+                  compiler: gcc
+                  script:
+                          - if [ -z "${DOCKER_NAMESPACE}" ] ; then export DOCKER_NAMESPACE="opendataplane"; fi
+                          - docker run --privileged -i -t
+                              -v `pwd`:/odp --shm-size 8g
+                              -e CC="${CC}"
+                              -e CONF=""
+                              -e ODP_SCHEDULER=sp
+                              ${DOCKER_NAMESPACE}/travis-odp-lng-ubuntu_16.04 /odp/scripts/ci/check.sh
+                - stage: test
+                  env: TEST=scheduler_scalable
+                  compiler: gcc
+                  script:
+                          - if [ -z "${DOCKER_NAMESPACE}" ] ; then export DOCKER_NAMESPACE="opendataplane"; fi
+                          - docker run --privileged -i -t
+                              -v `pwd`:/odp --shm-size 8g
+                              -e CC="${CC}"
+                              -e CONF=""
+                              -e ODP_SCHEDULER=scalable
+                              ${DOCKER_NAMESPACE}/travis-odp-lng-ubuntu_16.04 /odp/scripts/ci/check.sh
+                - stage: test
                   env: TEST=process_mode
                   install:
                           - true
--- a/.travis.yml
+++ b/.travis.yml
@@ -45,9 +45,11 @@ env:
         - CODECOV_TOKEN=a733c34c-5f5c-4ff1-af4b-e9f5edb1ab5e
         - UBUNTU_VERS="16.04"
         - BUILD_ONLY=0
+        - NETMAP=0
     matrix:
         - CONF=""
         - CONF="--disable-abi-compat"
+        - NETMAP=1 CONF=""
         - BUILD_ONLY=1 ARCH="arm64"
         - BUILD_ONLY=1 ARCH="armhf"
         - BUILD_ONLY=1 ARCH="powerpc"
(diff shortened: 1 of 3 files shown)


scripts/ci/distcheck.sh
```diff
@@ -14,8 +14,8 @@ cd "$(dirname "$0")"/../..
 # and systems might differ in performance.
 export CI="true"
 
-make distcheck
+make -j $(nproc) distcheck
```
muvarov: That is interesting moment. Travis is passed but wouldn't it bring some suprises in future?  Tests like pktio use veth devices and probably will not work correctly if different tests run in the same time.

.travis.yml
```diff
@@ -36,6 +36,10 @@ cache:
                 - netmap
                 - $HOME/doxygen-install
 
+compiler:
+        - gcc
+        - clang
+
```
muvarov: "travis: run all test with gcc first" patch looks like is not needed. I do not see in result matrix that all gcc tests run prior clang.

scripts/ci/check_sched_scalable.sh
```diff
@@ -0,0 +1,19 @@
...
+./configure
+make -j $(nproc)
+
+echo 1000 | tee /proc/sys/vm/nr_hugepages
+mkdir -p /mnt/huge
+mount -t hugetlbfs nodev /mnt/huge
+
+export CI="true"
+
+ODP_SCHEDULER=scalable make check
```
muvarov: no need to separate script file for each run. Just pass new environment variable to Docker. -e ODP_SCHEDULER="${ODP_SCHEDULER}" and in .travis.yaml env: setup it, like "env:  ODP_SCHEDULER=scalable"

.travis.yml
```diff
@@ -154,6 +154,16 @@ jobs:
...
+                - stage: test
+                  env: TEST=distcheck_nonabi
+                  compiler: gcc
+                  script:
+                          - if [ -z "${DOCKER_NAMESPACE}" ] ; then export DOCKER_NAMESPACE="opendataplane"; fi
+                          - docker run --privileged -i -t
+                              -v `pwd`:/odp --shm-size 8g
+                              -e CC="${CC}"
+                              -e CONF="${CONF}"
+                              ${DOCKER_NAMESPACE}/travis-odp-lng-ubuntu_16.04 /odp/scripts/ci/distcheck_nonabi.sh
```
muvarov: -e CONF="${CONF}"  actually sets up environment in Docker which is passed to run script. I.e. script has to be: ./configure ${CONF}

.travis.yml
```diff
@@ -76,13 +77,13 @@ matrix:
     env: BUILD_ONLY=1 ARCH="i386"
 
 install:
-        - sudo apt-get install linux-headers-`uname -r`
-        - if [ -z "${ARCH}" -o "${ARCH}" == "x86_64" ] ; then
-               echo "compilling netmap";
+        - if [ ${NETMAP} ] ; then
```
muvarov: better to compare it with 1. If somebody will set NETMAP=0 if statement will be true.

.travis.yml
psavol: Usually this change works like it should (all GCC first, then all clang), but on sometimes travis runs gcc/clang mixed. Also all travis examples files have "compiler" defined first and "env" after that.

.travis.yml
psavol: It that case shell complains when NETMAP is not defined, and I don't want to define NETMAP=0 for most test cases.

muvarov: No, in bash  echo ${NOT_DEFINED} is simple nothing. Bash never complains that variable is not defined. But if someone will define NETMAP=0 then it will be if [ "0" ]; then, which is actually true statement. I think after some time we can forgot that we or define it or not refine. I suggest here or define it as NETMAP=y or compare it with 1.

scripts/ci/check_sched_scalable.sh
psavol: Script files removed. Used -e ODP_SCHEDULER=sp. Didn't add to matrix (only make check with gcc).

scripts/ci/distcheck.sh
lumag: Then this needs to be fixed. Let's make parallel tests work in the end.

.travis.yml
lumag: @psavol I also see gcc/clang intermixed. Let's split this patch out, so that we can check if it works or not later.

scripts/ci/check_sched_scalable.sh
lumag: @psavol script files are still here.

scripts/ci/distcheck_nonabi.sh
```diff
@@ -0,0 +1,15 @@
...
+set -e
+
+if [ "${CC#clang}" != "${CC}" ] ; then
+	export CXX="clang++"
+fi
+
+cd "$(dirname "$0")"/../..
+./bootstrap
+./configure \
+	--enable-user-guides --disable-abi-compat
```
lumag: I'd prefer to see standard script extended, rather than having separate nonabi test script. Not to mention that in this case you configure main ODP instance without abi-compat, but the ODP which distcheck will actually check will be tested with default option (abi-compat). Please use `DISTCHECK__CONFIGURE_FLAGS`.

scripts/build-pktio-dpdk
```diff
@@ -28,7 +28,7 @@ sed -ri 's,(CONFIG_RTE_LIBRTE_PMD_PCAP=).*,\1y,' .config
 popd
 
 #Build DPDK
-make build O=${TARGET} EXTRA_CFLAGS="-fPIC"
+make -j $(nproc) build O=${TARGET} EXTRA_CFLAGS="-fPIC"
```
lumag: This script is not used during ODP Travis builds.

lumag:
@psavol Looks good overall, thanks for your work. Could you please fix `distcheck` issue.

scripts/ci/check_sched_scalable.sh
psavol: Will be removed in next version...

.travis.yml
psavol: I'm just running a test on v3 on private branch, and this works again OK. Anyway, I moved the patch the last in the set.

.travis.yml
psavol: Next version removes separate nonabi.sh and harmonizes CONF usage.

.travis.yml
psavol: if [ ${NOT_DEFINED} -eq 1]

Results this warning:
line 5: [: -eq: unary operator expected

Anyway, I'll try to add NETMAP=0 as global and then override that.

scripts/build-pktio-dpdk
psavol: It's not used anymore. But some dependency/readme file refer to it.

scripts/ci/distcheck.sh
psavol: I'll remove -j from distcheck as I don't have time to test it more today.

psavol:
V3: should fix all comments. Make -j not used in distcheck. Extra scripts removed. CONF env variable used. Compare bash if statement compare NETMAP/BUILD_ONLY to 1.

muvarov:
Merged.

PR #727, opened 2018-10-17
[PATCH v1] example: stop and close pktio devices on exit

MatiasElo:
Stop and close used pktio devices on exit to free used resources.


matiaselo:
@lumag Thanks!

PR #728, opened 2018-10-17
[PATCH v2] example: generator: use odp_wait_time_ns() instead of timers

matiaselo:
Simplify code by using odp_wait_time_ns() instead of timers.

V2:
- Remove printf() from tx loop


lumag:
If it doesn't harm scalability, it's definitely

matiaselo:
>If it doesn't harm scalability, it's definitely

While checking this I noticed that there was a printf() in every TX loop, so I added a second commit removing the print.

muvarov:
Merged

PR #729, opened 2018-10-17
Inline timer fixes

matiaselo:


platform/linux-generic/include/odp_queue_basic_internal.h
```diff
@@ -54,6 +54,7 @@ struct queue_entry_s {
 		ring_spsc_t  ring_spsc;
 	};
 
+	odp_atomic_u64_t     timers;
```
muvarov: I think it's better to name it timers_cnt or timers_num.

platform/linux-generic/odp_queue_basic.c
```diff
@@ -949,6 +950,20 @@ static int queue_api_enq_multi(odp_queue_t handle,
...
+
+	odp_atomic_inc_u64(&queue->s.timers);
+}
+
+static void queue_timer_rem(odp_queue_t handle)
+{
+	queue_entry_t *queue = qentry_from_handle(handle);
+
+	odp_atomic_dec_u64(&queue->s.timers);
+}
```
muvarov: these 2 functions repeated in all schedulers. how about to put them to .h ?

platform/linux-generic/odp_queue_scalable.c
```diff
@@ -844,6 +846,11 @@ static int queue_deq_multi(odp_queue_t handle, odp_event_t ev[], int num)
 		num = QUEUE_MULTI_MAX;
 
 	queue = qentry_from_ext(handle);
+
+	if (odp_global_rw->inline_timers &&
+	    odp_atomic_load_u64(&queue->s.timers))
+		timer_run();
```
muvarov: check for inline timers is not needed because it's done inside timer_run():
```
static inline unsigned timer_run(void)
{
	return inline_timers ? _timer_run() : 0;
}
```

platform/linux-generic/include/odp_queue_basic_internal.h
muvarov: or timers_ref_cnt

platform/linux-generic/odp_queue_scalable.c
matiaselo: The `odp_global_rw->inline_timers` check is here to avoid unnecessary atomic load in case of inline timer are not used. The compiler should be able to optimise away the second check from the inlined `timer_run()` function.

platform/linux-generic/odp_queue_basic.c
matiaselo: As the queue implementations should be independent I wouldn't like to add a common header file for such trivial functions.

platform/linux-generic/include/odp_queue_basic_internal.h
matiaselo: OK, will use use a more descriptive name.

Bill-Fischofer-Linaro: Looks like this name change is still pending?

platform/linux-generic/include/odp_queue_scalable_internal.h
```diff
@@ -35,6 +35,7 @@ struct queue_entry_s {
 	sched_elem_t     sched_elem;
 
 	odp_ticketlock_t ODP_ALIGNED_CACHE lock;
+	odp_atomic_u64_t timers;
```
Bill-Fischofer-Linaro: Same comment here as @muvarov mentioned earlier. `timers` here is too vague.  Something to indicate this is a count of the number of active timers would be better.

platform/linux-generic/odp_queue_basic.c
Bill-Fischofer-Linaro: Agree with @matiaselo on this. I'm OK with these in each scheduler.

platform/linux-generic/odp_timer.c
```diff
@@ -1093,6 +1114,11 @@ int odp_timer_capability(odp_timer_clk_src_t clk_src,
 odp_timer_pool_t odp_timer_pool_create(const char *name,
 				       const odp_timer_pool_param_t *param)
 {
+	if (odp_global_ro.init_param.not_used.feat.timer) {
+		ODP_ERR("Trying to use disabled ODP feature.\n");
+		return ODP_TIMER_POOL_INVALID;
```
Bill-Fischofer-Linaro: Worth an `ODP_ASSERT()` here? Behavior is undefined if an application tries to use a feature it's said it isn't going to use so these runtime courtesy checks aren't strictly needed, but an assert can be a useful debugging aid.

matiaselo: When trying to use a disabled feature the implementation behaviour is undefined and it may crash elsewhere. As `ODP_ASSERT()` won't check anything unless debug mode is enabled and this is not a fast path function I'd prefer to keep this check here.

PR #730, opened 2018-10-17
[PATCH v2] Pool config file option

psavol:
Add max num packets into config file as by default the max num packets needed to be limited to 256k. Changed max pool size back to 1M. So, that user can again create 1M packet pools (with config file change) if system memory allows.

Also cleaned up config header file as there was a scalable queue specific config.


psavol:
V2: improve config file version check output. Version number is incremented, so it's likely that users bump into version number mismatch (as did process mode validation test).

muvarov:
Merged.

PR #731, opened 2018-10-18
[PATCH v1] linux-gen: pool: increase minimum packet segment length

matiaselo:
Some DPDK NICs need at least 2176 byte buffers (2048B + headroom) to not
segment standard ethernet frames. Increase minimum segment length to avoid
this and add matching check to zero-copy dpdk pktio pool create.

Reported-by: P. Gyanesh Kumar Patra <pgyanesh.patra@gmail.com>


lumag:

muvarov:
Merged.

PR #732, opened 2018-10-18
[PATCH API-NEXT v2] api: init: add use_single_va member to odp_init_t

matiaselo:
Enables an application to support process mode without the need to modify ODP configuration file.

New odph_options() helper function is added to enable an application to find out if ODP should be started in process mode.

The first three commits are copied from master branch and are not part of this PR:
    linux-gen: queue: add internal interface for adding/removing inline timers
    linux-gen: timer: reduce inline timer overhead
    linux-gen: timer: run inline timers during queue dequeue operations

### DIFF
--- a/config/odp-linux-generic.conf
+++ b/config/odp-linux-generic.conf
@@ -16,7 +16,7 @@
 
 # Mandatory fields
 odp_implementation = "linux-generic"
-config_file_version = "0.1.3"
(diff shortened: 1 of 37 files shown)


Bill-Fischofer-Linaro:
@matias as we discussed today, please advise if you'd like to merge this and build on it or if you'd rather replace it with a different PR or set of PRs.

PR #734, opened 2018-10-22
[PATCH API-NEXT v19] api: schedule: add scheduler flow aware mode

lumag:
ODP scheduler configuration to support flow aware mode

### DIFF
--- a/example/packet/odp_pktio.c
+++ b/example/packet/odp_pktio.c
@@ -412,6 +412,9 @@ int main(int argc, char *argv[])
 	}
 	odp_pool_print(pool);
 
+	/* Config and start scheduler */
+	odp_schedule_config(NULL);
+
 	/* Create a pktio instance for each interface */
 	for (i = 0; i < args->appl.if_count; ++i)
 		create_pktio(args->appl.if_names[i], pool, args->appl.mode);
--- a/include/odp/api/spec/event.h
+++ b/include/odp/api/spec/event.h
@@ -210,6 +210,44 @@ void odp_event_free_multi(const odp_event_t event[], int num);
 void odp_event_free_sp(const odp_event_t event[], int num);
 
 /**
+ * Event flow id value
+ *
+ * Returns the flow id value set in the event.
+ * Usage of flow id enables scheduler to maintain multiple synchronization
+ * contexts per single queue. For example, when multiple flows are assigned to
+ * an atomic queue, events of a single flow (events from the same queue with
+ * the same flow id value) are guaranteed to be processed by only single thread
+ * at a time.  For packets received through packet input initial
+ * event flow id will be same as flow hash generated for packets. The hash
+ * algorithm and therefore the resulting flow id value is implementation
+ * specific. Use pktio API configuration options to select the fields used for
+ * initial flow id calculation. For all other events initial flow id is zero
+ * An application can change event flow id using odp_event_flow_id_set().
+ *
+ * @param	event	Event handle
+ *
+ * @return		Flow id of the event
+ *
+ */
+uint32_t odp_event_flow_id(odp_event_t event);
+
+/**
+ * Set event flow id value
+ *
+ * Store the event flow id for the event and sets the flow id flag.
+ * When scheduler is configured as flow aware, scheduled queue synchronization
+ * will be based on this id within each queue.
+ * When scheduler is configured as flow unaware, event flow id is ignored by
+ * the implementation.
+ * The value of flow id must be less than the number of flows configured in the
+ * scheduler.
+ *
+ * @param      event		Event handle
+ * @param      flow_id          Flow event id to be set.
+ */
+void odp_event_flow_id_set(odp_event_t event, uint32_t flow_id);
+
+/**
  * @}
  */
 
--- a/include/odp/api/spec/queue_types.h
+++ b/include/odp/api/spec/queue_types.h
@@ -139,7 +139,8 @@ typedef struct odp_queue_capability_t {
 	 * instead */
 	unsigned int ODP_DEPRECATE(max_sched_groups);
 
-	/** @deprecated Use prios field of odp_schedule_capability_t instead */
+	/** @deprecated Use max_prios field of odp_schedule_capability_t
+	 * instead */
 	unsigned int ODP_DEPRECATE(sched_prios);
 
 	/** Plain queue capabilities */
@@ -182,7 +183,8 @@ typedef struct odp_queue_capability_t {
 
 	} plain;
 
-	/** Scheduled queue capabilities */
+	/** @deprecated Use queue capabilities in odp_schedule_capability_t
+	 * instead */
 	struct {
 		/** Maximum number of scheduled (ODP_BLOCKING) queues of the
 		  * default size. */
@@ -220,7 +222,7 @@ typedef struct odp_queue_capability_t {
 
 		} waitfree;
 
-	} sched;
+	} ODP_DEPRECATE(sched);
 
 } odp_queue_capability_t;
 
--- a/include/odp/api/spec/queue_types.h
+++ b/include/odp/api/spec/queue_types.h
@@ -19,6 +19,7 @@ extern "C" {
 #endif
 
 #include <odp/api/schedule_types.h>
+#include <odp/api/deprecated.h>
 
 /** @addtogroup odp_queue
  *  @{
@@ -130,14 +131,16 @@ typedef struct odp_queue_capability_t {
 	  * types are used simultaneously. */
 	uint32_t max_queues;
 
-	/** Maximum number of ordered locks per queue */
-	uint32_t max_ordered_locks;
+	/** @deprecated Use max_ordered_locks field of
+	 * odp_schedule_capability_t instead */
+	uint32_t ODP_DEPRECATE(max_ordered_locks);
 
-	/** Maximum number of scheduling groups */
-	unsigned max_sched_groups;
+	/** @deprecated Use max_groups field of odp_schedule_capability_t
+	 * instead */
+	unsigned int ODP_DEPRECATE(max_sched_groups);
 
-	/** Number of scheduling priorities */
-	unsigned sched_prios;
+	/** @deprecated Use prios field of odp_schedule_capability_t instead */
+	unsigned int ODP_DEPRECATE(sched_prios);
 
 	/** Plain queue capabilities */
 	struct {
--- a/include/odp/api/spec/schedule.h
+++ b/include/odp/api/spec/schedule.h
@@ -258,6 +258,44 @@ int odp_schedule_default_prio(void);
 int odp_schedule_num_prio(void);
 
 /**
+ * Initialize schedule configuration options
+ *
+ * Initialize an odp_schedule_config_t to its default values.
+ *
+ * @param[out] config  Pointer to schedule configuration structure
+ */
+void odp_schedule_config_init(odp_schedule_config_t *config);
+
+/**
+ * Global schedule configuration
+ *
+ * Initialize and configure scheduler with global configuration options
+ * to schedule events across different scheduled queues.
+ * This function must be called before scheduler is used (any other scheduler
+ * function is called except odp_schedule_capability() and
+ * odp_schedule_config_init()) or any queues are created (by application itself
+ * or by other ODP modules).
+ * An application can pass NULL value to use default configuration. It will
+ * have the same result as filling the structure with
+ * odp_schedule_config_init() and then passing it to odp_schedule_config().
+ *
+ * The initialization sequeunce should be,
+ * odp_schedule_capability()
+ * odp_schedule_config_init()
+ * odp_schedule_config()
+ * odp_schedule()
+ *
+ * @param config   Pointer to scheduler configuration structure or NULL for the
+ *                 default configuration
+ *
+ * @retval 0 on success
+ * @retval <0 on failure
+ *
+ * @see odp_schedule_capability(), odp_schedule_config_init()
+ */
+int odp_schedule_config(const odp_schedule_config_t *config);
+
+/**
  * Query scheduler capabilities
  *
  * Outputs schedule capabilities on success.
--- a/include/odp/api/spec/schedule.h
+++ b/include/odp/api/spec/schedule.h
@@ -258,6 +258,18 @@ int odp_schedule_default_prio(void);
 int odp_schedule_num_prio(void);
 
 /**
+ * Query scheduler capabilities
+ *
+ * Outputs schedule capabilities on success.
+ *
+ * @param[out] capa   Pointer to capability structure for output
+ *
+ * @retval 0 on success
+ * @retval <0 on failure
+ */
+int odp_schedule_capability(odp_schedule_capability_t *capa);
+
+/**
  * Schedule group create
  *
  * Creates a schedule group with the thread mask. Only threads in the
--- a/include/odp/api/spec/schedule_types.h
+++ b/include/odp/api/spec/schedule_types.h
@@ -201,6 +201,25 @@ typedef struct odp_schedule_capability_t {
 } odp_schedule_capability_t;
 
 /**
+ * Schedule configuration
+ */
+typedef struct odp_schedule_config_t {
+	/** Maximum number of scheduled queues to be supported.
+	 *
+	 * @see odp_schedule_capability_t
+	 */
+	uint32_t num_queues;
+
+	/** Maximum number of events required to be stored simultaneously in
+	 * scheduled queue. This number must not exceed 'max_queue_size'
+	 * capability.  A value of 0 configures default queue size supported by
+	 * the implementation.
+	 */
+	uint32_t queue_size;
+
+} odp_schedule_config_t;
+
+/**
  * @}
  */
 
--- a/include/odp/api/spec/schedule_types.h
+++ b/include/odp/api/spec/schedule_types.h
@@ -166,6 +166,21 @@ typedef	struct odp_schedule_param_t {
 } odp_schedule_param_t;
 
 /**
+ * Scheduler capabilities
+ */
+typedef struct odp_schedule_capability_t {
+	/** Maximum number of ordered locks per queue */
+	uint32_t max_ordered_locks;
+
+	/** Maximum number of scheduling groups */
+	uint32_t max_groups;
+
+	/** Number of scheduling priorities */
+	uint32_t max_prios;
+
+} odp_schedule_capability_t;
+
+/**
  * @}
  */
 
--- a/include/odp/api/spec/schedule_types.h
+++ b/include/odp/api/spec/schedule_types.h
@@ -78,6 +78,9 @@ extern "C" {
  * requests another event from the scheduler, which implicitly releases the
  * context. User may allow the scheduler to release the context earlier than
  * that by calling odp_schedule_release_atomic().
+ * When scheduler is enabled as flow-aware, the event flow id value affects
+ * scheduling of the event and synchronization is maintained per flow within
+ * each queue.
  */
 
 /**
@@ -104,6 +107,9 @@ extern "C" {
  * (e.g. freed or stored) within the context are considered missing from
  * reordering and are skipped at this time (but can be ordered again within
  * another context).
+ * When scheduler is enabled as flow-aware, the event flow id value affects
+ * scheduling of the event and synchronization is maintained per flow within
+ * each queue.
  */
 
 /**
@@ -190,6 +196,13 @@ typedef struct odp_schedule_capability_t {
 	 * events. */
 	uint32_t max_queue_size;
 
+	/** Maximum supported flows per queue.
+	 * Specifies the maximum number of flows per queue supported by the
+	 * implementation.  A value of 0 indicates flow aware mode is not
+	 * supported.
+	 */
+	uint32_t max_flows;
+
 	/** Lock-free (ODP_NONBLOCKING_LF) queues support.
 	 * The specification is the same as for the blocking implementation. */
 	odp_support_t lockfree_queues;
@@ -216,6 +229,22 @@ typedef struct odp_schedule_config_t {
 	 * the implementation.
 	 */
 	uint32_t queue_size;
+
+	/** Number of flows per queue to be supported.  Scheduler enables flow
+	 * aware mode when flow count is configured greater than 1 (up to
+	 * 'max_flows' capability).
+	 *
+	 * Flows are lightweight entities and events can be assigned to
+	 * specific flows by the application using odp_event_flow_id_set()
+	 * before enqueuing the event into the scheduler. This value is ignored
+	 * unless scheduler supports flow aware mode.
+	 *
+	 * This number should be less than maximum flow supported by the
+	 * implementation. The default value is zero.
+	 *
+	 * @see odp_schedule_capability_t
+	 */
+	uint32_t num_flows;
 
 } odp_schedule_config_t;
 
--- a/include/odp/api/spec/schedule_types.h
+++ b/include/odp/api/spec/schedule_types.h
@@ -14,6 +14,8 @@
 #define ODP_API_SPEC_SCHEDULE_TYPES_H_
 #include <odp/visibility_begin.h>
 
+#include <odp/api/support.h>
+
 #ifdef __cplusplus
 extern "C" {
 #endif
@@ -177,6 +179,24 @@ typedef struct odp_schedule_capability_t {
 
 	/** Number of scheduling priorities */
 	uint32_t max_prios;
+
+	/** Maximum number of scheduled (ODP_BLOCKING) queues of the default
+	 * size. */
+	uint32_t max_queues;
+
+	/** Maximum number of events a scheduled (ODP_BLOCKING) queue can store
+	 * simultaneously. The value of zero means that scheduled queues do not
+	 * have a size limit, but a single queue can store all available
+	 * events. */
+	uint32_t max_queue_size;
+
+	/** Lock-free (ODP_NONBLOCKING_LF) queues support.
+	 * The specification is the same as for the blocking implementation. */
+	odp_support_t lockfree_queues;
+
+	/** Wait-free (ODP_NONBLOCKING_WF) queues support.
+	 * The specification is the same as for the blocking implementation. */
+	odp_support_t waitfree_queues;
 
 } odp_schedule_capability_t;
 
--- a/platform/linux-generic/odp_schedule_basic.c
+++ b/platform/linux-generic/odp_schedule_basic.c
@@ -1557,6 +1557,8 @@ static int schedule_capability(odp_schedule_capability_t *capa)
 	capa->max_ordered_locks = schedule_max_ordered_locks();
 	capa->max_groups = schedule_num_grps();
 	capa->max_prios = schedule_num_prio();
+	capa->max_queues = ODP_CONFIG_QUEUES - NUM_INTERNAL_QUEUES;
+	capa->max_queue_size = queue_glb->config.max_queue_size;
 
 	return 0;
 }
--- a/platform/linux-generic/odp_schedule_basic.c
+++ b/platform/linux-generic/odp_schedule_basic.c
@@ -598,6 +598,8 @@ static int schedule_init_queue(uint32_t queue_index,
 	int i;
 	int prio = prio_level_from_api(sched_param->prio);
 
+	ODP_ASSERT(_odp_schedule_configured);
+
 	pri_set_queue(queue_index, prio);
 	sched->queue[queue_index].grp  = sched_param->group;
 	sched->queue[queue_index].prio = prio;
@@ -794,6 +796,19 @@ static int schedule_term_local(void)
 		schedule_release_atomic();
 	else if (sched_local.sync_ctx == ODP_SCHED_SYNC_ORDERED)
 		schedule_release_ordered();
+
+	return 0;
+}
+
+static void schedule_config_init(odp_schedule_config_t *config)
+{
+	config->num_queues = ODP_CONFIG_QUEUES - NUM_INTERNAL_QUEUES;
+	config->queue_size = queue_glb->config.max_queue_size;
+}
+
+static int schedule_config(const odp_schedule_config_t *config)
+{
+	(void)config;
 
 	return 0;
 }
@@ -1587,6 +1602,8 @@ const schedule_fn_t schedule_basic_fn = {
 const schedule_api_t schedule_basic_api = {
 	.schedule_wait_time       = schedule_wait_time,
 	.schedule_capability      = schedule_capability,
+	.schedule_config_init     = schedule_config_init,
+	.schedule_config          = schedule_config,
 	.schedule                 = schedule,
 	.schedule_multi           = schedule_multi,
 	.schedule_multi_wait      = schedule_multi_wait,
--- a/platform/linux-generic/odp_schedule_basic.c
+++ b/platform/linux-generic/odp_schedule_basic.c
@@ -1545,7 +1545,7 @@ static int schedule_num_grps(void)
 	return NUM_SCHED_GRPS;
 }
 
-static void schedule_config(schedule_config_t *config)
+static void schedule_get_config(schedule_config_t *config)
 {
 	*config = *(&sched->config_if);
 };
@@ -1580,7 +1580,7 @@ const schedule_fn_t schedule_basic_fn = {
 	.order_lock = order_lock,
 	.order_unlock = order_unlock,
 	.max_ordered_locks = schedule_max_ordered_locks,
-	.config = schedule_config
+	.get_config = schedule_get_config
 };
 
 /* Fill in scheduler API calls */
--- a/platform/linux-generic/odp_schedule_basic.c
+++ b/platform/linux-generic/odp_schedule_basic.c
@@ -1548,6 +1548,17 @@ static int schedule_num_grps(void)
 static void schedule_config(schedule_config_t *config)
 {
 	*config = *(&sched->config_if);
+};
+
+static int schedule_capability(odp_schedule_capability_t *capa)
+{
+	memset(capa, 0, sizeof(odp_schedule_capability_t));
+
+	capa->max_ordered_locks = schedule_max_ordered_locks();
+	capa->max_groups = schedule_num_grps();
+	capa->max_prios = schedule_num_prio();
+
+	return 0;
 }
 
 /* Fill in scheduler interface */
@@ -1573,6 +1584,7 @@ const schedule_fn_t schedule_basic_fn = {
 /* Fill in scheduler API calls */
 const schedule_api_t schedule_basic_api = {
 	.schedule_wait_time       = schedule_wait_time,
+	.schedule_capability      = schedule_capability,
 	.schedule                 = schedule,
 	.schedule_multi           = schedule_multi,
 	.schedule_multi_wait      = schedule_multi_wait,
--- a/platform/linux-generic/odp_schedule_if.c
+++ b/platform/linux-generic/odp_schedule_if.c
@@ -25,6 +25,10 @@ extern const schedule_api_t schedule_scalable_api;
 const schedule_fn_t *sched_fn;
 const schedule_api_t *sched_api;
 
+#ifdef ODP_DEBUG
+int _odp_schedule_configured;
+#endif
+
 uint64_t odp_schedule_wait_time(uint64_t ns)
 {
 	return sched_api->schedule_wait_time(ns);
@@ -35,14 +39,46 @@ int odp_schedule_capability(odp_schedule_capability_t *capa)
 	return sched_api->schedule_capability(capa);
 }
 
+void odp_schedule_config_init(odp_schedule_config_t *config)
+{
+	memset(config, 0, sizeof(*config));
+
+	sched_api->schedule_config_init(config);
+}
+
+int odp_schedule_config(const odp_schedule_config_t *config)
+{
+	int ret;
+	odp_schedule_config_t defconfig;
+
+	ODP_ASSERT(!_odp_schedule_configured);
+
+	if (!config) {
+		odp_schedule_config_init(&defconfig);
+		config = &defconfig;
+	}
+
+	ret = sched_api->schedule_config(config);
+#ifdef ODP_DEBUG
+	if (ret >= 0)
+		_odp_schedule_configured = 1;
+#endif
+
+	return ret;
+}
+
 odp_event_t odp_schedule(odp_queue_t *from, uint64_t wait)
 {
+	ODP_ASSERT(_odp_schedule_configured);
+
 	return sched_api->schedule(from, wait);
 }
 
 int odp_schedule_multi(odp_queue_t *from, uint64_t wait, odp_event_t events[],
 		       int num)
 {
+	ODP_ASSERT(_odp_schedule_configured);
+
 	return sched_api->schedule_multi(from, wait, events, num);
 }
 
--- a/platform/linux-generic/odp_schedule_if.c
+++ b/platform/linux-generic/odp_schedule_if.c
@@ -30,6 +30,11 @@ uint64_t odp_schedule_wait_time(uint64_t ns)
 	return sched_api->schedule_wait_time(ns);
 }
 
+int odp_schedule_capability(odp_schedule_capability_t *capa)
+{
+	return sched_api->schedule_capability(capa);
+}
+
 odp_event_t odp_schedule(odp_queue_t *from, uint64_t wait)
 {
 	return sched_api->schedule(from, wait);
--- a/platform/linux-generic/odp_schedule_scalable.c
+++ b/platform/linux-generic/odp_schedule_scalable.c
@@ -2113,6 +2113,17 @@ static uint32_t schedule_max_ordered_locks(void)
 	return CONFIG_QUEUE_MAX_ORD_LOCKS;
 }
 
+static int schedule_capability(odp_schedule_capability_t *capa)
+{
+	memset(capa, 0, sizeof(odp_schedule_capability_t));
+
+	capa->max_ordered_locks = schedule_max_ordered_locks();
+	capa->max_groups = num_grps();
+	capa->max_prios = schedule_num_prio();
+
+	return 0;
+}
+
 const schedule_fn_t schedule_scalable_fn = {
 	.pktio_start	= pktio_start,
 	.thr_add	= thr_add,
@@ -2133,6 +2144,7 @@ const schedule_fn_t schedule_scalable_fn = {
 
 const schedule_api_t schedule_scalable_api = {
 	.schedule_wait_time		= schedule_wait_time,
+	.schedule_capability            = schedule_capability,
 	.schedule			= schedule,
 	.schedule_multi			= schedule_multi,
 	.schedule_multi_wait            = schedule_multi_wait,
--- a/platform/linux-generic/odp_schedule_scalable.c
+++ b/platform/linux-generic/odp_schedule_scalable.c
@@ -2120,6 +2120,8 @@ static int schedule_capability(odp_schedule_capability_t *capa)
 	capa->max_ordered_locks = schedule_max_ordered_locks();
 	capa->max_groups = num_grps();
 	capa->max_prios = schedule_num_prio();
+	capa->max_queues = ODP_CONFIG_QUEUES - NUM_INTERNAL_QUEUES;
+	capa->max_queue_size = 0;
 
 	return 0;
 }
--- a/platform/linux-generic/odp_schedule_scalable.c
+++ b/platform/linux-generic/odp_schedule_scalable.c
@@ -2000,6 +2000,19 @@ static int schedule_term_local(void)
 	return rc;
 }
 
+static void schedule_config_init(odp_schedule_config_t *config)
+{
+	config->num_queues = ODP_CONFIG_QUEUES - NUM_INTERNAL_QUEUES;
+	config->queue_size = 0; /* FIXME ? */
+}
+
+static int schedule_config(const odp_schedule_config_t *config)
+{
+	(void)config;
+
+	return 0;
+}
+
 static int num_grps(void)
 {
 	return MAX_SCHED_GROUP;
@@ -2147,6 +2160,8 @@ const schedule_fn_t schedule_scalable_fn = {
 const schedule_api_t schedule_scalable_api = {
 	.schedule_wait_time		= schedule_wait_time,
 	.schedule_capability            = schedule_capability,
+	.schedule_config_init		= schedule_config_init,
+	.schedule_config		= schedule_config,
 	.schedule			= schedule,
 	.schedule_multi			= schedule_multi,
 	.schedule_multi_wait            = schedule_multi_wait,
--- a/test/common/odp_cunit_common.c
+++ b/test/common/odp_cunit_common.c
@@ -87,6 +87,10 @@ static int tests_global_init(odp_instance_t *inst)
 		fprintf(stderr, "error: odp_init_local() failed.\n");
 		return -1;
 	}
+	if (0 != odp_schedule_config(NULL)) {
+		fprintf(stderr, "error: odp_schedule_config(NULL) failed.\n");
+		return -1;
+	}
 
 	return 0;
 }
--- a/test/performance/odp_cpu_bench.c
+++ b/test/performance/odp_cpu_bench.c
@@ -528,7 +528,7 @@ int main(int argc, char *argv[])
 	odp_cpumask_t cpumask;
 	odp_pool_capability_t pool_capa;
 	odp_pool_t pool;
-	odp_queue_capability_t queue_capa;
+	odp_schedule_capability_t schedule_capa;
 	odp_shm_t shm;
 	odp_shm_t lookup_tbl_shm;
 	odp_pool_param_t params;
@@ -607,27 +607,26 @@ int main(int argc, char *argv[])
 	printf("first CPU:          %i\n", odp_cpumask_first(&cpumask));
 	printf("cpu mask:           %s\n", cpumaskstr);
 
-	/* Create application queues */
-	if (odp_queue_capability(&queue_capa)) {
-		LOG_ERR("Error: odp_queue_capability() failed\n");
-		exit(EXIT_FAILURE);
+	if (odp_schedule_capability(&schedule_capa)) {
+		printf("Error: Schedule capa failed.\n");
+		return -1;
 	}
 
 	/* Make sure a single queue can store all the packets in a group */
 	pkts_per_group = QUEUES_PER_GROUP * PKTS_PER_QUEUE;
-	if (queue_capa.sched.max_size  &&
-	    queue_capa.sched.max_size < pkts_per_group)
-		pkts_per_group = queue_capa.sched.max_size;
+	if (schedule_capa.max_queue_size  &&
+	    schedule_capa.max_queue_size < pkts_per_group)
+		pkts_per_group = schedule_capa.max_queue_size;
 
 	/* Divide queues evenly into groups */
-	if (queue_capa.sched.max_num < QUEUES_PER_GROUP) {
+	if (schedule_capa.max_queues < QUEUES_PER_GROUP) {
 		LOG_ERR("Error: min %d queues required\n", QUEUES_PER_GROUP);
 		return -1;
 	}
-	num_queues = num_workers > queue_capa.sched.max_num ?
-			queue_capa.sched.max_num : num_workers;
+	num_queues = num_workers > schedule_capa.max_queues ?
+			schedule_capa.max_queues : num_workers;
 	num_groups = (num_queues + QUEUES_PER_GROUP - 1) / QUEUES_PER_GROUP;
-	if (num_groups * QUEUES_PER_GROUP > queue_capa.sched.max_num)
+	if (num_groups * QUEUES_PER_GROUP > schedule_capa.max_queues)
 		num_groups--;
 	num_queues = num_groups * QUEUES_PER_GROUP;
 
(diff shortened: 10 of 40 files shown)


include/odp/api/spec/event.h
```diff
@@ -209,6 +209,45 @@ void odp_event_free_multi(const odp_event_t event[], int num);
...
+
+/**
+ * Set event flow hash value
+ *
+ * Store the event flow hash for the event and sets the flow hash flag.
+ * When scheduler is configured as flow aware, schedule queue synchronization
+ * will be based on flow within each queue.
+ * When scheduler is configured as flow unaware, event flow hash is ignored by
+ * the implementation.
+ * The value of flow hash should not be greater than the max flow count
```
Bill-Fischofer-Linaro: This comment also supports calling this a flow id since hashes don't have an associated ordering. An id, by contrast, can be constrained to fall within a given range of valid values.

include/odp/api/spec/event.h
```diff
@@ -209,6 +209,45 @@ void odp_event_free_multi(const odp_event_t event[], int num);
...
+ * supported by the implementation.
+ *
+ * @param      event		Event handle
+ * @param      flow_hash        Hash value to set
+ *
+ * @note When scheduler is configured as flow unaware, overwriting the platform
+ * provided value doesn't change how the platform handles this packet after it.
```
Bill-Fischofer-Linaro: This note seem redundant since it's already covered in the descriptive text.

include/odp/api/spec/schedule.h
```diff
@@ -45,6 +45,110 @@ extern "C" {
...
+	/** Number of flows per queue to be supported
+	 * Scheduler enables flow aware mode when flow count is configured
+	 * greater than 1.
+	 *
+	 * Flows are lightweight entities and packets can be assigned to
+	 * specific flows by the application using odp_event_flow_hash_set()
+	 * before enqueuing the packet into the scheduler. This value is
+	 * ignored unless scheduler supports flow aware mode.
+	 *
+	 * Depening on the implementation this number might be rounded up to
```
Bill-Fischofer-Linaro: Typo `Depending`

include/odp/api/spec/schedule.h
```diff
@@ -45,6 +45,110 @@ extern "C" {
...
+	 * nearest supported value (e.g power of 2). odp_schedule_config() will
+	 * update passed config with the value actually used. This number
+	 * should be less than maximum flow supported by the implementation.
+	 *
+	 * @see odp_schedule_capability_t
+	 */
+	uint32_t flow_count;
```
Bill-Fischofer-Linaro: Again, this argues that we're dealing with flow ids (not hashes) that are in the range 0..`flow_count`-1.

include/odp/api/spec/schedule.h
```diff
@@ -45,6 +45,110 @@ extern "C" {
...
+	 * implementation.
+	 * A value of 0 indicates flow aware mode is not supported.
+	 */
+	uint32_t max_flow_count;
+
+	/** Maximum supported schedule queues
+	 * Specifies the maximum number of schedule queues supported by the
+	 * implementation.
+	 */
+	uint32_t max_queue_count;
```
Bill-Fischofer-Linaro: Do we need some convention for implementations that don't have a fixed upper limit but are limited only by available memory? Putting `UINT32_MAX` here might not imply that an application could create that many queues.

include/odp/api/spec/schedule.h
```diff
@@ -45,6 +45,110 @@ extern "C" {
...
+	 * A value of 0 indicates the implementation does not restrict the
+	 * queue size
+	 */
+	uint32_t max_queue_size;
+
+	/** Maximum schedule group
+	 * Specifies the maximum number of schedule group supported by the
+	 * implementation.
+	 */
+	uint32_t max_schedule_group;
```
Bill-Fischofer-Linaro: Does this also need a 0 = no fixed maximum convention?

include/odp/api/spec/schedule.h
```diff
@@ -45,6 +45,110 @@ extern "C" {
...
+/**
+ * Stop scheduler operations
+ *
+ * Stop scheduler module. The application should make sure there are no further
+ * events in the scheduler before calling odp_schedule_stop.
+ *
+ * @retval 0 on success
+ * @retval <0 on failure
+ */
+int odp_schedule_stop(void);
```
Bill-Fischofer-Linaro: I thought we decided against including this as stopping the scheduler cleanly is messy and the need for this API is unclear.

psavol:
I think it makes sense to split this patch into two. The first one introduces config/capa with old features ( queue size, etc) and the second one adds the new flow awareness feature.

include/odp/api/spec/schedule.h
```diff
@@ -45,6 +45,110 @@ extern "C" {
  * Maximum schedule group name length in chars including null char
  */
 
+/**
+ * Schedule configuration
+ * When schedule configuration is not called by the application, the scheduler
+ * behaves in flow-unaware mode and event flow hash values are ignored
+ */
```
psavol: Flow stuff needs to be removed from config_t intro. The same config is/will be used for many things - flow is only one of those.

include/odp/api/spec/schedule.h
psavol: 1) Most APIs use "num" instead of "count". So, this should be "num_flows"
2) Flow ID is on event level, so all references to "packets" should be converted to "events".
3) Maybe valid range needs to specified.
   * Application can set this to 1 ... capa.max_flows
   * Implementation may round up actual num_flows (how / where to read it)
   * Application/pktin/etc uses flow IDs between 0 and actual "num_flows" - 1

include/odp/api/spec/schedule.h
```diff
@@ -45,6 +45,110 @@ extern "C" {
...
+	 * @see odp_schedule_capability_t
+	 */
+	uint32_t flow_count;
+
+	/** Maximum number of schedule queues to be supported
+	 * Application configures the maximum number of schedule queues to be
+	 * supported by the implementation.
+	 * @see odp_queue_capability_t
+	 */
+	uint32_t queue_count;
```
psavol: 1) count => num: num_queues
2) Maximum number of scheduled queues to be used. (The second sentence is not needed)
3) queue capability three max_nums: blocking, non-blocking lockfree and waitfree. This should contain the same information (so that scheduler capa can be removed from queue API)

include/odp/api/spec/schedule.h
```diff
@@ -45,6 +45,110 @@ extern "C" {
...
+
+	/** Maximum number of events required to be stored simultaneously in
+	 * schedule queue. This number should be less than 'max_queue_size'
+	 * supported by the implementation.
+	 * A value of 0 configures default queue size supported by the
+	 * implementation.
+	 */
+	uint32_t queue_size;
```
psavol: "..should be less than.." => "... must not exceed..." (== max_size is a valid value)

include/odp/api/spec/event.h
```diff
@@ -209,6 +209,45 @@ void odp_event_free_multi(const odp_event_t event[], int num);
...
+ * @return		Hash value
+ *
+ * @note The hash algorithm and the header fields defining the flow (therefore
+ * used for hashing) is platform dependent. The initial event flow hash
+ * generated by the HW will be same for flow hash generated for packets.
+
+ * @note The returned hash is either the platform generated (if any), or if
+ * odp_event_flow_hash_set() were called then the value set there.
+ */
+uint32_t odp_event_flow_hash(odp_event_t event);
```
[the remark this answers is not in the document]
lumag: ack

include/odp/api/spec/event.h
lumag: ack

include/odp/api/spec/event.h
lumag: ack

include/odp/api/spec/schedule.h
lumag: ack

include/odp/api/spec/schedule.h
psavol: Agree. It's better to not add it now. We have already schedule_pause() which can used when a thread exits. Stop would hint that scheduler re-configure is possible, and I'd say it's difficult and rarely supported / needed in a live system.

include/odp/api/spec/schedule.h
```diff
@@ -45,6 +45,110 @@ extern "C" {
...
+	 * queue size
+	 */
+	uint32_t max_queue_size;
+
+	/** Maximum schedule group
+	 * Specifies the maximum number of schedule group supported by the
+	 * implementation.
+	 */
+	uint32_t max_schedule_group;
+} odp_schedule_capability_t;
```
psavol: Information / naming conventions in this type should match those in queue_capa so that scheduler portion of  queue_capa can be deprecated.

include/odp/api/spec/schedule.h
lumag: Hmm, I found no mention of that neither in GitHub comments nor in telco notes.

include/odp/api/spec/schedule.h
lumag: done

include/odp/api/spec/schedule.h
lumag: ack

include/odp/api/spec/schedule.h
lumag: @psavol do you suggest to move all `queue_capa.sched` to scheduler capabilities?

psavol: Actaully all scheduler capabilities:
.max_ordered_locks
.max_groups
.prios
.queue.max_num
.queue.max_size
.queue.lockfree.max_num
...

include/odp/api/spec/schedule.h
```diff
@@ -45,6 +45,94 @@ extern "C" {
...
+ * odp_schedule_capability()
+ * odp_schedule_config_init()
+ * odp_schedule_config()
+ * odp_schedule_start()
+ * odp_schedule()
+ *
+ * @retval 0 on success
+ * @retval <0 on failure
+ */
+int odp_schedule_start(void);
```
Bill-Fischofer-Linaro: I'm still wondering why this API is needed. Aside from backwards compatibility issues (Do all existing apps need to change to add this call if they want to continue using the ODP scheduler? That really seems a negative.) the only reason to have this API is if the idea is the app will launch multiple threads that will immediately call `odp_schedule()` before configuration has completed, at which point these calls would be expected to stall until the (main?) thread calls `odp_schedule_start()`. Such stalling could be complicated to implement safely. If we're not going to allow this behavior, then it's just as easy to say that `odp_schedule_config()` should be called prior to any `odp_schedule()` call, and that's sufficient.

psavol: When application can change config, some parts of scheduler global init needs to move later, but before the first schedule() call. So, a new step is needed on every application either it is schedule_config(NULL) or schedule_start(). I think it's more intuitive to keep config call optional and require only start() if application maintains the default config. It also enables later adding stop and re-config features if needed.

Bill-Fischofer-Linaro: How is it that the `odp_schedule_config()` call itself doesn't serve this purpose? If it's called, then a non-default set of configuration options is applied. If it isn't then the defaults remain.

What happens if `odp_schedule()` is called before `odp_schedule_config()` or `odp_schedule_start()`? Either we say that's not allowed, in which case there's no need for `odp_schedule_start()` as a sync point, or else the implication is that those calls(s) will stall until some other thread makes them, which is going to impose it's own implementation difficulties.

As far as reconfiguration goes, we already have `odp_schedule_pause()` and `odp_schedule_resume()` and we could consider `odp_schedule_stop()` and `odp_schedule_start()` to be global analogs of those thread-local calls. But note that `odp_schedule_resume()` is only needed to restart scheduling after `odp_schedule_pause()` is called. We do not require threads to call `odp_schedule_resume()` when they first start up. So we could have an `odp_schedule_start()` call that would only be needed after `odp_schedule_stop()`.

lumag:
@psavol I've split this PR into several patche to make it easier to review and implement.

Bill-Fischofer-Linaro:
Looks like this latest split has issues that need to be resolved before Travis can pass.

include/odp/api/spec/queue_types.h
```diff
@@ -180,7 +180,7 @@ typedef struct odp_queue_capability_t {
 
 	} plain;
 
-	/** Scheduled queue capabilities */
+	/** @deprecated Use odp_schedule_capbility_t.queue instead */
```
psavol: Typo: odp_schedule_capability_t

include/odp/api/spec/schedule_types.h
```diff
@@ -214,6 +214,27 @@ typedef struct odp_schedule_capability_t {
...
+/**
+ * Schedule configuration
+ */
+typedef struct odp_schedule_config_t {
+	/** Maximum number of schedule queues to be supported.
+	 * Application configures the maximum number of schedule queues to be
+	 * supported by the implementation.
+	 * @see odp_queue_capability_t
+	 */
+	uint32_t queue_count;
```
psavol: num_queues instead of count

... configures maximum number of scheduled queues required by the application. 

In general, "to be supported" or "to be supported by implementation" should be changed to "required" or "required by application".

include/odp/api/spec/queue_types.h
```diff
@@ -130,14 +131,14 @@ typedef struct odp_queue_capability_t {
 	  * types are used simultaneously. */
 	uint32_t max_queues;
 
-	/** Maximum number of ordered locks per queue */
-	uint32_t max_ordered_locks;
+	/** @deprecated  Use odp_schedule_capbility_t.max_ordered_locks */
```
psavol: typo: _capability_t

Also "_capability_t.foo" is not very readable. I'd prefer:

"Use 'foo' field of odp_schedule_capability_t instead"

include/odp/api/spec/schedule.h
lumag: @Bill-Fischofer-Linaro 
Without `odp_schedule_start() we will need:
- to check during each `odp_schedule()` call that scheduler was not configured/started and start it with default parameters.
- to disallow multiple `odp_schedule_config()` calls which can be used to negotiate number of flows.

lumag:
@Bill-Fischofer-Linaro this PR changes API and deprecates structure parts. Of course we expect it to fail afterwards. I will populate it with implementation changes after agreeing upon API.

Bill-Fischofer-Linaro:
@lumag We can't merge anything to API-NEXT if it breaks the build. Perhaps defer the deprecation until a later phase once the implementation is available? That way we can get the API changes in and then work on the implementation/validation tests separately. Otherwise this is going to have to grow into a very large PR.

include/odp/api/spec/queue_types.h
lumag: ack

include/odp/api/spec/schedule_types.h
lumag: ack

include/odp/api/spec/queue_types.h
lumag: ack

lumag:
@psavol @Bill-Fischofer-Linaro updated PR

platform/linux-generic/odp_schedule_scalable.c
```diff
@@ -1364,6 +1364,17 @@ static uint64_t schedule_wait_time(uint64_t ns)
 	return ns;
 }
 
+static int schedule_capability(odp_schedule_capability_t *capa)
+{
+	memset(capa, 0, sizeof(odp_schedule_capability_t));
+
+	capa->max_ordered_locks = sched_fn->max_ordered_locks();
+	capa->max_groups = sched_fn->num_grps();
+	capa->prios = odp_schedule_num_prio();
```
psavol: To avoid confusion, it's better to call these function directly (not through function pointer). For this file it means direct calls to schedule_max_ordered_locks(), schedule_num_grps() and schedule_num_prio().

platform/linux-generic/odp_schedule_basic.c
```diff
@@ -1249,6 +1249,17 @@ static uint64_t schedule_wait_time(uint64_t ns)
 	return ns;
 }
 
+static int schedule_capability(odp_schedule_capability_t *capa)
+{
+	memset(capa, 0, sizeof(odp_schedule_capability_t));
+
+	capa->max_ordered_locks = sched_fn->max_ordered_locks();
+	capa->max_groups = sched_fn->num_grps();
+	capa->prios = odp_schedule_num_prio();
```
psavol: To avoid confusion, it's better to call these function directly (not through function pointer). For this file it means direct calls to schedule_max_ordered_locks(), schedule_num_grps() and schedule_num_prio().

include/odp/api/spec/schedule_types.h
```diff
@@ -159,6 +165,107 @@ typedef	struct odp_schedule_param_t {
...
+ */
+typedef struct odp_schedule_capability_t {
+	/** Maximum number of ordered locks per queue */
+	uint32_t max_ordered_locks;
+
+	/** Maximum number of scheduling groups */
+	unsigned max_groups;
+
+	/** Number of scheduling priorities */
+	unsigned prios;
```
psavol: It would be a good opportunity to change this to max_prios. So, that the new config struct can have a num_prios field, which application may use to tell how many priorities are actually needed.

Also it would be good to change these "unsigned" types to uint32_t. At least some time back, checkpatch did not allow any more "unsigned" but "unsigned int" should be used. uint32_t is commonly same type as "unsigned", so it should not cause harm to applications if we change to uint32_t.

include/odp/api/spec/event.h
```diff
@@ -209,6 +209,41 @@ void odp_event_free_multi(const odp_event_t event[], int num);
...
+ *
+ * Returns the flow id value set in the event.
+ * Flow id value set in the event is used by scheduler to distribute the
+ * event across different flows. The returned id is either the platform
+ * generated (if any), or if odp_event_flow_id_set() were called then the value
+ * set there.
+ *
+ * @param	event	Event handle
+ *
+ * @return		Hash value
```
psavol: @return Flow id of the event

include/odp/api/spec/event.h
```diff
@@ -209,6 +209,41 @@ void odp_event_free_multi(const odp_event_t event[], int num);
...
+ *
+ * @note The initial event flow id generated by the HW will be same as flow
+ * hash generated for packets. The hash algorithm and the header fields
+ * defining the flow (therefore used for hashing) is platform dependent.
```
psavol: This note text can be moved above (to the specification text proper). Initial flow id is zero when not set. For packets received from packet input, initial flow id value is determined by the flow hash configuration...

So, it's only for input packets that flow id is set by pktin hash (if configured to do so).

include/odp/api/spec/schedule.h
```diff
@@ -187,6 +207,41 @@ void odp_schedule_prefetch(int num);
...
+ * Initialize and configure scheduler with global configuration options.
+ *
+ * @param[in,out] config   Pointer to scheduler configuration structure
+ *
+ * @retval 0 on success
+ * @retval <0 on failure
+ *
+ * @see odp_schedule_capability(), odp_schedule_config_init()
+ */
+int odp_schedule_config(odp_schedule_config_t *config);
```
psavol: I think output param is needed only for implementation to tell if flow id space is power of two (if it rounded it up). Now, application that sets flow id itself would not need to know that (it would just use it's own range). The problem is only with packet input, if implementation calculates hash and sets that as flow id, application may see larger flow ids than it itself configured here.

So, would it be better to have this as input param and add a pktio capability/config to fix how many bits packet input will use as flow id?

include/odp/api/spec/schedule_types.h
```diff
@@ -159,6 +165,107 @@ typedef	struct odp_schedule_param_t {
...
+} odp_schedule_capability_t;
+
+/**
+ * Schedule configuration
+ */
+typedef struct odp_schedule_config_t {
+	/** Maximum number of schedule queues to be supported.
+	 * Application configures the maximum number of scheduled queues
+	 * required by the application.
+	 * @see odp_queue_capability_t
```
psavol: @see odp_schedule_capability_t

include/odp/api/spec/schedule_types.h
```diff
@@ -159,6 +165,107 @@ typedef	struct odp_schedule_param_t {
...
+
+	} queue;
+
+} odp_schedule_capability_t;
+
+/**
+ * Schedule configuration
+ */
+typedef struct odp_schedule_config_t {
+	/** Maximum number of schedule queues to be supported.
```
psavol: scheduled queues

include/odp/api/spec/schedule_types.h
```diff
@@ -159,6 +165,107 @@ typedef	struct odp_schedule_param_t {
...
+	 * Application configures the maximum number of scheduled queues
+	 * required by the application.
+	 * @see odp_queue_capability_t
+	 */
+	uint32_t num_queues;
+
+	/** Maximum number of events required to be stored simultaneously in
+	 * schedule queue. This number must not exceed 'capa.queue.max_size'
```
psavol: scheduled queue

include/odp/api/spec/event.h
```diff
@@ -209,6 +209,41 @@ void odp_event_free_multi(const odp_event_t event[], int num);
...
+ */
+uint32_t odp_event_flow_id(odp_event_t event);
+
+/**
+ * Set event flow id value
+ *
+ * Store the event flow id for the event and sets the flow id flag.
+ * When scheduler is configured as flow aware, schedule queue synchronization
```
psavol: scheduled queue

include/odp/api/spec/schedule_types.h
```diff
@@ -159,6 +165,107 @@ typedef	struct odp_schedule_param_t {
...
+	 * Depending on the implementation this number might be rounded up to
+	 * nearest supported value (e.g power of 2). odp_schedule_config() will
+	 * update passed config with the value actually used. This number
+	 * should be less than maximum flow supported by the implementation.
+	 *
+	 * @see odp_schedule_capability_t
+	 */
+	uint32_t num_flows;
+
+} odp_schedule_config_t;
```
psavol: Maybe "struct { ... } queue" would make sense here also. Config would be closer to capa, and later we can add num_prio, etc and lockfree/waitfree queue params.

Bill-Fischofer-Linaro:
@lumag Based on discussions at today's ARCH call the consensus is that putting in just the API spec changes needed first would be helpful and permit easier review. This means deferring the parameter moves and deprecations until the implementation phase. Your thoughts on this?

include/odp/api/spec/schedule.h
lumag: @psavol also we can just restrict that flow id space should always be of a power of two. What do you think?

lumag:
@Bill-Fischofer-Linaro I think it is counterproductive: we need config and capa, changes are small enough, so there is no reason to split this PR further.

include/odp/api/spec/schedule_types.h
lumag: ack

include/odp/api/spec/schedule_types.h
lumag: ack

Bill-Fischofer-Linaro:
@lumag then the additional changes needed to enable Travis to run need to be part of this PR. We cannot merge this if it breaks the build.

include/odp/api/spec/event.h
lumag: @psavol I've changed the description, could you please take a look.

include/odp/api/spec/schedule.h
psavol: Actually would it work if implementation rounds flow IDs down to power of 2. E.g. max num flow id in capa is 128, application configures and uses itself only 100, implementation rounds down number of packet input flow IDs to 64.

lumag:
@Bill-Fischofer-Linaro build should be fixed now.

include/odp/api/spec/queue_types.h
```diff
@@ -130,14 +131,16 @@ typedef struct odp_queue_capability_t {
 	  * types are used simultaneously. */
 	uint32_t max_queues;
 
-	/** Maximum number of ordered locks per queue */
-	uint32_t max_ordered_locks;
+	/** @deprecated Use max_ordered_locks field of
+	 * odp_schedule_capability_t instead */
+	uint32_t ODP_DEPRECATE(max_ordered_locks);
```
Bill-Fischofer-Linaro: Checkpatch complains about this:
```
---------------------------------------------------------------
0001-api-queue-schedule-move-scheduler-capabilities-to-sc.patch
---------------------------------------------------------------
WARNING: function definition argument 'max_ordered_locks' should also have an identifier name
#37: FILE: include/odp/api/spec/queue_types.h:136:
+	uint32_t ODP_DEPRECATE(max_ordered_locks);
WARNING: Prefer 'unsigned int' to bare use of 'unsigned'
#43: FILE: include/odp/api/spec/queue_types.h:140:
+	unsigned ODP_DEPRECATE(max_sched_groups);
WARNING: function definition argument 'max_sched_groups' should also have an identifier name
#43: FILE: include/odp/api/spec/queue_types.h:140:
+	unsigned ODP_DEPRECATE(max_sched_groups);
WARNING: Prefer 'unsigned int' to bare use of 'unsigned'
#48: FILE: include/odp/api/spec/queue_types.h:143:
+	unsigned ODP_DEPRECATE(sched_prios);
WARNING: function definition argument 'sched_prios' should also have an identifier name
#48: FILE: include/odp/api/spec/queue_types.h:143:
+	unsigned ODP_DEPRECATE(sched_prios);
total: 0 errors, 5 warnings, 0 checks, 68 lines checked
0001-api-queue-schedule-move-scheduler-capabilities-to-sc.patch has style problems, please review.
```

test/performance/odp_sched_perf.c
```diff
@@ -279,19 +279,19 @@ static int create_queues(test_global_t *global)
...
-		printf("Error: Queue capa failed.\n");
+	if (odp_schedule_capability(&schedule_capa)) {
+		printf("Error: Schedule capa failed.\n");
 		return -1;
 	}
 
-	if (tot_queue > queue_capa.sched.max_num) {
-		printf("Max queues supported %u\n", queue_capa.sched.max_num);
+	if (tot_queue > schedule_capa.queue.max_num) {
+		printf("Max queues supported %u\n", schedule_capa.queue.max_num);
```
Bill-Fischofer-Linaro: Checkpatch complains about this:
```
---------------------------------------------------------------
0007-example-tests-move-scheduled-queue-capabilities-to-s.patch
---------------------------------------------------------------
#169: FILE: test/performance/odp_sched_perf.c:288:
+		printf("Max queues supported %u\n", schedule_capa.queue.max_num);
#242: FILE: test/performance/odp_scheduling.c:902:
+		globals->queues_per_prio = schedule_capa.queue.max_num / NUM_PRIOS;
#378: FILE: test/validation/api/scheduler/scheduler.c:1575:
+	if (sched_capa.queue.max_size && sched_capa.queue.max_size < BUFS_PER_QUEUE_EXCL) {
total: 0 errors, 3 warnings, 0 checks, 318 lines checked
0007-example-tests-move-scheduled-queue-capabilities-to-s.patch has style problems, please review.
----------------------------------------------------------
```

include/odp/api/spec/queue_types.h
lumag: @Bill-Fischofer-Linaro checkpatch does not understand `ODP_DEPRECATE()` syntax, thus it thinks that ODP_DEPRECATE is a function with K&R style arguments. Also I don't think we should change types of old fields.

test/performance/odp_sched_perf.c
lumag: Ack. Will update in the next update.

include/odp/api/spec/queue_types.h
psavol: "unsigned" is a synonym of "unsigned int", so we can update the type without any one noticing it.

test/performance/odp_sched_perf.c
lumag: fixed

include/odp/api/spec/queue_types.h
lumag: fixed `unsigned` -> `unsigned int`.

include/odp/api/spec/schedule_types.h
psavol: This needs to be odp_schedule_capability_t, not the old queue capa.

include/odp/api/spec/schedule_types.h
```diff
@@ -159,6 +165,109 @@ typedef	struct odp_schedule_param_t {
...
+		 * scheduler. This value is ignored unless scheduler supports
+		 * flow aware mode.
+		 *
+		 * This number should be less than maximum flow supported by
+		 * the implementation.
+		 *
+		 * @see odp_schedule_capability_t
+		 */
+		uint32_t num_flows;
+	} queue;
```
psavol: At least num and size need to be inside the same struct (blocking / LF / WF). Actually, flow support would be also per blocking level, since those may be different implementations.

In practice, HW scheduler may implement simultaneously all three blocking levels, but a SW scheduler only the first. So, it think it's fine that capa can tell if e.g. LF scheduler is available but extra config for it is not provided.

include/odp/api/spec/schedule.h
```diff
@@ -187,6 +187,51 @@ void odp_schedule_prefetch(int num);
...
+ * an application itself or any other ODP module creates any schedued queue.
+ * An application can pass NULL value to use default configuration.
+ *
+ * The initialization sequeunce should be,
+ * odp_schedule_capability()
+ * odp_schedule_config_init()
+ * odp_schedule_config()
+ * odp_schedule()
+ *
+ * @param config   Pointer to scheduler configuration structure
```
psavol: , or NULL for default configuration.

include/odp/api/spec/schedule.h
```diff
@@ -187,6 +187,51 @@ void odp_schedule_prefetch(int num);
...
+void odp_schedule_config_init(odp_schedule_config_t *config);
+
+/**
+ * Global schedule configuration
+ *
+ * Initialize and configure scheduler with global configuration options
+ * to schedule events across different schedule queues. The scheduler module
+ * should be started before creating any odp queues. This should happen before
+ * an application itself or any other ODP module creates any schedued queue.
+ * An application can pass NULL value to use default configuration.
```
psavol: config must be called before any other schedule API function is called, except capability.

Typo: "schedued queue"

include/odp/api/spec/event.h
```diff
@@ -209,6 +209,40 @@ void odp_event_free_multi(const odp_event_t event[], int num);
  */
 void odp_event_free_sp(const odp_event_t event[], int num);
 
+/**
+ * Event flow id value
+ *
+ * Returns the flow id value set in the event.
+ * Flow id value set in the event is used by scheduler to distribute the event
+ * across different flows. For packets received through packet input initial
```
psavol: This is not accurate specification of flow id usage:
"Flow id value set in the event is used by scheduler to distribute the event across different flows."

Scheduler uses queue + flow id to maintain sync context. So may be replace with this:
"Usage of flow id enables scheduler to maintain multiple synchronization contexts per single queue. For example, when multiple flows are assigned to an atomic queue, events of a single flow (events from the same queue with the same flow id value) are guaranteed to be processed by only single thread at a time."

include/odp/api/spec/event.h
```diff
@@ -209,6 +209,40 @@ void odp_event_free_multi(const odp_event_t event[], int num);
...
+ * event flow id will be same as flow hash generated for packets. The hash
+ * algorithm and the header fields defining the flow (therefore used for
+ * hashing) is platform dependent. For all other events initial flow id is zero
```
psavol: Not accurate: "The hash algorithm and the header fields defining the flow (therefore used for hashing) is platform dependent."

Hash algorithm and thus flow id value is implementation specific, but which header fields that are included into hash calculation are not (pktio config spec includes the options already).

So, replace with e.g. this:
"The hash algorithm and therefore the resulting flow id value is implementation specific. Use pktio API configuration options to select the fields used for initial flow id calculation."

include/odp/api/spec/queue_types.h
```diff
@@ -130,14 +131,16 @@ typedef struct odp_queue_capability_t {
...
-	/** Maximum number of scheduling groups */
-	unsigned max_sched_groups;
+	/** @deprecated Use max_groups field of odp_schedule_capability_t
+	 * instead */
+	unsigned int ODP_DEPRECATE(max_sched_groups);
 
-	/** Number of scheduling priorities */
-	unsigned sched_prios;
+	/** @deprecated Use prios field of odp_schedule_capability_t instead */
+	unsigned int ODP_DEPRECATE(sched_prios);
```
psavol: @deprecated Use **max_prios** field...

include/odp/api/spec/schedule.h
psavol: Still needs update about "config must be called before any other sched call". Also has still the typo.

Here's my suggestion:
"This function must be called before scheduler is used (any other scheduler function is called except odp_schedule_capability() and odp_schedule_config_init()) or any queues are created (by application itself or by  other ODP modules)."

include/odp/api/spec/schedule.h
psavol: NULL value specification is still needed (for @param config): "..., or NULL for default configuration."

include/odp/api/spec/schedule_types.h
```diff
@@ -79,6 +79,9 @@ extern "C" {
  * requests another event from the scheduler, which implicitly releases the
  * context. User may allow the scheduler to release the context earlier than
  * that by calling odp_schedule_release_atomic().
+ * When scheduler is enabled as flow-aware, the event flow hash value affects
+ * scheduling of the event and synchronization is maintained per flow within
+ * each queue.
```
psavol: hash -> id

"... the event flow hash value affects ..." -> "... the event flow id value affects ..."

include/odp/api/spec/schedule_types.h
```diff
@@ -105,6 +108,9 @@ extern "C" {
  * (e.g. freed or stored) within the context are considered missing from
  * reordering and are skipped at this time (but can be ordered again within
  * another context).
+ * When scheduler is enabled as flow-aware, the event flow hash value affects
+ * scheduling of the event and synchronization is maintained per flow within
+ * each queue.
```
psavol: Same thing: flow hash value -> flow id value

include/odp/api/spec/schedule_types.h
```diff
@@ -159,6 +165,109 @@ typedef	struct odp_schedule_param_t {
...
+} odp_schedule_capability_t;
+
+/**
+ * Schedule configuration
+ */
+typedef struct odp_schedule_config_t {
+	/** Maximum number of schedule queues to be supported.
+	 * Application configures the maximum number of scheduled queues
+	 * required by the application.
+	 * @see odp_queue_capability_t
```
psavol: odp_schedule_capability_t, not queue capa

include/odp/api/spec/schedule_types.h
```diff
@@ -159,6 +165,109 @@ typedef	struct odp_schedule_param_t {
...
+
+	} queue;
+
+} odp_schedule_capability_t;
+
+/**
+ * Schedule configuration
+ */
+typedef struct odp_schedule_config_t {
+	/** Maximum number of schedule queues to be supported.
```
psavol: typo: "schedule queues" -> "scheduled queues"

include/odp/api/spec/schedule_types.h
psavol: All three params (num/size/num_flows) should be on the same level. Actually, I think we could simplify capa/config by replacing lockfree/waitfree levels, with just booleans. The reasoning is that scheduler is complex module, and would not include multiple different implementations of scheduled queues. So, either all queues are lockfree or non are. Also all params affect the implementation (so e.g. num_prios would be also a per lockfree thing, etc).

So, "struct queue" can be removed and all config params would be on the same (base) level.

include/odp/api/spec/schedule_types.h
```diff
@@ -159,6 +165,109 @@ typedef	struct odp_schedule_param_t {
...
+		/** Maximum supported flows per queue.
+		 * Specifies the maximum number of flows per queue supported by
+		 * the implementation.  A value of 0 indicates flow aware mode
+		 * is not supported.
+		 */
+		uint32_t max_flows;
+
+	} queue;
+
+} odp_schedule_capability_t;
```
psavol: As per config_t comment: structs queue/lockfree/waitfree can be removed. So, that capa would be simply

typedef struct odp_schedule_capability_t {
uint32_t max_ordered_locks;
uint32_t max_groups;
uint32_t max_prios;
uint32_t max_queues;
uint32_t max_queue_size;
uint32_t max_flows;
odp_bool_t lockfree_queues;
odp_bool_t waitfree_queues;
} odp_schedule_capability_t;

include/odp/api/spec/schedule_types.h
```diff
@@ -159,6 +165,109 @@ typedef	struct odp_schedule_param_t {
...
+		 *
+		 * Flows are lightweight entities and events can be assigned to
+		 * specific flows by the application using
+		 * odp_event_flow_id_set() before enqueuing the event into the
+		 * scheduler. This value is ignored unless scheduler supports
+		 * flow aware mode.
+		 *
+		 * This number should be less than maximum flow supported by
+		 * the implementation.
+		 *
```
psavol: Add: "The default value is zero."

test/validation/api/scheduler/scheduler.c
```diff
@@ -141,6 +142,22 @@ static void release_context(odp_schedule_sync_t sync)
...
+static void scheduler_test_capa(void)
+{
+	odp_schedule_capability_t capa;
+	odp_queue_capability_t queue_capa;
+
+	memset(&capa, 0, sizeof(odp_schedule_capability_t));
+	CU_ASSERT_FATAL(odp_schedule_capability(&capa) == 0);
+	CU_ASSERT_FATAL(odp_queue_capability(&queue_capa) == 0);
+
+	CU_ASSERT(capa.max_ordered_locks != 0);
```
psavol: I think max ordered locks can be 0 == not supported.

include/odp/api/spec/schedule.h
lumag: ok

include/odp/api/spec/schedule_types.h
lumag: I hoped to retain close compatibility between old queue capa and scheduler capa.

test/validation/api/scheduler/scheduler.c
lumag: Hmm, true.

lumag:
I've updated this PR with `odp_schedule_config()` implementation. `odp_schedule_capability_t` change will come in next version.

platform/linux-generic/odp_schedule_if.c
```diff
@@ -25,11 +25,57 @@ extern const schedule_api_t schedule_scalable_api;
...
 	return sched_api->schedule_wait_time(ns);
 }
 
+int odp_schedule_capability(odp_schedule_capability_t *capa)
+{
+	return sched_api->schedule_capability(capa);
+}
+
+#ifdef ODP_DEBUG
+int odp_schedule_config(const odp_schedule_config_t *config)
```
psavol: ODP_DEBUG is always defined. Also it's better not to duplicate functions, but use the define value so that optimizer removes extra debug code:

if (ODP_DEBUG == 1 &&  _odp_schedule_configured)
		ODP_ABORT("Scheduler already configured!\n");

Actually, in this case ODP_ASSERT() would do the job (prints error and aborts).

platform/linux-generic/odp_schedule_if.c
```diff
@@ -121,6 +168,16 @@ void odp_schedule_order_unlock(uint32_t lock_index)
...
 }
 
+void odp_schedule_config_init(odp_schedule_config_t *config)
+{
+	odp_schedule_capability_t capa;
+
+	odp_schedule_capability(&capa);
+
+	memset(config, 0, sizeof(*config));
+	config->num_queues = capa.queue.max_num;
```
psavol: Default config is implementation specific. So, implementation of this function cannot be common to all implementations.

E.g. basic queue (basic and sp schedulers) has default queue size of 4k and max queue size of 8k in config file. So, queue size should be 4k after this call. Capability shows 8k. And user may change both through config file.

queue_basic: {
	# Maximum queue size. Value must be a power of two.
	max_queue_size = 8192

	# Default queue size. Value must be a power of two.
	default_queue_size = 4096
}

test/common/odp_cunit_common.c
```diff
@@ -87,6 +87,10 @@ static int tests_global_init(odp_instance_t *inst)
 		fprintf(stderr, "error: odp_init_local() failed.\n");
 		return -1;
 	}
+	if (0 != odp_schedule_config(NULL)) {
+		fprintf(stderr, "error: odp_schedule_config(NULL) failed.\n");
+		return -1;
+	}
```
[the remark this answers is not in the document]
psavol: Some validation test use scheduler, some don't. It should be ok always to configure (start) scheduler although it's not used.

test/performance/odp_cpu_bench.c
```diff
@@ -607,27 +607,28 @@ int main(int argc, char *argv[])
...
-	/* Create application queues */
-	if (odp_queue_capability(&queue_capa)) {
-		LOG_ERR("Error: odp_queue_capability() failed\n");
-		exit(EXIT_FAILURE);
+	if (odp_schedule_capability(&schedule_capa)) {
+		printf("Error: Schedule capa failed.\n");
+		return -1;
 	}
 
+	odp_schedule_config(NULL);
```
psavol: Here's an example of dilemma between capa and config. Capa tells max limits (e.g. queue size 8k), config(NULL) tells that defaults are ok (e.g. queue size 4k), should next capa call after config show 8k or 4k?

4k is now the configured maximum, queue creates over 4k would fail. Although, it would not be very intuitive if capa call before and after config returns different values - although config call is just there for that, change the limits to what's enough for the application.

On the other hand application changes are minimized if config affect capa and applications do config(NULL) *before* capa.

psavol: Maybe the most logical / easy to use option is to have two capa calls:

// max values for config
odp_schedule_config_capa()

// application configured maximums or defaults
odp_schedule_config()

// run time maximums for e.g. queue_create(). cannot be called before config()
odp_schedule_capa()

psavol: Then old applications would just call config(NULL) before any schedule calls (including schedule_capa), and would use new schedule_capa() just like old queue_capa.

platform/linux-generic/odp_schedule_if.c
lumag: Ack

test/performance/odp_cpu_bench.c
lumag: Well, in all other cases capa returns implementation capabilities, no matter what was configured, doesn't it? In this case also capa returns implementation capabilities. It is a task of an application to know what it has configured and which limitations apply because of this configuration.

Bill-Fischofer-Linaro: If we're treating config is an access function then by ODP convention `odp_schedule_config()` would be the getter and `odp_schedule_config_set()` would be the setter. That way there is no confusion.

psavol: typedef struct odp_schedule_config_t {
  uint32_t num_queues;
  uint32_t queue_size;
  uint32_t num_flows;
} odp_schedule_config_t;

typedef struct odp_schedule_capability_t {
  uint32_t max_ordered_locks;
  uint32_t max_groups;
  uint32_t max_prios;
  uint32_t max_queues;
  uint32_t max_queue_size;
  uint32_t max_flows;
  odp_support_t lockfree_queues;
  odp_support_t waitfree_queues;
} odp_schedule_capability_t


// maximums for the config call
int odp_schedule_config_capa(odp_schedule_capability_t *config_capa);

// set config before any other calls
int odp_schedule_config(const odp_schedule_config_t *config);

// run time capability (affected by config)
int odp_schedule_capability(odp_schedule_capability_t *capa);

How this looks like?
1) xxx_config_capa() would output limits for the xxx_config(). Only xxx_config_capa() and xxx_config_init() can be called before xxx_config().
2) xxx_config() must be called always (may accept NULL == all defaults) and before other calls.
3)  xxx_capability() would output limits for other calls (== maximums of the current config)

lumag: I do not like this approach. I still think that if application uses `odp_schedule_config(NULL)`, it doesn't really care about particular sizes/allowances. If it does, it will fill the config structure as it sees fit. Other approach might be to enforce that max == default.

Bill-Fischofer-Linaro: Having defaults be maximum values might not be reasonable either. An implementation-chosen "optimal" value would typically be the default, with capabilities returning the limits that the implementation can support. The application can then either accept the defaults or configure chosen values within the implementation-defined limits.

So `odp_schedule_capability()` might return min, max, and default values for relevant configurable parameters to give the application a complete picture of its range of options. `odp_schedule_config(NULL)` accepts the defaults while `odp_schedule_config_init()` sets parameters to their default values that the application can then override with a subsequent `odp_schedule_config()` call.

The only question remaining is whether the application can query the current scheduler configuration, and if so, how? Since we already have an `odp_schedule_group_info()` API we might want to use that as a model and simply have add `odp_schedule_config_info()` for that purpose.

psavol: // Max values. Those values not in config_t stay as limits also after config.
// Those values in config_t may change during config_t, use e.g. info_t to check those. 
typedef struct odp_schedule_capability_t {
uint32_t max_ordered_locks;
uint32_t max_groups;
uint32_t max_prios;
uint32_t max_queues;
uint32_t max_queue_size;
uint32_t max_flows;
odp_support_t lockfree_queues;
odp_support_t waitfree_queues;
} odp_schedule_capability_t

// input for _config()
typedef struct odp_schedule_config_t {
uint32_t num_queues;
uint32_t queue_size;
uint32_t num_flows;
} odp_schedule_config_t;

typedef struct odp_schedule_info_t {
uint32_t num_ordered_locks;
uint32_t num_groups;
uint32_t num_prios;
uint32_t num_queues;
uint32_t max_queue_size;
uint32_t num_flows;
bool_t lockfree_queues;
bool_t waitfree_queues;
} odp_schedule_info_t

// Max values for config. Some values are valid over config,
// some not (those included into config_t)
int odp_schedule_capability(odp_schedule_capability_t *capa)

// Set config and start scheduler before any other calls
int odp_schedule_config(const odp_schedule_config_t *config);

// check values of the current config. Some values are equal to other 
// get functions e.g. odp_schedule_num_prio() == info.num_prios
int odp_schedule_info(const odp_schedule_info_t *info);

This is also fine, but it means less trivial changes to current users of queue_capa: 

before:
queue_capa(&capa);
num = capa.sched.max_num

after:
sched_config(NULL);
sched_info(&info);
num = info.num_queues;

example/packet/odp_pktio.c
```diff
@@ -423,6 +425,7 @@ int main(int argc, char *argv[])
 	thr_params.thr_type = ODP_THREAD_WORKER;
 	thr_params.instance = instance;
 
+	/* Config and start scheduler */
```
psavol: hmm. comment left behind?

include/odp/api/spec/schedule_types.h
```diff
@@ -159,6 +167,87 @@ typedef	struct odp_schedule_param_t {
...
+		 * flow aware mode.
+		 *
+		 * This number should be less than maximum flow supported by
+		 * the implementation. The default value is zero.
+		 *
+		 * @see odp_schedule_capability_t
+		 */
+		uint32_t num_flows;
+
+	} queue;
```
psavol: Now when capa is flat, config should be also. So, "struct queue" needs to be removed. Just all three values on the same (base level). Later we'll add num_prios, num_groups, etc also into this config_t.

psavol:
Otherwise looks good, but config_t should be flat and the capa -> config -> info thing needs a resolution. If we go with info_t, then all queue_capa.sched.xxx needs to be changed to config(NULL) + info(), rather than current sched_capa.xxx.

include/odp/api/spec/schedule_types.h
lumag: ok

lumag:
@psavol @Bill-Fischofer-Linaro this PR should now contain all changes as agreed during telcos.

muvarov:
<pre>v16 checks:
checkpatch.pl - OK
</pre>

include/odp/api/spec/schedule_types.h
```diff
@@ -165,6 +173,81 @@ typedef	struct odp_schedule_param_t {
...
+	 * @see odp_schedule_capability_t
+	 */
+	uint32_t num_queues;
+
+	/** Maximum number of events required to be stored simultaneously in
+	 * scheduled queue. This number must not exceed 'capa.queue.max_size'
+	 * required by the application.  A value of 0 configures default queue
+	 * size supported by the implementation.
+	 */
+	uint32_t queue_size;
```
psavol: Broken sentence above:
"This number must not exceed 'capa.queue.max_size' required by the application."

Should be:
"This number must not exceed 'max_queue_size' capability. "

include/odp/api/spec/schedule_types.h
```diff
@@ -165,6 +173,81 @@ typedef	struct odp_schedule_param_t {
...
+
+	/** Number of flows per queue to be supported.  Scheduler enables flow
+	 * aware mode when flow count is configured greater than 1 (up to
+	 * 'capa.queue.max_flows').
```
psavol: ... up to 'max_flows' capability.

The .queue struct is not there any more.

test/validation/api/scheduler/scheduler.c
```diff
@@ -415,10 +432,11 @@ static void scheduler_test_queue_size(void)
...
 
 	CU_ASSERT_FATAL(odp_queue_capability(&queue_capa) == 0);
+	CU_ASSERT_FATAL(odp_schedule_capability(&schedule_capa) == 0);
 	queue_size = TEST_QUEUE_SIZE_NUM_EV;
-	if (queue_capa.sched.max_size &&
-	    queue_size > queue_capa.sched.max_size)
-		queue_size = queue_capa.sched.max_size;
+	if (schedule_capa.max_queue_size &&
+	    queue_size > schedule_capa.max_queue_size)
+		queue_size = schedule_capa.max_queue_size;
```
psavol: After config(NULL), this is a wrong assumption. Config(NULL) sets queue size to default. Since config() is shared over all validation test, easiest fix would be

odp_schedule_config_init(&default_config);

if (default_config.queue_size && queue_size > default_config.queue_size)
    queue_size = default_config.queue_size;

The same fix would be needed for all config(NULL) + comparison to capa.max_queue_size, but at least the validation test for scheduled queue size should be up to date.

Normally, application would do:
if (capa.max_queue_size && queue_size > capa.max_queue_size)
    queue_size = capa.max_queue_size;

config_init(&conf);
conf.queue_size = queue_size;
config(&conf);

psavol:
That's all: two API text fixes, and then correct usage of capa limits vs the default configuration limits. At least validation tests should follow the new spec "by the book". Examples/perf tests are not so critical to update, but will give bad example if config(NULL) assume max capacity.

include/odp/api/spec/schedule_types.h
lumag: ack

test/validation/api/scheduler/scheduler.c
lumag: @psavol Currently scalable scheduler will set `max_queue_size` to 0. How should it be fixed?

lumag:
@psavol I have updated both validation and performance tests to use schedule_config instead of capabilities where applicable. Could you please review?

include/odp/api/spec/schedule_types.h
psavol: Name of the capa field is "max_queue_size" (not max_size).

test/validation/api/scheduler/scheduler.c
psavol: 0 is valid value if queue size is "infinite" == only system memory / pool sizes limit. I'm not sure if scalable has infinite or finite size queues. Anyway, no need to fix it in this set.

psavol:
Otherwise OK, but found one API comment (max_size vs max_queue_size capa) and one place where config.num_queues should be used (odp_scheduling.c).

test/performance/odp_scheduling.c
```diff
@@ -899,10 +900,18 @@ int main(int argc, char *argv[])
...
+
+	odp_schedule_config(NULL);
+
 	globals->queues_per_prio = QUEUES_PER_PRIO;
 	num_queues = globals->queues_per_prio * NUM_PRIOS;
-	if (num_queues > capa.sched.max_num)
-		globals->queues_per_prio = capa.sched.max_num / NUM_PRIOS;
+	if (num_queues > schedule_capa.max_queues)
+		globals->queues_per_prio = schedule_capa.max_queues /
+			NUM_PRIOS;
```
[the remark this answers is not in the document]
lumag: Updated

test/validation/api/scheduler/scheduler.c
lumag: ok

lumag:
@psavol I've updated PR.

test/performance/odp_scheduling.c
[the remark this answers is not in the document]
psavol: I don't see a difference in this v18 ?! These lines should compare against schedule_config.num_queues instead of schedule_capa.max_queues.

lumag:
Fixed now, forgot to commit

lumag:
@psavol thank you!

muvarov:
Merged.

PR #736, opened 2018-10-24
[PATCH v2] Sched prefer ratio config option

psavol:
Change prefer ratio from fixed value to a config file option. Small value is usable when having only few threads (uneven service of queues mapped to different internal schedule queues). Larger values improve performance on with multiple threads (due to better scalability).

### DIFF
--- a/config/odp-linux-generic.conf
+++ b/config/odp-linux-generic.conf
@@ -16,7 +16,7 @@
 
 # Mandatory fields
 odp_implementation = "linux-generic"
-config_file_version = "0.1.1"
+config_file_version = "0.1.2"
 
 # Shared memory options
 shm: {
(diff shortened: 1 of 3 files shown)


config/odp-linux-generic.conf
```diff
@@ -87,6 +87,13 @@ sched_basic: {
 	# value is the number of threads using the scheduler.
 	prio_spread = 4
 
+	# Weight of the preferred scheduler internal queue over other queues.
+	# Each thread prefers one of the scheduler internal queues over others.
+	# This value controls how many times the preferred queue is polled
+	# between polls to other queues. A higher value typically improves
+	# parallelism, but causes uneven service level for low thread counts.
+	prio_spread_weight = 63
```
Bill-Fischofer-Linaro: Should have a better explanation of how this number is specified and used. What's the valid range? What are the different behaviors at the extreme ends of that range? Where does this default value sit in that range?

psavol: This is what is does: "...how many times the preferred queue is polled between polls to other queues". For example, poll the preferred queue 63 times, poll another queue one time, poll the preferred queue 63 times, poll another queue one time, ...

Since this is not API, user is expected to have a bit of knowledge on the implementation (code) before changing the values ... or just find optimal value through trial and error.

I have not copied range into config file documentation since it would become outdated easily. The code have range checks and prints an error (with min/max values) when range is exceeded.

psavol: Updated text for both prio_spread configs on v2. Hope, it's better now.

psavol:
v2: rebase, increment process mode config file version, updated config file comment text

config/odp-linux-generic.conf
Bill-Fischofer-Linaro: A config file should contain sufficient documentation so that an application developer can use it to tune their app without having to read the source code to try to figure that out. If ODP is installed on a system what most developers will have is the binaries, the user guide, and this config file. Those should be sufficient to be able to use ODP.

muvarov:
Merged.

PR #737, opened 2018-10-26
[PATCH v2] Sched automatic worker group config

psavol:
Add config file options to disable unused automatic schedule groups. These are enabled by default as API spec requires it. By default, scheduler polls/checks if there are active queues in both "all" and "worker" groups (for a worker thread). If only "all" groups is used, disabling "worker" group improves performance a bit. The improvement is significant when burst size is small (as there are more schedule() call cycles per event). With a sched_pktio test case (2 stage pipeline) and burst size forced to 1, improvement was 4-7%.

This config can be later added to schedule config API and this implementation is reusable for that (config file option changes to API option, but other code remain the same).

Sched_pktio burst size change was needed to test the performance effect.

This needs to be rebased on top of #736 once it's merged.

### DIFF
--- a/config/odp-linux-generic.conf
+++ b/config/odp-linux-generic.conf
@@ -16,7 +16,7 @@
 
 # Mandatory fields
 odp_implementation = "linux-generic"
-config_file_version = "0.1.2"
+config_file_version = "0.1.3"
 
 # Shared memory options
 shm: {
@@ -114,4 +114,15 @@ sched_basic: {
 	burst_size_default = [ 32,  32,  32,  32,  32, 16,  8, 4]
 	burst_size_max     = [255, 255, 255, 255, 255, 16, 16, 8]
 
+	# Automatically updated schedule groups
+	#
+	# API specification defines that ODP_SCHED_GROUP_ALL,
+	# _WORKER and _CONTROL are updated automatically. These options can be
+	# used to disable these group when not used. Set value to 0 to disable
+	# a group. Performance may improve when unused groups are disabled.
+	group_enable: {
+		all     = 1
+		worker  = 1
+		control = 1
+	}
 }
--- a/platform/linux-generic/include/odp_schedule_if.h
+++ b/platform/linux-generic/include/odp_schedule_if.h
@@ -19,6 +19,15 @@ extern "C" {
 /* Number of ordered locks per queue */
 #define SCHEDULE_ORDERED_LOCKS_PER_QUEUE 2
 
+typedef struct schedule_config_t {
+	struct {
+		int all;
+		int worker;
+		int control;
+	} group_enable;
+
+} schedule_config_t;
+
 typedef void (*schedule_pktio_start_fn_t)(int pktio_index,
 					 int num_in_queue,
 					 int in_queue_idx[],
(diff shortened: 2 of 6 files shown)


test/performance/odp_sched_pktio.c
```diff
@@ -534,6 +537,7 @@ static void print_usage(const char *progname)
 	       "  -i, --interface <name>    Packet IO interfaces (comma-separated, no spaces)\n"
 	       "  -c, --num_cpu <number>    Worker thread count. Default: 1\n"
 	       "  -q, --num_queue <number>  Number of pktio queues. Default: Worker thread count\n"
+	       "  -b, --burst <number>      Maximum number of events requested from scheduler. Default: 32\n"
```
Bill-Fischofer-Linaro: It's not clear that this patch is related to the others in this PR. It seems to have nothing to do with scheduler groups. This seems a reasonable improvement for this test but shouldn't this be its own PR?

config/odp-linux-generic.conf
```diff
@@ -100,4 +100,15 @@ sched_basic: {
 	burst_size_default = [ 32,  32,  32,  32,  32, 16,  8, 4]
 	burst_size_max     = [255, 255, 255, 255, 255, 16, 16, 8]
 
+	# Automatically updated schedule groups
+	#
+	# API specification defines that ODP_SCHED_GROUP_ALL,
+	# _WORKER and _CONTROL are updated automatically. These options can be
+	# used to disable these group when not used. Set value to 0 to disable
+	# a group. Performance may improve when unused groups are disabled.
```
Bill-Fischofer-Linaro: Threads are created as worker or control threads and queues are created in a specific scheduler group (default = `ODP_SCHED_GROUP_ALL`) so I'm not sure I understand the use cases envisioned here. What happens if `all = 0` is specified and `odp_queue_create()` is called to create a `SCHED` queue without an explicit scheduling group, so that by default it's assigned to group `ODP_SCHED_GROUP_ALL`?

Omitting the explicit worker and control groups might make sense if queues are either going to be in the default `all` group or explicitly named groups, but being able to disable the `all` group seems error-prone and it's not obvious under what conditions it would be useful.

platform/linux-generic/include/odp_schedule_if.h
```diff
@@ -44,7 +53,7 @@ typedef void (*schedule_order_unlock_lock_fn_t)(void);
 typedef void (*schedule_order_lock_start_fn_t)(void);
 typedef void (*schedule_order_lock_wait_fn_t)(void);
 typedef uint32_t (*schedule_max_ordered_locks_fn_t)(void);
-typedef void (*schedule_save_context_fn_t)(uint32_t queue_index);
+typedef void (*schedule_config_fn_t)(schedule_config_t *config);
```
Bill-Fischofer-Linaro: Looks like `schedule_save_context_fn_t` is unused, so this deletion of a stale symbol should technically be a separate patch. This patch should just be adding `schedule_config_fn_t`. Minor point and probably not worrying about here.

test/performance/odp_sched_pktio.c
psavol: Cover letter says: "Sched_pktio burst size change was needed to test the performance effect."

config/odp-linux-generic.conf
psavol: If user/application creates its own groups and places all queues in to those groups. "All" groups is not used or needed. User needs to be careful when changing configuration from the default.

platform/linux-generic/include/odp_schedule_if.h
psavol: I think it would be wasteful (for review and git history) to create an extra patch for this removal of dead code while adding a new line next to it.

test/performance/odp_sched_pktio.c
Bill-Fischofer-Linaro: OK

config/odp-linux-generic.conf
Bill-Fischofer-Linaro: Unless the application has complete control over all queue creation (which it does not, since ODP creates queues internally) it seems that allowing the `all` group to be removed is likely to cause problems. Is the performance gain so significant that this is worth it?

muvarov:
Merged.

PR #738, opened 2018-10-29
[PATCH v1] Remove pktin poll old

psavol:
Change SP scheduler to use the same pktin poll function as the basic scheduler. Remove the old pktin poll function.


muvarov:
Merged.

PR #739, opened 2018-10-31
[PATCH API-NEXT v1] api: scheduler priority functions

psavol:
Introduce functions for min/max/default priorities. Functions avoid fixed number of priorities in ABI compat mode. Also a scheduler may be configured with different number of priorities at start up. Macros are not yet deprecated but are turned into function calls.

### DIFF
--- a/include/odp/api/spec/schedule.h
+++ b/include/odp/api/spec/schedule.h
@@ -181,7 +181,48 @@ void odp_schedule_release_ordered(void);
 void odp_schedule_prefetch(int num);
 
 /**
+ * Maximum scheduling priority level
+ *
+ * This is the maximum value that can be set to 'prio' field in
+ * odp_schedule_param_t (e.g. @see odp_queue_create()). Queues with a higher
+ * priority value are served with higher priority than queues with a lower
+ * priority value.
+ *
+ * @return Maximum scheduling priority level
+ */
+int odp_schedule_max_prio(void);
+
+/**
+ * Minimum scheduling priority level
+ *
+ * This is the minimum value that can be set to 'prio' field in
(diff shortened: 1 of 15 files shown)


include/odp/api/spec/schedule.h
```diff
@@ -180,9 +180,50 @@ void odp_schedule_release_ordered(void);
...
+ * Maximum scheduling priority level
+ *
+ * This is the maximum value that can be set to 'prio' field in
+ * odp_schedule_param_t (e.g. @see odp_queue_create()). Queues with a higher
+ * priority value are served with higher priority than queues with a lower
+ * priority value.
+ *
+ * @return Maximum scheduling priority level
+ */
+int odp_schedule_max_prio(void);
```
Bill-Fischofer-Linaro: Since we're introducing an `odp_schedule_capability()` API, wouldn't these be better there rather than having three new APIs?

Bill-Fischofer-Linaro: OK, so this should be clarified as to how these relate to the capability/config APIs. As written this reads like a capability, not as a getter for what was configured.

psavol: There is no scheduler capa/config API yet. So, configuration is implementation specific (config file) currently. Still it enables run time configuration and ABI compatibility already now.

Bill-Fischofer-Linaro: Given the timing, this PR would be better written on top of PR #734 then. Otherwise we're just going to need a follow-on PR to modify this once that PR has been merged.

psavol: There should not be much to modify on this set as the change would be to add e.g. num_prio into capa/config. These return min/max priority regardless how those were configured.

muvarov:
Merged.

PR #740, opened 2018-11-02
[PATCH v3] linuxgen: add dumpconfig utility

muvarov:
it might be useful to package small binary which prints
platform default builtin config file.

### DIFF
--- a/platform/linux-generic/dumpconfig/dumpconfig.c
+++ b/platform/linux-generic/dumpconfig/dumpconfig.c
@@ -1,0 +1,43 @@
+/* Copyright (c) 2018, Linaro Limited
+ * All rights reserved.
+ *
+ * SPDX-License-Identifier:     BSD-3-Clause
+ */
+
+#include <stdio.h>
+#include <stdlib.h>
+#include <odp_libconfig_config.h>
+
+int main(void)
+{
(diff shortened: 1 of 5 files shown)


platform/linux-generic/dumpconfig/dumpconfig.c
```diff
@@ -0,0 +1,18 @@
...
+
+#include <stdio.h>
+#include <odp_libconfig_config.h>
+
+int main(void)
+{
+	unsigned int i;
+
+	for (i = 0; i < sizeof(config_builtin); i++)
+		printf("%c", config_builtin[i]);
```
Bill-Fischofer-Linaro: I'm not sure what this utility does. Is it just printing out the static config file found in `config/odp-linux-generic.conf`?  That doesn't seem terribly useful. A utility that printed out the effective configuration (taking the `ODP_CONFIG_FILE` into account) would be more useful, but this doesn't seem to do that.

muvarov: @Bill-Fischofer-Linaro yes odp apps can run without config. Then builtin config is used. And ODP_CONFIG_FILE can just override parts of default config.

muvarov:
Merged.

PR #742, opened 2018-11-05
[PATCH v8] linux-gen: ishm: use pre-reserved single va memory

matiaselo:
Reserve single VA memory in global init instead of only allocating the virtual address space. This enables reserving shm blocks in process mode after ODP process has been already forked.

V2:
- single_va_size -> single_va_size_kb (Bill & Maxim)

V5:
- Use MAP_POPULATE flag with mmap

V6:
- Fix new pool validation test running out of buffers

V7:
- Increase default single VA size to 128MB

V8:
- Rebase (Maxim)

### DIFF
--- a/config/odp-linux-generic.conf
+++ b/config/odp-linux-generic.conf
@@ -16,7 +16,7 @@
 
 # Mandatory fields
 odp_implementation = "linux-generic"
-config_file_version = "0.1.3"
+config_file_version = "0.1.4"
 
 # Shared memory options
 shm: {
@@ -37,6 +37,9 @@ shm: {
 	# Allocate internal shared memory using a single virtual address space.
 	# Set to 1 to enable using process mode.
 	single_va = 0
+
+ 	# Amount of memory pre-reserved for ODP_SHM_SINGLE_VA usage in kilobytes
+	single_va_size_kb = 131072
 }
 
 # Pool options
--- a/platform/linux-generic/test/process-mode.conf
+++ b/platform/linux-generic/test/process-mode.conf
@@ -1,10 +1,13 @@
 # Mandatory fields
 odp_implementation = "linux-generic"
-config_file_version = "0.1.3"
+config_file_version = "0.1.4"
 
 # Shared memory options
 shm: {
 	# Override default option and allocate internal shms using single
 	# virtual address space.
 	single_va = 1
+
+ 	# Increase the amount of single VA memory
+	single_va_size_kb = 1048576
 }
--- a/.travis.yml
+++ b/.travis.yml
@@ -214,7 +214,7 @@ jobs:
                           - true
                   script:
                           - if [ -z "${DOCKER_NAMESPACE}" ] ; then export DOCKER_NAMESPACE="opendataplane"; fi
-                          - docker run  -i -t -v `pwd`:/odp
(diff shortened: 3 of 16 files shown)


config/odp-linux-generic.conf
```diff
@@ -37,6 +37,9 @@ shm: {
 	# Allocate internal shared memory using a single virtual address space.
 	# Set to 1 to enable using process mode.
 	single_va = 0
+
+ 	# Amount of memory pre-reserved for ODP_SHM_SINGLE_VA usage in bytes
+	single_va_size = 67108864
```
Bill-Fischofer-Linaro: It's better to specify this in units of megabytes rather than bytes both for ease of use and to avoid the false impression that you're offering byte-level granular control. So `single_va_size = 64` would read much cleaner.  

Note that you're specifying 64MB here even though the commit comment says you're lowering the limit to 50MB, which shows how trying to specifying things in bytes vs. megabytes leads to specification errors.

platform/linux-generic/test/process-mode.conf
```diff
@@ -1,10 +1,13 @@
...
+config_file_version = "0.1.4"
 
 # Shared memory options
 shm: {
 	# Override default option and allocate internal shms using single
 	# virtual address space.
 	single_va = 1
+
+ 	# Increase the amount of single VA memory
+	single_va_size = 1073741824
```
Bill-Fischofer-Linaro: `single_va_size = 1024` would reach much easier here (see earlier comment).

config/odp-linux-generic.conf
Bill-Fischofer-Linaro: KB seems too small and artificially granular. Are there really any cases when we'd specify less than 1MB increments of shared memory? It also doesn't help much when specifying memory in the GB range. We could support MB/GB suffixes but that seems overkill on the parsing side.

matiaselo: In case of chip's internal memory, kilobyte level definition may be useful. Since this is a configuration option I'd prioritize configurability over readability.

Kilobytes are fine for linux:
E.g. /proc/meminfo
```
MemTotal:         492836 kB
MemFree:          304820 kB
MemAvailable:     414916 kB
```

Bill-Fischofer-Linaro: Linux's use of kilobytes reflects the legacy of 1990s-vintage small desktop systems as well as older embedded environments. But Linux itself needs several megabytes these days to boot, so it's a bit of an anachronism. 

In the case of ODP, are there any use cases where we'd want/need to specify sub-MB granularity? Especially for this intended use of shared large memory areas for supporting process mode, 1MB is actually a fairly small granule. The issue is what happens when due to typos or carelessness a `single_va_size = 67108872` is specified? Is that going to work wherever odp-linux finds itself running?

muvarov:
discussion call: single_va_size_kb and change memory to kbytes.

muvarov:
<pre>v7 checks:
checkpatch.pl - OK
</pre>

muvarov:
@matiaselo please rebase the latest (main) patch

matiaselo:
@muvarov rebase done

muvarov:
Merged.

PR #744, opened 2018-11-06
[PATCH API-NEXT v3] api: schedule: wait and no_wait

psavol:
Optimized versions of schedule_multi call for most common wait values. Schedule call is one of the most used API calls, so saving couple of if-clauses per every call may improve performance on some platforms. L2fwd with basic scheduler didn't show significant improvement.

I'll add test cases into v2, after we have agreed on the API.

### DIFF
--- a/include/odp/api/spec/schedule.h
+++ b/include/odp/api/spec/schedule.h
@@ -113,6 +113,35 @@ int odp_schedule_multi(odp_queue_t *from, uint64_t wait, odp_event_t events[],
 		       int num);
 
 /**
+ * Schedule, wait for events
+ *
+ * Like odp_schedule_multi(), but waits infinitely for events.
+ *
+ * @param[out] from    Output parameter for the source queue (where the event
+ *                     was dequeued from). Ignored if NULL.
+ * @param[out] events  Event array for output
+ * @param      num     Maximum number of events to output
+ *
+ * @return Number of events outputted (1 ... num)
+ */
+int odp_schedule_multi_wait(odp_queue_t *from, odp_event_t events[], int num);
+
+/**
+ * Schedule, do not wait for events
+ *
+ * Like odp_schedule_multi(), but does not wait for events.
+ *
+ * @param[out] from    Output parameter for the source queue (where the event
+ *                     was dequeued from). Ignored if NULL.
+ * @param[out] events  Event array for output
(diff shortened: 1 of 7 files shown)


include/odp/api/spec/schedule.h
```diff
@@ -112,6 +112,35 @@ odp_event_t odp_schedule(odp_queue_t *from, uint64_t wait);
...
+ * Like odp_schedule_multi(), but waits infinitely for events.
+ *
+ * @param[out] from    Output parameter for the source queue (where the event
+ *                     was dequeued from). Ignored if NULL.
+ * @param[out] events  Event array for output
+ * @param      num     Maximum number of events to output
+ *
+ * @return Number of events outputted (1 ... num)
+ */
+int odp_schedule_multi_wait(odp_queue_t *from, odp_event_t events[], int num);
```
Bill-Fischofer-Linaro: Why just variant forms for `odp_schedule_multi()`? Wouldn't there be similar benefits to having `odp_schedule_wait()` and `odp_schedule_no_wait()`?

psavol: Wouldn't add single versions at least at this time. Single versions are easy for application to create from multi versions:
num = odp_schedule_wait(&queue, &event, 1);

Wait/no_wait are bigger deal for implementation (e.g. prepare for power save (wait) vs. no need to worry about power save (no_wait)), than max number of events.

Bill-Fischofer-Linaro: I thought the whole point was of these was to save an `if` or two in the implementation. It seems that argument would be equally valid for both single and multi-event calls, especially since `odp_schedule()` is a simpler API than `odp_schedule_multi()`, and is the real fast-path use case.

psavol: Multi calls are more fast path oriented than single calls as API call overhead is divided over N events vs single event. So, a performance optimized application would use the multi call.

muvarov:
Proposal to make this functions Power Save.

Bill-Fischofer-Linaro:
@lumag please review this to see if these would cause any implementation concerns/issues for your platforms.

lumag:
@Bill-Fischofer-Linaro there should be no impact on our implemenation.

muvarov:
Merged.

PR #746, opened 2018-11-07
validation: crypto: capability call should not fail

psavol:
Capability call should always succeed. It's possible that no
algorithms are supported, but also then the call should succeed.

### DIFF
--- a/test/validation/api/crypto/odp_crypto_test_inp.c
+++ b/test/validation/api/crypto/odp_crypto_test_inp.c
@@ -647,8 +647,11 @@ static int check_alg_support(odp_cipher_alg_t cipher, odp_auth_alg_t auth)
 {
 	odp_crypto_capability_t capability;
 
-	if (odp_crypto_capability(&capability))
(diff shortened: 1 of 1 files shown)


lumag:

test/validation/api/crypto/odp_crypto_test_inp.c
```diff
@@ -647,8 +647,11 @@ static int check_alg_support(odp_cipher_alg_t cipher, odp_auth_alg_t auth)
 {
 	odp_crypto_capability_t capability;
 
-	if (odp_crypto_capability(&capability))
+	memset(&capability, 0, sizeof(odp_crypto_capability_t));
```
Bill-Fischofer-Linaro: `memset()` is unnecessary here as `odp_crypto_capability()` sets it.

psavol: This code checks which test cases are active. If the call would not set all bits, test results would be harder to understand if sometimes some cases are executed and sometimes not.

PR #747, opened 2018-11-09
[PATCH v2] Classifier shm usage optimization

psavol:
Combine three SHM blocks into one. Modified sysinfo example to print out shm blocks, so that it's easy to see how many blocks ODP uses internally. This is how I found out inefficient SHM usage of CLS.

### DIFF
--- a/platform/linux-generic/odp_classification.c
+++ b/platform/linux-generic/odp_classification.c
@@ -94,6 +94,7 @@ int odp_classification_init_global(void)
 
 	shm = odp_shm_reserve("_odp_cls_global", sizeof(cls_global_t),
 			      ODP_CACHE_LINE_SIZE, 0);
+
 	if (shm == ODP_SHM_INVALID)
 		return -1;
 


platform/linux-generic/odp_classification.c
```diff
@@ -80,102 +90,46 @@ pmr_t *get_pmr_entry_internal(odp_pmr_t pmr)
...
-	odp_shm_t queue_grp_shm;
+	odp_shm_t shm;
 	int i;
 
-	cos_shm = odp_shm_reserve("_odp_shm_odp_cos_tbl",
-				  sizeof(cos_tbl_t),
-				  sizeof(cos_t),
-				  _ODP_SHM_NO_HP);
+	shm = odp_shm_reserve("_odp_cls_global", sizeof(cls_global_t),
+			      ODP_CACHE_LINE_SIZE, 0);
```
muvarov: that is small mapping, why did you remove _ODP_SHM_NO_HP here?

psavol: Matias will send shortly PR which removes/replaces the flag and changes HP mapping. We did run into latency problems with the current NO_HP mappings. The PR will suggest to use normal pages for <64kB mappings, and HP for larger ones.

Also another aspect to this is number of page table entries used. The smaller the pages size the more page table entries are used, and it's more likely to run into TLB misses (which hurt performance and latency). 

This struct is about 150kB (38x4kB pages vs one 2MB page).

muvarov:
<pre>v1 checks:
checkpatch.pl - OK
</pre>

psavol:
v2: rebased

muvarov:
Merged.

PR #748, opened 2018-11-12
[PATCH v1] test: queue_perf: fix lockfree support check

psavol:
Fix segfault when -l/-w option is used, but lockfree/waitfree
queues are not supported.


muvarov:
Merged.

PR #751, opened 2018-11-14
[PATCH v4] linux-gen: timer: add inline timer config options

matiaselo:
V2:
- Free shm block on failure (Bill)

V3:
- Fixed timer_pool_new() not initializing allocated shm memory

V4:
- Always init odp_global_ro.init_param

### DIFF
--- a/platform/linux-generic/odp_timer.c
+++ b/platform/linux-generic/odp_timer.c
@@ -219,6 +219,7 @@ typedef struct timer_global_t {
 	_odp_atomic_flag_t ODP_ALIGNED_CACHE locks[NUM_LOCKS];
 #endif
 	odp_bool_t use_inline_timers;
+	int inline_poll_interval;
 } timer_global_t;
 
 static timer_global_t *timer_global;
@@ -871,8 +872,7 @@ static unsigned process_timer_pools(void)
 unsigned _timer_run(void)
 {
 	static __thread odp_time_t last_timer_run;
-	static __thread unsigned timer_run_cnt =
-		CONFIG_TIMER_RUN_RATELIMIT_ROUNDS;
+	static __thread unsigned timer_run_cnt = 1;
 	odp_time_t now;
 
 	if (timer_global->num_timer_pools == 0)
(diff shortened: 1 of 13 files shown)


matiaselo:
Travis hiccup

muvarov:
<pre>v1 checks:
checkpatch.pl - OK
</pre>

platform/linux-generic/odp_timer.c
```diff
@@ -217,7 +218,7 @@ typedef struct timer_global_t {
 	/* Multiple locks per cache line! */
 	_odp_atomic_flag_t ODP_ALIGNED_CACHE locks[NUM_LOCKS];
 #endif
-
+	odp_bool_t use_inline_timers;
```
lumag: Why do we need two separate variables: one here and one in `odp_global_rw`?

matiaselo: The value of `odp_global_rw.inline_timers` changes depending on if timer pools have been created to minimize overhead when timers are not actually used.

`timer_global_t.use_inline_timers` is set once during global init based on the config value. This avoids having to read the config file in `timer_pool_new()`.

lumag:

platform/linux-generic/odp_timer.c
```diff
@@ -1352,16 +1340,20 @@ int odp_timer_init_global(const odp_init_t *params)
 #else
 	ODP_DBG("Using lock-less timer implementation\n");
 #endif
+	conf_str =  "timer.inline";
+	if (!_odp_libconfig_lookup_int(conf_str, &val)) {
+		ODP_ERR("Config option '%s' not found.\n", conf_str);
+		return -1;
```
Bill-Fischofer-Linaro: You should back out the previous `odp_shm_reserve()` here before failing to avoid leaking `timer_global->shm`. Alternately, you could be a bit more forgiving and simply default to `use_inline_timers = false` if you can't find an override.

matiaselo: Will fix.

muvarov:
Merged.

PR #753, opened 2018-11-14
[PATCH v3] Build for native CPU arch 

psavol:
When in non abi compat mode, ODP can be build for native CPU architecture for optimal performance. Fix ARMv8.1 build issue. Without -march=native, ARM build was targeting ARMv8.0. Added sysinfo prints for checking ARM feature flags (ARMv8.0 vs 8.1).

### DIFF
--- a/platform/linux-generic/arch/aarch64/odp_sysinfo_parse.c
+++ b/platform/linux-generic/arch/aarch64/odp_sysinfo_parse.c
@@ -201,6 +201,76 @@ int cpuinfo_parser(FILE *file, system_info_t *sysinfo)
 
 void sys_info_print_arch(void)
 {
+	const char *ndef = "n/a";
+
+	/* Avoid compiler warning about unused variable */
+	(void)ndef;
+
+	/* See ARM C Language Extensions documentation for details */
+	printf("ARM FEATURES:\n");
+
+	printf("  __ARM_ARCH           ");
+#ifdef __ARM_ARCH
+	printf("%i\n", __ARM_ARCH);
+#else
+	printf("%s\n", ndef);
+#endif
+
+	printf("  __ARM_ARCH_ISA_A64   ");
+#ifdef __ARM_ARCH_ISA_A64
+	printf("%i\n", __ARM_ARCH_ISA_A64);
+#else
+	printf("%s\n", ndef);
+#endif
+
+#if defined(__ARM_ARCH) && __ARM_ARCH >= 8
+	/* Actually, this checks for new NEON instructions in
+	 * v8.1, but is currently the only way to distinguish
+	 * v8.0 and >=v8.1. */
+	printf("    ARMv8 ISA version  ");
+#ifdef __ARM_FEATURE_QRDMX
+	printf("v8.1 or higher\n");
+#else
+	printf("v8.0\n");
+#endif
+#endif
+
+#ifdef __ARM_FEATURE_QRDMX
+	/* Actually, this checks for new NEON instructions in
+	 * v8.1, but is currently the only way to distinguish
+	 * v8.0 and >=v8.1. */
+	printf("    ARMv8.1 instructions\n");
+#endif
+
+	printf("  __ARM_NEON           ");
+#ifdef __ARM_NEON
+	printf("%i\n", __ARM_NEON);
+#else
+	printf("%s\n", ndef);
+#endif
+
+	printf("  __ARM_FEATURE_IDIV   ");
+#ifdef __ARM_FEATURE_IDIV
+	printf("%i\n", __ARM_FEATURE_IDIV);
+#else
+	printf("%s\n", ndef);
+#endif
+
+	printf("  __ARM_FEATURE_CRYPTO ");
+#ifdef __ARM_FEATURE_CRYPTO
+	printf("%i\n", __ARM_FEATURE_CRYPTO);
+#else
+	printf("%s\n", ndef);
+#endif
+
+	printf("  __ARM_FEATURE_CRC32  ");
+#ifdef __ARM_FEATURE_CRC32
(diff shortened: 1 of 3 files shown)


platform/linux-generic/arch/aarch64/odp_sysinfo_parse.c
```diff
@@ -202,6 +202,49 @@ int cpuinfo_parser(FILE *file, system_info_t *sysinfo)
...
+#endif
+
+	value = -1;
+#ifdef __ARM_NEON
+	value = __ARM_NEON;
+#endif
+	printf("  __ARM_NEON           %i\n", value);
+
+	value = -1;
+#ifdef __ARM_FEATURE_IDIV
```
lumag: I don't quite like this kind of code. Maybe we can use the following template?

```c
#ifdef __ARM_FEATURE
   printf("  __ARM_FEATURE", __ARM_FEATURE);
#endif
```

lumag:
Looks good to me except the first patch.

platform/linux-generic/arch/aarch64/odp_sysinfo_parse.c
```diff
@@ -202,6 +202,49 @@ int cpuinfo_parser(FILE *file, system_info_t *sysinfo)
 
 void sys_info_print_arch(void)
 {
+	int value;
+
+	printf("ARM FEATURES:\n");
+	printf("  __ARM_ARCH           %i\n",  __ARM_ARCH);
+
```
Bill-Fischofer-Linaro: This is failing on Shippable:
```
  CC       arch/aarch64/odp_sysinfo_parse.lo
arch/aarch64/odp_sysinfo_parse.c: In function ‘sys_info_print_arch’:
arch/aarch64/odp_sysinfo_parse.c:207:41: error: ‘__ARM_ARCH’ undeclared (first use in this function)
  printf("  __ARM_ARCH           %i\n",  __ARM_ARCH);
                                         ^
arch/aarch64/odp_sysinfo_parse.c:207:41: note: each undeclared identifier is reported only once for each function it appears in
Makefile:1217: recipe for target 'arch/aarch64/odp_sysinfo_parse.lo' failed
make[1]: *** [arch/aarch64/odp_sysinfo_parse.lo] Error 1
```

platform/linux-generic/arch/aarch64/odp_sysinfo_parse.c
psavol: I want to print out outcome of all features tested (so that it's easy to compare the output). Most feature flags are not defined when the feature is disabled, and define 1 when it's enabled. So, above example would not print anything when it's disabled.

muvarov: I'm ok with current changes.  while value var is not needed here.  And ifdef can go inside print. But that is tiny cosmetic thing. I mean like that:

```
	printf("SOMETHING %d\n",
#if defined(SOMETHING)
	       SOMETHING
#else
	       -1
#endif
	      );
```

platform/linux-generic/arch/aarch64/odp_sysinfo_parse.c
psavol: Shippable has quite old GCC version: 4.8.5. Surprisingly it does not define ARM_ARCH. I'll ifdef it also. Should we update shippable SW level ?

muvarov: @psavol Shippable has available update. I clicked to install it. Will see if it will update gcc or not.

psavol:
v3: rebased and updated print format to print "n/a" when a feature define is not defined.

muvarov:
The code which prints define is ok. But does it actully mean that generated binary uses that cpu instructions? I.e. instructions can be used or can not . And it's not runtime thing. More clear place for it is ./configure. As I understand propose of it - it's just print some information about odp platform. Might be useful to debug flags propogated to compiller on compilation time.

platform/linux-generic/arch/aarch64/odp_sysinfo_parse.c
lumag: @muvarov please no. This is IMO the ugliest possible form.
Moreover if somebody uses `#define` to define printf, things will break, as preprocessors do not have to support directives inside macro arguments.

muvarov: @lumag  ok then looks like this PR can go in.

platform/linux-generic/arch/aarch64/odp_sysinfo_parse.c
lumag: @psavol @muvarov this is interesting. I could not find used gcc in the docker image mounts.

muvarov: ./configure output has to print it

lumag: Hmmm. @muvarov who controls `odp/dev` image? It is used for Shippable builds. Most probably we should switch to our images built from `Linaro/odp-docker-images`?

lumag: @muvarov could you please enable shippable build for odp-docker-image with empty .shippable.yml file for now? I'd try building new aarch64 image from that repo.

lumag:
Unless @Bill-Fischofer-Linaro has additional comments, I'd put my reviewed-by tag on this PR.

platform/linux-generic/arch/aarch64/odp_sysinfo_parse.c
muvarov: @lumag enabled in Shippable web interface

lumag:

muvarov:
Merged.

PR #754, opened 2018-11-16
[PATCH API-NEXT v3] api: init: add new mem_model member to odp_init_t

matiaselo:
Enables an application to support process mode without the need to modify ODP configuration file.

New odph_options() helper function is added to enable an application to find out if ODP should be started in process mode.

V2:
- Clarify that handle/pointer sharing works only within an ODP instance (Bill)

V3:
- Add default value statement (Petri)
- Use odp_mem_model_t with helpers (Petri)

### DIFF
--- a/include/odp/api/spec/init.h
+++ b/include/odp/api/spec/init.h
@@ -107,6 +107,32 @@ typedef int (*odp_log_func_t)(odp_log_level_t level, const char *fmt, ...);
 typedef void (*odp_abort_func_t)(void) ODP_NORETURN;
 
 /**
+ * Application memory model
+ */
+typedef enum {
+	/** Thread memory model: by default all memory is shareable between
+	 *  threads.
+	 *
+	 *  Within a single ODP instance all ODP handles and pointers to ODP
+	 *  allocated data may be shared amongst threads independent of data
+	 *  allocation time (e.g. before or after thread creation). */
+	ODP_MEM_MODEL_THREAD = 0,
+
+	/** Process memory model: by default all memory is not shareable between
+	 *  processes.
+	 *
+	 *  Within a single ODP instance all ODP handles and pointers to ODP
+	 *  allocated data (excluding non-single VA SHM blocks) may be shared
+	 *  amongst processes independent of data allocation time (e.g. before
+	 *  or after fork).
+	 *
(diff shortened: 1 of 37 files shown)


include/odp/api/spec/init.h
```diff
@@ -106,6 +106,26 @@ typedef int (*odp_log_func_t)(odp_log_level_t level, const char *fmt, ...);
...
 typedef void (*odp_abort_func_t)(void) ODP_NORETURN;
 
+/**
+ * Application memory model
+ */
+typedef enum {
+	/** All memory is shareable between threads. All ODP handles and
+	 *  pointers to ODP allocated data may be shared amongst threads
+	 *  independent of data allocation time. */
+	ODP_MEM_MODEL_THREAD = 0,
```
psavol: ... independent of data allocation time (e.g. before or after creation of a thread).

include/odp/api/spec/init.h
```diff
@@ -106,6 +106,26 @@ typedef int (*odp_log_func_t)(odp_log_level_t level, const char *fmt, ...);
...
+
+	/** By default all memory is not shareable between processes. All ODP
+	 *  handles and pointers to ODP allocated data (excluding non-single VA
+	 *  SHM blocks) may be shared amongst processes independent of data
+	 *  allocation time.
```
psavol: ... independent of data allocation time (e.g. before or after fork of a thread).

Bill-Fischofer-Linaro: If we want to clarify this, I'd clarify it to point out that sharing is only for threads/processes created after the `odp_init_global()` call, i.e., within a given ODP instance.

include/odp/api/spec/init.h
```diff
@@ -172,6 +198,12 @@ typedef struct odp_init_t {
 	 */
 	odp_feature_t not_used;
 
+	/** Application memory model. The main application thread has to call
+	 *  odp_init_global() and odp_init_local() before creating threads that
+	 *  share ODP data.
+	 */
+	odp_mem_model_t mem_model;
```
psavol: The default value needs to be defined: "The default value is ODP_MEM_MODEL_THREAD"

muvarov:
Merged.

PR #756, opened 2018-11-16
[PATCH v2] travis: define compiler for clang test

muvarov:
without specifying compiler CC falls back to default gcc
after it was initialized to clang. Current change forces
test to use clang.


lumag:
Please drop unnecessary `CC` settings. They are confusing.

muvarov:
@lumag that is confusing but it's nice view in Travis results output. I left them only to see that it's clang test.

muvarov:
@lumag updated. I think that can go in just after #757

muvarov:
Merged.

PR #757, opened 2018-11-19
[PATCH v2] configure: disable -march=native for clang

muvarov:
for clang we see strange optimizations for TM code which breaks
code flow execution. There might be number of such places. For
now it's better to disable march=native until we completely
test and validate clang support for this.


muvarov:
Merged.

PR #765, opened 2018-11-28
[PATCH v3] merge api-next to master (without flow aware and compression)

muvarov:


platform/linux-generic/m4/configure.m4
```diff
@@ -20,8 +20,6 @@ m4_include([platform/linux-generic/m4/odp_netmap.m4])
 m4_include([platform/linux-generic/m4/odp_dpdk.m4])
 ODP_SCHEDULER
 
-m4_include([platform/linux-generic/m4/performance.m4])
-
```
matiaselo: `configure.ac` line 413: `	test_perf_proc:		${test_perf_proc}` still has to be removed.

muvarov: updated in v3.

muvarov:
Merged.

PR #766, opened 2018-11-28
[PATCH v2] Update dpdk support to v18.11 level

psavol:
I have build this with v17.11.4, v18.02.2, v18.05.1 and v18.11. Make check passed with v17.11 locally, v18.11 failed one validation/api/pktio/pktio_run_dpdk.sh test case. L2fwd worked with v17.11 and v18.11.

This should not be included into ODP v1.20 as more testing is needed (e.g. v17.11 and v18.11 with zero copy).

### DIFF
--- a/DEPENDENCIES
+++ b/DEPENDENCIES
@@ -164,7 +164,8 @@ Prerequisites for building the OpenDataPlane (ODP) API
 
 3.4 DPDK packet I/O support (optional)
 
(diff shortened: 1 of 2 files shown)


psavol:
v2: rebased and tested with 17.11, 18.02, 18.05 and 18.11

muvarov:
Merged.

PR #773, opened 2018-12-14
[PATCH v1] linux-gen: dpdk: fix build with no pmd drivers

matiaselo:
Match pattern was added to DPDK_PMDS if no pmd drivers were found.


muvarov:
Merged

PR #774, opened 2018-12-17
[PATCH v1] linux-gen: socket_mmap: fix build with older clang versions

matiaselo:
Olders clang versions (at least 3.4.2 used by CentOS) don't handle
frame_map.v2.s_ll definition properly and throw invalid error.


muvarov:
interesting change :) merged.

PR #775, opened 2018-12-18
Flow aware schedule test and dummy implementation

psavol:


Bill-Fischofer-Linaro:
@lumag Please add your review, especially given the slight API changes incorporated here.

platform/linux-generic/odp_schedule_if.c
```diff
@@ -51,18 +48,20 @@ int odp_schedule_config(const odp_schedule_config_t *config)
 	int ret;
 	odp_schedule_config_t defconfig;
 
-	ODP_ASSERT(!_odp_schedule_configured);
+	if (_odp_schedule_configured) {
+		ODP_ERR("Scheduler has been configured already\n");
+		return -1;
```
muvarov: I think here should not be return. 2 odp_schedule_config() should be valid for user or unconfig() function has to be provided.

platform/linux-generic/odp_event.c
```diff
@@ -59,17 +59,14 @@ int odp_event_type_multi(const odp_event_t event[], int num,
 	return i;
 }
 
-/* For now ODP generic does not support flow awareness,
- * so all flow ids are zero. */
-uint32_t odp_event_flow_id(odp_event_t event ODP_UNUSED)
+uint32_t odp_event_flow_id(odp_event_t event)
```
muvarov: has it be uint64_t everywhere now?

platform/linux-generic/odp_schedule_if.c
psavol: There's no point of application calling config() multiple times as only the last config call would apply. Single config() call enable implementation to configure and start HW scheduler at this point. Typically, HW scheduler cannot be reconfigured without reboot.

I'll add note to the API spec that config() can be called only once.

platform/linux-generic/odp_event.c
psavol: The problem was that if ID is 32 bits, max count must be at least 33 bits. I'll change capa from max count to max ID, so that both can be the same size (uint32_t).

platform/linux-generic/odp_schedule_if.c
psavol: Config() can be called only once per instance (once after global_init()). API does not require synchronization between instances. The point is that this is not supported:

global_init()
...
config()
config()
config()
...
schedule()

platform/linux-generic/odp_event.c
psavol: v3 changes capa/config to max_flow_id. When max_flow_id is 0, flow aware mode is not supported.

psavol:
v3: change API capa/config to max_flow_id (uint32_t)
v4: correct test code max_flow_id usage

muvarov:
I'm ok with V4.

@lumag please do you review also.

platform/linux-generic/odp_schedule_if.c
lumag: @muvarov if you have such an applicaiton, it will call `capability()`, select optimum values and then call `config()`. Other subsystems (IPsec) follow same approach - you can call `config()` only once.

test/validation/api/scheduler/scheduler.c
```diff
@@ -2025,6 +2181,32 @@ odp_suiteinfo_t scheduler_suites[] = {
...
+	}
+
+	if (0 != odp_init_local(*inst, ODP_THREAD_CONTROL)) {
+		fprintf(stderr, "error: odp_init_local() failed.\n");
+		return -1;
+	}
+
+	return 0;
+}
+
```
lumag: There is no need to overwrite global init, isn't it?

psavol: The default validation test init contains schedule_config(NULL), so we override that in this suite to be able to call schedule_config() with non-default values.

lumag:

@muvarov please merge

PR #787, opened 2019-01-17
[PATCH v1] linux-gen: dpdk: add a fallback value for the number of numa nodes

matiaselo:
numa_num_configured_nodes() may return 0 on some platforms. In that case
use 1 as a replacement value.


muvarov:
Merged.

PR #810, opened 2019-02-07
[PATCH v4] API: queue parameter to disable reordering

psavol:
Application may do enqueues to multiple destination queues from within an ordered scheduling context. It's wasteful and sometimes not acceptable (due to potentially large reordering delay) to reorder all destination queues. E.g. application may use plain queues as temporary object storage, etc which do not benefit from enqueue operation reordering. Ordering context is released during the next schedule call (odp_schedule_release_ordered() is just a hint), so it would be complex for application to avoid unnecessary re-ordering.

### DIFF
--- a/configure.ac
+++ b/configure.ac
@@ -4,7 +4,7 @@ AC_PREREQ([2.5])
 ##########################################################################
 m4_define([odpapi_generation_version], [1])
 m4_define([odpapi_major_version], [21])
-m4_define([odpapi_minor_version], [0])
+m4_define([odpapi_minor_version], [1])
 m4_define([odpapi_point_version], [0])
 m4_define([odpapi_version],
     [odpapi_generation_version.odpapi_major_version.odpapi_minor_version.odpapi_point_version])
--- a/include/odp/api/spec/queue_types.h
+++ b/include/odp/api/spec/queue_types.h
@@ -123,6 +123,37 @@ typedef enum odp_nonblocking_t {
 } odp_nonblocking_t;
 
 /**
+ * Original event order maintenance options
+ *
+ * Options to keep or ignore the original event order of a source queue. This
+ * option is relevant for (plain or parallel scheduled) queues that are
+ * destinations for events enqueued while holding an ordered queue
+ * synchronization context. By default, an ordered context maintains original
+ * event order regardless of the destination queue type. Event re-ordering may
+ * cause extra synchronization effort for implementation and a long delay before
+ * application can receive a re-ordered event from the destination queue. This
+ * is wasteful and in some cases the extra delay is not acceptable for those
+ * destination queues that do not need to maintain the original event order.
+ * Application can use ODP_QUEUE_ORDER_IGNORE option to prevent implementation
+ * from performing unnecessary event re-ordering and negative side-effects of
+ * that.
+ */
+typedef enum odp_queue_order_t {
+	/** Keep original event order. Events enqueued into this queue while
+	 *  holding an ordered queue synchronization context maintain the
+	 *  original event order of the source queue.
+	 */
+	ODP_QUEUE_ORDER_KEEP = 0,
+
+	/** Ignore original event order. Events enqueued into this queue do not
+	 *  need to maintain the original event order of the source queue.
+	 *  Implementation must avoid significant event re-ordering delays.
+	 */
+	ODP_QUEUE_ORDER_IGNORE
+
+} odp_queue_order_t;
+
+/**
  * Queue capabilities
  */
 typedef struct odp_queue_capability_t {
@@ -261,6 +292,12 @@ typedef struct odp_queue_param_t {
 	  * These parameters are considered only when queue type is
 	  * ODP_QUEUE_TYPE_SCHED. */
 	odp_schedule_param_t sched;
+
+	/** Original event order maintenance
+	  *
+	  * Keep or ignore the original event order of a source queue.
+	  * The default value is ODP_QUEUE_ORDER_KEEP. */
+	odp_queue_order_t order;
 
 	/** Non-blocking level
 	  *
--- a/include/odp/api/spec/schedule_types.h
+++ b/include/odp/api/spec/schedule_types.h
@@ -77,7 +77,10 @@ extern "C" {
  * The atomic queue synchronization context is dedicated to the thread until it
  * requests another event from the scheduler, which implicitly releases the
  * context. User may allow the scheduler to release the context earlier than
- * that by calling odp_schedule_release_atomic().
+ * that by calling odp_schedule_release_atomic(). However, this call is just
+ * a hint to the implementation and the context may be held until the next
+ * schedule call.
+ *
  * When scheduler is enabled as flow-aware, the event flow id value affects
  * scheduling of the event and synchronization is maintained per flow within
  * each queue.
@@ -97,7 +100,9 @@ extern "C" {
  * queue synchronization context. A thread holds the context until it
  * requests another event from the scheduler, which implicitly releases the
  * context. User may allow the scheduler to release the context earlier than
- * that by calling odp_schedule_release_ordered().
+ * that by calling odp_schedule_release_ordered(). However, this call is just
+ * a hint to the implementation and the context may be held until the next
+ * schedule call.
  *
  * Events from the same (source) queue appear in their original order
  * when dequeued from a destination queue. The destination queue can have any
@@ -107,6 +112,11 @@ extern "C" {
  * (e.g. freed or stored) within the context are considered missing from
  * reordering and are skipped at this time (but can be ordered again within
  * another context).
+ *
+ * Unnecessary event re-ordering may be avoided for those destination queues
+ * that do not need to maintain the original event order by setting 'order'
+ * queue parameter to ODP_QUEUE_ORDER_IGNORE.
+ *
  * When scheduler is enabled as flow-aware, the event flow id value affects
  * scheduling of the event and synchronization is maintained per flow within
  * each queue.
--- a/test/validation/api/scheduler/scheduler.c
+++ b/test/validation/api/scheduler/scheduler.c
@@ -23,7 +23,7 @@
 #define NUM_GROUPS              2
 #define MAX_QUEUES              (64 * 1024)
 
-#define TEST_QUEUE_SIZE_NUM_EV  50
+#define DEFAULT_NUM_EV          50
 
 #define MAX_FLOWS               16
 #define FLOW_TEST_NUM_EV        (10 * MAX_FLOWS)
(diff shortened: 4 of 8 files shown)


include/odp/api/spec/schedule_types.h
```diff
@@ -77,7 +77,10 @@ extern "C" {
  * The atomic queue synchronization context is dedicated to the thread until it
  * requests another event from the scheduler, which implicitly releases the
  * context. User may allow the scheduler to release the context earlier than
- * that by calling odp_schedule_release_atomic().
+ * that by calling odp_schedule_release_atomic(). However, this call is just
+ * a hint to the implementation and the context may be helt until the next
```
Bill-Fischofer-Linaro: Typo: held

include/odp/api/spec/queue_types.h
```diff
@@ -122,6 +122,37 @@ typedef enum odp_nonblocking_t {
...
+typedef enum odp_queue_order_t {
+	/** Keep original event order. Events enqueued into this queue while
+	 *  holding an ordered queue synchronization context maintain the
+	 *  original event order of the source queue.
+	 */
+	ODP_QUEUE_ORDER_KEEP = 0,
+
+	/** Ignore original event order. Events enqueued into this queue do not
+	 *  need to maintain the original event order of the source queue.
+	 *  Implementation must avoid significant event re-ordering delays.
```
Bill-Fischofer-Linaro: Must seems a bit strong here. Shouldn't this be worded similar to other performance hints? Wording such as "Implementations may ignore ordering to provide enqueue improved performance" might be more consistent with other such hints.

include/odp/api/spec/schedule_types.h
```diff
@@ -97,7 +100,9 @@ extern "C" {
  * queue synchronization context. A thread holds the context until it
  * requests another event from the scheduler, which implicitly releases the
  * context. User may allow the scheduler to release the context earlier than
- * that by calling odp_schedule_release_ordered().
+ * that by calling odp_schedule_release_ordered(). However, this call is just
+ * a hint to the implementation and the context may be helt until the next
```
Bill-Fischofer-Linaro: Typo: held

configure.ac
```diff
@@ -4,7 +4,7 @@ AC_PREREQ([2.5])
 ##########################################################################
 m4_define([odpapi_generation_version], [1])
 m4_define([odpapi_major_version], [21])
-m4_define([odpapi_minor_version], [0])
+m4_define([odpapi_minor_version], [1])
```
Bill-Fischofer-Linaro: Is this PR intended to be its own release? If yes, then CHANGELOG updates should also be included in this PR.

include/odp/api/spec/queue_types.h
psavol: Application needs to be able to trust that an enqueue operation does not wait infinitely or does not have an unexpectedly long timeout for waiting other threads. So, "must" is needed in the wording (in one form or another) instead of "may".

Bill-Fischofer-Linaro: Do we have scenarios where use of ordered queues can result in deadlock or livelock? As long as individual threads don't stall for other reasons they'll eventually release the ordered context, which will permit any pending reorderings to be resolved.

include/odp/api/spec/queue_types.h
```diff
@@ -122,6 +122,37 @@ typedef enum odp_nonblocking_t {
...
+ * event order regardless of the destination queue type. Event re-ordering may
+ * cause extra synchronization effort for implementation and a long delay before
+ * application can receive a re-ordered event from the destination queue. This
+ * is wasteful and in some cases the extra delay is not acceptable for those
+ * destination queues that do not need to maintain the original event order.
+ * Application can use ODP_QUEUE_ORDER_IGNORE option to prevent implementation
+ * from performing unnecessary event re-ordering and negative side-effects of
+ * that.
+ */
+typedef enum odp_queue_order_t {
```
mvl-skardach: Clarification - what should happen to the events allocated by the thread? I.e.:
```c
odp_event_t e1 = odp_schedule(NULL, ODP_SCHED_NO_WAIT);
odp_event_t e2 = odp_buffer_to_event(odp_buffer_alloc(some_bufpool);

odp_queue_enq(q1, e2);
odp_queue_enq(q1, e1);
```
With above in mind, what is the required order of `e2`? What changes if `q1` is a `ODP_QUEUE_ORDER_IGNORE` queue?

test/validation/api/scheduler/scheduler.c
```diff
@@ -507,6 +507,120 @@ static void scheduler_test_queue_size(void)
...
+		if (ev == ODP_EVENT_INVALID)
+			continue;
+
+		CU_ASSERT(from == ordered);
+		ret = odp_queue_enq(plain, ev);
+
+		if (ret)
+			odp_event_free(ev);
+		else
+			num++;
```
mvl-skardach: You could break if num == queue_size. It will decrease test time.

test/validation/api/scheduler/scheduler.c
```diff
@@ -507,6 +508,120 @@ static void scheduler_test_queue_size(void)
...
+		ev = odp_queue_deq(plain);
+
+		if (ev == ODP_EVENT_INVALID)
+			continue;
+
+		odp_event_free(ev);
+		num++;
+	}
+
+	CU_ASSERT(num == queue_size);
```
mvl-skardach: There seems to be no observable behavior that would test platform's conformity to `ODP_QUEUE_ORDER_IGNORE`. In a single threaded test it is even impossible to get re-ordered events. Switching to multi-threaded scenario would also not be a solution (since re-ordering is a race-condition).
Therefore I'm not sure what does this test bring to the validation suite?

test/validation/api/scheduler/scheduler.c
```diff
@@ -1712,8 +1827,9 @@ static int create_queues(test_globals_t *globals)
 	num_sched = (prios * queues_per_prio * sched_types) + CHAOS_NUM_QUEUES;
 	num_plain = (prios * queues_per_prio);
 	while ((num_sched > default_config.num_queues ||
-		num_plain > capa.plain.max_num ||
-		num_sched + num_plain > capa.max_queues) && queues_per_prio) {
+		num_plain > queue_capa.plain.max_num ||
+		num_sched + num_plain > queue_capa.max_queues) &&
+		queues_per_prio) {
```
mvl-skardach: Unless I'm mistaken queues_per_prio should be 7 spaces inlined instead of 1tab==8 spaces. It is at the level of the whole `()`.

configure.ac
psavol: Not a release (no tag), but API development in master. Minor version number increment highlights that API is not any more at release 1.21 level, although it's backwards compatible to that. Next release will be tagged as v1.22.

include/odp/api/spec/queue_types.h
psavol: Nothing changes. e2 will be always before e1 as those are enqueued from the same thread and ordering context. Re-ordering apply between ordering contexts (threads).

mvl-skardach: OK, my proper code example would be:

```c
odp_event_t e1 = odp_schedule(NULL, ODP_SCHED_NO_WAIT);
odp_event_t e2 = odp_buffer_to_event(odp_buffer_alloc(some_bufpool);

odp_queue_enq(q1, e2);
odp_queue_enq(q2, e1);
```

And of course there are 2 threads which perform this action. My question is should the ordering be performed at enq to q2 or at enq to q1? Given that it's the thread that helds the ordering context, not an event.

include/odp/api/spec/queue_types.h
psavol: Deadlock will result e.g. when application protects a plain queue with a spinlock.

lock() -> enqueue() -> unlock() ...  lock() -> while(dequeue() == EVENT_INVALID) -> unlock()

If a thread holds a lock and waits in dequeue, other threads cannot enqueue and re-order never resolves...

include/odp/api/spec/schedule_types.h
```diff
@@ -97,7 +100,9 @@ extern "C" {
  * queue synchronization context. A thread holds the context until it
  * requests another event from the scheduler, which implicitly releases the
  * context. User may allow the scheduler to release the context earlier than
- * that by calling odp_schedule_release_ordered().
+ * that by calling odp_schedule_release_ordered(). However, this call is just
```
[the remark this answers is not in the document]
psavol: It relates to the fact that application cannot use release_ordered() to solve the deadlock problem. It cannot do plain queue enq/deq after context release easily (without calling sechedule).

test/validation/api/scheduler/scheduler.c
psavol: Schedule needs to be called at least once after last event to free the last context.

test/validation/api/scheduler/scheduler.c
psavol: This indent passes checkpatch, so the coding style should be OK.

include/odp/api/spec/queue_types.h
psavol: Thread successfully enqueued an event into the queue, so it should be able to dequeue it from there in a finite time. Locks can be used for many reasons to protect the queue and e.g. counters calculating how many events are in the queue.

include/odp/api/spec/schedule_types.h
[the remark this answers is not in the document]
mvl-skardach: Got it. That makes sense on `_enq()`. Though `_deq()` cannot maintain order since it can be done only on PLAIN queues. Or are we allowing scheduler implementation affect plain queue implementation.

test/validation/api/scheduler/scheduler.c
psavol: As commit log says, it's a simple test to ensure that we at least use the option in our test suite.

include/odp/api/spec/queue_types.h
mvl-skardach: OK. That makes sense.

include/odp/api/spec/schedule_types.h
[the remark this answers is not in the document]
psavol: Ordering spec requires that enqueues done inside ordering context appear in destination queue in correct order. Destination queue can be either sched or plain type. So, e.g. if single thread is doing all dequeues on the destination queue, it will see all events come out from the plain in the right order.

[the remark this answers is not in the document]
mvl-skardach: OK, I think we're both saying the same thing. What I mean is ordering can only happen on events retrieved via `odp_schedule()`, not ones coming from `odp_queue_deq()`.
Thread may `odp_queue_enq()` events coming from `odp_schedule()` to a `PLAIN` queue and those will be ordered.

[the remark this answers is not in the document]
psavol: yes

include/odp/api/spec/schedule_types.h
psavol: fixed in v2

include/odp/api/spec/schedule_types.h
psavol: fixed in v2

PR #814, opened 2019-02-20
[PATCH v2] example: check odp_shm_reserve() return value

matiaselo:


example/generator/odp_generator.c
```diff
@@ -1142,6 +1142,12 @@ int main(int argc, char *argv[])
 	/* Reserve memory for args from shared mem */
 	shm = odp_shm_reserve("shm_args", sizeof(args_t),
 			      ODP_CACHE_LINE_SIZE, 0);
+
+	if (shm == ODP_SHM_INVALID) {
+		EXAMPLE_ERR("Error: shared mem reserve failed.\n");
+		exit(EXIT_FAILURE);
+	}
+
```
Bill-Fischofer-Linaro: These extra tests are really unnecessary as `odp_shm_addr(ODP_SHM_INVALID)` will return `NULL` and be picked up by the guard that follows.

platform/linux-generic/test/mmap_vlan_ins/mmap_vlan_ins.c
```diff
@@ -172,6 +172,11 @@ int main(int argc, char **argv)
 	/* Reserve memory for args from shared mem */
 	shm = odp_shm_reserve("_appl_global_data", sizeof(global_data_t),
 			      ODP_CACHE_LINE_SIZE, 0);
+	if (shm == ODP_SHM_INVALID) {
+		printf("Error: shared mem alloc failed.\n");
+		exit(EXIT_FAILURE);
+	}
+
```
Bill-Fischofer-Linaro: Same comments for these tests.

example/generator/odp_generator.c
matiaselo: The issue here is that the API definition doesn't state what happens if you pass an invalid shm handle to `odp_shm_addr()`(e.g. implementation may crash). Linux-generic handles it properly and returns NULL as you stated, but I think the example applications shouldn't make this assumption.

Bill-Fischofer-Linaro: The API spec for `odp_shm_addr()` is pretty simple:
```
/**
 * Shared memory block address
 *
 * @param shm    Block handle
 *
 * @return Memory block address
 * @retval NULL on failure
 */
void *odp_shm_addr(odp_shm_t shm);
```
Note that there are no "results are undefined" outputs, so passing `ODP_SHM_INVALID` to this API should return `NULL`. This should probably be added as an explicit validation test.

matiaselo: I think the standard assumption is that passing an invalid handle as function parameter is not OK, unless it is allowed in the function specification?

Bill-Fischofer-Linaro: Consider these two other APIs:
```
/**
 * Perform full packet validity check
 *
 * The operation may consume considerable number of cpu cycles depending on
 * the check level.
 *
 * @param pkt  Packet handle
 *
 * @retval 0 Packet is not valid
 * @retval 1 Packet is valid
 */
int odp_packet_is_valid(odp_packet_t pkt);
```
and 
``/**
 * Packet data length
 *
 * Returns total data length over all packet segments. This equals the sum of
 * segment level data lengths (odp_packet_seg_data_len()).
 *
 * @param pkt  Packet handle
 *
 * @return Packet data length
 *
 * @see odp_packet_seg_len(), odp_packet_data(), odp_packet_seg_data_len()
 */
uint32_t odp_packet_len(odp_packet_t pkt);`
```
Passing `ODP_PACKET_INVALID` to the first is OK, but not to the second. The reason is that the first fully specifies the output values while the second only specifies the intended output. 

`odp_shm_addr()` falls into the first category since it explicitly says that it returns `NULL` on any error. So as long as you're passing an input of type `odp_shm_t` (and `ODP_SHM_INVALID` is of type `odp_shm_t`) you're going to get a defined output from that API.

psavol: "Returns NULL on failure" does not specify which failures are checked and which are not. By default, our APIs leave error check level implementation specific. E.g. odp_shm_addr() may or may not check against ODP_SHM_INVALID, ODP_SHM_INVALID + 1, ODP_SHM_INVALID + 2, etc. So, application should not rely on check against ODP_SHM_INVALID.

Actually, even odp_packet_is_valid() spec does not promise INVALID/random handle check. It will check if packet metadata is OK, but random handle could crash it.

PR #820, opened 2019-03-13
[PATCH v3] api: shm: add ODP_SHM_HP flag

matiaselo:
When set, this flag guarantees that the memory reserved by odp_shm_reserve() is allocated from huge pages. The call will fail if enough huge page memory is not available.

V2:
- ODP_SHM_HP comment wording and comment clean-up (Petri)
- Removed internal _odp_shm_reserve() function

### DIFF
--- a/include/odp/api/spec/shared_memory.h
+++ b/include/odp/api/spec/shared_memory.h
@@ -1,4 +1,5 @@
-/* Copyright (c) 2013-2018, Linaro Limited
+/* Copyright (c) 2019, Nokia
+ * Copyright (c) 2013-2018, Linaro Limited
  * All rights reserved.
  *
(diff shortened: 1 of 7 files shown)


include/odp/api/spec/shared_memory.h
```diff
@@ -62,6 +63,15 @@ extern "C" {
  */
 #define ODP_SHM_EXPORT		0x08
 
+/**
+ * Use huge pages
+ *
+ * When set, this flag guarantees that the memory reserved by odp_shm_reserve()
+ * is allocated from huge pages. The call will fail if enough huge page memory
+ * is not available.
```
psavol: Prefered wording: "The call returns failure if ..."

PR #823, opened 2019-03-13
[PATCH v7] build: support platforms that do not use config

lumag:
Support platforms that do not use config file.

Change-Id: I9f82337bcd547e0757cc1b5b0c9cdbcf99758b63
Signed-off-by: Dmitry Eremin-Solenikov <deremin-solenikov@cavium.com>

### DIFF
--- a/configure.ac
+++ b/configure.ac
@@ -231,6 +231,8 @@ AS_IF([test "${with_platform}" = "linux-generic"],
 AC_DEFINE_UNQUOTED([ODP_IMPLEMENTATION_NAME], ["$ODP_IMPLEMENTATION_NAME"],
 		   [Define to the name of the implementation])
 
+AM_CONDITIONAL([ODP_USE_CONFIG], [test "x$odp_use_config" = "xtrue"])
+
 ##########################################################################
 # Build examples/tests dynamically
 ##########################################################################
--- a/Makefile.am
+++ b/Makefile.am
@@ -28,7 +28,7 @@ endif
 
 @DX_RULES@
 
-EXTRA_DIST = bootstrap CHANGELOG config/README config/odp-$(with_platform).conf
+EXTRA_DIST = bootstrap CHANGELOG config/README
 
 distcheck-hook:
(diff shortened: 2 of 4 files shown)


configure.ac
```diff
@@ -231,6 +231,8 @@ AS_IF([test "${with_platform}" = "linux-generic"],
 AC_DEFINE_UNQUOTED([ODP_IMPLEMENTATION_NAME], ["$ODP_IMPLEMENTATION_NAME"],
 		   [Define to the name of the implementation])
 
+AM_CONDITIONAL([ODP_USE_CONFIG], [test "xodp_use_config" = "true"])
```
matiaselo: Shouldn't the comparison be `[test "x$odp_use_config" = "xtrue"]`?

lumag: Hmm, true. I tested only for my platform, which doesn't use config. Sorry.

configure.ac
```diff
@@ -231,6 +231,8 @@ AS_IF([test "${with_platform}" = "linux-generic"],
 AC_DEFINE_UNQUOTED([ODP_IMPLEMENTATION_NAME], ["$ODP_IMPLEMENTATION_NAME"],
 		   [Define to the name of the implementation])
 
+AM_CONDITIONAL([ODP_USE_CONFIG], [test "xodp_use_config" = "xtrue"])
```
matiaselo: The $ sign is still missing in front of `odp_use_config`.

lumag: fixed

matiaselo:
Reviewed-by tags missing, otherwise ready to merge.

lumag:
Unfortunately this is not enough: `EXTRA_DIST` in main `Makefile.am` references `config/odp-$(with_platform).conf`. If bundling several platforms, this variable will not be multi-substituted (for each platform), so the generated tarball will have one config file at most.
Updated PR with proposed solution.

matiaselo:
@lumag Looks good, please add the reviewed-by tags from me and Stand, so I can merge this PR.

PR #825, opened 2019-03-20
[PATCH v3] example: add new pipeline example application

matiaselo:
Add new simple pipeline example application. The application receives
packets from one interface and passes them through 0-N worker stages before
outputting them from a second network interface. The RX, worker, and TX
stages are connected using plain queues and each stage is run on a separate
CPU thread. The workers stages calculate CRC-32C over packet data.

V2:
- Make calculating CRC-32C optional (Petri)


example/simple_pipeline/odp_simple_pipeline.c
```diff
@@ -0,0 +1,938 @@
...
+	while (!global->exit_threads) {
+		events = odp_queue_deq_multi(rx_queue, event_tbl,
+					     MAX_PKT_BURST);
+
+		if (odp_unlikely(events <= 0))
+			continue;
+
+		stats->s.rx_cnt += events;
+
+		work_on_events(event_tbl, events);
```
psavol: This could be enabled/disabled with an option. CRC cycles get dominant with large packet sizes.

PR #829, opened 2019-04-03
[PATCH v2] linux-gen: crypto: disable openssl locking callbacks

matiaselo:
Starting from OpenSSL versions 1.1.0 the locking callbacks are no longer
required.

Fixes: https://github.com/OpenDataPlane/odp/issues/817

Reported-by: Jussi Kerttula <jussi.kerttula@nokia.com>

### DIFF
--- a/platform/linux-generic/odp_crypto_openssl.c
+++ b/platform/linux-generic/odp_crypto_openssl.c
@@ -1828,6 +1828,7 @@ odp_crypto_operation(odp_crypto_op_param_t *param,
 	return 0;
 }
 
+#if OPENSSL_VERSION_NUMBER < 0x10100000L
 static void ODP_UNUSED openssl_thread_id(CRYPTO_THREADID ODP_UNUSED *id)
 {
 	CRYPTO_THREADID_set_numeric(id, odp_thread_id());
@@ -1842,6 +1843,7 @@ static void ODP_UNUSED openssl_lock(int mode, int n,
 	else
 		odp_ticketlock_unlock(&global->openssl_lock[n]);
 }
+#endif
 
 int
 odp_crypto_init_global(void)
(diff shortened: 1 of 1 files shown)


lumag:
Hmm, strange. Code from master branch compiles for me for OpenSSL 1.1.0/1.1.1

matiaselo:
> Hmm, strange. Code from master branch compiles for me for OpenSSL 1.1.0/1.1.1

Strange, for me it fails (OpenSSL 1.1.1b, GCC 7.3.0):
```
In file included from /home/nnn/openssl_install/include/openssl/bio.h:20:0,
                 from /home/nnn/openssl_install/include/openssl/evp.h:16,
                 from /home/nnn/openssl_install/include/openssl/hmac.h:15,
                 from odp_crypto_openssl.c:31:
odp_crypto_openssl.c: In function \u2018odp_crypto_init_global\u2019:
/home/nnn/openssl_install/include/openssl/crypto.h:235:57: error: statement with no effect [-Werror=unused-value]
 #  define CRYPTO_THREADID_set_callback(threadid_func)   (0)
                                                         ^
odp_crypto_openssl.c:1883:3: note: in expansion of macro \u2018CRYPTO_THREADID_set_callback\u2019
   CRYPTO_THREADID_set_callback(openssl_thread_id);
```

mvl-skardach:
Those are results of `-Wall -Werror`. If you change gcc to 8.3.0 you'll see even more warnings.
I would prefer to disable this warning (`Wunused-value`) rather than adding version checks. This warning is not a very useful one anyway.

matiaselo:
> Those are results of `-Wall -Werror`. If you change gcc to 8.3.0 you'll see even more warnings.
> I would prefer to disable this warning (`Wunused-value`) rather than adding version checks. This warning is not a very useful one anyway.

Are these warnings caused by OpenSSL or are they from our code (I haven't tested 8 series gcc yet)? In case they are from our code we should fix these warnings. In my opinion disabling `-Wunused-value` warnings should be the last option.

PR #830, opened 2019-04-06
[PATCH v3] linux-gen: crypto: fix AES-GMAC with OpenSSL 1.1.1b

lumag:
OpenSSL 1.1.1b (and master branches) have changed semantics of
EVP_En/Decrypt operations by diallowing now to call EVP_Decrypt on the
context initialized with EVP_EncryptInit_ex (which was allowed before).

Fix this by calling corresponding function for data processing.

Signed-off-by: Dmitry Eremin-Solenikov <deremin-solenikov@cavium.com>

### DIFF
--- a/platform/linux-generic/odp_crypto_openssl.c
+++ b/platform/linux-generic/odp_crypto_openssl.c
@@ -621,7 +621,8 @@ odp_crypto_alg_err_t auth_cmac_check(odp_packet_t pkt,
 static
 int internal_aad(EVP_CIPHER_CTX *ctx,
 		 odp_packet_t pkt,
-		 const odp_crypto_packet_op_param_t *param)
+		 const odp_crypto_packet_op_param_t *param,
+		 odp_bool_t encrypt)
 {
 	uint32_t offset = param->auth_range.offset;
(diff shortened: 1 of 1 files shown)


platform/linux-generic/odp_crypto_openssl.c
```diff
@@ -621,7 +621,8 @@ odp_crypto_alg_err_t auth_cmac_check(odp_packet_t pkt,
 static
 int internal_aad(EVP_CIPHER_CTX *ctx,
 		 odp_packet_t pkt,
-		 const odp_crypto_packet_op_param_t *param)
+		 const odp_crypto_packet_op_param_t *param,
+		 odp_bool_t gen)
```
matiaselo: Based on the following function calls (`EVP_EncryptUpdate()`, `EVP_DecryptUpdate()`) `encrypt` (or `decrypt`) would in my opinion be a more descriptive variable name. Some specific reason to use `gen` here?

lumag: This was related to `gen`/`check` phases. I'm changing that to `encrypt`/`decrypt` correspondingly.

PR #834, opened 2019-04-16
[PATCH v2] Improve global init API specification

psavol:
Improve API specification text to be more precise e.g. about which functions may be used before init_global call. API functionality is not changed.

Also fix/improve init validation test suite. Fixed also a bug found in buffer validation test status output.

PR #835, opened 2019-04-16
[PATCH v2] api: queue context is null by default

psavol:
Improve documentation and testing of queue context being NULL by default. It was already defined under odp_queue_param_t documentation.

This PR should be merged after #834 to maintain API version number increments in order.

### DIFF
--- a/include/odp/api/spec/queue.h
+++ b/include/odp/api/spec/queue.h
@@ -116,10 +116,16 @@ int odp_queue_context_set(odp_queue_t queue, void *context, uint32_t len);
 /**
  * Get queue context
  *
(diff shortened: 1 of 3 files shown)


include/odp/api/spec/queue.h
```diff
@@ -116,6 +116,10 @@ int odp_queue_context_set(odp_queue_t queue, void *context, uint32_t len);
 /**
  * Get queue context
  *
+ * Returns previously stored queue context pointer from the application.
```
matiaselo: `from the application` part seems a bit out of place here.

psavol: Fixed.

PR #837, opened 2019-04-21
[PATCH v8] Additional crypto algorithms support

lumag:
Add support for:
 - 3DES in ECB mode
 - AES in ECB, CFB128 and XTS modes
 - AES as used by 3GPP 128-EEA2/128-EIA2 algorithms


include/odp/api/spec/crypto.h
```diff
@@ -262,6 +262,24 @@ typedef enum {
 	 */
 	ODP_AUTH_ALG_ZUC_EIA3,
 
+	/** MD5 algorithm */
+	ODP_AUTH_ALG_MD5,
```
Bill-Fischofer-Personal: MD5 has been deprecated for some years now. Is this being added for historical reasons?

mvl-skardach: It has been though there is still hardware supporting it. For the sake of legacy systems.

include/odp/api/spec/crypto.h
```diff
@@ -154,6 +170,13 @@ typedef enum {
 	 */
 	ODP_AUTH_ALG_SHA1_HMAC,
 
+	/** HMAC-SHA-224
+	 *
+	 *  SHA-224 algorithm in HMAC mode
+	 */
+
```
psavol: Remove this extra new line

psavol:
Since these API changes are backward compatible, we can merge them to master but version number needs to be incremented (in configure.ac). Add a commit to the patch set to increment it.

E.g.: api: update version number to 1.21.5

include/odp/api/spec/crypto.h
lumag: Dine

psavol:
Build fails in Travis:

  CC       odp_weak.lo
  CC       odp_crypto_openssl.lo
odp_crypto_openssl.c: In function '_odp_crypto_init_local':
odp_crypto_openssl.c:2554:21: error: implicit declaration of function 'EVP_MD_CTX_new' [-Werror=implicit-function-declaration]
   local.md_ctx[i] = EVP_MD_CTX_new();

Likely the openssl version is older in Travis and some openssl version ifdefs needs to be added.

lumag:
@psavol should be fixed now.

psavol:
Agree, all checkpatch warnings except CamelCase ones should be easy to fix.

Also, the ODP API version number increment patch is needed.

lumag:
@matiaselo @psavol 
- Fixed all checkpatch issue except CamelCase and another CHECK issue, where checkpatch incorrectly parses variable definition.
- Added API version bump

include/odp/api/spec/crypto.h
```diff
@@ -76,12 +76,25 @@ typedef enum {
...
 	ODP_CIPHER_ALG_AES_CBC,
 
 	/** AES with counter mode */
 	ODP_CIPHER_ALG_AES_CTR,
 
+	/** AES with electronic codebook */
+	ODP_CIPHER_ALG_AES_ECB,
+
+	/** AES with 128-bit cipher feedback */
+	ODP_CIPHER_ALG_AES_CFB128,
```
psavol: The shift size could be added into session parameters (like digest_len and aad_len). This would avoid later enum additions for different shift sizes. How likely it is that e.g. 1 or 8 bit shift needs to be supported, or other chiphers would be used in CFB mode?

E.g. openSSL seems to have 1, 8 and 128 bit shift sizes supported.

EVP_aes_128_cfb(), EVP_aes_192_cfb(), EVP_aes_256_cfb(), EVP_aes_128_cfb1(), EVP_aes_192_cfb1(), EVP_aes_256_cfb1(), EVP_aes_128_cfb8(), EVP_aes_192_cfb8(), EVP_aes_256_cfb8(), EVP_aes_128_cfb128(), EVP_aes_192_cfb128(), EVP_aes_256_cfb128()

include/odp/api/spec/crypto.h
```diff
@@ -127,6 +140,9 @@ typedef enum {
 	 */
 	ODP_CIPHER_ALG_ZUC_EEA3,
 
+	/** AES_CTR in EEA2 mode */
+	ODP_CIPHER_ALG_AES_CTR_EEA2,
```
psavol: This should be aligned with current 128-EEA1 and 128-EEA3 definitions:
1) It would be better to have it defined between ODP_CIPHER_ALG_SNOW3G_UEA2 and ODP_CIPHER_ALG_ZUC_EEA3
2) CTR is not needed in the name as EEA2 spec define the AES mode: better name is  ODP_CIPHER_ALG_AES_EEA2
3) Mention 128-EEA2 in comment

ODP is maintaining the counter value, which includes five "bearer" bits. How application passes bearer to the  crypto operation / session ?

include/odp/api/spec/crypto.h
```diff
@@ -252,6 +274,27 @@ typedef enum {
 	 */
 	ODP_AUTH_ALG_ZUC_EIA3,
 
+	/** AES_CMAC in EIA2 mode */
+	ODP_AUTH_ALG_AES_CMAC_EIA2,
```
psavol: Similar comment:
1) Move before ODP_AUTH_ALG_ZUC_EIA3
2) CMAC not needed in enum name: ODP_AUTH_ALG_AES_EIA2 is enough
3) Mention 128-EIA2 in comment (similarly to 128-EIA3 reference of ZUC)

There's bug in current documentation of ODP_AUTH_ALG_SNOW3G_UIA2: 128-EEA1 in heading vs. 128-EIA1 in body text.

include/odp/api/spec/crypto.h
lumag: Shift size is quite unique to cfb mode. It takes only several fixed values. So I do not think that it makes sense to move it to session params.

include/odp/api/spec/crypto.h
lumag: Ok

Counter and bearer are passed as pre-formatted IV blocks. If we need higher level API, it can be created on top crypto API. Do you have any suggestions on how this high level API should look?

include/odp/api/spec/crypto.h
lumag: Ok

lumag:
Updated pr to reflect review.

include/odp/api/spec/crypto.h
psavol: OK.

include/odp/api/spec/crypto.h
psavol: API spec needs some specification about how bearer is passed / IV is used.

We can refer to 3GPP spec (33.401: chapter B.1.3 128-EEA2 ?) format. According to that CTR = COUNT[0..31] | BEARER[0..4] | DIRECTION | 26 bits zero | 64bit block counter. User would need to pass COUNT and BEARER for every operation through IV.

Other option would be to add COUNT/BEARER to crypto session params so that ODP would update/fill those automatically.

lumag: @psavol does the following sound good to you:

```
For 3GPP algorithms users of crypto API should format IV blocks by hand according to the 3GPP specificatons using COUNTER/BEARER/DIRECTION.
```

lumag:
For the reference. NXP had the following API addition:
```c
/**
 * PDCP CONFIG API session parameters
 */
typedef struct odp_pdcp_params_s {
        int8_t bearer;  /**< PDCP bearer ID */
        int8_t pkt_dir; /**< PDCP Frame Direction 0:UL 1:DL*/
        int8_t hfn_ovd; /**< Overwrite HFN per operation.
                         * Use override_iv_ptr for overriding*/
        uint8_t sn_size; /**< Sequence number size, 5/7/12/15 */
        uint32_t hfn;   /**< Hyper Frame Number */
        uint32_t hfn_threshold; /**< HFN Threashold for key renegotiation */
} odp_pdcp_params_t;

int odp_crypto_session_config_pdcp(odp_crypto_session_t session,
                                   odp_pdcp_mode_t pdcp_mode,
                                    odp_pdcp_params_t* pdcp_params);
```

I think that reusing crypto sessions for PDCP is incorrect. I'd suggest to use their definition of `odp_pdcp_params_t, extend it with cipher/auth algorithms` and add `odp_pdcp_session_create()/destroy()/op()/op_enq()` on top of it.

include/odp/api/spec/crypto.h
psavol: Maybe a bit more specific:

"3GPP algorithms require initialization vector to be formatted as specified in 3GPP standards (see e.g. TS 33.401). COUNT/BEARER/DIRECTION information needs to be packed into IV bits as standards specify." 

Then also specification if override or session IV may be used.
"128-EEA2 require IV to be set per operation (due to unique COUNT value)"

Did not check it, but may be e.g. ZUC can be used also with session IV, if the format does not have a counter.

lumag: All 3gpp algorithms use `COUNT`/`BEARER`/`DIRECTION` (and some of them require random called `FRESH`.

lumag: @psavol I ended up describing IV format for all 8 algorithms.

psavol:
Patch set is otherwise OK, but it seems that "bump ODP API version" is not the last patch. Please move it to the last, if that's not the case. Also add my reviewed-by tags and submit v8.

psavol:
I think IV content documentation is enough at this point. PDCP specific parameters/functions etc could follow later.

lumag:
> Patch set is otherwise OK, but it seems that "bump ODP API version" is not the last patch. Please move it to the last, if that's not the case. Also add my reviewed-by tags and submit v8.

Done, thank you.

PR #841, opened 2019-05-07
[PATCH v3] New helpers for thread create and join

psavol:
Current thread create and join helpers are not easy to use when threads are created/joined in multiple steps. API does not specify if it's possible and implementation is based on cpu count in cpumask and hidden "last" flag.

Added new implementations, with improved specification, parameter set and new features. Added also version number support for helper library, so that helper API changes are easy to track from application.

Old create/join helpers are not modified. Some test applications (incl. all validation test) was already moved to use the new functions, the rest will be moved soon.

### DIFF
--- a/helper/include/odp/helper/threads.h
+++ b/helper/include/odp/helper/threads.h
@@ -54,18 +54,26 @@ typedef struct {
 	int   status;   /**< Process state change status */
 } odph_linux_process_t;
 
-/** odpthread parameters for odp threads (pthreads and processes) */
+/** Thread parameters (pthreads and processes) */
 typedef struct {
-	int (*start)(void *);       /**< Thread entry point function */
-	void *arg;                  /**< Argument for the function */
-	odp_thread_type_t thr_type; /**< ODP thread type */
-	odp_instance_t instance;    /**< ODP instance handle */
-} odph_odpthread_params_t;
+	/** Thread entry point function */
+	int (*start)(void *arg);
+
+	/** Argument for the function */
+	void *arg;
+
+	/** ODP thread type */
+	odp_thread_type_t thr_type;
+
+	/** @deprecated ODP instance handle for odph_odpthreads_create(). */
+	odp_instance_t instance;
+
+} odph_thread_param_t;
 
 /** The odpthread starting arguments, used both in process or thread mode */
 typedef struct {
 	odp_mem_model_t mem_model;          /**< process or thread */
-	odph_odpthread_params_t thr_params; /**< odpthread start parameters */
+	odph_thread_param_t thr_params; /**< odpthread start parameters */
 } odph_odpthread_start_args_t;
 
 /** Linux odpthread state information, used both in process or thread mode */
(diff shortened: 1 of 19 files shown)


helper/include/odp/helper/threads.h
```diff
@@ -86,13 +94,127 @@ typedef struct {
...
+ * may be overridden with ODP helper options. See e.g. --odph_proc under
+ * odph_options() documentation.
+ *
+ * Thread creation may be synchronized by setting 'sync' parameter. It
+ * serializes thread start up (odp_init_local() calls), which helps to
+ * stabilize application start up sequence.
+ *
+ * By default, the thread parameter table contains 'num' elements, one for
+ * each thread to be created. However, all threads may be created
+ * with the same parameters by setting 'share_param' parameter.
```
mvl-skardach: It is a little unclear how many elements should `thr_param[]` have in `share_param` case. I presume the answer is 1.

helper/include/odp/helper/threads.h
```diff
@@ -86,13 +94,127 @@ typedef struct {
...
+		       const odph_thread_common_param_t *param,
+		       const odph_thread_param_t thr_param[],
+		       int num);
+
+/**
+ * Wait previously launched threads to exit
+ *
+ * This is an updated version of odph_odpthreads_join() call. It waits for
+ * threads launched with odph_thread_create() to exit. It may be called
+ * multiple times to wait thread exits in steps.
```
mvl-skardach: Is it allowed to wait on the same thread multiple times (as pthread allows)? I'm fine with both answers. :)

helper/include/odp/helper/threads.h
```diff
@@ -118,6 +242,7 @@ int odph_odpthreads_create(odph_odpthread_t *thread_tbl,
  * (error occurs if any of the start_routine return non-zero or if
  *  the thread join/process wait itself failed -e.g. as the result of a kill)
  *
+ * @note Will be deprecated. Use odph_thread_join() instead.
```
mvl-skardach: I think we should use `@deprecated`: http://www.doxygen.nl/manual/commands.html#cmddeprecated

helper/include/odp/helper/threads.h
```diff
@@ -103,6 +225,8 @@ typedef struct {
  * @param thr_params    ODP thread parameters
  *
  * @return Number of threads created
+ *
+ * @note Will be deprecated. Use odph_thread_create() instead.
```
mvl-skardach: Ditto.

helper/include/odp/helper/threads.h
psavol: Fixed in v2. It was documented in 'share_param' documentation, but now also here.

helper/include/odp/helper/threads.h
psavol: Fixed in v2. A thread can be waited only once.

helper/include/odp/helper/threads.h
psavol: Fixed in v2. These helpers are documented now as deprecated, instead of "will be deprecated".

helper/include/odp/helper/threads.h
psavol: Fixed in v2.

PR #846, opened 2019-05-16
[PATCH v6] Timer and time API additions

psavol:
Clarifies current timer API functionality (set/reset operation and event type), adds timer resolution capability and adds time defines for minute/hour.

Timer validation test of minimum timeout needed fixing posix timer overrun issue in timer pool start. It seems Linux (scheduler) has about 15ms latency before the first posix timer signal is received on the timer pthread.

### DIFF
--- a/include/odp/api/spec/timer.h
+++ b/include/odp/api/spec/timer.h
@@ -69,26 +69,29 @@ typedef enum {
  * Return values of timer set calls.
  */
 typedef enum {
-/**
- * Timer set operation succeeded
- */
+	/**
+	 * Timer set operation succeeded
+	 */
 	ODP_TIMER_SUCCESS = 0,
-/**
- * Timer set operation failed, expiration too early.
- * Either retry with a later expiration time or process the timeout
- * immediately. */
+
+	/**
+	 * Timer set operation failed, expiration too early.
+	 * Either retry with a later expiration time or process the timeout
+	 * immediately. */
 	ODP_TIMER_TOOEARLY = -1,
 
-/**
- * Timer set operation failed, expiration too late.
- * Truncate the expiration time against the maximum timeout for the
- * timer pool. */
+	/**
+	 * Timer set operation failed, expiration too late.
+	 * Truncate the expiration time against the maximum timeout for the
+	 * timer pool. */
 	ODP_TIMER_TOOLATE = -2,
-/**
- * Timer set operation failed because no event specified and no event present
- * in the timer (timer inactive/expired).
- */
+
+	/**
+	 * Timer set operation failed because no event specified and no event
+	 * present in the timer (timer inactive/expired).
+	 */
 	ODP_TIMER_NOEVENT = -3
+
 } odp_timer_set_t;
 
 /**
@@ -96,8 +99,8 @@ typedef enum {
  * Maximum timer pool name length in chars including null char
  */
 
-/** Timer pool parameters
- * Timer pool parameters are used when creating and querying timer pools.
+/**
+ * Timer pool parameters
  */
 typedef struct {
 	/** Timeout resolution in nanoseconds. Timer pool must serve timeouts
@@ -187,9 +190,8 @@ int odp_timer_capability(odp_timer_clk_src_t clk_src,
  * @return Timer pool handle on success
  * @retval ODP_TIMER_POOL_INVALID on failure and errno set
  */
-odp_timer_pool_t
-odp_timer_pool_create(const char *name,
-		      const odp_timer_pool_param_t *params);
+odp_timer_pool_t odp_timer_pool_create(const char *name,
+				       const odp_timer_pool_param_t *params);
 
 /**
  * Start a timer pool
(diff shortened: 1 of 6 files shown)


include/odp/api/spec/timer.h
```diff
@@ -287,80 +361,94 @@ odp_timer_t odp_timer_alloc(odp_timer_pool_t tpid,
...
+ * reset operation with a new event returns the old event.
  *
- * @param tim      Timer
- * @param abs_tck  Expiration time in absolute timer ticks
- * @param[in,out] tmo_ev  Reference to an event variable that points to
- * timeout event or NULL to reuse the existing timeout event. Any existing
- * timeout event that is replaced by a successful set operation will be
- * returned here.
+ * The user provided event can be of any event type, but only ODP_EVENT_TIMEOUT
+ * type events (odp_timeout_t) carry timeout specific metadata. Also timer
```
matiaselo: `Also timer...`, missing comma or replace with for example` Furthermore,`.

include/odp/api/spec/timer.h
```diff
@@ -173,6 +220,28 @@ typedef struct {
...
 			 odp_timer_capability_t *capa);
 
+/**
+ * Timer resolution capability
+ *
+ * This function fills in capability limits for timer pool resolution and
+ * min/max timeout values, based on either resolution or maximum timeout.
+ * Set the required value to 'res_ns' or 'max_tmo', and set other fields to
+ * zero. A successful call fills in the other two fields. The call returns
+ * a failure, if the user defined value exceeds (max_res or max_tmo) capability
```
matiaselo: Move `(max_res or max_tmo)` before exceeds and add missing apostrophes.

include/odp/api/spec/timer.h
```diff
@@ -287,80 +361,94 @@ odp_timer_t odp_timer_alloc(odp_timer_pool_t tpid,
...
- * late
- * @retval ODP_TIMER_NOEVENT Operation failed because timeout event not
- * specified in odp_timer_set call and not present in timer
+ * @param         timer    Timer
+ * @param         abs_tick Absolute expiration time in timer ticks
+ * @param[in,out] tmo_ev   Pointer to an event handle. The event is enqueued
+ *                         when the timer expires. Use NULL when resetting the
+ *                         timer without changing the event. When resetting the
+ *                         timer with a new event, a successful operation
+ *                         returns the old event here.
```
matiaselo: Perhaps replace `returns` with stores/outputs.

include/odp/api/spec/timer.h
```diff
@@ -287,80 +361,94 @@ odp_timer_t odp_timer_alloc(odp_timer_pool_t tpid,
...
  *
- * Set (arm) the timer to expire at a relative future time.
+ * @param         timer    Timer
+ * @param         rel_tick Expiration time relative to current time of
+ *                         the timer pool in timer ticks
+ * @param[in,out] tmo_ev   Pointer to an event handle. The event is enqueued
+ *                         when the timer expires. Use NULL when resetting the
+ *                         timer without changing the event. When resetting the
+ *                         timer with a new event, a successful operation
+ *                         returns the old event here.
```
matiaselo: Perhaps replace `returns` with stores/outputs.

include/odp/api/spec/timer.h
psavol: Fixed

include/odp/api/spec/timer.h
psavol: Fixed

include/odp/api/spec/timer.h
psavol: Fixed

include/odp/api/spec/timer.h
psavol: Fixed

psavol:
v3: rebased

include/odp/api/spec/timer.h
```diff
@@ -294,73 +296,86 @@ odp_timer_t odp_timer_alloc(odp_timer_pool_t tpid,
...
- * late
- * @retval ODP_TIMER_NOEVENT Operation failed because timeout event not
- * specified in call and not present in timer
+ * @retval ODP_TIMER_SUCCESS  Success
+ * @retval ODP_TIMER_TOOEARLY Failure. Expiration time is too near to
+ *                            the current time.
+ * @retval ODP_TIMER_TOOLATE  Failure. Expiration time is too far from
+ *                            the current time.
+ * @retval ODP_TIMER_NOEVENT  Failure. Set operation: no event provided.
+ *                            Reset operation: timer already expired.
```
mvl-skardach: I'm not sure I understand the `ODP_TIMER_NOEVENT` case with Reset operation. In this case user gives a "valid" `rel_tick` but during the set/reset operation timer fires and therefore there was no event to "reset"?
Wouldn't that normally mean that platform could simply fire the new event and old event won't be available?
If "inserting" and event into the timer would mean software blocking timer HW from working/ticking, it means that SW will have to take care of any timers that have fired while it processes (i.e. our platform goes through a current bucket and manually fires all events).

That would be a "success" case since the old event is already pushed on a queue.

If that is a failure case then is my assumption correct that implementation should NOT arm the timer?

mvl-skardach:
I think I screwed up the "approval" label. I thought it was concerning a particular commit, not all commits. So please discard it if possible. :)

include/odp/api/spec/timer.h
psavol: Yes. In reset case, functions arguments were OK (expiration tick value and event), but reset is not possible as the timer cannot be modified any more (has fired already or will fire soon). Reset operation failed, but timer has fired (will fire soon) with the previous settings. The "old" event will be received normally through the queue and the new event is not consumed. So, implementation does not have to do anything extra if it was too late to reset HW timer, just return this failure and the old event is processed normally. Timeout can be marked stale, if there's no danger of race condition (should not write to a memory location that is not any more owned by the implementation).

mvl-skardach: Then perhaps it would be a bit clearer to say "timer expired during the call".

psavol: The timer may have expired already, or it expires soon (e.g. 1us) enough that it cannot be modified anymore. So, it may not exactly expire during the call...

mvl-skardach: Fair enough. Then please change this to "Too late to modify the timer".

psavol: Corrected in v4. Also I added this to function description: "A failed reset operation does not modify the timer."

PR #853, opened 2019-06-20
[PATCH v4] linux-gen: timer: don't reserve any resources when timers are disabled

matiaselo:
Don't reserve any memory or perform initialization if timer feature
(odp_feature_t.feat.timer) is marked as not used.

### DIFF
--- a/platform/linux-generic/odp_timer.c
+++ b/platform/linux-generic/odp_timer.c
@@ -1413,6 +1413,12 @@ int odp_timer_init_global(const odp_init_t *params)
 	const char *conf_str;
 	int val = 0;
 
+	if (params && params->not_used.feat.timer) {
+		ODP_DBG("Timers disabled\n");
+		timer_global = NULL;
+		return 0;
+	}
+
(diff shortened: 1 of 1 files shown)


platform/linux-generic/odp_timer.c
```diff
@@ -1413,6 +1413,11 @@ int odp_timer_init_global(const odp_init_t *params)
 	const char *conf_str;
 	int val = 0;
 
+	if (params && params->not_used.feat.timer) {
+		ODP_DBG("Timers disabled\n");
+		return 0;
```
Bill-Fischofer-Personal: Since `odp_timer_term_global()` assumes that `timer_global == NULL` if timers are not being used, you should explicitly set `timer_global` to `NULL` here to avoid problems if an ODP instance with timers is followed by an ODP instance without timers.

matiaselo: Thanks, will fix. Thinking about it we may have the same assumption (static pointers initialized always to NULL) also in other places.

matiaselo:
Is this PR now OK for you @Bill-Fischofer-Personal?

PR #854, opened 2019-06-24
[PATCH v4] Don't set -march=native when user has set CFLAGS

psavol:
User may set -march option with CFLAGS. Don't set -march when user has defined CFLAGS.

### DIFF
--- a/configure.ac
+++ b/configure.ac
@@ -34,6 +34,13 @@ AC_SUBST(ODPH_VERSION_MAJOR)
 ODPH_VERSION_MINOR=odph_version_minor
 AC_SUBST(ODPH_VERSION_MINOR)
 
+##########################################################################
+# Test if user has set CFLAGS. Automake initializes CFLAGS to "-g -O2"
+# by default.
+##########################################################################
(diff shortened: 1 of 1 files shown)


configure.ac
```diff
@@ -34,6 +34,20 @@ AC_SUBST(ODPH_VERSION_MAJOR)
...
+# automake will initialize CFLAGS to "-g O2", and it would not be possible
+# to check if CFLAGS are user defined.
+##########################################################################
+: ${CFLAGS=""}
+if test -z "$CFLAGS"; then
+	user_cflags=0
+	CFLAGS="-O2 -g"
+else
+	user_cflags=1
+fi
```
lumag: `AS_IF([test "$ac_cv_env_CFLAGS_set" = ""], [...], [...])`

lumag: Then you don't even have to set `CFLAGS` to `-g -O2`

psavol: Fixed and added your review tag in v4. Thanks.

lumag: Thank you.

PR #857, opened 2019-07-04
[PATCH v2] New options to sched_perf test

psavol:
Group options -g and -j manage how many groups are created and how thread join into those groups. Wait time option -w specifies how much time application spends between schedule and queue enqueue calls.

### DIFF
--- a/test/performance/odp_sched_perf.c
+++ b/test/performance/odp_sched_perf.c
@@ -43,17 +43,23 @@ typedef struct test_stat_t {
 
 } test_stat_t;
 
+typedef struct thread_arg_t {
+	void *global;
+	int first_group;
+
+} thread_arg_t;
+
 typedef struct test_global_t {
 	test_options_t test_options;
-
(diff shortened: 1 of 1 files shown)


test/performance/odp_sched_perf.c
```diff
@@ -150,6 +169,18 @@ static int parse_options(int argc, char *argv[], test_options_t *test_options)
 		ret = -1;
 	}
 
+	if (test_options->num_group > MAX_GROUPS) {
+		printf("Error: Too many groups. Max supported %i\n.",
+		       MAX_GROUPS);
+		ret = -1;
+	}
+
+	if (test_options->num_join > test_options->num_group) {
```
mvl-skardach: I think we need another warning on: `if (test_options->num_join * test_options->num_cpu < test_options->num_group)`.

psavol: Fixed. Now warns if some groups are not served, and in addition errors if forwarding was set.

PR #871, opened 2019-09-17
[PATCH v2] Use helper library debug macros

matiaselo:
Use ODP helper library debug macros instead of reimplementing the same functionality in separate application headers.

Duplicate ODPH_UNUSED macro has been removed. ODP_UNUSED should be used instead.


configure.ac
```diff
@@ -22,7 +22,7 @@ AC_SUBST(ODP_VERSION_API_MINOR)
 ##########################################################################
 m4_define([odph_version_generation], [1])
 m4_define([odph_version_major], [0])
-m4_define([odph_version_minor], [2])
+m4_define([odph_version_minor], [3])
```
muvarov: version change looks strange here.

matiaselo: The duplicate ODPH_UNUSED macro was removed; hence the version change.

PR #876, opened 2019-10-03
[PATCH v1] Fix build with GCC 9

psavol:
GCC 9 includes a new warning (-Waddress-of-packed-member). Handle it only as a warning, not as an error. Enabled struct initializations warnings again and fixed some struct init issues that an old compiler complained about.

### DIFF
--- a/example/ipfragreass/odp_ipfragreass.c
+++ b/example/ipfragreass/odp_ipfragreass.c
@@ -230,7 +230,7 @@ int main(void)
 	odp_pool_t fragment_pool;
 	odp_shm_t shm;
 	odp_cpumask_t cpumask;
-	odph_odpthread_t threads[MAX_WORKERS] = {};
+	odph_odpthread_t threads[MAX_WORKERS];
 	odph_odpthread_params_t thread_params;
 	odp_packet_t dequeued_pkts[NUM_PACKETS];
 	odp_event_t ev;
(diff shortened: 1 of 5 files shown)


test/validation/api/ipsec/ipsec_test_out.c
```diff
@@ -45,16 +45,17 @@ static void test_out_ipv4_ah_sha256(void)
 
 static void test_out_ipv4_ah_sha256_tun_ipv4(void)
 {
-	uint32_t src = IPV4ADDR(10, 0, 111, 2);
-	uint32_t dst = IPV4ADDR(10, 0, 222, 2);
-	odp_ipsec_tunnel_param_t tunnel = {
```
mvl-skardach: I'm curious - what was the compiler error message here?

example/ipfragreass/odp_ipfragreass.c
```diff
@@ -230,7 +230,7 @@ int main(void)
 	odp_pool_t fragment_pool;
 	odp_shm_t shm;
 	odp_cpumask_t cpumask;
-	odph_odpthread_t threads[MAX_WORKERS] = {};
+	odph_odpthread_t threads[MAX_WORKERS];
```
mvl-skardach: What is the added value of removing this warning? It is rather overly paranoid.

test/validation/api/ipsec/ipsec_test_out.c
psavol: Missing initialization on some fields (-Wmissing-field-initializers). The warning was disabled and I enable it again in this patch set, which causes an older version of GCC in CentOS 7 to fail here.

example/ipfragreass/odp_ipfragreass.c
psavol: These also resulted -Wmissing-field-initializers

PR #878, opened 2019-10-07
[PATCH v2] New odp_ping example application

psavol:
Application replies to ICMPv4 ping requests. Sender side ARP table needs to be setup manually as the application does not reply to ARP.


matiaselo:
Please add a short application description to examples.dox.

PR #886, opened 2019-10-23
[PATCH v1] travis: add em-odp build test

matiaselo:
Add EM-ODP project build test. The test is allowed to fail since ODP API
changes have to be ported into EM-ODP project.


mvl-skardach:
Seems rather strange to put this here. EM is dependent on ODP, not the other way around.

matiaselo:
> Seems rather strange to put this here. EM is dependent on ODP, not the other way around.

What I'm trying to achieve here is to get an early warning of changes breaking EM-ODP integration, so we can warn the EM guys beforehand or avoid problems altogether.

I'm also adding build tests to the EM-ODP repo.

PR #889, opened 2019-10-31
[PATCH v2] Timer control thread poll configure option

psavol:
Added configure file option for inline timer (inline_thread_type) to select if control threads poll shared timer pools or not. When control threads still poll their private pools. With this user can configure control and worker threads to use separate timer pools.

### DIFF
--- a/config/odp-linux-generic.conf
+++ b/config/odp-linux-generic.conf
@@ -16,7 +16,7 @@
 
 # Mandatory fields
 odp_implementation = "linux-generic"
-config_file_version = "0.1.9"
+config_file_version = "0.1.10"
 
 # Shared memory options
 shm: {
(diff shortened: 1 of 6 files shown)


config/odp-linux-generic.conf
```diff
@@ -168,6 +171,17 @@ timer: {
...
+	# Inline timer use of threads
+	#
+	# Select which thread types process non-private timer pools in inline
+	# timer implementation. Thread type does not affect private timer
+	# pool procesessing, those are always processed by the thread which
+	# created the pool. Ignored when inline timer is not used.
+	#
+	# 0: Both control and worker threads process non-private timer pools
+	# 1: Only worker threads process non-private timer pools
+	inline_thread_type = 0
```
matiaselo: You could also add  option e.g.:
# 2: Only control threads process non-private timer pools

PR #891, opened 2019-11-01
[PATCH v3] validation: scheduler: add scheduled and plain queue tests

matiaselo:
Add new validation tests to verify that packet order is maintained when
events are enqueued and dequeued to/from a plain queue while maintaining
atomic/ordered scheduling context.

### DIFF
--- a/test/validation/api/scheduler/scheduler.c
+++ b/test/validation/api/scheduler/scheduler.c
@@ -55,6 +55,8 @@
 #define WAIT_TOLERANCE (150 * ODP_TIME_MSEC_IN_NS)
 #define WAIT_1MS_RETRIES 1000
 
+#define SCHED_AND_PLAIN_ROUNDS 10000
+
 /* Test global variables */
 typedef struct {
 	int num_workers;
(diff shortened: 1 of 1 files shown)


test/validation/api/scheduler/scheduler.c
```diff
@@ -1780,6 +1786,228 @@ static void scheduler_test_ordered_lock(void)
...
+		if (buf2 == ODP_BUFFER_INVALID) {
+			odp_buffer_free(buf1);
+			break;
+		}
+		buf_data = odp_buffer_addr(buf1);
+		*buf_data = seq;
+		buf_data = odp_buffer_addr(buf2);
+		*buf_data = seq;
+
+		CU_ASSERT_FATAL(!odp_queue_enq(sched_queue,
```
mvl-skardach: Maybe important comment to add here is that this test depends on all events being of the same flow in case of a multi-flow scheduler.

matiaselo: The flow id is zero by default, so multi-flow scheduling has no effect here.

mvl-skardach: I know it has but it might be not so obvious to a person just reading the unit test. If you think the implied 0 is enough then please feel free to discard my comment.

matiaselo: Okay, we are currently not referencing to multi-flow aware scheduler anywhere else except scheduler_test_flow_aware(), so adding the comment felt a bit out of place. But you gotta start improvements from somewhere, so I'll add a comment here.

PR #896, opened 2019-11-08
[PATCH v2] test: bench_packet: add tests for missing packet functions

matiaselo:
Added tests for the following packet functions:
	odp_packet_data_seg_len()
	odp_packet_free_sp()
	odp_packet_from_event_multi()
	odp_packet_to_event_multi()
	odp_packet_subtype()
	odp_packet_parse()
	odp_packet_parse_multi()

### DIFF
--- a/test/performance/odp_bench_packet.c
+++ b/test/performance/odp_bench_packet.c
@@ -17,6 +17,8 @@
 #include <inttypes.h>
 #include <signal.h>
 
+#include <test_packet_parser.h>
+
 #include <odp_api.h>
 #include <odp/helper/odph_api.h>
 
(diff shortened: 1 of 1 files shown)


test/performance/odp_bench_packet.c
```diff
@@ -587,6 +713,18 @@ static int bench_packet_data(void)
...
 }
 
+static int bench_packet_data_seg_len(void)
+{
+	odp_packet_t *pkt_tbl = gbl_args->pkt_tbl;
+	uint32_t *output_tbl = gbl_args->output_tbl;
+	int i;
+
+	for (i = 0; i < TEST_REPEAT_COUNT; i++)
+		gbl_args->ptr_tbl[i] = odp_packet_data_seg_len(pkt_tbl[i],
```
psavol: Better to increment a length variable to avoid extra data movement, like packet len:
ret += odp_packet_len()

matiaselo: odp_packet_data_seg_len() returns a data pointer, not the length.

PR #899, opened 2019-11-13
[PATCH v3] api: compression feature disable

psavol:
Add compression feature flag into API. Update feature disable implementation. Disabling crypto, ipsec, compression and traffic manager when not used saves over 20MB of shared memory.

### DIFF
--- a/platform/linux-generic/include/odp_global_data.h
+++ b/platform/linux-generic/include/odp_global_data.h
@@ -39,7 +39,7 @@ typedef struct {
 
 /* Read-only global data. Members should not be modified after global init
  * to enable process more support. */
-struct odp_global_data_ro_t {
+typedef struct odp_global_data_ro_t {
 	odp_init_t init_param;
 	/* directory for odp mmaped files */
 	char *shm_dir;
@@ -60,17 +60,19 @@ struct odp_global_data_ro_t {
 	config_t libconfig_default;
 	config_t libconfig_runtime;
 	odp_random_kind_t ipsec_rand_kind;
-};
+
+} odp_global_data_ro_t;
 
 /* Modifiable global data. Memory region is shared and synchronized amongst all
  * worker processes. */
-struct odp_global_data_rw_t {
+typedef struct odp_global_data_rw_t {
 	odp_bool_t dpdk_initialized;
 	odp_bool_t inline_timers;
-};
 
-extern struct odp_global_data_ro_t odp_global_ro;
-extern struct odp_global_data_rw_t *odp_global_rw;
+} odp_global_data_rw_t;
+
+extern odp_global_data_ro_t odp_global_ro;
+extern odp_global_data_rw_t *odp_global_rw;
 
 #ifdef __cplusplus
 }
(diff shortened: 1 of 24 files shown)


platform/linux-generic/odp_init.c
```diff
@@ -45,8 +45,8 @@ enum init_stage {
 	ALL_INIT      /* All init stages completed */
 };
 
-struct odp_global_data_ro_t odp_global_ro;
-struct odp_global_data_rw_t *odp_global_rw;
+odp_global_data_ro_t odp_global_ro;
+odp_global_data_rw_t *odp_global_rw;
```
matiaselo: These global variables should be prefixed with underscore.

platform/linux-generic/include/odp_global_data.h
```diff
@@ -61,6 +61,15 @@ typedef struct odp_global_data_ro_t {
 	config_t libconfig_runtime;
 	odp_random_kind_t ipsec_rand_kind;
 
+	/* Disabled features during global init */
+	struct {
+		uint8_t compress;
+		uint8_t crypto;
+		uint8_t ipsec;
+		uint8_t traffic_mng;
```
matiaselo: Perhaps tm or traffic_mngr would be more consistent with the current internal naming.

platform/linux-generic/odp_comp.c
```diff
@@ -610,11 +610,17 @@ int _odp_comp_init_global(void)
 	odp_shm_t shm;
 	int idx;
 
+	if (odp_global_ro.disable.compress) {
```
matiaselo: Missing odp_global_data.h include.

platform/linux-generic/odp_init.c
psavol: It's better to do variable renaming after this as touches many files not related to this patch set.

platform/linux-generic/odp_traffic_mngr.c
```diff
@@ -2918,7 +2918,7 @@ odp_tm_t odp_tm_create(const char            *name,
 	uint32_t max_tm_queues, max_sorted_lists;
 	int rc;
 
-	if (odp_global_ro.init_param.not_used.feat.tm) {
+	if (odp_global_ro.disable.traffic_mng) {
```
matiaselo: Missing odp_global_data.h include.

platform/linux-generic/include/odp_global_data.h
psavol: OK

platform/linux-generic/odp_init.c
```diff
@@ -48,6 +48,27 @@ enum init_stage {
 odp_global_data_ro_t odp_global_ro;
 odp_global_data_rw_t *odp_global_rw;
 
+static int disable_features(odp_global_data_ro_t *global_ro,
```
matiaselo: Return value not needed.

platform/linux-generic/odp_comp.c
psavol: OK

platform/linux-generic/odp_traffic_mngr.c
psavol: OK

platform/linux-generic/odp_init.c
psavol: OK

PR #907, opened 2019-12-05
[PATCH v2] example: timer_acc: add burst gap option

psavol:
Added option (-g) to control offset (gap) between timers within
a burst. This way many periodic (restarted) timers can be tested
with different expiration times.

Signed-off-by: Petri Savolainen <petri.savolainen@nokia.com>

### DIFF
--- a/example/timer/odp_timer_accuracy.c
+++ b/example/timer/odp_timer_accuracy.c
@@ -73,6 +73,7 @@ typedef struct test_global_t {
 	uint64_t         start_tick;
 	uint64_t         start_ns;
 	uint64_t         period_tick;
+	odp_shm_t        log_shm;
 	test_log_t	*log;
 	FILE            *file;
 	char		 filename[MAX_FILENAME + 1];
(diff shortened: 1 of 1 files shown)


example/timer/odp_timer_accuracy.c
```diff
@@ -691,13 +692,22 @@ int main(int argc, char *argv[])
...
-			printf("Test log calloc failed.\n");
+		shm = odp_shm_reserve("timer_accuracy_log", size,
+				      sizeof(test_log_t), 0);
+
+		if (shm == ODP_SHM_INVALID) {
+			printf("Test log alloc failed.\n");
 			goto quit;
 		}
+
+		addr = odp_shm_addr(shm);
```
matiaselo: Since this is an example application, I would still check the return value here.

psavol: Shm handle is checked before the call, and memset(0) after the call crashes if some address of the memory block  cannot be written. So, I didn't think that addr != NULL check would have a benefit here.

matiaselo: Yes, the check is rather unnecessary here. I just had a faint memory that we previously changed all example applications to check odp_shm_addr() return value, since according to api it may fail. I'm good either way.

psavol: I think those apps didn't check either shm handle or pointer. I think it's enough to check handle or pointer, but no need to check both.

matiaselo: OK

PR #908, opened 2019-12-10
[PATCH v1] test/validation/scheduler: fix plain+sched test

mvl-skardach:
Whenever order locks are used, user should ensure that he follows the
same sequence of ordered locks when processing events that have to be
ordered. That includes a drain loop when scheduler is paused.


mvl-skardach:
Wrong patch, sorry.

PR #909, opened 2019-12-10
[PATCH v3] test/validation/scheduler: fix plain+sched test

mvl-skardach:
Whenever order locks are used, user should ensure that he follows the
same sequence of ordered locks when processing events that have to be
ordered. That includes a drain loop when scheduler is paused.

### DIFF
--- a/test/validation/api/scheduler/scheduler.c
+++ b/test/validation/api/scheduler/scheduler.c
@@ -1831,8 +1831,20 @@ static int sched_and_plain_thread(void *arg)
 
 	/* Make sure scheduling context is released */
 	odp_schedule_pause();
-	while ((ev1 = odp_schedule(NULL, wait)) != ODP_EVENT_INVALID)
+	while ((ev1 = odp_schedule(NULL, wait)) != ODP_EVENT_INVALID) {
+		if (sync == ODP_SCHED_SYNC_ORDERED)
+			odp_schedule_order_lock(0);
+
+		ev2 = odp_queue_deq(plain_queue);
+		CU_ASSERT_FATAL(ev2 != ODP_EVENT_INVALID);
+
+		CU_ASSERT_FATAL(!odp_queue_enq(plain_queue, ev2));
+
+		if (sync == ODP_SCHED_SYNC_ORDERED)
+			odp_schedule_order_unlock(0);
+
 		CU_ASSERT_FATAL(!odp_queue_enq(sched_queue, ev1));
+	}
 
 	/* Don't resume scheduling until all threads have finished */
 	odp_barrier_wait(&globals->barrier);


test/validation/api/scheduler/scheduler.c
```diff
@@ -1831,8 +1831,16 @@ static int sched_and_plain_thread(void *arg)
...
 	odp_schedule_pause();
-	while ((ev1 = odp_schedule(NULL, wait)) != ODP_EVENT_INVALID)
+	while ((ev1 = odp_schedule(NULL, wait)) != ODP_EVENT_INVALID) {
+		/* Perform same sequence of ordered locks with all events
+		 * which are part of the same application ordering flow.
+		 */
+		if (sync == ODP_SCHED_SYNC_ORDERED) {
+			odp_schedule_order_lock(0);
+			odp_schedule_order_unlock(0);
+		}
```
Bill-Fischofer-Personal: Is this really necessary? By design ordered locks can be skipped and the effect is the same as if they were acquired/released as done here. The act of calling the scheduler following event receipt is an implicit "pass" on any ordered locks associated with that previous ordered context.

mvl-skardach: In all the hardware platforms I've worked with the act of taking an ordered lock means moving packet from one ordering phase to another. Skipping one of such phases for some packets meant that all the packets which skipped would be ordered in relation to each other but not in relation to the packets which did not skip processing phases.
Naturally one could implement implicit passing through all the phases/ordered-locks on `odp_queue_enq()`, however that is sub-optimal.
Also it seems that from application standpoint it is rather strange that user would process the same "flow" of events with different sequences of ordered locks.

Naturally if you could point me to a hardware platform that would support the semantics you describe (with multiple ordered locks) I'll be very interested.

Bill-Fischofer-Personal: If you look at the linux-generic implementation of ordered locks they are always optional, meaning that when a thread releases an ordered context, any ordered locks associated with that context that the thread "skipped" behave as if the thread acquired/released them. This is to avoid unnecessary lock/unlock calls. So the above change is extraneous since that functionality is implicit if the lock had not already been acquired/released earlier in that context. 

Had the locks been previously used in this context, then attempting to acquire them again in the same context would be an error since an ordered lock can only be used once by a thread in a given context.

This isn't just an artifact of the linux-generic implementation as it's part of the semantics of ODP ordered locks. How each implementation realizes those semantics will differ but the semantics remain the same across all implementations.

mvl-skardach: > Had the locks been previously used in this context, then attempting to acquire them again in the same context would be an error since an ordered lock can only be used once by a thread in a given context.

Here it is not anymore the same context as there is an odp_schedule() called in line 1834.

> This isn't just an artifact of the linux-generic implementation as it's part of the semantics of ODP ordered locks.
That is why I asked for an actual hardware platform implementation example. To me the semantics you describe seem artificial. Perhaps I'm missing the use-case where application would need to call different order lock sequences for packets that need to be ordered between each other.

My understanding is: a typical ODP application would use classification to put packets of different classes of packets to different ODP queues. Then use order locks for critical sections.
A sequence of ordered locks would define a processing path of the packet. If application thread needs to call different paths in scope of the same queue, it means that those packets most likely should be on 2 separate queues and therefore ordered separately.

Bill-Fischofer-Personal: Yes, the `odp_schedule()` call is an implicit release of any currently held context. My point was that for ordered contexts, that release implies an explicit "skip" of any unused ordered locks associated with that context.

mvl-skardach: That is the point that I'm contesting. Perhaps I should propose a spec change then. @psavol What do you think?

Bill-Fischofer-Personal: @mvl-skardach I'm not sure what benefit there would be to forcing a thread to cycle through unused ordered locks before allowing it to release an order context. In addition to backwards-compatibility concerns that requirement would be error-prone. What happens if a thread fails to observe that protocol? Having context release imply a "skip" for all unused ordered locks is clean and is guaranteed to avoid potential deadlocks associated with adding ordered locks to ordered queues.

mvl-skardach: Oh, I'm not implying that thread needs to. `odp_schedule()` should still release the context irrespectively of the order locks being used or not.
What I'm suggesting is that for a processing loop, when there are 3 processing paths in schedule loop:
a. rx -> `odp_schedule_order_lock(0)` -> tx
b. rx -> odp_schedule_order_lock(1) -> tx
c. rx -> tx
Then packets going through `a`, `b` and `c` will be ordered inside same processing path but ordering between packets going through `a` and going through `b` is not guaranteed.

Bill-Fischofer-Personal: Ordered locks within a given ordered context are independent and any usage protocol among them is up to the application. The main reason for having multiple ordered locks within an ordered context would be to permit packet flow to be processed in parallel between multiple ordered critical sections. Who gets "first shot" at a lock depends on the originating event sequence on the ordered queue, not the order in which the thread calls `odp_schedule_order_lock()` if there are a multiple ordered locks in play.

So if there are N ordered locks in the context and the input ordered queue has events `ev1`, `ev2`, and `ev3`. Then the thread the dequeues `ev1` gets every ordered lock in that context that it chooses to use before any ordered lock requested by the thread that dequeues `ev2`, etc. There's no architectural requirement that the `ev1` thread request locks in any particular order if there are more than one to choose from. None of `ev2`'s lock requests can be satisfied until either the lock that the `ev2` thread is requesting has been explicitly released by the `ev1` thread after it's done using it, or until it is implicitly skipped by virtue of the `ev1` thread releasing the ordered context by another `odp_schedule()` call.

This is why ordered locks are immune from deadlocks and callers need not observe any particular locking hierarchy when using them.

psavol: For correct test result, the paused schedule loop should do the same thing as the main loop. The paused loop just handles potential, last few stashed packets before the thread exits. Random wait can be skipped, it's in the main loop to increase chance of events getting out-of-order if lock implementation does not work correctly.

ev1 = odp_schedule(NULL, wait);
if (ev1 == ODP_EVENT_INVALID)
			break;

if (sync == ODP_SCHED_SYNC_ORDERED)
	odp_schedule_order_lock(0);

ev2 = odp_queue_deq(plain_queue);
CU_ASSERT_FATAL(ev2 != ODP_EVENT_INVALID);

CU_ASSERT_FATAL(!odp_queue_enq(plain_queue, ev2));

if (sync == ODP_SCHED_SYNC_ORDERED)
		odp_schedule_order_unlock(0);

CU_ASSERT_FATAL(!odp_queue_enq(sched_queue, ev1));

psavol: Commit log subject line should use ":" as delimitters, and can be shorted to:

validation: scheduler: ...

I use mainly these as the first "tag": api, validation, linux-gen, example and test (performance and other test applications than validation and examples).

Especially, consistent "api" and "validation" tagging should help different implementations to pick-up API spec related commits easily.

mvl-skardach: Will fix both in V2. "validation" sounds perfect.

PR #910, opened 2019-12-12
[PATCH v3] Enable DPDK v19.11 usage

psavol:
Fix build issues with DPDK v19.11. Also DPDK mempool allocation behavior has changed so that  zero copy packet IO needs MEMPOOL_F_NO_IOVA_CONTIG flag defined.

### DIFF
--- a/platform/linux-generic/pktio/dpdk.c
+++ b/platform/linux-generic/pktio/dpdk.c
@@ -1525,6 +1525,33 @@ static void dpdk_init_capability(pktio_entry_t *pktio_entry,
 		capa->config.pktout.bit.tcp_chksum;
 }
 
+/* Some DPDK PMD virtual devices, like PCAP, do not support promisc
+ * mode change. Use system call for them. */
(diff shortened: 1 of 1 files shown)


platform/linux-generic/pktio/dpdk.c
```diff
@@ -339,12 +340,18 @@ static struct rte_mempool *mbuf_pool_create(const char *name,
...
+	flags |= MEMPOOL_F_NO_IOVA_CONTIG;
+#endif
+
 	/* Skipped buffers have to be taken into account to populate pool
 	 * properly. */
 	mp = rte_mempool_create_empty(name, num + pool_entry->skipped_blocks,
 				      elt_size, cache_size(num),
 				      sizeof(struct rte_pktmbuf_pool_private),
-				      rte_socket_id(), 0);
+				      rte_socket_id(), flags);
```
matiaselo: According to DPDK API the same flags should also be used for `rte_mempool_calc_obj_size()`.

psavol: fixed in v2

PR #914, opened 2020-01-07
[PATCH v2] API: add version and cache line round up macros

psavol:
Added new macros into API for easier:
* API version comparison: ODP_VERSION_API_NUM, ODP_VERSION_API
* Cache line size round up in buffer size calculations, etc: ODP_CACHE_LINE_ROUNDUP()

### DIFF
--- a/include/odp/api/spec/align.h
+++ b/include/odp/api/spec/align.h
@@ -48,12 +48,12 @@ extern "C" {
 
 /**
  * @def ODP_CACHE_LINE_SIZE
- * Cache line size
+ * Cache line size in bytes
  */
 
 /**
  * @def ODP_PAGE_SIZE
- * Page size
+ * Page size in bytes
  */
 
 /**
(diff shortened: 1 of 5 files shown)


include/odp/api/spec/align.h
```diff
@@ -66,6 +66,15 @@ extern "C" {
  * Defines type/struct/variable to be page size aligned
  */
 
+/**
+ * @def ODP_CACHE_LINE_ROUNDUP
```
mvl-skardach: Just a suggestion. Given that there is really a limited number of ways to implement this macro, why not move it to the helper?

psavol: Yes, there are not many ways to implement it, but cache alignment is so commonly needed by applications that it's better to have this as a proper API.

Helpers are something that our test applications commonly re-use (e.g. protocol headers), but other applications could choose not to use (e.g. as they have their own definitions already).

PR #917, opened 2020-01-16
[PATCH v1] Prevent creation of zero length packets

matiaselo:


include/odp/api/spec/packet.h
```diff
@@ -318,11 +318,10 @@ void odp_packet_free_sp(const odp_packet_t pkt[], int num);
  *
  * Resets all packet metadata to their default values. Packet length is used
  * to initialize pointers and lengths. It must be less than the total buffer
- * length of the packet minus the default headroom length. Packet is not
- * modified on failure.
+ * length of the packet. Packet is not modified on failure.
  *
  * @param pkt           Packet handle
- * @param len           Packet data length
+ * @param len           Packet data length (1 ... odp_packet_buf_len())
```
Bill-Fischofer-Personal: What is problematic about zero-length packets? These have been supported by ODP since it's inception and are quite useful in many cases where packets are created dynamically by sweeping out tailroom. What is motivating this API change?

matiaselo: We are unifying the valid packet length definitions across all function definitions. `odp_packet_alloc()` has defined valid packet length as `(1 ... pool max_len)` for a long time now and also some function like `odp_packet_pull_head()` or `odp_packet_pull_tail()` reflect this already.

When working with packets you usually always start with some headers/payload and modify packet from there. On some HW platforms zero length packets are problematic and handling them would require adding extra checks to the fast path functions.

Bill-Fischofer-Personal: I've always viewed zero-length packets as no different from zero-length strings or null sets, both of which are standard features in languages that support them, but thanks for the clarification.

PR #918, opened 2020-01-20
[PATCH v2] API: packet length zero and packet pool align

psavol:
Merged packet API modifications into a single patch set. Clarify that application cannot create zero length packets. Add packet pool parameter to control minimum alignment of packet data when application allocates packets from a pool.

### DIFF
--- a/include/odp/api/spec/pool.h
+++ b/include/odp/api/spec/pool.h
@@ -92,6 +92,12 @@ typedef struct odp_pool_capability_t {
 		 * memory size for the pool. */
 		uint32_t max_num;
 
+		/** Maximum packet data alignment in bytes
+		 *
+		 * This is the maximum value of packet pool alignment
+		 * (pkt.align) parameter. */
+		uint32_t max_align;
+
 		/** Minimum packet level headroom length in bytes
 		 *
 		 * The minimum number of headroom bytes that newly created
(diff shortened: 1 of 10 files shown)


include/odp/api/spec/pool.h
```diff
@@ -92,6 +92,12 @@ typedef struct odp_pool_capability_t {
 		 * memory size for the pool. */
 		uint32_t max_num;
 
+		/** Maximum packet data alignment in bytes
+		 *
+		 * This is the maximum value of packet pool alignment
+		 * (pkt.align) parameter. */
```
mvl-skardach: Shouldn't `*/` be on the next line?

platform/linux-generic/odp_pool.c
```diff
@@ -503,6 +503,15 @@ static odp_pool_t pool_create(const char *name, odp_pool_param_t *params,
 	align = 0;
 
 	if (params->type == ODP_POOL_PACKET) {
+		uint32_t align_req = params->pkt.align;
```
mvl-skardach: You read this but don't use it. Shouldn't it be set to `align` if not 0?

include/odp/api/spec/pool.h
psavol: This style is has been used on this capa struct. So, used the same here as well.

platform/linux-generic/odp_pool.c
psavol: Since this is reference implementation, I'm checking the param for errors ... but decided to minimize changes to packet buffering to avoid new bugs / change of behaviour. So, all packet data is "base_align" aligned as it used to be.

PR #934, opened 2020-03-12
[PATCH v2] validation: packet: fix max_num expectations

mvl-skardach:
Some tests are missing params.pkt.max_num value and yet expect that no
more than params.pkt.num packets will be present in the pool. That is
not required by spec. Instead max_num parameter should be used for that.

Change-Id: I63901f564a634355267f1243cf7c12379945791c


matiaselo:
When you add the reviewed-by tag please remove the `Change-Id: I63901f564a634355267f1243cf7c12379945791c` from the commit message.

mvl-skardach:
> When you add the reviewed-by tag please remove the `Change-Id: I63901f564a634355267f1243cf7c12379945791c` from the commit message.

Will do in V2

PR #939, opened 2020-03-31
[PATCH v3] validation: scheduler: add new stress test

matiaselo:
New scheduler stress test creates the maximum number of scheduled queues
returned by odp_schedule_config_init(), fills them with events, and runs a
number of test scheduling rounds. The test is repeated for all scheduled
queue types.

V2:
- Fixed Petri's comments
- Added odp_schedule_group_create() test improvement commit

### DIFF
--- a/test/validation/api/scheduler/scheduler.c
+++ b/test/validation/api/scheduler/scheduler.c
@@ -685,7 +685,7 @@ static void scheduler_test_groups(void)
 	CU_ASSERT(lookup == ODP_SCHED_GROUP_INVALID);
 
 	/* Now create it and verify we can find it */
-	mygrp2 = odp_schedule_group_create("Test Group 2", &zeromask);
+	mygrp2 = odp_schedule_group_create("Test Group 2", &mymask);
 	CU_ASSERT_FATAL(mygrp2 != ODP_SCHED_GROUP_INVALID);
 
 	lookup = odp_schedule_group_lookup("Test Group 2");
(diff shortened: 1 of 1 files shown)


test/validation/api/scheduler/scheduler.c
```diff
@@ -493,6 +495,133 @@ static void scheduler_test_queue_size(void)
...
+			ret = odp_queue_enq(src_queue, ev);
+			CU_ASSERT(ret == 0);
+			if (ret) {
+				odp_event_free(ev);
+				num_enq--;
+			}
+		}
+		/* Clean-up */
+		for (j = 0; j < num_bufs; j++) {
+			ev = odp_schedule(NULL, 10 * ODP_TIME_MSEC_IN_NS);
```
psavol, line 607: Need to use odp_schedule_wait_time() to convert nsec to wait time.

test/validation/api/scheduler/scheduler.c
```diff
@@ -493,6 +495,133 @@ static void scheduler_test_queue_size(void)
...
+					continue;
+				}
+				num_enq++;
+			}
+		}
+		/* Run normal scheduling rounds */
+		for (j = 0; j < num_bufs; j++) {
+			odp_queue_t src_queue;
+
+			ev = odp_schedule(&src_queue, ODP_SCHED_NO_WAIT);
```
psavol, line 594: HW may have small delay from enqueue to schedule. May be this schedule call could also have a small wait time to ensure that test won't fail due to HW delays.

test/validation/api/scheduler/scheduler.c
```diff
@@ -493,6 +495,133 @@ static void scheduler_test_queue_size(void)
...
+
+	/* Scheduler has been already configured. Use default config as queue
+	 * size and queue count. */
+	odp_schedule_config_init(&default_config);
+	if (default_config.queue_size)
+		queue_size = default_config.queue_size;
+	num_queues = default_config.num_queues;
+
+	/* Queues reserved by create_queues() in suite init */
+	num_queues -= globals->num_queues;
```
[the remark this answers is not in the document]
matiaselo: I guess it's better to check it here and exit cleanly. Though, several other tests will also fail if they cannot create new queues.

test/validation/api/scheduler/scheduler.c
matiaselo: Will fix.

test/validation/api/scheduler/scheduler.c
matiaselo: For that reason there is no CU_ASSERT check for this schedule call. The clean-up loop, which uses wait time, checks that all enqueued events are received.

PR #943, opened 2020-04-23
[PATCH v5] fix scalable scheduler dead-loop issue during no traffic

jeckypei:
…lem maybe moved to another scheduler queue by other CPU core,

### DIFF
--- a/configure.ac
+++ b/configure.ac
@@ -4,7 +4,7 @@ AC_PREREQ([2.5])
 ##########################################################################
 m4_define([odpapi_generation_version], [1])
 m4_define([odpapi_major_version], [23])
-m4_define([odpapi_minor_version], [3])
+m4_define([odpapi_minor_version], [4])
 m4_define([odpapi_point_version], [0])
 m4_define([odpapi_version],
     [odpapi_generation_version.odpapi_major_version.odpapi_minor_version.odpapi_point_version])


matiaselo:
Hi,

Could you describe how we can reproduce this bug in our system?

jeckypei:
hi matiaselo 
      create many pktio and pktin queues , run odp_schedule_multi() in >= 2 DP cores  with a little traffic .
      the code logic is very clear.  all DP cores can traverse one same scheudler queue.   
 example : 
  Core A is traversing sched-queue1 :
         (1)  support pktio-rx-queue1 is the first elem,  first is set to judge dead-loop later . fetch a few packets, and then put it back to  sched-queue1.
  
 Core B also traversing  sched-queue1 :   
         (2)pop  pktio-rx-queue1 , fetch no packet.  then push it to another sheduler queue . 

  Core A is still traversing sched-queue1 :
          (3) there is other rx-queues in sched-queue1,  suppose these packets with no traffic now .  do it repeatedly.  try to use first to judge dead-loop, but it will not in sched-queue1, then dead-loop happen on CoreA, it will block timer and other application things, even if you use ODP_SCHED_NO_WAIT.

matiaselo:
@jeckypei Thanks for the comprehensive description. It sums up the problem clearly.

I guess the current commit in this PR is there more to describe the problem and not actually fix it yet. Are you planning to update it?

Btw. in most cases the default scheduler implementation should perform better than this one. There are also additional options in the config file (e.g. prio_spread) to further tweak the default scheduler operation.

jeckypei:
@matiaselo   
         i don't think current code is better than my patch . my patch check loop perfectly with similar time cost .   get timestamp only use one or two assembly instructions. 
    actually, i tested the performance with my patch . it reach the highest performance and it is very stable  on arm64 board,  no much fluctuation .  
     it is a basic code, should fix dead-loop issue with any configuration, right  ?

matiaselo:
> i don't think current code is better than my patch . my patch check loop perfectly with similar time cost . get timestamp only use one or two assembly instructions.
> actually, i tested the performance with my patch . it reach the highest performance and it is very stable on arm64 board, no much fluctuation .
> it is a basic code, should avoid dead-loop issue with any configuration, right ?

The bug should definitely be fixed. My earlier question was because in the current form your PR won't compile ([CI results](https://travis-ci.org/github/OpenDataPlane/odp/builds/678462599)).

jeckypei:
@matiaselo 
    will pull again. sorry my mistake.

platform/linux-generic/odp_schedule_scalable.c
```diff
@@ -950,18 +950,19 @@ static int _schedule(odp_queue_t *from, odp_event_t ev[], int num_evts)
...
 		if (odp_unlikely(elem == NULL)) {
 			/* Schedq empty, look at next one. */
 			continue;
 		}
 		if (is_pktin(elem)) {
-			/* Pktio ingress queue */
-			if (first == NULL)
-				first = elem;
-			else if (elem == first) /* Wrapped around */
+			if (elem->loop_check[odp_cpu_id()] == time_stamp) { /* Wrapped around */
```
matiaselo, line 961: Scheduling elements are currently mostly read-only, so adding a write operation for each element (plus `CONFIG_NUM_CPU` x 8B extra data) adds a considerable overhead. On the other hand, I'm finding it hard to figure out a solution with less overhead. Would adding an internal define for the max number of pktio queues to poll per scheduling round be enough to avoid most of the problems you have described?

jeckypei: hi @matiaselo 
     because ODP don't have   DP NUMBER and DP core id design.   so use CONFIG_NUM_CPU   ,maybe waste some memory. But it is very a little,relative to pkt_priv a big memory waste.  it is not a big thing .
unsigned char ODP_ALIGNED_CACHE pkt_priv[PKTIO_PRIVATE_SIZE]; //great memory waste is here .      you can ignore schedule elem .

jeckypei: maybe we can use unsigned __u16 to implement the same feature,   use a thread variable to replace __u64 type timestamp .

matiaselo: > because ODP don't have DP NUMBER and DP core id design. so use CONFIG_NUM_CPU ,maybe waste some memory. But it is very a little,relative to pkt_priv a big memory waste. it is not a big thing .

The size of sched_elem_t is multiplied with queue count, but anyway that is a secondary issue. What I'm worried about is writing to each `sched_elem_t` element during the scheduling loop and causing additional cache line invalidations. Currently element data is only read if no packets are received.

matiaselo: One option would be to always schedule pktin queues using the original scheduling queue and not move empty queues to a lower priority sched queue. This would improve QoS seen by the input queues but decrease max throughput if not all input queues are receiving traffic.

jeckypei: for hw queues, we can not know which one have traffic with high priority, sometime no traffic , other time  many traffic, it is normal status in network . we depend on dynamical priority.     ignore my patch3 uint16 which is not perfect to avoid loop.  uint64 is still perfect .

regarding to  cacheline , we can test the real performance difference , is it a big thing , to check. in my test setup,  the performance is hightest value after using my fix .

matiaselo: Isn't it so, that the dead lock actually happens when multiple pktio queue elements have already been "demoted" to the lower priority idle scheduling queue? When looping through this idle queue and some other thread moves the 'first' element we end up in a dead lock. However, in case of a normal priority scheduling queue the idle pktio queue elements will always be demoted to the lower priority queue and the loop will eventually break since there are no more queue elements left in the current scheduling queue.

Now, would it be acceptable to simply add a define (or dynamically calculated value) to set the maximum number of "idle" pktio queue elements to poll per scheduling loop? This would minimize the overhead seen by normal priority queues.

matiaselo: Thinking more about it, adding one more if clause (`if (elem->schedq != schedq)`) removes most of the overhead for the non-idle pktin queues.

E.g.
```
if (is_pktin(elem)) {
	/* Check idle pktin queue for wraparound */
	if (elem->schedq != schedq) {
		int thread_id = odp_thread_id();

		if (elem->loop_check[thread_id] == loop_cnt)
			continue;
		elem->loop_check[thread_id] = loop_cnt;
	}
...
```

platform/linux-generic/include/odp_schedule_scalable.h
```diff
@@ -89,6 +89,7 @@ typedef struct ODP_ALIGNED_CACHE {
 #define cons_type qschst_type
 #endif
 	odp_schedule_group_t sched_grp;
+	uint64_t loop_check[CONFIG_NUM_CPU];
```
matiaselo: `ODP_THREAD_COUNT_MAX` should be used here.

platform/linux-generic/odp_schedule_scalable.c
```diff
@@ -950,18 +950,19 @@ static int _schedule(odp_queue_t *from, odp_event_t ev[], int num_evts)
...
 				continue; /* Go to next schedq */
+			} else {
+				elem->loop_check[odp_cpu_id()] = time_stamp;
```
matiaselo: `odp_thread_id()` should be used here. Its range is always from 0 to odp_thread_count_max() - 1.

platform/linux-generic/odp_schedule_scalable.c
```diff
@@ -67,6 +67,7 @@ typedef struct {
 static sched_global_t *global;
 
 __thread sched_scalable_thread_state_t *sched_ts;
+__thread uint16_t loop_cnt;
```
matiaselo, line 70: Should be moved inside `sched_scalable_thread_state_t `.

matiaselo:
All the commits should be squashed into a single commit with an appropriate name.
E.g.
linux-gen: sched scalable: fix pkin queue polling dead lock

platform/linux-generic/include/odp_schedule_scalable.h
jeckypei: generally, ODP_THREAD_COUNT_MAX   is two bigger than CONFIG_NUM_CPU

matiaselo: `ODP_THREAD_COUNT_MAX` doesn't refer directly to the number CPU threads. Instead, it is the number of ODP application threads (or processes), where `odp_init_local()` has been called. `ODP_THREAD_COUNT_MAX ` should be used with the `odp_thread_id()` API calls. By default, the values of `ODP_THREAD_COUNT_MAX` and `CONFIG_NUM_CPU` are the same.

jeckypei: in our case
ODP_THREAD_COUNT_MAX  is  about 300
CONFIG_NUM_CPU is about 2-8 
very different

platform/linux-generic/odp_schedule_scalable.c
jeckypei: hi matiaselo
   it have a very little risk to use uint16.  
  if high priority sched queue always return packets, maybe increase the loop_cnt one 
 or more cycle. when the low priority sched queue is called in the next time, maybe the loop_cnt is equal to the old value , cause scheduler ignore the queue directly . But  the probability is very very small.   if we use uint64, there is not probability to happen the coincidence, for counter loop need many years  .  uint32     
 is  the middle value to reduce the  coincidence probability, uint32 can make sure time of  the coincidence is second level.   which one(uint16 uint32 uint64) do you want to accept from you view  ?

matiaselo: Hi,

As you have noted, the probability of this happening is really small. And even if it did, the result would not be a critical failure. So, I think using uint64 would be overkill. Since you are using this code in your application you can better analyze the potential performance differences between uint16 and uint32, I'm good with either option.

platform/linux-generic/include/odp_schedule_scalable.h
matiaselo: There is a potential issue if multiple threads are running on the same core (allowed by API), since they end up sharing the same `loop_check`value but use a thread local `loop_cnt`. However, the potential value clash only causes a missed schedule round for low priority queues so I think we can live with this unless someone else has an objection. Calling schedule on multiple threads using the same core doesn't make much sense anyway.

Btw. `CONFIG_NUM_CPU` will be renamed to `CONFIG_NUM_CPU_IDS` in the head of master branch.

PR #950, opened 2020-05-14
[PATCH v2] linux-gen: pool: add skip_cross_page in odp_pool_param_t

jeckypei:


platform/linux-generic/odp_pool.c
```diff
@@ -641,9 +642,9 @@ static odp_pool_t pool_create(const char *name, odp_pool_param_t *params,
 		block_size = ROUNDUP_CACHE_LINE(hdr_size + align_pad + seg_len);
 	}
 
-	/* Allocate extra memory for skipping packet buffers which cross huge
+	/* Allocate extra memory for skipping buffers which cross huge
 	 * page boundaries. */
-	if (params->type == ODP_POOL_PACKET) {
+	if (params->skip_cross_page) {
```
matiaselo, line 647: At a quick glance it seems we could make this the default operation for all pool types, so no API changes would be required.

jeckypei: only in the situation:
   buffer is shared to kernel space .  kernel convert phy-addr to virt-addr to access memory, it ask buffer don't cross page .
Generally , need not to skip cross page to save memory .

include/odp/api/spec/pool.h
```diff
@@ -200,6 +200,9 @@ typedef struct odp_pool_param_t {
 	/** Pool type */
 	int type;
 
+	/** skip buffers which cross huge page boundaries */
+	odp_bool_t skip_cross_page;
```
psavol, line 204: It is not significant for an application, if a buffer is mapped over single or multiple physical pages. So, configuration options for this should be added into config file, rather than into the API.

platform/linux-generic/odp_pool.c
psavol: This could be the default operation with huge pages for all pool types. E.g. with 2MB huge page size and 2kB buffer size the memory loss would 1/1000, which is not very significant.

jeckypei: ok ,let's  make it default .

include/odp/api/spec/pool.h
jeckypei: that make sense

PR #952, opened 2020-05-14
[PATCH v2] test: l2fwd: add packet copy option

psavol:
Added -p option to enable performance testing with packet copy.
This stresses packet alloc and free operations, as well as CPU's
interconnection toward caches and memory.

### DIFF
--- a/test/performance/odp_l2fwd.c
+++ b/test/performance/odp_l2fwd.c
@@ -1,5 +1,5 @@
 /* Copyright (c) 2014-2018, Linaro Limited
- * Copyright (c) 2019, Nokia
+ * Copyright (c) 2019-2020, Nokia
  * All rights reserved.
  *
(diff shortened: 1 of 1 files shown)


test/performance/odp_l2fwd.c
```diff
@@ -1150,6 +1179,9 @@ static void usage(char *progname)
 	       "                          num: must not exceed number of interfaces or workers\n"
 	       "  -b, --burst_rx <num>    0:   Use max burst size (default)\n"
 	       "                          num: Max number of packets per receive call\n"
+	       "  -p, --packet_copy       0: Don't copy packet\n"
```
matiaselo, line 1182: (default)

psavol: Fixed in v2

PR #953, opened 2020-05-15
[PATCH v3] linux-gen: add configure option for setting path to the default config file

matiaselo:
Add '--with-config=FILE' configuration option for setting path to the default configuration file. The default configuration file has to include all configuration options.

V2:
- Renamed `--with-config` to `--with-config-file`. (Petri)

### DIFF
--- a/platform/linux-generic/m4/odp_libconfig.m4
+++ b/platform/linux-generic/m4/odp_libconfig.m4
@@ -1,0 +1,18 @@
+##########################################################################
+# Configuration file version
+##########################################################################
+m4_define([_odp_config_version_generation], [0])
+m4_define([_odp_config_version_major], [1])
+m4_define([_odp_config_version_minor], [13])
+
+m4_define([_odp_config_version],
+          [_odp_config_version_generation._odp_config_version_major._odp_config_version_minor])
+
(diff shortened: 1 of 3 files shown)


platform/linux-generic/m4/odp_libconfig.m4
```diff
@@ -15,4 +15,14 @@ AC_SUBST(_ODP_CONFIG_VERSION_MAJOR)
...
 AC_SUBST(_ODP_CONFIG_VERSION_MINOR)
 
-ODP_LIBCONFIG([$with_platform])
+##########################################################################
+# Set optional path for the default configuration file
+##########################################################################
+default_config_path="${srcdir}/config/odp-$with_platform.conf"
+
+AC_ARG_WITH([config],
+AS_HELP_STRING([--with-config=FILE      path to the default configuration file],
```
psavol, line 24: I'd prefer to name this --with-config-file. It would be more clear what the option controls and would match ODP_CONFIG_FILE env variable naming.

matiaselo: Makes sense. Will fix in V2.

PR #955, opened 2020-05-20
example: hello: honor -c option

jerinjacobk:
In the existing application, -c is not honored, instead it picks the
first available CPU from odp_cpumask_default_control().

Changed the logic to iterate  over available CPU to application
and to find requested CPU available and run on the specific CPU.

Changed default CPU as 1 for more sane default value as
linux generic implementation does not have CPU 0 in
odp_cpumask_all_available().

### DIFF
--- a/example/hello/odp_hello.c
+++ b/example/hello/odp_hello.c
@@ -56,15 +56,16 @@ static int parse_args(int argc, char *argv[], options_t *opt)
 
 int main(int argc, char *argv[])
 {
+	char mask_str[ODP_CPUMASK_STR_SIZE];
 	odp_instance_t inst;
 	options_t opt;
 	pid_t pid;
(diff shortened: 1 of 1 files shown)


psavol:
Thanks for spotting and fixing this. Do you need the -c option? After looking into the hello world code, I think we could just remove the -c option. It's pretty unnecessary for hello world to demonstrate CPU pinning. It should demonstrate a minimum functionality application. User can run the application with taskset command, if it needs to run on a particular CPU. Also, the proposed default CPU ID of 1 would not work on single CPU machines, fixing that would complicate the cpu selection part even more.

jerinjacobk:
> Thanks for spotting and fixing this. Do you need the -c option?

Yes. it makes sense to remove -c option for hello application. I will send another patch to remove -c option if there no other comments?

PR #956, opened 2020-05-22
[PATCH v9] build: explicitly list dependency libraries

mvl-skardach:
build: explicitly list dependency libraries

Debian-based Linux distributions have introduced a while ago a very
specific patch to libtool generation which disables libtool generating
dependent libraries list to the linker (distribution libtool.m4 has
link_all_deplibs forced to no). This causes issues when cross-compiling
while using dependent libraries compiled and installed manually to
non-sysroot location (see [1]).

To reproduce the issue, one has to fetch a toolchain which does not
contain pkg-config in it. Debian/Ubuntu aarch64 toolchains do contain
version of that tool (aarch64-linux-gnu-pkg-config) which does not have
the change in question.

Reproduction steps:

1. Set env variables used throughout other steps:
```
   # Cross-compiler without modified pkg-config -> using system one.
   CROSS_COMPILE=aarch64-marvell-linux-gnu
   LIBCONFIG_DIR=/tmp/odp/libconfig
   DEPS_DIR=/tmp/odp/deps-prefix
   ODP_DIR=`pwd`
   BUILD_DIR=/tmp/odp/odp-build
   PREFIX=/tmp/odp/prefix
```
2. Build libconfig in separate dir:
```
   mkdir -p $LIBCONFIG_DIR
   cd $LIBCONFIG_DIR
   wget https://hyperrealm.github.io/libconfig/dist/libconfig-1.7.2.tar.gz
   tar -zxvf libconfig-1.7.2.tar.gz --strip-components=1
   ./configure --host=${CROSS_COMPILE} --prefix=${DEPS_DIR}
   make -j4 all
   make install
```
3. Cross-compile ODP linux-generic with shared libraries enabled:
```
   cd $ODP_DIR
   ./bootstrap
   mkdir -p $BUILD_DIR
   cd $BUILD_DIR
   PKG_CONFIG_PATH=$DEPS_DIR/lib/pkgconfig \
     $ODP_DIR/configure \
       --with-platform=linux-generic \
       --host=${CROSS_COMPILE} \
       --build=x86_64-linux-gnu \
       --enable-shared \
       --disable-static \
       --prefix=$PREFIX \
       --without-openssl \
       --disable-test-vald
   make all -j4 V=1
```
This will fail at odp_classifier linking with the following error::
```
  make[2]: Entering directory '/tmp/odp-build/example/classifier'
    CC       odp_classifier.o
    CCLD     odp_classifier
  /opt/marvell-tools/.../ld: warning: libconfig.so.11,
     needed by ../../lib/.libs/libodp-linux.so, not found
     (try using -rpath or -rpath-link)
  ../../lib/.libs/libodp-linux.so: undefined reference to `config_init'
  ../../lib/.libs/libodp-linux.so: undefined reference to `config_read_file'
  ../../lib/.libs/libodp-linux.so: undefined reference to `config_lookup'
  ../../lib/.libs/libodp-linux.so: undefined reference to `config_lookup_int'
  ../../lib/.libs/libodp-linux.so: undefined reference to `config_lookup_string'
  ../../lib/.libs/libodp-linux.so: undefined reference to `config_setting_get_int_elem'
  ../../lib/.libs/libodp-linux.so: undefined reference to `config_read_string'
  ../../lib/.libs/libodp-linux.so: undefined reference to `config_setting_length'
  ../../lib/.libs/libodp-linux.so: undefined reference to `config_destroy'
collect2: error: ld returned 1 exit status
make[2]: *** [Makefile:533: odp_classifier] Error 1
make[2]: Leaving directory '/tmp/odp-build/example/classifier'
make[1]: *** [Makefile:468: all-recursive] Error 1
make[1]: Leaving directory '/tmp/odp-build/example'
make: *** [Makefile:551: all-recursive] Error 1
```
There are several solutions to fix this:
1. Live-patch `libtool.m4` in configure stage.
2. Explicitly add dependent libraries to examples and tests.
3. Commit a pre-patched `m4/libtool.m4` to the repository.

First approach causes configure to be run twice (due to live patching).
Third approach requires locking `libtool.m4` to a specific version which
may result in issues with some older autotools as well as produces a
large blob-like patch.
This patch implements the second approach by adding `PLAT_DEP_LIBS`
variable which platform should fill with libraries that can be added
from non-standard locations (such as `libconfig` and `libcrypto` for
linux-generic).

[1] https://bugs.debian.org/cgi-bin/bugreport.cgi?bug=297726

### DIFF
--- a/m4/odp_dpdk.m4
+++ b/m4/odp_dpdk.m4
@@ -31,10 +31,13 @@ AC_DEFUN([_ODP_DPDK_SET_LIBS], [dnl
 ODP_DPDK_PMDS([$DPDK_PMD_PATH])
 DPDK_LIB="-Wl,--whole-archive,-ldpdk,--no-whole-archive"
 AS_IF([test "x$DPDK_SHARED" = "xyes"], [dnl
-    # applications don't need to be linked to anything, just rpath
-    DPDK_LIBS_LT="$DPDK_RPATH_LT"
-    # static linking flags will need -ldpdk
-    DPDK_LIBS_LT_STATIC="$DPDK_LDFLAGS $DPDK_LIB $DPDK_LIBS"
+    if test x$enable_static_applications != xyes; then
+      # applications don't need to be linked to anything, just rpath
+      DPDK_LIBS_LT="$DPDK_RPATH_LT"
+    else
+      # static linking flags will need -ldpdk
+      DPDK_LIBS_LT="$DPDK_LDFLAGS $DPDK_LIB $DPDK_LIBS"
+    fi
     DPDK_LIBS="-Wl,--no-as-needed,-ldpdk,--as-needed,`echo $DPDK_LIBS | sed -e 's/ /,/g'`"
     DPDK_LIBS="$DPDK_LDFLAGS $DPDK_RPATH $DPDK_LIBS"
     # link libodp-linux with -ldpdk
@@ -44,7 +47,6 @@ AS_IF([test "x$DPDK_SHARED" = "xyes"], [dnl
     # rearranged by libtool
     DPDK_LIBS_LT="`echo $DPDK_LIBS | sed -e 's/^/-Wc,/' -e 's/ /,/g'`"
     DPDK_LIBS_LT="$DPDK_LDFLAGS $DPDK_LIB $DPDK_LIBS_LT $DPDK_LIBS"
-    DPDK_LIBS_LT_STATIC="$DPDK_LIBS_LT"
     # static linking flags follow the suite
     DPDK_LIBS="$DPDK_LDFLAGS $DPDK_LIB $DPDK_LIBS"
     # link libodp-linux with libtool linking flags
@@ -62,7 +64,6 @@ LIBS=$OLD_LIBS
 AC_SUBST([DPDK_LIBS])
 AC_SUBST([DPDK_LIBS_LIBODP])
 AC_SUBST([DPDK_LIBS_LT])
-AC_SUBST([DPDK_LIBS_LT_STATIC])
 ])
 
 # _ODP_DPDK_CHECK_LIB(LDFLAGS, [LIBS])
@@ -189,8 +190,6 @@ DPDK_PKG=", libdpdk"
 AC_SUBST([DPDK_PKG])
 # applications don't need to be linked to anything, just rpath
 DPDK_LIBS_LT=""
-# compile all static flags into single argument to fool libtool
-DPDK_LIBS_LT_STATIC="-pthread -Wl,`echo $DPDK_STATIC_LIBS | sed -e 's/-pthread//g' -e 's/ \+/,/g' -e 's/-Wl,//g'`"
 # FIXME: this might need to be changed to DPDK_LIBS_STATIC
 DPDK_LIBS_LIBODP="$DPDK_LIBS"
 DPDK_LIBS=""
--- a/platform/linux-generic/m4/configure.m4
+++ b/platform/linux-generic/m4/configure.m4
@@ -30,6 +30,8 @@ m4_include([platform/linux-generic/m4/odp_netmap.m4])
 m4_include([platform/linux-generic/m4/odp_dpdk.m4])
 ODP_SCHEDULER
 
+AS_VAR_APPEND([PLAT_DEP_LIBS], ["${LIBCONFIG_LIBS} ${OPENSSL_LIBS} ${DPDK_LIBS_LT}"])
+
 AC_CONFIG_COMMANDS_PRE([dnl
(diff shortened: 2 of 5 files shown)


mvl-skardach:
Let's track the discussion here. Last time we spoke @psavol you expressed preference of method 2 above, correct? To summarize this will require adding a build system variable that platforms would set which would list all required libraries, correct?

platform/linux-generic/m4/configure.m4
```diff
@@ -29,6 +29,8 @@ m4_include([platform/linux-generic/m4/odp_netmap.m4])
 m4_include([platform/linux-generic/m4/odp_dpdk.m4])
 ODP_SCHEDULER
 
+AS_VAR_APPEND([PLAT_DEP_LIBS], ["${LIBCONFIG_LIBS} ${OPENSSL_LIBS}"])
```
matiaselo, line 32: Should the rest of linux-generic dependency libraries be also listed here? E.g. pcap, rt, pthread, and atomic.

matiaselo:
The commit message can be cleaned to only include the selected solution.

platform/linux-generic/m4/configure.m4
mvl-skardach: I have not noticed any issues with rt, pthread or atomic.

Though I see you've had some issues there:
```
commit 324a2a2c7c1a9d35323483b679eedfebcb310827
Author: Matias Elo <matias.elo@nokia.com>
Date:   Thu Feb 13 10:39:56 2020 +0200

    build: check if libatomic is needed for __atomic_compare_exchange_n
    
    Double wide __atomic_compare_exchange_n() is required by ipfragreass
    example.
    
```

I'd have to re-check with pcap. Perhaps I've missed it.

mvl-skardach: OK, I need a bit of help with pcap. If I'm looking at the current code right - there's no way to specify a cross-compile path for pcap, correct? `AC_ARG_WITH([pcap]` is really used for disabling pcap only, the `ODP_PCAP` just assumes `-lpcap` is available in default linking path.

Am I missing something?

matiaselo: You are correct regarding the libpcap path. Now, reading the commit message properly, the libtool bug only affects libs with custom install paths. So in that case I guess you are only missing DPDK_LIBS.

mvl-skardach: Fixed in V5.
Bit related question - is there a way to run all docker builds locally without invoking Travis? I'd prefer to check those kind of commits against the whole build matrix but without necessarily pushing them here and hogging Your resources.

matiaselo: I've sometimes run the docker images locally for debugging purposes, but there doesn't seem to be an easy method to run the whole build matrix locally.

When Travis is running in your fork of ODP main repo it's not using ODP project resources. So in case you've already submitted a PR, you can create a new branch and cherry pick the updated PR commits (not pushed yet) there. Pushing this new branch to GitHub will trigger Travis in your repo and doesn't take any resources from the ODP project.

matiaselo: It seems like you haven't yet enabled Travis in you ODP fork. Travis can be enabled by following [these](https://github.com/OpenDataPlane/odp/wiki/Using-GitHub-and-Travis-for-ODP-Development) instructions.

mvl-skardach:
@matiaselo Seems like traffic manager is acting up again :(

matiaselo:
> @matiaselo Seems like traffic manager is acting up again :(

@mvl-skardach  I restarted the test.

m4/odp_dpdk.m4
```diff
@@ -44,7 +47,6 @@ AS_IF([test "x$DPDK_SHARED" = "xyes"], [dnl
     # rearranged by libtool
     DPDK_LIBS_LT="`echo $DPDK_LIBS | sed -e 's/^/-Wc,/' -e 's/ /,/g'`"
     DPDK_LIBS_LT="$DPDK_LDFLAGS $DPDK_LIB $DPDK_LIBS_LT $DPDK_LIBS"
-    DPDK_LIBS_LT_STATIC="$DPDK_LIBS_LT"
```
matiaselo, line 47: It seems like `DPDK_LIBS_LT_STATIC` is not used anymore, so it could be removed completely (still couple usages in this file).

matiaselo:
One minor comments but otherwise looking good.

m4/odp_dpdk.m4
mvl-skardach: Done in V8

matiaselo (approved):

PR #957, opened 2020-05-28
[PATCH v2] api: sched: config options for event accumulation

vvelumuri:
Currently ODP assumes that scheduler implementation will use best
parameters for burst accumulation size and timeout. However HW
schedulers might require this to be setup in control path and not on
each call of odp_schedule_multi().

Exposing such parameters to the application allows for better control of
scheduler latency and throughput as well as benefit from more effective
SIMD processing.

It proposes the following configuration options to configure the
multi-event aspects of the scheduler API.

1. Number of events to accumulate
The scheduler will asynchronously accumulate upto these
many number of events and deliver the events on the subsequent
invocation of odp_schedule_multi().

Invoking odp_schedule_multi() with burst less than this value is
not allowed.

2. Timeout for accumulation:
The scheduler will asynchronously accumulate events for this
time period before delivering the batch of events.

### DIFF
--- a/include/odp/api/spec/schedule_types.h
+++ b/include/odp/api/spec/schedule_types.h
@@ -160,6 +160,39 @@ extern "C" {
  */
 typedef int odp_schedule_prio_t;
 
+/**
+ * Multi event config parameters
+ */
+typedef struct odp_multi_event_param_t {
+	/** Number of events to accumulate
+	 *
+	 * The scheduler will asynchronously accumulate upto these many number
+	 * of events and deliver the events on the subsequent invocation of
+	 * odp_schedule_multi().
+	 * This value should be less than or equal to max_multi_event_count.
+	 *
+	 * 0:     Asynchronous accumulation is disabled.
+	 * >= 1:  Asynchronous accumulation is enabled.
+	 *
+	 * @see odp_schedule_capability_t::max_multi_event_count.
+	 *
+	 * Default value is 0. */
+	unit32_t   multi_event_count;
+
+	/** Timeout for accumulation
+	 *
+	 * The scheduler will asynchronously accumulate events for this time
+	 * period before delivering the batch of events.
+	 * This value should be in the range defined by
+	 * multi_event_min_async_wait and multi_event_max_async_wait.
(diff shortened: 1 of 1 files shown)


include/odp/api/spec/schedule_types.h
```diff
@@ -181,6 +214,9 @@ typedef	struct odp_schedule_param_t {
 	  *
 	  * Default value is 0. */
 	uint32_t lock_count;
+
+	/** Multi event config options */
+	odp_multi_event_param_t multi_event_param;
```
psavol, line 219: schedule_param_t is a queue create parameter. Scheduler burst size config belongs to odp_schedule_config_t, which has global scheduler configuration parameters.

include/odp/api/spec/schedule_types.h
```diff
@@ -160,6 +160,39 @@ extern "C" {
...
+	 * period before delivering the batch of events.
+	 * This value should be in the range defined by
+	 * multi_event_min_async_wait and multi_event_max_async_wait.
+	 *
+	 * @see odp_schedule_capability_t::multi_event_min_async_wait.
+	 * @see odp_schedule_capability_t::multi_event_max_async_wait.
+	 *
+	 * Default value is 0. */
+	uint32_t  multi_event_async_wait_ns;
+} odp_multi_event_param_t;
```
psavol, line 194: This specification does not bring much new functionality or guarantees to the application. Biggest difference would be to an implementation which receives expected burst size (and timeout) in scheduler configuration phase. The same could be achieved with implementation specific configuration file. E.g. application would still need to prepare to receive 1...N events from each schedule multi call.

An schedule burst size API change should probably specify more functionality / limitations e.g.:
 * promise that exactly N events is returned if there was no timeout
 * specify that application must call odp_schedule_multi(), with 'num' >= N. odp_schedule() could not be used when N > 1.
 * how this wait time relates to schedule call wait time. Is this max (or min) schedule wait time that application will set into a schedule call.

psavol:
Replaced by https://github.com/OpenDataPlane/odp/pull/1062

PR #960, opened 2020-05-28
[PATCH v3] L2fwd pool options

psavol:
Added new options to control number of pools (-y) and number of packets per pool (-n). Depending on implementation these settings may have performance impacts.

  -y, --pools             0: A pool per interface
                          1: A single pool for all interfaces (default)
  -n, --num_pkt <num>     Number of packets per pool. Default is 16k or
                          the maximum capability. Use 0 for the default.

### DIFF
--- a/platform/linux-generic/odp_pool.c
+++ b/platform/linux-generic/odp_pool.c
@@ -513,9 +513,10 @@ static odp_pool_t pool_create(const char *name, const odp_pool_param_t *params,
 	uint32_t max_len, cache_size, burst_size;
 	uint32_t ring_size;
 	uint32_t num_extra = 0;
-	int name_len;
-	const char *postfix = "_uarea";
-	char uarea_name[ODP_POOL_NAME_LEN + sizeof(postfix)];
+	const char *max_prefix = "pool_000_uarea_";
+	int max_prefix_len = strlen(max_prefix);
+	char shm_name[ODP_POOL_NAME_LEN + max_prefix_len];
+	char uarea_name[ODP_POOL_NAME_LEN + max_prefix_len];
 
 	align = 0;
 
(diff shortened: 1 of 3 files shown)


platform/linux-generic/odp_pool.c
```diff
@@ -622,9 +623,9 @@ static odp_pool_t pool_create(const char *name, const odp_pool_param_t *params,
 		pool->name[ODP_POOL_NAME_LEN - 1] = 0;
 	}
 
-	name_len = strlen(pool->name);
-	memcpy(uarea_name, pool->name, name_len);
-	strcpy(&uarea_name[name_len], postfix);
+	/* Format SHM names from prefix, pool index and pool name. */
+	sprintf(shm_name,   "pool_%03i_%s", pool->pool_idx, pool->name);
+	sprintf(uarea_name, "pool_%03i_uarea_%s", pool->pool_idx, pool->name);
```
matiaselo, line 628: One issue with this change is that it breaks IPC pktio if pools aren't created in the same order in both processes.

test/performance/odp_l2fwd.c
```diff
@@ -91,6 +91,7 @@ typedef struct {
 	int sched_mode;         /* Scheduler mode */
 	int num_groups;         /* Number of scheduling groups */
 	int burst_rx;           /* Receive burst size */
+	int pools;              /* Pools per interface option */
```
matiaselo, line 94: Naming of this variable makes reading the code a bit confusing. E.g. `if (gbl_args->appl.pools == 0)` -> use multiple pools. For example pool_per_if would be easier to understand.

psavol: fixed in v2

PR #961, opened 2020-06-02
[PATCH v3] Fix compilation with GCC 10

JannePeltonen:
GCC 10 warns if the bounds of interior zero length arrays are exceeded, but that is done intentionally in ODP in a few places. Ignore the warning in those pieces of code.

The current GCC 10 also seems to have a bug that causes bogus array-bounds and stringop-overflow warnings. As a workaround, globally exclude those warnings from -Werror. Both are in the same commit as the affected code is the same and the issue triggering the warnings appears to be the same.

### DIFF
--- a/configure.ac
+++ b/configure.ac
@@ -133,6 +133,15 @@ ODP_CHECK_CFLAG([-Wformat-overflow=0])
 # header structures). Generate only warnings on those, not errors.
 ODP_CHECK_CFLAG([-Wno-error=address-of-packed-member])
 
+# GCC 10 sometimes gets confused about object sizes and gives bogus warnings.
+# Make the affected warnings generate only warnings, not errors.
+AS_IF([test "$GCC" == yes],
+      AS_IF([test `$CC -dumpversion` -ge 10],
+	    ODP_CHECK_CFLAG([-Wno-error=array-bounds])
+	    ODP_CHECK_CFLAG([-Wno-error=stringop-overflow])
+      )
+)
+
 ODP_CFLAGS="$ODP_CFLAGS -std=c99"
 ODP_CXXFLAGS="$ODP_CXXFLAGS -std=c++11"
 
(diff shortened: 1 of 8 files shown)


configure.ac
```diff
@@ -133,6 +133,14 @@ ODP_CHECK_CFLAG([-Wformat-overflow=0])
 # header structures). Generate only warnings on those, not errors.
 ODP_CHECK_CFLAG([-Wno-error=address-of-packed-member])
 
+# GCC 10 sometimes gets confused about oject sizes and gives bogus warnings.
+# Make the affected warnings generate only warnings, not errors.
+case `($CC --version) 2>/dev/null` in
+     'gcc (GCC) 10'*)
```
matiaselo, line 139: This check doesn't work on Ubuntu 20.04.

```
$ gcc-10 --version
gcc-10 (Ubuntu 10-20200411-0ubuntu1) 10.0.1 20200411 (experimental) [master revision bb87d5cc77d:75961caccb7:f883c46b4877f637e0fa5025b4d6b5c9040ec566]
```

matiaselo:
You can add a Travis build test for gcc 10 for example like [this](https://github.com/matiaselo/odp/commit/a6bd4c1acec530297e763bddb4bace1bc815d574).

configure.ac
JannePeltonen: Fixed in v2.

configure.ac
```diff
@@ -133,6 +133,15 @@ ODP_CHECK_CFLAG([-Wformat-overflow=0])
 # header structures). Generate only warnings on those, not errors.
 ODP_CHECK_CFLAG([-Wno-error=address-of-packed-member])
 
+# GCC 10 sometimes gets confused about oject sizes and gives bogus warnings.
```
matiaselo, line 136: typo: oject

JannePeltonen:
V3: Fixed the typo and the commit message as requested and added reviewed-by tags.

PR #965, opened 2020-06-08
[PATCH v4] API: timer resolution in Hz

psavol:


include/odp/api/spec/timer.h
```diff
@@ -186,8 +195,9 @@ typedef struct {
 	 *
 	 * This defines the highest resolution supported by a timer, with
 	 * limits to min/max timeout values. The highest resolution for a timer
-	 * pool is defined by 'max_res.res_ns', therefore it's the minimum value
-	 * for 'res_ns' timer pool parameter. When this resolution is used:
+	 * pool are defined by 'max_res.res_ns' in nanoseconds and
```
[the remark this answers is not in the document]
psavol: fixed in v2

include/odp/api/spec/timer.h
```diff
@@ -226,16 +237,16 @@ int odp_timer_capability(odp_timer_clk_src_t clk_src,
...
  * This function fills in capability limits for timer pool resolution and
  * min/max timeout values, based on either resolution or maximum timeout.
- * Set the required value to 'res_ns' or 'max_tmo', and set other fields to
- * zero. A successful call fills in the other two fields. The call returns
- * a failure, if the user defined value ('res_ns' or 'max_tmo)' exceeds
- * capability limits. Outputted values are minimums for 'res_ns' and 'min_tmo',
- * and a maximum for 'max_tmo'.
+ * Set the required value to a resolution field (res_ns or res_hz) or to the
+ * maximum timeout field (max_tmo), and set other fields to zero. A successful
+ * call fills in the other two fields. The call returns a failure, if the user
```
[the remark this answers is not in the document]
psavol: fixed in v2

include/odp/api/spec/timer.h
```diff
@@ -106,14 +106,20 @@ typedef enum {
...
+	 *  required resolution in hertz instead of nanoseconds. When this
+	 *  parameter is used, set 'res_ns' to zero. */
+	uint64_t res_hz;
+
 	/** Minimum relative timeout in nanoseconds. All requested timeouts
 	 *  will be at least this many nanoseconds after the current
 	 *  time of the timer pool. Timer set functions return an error, if too
-	 *  short timeout was requested. The value may be also less than
-	 *  'res_ns'. */
+	 *  short timeout was requested. The value may be also smaller than
```
mvl-skardach, line 121: A potential problem came to mind with specifying timeout always in ns. By the nature of ns vs Hz you'll always have rounding errors if timeout values are unsigned integers. That's because `Hz = 1/s`. I only see 3 ways of getting out of it (in no particular order of preference):

1. Accept this and note in documentation.
2. Specify parallel set of APIs for tick-based timeout specification (100Hz res gives 100 ticks available through API).
3. Specify that "timeout" values are specified in ns if `res_ns` was used or in ticks if `res_hz` was used.

psavol: I think it's not a problem in practice as a user accepts a resolution size inaccuracy on timeouts anyway. E.g. minimum resolution = 30MHz = 33.333ns, timeout 7 ticks = 233ns, should arrive after 7 or 8 ticks (233.3ns - 266.6ns), depending on when timer set call was made compared to current tick period. So, in this case additional inaccuracy from nsec vs tick API is 0.3ns (233.3ns vs 233ns), but resolution based inaccuracy is 33.333 ns (100x larger).

So, I would not change timeout units right now. We'd likely need to continue towards periodic timer API definition later on and can revisit this then. I think hz as timeout value makes sense with periodic timers.

mvl-skardach: OK.

PR #966, opened 2020-06-08
[PATCH v5] API: time stamp in nsec

psavol:


platform/linux-generic/include/odp/api/plat/time_inlines.h
```diff
@@ -44,14 +44,42 @@ static inline odp_time_t _odp_time_cur(void)
...
+	uint64_t freq_hz = _odp_time_glob.hw_freq_hz;
+	uint64_t count = time.count;
+	uint64_t sec = 0;
+
+	if (count >= freq_hz) {
+		sec   = count / freq_hz;
+		count = count - sec * freq_hz;
+	}
+
+	nsec = (1000000000ULL * count) / freq_hz;
```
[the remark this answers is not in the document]
psavol: Fixed in v2. Defined internal GIGA_HZ in this file. Time API defines (e.g. SEC_IN_NS) is not visible in this file, since the API file is included after this "inlines.h" file.

include/odp/api/spec/time.h
```diff
@@ -76,6 +76,24 @@ odp_time_t odp_time_local(void);
...
 odp_time_t odp_time_global(void);
 
+/**
+ * Current local time in nanoseconds
+ *
+ * Like odp_time_local(), but time stamp value is converted to nanoseconds.
+ *
+ * @return Local time stamp in nanoseconds
+ */
+uint64_t odp_time_local_ns(void);
```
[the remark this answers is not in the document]
psavol: The current time API handles direct "tick" time and calculations on it. Abstract odp_time_t + functions provide portability since "tick" time may consist of multiple counters (like in Posix API sec + nsec).

include/odp/api/spec/time.h
```diff
@@ -76,6 +77,24 @@ odp_time_t odp_time_local(void);
...
+
+/**
+ * Current global time in nanoseconds
+ *
+ * Like odp_time_global(), but time stamp value is converted to nanoseconds.
```
matiaselo, line 92: Perhaps  `..., but the time stamp value is converted into nanoseconds`. Same for `odp_time_local_ns ()`.

psavol: fixed in v3

configure.ac
```diff
@@ -4,7 +4,7 @@ AC_PREREQ([2.5])
 ##########################################################################
 m4_define([odpapi_generation_version], [1])
 m4_define([odpapi_major_version], [23])
-m4_define([odpapi_minor_version], [5])
+m4_define([odpapi_minor_version], [6])
```
mvl-skardach, line 7: Don't we want to wait for other pull requests (i.e. #965 and #972) before bumping the version? Version bump could be a separate commit.

psavol: When those are ready, I'll push 972, 965 and then this patch set (back-to-back).

PR #51, opened 2017-06-15
[PATCH API-NEXT v1] linux-generic: crypto: adapt HMAC code to OpenSSL 1.1.x

lumag:
OpenSSL 1.1.x has changed HMAC API in an incompatible way. Let's adapt
to it by providing version-dependent wrapper around HMAC calculation.

Signed-off-by: Dmitry Eremin-Solenikov <dmitry.ereminsolenikov@linaro.org>

PR #967, opened 2020-06-08
[PATCH v2] Miscellaneous fixes and clean up

JannePeltonen:
Fix a bunch of issues (both actual bugs and cosmetic issues) found by a static analyzer.

### DIFF
--- a/helper/cuckootable.c
+++ b/helper/cuckootable.c
@@ -260,7 +260,11 @@ odph_cuckoo_table_create(
 	pool = odp_pool_lookup(pool_name);
 
 	if (pool != ODP_POOL_INVALID)
-		odp_pool_destroy(pool);
+		if (odp_pool_destroy(pool)) {
+			odp_shm_free(shm_tbl);
+			ODPH_DBG("failed to destroy pre-existing pool\n");
+			return NULL;
+		}
 
 	odp_pool_param_init(&param);
 	param.type = ODP_POOL_BUFFER;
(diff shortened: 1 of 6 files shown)


helper/cuckootable.c
```diff
@@ -285,7 +289,7 @@ odph_cuckoo_table_create(
 	queue = odp_queue_create(queue_name, &qparam);
 	if (queue == ODP_QUEUE_INVALID) {
 		ODPH_DBG("failed to create free_slots queue\n");
-		odp_pool_destroy(pool);
+		(void) odp_pool_destroy(pool);
```
psavol, line 292: This line has a checkpatch warning:
CHECK: No space is necessary after a cast
#38: FILE: helper/cuckootable.c:292:
+		(void) odp_pool_destroy(pool);


You can run checkpatch locally e.g. like this:
git format-patch -o ../outgoing/ -6
perl scripts/checkpatch.pl ../outgoing/*

Travis also runs it as the last test case. However, Travis status ignores failure in checkpatch case.

helper/iplookuptable.c
```diff
@@ -150,7 +150,7 @@ cache_destroy(odph_iplookup_table_impl *impl)
 			sprintf(
 					pool_name, "%s_%d_%d",
 					impl->name, i, count);
-			odp_pool_destroy(odp_pool_lookup(pool_name));
+			(void) odp_pool_destroy(odp_pool_lookup(pool_name));
```
psavol, line 153: CHECK: No space is necessary after a cast
#51: FILE: helper/iplookuptable.c:153:
+			(void) odp_pool_destroy(odp_pool_lookup(pool_name));

psavol:
Git log headings could be extended with the module name when only one module is touched. E.g. "example: timer_accuracy: fix null pointer dereference when quitting after error",  "helper: cockootable: do not ignore the return value of odp_pool_destroy()", etc

This way it is easier to find e.g. l2fwd application changes when browsing through git log.

PR #969, opened 2020-06-09
[PATCH v2] api: ipsec: add parse_level IPsec sa param

vvelumuri:
Currently ODP provides selection of header parsing level as a global ipsec
configuration. Some hardawares can parse the layer headers after ipsec
transformation. As pktio information is not available during SA creation,
parse level is added as a parameter to SA creation.

This set of changes provides the configuration of parse level to each
IPsec inbound SA. It overrides the parse level of pktio with this value
for the packets specific to this SA.


JannePeltonen:
I did not quite understand the rationale for the API change. Is the global parse level setting too coarse and SA level granularity is needed for some reason?

I do not see how pktio is related to this change since the IPsec parse level in the API is not tied to pktio and since, to my understanding, the existing global IPsec setting already overrids the parse layer setting of pktio (the existing API could be more clear about this, especially for transport mode). Also, packets for the same SA can come in through any pktio.

The new parse level setting overlaps the one in global IPsec config. The API does not say which one takes affect, but it seems the intention is that the new setting takes precedence as it is more specific. But this means that the existing global setting becomes totally unnecessary in the inline op mode case as it is not possible to leave the SA specific setting unset. This should be clarified in the API.

Limiting the new setting only to the inline op mode case seems arbitrary. What is the reason for that?

Overall, I am not sure that this change is needed and that it should look quite like this if it turns out to be needed.

vvelumuri:
The granularity for parse_level at SA level provides application in controlling the parsing of packets on SA basis. 

Updated the description in the API when and which parse level will take precedence. 
Removed the limitation of this config only for inline case.

PR #970, opened 2020-06-09
[PATCH v5] build: allow per-platform lib name

mvl-skardach:
So far the build system (helper, examples, tests) expected that the ODP
library is named libodp-linux. However that's only a recommended library
name. Therefore move the library name specification to the platform
automake files. More precisely platform has to define the following:
1. In configure.m4:
   ODP_LIB_NAME=<name>
2. In Makefile.am:
   pkgconfig_DATA += lib<name>.pc
   lib_LTLIBRARIES += $(LIB)/lib<name>.la

Where <name> is the library name. Please note that lib_LTLIBRARIES has
to be explicitly named (cannot use ODP_LIB_NAME variable), otherwise
automake will fail to generate compile target.
Trying to perform a substitutions in it is only possible with full
library name substitution (see [1]), however then all possible library
names have to be defined in EXTRA_LTLIBRARIES which cannot contain
substitutions. This would effectively replace one hard-coded place with
another.

[1] https://www.gnu.org/software/automake/manual/html_node/Conditional-Libtool-Libraries.html


matiaselo:
Just a quick idea. What if we would name the pkt-config file the same on all platforms (e.g. libodp.pc)? Based on quick testing, this way the application wouldn't have to define the actual platform library name during configure and it could just use `PKG_CHECK_MODULES` as it is done today. Or does this just create confusion and possible other side effects?

psavol:
We'll merge first PR 924. This needs to be rebased then on top of that since it changes the same Makefile.inc lines.

matiaselo:
@mvl-skardach Needs rebase.

PR #971, opened 2020-06-09
[PATCH v4] Fix various LTO compile errors

mvl-skardach:
This patchset fixes several LTO related compile errors which I've noticed when building an empty platform skeleton.

### DIFF
--- a/platform/linux-generic/test/pktio_ipc/ipc_common.c
+++ b/platform/linux-generic/test/pktio_ipc/ipc_common.c
@@ -22,6 +22,7 @@ int ipc_odp_packet_send_or_free(odp_pktio_t pktio,
 	odp_pktout_queue_t pktout;
 	int i;
 
+	memset(&pktout, 0, sizeof(pktout));
 	start_time = odp_time_local();
 	wait = odp_time_local_from_ns(ODP_TIME_SEC_IN_NS);
 	end_time = odp_time_sum(start_time, wait);
--- a/test/performance/odp_packet_gen.c
+++ b/test/performance/odp_packet_gen.c
@@ -896,6 +896,9 @@ static int init_packets(test_global_t *global, int pktio,
 	uint32_t udp_src_cnt = 0;
 	uint32_t udp_dst_cnt = 0;
 
+	if (num_vlan > MAX_VLANS)
+		num_vlan = MAX_VLANS;
+
 	for (i = 0; i < num; i++) {
 		pkt = packet[i];
 		pkt_len = odp_packet_len(pkt);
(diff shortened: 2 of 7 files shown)


psavol:
Couple of old compilers (in ubuntu 16.04 and centos 7) complain about struct auto-initializations. 

make[4]: Entering directory '/odp/test/validation/api/pool'
  CC       pool.o
pool.c:83:27: error: missing field 'params' initializer
      [-Werror,-Wmissing-field-initializers]
        odp_pool_info_t info = {0};

mvl-skardach:
> Couple of old compilers (in ubuntu 16.04 and centos 7) complain about struct auto-initializations.
> 
> make[4]: Entering directory '/odp/test/validation/api/pool'
> CC pool.o
> pool.c:83:27: error: missing field 'params' initializer
> [-Werror,-Wmissing-field-initializers]
> odp_pool_info_t info = {0};

Yes, I've noticed that. It is a false positive according to C standard (http://www.open-std.org/jtc1/sc22/wg14/www/docs/n1548.pdf). §6.7.9 verse 21 says:

> If there are fewer initializers in a brace-enclosed list than there are elements or members
of an aggregate, or fewer characters in a string literal used to initialize an array of known
size than there are elements in the array, the remainder of the aggregate shall be
initialized implicitly the same as objects that have static storage duration.

I'll check manually whether `{}` produces same warnings. Otherwise we'll have a decision to make - either explicitly initialize all or disable this warning.

mvl-skardach:
Now only centos_7 complains. I'll check next week with explicit struct initializers but it does seem like a compiler strictness in CentOS. I dug out an old bug report: https://gcc.gnu.org/bugzilla/show_bug.cgi?id=36750. I'll check next week why is Ubuntu 16.04 gcc 5.4.0 not working with `{0}` as this is a valid initializer ([this comment](https://gcc.gnu.org/bugzilla/show_bug.cgi?id=36750#c9) suggest it should).

psavol:
pool.c:83:2: error: missing initializer for field 'name' of 'odp_pool_info_t' [-Werror=missing-field-initializers]
  odp_pool_info_t info = {};

I guess the easiest way is to just memset(&info, 0, sizeof(info)). I did bunch of those fixes some time back. We need to keep centos7 building, but I would not suppress the warning for all compiler versions if one has a bug.

psavol:
Upgraded checkpatch requires commit log text in addition to the subject line.

mvl-skardach:
@matiaselo Did this fail due to some timing error? I have not touched traffic manager related files here at all.

matiaselo:
> Did this fail due to some timing error? I have not touched traffic manager related files here at all.

@mvl-skardach Yep, the TM test does this occasionally. I restarted the failed test.

PR #972, opened 2020-06-09
[PATCH v4] Create missing TM print APIs

mvl-skardach:
Various Traffic Manager objects are defined in spec as separate types defined by the implementation. However while printing in unit-tests, handles are directly printed as `PRIx64` forcing implementations to utilize uint64_t (or related) type to be used for TM handles.
This patchset proposes new APIs to introduce TM debug print APIs to allow more freedom for implementation in choosing types for TM handles.

What remains to be discussed is ODP_TM_INVALID which according to spec indicates an invalid handle for each of Traffic Manager types.

### DIFF
--- a/test/validation/api/traffic_mngr/traffic_mngr.c
+++ b/test/validation/api/traffic_mngr/traffic_mngr.c
@@ -1324,7 +1324,8 @@ static int create_tm_queue(odp_tm_t         odp_tm,
 	queue_desc->tm_queues[priority] = tm_queue;
 	rc = odp_tm_queue_connect(tm_queue, tm_node);
 	if (rc != 0) {
-		ODPH_ERR("odp_tm_queue_connect() failed\n");
+		ODPH_ERR("odp_tm_queue_connect() failed for queue %" PRIx64
+			"\n", odp_tm_queue_to_u64(tm_queue));
 		odp_tm_queue_destroy(tm_queue);
 		return -1;
 	}
@@ -1652,7 +1653,8 @@ static int create_tm_system(void)
 	/* Test odp_tm_capability and odp_tm_find. */
 	rc = odp_tm_capability(odp_tm, &tm_capabilities);
 	if (rc != 0) {
-		ODPH_ERR("odp_tm_capability() failed\n");
+		ODPH_ERR("odp_tm_capability() failed for tm: %" PRIx64 "\n",
+			 odp_tm_to_u64(odp_tm));
 		return -1;
 	}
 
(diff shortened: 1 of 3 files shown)


psavol:
Commit log text needed. Also commit log subject line of the API patch should tag traffic manager: e.g. "api: tm: ..."

test/validation/api/traffic_mngr/traffic_mngr.c
```diff
@@ -1677,7 +1677,7 @@ static void dump_tm_subtree(tm_node_desc_t *node_desc)
 	rc = odp_tm_node_info(node_desc->node, &node_info);
 	if (rc != 0) {
 		ODPH_ERR("odp_tm_node_info failed for tm_node=0x%" PRIX64 "\n",
-			 node_desc->node);
+			 odp_tm_node_to_u64(node_desc->node));
```
psavol, line 1680: In addition to node_to_u64(), all new conversion functions should be used at least once (dummy usage if not really needed by the test suite). Validation tests need to in minimum verify that those functions are accessible from an application.

mvl-skardach: Fixed in v2

test/validation/api/traffic_mngr/traffic_mngr.c
```diff
@@ -1968,19 +1980,20 @@ static int destroy_all_threshold_profiles(void)
...
 	for (idx = 0; idx < NUM_WRED_PROFILES; idx++) {
 		for (color = 0; color < ODP_NUM_PKT_COLORS; color++) {
-			wred_profile = wred_profiles[idx][color];
-			if (wred_profile != ODP_TM_INVALID) {
-				rc = odp_tm_wred_destroy(wred_profile);
-				if (rc != 0) {
+			wred_prof = wred_profiles[idx][color];
+			if (wred_prof != ODP_TM_INVALID) {
+				rc = odp_tm_wred_destroy(wred_prof);
+				if (rc == 0) {
```
psavol, line 1992: Seems accidental change: rc != 0 --> rc == 0

Function return value is 0 on success.

test/validation/api/traffic_mngr/traffic_mngr.c
```diff
@@ -2211,7 +2224,8 @@ static void traffic_mngr_test_sched_profile(void)
 	for (idx = 1; idx <= NUM_SCHED_TEST_PROFILES; idx++) {
 		snprintf(sched_name, sizeof(sched_name),
 			 "sched_profile_%" PRIu32, idx);
-		for (priority = 0; priority < 16; priority++) {
+		for (priority = 0; priority < ODP_TM_MAX_PRIORITIES;
+		     priority++) {
```
psavol, line 2228: This fits into the same line now, as the max line length is 100.

test/validation/api/traffic_mngr/traffic_mngr.c
mvl-skardach: Will fix in V3

PR #973, opened 2020-06-09
[PATCH v3] Allow inlining of hash and errno APIs

mvl-skardach:
Both APIs are missing respective ABI files where inline implementation could be placed.


psavol:
I'd start the git log subject line with "abi:" instead of "api:". These do not change API specification, but ABI file structures. Application writer does not need to worry about these changes.

matiaselo:
The commits still require reviewed-by tags.

mvl-skardach:
Right, wanted to add them once I get an approval after the api->abi change.

PR #974, opened 2020-06-09
[PATCH v7] Clarify random API return expectations

mvl-skardach:
It was implied but not explicitly stated that `odp_random_data()` may return less than requested number of bytes of data in a single call (i.e. when there was entropy missing). This patchset clarifies this and adapts unit test to expect such scenario.

### DIFF
--- a/test/validation/api/random/random.c
+++ b/test/validation/api/random/random.c
@@ -9,11 +9,31 @@
 
 static void random_test_get_size(void)
 {
-	int32_t ret;
+	/* odp_random_data may fail to return data on every call (i.e. lack of
+	 * entropy). Therefore loop with some sane loop timeout value. Note that
+	 * it is not required for implementation to return data in the "timeout"
+	 * amount of steps. Rather it is a way for preventing the test to loop
+	 * forever.
+	 * Also note that the timeout value here is chosen completely
+	 * arbitrarily (although considered sane) and neither platforms or
+	 * applications are not required to use it.
+	 */
+	int32_t ret, timeout_ns = 1 * ODP_TIME_MSEC_IN_NS, sleep_ns = 100;
+	uint32_t bytes = 0;
 	uint8_t buf[32];
 
-	ret = odp_random_data(buf, sizeof(buf), ODP_RANDOM_BASIC);
-	CU_ASSERT(ret == sizeof(buf));
+	do {
+		ret = odp_random_data(buf + bytes, sizeof(buf) - bytes,
+				      ODP_RANDOM_BASIC);
+		bytes += ret;
+		if (ret < 0 || bytes >= sizeof(buf))
+			break;
+		odp_time_wait_ns(sleep_ns);
+		timeout_ns -= sleep_ns;
+	} while (timeout_ns > 0);
+
+	CU_ASSERT(ret > 0);
+	CU_ASSERT(bytes == (int32_t)sizeof(buf));
 }
 
 static void random_test_kind(void)
--- a/include/odp/api/spec/random.h
+++ b/include/odp/api/spec/random.h
@@ -72,7 +72,7 @@ odp_random_kind_t odp_random_max_kind(void);
  *                      is expected to fail if the implementation is unable to
  *                      provide the requested type.
  *
- * @return Number of bytes written
+ * @return Number of bytes written (0...len).
  * @retval <0 on failure
  */
 int32_t odp_random_data(uint8_t *buf, uint32_t len, odp_random_kind_t kind);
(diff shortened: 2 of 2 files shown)


test/validation/api/random/random.c
```diff
@@ -9,11 +9,25 @@
...
 	uint8_t buf[32];
 
-	ret = odp_random_data(buf, sizeof(buf), ODP_RANDOM_BASIC);
-	CU_ASSERT(ret == sizeof(buf));
+	do {
+		ret = odp_random_data(buf, sizeof(buf), ODP_RANDOM_BASIC);
+		if (ret < 0)
+			break;
+		bytes += ret;
+	} while (bytes < sizeof(buf) && timeout > 0);
```
Bill-Fischofer-Personal, line 27: Two points:

1. Since `timeout` is never changed from its initial value of 100, how is the `timeout > 0` clause adding anything to this test?

2. This accumulates the bytes returned, but if less than the initial request is returned, subsequent calls overwrite `buf`, meaning it's not properly filled. Need to increment the point where new bytes are inserted into `buf` as well to successfully accumulate the requested number of bytes.

mvl-skardach: Correct. Will fix in V2.

psavol:
How about odp_random_test_data() function. It is only for testing. Usage is easier if it always  returns "num" bytes. Pseudo random function should be possible to do that always.

On the other hand, it would be intuitive if both work the same way.

mvl-skardach:
> How about odp_random_test_data() function. It is only for testing. Usage is easier if it always returns "num" bytes. Pseudo random function should be possible to do that always.

I was thinking the same, hence I did not change it. I don't have a strong opinion here.

Bill-Fischofer-Personal:
Adding another API is just introducing something else to test and does not address the base question of how to properly validate the existing `odp_random_data()` API. It seems the other alternative is to either change (or clarify) the definition of `odp_random_data()` to not return until all requested bytes are provided, or else introduce an additional argument to the API that gives the minimum number of bytes needed before returning. 

The former is probably preferable since applications tend not to request more data than they need and most implementations have no problem producing the requested amount of data in the first place, so this would just avoid having to include additional logic in every application call.

psavol:
Bill, we have the odp_random_test_data() already. Both functions are specified to return "Number of bytes written". This (0 ... num) correction fits better the real random generation  than pseudo random generation. So, I'm wondering should we add (0 ... num) to random_test_data() as well, or change it to always return num.

Bill-Fischofer-Personal:
Admittedly I haven't been following the evolution of ODP closely of late, but it's not clear why the distinction between `odp_random_data()` and `odp_random_test_data()` is necessary. It's also not clear under what circumstances applications would prefer to get fewer bytes than requested, so it seems that adding that complication just makes robust invocation of these APIs that much more difficult for applications, since then every call needs to either involve these sort of loops or else needs to create its own wrapper function around them to get what it needs.

It's much simpler to say request the number of bytes you need and the API will deliver them.

mvl-skardach:
Similarly to `getrandom()` in Linux there's never a guarantee random source would return exactly the requested amount. If not then application could perform some useful operation with such "degraded" result or wait if that's needed.
By requiring looping in implementation we effectively remove such choice from application, which I believe is wrong.

Bill-Fischofer-Personal:
In that case, why not an additional parameter that specifies the minimum number of bytes needed if the application is OK with less than requested? One could even stipulate that the default minimum is 1 byte for compatibility with existing behavior. That puts the application in full control.

My expectation is that in almost all cases applications would simply specific a minimum equal to the requested length.

The alternative is to split the API into two forms: one that always returns the requested length, and a second form that allows the specification of a minimum length required as well.

mvl-skardach:
Except that for any reason (i.e. not available bandwidth, unit busy, etc.) 1 (or any minimum) might not be available and result in polling. That's why `getrandom()` is done as it is (and in fact so are `read()` calls). That's a well-known usage model, why not use it?
On the other hand most of those functions have a flag specifying requested behavior (`GRND_NONBLOCK` in `getrandom()` case) in case of no data. That would be reasonable too I believe.

Bill-Fischofer-Personal:
Then set the default minimum to 0 bytes. 

But all of this is sort of contrary to the spirit of ODP, where the philosophy is that the implementation conforms to the needs of the application rather than the other way around. It's precisely to provide a more usable and portable layer to insulate applications from platform idiosyncrasies that motivated ODP's development in the first place.

mvl-skardach:
I'm not motivated by any particular platform in this case. Even linux-generic may run out of entropy at one point in the system. My understanding of dataplane applications was that they strive to eliminate blocking at each possible step (starting from RX/TX). This decreases processing latency and gives applications ability to do useful work than yielding to the implementation.
Wasn't that one of the main reasons for introducing `odp_schedule_multi_no_wait()` and `ODP_SCHED_NO_WAIT`?
That's why I proposed adding a flag for blocking vs non-blocking random choice.

psavol:
Backwards compatibility of these two functions should be maintained. A new function could be introduced for waiting random data (likely needs timeout value also). We'd need to define at minimum what these two existing functions return. I'd suggest that we update test_data() to always return "len" on success as by definition it is pseudo random and should not have problems to return all bytes in one go.

mvl-skardach:
That should be enough. I'll update that text too.

PR #975, opened 2020-06-09
[PATCH v7] Clean-up platform-specific autoconf argument handling

mvl-skardach:
Currently all autoconf arguments appear as if they are relevant for each platform to be placed in side `platform/` directory. To allow multiple platforms to co-exist within a single repository while easily migrating to newer versions of ODP, this patchset ensures that:

1. Platform specific automake macros are invoked only for that platform.
2. Global-scope automake macros are invoked from global scope in `configure.ac`.
3. Platform specific `./configure` arguments are marked so in the help text.
4. Platforms can emit platform-specific configure status text.
5. automake does not mess-up the platform-specific AM_CONDITIONAL checking.


mvl-skardach:
@matiaselo @psavol This one should wait until #956 is integrated and then I'll have to rebase it to adapt to the shifts in configure.ac there.

example/l2fwd_simple/Makefile.am
```diff
@@ -4,9 +4,11 @@ bin_PROGRAMS = odp_l2fwd_simple
 
 odp_l2fwd_simple_SOURCES = odp_l2fwd_simple.c
 
+if PLATFORM_IS_LINUX_GENERIC
```
matiaselo, line 7: Could we just replace `HAVE_PCAP` with e.g. `_ODP_PKTIO_PCAP` (we already have this as a define)? This way each platform which supports pcap based pktio (e.g. odp-dpdk) could use this file without modifications.

platform/linux-generic/m4/configure.m4
```diff
@@ -5,30 +5,33 @@ ODP_ATOMIC
...
-AS_IF([test "$with_openssl" != "no"],
-      [ODP_OPENSSL])
-AM_CONDITIONAL([WITH_OPENSSL], [test x$with_openssl != xno])
+m4_include([platform/linux-generic/m4/odp_pcap.m4])
+m4_include([platform/linux-generic/m4/odp_scheduler.m4])
 
 AC_ARG_WITH([pcap],
 	    [AS_HELP_STRING([--without-pcap],
-			    [compile without PCAP])],
+			    [(linux-generic) compile without PCAP])],
```
matiaselo, line 13: I'd prefer moving the linux-generic text after the description with square brackets to distinguish it from default value.

matiaselo:
Couple minor comments but mostly looking OK. Needs rebase.

platform/linux-generic/m4/configure.m4
mvl-skardach: In that case shouldn't we change the default value texts to follow autoconf built-in options? I.e. running `./configure --help` gives me:
```
...
  --enable-shared[=PKGS]  build shared libraries [default=yes]
  --enable-static[=PKGS]  build static libraries [default=yes]
  --enable-fast-install[=PKGS]
                          optimize for fast installation [default=yes]
...

```

example/l2fwd_simple/Makefile.am
mvl-skardach: Sure, will fix in V5

platform/linux-generic/m4/configure.m4
matiaselo: Yes, you'r correct. So something like:
`--enable-scheduler-default  choose default scheduler [default=basic] (linux-generic)`

mvl-skardach: Right. I'm going through the current `AC_ARG_*` and re-formatting them.

platform/linux-generic/m4/odp_scheduler.m4
```diff
@@ -4,8 +4,8 @@
 AC_DEFUN([ODP_SCHEDULER], [dnl
 AC_ARG_ENABLE([scheduler-default],
 	      [AS_HELP_STRING([enable-scheduler-default],
-			      [Choose default scheduler (default is basic)])],
+			      [choose default scheduler [default=basic] (linux-generic)])],
 	      [], [enable_scheduler_default=basic])
 AC_DEFINE_UNQUOTED([_ODP_SCHEDULE_DEFAULT], ["$enable_scheduler_default"],
-		   [Define to name default scheduler])
+		   [define to name default scheduler])
```
matiaselo, line 10: No need to change this. Used for config.h (autoheader_build.h) generation, where descriptions start with capital letters.

platform/linux-generic/m4/odp_scheduler.m4
```diff
@@ -4,8 +4,8 @@
 AC_DEFUN([ODP_SCHEDULER], [dnl
 AC_ARG_ENABLE([scheduler-default],
 	      [AS_HELP_STRING([enable-scheduler-default],
```
matiaselo, line 6: Should be `[--enable-scheduler-default]`

matiaselo (approved):
Two minor comments but otherwise looking good.

PR #976, opened 2020-06-09
[PATCH v5] Various build-system fixes

mvl-skardach:
While integrating our implementations with other build-systems, such as buildroot and adding automated remote hardware testing, we've noticed several minor issues with current automake files. This patchset gathers together fixes for those issues. Related to this pull request are: https://github.com/OpenDataPlane/odp/pull/924 and https://github.com/OpenDataPlane/odp/pull/956 which I think should be handled separately.

### DIFF
--- a/m4/odp_atomic.m4
+++ b/m4/odp_atomic.m4
@@ -126,12 +126,13 @@ AC_CACHE_CHECK([whether -latomic is needed for 64-bit atomic compare exchange],
 	       [odp_cv_atomic_needed_64bit_cmp_exc], [dnl
 AC_LINK_IFELSE(
   [AC_LANG_SOURCE([[
+    #include <stdbool.h>
     #include <stdint.h>
     static uint64_t loc;
     int main(void)
     {
         uint64_t exp = 0;
-        uint64_t = __atomic_compare_exchange_n(&loc, &exp, 1, 1,
+        bool res = __atomic_compare_exchange_8(&loc, &exp, 1, 1,
                                                __ATOMIC_ACQUIRE,
                                                __ATOMIC_RELAXED);
         return 0;
@@ -157,14 +158,14 @@ AC_CACHE_CHECK([whether -latomic is needed for 128-bit atomic compare exchange],
 	       [odp_cv_atomic_needed_128bit_cmp_exc], [dnl
 AC_LINK_IFELSE(
   [AC_LANG_SOURCE([[
-    #include <stdint.h>
+    #include <stdbool.h>
     static __int128 loc;
     int main(void)
     {
         __int128 exp = 0;
-        __int128 = __atomic_compare_exchange_n(&loc, &exp, 1, 1,
-                                               __ATOMIC_ACQUIRE,
-                                               __ATOMIC_RELAXED);
+        bool res = __atomic_compare_exchange_16(&loc, &exp, 1, 1,
+                                                __ATOMIC_ACQUIRE,
+                                                __ATOMIC_RELAXED);
         return 0;
     }
     ]])],
--- a/bootstrap
+++ b/bootstrap
@@ -1,7 +1,7 @@
 #! /bin/sh
 set -x
 aclocal -I config -I m4
-libtoolize --copy
+libtoolize --copy --force
 autoheader
 automake --add-missing --copy --warnings=all
 autoconf --force
--- a/test/Makefile.inc
+++ b/test/Makefile.inc
@@ -29,7 +29,7 @@ endif
 AM_CFLAGS = $(CUNIT_CFLAGS)
 
 if STATIC_APPS
-AM_LDFLAGS = -L$(LIB) -static
+AM_LDFLAGS = -L$(LIB)
 else
 AM_LDFLAGS =
 endif


m4/odp_atomic.m4
```diff
@@ -126,12 +126,13 @@ AC_CACHE_CHECK([whether -latomic is needed for 64-bit atomic compare exchange],
...
 AC_LINK_IFELSE(
   [AC_LANG_SOURCE([[
+    #include <stdbool.h>
     #include <stdint.h>
     static uint64_t loc;
     int main(void)
     {
         uint64_t exp = 0;
-        uint64_t = __atomic_compare_exchange_n(&loc, &exp, 1, 1,
+        bool res = __atomic_compare_exchange_n(&loc, &exp, 1, 1,
```
mvl-skardach, line 135: Note that this was broken from the initial commit of this code. That is this code was never compiling (because it had invalid syntax). After fixing the syntax it seems clang linking fails on `odp_ipfragreass_atomics.h`. Naturally changing this back to effectively force `-latomic` makes the compilation pass but that is not the correct approach.

m4/odp_atomic.m4
```diff
@@ -126,12 +126,13 @@ AC_CACHE_CHECK([whether -latomic is needed for 64-bit atomic compare exchange],
...
 AC_LINK_IFELSE(
   [AC_LANG_SOURCE([[
+    #include <stdbool.h>
     #include <stdint.h>
     static uint64_t loc;
     int main(void)
     {
         uint64_t exp = 0;
-        uint64_t = __atomic_compare_exchange_n(&loc, &exp, 1, 1,
+        bool res = __atomic_compare_exchange_8(&loc, &exp, 1, 1,
```
psavol, line 135: Would addition of the bool variable fix the issue. Our code base uses compare_exchange_n(), but not compare_exchange_8(). So, I'd vote for the minimal change here.

bool res = __atomic_compare_exchange_n(...)

psavol: Noticed that commit log mentions that compare_exchange_n() didn't work on some cross compile case. If this is the case, maybe we then merge this compare_exchange_8() version and change back to compare_exchange_n() if problems arise.

mvl-skardach: Yep, it was clang that failed. Just to be sure I've added a temporary build on [my travis instance](https://travis-ci.org/github/mvl-skardach/odp/builds/703525499) changing back to _n(). Just to see if that previous failure was not some artifact.

mvl-skardach: Still fails in clang if using _n(): https://travis-ci.org/github/mvl-skardach/odp/jobs/703525540#L9500

JannePeltonen: I do not understand this and I guess I missed the explanation earlier today. Why _n is not enough? Using the explicit functions look a bit questionable to me. Are we supposed to even know they are there and named like that? And if the actual code is using _n, why is that not sufficient here? Could you elaborate?

psavol: ipfragreass code assumes that when __SIZEOF_POINTER__ == 4, compare_exchange_n() will be able to handle uint64_t types. Maybe this support should be checked instead, and disable ipfragreass compilation when it is not supported.

It may be compare_exchange_n() has additional checks (64 bit atomics support on 32 bit) and it won't compile, but compare_exchange_8() don't have such checks (as the front end _n() function have already done that). So, there may be a valid reason _n() won't compile, but direct _8() does (but could fail to a race condition on run time).

JannePeltonen: That would make sense.

mvl-skardach: > Using the explicit functions look a bit questionable to me.

Explicit `compare_exchange_n()` is used just several lines below (line [144](https://github.com/OpenDataPlane/odp/pull/976/files#diff-363fe833c8e839044373736c07845f58R144)):

```
if test "x$odp_cv_atomic_needed_64bit_cmp_exc" = "xyes" ; then
   AC_CHECK_LIB(	   AC_CHECK_LIB(
     [atomic], [__atomic_compare_exchange_8],	     [atomic], [__atomic_compare_exchange_8],
     [m4_default([$1], [:])],	     [m4_default([$1], [:])],
     [m4_default([$2], [:])])	     [m4_default([$2], [:])])
fi
```

So question is whether that check is also questionable and if so then whole odp_atomic.m4 should be reviewed. I've started with an assumption that it's acceptable.

> ipfragreass code assumes that when SIZEOF_POINTER == 4, compare_exchange_n() will be able to handle uint64_t types. Maybe this support should be checked instead, and disable ipfragreass compilation when it is not supported.

I don't think it's a problem of `compare_exchange_n()` not handling the uint64_t because it does as soon as you add `-latomic`. The problem is with clang and `_n()` version (see the build [here](https://travis-ci.org/github/mvl-skardach/odp/jobs/703525540#L445)) is that it passes the `AC_LINK_IFELSE` [check](https://travis-ci.org/github/mvl-skardach/odp/jobs/703525540#L11714) which causes `-latomic` not to be added to `ATOMIC_LIBS_OPT`. Later on it causes the ipfragreass to fail because it's missing the `-latomic`.

PR #978, opened 2020-06-11
[PATCH v3] test: packet_gen: add periodic statistics

psavol:
Added -u option to enable periodic statistics printing. Wall clock
time (since test start) is printed with per interface TX/RX
cumulative packet counts. User can calculate packet and bit rates
from these.

### DIFF
--- a/test/performance/odp_packet_gen.c
+++ b/test/performance/odp_packet_gen.c
@@ -4,6 +4,11 @@
  * SPDX-License-Identifier:     BSD-3-Clause
  */
 
+/* enable usleep */
+#ifndef _GNU_SOURCE
+#define _GNU_SOURCE
+#endif
+
 #include <stdio.h>
 #include <string.h>
 #include <stdint.h>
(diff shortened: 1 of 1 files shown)


test/performance/odp_packet_gen.c
```diff
@@ -1378,6 +1446,32 @@ int main(int argc, char **argv)
...
+		printf("\n");
+
+		t1 = odp_time_local();
+		while (odp_atomic_load_u32(&global->exit_test) == 0) {
+			usleep(1000 * global->test_options.update_msec);
+			t2 = odp_time_local();
+			nsec = odp_time_diff_ns(t2, t1);
+			print_periodic_stat(global, nsec);
+		}
+	}
```
matiaselo, line 1473: I would prefer having all printing code inside `print_periodic_stat()` function. The PR looks good othwewise.

psavol: fixed in v3

PR #979, opened 2020-06-18
[PATCH v6] api: pktio: enable reading link status information

matiaselo:
Add new function odp_pktio_link_info() and data structure
odp_pktio_link_info_t for reading current link status information:
- Autonegotiation mode (enabled/disabled)
- Duplex mode (unknown/half duplex/full duplex)
- Speed (Mbps)
- Link status (up/down)

### DIFF
--- a/include/odp/api/spec/packet_io.h
+++ b/include/odp/api/spec/packet_io.h
@@ -1,4 +1,5 @@
 /* Copyright (c) 2013-2018, Linaro Limited
+ * Copyright (c) 2020, Nokia
  * All rights reserved.
  *
  * SPDX-License-Identifier:     BSD-3-Clause
@@ -1208,6 +1209,13 @@ void odp_pktio_config_init(odp_pktio_config_t *config);
  */
 void odp_pktio_print(odp_pktio_t pktio);
 
+/** Link status */
+typedef enum odp_pktio_link_status_t {
+	ODP_PKTIO_LINK_STATUS_UNKNOWN = -1,
+	ODP_PKTIO_LINK_STATUS_DOWN = 0,
+	ODP_PKTIO_LINK_STATUS_UP = 1
+} odp_pktio_link_status_t;
+
 /**
  * Determine pktio link is up or down for a packet IO interface.
  *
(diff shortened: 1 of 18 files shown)


include/odp/api/spec/packet_io.h
```diff
@@ -1255,6 +1256,77 @@ typedef struct odp_pktio_info_t {
...
+	ODP_PKTIO_LINK_DUPLEX_UNKNOWN = 0,
+	ODP_PKTIO_LINK_DUPLEX_HALF = 1,
+	ODP_PKTIO_LINK_DUPLEX_FULL = 2
+} odp_pktio_link_duplex_t;
+
+/** Link status */
+typedef enum odp_pktio_link_status_t {
+	ODP_PKTIO_LINK_DOWN = 0,
+	ODP_PKTIO_LINK_UP = 1
+} odp_pktio_link_status_t;
```
psavol, line 1299: With addition of this enum, the current link status function could be changed to return the same type.

 * @retval  1 link is up
 * @retval  0 link is down
 * @retval <0 on failure
*/
int odp_pktio_link_status(odp_pktio_t pktio);

include/odp/api/spec/packet_io.h
```diff
@@ -1255,6 +1256,77 @@ typedef struct odp_pktio_info_t {
...
+	ODP_PKTIO_LINK_SPEED_400G   = 400000
+} odp_pktio_link_speed_t;
+
+/** Autonegotiation mode */
+typedef enum odp_pktio_link_autoneg_t {
+	/** Autonegotiation disabled */
+	ODP_PKTIO_LINK_FIXED = 0,
+	/** Autonegotiation enabled */
+	ODP_PKTIO_LINK_AUTONEG  = 1
+} odp_pktio_link_autoneg_t;
```
psavol, line 1286: Instead of "fixed", these could be named "autoneg_off, autoneg_on", or something similar

include/odp/api/spec/packet_io.h
```diff
@@ -1255,6 +1256,77 @@ typedef struct odp_pktio_info_t {
...
+typedef struct odp_pktio_link_info_t {
+	/** Link autonegotiation */
+	odp_pktio_link_autoneg_t autoneg;
+	/** Duplex mode */
+	odp_pktio_link_duplex_t duplex;
+	/** Link speed in Mbps */
+	odp_pktio_link_speed_t speed;
+	/** Link status */
+	odp_pktio_link_status_t status;
+} odp_pktio_link_info_t;
```
psavol, line 1313: Link flow control status could be also added here (pause frame on vs off)

include/odp/api/spec/packet_io.h
```diff
@@ -1255,6 +1256,77 @@ typedef struct odp_pktio_info_t {
...
+/**
+ * Packet IO link information
+ */
+typedef struct odp_pktio_link_info_t {
+	/** Link autonegotiation */
+	odp_pktio_link_autoneg_t autoneg;
+	/** Duplex mode */
+	odp_pktio_link_duplex_t duplex;
+	/** Link speed in Mbps */
+	odp_pktio_link_speed_t speed;
```
psavol, line 1310: In addition to enum values, other int speed values could be also possible. E.g. in wifi speed varies and SoC internal links can have non-standard link speeds.

include/odp/api/spec/packet_io.h
jerinjacobk: Not sure, drivers can reliability get the exact "live" status of the pause frame.  i.e by the time odp_pktio_link_status() sync call get completed, it may not report the exact status of pause frame as it will keep changing based on the peer load, unlike link up/down event where up/down takes time to transition from one state or another state.

include/odp/api/spec/packet_io.h
```diff
@@ -1255,6 +1256,77 @@ typedef struct odp_pktio_info_t {
...
+ * May be called any time with a valid pktio handle. The call is not intended
+ * for fast path use. The info structure is written only on success.
+ *
+ * @param      pktio   Packet IO handle
+ * @param[out] info    Pointer to packet IO link info struct for output
+ *
+ * @retval  0 on success
+ * @retval <0 on failure
+ */
+int odp_pktio_link_info(odp_pktio_t pktio, odp_pktio_link_info_t *info);
```
psavol, line 1328: When only some values were successfully updated, function return success and sets unknowns to other fields, Returns failure, if all are unknowns (?)

include/odp/api/spec/packet_io.h
matiaselo: I guess Petri was thinking more about the configured value (enabled/disabled) and not the live status. The current plan is to add further configuration APIs in the near future. Perhaps it would make sense to wait with the pause status until we are clear how the matching configure path is implemented.

jerinjacobk: OK. If it is not the live status then fine.

psavol: Another thing to add could be link media: e.g. twisted pair/copper, fiber, etc. Enumeration may be tricky. Copper vs fiber is likely the most common and usable values. E.g. values could be: unknown, copper, fiber, wireless, other.

include/odp/api/spec/packet_io.h
```diff
@@ -1255,6 +1256,106 @@ typedef struct odp_pktio_info_t {
...
+	ODP_PKTIO_LINK_AUTONEG_UNKNOWN = -1,
+	/** Autonegotiation disabled */
+	ODP_PKTIO_LINK_AUTONEG_OFF = 0,
+	/** Autonegotiation enabled */
+	ODP_PKTIO_LINK_AUTONEG_ON  = 1
+} odp_pktio_link_autoneg_t;
+
+/** Duplex mode */
+typedef enum odp_pktio_link_duplex_t {
+	ODP_PKTIO_LINK_DUPLEX_UNKNOWN = 0,
```
psavol, line 1298: All unknown values should be the same: in practice -1

include/odp/api/spec/packet_io.h
```diff
@@ -1255,6 +1256,106 @@ typedef struct odp_pktio_info_t {
...
+	ODP_PKTIO_LINK_SPEED_10G    = 10000,
+	ODP_PKTIO_LINK_SPEED_20G    = 20000,
+	ODP_PKTIO_LINK_SPEED_25G    = 25000,
+	ODP_PKTIO_LINK_SPEED_40G    = 40000,
+	ODP_PKTIO_LINK_SPEED_50G    = 50000,
+	ODP_PKTIO_LINK_SPEED_56G    = 56000,
+	ODP_PKTIO_LINK_SPEED_100G   = 100000,
+	ODP_PKTIO_LINK_SPEED_200G   = 200000,
+	ODP_PKTIO_LINK_SPEED_400G   = 400000
+} odp_pktio_link_speed_t;
```
psavol, line 1284: Agreed to change speeds to defines and speed field to uint32_t

platform/linux-generic/pktio/socket_common.c
```diff
@@ -160,3 +162,131 @@ int link_status_fd(int fd, const char *name)
...
+
+	if (hcmd.link_mode_masks_nwords >= 0 || hcmd.cmd != ETHTOOL_GLINKSETTINGS) {
+		ODP_ERR("ETHTOOL_GLINKSETTINGS handshake failed\n");
+		return -1;
+	}
+	hcmd.link_mode_masks_nwords = -hcmd.link_mode_masks_nwords;
+
+	/* Read link information */
+	uint8_t ODP_ALIGNED_CACHE data[offsetof(struct ethtool_link_settings, link_mode_masks) +
+				       (3 * sizeof(uint32_t) * hcmd.link_mode_masks_nwords)];
```
psavol, line 245: Some comment about max data size calculation would be helpful. Maybe few sentences from ethtool documentation.

platform/linux-generic/odp_packet_io.c
```diff
@@ -1168,23 +1168,23 @@ int odp_pktio_mac_addr_set(odp_pktio_t hdl, const void *mac_addr, int addr_size)
 	return ret;
 }
 
-int odp_pktio_link_status(odp_pktio_t hdl)
+odp_pktio_link_status_t odp_pktio_link_status(odp_pktio_t hdl)
 {
 	pktio_entry_t *entry;
 	int ret = -1;
```
psavol, line 1174: Should initialize ret to ODP_PKTIO_LINK_STATUS_UNKNOWN

platform/linux-generic/pktio/socket_common.c
```diff
@@ -157,7 +157,7 @@ int link_status_fd(int fd, const char *name)
 		__odp_errno = errno;
 		ODP_DBG("ioctl(SIOCGIFFLAGS): %s: \"%s\".\n", strerror(errno),
 			ifr.ifr_name);
-		return -1;
+		return ODP_PKTIO_LINK_STATUS_UNKNOWN;
 	}
 
 	return !!(ifr.ifr_flags & IFF_RUNNING);
```
psavol, line 163: This would need an if-else to return STATUS_UP vs _DOWN (instead of 0 vs 1)

test/validation/api/pktio/pktio.c
```diff
@@ -1480,6 +1481,43 @@ static void pktio_test_info(void)
...
+	odp_pktio_t pktio;
+	odp_pktio_link_info_t link_info;
+	int i;
+
+	for (i = 0; i < num_ifaces; i++) {
+		pktio = create_pktio(i, ODP_PKTIN_MODE_QUEUE,
+				     ODP_PKTOUT_MODE_DIRECT);
+		CU_ASSERT_FATAL(pktio != ODP_PKTIO_INVALID);
+
+		CU_ASSERT_FATAL(odp_pktio_link_info(pktio, &link_info) == 0);
```
psavol, line 1495: memset(&link_info, 0, sizeof(link_info))

would ensure that each successful call really fills in the info

PR #980, opened 2020-06-18
[PATCH v2] example: timer: fix LTO compilation

mvl-skardach:
GCC7.3 warns about uninitialized members of timer pool parameters.
Application should take care to ensure clean parameter initialization.

### DIFF
--- a/example/timer/odp_timer_simple.c
+++ b/example/timer/odp_timer_simple.c
@@ -69,6 +69,7 @@ int main(int argc ODP_UNUSED, char *argv[] ODP_UNUSED)
 		ret += 1;
 		goto err_tp;
 	}
+	memset(&tparams, 0, sizeof(tparams));
 	tparams.res_ns = MAX(10 * ODP_TIME_MSEC_IN_NS,
 			     timer_capa.highest_res_ns);
 	tparams.min_tmo = 10 * ODP_TIME_MSEC_IN_NS;


example/timer/odp_timer_simple.c
```diff
@@ -69,6 +69,7 @@ int main(int argc ODP_UNUSED, char *argv[] ODP_UNUSED)
 		ret += 1;
 		goto err_tp;
 	}
+	memset(&tparams, 0, sizeof(tparams));
```
Bill-Fischofer-Personal, line 72: Isn't this really exposing the omission of an `odp_timer_pool_param_init()` API?

mvl-skardach: Sure, though that API would have to zero out all parameters in light of #965 - since user has to choose which resolution he wants. I don't have a strong opinion here though. Petri might have more insight why that API is not there.

matiaselo: Bill may know better, but I suspect `odp_timer_pool_create()` API was created before we started the practice of adding init functions for parameters. At least from API consistency point of view adding `odp_timer_pool_param_init()` could be beneficial . I'll add it the API todo-list.

For now this fix is needed.

Bill-Fischofer-Personal: @matiaselo Yes, you are correct. The init for the `odp_xxx_param_init()` functions was identified later but I had though by now ODP had added them all. Glad to hear this will be on the to-do list as this one clearly was an omission to that practice.

PR #982, opened 2020-06-25
[PATCH v3] GCC10 LTO compilation fixes

mvl-skardach:
While testing compilation against gcc 10.0.1 I've noticed several issues which this patchset is trying to address:

1. GCC 10 version detection is done improperly.
2. Some tests have odp_pktout_queue_t and odp_queue_t left uninitialized in some paths detected by the compiler. For safety initialize them to known invalid values.

### DIFF
--- a/test/validation/api/traffic_mngr/traffic_mngr.c
+++ b/test/validation/api/traffic_mngr/traffic_mngr.c
@@ -3292,6 +3292,7 @@ static int test_query_functions(const char *shaper_name,
 	expected_pkt_cnt  = num_pkts - 2;
 	expected_byte_cnt = expected_pkt_cnt * PKT_BUF_SIZE;
 
+	memset(&query_info, 0, sizeof(query_info));
 	rc = odp_tm_queue_query(tm_queue,
 				ODP_TM_QUERY_PKT_CNT | ODP_TM_QUERY_BYTE_CNT,
 				&query_info);
@@ -3301,6 +3302,7 @@ static int test_query_functions(const char *shaper_name,
 	CU_ASSERT(query_info.total_byte_cnt_valid);
 	CU_ASSERT(expected_byte_cnt < query_info.total_byte_cnt);
 
+	memset(&query_info, 0, sizeof(query_info));
 	rc = odp_tm_priority_query(odp_tm_systems[0], priority,
 				   ODP_TM_QUERY_PKT_CNT | ODP_TM_QUERY_BYTE_CNT,
 				   &query_info);
@@ -3310,6 +3312,7 @@ static int test_query_functions(const char *shaper_name,
 	CU_ASSERT(query_info.total_byte_cnt_valid);
 	CU_ASSERT(expected_byte_cnt < query_info.total_byte_cnt);
 
+	memset(&query_info, 0, sizeof(query_info));
 	rc = odp_tm_total_query(odp_tm_systems[0],
 				ODP_TM_QUERY_PKT_CNT | ODP_TM_QUERY_BYTE_CNT,
 				&query_info);
--- a/configure.ac
+++ b/configure.ac
@@ -136,7 +136,7 @@ ODP_CHECK_CFLAG([-Wno-error=address-of-packed-member])
 # GCC 10 sometimes gets confused about object sizes and gives bogus warnings.
 # Make the affected warnings generate only warnings, not errors.
 AS_IF([test "$GCC" == yes],
-      AS_IF([test `$CC -dumpversion` -ge 10],
+      AS_IF([test `$CC -dumpversion | cut -d '.' -f 1` -ge 10],
 	    ODP_CHECK_CFLAG([-Wno-error=array-bounds])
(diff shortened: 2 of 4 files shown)


JannePeltonen:
The commit comments say that potentially un-initialized variables get fixed. It is somewhat misleading since in all of the cases the variables are, AFAICS, initialized before first use. Apparently GCC could not analyse the code deep enough. So I think the commit comments should be changed to make it clear that this is not about fixing actual bugs but making the compiler happy.

But I wonder to what extent we want to sprinkle the code with unnecessary clutter just to make random static analyzers less confused. Where does it stop? Is there no better way in this case?

I suppose the subject lines should include module tags as is customary in ODP.

test/validation/api/traffic_mngr/traffic_mngr.c
```diff
@@ -3289,6 +3289,7 @@ static int test_query_functions(const char *shaper_name,
 	expected_pkt_cnt  = num_pkts - 2;
 	expected_byte_cnt = expected_pkt_cnt * PKT_BUF_SIZE;
 
+	memset(&query_info, 0, sizeof(query_info));
 	rc = odp_tm_queue_query(tm_queue,
```
matiaselo, line 3293: This seems more like a bug in the test itself. If `odp_tm_queue_query()` fails the test still tries to read `query_info` contents. Does replacing the following CU_ASSERT with CU_ASSERT_FATAL fix the warning? The same issue is repeated with each query call in `test_query_functions`.

mvl-skardach: Nope, doesn't help.

mvl-skardach:
> The commit comments say that potentially un-initialized variables get fixed. It is somewhat misleading since in all of the cases the variables are, AFAICS, initialized before first use. Apparently GCC could not analyse the code deep enough. So I think the commit comments should be changed to make it clear that this is not about fixing actual bugs but making the compiler happy.
Fair enough. Will fix in V2.

> But I wonder to what extent we want to sprinkle the code with unnecessary clutter just to make random static analyzers less confused. Where does it stop? Is there no better way in this case?
Aside of disabling `-Wmaybe-uninitialized` globally (or locally per test) I don't know any other solution. I'd welcome any suggestions.

> I suppose the subject lines should include module tags as is customary in ODP.
Decided not to add them because I'd have to add multiple modules for the validation patch. That would make title more confusing to me but if you prefer it that way, I'll add them.

mvl-skardach:
@JannePeltonen @matiaselo Before I go around making V2 let's decide. Should I:
1. Disable `-Wmaybe-uninitialized` globally.
2. Disable `-Wmaybe-uninitialized` per-test (with `#pragma`).
3. Continue with `memset()` approach.

JannePeltonen:
I was just about to suggest -Wnoerror=maybe-uninitialized globally if GCC version is 10 and LTO is enabled. I think the pragma may not even work for the link time phase.

I am just trying LTO compilation myself and I get additional errors/warnings from the TM test. When I turned the warning to just warning with the -Wnoerror flag the build failed with and internal compiler error with the GCC 10.0.1 of ubuntu but succeeded with GCC 10.1 in Fedora. I am not yet sure what happened.

JannePeltonen:
I gave up trying to use the experimental 10.0.1 version of GCC.

With GCC 10.1 I did not need all these patches for the lto build to succeed. I needed the traffic_mngr.c change that is part of the third commit. Then I needed an additional change in traffic_mngr.c: memsetting shaper_params to zero in check_shaper_profile(). That is all. If this were all that is needed, then I would not touch the warning settings.

I wonder what is different in your environment. I tested in Fedora 32 on x86 with only --enable-lto config flag.

mvl-skardach:
> I wonder what is different in your environment. I tested in Fedora 32 on x86 with only --enable-lto config flag.

I'm cross-compiling on x86 for aarch64. As for flags: --enable-lto and -O3. I'll re-check on Monday whether I have not put some extra changes which are not needed.

JannePeltonen:
I just made PR #987 that improves CU_ASSERT_FATAL() among other things and should help getting rid of many of the bogus warnings in code paths following CU_ASSERT_FATAL(). If that PR gets accepted, the pktio.c changes of the third commit of this PR should not be needed.

I think we may not need -Wno-error=maybe-uninitialized after all.

test/validation/api/traffic_mngr/traffic_mngr.c
JannePeltonen: If PR #987 gets merged, CU_ASSERT_FATAL should get rid of the warning here.

However, there is an even bigger bug in the test: It does not actually test if the latter query functions write the query info at all. If the first query succeeds, query_info already contains the right content even if the implementation of the later called query functions were simply {return 0;}. The test would not notice anything wrong. So clearing the query info content is needed because of that and then the CU_ASSERT versus CU_ASSERT_FATAL issue becomes moot.

mvl-skardach:
I've checked against PR #987 and eliminated not necessary commits.

PR #984, opened 2020-06-26
[PATCH v4] linux_gen: example: ipsec: fix stream input packet allocation

matiaselo:
The application tried to allocate packets of length zero, which is not
allowed by the API.

### DIFF
--- a/example/ipsec_api/odp_ipsec.c
+++ b/example/ipsec_api/odp_ipsec.c
@@ -1020,6 +1020,10 @@ main(int argc, char *argv[])
 	/* If we have test streams build them before starting workers */
 	resolve_stream_db();
 	stream_count = create_stream_db_inputs();
+	if (stream_count < 0) {
+		ODPH_ERR("Error: creating input packets failed\n");
+		exit(EXIT_FAILURE);
+	}
 
 	/*
 	 * Create and init worker threads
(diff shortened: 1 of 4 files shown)


example/ipsec_api/odp_ipsec.c
```diff
@@ -109,6 +109,10 @@ static void usage(char *progname);
 #define SHM_PKT_POOL_BUF_SIZE  4096
 #define SHM_PKT_POOL_SIZE      (SHM_PKT_POOL_BUF_COUNT * SHM_PKT_POOL_BUF_SIZE)
 
+/* Test stream input packet size */
```
JannePeltonen, line 112: This does not appear to be the actual packet size but the (initial) size of the packet buffer that then gets filled with the packet of whatever length it happens to have.

The IPv4 packet creation function still looks fishy: at one point ICMP data of stream->length bytes gets written into the packet. I suppose the length could exceed the fixed packet buffer size as there is no check and as the length appears to come from the user. I also did not see any check that the packet data is actually contiguous in memory.

matiaselo: Yes, it's the initial packet length. Now when you mentioned it, I finally found the undocumented '-s' command line argument, which can be used to select the test packet payload length. Will post V2.

PR #985, opened 2020-06-26
[PATCH v4] api: tm: info structures are written only on success

matiaselo:
Clarify that info structures (odp_tm_node_info_t, odp_tm_node_fanin_info_t,
odp_tm_queue_info_t, odp_tm_query_info_t) are only written on success.
Invalid out tags were removed from query flags.


psavol:
The commit log heading could be shorter, so that it fits in single line.

Now:
 api: traffic_mngr: clarify that info structures are only written on s…
…uccess

Could be e.g.:
api: tm; info structures are written only on success

PR #986, opened 2020-06-28
[PATCH v2] Use atomic flag variable to trigger thread exit

JannePeltonen:
Currently various test and example apps use a normal variable without synchronization to communicate to worker threads that they should exit. Even though this is wrong (it is a data race and thus allows undefined behavior) it often works in practice unless the compiler performs so aggressive optimizations that the flag check moves away from the thread main loop. Something like this actually happened to me with GCC-10 and link time optimization, causing make check to never complete.

Some apps use volatile flag variable which probably always works in practice even though it is still wrong (it is still a data race as volatile does not imply synchronization or atomicity). The commits of this PR do not touch the volatile flags.

### DIFF
--- a/example/l2fwd_simple/odp_l2fwd_simple.c
+++ b/example/l2fwd_simple/odp_l2fwd_simple.c
@@ -23,7 +23,7 @@ typedef struct {
 	odp_pktout_queue_t if0out, if1out;
 	odph_ethaddr_t src, dst;
 	odp_shm_t shm;
-	int exit_thr;
+	odp_atomic_u32_t exit_thr;
 	int wait_sec;
 } global_data_t;
 
@@ -34,7 +34,7 @@ static void sig_handler(int signo ODP_UNUSED)
 	printf("sig_handler!\n");
 	if (global == NULL)
 		return;
-	global->exit_thr = 1;
+	odp_atomic_store_u32(&global->exit_thr, 1);
 }
 
 static odp_pktio_t create_pktio(const char *name, odp_pool_t pool,
@@ -105,7 +105,7 @@ static int run_worker(void *arg ODP_UNUSED)
 	printf("started output interface\n");
 	printf("started all\n");
 
-	while (!global->exit_thr) {
+	while (!odp_atomic_load_u32(&global->exit_thr)) {
 		pkts = odp_pktin_recv_tmo(global->if0in, pkt_tbl, MAX_PKT_BURST,
 					  wait_time);
 
@@ -189,6 +189,7 @@ int main(int argc, char **argv)
 	}
 
 	memset(global, 0, sizeof(global_data_t));
+	odp_atomic_init_u32(&global->exit_thr, 0);
 	global->shm = shm;
 
 	if (argc > 7 ||
--- a/example/l3fwd/odp_l3fwd.c
+++ b/example/l3fwd/odp_l3fwd.c
@@ -90,7 +90,7 @@ typedef struct {
 	/** Shm for storing global data */
 	odp_shm_t shm;
 	/** Break workers loop if set to 1 */
-	int exit_threads;
+	odp_atomic_u32_t exit_threads;
 
 	/* forward func, hash or lpm */
 	int (*fwd_func)(odp_packet_t pkt, int sif);
@@ -323,7 +323,7 @@ static int run_worker(void *arg)
 
 	odp_barrier_wait(&global->barrier);
 
-	while (!global->exit_threads) {
+	while (!odp_atomic_load_u32(&global->exit_threads)) {
 		if (num_pktio > 1) {
 			if_idx = input_ifs[pktio];
 			inq = input_queues[pktio];
@@ -972,6 +972,7 @@ int main(int argc, char **argv)
 	}
 
 	memset(global, 0, sizeof(global_data_t));
+	odp_atomic_init_u32(&global->exit_threads, 0);
 	global->shm = shm;
 
 	/* Initialize the dest mac as 2:0:0:0:0:x */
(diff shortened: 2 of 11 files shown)


mvl-skardach:
LGTM. The volatile is a necessary change, atomic may be an overkill.
Reason being that in both cases you need another synchronization method to wait for threads to actually exit before doing any changes that might affect them.
Atomicity is not required because there is only 1 writer to that variable, there's no compare-and-swap logic. What is necessary is marking that this variable is not cache'able for read/write. Hence volatile is enough.

JannePeltonen:
Volatile alone is not enough since it is not meant for inter-thread communication. Having one thread write to a volatile variable and another one read it at the same time without any synchronization is a data race and thus technically can cause undefined behavior. In practice, in this case it works even  though it is wrong.

The atomic variable is not for the atomicity per-se but to allow sharing the variable between the threads in a correct way (from C std point-of-view), i.e. without introducing a data race.

Changing the flag variable to atomic also prevents the compiler from moving the load outside the main loop of the thread. This is what happens in practice, but now that I think of it more, it does not actually seem to be guaranteed by the C standard unless the main loop contains some synchronization ops that prevent such transformation (which is very likely). To make sure the flag load cannot be moved outside the loop no matter how simple the loop body is, I think we would have to use sequentially consistent memory order with it or use a suitable barrier.

Making the flag volatile in addition to atomic would work too but we cannot do it with ODP atomics because the volatile qualifier would get discarded when calling the functions and that would fail the build.

Other synchronization needs than signaling the need for termination is another issue and needs separate handling. I think in most cases odph_odpthreads_join() and similar take care of it.

mvl-skardach:
> The atomic variable is not for the atomicity per-se but to allow sharing the variable between the threads in a correct way (from C std point-of-view), i.e. without introducing a data race.

Right, but do we care for preventing the race condition? The worst case that can happen is one thread reading an old value or perhaps a corrupted non-zero value. This will also happen only when application should exit. Since in all cases we're continuing the loop on 0 and ending on non-zero value the thread will either exit early or 1 loop later. In both cases odph_odpthreads_join() will act as a synchronization which catches both cases.
That said atomic variable is also correct.

> Changing the flag variable to atomic also prevents the compiler from moving the load outside the main loop of the thread.

If volatile read is done inside the loop (and doing it in `while(reading_the_var)` counts as part of the loop`) has the same effect. Otherwise compiler would violate volatile keyword. I mean otherwise how would it guarantee that `reading_the_var` is not done from a core register but rather from a memory (even L2 cache)?

> Making the flag volatile in addition to atomic would work too but we cannot do it with ODP atomics because the volatile qualifier would get discarded when calling the functions and that would fail the build.

atomic implies volatile, otherwise it makes no sense.

JannePeltonen:
> Right, but do we care for preventing the race condition?

I think at least in example apps it is not very nice to have a construct that causes undefined behavior from language definition point-of-view.

> The worst case that can happen is one thread reading an old value or perhaps a corrupted non-zero value.

In theory anything can happen (like nethack starting), but in practice in this case I cannot see how anything problematic could happen (apart from possibly restricting optimization slightly more than necessary). I just did not want to do a change that is wrong from language spec perspective.

> If volatile read is done inside the loop (and doing it in `while(reading_the_var)` counts as part of the loop) has the same effect.

Yes, it does, I am not arguing that. I meant that the use of atomic compared to a normal non-volatile and non-atomic variable prevents the moving the load out of the loop (except not always, in theory).

> atomic implies volatile, otherwise it makes no sense.

No, it does not imply it. Volatile and atomic are different, quite orthogonal concepts, even though the have some similarity. Accesses of volatile variable specify are part of the externally observable boundary of the abstract machine but atomic variables are for multi-threading inside the abstract machine. Volatile variables, atomic variables and volatile atomic variables are all different from each other.

So, I think we can do one of:

1) Use volatile. It would work in practice but it would be wrong.
2) Use relaxed atomic. It would work in practice and be ok from language point-of-view. Theoretically the compiler would be allowed (depending on other things in the main loop) to optimize the load out of the loop but I do not think any compiler would do it.
3) Maybe use sequentially consistent load, or relaxed one with mb_full(), or something similar, but I think that would be overkill.
4) Use volatile atomic variable. This would work but we cannot do it using the atomic operations provided by ODP API.

I am suggesting 2).

mvl-skardach:
I don't mind the solution nb 2.

> Use relaxed atomic. It would work in practice and be ok from language point-of-view. Theoretically the compiler would be allowed (depending on other things in the main loop) to optimize the load out of the loop but I do not think any compiler would do it.

I don't believe this assumption about the compiler is true at all. In the following code:
```
while (!my_atomic_load(&exit_threads)) {
  // do something
}
```
1. If compiler would move the `my_atomic_load()` out of the loop the code becomes semantically different.
2. Compiler has no knowledge on whether `exit_threads` is modified in another thread or not. The whole point of atomic variable is to assume it is, so again moving it out of the loop is not valid.
3. C11 atomics take [volatile parameters](https://en.cppreference.com/w/c/atomic/atomic_load).
4. When looking at how atomic variable accesses are actually implemented, they are a combination of exclusive load/stores and/or memory barriers (all depending on a platform). Similarly implementation of volatile involves always reading the variable from memory, never from cached register value. One CANNOT implement atomic operations without ensuring their results are observable, not only to current thread but also to others. In other words both volatile and atomic never rely on register-cached values.
5. Compiler may definitely re-order volatile and other instructions in the loop. Unlikely such thing would be done for atomics since memory barriers are also compiler barriers.

JannePeltonen:
What is intuitively right is not necessarily what the standard guarantees. Atomic operations are allowed to be rearranged a lot more than volatile accesses, even though a good compiler might choose not to do so. See e.g. https://stackoverflow.com/questions/45960387/why-dont-compilers-merge-redundant-stdatomic-writes and the links in the answer.

There are no memory barriers and such involved with relaxed atomic ops since they are not synchronization operations at all.

The fact that the atomic operations of the standard library have volatile qualifiers for the arguments is so that they can take both volatile and non-volatile parameters. It does not, AFAIK, imply that non-volatile atomic variables get volatile semantics.

mvl-skardach:
Thank you, that was quite informative. Atomics seem to be the most proper from the C/C++ standard point of view and what we can express.

PR #988, opened 2020-06-29
[PATCH v9] api: ipsec: add tm_queue as parameter for inline outbound

anoobj:
Inline outbound requires pktio output queue to allow application to
control the traffic distribution and precedence. The option to specify
output queue is required to perform QoS post inline IPsec processing.

### DIFF
--- a/include/odp/api/spec/ipsec.h
+++ b/include/odp/api/spec/ipsec.h
@@ -22,6 +22,7 @@ extern "C" {
 #include <odp/api/support.h>
 #include <odp/api/packet_io.h>
 #include <odp/api/classification.h>
+#include <odp/api/traffic_mngr.h>
 
 /** @defgroup odp_ipsec ODP IPSEC
  *  IPSEC protocol offload.
@@ -261,6 +262,15 @@ typedef struct odp_ipsec_capability_t {
 
 	/** Supported authentication algorithms */
 	odp_crypto_auth_algos_t   auths;
+
+	/** Support of traffic manager (TM) after inline outbound IPSEC
+	 *  processing. On unsupported platforms, application is not allowed
+	 *  to use a TM enabled pktio (ODP_PKTOUT_MODE_TM) with outbound
+	 *  inline IPSEC.
+	 *
+	 *  @see odp_pktio_open(), odp_pktio_param_t
+	 */
+	odp_support_t inline_ipsec_tm;
 
 } odp_ipsec_capability_t;
 
@@ -1148,13 +1158,28 @@ typedef struct odp_ipsec_in_param_t {
  * Outbound inline IPSEC operation parameters
  */
 typedef struct odp_ipsec_out_inline_param_t {
-	/** Packet output interface for inline output operation
+	/** Packet output interface for inline outbound operation without TM
 	 *
 	 *  Outbound inline IPSEC operation uses this packet IO interface to
 	 *  output the packet after a successful IPSEC transformation. The pktio
 	 *  must have been configured to operate in inline IPSEC mode.
+	 *
+	 *  The pktio must not have been configured with ODP_PKTOUT_MODE_TM.
+	 *  For IPSEC inline output to TM enabled interfaces set this field
+	 *  to ODP_PKTIO_INVALID and specify the TM queue to be used through
+	 *  the tm_queue parameter. Inline IPSEC output through TM can be
+	 *  done only if the platform has inline_ipsec_tm capability.
 	 */
 	odp_pktio_t pktio;
+
+	/** TM queue for inline outbound operation
+	 *
+	 *  TM queue to be used for inline IPSEC output when pktio field
+	 *  is ODP_PKTIO_INVALID, indicating use of TM. Otherwise ignored.
+	 *
+	 *  @see odp_ipsec_capability()
+	 */
+	odp_tm_queue_t tm_queue;
 
 	/** Outer headers for inline output operation
 	 *
@@ -1489,8 +1514,8 @@ int odp_ipsec_out_enq(const odp_packet_t pkt[], int num,
  *
  * This operation does outbound inline IPSEC processing for the packets. It's
  * otherwise identical to odp_ipsec_out_enq(), but outputs all successfully
- * transformed packets to the specified output interface, instead of generating
- * events for those.
+ * transformed packets to the specified output interface (or tm_queue), instead of
+ * generating events for those.
  *
  * Inline operation parameters are defined per packet. The array of parameters
  * must have 'num' elements and is pointed to by 'inline_param'.
--- a/include/odp/api/spec/packet_io.h
+++ b/include/odp/api/spec/packet_io.h
@@ -516,13 +516,14 @@ typedef struct odp_pktio_config_t {
 	 *  to the pktio interface for output. IPSEC configuration is done
 	 *  through the IPSEC API.
 	 *
-	 *  Outbound IPSEC inline operation cannot be combined with traffic
-	 *  manager (ODP_PKTOUT_MODE_TM).
+	 *  Support of outbound IPSEC inline operation with traffic manager
+	 *  (ODP_PKTOUT_MODE_TM) can be queried with odp_ipsec_capability().
 	 *
-	 *  0: Disable outbound IPSEC inline operation (default)
-	 *  1: Enable outbound IPSEC inline operation
+	 * * 0: Disable outbound IPSEC inline operation (default)
+	 * * 1: Enable outbound IPSEC inline operation
 	 *
 	 *  @see odp_ipsec_config(), odp_ipsec_sa_create()
+	 *  odp_ipsec_out_inline()
 	 */
 	odp_bool_t outbound_ipsec;
 
--- a/platform/linux-generic/odp_ipsec.c
+++ b/platform/linux-generic/odp_ipsec.c
@@ -63,6 +63,7 @@ int odp_ipsec_capability(odp_ipsec_capability_t *capa)
 		return rc;
 
 	capa->max_queues = queue_capa.max_queues;
+	capa->inline_ipsec_tm = ODP_SUPPORT_NO;
 
 	return 0;
 }
@@ -1821,6 +1822,7 @@ int odp_ipsec_out_inline(const odp_packet_t pkt_in[], int num_in,
 	unsigned opt_inc = (param->num_opt > 1) ? 1 : 0;
 
 	ODP_ASSERT(param->num_sa != 0);
+	ODP_ASSERT(inline_param->pktio != ODP_PKTIO_INVALID);
 
 	while (in_pkt < num_in) {
 		odp_packet_t pkt = pkt_in[in_pkt];
--- a/test/validation/api/ipsec/ipsec.c
+++ b/test/validation/api/ipsec/ipsec.c
@@ -760,6 +760,7 @@ static int ipsec_send_out_one(const ipsec_test_part *part,
 			memset(hdr, 0xff, hdr_len);
 		}
 		inline_param.pktio = suite_context.pktio;
+		inline_param.tm_queue = ODP_TM_INVALID;
 		inline_param.outer_hdr.ptr = hdr;
 		inline_param.outer_hdr.len = hdr_len;
 


include/odp/api/spec/ipsec.h
```diff
@@ -1145,16 +1146,25 @@ typedef struct odp_ipsec_in_param_t {
 } odp_ipsec_in_param_t;
 
 /**
- * Outbound inline IPSEC operation parameters
+ * Outbound inline IPsec operation parameters
```
psavol, line 1149: You are right that IPsec is better spelling. However, I'd prefer that spelling would be changed once (for the entire file) in a separate patch.

anoobj: RFC (https://tools.ietf.org/html/rfc4301) deprecates the usage of IPSEC. So using IPSEC to in the new code added didn't seem right. I can switch to IPSEC for all usages. Or I can submit a parent patch fixing the entire file. Can you suggest what would be the right approach?

JannePeltonen:
I wonder if it is good to have the possibility to specify the output TM queue as part of the output operation instead of setting it for the SA. It is more difficult for the implementation and it is generally not advisable to use the same SA for different QoS classes because packet reordering is problematic with replay prevention. IPsec RFCs suggest having a separate SA for each QoS class.

OTOH, if per-packet TM queue selection turns out problematic with some implementation in the future, maybe we could then add per-SA selection and some capability flags to allow the simpler model.

On the third hand, the existing pktio selection is per operation to allow interface changes on the fly (e.g. for routing changes or LAG), so maybe the same should apply to the TM queue selection.

JannePeltonen:
I am not sure what is the reason for making it possible to specify pktout queue in the inline output operation. What is the intended externally observable behavior that is sought with it?

Different SAs can already use different implementation-internal output queues for e.g. better parallelism even if in the API the application configures single output queue for the pktout. The queues used internally between inline IPsec and the actual packet output do not necessarily have to match the API visible output queues, or if they had to, the API would have to explain why and how.

If one needs to extract parallelism (e.g. for multiple IPsec processing engines) within single SA, then maybe instead of an explicit queue parameter there could be some sort of flow hash or entropy parameter that the implementation could use without the application having to know how many internal queues there really are? OTOH, having different queuing behavior for different packets of the same SA is problematic for replay prevention.

JannePeltonen:
Does the API need some capability flags that tell if combining inline IPsec with TM is possible?

anoobj:
> I am not sure what is the reason for making it possible to specify pktout queue in the inline output operation. What is the intended externally observable behavior that is sought with it?

When IPsec is done in non-inline mode, application is allowed to use any pktout_queue or tm_queue for sending the packet out. The QoS would be done based on DSCP field and so packets from SA may or may not end up on the same queue (be it pktout_queue/tm_queue).

Intent of this patch is to allow the same behavior with inline IPsec in outbound. The current spec already allows packets from the same SA be send out on different pktio interfaces. This proposal is to change the granularity to queue level so that application can decide further on the packet transmit parameters.

anoobj:
> Does the API need some capability flags that tell if combining inline IPsec with TM is possible?

Yes. That would be useful. Also, to avoid ambiguity in choosing tm_queue vs pktout_queue, I'll try to introduce a config flag as well.

include/odp/api/spec/ipsec.h
anoobj: I've replaced IPsec with IPSEC as part of v4.

matiaselo:
FYI the Travis i386 test failures are not caused by this PR and will be fixed by PR #1016.

JannePeltonen:
Maybe I start to sound like a broken record but what is the intended externally observable effect (that applies to all ODP implementations) that you would get when specifying pktout queue in the ipsec inline output op? Why is pktout queue parameter needed, i.e. what problem does it solve?

The point being able to specify TM queue for combining inline IPsec with TM is more clear.

include/odp/api/spec/packet_io.h
```diff
@@ -512,18 +512,38 @@ typedef struct odp_pktio_config_t {
...
+	 *  When enabled, odp_ipsec_out_inline() would use TM queue for packet
+	 *  transmission.
+	 *
+	 * * 0: Disable outbound IPSEC inline operation with TM queues (default)
+	 * * 1: Enable outbound IPSEC inline operation with TM queues
+	 *
+	 *  @see odp_ipsec_config(), odp_ipsec_sa_create(),
+	 *  odp_ipsec_out_inline()
+	 */
+	odp_bool_t outbound_ipsec_tm_queue;
```
JannePeltonen, line 546: Why does this need to be introduced instead of relying on the out_mode (e.g. ODP_PKTOUT_MODE) configured for the pktout? It would seem simpler and more clear that pktout either uses or does not use TM based on the existing mode config and inline IPsec just follows that.

This additional setting would seem to allow one to configure the use or non-use of TM separately for normal output operations and IPsec inline output operations. I am not sure if that is a good thing. It may make implementation harder in some environments.

anoobj: That should be fine. Thanks for the suggestion. So what you propose is 

if (odp_pktio_param_t.out_mode == ODP_PACKET_MODE_TM) {
    /* Use TM queues */
} else {
    /* Use pktout */
}

Above should be fine. 

To answer your other question,

> Why is pktout queue parameter needed, i.e. what problem does it solve?

Because for a regular lookaside mode of operation, the packet being transmitted after IPsec can be at pktout queue level. With inline that will not be possible and any application which is trying to use the pktout queues for distribution may find that missing.

anoobj:
@JannePeltonen @psavol Any comments on the latest version?

include/odp/api/spec/ipsec.h
```diff
@@ -262,6 +263,14 @@ typedef struct odp_ipsec_capability_t {
 	/** Supported authentication algorithms */
 	odp_crypto_auth_algos_t   auths;
 
+	/** Support of traffic manager (TM) after inline IPSEC processing. On
```
JannePeltonen, line 266: It could be mentioned that this is about outbound IPsec only: i.e. "...after outbound inline IPsec processing"

include/odp/api/spec/ipsec.h
```diff
@@ -262,6 +263,14 @@ typedef struct odp_ipsec_capability_t {
...
+	 *  unsupported platforms, inline IPSEC outbound operations would be
```
JannePeltonen, line 267: This second sentence could be just removed as we do not need to promise this sort of behavior. The two basics behavior we can choose from for the case the capability is not there, are: 1) Application is not allowed to use a pktio with TM enabled with inline IPsec. 2) Application is allowed to use such pktio with inline IPsec but the IPsec packets do not go through TM. Petri an I think that the first one is the better choice. My other comments are based on that.

include/odp/api/spec/ipsec.h
```diff
@@ -1148,14 +1157,29 @@ typedef struct odp_ipsec_in_param_t {
...
-	/** Packet output interface for inline output operation
+	/** Packet output interface for inline outbound operation
 	 *
 	 *  Outbound inline IPSEC operation uses this packet IO interface to
 	 *  output the packet after a successful IPSEC transformation. The pktio
 	 *  must have been configured to operate in inline IPSEC mode.
 	 */
 	odp_pktio_t pktio;
 
+	/** Packet output TM queue for inline outbound operation
```
JannePeltonen, line 1168: Packets are sent out through traffic manager if the packet output mode of the pktout is ODP_PKTOUT_MODE_TM. Such pktout can be used with inline IPsec only if the platform has inline_ipsec_tm capability.

include/odp/api/spec/ipsec.h
```diff
@@ -1148,14 +1157,29 @@ typedef struct odp_ipsec_in_param_t {
...
+	 *
+	 *  If pktio is configured to use traffic manager (ODP_PKTOUT_MODE_TM),
+	 *  then application is required to provide 'tm_queue' to be used after
+	 *  successful IPSEC transformation. The pktio must have been configured to
+	 *  operate in inline IPSEC mode.
+	 *
+	 *  If application does not configure all pktio to use traffic manager, then
```
JannePeltonen, line 1175: This would not be needed since now the pktio indicates (through its output mode) wheter TM is to be used with inline IPsec or not.

include/odp/api/spec/ipsec.h
```diff
@@ -1148,14 +1157,29 @@ typedef struct odp_ipsec_in_param_t {
...
 	 *  must have been configured to operate in inline IPSEC mode.
 	 */
 	odp_pktio_t pktio;
 
+	/** Packet output TM queue for inline outbound operation
+	 *
+	 *  If pktio is configured to use traffic manager (ODP_PKTOUT_MODE_TM),
+	 *  then application is required to provide 'tm_queue' to be used after
+	 *  successful IPSEC transformation. The pktio must have been configured to
+	 *  operate in inline IPSEC mode.
```
JannePeltonen, line 1173: It could be set that this parameter is ignored if the pktio is not configured to use TM.

The comment about the need to enable inline IPsec mode in the pktio does not seem to belong here as it is a general requirement to inline IPsec.

include/odp/api/spec/ipsec.h
anoobj: Agreed. Will fix this as part of next version.

include/odp/api/spec/ipsec.h
anoobj: I'll reword to, 

"On unsupported platforms, application is not allowed to use a TM enabled pktio with inline IPSEC."

Can you confirm if this is fine?

include/odp/api/spec/ipsec.h
anoobj: Did you mean add this description after the title? If yes, I'll make this change in next version.

include/odp/api/spec/ipsec.h
anoobj: I'll reword this section to.

/** Packet output TM queue for inline outbound operation
 *
 *  Packets are send out through traffic manager if the packet output
 *  mode of the pktio is ODP_PKTOUT_MODE_TM. Such pktio interface can
 *  be used with inline IPSEC only if the platform has inline_ipsec_tm
 *  capability.
 *
 *  Application must set pktio as ODP_PKTIO_INVALID to indicate that
 *  the implementation should use tm_queue. This parameter would be
 *  ignored if the pktio is not configured for TM.
 *
 *  @see odp_ipsec_capability()
 */
odp_tm_queue_t tm_queue

Can you confirm is this looks fine?

psavol:
Please, add also a minimal linux-gen implementation which sets capa to not supported, and e.g. uses ODP_ASSERT to check if application tries to use inline IPSec with TM.

Also, some validation test case could be updated to (in minimum) refer to the new capa/param fields, so that test passing API header file has those new fields.

anoobj:
> 
> 
> Please, add also a minimal linux-gen implementation which sets capa to not supported, and e.g. uses ODP_ASSERT to check if application tries to use inline IPSec with TM.

Will submit this with next version.

> 
> Also, some validation test case could be updated to (in minimum) refer to the new capa/param fields, so that test passing API header file has those new fields.

The current validation app for inline IPsec outbound doesn't initialize TM. And in the spec, we have specified that tm_queue is ignored if pktout is not configured for TM. So do you suggest making a change to set tm_queue = ODP_TM_INVALID in the existing pktio cases?

psavol:
> The current validation app for inline IPsec outbound doesn't initialize TM. And in the spec, we have specified that tm_queue is ignored if pktout is not configured for TM. So do you suggest making a change to set tm_queue = ODP_TM_INVALID in the existing pktio cases?

Yes, at least some test case should set the field to invalid, so that it does not build if the field is missing.

anoobj:
> 
> 
> > The current validation app for inline IPsec outbound doesn't initialize TM. And in the spec, we have specified that tm_queue is ignored if pktout is not configured for TM. So do you suggest making a change to set tm_queue = ODP_TM_INVALID in the existing pktio cases?
> 
> Yes, at least some test case should set the field to invalid, so that it does not build if the field is missing.

Agreed. I have pushed the changes. I wasn't sure whether I should push as one patch or 3 patches. I've pushed as 3. Please squash if required.

anoobj:
@matiaselo Can you re trigger the CI?

matiaselo:
> @matiaselo Can you re trigger the CI?

Done

JannePeltonen:
This looks otherwise ok to me, but I think the description of the inline output parameters is still not as clear as it could. How about this:

        /** Packet output interface for inline outbound operation without TM
         *
         *  Outbound inline IPSEC operation uses this packet IO interface to
         *  output the packet after a successful IPSEC transformation. The
         *  pktio must have been configured to operate in inline IPSEC mode.
         *
         *  The pktio must not have been configured with ODP_PKTOUT_MODE_TM.
         *  For IPsec inline output to TM enabled interfaces set this field
         *  to ODP_PKTIO_INVALID and specify the TM queue to be used through
         *  the tm_queue parameter. Inline IPsec output through TM can be
         *  done only if the platform has inline_ipsec_tm capability.
         */
        odp_pktio_t pktio;

        /** TM queue for inline outbound operation
         *
         *  TM queue to be used for inline IPsec output when pktio field
         *  is ODP_PKTIO_INVALID, indicating use of TM. Otherwise ignored.
         */
        odp_tm_queue_t tm_queue;

JannePeltonen:

psavol (approved):

PR #989, opened 2020-06-29
[PATCH v1] linux-gen: add function to read string from conf

pansiot:
Needed in a specific PacketIO

Signed-off-by: Thomas Pansiot <thomas.pansiot@nokia.com>


matiaselo:
@pansiot Hi, are you planning to submit a follow-up PR for this one? We don't normally like to merge unused dead code to the repo.

PR #990, opened 2020-06-30
[PATCH v3] build: simplify remote target testing

mvl-skardach:
It is often beneficial to run "make check" on a remote target platform.
However in order to run it, a full build-system has to be present which
may not be efficient or possible for embedded platforms.
Alternatively autotools offer a mechanism which may be used:
LOG_COMPILER environment variable. If specified, all test invocations
are prepended with its contents. This allows injecting a wrapper script
which will ensure the test is run remotely (i.e. through ssh).

Prior to that the binaries have to be transported to the remote system.
This cannot be done in the script itself as it has no knowledge on the
dependencies of the test program, only its path. Test could be relying
on some files stored in srcdir or builddir.

An issue exists that if ODP is built in out-of-tree method (builddir !=
srcdir), only the compiled test programs are copied to the builddir.
scripts and script data files remain in srcdir and wrapper script is
called with an absolute path to the script located in srcdir.
This would require to copy the srcdir to exactly the same path on the
target as on build host. This may alias with some paths already there.

To prevent this from happening add make target wrappers which copy
scripts and script data to build directory which allows the autotools
test harness to produce a proper wrapper script invocation with a
relative path and allows to only copy the builddir to the target to a
strictly controlled location.

### DIFF
--- a/example/generator/Makefile.am
+++ b/example/generator/Makefile.am
@@ -11,3 +11,23 @@ TESTS = generator_null_test.sh
 TESTS_ENVIRONMENT += ODP_PLATFORM=$(with_platform)
 endif
 EXTRA_DIST = generator_null_test.sh
+
+# If building out-of-tree, make check will not copy the scripts and data to the
+# $(builddir) assuming that all commands are run locally. However this prevents
+# running tests on a remote target using LOG_COMPILER.
+# So copy all script and data files explicitly here.
+all-local:
+	if [ "x$(srcdir)" != "x$(builddir)" ]; then \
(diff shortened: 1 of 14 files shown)


matiaselo:
I guess you tried to find a method to avoid having to copy the same script patterns to each Makefile? I also tried this but didn't find a way.

A rebase is required for the PR.

mvl-skardach:
Exactly, I haven't found it but I'm open to suggestions. :)

matiaselo (approved):
I've no better suggestions, so let's go with this one.

PR #1002, opened 2020-07-29
[PATCH v3] test: packet_gen: misc fixes

matiaselo:
V2:
- Moved wait logic to rx threads (Janne)

### DIFF
--- a/test/performance/odp_packet_gen.c
+++ b/test/performance/odp_packet_gen.c
@@ -803,7 +803,7 @@ static int rx_thread(void *arg)
 	int i, thr, num;
 	uint32_t exit_test;
 	uint64_t bytes;
-	odp_time_t t1, t2;
+	odp_time_t t1, t2, exit_time;
 	odp_packet_t pkt;
 	thread_arg_t *thread_arg = arg;
 	test_global_t *global = thread_arg->global;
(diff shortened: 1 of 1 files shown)


JannePeltonen:
After this the RX packet rate statistics would be a bit off. Also, the delay is not done if the program is stopped through ctrl-c. Would this kind of logic be better:

- TX thread immediately sets the exit_test flag as it used to do.
- RX thread, after seeing the exit_test flag set, sets t2 but continues to run for still 1 second, updating t2 for every packet received. IOW, the RX stop time would be the time the last packet was received or the time the exit_test flag was seen, whichever is later.

matiaselo:
> After this the RX packet rate statistics would be a bit off. Also, the delay is not done if the program is stopped through ctrl-c. Would this kind of logic be better:
> 
> * TX thread immediately sets the exit_test flag as it used to do.
> * RX thread, after seeing the exit_test flag set, sets t2 but continues to run for still 1 second, updating t2 for every packet received. IOW, the RX stop time would be the time the last packet was received or the time the exit_test flag was seen, whichever is later.

My original thinking was that the extra delay is only needed when the application is run for a specified time (with -q arg). However, it makes sense that the app operates in the same manner no matter how it closes. Will fix in V2.

PR #1003, opened 2020-07-29
validation: pool: honor max_align capability

jerinjacobk:
Honor the odp_pool_capability_t::buf.max_align attribute while
creating the pool.

Some implementation chooses to optimize odp_buffer_t handling in
such a way that implementation-specific metadata and actual
data can be in the same cache-line as a performance optimization
while accessing the odp_buffer_t.

As a fix, except helpers, let the application uses the default
alignment from odp_pool_param_init().

### DIFF
--- a/test/validation/api/queue/queue.c
+++ b/test/validation/api/queue/queue.c
@@ -83,7 +83,6 @@ static int queue_suite_init(void)
 	odp_pool_param_init(&params);
 
 	params.buf.size  = 4;
-	params.buf.align = ODP_CACHE_LINE_SIZE;
 	params.buf.num   = MAX_NUM_EVENT;
 	params.type      = ODP_POOL_BUFFER;
 
--- a/example/time/time_global_test.c
+++ b/example/time/time_global_test.c
@@ -330,7 +330,6 @@ int main(int argc, char *argv[])
 	odp_pool_param_init(&pool_param);
 
 	pool_param.buf.size  = sizeof(timestamp_event_t);
-	pool_param.buf.align = ODP_CACHE_LINE_SIZE;
 	pool_param.buf.num   = MAX_NUM_BUF;
 	pool_param.type      = ODP_POOL_BUFFER;
 
--- a/helper/cuckootable.c
+++ b/helper/cuckootable.c
@@ -269,7 +269,8 @@ odph_cuckoo_table_create(
 	odp_pool_param_init(&param);
 	param.type = ODP_POOL_BUFFER;
 	param.buf.size = kv_entry_size;
-	param.buf.align = ODP_CACHE_LINE_SIZE;
+	if (pcapa.buf.max_align >= ODP_CACHE_LINE_SIZE)
+		param.buf.align = ODP_CACHE_LINE_SIZE;
 	param.buf.num = capacity;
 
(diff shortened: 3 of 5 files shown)


JannePeltonen:
This looks good to me but there appears to be similar problem also in these functions:
test/validation/api/queue.c:queue_suite_init
example/time/time_global_test.c:main
helper/cuckootable.c:odph_cuckoo_table_create
helper/iplookuptable.c:cache_alloc_new_pool

Perhaps you could fix those too now that you are at it?

jerinjacobk:
> Perhaps you could fix those too now that you are at it?

Sure . I will send the next version with that.

test/validation/api/queue/queue.c
```diff
@@ -55,12 +55,18 @@ static void generate_name(char *name, uint32_t index)
...
 static int queue_suite_init(void)
 {
+	odp_pool_capability_t pool_capa;
 	odp_shm_t shm;
 	test_globals_t *globals;
 	odp_pool_param_t params;
 	int num_workers;
 	odp_cpumask_t mask;
 
+	memset(&pool_capa, 0, sizeof(odp_pool_capability_t));
```
JannePeltonen, line 65: memset() before odp_pool_capability() is not needed (unless one is testing that odp_pool_capability() works, which is not the point of this test). So this could be removed (but see my other comment).

Btw, if the memset was needed, it would be better to use sizeof(pool_capa) to avoid the explicit type name that must be kept in sync with pool_capa definition.

test/validation/api/queue/queue.c
```diff
@@ -83,7 +89,8 @@ static int queue_suite_init(void)
 	odp_pool_param_init(&params);
 
 	params.buf.size  = 4;
-	params.buf.align = ODP_CACHE_LINE_SIZE;
```
JannePeltonen, line 86: Removal of this line would be the only change needed in this file. Adding code for reading pool capabilities etc seems unnecessary bloat here. There is nothing that actually relies on better buffer alignment than the default (and if there were, this patch would not handle it anyway).

I think the changes in the other files are ok even though there too better alignment than default is not really needed. The other changes are small and may improve performance (in the helpers). And in the pool validation case it is good that, at least in some cases, the alignment parameter is set.

JannePeltonen: I would leave the changes in the helpers as they are now (i.e. cache line alignment if possible, default otherwise). Maybe it helps performance.

In time_global_test.c and queue.c I would just remove the whole alignment setting and rely on the default (which is at least 8), as I do not see any point requesting a larger alignment.

In pool.c setting the alignment might slightly increase test coverage, but since there are alignment tests in buffer and packet validation tests, I would just remove the explicit alignment setting from pool.c too.

Thus, the way I would do this would be to leave the helpers as they are in the latest patch and simply remove the alignment setting line from the other files so that the parameter gets set by odp_pool_param_init().

jerinjacobk: Ack. I wil send the next version with above change

JannePeltonen (commented):
Looks good to me. Just add the reviewed-by tag.

JannePeltonen (approved):

PR #1007, opened 2020-08-10
[PATCH v3] test: packet_gen: add pktio link status check option

matiaselo:
Add new command line option '-w <sec>' for waiting up to <sec> seconds
for all pktio links to be up before starting the worker threads. Prints
also link information.

V2:
- Mbps -> Mbit/s (Janne)
- Don't reset `link_wait` for each interface (Janne)

### DIFF
--- a/test/performance/odp_packet_gen.c
+++ b/test/performance/odp_packet_gen.c
@@ -48,6 +48,7 @@ typedef struct test_options_t {
 	uint32_t ipv4_dst;
 	uint16_t udp_src;
 	uint16_t udp_dst;
+	uint32_t wait_sec;
 
 	struct vlan_hdr {
 		uint16_t tpid;
@@ -163,6 +164,8 @@ static void print_usage(void)
 	       "  -u, --update_stat <msec>  Update and print statistics every <msec> milliseconds.\n"
 	       "                            0: Don't print statistics periodically (default)\n"
 	       "  -h, --help                This help\n"
+	       "  -w, --wait <sec>          Wait up to <sec> seconds for network links to be up.\n"
+	       "                            Default: 0 (don't check link status)\n"
 	       "\n");
 }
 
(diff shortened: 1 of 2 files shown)


test/performance/odp_packet_gen.c
```diff
@@ -724,6 +732,32 @@ static int open_pktios(test_global_t *global)
...
+	       (info.duplex == ODP_PKTIO_LINK_DUPLEX_HALF ? "half" :
+	       (info.duplex == ODP_PKTIO_LINK_DUPLEX_FULL ? "full" : "unknown")));
+	printf("  media       %s\n", info.media);
+	printf("  pause_rx    %s\n",
+	       (info.pause_rx == ODP_PKTIO_LINK_PAUSE_ON ? "on" :
+	       (info.pause_rx == ODP_PKTIO_LINK_PAUSE_OFF ? "off" : "unknown")));
+	printf("  pause_tx    %s\n",
+	       (info.pause_tx == ODP_PKTIO_LINK_PAUSE_ON ? "on" :
+	       (info.pause_tx == ODP_PKTIO_LINK_PAUSE_OFF ? "off" : "unknown")));
+	printf("  speed(Mbps) %" PRIu32 "\n\n", info.speed);
```
JannePeltonen, line 757: I would suggest using the more standard Mbit/s instead of Mbps here.

test/performance/odp_packet_gen.c
```diff
@@ -741,6 +775,34 @@ static int start_pktios(test_global_t *global)
 		global->pktio[i].started = 1;
 	}
 
+	if (!test_options->wait_sec)
+		return 0;
+
+	/* Wait until all links are up */
+	for (i = 0; i < num_pktio; i++) {
+		uint32_t link_wait = 0;
```
JannePeltonen, line 783: Now each pktio may wait separately up to the max wait time instead of there being max total wait time as one would expect. I would propose moving this outside of the loop as all the pktios have already been started when entering the loop.

JannePeltonen (approved):

PR #1009, opened 2020-08-13
[PATCH v3] validation: ipsec: fix async mode test

asasidharan:
* Fix the number of iterations for which the test
  expects an ODP_IPSEC_EVENT.
* When the odp_ipsec_in_enq() or odp_ipsec_out_enq()
  was not successful, the implementation won't deliver
  an ODP_IPSEC_EVENT corresponding to the failed packet.

### DIFF
--- a/test/validation/api/ipsec/ipsec.c
+++ b/test/validation/api/ipsec/ipsec.c
@@ -566,7 +566,10 @@ static int ipsec_send_in_one(const ipsec_test_part *part,
 		CU_ASSERT_EQUAL(num_out, part->out_pkt);
 		CU_ASSERT(odp_packet_subtype(*pkto) == ODP_EVENT_PACKET_IPSEC);
 	} else if (ODP_IPSEC_OP_MODE_ASYNC == suite_context.inbound_op_mode) {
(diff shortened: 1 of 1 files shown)


JannePeltonen:
Since this is a validation test, it would be good to not trust too much that the API functions work correctly but assume that they may behave in unexpected ways.

The current async mode test in master may get stuck if the ipsec enqueue function returns zero. The proposed patch fixes that but may get stuck or may even crash if the return value is greater than one.

asasidharan:
> 
> 
> Since this is a validation test, it would be good to not trust too much that the API functions work correctly but assume that they may behave in unexpected ways.
> 
> The current async mode test in master may get stuck if the ipsec enqueue function returns zero. The proposed patch fixes that but may get stuck or may even crash if the return value is greater than one.

In that case, I can make the change such that the test would look for ODP_PACKET_EVENT_IPSEC type event only if the num_out value returned by the IPsec enqueue API is 1.
This way we can make sure that the test function won't look for the IPsec event if an unexpected value is returned.

JannePeltonen (approved):

PR #1010, opened 2020-08-18
[PATCH v5] api: cls: introduce mark support

jerinjacobk:
## api: cls: introduce mark support

In order to accelerate the lookup function, typical classification
HW gives a feature to embedded a custom application-defined value as
metadata when a specific flow matches the PMR rule.

Introduce a new feature called mark to enable this use-case where
the application can create a PMR with mark and configure the application
defined values as struct odp_pmr_create_opt_t::mark for the given
PMR, and then application retrieve the value through odp_packet_cls_mark()
in fast-path to achieve the first level HW accelerated lookup.

Since the existing odp_cls_pmr_create() API does not support
structure version to accept new arguments, introduced
odp_cls_pmr_create_opt() API provide additional options for
creating the PMR which include the mark support.

## linux-gen: cls: dummy implementation of mark support

Added a dummy implementation for odp_packet_mark()
and odp_cls_pmr_create_opt() APIs.

### DIFF
--- a/include/odp/api/spec/classification.h
+++ b/include/odp/api/spec/classification.h
@@ -204,6 +204,9 @@ typedef struct odp_cls_capability_t {
 
 	/** Supported threshold type for BP */
 	odp_threshold_types_t threshold_bp;
+
+	/** Maximum value of odp_pmr_create_opt_t::mark */
+	uint64_t max_mark;
 } odp_cls_capability_t;
 
 /**
@@ -606,6 +609,24 @@ typedef struct odp_pmr_param_t {
 } odp_pmr_param_t;
 
 /**
+ * Packet Matching Rule creation options
+ */
+typedef struct odp_pmr_create_opt_t {
+	/** Array of odp_pmr_param_t entries, one entry per term desired.
+	 * Use odp_cls_pmr_param_init() to initialize parameters into their default values.
+	 */
+	odp_pmr_param_t *terms;
+	/** Number of terms in the match rule. */
+	int num_terms;
+	/** Value to be set in the CLS mark of a packet when the packet matches this
+	 * Packet Matching Rule.
+	 * The default value is zero.
+	 * The maximum value is indicated in odp_cls_capability_t::max_mark capability.
+	 */
+	uint64_t mark;
+} odp_pmr_create_opt_t;
+
+/**
  * Initialize packet matching rule parameters
  *
  * Initialize an odp_pmr_param_t to its default values for all fields
@@ -613,6 +634,15 @@ typedef struct odp_pmr_param_t {
  * @param param        Address of the odp_pmr_param_t to be initialized
  */
 void odp_cls_pmr_param_init(odp_pmr_param_t *param);
+
+/**
+ * Initialize packet matching rule creation option
+ *
+ * Initialize an odp_pmr_create_opt_t to its default values for all fields
+ *
+ * @param opt       Address of the odp_pmr_create_opt_t to be initialized
+ */
+void odp_cls_pmr_create_opt_init(odp_pmr_create_opt_t *opt);
 
 /**
  * Create a packet matching rule
@@ -640,10 +670,36 @@ void odp_cls_pmr_param_init(odp_pmr_param_t *param);
  *
  * @return Handle to the Packet Match Rule.
  * @retval ODP_PMR_INVALID on failure
+ *
+ * @note Matching PMR rule created through this function sets the CLS mark metadata
+ * of the packet to zero.
+ *
+ * @note Rules created through this function are equivalent to rules created through
+ * odp_cls_pmr_create_opt() with the same PMR terms and with the additional option
+ * fields set to their default values.
  */
 odp_pmr_t odp_cls_pmr_create(const odp_pmr_param_t *terms, int num_terms,
 			     odp_cos_t src_cos, odp_cos_t dst_cos);
 
+/**
+ * Create a packet matching rule with options
+ *
+ * Similar to odp_cls_pmr_create() function with additional PMR creation
+ * options specified through odp_pmr_create_opt_t.
+ *
+ * Use odp_cls_pmr_create_opt_init() to initialize options into their default
+ * values.
+ *
+ * @param opt	       points to PMR create options
+ * @param src_cos      source CoS handle
+ * @param dst_cos      destination CoS handle
+ *
+ * @return Handle to the Packet Match Rule.
+ * @retval ODP_PMR_INVALID on failure
+ *
+ */
+odp_pmr_t odp_cls_pmr_create_opt(const odp_pmr_create_opt_t *opt,
+				 odp_cos_t src_cos, odp_cos_t dst_cos);
 /**
  * Function to destroy a packet match rule
  * Destroying a PMR removes the link between the source and destination
--- a/include/odp/api/spec/packet.h
+++ b/include/odp/api/spec/packet.h
@@ -1990,6 +1990,20 @@ int8_t odp_packet_shaper_len_adjust(odp_packet_t pkt);
  */
 void odp_packet_shaper_len_adjust_set(odp_packet_t pkt, int8_t adj);
 
+/**
+ * Get CLS mark value from the odp_packet_t
+ *
+ * Get the value of the CLS mark of the packet.
+ * The mark value is zero by default but may be changed by the classification subsystem.
+ *
+ * @param pkt Packet handle
+ * @return mark value
+ * @see odp_cls_capability_t::max_mark
+ * @see odp_pmr_create_opt_t::mark
+ * @see odp_cls_pmr_create_opt()
+ */
+uint64_t odp_packet_cls_mark(odp_packet_t pkt);
+
 /*
  *
  * Debugging
--- a/platform/linux-generic/odp_classification.c
+++ b/platform/linux-generic/odp_classification.c
@@ -180,7 +180,15 @@ int odp_cls_capability(odp_cls_capability_t *capability)
 	capability->threshold_red.all_bits = 0;
 	capability->threshold_bp.all_bits = 0;
 	capability->max_hash_queues = CLS_COS_QUEUE_MAX;
+	capability->max_mark = 0;
 	return 0;
+}
+
+void odp_cls_pmr_create_opt_init(odp_pmr_create_opt_t *opt)
+{
+	opt->terms = NULL;
+	opt->num_terms = 0;
+	opt->mark = 0;
 }
 
 static void _odp_cls_update_hash_proto(cos_t *cos,
@@ -736,6 +744,19 @@ odp_pmr_t odp_cls_pmr_create(const odp_pmr_param_t *terms, int num_terms,
 
 	UNLOCK(&pmr->s.lock);
 	return id;
+}
+
+odp_pmr_t odp_cls_pmr_create_opt(const odp_pmr_create_opt_t *opt,
+				 odp_cos_t src_cos, odp_cos_t dst_cos)
+{
+	if (opt == NULL)
+		return ODP_PMR_INVALID;
+
+	/* Current implementation does not support mark > 0*/
+	if (opt->mark != 0)
+		return ODP_PMR_INVALID;
+
+	return odp_cls_pmr_create(opt->terms, opt->num_terms, src_cos, dst_cos);
 }
 
 int odp_cls_cos_pool_set(odp_cos_t cos_id, odp_pool_t pool)
--- a/platform/linux-generic/odp_packet.c
+++ b/platform/linux-generic/odp_packet.c
@@ -2843,3 +2843,10 @@ odp_proto_l4_type_t odp_packet_l4_type(odp_packet_t pkt)
 
 	return ODP_PROTO_L4_TYPE_NONE;
 }
+
+uint64_t odp_packet_cls_mark(odp_packet_t pkt)
+{
+	(void)pkt;
+
+	return 0;
+}


JannePeltonen:
I have a generic comment that the "mark" and "id" in mark_id sound quite redundant to me. I would suggest cos_mark or just mark instead,

include/odp/api/spec/classification.h
```diff
@@ -55,6 +55,13 @@ extern "C" {
  * @deprecated Use ODP_PMR_INVALID instead
  */
 
+/**
+ * @def ODP_MARK_ID_INVALID
```
JannePeltonen, line 59: Since CoS mark is not an opaque type, I do not think the definition of ODP_MARK_ID_INVALID can be left to the implementation or ABI. An application, even if not interested in the actual value of ODP_MARK_ID_INVALID, needs to know which mark values it _can_ use without clashing with ODP_MARK_ID_INVALID. It gets quite inconvenient if the value can be anything (like 7 in one implementation, 42 in another).

include/odp/api/spec/classification.h
```diff
@@ -164,6 +171,21 @@ typedef struct odp_bp_param_t {
 	odp_threshold_t threshold;
 } odp_bp_param_t;
 
+/** Mark support
+ * When mark support is enabled for a particular flow, the implementation shall
```
JannePeltonen, line 175: > When mark support is enabled for a particular flow,

That may be slightly confusing since there is no separate enable control but just the desired mark value. I would also change "insert" to "set" because there cannot be more than one CoS mark set in packet metadata.

The default value set by odp_cls_cos_param_init() should be defined.

I would rephrase this somehow like this:

CoS mark support

When a packet is processed by a class-of-service, CoS mark in packet metadata is set to the value configured for the class-of-service through odp_cls_cos_param_t.

include/odp/api/spec/classification.h
```diff
@@ -204,6 +226,12 @@ typedef struct odp_cls_capability_t {
 
 	/** Supported threshold type for BP */
 	odp_threshold_types_t threshold_bp;
+
+	/** Maximum number of mark id value.
+	 * Value zero represents the mark based CoS is not supported.
```
JannePeltonen, line 231: Something like this would sound better to me:
"Zero value means that CoS marking is not supported"

include/odp/api/spec/packet.h
```diff
@@ -1979,6 +1979,23 @@ int8_t odp_packet_shaper_len_adjust(odp_packet_t pkt);
  */
 void odp_packet_shaper_len_adjust_set(odp_packet_t pkt, int8_t adj);
 
+/**
+ * Get mark id value from the odp_packet_t
+ *
+ * When a flow configured with mark CoS, Upon hitting the mark CoS,
```
JannePeltonen, line 1985: Maybe we should not discuss the classifier details too much in here as from packet point of view this just somewhat arbitrary metadata that can perhaps be used in more ways later (like allowing the application set the value directly).

I would reduce the explanation to something shorter, like this, but add a comment about the initial value in case CoS is not even involved (e.g. someone just allocs a packet and later reads the CoS mark):

"CoS class can be set by the classifier to an application defined value. The default value is ODP_COS_MARK_INVALID."

platform/linux-generic/odp_packet.c
```diff
@@ -2843,3 +2843,10 @@ odp_proto_l4_type_t odp_packet_l4_type(odp_packet_t pkt)
 
 	return ODP_PROTO_L4_TYPE_NONE;
 }
+
+uint64_t odp_packet_mark_id(odp_packet_t pkt)
+{
+	(void)pkt;
```
JannePeltonen, line 2849: ODP seems to typically use ODP_UNUSED attribute in the parameter list instead of this.

psavol: I think both ways are used, and I'm OK with either way. Personally, I prefer cast to void as it's standard C language (attribute supports is compiler specific).

psavol:
General comment on API patch sets. Generally, those should contain three types of patches:
1) Edit odp/api/spec header files only (or edit as few other files as possible)
2) Add implementation (ABI, linux-gen files)
3) Add validation tests. In minimum, use the new type or field, dummy function call of a new function.

jerinjacobk:
> I have a generic comment that the "mark" and "id" in mark_id sound quite redundant to me. I would suggest cos_mark or just mark instead,

I will stick with mark then.

jerinjacobk:
> I have a generic comment that the "mark" and "id" in mark_id sound quite redundant to me. I would suggest cos_mark or just mark instead,

include/odp/api/spec/classification.h
jerinjacobk: OK. How about setting ODP_MARK_ID_INVALID as zero and change the comments accordingly.

include/odp/api/spec/classification.h
jerinjacobk: > > When mark support is enabled for a particular flow,
> 
> That may be slightly confusing since there is no separate enable control but just the desired mark value. I would also change "insert" to "set" because there cannot be more than one CoS mark set in packet metadata.

Yes. "set" makes sense.

> 
> The default value set by odp_cls_cos_param_init() should be defined.

Ack.

> 
> I would rephrase this somehow like this:
> 
> CoS mark support
> 
> When a packet is processed by a class-of-service, CoS mark in packet metadata is set to the value configured for the class-of-service through odp_cls_cos_param_t.

After implementing this patch on one of the Marvell platform, I realized that, Hooking Mark in CoS would waste the HW resources like RQ in the case where "single" CoS preferred for application and just because of  CoS has match value, application needs to create another CoS and it have bearing on HW resources that way it is mapped to Implementation. Look like the ideal mapping is NOT CoS for mark, Instead have new variant of odp_cls_pmr_create() as odp_cls_pmr_create_with_mark().

/**
 * Create a packet matching rule with mark id
 *
 * It is the mark feature variant of odp_cls_pmr_create().
 * When a PMR is created with mark support, the implementation shall insert
 * the specified mark_id in the packet metadata. If the packet matches the PMR
 * then the application can retrieve it through odp_packet_mark_id().
 *
 * @param terms        Array of odp_pmr_param_t entries, one entry per term
 *                     desired.
 * @param num_terms    Number of terms in the match rule.
 * @param mark_id      mark id to insert in the packet metadata on match.
 * @param src_cos      source CoS handle
 * @param dst_cos      destination CoS handle
 *
 * @return Handle to the Packet Match Rule.
 * @retval ODP_PMR_INVALID on failure
 * @note The mark id value is arbitrary and application-defined.
 * The maximum allowed value depends on the underlying implementation.
 * @see odp_cls_capability_t::max_mark_id for maximum mark id value
 * @see odp_packet_mark_id() to get the mark id for the packet.
 */
odp_pmr_t odp_cls_pmr_create_with_mark(const odp_pmr_param_t *terms, int num_terms,
                                       uint64_t mark_id, odp_cos_t src_cos, odp_cos_t dst_cos);

include/odp/api/spec/classification.h
jerinjacobk: Ack

include/odp/api/spec/packet.h
jerinjacobk: I will wait for your comments on new API proposal as odp_cls_pmr_create_with_mark() to reword the same.

platform/linux-generic/odp_packet.c
jerinjacobk: OK. That case, I prefer to keep (void).

jerinjacobk:
> General comment on API patch sets. Generally, those should contain three types of patches:
> 
> 1. Edit odp/api/spec header files only (or edit as few other files as possible)

OK

> 2. Add implementation (ABI, linux-gen files)

OK. I assume a dummy implementation here. So that linux-generic platform owners can implement the required implement as 
and when needed. 

> 3. Add validation tests. In minimum, use the new type or field, dummy function call of a new function.

I thought to update the validation test, But in that case, validation test would fail. Right? and CI will fail too. Is OK to have validation test case and due to lack of implementation in linux-generic fails?

include/odp/api/spec/classification.h
JannePeltonen: Yes, defining it as zero in the API itself would solve this.

include/odp/api/spec/classification.h
JannePeltonen: I am not yet sure, but I think I am in favor of putting mark somehow in the PMR (but with a cleaner and more generic API that the odp_cls_pmr_create_with_mark()).

Initially I thought that you could just map the abstract API level PMR + CoS configuration to somewhat different HW level "PMR" + "CoS" configuration. For instance, if an application would like to set mark based on IP address and select destination queue based on DSCP, the application would just create all the different CoS combinations, but the implementation could notice when a new CoS differs from an existing one only in mark and then reuse the existing CoS and do the marking as part of the different packet patching rules that now point to the shared "CoS" in the HW.

This way PMR rules would remain as pure matchers and all actions would be defined in the CoS objects.

But perhaps that is not the right model. Both Linux netfilter and DPDK flow API allow mark (and other) actions as part of the matching rule. We could do the same in ODP, but that would be a significant conceptual change that needs to be discussed with Petri too.

Btw, there is an additional problem in putting mark in CoS and having application create all the combinations even if the implementation could optimize the redundancy away: When using multiple queues with hashing in a CoS, the implementation would have to create a separate set of queues for each QoS. Thus an application can create multiple CoS which differ only in mark only if it can specify the exact same queue(s) for the CoS and that is now possible only with single queue per CoS.

psavol: I discussed with Janne about this, and it seems that adding mark to pmr is the best solution. Unfortunately, pmr create does not have struct that we could modify, so a new create function seems the only option. For future proof design, it would be good to introduce a structure this time. So, we'd suggest something like:

typedef struct odp_pmr_create_opt_t {
    const odp_pmr_param_t *terms;
    int num_terms;
    uint64_t mark_id;
    odp_cos_t src_cos;
    odp_cos_t dst_cos;
} odp_pmr_create_opt_t;

odp_pmr_t odp_cls_pmr_create_opt(const odp_pmr_create_opt_t *opt);

src_cos and dst_cos could be also direct function parameters. That would keep current look/feel of the function, where the call connects src_cos to dst_cos. 

odp_pmr_t odp_cls_pmr_create_opt(const odp_pmr_create_opt_t *opt, odp_cos_t src_cos, odp_cos_t dst_cos);

Also capability is needed for the maximum mark size (0 ... max_mark_id).

jerinjacobk: Ack. How about odp_pmr_create_meta_t instead of odp_pmr_create_opt_t and odp_cls_pmr_create_with_meta() to denote this variant of function would be useful to create the PMR with metadata(the content which not available in the packet unlike normal terms like match id and future some other extension like IPsec second pass metadata from HW accelerators etc). Thoughts on the name?

jerinjacobk: OR
typedef struct odp_pmr_apply_meta_t {
             uint64_t mark_id;
} odp_pmr_apply_meta_t;
odp_cls_pmr_apply_meta(odp_pmr_t pmr, odp_pmr_apply_meta_t meta);
This would apply meta based patterns to exisitng odp_pmr_t  the one created with odp_cls_pmr_create().

JannePeltonen: I would prefer opt over meta as it is maybe more generic. Maybe there will later be need to add something that is not metadata specific at all. We even thought about odp_cls_pmr_create2().

Postponing the configuration from PMR creation to a later attach/apply step is problematic because it would require the implementation be able to modify an existing PMR on the fly when it is already being used for packet matching. The implementation cannot postpone the creation until the attach/apply phase because it does not know if it ever comes.

jerinjacobk: OK. Makes sense. In order to add comments, I would like to know the full form of "opt" , Is it "option"?
Is there any name conversion on the table for the new variant? Is extension -> odp_cls_pmr_create_extn(), alternative -> odp_cls_pmr_create_alt(), varint -> odp_cls_pmr_create_varint() can be used instead of option. No strong option on the name, Waiting for the feedback on the name for v2.

JannePeltonen: I think Petri should comment.

I am not a big fan of "opt", but there are already various _opt_t types in ODP for miscellaneous config options so it seems to me that opt would be in line with those. "alt", "variant", "ext" and such would be totally new. That said, I do not have a strong opinion on naming.

jerinjacobk: I will stick with odp_cls_pmr_create_opt() then.

jerinjacobk: Sent the  V2 with _opt scheme. Resolving this one.

include/odp/api/spec/classification.h
```diff
@@ -55,6 +55,14 @@ extern "C" {
  * @deprecated Use ODP_PMR_INVALID instead
  */
 
+/**
+ * Invalid mark value.
+ * This value returned from odp_packet_mark() when the mark
+ * is not valid for the packet or mark support is not available.
+ * @see odp_cls_capability_t::max_mark
+ */
+#define ODP_CLS_MARK_INVALID 0
```
psavol, line 64: Usually, invalid values are needed with handles. I think we don't need explicit invalid value for the mark, but just define that the default value for mark is 0.

include/odp/api/spec/classification.h
```diff
@@ -204,6 +212,12 @@ typedef struct odp_cls_capability_t {
 
 	/** Supported threshold type for BP */
 	odp_threshold_types_t threshold_bp;
+
+	/** Maximum number of mark value.
+	 * ODP_CLS_MARK_INVALID means that marking is not supported.
+	 * @see odp_pmr_create_opt_t::mark
+	 */
+	uint64_t max_mark;
```
psavol, line 220: If maximum is 0, then packet mark value remain always in it's default value (==0). So, marking is kind of always  supported, when max is 0 it's the only supported value.

include/odp/api/spec/classification.h
```diff
@@ -605,6 +619,28 @@ typedef struct odp_pmr_param_t {
...
+	/** When a PMR is created with mark support, the implementation shall set
+	 * the specified mark in the packet metadata if PMR matches. In the event
+	 * of not matching the PMR, implementation set the ODP_COS_MARK_INVALID in
+	 * the packet metadata. The application can retrieve mark value set by
+	 * the implementation through odp_packet_mark().
+	 * The default value is ODP_COS_MARK_INVALID which means marking is
+	 * disabled. Application can set any value from ODP_COS_MARK_INVALID(0) to
+	 * odp_cls_capability_t::max_mark
+	 */
+	uint64_t mark;
```
psavol, line 641: This also get shorter if invalid and enable/disable specification is removed.

- default / initial value is zero
- this value replaces current mark value when rule matches
- user can use values 0 ... max
- what happens when rule does not match? Current value is overwritten with 0, or not changed

include/odp/api/spec/classification.h
```diff
@@ -644,6 +689,24 @@ void odp_cls_pmr_param_init(odp_pmr_param_t *param);
...
+ *
+ * @param opt	       points to PMR create options
+ * @param src_cos      source CoS handle
+ * @param dst_cos      destination CoS handle
+ *
+ * @return Handle to the Packet Match Rule.
+ * @retval ODP_PMR_INVALID on failure
+ */
+odp_pmr_t odp_cls_pmr_create_opt(const odp_pmr_create_opt_t *opt,
+				 odp_cos_t src_cos, odp_cos_t dst_cos);
```
psavol, line 709: Need to document if old pmr_create and new pmr_create_opt calls can be mixed. I think those could be mixed, just need to document what happens to mark when PMR created with the old function hits/misses. Is the current  mark value set to zero, or not changed.

include/odp/api/spec/packet.h
```diff
@@ -1990,6 +1990,23 @@ int8_t odp_packet_shaper_len_adjust(odp_packet_t pkt);
...
+ * When mark is not supported or configured mark is not matched
+ * the PMR then this function shall return ODP_CLS_MARK_INVALID.
+ *
+ * @param pkt Packet handle
+ * @return mark value
+ * @see odp_cls_capability_t::max_mark
+ * @see odp_cls_pmr_create_opt()
+ * @see ODP_CLS_MARK_INVALID
+ */
+uint64_t odp_packet_mark(odp_packet_t pkt);
```
psavol, line 2008: This could be called also odp_packet_cls_mark(). That would highlight that the mark is set by CLS (user defined value through CLS API). On the other hand, if some other ODP API could change the (same) mark, this more generic name would fit better.

include/odp/api/spec/classification.h
JannePeltonen: If a rule does not match, then I think the mark that is already in the packet should not change in any way .

The open issue that we have about the semantics and that Petri refers to above an in other comments is whether we have
1) Only one kind of PMR rules, that always set the mark upon rule match. The mark value of the last matching rule is the one that the application will see. Rules created with the old create command are equivalent to rules created with the create_opt() command with mark value zero.
2) Two kind of PMT rules: rules that set the mark and rules that do not. Rules created with the old command or with mark value zero do not change the mark of the packet. The mark that the application sees is the mark value of that last matching rule that had marking enabled, or the default mark value if no such rule matched. Rules created with the old command are equivalent to rules created with the new command with mark value 0.

The two alternatives are equivalent if there is only one layer of rules, but not if there is a multi level hierarchy of rules.

Which one do we want? If we want the latter, then the mark-setting-enable config could be a separate flag in the opt struct instead of having a special mark value designate it. And if so, then alternative 2 can be seen as a superset of alternative 1 and can be added later in a backward compatible manner.

include/odp/api/spec/classification.h
```diff
@@ -204,6 +212,12 @@ typedef struct odp_cls_capability_t {
 
 	/** Supported threshold type for BP */
 	odp_threshold_types_t threshold_bp;
+
+	/** Maximum number of mark value.
```
JannePeltonen, line 216: Should be: Maximum mark value.

include/odp/api/spec/packet.h
```diff
@@ -1990,6 +1990,23 @@ int8_t odp_packet_shaper_len_adjust(odp_packet_t pkt);
  */
 void odp_packet_shaper_len_adjust_set(odp_packet_t pkt, int8_t adj);
 
+/**
+ * Get mark value from the odp_packet_t
+ *
+ * When a flow is configured with mark PMR, Upon matching the rule,
```
JannePeltonen, line 1996: As I commented earlier, details on how the classifier sets the mark is a bit out-of-place here. From packet point-of-view the mark is just metadata and it should suffice to just briefly mention here that classifier can set it.

The default value must be specified here as one can call this function for any packet, including those that never went through the classifier.

include/odp/api/spec/classification.h
jerinjacobk: I will change to ODP_CLS_MARK_DEFAULT in next version.

include/odp/api/spec/classification.h
jerinjacobk: Makes sense. I will reword the "supported" aspect in the next version.

include/odp/api/spec/classification.h
jerinjacobk: We can stick with option 1. I will send the version based on that.

include/odp/api/spec/classification.h
jerinjacobk: Yes. We can document the old and new API can be mixed. If we use old API then odp_packet_mark() shall return ODP_CLS_MARK_DEFAULT;

include/odp/api/spec/packet.h
jerinjacobk: It may possible that from another subsystem like ipsec can produce mark, So I will go with your first suggestion i.e rename to odp_packet_cls_mark()

include/odp/api/spec/classification.h
jerinjacobk: Will be fixed in next version.

include/odp/api/spec/packet.h
jerinjacobk: OK. I will reword it next version. Please check the next version, if it is not suffice, You could share the exact text to replace.

include/odp/api/spec/classification.h
jerinjacobk: Updated to following in the next version.
        /** The mark value to retrieve it through odp_packet_cls_mark() on PMR match.
         * On PMR miss, odp_packet_cls_mark() shall return ODP_COS_MARK_DEFAULT.
         * The default mark value is ODP_COS_MARK_DEFAULT.
         * 0(ODP_COS_MARK_DEFAULT) to odp_cls_capability_t::max_mark is the valid
         * value range for mark.
         */

include/odp/api/spec/classification.h
jerinjacobk: Added the following note in the next version.
 * @note Application can mix the usage of odp_cls_pmr_create() and
 * odp_cls_pmr_create_opt(), when mixed and the PMR is created with
 * odp_cls_pmr_create() then it will be similar to PMR creation using
 * odp_cls_pmr_create_opt() with default options.

include/odp/api/spec/classification.h
```diff
@@ -55,6 +55,11 @@ extern "C" {
  * @deprecated Use ODP_PMR_INVALID instead
  */
 
+/**
+ * Default classification mark value.
+ */
+#define ODP_CLS_MARK_DEFAULT 0
```
JannePeltonen, line 61: Do we really need this? I thought we would just say that the default value is zero and that is it.

include/odp/api/spec/classification.h
```diff
@@ -204,6 +209,11 @@ typedef struct odp_cls_capability_t {
 
 	/** Supported threshold type for BP */
 	odp_threshold_types_t threshold_bp;
+
+	/** Maximum mark value.
+	 * odp_pmr_create_opt_t::mark value must be less than this value.
```
JannePeltonen, line 214: So now max_mark would not be maximum mark value (but maximum mark value plus one). That is unclear naming and there is no good reason for the "less than" requirement.

I would rephrase: "maximum value of odp_pmr_create_opt_t::mark".

include/odp/api/spec/classification.h
```diff
@@ -605,6 +615,25 @@ typedef struct odp_pmr_param_t {
...
+ * Packet Matching Rule creation options
+ */
+typedef struct odp_pmr_create_opt_t {
+	/** Array of odp_pmr_param_t entries, one entry per term desired.
+	 * Use odp_cls_pmr_param_init() to initialize parameters into their default values.
+	 */
+	odp_pmr_param_t *terms;
+	/** Number of terms in the match rule. */
+	int num_terms;
+	/** The mark value to retrieve it through odp_packet_cls_mark() on PMR match.
```
JannePeltonen, line 628: The grammar needs to be fixed. In addition, it is IMHO inaccurate to say that mark will be zero (or default) on PMR miss when in reality the mark in that case is not changed but may already have a non-default value as set by some other PMR.

I would rephrase the whole thing with something like this:
"Value to be set in the CLS mark of a packet when the packet matches this Packet Matching Rule. The default value is zero. The maximum value is indicated in 'max_mark' capability."

include/odp/api/spec/classification.h
```diff
@@ -644,6 +682,29 @@ void odp_cls_pmr_param_init(odp_pmr_param_t *param);
...
+ * values.
+ *
+ * @param opt	       points to PMR create options
+ * @param src_cos      source CoS handle
+ * @param dst_cos      destination CoS handle
+ *
+ * @return Handle to the Packet Match Rule.
+ * @retval ODP_PMR_INVALID on failure
+ *
+ * @note Application can mix the usage of odp_cls_pmr_create() and
```
JannePeltonen, line 701: I would actually add this in the documentation of odp_cls_pmr_create():

"Matching PMR rule created through this function sets the CLS mark metadata of the packet to zero.

Rules created through this function are equivalent to rules created through odp_cls_pmr_create() with the same PMR terms and with the additional option fields set to their default values."

After that, I am not sure if this note is needed anymore. If it is, the grammar needs to be fixed. I would simply say: "Rules can be created through both odp_cls_pmr_create() odp_cls_pmr_create_opt() in the same application."

include/odp/api/spec/packet.h
```diff
@@ -1990,6 +1990,22 @@ int8_t odp_packet_shaper_len_adjust(odp_packet_t pkt);
  */
 void odp_packet_shaper_len_adjust_set(odp_packet_t pkt, int8_t adj);
 
+/**
+ * Get mark value from the odp_packet_t
+ *
+ * Retrieve the mark value set by classification subsystem.
```
JannePeltonen, line 1996: As I commented earlier, the logic of packet marking should be explained in the classification API and here we only need to tell that this metadata exists, has certain default value and can be retrieved through this function. Thus, I would rephrase like this:

"Get the value of the CLS mark of the packet. The mark value is zero by default but may be changed by the classification subsystem."

platform/linux-generic/odp_classification.c
```diff
@@ -180,9 +180,17 @@ int odp_cls_capability(odp_cls_capability_t *capability)
...
 	capability->max_hash_queues = CLS_COS_QUEUE_MAX;
+	capability->max_mark = ODP_CLS_MARK_DEFAULT;
 	return 0;
 }
 
+void odp_cls_pmr_create_opt_init(odp_pmr_create_opt_t *opt)
+{
+	opt->terms = NULL;
+	opt->num_terms = 0;
+	opt->mark = ODP_CLS_MARK_DEFAULT;
```
JannePeltonen, line 191: This would actually be an illegal value with the proposed definition of max_mark. But see my comment about changing max_mark to actually mean max mark value.

platform/linux-generic/odp_classification.c
```diff
@@ -738,6 +746,18 @@ odp_pmr_t odp_cls_pmr_create(const odp_pmr_param_t *terms, int num_terms,
...
 }
 
+odp_pmr_t odp_cls_pmr_create_opt(const odp_pmr_create_opt_t *opt,
+				 odp_cos_t src_cos, odp_cos_t dst_cos)
+{
+	(void)opt;
+	(void)src_cos;
+	(void)dst_cos;
+
+	ODP_UNIMPLEMENTED();
```
JannePeltonen, line 756: This could just call odp_cls_pmr_create() and then the API implementation would be complete (although a bit limited in the allowed mark range).

JannePeltonen:
I think we are getting close to conclusion. I made a bunch of comments but they were all about minor issues mostly related to the API documentation.

include/odp/api/spec/classification.h
jerinjacobk: OK. I will remove this.

include/odp/api/spec/classification.h
jerinjacobk: Make sense. I will fix it in next version.

include/odp/api/spec/classification.h
jerinjacobk: Changed to :
        /** Value to be set in the CLS mark of a packet when the packet matches this
         * Packet Matching Rule.
         * The default value is zero.
         * The maximum value is indicated in odp_cls_capability_t::max_mark capability.
         */

include/odp/api/spec/classification.h
jerinjacobk: makes sense. Update odp_cls_pmr_create()'s note to.
+ *
+ * @note Matching PMR rule created through this function sets the CLS mark metadata
+ * of the packet to zero.
+ *
+ * @note Rules created through this function are equivalent to rules created through
+ * odp_cls_pmr_create() with the same PMR terms and with the additional option fields
+ * set to their default values.

include/odp/api/spec/packet.h
jerinjacobk: Will be fixed in next version.

platform/linux-generic/odp_classification.c
jerinjacobk: The default value will be 0 right? which is the one we need to set in odp_cls_pmr_create_opt_init(), Not sure what is illegal here? Please comment.

platform/linux-generic/odp_classification.c
jerinjacobk: Are we on same page wrt implemetion?
`odp_pmr_t odp_cls_pmr_create_opt(const odp_pmr_create_opt_t *opt,
                                 odp_cos_t src_cos, odp_cos_t dst_cos)
{

        if (opt == NULL)
                return ODP_PMR_INVALID;

        if (opt->mark != 0) {
                ODP_UNIMPLEMENTED();
                return ODP_PMR_INVALID;
        }

        return odp_cls_pmr_create(opt->terms, opt->num_terms, src_cos, dst_cos);
}`

platform/linux-generic/odp_classification.c
JannePeltonen: If max_mark capability gets fixed so that it defines the maximum value that is still allowed, then this is ok. But if max_mark definition was not fixed, then it would require that the mark value in the rule parameters was at most _one less_ than zero (the default). Thus in that case this init function would set the mark value to something that one would not be allowed to pass to PMR creation.

So this comment was just to show that the current definition of max_mark in the capability struct just causes problems. But when that gets improved, this init function is ok.

platform/linux-generic/odp_classification.c
JannePeltonen: There is no need to check that opt->mark is zero because it is required to be due to the max_mark capability value being zero. If you leave the check (I would just remove it), then I do not think ODP_UNIMPLEMENTED() is the right thing to do (since the function is implemented and does work). It is just the case that too large a mark value was passed, so returning ODP_PMR_INVALID would seem like the right thing to do.

platform/linux-generic/odp_classification.c
jerinjacobk: If I understand it correctly init will be filled with the default value(which is valid for all cases). So that application will first doodp_cls_pmr_create_opt_init() followed by if needs to change the mark value.

If we don't set the value as 0 then it can be any random value. Right? We need to avoid that. Right?

platform/linux-generic/odp_classification.c
jerinjacobk: I thought of checking opt->mark against zero to catch rouge application which is not honoring the max_mark value.
In the current implementation the max_mark = 0, But In the future when we add support for the real mark, It needs to be compared against odp_cls_capability_t::max_mark. Is it adding up? Agree on ODP_UNIMPLEMENTED(), I can remove that

platform/linux-generic/odp_classification.c
JannePeltonen: Yes, all the fields need to be set by init() and zeroing the mark field is the right thing to do.

I meant that the code in this version of this PR sets max_mark capability to zero at the same time when the API says: "odp_pmr_create_opt_t::mark value must be less than this value.". Since zero is not less than zero, it is not ok to pass the options initialized by the init function to odp_cls_pmr_create() (I suppose you would not be able to call the function at all).

But this gets fixed when you fix the definition of max_mark so that it is the maximum mark value that is allowed. Then zero becomes an allowed value (and the only one) in this minimal implementation.

platform/linux-generic/odp_classification.c
JannePeltonen: Yes, that is fine.

platform/linux-generic/odp_classification.c
jerinjacobk: Ack. I rephrased to "maximum value of odp_pmr_create_opt_t::mark". So code and doc will be aligned in the next version.

include/odp/api/spec/classification.h
```diff
@@ -640,10 +670,36 @@ void odp_cls_pmr_param_init(odp_pmr_param_t *param);
  *
  * @return Handle to the Packet Match Rule.
  * @retval ODP_PMR_INVALID on failure
+ *
+ * @note Matching PMR rule created through this function sets the CLS mark metadata
+ * of the packet to zero.
+ *
+ * @note Rules created through this function are equivalent to rules created through
+ * odp_cls_pmr_create() with the same PMR terms and with the additional option fields
```
JannePeltonen, line 678: Should be odp_cls_pmr_create_opt().

I made the mistake in my comment and it ended up here. Sorry about that.

JannePeltonen:
There was one typo in the last version, but otherwise this looks ok to me. You can add my reviewed-by tag when convenient. I think we need to get Petri's review too.

include/odp/api/spec/classification.h
jerinjacobk: Will be fixed in next version.

JannePeltonen (approved):

psavol:
CLS mark API was merged in 1.24.2

PR #1012, opened 2020-08-20
[PATCH v1] api: ipsec: add parse_level IPsec sa param

vvelumuri:
Currently ODP provides selection of header parsing level as a global ipsec
configuration. Some hardawares can parse the layer headers after ipsec
transformation.

This set of changes provides the configuration of parse level to each
IPsec inbound SA. It overrides the parse level of pktio with this value
for the packets specific to this SA.


vvelumuri:
Closed the previous pull request: #969 and created this new one.

PR #1013, opened 2020-08-21
[PATCH v3] api: ipsec: new apis for IPsec stats

vvelumuri:
Present ODP spec does not have a provision to get the stats of IPsec.
Adding new APIs to get and clear the IPsec stats.

These APIs can get and clear the IPsec stats for multiple SAs.
Multiple SA support is provided to reduce the invocation of these APIs
in datapath, when there are huge number of SAs.

Application need to allocate the memory for stats based on number of
SAs it queries for. API will populate the memory with stats of each
specified SA.

### DIFF
--- a/include/odp/api/spec/ipsec.h
+++ b/include/odp/api/spec/ipsec.h
@@ -1264,6 +1264,81 @@ typedef struct odp_ipsec_status_t {
 } odp_ipsec_status_t;
 
 /**
+ * IPsec statistics
+ */
+typedef struct odp_ipsec_stats_t {
+	/** All packets of an SA received, decrypted
+	 * and authenticated successfully */
+	uint64_t in_sa_pkts;
+
+	/** All packets of an SA received but
+	 * discarded due to anti replay error
+	 */
+	uint64_t in_ar_err_pkts;
+
+	/** All packets of an SA but discarded
+	 * due to decryption or authentication errors.
+	 */
+	uint64_t in_dec_err_pkts;
+
+	/** All packets of an SA encrypted and
+	 * sent successfully
+	 */
+	uint64_t out_sa_pkts;
+
+	/** All packets of an SA with
+	 * encryption failures
+	 */
+	uint64_t out_enc_err_pkts;
+
+	/** Pending counter for soft byte limit of the SA */
+	uint64_t byte_limit_soft_pending;
+
+	/** Pending counter for hard byte limit of the SA */
+	uint64_t byte_limit_hard_pending;
+
+	/** Pending counter for soft pkt limit of the SA */
+	uint64_t pkt_limit_soft_pending;
+
+	/** Pending counter for hard pkt limit of the SA */
+	uint64_t pkt_limit_hard_pending;
+} odp_ipsec_stats_t;
+
+/**
+ * Get IPsec stats.
+ *
+ * Get IPsec stats for multiple SAs.
+ *
+ * @param          sa       Array of IPsec SA handles.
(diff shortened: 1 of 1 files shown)


JannePeltonen:
> Present ODP spec does not have a provision to get the stats of IPsec

This is not totally accurate since the current IPsec API enables statistics gathering by the application. The API provides status and status events about errors so that the application can maintain statistics at the granularity it wishes (there may perhaps be a gap in the inline input case since the API does not seem to specify for which incoming packets ipsec events must be generated.)

That said, an API addition like this might be useful for performance reasons.

It needs to be documented how this API works with respect to SA creation and deletion, i.e. when the stats function may be called.

I do not see what purpose the clearing function serves. It is not synchronized with reading the statistics so it is hard to see how an application could use it without possibly losing part of the statistics.

The definition of the actual counters requires some more work. e.g.: Only ESP is included and AH is missing, but it does not seem necessary to have such separate counters in the first place since an SA is anyway for a specific IPsec transformation. I did not understand the bytelimit and pkt_limit statistics.

vvelumuri:
> 
> 
> > Present ODP spec does not have a provision to get the stats of IPsec
> 
> This is not totally accurate since the current IPsec API enables statistics gathering by the application. The API provides status and status events about errors so that the application can maintain statistics at the granularity it wishes (there may perhaps be a gap in the inline input case since the API does not seem to specify for which incoming packets ipsec events must be generated.)
> 
Will modify the description. 

> That said, an API addition like this might be useful for performance reasons.
> 
> It needs to be documented how this API works with respect to SA creation and deletion, i.e. when the stats function may be called.
> 
Ack. Will update the documentation with required information. 

> I do not see what purpose the clearing function serves. It is not synchronized with reading the statistics so it is hard to see how an application could use it without possibly losing part of the statistics.
> 
clearing function can be useful for diagnosis purpose.

> The definition of the actual counters requires some more work. e.g.: Only ESP is included and AH is missing, but it does not seem necessary to have such separate counters in the first place since an SA is anyway for a specific IPsec transformation. I did not understand the bytelimit and pkt_limit statistics.
Ack, will change the naming convention of counters as they are not specific to ESP. 
the intention for limit counters is to indicate the pending limits. Will change the documentation and naming convention accordingly.

vvelumuri:
> 
> 
> > > Present ODP spec does not have a provision to get the stats of IPsec
> > 
> > 
> > This is not totally accurate since the current IPsec API enables statistics gathering by the application. The API provides status and status events about errors so that the application can maintain statistics at the granularity it wishes (there may perhaps be a gap in the inline input case since the API does not seem to specify for which incoming packets ipsec events must be generated.)
> 
> Will modify the description.
> 
> > That said, an API addition like this might be useful for performance reasons.
> > It needs to be documented how this API works with respect to SA creation and deletion, i.e. when the stats function may be called.
> 
> Ack. Will update the documentation with required information.
> 
> > I do not see what purpose the clearing function serves. It is not synchronized with reading the statistics so it is hard to see how an application could use it without possibly losing part of the statistics.
> 
> clearing function can be useful for diagnosis purpose.
> 
> > The definition of the actual counters requires some more work. e.g.: Only ESP is included and AH is missing, but it does not seem necessary to have such separate counters in the first place since an SA is anyway for a specific IPsec transformation. I did not understand the bytelimit and pkt_limit statistics.
Ack, will change the naming convention of counters as they are not specific to ESP.
The intention for limit counters is to indicate the pending limits. Will change the documentation and naming convention accordingly.

PR #1014, opened 2020-08-25
[PATCH v2] api: ipsec: extended stats framework for IPsec.

vvelumuri:
New framework to get and clear complete stats for an IPsec SA.
Platforms can maintain various stats corresponding to an SA.
These stats or counters may differ from platform to platform.
This framework helps application to get the stats agnostick of platform
implimentation.

The framework provide two APIs. odp_ipsec_xstats_names() will return
an array of all the names of supported stats. odp_ipsec_xstats() will
return an array of values for all supported stats. Both arrays are
matched by array index. i.e: xstats[i].value is correspond to the
xstat_names[i].name.

IPsec capability exposes the number of extended stats supported by the
platform. The value 0 indicates the platform does not support the
extended cpabilities.

Application need to allocate the memmory for the array of names and
values before invoking the APIs. Application can invoke these APIs
in any order. The preferred order of invoking is:
odp_ipsec_xstats_names() followed by odp_ipsec_xstats().

### DIFF
--- a/include/odp/api/spec/ipsec.h
+++ b/include/odp/api/spec/ipsec.h
@@ -33,6 +33,11 @@ extern "C" {
  * IPSEC Security Association (SA)
  */
 
+/**
+ * Maximum length for extended statistics names
+ */
+#define ODP_IPSEC_XSTATS_NAME_MAX_LEN 64
+
  /**
  * @def ODP_IPSEC_SA_INVALID
  * Invalid IPSEC SA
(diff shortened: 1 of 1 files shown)


psavol:
Is this implementation specific IPsec statistics still needed when we have standard stats API (https://github.com/OpenDataPlane/odp/pull/1060) coming ? If need, could we turn this API into a implementation specific print as output instead of the struct?

PR #1015, opened 2020-08-25
[PATCH v3] Misc fixes to support older kernel versions

matiaselo:
V2:

- Added debug print about recreated TPID (Janne)
- Handle endianness in random_null (Janne)
- Updated Linux dependency (new commit)

V3:
- Added check that RAND_MAX is a power of two minus 1 and cleaned commit message (Janne)

### DIFF
--- a/platform/linux-generic/odp_random_null.c
+++ b/platform/linux-generic/odp_random_null.c
@@ -8,10 +8,16 @@
 #include <stdint.h>
 #include <stdlib.h>
 #include <odp/api/random.h>
+#include <odp/api/byteorder.h>
 #include <odp/api/cpu.h>
+#include <odp/api/debug.h>
 #include <odp_init_internal.h>
 
 #include <time.h>
+
+/* Assume at least two rand bytes are available and RAND_MAX is power of two - 1 */
+ODP_STATIC_ASSERT(RAND_MAX >= UINT16_MAX, "RAND_MAX too small");
+ODP_STATIC_ASSERT((RAND_MAX & (RAND_MAX + 1))  ==  0, "RAND_MAX not power of two - 1");
 
 odp_random_kind_t odp_random_max_kind(void)
 {
@@ -24,12 +30,14 @@ static int32_t _random_data(uint8_t *buf, uint32_t len, uint32_t *seed)
 		uint32_t rand_word;
 		uint8_t rand_byte[4];
 	} u;
-	uint32_t i = 0, j;
+	uint32_t i = 0, j, k;
 
 	while (i < len) {
 		u.rand_word = rand_r(seed);
 
-		for (j = 0; j < 4 && i < len; j++, i++)
+		/* Use two least significant bytes */
+		j = ODP_LITTLE_ENDIAN ? 0 : 2;
+		for (k = 0; k < 2 && i < len; i++, j++, k++)
 			*buf++ = u.rand_byte[j];
 	}
 
--- a/platform/linux-generic/pktio/socket_mmap.c
+++ b/platform/linux-generic/pktio/socket_mmap.c
@@ -42,7 +42,11 @@
 #include <protocols/ip.h>
 
 /* VLAN flags in tpacket2_hdr status */
+#ifdef TP_STATUS_VLAN_TPID_VALID
 #define VLAN_VALID (TP_STATUS_VLAN_VALID | TP_STATUS_VLAN_TPID_VALID)
+#else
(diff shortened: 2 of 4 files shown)


platform/linux-generic/pktio/socket_mmap.c
```diff
@@ -243,7 +247,21 @@ static inline unsigned pkt_mmap_v2_rx(pktio_entry_t *pktio_entry,
...
 			type  = (uint16_t *)(uintptr_t)
 				(mac + 2 * _ODP_ETHADDR_LEN);
+
+			#ifdef TP_STATUS_VLAN_TPID_VALID
 			*type = odp_cpu_to_be_16(tp_hdr->tp_vlan_tpid);
+			#else
+			/* Fallback for old kernels (< v3.14) */
+			uint16_t *type2;
+
+			type2 = (uint16_t *)(uintptr_t)(mac + (2 * _ODP_ETHADDR_LEN) + vlan_len);
```
JannePeltonen, line 257: What is the point of the intermediate cast to uintptr_t?

Btw, this line assumes mac is correctly aligned for the alignment requirement of uint16_t and the uintptr_t does not change that. But it is not a new assumption so it should be ok here too.

platform/linux-generic/pktio/socket_mmap.c
```diff
@@ -243,7 +247,21 @@ static inline unsigned pkt_mmap_v2_rx(pktio_entry_t *pktio_entry,
...
+			/* Recreate TPID 0x88a8 for double tagged and 0x8100 for single tagged */
+			if (*type2 == odp_cpu_to_be_16(0x8100))
+				*type = odp_cpu_to_be_16(0x88a8);
```
JannePeltonen, line 260: This seems to be just an educated guess. Isn't there any better way? Sometimes TPID values 0x8100 and even 0x9100 are used in the outer VLAN tag with double tagging,

platform/linux-generic/odp_random_null.c
```diff
@@ -29,7 +34,7 @@ static int32_t _random_data(uint8_t *buf, uint32_t len, uint32_t *seed)
 	while (i < len) {
 		u.rand_word = rand_r(seed);
 
-		for (j = 0; j < 4 && i < len; j++, i++)
+		for (j = 0; j < NUM_RAND_BYTES && i < len; j++, i++)
```
JannePeltonen, line 37: This does not work correctly with big endian byte order as it would use the most significant bits returned by rand_r(). I would suggest either changing the code to endian-neutal or adding an assert about byte order.

platform/linux-generic/pktio/socket_mmap.c
matiaselo: Some clang versions in the CI are trigger happy to complain about casts, so used the same patter as couple lines above.

platform/linux-generic/pktio/socket_mmap.c
matiaselo: The necessary header information is already lost at this point. I added a debug print about this.

platform/linux-generic/odp_random_null.c
matiaselo: Fixed in V2.

platform/linux-generic/odp_random_null.c
```diff
@@ -24,12 +29,14 @@ static int32_t _random_data(uint8_t *buf, uint32_t len, uint32_t *seed)
...
 		uint8_t rand_byte[4];
 	} u;
-	uint32_t i = 0, j;
+	uint32_t i = 0, j, k;
 
 	while (i < len) {
 		u.rand_word = rand_r(seed);
 
-		for (j = 0; j < 4 && i < len; j++, i++)
+		/* Use two least significant bytes */
```
JannePeltonen, line 37: This actually requires that RAND_MAX is a power of two minus 1. Otherwise you get a biased distribution and may see the original problem in some form. In practice RAND_MAX satisfies the requirement in a typical Linux distro, so maybe the easiest fix is to just additionally assert that (RAND_MAX & (RAND_MAX + 1)) == 0. If you did not want to do that, you would probably have to discard rand_r results that exceed the highest power-of-two-minus-one that is less or equal to RAND_MAX and then retry rand_r. But maybe it is not worth the trouble for this piece of code.

You might also want to tune the commit message a little (it talks about two "first" bytes). And no need for the suggested-by tag.

matiaselo: Fixed in V3

JannePeltonen (approved):

PR #1016, opened 2020-08-26
[PATCH v3] linux-gen: pktio: use odp atomics for in_discards counter

matiaselo:
Previously, build would fail on 32-bit systems since __atomic_fetch_add()
was called using a 64-bit variable.

V2:
- Updated commit message (Janne)

### DIFF
--- a/platform/linux-generic/include/odp_packet_io_internal.h
+++ b/platform/linux-generic/include/odp_packet_io_internal.h
@@ -93,7 +93,12 @@ struct pktio_entry {
 	} state;
 	odp_pktio_config_t config;	/**< Device configuration */
 	classifier_t cls;		/**< classifier linked with this pktio*/
-	odp_pktio_stats_t stats;	/**< statistic counters for pktio */
+	/* Driver level statistics counters */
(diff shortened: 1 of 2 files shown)


JannePeltonen:
Aside from the compilation issue, that I did not look at yet, this change seems to fix a couple of other problems in all environments:

1) It fixes the data race between the scheduling CB and the pktio drivers.

2) it fixes the in_discard counting otherwise too because currently the odp_pktio_stats contained in pktio_entry are handled in pktio specific ways and one cannot just assume that incrementing the in_discard counter there would result in the right thing. Sometimes (like for loopback) the stats in the pktio_entry are the actual counters (so the current code sort-of works) but sometimes they record the lower layer counter values at counter reset time (like for socket IO) so that the increment of in_discard in the sched CB results in the user visible in_discard count actually decreasing. And sometimes (with DPDK) the stats in pktio_entry are not used at all.

This patch separates the stats done by each driver in its own way from the in_discard at the generic layer and fixes all the above issues AFAICS. You may want to mention something about this goodness in the commit comment,

JannePeltonen (approved):

PR #1017, opened 2020-08-26
[PATCH v1] api: ipsec: add ipsec operating mode to sa

vvelumuri:
With the IPsec mode in global config set to inline, packets can be
processed in inline mode or in asynchronous mode. When a platform has
the ability to support both inline and asynchronous modes in parallel,
application can choose to process some SAs in inline mode and some in
asynchronous mode.

This change set adds the provision to configure the IPsec mode for an SA.
With this provision application can configure the SA to either inline or
asynchronous mode when global IPsec mode is inline.

### DIFF
--- a/include/odp/api/spec/ipsec.h
+++ b/include/odp/api/spec/ipsec.h
@@ -647,6 +647,20 @@ typedef enum odp_ipsec_ip_version_t {
  * IPSEC Security Association (SA) parameters
  */
 typedef struct odp_ipsec_sa_param_t {
+	/** IPsec operation mode of SA
+	 *
+	 * When the IPsec mode in global config is: ODP_IPSEC_OP_MODE_INLINE,
+	 * this value takes precedence over the global config value.
+	 * In all other cases, IPsec mode in global config will take
+	 * precedence over this value.
+	 *
+	 * With IPsec mode in global config set to ODP_IPSEC_OP_MODE_INLINE,
+	 * value for this can be set to either ODP_IPSEC_OP_MODE_INLINE
+	 * or ODP_IPSEC_OP_MODE_ASYNC.
+	 *
+	 */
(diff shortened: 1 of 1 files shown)


JannePeltonen:
This PR would need a clear rationale for why the proposed change is needed or useful.

Current ODP API allows both asynchronous and inline operations even for the same SA when operating in the inline mode. This is intentional to make it possible to fall back look-a-side IPsec processing in cases where inline processing cannot be done. For instance, if incoming packets for the same SA may come through different paths (e.g. different interfaces or different encapsulation) and one of the paths may not support inline processing but we still need the packets to be processed using the same SA.

Maybe the sync/async/inline mode setting does not make the above very clear, even though the text of the API spec does say that async ops are allowed in inline mode. Basically the whole IPsec subsystem works either in sync or async "mode" so that async mode always supports look-a-side operations and optionally also inline operations (and then we call the combination inline mode).

The proposed change does not bring any new functionality to the API but adds a restriction that a given SA must be used either in look-a-side manner or inline manner for its whole life time. Adding such a restriction would require a good reason.

Is there some HW restriction that makes it impossible to support the current IPsec API? If so, and if a change like this is really necessary, then the current text regarding simultaneous use of inline and look-a-side type async operations would need to be updated. And maybe it would make sense to add some capability flags and make the restriction depend on the implementation.

A somewhat less severe restriction which might not be such a big problem to some of the use cases would be to require an SA to be either in look-a-side or inline mode (allowing only the respective ops) but making it possible to change the mode on the fly. But that might be difficult to do from concurrency and synchronization perspective.

vvelumuri:
As suggested, lookup mode in the sa param can be used to configure the SA for inline or lookaside mode.  The SA with lookup_mode = ODP_IPSEC_LOOKUP_DISABLED can be treated for inline.

PR #1018, opened 2020-08-26
[PATCH v3] api: packet vector producer support

jerinjacobk:
## api: pool: introduce vector pool

Introduce the vector pool to hold the vector of a generic type such
as packets. Updated odp_pool_capability_t to advertise the new vector
pool capabilities and odp_pool_param_t to configure the vector pool.
The application shall use the max_vector_size parameter to configure
the max number of general types such as odp_packet_t that
can hold in the vector.

When vector pool used as odp_packet_t container,
either pktin subsystem can allocate a buffer to hold/form
the vector in packet input path as pktin producer or
the application can allocate the vector pool using
odp_packet_vector_alloc() from packet API to create a vector
of packets to store an array of packets as a vector.

## api: event: introduce packet vector event

Introduce packet vector events which hold an array of odp_packet_t.

This patch also introduce the odp_packet_vector_from_event()
and odp_packet_vector_to_event() event conversion APIs.

In order to build the linux-generic in --disable-abi-compat mode,
A dummy definition of odp_packet_vector_t has been added.

## api: pktio: configure pktin vector producer

There are two configuration aspects of the pktin vector producer.

First, When pktin subsystem produces the vector of packets,
it needs to choose the ODP_VECTOR_POOL to hold/form
the vector.

Second, The max timeout interval to accumulate a vector for
packets.

The odp_pktin_vector_config_t depicts the configuration
parameters.

Either it max timeout reaches or max_vector_size configured
in the pool reaches the value, the pktin subsystem forms the
odp_packet_vector_t and inject as an event to the event queue.

The pktin configuration available either through pktio or
classification subsystem based on the final odp queue delivery.

This patch also introduces odp_pktin_vector_capability_t capability
parameter to probe the pktin vector producer availability in the
implementation as well as to know the max and min timeout value
for accumulating the packet by the pktin vector producer and
also the max_vector_size.

## api: packet: introduce packet vector handling routines

Introduce odp_packet_vector_t type handling functions such
alloc and free the vector from packet vector pool,
get and set the vector size and get the vector packets array
for creating, manipulating the packet vector abstracted
using the odp_packet_vector_t.

### DIFF
--- a/include/odp/api/abi-default/packet.h
+++ b/include/odp/api/abi-default/packet.h
@@ -19,16 +19,21 @@ typedef struct { char dummy; /**< @internal Dummy */ } _odp_abi_packet_t;
 /** @internal Dummy  type for strong typing */
 typedef struct { char dummy; /**< *internal Dummy */ } _odp_abi_packet_seg_t;
 
+/** @internal Dummy type for strong typing */
+typedef struct { char dummy; /**< *internal Dummy */ } _odp_abi_packet_vector_t;
+
 /** @ingroup odp_packet
  *  @{
  */
 
 typedef _odp_abi_packet_t *odp_packet_t;
 typedef _odp_abi_packet_seg_t *odp_packet_seg_t;
+typedef _odp_abi_packet_vector_t *odp_packet_vector_t;
 
 #define ODP_PACKET_INVALID        ((odp_packet_t)0)
 #define ODP_PACKET_SEG_INVALID    ((odp_packet_seg_t)0)
 #define ODP_PACKET_OFFSET_INVALID 0xffff
+#define ODP_PACKET_VECTOR_INVALID   ((odp_packet_vector_t)0)
 
 typedef uint8_t odp_proto_l2_type_t;
 
--- a/include/odp/api/spec/packet.h
+++ b/include/odp/api/spec/packet.h
@@ -2028,6 +2028,147 @@ odp_packet_vector_t odp_packet_vector_from_event(odp_event_t ev);
  */
 odp_event_t odp_packet_vector_to_event(odp_packet_vector_t vpkt);
 
+/**
+ * Allocate a packet vector from a packet vector pool
+ *
+ * Allocates a packet vector from the specified packet vector pool.
+ * The pool must have been created with the ODP_POOL_VECTOR type.
+ *
+ * @param pool Packet vector pool handle
+ *
+ * @return Handle of allocated packet vector
+ * @retval ODP_PACKET_VECTOR_INVALID  Packet vector could not be allocated
+ *
+ * @note The newly allocated vector shall not contain any packets, instead, alloc
+ * operation shall reserve the space for odp_pool_param_t::vector::max_vector_size packets.
+ */
+odp_packet_vector_t odp_packet_vector_alloc(odp_pool_t pool);
+
+/**
+ * Free packet vector
+ *
+ * Frees the packet vector into the packet vector pool it was allocated from.
+ *
+ * @param vpkt Packet vector handle
+ *
+ * @note This API just frees the vector, not any packets inside the vector.
+ * Application can use odp_event_free() to free the vector and packets inside
+ * the vector.
+ */
+void odp_packet_vector_free(odp_packet_vector_t vpkt);
+
+/**
+ * Get packet vector table
+ *
+ * Packet vector table is a array of packets(odp_packet_t) stored
+ * in contiguous memory location. Upon completion of this API,
+ * the implementation returns the packet table pointer in pkts_tbl.
+ *
+ * @param vpkt Packet vector handle
+ * @param [out] pkts_tbl Points to packet vector table
+ *
+ * @return Number of packets available in the vector.
+ *
+ * @note When pktin subsystem is producing the packet vectors,
+ * odp_pktin_vector_config_t::vector_pool shall be used to configure
+ * the pool to form the vector table.
+ *
+ * @note The maximum number of packets can hold this vector is defined
+ * by odp_pool_param_t:vector:max_vector_size. The return value of this
+ * function will not be greater than odp_pool_param_t:vector:max_vector_size
+ *
+ * @note The pkts_tbl points to the packet vector table. Application can edit the packet
+ * handles in the table directly(up to odp_pool_param_t::vector::max_vector_size).
+ * Application must update the size of the table using odp_packet_vector_size_set() when
+ * there is a change in the size of the vector.
+ *
+ * @note Invalid packet handle(ODP_PACKET_INVALID) is not allowed to store in the table to
+ * allow consumers of odp_packet_vector_t handle to have optimized implementation.
+ * So consumption of packets in the middle of the vector would call for moving
+ * remaining packets up to form the contiguous array of packets and update
+ * the size of the new vector using odp_packet_vector_size_set().
+ *
+ * @note The table memory is backed by the vector pool buffer. The ownership of the
+ * table memory is linked to ownership of the event. i.e after sending the event to a queue,
+ * the sender loses ownership to the table also.
+ *
+ */
+uint32_t odp_packet_vector_tbl(odp_packet_vector_t vpkt, odp_packet_t **pkts_tbl);
+
+/**
+ * Get number of packets in a vector
+ *
+ * @param vpkt Packet vector handle
+ *
+ * @return The number of packets available in the vector
+ */
+uint32_t odp_packet_vector_size(odp_packet_vector_t vpkt);
+
+/**
+ * Set the number of packets stored in a vector
+ *
+ * Update the number of packets stored in a vector. When the application is
+ * producing the packet vector, this function shall be used by the application
+ * to set the number of packets available in this vector.
+ *
+ * @param vpkt Packet vector handle
+ * @param size Number of packets in this vector
+ *
+ * @note The maximum number of packets can hold this vector is defined
+ * by odp_pool_param_t::vector::max_vector_size. The size value should not be
+ * greater than odp_pool_param_t::vector::max_vector_size
+ *
+ * @note (0 .. size - 1) handles in the vector table needs to be valid packet handles.
+ * @see odp_packet_vector_tbl()
+ *
+ */
+void odp_packet_vector_size_set(odp_packet_vector_t vpkt, uint32_t size);
+
+/**
+ * Perform full packet vector validity check
+ *
+ * The operation may consume considerable number of cpu cycles depending on
+ * the check level.
+ *
+ * @param vpkt Packet vector handle
+ *
+ * @retval 0 Packet vector is not valid
+ * @retval 1 Packet vector is valid
+ */
+int odp_packet_vector_valid(odp_packet_vector_t vpkt);
+
+/**
+ * Get packet vector pool
+ *
+ * Returns handle to the packet vector pool where the packet vector was allocated from.
+ *
+ * @param vpkt Packet vector handle
+ *
+ * @return Packet vector pool handle
+ */
+odp_pool_t odp_packet_vector_pool(odp_packet_vector_t vpkt);
+
+/**
+ * Print packet vector debug information
+ *
+ * Print all packet vector debug information to the ODP log.
+ *
+ * @param vpkt Packet vector handle
+ */
+void odp_packet_vector_print(odp_packet_vector_t vpkt);
+
+/**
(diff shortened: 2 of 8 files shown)


include/odp/api/spec/packet_io.h
```diff
@@ -159,6 +159,18 @@ typedef struct odp_pktin_queue_param_ovr_t {
...
+ * Packet vector producer configurations
+ */
+typedef struct odp_pktin_vector_prod_config_t {
+	/** Packet vector pool */
+	odp_pool_t packet_vect_pool;
+	/** Maximum timeout in nanosec to wait for producer form the vector of
+	 * packet event(odp_packet_vect_t).
+	 */
+	uint64_t max_vector_prod_tmo_ns;
+} odp_pktin_vector_prod_config_t;
```
psavol, line 172: I think "prod" part can be dropped from type name: odp_pktin_vector_config_t

jerinjacobk: Ack

include/odp/api/spec/packet_io.h
```diff
@@ -568,6 +583,15 @@ typedef struct odp_pktio_capability_t {
...
 	odp_bool_t ODP_DEPRECATE(loop_supported);
+
+	/** Maximum timeout in nanosec to wait for producer form the vector of packets
+	 * event(odp_packet_vect_t).
+	 * The value zero denotes the producer is not capablable of producing
+	 * the vector packet events a.k.a odp_packet_vect_t.
+	 * @see odp_packet_vect_t
+	 * @see odp_pktin_vector_prod_config_t::max_vector_prod_tmo_ns
+	 */
+	uint64_t max_vector_prod_tmo_ns;
```
psavol, line 594: Maybe a struct is needed instead of using tmo capa for supported/not supported. Max vector size in packet input could differ from max vector size in other parts of the system, so size is likely needed here as well.

struct {
    odp_bool_t supported;
    uint32_t max_size;
    uint64_t max_tmo_ns;
} pktin_vector;

include/odp/api/spec/packet.h
```diff
@@ -1979,6 +1989,147 @@ int8_t odp_packet_shaper_len_adjust(odp_packet_t pkt);
...
+ * @param vpkt Packet vector handle
+ */
+void odp_packet_vect_free(odp_packet_vect_t vpkt);
+
+/**
+ * Get packet vector table
+ *
+ * Packet vector table is a array of packets(odp_packet_t) stored
+ * in contiguous memory location. Upon completion of this API,
+ * the implementation returns the packet table pointer in pkts_tbl.
```
psavol, line 2047: Usually, we have get/set functions to read and write event metadata. Since this function differs from that, it should be documented that application can edit packet handles in the table directly (up to max_size) and must set new vector size if number of handles was changed.

jerinjacobk: Ack

include/odp/api/spec/packet_io.h
jerinjacobk: Ack

include/odp/api/spec/packet.h
```diff
@@ -1979,6 +1989,147 @@ int8_t odp_packet_shaper_len_adjust(odp_packet_t pkt);
...
+ *
+ * Allocates a packet vector from the specified packet vector pool.
+ * The pool must have been created with the ODP_POOL_PACKET_VECTOR type.
+ *
+ * @param pool Packet vector pool handle
+ *
+ * @return Handle of allocated packet vector
+ * @retval ODP_PACKET_VECT_INVALID  Packet vector could not be allocated
+ */
+odp_packet_vect_t odp_packet_vect_alloc(odp_pool_t pool);
```
psavol, line 2031: For clarity, this could mention that vector is empty after alloc (no packets were allocated, just the container)

jerinjacobk: Ack

include/odp/api/spec/packet.h
```diff
@@ -1979,6 +1989,147 @@ int8_t odp_packet_shaper_len_adjust(odp_packet_t pkt);
...
+
+/**
+ * Free packet vector
+ *
+ * Frees the packet vector into the packet vector pool it was allocated from.
+ *
+ * @param vpkt Packet vector handle
+ */
+void odp_packet_vect_free(odp_packet_vect_t vpkt);
```
psavol, line 2040: Need to document if packets in the vector are freed or not. I think this basic free call would just free the vector event (ignores all packet handles). We could then have another free call that frees all packets and the vector.

jerinjacobk: How about adding an argument for this function instead of a new function. say bool free_elements or so?

psavol:
General question: we may want to expand vectors also to buffers (SW messages), or other event types later on. How we could avoid unnecessary repetition on APIs (e.g. pool) ?

Options on re-use level:
1) There could be e.g. only one pool type for all vector events. There would be a single set of pool params / capa for all vector events. 
2) In addition, there could be single event type (odp_event_vector_t) for vector events, but different type codes (ODP_EVENT_PACKET_VECTOR, ODP_EVENT_BUFFER_VECTOR, ...). This would re-use alloc/free calls. Accessor functions could be still different (e.g. odp_packet_vector_tbl(odp_event_vector_t v, odp_packet_t **pkts) vs, odp_buffer_vector_tlb(odp_event_vector_t v, odp_buffer_t **bufs).

jerinjacobk:
> General question: we may want to expand vectors also to buffers (SW messages), or other event types later on. How we could avoid unnecessary repetition on APIs (e.g. pool) ?

Yes. The pool can be made as common. 

> 
> Options on re-use level:
> 
> 1. There could be e.g. only one pool type for all vector events. There would be a single set of pool params / capa for all vector events.

Ack

> 2. In addition, there could be single event type (odp_event_vector_t) for vector events, but different type codes (ODP_EVENT_PACKET_VECTOR, ODP_EVENT_BUFFER_VECTOR, ...).

You mean, odp_packet_vect_tbl() takes odp_event_vector_t argument instead of odp_packet_vect_t. Right? If so,
It works for me.
 
> This would re-use alloc/free calls. Accessor functions could be still different (e.g. odp_packet_vector_tbl(odp_event_vector_t v, odp_packet_t **pkts) vs, odp_buffer_vector_tlb(odp_event_vector_t v, odp_buffer_t **bufs).

Yes, it makes sense to keep different Accessor functions.

I will send the next version with the above comments then.

include/odp/api/spec/packet_io.h
psavol: "vector" can be also removed from config struct field names

include/odp/api/spec/packet_io.h
matiaselo: `uint32_t min_size` could be also useful.

include/odp/api/spec/packet.h
psavol: odp_event_free() should work correctly for any event, so it should free all packets inside a vector.

psavol: odp_packet_vector_free() would be the basic form that frees only the container. odp_packet_vector_free_all() ??? could be added later for freeing both container and all remaining packet.

include/odp/api/spec/packet.h
psavol: ownership of the table is linked to the ownership of the event. e.g. after sending the event to a queue, sender looses ownership to the table also.

psavol: packet handles must be packed into the start of the table (?)

include/odp/api/spec/packet.h
```diff
@@ -1979,6 +1989,147 @@ int8_t odp_packet_shaper_len_adjust(odp_packet_t pkt);
...
+ * to set the number of packets available in this vector.
+ *
+ * @param vpkt Packet vector handle
+ * @param size Number of packets in this vector
+ *
+ * @note The maximum number of packets can hold this vector is defined
+ * by odp_pool_param_t::pktv::max_vect_size. The size value should not be
+ * greater than odp_pool_param_t::pktv::max_vect_size
+ */
+void odp_packet_vect_size_set(odp_packet_vect_t vpkt, uint32_t size);
```
psavol, line 2086: if packet_invalids are allowed then this size includes also invalids (?)

jerinjacobk: For the optimized odp_packet_vect_t consumer implementation it is better to have only valid packet handles like other vector processing frameworks.

psavol: OK. Remember to document that 0 ... size - 1 handles in the vector table needs to be valid handles (when application edits the table).

JannePeltonen: I think we should allow application to write anything in the vector as long as the handles are valid when the vector is passed to ODP (through the size_set function or otherwise). It is just direct memory access after all.

Allowing invalid handles in the vector in size_set() and when converting the vector to an event and passing the event through a queue might be useful in some cases as we discussed, but it will be possible
to add it later without breaking backward compatibility.

jerinjacobk: Yes. It is possible to add it later when we understand the implications without ABI change. For now, I will document it is allowed to have valid handles. The current use case may just to give the table to pktout_send() function to transmit

include/odp/api/spec/packet.h
```diff
@@ -167,6 +167,16 @@ extern "C" {
  * Layer 4 protocol type is ROHC
  */
 
+/**
+ * @typedef odp_packet_vector_t
+ * ODP packet vector
+ */
+
+/**
+ * @def ODP_PACKET_VECT_INVALID
```
psavol, line 176: Should be ODP_PACKET_VECTOR_INVALID (use "vector" instead of "vect" in type names)

include/odp/api/abi-default/packet.h
```diff
@@ -19,16 +19,21 @@ typedef struct { char dummy; /**< @internal Dummy */ } _odp_abi_packet_t;
 /** @internal Dummy  type for strong typing */
 typedef struct { char dummy; /**< *internal Dummy */ } _odp_abi_packet_seg_t;
 
+/** @internal Dummy type for strong typing */
+typedef struct { char dummy; /**< *internal Dummy */ } _odp_abi_packet_vect_t;
+
```
psavol, line 24: "vect" -> "vector"

include/odp/api/abi-default/packet.h
```diff
@@ -19,16 +19,21 @@ typedef struct { char dummy; /**< @internal Dummy */ } _odp_abi_packet_t;
...
  */
 
 typedef _odp_abi_packet_t *odp_packet_t;
 typedef _odp_abi_packet_seg_t *odp_packet_seg_t;
+typedef _odp_abi_packet_vect_t *odp_packet_vector_t;
 
 #define ODP_PACKET_INVALID        ((odp_packet_t)0)
 #define ODP_PACKET_SEG_INVALID    ((odp_packet_seg_t)0)
 #define ODP_PACKET_OFFSET_INVALID 0xffff
+#define ODP_PACKET_VECT_INVALID   ((odp_packet_vector_t)0)
```
psavol, line 36: "VECT" -> "VECTOR"

include/odp/api/spec/packet.h
jerinjacobk: Will be fixed in next version.

include/odp/api/abi-default/packet.h
jerinjacobk: Will be fixed in the next version.

include/odp/api/abi-default/packet.h
jerinjacobk: Will be fixed in next version.

include/odp/api/spec/packet_io.h
```diff
@@ -159,6 +159,34 @@ typedef struct odp_pktin_queue_param_ovr_t {
 	odp_schedule_group_t group;
 } odp_pktin_queue_param_ovr_t;
 
+/**
+ * Packet input vector configuration
+ */
+typedef struct odp_pktin_vector_config_t {
```
PavanNikhilesh, line 165: I think we need to have a enable flag here to indicate that the vector config is valid. Makes config validation simpler,

jerinjacobk: Makes sense. I will add "enable" flag in next version.

matiaselo:
Replaced by #1062 .

Issue #800, opened 2019-02-05
Unexpected IV causes IPsec API validation to fail


matiaselo:
Janne Peltonen 2018-10-08 12:15:36 UTC
IPsec validation code expects ODP implementation to use the same IV as in the test vectors in the output ESP and AH packets. This causes valid implementations to fail validation since the implementation can pick the IV value in any standard compliant way it wants.

Comment 1 Bill Fischofer 2018-10-09 12:17:01 UTC
Reassigning to Dmitry for review/comment.

Comment 2 Dmitry Eremin-Solenikov 2018-12-06 13:17:50 UTC
Janne, could you please specify, which test cases are you referring to?
I've tried not to enforce IV.

Comment 3 Janne Peltonen 2018-12-10 13:23:23 UTC
I am referring to all IPsec output tests. The test cases compare the output packet produced by the IPsec implementation against the expected output packet defined in the test vector. The packets match only if the IPsec implementation chooses the same IV that is in the test vector as otherwise the encrypted bytes are totally different.

You can see which test cases could fail with other implementations by changing the IV generation a little, like this, and running make check:
```
--- a/platform/linux-generic/odp_ipsec.c
+++ b/platform/linux-generic/odp_ipsec.c
@@ -1032,6 +1032,7 @@ static int ipsec_out_iv(ipsec_state_t *state,
                        return -1;
        }

+       state->iv[0] += 1;
        return 0;
 }
```
Comment 4 Dmitry Eremin-Solenikov 2018-12-10 14:44:58 UTC
Janne,

I've rechecked IPsec testsuite. Please correct me if I'm wrong. Output-with-compare tests fall into two major groups:

- Null encryption + HMAC-SHA-somthing. These test vectors do not require IV at all and thus should generate the same package in all circumstances

- Null encryption + AES-GMAC. This really looks like a mistake from my side, I will correct these test cases not to use predefined test packages to compare against.

Did I miss any of the tests which use ipsec_check_out_one?

Comment 5 Janne Peltonen 2018-12-10 15:29:38 UTC
Sound right to me.

Maybe the corrected test cases could still do the full packet comparison if they notice that the IV is the same as in the test vector. This way the current implementation would get some extra test coverage.

PR #1019, opened 2020-08-28
[PATCH v5] IPsec validation test application fixes

asasidharan:
This PR consists of following two fixes to IPsec validation test applications:

1.  Fix test_out_ipv4_esp_null_aes_gmac_128() test function:
Perform both outbound and inbound processing similar to test_out_ipv4_esp_aes_gcm128. This is required as AES-GMAC algorithm uses IV for ICV generation and is generated by the ODP IPsec implementation during the IPsec outbound processing. The IV value is not known to the receiver until it receives the packet, but the test function expects a predefined IV value in the resulting packet.

2. Set L3 type for a test packet in ipsec_packet()
Similar to setting the l3 offset, set the L3 type to IPv4/IPv6 in ipsec_packet()

### DIFF
--- a/test/validation/api/ipsec/ipsec_test_out.c
+++ b/test/validation/api/ipsec/ipsec_test_out.c
@@ -453,7 +453,8 @@ static void test_out_ipv4_esp_aes_gcm128(void)
 static void test_out_ipv4_ah_aes_gmac_128(void)
 {
 	odp_ipsec_sa_param_t param;
-	odp_ipsec_sa_t sa;
+	odp_ipsec_sa_t out_sa;
+	odp_ipsec_sa_t in_sa;
 
 	ipsec_sa_param_fill(&param,
 			    false, true, 123, NULL,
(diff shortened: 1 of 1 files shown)


test/validation/api/ipsec/ipsec.c
```diff
@@ -447,6 +449,12 @@ odp_packet_t ipsec_packet(const ipsec_test_packet *itp)
 
 	odp_packet_user_ptr_set(pkt, PACKET_USER_PTR);
 
+	eth_hdr = (odph_ethhdr_t *)odp_packet_l2_ptr(pkt, NULL);
+	type = odp_be_to_cpu_16(eth_hdr->type);
+	if (type == ODPH_ETHTYPE_IPV4)
+		odp_packet_has_ipv4_set(pkt, 1);
```
JannePeltonen, line 455: This is not right. IPsec operations do not require the has_ipv4 or has_ipv6 flags to be set (see the api spec, e.g. ipsec.h:1287). The implementation is expected to look into the packet data to see whether the packet is IPv4 or IPv4. Setting the has_ipv4 or has_ipv6 flags here would make the validation test accept incorrect implementations that require the flags to be set according to the real packet type.

JannePeltonen:
The AES-GMAC fix looks good.

The problem is actually reported in issue #800. There is similar problem in  test_out_ipv4_esp_null_aes_gmac_128(). Could you perhaps fix that too? And please add this tag in the commit log: Fixes: https://github.com/OpenDataPlane/odp/issues/800

Please remove the second commit (the has_ipv4 & has_ipv6 addition) from this PR.

test/validation/api/ipsec/ipsec.c
asasidharan: Abandoning this change as pointed out, as per the spec only l3 and l4 offsets need to have valid values.

asasidharan:
> The problem is actually reported in issue #800. There is similar problem in test_out_ipv4_esp_null_aes_gmac_128(). Could you perhaps fix that too? And please add this tag in the commit log: Fixes: #800
> 
> Please remove the second commit (the has_ipv4 & has_ipv6 addition) from this PR.

Updated the pull request with fix for AES-GMAC test functions in both ESP and AH mode along with the tag "Fixes: #800". Also removed the second commit as suggested.

JannePeltonen (approved):

matiaselo:
Hi @asasidharan,

Please add Janne's reviewed-by tag to the commit message and do a rebase. After this I can merge the PR.

PR #1023, opened 2020-09-02
[PATCH v6] api: crypto: fix bit mode for aes eea2

asasidharan:
This pull request has two interrelated commits:

**1. Fix to length field in EEA2 test vectors.**
- The test vectors for EEA2 has incorrect length values configured in the current ODP.
- This fix updates the length values properly.

**2, Add bit-mode support to validation test result comparison**
- The first commit in this PR causes the linux-generic implementation to fail for the EEA2 tests.
- The reason being, the packet_cmp_mem() function which compares the result of crypto operation with the test vector does a memcmp() for the rounded up byte length.  But the length in bit for the test cases are not a multiple of 8.
- This commit adds support for comparison of bit mode crypto operation's result where the last byte is masked according to the actual length in bits and compared with the test vector's expected result.
- Here I have added two functions packet_cmp_mem_bytes() and packet_cmp_mem_bits() for byte mode and bit mode operations respectively.
- The existing packet_cmp_mem() function now calls one of the byte/bit mode function based on newly added  bit_mode parameter.
```
static int packet_cmp_mem(odp_packet_t pkt, uint32_t offset,
			              uint8_t *s, uint32_t len, odp_bool_t bit_mode) 
```
- With this second commit in the PR, the EEA2 tests passes as expected for linux-generic implementation.


JannePeltonen:
This looks basically ok to me, except that there is a problem with the crypto API: The API says that in bit_mode, cipher data length must be multiple of 8 or unspecified behavior ensues. So to remove the API violation, either the API needs to be changed or some workaround needs to be added in the test (maybe just passing the next largest multiple of 8 to the crypto op would be enough).

Now when looking at the API, I wonder if the bit order within bytes in the context of bit mode crypto operations should be better specified. Now there is no mention of it.

test/validation/api/crypto/odp_crypto_test_inp.c
```diff
@@ -23,8 +23,32 @@ struct suite_context_s {
...
 static struct suite_context_s suite_context;
 
-static int packet_cmp_mem(odp_packet_t pkt, uint32_t offset,
-			  void *s, uint32_t len)
+static int packet_cmp_mem_bits(odp_packet_t pkt, uint32_t offset,
+			       uint8_t *s, uint32_t len)
+{
+	int rc = -1;
+	uint32_t len_bytes = ((len + 7) >> 3);
+	uint8_t leftover_bits = len & 0x7;
```
JannePeltonen, line 31: A very minor comment: I think this would be more readable:
uint32_t len_bytes = (len + 7) / 8;
uint8_t leftover_bits = len % 8;

asasidharan:
Thank you for reviewing the changes.

>
> The API says that in bit_mode, cipher data length must be multiple of 8 or unspecified behavior ensues. So to remove the API violation, either the API needs to be changed or some workaround needs to be added in the test (maybe just passing the next largest multiple of 8 to the crypto op would be enough).

Would updating the Note2 for bit_mode field in the API spec as below be enough:
```
Note2:  When the data length is not a multiple of 8, the bits in the last byte that falls beyond the data length is undefined."
```

> 
> Now when looking at the API, I wonder if the bit order within bytes in the context of bit mode crypto operations should be better specified. Now there is no mention of it.

Bit order matters when the value being accessed spans more than 1 byte. Within a byte, masking of the bits should be fine for any bit order is what I understand.

test/validation/api/crypto/odp_crypto_test_inp.c
asasidharan: Will update this in v4.

JannePeltonen:
> Would updating the Note2 for bit_mode field in the API spec as below be enough:

Maybe just leaving out the note altogether would be fine. If the API does not promise anything about bits or bytes that are not part of the encryption results (which depends on the input data length but maybe always not directly) then they are undefined.

But I wonder why the note has been written in the first place. Maybe there is some limitation in the current ODP linux implementation for some algorithm and it has to be fixed if the API is changed?

JannePeltonen:
> Within a byte, masking of the bits should be fine for any bit order is what I understand.

If the length is not a multiple of 8, then it is not stated if the remaining bits are the most significant or the least significant bits of the byte.

asasidharan:
> Maybe just leaving out the note altogether would be fine. If the API does not promise anything about bits or bytes that are not part of the encryption results (which depends on the input data length but maybe always not directly) then they are undefined.

Agreed.

> But I wonder why the note has been written in the first place. Maybe there is some limitation in the current ODP linux implementation for some algorithm and it has to be fixed if the API is changed?

What I have observed with the EEA2 implementation on linux-generic is that, since it uses the openssl library which works at byte level granularity, the same limitation exists in linux-generic. But as you said if the API doesn't promise anything about bits or bytes that are not part of the encryption results, this shouldn't be considered as a limitation.

For EEA2/AES-CTR, the cipher text is obtained by simple XOR of the plain text blocks with key stream bits(generated using AES encryption of counter). And for the last block where the plain text is not aligned with the block size, the key stream block is truncated to same length as the final plaintext block:
`      CT[n] := PT[n] XOR TRUNC(AES(CTRBLK))`

As it is a simple XOR, the bits that are part of the encryption result will always have the correct result even if the TRUNCATION of the key stream didn't happen at bit level.

Hence I believe that if the current linux-generic implementation for EEA2 is fine if the API does not promise anything about bits that are not part of the encryption results.

asasidharan:
> 
> 
> > Within a byte, masking of the bits should be fine for any bit order is what I understand.
> 
> If the length is not a multiple of 8, then it is not stated if the remaining bits are the most significant or the least significant bits of the byte.

Now I realize your point. The remaining bits for the data is same as the most significant bits in the last byte.

If we consider the TRUNC function for the last block in AES-CTR, it returns the most significant bits of the block which is eventually XORed with the last plain text block.

But what would be the best place to state this? In the API spec or the validation test sources as a comment?

JannePeltonen:
I think we could remove the current Note2 from  odp_crypto_cipher_capability_t and add the following note in both odp_crypto_cipher_capability_t and odp_crypto_auth_capability_t:

Note2: If the data length is not a multiple of 8, the last input and output bytes will contain the input and output in the most significant bits, i.e. the most significant bit is considered to be the first bit of a byte for the purpose of input and output data range. The output bits that fall out of the output range are undefined.

Or feel free to suggest better wording if you like.

The API change should be a separate patch. You can add it in this PR, or if you prefer to have it in a separate PR or have it done by me, that too is fine with me. Let me know.

asasidharan:
Thanks you for the note and the suggestions. I will update the suggested API change along with this PR.

asasidharan:
> 
> Note2: If the data length is not a multiple of 8, the last input and output bytes will contain the input and output in the most significant bits, i.e. the most significant bit is considered to be the first bit of a byte for the purpose of input and output data range. The output bits that fall out of the output range are undefined.
> 
> Or feel free to suggest better wording if you like.

I have slightly updated the first part of the suggested Note2:

	 * Note2: If the data length is not a multiple of 8, the remaining
	 * bits of the data in the last byte of the input/output will be the
	 * most significant bits, i.e. the most significant bit is considered
	 * to be the first bit of a byte for the purpose of input and output
	 * data range. The output bits that fall out of the output range are
	 * undefined.

Let me know if this looks fine.

JannePeltonen:
That is fine.

asasidharan:
Added "Reviewed-by:" tag to the commits in V5.

JannePeltonen (approved):

psavol (approved):

asasidharan:
I have pushed a V6 for the commits with Petri's "Reviewed-by" tag.

But somehow the travis-ci is failing with the following error:
```
  Test: traffic_mngr_test_scheduler ...../../../../../../test/validation/api/traffic_mngr/traffic_mngr.c:2728:test_sched_queue_priority():pkts_sent=44 pkt_cnt=40 num_rcv_pkts=44 rcvd_in_order=37

FAILED
```

The error doesn't seem to be related to this PR.
@matiaselo Could you please help me in relaunching the travis-CI for this PR?

matiaselo:
> @matiaselo Could you please help me in relaunching the travis-CI for this PR?

Done

psavol:
We'll merge this soon (with other API PRs as v1.24.2).

asasidharan:
@matiaselo Thanks for relaunching.
@psavol Thank you for letting me know the plan.

PR #1027, opened 2020-09-07
[PATCH v1] api: pktio: add DMAC and l3_custom based flow hash

jerinjacobk:
Add DMAC and l3_custom based flow hash configuration.
When DMAC is selected, the implementation uses 6B destination
mac address from the packet to form the hash value.
When l3_custom is selected, implementation can select any of
the fields in the L3 layer to form the hash. This would enable,
custom protocols in the L3 layer from the hash.

### DIFF
--- a/include/odp/api/spec/packet_io.h
+++ b/include/odp/api/spec/packet_io.h
@@ -121,6 +121,10 @@ typedef union odp_pktin_hash_proto_t {
 		uint32_t ipv6_tcp : 1;
 		/** IPv6 addresses */
 		uint32_t ipv6     : 1;
+		/** Destination MAC address */
+		uint32_t dmac	  : 1;
+		/** Custom L3 fields */
+		uint32_t l3_custom: 1;
 	} proto;
 
 	/** All bits of the bit field structure */


include/odp/api/spec/packet_io.h
```diff
@@ -121,6 +121,10 @@ typedef union odp_pktin_hash_proto_t {
 		uint32_t ipv6_tcp : 1;
 		/** IPv6 addresses */
 		uint32_t ipv6     : 1;
+		/** Destination MAC address */
+		uint32_t dmac	  : 1;
+		/** Custom L3 fields */
```
JannePeltonen, line 126: This would require more explanation. You have some in the commit message but it is needed here. Custom sounds that there is some configuration hidden somewhere  to select the fields. Implementation selecting the fields it wishes sounds more like 'automatic' to me.

Can 'custom' be combined with other fields and what would it mean?

include/odp/api/spec/packet_io.h
```diff
@@ -121,6 +121,10 @@ typedef union odp_pktin_hash_proto_t {
 		uint32_t ipv6_tcp : 1;
 		/** IPv6 addresses */
 		uint32_t ipv6     : 1;
+		/** Destination MAC address */
+		uint32_t dmac	  : 1;
```
JannePeltonen, line 125: If the HW is capable of including destination MAC address and upper layer fields (like IP addresses, L4 ports) in the same hash, then that is what I might want to use to get even more flow entropy. But how could I express that? Having only the destination MAC address in the flow hash seems a bit limiting to me.

The current hash config looks a bit weird to me since, despite being a bit field, it tries to specify the hash strategies separately (e.g. ipv4_tcp) instead of as desired fields to include in the hash. So in keeping with that the dmac combined with other fields would give rise to quite many new combinations. Maybe one possibility would be to move dmac to a separate "layer" of the config so that the new "L2 layer hash config" and the current "L3+L4 layer hash config" would be combined.

I am not sure dmac is a good name. dst_mac would sound more clear to me.

Maybe there should be source MAC too.

jerinjacobk:
I am closing this now. I will reopen if this patch is an absolute requirement. Thanks for the comments.

PR #1029, opened 2020-09-07
[PATCH v3] test: fix tcp sample packet header

nithind1988:
test: fix tcp sample packet header

TCP sample packet header used by pktio_main has TCP
flags set to zero. Flags set to zero can be generally
considered illegal and treated as a bad packet by some HW.
RFC 793 also does not have any use case as such with all the
flags clear. Hence this patch sets SYN flag in the sample pkt.


JannePeltonen:
Setting the SYN flag is reasonable, but you have to update the TCP checksum fields too after altering the headers. Now the checksums are incorrect.

nithind1988:
> 
> 
> Setting the SYN flag is reasonable, but you have to update the TCP checksum fields too after altering the headers. Now the checksums are incorrect.

Ack. I missed that as pktio_main was not checking for checksum in test case of parser_test_ipv4_tcp(). Will send a V2 with the fix.

JannePeltonen:
Ok, now the checksums look correct to me. Please add reviewed-by tag.

JannePeltonen (approved):

PR #1034, opened 2020-09-11
[PATCH v2] example: debug: add new debug example application

psavol:
This application can be used to test various debug print
functions of ODP API. Debug functions to be called are
selected with command line options. E.g. -s option reserves
a SHM area and calls odp_shm_print() to print debug
information about it.

Signed-off-by: Petri Savolainen <petri.savolainen@nokia.com>

### DIFF
--- a/example/debug/odp_debug.c
+++ b/example/debug/odp_debug.c
@@ -1,0 +1,335 @@
+/* Copyright (c) 2020, Nokia
+ * All rights reserved.
+ *
+ * SPDX-License-Identifier:     BSD-3-Clause
+ */
+
+#include <stdio.h>
+#include <string.h>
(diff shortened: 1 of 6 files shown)


example/debug/odp_debug.c
```diff
@@ -0,0 +1,334 @@
...
+	int queue;
+
+} test_global_t;
+
+static test_global_t test_global;
+
+static void print_usage(void)
+{
+	printf("\n"
+	       "ODP debug example. Prints out debug information.\n"
```
matiaselo, line 26: No need to repeat "ODP debug example." here.

matiaselo (approved):
One minor comment.

PR #59, opened 2017-06-26
[PATCH v2] Master timer debug

muvarov:
do timer debug with no code change.

PR #1039, opened 2020-09-15
[PATCH v3] Ipsec validation test improvements

JannePeltonen:
This patch set adds IPsec output-input tests for many algorithm and algorithm combinations, with the goal that all algorithms supported by the IPsec API would be included. Currently some algorithms and key lenghts are not tested at all, even though apparently supported by the linux generic implementation.

The output-input tests perform IPsec output processing for a packet and then IPsec input processing to see if the original packet is recovered. This obviously does not test that the algorithms work correctly or even do anything at all but it hopefully improves test coverage a bit.


test/validation/api/ipsec/test_vectors.h
```diff
@@ -18,6 +18,12 @@ KEY(key_a5_128, 0xa5, 0xa5, 0xa5, 0xa5, 0xa5, 0xa5, 0xa5, 0xa5,
 		0xa5, 0xa5, 0xa5, 0xa5, 0xa5, 0xa5, 0xa5, 0xa5);
 KEY(key_5a_128, 0x5a, 0x5a, 0x5a, 0x5a, 0x5a, 0x5a, 0x5a, 0x5a,
 		0x5a, 0x5a, 0x5a, 0x5a, 0x5a, 0x5a, 0x5a, 0x5a);
+KEY(key_5a_160, 0x5a, 0x5a, 0x5a, 0x5a, 0x5a, 0x5a, 0x5a, 0x5a,
+		0x5a, 0x5a, 0x5a, 0x5a, 0x5a, 0x5a, 0x5a, 0x5a,
+		0x5a, 0x5a, 0x5a, 0x5a);
+KEY(key_a5_192, 0xa6, 0xa5, 0xa5, 0xa5, 0xa5, 0xa5, 0xa5, 0xa5,
+		0xa6, 0xa5, 0xa5, 0xa5, 0xa5, 0xa5, 0xa5, 0xa5,
+		0xa5, 0xa5, 0xa5, 0xa5, 0xa5, 0xa5, 0xa5, 0xa5);
```
psavol, line 26: CHECK: Alignment should match open parenthesis
#59: FILE: test/validation/api/ipsec/test_vectors.h:22:
+KEY(key_5a_160, 0x5a, 0x5a, 0x5a, 0x5a, 0x5a, 0x5a, 0x5a, 0x5a,
+		0x5a, 0x5a, 0x5a, 0x5a, 0x5a, 0x5a, 0x5a, 0x5a,

Checkpatch warnings could be fixed like this:
KEY(key_5a_160,
      0x5a, 0x5a, 0x5a, 0x5a, 0x5a, 0x5a, 0x5a, 0x5a,
      0x5a, 0x5a, 0x5a, 0x5a, 0x5a, 0x5a, 0x5a, 0x5a,

psavol:
General comments:
- Typo in commit log subject "validation: ipsec: test many alorithms in single out-in test"
- Update copyright of all the files as changes are non-trivial

test/validation/api/ipsec/test_vectors.h
JannePeltonen: I actually tried that and it did not work: "WARNING: please, no spaces at the start of a line".

If KEY was at least 7 characters, it would work (I could "#define KEY_DEF DEF" and use that to get a tab in front).

In the end I decided to just ignore checkpatch in this case and use the same formatting as the existing definitions.

test/validation/api/ipsec/ipsec_test_out.c
```diff
@@ -1046,5 +1088,6 @@ odp_testinfo_t ipsec_out_suite[] = {
 				  ipsec_check_esp_null_sha256),
 	ODP_TEST_INFO_CONDITIONAL(test_out_ipv4_udp_esp_null_sha256,
 				  ipsec_check_esp_null_sha256),
+	ODP_TEST_INFO(test_esp_out_in_all),
```
psavol, line 1217: I'd prefer having individual test cases as it's easier to control which ones to run (and in which order) while developing an implementation (or debugging a failing test case). Some times a bug (e.g. memory leak) in a previous test case may cause a latter test case to fail.

JannePeltonen:
v2: Updated the copyrights and fixed the typo.

As we discussed, I did not change the combined test case nor did I try to stop checkpatch complaining about the added key definitions,

test/validation/api/ipsec/test_vectors.h
psavol: ok, all KEY macros could be renamed later to match 7 char length.

psavol (approved):

PR #1042, opened 2020-09-16
[PATCH v3] IPsec example improvements

matiaselo:
V2:

- Improve README intro chapter (Janne)

- Fail `create_sp_db_entry ()` if used with "both" option from ipsec_api example (Janne)

V3:
- Removed comment that tunnel mode is not supported from README


JannePeltonen:
The app still accepts "both" (in create_sp_db_entry()), which would then make the example misbehave.

example/ipsec_api/README
```diff
@@ -0,0 +1,154 @@
+Copyright (c) 2014-2018, Linaro Limited
+Copyright (c) 2020, Nokia
+All rights reserved.
+
+SPDX-License-Identifier:        BSD-3-Clause
+
+1. Intro
+
+The IPsec API example application "odp_ipsec_api" functions as a simple L3 IPv4
+router which supports IPsec 3DES cipher and HMAC-MD5 authentication in both the
```
JannePeltonen, line 10: Yes, 3DES and HMAC-MD5 are supported but not at the same time as one might expect. It might be useful to state what is supported:
- ESP with 3DES or NULL encryption algorithm, with NULL authentication
- AH with HMAC-MD5 or HMAC-SHA-256 integrity algorithm

I get the feeling that the original author(s) may have confused AH and the authentication algorithm of ESP, resulting in a bit weird examples that attempt to support combined ESP+AH but not ESP with both confidentiality and integrity.

example/ipsec_api/README
```diff
@@ -0,0 +1,154 @@
...
+transmit and receive directions. Note that only IPsec "transport" mode is
```
JannePeltonen, line 11: Tunnel mode appears to be supported too.

matiaselo: Yes it appears so, but I didn't want to mention it without verifying that it actually works. The tunnel mode argument `-t` is missing also from the usage print.

example/ipsec_api/README
matiaselo: Okay, will add this is V2.

matiaselo:
@JannePeltonen All comments have been fixed. Now OK to merge?

JannePeltonen:
Looks ok.

JannePeltonen (approved):

PR #1043, opened 2020-09-18
[PATCH v5] validation: cls: fix error packet generation

kkirankk:
TCP sample packet header used by classification has TCP
flags set to zero and IP checksum is not being calculated
properly in some cases. Flags set to zero can be generally
considered illegal and treated as a bad packet by some HW.
RFC 793 also does not have any use case as such with all the
flags clear. Hence this patch sets ACK flag in the sample pkt.

### DIFF
--- a/test/validation/api/classification/odp_classification_common.c
+++ b/test/validation/api/classification/odp_classification_common.c
@@ -316,11 +316,13 @@ odp_packet_t create_packet(cls_packet_info_t pkt_info)
 		ip->src_addr = odp_cpu_to_be_32(addr);
 		ip->ver_ihl = ODPH_IPV4 << 4 | ODPH_IPV4HDR_IHL_MIN;
 		ip->id = odp_cpu_to_be_16(seqno);
-		odph_ipv4_csum_update(pkt);
 		ip->proto = next_hdr;
 		ip->tot_len = odp_cpu_to_be_16(l3_len);
 		ip->ttl = DEFAULT_TTL;
+		ip->frag_offset = 0;
+		ip->tos = 0;
 		odp_packet_has_ipv4_set(pkt, 1);
+		odph_ipv4_csum_update(pkt);
 	} else {
 		/* ipv6 */
 		odp_packet_has_ipv6_set(pkt, 1);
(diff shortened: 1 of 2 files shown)


test/validation/api/classification/odp_classification_common.c
```diff
@@ -316,11 +316,12 @@ odp_packet_t create_packet(cls_packet_info_t pkt_info)
 		ip->src_addr = odp_cpu_to_be_32(addr);
 		ip->ver_ihl = ODPH_IPV4 << 4 | ODPH_IPV4HDR_IHL_MIN;
 		ip->id = odp_cpu_to_be_16(seqno);
-		odph_ipv4_csum_update(pkt);
 		ip->proto = next_hdr;
 		ip->tot_len = odp_cpu_to_be_16(l3_len);
 		ip->ttl = DEFAULT_TTL;
+		ip->frag_offset = 0;
```
JannePeltonen, line 322: Could you also initialize the tos field so that it is not left to some random value.

test/validation/api/classification/odp_classification_common.c
```diff
@@ -358,6 +359,9 @@ odp_packet_t create_packet(cls_packet_info_t pkt_info)
 		tcp->src_port = odp_cpu_to_be_16(CLS_DEFAULT_SPORT);
 		tcp->dst_port = odp_cpu_to_be_16(CLS_DEFAULT_DPORT);
 		tcp->hl = ODPH_TCPHDR_LEN / 4;
+		tcp->ack = 1;
+		tcp->fin = 0;
+		tcp->syn = 0;
```
JannePeltonen, line 364: Could you please also initialize the other currently uninitialized fields of the TCP header. It is still possible to get an invalid TCP header out of this if the other fields (e.g. reserved bits that should be zero) happen to get certain values.

JannePeltonen:
After the latest patch I think all the TCP packets are well formed. But as I already commented, I would prefer to have all the currently uninitialized fields to be initialized to some known values (that includes seq_no, ack_no, window and urgptr).

JannePeltonen:
You could also fix the commit subject line. This patch is not fixing the generation of "error packets" as the current subject line says. Just drop "error" from the subject and it becomes ok.

JannePeltonen:

JannePeltonen (approved):

PR #1044, opened 2020-09-21
[PATCH v1] api: new cls terms

jerinjacobk:
Pull request to add a few more terms to leverge HW classfication support.

## ap: cls: introduce IGMP group addr term

Introduce IGMP group address term.

## api: cls: introduce ICMP terms

Introduce ICMP identifier and type terms.

## api: cls: introduce SCTP terms

Introduce SCTP source and destination port terms.

## api: cls: introduce GTPU tunnel endpoint id term

Introduce GTPU tunnel endpoint identifier term.

### DIFF
--- a/include/odp/api/spec/classification.h
+++ b/include/odp/api/spec/classification.h
@@ -109,6 +109,10 @@ typedef union odp_cls_pmr_terms_t {
 		uint64_t	icmp_id:1;
 		/** ICMP type */
 		uint64_t	icmp_type:1;
+		/** Source SCTP port */
+		uint64_t	sctp_sport:1;
+		/** Destination SCTP port */
+		uint64_t	sctp_dport:1;
 
 	} bit;
 	/** All bits of the bit field structure */
@@ -562,6 +566,12 @@ typedef enum {
 
 	/** ICMP type (val_sz = 2) */
 	ODP_PMR_ICMP_TYPE,
+
+	/** Source SCTP port (val_sz = 2) */
+	ODP_PMR_SCTP_SPORT,
+
+	/** Destination SCTP port (val_sz = 2) */
+	ODP_PMR_SCTP_DPORT,
 
 	/** Inner header may repeat above values with this offset */
 	ODP_PMR_INNER_HDR_OFF = 32
--- a/include/odp/api/spec/classification.h
+++ b/include/odp/api/spec/classification.h
@@ -113,6 +113,8 @@ typedef union odp_cls_pmr_terms_t {
 		uint64_t	sctp_sport:1;
 		/** Destination SCTP port */
 		uint64_t	sctp_dport:1;
+		/** GTPU Tunnel endpoint identifier */
+		uint64_t	gtpu_teid:1;
 
 	} bit;
 	/** All bits of the bit field structure */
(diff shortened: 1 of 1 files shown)


include/odp/api/spec/classification.h
```diff
@@ -105,6 +105,10 @@ typedef union odp_cls_pmr_terms_t {
 		uint64_t        custom_l3:1;
 		/** IGMP Group address */
 		uint64_t	igmp_grp_addr:1;
+		/** ICMP identifier */
+		uint64_t	icmp_id:1;
```
JannePeltonen, line 109: I wonder if this is useful or just API bloat, i.e. are there use cases where it is necessary to be able to separate ICMP messages of the same type based on the identification value already in the classifier and not later in the SW? I also wonder why the identification field is important enough to get added here but other fields, like ICMP code, are not?

One can also use custom PMR terms to match things that are not explicitly supported, so maybe every possible protocol and field does not need to be added.

include/odp/api/spec/classification.h
```diff
@@ -573,6 +575,9 @@ typedef enum {
 	/** Destination SCTP port (val_sz = 2) */
 	ODP_PMR_SCTP_DPORT,
 
+	/** GTPU Tunnel endpoint identifier  (val_sz = 4)*/
+	ODP_PMR_GTPU_TEID,
```
JannePeltonen, line 579: I wonder how GTP TEID matching really works. To my understanding (but I may be wrong) the existing terms, like ODP_PMR_TCP_DPORT, match if ODP has determined that the packet is of the right protocol (TCP) and that the protocol specific matching criteria (port number matches) is satisfied.

In case of GTP, ODP cannot distinguish whether a received UDP packet is GTP or something else since the interpretation is up to the application. Thus the GTP TEID term requires better, unambiguous definition of when exactly it matches.

If the GTP TEID PMR term just looks at UDP packets at certain offset, then does it bring much value over the existing custom PMR term (though perhaps custom L4 term would be more convenient).

include/odp/api/spec/classification.h
```diff
@@ -103,6 +103,18 @@ typedef union odp_cls_pmr_terms_t {
...
+		/** ICMP identifier */
+		uint64_t	icmp_id:1;
+		/** ICMP type */
+		uint64_t	icmp_type:1;
+		/** Source SCTP port */
+		uint64_t	sctp_sport:1;
+		/** Destination SCTP port */
+		uint64_t	sctp_dport:1;
+		/** GTPU Tunnel endpoint identifier */
+		uint64_t	gtpu_teid:1;
```
psavol, line 117: What UDP port number is matched as GTPU?

include/odp/api/spec/classification.h
jerinjacobk: We will send the next version with the definition of GTPU in the comment.

include/odp/api/spec/classification.h
jerinjacobk: We will send the next version with the definition of GTPU in the comment.

include/odp/api/spec/classification.h
jerinjacobk: I think, we can add new terms based on the need.

jerinjacobk:
Kiran will be resending this series, addressing the comments and validation and helper functions for those new TERMS. I am closing here so that Kiran can start a new branch for the same,

PR #1045, opened 2020-09-21
[PATCH v7] api: pktio: change Tx checksum insertion API semantics 

nithind1988:
api: pktio: change Tx checksum insertion API semantics

Existing API spec requires application to just provide L3 offset and
and L4 offset and expects platform either in SW or HW parse the packet in
Tx and identify L3 and L4 protocols and then generate checksum if
requested. Since not all platforms have support for packet parsing in HW
in Tx, they depend on software to provide L3 and L4 types in addition to
existing L3 and L4 protocol offsets. Doing the parsing in platform SW for
such cases brings down performance and is counter productive.
Since similar to L3 and L4 offsets, even L3 and L4 proto types are provided
by HW parser in packet ingress, making use of the same information in Tx is
benefitial for performance.

So this patch changes Tx checksum insertion semantics by mandating that
application should have valid packet metadata for L3 and L4 type
along with existing requirement of L3 and L4 offsets, and platform will
make use of the same information for checksum calculation and insertion.

doc: userguide: change checksum insertion API documentation

Change checksum insertion API documentation with updated
API semantics.

validation: pktio: update l3 and l4 type of created pkts

For checksum insertion in pktio Tx, as per new API spec,
application needs to keep L3 and L4 type in packet metadata
valid reflecting actual packet content. This patch updates
the same in packets created for Tx checksum insertion validation
test case.

### DIFF
--- a/include/odp/api/spec/packet_io.h
+++ b/include/odp/api/spec/packet_io.h
@@ -365,14 +365,16 @@ typedef union odp_pktin_config_opt_t {
  * These defaults may be overridden on per packet basis using e.g.
  * odp_packet_l4_chksum_insert().
  *
- * For correct operation, packet metadata must provide valid offsets for the
- * appropriate protocols. For example, UDP checksum calculation needs both L3
- * and L4 offsets (to access IP and UDP headers). When application
- * (e.g. a switch) does not modify L3/L4 data and thus checksum does not need
- * to be updated, checksum insertion should be disabled for optimal performance.
- *
- * Packet flags (odp_packet_has_*()) are ignored for the purpose of checksum
- * insertion in packet output.
+ * For correct operation, packet metadata must provide valid offsets and type
+ * flags for the appropriate layer 3 and layer 4 protocols.  L3 and L4 offsets
+ * can be updated with odp_packet_l3_offset_set() and odp_packet_l4_offset_set()
+ * calls. L3 and L4 type flags can be updated using odp_packet_has_*_set() calls
+ * For example, UDP checksum calculation needs both L3 and L4 types (IP and UDP) and
+ * L3 and L4 offsets (to access IP and UDP headers), while IP checksum
+ * calculation only needs L3 type (IP) and L3 offset (to access IP header).
+ * When application (e.g. a switch) does not modify L3/L4 data and thus checksum
+ * does not need to be updated, checksum insertion should be disabled for optimal
+ * performance.
  *
  * UDP, TCP and SCTP checksum insertion must not be requested for IP fragments.
  * Use checksum override function (odp_packet_l4_chksum_insert()) to disable
--- a/test/validation/api/pktio/pktio.c
+++ b/test/validation/api/pktio/pktio.c
@@ -2211,6 +2211,12 @@ static void pktio_test_chksum(void (*config_fn)(odp_pktio_t, odp_pktio_t),
 		return;
 	}
 
+	/* Provide L3 and L4 proto for pktout HW checksum generation */
+	for (i = 0; i < TX_BATCH_LEN; i++) {
+		odp_packet_has_ipv4_set(pkt_tbl[i], true);
+		odp_packet_has_udp_set(pkt_tbl[i], true);
+	}
+
 	ret = odp_pktout_queue(pktio_tx, &pktout_queue, 1);
 	CU_ASSERT_FATAL(ret > 0);
 
@@ -2278,6 +2284,12 @@ static void pktio_test_chksum_sctp(void (*config_fn)(odp_pktio_t, odp_pktio_t),
 			CU_ASSERT_FATAL(odp_pktio_close(pktio[i]) == 0);
 		}
 		return;
+	}
+
+	/* Provide L3 and L4 proto for pktout HW checksum generation */
+	for (i = 0; i < TX_BATCH_LEN; i++) {
+		odp_packet_has_ipv4_set(pkt_tbl[i], true);
+		odp_packet_has_sctp_set(pkt_tbl[i], true);
 	}
 
 	ret = odp_pktout_queue(pktio_tx, &pktout_queue, 1);
--- a/doc/users-guide/users-guide-pktio.adoc
+++ b/doc/users-guide/users-guide-pktio.adoc
@@ -371,14 +371,16 @@ transmitted:
  * These defaults may be overridden on per packet basis using e.g.
  * odp_packet_l4_chksum_insert().
  *
- * For correct operation, packet metadata must provide valid offsets for the
- * appropriate protocols. For example, UDP checksum calculation needs both L3
- * and L4 offsets (to access IP and UDP headers). When application
- * (e.g. a switch) does not modify L3/L4 data and thus checksum does not need
- * to be updated, checksum insertion should be disabled for optimal performance.
- *
- * Packet flags (odp_packet_has_*()) are ignored for the purpose of checksum
- * insertion in packet output.
+ * For correct operation, packet metadata must provide valid offsets and type
+ * flags for the appropriate layer 3 and layer 4 protocols.  L3 and L4 offsets
+ * can be updated with odp_packet_l3_offset_set() and odp_packet_l4_offset_set()
+ * calls. L3 and L4 type flags can be updated using odp_packet_has_*_set() calls
+ * For example, UDP checksum calculation needs both L3 and L4 types (IP and UDP) and
+ * L3 and L4 offsets (to access IP and UDP headers), while IP checksum
+ * calculation only needs L3 type (IP) and L3 offset (to access IP header).
+ * When application (e.g. a switch) does not modify L3/L4 data and thus checksum
+ * does not need to be updated, checksum insertion should be disabled for optimal
+ * performance.
  *
  * UDP, TCP and SCTP checksum insertion must not be requested for IP fragments.
  * Use checksum override function (odp_packet_l4_chksum_insert()) to disable
--- a/example/generator/odp_generator.c
+++ b/example/generator/odp_generator.c
@@ -405,8 +405,10 @@ static int setup_udp_pkt(odp_packet_t pkt, odp_pktout_config_opt_t *pktout_cfg,
 	if (pktout_cfg->bit.ipv4_chksum || pktout_cfg->bit.udp_chksum) {
 		odp_packet_l2_offset_set(pkt, 0);
 		odp_packet_l3_offset_set(pkt, ODPH_ETHHDR_LEN);
+		odp_packet_has_ipv4_set(pkt, 1);
 		odp_packet_l4_offset_set(pkt, ODPH_ETHHDR_LEN +
 					 ODPH_IPV4HDR_LEN);
+		odp_packet_has_udp_set(pkt, 1);
 	}
 	return 0;
 }
@@ -518,6 +520,7 @@ static int setup_icmp_pkt(odp_packet_t pkt,
 	if (pktout_cfg->bit.ipv4_chksum) {
 		odp_packet_l2_offset_set(pkt, 0);
 		odp_packet_l3_offset_set(pkt, ODPH_ETHHDR_LEN);
+		odp_packet_has_ipv4_set(pkt, 1);
 		odp_packet_l4_offset_set(pkt, ODPH_ETHHDR_LEN +
 					 ODPH_IPV4HDR_LEN);
 	}
--- a/include/odp/api/spec/packet.h
+++ b/include/odp/api/spec/packet.h
@@ -1830,9 +1830,14 @@ odp_packet_chksum_status_t odp_packet_l4_chksum_status(odp_packet_t pkt);
  * inserted if the packet is output through a pktio that does not have
  * the relevant checksum insertion enabled.
  *
+ * L3 type and L3 offset in packet metadata should provide valid protocol
+ * and header offset for checksum insertion purposes.
+ *
  * @param pkt     Packet handle
  * @param insert  0: do not insert L3 checksum
  *                1: insert L3 checksum
+ *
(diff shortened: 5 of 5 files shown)


JannePeltonen:
The API spec says that packet flags are ignored for the purpose of checksum insertion. If the current validation test fails, it just shows that the implementation makes too many assumptions no the packet flags.

I think the implementation can assume that L3 offset always points to IP header when checksum insertion is requested. The implementation can then check the IP protocol version from the IP version field and after that it can figure out the L4 protocol too.

nithind1988:
> 
> 
> The API spec says that packet flags are ignored for the purpose of checksum insertion. If the current validation test fails, it just shows that the implementation makes too many assumptions no the packet flags.
> 
> I think the implementation can assume that L3 offset always points to IP header when checksum insertion is requested. The implementation can then check the IP protocol version from the IP version field and after that it can figure out the L4 protocol too.

But then in default checksum insert case, user will not call odp_packet_l3|l4_chksum_insert(). So we will not have a chance to detect L3 type and hence l4 type right ? 

# By making user explicitly say what is the l3 type and l4 type for packets created by him, we can have a clear assurance from user that this is so and so packet. 
# This will also help in cases where after receiving a packet, user changes packet content to a different l4 type before pktout. 
I can send a spec change to update the text if you think it is fine.

JannePeltonen:
You just have to assume that L3 type is IPv4 or IPv6 when checksum insertion has been requested for a packet.

nithind1988:
We did it this way to avoid a performance hit in fast path in forwarding cases and we thought it will be the same in other platforms too. Our HW needs to know the exact l3/l4 types along with offsets in order to perform checksum insertion.

As per current spec,
- Platform will be using odp_packet_l3_offset() and odp_packet_l4_offset() which can either be the ones that came in from packet ingress or, user has explicitly overridden values.
- Also the documentation says that it is user who has to make sure those offsets are right but doesn't always say they need to be always overridden with new values as after packet ingress, they are based on packet ingress parsing.

So if we can also get l3/l4 types also based on packet ingress parsing or explicitly overridden by user, we can avoid performance hit that comes by manual parsing in platform implementation.

For forwarding cases, since user only modifies minor data like, IPv4 TTL, TCP/UDP ports for NAT etc, he will not need to call 
odp_packet_has_*_set() API's. This is the same behavior with l3/l4 offset set API's.

Another performance disadvantage of parsing in SW to identify L4 proto is in case of IPv6. The L4 proto is not at fixed or directly computable offset and we have to walk through every extension header in a Loop for every packet to identify L4 protocol.

Other platforms that can parse the packet in egress in HW can ignore the flags provided by user and there is no change in that.

JannePeltonen:
ODP implementations must adhere to the API specification. Thus, as long as the API specification says that the packet flags are ignored, applications must be able to trust that implementations indeed ignore the packet flags and rely only on the L3 and L4 offsets. The current validation test is according to the API spec and adding the packet flag setting in it would not be a fix but counterproductive as it would allow previously detected non-compliant implementations pass the test more easily.

So what you really seem to want is an API change for potentially better performance (not for functionality). If such an API change is made (perhaps you might want to create a PR about it) then the validation check will need to be updated. But the API definition cannot be changed by just changing the validation test.

I can see that in IPv6 case where L4 checksum is requested the current API requires implementations to parse through the set of possible extension headers, which is not very nice. Maybe there is room for improvement in the API or maybe the performance advantage is not big enough to justify the change?

Note that requiring the has_ipv4 etc flags to be "properly" set may not be as clean and unambiguous as it sounds. The current API, to my understanding, allows checksum offloading in various tunneling cases (e.g. TCP inside VxLAN or IPv6 inside IPv4) as the application just needs to make the offsets point to the right headers. In those cases the "has_ipv4," "has_tcp" etc flags would need to be about those headers and not the outmost headers of the packet.

Btw, I looked at Linux generic ODP and noticed that it gives up IPv6 parsing (won't find L4 protocol) if there are extension headers. So even the reference implementation is not fully compliant to the API spec.

nithind1988:
> 
> 
> ODP implementations must adhere to the API specification. Thus, as long as the API specification says that the packet flags are ignored, applications must be able to trust that implementations indeed ignore the packet flags and rely only on the L3 and L4 offsets. The current validation test is according to the API spec and adding the packet flag setting in it would not be a fix but counterproductive as it would allow previously detected non-compliant implementations pass the test more easily.
> 
> So what you really seem to want is an API change for potentially better performance (not for functionality). If such an API change is made (perhaps you might want to create a PR about it) then the validation check will need to be updated. But the API definition cannot be changed by just changing the validation test.
> 
> I can see that in IPv6 case where L4 checksum is requested the current API requires implementations to parse through the set of possible extension headers, which is not very nice. Maybe there is room for improvement in the API or maybe the performance advantage is not big enough to justify the change?
> 

- In CN96xx, I see a ~51% drop in performance for IPv6 packet with one extension header(IPv6+Hop-by-Hop+UDP) tested using test/performance/l2fwd with checksum enabled option. 
- Though our HW parses packet and provides l3/l4 types along with l3/l4 offset, 
with current spec and below logic of recomputing l3/l4 types from packet content, I see lot of performance lost.

```
void
odp_packet_l4_chksum_insert(odp_packet_t pkt, int insert)
{
        PLT_ASSERT_FP(odp_packet_is_valid(pkt) != 0);

#ifdef USER_SET_L3L4_TYPE
        pkt->w5.l4_chksum_ovr = 1;
        pkt->w5.l4_chksum = !!insert;
#else
        if (insert) {
                uint8_t *l3_hdr = ((uint8_t *)pkt->w[1] + pkt->w4.lcptr);
                uint8_t l4_proto = 0;

                if ((*l3_hdr & 0xF0) == 0x40) {
                        _odp_ipv4hdr_t *hdr = (_odp_ipv4hdr_t *)l3_hdr;

                        pkt->w0.lctype = NPC_LT_LC_IP;
                        l4_proto = hdr->proto; 
                } else if ((*l3_hdr & 0xF0) == 0x60) {
                        _odp_ipv6hdr_t *hdr = (_odp_ipv6hdr_t *)l3_hdr;
                        uint8_t next_hdr;
                        uint8_t *hdr_ext;

                        pkt->w0.lctype = NPC_LT_LC_IP6;

                        next_hdr = hdr->next_hdr; 
                        hdr_ext = (uint8_t *)(hdr + 1);

                        while (next_hdr != _ODP_IPPROTO_NONH) {
                                _odp_ipv6hdr_ext_t *ext = (_odp_ipv6hdr_ext_t *)hdr_ext;

                                switch (next_hdr) {
                                case _ODP_IPPROTO_HOPOPTS:
                                case _ODP_IPPROTO_ROUTE:
                                case _ODP_IPPROTO_FRAG:
                                        next_hdr = ext->next_hdr; 
                                        hdr_ext += ((ext->ext_len + 1)<< 3);
                                        break;
                                default:
                                        /* Break then extension walk */
                                        l4_proto = ext->next_hdr;
                                        next_hdr = _ODP_IPPROTO_NONH;
                                        break;
                                }
                        }

                }

                if (l4_proto == _ODP_IPPROTO_TCP)
                        pkt->w0.ldtype = NPC_LT_LD_TCP;
                else if (l4_proto == _ODP_IPPROTO_UDP)
                        pkt->w0.ldtype = NPC_LT_LD_UDP;
                else if (l4_proto == _ODP_IPPROTO_SCTP)
                        pkt->w0.ldtype = NPC_LT_LD_SCTP;
        }

        pkt->w5.l4_chksum_ovr = 1;
        pkt->w5.l4_chksum = !!insert;
#endif
```



> Note that requiring the has_ipv4 etc flags to be "properly" set may not be as clean and unambiguous as it sounds. The current API, to my understanding, allows checksum offloading in various tunneling cases (e.g. TCP inside VxLAN or IPv6 inside IPv4) as the application just needs to make the offsets point to the right headers. In those cases the "has_ipv4," "has_tcp" etc flags would need to be about those headers and not the outmost headers of the packet.
> 
Agreed. I can send a PR for API change including addition of new API's to handle tunneling scenario's like below that enhances the spec. We can continue the discussion in the new PR if it makes sense.

- Packet flag api's for inner headers.
```
odp_packet_inner_l3_chksum_insert()
odp_packet_inner_l4_chksum_insert()
odp_packet_has_inner_tcp_set()
odp_packet_has_inner_udp_set()
odp_packet_has_inner_sctp_set()
odp_packet_has_inner_ipv4_set()
odp_packet_has_inner_ipv6_set()
odp_packet_has_inner_tcp()
odp_packet_has_inner_udp()
odp_packet_has_inner_sctp()

```
- Modifications to odp_pktio_parser_config_t to include Inner L2, Inner l3, Inner l4 etc to allow parsing inner headers, etc.

This will also be needed in the example of VXLAN and if we want HW to compute bother inner TCP and outer UDP checksum but have no way currently to provide inner and outer l3/l4 offsets.


> Btw, I looked at Linux generic ODP and noticed that it gives up IPv6 parsing (won't find L4 protocol) if there are extension headers. So even the reference implementation is not fully compliant to the API spec.

JannePeltonen:
> odp_packet_l4_chksum_insert(odp_packet_t pkt, int insert)

Note that this function cannot actually calculate the checksum yet (like the attached code does) but merely mark that the checksum needs to be calculated and inserted when the packet is sent out. After the applications calls this function and before it sends the packet out, it may still modify the packet content and change the offsets.

test/validation/api/pktio/pktio.c
```diff
@@ -269,13 +269,15 @@ static void pktio_init_packet_eth_ipv4(odp_packet_t pkt, uint8_t proto)
 
 	/* Ethernet */
 	odp_packet_l2_offset_set(pkt, 0);
+	odp_packet_has_eth_set(pkt, true);
```
psavol, line 272: This is not required by the API spec. Application is free to set it and not set it. Maybe validation test should not set it (do only minimum settings), so that implementations cannot depend on it.

psavol:
In addition to validation tests, all our test applications that use checksum offload need to be updated.

nithind1988:
> 
> 
> In addition to validation tests, all our test applications that use checksum offload need to be updated.

Ack, Currently it is only used in test/performance/l2fwd and test/validation/api/pktio and examples/generator.
For l2fwd, there is no need to update as it currently doesn't update l3 offset(uses existing value) and hence l3 type is also same.

For generator, I'll send a next version.

test/validation/api/pktio/pktio.c
nithind1988: Ack.

include/odp/api/spec/packet_io.h
```diff
@@ -365,14 +365,16 @@ typedef union odp_pktin_config_opt_t {
...
- * to be updated, checksum insertion should be disabled for optimal performance.
- *
- * Packet flags (odp_packet_has_*()) are ignored for the purpose of checksum
- * insertion in packet output.
+ * For correct operation, packet metadata must provide valid offsets and type
+ * flags for the appropriate layer 3 and layer 4 protocols.  L3 and L4 offsets
+ * can be updated with odp_packet_l3_offset_set() and odp_packet_l4_offset_set()
+ * calls. L3 and L4 type flags can be updated using odp_packet_has_*_set() calls
+ * For example, UDP checksum calculation needs both L3 and L4 types (IP and UDP) and
+ * L3 and L4 offsets (to access IP and UDP headers). While IP checksum
```
JannePeltonen, line 373: The sentence starting with "While" could be fixed grammatically. Either by combining it with the previous sentence: "... headers), while IP checksum..." or by just removing "While".

JannePeltonen:
Looks ok to me (I made a very minor grammar comment that could still be fixed), but I would prefer the patches to be ordered so that the example app changes come first and the API change after those. Then all revisions of ODP stay internally consistent.

test/validation/api/pktio/pktio.c
```diff
@@ -276,6 +276,7 @@ static void pktio_init_packet_eth_ipv4(odp_packet_t pkt, uint8_t proto)
 
 	/* IP */
 	odp_packet_l3_offset_set(pkt, ODPH_ETHHDR_LEN);
+	odp_packet_has_ipv4_set(pkt, true);
```
JannePeltonen, line 279: I think setting of the metadata required by TX checksumming should be done only for the test cases that do TX checksumming and not globally for all tests. If TX checksum is not requested, the L3 & L4 type metadata should not be needed by the ODP implementation being tested, so adding the metadata in that case can reduce the effectiveness of the API validation test.

nithind1988: PKT metadata like l3 offset, l4 offset  are already used for checksum fixing up (in non check sum test cases) and also in 
pktio_pkt_set_seq() sequence number. Now it is just in addition to existing metadata we are also updating L3 and L4 type meta data. Since even l3 and l4 offsets are also not needed in non Tx HW checksum generation test cases, just removing L3 and L4 type from common packet initialization is also not full solution if we want to test without pkt metadata.

JannePeltonen: Even if the offsets are now set even when they are not strictly needed for the ODP API under test, we do not need to do the same for the packet types. As far as I can see it would be sufficient to set the types in pktio_test_chksum() and pktio_test_chksum_sctp() instead of the generic packet creation functions to restrict the type metadata only to the checksum tests. So why not do it that way?

nithind1988: Ok. If setting them outside is fine, then I'm fine. I was only thinking it might be out of context if we don't set it at packet creation. Other way is propagate fix_cs flag upto packet creation and then set it and clear it based on whether we are doing HW checksum offload or not. This second approach would involve changes at lot of places as it is generic packet creation code.

JannePeltonen: There is no need to set the flags in packet creation and unnecessarily complicate the code. The flags are needed only by checksum offload, so they are related to checksumming in the context of this application and are thus not out of place in the checksum test functions.

JannePeltonen (approved):

PR #1046, opened 2020-09-22
[PATCH v3] helper: fix compiler warning in chksum.c

xvikas:
compiler warning:
warning: taking address of packed member of
‘struct <anonymous>’ may result in an unaligned
pointer value

Fix:
odph_udphdr_t, odph_tcphdr_t are defined as packed
structs and when we try to access the member of the
packed struct using a typecasted pointer, gcc10 reports
warning. To fix the issue, use the l4 pointer and add
offset of the checksum

Signed-off-by: Vikas Kumar Singh <svikaskumar@marvell.com>

### DIFF
--- a/helper/chksum.c
+++ b/helper/chksum.c
@@ -98,14 +98,13 @@ static inline int odph_process_l4_hdr(odp_packet_t      odp_pkt,
 				      uint32_t         *l4_len_ptr,
 				      odp_bool_t       *split_l4_hdr_ptr,
 				      odp_bool_t       *is_tcp_ptr,
-				      uint32_t         *pkt_chksum_offset_ptr,
-				      uint16_t        **pkt_chksum_ptr_ptr)
+				      uint32_t         *pkt_chksum_offset_ptr)
 {
-	odph_udphdr_t  *udp_hdr_ptr;
-	odph_tcphdr_t  *tcp_hdr_ptr;
+	odph_udphdr_t  *udp_hdr_ptr = NULL;
+	odph_tcphdr_t  *tcp_hdr_ptr = NULL;
 	odp_bool_t      split_l4_hdr, is_tcp;
 	uint32_t        l4_offset, l4_len, pkt_chksum_offset;
-	uint16_t       *pkt_chksum_ptr;
+	uint16_t        pkt_chksum;
 	uint8_t        *l4_ptr;
 	uint32_t hdr_len = 0;
 
@@ -130,7 +129,7 @@ static inline int odph_process_l4_hdr(odp_packet_t      odp_pkt,
 		 * should come from the udp header, unlike for TCP where is
 		 * derived. */
 		l4_len            = odp_be_to_cpu_16(udp_hdr_ptr->length);
-		pkt_chksum_ptr    = (uint16_t *)(void *)&udp_hdr_ptr->chksum;
+		pkt_chksum        = udp_hdr_ptr->chksum;
 		pkt_chksum_offset = l4_offset + offsetof(odph_udphdr_t, chksum);
 	} else if (odp_packet_has_tcp(odp_pkt)) {
 		tcp_hdr_ptr  = (odph_tcphdr_t *)l4_ptr;
@@ -141,7 +140,7 @@ static inline int odph_process_l4_hdr(odp_packet_t      odp_pkt,
 					       ODPH_TCPHDR_LEN, tcp_hdr_ptr);
 		}
 
-		pkt_chksum_ptr    = (uint16_t *)(void *)&tcp_hdr_ptr->cksm;
+		pkt_chksum        = tcp_hdr_ptr->cksm;
 		pkt_chksum_offset = l4_offset + offsetof(odph_tcphdr_t, cksm);
 		is_tcp            = true;
 	} else {
@@ -151,25 +150,24 @@ static inline int odph_process_l4_hdr(odp_packet_t      odp_pkt,
 	/* Note that if the op is ODPH_CHKSUM_VERIFY and the existing
 	 * chksum field is 0 and this is a UDP pkt and the chksum_ptr is NULL
 	 * then skip the rest of the chksum calculation, returning 1 instead. */
-	if ((op == ODPH_CHKSUM_VERIFY) && (*pkt_chksum_ptr == 0) &&
+	if ((op == ODPH_CHKSUM_VERIFY) && (pkt_chksum == 0) &&
 	    (!is_tcp) && (chksum_ptr == NULL))
 		return 1;
 
 	/* If we are doing a ODPH_CHKSUM_GENERATE op, then make sure that the
 	 * existing chksum field has been set to zeros. */
-	if ((op == ODPH_CHKSUM_GENERATE) && (*pkt_chksum_ptr != 0)) {
+	if ((op == ODPH_CHKSUM_GENERATE) && (pkt_chksum != 0)) {
 		if (split_l4_hdr)
 			odp_packet_copy_from_mem(odp_pkt, pkt_chksum_offset,
 						 2, ZEROS);
 		else
-			*pkt_chksum_ptr = 0;
+			(is_tcp  ? (tcp_hdr_ptr->cksm = 0) : (udp_hdr_ptr->chksum = 0));
 	}
 
 	*l4_len_ptr            = l4_len;
 	*split_l4_hdr_ptr      = split_l4_hdr;
 	*is_tcp_ptr            = is_tcp;
 	*pkt_chksum_offset_ptr = pkt_chksum_offset;
-	*pkt_chksum_ptr_ptr    = pkt_chksum_ptr;
 	return 0;
 }
 
@@ -184,10 +182,10 @@ static inline int odph_process_l3_hdr(odp_packet_t odp_pkt,
 	odph_ipv6hdr_t *ipv6_hdr_ptr, ipv6_hdr;
 	odp_bool_t      split_l3_hdr;
 	swap_buf_t      swap_buf;
-	uint32_t        l3_offset, l4_offset, l3_hdrs_len, addrs_len;
-	uint32_t        protocol, l3_len, l4_len, idx, ipv6_payload_len, sum;
-	uint16_t       *addrs_ptr;
-	uint32_t hdr_len = 0;
+	uint32_t        l3_offset, l4_offset, l3_hdrs_len;
+	uint32_t        protocol, l3_len, l4_len, idx, ipv6_payload_len;
+	uint32_t        sum = 0;
+	uint32_t        hdr_len = 0;
 
 	/* The following computation using the l3 and l4 offsets handles both
 	 * the case of IPv4 options and IPv6 extension headers uniformly. */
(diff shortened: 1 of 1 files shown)


helper/chksum.c
```diff
@@ -130,7 +130,7 @@ static inline int odph_process_l4_hdr(odp_packet_t      odp_pkt,
 		 * should come from the udp header, unlike for TCP where is
 		 * derived. */
 		l4_len            = odp_be_to_cpu_16(udp_hdr_ptr->length);
-		pkt_chksum_ptr    = (uint16_t *)(void *)&udp_hdr_ptr->chksum;
+		pkt_chksum_ptr    = (uint16_t *)&l4_ptr[offsetof(odph_udphdr_t, chksum)];
```
psavol, line 133: Did you try cast to uintptr_t?

pkt_chksum_ptr = (uint16_t *)(uintptr_t)&udp_hdr_ptr->chksum;

xvikas: yes.. that does not help: 

chksum.c:133:46: warning: taking address of packed member of ‘struct <anonymous>’ may result in an unaligned pointer value [-Waddress-of-packed-member]
  133 |   pkt_chksum_ptr    = (uint16_t *)(uintptr_t)&udp_hdr_ptr->chksum;

Moreover I feel that from compiler perspective, it will treated similar to (void *) for packed structures.

xvikas: I had another thought about this. Compiler warning it because because there is no packed structure members indeed may be unaligned since compiler did not try to align it (compiler tried to put no extra padding between them and minimized the space consumed) but since base type is also not ensured to be aligned, overall we might endup having a unaligned pointer. If I align the base type also, then warning disappears: 

/** UDP header */
typedef struct **ODP_ALIGNED(4)** ODP_PACKED {
	odp_u16be_t src_port; /**< Source port */
	odp_u16be_t dst_port; /**< Destination port */
	odp_u16be_t length;   /**< UDP datagram length in bytes (header+data) */
	odp_u16be_t chksum;   /**< UDP header and data checksum (0 if not used)*/
} odph_udphdr_t;

Not sure which is better solution to address this.

Malvika-Gupta: I had run into this issue as well. As a matter of fact, I saw this issue another file too, odp_classification_common.c under test/validation. The highlighted lines were:
vlan_type = (odp_u16be_t *)(void *)&ethhdr->type;
vlan_type = (uint16_t *)(void *)&vlan_hdr->type;

I found a workaround for this issue (don't know if it is correct or valid) by void casting the pointer to a local
variable before using it, like shown below:
void *ethhdr_type = (void *)&ethhdr->type;
vlan_type = (odp_u16be_t *)ethhdr_type;

Doing this for all the conflicting lines seems to suppress the warning and I don't see it during compilation anymore.

JereLeppanen: I think a better solution would be to completely avoid taking the address by reorganizing the code a little.

psavol: To me usage of packed attribute in protocol headers is unnecessary and should be removed. Protocol header fields are "packed" by the specifications. E.g. UDP header has four uint16_t fields and thus "packed" should not have any effect on alignment of those fields.

Currently, this warning is suppressed in odp-linux build (as GCC10 complains those).

xvikas: I agree with you @psavol. There is no need to pack udp header (struct odph_udphdr_t) since all the members are 16-bits.  odph_tcphdr_t also looked to be well packed naturally, but its using the bit-fields. 
So shall we go ahead and remove the packed attribute from both udp header and tcp header struct declarations?

Malvika-Gupta: @psavol and @xvikas, we would need to consider if the above approach can probably take care of the conflicting lines in odp_classification_common.c as well. odph_vlanhdr_t doesn't need to be packed since both its members are 16-bits. As for odph_ethhdr_t, I am unsure if the above approach will work as expected as the source and destination MAC addresses are stored as array of 6 octets and alignment issues may arise.

typedef struct ODP_PACKED {
	uint8_t addr[ODPH_ETHADDR_LEN]; /**< @private Address */
} odph_ethaddr_t

typedef struct ODP_PACKED {
	odph_ethaddr_t dst; /**< Destination address */
	odph_ethaddr_t src; /**< Source address */
	odp_u16be_t type;   /**< EtherType */
} odph_ethhdr_t;

JannePeltonen: We need the packed attribute not only for getting rid of possible padding (which in most cases would not even be added) but also for reducing the minimum alignment of the structures and their members to 1 so that we can use the corresponding struct pointer types to access memory buffers (i.e. packet data) in arbitrary, possibly unaligned, memory locations.

xvikas:
Re-arranged the code to to avoid using the pointers for the packed structures (use the member values directly).

JannePeltonen:
Fixes in the helper function should be separated from additions to the external helper API and helper test improvements. i.e. they should be put in separate commits. That would also fix the problem that the commit message only mentioned the warning fix and not the other stuff.

Having everything in single commit is harder to read and selectively revert, should the need arise. It also has the danger that new bugs in the helper implementation and new bugs in its test may cancel each other out.

JannePeltonen:
Commit subject line could be more descriptive. Maybe something like this:

helper: fix address-of-packed-member compiler warning

> gcc10 reports warning.

I think this happens with gcc-9 too.

> To fix the issue, use the l4 pointer and add offset of the checksum

I did not understand this explanation.

helper/chksum.c
```diff
@@ -206,10 +204,14 @@ static inline int odph_process_l3_hdr(odp_packet_t odp_pkt,
...
 		}
 
-		addrs_ptr = (uint16_t *)(void *)&ipv4_hdr_ptr->src_addr;
-		addrs_len = 2 * ODPH_IPV4ADDR_LEN;
 		protocol  = ipv4_hdr_ptr->proto;
 		l3_len    = odp_be_to_cpu_16(ipv4_hdr_ptr->tot_len);
+
+		/* calculate IP pseudo header src and dest addr checksum */
+		sum       = ipv4_hdr_ptr->src_addr & 0xFFFF;
+		sum      += ipv4_hdr_ptr->src_addr >> 16 & 0xFFFF;
```
JannePeltonen, line 212: The "& 0xFFFF" part is unnecessary here. And the same applies to other occurences of this pattern.

helper/chksum.c
```diff
@@ -219,11 +221,15 @@ static inline int odph_process_l3_hdr(odp_packet_t odp_pkt,
 			ipv6_hdr_ptr = &ipv6_hdr;
 		}
 
-		addrs_ptr        = (uint16_t *)(void *)&ipv6_hdr_ptr->src_addr;
-		addrs_len        = 2 * ODPH_IPV6ADDR_LEN;
 		protocol         = ipv6_hdr_ptr->next_hdr;
 		ipv6_payload_len = odp_be_to_cpu_16(ipv6_hdr_ptr->payload_len);
 		l3_len           = ipv6_payload_len + ODPH_IPV6HDR_LEN;
+		for (idx = 0; idx < 8; idx++) {
```
JannePeltonen, line 227: The indexing is wrong. Only the first 9 bytes of the addresses are looked at and most of them twice.

This would not look right to me even with fixed indexing. This code produces the same result regardless of the endianess of the CPU whereas the old code produces different results in big and little endian CPUs.

helper/chksum.c
```diff
@@ -280,11 +280,12 @@ int odph_udp_tcp_chksum(odp_packet_t     odp_pkt,
 
 	/* First parse and process the l4 header */
 	rc = odph_process_l4_hdr(odp_pkt, op, &udp_tcp_hdr, chksum_ptr, &l4_len,
-				 &split_l4_hdr, &is_tcp, &pkt_chksum_offset,
-				 &pkt_chksum_ptr);
+				 &split_l4_hdr, &is_tcp, &pkt_chksum_offset);
 	if (rc != 0)
 		return rc;
 
+	pkt_chksum_ptr = (uint16_t *)(((uint8_t *)odp_packet_data(odp_pkt)) + pkt_chksum_offset);
```
JannePeltonen, line 287: This loses information about the minimum guaranteed aligment of the checksum field (which is 1). pkt_chksum_ptr is uint16_t pointer which has guaranteed alignment of 2. The cast through uint8_t* hides the problem in most platforms (except if --Wcast-align=strict is used, I think).

The overall approach for not taking address of a packed struct member looks ok to me, except for this line which does basically the same potentially dangerous thing but in a different way to hide it from view.

My point is that if we wanted to silence the warning without really fixing anything, we could have just added a suitable pragma and be done with it. But if we want to silence the warning by getting rid of the alignment issue that causes it, then it is not good to introduce another, similar but subtly different, alignment issue in the code as part of the fix (even if there may be other alignment problems, not warned about by the compiler, in the existing code).

helper/ip.c
```diff
@@ -28,3 +28,23 @@ int odph_ipv4_addr_parse(uint32_t *ip_addr, const char *str)
...
+
+int odph_ipv6_addr_parse(uint8_t * const ip_addr, const char * const str)
+{
+	int i, j;
+	uint16_t hextet[8];
+
+	if ((NULL == ip_addr) || (NULL == str))
+		return -1;
+
+	if (sscanf(str, "%hx:%hx:%hx:%hx:%hx:%hx:%hx:%hx", &hextet[0], &hextet[1], &hextet[2],
```
JannePeltonen, line 40: "hx" is not portable for scanning into uint16_t as there is no guarantee that unsigned short int is the same as uint16_t. You should use the format specifier defined by the SCNx16 macro.

helper/ip.c
```diff
@@ -28,3 +28,23 @@ int odph_ipv4_addr_parse(uint32_t *ip_addr, const char *str)
...
+		   &hextet[3], &hextet[4], &hextet[5], &hextet[6],
+		   &hextet[7]) != (ODPH_IPV6ADDR_LEN / 2))
+		return -1;
+
+	for (i = 0, j = 0; i < 16; i += 2, j++) {
+		ip_addr[i] = hextet[j] & 0xFF;
```
JannePeltonen, line 46: Byte order looks incorrect. If we consider address "AABB::", 0xBB ends up in the first memory location and 0xAA in the second.

xvikas:
> Fixes in the helper function should be separated from additions to the external helper API and helper test improvements. i.e. they should be put in separate commits. That would also fix the problem that the commit message only mentioned the warning fix and not the other stuff.
> 
> Having everything in single commit is harder to read and selectively revert, should the need arise. It also has the danger that new bugs in the helper implementation and new bugs in its test may cancel each other out.

New patch has only the compiler warning fix in chksum.c. I will push two other patches for the test enhancements

helper/chksum.c
xvikas: It's more of the code readability.. I believe its done in other places in code too. But I have removed unnecessary masking

helper/chksum.c
xvikas: Yes.. Will be fixed in the new patch-set

helper/chksum.c
xvikas: Fixed

helper/ip.c
xvikas: Yes.. I will fix this in another change-set. I can completely avoid using sscanf() and just parse it like a string. Is that fine?

helper/ip.c
xvikas: Yes.. I will fix this in another change-set

psavol:
Should we close this one?

psavol:
The original build issue has been fixed already. Closing this one since no progress for two months. Can be re-opened later if needed.

PR #1047, opened 2020-09-23
[PATCH v1] api: add packet inner layer identification API and inner layer parsing, checksum generation and validation

nithind1988:
This patchset expands pktio and packet spec to allow inner layer parsing, packet identification, checksum generation, validation.
Inner layer API are needed for inner protocol identification, parsing, checksum validation and generation for tunneled packets.
For example, VXLAN tunnel packets will have outer IPv4, IPv6, UDP and then Inner IPv4/IPv6, Inner TCP/UDP. Parsing and checksum generation for both the layers is a must for many usecases.

Also this patchset changes existing behavior of Tx checksum generation dependent on only L3 and L4 offsets in metadata to include L3 and L4 proto's of meta data. For platforms that doesn't support HW parsing of L3 and L4 proto in Tx, platform SW needs to parse them and it effects performance.
Since L3 and L4 offsets are already allowed to be taken from pkt metadata and they are initially filled by Ingress pkt HW parsing,
taking L3 and L4 proto types as well from pkt metadata is beneficial for performance as they are also filled by Ingress pkt HW parsing.

### DIFF
--- a/doc/users-guide/users-guide-packet.adoc
+++ b/doc/users-guide/users-guide-packet.adoc
@@ -131,7 +131,10 @@ application changes the Layer 3 offset by using the `odp_packet_l3_offset_set()`
 API, the subsequent calls to `odp_packet_l3_ptr()` will return an address
 starting from that changed offset, changing an attribute like
 `odp_packet_has_udp_set()` will not, by itself, turn a non-UDP packet into
-a valid UDP packet. Applications are expected to exercise appropriate care
+a valid UDP packet. The same applies to inner layers and its offsets which
+can be modified using equivalent inner layer API such as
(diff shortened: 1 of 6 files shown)


nithind1988:
@JannePeltonen @psavol

JannePeltonen:
I think the meaning of "inner header", "inner L2", "inner L3" may not be sufficiently unambiguously defined. What if there are multiple tunneling encapsulations in the same packet? What if the tunnel encapsulation is not recognized? Are there some encapsulations that will always be recognized and parsed?

VxLAN is quite a difficult case since one cannot be sure whether a packet is a VxLAN packet or some random UDP packet by just looking at the packet. The interpretation depends on the local configuration,

JannePeltonen:

PR #1049, opened 2020-09-24
[PATCH v2] api: CLS mark support (with implementation)

psavol:
This PR finalizes the mark API changes of https://github.com/OpenDataPlane/odp/pull/1010

### DIFF
--- a/platform/linux-generic/include/odp_packet_internal.h
+++ b/platform/linux-generic/include/odp_packet_internal.h
@@ -123,6 +123,9 @@ typedef struct odp_packet_hdr_t {
 	/* Flow hash value */
 	uint32_t flow_hash;
 
+	/* Classifier mark */
+	uint16_t cls_mark;
+
 	union {
 		struct {
 			/* Result for crypto packet op */
@@ -198,8 +201,9 @@ static inline void packet_init(odp_packet_hdr_t *pkt_hdr, uint32_t len)
 		last->seg_len = seg_len;
 	}
 
-	pkt_hdr->p.input_flags.all  = 0;
-	pkt_hdr->p.flags.all_flags  = 0;
+	/* Clear all flags. Resets also return value of cls_mark, user_ptr, etc. */
+	pkt_hdr->p.input_flags.all = 0;
+	pkt_hdr->p.flags.all_flags = 0;
 
 	pkt_hdr->p.l2_offset = 0;
 	pkt_hdr->p.l3_offset = ODP_PACKET_OFFSET_INVALID;
@@ -233,6 +237,7 @@ static inline void copy_packet_cls_metadata(odp_packet_hdr_t *src_hdr,
 	dst_hdr->dst_queue = src_hdr->dst_queue;
 	dst_hdr->flow_hash = src_hdr->flow_hash;
 	dst_hdr->timestamp = src_hdr->timestamp;
+	dst_hdr->cls_mark  = src_hdr->cls_mark;
 }
 
 static inline void *packet_data(odp_packet_hdr_t *pkt_hdr)
--- a/platform/linux-generic/odp_packet.c
+++ b/platform/linux-generic/odp_packet.c
@@ -1579,6 +1579,8 @@ void odp_packet_print(odp_packet_t pkt)
 	}
 	len += snprintf(&str[len], n - len, "  flags        0x%" PRIx32 "\n",
 			hdr->p.flags.all_flags);
+	len += snprintf(&str[len], n - len, "  cls_mark     %" PRIu64 "\n",
+			odp_packet_cls_mark(pkt));
 	len += snprintf(&str[len], n - len,
 			"  l2_offset    %" PRIu32 "\n", hdr->p.l2_offset);
 	len += snprintf(&str[len], n - len,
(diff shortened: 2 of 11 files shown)


psavol:
I did couple of white space and doxygen style changes (the first sentence after doxygen tag is the short description) to the API patch, but other than that it's the same as in PR 1010.

platform/linux-generic/include/odp_packet_internal.h
```diff
@@ -123,6 +123,9 @@ typedef struct odp_packet_hdr_t {
 	/* Flow hash value */
 	uint32_t flow_hash;
 
+	/* Classifier mark */
+	uint16_t cls_mark;
```
JannePeltonen, line 127: Initialization to zero is missing from packet_init().

platform/linux-generic/odp_packet.c
```diff
@@ -1579,6 +1579,7 @@ void odp_packet_print(odp_packet_t pkt)
 	}
 	len += snprintf(&str[len], n - len, "  flags        0x%" PRIx32 "\n",
 			hdr->p.flags.all_flags);
+	len += snprintf(&str[len], n - len, "  cls_mark     %" PRIu16 "\n", hdr->cls_mark);
```
JannePeltonen, line 1582: Ok for this patch since it just follows the existing pattern. But the code is already broken because it does not take into account that the return value of snprintf is the number of characters that would have been written if the buffer were long enough and not the number actually written. But since that bug does not have anything to do with this patch, let's deal with it separately.

test/validation/api/classification/odp_classification_test_pmr.c
```diff
@@ -2477,6 +2522,20 @@ static int check_capa_pmr_series(void)
...
 
+static int check_capa_pmr_marking(void)
+{
+	uint64_t terms;
+
+	terms = cls_capa.supported_terms.bit.dip_addr &&
+		cls_capa.supported_terms.bit.udp_dport;
+
+	/* one PMR for IP, MAX_NUM_UDP PMRs for UDP */
+	if (terms && cls_capa.max_mark >= (MAX_NUM_UDP + 1))
```
JannePeltonen, line 2533: The comparison should be with MARK_UDP + MAX_NUM_UDP - 1.

test/validation/api/classification/odp_classification_test_pmr.c
```diff
@@ -2237,7 +2242,17 @@ static void test_pmr_series(const int num_udp)
 	pmr_param.val_sz      = sizeof(dst_addr_be);
 	pmr_param.offset      = 0;
 
-	pmr_ip = odp_cls_pmr_create(&pmr_param, 1, default_cos, cos_ip);
+	if (marking) {
```
JannePeltonen, line 2245: It would be nicer to have the repeating code put in a function that would take the "mark" and "marking" parameters in addition to the existing stuff. The test function is already quite long.

test/validation/api/classification/odp_classification_test_pmr.c
```diff
@@ -2293,6 +2317,14 @@ static void test_pmr_series(const int num_udp)
 	CU_ASSERT_FATAL(pkt != ODP_PACKET_INVALID);
 	CU_ASSERT(seqno == cls_pkt_get_seq(pkt));
 	CU_ASSERT(retqueue == queue_ip);
+
+	if (marking) {
```
JannePeltonen, line 2321: I am just thinking that should there be a test case also for the case where first level PMR matches but the second level PMRs do not. All PMRs would be configured to do the marking and the expected mark value would be the one of the firt level PMR. The current test would not notice if the implementation wrote the mark whenever it compared the packet with a PMR instead of only setting the mark upon PMR match.

platform/linux-generic/include/odp_packet_internal.h
psavol: Done in v2. Uses a flag for maintaining packet input / alloc performance.

platform/linux-generic/odp_packet.c
psavol: Our snprintf() usage seems to rely on large enough str[] buffer size (4kB in this case). A macro/function could be used to wrap snprintf() return value check.

test/validation/api/classification/odp_classification_test_pmr.c
psavol: fixed in v2

test/validation/api/classification/odp_classification_test_pmr.c
psavol: I'd keep this code flat, since a generic function for IP/UDP PMR create would save only 10 lines. Also, flat validation test code is easier to use when debugging implementation as asserts/printfs can be inserted directly to the failing branch.

test/validation/api/classification/odp_classification_test_pmr.c
psavol: I think this could be another test case for parallel/serial PMR usage in general, and include marking test cases into it. However, I'll add cls_mark default value tests to packet validation suite.

JannePeltonen (approved):

PR #1052, opened 2020-09-29
[PATCH v3] test: performance: add odp_sys_info_print() to all performance tests

Malvika-Gupta:
The first patch adds the odp_sys_info_print() to all the performance tests which currently do not display this info. The ODP system info contains important information such as CPU model, frequency, cache etc. which is needed to correctly analyze and interpret the results, especially those which are given in terms of CPU cycles. As of now, this information is shown for only half of the performance tests.
The second patch fixes a spelling error found in the results of the queue performance test.

### DIFF
--- a/test/performance/odp_crypto.c
+++ b/test/performance/odp_crypto.c
@@ -1064,6 +1064,8 @@ int main(int argc, char *argv[])
 	/* Init this thread */
 	odp_init_local(instance, ODP_THREAD_WORKER);
 
+	odp_sys_info_print();
+
 	if (odp_crypto_capability(&crypto_capa)) {
 		app_err("Crypto capability request failed.\n");
 		exit(EXIT_FAILURE);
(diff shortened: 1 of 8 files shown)


Malvika-Gupta:
@psavol @JannePeltonen A gentle reminder and request to please review these patches. Thank you!

JereLeppanen (approved):
Looks OK to me. For consistency, the odp_sys_info_print() call could perhaps be right after odp_init_local() and parse_options() calls, whichever comes last.

Reviewed-by: Jere Leppänen <jere.leppanen@nokia.com>

Malvika-Gupta:
Thank you so much for the review @JereLeppanen. I will send a 2nd version of the patch with your reviewed-by tag and will also check the patch to ensure that odp_sys_info_print() calls and occur after either odp_init_local() or parse_options().

PR #1053, opened 2020-09-29
[PATCH v2] test: packet_gen: add random packet length option

matiaselo:
Add new packet length range option (-L, --len_range <min,max,bins>) to
enable testing with random packet lengths. To reduce pool size requirement
the length range can be split into even sized bins. Min and max size
packets are always used and one size is selected from each other bin. Bin
value of 0 means that each packet length is used. This option overrides the
standard packet length command line option.

### DIFF
--- a/test/performance/odp_packet_gen.c
+++ b/test/performance/odp_packet_gen.c
@@ -1514,13 +1514,13 @@ int main(int argc, char **argv)
 	/* Init ODP before calling anything else */
 	if (odp_init_global(&instance, &init, NULL)) {
 		printf("Error: Global init failed.\n");
-		return -1;
+		return 1;
 	}
 
 	/* Init this thread */
 	if (odp_init_local(instance, ODP_THREAD_CONTROL)) {
 		printf("Error: Local init failed.\n");
-		return -1;
+		return 1;
 	}
 
 	shm = odp_shm_reserve("packet_gen_global", sizeof(test_global_t),
@@ -1528,7 +1528,7 @@ int main(int argc, char **argv)
 
 	if (shm == ODP_SHM_INVALID) {
 		printf("Error: SHM reserve failed.\n");
-		return -1;
+		return 1;
 	}
 
 	global = odp_shm_addr(shm);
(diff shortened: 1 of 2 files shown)


test/performance/odp_packet_gen.c
```diff
@@ -145,6 +154,14 @@ static void print_usage(void)
...
 	       "  -n, --num_pkt             Number of packets in the pool. Default: 1000\n"
 	       "  -l, --len                 Packet length. Default: 512\n"
+	       "  -L, --len_range <min,max,bins>\n"
+	       "                            Random packet length. Specify the minimum and maximum\n"
+	       "                            packet lengths and the number of bins. To reduce pool size\n"
+	       "                            requirement the length range can be divined into even sized\n"
+	       "                            bins. Min and max size packets are always used and one size\n"
+	       "                            is selected from each other bin. Bin value of 0 means that\n"
+	       "                            each packet length is used. Comma-separated (no spaces).\n"
+	       "                            Overrides standard packet length option.\n"
```
psavol, line 164: Number of bins documentation could be more clear e.g.:
"... Min and max size packets are always used and included into number of bins (bins >= 2) ..."

test/performance/odp_packet_gen.c
```diff
@@ -1115,12 +1229,69 @@ static inline int send_burst(odp_pktout_queue_t pktout, odp_packet_t pkt[],
...
+	if (test_options->use_rand_pkt_len) {
+		int i, j;
+		int num_pktio = test_options->num_pktio;
+		uint32_t pkt_bins = test_options->rand_pkt_len_bins;
+		uint32_t pkt_len_min = test_options->rand_pkt_len_min;
+		uint32_t pkt_len_max = test_options->rand_pkt_len_max;
+		uint32_t bin_size = 1;
+
+		if (pkt_bins)
+			bin_size = (pkt_len_max - pkt_len_min + 1) / pkt_bins;
```
psavol, line 1255: Should division be done with (pkt_bins - 1)?

E.g. max = 20, min = 10, bins = 3: leeds now to packet lengths of 10, 13 and 20

psavol (approved):

PR #1054, opened 2020-09-30
[PATCH v3] test: l2fwd: add loop to drain prefetched events

shijith:
Fix l2fwd mode with ordered queues for schedulers with prefetch
capability. If prefetch is enabled, prefetched event is not received by
a worker as exit_threads is checked in l2fwd loop. The prefetched event
could block the forward progress of other workers because of ordering
constraints. This patch adds a loop to drain prefetched event before
entering barrier to wait for other threads.

Signed-off-by: Shijith Thotton <sthotton@marvell.com>

### DIFF
--- a/test/performance/odp_l2fwd.c
+++ b/test/performance/odp_l2fwd.c
@@ -434,6 +434,22 @@ static int run_worker_sched_mode(void *arg)
 		stats->s.packets += pkts;
 	}
 
+	/*
+	 * Free prefetched packets before entering the thread barrier.
+	 * Such packets can block sending of later packets in other threads
+	 * that then would never enter the thread barrier and we would
+	 * end up in a dead-lock.
+	 */
+	odp_schedule_pause();
+	while (1) {
+		odp_event_t  ev;
+
+		ev = odp_schedule(NULL, ODP_SCHED_NO_WAIT);
+		if (ev == ODP_EVENT_INVALID)
+			break;
+		odp_event_free(ev);
+	}
+
 	/* Make sure that latest stat writes are visible to other threads */
 	odp_mb_full();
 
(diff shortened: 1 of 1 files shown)


shijith:
@psavol This is an initial fix for an issue which we found on scheduler with prefetch capability. We can discuss and reach a better solution.

test/performance/odp_l2fwd.c
```diff
@@ -434,6 +434,18 @@ static int run_worker_sched_mode(void *arg)
...
+	odp_schedule_pause();
+	while (1) {
+		odp_event_t  ev;
+
+		ev = odp_schedule(NULL, ODP_SCHED_NO_WAIT);
+		if (ev == ODP_EVENT_INVALID)
+			break;
+		odp_event_free(ev);
+	}
+	odp_schedule_resume();
```
JannePeltonen, line 447: At this point pktios have not necessarily been stopped yet, so new events can be available and maybe the scheduler could do some prefetching even without an explicit schedule call by this thread (e.g. HW scheduler prefetching event asynchronously or SW scheduler or SW packet reception in some other thread pushing packets to this thread. Then we would be in the same state as before the pause-resume dance.

ODP API guarantees that stale locally prefetched events will not be left dangling is one first calls pause and then drains the scheduler locally by calling schedule. I would not like to depend on the exact behavior when not following that sequence. If the resume call were moved little bit forward so that it comes after the barrier wait, then this issues
would seem to go away as at that point the pktios have already been stopped and all threads have come out of the forwarding loop.

shijith: Moving resume call below barrier makes sense. Will update the patch.

JannePeltonen:
This looks ok to me know, but it may be far from obvious to a casual reader why there are two event draining loops almost back to back like that and why the first one could not be simply removed. Could you add a clarifying comment in the code, perhaps something like this:

/*
 * Free prefetched packets before entering the thread barrier. 
 * Such packets can block sending of later packets in other threads
 * that then would never enter the thread barrier and we would
 * end up in a dead-lock.
 */

You can also already add my reviewed-by tag.

shijith:
@JannePeltonen Thanks for the review. I will send v3 with the comment and tag.

JannePeltonen (approved):

PR #1055, opened 2020-09-30
[PATCH v8] api: pktio: support for HW Tx timestamp offload

nithind1988:
Current ODP Packet-IO and ODP Packet spec has support to enable HW timestamping offload on Rx packets and retrieve them respectively. Since in some usecases like debugging/PTP, Tx timestamping is also needed, this series adds modification to 
Packet-IO spec to allowing enabling Tx Timestamp offload via odp_pktio_config() and enable per packet timestamp insertion using
odp_packet_*() API. Since transmission by HW on link is asynchronous, application has to poll for timestamp using odp_pktout_ts() API to get last timestamped packet's timestamp.


JannePeltonen:
General comment about RX and TX time stamps:

I think ODP API should be more clear about the clocks and time scales that apply to RX and TX time stamping. I think every pktio should have a separate clock not synced to any other clock (not in absolute time nor in frequency). RX and TX time stamps use the time from the clock of the pktio and cannot directly be compared to the RX and TX time stamps of other pktios nor the global or local ODP time.

This model would match HW (like discrete NICs) where network I/O ports have their own clock and would leave all clock synchronization issues to the application. Maybe this is already the existing model since ODP API refers to "packet input time source" in odp_pktin_ts_from_ns() documentation, but other than that, the nature of the RX time stamps and their time scale are not discussed.

If the above is the model, I think we need an additional API function that provides the current time of a pktio clock (maybe as offset to odp_time_global()), to make it easier to implement PTP.

doc/users-guide/users-guide-pktio.adoc
```diff
@@ -390,10 +390,21 @@ transmitted:
  * contain at least as many data bytes after L3/L4 offsets as the headers
  * indicate. Other data bytes of the packet are ignored for the checksum
  * insertion.
+ *
+ * Packet Tx timestamp insertion may be enabled or disabled.
```
JannePeltonen, line 394: Talking about timestamp insertion is misleading since the timestamp is not inserted into the packet (like checksum is) but just reported to the application later. There are probably also other things in the API text that need some improvement but I think these can be postponed until the API design itself is clear.

include/odp/api/spec/packet_io.h
```diff
@@ -1394,6 +1397,18 @@ uint64_t odp_pktin_ts_res(odp_pktio_t pktio);
...
 
+/**
+ * Get Tx timestamp inserted for a packet if available and clear it for
+ * next timestamp.
+ *
+ * @param pktio handle identifying pktio
+ * @param ts    Pointer to odp_time_t where timestamp will be written
+ *
+ * @retval zero     Successfully returns last pkt timestamp.
+ * @retval negative error code on failure.
```
JannePeltonen, line 1408: We probably should distinguish between different kind of errors: 1) Time stamp not available because none has been requested, 2) Time stamp not yet available for a sent packet, try again later, 3) Time stamping failed or other error occured, retrying is useless.

Maybe 1) and 2) could be the same error code.

include/odp/api/spec/packet_io.h
```diff
@@ -1394,6 +1397,18 @@ uint64_t odp_pktin_ts_res(odp_pktio_t pktio);
...
+ */
+int odp_pktout_ts(odp_pktio_t pktio, odp_time_t *ts);
```
JannePeltonen, line 1410: I was thinking if this is a good API instead of defining an event and/or making it possible to have multiple timestamped packets in flight in the same pktio (as supported by some HW). But maybe this is better as it is simple.

nithind1988: I also think, since PTP packets are not that frequent and also we need them to work even when application cannot support schedule mode, this API is simple and sufficient for the usecase.
We can also extended this in future to generate events if any vendor hardware supports it.

include/odp/api/spec/packet_io.h
nithind1988: Ok. Will fix it with distinguishing between 1), 2) error and 3) error.

nithind1988:
> 
> 
> General comment about RX and TX time stamps:
> 
> I think ODP API should be more clear about the clocks and time scales that apply to RX and TX time stamping. I think every pktio should have a separate clock not synced to any other clock (not in absolute time nor in frequency). RX and TX time stamps use the time from the clock of the pktio and cannot directly be compared to the RX and TX time stamps of other pktios nor the global or local ODP time.
> 
> This model would match HW (like discrete NICs) where network I/O ports have their own clock and would leave all clock synchronization issues to the application. Maybe this is already the existing model since ODP API refers to "packet input time source" in odp_pktin_ts_from_ns() documentation, but other than that, the nature of the RX time stamps and their time scale are not discussed.
> 
> If the above is the model, I think we need an additional API function that provides the current time of a pktio clock (maybe as offset to odp_time_global()), to make it easier to implement PTP.

Agreed. Can I add two new API's odp_pktio_ts_from_ns() and odp_pktio_ns_from_ts() and update documentation that says that packet input and packet output timestamp will be based on PKTIO specific clock ?
We may have to deprecate odp_pktin_ts_from_ns() as it will become redundant given that we are making time source to be pktio specific and not pktin or pktout specific.

JannePeltonen:
> Agreed. Can I add two new API's odp_pktio_ts_from_ns() and odp_pktio_ns_from_ts() and update documentation that says that packet input and packet output timestamp will be based on PKTIO specific clock ?
> We may have to deprecate odp_pktin_ts_from_ns() as it will become redundant given that we are making time source to be pktio specific and not pktin or pktout specific.

That sounds reasonable. I think we need to call the pktio clock "pktio time source" in accordance with other time documentation (see time.h).

Then we would need a function called something like odp_time_pktio(pktio) that returns time from a pktio time source. But for PTP that would be inconvenient as there would be delay in reading and processing the time so it would be good to have a function that somehow returns the difference (at that point of time) of global time and pktio time. Or maybe a function that returns pktio time _and_ global time, read at the same moment?

nithind1988:
> 
> 
> > Agreed. Can I add two new API's odp_pktio_ts_from_ns() and odp_pktio_ns_from_ts() and update documentation that says that packet input and packet output timestamp will be based on PKTIO specific clock ?
> > We may have to deprecate odp_pktin_ts_from_ns() as it will become redundant given that we are making time source to be pktio specific and not pktin or pktout specific.
> 
> That sounds reasonable. I think we need to call the pktio clock "pktio time source" in accordance with other time documentation (see time.h).
> 
> Then we would need a function called something like odp_time_pktio(pktio) that returns time from a pktio time source. But for PTP that would be inconvenient as there would be delay in reading and processing the time so it would be good to have a function that somehow returns the difference (at that point of time) of global time and pktio time. Or maybe a function that returns pktio time _and_ global time, read at the same moment?

Ack, Will sent next version with three API's odp_pktio_ts_from_ns() , odp_pktio_ns_from_ts() and odp_time_pktio(odp_time_t *pktio_clk, odp_time_t *global_clk) that gives both global time and pktio time at same point of time.

JannePeltonen:
> Ack, Will sent next version with three API's odp_pktio_ts_from_ns() , odp_pktio_ns_from_ts() and odp_time_pktio(odp_time_t *pktio_clk, odp_time_t *global_clk)

odp_pktio_ns_from_ts() is not needed since there is already a generic function for that: odp_time_to_ns(). That does not need to be time source specific since either time stamps of all time sources can be converted in the same way or odp_time_t carries information on what its time source is.

nithind1988:
> 
> 
> > Ack, Will sent next version with three API's odp_pktio_ts_from_ns() , odp_pktio_ns_from_ts() and odp_time_pktio(odp_time_t *pktio_clk, odp_time_t *global_clk)
> 
> odp_pktio_ns_from_ts() is not needed since there is already a generic function for that: odp_time_to_ns(). That does not need to be time source specific since either time stamps of all time sources can be converted in the same way or odp_time_t carries information on what its time source is.

Ack. Missed that point.

doc/users-guide/users-guide.adoc
```diff
@@ -404,7 +404,7 @@ To get rate of time source `odp_time_local_res()`, `odp_time_global_res()`
 are used. To wait, `odp_time_wait_ns()` and `odp_time_wait_until()` are used,
 during which a thread potentially busy loops the entire wait time.
 
-The `odp_time_t` opaque type represents local or global timestamps.
+The `odp_time_t` opaque type represents local, global and PktIO timestamps.
```
psavol, line 407: When possible, an API patch should contain only include/odp/api/spec/ header file changes. It's easier to merge patches to other repos when changes done by one patch is limited as few files as possible. E.g. these user guide changes can go to a separate patch.

include/odp/api/spec/packet_io.h
```diff
@@ -1394,6 +1397,69 @@ uint64_t odp_pktin_ts_res(odp_pktio_t pktio);
...
+
+/**
+ * Get last captured Tx timestamp of a packet if available and clear it for
+ * next timestamp.
+ *
+ * @param pktio handle identifying pktio
+ * @param ts    Pointer to odp_time_t where timestamp will be written
+ *
+ * @retval zero     Successfully returns last pkt timestamp.
+ * @retval positive Timestamp not available either because none as been
```
JannePeltonen, line 1461: Typo: as -> has

include/odp/api/spec/packet_io.h
```diff
@@ -1394,6 +1397,69 @@ uint64_t odp_pktin_ts_res(odp_pktio_t pktio);
  */
 odp_time_t odp_pktin_ts_from_ns(odp_pktio_t pktio, uint64_t ns);
 
+/**
+ * Packet IO timestamp resolution in hertz
+ *
+ * This is the resolution of packet input and packet output timestamps.
+ * using Packet IO time source. Returns zero on a failure or when both
+ * packet in and packet out timestamping is disabled.
```
JannePeltonen, line 1411: typo: is ->are

Maybe also use the terms RX and TX timestamping (instead of packet in and packet out timestampig) here too, for consistency.

platform/linux-generic/odp_packet_io.c
```diff
@@ -1288,12 +1288,44 @@ odp_time_t odp_pktin_ts_from_ns(odp_pktio_t hdl, uint64_t ns)
...
+	/* Callback if present */
+	if (entry->s.ops->pktio_time)
+		return entry->s.ops->pktio_time(entry, global_ts);
+
+	/* By default both Packet IO time source and
+	 * global time source are same.
+	 */
+	if (global_ts)
+		*global_ts = odp_time_global();
+	return *global_ts;
```
JannePeltonen, line 1316: This is a null pointer dereference if the caller is not interested in global time.

include/odp/api/spec/packet_io.h
```diff
@@ -1394,6 +1397,69 @@ uint64_t odp_pktin_ts_res(odp_pktio_t pktio);
...
+ * Returns current Packet IO time and optionally
+ * global time. This global time is that of global time source
+ * where as Packet IO timestamp is of Packet IO time source that
+ * is used to timestamp incoming and outgoing packets.
+ *
+ * @param      pktio        Packet IO handle
+ * @param      ts_global    Pointer to odp_time_t.
+ *                          When not NULL, on success, global timestamp
+ *			    will be updated taken at same point of time
+ *			    as Packet IO time.
```
psavol, line 1446: Indent comment text using spaces only. Mix of tabs and spaces does not show clean on different editors/web view/etc.

include/odp/api/spec/packet_io.h
psavol: How application is able to be sure which (outgoing) packet the timestamp refers to. I think application would need to mark the packet that requests a time stamp during output. E.g.

/**
 *  Request packet output to record a time stamp
 *
 * Time stamp recorded when this packet is output on the interface. Time stamp 
 * can be retrieved with odp_pktout_ts().
*/
int odp_packet_ts_request(odp_packet_t pkt);

psavol: odp_pktout_ts() is a bit short / not very descriptive name. Would e.g. odp_pktout_ts_poll() or odp_pktout_ts_read() tell that this function is used asynchronously to poll the interface after packet output. It's part of normal operation that application needs to poll for an implementation specific time before the time stamp is available.

include/odp/api/spec/packet.h
```diff
@@ -1940,6 +1940,22 @@ odp_time_t odp_packet_ts(odp_packet_t pkt);
...
+ * called always but Tx timestamping capture is not enabled if pktio is not
+ * configured to enable timestamp.
+ *
+ * @param pkt     Packet handle
+ * @param enable  0: do not capture timestamp on Tx
+ *                1: capture timestamp on Tx
+ *
+ * @see odp_pktout_ts()
+ */
+void odp_packet_ts_enable(odp_packet_t pkt, int enable);
```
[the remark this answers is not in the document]
JannePeltonen: I concur, I was just about to write essentially the same comment.

include/odp/api/spec/packet_io.h
nithind1988: Ok. Will change to odp_pktout_ts_read() then.

test/validation/api/pktio/pktio.c
```diff
@@ -2165,6 +2165,113 @@ static void pktio_test_pktin_ts(void)
...
+	pktio_info_t pktio_rx_info;
+	odp_pktio_capability_t capa;
+	odp_pktio_config_t config;
+	odp_time_t ts_prev;
+	odp_time_t ts;
+	int num_rx = 0;
+	int ret;
+	int i;
+
+	memset(&pktout_queue, 0, sizeof(pktout_queue));
```
JannePeltonen, line 2207: This memset is not needed.

test/validation/api/pktio/pktio.c
```diff
@@ -2165,6 +2165,113 @@ static void pktio_test_pktin_ts(void)
...
+						&pkt_tbl[i], 1) == 1);
+		ret = wait_for_packets(&pktio_rx_info, &pkt_tbl[i], &pkt_seq[i],
+				       1, TXRX_MODE_SINGLE, ODP_TIME_SEC_IN_NS);
+		if (ret != 1)
+			break;
+
+		/* Since we got packet back, check for sent ts */
+		CU_ASSERT_FATAL(odp_pktout_ts_read(pktio_tx, &ts) == 0);
+
+		CU_ASSERT(odp_time_cmp(ts, ts_prev) > 0);
```
JannePeltonen, line 2258: In the first round this compares the TX timestamp with ODP_TIME_NULL, but the return value of odp_time_cmp() is not well defined in that case (if it is even ok to make such a call in the first place).

This can be fixed by initializing ts_prev() using odp_pktio_time() (and having the PKTIO_TS_INTERVAL delay after it), which has the added benefit of testing odp_pktio_time() a little bit.

test/validation/api/Makefile.am
```diff
@@ -56,6 +56,7 @@ TESTS = \
 	ipsec/ipsec_inline_out$(EXEEXT) \
 	lock/lock_main$(EXEEXT) \
 	packet/packet_main$(EXEEXT) \
+	pktio/pktio_main$(EXEEXT) \
```
matiaselo, line 59: This test will default to using "loop" interface. So in case of linux-generic you could remove the now unnecessary loop pktio test in `platform/linux-generic/test/validation/api/pktio/pktio_run.sh`.

JannePeltonen:

matiaselo:
Replaced by #1083.

PR #1057, opened 2020-10-05
[PATCH v1] test: packet_gen: fix error check for less rounds

nithind1988:
Fix error check for case when total tx packets is
less than MIN_RX_PACKETS_CI.

### DIFF
--- a/test/performance/odp_packet_gen.c
+++ b/test/performance/odp_packet_gen.c
@@ -1475,7 +1475,7 @@ static int print_final_stat(test_global_t *global)
 	printf("  tx Mbit/s:                  %.1f\n", tx_mbit_per_sec);
 	printf("\n");
 
-	if (rx_pkt_sum < MIN_RX_PACKETS_CI)
+	if (rx_pkt_sum < MIN_RX_PACKETS_CI && rx_pkt_sum < tx_pkt_sum)
 		return -1;
 
 	return 0;


JereLeppanen (commented):
I suppose this would be fine. But since this only affects the exit code, for CI, what is the actual benefit of this change?

nithind1988:
> 
> 
> I suppose this would be fine. But since this only affects the exit code, for CI, what is the actual benefit of this change?

It is needed for CI tests one of our simulators, where we cannot run default number of rounds i.e 500.

JereLeppanen:
> It is needed for CI tests one of our simulators, where we cannot run default number of rounds i.e 500.

Don't you need to modify the packet_gen parameters as well? And don't you have similar problems with other tests?

nithind1988:
> 
> 
> > It is needed for CI tests one of our simulators, where we cannot run default number of rounds i.e 500.
> 
> Don't you need to modify the packet_gen parameters as well? And don't you have similar problems with other tests?

That parameter we are overriding using our platform specific pktio_env and a wrapper script around odp_packet_gen

JereLeppanen:
> That parameter we are overriding using our platform specific pktio_env and a wrapper script around odp_packet_gen

I see, thanks for the explanation.

But on second thought, actually this change could make CI unable to catch error situations where tx is super slow or stuck. Maybe MIN_RX_PACKETS_CI should be a parameter instead of a define.

matiaselo:
> But on second thought, actually this change could make CI unable to catch error situations where tx is super slow or stuck. Maybe MIN_RX_PACKETS_CI should be a parameter instead of a define.

I agree with this. We have similar checks also in several other examples (e.g. l2fwd, l2fwd_simple). For example `-C --ci-limit` should be available to use with all examples.

nithind1988:
I agree. I'll drop this pull request for now as we have found another way to speed up our simulator and it seems to work fine.
I'll work on adding -C option in future if we again face similar issue.
Thanks

matiaselo:
> I agree. I'll drop this pull request for now as we have found another way to speed up our simulator and it seems to work fine.
> I'll work on adding -C option in future if we again face similar issue.
> Thanks

OK, I'll close the PR then.

PR #1058, opened 2020-10-06
[PATCH v3] api: pktio: add pktio capability for skip offset support

nithind1988:
api: pktio: add pktio capability for skip offset support

Current odp_pktio_skip_set() is supposed to work on all
platforms. Since this API is HW dependent as it asks ingress
parser to skip X bytes before starting to parse packet for standard
L2/L3 protocols, it may not be supported by every platforms.

Hence adding a set_op capability in packet_io for reporting
if a given platform supports it or not.

### DIFF
--- a/platform/linux-generic/odp_packet_io.c
+++ b/platform/linux-generic/odp_packet_io.c
@@ -1445,9 +1445,12 @@ int odp_pktio_capability(odp_pktio_t pktio, odp_pktio_capability_t *capa)
 	else
 		ret = single_capability(capa);
 
-	/* The same parser is used for all pktios */
-	if (ret == 0)
+	if (ret == 0) {
+		/* The same parser is used for all pktios */
 		capa->config.parser.layer = ODP_PROTO_LAYER_ALL;
+		/* Header skip is not supported */
+		capa->set_op.op.skip_offset = 0;
+	}
 
 	return ret;
 }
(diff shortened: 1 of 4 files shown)


platform/linux-generic/odp_packet_io.c
```diff
@@ -1445,9 +1445,12 @@ int odp_pktio_capability(odp_pktio_t pktio, odp_pktio_capability_t *capa)
 	else
 		ret = single_capability(capa);
 
-	/* The same parser is used for all pktios */
-	if (ret == 0)
+	if (ret == 0) {
+		/* The same parser is used for all pktios */
 		capa->config.parser.layer = ODP_PROTO_LAYER_ALL;
+		/* Header skip is supported */
+		capa->set_op.op.skip_offset = 1;
```
psavol, line 1452: It seems that classifier does not actually use skip offset when processing packets. It is set, but not used.

It would be good to implement it. Otherwise, we need to set it "unsupported" here.

nithind1988: Ok. I didn't look at implementation in detail and thought it is supported as the validation test was added. 
I'll mark it as not supported.

psavol: Also, current API spec for skip offset seems not be complete ("Number of bytes the classifier must skip"). To me it should be documented that this setting affect also parser (all input side offloads in general ?). I think the use case of this offset is to ODP packet input HW to ignore any proprietary headers (e.g. used by backplane switching) before Ethernet frame start.

nithind1988: I agree. Will update the text as above.

psavol (approved):
Please add reviewed-by tag

PR #1059, opened 2020-10-06
[PATCH v6] api: pktio: add pktout no packet refs

nithind1988:
Currently ODP spec provides support for holding references
to packets and thereby those packets are not freed by platform
after transmission if they have more than one reference.

This support is useful when using stack that needs retransmissions
like TCP or some fragmentation implementations.

Most of the other forwarding/pipeline applications will not be using
this support as they don't hold any references and forward traffic
after packet manipulations.

Since there is a performance hit in some platforms for supporting
ref count update and validation in odp_pktout_send(),
this patch adds support to indicate via pktio config if PKTIO
can optimize the fastpath for case where packets sent for
transmission will not have multiple references and platform/HW
can always free them post transmission.

In our CN9K platform, we are seeing a performance hit of ~8 % with
odp_l2fwd even thought it is not using pkt references.

### DIFF
--- a/doc/users-guide/users-guide-pktio.adoc
+++ b/doc/users-guide/users-guide-pktio.adoc
@@ -392,6 +392,11 @@ transmitted:
  * contain at least as many data bytes after L3/L4 offsets as the headers
  * indicate. Other data bytes of the packet are ignored for the checksum
  * insertion.
+ *
+ * No packet refs is an offload when set indicates that this pktio
+ * can always free the packet buffer after transmission and need not check
+ * for packet references and application will make sure that a packet
+ * transmitted via this pktio will always have only single reference.
  */
 typedef union odp_pktout_config_opt_t {
 	/** Option flags for packet output */
@@ -419,6 +424,20 @@ typedef union odp_pktout_config_opt_t {
 
 		/** Insert SCTP checksum on packet by default */
 		uint64_t sctp_chksum     : 1;
+
+		/** Packet references not used on packet output
+		 *
+		 * When set, application indicates that it will not transmit
+		 * packet references on this packet IO interface.
+		 * Since every ODP implementation supports it, it is always
+		 * ok to set this flag.
+		 *
+		 * 0: Packet references may be transmitted on the
+		 *    interface (the default value).
+		 * 1: Packet references will not be transmitted on the
+		 *    interface.
+		 */
+		uint64_t no_packet_refs  : 1;
 
 	} bit;
 
--- a/include/odp/api/spec/packet_io.h
+++ b/include/odp/api/spec/packet_io.h
@@ -414,6 +414,20 @@ typedef union odp_pktout_config_opt_t {
 		/** Insert SCTP checksum on packet by default */
 		uint64_t sctp_chksum     : 1;
 
+		/** Packet references not used on packet output
+		 *
+		 * When set, application indicates that it will not transmit
+		 * packet references on this packet IO interface.
+		 * Since every ODP implementation supports it, it is always
+		 * ok to set this flag.
+		 *
+		 * 0: Packet references may be transmitted on the
+		 *    interface (the default value).
+		 * 1: Packet references will not be transmitted on the
+		 *    interface.
+		 */
+		uint64_t no_packet_refs  : 1;
+
 	} bit;
 
 	/** All bits of the bit field structure
--- a/test/performance/odp_l2fwd.c
+++ b/test/performance/odp_l2fwd.c
@@ -790,6 +790,9 @@ static int create_pktio(const char *dev, int idx, int num_rx, int num_tx,
 		config.pktout.bit.tcp_chksum_ena  = 1;
 	}
 
+	/* Provide hint to pktio that packet references are not used */
+	config.pktout.bit.no_packet_refs = 1;
+
 	odp_pktio_config(pktio, &config);
 
 	if (gbl_args->appl.promisc_mode) {
--- a/example/l3fwd/odp_l3fwd.c
+++ b/example/l3fwd/odp_l3fwd.c
@@ -128,6 +128,10 @@ static int create_pktio(const char *name, odp_pool_t pool,
 	config.parser.layer = global->cmd_args.error_check ?
 			ODP_PROTO_LAYER_ALL :
 			ODP_PROTO_LAYER_L4;
+
+	/* Provide hint to pktio that packet references are not used */
+	config.pktout.bit.no_packet_refs = 1;
+
 	odp_pktio_config(pktio, &config);
 
 	fwd_pktio->nb_rxq = (int)capa.max_input_queues;
--- a/platform/linux-generic/odp_packet_io.c
+++ b/platform/linux-generic/odp_packet_io.c
@@ -1450,6 +1450,10 @@ int odp_pktio_capability(odp_pktio_t pktio, odp_pktio_capability_t *capa)
 		capa->config.parser.layer = ODP_PROTO_LAYER_ALL;
 		/* Header skip is not supported */
 		capa->set_op.op.skip_offset = 0;
+		/* Irrespective of whether we optimize the fast path or not,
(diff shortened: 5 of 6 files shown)


JannePeltonen:
I wonder why there is such a big performance hit. Couldn't the code just perform single atomic read and comparison to see if there are any other references than the one being sent?

Btw, I think they way dynamic packet references are defined in ODP API may be bit problematic, making the implementation harder than it would otherwise be. Maybe partly for this reason the Linux generic ODP implements dynamic references by simply copying the packet, which is not much better than having a "packet references" capability flag in the API and not having the implementation at all.

include/odp/api/spec/packet_io.h
```diff
@@ -412,6 +412,9 @@ typedef union odp_pktout_config_opt_t {
 		/** Insert SCTP checksum on packet by default */
 		uint64_t sctp_chksum     : 1;
 
+		/** Packet fast free */
+		uint64_t fast_free       : 1;
```
JannePeltonen, line 416: In my opinion this should not be called fast_free but something more descriptive (packet_references_enabled?).

If someone comes up with another flag like this, indicating, say, that there are no multi-segment packets, we do not want to add a flag called fast_free2 or something,

nithind1988:
> 
> 
> I wonder why there is such a big performance hit. Couldn't the code just perform single atomic read and comparison to see if there are any other references than the one being sent?
> 
> Btw, I think they way dynamic packet references are defined in ODP API may be bit problematic, making the implementation harder than it would otherwise be. Maybe partly for this reason the Linux generic ODP implements dynamic references by simply copying the packet, which is not much better than having a "packet references" capability flag in the API and not having the implementation at all.

We use HW to free packet after transmission and hence need to provide free/don't free info as part of Tx descriptor.

With Static references, along with providing info in Tx descriptor whether to free/don't free based on ref count,
we have to update ref count in either atomic or non-atomic method per packet. Per packet atomic operations are costly in our platform. 
Even if we choose to update it non-atomically, since our CN9K platform is a weakly memory ordered(ARMv8), we need to have a write memory barrier per packet. Normally without this, since there is no other packet update needed, we have single write memory barrier per burst.

Dynamic packet references as you mentioned needs even more logic in place than static references and hence is not so performance friendly for forwarding use cases.

include/odp/api/spec/packet_io.h
nithind1988: Ack, will change it to 'packet_refs_enabled'

JannePeltonen:
> we have to update ref count

Why do you have to do that? Wouldn't it be sufficient to do a relaxed atomic load of the refcount? If the refcount is one, no other thread is supposed to use the packet handle any more so nobody else than the current thread even cares about the refcount value at that point.

nithind1988:
> 
> 
> > we have to update ref count
> 
> Why do you have to do that? Wouldn't it be sufficient to do a relaxed atomic load of the refcount? If the refcount is one, no other thread is supposed to use the packet handle any more so nobody else than the current thread even cares about the refcount value at that point.

Agreed, we need to update refcount and have barrier only when > 1. We are already doing that but even with unlikely branch, we see the above said performance drop. It could be due to assembly code layout changes and such.

If we have this offload support, it helps us because, we generally create function pointers for different combinations of offloads.
Each function pointer has code only for the offloads it supports. Like say, 
# xmit_pkts_l3l4csum() is a function pointer that is choosen when only Tx L3/L4 checksum is enabled 
# xmit_pkts_no_offloads() is choosen, when there is no Tx offload enabled(l2fwd). 

We choose function pointer based on offloads configured on PKTIO.

JannePeltonen:
> Agreed, we need to update refcount and have barrier only when > 1. We are already doing that but even with unlikely branch, we see the above said performance drop. It could be due to assembly code layout changes and such.

Ok, I was just wondering.

I think making it possible for an application to tell ODP that it won't use packet references makes sense, so please send an updated version. I wonder if something similar should be done with IPsec or maybe even globally.

include/odp/api/spec/packet_io.h
```diff
@@ -412,6 +412,9 @@ typedef union odp_pktout_config_opt_t {
 		/** Insert SCTP checksum on packet by default */
 		uint64_t sctp_chksum     : 1;
 
+		/** Packet references disabled */
+		uint64_t no_packet_refs  : 1;
```
psavol, line 416: This part needs more description here in the API spec (user guide is not spec but just additional information how to operate the API). Something like this would be needed:

"Packet references not used on packet output

When set, application indicates that it will not transmit packet references on this packet IO interface.

0: Packet references may be transmitted on the interface (the default value).
1: Packet references will not be transmitted on the interface "

doc/users-guide/users-guide-pktio.adoc
```diff
@@ -390,6 +390,11 @@ transmitted:
  * contain at least as many data bytes after L3/L4 offsets as the headers
  * indicate. Other data bytes of the packet are ignored for the checksum
  * insertion.
+ *
+ * No packet refs is an offload when set indicates that this pktio
+ * can always free the packet buffer after transmission and need not check
+ * for packet references and application will make sure that a packet
+ * transmitted via this pktio will always have only single reference.
```
psavol, line 397: Please, split user doc changes and API changes into separate patches.

test/performance/odp_l2fwd.c
```diff
@@ -789,6 +789,10 @@ static int create_pktio(const char *dev, int idx, int num_rx, int num_tx,
 		config.pktout.bit.tcp_chksum_ena  = 1;
 	}
 
+	/* Provide hint to pktio that packet references are not used */
+	if (pktio_capa.config.pktout.bit.no_packet_refs)
```
JannePeltonen, line 793: The capability check here is required by the current API, but I feel it is a bit silly that it has to be done. This setting is just making life easier to an ODP implementation and every implementation obviously has to support sending non-referenced packets anyway.

Maybe we could add a note in odp_pktout_config_opt_t that it is always ok to set the no_packet_refs flag, i.e. every ODP implementation supports it?

nithind1988: Ok. I'll add a comment on top of odp_pktout_config_opt_t:no_packet_refs field mentioning the same.

psavol:
Also some (existing) validation test needs to set this flag, so that we verify that API header has the field.

nithind1988:
> 
> 
> Also some (existing) validation test needs to set this flag, so that we verify that API header has the field.

You mean, in pktio_main ?
It is however already part of test/performance/odp_l2fwd_run.sh  test

include/odp/api/spec/packet_io.h
```diff
@@ -412,6 +412,20 @@ typedef union odp_pktout_config_opt_t {
 		/** Insert SCTP checksum on packet by default */
 		uint64_t sctp_chksum     : 1;
 
+		/** Packet references not used on packet output
+		 *
+		 * When set, application indicates that it will not transmit
+		 * packet references on this packet IO interface.
+		 * Since every ODP implementation supports it, it is always
+		 * ok to set this flag when needed.
```
JannePeltonen, line 420: I would like to have the "when needed" part removed as it implies that the flag is sometimes _needed_ and in (only?) those particular cases it is ok to set it. But the flag is always optional and never strictly needed.

JannePeltonen:
> > Also some (existing) validation test needs to set this flag, so that we verify that API header has the field.

> You mean, in pktio_main ?
> It is however already part of test/performance/odp_l2fwd_run.sh test

Yes, Petri meant that it would be good to have it as part of API validation test suite, not just in the performance test. So if you add explicit setting of the flag to zero in there, the test stays otherwise unchanged but it checks that the new field really exists in the set of ODP headers files that the test is being compiled against.

JannePeltonen:
If you fix the two minor open issues, you could add my reviewed-by tag already at the same time as this otherwise looks ok to me.

JannePeltonen (approved):

PR #1060, opened 2020-10-07
[PATCH v14] api: ipsec: add IPsec stats

anoobj:
Add support for IPsec stats collection and reporting. This would allow
applications to use ODP APIs for stats collection and retrieval. Also
implementations can optimize stats reporting, which would otherwise
involve inspecting every packet and updating stats atomically.

Stats collection can be disabled using the 'stats_en' option added in
odp_ipsec_config_t.

### DIFF
--- a/include/odp/api/spec/ipsec.h
+++ b/include/odp/api/spec/ipsec.h
@@ -320,6 +320,14 @@ typedef struct odp_ipsec_config_t {
 	/** IPSEC outbound processing configuration */
 	odp_ipsec_outbound_config_t outbound;
 
+	/** Enable stats collection
+	 *
+	 *  Default value is false (stats collection disabled).
+	 *
+	 *  @see odp_ipsec_stats(), odp_ipsec_stats_multi()
+	 */
+	odp_bool_t stats_en;
+
 } odp_ipsec_config_t;
 
 /**
@@ -771,6 +779,35 @@ typedef struct odp_ipsec_sa_param_t {
 	};
 
 } odp_ipsec_sa_param_t;
+
+/**
+ * IPSEC stats content
+ */
+typedef struct odp_ipsec_stats_t {
+	/** Number of packets processed successfully */
+	uint64_t success;
+
+	/** Number of packets with protocol errors */
+	uint64_t proto_err;
+
+	/** Number of packets with authentication errors */
+	uint64_t auth_err;
+
+	/** Number of packets with antireplay check failures */
+	uint64_t antireplay_err;
+
+	/** Number of packets with algorithm errors */
+	uint64_t alg_err;
+
+	/** Number of packes with MTU errors */
+	uint64_t mtu_err;
+
+	/** Number of packets with hard lifetime(bytes) expired */
+	uint64_t hard_exp_bytes_err;
+
+	/** Number of packets with hard lifetime(packets) expired */
+	uint64_t hard_exp_pkts_err;
+} odp_ipsec_stats_t;
 
 /**
  * Query IPSEC capabilities
@@ -1642,6 +1679,29 @@ void odp_ipsec_print(void);
  * Print implementation-defined IPSEC SA debug information to the ODP log.
  */
 void odp_ipsec_sa_print(odp_ipsec_sa_t sa);
+
+/**
+ * Get IPSEC stats for the IPSEC SA handle
+ *
+ * @param          sa       IPSEC SA handle
+ * @param[out]     stats    Stats output
+ *
+ * @retval 0 on success
+ * @retval <0 on failure
+ */
+int odp_ipsec_stats(odp_ipsec_sa_t sa, odp_ipsec_stats_t *stats);
+
+/**
+ * Get IPSEC stats for multiple IPSEC SA handles
+ *
+ * @param          sa       Array of IPSEC SA handles
+ * @param[out]     stats    Stats array for output
+ * @param          num      Number of SA handles
+ *
+ * @retval 0 on success
+ * @retval <0 on failure
+ */
+int odp_ipsec_stats_multi(odp_ipsec_sa_t sa[], odp_ipsec_stats_t stats[], int num);
 
 /**
  * @}
--- a/platform/linux-generic/odp_ipsec.c
+++ b/platform/linux-generic/odp_ipsec.c
@@ -24,6 +24,7 @@
 #include <protocols/ipsec.h>
 #include <protocols/udp.h>
 
+#include <errno.h>
 #include <string.h>
 
 static odp_ipsec_config_t *ipsec_config;
@@ -217,6 +218,7 @@ void odp_ipsec_config_init(odp_ipsec_config_t *config)
 	config->inbound.default_queue = ODP_QUEUE_INVALID;
 	config->inbound.lookup.min_spi = 0;
 	config->inbound.lookup.max_spi = UINT32_MAX;
+	config->stats_en = false;
 }
 
 int odp_ipsec_config(const odp_ipsec_config_t *config)
@@ -665,6 +667,37 @@ static int ipsec_in_ah_post(odp_packet_t pkt,
 	return 0;
 }
 
+static void
+ipsec_sa_err_stats_update(ipsec_sa_t *sa, odp_ipsec_op_status_t *status)
+{
+	if (odp_likely(ODP_IPSEC_OK == status->error.all))
+		return;
+
+	if (NULL == sa)
+		return;
+
+	if (status->error.proto)
+		odp_atomic_inc_u64(&sa->stats.proto_err);
+
+	if (status->error.auth)
+		odp_atomic_inc_u64(&sa->stats.auth_err);
+
+	if (status->error.antireplay)
+		odp_atomic_inc_u64(&sa->stats.antireplay_err);
+
+	if (status->error.alg)
+		odp_atomic_inc_u64(&sa->stats.alg_err);
+
+	if (status->error.mtu)
+		odp_atomic_inc_u64(&sa->stats.mtu_err);
+
+	if (status->error.hard_exp_bytes)
+		odp_atomic_inc_u64(&sa->stats.hard_exp_bytes_err);
+
+	if (status->error.hard_exp_packets)
+		odp_atomic_inc_u64(&sa->stats.hard_exp_pkts_err);
+}
+
 static ipsec_sa_t *ipsec_in_single(odp_packet_t pkt,
 				   odp_ipsec_sa_t sa,
 				   odp_packet_t *pkt_out,
@@ -702,7 +735,7 @@ static ipsec_sa_t *ipsec_in_single(odp_packet_t pkt,
 	if (rc < 0 ||
 	    state.ip_tot_len + state.ip_offset > odp_packet_len(pkt)) {
 		status->error.alg = 1;
-		goto err;
+		goto exit;
 	}
 
 	/* Check IP header for IPSec protocols and look it up */
@@ -713,18 +746,18 @@ static ipsec_sa_t *ipsec_in_single(odp_packet_t pkt,
 		rc = ipsec_in_ah(&pkt, &state, &ipsec_sa, sa, &param, status);
 	} else {
 		status->error.proto = 1;
-		goto err;
+		goto exit;
 	}
 	if (rc < 0)
-		goto err;
+		goto exit;
 
 	if (_odp_ipsec_sa_replay_precheck(ipsec_sa,
 					  state.in.seq_no,
 					  status) < 0)
-		goto err;
+		goto exit;
 
 	if (_odp_ipsec_sa_stats_precheck(ipsec_sa, status) < 0)
-		goto err;
+		goto exit;
 
 	param.session = ipsec_sa->session;
 
@@ -732,14 +765,14 @@ static ipsec_sa_t *ipsec_in_single(odp_packet_t pkt,
 	if (rc < 0) {
 		ODP_DBG("Crypto failed\n");
 		status->error.alg = 1;
-		goto err;
+		goto exit;
 	}
 
 	rc = odp_crypto_result(&crypto, pkt);
 	if (rc < 0) {
 		ODP_DBG("Crypto failed\n");
 		status->error.alg = 1;
-		goto err;
+		goto exit;
 	}
 
 	if (!crypto.ok) {
@@ -751,18 +784,18 @@ static ipsec_sa_t *ipsec_in_single(odp_packet_t pkt,
 		else
 			status->error.alg = 1;
 
-		goto err;
+		goto exit;
 	}
 
 	if (_odp_ipsec_sa_replay_update(ipsec_sa,
 					state.in.seq_no,
 					status) < 0)
-		goto err;
+		goto exit;
 
-	if (_odp_ipsec_sa_stats_update(ipsec_sa,
-				       state.stats_length,
-				       status) < 0)
-		goto err;
+	if (_odp_ipsec_sa_lifetime_update(ipsec_sa,
+					  state.stats_length,
+					  status) < 0)
+		goto post_lifetime_err_cnt_update;
 
 	state.ip = odp_packet_l3_ptr(pkt, NULL);
 
@@ -774,12 +807,12 @@ static ipsec_sa_t *ipsec_in_single(odp_packet_t pkt,
 		rc = -1;
 	if (rc < 0) {
 		status->error.proto = 1;
-		goto err;
+		goto post_lifetime_err_cnt_update;
 	}
 
 	if (odp_packet_trunc_tail(&pkt, state.in.trl_len, NULL, NULL) < 0) {
 		status->error.alg = 1;
-		goto err;
+		goto post_lifetime_err_cnt_update;
 	}
 	state.ip_tot_len -= state.in.trl_len;
 
@@ -793,7 +826,7 @@ static ipsec_sa_t *ipsec_in_single(odp_packet_t pkt,
 					  state.in.hdr_len,
 					  NULL, NULL) < 0) {
 			status->error.alg = 1;
-			goto err;
+			goto post_lifetime_err_cnt_update;
 		}
 		state.ip_tot_len -= state.ip_hdr_len + state.in.hdr_len;
 		if (_ODP_IPPROTO_IPIP == state.ip_next_hdr) {
@@ -807,7 +840,7 @@ static ipsec_sa_t *ipsec_in_single(odp_packet_t pkt,
 			state.is_ipv6 = 0;
 		} else {
 			status->error.proto = 1;
-			goto err;
+			goto post_lifetime_err_cnt_update;
 		}
 	} else {
 		odp_packet_move_data(pkt, state.in.hdr_len, 0,
@@ -815,7 +848,7 @@ static ipsec_sa_t *ipsec_in_single(odp_packet_t pkt,
 		if (odp_packet_trunc_head(&pkt, state.in.hdr_len,
 					  NULL, NULL) < 0) {
 			status->error.alg = 1;
-			goto err;
+			goto post_lifetime_err_cnt_update;
 		}
 		state.ip_tot_len -= state.in.hdr_len;
 	}
@@ -840,7 +873,7 @@ static ipsec_sa_t *ipsec_in_single(odp_packet_t pkt,
 			ipv6hdr->hop_limit -= ipsec_sa->dec_ttl;
 	} else if (state.ip_next_hdr != _ODP_IPPROTO_NO_NEXT) {
 		status->error.proto = 1;
-		goto err;
+		goto post_lifetime_err_cnt_update;
 	}
 
 	if (_ODP_IPPROTO_NO_NEXT == state.ip_next_hdr &&
@@ -863,8 +896,17 @@ static ipsec_sa_t *ipsec_in_single(odp_packet_t pkt,
 		odp_packet_parse(pkt, state.ip_offset, &parse_param);
 	}
 
-err:
+	goto exit;
+
+post_lifetime_err_cnt_update:
+	if (ipsec_config->stats_en)
+		odp_atomic_inc_u64(&ipsec_sa->stats.post_lifetime_err_pkts);
+
+exit:
 	*pkt_out = pkt;
+
+	if (ipsec_config->stats_en)
+		ipsec_sa_err_stats_update(ipsec_sa, status);
 
 	return ipsec_sa;
 }
@@ -1522,7 +1564,7 @@ static ipsec_sa_t *ipsec_out_single(odp_packet_t pkt,
 			rc = -1;
 		if (rc < 0) {
 			status->error.alg = 1;
-			goto err;
+			goto exit;
 		}
 
 		ipsec_out_checksums(pkt, &state);
@@ -1540,7 +1582,7 @@ static ipsec_sa_t *ipsec_out_single(odp_packet_t pkt,
 	}
 	if (rc < 0) {
 		status->error.alg = 1;
-		goto err;
+		goto exit;
 	}
 
 	if (ODP_IPSEC_ESP == ipsec_sa->proto) {
@@ -1550,16 +1592,16 @@ static ipsec_sa_t *ipsec_out_single(odp_packet_t pkt,
 		rc = ipsec_out_ah(&pkt, &state, ipsec_sa, &param, status, mtu);
 	} else {
 		status->error.alg = 1;
-		goto err;
+		goto exit;
 	}
 	if (rc < 0)
-		goto err;
+		goto exit;
 
 	/* No need to run precheck here, we know that packet is authentic */
-	if (_odp_ipsec_sa_stats_update(ipsec_sa,
-				       state.stats_length,
-				       status) < 0)
-		goto err;
+	if (_odp_ipsec_sa_lifetime_update(ipsec_sa,
+					  state.stats_length,
+					  status) < 0)
+		goto post_lifetime_err_cnt_update;
 
 	param.session = ipsec_sa->session;
 
@@ -1579,19 +1621,19 @@ static ipsec_sa_t *ipsec_out_single(odp_packet_t pkt,
 	if (rc < 0) {
 		ODP_DBG("Crypto failed\n");
 		status->error.alg = 1;
-		goto err;
+		goto post_lifetime_err_cnt_update;
 	}
 
 	rc = odp_crypto_result(&crypto, pkt);
 	if (rc < 0) {
 		ODP_DBG("Crypto failed\n");
 		status->error.alg = 1;
-		goto err;
+		goto post_lifetime_err_cnt_update;
 	}
 
 	if (!crypto.ok) {
 		status->error.alg = 1;
-		goto err;
+		goto post_lifetime_err_cnt_update;
 	}
 
 	/* Finalize the IP header */
@@ -1600,7 +1642,13 @@ static ipsec_sa_t *ipsec_out_single(odp_packet_t pkt,
 	else if (ODP_IPSEC_AH == ipsec_sa->proto)
 		ipsec_out_ah_post(&state, pkt);
 
-err:
+	goto exit;
+
+post_lifetime_err_cnt_update:
+	if (ipsec_config->stats_en)
+		odp_atomic_inc_u64(&ipsec_sa->stats.post_lifetime_err_pkts);
+
+exit:
 	*pkt_out = pkt;
 	return ipsec_sa;
 }
@@ -1699,6 +1747,9 @@ int odp_ipsec_out(const odp_packet_t pkt_in[], int num_in,
 		memset(result, 0, sizeof(*result));
 		result->status = status;
 		result->sa = ipsec_sa->ipsec_sa_hdl;
+
+		if (ipsec_config->stats_en)
+			ipsec_sa_err_stats_update(ipsec_sa, &status);
 
 		pkt_out[out_pkt] = pkt;
 		in_pkt++;
@@ -1804,6 +1855,9 @@ int odp_ipsec_out_enq(const odp_packet_t pkt_in[], int num_in,
 		result->status = status;
 		result->sa = ipsec_sa->ipsec_sa_hdl;
 		queue = ipsec_sa->queue;
+
+		if (ipsec_config->stats_en)
+			ipsec_sa_err_stats_update(ipsec_sa, &status);
 
 		if (odp_queue_enq(queue, odp_ipsec_packet_to_event(pkt))) {
 			odp_packet_free(pkt);
@@ -1942,6 +1996,9 @@ int odp_ipsec_out_inline(const odp_packet_t pkt_in[], int num_in,
 		} else {
 			odp_queue_t queue;
 err:
+			if (ipsec_config->stats_en)
+				ipsec_sa_err_stats_update(ipsec_sa, &status);
+
 			packet_subtype_set(pkt, ODP_EVENT_PACKET_IPSEC);
 			result = ipsec_pkt_result(pkt);
 			memset(result, 0, sizeof(*result));
@@ -1987,6 +2044,33 @@ odp_packet_t odp_ipsec_packet_from_event(odp_event_t ev)
 odp_event_t odp_ipsec_packet_to_event(odp_packet_t pkt)
 {
 	return odp_packet_to_event(pkt);
+}
+
+int odp_ipsec_stats(odp_ipsec_sa_t sa, odp_ipsec_stats_t *stats)
+{
+	ipsec_sa_t *ipsec_sa;
+
+	if (ODP_IPSEC_SA_INVALID == sa)
+		return -EINVAL;
+
+	if (!ipsec_config->stats_en)
+		return -ENOTSUP;
+
+	ODP_ASSERT(NULL != stats);
+
+	ipsec_sa = _odp_ipsec_sa_entry_from_hdl(sa);
+	ODP_ASSERT(NULL != ipsec_sa);
+
+	stats->success = _odp_ipsec_sa_stats_pkts(ipsec_sa);
+	stats->proto_err = odp_atomic_load_u64(&ipsec_sa->stats.proto_err);
+	stats->auth_err = odp_atomic_load_u64(&ipsec_sa->stats.auth_err);
+	stats->antireplay_err = odp_atomic_load_u64(&ipsec_sa->stats.antireplay_err);
+	stats->alg_err = odp_atomic_load_u64(&ipsec_sa->stats.alg_err);
+	stats->mtu_err = odp_atomic_load_u64(&ipsec_sa->stats.mtu_err);
+	stats->hard_exp_bytes_err = odp_atomic_load_u64(&ipsec_sa->stats.hard_exp_bytes_err);
+	stats->hard_exp_pkts_err = odp_atomic_load_u64(&ipsec_sa->stats.hard_exp_pkts_err);
+
+	return 0;
 }
 
 int _odp_ipsec_init_global(void)
--- a/platform/linux-generic/odp_ipsec_sad.c
+++ b/platform/linux-generic/odp_ipsec_sad.c
@@ -61,7 +61,7 @@ typedef struct sa_thread_local_s {
 	 * Packets that can be processed in this thread before looking at
 	 * the SA-global packet counter and checking hard and soft limits.
 	 */
-	uint32_t packet_quota;
+	odp_atomic_u32_t packet_quota;
 	/*
 	 * Bytes that can be processed in this thread before looking at
 	 * at the SA-global byte counter and checking hard and soft limits.
@@ -125,7 +125,7 @@ static void init_sa_thread_local(ipsec_sa_t *sa)
 
 	for (n = 0; n < ODP_THREAD_COUNT_MAX; n++) {
 		sa_tl = &ipsec_sa_tbl->per_thread[n].sa[sa->ipsec_sa_idx];
-		sa_tl->packet_quota = 0;
+		odp_atomic_init_u32(&sa_tl->packet_quota, 0);
 		sa_tl->byte_quota = 0;
 		sa_tl->lifetime_status.all = 0;
 	}
@@ -458,6 +458,15 @@ odp_ipsec_sa_t odp_ipsec_sa_create(const odp_ipsec_sa_param_t *param)
 	ipsec_sa->hard_limit_bytes = param->lifetime.hard_limit.bytes;
 	ipsec_sa->hard_limit_packets = param->lifetime.hard_limit.packets;
 
+	odp_atomic_init_u64(&ipsec_sa->stats.proto_err, 0);
+	odp_atomic_init_u64(&ipsec_sa->stats.auth_err, 0);
+	odp_atomic_init_u64(&ipsec_sa->stats.antireplay_err, 0);
+	odp_atomic_init_u64(&ipsec_sa->stats.alg_err, 0);
+	odp_atomic_init_u64(&ipsec_sa->stats.mtu_err, 0);
+	odp_atomic_init_u64(&ipsec_sa->stats.hard_exp_bytes_err, 0);
+	odp_atomic_init_u64(&ipsec_sa->stats.hard_exp_pkts_err, 0);
+	odp_atomic_init_u64(&ipsec_sa->stats.post_lifetime_err_pkts, 0);
+
 	if (ODP_IPSEC_MODE_TUNNEL == ipsec_sa->mode &&
 	    ODP_IPSEC_DIR_OUTBOUND == param->dir) {
 		if (ODP_IPSEC_TUNNEL_IPV4 == param->outbound.tunnel.type) {
@@ -781,17 +790,19 @@ int _odp_ipsec_sa_stats_precheck(ipsec_sa_t *ipsec_sa,
 	return rc;
 }
 
-int _odp_ipsec_sa_stats_update(ipsec_sa_t *ipsec_sa, uint32_t len,
-			       odp_ipsec_op_status_t *status)
+int _odp_ipsec_sa_lifetime_update(ipsec_sa_t *ipsec_sa, uint32_t len,
+				  odp_ipsec_op_status_t *status)
 {
 	sa_thread_local_t *sa_tl = ipsec_sa_thread_local(ipsec_sa);
 	uint64_t packets, bytes;
+	uint32_t tl_pkt_quota;
 
-	if (odp_unlikely(sa_tl->packet_quota == 0)) {
+	tl_pkt_quota = odp_atomic_load_u32(&sa_tl->packet_quota);
+	if (odp_unlikely(tl_pkt_quota == 0)) {
 		packets = odp_atomic_fetch_add_u64(&ipsec_sa->hot.packets,
 						   SA_LIFE_PACKETS_PREALLOC);
 		packets += SA_LIFE_PACKETS_PREALLOC;
-		sa_tl->packet_quota += SA_LIFE_PACKETS_PREALLOC;
+		tl_pkt_quota += SA_LIFE_PACKETS_PREALLOC;
 
 		if (ipsec_sa->soft_limit_packets > 0 &&
 		    packets >= ipsec_sa->soft_limit_packets)
@@ -801,7 +812,8 @@ int _odp_ipsec_sa_stats_update(ipsec_sa_t *ipsec_sa, uint32_t len,
 		    packets >= ipsec_sa->hard_limit_packets)
 			sa_tl->lifetime_status.error.hard_exp_packets = 1;
 	}
-	sa_tl->packet_quota--;
+	tl_pkt_quota--;
+	odp_atomic_store_u32(&sa_tl->packet_quota, tl_pkt_quota);
 
 	if (odp_unlikely(sa_tl->byte_quota < len)) {
 		bytes = odp_atomic_fetch_add_u64(&ipsec_sa->hot.bytes,
@@ -909,4 +921,31 @@ uint16_t _odp_ipsec_sa_alloc_ipv4_id(ipsec_sa_t *ipsec_sa)
 
 	/* No need to convert to BE: ID just should not be duplicated */
 	return tl->next_ipv4_id++;
+}
+
+uint64_t _odp_ipsec_sa_stats_pkts(ipsec_sa_t *sa)
+{
+	uint64_t tl_pkt_quota = 0;
+	sa_thread_local_t *sa_tl;
+	int n;
+
+	/*
+	 * Field 'hot.packets' tracks SA lifetime. The same field is being used
+	 * to track the number of success packets.
+	 *
+	 * SA lifetime tracking implements a per thread packet quota to allow
+	 * less frequent updates to the hot field. The per thread quota need
+	 * to be decremented. In addition, SA lifetime gets consumed for any
+	 * errors occurring after lifetime check is done. Those packets also
+	 * need to be accounted for.
+	 */
+
+	for (n = 0; n < ODP_THREAD_COUNT_MAX; n++) {
+		sa_tl = &ipsec_sa_tbl->per_thread[n].sa[sa->ipsec_sa_idx];
+		tl_pkt_quota += odp_atomic_load_u32(&sa_tl->packet_quota);
+	}
+
+	return odp_atomic_load_u64(&sa->hot.packets)
+	       - odp_atomic_load_u64(&sa->stats.post_lifetime_err_pkts)
+	       - tl_pkt_quota;
 }
--- a/test/validation/api/ipsec/ipsec.c
+++ b/test/validation/api/ipsec/ipsec.c
@@ -591,7 +591,7 @@ static int ipsec_send_in_one(const ipsec_test_part *part,
 	pkt = ipsec_packet(part->pkt_in);
 
 	memset(&param, 0, sizeof(param));
-	if (!part->lookup) {
+	if (!part->flags.lookup) {
 		param.num_sa = 1;
 		param.sa = &sa;
 	} else {
--- a/test/validation/api/ipsec/ipsec.c
+++ b/test/validation/api/ipsec/ipsec.c
@@ -1,5 +1,6 @@
 /* Copyright (c) 2017-2018, Linaro Limited
  * Copyright (c) 2019-2020, Nokia
+ * Copyright (c) 2020, Marvell
  * All rights reserved.
  *
  * SPDX-License-Identifier:	 BSD-3-Clause
@@ -332,12 +333,6 @@ int ipsec_check_esp_aes_gcm_128(void)
 				ODP_AUTH_ALG_AES_GCM, 0);
 }
 
-int ipsec_check_esp_aes_gcm_192(void)
-{
-	return  ipsec_check_esp(ODP_CIPHER_ALG_AES_GCM, 192,
-				ODP_AUTH_ALG_AES_GCM, 0);
-}
-
 int ipsec_check_esp_aes_gcm_256(void)
 {
 	return  ipsec_check_esp(ODP_CIPHER_ALG_AES_GCM, 256,
@@ -378,24 +373,6 @@ int ipsec_check_esp_null_aes_gmac_256(void)
 {
 	return  ipsec_check_esp(ODP_CIPHER_ALG_NULL, 0,
 				ODP_AUTH_ALG_AES_GMAC, 256);
-}
-
-int ipsec_check_esp_aes_ccm_128(void)
-{
-	return  ipsec_check_esp(ODP_CIPHER_ALG_AES_CCM, 128,
-				ODP_AUTH_ALG_AES_CCM, 0);
-}
-
-int ipsec_check_esp_aes_ccm_192(void)
-{
-	return  ipsec_check_esp(ODP_CIPHER_ALG_AES_CCM, 192,
-				ODP_AUTH_ALG_AES_CCM, 0);
-}
-
-int ipsec_check_esp_aes_ccm_256(void)
-{
-	return  ipsec_check_esp(ODP_CIPHER_ALG_AES_CCM, 256,
-				ODP_AUTH_ALG_AES_CCM, 0);
 }
 
 int ipsec_check_esp_chacha20_poly1305(void)
--- a/test/validation/api/ipsec/ipsec.c
+++ b/test/validation/api/ipsec/ipsec.c
@@ -584,7 +584,7 @@ static int ipsec_send_in_one(const ipsec_test_part *part,
 			     odp_packet_t *pkto)
 {
 	odp_ipsec_in_param_t param;
-	int num_out = part->out_pkt;
+	int num_out = part->num_pkt;
 	odp_packet_t pkt;
 	int i;
 
@@ -600,10 +600,10 @@ static int ipsec_send_in_one(const ipsec_test_part *part,
 	}
 
 	if (ODP_IPSEC_OP_MODE_SYNC == suite_context.inbound_op_mode) {
-		CU_ASSERT_EQUAL(part->out_pkt, odp_ipsec_in(&pkt, 1,
+		CU_ASSERT_EQUAL(part->num_pkt, odp_ipsec_in(&pkt, 1,
 							    pkto, &num_out,
 							    &param));
-		CU_ASSERT_EQUAL(num_out, part->out_pkt);
+		CU_ASSERT_EQUAL(num_out, part->num_pkt);
 		CU_ASSERT(odp_packet_subtype(*pkto) == ODP_EVENT_PACKET_IPSEC);
 	} else if (ODP_IPSEC_OP_MODE_ASYNC == suite_context.inbound_op_mode) {
 		num_out = odp_ipsec_in_enq(&pkt, 1, &param);
@@ -653,7 +653,7 @@ static int ipsec_send_in_one(const ipsec_test_part *part,
 						odp_event_types(ev, &subtype));
 				CU_ASSERT_EQUAL(ODP_EVENT_PACKET_BASIC,
 						subtype);
-				CU_ASSERT(part->out[i].status.error.sa_lookup);
+				CU_ASSERT(part->in[i].status.error.sa_lookup);
 
 				pkto[i++] = odp_ipsec_packet_from_event(ev);
 				continue;
@@ -666,7 +666,7 @@ static int ipsec_send_in_one(const ipsec_test_part *part,
 						odp_event_types(ev, &subtype));
 				CU_ASSERT_EQUAL(ODP_EVENT_PACKET_IPSEC,
 						subtype);
-				CU_ASSERT(!part->out[i].status.error.sa_lookup);
+				CU_ASSERT(!part->in[i].status.error.sa_lookup);
 
 				pkto[i] = odp_ipsec_packet_from_event(ev);
 				CU_ASSERT(odp_packet_subtype(pkto[i]) ==
@@ -685,7 +685,7 @@ static int ipsec_send_out_one(const ipsec_test_part *part,
 			      odp_packet_t *pkto)
 {
 	odp_ipsec_out_param_t param;
-	int num_out = part->out_pkt;
+	int num_out = part->num_pkt;
 	odp_packet_t pkt;
 	int i;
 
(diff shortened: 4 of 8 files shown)


include/odp/api/spec/ipsec.h
```diff
@@ -762,6 +768,38 @@ typedef struct odp_ipsec_sa_param_t {
...
+	uint64_t num_alg_err;
+
+	/** Number of packes with MTU errors */
+	uint64_t num_mtu_err;
+
+	/** Number of packets with hard lifetime(bytes) expired */
+	uint64_t num_hard_exp_bytes_err;
+
+	/** Number of packets with hard lifetime(packets) expired */
+	uint64_t num_hard_exp_pkts_err;
```
psavol, line 800: num_ prefix is not needed for stat counters

include/odp/api/spec/ipsec.h
```diff
@@ -310,6 +310,12 @@ typedef struct odp_ipsec_config_t {
 	/** IPSEC outbound processing configuration */
 	odp_ipsec_outbound_config_t outbound;
 
+	/** Enable stats collection
+	 *
+	 * @see odp_ipsec_stats(), odp_ipsec_stats_multi()
```
JannePeltonen, line 315: This could say that the default value is false.

include/odp/api/spec/ipsec.h
```diff
@@ -1601,6 +1639,25 @@ int odp_ipsec_sa_mtu_update(odp_ipsec_sa_t sa, uint32_t mtu);
...
 void *odp_ipsec_sa_context(odp_ipsec_sa_t sa);
 
+/**
+ * Get IPSEC stats for the IPSEC SA handle
+ *
+ * @param          sa       IPSEC SA handle
+ * @param[out]     stats    Stats output
+ *                          Stats returned will be 0 for invalid SA
+ */
+void odp_ipsec_stats(odp_ipsec_sa_t sa, odp_ipsec_stats_t *stats);
```
JannePeltonen, line 1649: Maybe both of these functions should return success status?

JannePeltonen:
Other than the minor issues already commented, this looks quite ok to me.

include/odp/api/spec/ipsec.h
anoobj: Agreed. Will make this change with next version.

include/odp/api/spec/ipsec.h
anoobj: Agreed. Will make this change with next version.

include/odp/api/spec/ipsec.h
anoobj: Agreed. What would be better,
1. Return status (success/failure)
2. Return number of SAs processed (0 indicating failure)

JannePeltonen: success/failure would seem more natural since the SAs do not form an ordered batch like packets do in other APIs. But it is still useful to be able to have the lookup fail for some SAs. So maybe it should work so that if return status is 0, then stats could be returned at least for some SAs (and others would show zeroes)?

anoobj: I've updated this in v3.

JannePeltonen:

include/odp/api/spec/ipsec.h
psavol: I think return value of the stats_multi call should tell which statistics where successfully updated. Either all/nothing (0 vs -1), or N = number of stats successfully written (starting from stats[0]). Otherwise, application could not tell if stats of all zeros did not receive any packets, or if there was an error reading those stats.

Usually the call should always succeed (application gives valid SA handles). So, in practice the difference would be that retval N would report the first failing SA vs 0/-1 would just tell that some SA failed.

I'd vote retval N. It's consistent with other multi APIs (although here application would not retry a failing stats read) and has the extra ability to return (the first) failing SA.

Another option would be to add a success / failure flag into stats struct, which application could check if N != num. This would tell all failing SAs (normally N==num and no need to check the flag).

JannePeltonen: IMHO, success/failure flag in the SA stats struct would be better than having the return value tell that the _first_ N SA stats were ok. The SAs are not ordered like packets are, so problematic SA *) in the beginning would prevent getting stats of the later SAs unless the app reorders the SA request list. Such reordering or removal of packets in packet APIs happens naturally but not, in my view, for SAs in this SA stats case.

*) Actually, if the only failure case would be invalid SA handle, then we could just tell that invalid SA handles are not ok or have the whole function fail without returning anything until the application gets its act together. But maybe there could be other, transient, issues in fetching the stats (e.g. when SAs are being rapidly created and destroyed) so the success/failure flag could be useful.

anoobj: Initial plan was to have all stats to say 0 for an invalid SA. But looking at the discussion, I think it's better to fail the operation in such cases. I can remove the line,
`
Stats returned will be 0 for invalid SA`

And state that -EINVAL would be returned in case any SA is invalid. @psavol do you agree to this approach?

psavol: Agree that this stats_multi call could just either succeed (return 0), or fail (<0). Application should pass only valid SAs to the call, and thus a failure should be very rare result.

For failure, specify <0 instead of a specific value (usually we return -1, but spec says <0).

anoobj: @psavol Can you take a look at latest patch? I had made this change (returning 0 for success & <0 for failure). I'll submit a new version to state that application should pass only valid SAs to this API.

JannePeltonen: I do not think there is need to state that an application should pass only valid SAs to this API as that can be expected anyway. But the point is that if the call (single or multi-SA variant) succeeds, then stats are returned for all SAs. The possibility to return stats only for some SAs is thus removed, so these sentences should be removed: "Stats returned will be 0 for any invalid SA".

anoobj: @JannePeltonen Agreed. I'll make this change in next version. Shall I add your "Reviewed-by" tag in next version?

JannePeltonen: Yes, please add the tag.

psavol:
API spec is OK.

anoobj:
> 
> 
> v3 API spec is OK. Reviewed-by: Petri Savolainen [petri.savolainen@nokia.com](mailto:petri.savolainen@nokia.com)
> 
> Do you plan to do implementation and validation tests as well?

We do have plans to introduce validation tests. What is the strategy you recommend (with regards to implementation & validation tests)? Can that be taken up after we merge in the spec change?

psavol:
>  Can that be taken up after we merge in the spec change

We should include at least minimal implementation and test cases with the API change. Could you provide it, if it's e.g. a days job to complete the PR with implementation and test case.

anoobj:
> 
> 
> > Can that be taken up after we merge in the spec change
> 
> We should include at least minimal implementation and test cases with the API change. Could you provide it, if it's e.g. a days job to complete the PR with implementation and test case.

Let me try to come up with a minimal set of validation tests. As I was implementing, came across a minor change required in the spec. Will comment in the code.

include/odp/api/spec/ipsec.h
```diff
@@ -762,6 +770,38 @@ typedef struct odp_ipsec_sa_param_t {
...
+ */
+typedef struct odp_ipsec_stats_t {
+	/** Number of packets processed successfully */
+	uint64_t success;
+
+	/** Number of packets with protocol errors */
+	uint64_t proto_err;
+
+	/** Number of packets with SA lookup failures */
+	uint64_t sa_lookup_err;
```
anoobj, line 784: @JannePeltonen I think we should remove sa_lookup_err from this list. Reason, we are collecting stats per SA. An SA lookup error cannot be associated to an SA. Can you confirm I can make this change?

JannePeltonen: Yes, of course. Obviously lookup failure cannot be in per-SA stats and somehow we missed it.

anoobj:
@JannePeltonen @psavol I've submitted a basic linux generic implementation and added basic tests for the stats of success packets. I tried extending the current functions rather than introducing a new function altogether for testing stats. Can you take a look at the approach and see if this aligns with overall direction that we are taking.

platform/linux-generic/odp_ipsec.c
```diff
@@ -1917,6 +1919,41 @@ int odp_ipsec_out_inline(const odp_packet_t pkt_in[], int num_in,
...
 }
 
+static void
+ipsec_sa_err_stats_update(odp_ipsec_packet_result_t *result)
+{
+	ipsec_sa_t *sa;
+
+	ODP_ASSERT(result->sa != ODP_IPSEC_SA_INVALID);
+
+	if (result->status.all == ODP_IPSEC_OK)
```
JannePeltonen, line 1929: This probably should be odp_likely().

platform/linux-generic/odp_ipsec.c
```diff
@@ -1928,6 +1965,9 @@ int odp_ipsec_result(odp_ipsec_packet_result_t *result, odp_packet_t packet)
 	/* FIXME: maybe postprocess here, setting alg error in case of crypto
 	 * error instead of processing packet fully in ipsec_in/out_single */
 
+	if (ipsec_config->stats_en)
+		ipsec_sa_err_stats_update(res);
```
JannePeltonen, line 1969: Doing the update here means that statistics will not get updated if the application does not call odp_ipsec_result(). I think the update should be done in ipsec_in_single() and in ipsec_out_single().

platform/linux-generic/odp_ipsec_sad.c
```diff
@@ -900,3 +908,18 @@ uint16_t _odp_ipsec_sa_alloc_ipv4_id(ipsec_sa_t *ipsec_sa)
 	/* No need to convert to BE: ID just should not be duplicated */
 	return tl->next_ipv4_id++;
 }
+
+uint64_t _odp_ipsec_sa_stats_pkts(ipsec_sa_t *sa)
+{
+	uint64_t per_thread_pkt_cnt = 0;
```
JannePeltonen, line 914: Please rename the variable as it is not a per-thread count. Actually, it is not really a packet count either, but the number of future packets already added to the per-SA packet count but not yet processed.

The comment line below ("aggregate per thread packet counts") may also mislead the casual reader so I would suggest either just removing the comment or rephrasing it.

test/validation/api/ipsec/ipsec_test_out.c
```diff
@@ -375,6 +407,10 @@ static void test_esp_out_in_all(void)
 	for (c = 0; c < ARRAY_SIZE(ciphers); c++)
 		for (a = 0; a < ARRAY_SIZE(auths); a++)
 			test_esp_out_in(&ciphers[c], &auths[a]);
+
+	for (c = 0; c < ARRAY_SIZE(cipher_auth_comb); c++)
+		test_esp_out_in(&cipher_auth_comb[c].cipher,
```
JannePeltonen, line 431: AFAICS, this does not add any new tests but duplicates existing tests that are currently separate test cases and not part of test_esp_out_in_all().

This change might have some merit as a code cleanup patch if it also removed the existing test cases.

test/validation/api/ipsec/ipsec_test_out.c
```diff
@@ -330,9 +330,11 @@ static void test_out_in_common(odp_bool_t ah,
 			       odp_auth_alg_t auth,
 			       const odp_crypto_key_t *auth_key,
 			       const odp_crypto_key_t *cipher_key_extra,
-			       const odp_crypto_key_t *auth_key_extra)
+			       const odp_crypto_key_t *auth_key_extra,
+			       ipsec_test_flags *flags)
```
JannePeltonen, line 334: If there is this new flags parameter, would it not make sense to include the ah versus esp flag in it too? And I am not sure if the flags parameter should be a struct pointer like that or if it would be simpler here to just have a bitmap of flags defined as ENUM.

test/validation/api/ipsec/ipsec_test_out.c
```diff
@@ -1181,6 +1201,33 @@ static void ipsec_test_capability(void)
 	CU_ASSERT(odp_ipsec_capability(&capa) == 0);
 }
 
+static void test_ipsec_stats(void)
```
JannePeltonen, line 1204: This is now quite similar to test_esp_out_in_all() except that statistics checking is also enabled. I do not see why we shouldn't just have test_esp_out_in_all() enable statistics checking and be done with it?

test/validation/api/ipsec/ipsec_test_out.c
```diff
@@ -1181,6 +1201,33 @@ static void ipsec_test_capability(void)
...
+	ipsec_test_flags flags = {0};
+	uint32_t a, c;
+	int rc = -1;
+
+	flags.stats_success = 1;
+
+	for (c = 0; c < ARRAY_SIZE(ciphers); c++)
+		for (a = 0; a < ARRAY_SIZE(auths); a++) {
+			rc = test_esp_out_in(&ciphers[c], &auths[a], &flags);
+			if (rc == 0)
```
JannePeltonen, line 1215: Is the intent here that stats are checked only for the first supported algorithm combo and not the others? Why? Would it not be much simpler to just check the stats as part of test_esp_out_in_all()?

test/validation/api/ipsec/ipsec_test_out.c
```diff
@@ -1181,6 +1201,33 @@ static void ipsec_test_capability(void)
...
+
+	for (c = 0; c < ARRAY_SIZE(cipher_auth_comb); c++) {
+		rc = test_esp_out_in(&cipher_auth_comb[c].cipher,
+				     &cipher_auth_comb[c].auth,
+				     &flags);
+		if (rc == 0)
+			return;
+	}
+
+	CU_ASSERT_NOT_EQUAL_FATAL(rc, 0);
```
JannePeltonen, line 1227: rc cannot ever be zero here.

platform/linux-generic/odp_ipsec.c
anoobj: Agreed. Will fix in next version.

platform/linux-generic/odp_ipsec.c
anoobj: I did ask the same question myself while making this change. For linux-generic it should be fine either way, but with implementations having hardware accelerators, it may not be possible to do it that way (as in, state of the operation may not be established by the time the API returns). So ipsec result check would be kind of mandatory in such cases. 

But for linux-generic, I don't mind either way. Please let me know your preference and I can rework.

platform/linux-generic/odp_ipsec_sad.c
anoobj: Will remove the comment. As for variable name, do you have any suggestions? 

My suggestion: 'tl_pkt_quota'

test/validation/api/ipsec/ipsec_test_out.c
anoobj: I saw that there were tests added to make sure all tests use a common framework. Thought I can chip in by extending it to include combined mode algos as well. I can remove the existing cases when we decide to remove the ones covered with the first loop (non-combined mode cipher & auth).

I saw that earlier every combination was tried as a separate test case, while with this approach, the skipped tests are not really counted. So I wasn't sure if this is still WIP. IMO, IPsec tests can focus on other IPsec features rather on combinations of algos, and the direction looks sensible to me.

test/validation/api/ipsec/ipsec_test_out.c
anoobj: Yes. It makes sense to include AH vs ESP also. I'm also considering using an ENUM instead. I've added one more test case (stats of proto error). Please have a look at it when I push next version. In general, can I assume that you are okay with one framework doing all tests based on flags?

test/validation/api/ipsec/ipsec_test_out.c
anoobj: That is true. May be let me highlight why I intend to do with different function.

1. Currently only success case is covered. I would like to cover other cases also. (proto error, auth error etc), which will look like deviating from what test_esp_out_in_all() is supposed to do. I'll push another version which would cover proto_err, then the change might be more clear to you.
2. We cannot try both success and error case with same run. Multiple runs of one algo is required, but not necessarily all algos as in test_esp_out_in_all()

test/validation/api/ipsec/ipsec_test_out.c
anoobj: That is correct. We can technically check success stats for all combination in test_esp_out_in_all() but error cases are the trickier ones.

test/validation/api/ipsec/ipsec_test_out.c
anoobj: Idea is to make sure we find atleast one supported algo. Let me recheck if we need this.

platform/linux-generic/odp_ipsec.c
JannePeltonen: The problem is that the API does not mandate that the application calls odp_ipsec_result() right after IPsec processing or ever. In practise it might be a rare case that the application never calls the function but I can imagine that some application may e.g. queue the packet and call the result function much later in another thread.

So the question is, what is the state of the packet or event returned by your IPsec implementation before odp_ipsec_result() gets called? Does skipping the odp_ipsec_result() call screw up merely the statistics or something else too?

Why cannot IPsec processing be fully done already when the application gets the result packet? Too costly to check for IPsec processing completion need for every packet in some generic scheduling or dequeueing functions so you would like to postpone that to odp_ipsec_result()?

If you really require that odp_ipsec_result() is always called or bad things (or just statistics miscounts) will happen, then I think the API needs to be updated.

platform/linux-generic/odp_ipsec_sad.c
JannePeltonen: I do not have a good suggestion (maybe precounted_packets?). I would not complain about tl_pkt_quota.

test/validation/api/ipsec/ipsec_test_out.c
JannePeltonen: > I can remove the existing cases when we decide to remove the ones covered with the first loop (non-combined mode cipher & auth).

I do not think there are any existing separate test cases that are also covered buy the common out+in test.

A while ago I changed the code to get rid of duplication and to be able to add more algorithm combos without combinatoric explosion of the number of separate test cases. I am not working on further refactoring, but some may be needed when further tests are added (e.g. if we start supporting more than one ICV length per algorithm).

test/validation/api/ipsec/ipsec_test_out.c
JannePeltonen: > In general, can I assume that you are okay with one framework doing all tests based on flags?

Yes, as I suppose it avoids having to duplicate the same code multiple times. Just make sure that it is clear from test output what exactly failed in a test failure.

test/validation/api/ipsec/ipsec_test_out.c
JannePeltonen: Ok, let me see a version with some error stat test and it may become more clear.

test/validation/api/ipsec/ipsec_test_out.c
JannePeltonen: But rc is always -1 if the code ends up here, no matter what is supported and what the ODP implementation does.

platform/linux-generic/odp_ipsec.c
anoobj: > 
> The problem is that the API does not mandate that the application calls odp_ipsec_result() right after IPsec processing or ever. In practise it might be a rare case that the application never calls the function but I can imagine that some application may e.g. queue the packet and call the result function much later in another thread.
> 
> So the question is, what is the state of the packet or event returned by your IPsec implementation before odp_ipsec_result() gets called? Does skipping the odp_ipsec_result() call screw up merely the statistics or something else too?

Isn't IPsec result check mandatory? Like, in case of ICV errors, it is an auditable event.

https://tools.ietf.org/html/rfc4303#page-30

> 
> Why cannot IPsec processing be fully done already when the application gets the result packet? Too costly to check for IPsec processing completion need for every packet in some generic scheduling or dequeueing functions so you would like to postpone that to odp_ipsec_result()?

Yes. It would make the scheduling/dequeuing functions to grow and become rather inefficient for non-IPsec use cases.

> 
> If you really require that odp_ipsec_result() is always called or bad things (or just statistics miscounts) will happen, then I think the API needs to be updated.

From the RFC it looks like, result check is mandatory. So do you recommend documenting in the API? I can submit the patch if you suggest so.

test/validation/api/ipsec/ipsec_test_out.c
anoobj: Yeah. The returns had to be changed. The returns were a temporary thing which I missed. Will have it addressed with next version. Thanks.

platform/linux-generic/odp_ipsec.c
JannePeltonen: > Isn't IPsec result check mandatory? Like, in case of ICV errors, it is an auditable event.

IPsec RFC compliancy matters do not tell how the application is allowed to use the API and one cannot deduce that based on certain RFC clause an ODP application must always all odp_ipsec_result(). Maybe the application does not even want to be RFC compliant? Maybe the application uses odp_packet_has_error() to check when it needs to call odp_ipsec_result() (yes, there is ongoing discussion about how odp_packet_has_error() should work).

> So do you recommend documenting in the API? I can submit the patch if you suggest so.

The API needs to fully document how it must or must not be used, so please suggest a change if you think it is needed. But try to define very carefully when exactly the application must call odp_ipsec_result() and what happens if it does not (undefined behaviour / inaccurate statistics / some functions may not be called for the packet). I think one use case that probably needs to work is reception of IPsec events and then freeing them without any processing at all.

anoobj: > 
> 
> > Isn't IPsec result check mandatory? Like, in case of ICV errors, it is an auditable event.
> 
> IPsec RFC compliancy matters do not tell how the application is allowed to use the API and one cannot deduce that based on certain RFC clause an ODP application must always all odp_ipsec_result(). Maybe the application does not even want to be RFC compliant? Maybe the application uses odp_packet_has_error() to check when it needs to call odp_ipsec_result() (yes, there is ongoing discussion about how odp_packet_has_error() should work).

Understood. Should I update the ipsec_stats documentation to reflect that application is expected to call odp_ipsec_result() for stats to be updated properly? Putting it in schedule/dequeue APIs will mean non-IPsec packets will have an unnecessary check and I would like to avoid that.

> 
> > So do you recommend documenting in the API? I can submit the patch if you suggest so.
> 
> The API needs to fully document how it must or must not be used, so please suggest a change if you think it is needed. But try to define very carefully when exactly the application must call odp_ipsec_result() and what happens if it does not (undefined behaviour / inaccurate statistics / some functions may not be called for the packet). I think one use case that probably needs to work is reception of IPsec events and then freeing them without any processing at all.

Agreed. IMO availability of APIs like odp_ipsec_result() allows schedule/dequeue APIs to be light while allowing various module specific processing to be done separately. Let me recheck the documentation. I can submit API documentation changes if required. For this case, I can update the ipsec_stats documentation to reflect the dependency on odp_ipsec_result().

anoobj:
@JannePeltonen Did you get chance to review the latest changes? I've added test case with protocol errors.

psavol:
Depends on PR1095

platform/linux-generic/odp_ipsec_sad.c
```diff
@@ -900,3 +908,18 @@ uint16_t _odp_ipsec_sa_alloc_ipv4_id(ipsec_sa_t *ipsec_sa)
 	/* No need to convert to BE: ID just should not be duplicated */
 	return tl->next_ipv4_id++;
 }
+
+uint64_t _odp_ipsec_sa_stats_pkts(ipsec_sa_t *sa)
```
JannePeltonen, line 912: This does not actually return accurate success stats now because this is based on life time counters and there can be an IPsec processing error  after the life time counter has already been updated. Moving the life time update later in the processing path would not be ok since life time is about how much the SA has been used.

I think you need to track whether an error occurs before or after life time update (e.g. have a new internal counter for that) and adjust the statistics accordingly.

platform/linux-generic/odp_ipsec_sad.c
```diff
@@ -900,3 +908,18 @@ uint16_t _odp_ipsec_sa_alloc_ipv4_id(ipsec_sa_t *ipsec_sa)
...
+{
+	uint64_t tl_pkt_quota = 0;
+	sa_thread_local_t *sa_tl;
+	int n;
+
+	/* Aggregate per thread packet counts */
+	for (n = 0; n < ODP_THREAD_COUNT_MAX; n++) {
+		sa_tl = &ipsec_sa_tbl->per_thread[n].sa[sa->ipsec_sa_idx];
+		tl_pkt_quota += sa_tl->packet_quota;
```
JannePeltonen, line 921: While in practice this probably would not cause any problems, this is still a data race (other thread writing to the variable while we are reading it). To make things right, sa_tl->packet_quota should be an atomic variable. It can be accessed using relaxed ordering.

test/validation/api/ipsec/ipsec_test_out.c
```diff
@@ -55,6 +55,38 @@ static struct auth_param auths[] = {
...
+	},
+	{
+		ALG(ODP_CIPHER_ALG_NULL, NULL, NULL),
+		ALG(ODP_AUTH_ALG_AES_GMAC, &key_a5_192, &key_mcgrew_gcm_salt_2),
+	},
+	{
+		ALG(ODP_CIPHER_ALG_NULL, NULL, NULL),
+		ALG(ODP_AUTH_ALG_AES_GMAC, &key_a5_256, &key_mcgrew_gcm_salt_2),
+	},
+};
```
JannePeltonen, line 88: This could cover all combined mode algorithms used in ESP (as the commit message suggests). AES-CCM and CHACHA20-POLY1305 are now missing. Adding those too and removing the individual tests will also make the later patches a tad smaller.

test/validation/api/ipsec/ipsec_test_out.c
JannePeltonen: I still do not see the benefit of this separate test function. Now the code spends a lot of effort in making it so that statistics are tested only with single algorithm combo when it could just run the test with all algorithms and have better coverage. Why is this better than e.g. having test_esp_out_in() test the statistics?

test/validation/api/ipsec/ipsec.c
```diff
@@ -807,6 +807,13 @@ static int ipsec_send_out_one(const ipsec_test_part *part,
 	return num_out;
 }
 
+static void ipsec_pkt_proto_err_set(odp_packet_t pkt)
+{
+	uint8_t *data = (uint8_t *)odp_packet_data(pkt);
+	data += odp_packet_len(pkt) - 4;
+	data[0] = ~data[0];
```
JannePeltonen, line 814: It would be good to have a comment what this does. This appears to me to write into the ICV area of the packet in most (all?) cases, so why do we not get an auth error instead of proto error?

This may access data outside the segment, depending on the packet length. Or maybe is won't if there is suitable pool config and packet construction elsewhere in the code. Using odp_packet_copy_from_mem() would be more robust.

test/validation/api/ipsec/ipsec_test_out.c
```diff
@@ -375,31 +378,60 @@ static void test_out_in_common(ipsec_test_flags *flags,
...
+
+	if (flags->stats == IPSEC_TEST_STATS_PROTO_ERR)
+		test.in[0].status.error.proto = 1;
+
+	if (flags->stats == IPSEC_TEST_STATS_SUCCESS) {
+		CU_ASSERT_EQUAL(odp_ipsec_stats(sa_out, &stats), 0);
+		CU_ASSERT_EQUAL(stats.success, 0);
+		CU_ASSERT_EQUAL(odp_ipsec_stats(sa_in, &stats), 0);
+		CU_ASSERT_EQUAL(stats.success, 0);
+	}
```
JannePeltonen, line 391: Why not check here the initial values of all counters, not just the success counter, and in all cases, not just when testing the success counter?

test/validation/api/ipsec/ipsec_test_out.c
```diff
@@ -375,31 +378,60 @@ static void test_out_in_common(ipsec_test_flags *flags,
...
+		CU_ASSERT_EQUAL(odp_ipsec_stats(sa_out, &stats), 0);
+		CU_ASSERT_EQUAL(stats.success, 1);
+		CU_ASSERT_EQUAL(odp_ipsec_stats(sa_in, &stats), 0);
+		CU_ASSERT_EQUAL(stats.success, 1);
+	}
+
+	if (flags->stats == IPSEC_TEST_STATS_PROTO_ERR) {
+		CU_ASSERT_EQUAL(odp_ipsec_stats(sa_in, &stats), 0);
+		CU_ASSERT_EQUAL(stats.proto_err, 1);
+	}
```
JannePeltonen, line 405: Here the success case could test that error counters are zero and the error case that the success counters are zero,

test/validation/api/ipsec/ipsec_test_out.c
anoobj: Currently test_esp_out_in() would only try for success cases. So, yes, we can have stats run for every success algo. But then, if we want stats of any error case, then the entire suite need to be re-run with error simulated for each case, right? The effort in making sure only once the test case is run is actually minimal. Majority of the changes proposed is to make sure we don't replicate the code for just this test case.

Currently, we test IPsec features with random algo combinations. Like, few tests use AES-GCM, while few others use NULL+SHA256. What if the implementation doesn't support that combination? I was trying to answer that. Also, I saw that test_esp_out_in() routine was actually an attempt to have a common framework that iterates through all supported ciphers. The intention is to extend the same framework but pick a supported algo combination for all further tests.

test/validation/api/ipsec/ipsec_test_out.c
```diff
@@ -1157,6 +1193,49 @@ static void ipsec_test_capability(void)
...
+	for (c = 0; c < ARRAY_SIZE(cipher_auth_comb); c++) {
+		rc = test_esp_out_in(&cipher_auth_comb[c].cipher,
+				     &cipher_auth_comb[c].auth,
+				     &flags);
+		if (rc == 0)
+			goto stats_tests;
+	}
+
+stats_tests:
+	CU_ASSERT_EQUAL_FATAL(rc, 0);
```
JannePeltonen, line 1220: I do not think it is a fatal error if there are no supported algorithms. I suppose that is the case with null cypto, i.e. when not compiling linux-gen with openssl.

test/validation/api/ipsec/ipsec_test_out.c
JannePeltonen: We are discussing two related points here: 1) whether statistics should be tested with all algorithm combos, 2) whether the statistics test could be combined with the existing test.

For 1) I would say we should test with all algorithms, since we can do it basically without any extra effort. Just picking the first algorithm that happens to be supported may not give good coverage. E.g. combined mode algorithms might behave a bit differently from the traditional ones for some statistics. The current proposal in this patch would actually run the test with null encryption and null authentication which is not good.

Then regarding 2), I think we could do it because of 1). But If you want to keep it as a separate test, that is ok too. In this case the code gets still simpler because because of 1) you would need the separate phase for finding the first supported algorithm (which should anyway have been done using ipsec_check_esp() instead of rerunning the original test).

anoobj: For 1) Yes. You are right. Different algos could be done differently. So running for stats success case for all ciphers do make sense. I'll have this addressed in next version.

For 2) My only problem is the simulation of an error case. Right now, if I add the flags for proto_err, then the success case won't be tested. So only option is to first run the cases where packets would be successfully processed and then proceed to error cases (one by one). Hence my preference to introduce a separate function for stats testing.

Now regarding testing proto_err stats, do you recommend testing it for specific algos (like let's say AES-GCM) or any supported algo or for all algos?

JannePeltonen: > Now regarding testing proto_err stats, do you recommend testing it for specific algos (like let's say AES-GCM) or any supported algo or for all algos?

I would run all stats tests for all algos because it can be easily done. Except maybe not everything (or anything) for null algos (e.g. there cannot be auth error with null authentication, so you cannot test it with that).

anoobj: Agreed. I can try to extend the error cases to run all algos.

platform/linux-generic/odp_ipsec_sad.c
anoobj: For such cases, we can just subtract the 'hard_exp_pkt_err', right?

`+       stats->success = _odp_ipsec_sa_stats_pkts(ipsec_sa) - stats->hard_exp_pkts_err; `

JannePeltonen: You are actually thinking about a different issue. Packets with hard lifetime expiration are should not be counted as succesfuly received packets so something needs to be done about it. Just doing the subtraction like that will not work correctly if byte based life time expires before the packet based.

What I was thinking was the case when lifetime is ok but there is an error later. There are many possibilities for such later error in ipsec_in_single().

Maybe a simple way to fix this could be to add a new flag in the state struct indicating if lifetime counting has already happened and then just before the common error return increment a new internal counter that counts such errored packets. The statistics query could then properly subtract those packets from the success count. Just a quick thought.

By the way, I noticed a problem that lifetime tracking state gets updated before the final anti-replay check, thus potentially allowing replayed packets alter the SA state. See PR 1105 for the fix that touches the same code that we are discussing.

anoobj: Thinking again, do you think we should introduce a different counter for success stats rather than piggybacking on lifetime tracks? Your suggestion looks good otherwise. Let me see how straight forward it is.

Saw the PR. Thanks for fixing it.

JannePeltonen: If it gets very messy piggybacking on lifetime counters then I think we can add a new counter for the success case and optimize things later as necessary.

platform/linux-generic/odp_ipsec_sad.c
anoobj: Agreed. Will have this addressed in next version.

test/validation/api/ipsec/ipsec_test_out.c
anoobj: Agreed. Only issue is, CCM support has some problem. Capabilities says CCM support is there but the current unit test itself skips the tests after printing "Wrong auth capabilities"

Since I'm covering all algos for all tests, the log would look like,

```
    ODP_CIPHER_ALG_NULL (keylen 0) ODP_AUTH_ALG_AES_GMAC (keylen 256) Wrong auth capabilities
Wrong auth capabilities
Wrong auth capabilities

    ODP_CIPHER_ALG_CHACHA20_POLY1305 (keylen 256) ODP_AUTH_ALG_CHACHA20_POLY1305 (keylen 0)
  passed
  Test: test_ah_out_in_all ...
    ODP_AUTH_ALG_NULL (keylen 0)
    ODP_AUTH_ALG_MD5_HMAC (keylen 128)
    ODP_AUTH_ALG_SHA1_HMAC (keylen 160)
    ODP_AUTH_ALG_SHA256_HMAC (keylen 256)
    ODP_AUTH_ALG_SHA384_HMAC (keylen 384)
    ODP_AUTH_ALG_SHA512_HMAC (keylen 512)
    ODP_AUTH_ALG_AES_XCBC_MAC (keylen 128)
  passed
  Test: test_ipsec_stats ...
        Stats : successWrong auth capabilities
Wrong auth capabilities
Wrong auth capabilities

        Stats : proto errWrong auth capabilities
Wrong auth capabilities
Wrong auth capabilities

  passed
```
I'll submit this in the next version.

JannePeltonen: Yes, I know there is a problem with CCM in linux-gen IPsec. The short story is that it is not supported in linux-gen even though odp_ipsec_capability() claims so.

The long story:

The IPsec implementation in Linux-gen wants to use AES-CCM with 16 byte ICV length but the AES-CCM implementation in the crypto subsystem does not support that ICV length. The IPsec implementation does not realize this in the capability call and returns AES-CCM in the list of supported algorithms. Then odp_ipsec_auth_capability() checks the ICV length and does not return anything for AES-CCM, giving rise to the error you see.

An underlying problem here is that the ODP IPsec API does not support user-specified ICV lengths when an algorithm (or its application to IPsec) supports multiple lengths and leaves it unspecified which length is used, AES-GCM supports multiple lengths, but only one (16 bytes) is mandatory, so the current API is ok'ish with the user having to deduce or assume that ICV will be 16 bytes. or by adding a note about the ICV length in the API. But AES-CCM has two mandatory ICV lengths, so fixing ODP API to support only one of them would not be very good and the only real solution seems to be to add provisions for the user to specify the ICV length. I am planning to submit a PR for such an API change.

In the mean time, I think you could just ignore the error message as it does not cause validation test failure (it probably should) and it is not a problem in the validation test itself.

anoobj: Understood. Thanks for the explanation.

test/validation/api/ipsec/ipsec.c
anoobj: You are right (on both fronts). Will have these addressed in the next version.

test/validation/api/ipsec/ipsec_test_out.c
anoobj: Agreed.

test/validation/api/ipsec/ipsec_test_out.c
anoobj: Agreed. Fixed in v7

test/validation/api/ipsec/ipsec_test_out.c
anoobj: This is actually removed in the latest version (since we are running for all algos).

anoobj:
@JannePeltonen @psavol Following is the summary of changes,

1. In linux-gen I had to introduce a new patch after rebasing to ToT.
        linux-gen: fix exit label in ipsec_in_single
2. Extended the common framework to run AES-CCM & CHACHA-POLY algos
3. Made stats to run for all supported algos
4. Enhancements to error packet cases (now supports auth_err & proto err).

There is a checkpatch warning, which if I fix would give me build errors.

WARNING: braces {} are not necessary for any arm of this statement
#135: FILE: test/validation/api/ipsec/ipsec_test_out.c:358:
+       if (test == IPSEC_TEST_STATS_SUCCESS) {
[...]
+       } else {
[...]

JannePeltonen:
The reuse of lifetime counter for success packet counting is getting a bit messy for maybe a questionable benefit. Maybe it would be better to just have a separate counter (put I think it should be per-thread) for the succesfuly processed packets. It would also be somewhat less racy (e.g. success count would always be monotonic).

We can optimize things later (e.g. by reusing lifetime counters or reusing the sequence number) if it seems worthwhile.

I will comment the implementation patch about things that would have to be fixed or improved, but most, if not all, of those comments become moot if we switch to a separate packet counter.

platform/linux-generic/odp_ipsec.c
```diff
@@ -798,6 +800,12 @@ static ipsec_sa_t *ipsec_in_single(odp_packet_t pkt,
 		odp_packet_parse(pkt, state.ip_offset, &parse_param);
 	}
 
+	goto exit;
+
+post_ipsec_err_cnt_update:
```
JannePeltonen, line 805: If _odp_ipsec_sa_stats_update() (or _odp_ipsec_sa_replay_update() before the antireplay check ordering fix is included) fails, we should come here to not count the packets as successes.

This whole mechanism would also be needed in outbound processing. There can be errors after lifetime updating there too.

platform/linux-generic/odp_ipsec.c
```diff
@@ -1908,6 +1916,41 @@ int odp_ipsec_out_inline(const odp_packet_t pkt_in[], int num_in,
...
 }
 
+static void
+ipsec_sa_err_stats_update(odp_ipsec_packet_result_t *result)
+{
+	ipsec_sa_t *sa;
+
+	ODP_ASSERT(result->sa != ODP_IPSEC_SA_INVALID);
+
+	if (odp_likely(ODP_IPSEC_OK == result->status.all))
```
JannePeltonen, line 1926: It would be better to check if result->status.error.all is zero as we do not count soft lifetime expirations.

platform/linux-generic/odp_ipsec_sad.c
```diff
@@ -801,7 +810,8 @@ int _odp_ipsec_sa_stats_update(ipsec_sa_t *ipsec_sa, uint32_t len,
 		    packets >= ipsec_sa->hard_limit_packets)
 			sa_tl->lifetime_status.error.hard_exp_packets = 1;
 	}
-	sa_tl->packet_quota--;
+
+	odp_atomic_dec_u32(&sa_tl->packet_quota);
```
JannePeltonen, line 814: Atomic incs and decs are actually unnecessarily strong here as this thread is the only writer. IOW, we do not need atomicity of the whole read-modify-write sequence but only atomicity of the store so that the stats query thread can load the value without a data race. Atomic load, followed by increment/decrement, followed by atomic store should be faster than an atomic inc/dec in many systems.

test/validation/api/ipsec/ipsec.c
```diff
@@ -269,10 +269,8 @@ int ipsec_check(odp_bool_t ah,
 	}
 
 	num = odp_ipsec_auth_capability(auth, auth_capa, MAX_ALG_CAPA);
-	if (num <= 0) {
-		fprintf(stderr, "Wrong auth capabilities\n");
+	if (num <= 0)
```
JannePeltonen, line 272: I think this should result in validation test failure instead of just inactivation of the test. At this point ODP has already claimed to support the algorithm when queried through odp_ipsec_capability() and now when the test code wants to know more, ODP basically says that it does not support any key length with the algorithm.

This just cannot be fixed without fixing the CCM implementation problem that I discussed earlier or the validation tests would start failing (which would be right, but break the CI for everybody). I would leave this alone for now as it is not really related to the code changes in this patch.

test/validation/api/ipsec/ipsec.c
```diff
@@ -781,6 +781,30 @@ static int ipsec_send_out_one(const ipsec_test_part *part,
...
+static void ipsec_pkt_proto_err_set(odp_packet_t pkt)
+{
+	odph_ipv4hdr_t *ip = odp_packet_l3_ptr(pkt, NULL);
+
+	/* Simulate proto error by corrupting protocol field */
+
+	if (ip->proto == ODPH_IPPROTO_ESP)
+		ip->proto = ODPH_IPPROTO_AH;
+	else
+		ip->proto = ODPH_IPPROTO_ESP;
```
JannePeltonen, line 793: This may write outside the packet segment if there is a segment boundary before the protocol field. The copy-to-and-from-mem pattern would fix this, or alternatively, maybe you could just pass non-null second parameter to odp_packet_l3_ptr() and then assert that the contiguous length is sufficient.

test/validation/api/ipsec/ipsec_test_out.c
```diff
@@ -349,9 +389,23 @@ static void test_out_in_common(ipsec_test_flags *flags,
...
 	odp_ipsec_sa_param_t param;
+	odp_ipsec_stats_t stats;
 	odp_ipsec_sa_t sa_out;
 	odp_ipsec_sa_t sa_in;
 
+	CU_ASSERT_NOT_EQUAL_FATAL(flags, NULL);
+
+	/*
+	 * ICV won't be generated for NULL AUTH. For AES-GCM & CHACHA-POLY
+	 * any data corruption would lead to alg error.
```
JannePeltonen, line 400: I wonder if this is ok and according to the API spec or a bug in the implementation.

platform/linux-generic/odp_ipsec.c
anoobj: > If _odp_ipsec_sa_stats_update() (or _odp_ipsec_sa_replay_update() before the antireplay check ordering fix is included) fails, we should come here to not count the packets as successes.

Correct. Actually I was thinking that I would rebase after the other PR is merged. So missed this part.

> 
> This whole mechanism would also be needed in outbound processing. There can be errors after lifetime updating there too.

Agreed. Will add this in next version.

Do you agree with the change in labels? (Parent patch to this change). Outbound also would need a similar adjustment. IMO, calling the label 'err' and success also taking that code path may not look right.

platform/linux-generic/odp_ipsec_sad.c
anoobj: Agreed. Will have this fixed in next version.

test/validation/api/ipsec/ipsec.c
anoobj: Missed this. copy-to-and-from-mem looks good. I'll make that change.

test/validation/api/ipsec/ipsec_test_out.c
anoobj: I'm not sure if it's a bug or if that's the intended usage.

In linux-gen, when crypto status is not ok, 
```
if (cipher_status.alg_err)
    status->error.alg = 1;

if (auth_status.alg_err)
    status->error.auth = 1;
```

It's more like, when crypto & auth are separate operations, crypto(cipher_status.alg_err) will not be flagged for any data corruption. It's the auth check which allows us to detect errors. But when we do combined operation, crypto(cipher_status.alg_err) gets reported.

IMO, the above code snippet is not correct. It should be auth_err for either case. I can submit a fix for this and correct UT as well if you can confirm that's the intended usage.

platform/linux-generic/odp_ipsec.c
JannePeltonen: > Do you agree with the change in labels?

I do not think I would have made the change. I do not think it is a problem that the success path falls through to the same code that comes after the err label. Only the error cases actually use the label. If the label were already named exit, I would not change it either.

So, I would slightly prefer not changing the label, but it is a matter of taste and if you want to change it I am ok with it (which is why I did not comment it in the review).

anoobj: Understood. Without the label rename, my code change in case of success would have had like,
```
	goto err;

post_ipsec_err_cnt_update:
```

Or duplicate what is being done as part of err(/exit). Let me recheck if similar kind of change is required in the out function as well.

JannePeltonen: Right, now I see it. I was looking at just that patch when making the previous comment, but after the latter patch the exit label makes more sense.

platform/linux-generic/odp_ipsec.c
anoobj: @JannePeltonen Since the API for post-processing may need some more iterations, shall we merge this now and I can take up moving stats collection to the API that we decide for post-processing, once we agree on that PR.

JannePeltonen: For us to merge this before changing the API to allow post processing we need to move the ipsec_sa_err_stats_update() call from odp_ipsec_result() to somewhere like ipsec_in_single() and ipsec_out_single(). I think we should do that change and try to get this merged soon.

test/validation/api/ipsec/ipsec_test_out.c
JannePeltonen: Yes, I think we should get auth error if the ICV is wrong. If you could submit a fix, that would be great.

test/validation/api/ipsec/ipsec.c
JannePeltonen: Could you remove this change?

I am going to soon send a patch to fix the capability issue and will put a CU_ASSERT in this location to make sure an implementation does not shy away from supporting an algorithm that it already advertised in capabilities.

JannePeltonen:
> There is a checkpatch warning, which if I fix would give me build errors.

This is caused by buggy macro definition in CUnit. We just have to ignore the warning.

JannePeltonen:
Minor points: There is a typo in a commit message ("oepration"). _odp_ipsec_sa_stats_update() could be renamed to more descriptive_odp_ipsec_sa_lifetime_update() if you would like to do that change in this patch set.

test/validation/api/ipsec/ipsec_test_out.c
anoobj: Done.

https://github.com/OpenDataPlane/odp/pull/1122

test/validation/api/ipsec/ipsec.c
anoobj: Sure.

anoobj:
> Minor points: There is a typo in a commit message ("oepration").

Will have this fixed with next version.

> _odp_ipsec_sa_stats_update() could be renamed to more descriptive_odp_ipsec_sa_lifetime_update() if you would like to do that change in this patch set.

I can do that. Should the rename be a separate patch? Or I can put it as part of the label name change. Can you suggest?

JannePeltonen:
> I can do that. Should the rename be a separate patch? Or I can put it as part of the label name change. Can you suggest?

I would actually put it in the implementation patch and also squash the label name change into it.

platform/linux-generic/odp_ipsec_sad.c
```diff
@@ -458,6 +458,15 @@ odp_ipsec_sa_t odp_ipsec_sa_create(const odp_ipsec_sa_param_t *param)
...
 	ipsec_sa->hard_limit_packets = param->lifetime.hard_limit.packets;
 
+	odp_atomic_store_u64(&ipsec_sa->stats.proto_err, 0);
+	odp_atomic_store_u64(&ipsec_sa->stats.auth_err, 0);
+	odp_atomic_store_u64(&ipsec_sa->stats.antireplay_err, 0);
+	odp_atomic_store_u64(&ipsec_sa->stats.alg_err, 0);
+	odp_atomic_store_u64(&ipsec_sa->stats.mtu_err, 0);
+	odp_atomic_store_u64(&ipsec_sa->stats.hard_exp_bytes_err, 0);
+	odp_atomic_store_u64(&ipsec_sa->stats.hard_exp_pkts_err, 0);
+	odp_atomic_store_u64(&ipsec_sa->stats.post_lifetime_err_pkts, 0);
```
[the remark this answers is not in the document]
anoobj: Fixed the ones introduced with this patch. Other occurrences we can take up with a new PR.

platform/linux-generic/odp_ipsec.c
```diff
@@ -1521,17 +1521,13 @@ static ipsec_sa_t *ipsec_out_single(odp_packet_t pkt,
...
-			status->error.alg = 1;
-
-		if ((crypto.auth_status.alg_err !=
-		     ODP_CRYPTO_ALG_ERR_NONE) ||
-		    (crypto.auth_status.hw_err !=
-		     ODP_CRYPTO_HW_ERR_NONE))
+		if ((crypto.cipher_status.alg_err ==
+		     ODP_CRYPTO_ALG_ERR_ICV_CHECK) ||
+		    (crypto.auth_status.alg_err ==
+		     ODP_CRYPTO_ALG_ERR_ICV_CHECK))
```
JannePeltonen, line 1571: Authentication error does not make much sense to me in outbound processing so I would rather have this always return alg error when !crypto.ok.

platform/linux-generic/odp_ipsec.c
```diff
@@ -596,6 +598,42 @@ static int ipsec_in_ah_post(odp_packet_t pkt,
 	return 0;
 }
 
+static void
+ipsec_sa_err_stats_update(odp_ipsec_packet_result_t *result)
```
JannePeltonen, line 602: Now that the call sites of this changed, this could take SA (as ipsec_sa_t) and status (odp_ipsec_op_status_t) as parameters and get rid of the _odp_ipsec_sa_entry_from_hdl() call and all the result dereferencing.

Is there some benefit in calling this from the top level API functions instead of ipsec_in_single() and ipsec_out_single()? That puts the call in 6 call sites instead of 2.

platform/linux-generic/odp_ipsec.c
```diff
@@ -596,6 +598,42 @@ static int ipsec_in_ah_post(odp_packet_t pkt,
...
+{
+	ipsec_sa_t *sa;
+
+	if (odp_likely(ODP_IPSEC_OK == result->status.error.all))
+		return;
+
+	if (result->status.error.sa_lookup)
+		return;
```
JannePeltonen, line 610: There are also other cases where result->sa is not valid. For instance, if packet parsing before SA lookup fails, ipsec_in_single() sets alg error. It would also be more robust in general to check the validity of the SA directly.

platform/linux-generic/odp_ipsec.c
```diff
@@ -682,18 +720,18 @@ static ipsec_sa_t *ipsec_in_single(odp_packet_t pkt,
...
 	}
 
 	if (_odp_ipsec_sa_replay_update(ipsec_sa,
 					state.in.seq_no,
 					status) < 0)
-		goto err;
+		goto exit;
 
-	if (_odp_ipsec_sa_stats_update(ipsec_sa,
+	if (_odp_ipsec_sa_lifetime_update(ipsec_sa,
```
JannePeltonen, line 731: Indentation of the unchanged lines should be fixed. Checkpatch also complains about it.

platform/linux-generic/odp_ipsec.c
anoobj: Agreed. Do you want me to push as a separate patch or shall I make this change as part of stats implementation?

platform/linux-generic/odp_ipsec.c
anoobj: > Now that the call sites of this changed, this could take SA (as ipsec_sa_t) and status (odp_ipsec_op_status_t) as parameters and get rid of the _odp_ipsec_sa_entry_from_hdl() call and all the result dereferencing.

Agreed.
 
> Is there some benefit in calling this from the top level API functions instead of ipsec_in_single() and ipsec_out_single()? That puts the call in 6 call sites instead of 2.

In odp_ipsec_out_inline(), there are errors set after ipsec_out_single() call. For inbound, we can put the call in ipsec_in_single().

platform/linux-generic/odp_ipsec.c
anoobj: Agreed.

platform/linux-generic/odp_ipsec.c
anoobj: Missed this one during rebase. Will be fixed in next version.

platform/linux-generic/odp_ipsec.c
anoobj: Submitted a new PR as it is unrelated to stats implementation.

https://github.com/OpenDataPlane/odp/pull/1123

platform/linux-generic/odp_ipsec.c
JannePeltonen: I think the call to ipsec_sa_err_stats_update() in odp_ipsec:_out_inline() should be made a bit earlier. It should be in the preceding else branch before the error result queuing. That way error counters get updated even if the application lets the result queue sometimes become full. The call also otherwise fits better there since we do not need to update error stats for non-errors.

test/validation/api/ipsec/ipsec_test_out.c
```diff
@@ -410,8 +411,10 @@ static void test_esp_out_in(struct cipher_param *cipher,
 	printf("\n    %s (keylen %d) %s (keylen %d) ",
 	       cipher->name, cipher_keylen, auth->name, auth_keylen);
 
-	test_out_in_common(false /* ESP */,
-			   cipher->algo, cipher->key,
+	memset(&flags, 0, sizeof(ipsec_test_flags));
```
JannePeltonen: I wonder why you changed the initialization to this memset? I think the previous version (ipsec_test_flags flags = {0}) is better. If the memset were to stay, I would prefer it like this: memset(&flags, 0, sizeof(flags)). But I do not see why not to use the initializer like in the previous version.

JannePeltonen:
You probably should update the copyrights of the relevant files. And please rebase on latest master now that PR 1123 has been merged.

platform/linux-generic/odp_ipsec.c
anoobj: Agreed.

test/validation/api/ipsec/ipsec_test_out.c
anoobj: Couple of stages in CI were complaining. Personally I also prefer it over memset, but I've seen similar issues reported with older toolchain versions. Earlier ODP CI wasn't complaining. But now it is. So I don't think we have an option.

>  I would prefer it like this: memset(&flags, 0, sizeof(flags))
I was just sticking to what was being followed in the rest of the file. Will have it changed.

anoobj:
> You probably should update the copyrights of the relevant files. And please rebase on latest master now that PR 1123 has been merged.

Which all files I need to update? Do I need to update all files I've edited?

JannePeltonen:
> > You probably should update the copyrights of the relevant files. And please rebase on latest master now that PR 1123 has been merged.

> Which all files I need to update? Do I need to update all files I've edited?

I would say it is up to you and probably depends on how big changes you have made in each file. So use your judgement.

test/validation/api/ipsec/ipsec_test_out.c
JannePeltonen: I see.

JannePeltonen:
You could also already add my reviewed-by tag in the next version as it is quite likely that it will be the last.

JannePeltonen:
Did this change get in by accident or was it intentional:
```
-       if (flags->display_algo)
-               printf("\n    %s (keylen %d) %s (keylen %d) ",
-                      cipher->name, cipher_keylen, auth->name, auth_keylen);
+       printf("\n    %s (keylen %d) %s (keylen %d) ",
+              cipher->name, cipher_keylen, auth->name, auth_keylen);
```

anoobj:
> 
> 
> Did this change get in by accident or was it intentional:
> 
> ```
> -       if (flags->display_algo)
> -               printf("\n    %s (keylen %d) %s (keylen %d) ",
> -                      cipher->name, cipher_keylen, auth->name, auth_keylen);
> +       printf("\n    %s (keylen %d) %s (keylen %d) ",
> +              cipher->name, cipher_keylen, auth->name, auth_keylen);
> ```

It was an accident during rebase. Let me have it fixed. Thanks for checking.

JannePeltonen (approved):

anoobj:
@matiaselo There is one CI failure on this patch. But it looks unrelated. Will you be able to re-trigger the tests?

```
 ../../../../../../test/validation/api/traffic_mngr/traffic_mngr.c:2720:test_sched_queue_priority():pkts_sent=44 pkt_cnt=40 num_rcv_pkts=44 rcvd_in_order=37
  Test: traffic_mngr_test_scheduler ...FAILED
    1. ../../../../../../test/validation/api/traffic_mngr/traffic_mngr.c:2723  - pkts_in_order == pkt_cnt
```

matiaselo:
> @matiaselo There is one CI failure on this patch. But it looks unrelated. Will you be able to re-trigger the tests?

Done

psavol:
Checkpatch fails due to extra braces. @anoobj please rebase and fix the checkpatch warnings.

WARNING: braces {} are not necessary for any arm of this statement
#141: FILE: test/validation/api/ipsec/ipsec_test_out.c:359:
+ if (test == IPSEC_TEST_STATS_SUCCESS) {
[...]
+ } else {
[...]
Error: WARNING: braces {} are not necessary for any arm of this statement

anoobj:
> Checkpatch fails due to extra braces. @anoobj please rebase and fix the checkpatch warnings.

But if I remove the braces, then the build would fail. I had commented about this earlier and @JannePeltonen had confirmed that we can ignore this checkpatch warning.

Do you want me to rebase?

psavol:
> 
> 
> > Checkpatch fails due to extra braces. @anoobj please rebase and fix the checkpatch warnings.
> 
> But if I remove the braces, then the build would fail. I had commented about this earlier and @JannePeltonen had confirmed that we can ignore this checkpatch warning.
> 
> Do you want me to rebase?

OK, braces are due to the CUnit macro. I have put a comment ("braces needed by CU macro") inside the braces of the if-statement in these cases. Then checkpatch won't warn as you have at least two lines within one of the arms of the statement. You could do the same to avoid warnings when someone touches near those lines the next time.

Yes, rebase is needed now as this conflicts with the current master. I just merged the ipsec debug info PR.

JannePeltonen:
> But if I remove the braces, then the build would fail. I had commented about this earlier and @JannePeltonen had confirmed that we can ignore this checkpatch warning.

Yes. The braces are necessary since CUnit macros are not defined properly (they lack fo do { } while (0) idiom).

Please rebase to resolve the conflict.

anoobj:
> OK, braces are due to the CUnit macro. I have put a comment ("braces needed by CU macro") inside the braces of the if-statement in these cases. Then checkpatch won't warn as you have at least two lines within one of the arms of the statement. You could do the same to avoid warnings when someone touches near those lines the next time.

Agreed. Will add it.
 
> Yes, rebase is needed now as this conflicts with the current master. I just merged the ipsec debug info PR.

Will do that. Didn't notice that the other PR was merged.

anoobj:
@JannePeltonen @psavol I've added comments as you suggested. Rebased patches as well.

PR #1061, opened 2020-10-07
[PATCH v4] test: l2fwd: enable promisc mode for dev that support it

nithind1988:
Ideally l2fwd should have promiscuous enabled.
This change is inline with example/l2fwd_simple application.

### DIFF
--- a/test/performance/odp_l2fwd.c
+++ b/test/performance/odp_l2fwd.c
@@ -94,6 +94,7 @@ typedef struct {
 	int pool_per_if;        /* Create pool per interface */
 	uint32_t num_pkt;       /* Number of packets per pool */
 	int verbose;		/* Verbose output */
+	int promisc_mode;       /* Promiscuous mode enabled */
 } appl_args_t;
 
 /* Statistics */
@@ -791,6 +792,20 @@ static int create_pktio(const char *dev, int idx, int num_rx, int num_tx,
 
 	odp_pktio_config(pktio, &config);
 
+	if (gbl_args->appl.promisc_mode) {
+		if (!pktio_capa.set_op.op.promisc_mode) {
+			ODPH_ERR("Error: promisc mode set not supported %s\n",
+				 dev);
+			return -1;
+		}
+
+		/* Enable promisc mode */
+		if (odp_pktio_promisc_mode_set(pktio, true)) {
+			ODPH_ERR("Error: promisc mode enable failed %s\n", dev);
+			return -1;
+		}
+	}
+
 	odp_pktin_queue_param_init(&pktin_param);
 	odp_pktout_queue_param_init(&pktout_param);
 
@@ -1229,6 +1244,7 @@ static void usage(char *progname)
 	       "                          1: Create a pool per interface\n"
 	       "  -n, --num_pkt <num>     Number of packets per pool. Default is 16k or\n"
 	       "                          the maximum capability. Use 0 for the default.\n"
+	       "  -P, --promisc_mode      Enable promiscuous mode.\n"
 	       "  -v, --verbose           Verbose output.\n"
 	       "  -h, --help              Display help and exit.\n\n"
 	       "\n", NO_PATH(progname), NO_PATH(progname), MAX_PKTIOS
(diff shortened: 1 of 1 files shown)


JereLeppanen (commented):
This may be pedantic, so take this as optional, but it would be logical to set promisc mode right after the call to odp_pktio_config(), not between the queue setup and printing a message about queue setup. Also, the commit comment could just state the actual reason for this change, i.e. that we want to receive all packets.

nithind1988:
> 
> 
> This may be pedantic, so take this as optional, but it would be logical to set promisc mode right after the call to odp_pktio_config(), not between the queue setup and printing a message about queue setup. Also, the commit comment could just state the actual reason for this change, i.e. that we want to receive all packets.

Ack. Will fix it in V2.

JereLeppanen (commented):
`Reviewed-by: Jere Leppänen <jere.leppanen@nokia.com>`

test/performance/odp_l2fwd.c
```diff
@@ -791,6 +791,11 @@ static int create_pktio(const char *dev, int idx, int num_rx, int num_tx,
 
 	odp_pktio_config(pktio, &config);
 
+	/* Explicitly enable promisc mode for devices that support
+	 * since we need to perform L2 forwarding.
+	 */
+	odp_pktio_promisc_mode_set(pktio, true);
```
matiaselo, line 797: `odp_pktio_capability_t.set_op.op.promisc_mode` can be used to check if promisc mode is supported. After checking the capability, `odp_pktio_promisc_mode_set()`return value should also be checked.

The remaining question is do we want to always enable promiscuous mode and hence change the application default behavior? To maintain current default behavior promiscuous mode could be enabled using a command line option.

test/performance/odp_l2fwd.c
```diff
@@ -791,6 +791,11 @@ static int create_pktio(const char *dev, int idx, int num_rx, int num_tx,
 
 	odp_pktio_config(pktio, &config);
 
+	/* Explicitly enable promisc mode for devices that support
```
matiaselo, line 794: `it` missing

test/performance/odp_l2fwd.c
nithind1988: Ok. I can check for the promisc mode set op capability and return value. But do we need to leave it as an option ?
Ideally shouldn't we have L2 fwd with promic mode ?

example/l2fwd_simple also by default enables promisc

test/performance/odp_l2fwd.c
nithind1988: Ack.

test/performance/odp_l2fwd.c
matiaselo: We are using  l2fwd app widely for testing and would like to keep the processing similar than what a "normal" application would do. In most cases promisc mode wouldn't be enabled.

Also, making promisc mode default would decrease CI coverage since packets with invalid MAC addresses would be accepted and possible problems in the packet generator application would go unnoticed.

psavol: Agree with Matias that promisc mode should not be required/used by default, but should be a command line option. For example, "-P" character could be used for it "-P, --promisc_mode".

nithind1988:
@matiaselo It seems to be some unrelated CI failure. Can I retrigger it ?

matiaselo:
> It seems to be some unrelated CI failure. Can I retrigger it ?

Done

test/performance/odp_l2fwd.c
```diff
@@ -791,6 +792,18 @@ static int create_pktio(const char *dev, int idx, int num_rx, int num_tx,
...
+	if (pktio_capa.set_op.op.promisc_mode) {
+		bool promisc_mode = gbl_args->appl.promisc_mode;
+
+		/* Enable/Disable promisc mode */
+		if (odp_pktio_promisc_mode_set(pktio, promisc_mode)) {
+			ODPH_ERR("Error: promisc mode set failed %s\n", dev);
+			return -1;
+		}
+		printf("Promisc mode %s\n",
+		       promisc_mode ? "enabled" : "disabled");
```
matiaselo, line 804: This print would fit better inside `print_info()` function.

test/performance/odp_l2fwd.c
```diff
@@ -94,6 +94,7 @@ typedef struct {
 	int pool_per_if;        /* Create pool per interface */
 	uint32_t num_pkt;       /* Number of packets per pool */
 	int verbose;		/* Verbose output */
+	int promisc_mode;      /* Promiscuous mode enabled */
```
matiaselo, line 97: One space missing from comment alignment.

test/performance/odp_l2fwd.c
```diff
@@ -791,6 +792,18 @@ static int create_pktio(const char *dev, int idx, int num_rx, int num_tx,
 
 	odp_pktio_config(pktio, &config);
 
+	if (pktio_capa.set_op.op.promisc_mode) {
```
matiaselo, line 795: Simply fail if user has enabled promisc mode but the implementation doesn't support it.

matiaselo:
@nithind1988 I had couple minor comments. You can add my reviewed-by tags already in the next PR version.

test/performance/odp_l2fwd.c
nithind1988: Ok.

test/performance/odp_l2fwd.c
nithind1988: Ok.

test/performance/odp_l2fwd.c
nithind1988: ok.

matiaselo (approved):

PR #1063, opened 2020-10-09
[PATCH v3] validation: ipsec: fix uninitized access of odp_packet

asasidharan:
The field, pkto, is used to hold packets after IPsec processing.
Performing subtype check without checking the number of packets actually
processed by the API can lead to uninitialized access.

Also, changed the checks to compare against the actual number of
expected packets (ie, 1). Comparing against the variable can lead to
added complexity and incorrect results.


JannePeltonen:
Looks good to me.

asasidharan:
Fixed typo in commit short log and added "Reviewed-by" tag in V3.

JannePeltonen (approved):

PR #1065, opened 2020-10-12
[PATCH v2] test: packet_gen: use correct packet length when creating packet pool

matiaselo:
When using random packet lengths the pool has to be able to allocate
max length packets.


JereLeppanen (commented):
`Reviewed-by: Jere Leppänen <jere.leppanen@nokia.com>`

PR #1067, opened 2020-10-12
[PATCH v1] validation: ipsec: fix result check for outbound

asasidharan:
Fix result check for outbound packet in the
ipsec_check_out_in_one() function so that it checks only
whether the IPSEC processing has any error and not look for
specific bits in result.status.error meant for the inbound
test case.

### DIFF
--- a/test/validation/api/ipsec/ipsec.c
+++ b/test/validation/api/ipsec/ipsec.c
@@ -919,8 +919,7 @@ void ipsec_check_out_in_one(const ipsec_test_part *part,
 		} else {
 			/* IPsec packet */
 			CU_ASSERT_EQUAL(0, odp_ipsec_result(&result, pkto[i]));
-			CU_ASSERT_EQUAL(part->out[i].status.error.all,
-					result.status.error.all);
+			CU_ASSERT_EQUAL(0, result.status.error.all);
 			CU_ASSERT_EQUAL(sa, result.sa);
 			CU_ASSERT_EQUAL(IPSEC_SA_CTX,
 					odp_ipsec_sa_context(sa));


JannePeltonen:
I could not see what problem gets fixed by this. If this does fix a problem, the commit message could describe it better.

> not look for specific bits in result.status.error meant for the inbound test case.

I did not understand this. The current code is looking at all the bits, not any specific ones.

asasidharan:
> 
> 
> I could not see what problem gets fixed by this. If this does fix a problem, the commit message could describe it better.
> 
> > not look for specific bits in result.status.error meant for the inbound test case.
> 
> I did not understand this. The current code is looking at all the bits, not any specific ones.

Now when I think through, the real issue seems to be lack of information present in ipsec_test_part structure about the test operations covered. The ipsec_check_out_in_one() performs both outbound and inbound IPsec operations, but it has only one status structure as part of ipsec_test_part parameter to be compared against the result of both the outbound and the inbound operation:
ipsec_test_part::out::status

I was trying to leverage the existing ipsec_check_out_in_one() function to add some negative test cases for antireplay in IPsec IN Suite. I am setting the ipsec_test_part::out::status.error.antireplay bit for the negative test cases.

But with the existing version of ipsec_check_out_one() function, the test fails because the antireplay error set for the test affects the outbound operations result comparison. Event though we don't expect any error in the outbound processed packet, the check in question mandates the outbound processed packet also has the all the error bits set exactly as that is specified in ipsec_test_part::out::status.error,

Following are the paths I can think of to move us forward:
1. To allow outbound + inbound processing as a single function, add an additional ipsec_test_part::in::status structure which would represent the status expected for inbound processed packets.
2. Add another variant of ipsec_check_out_in_one() function to be used for the negative tests in IN Suite

Let me know your thoughts on this.

JannePeltonen:
Option 2 sounds like there would be two functions that are almost similar but not quite, which may lead to unnecessary duplicated code.

It is a bit hard to have an opinion on this patch in isolation. This probably should be part of a patch series that introduces the new tests that you are referring to, or even squashed into another commit. Maybe you original change is ok too, but then it would not be called a fix but something like a cleanup or preparation for subsequent commits.

In general I do not have anything against even big changes to this validation test code if you feel like refactoring it.

JannePeltonen:
It seems this has been made obsolete by the changes coming through PR 1060 (api: ipsec: add IPsec stats), thus I am closing this. If you disagree, please reopen.

PR #1068, opened 2020-10-13
[PATCH v13] api: ipsec: add odp_ipsec_sa_info

asasidharan:
The proposed API, odp_ipsec_sa_info(), allows application to check
parameters associated with an IPsec SA. Retrieved information
covers constant fields that application had provided while
creating session (like details of cipher choices) and variables
associated with IPsec (like sequence number, anti-replay parameters).
Also, this API allows application to check session constants which
implementation is allowed to update based on the support available
(like anti-replay window)

This pull request consists of:
1. The API spec update for odp_ipsec_sa_info()
2. An implementation of odp_ipsec_sa_info() for linux-generic platform
3. An oubound test function to verify the API.

### DIFF
--- a/include/odp/api/spec/ipsec.h
+++ b/include/odp/api/spec/ipsec.h
@@ -21,6 +21,7 @@ extern "C" {
 #include <odp/api/crypto.h>
 #include <odp/api/support.h>
 #include <odp/api/packet_io.h>
+#include <odp/api/protocols.h>
 #include <odp/api/classification.h>
 #include <odp/api/traffic_mngr.h>
 
@@ -808,6 +809,76 @@ typedef struct odp_ipsec_stats_t {
 	/** Number of packets with hard lifetime(packets) expired */
 	uint64_t hard_exp_pkts_err;
 } odp_ipsec_stats_t;
+
+/**
+ * IPSEC SA information
+ */
+typedef struct odp_ipsec_sa_info_t {
+	/** Copy of IPSEC Security Association (SA) parameters */
+	odp_ipsec_sa_param_t param;
+
+	/** IPSEC SA direction dependent parameters */
+	union {
+		/** Inbound specific parameters */
+		struct {
+			/** Additional SA lookup parameters. */
+			struct {
+				/** IP destination address (NETWORK ENDIAN) to
+				 *  be matched in addition to SPI value. */
+				uint8_t dst_addr[ODP_IPV6_ADDR_SIZE];
+			} lookup_param;
+
+			/** Antireplay window size
+			 *
+			 * Antireplay window size configured for the SA.
+			 * This value can be different from what application
+			 * had requested.
+			 */
+			uint32_t antireplay_ws;
+
+			/** Antireplay window top
+			 *
+			 * Sequence number representing a recent top of the
+			 * anti-replay window. There may be a delay before the
+			 * SA state is reflected in the value. The value will be
+			 * zero if no packets have been processed or if the
+			 * anti-replay service is not enabled.
+			 */
+			uint64_t antireplay_window_top;
+		} inbound;
+
+		/** Outbound specific parameters */
+		struct {
+			/** Sequence number
+			 *
+			 * Sequence number used for a recently processed packet.
+			 * There may be a delay before the SA state is reflected
+			 * in the value. When no packets have been processed,
+			 * the value will be zero.
+			 */
+			uint64_t seq_num;
+
+			/** Tunnel IP address */
+			union {
+				/** IPv4 */
+				struct {
+					/** IPv4 source address */
+					uint8_t src_addr[ODP_IPV4_ADDR_SIZE];
+					/** IPv4 destination address */
+					uint8_t dst_addr[ODP_IPV4_ADDR_SIZE];
+				} ipv4;
+
+				/** IPv6 */
+				struct {
+					/** IPv6 source address */
+					uint8_t src_addr[ODP_IPV6_ADDR_SIZE];
+					/** IPv6 destination address */
+					uint8_t dst_addr[ODP_IPV6_ADDR_SIZE];
+				} ipv6;
+			} tunnel;
+		} outbound;
+	};
+} odp_ipsec_sa_info_t;
 
 /**
  * Query IPSEC capabilities
@@ -1702,6 +1773,25 @@ int odp_ipsec_stats(odp_ipsec_sa_t sa, odp_ipsec_stats_t *stats);
  * @retval <0 on failure
  */
 int odp_ipsec_stats_multi(odp_ipsec_sa_t sa[], odp_ipsec_stats_t stats[], int num);
+
+/**
+ * Retrieve information about an IPSEC SA
+ *
+ * The cipher and auth key data(including key extra) will not be exposed and
+ * the corresponding pointers will be set to NULL. The IP address pointers
+ * will point to the corresponding buffers available in the SA info structure.
+ *
+ * The user defined SA context pointer is an opaque field and hence the value
+ * provided during the SA creation will be returned.
+ *
+ * @param      sa       The IPSEC SA for which to retrieve information
+ * @param[out] sa_info  Pointer to caller allocated SA info structure to be
+ *                      filled in
+ *
+ * @retval 0            On success
+ * @retval <0           On failure
+ **/
+int odp_ipsec_sa_info(odp_ipsec_sa_t sa, odp_ipsec_sa_info_t *sa_info);
 
 /**
  * @}
--- a/platform/linux-generic/include/odp_ipsec_internal.h
+++ b/platform/linux-generic/include/odp_ipsec_internal.h
@@ -207,6 +207,8 @@ struct ipsec_sa_s {
 		 */
 		odp_atomic_u64_t post_lifetime_err_pkts;
 	} stats;
+
+	odp_ipsec_sa_param_t param;
 };
 
 /**
--- a/test/validation/api/ipsec/ipsec_test_out.c
+++ b/test/validation/api/ipsec/ipsec_test_out.c
@@ -1230,6 +1230,127 @@ static void test_out_ipv4_udp_esp_null_sha256(void)
 	ipsec_sa_destroy(sa);
 }
 
+static void test_sa_info(void)
+{
+	uint32_t src = IPV4ADDR(10, 0, 111, 2);
+	uint32_t dst = IPV4ADDR(10, 0, 222, 2);
+	odp_ipsec_tunnel_param_t tunnel_out;
+	odp_ipsec_tunnel_param_t tunnel_in;
+	odp_ipsec_sa_param_t param_out;
+	odp_ipsec_sa_param_t param_in;
+	odp_ipsec_sa_info_t info_out;
+	odp_ipsec_sa_info_t info_in;
+	odp_ipsec_capability_t capa;
+	odp_ipsec_sa_t sa_out;
+	odp_ipsec_sa_t sa_in;
+
+	CU_ASSERT_EQUAL(0, odp_ipsec_capability(&capa));
+
+	memset(&tunnel_out, 0, sizeof(tunnel_out));
+	memset(&tunnel_in, 0, sizeof(tunnel_in));
+
+	tunnel_out.type = ODP_IPSEC_TUNNEL_IPV4;
+	tunnel_out.ipv4.src_addr = &src;
+	tunnel_out.ipv4.dst_addr = &dst;
+
+	ipsec_sa_param_fill(&param_out,
+			    false, false, 123, &tunnel_out,
+			    ODP_CIPHER_ALG_AES_CBC, &key_a5_128,
+			    ODP_AUTH_ALG_SHA1_HMAC, &key_5a_160,
+			    NULL, NULL);
+
+	sa_out = odp_ipsec_sa_create(&param_out);
+
+	CU_ASSERT_NOT_EQUAL_FATAL(ODP_IPSEC_SA_INVALID, sa_out);
+
+	ipsec_sa_param_fill(&param_in,
+			    true, false, 123, &tunnel_in,
+			    ODP_CIPHER_ALG_AES_CBC, &key_a5_128,
+			    ODP_AUTH_ALG_SHA1_HMAC, &key_5a_160,
+			    NULL, NULL);
+
+	param_in.inbound.antireplay_ws = 32;
+	sa_in = odp_ipsec_sa_create(&param_in);
+
+	memset(&info_out, 0, sizeof(info_out));
+	CU_ASSERT_EQUAL_FATAL(0, odp_ipsec_sa_info(sa_out, &info_out));
+
+	CU_ASSERT_EQUAL(info_out.param.dir, param_out.dir);
+	CU_ASSERT_EQUAL(info_out.param.proto, param_out.proto);
+	CU_ASSERT_EQUAL(info_out.param.mode, param_out.mode);
+
+	CU_ASSERT_EQUAL(info_out.param.crypto.cipher_alg,
+			param_out.crypto.cipher_alg);
+	CU_ASSERT_EQUAL(info_out.param.crypto.auth_alg,
+			param_out.crypto.auth_alg);
+	CU_ASSERT_EQUAL(info_out.param.opt.udp_encap, param_out.opt.udp_encap);
+	CU_ASSERT_EQUAL(info_out.param.spi, param_out.spi);
+	CU_ASSERT_EQUAL(info_out.param.opt.esn, param_out.opt.esn);
+	CU_ASSERT_EQUAL(info_out.param.opt.udp_encap, param_out.opt.udp_encap);
+	CU_ASSERT_EQUAL(info_out.param.opt.copy_dscp, param_out.opt.copy_dscp);
+	CU_ASSERT_EQUAL(info_out.param.opt.copy_flabel, param_out.opt.copy_flabel);
+	CU_ASSERT_EQUAL(info_out.param.opt.copy_df, param_out.opt.copy_df);
+
+	CU_ASSERT_EQUAL(ODP_IPSEC_MODE_TUNNEL, info_out.param.mode);
+
+	CU_ASSERT_EQUAL(info_out.param.outbound.tunnel.type,
+			param_out.outbound.tunnel.type);
+	CU_ASSERT_EQUAL(info_out.param.outbound.tunnel.ipv4.dscp,
+			param_out.outbound.tunnel.ipv4.dscp);
+	CU_ASSERT_EQUAL(info_out.param.outbound.tunnel.ipv4.df,
+			param_out.outbound.tunnel.ipv4.df);
+	CU_ASSERT_NOT_EQUAL_FATAL(NULL,
+				  info_out.param.outbound.tunnel.ipv4.src_addr);
+	CU_ASSERT_EQUAL(0, memcmp(info_out.param.outbound.tunnel.ipv4.src_addr,
+				  param_out.outbound.tunnel.ipv4.src_addr,
+				  ODP_IPV4_ADDR_SIZE));
+	CU_ASSERT_NOT_EQUAL_FATAL(NULL,
+				  info_out.param.outbound.tunnel.ipv4.dst_addr);
+	CU_ASSERT_EQUAL(0, memcmp(info_out.param.outbound.tunnel.ipv4.dst_addr,
+				  param_out.outbound.tunnel.ipv4.dst_addr,
+				  ODP_IPV4_ADDR_SIZE));
+
+	CU_ASSERT_EQUAL(info_out.param.lifetime.soft_limit.bytes,
+			param_out.lifetime.soft_limit.bytes);
+	CU_ASSERT_EQUAL(info_out.param.lifetime.hard_limit.bytes,
+			param_out.lifetime.hard_limit.bytes);
+	CU_ASSERT_EQUAL(info_out.param.lifetime.soft_limit.packets,
+			param_out.lifetime.soft_limit.packets);
+	CU_ASSERT_EQUAL(info_out.param.lifetime.hard_limit.packets,
+			param_out.lifetime.hard_limit.packets);
+
+	CU_ASSERT_EQUAL(0, info_out.outbound.seq_num);
+
+	memset(&info_in, 0, sizeof(info_in));
+	CU_ASSERT_EQUAL_FATAL(0, odp_ipsec_sa_info(sa_in, &info_in));
+	CU_ASSERT_EQUAL(0, info_in.inbound.antireplay_window_top);
+
+	ipsec_test_part test = {
+		.pkt_in = &pkt_ipv4_icmp_0,
+		.num_pkt = 1,
+		.out = {
+			{ .status.warn.all = 0,
+			  .status.error.all = 0,
+			  .l3_type = ODP_PROTO_L3_TYPE_IPV4,
+			  .l4_type = ODP_PROTO_L4_TYPE_ICMPV4,
+			  .pkt_res = &pkt_ipv4_icmp_0 },
+		},
+	};
+
+	ipsec_check_out_in_one(&test, sa_out, sa_in);
+
+	memset(&info_out, 0, sizeof(info_out));
+	CU_ASSERT_EQUAL_FATAL(0, odp_ipsec_sa_info(sa_out, &info_out));
+	CU_ASSERT_EQUAL(1, info_out.outbound.seq_num);
+
+	memset(&info_in, 0, sizeof(info_in));
+	CU_ASSERT_EQUAL_FATAL(0, odp_ipsec_sa_info(sa_in, &info_in));
+	CU_ASSERT_EQUAL(1, info_in.inbound.antireplay_window_top);
+
+	ipsec_sa_destroy(sa_out);
+	ipsec_sa_destroy(sa_in);
+}
+
 static void ipsec_test_capability(void)
 {
 	odp_ipsec_capability_t capa;
@@ -1308,6 +1429,8 @@ odp_testinfo_t ipsec_out_suite[] = {
 				  ipsec_check_esp_null_sha256),
 	ODP_TEST_INFO_CONDITIONAL(test_out_ipv4_udp_esp_null_sha256,
 				  ipsec_check_esp_null_sha256),
+	ODP_TEST_INFO_CONDITIONAL(test_sa_info,
+				  ipsec_check_esp_aes_cbc_128_sha1),
 	ODP_TEST_INFO(test_esp_out_in_all_basic),
 	ODP_TEST_INFO(test_ah_out_in_all),
 	ODP_TEST_INFO(test_ipsec_stats),
--- a/include/Makefile.am
+++ b/include/Makefile.am
@@ -34,6 +34,7 @@ odpapiinclude_HEADERS = \
 	odp/api/packet_flags.h \
 	odp/api/packet_io.h \
 	odp/api/packet_io_stats.h \
+	odp/api/protocols.h \
 	odp/api/pool.h \
 	odp/api/queue.h \
 	odp/api/random.h \
@@ -84,6 +85,7 @@ odpapispecinclude_HEADERS = \
 		  odp/api/spec/packet_flags.h \
 		  odp/api/spec/packet_io.h \
 		  odp/api/spec/packet_io_stats.h \
+		  odp/api/spec/protocols.h \
 		  odp/api/spec/pool.h \
 		  odp/api/spec/queue.h \
 		  odp/api/spec/queue_types.h \
--- a/include/odp/api/protocols.h
+++ b/include/odp/api/protocols.h
@@ -1,0 +1,26 @@
+/* Copyright (c) 2020, Marvell
+ * All rights reserved.
+ *
+ * SPDX-License-Identifier:     BSD-3-Clause
+ */
+
+/**
+ * @file
+ *
+ * ODP protocols
+ */
+
+#ifndef ODP_API_PROTOCOLS_H_
+#define ODP_API_PROTOCOLS_H_
+
+#ifdef __cplusplus
+extern "C" {
+#endif
+
+#include <odp/api/spec/protocols.h>
+
+#ifdef __cplusplus
+}
+#endif
+
+#endif
--- a/include/odp/api/spec/protocols.h
+++ b/include/odp/api/spec/protocols.h
@@ -1,0 +1,44 @@
+/* Copyright (c) 2020, Marvell
+ * All rights reserved.
+ *
+ * SPDX-License-Identifier:     BSD-3-Clause
+ */
+
+/**
+ * @file
+ *
+ * ODP protocols
+ */
+
+#ifndef ODP_API_SPEC_PROTOCOLS_H_
+#define ODP_API_SPEC_PROTOCOLS_H_
+#include <odp/visibility_begin.h>
+
+#ifdef __cplusplus
+extern "C" {
+#endif
+
+/**
+ * @addtogroup odp_protocols
+ * @details
+ * <b> Protocols </b>
+ *
+ *  @{
+ */
+
+/** IPv4 address size */
+#define ODP_IPV4_ADDR_SIZE 4
+
+/** IPv6 address size */
+#define ODP_IPV6_ADDR_SIZE 16
+
+/**
+ * @}
+ */
+
+#ifdef __cplusplus
+}
+#endif
+
+#include <odp/visibility_end.h>
+#endif
--- a/platform/linux-generic/odp_ipsec_sad.c
+++ b/platform/linux-generic/odp_ipsec_sad.c
@@ -467,6 +467,20 @@ odp_ipsec_sa_t odp_ipsec_sa_create(const odp_ipsec_sa_param_t *param)
 	odp_atomic_init_u64(&ipsec_sa->stats.hard_exp_pkts_err, 0);
 	odp_atomic_init_u64(&ipsec_sa->stats.post_lifetime_err_pkts, 0);
 
+	/* Copy application provided parameter values. */
+	ipsec_sa->param = *param;
+
+	/* Set all the key related pointers and ip address pointers to null. */
+	ipsec_sa->param.crypto.cipher_key.data = NULL;
+	ipsec_sa->param.crypto.cipher_key_extra.data = NULL;
+	ipsec_sa->param.crypto.auth_key.data = NULL;
+	ipsec_sa->param.crypto.auth_key_extra.data = NULL;
+	ipsec_sa->param.inbound.lookup_param.dst_addr = NULL;
+	ipsec_sa->param.outbound.tunnel.ipv4.src_addr = NULL;
+	ipsec_sa->param.outbound.tunnel.ipv4.dst_addr = NULL;
+	ipsec_sa->param.outbound.tunnel.ipv6.src_addr = NULL;
+	ipsec_sa->param.outbound.tunnel.ipv6.dst_addr = NULL;
+
 	if (ODP_IPSEC_MODE_TUNNEL == ipsec_sa->mode &&
 	    ODP_IPSEC_DIR_OUTBOUND == param->dir) {
 		if (ODP_IPSEC_TUNNEL_IPV4 == param->outbound.tunnel.type) {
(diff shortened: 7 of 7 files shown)


include/odp/api/spec/ipsec.h
```diff
@@ -762,6 +762,37 @@ typedef struct odp_ipsec_sa_param_t {
...
+	union {
+		/** Inbound specific internal parameters */
+		struct {
+			/** Antireplay window top
+			 *
+			 * Sequence number representing the top of the
+			 * antireplay window.
+			 */
+			uint64_t ar_top;
+		} inb;
```
psavol, line 782: Similarly to the union in odp_ipsec_sa_param_t, I'd prefer to use full words as struct named for "inbound" and "outbound" (inb->inbound).

include/odp/api/spec/ipsec.h
```diff
@@ -1601,6 +1632,18 @@ int odp_ipsec_sa_mtu_update(odp_ipsec_sa_t sa, uint32_t mtu);
  */
 void *odp_ipsec_sa_context(odp_ipsec_sa_t sa);
 
+/**
+ * Retrieve information about an IPSEC SA
+ *
+ * @param      sa       The IPSEC SA for which parameter value to be set
```
psavol, line 1638: "... for which to retrieve information"

include/odp/api/spec/ipsec.h
psavol: Also ar_top should be extended to "antireplay_top" or something similar. We use full word "antireplay" in other parts of the API.

include/odp/api/spec/ipsec.h
asasidharan: Will fix it in the next revision.

include/odp/api/spec/ipsec.h
asasidharan: > 
> 
> Similarly to the union in odp_ipsec_sa_param_t, I'd prefer to use full words as struct named for "inbound" and "outbound" (inb->inbound).

Agreed. Will update it to inbound and outbound in the next revision.


> 
> 
> Also ar_top should be extended to "antireplay_top" or something similar. We use full word "antireplay" in other parts of the API.

Would "antireplay_window_top" be fine?

JannePeltonen:
Why do we need this API?

The part that returns SA creation parameters looks redundant to me. The application has created the SA and should know the parameters it used at creation time. We do not have this kind of parameter query API for things like queues or crypto sessions either.

The part that reports the replay window status and the latest sequence number looks potentially problematic to me because of the need to synchronize between this API call and the actual data path processing. It might help if the API was not required to return the latest status but some past status (that may or may not be the latest).

asasidharan:
> 
> 
> Why do we need this API?
> 

The API is needed for validation and debug purposes. The API is not intended for use in fast path.

> The part that returns SA creation parameters looks redundant to me. The application has created the SA and should know the parameters it used at creation time.

We had referred to odp_pool_info() API while drafting this change.
ODP pool spec defines certain fields in pool parameters where the application requests pool creation with a minimal required value. The implementation then can configure the pool which meets the requirement but need not be the exact same configuration requested by the application. One of such fields  in pool parameter is:
`uint32_t   align  | Minimum buffer alignment in bytes`

We have a similar situation with antireplay window size where the application can request for a minimal required window size, but if the implementation doesn't support that exact value, it would go for a higher window size value.

>  We do not have this kind of parameter query API for things like queues or crypto sessions either.

But in the current spec, I can see that the following odp_xyz_info() APIs do have their corresponding odp_xyz_param_t structure embedded in it's odp_xyz_info_t:
odp_pool_info()
odp_pktio_info()
odp_queue_info()

> 
> The part that reports the replay window status and the latest sequence number looks potentially problematic to me because of the need to synchronize between this API call and the actual data path processing. It might help if the API was not required to return the latest status but some past status (that may or may not be the latest).

Agreed. I can update the spec like following if that is fine:
```
		struct {
			/** Antireplay window top
			 *
			 * Sequence number representing a recent top of the
			 * antireplay window.
			 */
			uint64_t ar_top;
		} inb;

		/** Outbound specific internal parameters */
		struct {
			/** Sequence number value
			 *
			 * Sequence number used for a recently processed
			 * packet by the IPSEC SA. Need not be the latest.
			 */
			uint64_t seq_num;
		} outb;
```

JannePeltonen:
Sorry, I missed those existing info APIs. The proposed SA info API seems to follow the pattern, so it should be ok.

> We had referred to odp_pool_info() API while drafting this change.

odp_pool_info() documentation says that the returned params is "Copy of pool parameters"- As an application writer I would expect that field then to be a copy of the pool parameters that I supplied when I created the pool. From application point-of-view there are no other pool parameters. I would argue that the current implementation of the pool info is not according to the spec (or rather, the spec does not convey the intended behaviour).

I suppose the lesson here is that the API spec should be very clear whether the parameters returned by the info call are a copy of the original creation parameters or something else. This is something that should be improved in this patch for the IPsec info API.

JannePeltonen:
> > It might help if the API was not required to return the latest status but some past status (that may or may not be the latest).

> Agreed. I can update the spec like following if that is fine:

That looks good to me.

include/odp/api/spec/ipsec.h
```diff
@@ -762,6 +762,37 @@ typedef struct odp_ipsec_sa_param_t {
 
 } odp_ipsec_sa_param_t;
 
+/**
+ * Retrieve IPSEC SA information
+ */
+typedef struct odp_ipsec_sa_info_t {
+	/** IPSEC Security Association (SA) parameters */
+	odp_ipsec_sa_param_t param;
```
psavol, line 770: As discussed last week, current pattern in ODP info APIs is that if there's param structure included, it is a copy of parameters from application (from sa create). Additional info fields represent data that is dynamic/implementation picked values.

Params may be included here or not (no strict opinion either way), since application should be able to remember param values used in sa creation phase.

include/odp/api/spec/ipsec.h
asasidharan: In V2, updated the following:
1. struct names in sa_info now uses full words: inbound and outbound.
2. Renamed ar_top to antireplay_window_top.

include/odp/api/spec/ipsec.h
asasidharan: Fixed this in V2.

include/odp/api/spec/ipsec.h
asasidharan: Agreed that the param structure in sa_info should be a copy of parameters provided by the application.
In V2, I am updating the description of the field to:
`       /** Copy of IPSEC Security Association (SA) parameters */ `

Let me know if this would be enough.

asasidharan:
In V2, introducing additional uint8_t buffers for returning cryptographic key data and for tunnel src and dst ip addresses in case of outbound sa. The intention of introducing these buffers is to simplify the handling of pointer fields in the odp_ipsec_sa_param_t structure. The API would update the pointer fields with the address of corresponding buffer in sa info structure .

include/odp/api/spec/ipsec.h
```diff
@@ -762,6 +762,79 @@ typedef struct odp_ipsec_sa_param_t {
...
 
+/**
+ * Retrieve IPSEC SA information
+ */
+typedef struct odp_ipsec_sa_info_t {
+	/** Copy of IPSEC Security Association (SA) parameters */
+	odp_ipsec_sa_param_t param;
+
+	/** Cipher key */
+	uint8_t cipher_key_data[32];
```
JannePeltonen, line 773: I do not think the API should provide any key data like this as it is not good practice from security point-of-view. This would prevent SW from forgetting the keys after writing them in write-only space of some hypothetical HW device.

As discussed, I do not think returning the param struct is necessary at all, but if others would like to have it then it is ok to me as long as it does not have the keys. If the param struct stays in here, I think the API could say that the struct content is the same as provided in the SA creation except that all pointers to key material are null.

The fixed sizes for the key arrays in this struct are not very nice either.

anoobj: > 
> I do not think the API should provide any key data like this as it is not good practice from security point-of-view. This would prevent SW from forgetting the keys after writing them in write-only space of some hypothetical HW device.
> 

Though we typically treat keys as a security concern, keys used in IPsec are more or less transient keys and is unlike HSMs where the keys are supposed to be inaccessible outside the system. Also, IPsec implementation in linux typically prints everything including keys. The intent was more to be inline with that.

Said that, if you think keys should not be exposed, we will try to make that as application responsibility rather than getting via sa_info. This will obviously cause more memory footprint since the same data need to be replicated for consumption of hardware and application book keeping. 

> As discussed, I do not think returning the param struct is necessary at all, but if others would like to have it then it is ok to me as long as it does not have the keys. If the param struct stays in here, I think the API could say that the struct content is the same as provided in the SA creation except that all pointers to key material are null.

Did you mean everything related to key (including salt) should not be exposed?

> The fixed sizes for the key arrays in this struct are not very nice either.

Since these are transient buffers, isn't it better to bundle everything into one entity (or structure) as done here. Other option is application passing a pointer of a buffer, but unlike SA creation, application doesn't know the length of key. So application would be required to create a buffer that can accommodate max key length. Also, it may not be possible to prevent possible buffer overflows.

I agree that the array width should not be hardcoded. It should be some macro.

JannePeltonen: I just do not see any need for returning the keys (or salt) to the application. If the application needs to remember the keys, it can store them itself.

Having the keys in the API would require an ODP implementation to keep the keys around even if it could avoid if (e.g. after loading them in some device).

If we do not return the keys, we avoid having to define the returning mechanism, including the allocation and ownership of the storage.

If we do end up returning the keys, then I think one viable option would be to have the application allocate the storage and provide max length to ODP.

anoobj: ODP implementations can always try to read back from device. That way application can be sure that parameters are setup correctly (like in the case of pool_info()).

But since you think keys should not be exposed, we will have this reworked.

JannePeltonen: > But since you think keys should not be exposed

Yes, I think so, but this is mostly still about not seeing the use case so that we should spend effort in defining and implementing the API.

> That way application can be sure that parameters are setup correctly (like in the case of pool_info()).

But if the implementation just stores a copy of the parameters and returns the copy to the application, it does not really tell if the parameters were translated correctly to implementation domain and set up correctly in the HW. So the API does not seem terribly useful to me to see that ODP is doing what it should.

But all of this is just my opinion. Maybe there are other opinions. @psavol?

asasidharan: Updated V4 with following change:
1. API requires that all the pointers in sa_param to be set to null.
2. cipher/auth key data buffers are not exposed with this API.

include/odp/api/spec/ipsec.h
```diff
@@ -772,6 +772,76 @@ typedef struct odp_ipsec_sa_param_t {
...
 } odp_ipsec_sa_param_t;
 
+/** IPv4 address size */
+#define ODP_IP4ADDR_SIZE 4
+
+/** IPv6 address size */
+#define ODP_IP6ADDR_SIZE 16
+
+/**
+ * Retrieve IPSEC SA information
```
JannePeltonen, line 782: This sounds like a description of a function, not a type. Just "SA information" would be better.

include/odp/api/spec/ipsec.h
```diff
@@ -772,6 +772,76 @@ typedef struct odp_ipsec_sa_param_t {
...
+ */
+typedef struct odp_ipsec_sa_info_t {
+	/** Copy of IPSEC Security Association (SA) parameters
+	 *
+	 * All pointers will be set to NULL.
```
JannePeltonen, line 787: Maybe the context pointer should not be set to null? It is opaque data to ODP and could as well be returned.

I am not sure this this ponters-set-to-null should even be described here instead of as part of odp_ipsec_sa_info() documentation.

Then I wonder if inbound.lookup_param.dst_addr should be returned since also tunnel addresses are returned?

include/odp/api/spec/ipsec.h
```diff
@@ -772,6 +772,76 @@ typedef struct odp_ipsec_sa_param_t {
...
+	 */
+	odp_ipsec_sa_param_t param;
+
+	/** Recent value of SA lifetime limits */
+	odp_ipsec_lifetime_t lifetime_recent;
```
JannePeltonen, line 792: I do not understand this. Lifetime limits do not change. They are set at SA creation time and it is not possible to alter them later. So the "recent" limits are the same as the initial ones in SA parameters and this field is redundant.

But maybe you think that this tells the *remaining* lifetime, i.e. the original packet and byte lifetimes subtracted by the number of packets and bytes processed, respectively? That would seem to duplicate the functionality of the statistics API that returns the number of packets processed (not bytes in current proposal, IIRC, but it can be added if needed). So even with this interpretation I do not see why this field would be needed here.

include/odp/api/spec/ipsec.h
```diff
@@ -1626,6 +1696,18 @@ int odp_ipsec_sa_mtu_update(odp_ipsec_sa_t sa, uint32_t mtu);
  */
 void *odp_ipsec_sa_context(odp_ipsec_sa_t sa);
 
+/**
+ * Retrieve information about an IPSEC SA
+ *
+ * @param      sa       The IPSEC SA for which to retrieve information
+ * @param[out] sa_info  Pointer to the structure which stores the IPSEC SA
```
JannePeltonen, line 1703: This wording makes it a bit unclear who allocates and owns the memory pointed to by sa_info. How about "pointer to caller allocated SA info structure to be filled in"?

platform/linux-generic/odp_ipsec.c
```diff
@@ -1917,6 +1917,57 @@ int odp_ipsec_out_inline(const odp_packet_t pkt_in[], int num_in,
...
+{
+	ipsec_sa_t *ipsec_sa;
+	odp_ipsec_dir_t dir;
+
+	ipsec_sa = _odp_ipsec_sa_entry_from_hdl(sa);
+
+	ODP_ASSERT(ipsec_sa != NULL);
+	ODP_ASSERT(sa_info != NULL);
+
+	memset(sa_info, 0, sizeof(odp_ipsec_sa_info_t));
```
JannePeltonen, line 1930: I would prefer memset(sa_info, 0, sizeof(*sa_info));

And the same comment for other occurences of this pattern.

platform/linux-generic/include/odp_ipsec_internal.h
```diff
@@ -190,6 +190,7 @@ struct ipsec_sa_s {
 			};
 		} out;
 	};
+	odp_ipsec_sa_param_t param;
```
JannePeltonen, line 193: This duplicates many of the parameters that are already stored in the SA,.

platform/linux-generic/odp_ipsec.c
```diff
@@ -1917,6 +1917,57 @@ int odp_ipsec_out_inline(const odp_packet_t pkt_in[], int num_in,
...
+
+	dir = ipsec_sa->param.dir;
+
+	sa_info->param = (odp_ipsec_sa_param_t)ipsec_sa->param;
```
JannePeltonen, line 1934: Now the pointers that were supposed to be null are set to their original non-null values.

platform/linux-generic/odp_ipsec.c
```diff
@@ -1917,6 +1917,57 @@ int odp_ipsec_out_inline(const odp_packet_t pkt_in[], int num_in,
...
+				memcpy(src,
+				       &ipsec_sa->out.tun_ipv6.src_ip,
+				       _ODP_IPV6ADDR_LEN);
+				memcpy(dst,
+				       &ipsec_sa->out.tun_ipv6.dst_ip,
+				       _ODP_IPV6ADDR_LEN);
+			}
+		}
+	} else {
+		ODP_ERR("Feature to be added\n");
```
JannePeltonen, line 1965: I think this needs to be implemented before this can be merged.

test/validation/api/ipsec/ipsec_test_out.c
```diff
@@ -1138,6 +1138,90 @@ static void test_out_ipv4_udp_esp_null_sha256(void)
 	ipsec_sa_destroy(sa);
 }
 
+static void test_out_ipv4_esp_aes_gcm_128_sa_info(void)
```
JannePeltonen, line 1141: How about naming this simply test_sa_info? I suppose the algorithm chosen for this test is not an essential and you are not planning to add a bunch of similar tests with different algorithms.

This test does not test the dynamic state info like recent sequence number.

include/odp/api/spec/ipsec.h
asasidharan: Agreed. Will update in the next version.

include/odp/api/spec/ipsec.h
asasidharan: > 
> 
> Maybe the context pointer should not be set to null? It is opaque data to ODP and could as well be returned.

Missed the context pointer. Agreed that it should be returned as it is.

> 
> I am not sure this this ponters-set-to-null should even be described here instead of as part of odp_ipsec_sa_info() documentation.

Agreed. Will move it to the odp_ipsec_sa_info() documentation.

> 
> Then I wonder if inbound.lookup_param.dst_addr should be returned since also tunnel addresses are returned?

Missed that. I will add buffer space for inbound.lookup_param.dst_addr as well.

include/odp/api/spec/ipsec.h
asasidharan: Agreed. Will update it to "Pointer to caller allocated SA info structure to be filled in".

platform/linux-generic/odp_ipsec.c
asasidharan: Agreed with the use of memset. Will update here and all other occurrence in this change.

platform/linux-generic/include/odp_ipsec_internal.h
asasidharan: This was just a sample implementation while we wait for the API to get mature. Agreed that params are getting repeated.

test/validation/api/ipsec/ipsec_test_out.c
asasidharan: > 
> 
> How about naming this simply test_sa_info? I suppose the algorithm chosen for this test is not an essential and you are not planning to add a bunch of similar tests with different algorithms.

test_sa_info sounds good. Yes, there is no importance on the algorithm chosen as long as the algorithm is supported on the platform.

> 
> This test does not test the dynamic state info like recent sequence number.

We can do something like, send one packet and then do an sa_info call to see that the seq_num was incremented in the same test.

include/odp/api/spec/ipsec.h
asasidharan: Remaining lifetime you just described was the idea. But agreed that the same can be achieved with the statistics API. I will drop the lifetime remaining field from the SA info structure.

asasidharan:
V5: Updated API spec with the followiing:
- Removed lifetime remaining fields
- Moved cipher/auth key data pointer-set-to-null documentation to the odp_ipsec_sa_info() specification
- Added buffer space for inbound.lookup.dst_addr in sa_info
- Other improvements to the documentation as agreed.

Let me know if the API spec looks OK now.

JannePeltonen:
> Let me know if the API spec looks OK now.

How should the other pointers than those for keys be set in the info structure? We discussed that the context pointer should be returned as is. That could be mentioned. But what about the tunnel address and lookup address pointers? Should they be returned as they were in the SA creation call (this is how I understand the current API, although ODP cannot know if the pointers are still valid) or should they point to the arrays that are part of the info structure or should they be set to null or should they have undefined values? I would not like the first alternative as it would require ODP to remember the original pointer values even though in many applications the values would be stale.

include/odp/api/spec/ipsec.h
```diff
@@ -1626,6 +1696,21 @@ int odp_ipsec_sa_mtu_update(odp_ipsec_sa_t sa, uint32_t mtu);
  */
 void *odp_ipsec_sa_context(odp_ipsec_sa_t sa);
 
+/**
+ * Retrieve information about an IPSEC SA
+ *
+ * The pointers for cipher and auth key data in odp_ipsec_sa_param_t field will
+ * be set to NULL.
```
psavol, line 1703: Please, mention also that IP address pointers point to sa_info data.

asasidharan: > 
> 
> > Let me know if the API spec looks OK now.
> 
> How should the other pointers than those for keys be set in the info structure? We discussed that the context pointer should be returned as is. That could be mentioned. But what about the tunnel address and lookup address pointers? Should they be returned as they were in the SA creation call (this is how I understand the current API, although ODP cannot know if the pointers are still valid) or should they point to the arrays that are part of the info structure or should they be set to null or should they have undefined values? I would not like the first alternative as it would require ODP to remember the original pointer values even though in many applications the values would be stale.

V6: Updated the spec saying:
1. context pointer is returned as it was programmed during sa creation
2. all the IP address pointers points to the corresponding buffers available in the SA info structure.

> 
> 
> Please, mention also that IP address pointers point to sa_info data.

Done.

JannePeltonen:
I do not have further comments on the API.

include/odp/api/spec/ipsec.h
```diff
@@ -772,6 +772,76 @@ typedef struct odp_ipsec_sa_param_t {
 
 } odp_ipsec_sa_param_t;
 
+/** IPv4 address size */
+#define ODP_IP4ADDR_SIZE 4
```
psavol, line 776: These could be defined in a new file (odp/api/spec/protocols.h).  ODP_IPV4_ADDR_SIZE, ODP_IPV6_ADDR_SIZE. So, that naming convention for IPv4 would be ODP_IPV4_XXX.

asasidharan: > 
> 
> These could be defined in a new file (odp/api/spec/protocols.h). ODP_IPV4_ADDR_SIZE, ODP_IPV6_ADDR_SIZE. So, that naming convention for IPv4 would be ODP_IPV4_XXX.

With V8, added separate commit adding protocols.h. Also as suggested, renamed the macros to ODP_IPV4_ADDR_SIZE and ODP_IPV6_ADDR_SIZE.

Let me know if the API spec for SA info looks fine now. I can continue improving on the linux-generic implementation if you can confirm the spec.

include/odp/api/spec/ipsec.h
```diff
@@ -772,6 +773,71 @@ typedef struct odp_ipsec_sa_param_t {
...
+	/** IPSEC SA direction dependent internal parameters */
+	union {
+		/** Inbound specific internal parameters */
+		struct {
+			/** Additional SA lookup parameters. */
+			struct {
+				/** IP destination address (NETWORK ENDIAN) to
+				 *  be matched in addition to SPI value. */
+				uint8_t dst_addr[ODP_IPV6_ADDR_SIZE];
+			} lookup_param;
```
JannePeltonen, line 792: How does the caller know if the address is IPv4 or IPv6? In the corresponding config API there is also IP version field.

include/odp/api/spec/ipsec.h
```diff
@@ -772,6 +773,71 @@ typedef struct odp_ipsec_sa_param_t {
...
 } odp_ipsec_sa_param_t;
 
+/**
+ * IPSEC SA information
+ */
+typedef struct odp_ipsec_sa_info_t {
+	/** Copy of IPSEC Security Association (SA) parameters */
+	odp_ipsec_sa_param_t param;
+
+	/** IPSEC SA direction dependent internal parameters */
```
JannePeltonen, line 783: Looking at this patch now, I think the term "internal parameters" that appear here and elsewhere is not very clear. Why not simply "parameters"?

include/odp/api/spec/ipsec.h
asasidharan: Didn't include the IP version separately as it is already available in the copy of SA param(param.inbound.lookup_param.ip_version).

include/odp/api/spec/ipsec.h
asasidharan: Ack. Made it simply "parameters" in v9.

include/odp/api/spec/ipsec.h
```diff
@@ -772,6 +773,71 @@ typedef struct odp_ipsec_sa_param_t {
...
+		} inbound;
+
+		/** Outbound specific parameters */
+		struct {
+			/** Sequence number
+			 *
+			 * Sequence number used for a recently processed packet
+			 * by the IPSEC SA. Need not be the latest.
+			 */
+			uint64_t seq_num;
```
JannePeltonen, line 817: One more comment on the API: This is not well defined when no packets have yet been processed. Should this be zero then? Or maybe this field could be the sequence number that will be used for the next successfully transformed packet?

platform/linux-generic/include/odp_ipsec_internal.h
JannePeltonen: The duplication of data is still there. Maybe it is ok for now until the max SA count in linux-gen is made bigger.

platform/linux-generic/odp_ipsec.c
```diff
@@ -1963,6 +1963,63 @@ int odp_ipsec_out_inline(const odp_packet_t pkt_in[], int num_in,
...
+
+	/* Set all the key related pointers to null. */
+	sa_info->param.crypto.cipher_key.data = NULL;
+	sa_info->param.crypto.cipher_key_extra.data = NULL;
+	sa_info->param.crypto.auth_key.data = NULL;
+	sa_info->param.crypto.auth_key_extra.data = NULL;
+
+	if (dir == ODP_IPSEC_DIR_OUTBOUND) {
+		sa_info->outbound.seq_num =
+			(uint64_t)odp_atomic_load_u64(&ipsec_sa->hot.out.seq);
```
JannePeltonen, line 1990: This returns the next sequence number that will be used, not adhering to the API spec in the first patch.

platform/linux-generic/odp_ipsec.c
JannePeltonen: This is still needed.

platform/linux-generic/odp_ipsec.c
```diff
@@ -1963,6 +1963,63 @@ int odp_ipsec_out_inline(const odp_packet_t pkt_in[], int num_in,
...
+	ipsec_sa = _odp_ipsec_sa_entry_from_hdl(sa);
+
+	ODP_ASSERT(ipsec_sa != NULL);
+	ODP_ASSERT(sa_info != NULL);
+
+	memset(sa_info, 0, sizeof(*sa_info));
+
+	dir = ipsec_sa->param.dir;
+
+	sa_info->param = (odp_ipsec_sa_param_t)ipsec_sa->param;
```
JannePeltonen, line 1980: The cast is unnecessary. ipsec_sa->param is of odp_ipsec_sa_param_t type.

test/validation/api/ipsec/ipsec_test_out.c
```diff
@@ -1138,6 +1138,128 @@ static void test_out_ipv4_udp_esp_null_sha256(void)
...
+			param_out.crypto.auth_alg);
+	CU_ASSERT_EQUAL(sa_info.param.opt.udp_encap, param_out.opt.udp_encap);
+	CU_ASSERT_EQUAL(sa_info.param.spi, param_out.spi);
+	CU_ASSERT_EQUAL(sa_info.param.opt.esn, param_out.opt.esn);
+	CU_ASSERT_EQUAL(sa_info.param.opt.udp_encap, param_out.opt.udp_encap);
+	CU_ASSERT_EQUAL(sa_info.param.opt.copy_dscp, param_out.opt.copy_dscp);
+	CU_ASSERT_EQUAL(sa_info.param.opt.copy_flabel, param_out.opt.copy_flabel);
+	CU_ASSERT_EQUAL(sa_info.param.opt.copy_df, param_out.opt.copy_df);
+
+	if (param_out.mode == ODP_IPSEC_MODE_TUNNEL) {
```
JannePeltonen, line 1201: Tunnel mode SA was created, so this should be an assert.

test/validation/api/ipsec/ipsec_test_out.c
```diff
@@ -1138,6 +1138,128 @@ static void test_out_ipv4_udp_esp_null_sha256(void)
...
+		CU_ASSERT_EQUAL(sa_info.param.outbound.tunnel.type,
+				param_out.outbound.tunnel.type);
+		if (param_out.outbound.tunnel.type == ODP_IPSEC_TUNNEL_IPV4) {
```
JannePeltonen, line 1204: IPv4 tunnel was created, so this should be an assert. The else branch is dead code.

include/odp/api/spec/ipsec.h
asasidharan: Yes, agreed that this case should be defined. When no packets have been processed, it should be zero. Shall I update the documentation as follows?:

```
 * Sequence number used for a recently processed packet(need not be the latest)
 * by the IPSEC SA. When no packets have been processed, this value will be zero.
 */
```

platform/linux-generic/odp_ipsec.c
asasidharan: Thanks for pointing this out. I will update the implementation to return one less than the value in ipsec_sa->hot.out.seq.

platform/linux-generic/odp_ipsec.c
asasidharan: Agreed.

test/validation/api/ipsec/ipsec_test_out.c
asasidharan: Ack.

test/validation/api/ipsec/ipsec_test_out.c
asasidharan: Ack.

platform/linux-generic/odp_ipsec.c
asasidharan: In v10, implemented inbound specific param retrieval for the SA info API.

platform/linux-generic/odp_ipsec.c
asasidharan: Updated the linux-gen implementation such that immediately after copying the original param during SA create, the key and IP address related pointers are initialized to NULL.

platform/linux-generic/odp_ipsec.c
```diff
@@ -1963,6 +1963,81 @@ int odp_ipsec_out_inline(const odp_packet_t pkt_in[], int num_in,
...
+	}
+
+	if (ipsec_sa->antireplay) {
+		sa_info->inbound.antireplay_ws = IPSEC_ANTIREPLAY_WS;
+		sa_info->inbound.antireplay_window_top =
+			_odp_ipsec_sa_antireplay_max_seq(ipsec_sa);
+	}
+}
+
+int odp_ipsec_sa_info(odp_ipsec_sa_t sa, odp_ipsec_sa_info_t  *sa_info)
```
JannePeltonen, line 2020: Maybe a better place for this API function and its callees would be odp_ipsec_sad.c where SA creation, anti-replay window update etc reside?

platform/linux-generic/odp_ipsec.c
```diff
@@ -1963,6 +1963,81 @@ int odp_ipsec_out_inline(const odp_packet_t pkt_in[], int num_in,
...
+	if (ipsec_sa->param.mode == ODP_IPSEC_MODE_TUNNEL) {
+		uint8_t *dst = sa_info->inbound.lookup_param.dst_addr;
+
+		if (ipsec_sa->param.inbound.lookup_param.ip_version ==
+		    ODP_IPSEC_IPV4)
+			memcpy(dst, &ipsec_sa->in.lookup_dst_ipv4,
+			       _ODP_IPV4ADDR_LEN);
+		else
+			memcpy(dst, &ipsec_sa->in.lookup_dst_ipv6,
+			       _ODP_IPV6ADDR_LEN);
```
JannePeltonen, line 2008: Shouldn't these be now ODP_IPV4_ADDR_SIZE and ODP_IPV6_ADDR_SIZE as those are the lengths of the arrays in the SA info?

include/odp/api/spec/ipsec.h
```diff
@@ -772,6 +773,75 @@ typedef struct odp_ipsec_sa_param_t {
...
+			uint32_t antireplay_ws;
+
+			/** Antireplay window top
+			 *
+			 * Sequence number representing a recent top of the
+			 * antireplay window(need not be the latest).
+			 * When no packets have been processed, this value will
+			 * be zero.
+			 */
+			uint64_t antireplay_window_top;
```
JannePeltonen, line 809: This should also tell what is returned if the anti-replay service is disabled. There is a space missing before the opening parenthesis. Maybe this could be worded like this:

Sequence number representing a recent top of the anti-replay window. There may be a delay before the SA state is reflected in the value. The value will be zero if no packets have been processed or if the anti-replay service is not enabled.

include/odp/api/spec/ipsec.h
JannePeltonen: Space missing here too. Maybe similar rewording here too:

Sequence number used for a recently processed packet. There may be a delay before the SA state is reflected in the value. When no packets have been processed, the value will be zero.

test/validation/api/ipsec/ipsec_test_out.c
```diff
@@ -1138,6 +1138,127 @@ static void test_out_ipv4_udp_esp_null_sha256(void)
...
+	memset(&tunnel_in, 0, sizeof(tunnel_in));
+
+	tunnel_out.type = ODP_IPSEC_TUNNEL_IPV4;
+	tunnel_out.ipv4.src_addr = &src;
+	tunnel_out.ipv4.dst_addr = &dst;
+
+	ipsec_sa_param_fill(&param_out,
+			    false, false, 0x4a2cbfe3, &tunnel_out,
+			    ODP_CIPHER_ALG_AES_GCM, &key_mcgrew_gcm_3,
+			    ODP_AUTH_ALG_AES_GCM, NULL,
```
JannePeltonen, line 1167: The algorithms could be AES-CBS-128 and HMAC-SHA1-96 as those are mandatory for every IPsec implementation (same change was proposed in PR 1127).

JannePeltonen:
I still asked for one more clarification in the API (for the case anti-replay is disabled). Other comments were minor and rather cosmetic and I would think the next version of the PR will probably be the final one.

You could also check if you want to fine tune the commit messages. A couple of small notes about them (but let's not spend too much time on this):

> The proposed API

It is no longer a proposed API when it gets merged, so "The new API" or something like that would fit better in the commit log (initially everything has "proposed" status anyway).

> to verify the parameter set by the application is properly updated in the implementation.

I am not sure if "verifying" that parameters are "properly updated" is accurate as for most parameters the implementation just remembers the parameters passed and returns them back when asked. 

> Add outbound test to verify odp_ipsec_sa_info() API.

The test has both outbound and inbound processing and SA info checks, so maybe just remove the work "outbound".

include/odp/api/spec/ipsec.h
asasidharan: Updated the API description as suggested.

platform/linux-generic/odp_ipsec.c
asasidharan: Agreed. Moved the functions to odp_ipsec_sad.c.

platform/linux-generic/odp_ipsec.c
asasidharan: Agreed. Updated v11 to use the new macros.

include/odp/api/spec/ipsec.h
asasidharan: Ack.

test/validation/api/ipsec/ipsec_test_out.c
asasidharan: Ack. I had to add the same check function introduced in PR 1127 - ipsec_check_esp_aes_cbc_128_sha1 here as well for this. Once the PR 1127(IPsec Print) is merged, I can drop the definition of ipsec_check_esp_aes_cbc_128_sha1() function from this PR.

asasidharan:
Summary of changes in v11:

- API: Updated the description for antireplay_window_top and seq_num fields as suggested.
- linux-gen: Moved the function definitions to odp_ipsec_sad.c.
- validation: Updated the test to use AES-CBC cipher and SHA1 auth.
- Also improved the commit messages as per the review comments.

JannePeltonen:
Looks ok for me, except for one minor thing: Now that the code moved to odp_ipsec_sad.c, the function _odp_ipsec_sa_antireplay_max_seq should be static and not declared in the header file. And since it will not be visible outside the file, it can have a simpler name (I would suggest dropping the _odp_ -prefix).

You can add my reviewed-by tag in the next version.

asasidharan:
Updated in v12:

- Renamed function to ipsec_sa_antireplay_max_seq() and made it static.
- Added reviewed-by tag for Janne.

JannePeltonen (approved):

JannePeltonen:
It seems there will be merge conflicts with PR 1060, so one more rebase will be needed once PR 1060 gets pulled.

asasidharan:
> 
> 
> It seems there will be merge conflicts with PR 1060, so one more rebase will be needed once PR 1060 gets pulled.

Thank you for your time in reviewing this PR @JannePeltonen and @psavol . I will rebase this change once PR 1060 gets merged.

matiaselo:
Hi @asasidharan, I merged #1060, so you can now do a rebase.

asasidharan:
v13: Rebased the PR to current master.

PR #1069, opened 2020-10-13
[PATCH v2] api: make odp_likely() work also with non-boolean values

JereLeppanen:
A change to the behavior of odp_likely(), and a couple of fixes to odp_likely() calls.


psavol:
Commit log entry should indicate the API "module" that was modified. So, please reword API patch log entry to start with "api: hints: ..." in v2. Otherwise OK and approved.

psavol (approved):

PR #1070, opened 2020-10-14
[PATCH v6] api: packet: specify errors returned in case of IPsec

asasidharan:
In case of IPsec packets, odp_packet_has_error() can be used to check parse
errors after IPsec processing. All IPsec errors/warnings need to be
checked using odp_ipsec_result().

### DIFF
--- a/test/validation/api/ipsec/ipsec.c
+++ b/test/validation/api/ipsec/ipsec.c
@@ -831,8 +831,9 @@ void ipsec_check_in_one(const ipsec_test_part *part, odp_ipsec_sa_t sa)
 			CU_ASSERT_EQUAL(0, odp_ipsec_result(&result, pkto[i]));
 			CU_ASSERT_EQUAL(part->out[i].status.error.all,
 					result.status.error.all);
-			CU_ASSERT(!result.status.error.all ==
-				  !odp_packet_has_error(pkto[i]));
+			if (0 == result.status.error.all)
+				CU_ASSERT_EQUAL(0,
+						odp_packet_has_error(pkto[i]));
 			CU_ASSERT_EQUAL(suite_context.inbound_op_mode ==
 					ODP_IPSEC_OP_MODE_INLINE,
 					result.flag.inline_mode);
@@ -881,8 +882,9 @@ void ipsec_check_out_one(const ipsec_test_part *part, odp_ipsec_sa_t sa)
 			CU_ASSERT_EQUAL(0, odp_ipsec_result(&result, pkto[i]));
 			CU_ASSERT_EQUAL(part->out[i].status.error.all,
 					result.status.error.all);
-			CU_ASSERT(!result.status.error.all ==
-				  !odp_packet_has_error(pkto[i]));
+			if (0 == result.status.error.all)
+				CU_ASSERT_EQUAL(0,
+						odp_packet_has_error(pkto[i]));
 			CU_ASSERT_EQUAL(sa, result.sa);
 			CU_ASSERT_EQUAL(IPSEC_SA_CTX,
 					odp_ipsec_sa_context(sa));
--- a/include/odp/api/spec/packet_flags.h
+++ b/include/odp/api/spec/packet_flags.h
@@ -27,13 +27,17 @@ extern "C" {
  */
 
 /**
- * Check for all errors in packet
+ * Check for all parse errors in packet
  *
  * Check if packet parsing has found any errors in the packet. The level of
  * error checking depends on the parse configuration (e.g. included layers and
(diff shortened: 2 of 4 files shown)


JereLeppanen:
Wait, how is this redundant? Isn't it useful to check that odp_packet_has_error() also returns what we expect?

Similar check is also done in ipsec_check_out_one(). If this one is changed, then that one should probably be changed as well.

asasidharan:
> 
> 
> Wait, how is this redundant? Isn't it useful to check that odp_packet_has_error() also returns what we expect?

API spec for odp_packet_has_error() says: `Check if packet parsing has found any errors in the packet. `
But it is not clear from the current spec whether IPsec errors like anti-replay is to be considered as a parse error.

Since such errors specific to IPsec are already being reported by odp_ipsec_result() this seems redundant for me.
Or is there any other utility where we really need to make it appear as a parse/packet error?
In other words is it really necessary for an implementation to set packet error for all the IPsec operation error cases?

> 
> Similar check is also done in ipsec_check_out_one(). If this one is changed, then that one should probably be changed as well.

Agreed. Will update this in the next revision.

JereLeppanen:
> But it is not clear from the current spec whether IPsec errors like anti-replay is to be considered as a parse error.

I have to agree with you there, I can't tell either how this should be interpreted.

Looks like currently linux-gen does report ipsec errors via odp_packet_has_error().

asasidharan:
> 
> Looks like currently linux-gen does report ipsec errors via odp_packet_has_error().

If we can confirm that  anti-replay is not treated as a parsing error, then I can submit a fix for linux-generic for this.

asasidharan:
Updated V2 with odp_packet_has_error() check removed in ipsec_checkout_one as well.

JannePeltonen:
When I read the commit message I immediately thought that the change is wrong. What is redundant in a normal application can be useful and can add test coverage in an API validation test. If the API spec says that an error can be reported in two ways, then an API validation check should check both ways even if any normal application would not.

If a validation test checks something that is not guaranteed by the API, then it is just a bug that should be fixed, not merely a redundant check.

JannePeltonen:
It seems that the API spec is not very clear about what odp_packet_has_error() reports after IPsec processing.

The way I read it is that it reports parsing errors and not things like "SA lifetime exceeded" or "SA not found", but that is just my interpretation and even then it is hard to draw the line.

I also think the API spec implies that an implementation can report only the errors it has found and is not required to report all errors. That would mean that an application cannot really rely on odp_packet_has_error() for much.

Interestingly, there is a commit that specifically adds IPsec errors in the generic packet error flags:
56432c440 linux-gen: ipsec: mark IPsec packets with errors with error flag

I think the specification of odp_packet_has_error() needs to be improved, but I am not sure how.

psavol:
Agree that odp_packet_has_error(), odp_packet_has_l3_error(), etc API may need better specification (or capability) what exact errors are checked from the packet.

anoobj:
> 
> Agree that odp_packet_has_error(), odp_packet_has_l3_error(), etc API may need better specification (or capability) what exact errors are checked from the packet.

May be the question is, since the API is unclear about the error reporting via odp_packet_has_error(), do we want the validation test to be checking for the same. IMO, IPsec result check is what is mandated by the spec and validation suite should check that and skip tests that are not mandated.

test/validation/api/ipsec/ipsec.c
```diff
@@ -831,8 +831,6 @@ void ipsec_check_in_one(const ipsec_test_part *part, odp_ipsec_sa_t sa)
 			CU_ASSERT_EQUAL(0, odp_ipsec_result(&result, pkto[i]));
 			CU_ASSERT_EQUAL(part->out[i].status.error.all,
 					result.status.error.all);
-			CU_ASSERT(!result.status.error.all ==
-				  !odp_packet_has_error(pkto[i]));
```
JannePeltonen, line 835: This fixes the problem that the validation test checks that odp_packet_has_error() reports an error when odp_ipsec_result() reports an error. But in case odp_ipsec_result() reports success, maybe the validation check should still check that also odp_packet_has_error() reports no-error. So instead of removing the check completely, maybe it should be made conditional on IPsec processing success?

JannePeltonen:
Looks ok to me except for a minor comment about the validation test.

test/validation/api/ipsec/ipsec.c
asasidharan: Agreed. Updated V4 such that odp_packet_has_error() is checked conditionally when IPsec processing is a success.
Additionally on linux-generic, fixed adjusting the bit fields while removing ipsec_err bit from _odp_packet_flags_t.

JannePeltonen (approved):

PR #1072, opened 2020-10-14
[PATCH v16] api: timer: add timer pool queue capability

PavanNikhilesh:
## api: timer: add timer pool queue capability

In some ODP implementations, plain queues might be using software
primitives such as a simple ring and the ODP timer hardware pools
can enqueue timeouts only into schedule queues.
Introducing capability `queue_type_sched` and `queue_type_plain`
which define the queue types that can be the destination of the
timeout events.

## linux-gen: timer: destination queue type capability

Add support for the new timer pool destination queue type
capability.

## test: validation: timer: check timer capability

Check timer pool capability and set tests as inactive if destination
queue type is not supported.

### DIFF
--- a/include/odp/api/spec/timer.h
+++ b/include/odp/api/spec/timer.h
@@ -216,6 +216,27 @@ typedef struct {
 	 */
 	odp_timer_res_capability_t max_tmo;
 
+	/**
+	 * Scheduled queue destination support
+	 *
+	 * This defines whether schedule queues are supported as timeout
+	 * destination queues.
+	 * 0: Scheduled queues are not supported as timeout destination queues
+	 * 1: Scheduled queues are supported as timeout destination queues
+	 * @see odp_timer_alloc()
+	 */
+	odp_bool_t queue_type_sched;
+
+	/**
+	 * Plain queue destination support
+	 *
+	 * This defines whether plain queues are supported as timeout
+	 * destination queues.
+	 * 0: Plain queues are not supported as timeout destination queues
+	 * 1: Plain queues are supported as timeout destination queues
+	 * @see odp_timer_alloc()
+	 */
+	odp_bool_t queue_type_plain;
 } odp_timer_capability_t;
 
 /**
--- a/platform/linux-generic/odp_timer.c
+++ b/platform/linux-generic/odp_timer.c
@@ -1254,6 +1254,8 @@ int odp_timer_capability(odp_timer_clk_src_t clk_src,
 	capa->max_tmo.res_hz  = timer_global->highest_res_hz;
 	capa->max_tmo.min_tmo = 0;
 	capa->max_tmo.max_tmo = MAX_TMO_NSEC;
+	capa->queue_type_sched = true;
+	capa->queue_type_plain = true;
 
 	return 0;
 }
--- a/test/validation/api/timer/timer.c
+++ b/test/validation/api/timer/timer.c
@@ -41,12 +41,19 @@ struct test_timer {
 	uint64_t tick; /* Expiration tick or TICK_INVALID */
 };
 
+struct thread_args {
+	pthrd_arg thrdarg;
+	odp_queue_type_t queue_type;
+};
+
 typedef struct {
 	/* Default resolution / timeout parameters */
 	struct {
 		uint64_t res_ns;
 		uint64_t min_tmo;
 		uint64_t max_tmo;
+		odp_bool_t queue_type_sched;
+		odp_bool_t queue_type_plain;
 	} param;
 
 	/* Timeout pool handle used by all threads */
@@ -152,6 +159,8 @@ static int timer_global_init(odp_instance_t *inst)
 	global_mem->param.res_ns  = res_ns;
 	global_mem->param.min_tmo = min_tmo;
 	global_mem->param.max_tmo = max_tmo;
+	global_mem->param.queue_type_plain = capa.queue_type_plain;
+	global_mem->param.queue_type_sched = capa.queue_type_sched;
 
 	return 0;
 }
@@ -177,6 +186,24 @@ static int timer_global_term(odp_instance_t inst)
 	}
 
 	return 0;
+}
+
+static int
+check_sched_queue_support(void)
+{
+	if (global_mem->param.queue_type_sched)
+		return ODP_TEST_ACTIVE;
+
+	return ODP_TEST_INACTIVE;
+}
+
+static int
+check_plain_queue_support(void)
+{
+	if (global_mem->param.queue_type_plain)
+		return ODP_TEST_ACTIVE;
+
+	return ODP_TEST_INACTIVE;
 }
 
 static void timer_test_capa(void)
@@ -323,12 +350,26 @@ static void timer_test_timeout_pool_free(void)
 static void timer_pool_create_destroy(void)
 {
 	odp_timer_pool_param_t tparam;
+	odp_queue_param_t queue_param;
+	odp_timer_capability_t capa;
 	odp_timer_pool_info_t info;
 	odp_timer_pool_t tp[2];
 	odp_timer_t tim;
 	odp_queue_t queue;
+	int ret;
 
-	queue = odp_queue_create("timer_queue", NULL);
+	memset(&capa, 0, sizeof(capa));
+	ret = odp_timer_capability(ODP_CLOCK_CPU, &capa);
+	CU_ASSERT_FATAL(ret == 0);
+
+	odp_queue_param_init(&queue_param);
+	if (capa.queue_type_plain) {
+		queue_param.type = ODP_QUEUE_TYPE_PLAIN;
+	} else if (capa.queue_type_sched) {
+		queue_param.type = ODP_QUEUE_TYPE_SCHED;
+		queue_param.sched.sync = ODP_SCHED_SYNC_ATOMIC;
+	}
+	queue = odp_queue_create("timer_queue", &queue_param);
 	CU_ASSERT_FATAL(queue != ODP_QUEUE_INVALID);
 
 	memset(&tparam, 0, sizeof(odp_timer_pool_param_t));
@@ -391,6 +432,7 @@ static void timer_pool_max_res(void)
 {
 	odp_timer_capability_t capa;
 	odp_timer_pool_param_t tp_param;
+	odp_queue_param_t queue_param;
 	odp_timer_pool_t tp;
 	odp_timer_t timer;
 	odp_pool_param_t pool_param;
@@ -411,7 +453,14 @@ static void timer_pool_max_res(void)
 	pool = odp_pool_create("timeout_pool", &pool_param);
 	CU_ASSERT_FATAL(pool != ODP_POOL_INVALID);
 
-	queue = odp_queue_create("timer_queue", NULL);
+	odp_queue_param_init(&queue_param);
+	if (capa.queue_type_plain) {
+		queue_param.type = ODP_QUEUE_TYPE_PLAIN;
+	} else if (capa.queue_type_sched) {
+		queue_param.type = ODP_QUEUE_TYPE_SCHED;
+		queue_param.sched.sync = ODP_SCHED_SYNC_ATOMIC;
+	}
+	queue = odp_queue_create("timer_queue", &queue_param);
 	CU_ASSERT_FATAL(queue != ODP_QUEUE_INVALID);
 
 	/* Highest resolution: first in nsec, then in hz */
@@ -824,6 +873,8 @@ static void timer_test_cancel(void)
 	odp_pool_t pool;
 	odp_pool_param_t params;
 	odp_timer_pool_param_t tparam;
+	odp_queue_param_t queue_param;
+	odp_timer_capability_t capa;
 	odp_timer_pool_t tp;
 	odp_queue_t queue;
 	odp_timer_t tim;
@@ -831,6 +882,11 @@ static void timer_test_cancel(void)
 	odp_timeout_t tmo;
 	odp_timer_set_t rc;
 	uint64_t tick;
+	int ret;
+
+	memset(&capa, 0, sizeof(capa));
+	ret = odp_timer_capability(ODP_CLOCK_CPU, &capa);
+	CU_ASSERT_FATAL(ret == 0);
 
 	odp_pool_param_init(&params);
 	params.type    = ODP_POOL_TIMEOUT;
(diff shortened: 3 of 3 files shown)


include/odp/api/spec/timer.h
```diff
@@ -216,6 +234,20 @@ typedef struct {
...
+	 * delevering events to.
+	 * When this values is:
+	 * - `ODP_TIMER_QUEUE_TYPE_SCHED` odp_timer_alloc() can only be called
+	 *				  with sched queue.
+	 * - `ODP_TIMER_QUEUE_TYPE_PLAIN` odp_timer_alloc() can only be called
+	 *				  with plain queue.
+	 * - `ODP_TIMER_QUEUE_TYPE_ANY`   odp_timer_alloc() can be called with
+	 *				  any type of queue.
+	 */
+	odp_timer_queue_type_t queue_type;
```
psavol, line 255: Instead of creating a new type for timer queue capability, could you use e.g. two booleans to indicate if sched/plain queues are supported with timer.

include/odp/api/spec/timer.h
```diff
@@ -42,7 +42,9 @@ typedef enum {
 	/** Use CPU clock as clock source for timers */
 	ODP_CLOCK_CPU,
 	/** Use external clock as clock source for timers */
-	ODP_CLOCK_EXT
+	ODP_CLOCK_EXT,
+	/* Use external clock with constant frequency as clock source for timers */
+	ODP_CLOCK_EXT_CONST_HZ
```
psavol, line 47: This feels unnecessary enumeration. The new external clock frequency parameter under can indicate if application is able to pass the clock frequency. This enumeration is for selecting between different clock sources (from multiple internal/external options).

include/odp/api/spec/timer.h
```diff
@@ -158,6 +160,9 @@ typedef struct {
 	/** Clock source for timers */
 	odp_timer_clk_src_t clk_src;
 
+	/** Frequency of the constant external clock source.
+	 *  Only valid when clk_src is 'ODP_CLOCK_EXT_CONST_HZ'. */
+	uint64_t const_ext_hz;
```
psavol, line 165: This could be simply clock frequency, which is by default zero (implementation knows the freq).

include/odp/api/spec/timer.h
PavanNikhilesh: Will remove this in the next iteration and use two booleans instead.

include/odp/api/spec/timer.h
PavanNikhilesh: The idea was that EXT clock source might not have a fixed frequency, In such cases application should use ODP_CLOCK_EXT.
When EXT clock source is know to be of constant frequency ODP_CLOCK_EXT_CONST_HZ can be used as a hint to the underlying timer HW accelerator.

include/odp/api/spec/timer.h
PavanNikhilesh: If external frequency varies, then I believe passing clock frequency would not be helpful to the underlying timer hardware.

psavol: My idea is that the enumeration is not changed above, but this additional parameter would tell if application knows the ext clock frequency to be fixed. CLOCK_EXT + non-zero clock_ext_hz value tells that the frequency is fixed. Value of zero means that application does not know the frequency, or it may vary.

Maybe clock_ext_hz would be better name for this than const_ext_hz.

Clock source enumeration may need to be extended later to cover multiple external (or internal) clock sources. We may then add more enums (e.g. ODP_CLOCK_EXT1, ODP_CLOCK_EXT2, ...) or another variable here (e.g. clock_ext_id).

PavanNikhilesh: Makes sense, I removed the macro change in v4. Thanks.

include/odp/api/spec/timer.h
```diff
@@ -140,6 +140,8 @@ typedef struct {
 	/** Clock source for timers */
 	odp_timer_clk_src_t clk_src;
 
+	/** Frequency of the constant clock source. */
+	uint64_t const_hz;
```
PavanNikhilesh, line 144: We believe that this parameter is not really helpful as it needs to be known while calling `odp_timer_capability` and `odp_timer_res_capability`  functions, It is the responsibility of the implementation to figure out the EXT_CLK frequency.
Will drop this and send a v7.

include/odp/api/spec/timer.h
```diff
@@ -216,6 +216,23 @@ typedef struct {
...
 	odp_timer_res_capability_t max_tmo;
 
+	/**
+	 * Enqueue queue type supported : Schedule queue.
+	 *
+	 * Defines the odp_queue_t type that the timer is capable of
+	 * delevering events to.
+	 * When set odp_timer_alloc() can be called with sched queue.
+	 */
+	odp_bool_t sched_queue_enq;
```
psavol, line 226: Timer API uses term "destination queue for timeout". Both of these capabilities should refer to "destination queue" term. For example:
```
Scheduled destination queue support
0: Scheduled queues are not supported as timeout destination queues
1: Scheduled queues are supported as timeout destination queues
@see odp_timer_alloc()
odp_bool_t queue_type_sched;
```

platform/linux-generic/odp_timer.c
```diff
@@ -1254,6 +1254,8 @@ int odp_timer_capability(odp_timer_clk_src_t clk_src,
 	capa->max_tmo.res_hz  = timer_global->highest_res_hz;
 	capa->max_tmo.min_tmo = 0;
 	capa->max_tmo.max_tmo = MAX_TMO_NSEC;
+	capa->sched_queue_enq = true;
+	capa->plain_queue_enq = true;
```
psavol, line 1258: API and implementation changes should be in separate patches. So, move this into a "linux-gen: timer: ..." patch.

include/odp/api/spec/timer.h
PavanNikhilesh: Sure, will send next version.
Thanks.

platform/linux-generic/odp_timer.c
PavanNikhilesh: Will split the pact in next verison.
Thanks

test/validation/api/timer/timer.c
```diff
@@ -488,6 +511,23 @@ static void timer_test_event_type(odp_queue_type_t queue_type,
 	int num = 5;
 	odp_timer_t timer[num];
 
+	memset(&capa, 0, sizeof(capa));
+	ret = odp_timer_capability(ODP_CLOCK_CPU, &capa);
+	CU_ASSERT_FATAL(ret == 0);
+
+	if (!(capa.queue_type_plain && capa.queue_type_sched)) {
```
psavol, line 518: Test cases should be modified to be conditional based on capability. If a test case returns without testing anything, it still looks like a passed validation test case.

You can take example from other suites how ODP_TEST_INFO_CONDITIONAL() macro is used. Basically, e.g. ODP_TEST_INFO(timer_test_pkt_event_sched) would change to ODP_TEST_INFO_CONDITIONAL(timer_test_pkt_event_sched, check_sched_queue_support). Suite init phase would record capabilities, and a check function would read those and return ACTIVE / INACTIVE status.

For example,
```
static int check_sched_queue_support(void)
{
	if (globals->capa.queue_type_sched)
		return ODP_TEST_ACTIVE;

	return ODP_TEST_INACTIVE;
}
```

PavanNikhilesh: Thanks for the pointers, I will modify it in the next version.

test/validation/api/timer/timer.c
```diff
@@ -1218,8 +1277,31 @@ static int worker_entrypoint(void *arg ODP_UNUSED)
...
+	odp_queue_param_init(&queue_param);
+	if (capa.queue_type_plain) {
+		queue_param.type = ODP_QUEUE_TYPE_PLAIN;
+		queue_type = ODP_QUEUE_TYPE_PLAIN;
+	} else if (capa.queue_type_sched) {
+		odp_thrmask_set(&thr_mask, odp_thread_id());
+		queue_param.type = ODP_QUEUE_TYPE_SCHED;
+		queue_param.sched.sync = ODP_SCHED_SYNC_PARALLEL;
+		queue_type = ODP_QUEUE_TYPE_SCHED;
+		queue_param.sched.group = odp_schedule_group_create(NULL,
```
psavol, line 1300: This requires "num_workers" (max 32 ?) schedule groups supported. A system may have limited number of schedule groups available. So, this test case (timer_test_all) should limit num_workers to max scheduler groups (in schedule mode test).

test/validation/api/timer/timer.c
```diff
@@ -1218,8 +1277,31 @@ static int worker_entrypoint(void *arg ODP_UNUSED)
...
-	queue = odp_queue_create("timer_queue", NULL);
+	memset(&capa, 0, sizeof(capa));
+	ret = odp_timer_capability(ODP_CLOCK_CPU, &capa);
+	CU_ASSERT_FATAL(ret == 0);
+
+	odp_queue_param_init(&queue_param);
+	if (capa.queue_type_plain) {
+		queue_param.type = ODP_QUEUE_TYPE_PLAIN;
+		queue_type = ODP_QUEUE_TYPE_PLAIN;
+	} else if (capa.queue_type_sched) {
```
psavol, line 1295: This if-else if tests only plain queues when both are supported. So, reference implementation runs only the plain queue path of the test. The scheduled queue path should be also tested with reference implementation (to be  sure it works and will remain working in the future).

Could you add queue type parameter to timer_test_all, and run it twice in the suite (once for plain and once for sched).

psavol:
Please remove the last patch (merge master to ...). Also, the two last patches are are bug corrections not related to the API change. Could you pick those into another PR, so that we can merge those ASAP (before the next API version).

test/validation/api/timer/timer.c
PavanNikhilesh: Sure, I will modify it in the next version.

test/validation/api/timer/timer.c
PavanNikhilesh: Makes sense, will split the test case.

PavanNikhilesh:
> Please remove the last patch (merge master to ...). Also, the two last patches are are bug corrections not related to the API change. Could you pick those into another PR, so that we can merge those ASAP (before the next API version).

My bad I thought it was a way to retrigger CI as it failed with false positives.
I will split the pull-requests.

test/validation/api/timer/timer.c
```diff
@@ -1411,23 +1510,30 @@ static void timer_test_all(void)
...
-	int num_workers = odp_cpumask_default_worker(&unused, 0);
+	num_workers = odp_cpumask_default_worker(&unused, 0);
 
 	/* force to max CPU count */
 	if (num_workers > MAX_WORKERS)
 		num_workers = MAX_WORKERS;
 
+	if (queue_type == ODP_QUEUE_TYPE_SCHED &&
+	    num_workers > sched_capa.max_groups)
+		num_workers = sched_capa.max_groups;
```
psavol, line 1535: Scheduler API has 3 automatic groups (ODP_SCHED_GROUP_ALL, etc). So, application may be able to create only max_groups - 3 groups.

psavol:
SP scheduler test segfaults (may be relater to num groups).

Inline timer test case fails due to non-empty queues. Test case clean up may need to run scheduler a while and drop all events.

PavanNikhilesh:
> SP scheduler test segfaults (may be relater to num groups).
> 
> Inline timer test case fails due to non-empty queues. Test case clean up may need to run scheduler a while and drop all events.

I tried with only one worker and test case still fails with sp scheduler.

```
Test: timer_test_plain_all ...num_workers 1
passed
  Test: timer_test_sched_all ...num_workers 1
odp_queue_basic.c:418:queue_destroy():queue "timer_queue" not empty
threads.c:211:wait_pthread():Bad exit status cpu #27 0x2
error: odph_thread_join() failed.
FAILED
    1. timer.c:1478  - CU_FAIL("Failed to receive stale timeout")
    2. timer.c:1497  - rc == 0
```

If I set the sched queue group 
`queue_param.sched.group = ODP_SCHED_GROUP_ALL;` 
the test passes
```
Test: timer_test_plain_all ...num_workers 1
passed
  Test: timer_test_sched_all ...num_workers 1
passed
```

Seems like an issue with schedule sp itself?

psavol:
I found the bug in SP scheduler. I'll send the fix as a PR and update validation tests to catch it.

test/validation/api/timer/timer.c
```diff
@@ -1218,8 +1282,26 @@ static int worker_entrypoint(void *arg ODP_UNUSED)
...
+	struct thread_args *thr_args = arg;
+	uint64_t sched_tmo;
 
-	queue = odp_queue_create("timer_queue", NULL);
+	odp_queue_param_init(&queue_param);
+	if (thr_args->queue_type == ODP_QUEUE_TYPE_PLAIN) {
+		queue_param.type = ODP_QUEUE_TYPE_PLAIN;
+		queue_type = ODP_QUEUE_TYPE_PLAIN;
+	} else {
+		odp_thrmask_set(&thr_mask, odp_thread_id());
```
psavol, line 1296: Thread mask needs to be init to zero (odp_thrmask_zero(&thr_mask)) before adding the thread id.

jerinjacobk:
Can this spec change merged as part of 1.25.2?

psavol:
> 
> 
> @psavol @matiaselo Can this spec change merged as part of 1.25.2?

@jerinjacobk Yes, but first needs to be rebased, thread mask zeroing fixed and pass all tests.

test/validation/api/timer/timer.c
```diff
@@ -1218,8 +1282,27 @@ static int worker_entrypoint(void *arg ODP_UNUSED)
...
+	if (thr_args->queue_type == ODP_QUEUE_TYPE_PLAIN) {
+		queue_param.type = ODP_QUEUE_TYPE_PLAIN;
+		queue_type = ODP_QUEUE_TYPE_PLAIN;
+	} else {
+		odp_thrmask_zero(&thr_mask);
+		odp_thrmask_set(&thr_mask, odp_thread_id());
+		queue_param.type = ODP_QUEUE_TYPE_SCHED;
+		queue_param.sched.sync = ODP_SCHED_SYNC_PARALLEL;
+		queue_type = ODP_QUEUE_TYPE_SCHED;
+		queue_param.sched.group = odp_schedule_group_create(NULL,
```
matiaselo, line 1301: Please check the return value here. There was a bug in the basic scheduler implementation, which caused `odp_schedule_group_create()` to fail when trying to create a large number of groups.

This PR needs to be rebased once Petri has created a fix commit.

matiaselo: The subject line of the validation test commit should be `validation: timer: check timer capability`

PavanNikhilesh: The caller function limits the number of threads to number of available schedule groups. But it makes sense to error check here just to be safe and prevent ub.
I will send the updated version, fixing the commit title too.

Thanks for reviewing

psavol: Here's PR for the fix
https://github.com/OpenDataPlane/odp/pull/1144

PavanNikhilesh: I will wait for the fix to be merged and then rebase my PR. Thanks!

psavol: Fix is now merged. Add my review-by tag to the next version. If it passes tests, it would be then ready to be merged.

PavanNikhilesh: Pushed new version with reviewed-by tag.
Thanks.

PavanNikhilesh:
@matiaselo any reason why shippable got cancelled?

matiaselo:
>  any reason why shippable got cancelled?
@PavanNikhilesh There was (/is) some issues with the server, which caused tests to run really slowly. I'll restart all the tasks once the server is working properly again.

PR #1074, opened 2020-10-16
[PATCH v1] example: atomic_strong_cas_dblptr() using casp for aarch64

ashwinyes:
For aarch64, if __ARM_FEATURE_ATOMICS is defined then use casp instruction
for implementing atomic_strong_cas_dblptr().


matiaselo:
Instead of adding more ARM specific code to this example we should clean it and remove all architecture specific code. ARM is currently working on adding 128bit atomics to the ODP API.

ashwinyes:
Okay. Abandoning PR.

PR #1081, opened 2020-10-23
[PATCH v7] api: system info call

psavol:
System info may be used to retrieve various system information
with a single call. Content of the system info structure may be
extended later, but it contains initially e.g. CPU ISA family
and version.

### DIFF
--- a/include/odp/api/spec/system_info.h
+++ b/include/odp/api/spec/system_info.h
@@ -242,6 +242,15 @@ int odp_sys_cache_line_size(void);
 void odp_sys_info_print(void);
 
 /**
+ * Print configuration
+ *
+ * Print out implementation defined information about selected configuration options. This
+ * information is intended for debugging purposes and may contain e.g. content of various
+ * configuration files, environment variables and configuration options of ODP API.
+ */
+void odp_sys_config_print(void);
+
+/**
  * @}
  */
 
--- a/include/odp/api/spec/system_info.h
+++ b/include/odp/api/spec/system_info.h
@@ -1,4 +1,5 @@
 /* Copyright (c) 2013-2018, Linaro Limited
+ * Copyright (c) 2020, Nokia
  * All rights reserved.
  *
  * SPDX-License-Identifier:     BSD-3-Clause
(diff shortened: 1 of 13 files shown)


psavol:
v2: rebased

psavol:
v3: added odp_sys_config_print() API. This is useful information e.g. when user needs to check which configuration options were selected for a build or a test run.

matiaselo:
Hi @m-govind, could you please take a look if this PR looks OK from ARM ISA point of view?

include/odp/api/spec/system_info.h
```diff
@@ -23,6 +24,165 @@ extern "C" {
...
+	 */
+	odp_cpu_arch_t cpu_arch;
+
+	/**
+	 * ISA version of ODP software
+	 *
+	 * Defines the ISA version that was used to build the ODP library. Note that CPU hardware
+	 * may support a higher ISA version than this.
+	 */
+	odp_cpu_arch_isa_t cpu_isa_sw;
```
matiaselo, line 169: It could be useful to have also `cpu_isa_hw ` variant. `Unknown` would be a valid value for this member since there may not always be a way to know the HW ISA.

platform/linux-generic/odp_system_info.c
```diff
@@ -527,6 +527,18 @@ int odp_cpu_count(void)
...
 }
 
+int odp_system_info(odp_system_info_t *info)
+{
+	system_info_t *sys_info = &odp_global_ro.system_info;
+
+	memset(info, 0, sizeof(odp_system_info_t));
+
+	info->cpu_arch   = sys_info->cpu_arch;
+	info->cpu_isa_sw = sys_info->cpu_isa_sw;
```
matiaselo, line 538: Should these values be initialized also for the default platform? E.g. `_odp_dummy_cpuinfo()`

m-govind:
> Hi @m-govind, could you please take a look if this PR looks OK from ARM ISA point of view?

Hi Matias/Petri, How the ARM ISA version information will be used in ODP ? Whether the ISA versions will be used to choose specific features from an ISA version ? If yes, it is not recommended to use __ARM_ARCH predefined macro. We have to use feature specific predefined macros defined in the chapter 5 in ACLE document (attached here).
[ACLE_Q3_2020.pdf](https://github.com/OpenDataPlane/odp/files/5502889/ACLE_Q3_2020.pdf)

Also, Few versions of GCC did not implement the ISA version information correctly and report a value of 8 for this macro even when compiling for v8.1 and above.

psavol:
> 
> 
> > Hi @m-govind, could you please take a look if this PR looks OK from ARM ISA point of view?
> 
> Hi Matias/Petri, How the ARM ISA version information will be used in ODP ? Whether the ISA versions will be used to choose specific features from an ISA version ? If yes, it is not recommended to use __ARM_ARCH predefined macro. We have to use feature specific predefined macros defined in the chapter 5 in ACLE document (attached here).
> [ACLE_Q3_2020.pdf](https://github.com/OpenDataPlane/odp/files/5502889/ACLE_Q3_2020.pdf)
> 
> Also, Few versions of GCC did not implement the ISA version information correctly and report a value of 8 for this macro even when compiling for v8.1 and above.

This is a new API, so it's for application use (ODP can use internally those feature macros). First, application (developer) can check the ISA version that was used during compilation (print it e.g. into log). It's an easy mistake not to define -march for an aarch64 build and GCC defaults to v8.0 even when building on a >v8.0 machine. Secondly, when building against v8.0 application may use a different strategy in using atomic APIs, than when building against >v8.0 as in memory atomics were introduced in v8.1.

This patch set handles __ARM_ARCH == 8 vs __ARM_ARCH == 8 +__ARM_FEATURE_QRDMX vs __ARM_ARCH >= 800. So, at least the important bit: version == v8.0 vs version >= v8.1 should be there.

platform/linux-generic/odp_system_info.c
psavol: Initialized sys_info->cpu_arch to ODP_CPU_ARCH_UNKNOWN. sys_info->cpu_isa_sw is memset to zero and should not be used when cpu_arch is unknown.

include/odp/api/spec/system_info.h
psavol: Added in v7.

m-govind:
> > > Hi @m-govind, could you please take a look if this PR looks OK from ARM ISA point of view?
> > 
> > 
> > Hi Matias/Petri, How the ARM ISA version information will be used in ODP ? Whether the ISA versions will be used to choose specific features from an ISA version ? If yes, it is not recommended to use __ARM_ARCH predefined macro. We have to use feature specific predefined macros defined in the chapter 5 in ACLE document (attached here).
> > [ACLE_Q3_2020.pdf](https://github.com/OpenDataPlane/odp/files/5502889/ACLE_Q3_2020.pdf)
> > Also, Few versions of GCC did not implement the ISA version information correctly and report a value of 8 for this macro even when compiling for v8.1 and above.
> 
> This is a new API, so it's for application use (ODP can use internally those feature macros). 
First, application (developer) can check the ISA version that was used during compilation (print it e.g. into log). It's an easy mistake not to define -march for an aarch64 build and GCC defaults to v8.0 even when building on a >v8.0 machine. 

Agree. ODP code can be compiled for portability (Run across different ISA versions) or for performance (Specific ISA version and micro architecture). Developer can use the ISA version info from log message to check whether proper GCC options are used for best performance. The message will be helpful only to use the right ISA version through march. There are other options like mtune for microarchitecture related optimizations which cannot be detected by reading the ISA version. Following blog explain about the usage of GCC options march, mtune and mcpu. 
https://community.arm.com/developer/tools-software/tools/b/tools-software-ides-blog/posts/compiler-flags-across-architectures-march-mtune-and-mcpu
As mentioned before, few gcc versions did not implement the ISA versions correctly for __ARM_ARCH. As you mentioned below, with __ARM_FEATURE_QRDMX, only v8.0 and v8.1 only can be differentiated. Other minor versions may not be detected using __ARM_ARCH.

Secondly, when building against v8.0 application may use a different strategy in using atomic APIs, than when building against >v8.0 as in memory atomics were introduced in v8.1.

Regarding the feature selection, __ARM_ARCH is not recommended because, some of the CPU can choose to implement the features from next ISA version also. For example, CPU with base ISA v8.0, can pull in some of the features (like atomics) from ISA v8.1. Please refer the link below for more information.
https://developer.arm.com/architectures/learn-the-architecture/understanding-the-armv8-x-extensions/single-page

> 
> This patch set handles __ARM_ARCH == 8 vs __ARM_ARCH == 8 +__ARM_FEATURE_QRDMX vs __ARM_ARCH >= 800. So, at least the important bit: version == v8.0 vs version >= v8.1 should be there.

psavol:
ISA version enum should follow ARM ISA level specification, so that e.g. v8.1 tells that all mandatory features of v8.0 and v8.1 are implemented, in addition some features from v8.2 could be there (but application can not be sure without checking feature flags). This API enables application/developer to check basics about the ISA without ifdefs and knowledge about feature flags (such as QRDMX) and definition of flags in various compiler versions.

I'd suggest that specific feature flags (for arm and x86) could be added later (as an union of bit fields in system info struct).

Are you OK with the current API spec and implementation?

m-govind:
> ISA version enum should follow ARM ISA level specification, so that e.g. v8.1 tells that all mandatory features of v8.0 and v8.1 are implemented, in addition some features from v8.2 could be there (but application can not be sure without checking feature flags). This API enables application/developer to check basics about the ISA without ifdefs and knowledge about feature flags (such as QRDMX) and definition of flags in various compiler versions.
> 
> I'd suggest that specific feature flags (for arm and x86) could be added later (as an union of bit fields in system info struct).
> 
> Are you OK with the current API spec and implementation?

Looks good Petri.

matiaselo (approved):

PR #1082, opened 2020-10-28
[PATCH v3] api: unaligned int types

psavol:
This is one option to handle GCC9 warnings of taking a pointer to packed structure. We could introduce unaligned integer types as part of ODP API.

PR https://github.com/OpenDataPlane/odp/pull/1046 is solving the same warnings with re-write of the checksum code.

### DIFF
--- a/helper/chksum.c
+++ b/helper/chksum.c
@@ -105,7 +105,7 @@ static inline int odph_process_l4_hdr(odp_packet_t      odp_pkt,
 	odph_tcphdr_t  *tcp_hdr_ptr;
 	odp_bool_t      split_l4_hdr, is_tcp;
 	uint32_t        l4_offset, l4_len, pkt_chksum_offset;
-	uint16_t       *pkt_chksum_ptr;
+	odp_una_u16_t  *pkt_chksum_ptr;
 	uint8_t        *l4_ptr;
 	uint32_t hdr_len = 0;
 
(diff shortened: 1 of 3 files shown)


include/odp/api/spec/std_types.h
```diff
@@ -40,6 +41,15 @@ extern "C" {
  */
 typedef uint32_t odp_percent_t;
 
+/** Unaligned uint16_t type */
+typedef uint16_t ODP_ALIGNED(1) odp_una_u16_t;
```
JannePeltonen, line 45: Looking at GCC documentation it seems to me that in this kind of typedef the type attribute should be at the end, i.e.: typedef uint16_t odp_una_u16_t ODP_ALIGNED(1); With structs and unions it is different.

helper/chksum.c
```diff
@@ -130,7 +130,7 @@ static inline int odph_process_l4_hdr(odp_packet_t      odp_pkt,
 		 * should come from the udp header, unlike for TCP where is
 		 * derived. */
 		l4_len            = odp_be_to_cpu_16(udp_hdr_ptr->length);
-		pkt_chksum_ptr    = (uint16_t *)(void *)&udp_hdr_ptr->chksum;
+		pkt_chksum_ptr    = (odp_una_u16_t *)(void *)&udp_hdr_ptr->chksum;
```
JannePeltonen, line 133: Is the intermediate cast through (void *) needed for some reason? Cannot we just drop it now?

JannePeltonen:
I made a couple of minor comments, but overall this looks good to me. This patch set is about the minimal code change to get rid of the compiler warning in a way that also fixes the underlying issue (information about the (lack of) alignment getting lost).

include/odp/api/spec/std_types.h
psavol: fixed in v2

helper/chksum.c
psavol: fixed in v2

JannePeltonen (approved):

PR #1083, opened 2020-10-28
[PATCH v2]  api: pktio: support for HW Tx timestamp offload 

matiaselo:
This PR adds reference implementation to the original API PR #1055.

Some of the original commits have been modified (mainly for readability, no functional API changes). The original commits which have been modified can be identified by the new signed-off-by tags.


platform/linux-generic/include/odp_packet_io_internal.h
```diff
@@ -68,9 +68,13 @@ struct pktio_entry {
 	/* These two locks together lock the whole pktio device */
 	odp_ticketlock_t rxl;		/**< RX ticketlock */
 	odp_ticketlock_t txl;		/**< TX ticketlock */
-	uint8_t cls_enabled;            /**< classifier enabled */
-	uint8_t chksum_insert_ena;      /**< pktout checksum offload enabled */
 	uint16_t pktin_frame_offset;
+	struct {
```
JannePeltonen, line 72: I wonder how to comment a commit comment in github, but here "Reserve memory" should probably be "save memory" or similar.

platform/linux-generic/include/odp_packet_io_internal.h
```diff
@@ -231,6 +235,18 @@ static inline void pktio_cls_enabled_set(pktio_entry_t *entry, int ena)
 	entry->s.enabled.cls = !!ena;
 }
 
+static inline int _odp_pktio_tx_ts_enabled(pktio_entry_t *entry)
+{
+	return entry->s.enabled.tx_ts;
+}
+
+static inline void _odp_pktio_tx_ts_set(pktio_entry_t *entry)
```
JannePeltonen, line 243: This might be more useful in the future if it allowed the devide code to specify the time, e.g. by getting time from odp_time_global() only when time is not provided by the caller.

platform/linux-generic/pktio/loop.c
```diff
@@ -315,6 +317,11 @@ static int loopback_send(pktio_entry_t *pktio_entry, int index ODP_UNUSED,
 		 */
 		out_octets_tbl[i] = bytes;
 		nb_tx++;
+
+		if (tx_ts_enabled && tx_ts_idx == 0) {
+			if (odp_unlikely(packet_hdr(pkt_tbl[i])->p.flags.ts_set))
```
JannePeltonen, line 322: Wouldn't it be better to first check the unlikely packet header flag and only then tx_ts_enabled and tx_ts_idx? If time stamping is not enabled, the check for tx_ts_enabled would change to a check of the ts_set flag, which is should not be significantly more costly (wouldn't the flags be hot in the cache already?), but when time stamping is enabled, we would still need only single comparison instead of three for the vast majority of packets.

JannePeltonen: And the same comment for the other occurences of the similar code in other commits in performance relevant pktios.

platform/linux-generic/pktio/loop.c
```diff
@@ -339,6 +346,9 @@ static int loopback_send(pktio_entry_t *pktio_entry, int index ODP_UNUSED,
 	ret = odp_queue_enq_multi(queue, (odp_event_t *)hdr_tbl, nb_tx);
 
 	if (ret > 0) {
+		if (odp_unlikely(tx_ts_idx && ret >= tx_ts_idx))
```
JannePeltonen, line 349: I think it would be better to have odp_unlikely() apply already to tx_ts_idx alone, i.e. if (odp_unlikely(tx_ts_idx) && ret >= tx_ts_idx).

platform/linux-generic/pktio/dpdk.c
```diff
@@ -956,17 +966,24 @@ static inline int pkt_to_mbuf_zero(pktio_entry_t *pktio_entry,
...
 			/* Fall back to packet copy */
 			if (odp_unlikely(pkt_to_mbuf(pktio_entry, &mbuf,
-						     &pkt, 1) != 1))
+						     &pkt, 1, &dummy_idx) != 1))
 				goto fail;
 			(*copy_count)++;
 		}
 		mbuf_table[i] = mbuf;
+
+		if (odp_unlikely(tx_ts_enabled)) {
```
JannePeltonen, line 983: I am not sure this should be marked unlikely.

platform/linux-generic/pktio/dpdk.c
```diff
@@ -956,17 +966,24 @@ static inline int pkt_to_mbuf_zero(pktio_entry_t *pktio_entry,
...
+			if (odp_unlikely(*tx_ts_idx == 0 && pkt_hdr->p.flags.ts_set))
```
JannePeltonen, line 984: I wonder if it would be better to have a local variable for tx_ts_idx to help the compiler in optimization. But if this gets changed as suggested in the comment for loop.c, it wouldn't matter anymore.

platform/linux-generic/include/odp_packet_io_internal.h
```diff
@@ -231,6 +235,18 @@ static inline void pktio_cls_enabled_set(pktio_entry_t *entry, int ena)
 	entry->s.enabled.cls = !!ena;
 }
 
+static inline int _odp_pktio_tx_ts_enabled(pktio_entry_t *entry)
```
JannePeltonen, line 238: Is it ok to have a leading underscore in the name?

JannePeltonen:
I made a couple of minor comments, but otherwise this looks good.

platform/linux-generic/include/odp_packet_io_internal.h
matiaselo: Ok

platform/linux-generic/include/odp_packet_io_internal.h
matiaselo: The project convention is to use `_odp` -prefix for implemention internal functions.

platform/linux-generic/pktio/loop.c
matiaselo: Based on testing reading the `pkt_tbl[i])->p.flags.ts_set` first is more costly compared to the current implementation.

platform/linux-generic/include/odp_packet_io_internal.h
matiaselo: I'd prefer adding the extra parameter if/when it's actually needed. This shouldn't be a complex change if the function's external functionality stays the same.

platform/linux-generic/pktio/dpdk.c
matiaselo: Removing the `odp_unlikely()` causes a measurable performance penalty in the common processing path.

platform/linux-generic/pktio/loop.c
matiaselo: OK, will fix.

psavol (approved):

PR #1084, opened 2020-10-29
[PATCH v4] api: Pull request for pool enhancement APIs for VPP

jerinjacobk:
v4:
- Rebase to master
- Fix build issue with clang by avoid forward declaration with typedef ie. introding
callback arguments as "void *" and allows any opaque type.

v3:
- Rebase to master
- Update sorting entry table configuration definition. Reason for the change as default pool should
be able to handle all packet lengths in the sort table.

v2:
- Split debug API to different PR ( https://github.com/OpenDataPlane/odp/pull/1100 )

## api: pool: introduce external memory allocator

Some of the ODP application manages the packet io memory
for better control of memory management.
Introduce an external memory allocator in the ODP specification
to use that memory for the pool memory.

Introducing an external memory callback for allocator and
deallocator to allocate and deallocate the pool memory.

This callback will be called once in the pool create and destroy
to allow ODP application to allocate and destroy the pool memory.

## api: pool: introduce pool type iterators

Some of the ODP application needs to store the
application-specific data in the pool creation time.

Introducing pool type iterators to allow getting
a callback to update/store in the application-specific
metadata in the pool creation phase on different types.

## api: pktio: introduce buffer sort capability

Some of the HW has support for allocating the packet buffer
based on the size.

This is often useful for saving the memory where the application
can create a different pool to steer the specific size of
the packet, thus enabling effective use of memory.

A similar use case can be achieved through the sub pool scheme,
however, if the application needs finer control on the number of
pools and to give the very same pool to different packet
in sort entries then this capability will be handy.

Packet input buffer sorting entries define the sorting requirements.
When sorting is enabled and packet received in given a packet io,
the implementation shall scan the configured entries, if the packet size
is less than or equal odp_pktin_sort_desc_t::len, it allocates
the packets from odp_pktin_sort_desc_t::pool.

odp_pktin_sort_desc_t::len value in each entry must be unique and must
be sorted in ascending order. Otherwise, the behavior is undefined.

Example sorting entry table configuration:

- (0B <= packet_size <= 1500B), steer to pool a
- (1500B < packet_size <= 4000B), steer to pool b
- (4000B < packet_size <= 7000B), steer to pool c
- (7000B < packet_size <= default_pool::param::max_len), steer to
default pool configured for the pktio.

tbl[0].len = 1500;
tbl[0].pool = pool_a;
tbl[1].len = 4000;
tbl[1].pool = pool_b;
tbl[2].len = 7000;
tbl[2].pool = pool_c;

## api: packet: introduce packet handle get from headroom

The use case where odp_packet_t converted as the application-specific
object on the Rx path and application-specific object to odp_packet_t on
Tx path would need a get function for getting odp_packet_t
from the predefined location from the packet.

odp_packet_from_head() enables such use case in Tx path
where, if application objects stores in place in headroom location
and the application wants to convert the application-specific
object to odp_packet_t before transmitting the packet.

## api: packet: introduce packet handle get from user area

The use case where odp_packet_t converted as the application-specific
object on the Rx path and application-specific object to odp_packet_t
on Tx path would need a get function for getting
odp_packet_t from the predefined location from the packet.

odp_packet_from_user_area() enables such use case in Tx path
where, if application objects stores in place in user area location
and the application wants to convert the application-specific
object to odp_packet_t before transmitting the packet.

### DIFF
--- a/include/odp/api/spec/pool.h
+++ b/include/odp/api/spec/pool.h
@@ -274,6 +274,42 @@ typedef uintptr_t (*odp_pool_ext_mem_allocate_t)(void *arg, uint32_t total_sz, u
 typedef void (*odp_pool_ext_mem_deallocate_t)(void *arg, uintptr_t ptr);
 
 /**
+ * ODP pool buffer type iterator callback.
+ *
+ * This callback shall be invoked by the implementation when each odp_buffer_t type is freed
+ * to the pool in odp_pool_create().
+ *
+ * @param arg    Opaque pointer registered by the user with odp_pool_param_t::buf::arg
+ * @param data_addr Pointer to the buffer start address.
+ *
+ * On successful pool creation, the application can access the same buffer
+ * address using odp_buffer_addr().
+ *
+ */
+typedef void (*odp_pool_buffer_iterator_t)(void *arg, void *data_addr);
+
+/**
+ * ODP pool packet type iterator callback.
+ *
+ * This callback shall be invoked by the implementation when each odp_packet_seg_t
+ * type is freed to the pool on odp_pool_create().
+ *
+ * @param headroom_addr Pointer to the start address packet headroom.
+ * On successful pool creation, the application can access the same headroom address
+ * using odp_packet_head().
+ * Value NULL will be passed to the callback when packet headroom of size zero.
+ *
+ * @param arg    Opaque pointer registered by the user with odp_pool_param_t::pkt::arg
+ * @param uarea_addr Pointer to the start address packet user area.
+ *
+ * On successful pool creation, the application can access the same user area address
+ * using odp_packet_user_area().
+ * Value NULL will be passed to the callback when the packet user area of size zero.
+ *
+ */
+typedef void (*odp_pool_packet_iterator_t)(void *arg, void *headroom_addr, void *uarea_addr);
+
+/**
  * Pool parameters
  */
 typedef struct odp_pool_param_t {
@@ -311,6 +347,12 @@ typedef struct odp_pool_param_t {
 		 *  implementation specific and set by odp_pool_param_init().
 		 */
 		uint32_t cache_size;
+
+		/** Opaque pointer to be passed in buf_iter invocation */
+		void *arg;
+
+		/** Buffer type iterator callback */
+		odp_pool_buffer_iterator_t buf_iter;
 	} buf;
 
 	/** Parameters for packet pools */
@@ -411,6 +453,12 @@ typedef struct odp_pool_param_t {
 		 *  See buf.cache_size documentation for details.
 		 */
 		uint32_t cache_size;
+
+		/** Opaque pointer to be passed in pkt_iter invocation */
+		void *arg;
+
+		/** Packet segment type iterator callback */
+		odp_pool_packet_iterator_t pkt_iter;
 	} pkt;
 
 	/** Parameters for timeout pools */
--- a/include/odp/api/spec/pool.h
+++ b/include/odp/api/spec/pool.h
@@ -20,6 +20,7 @@ extern "C" {
 #endif
 
 #include <odp/api/std_types.h>
+#include <odp/api/support.h>
 
 /** @defgroup odp_pool ODP POOL
  *  Packet and buffer (event) pools.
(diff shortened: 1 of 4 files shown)


include/odp/api/spec/pool.h
```diff
@@ -251,6 +251,37 @@ typedef uintptr_t (*odp_pool_ext_mem_allocate_t)(uint32_t total_sz, unsigned int
...
+ * ODP pool buffer type iterator callback.
+ *
+ * This callback shall be invoked by the implementation when each odp_buffer_t type is freed
+ * to the pool in odp_pool_create().
+ *
+ * @param data_addr Pointer to the buffer start address. On successful pool creation,
+ * the application can access the same buffer address using odp_buffer_addr().
+ *
+ */
+typedef void (*odp_pool_buffer_iterator_t)(void *data_addr);
```
psavol, line 264: Some for of user context would be needed as well. Maybe a pointer to copy of pool parameters.

include/odp/api/spec/pool.h
```diff
@@ -604,6 +604,32 @@ unsigned int odp_pool_max_index(void);
...
+ *
+ * Get the count of available entries in the pool. When the local cache is enabled,
+ * the count will account for the local cache entries per CPU as well.
+ *
+ * @param pool      Pool handle
+ *
+ * @return Return the number of available entries in the pool
+ * @retval <0     On failure (e.g., handle not valid)
+ */
+int odp_pool_available_count(odp_pool_t pool);
```
psavol, line 618: These counts could be combined to common odp_pool_stat(pool, &stat) API. Pool statistics could contain potentially also other counters. Max utilization could be provided also. 

Application could enable / disable statistics as needed.

psavol:
Pool index and stats patches could be a separate PR. That way those less difficult APIs could be advanced and merged earlier.

include/odp/api/spec/pool.h
jerinjacobk: I will send a separate PR for the same.

include/odp/api/spec/pool.h
jerinjacobk: Addressed in v3.

psavol:
Seems that pool capa/param needs to added to guarantee that headroom content is preserved while implementation owns the packet: during packet copy, when sending packet through queues, etc.

psavol:
patch "api: pktio: introduce buffer sort capability" could be moved to another PR

include/odp/api/spec/pool.h
```diff
@@ -273,6 +273,42 @@ typedef uintptr_t (*odp_pool_ext_mem_allocate_t)(void *arg, uint32_t total_sz, u
...
+ * to the pool in odp_pool_create().
+ *
+ * @param arg    Opaque pointer registered by the user with odp_pool_param_t::buf::arg
+ * @param data_addr Pointer to the buffer start address.
+ *
+ * On successful pool creation, the application can access the same buffer
+ * address using odp_buffer_addr().
+ *
+ */
+typedef void (*odp_pool_buffer_iterator_t)(void *arg, void *data_addr);
```
JannePeltonen, line 289: I would rather call this odp_pool_buffer_iterator_cb_t instead, as it is an iteration callback and not an iterator (an object holding iteration state).

JannePeltonen:
Just recording here the offline dicussions that it would probably be better to store VPP related application specific metadata in packet user area instead of in the packet headeroom. As we have discussed, user area is intended for application specific data whereas headroom is used as space for packet data to grow into. Headroom does not get used only explicitly through odp_packet_push_head() and odp_packet_extend_head() and other packet operations but it may also get used indirectly e.g. in ODP crypto or IPsec operations, depending on the ODP implementation.

Could you perhaps draft a version of this PR that would enable the use of user areas instead of headroom? I think that for the use of user area you need a couple of additional guarantees (and the associated pool capabilities and config options) by ODP: that the user area is at a constant offset of packet head and that the content of user area is preserved over odp_packet_free-odp_packet_alloc sequence.

include/odp/api/spec/pool.h
jerinjacobk: Ack

JannePeltonen:
This is superseded by PR 1260 (external memory pool). Marked this as duplicate.

Can we close this PR now?

jerinjacobk:
> This is superseded by PR 1260 (external memory pool). Marked this as duplicate.
> 
> Can we close this PR now?

Yes. Closing now. Created a separate PR as https://github.com/OpenDataPlane/odp/pull/1290 which is not addressed in the duplicate one

PR #1086, opened 2020-10-30
[PATCH v4] Pull request for maxlen_packet_tests

harman-marvell:
## test: bench_packet: change packet length pool param

Changing the value of packet length pool parameter as twice of
TEST_MAX_PKT_SIZE (viz 2048), so that packet of length up to
2 * 2048 can be allocated from the pool.
This change is required to honour certain test cases like
"bench_packet_ref_pkt", where for 2048 byte packet test iteration,
header packet of same length is prefixed to the base packet,
leading to the failure as with current pool configuration packets
allocated from the pool can be of maximum 2048 bytes.

## validation: packet: add maxlen pool param

Defining maximum length pool parameter while create pool
to ensure packet of length max_len will never fail.
As in default pool creation, segmented_packet_len gets the
value from pool capability to allocate maximum packet length.
So same should be used as max_len while creating the pool.

### DIFF
--- a/test/performance/odp_bench_packet.c
+++ b/test/performance/odp_bench_packet.c
@@ -1823,7 +1823,8 @@ int main(int argc, char *argv[])
 		ODPH_ERR("Error: packet pool size not supported.\n");
 		printf("MAX: %" PRIu32 "\n", capa.pkt.max_num);
 		exit(EXIT_FAILURE);
-	} else if (capa.pkt.max_len && capa.pkt.max_len < TEST_MAX_PKT_SIZE) {
+	} else if (capa.pkt.max_len &&
+		   capa.pkt.max_len < 2 * TEST_MAX_PKT_SIZE) {
 		ODPH_ERR("Error: packet length not supported.\n");
 		exit(EXIT_FAILURE);
 	} else if (capa.pkt.max_seg_len &&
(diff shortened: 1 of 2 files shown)


test/performance/odp_bench_packet.c
```diff
@@ -1840,6 +1840,7 @@ int main(int argc, char *argv[])
 	odp_pool_param_init(&params);
 	params.pkt.seg_len = PKT_POOL_SEG_LEN;
 	params.pkt.len     = TEST_MAX_PKT_SIZE;
+	params.pkt.max_len = 2 * TEST_MAX_PKT_SIZE;
```
matiaselo, line 1843: A problem with this solution is that the API promises only that the pool is able to allocate at least one `max_len` packet. In case of "bench_packet_ref_pkt" test, TEST_REPEAT_COUNT (=1000) packets are required.

A cleaner solution could be to use a larger pool, where `params.pkt.len = 2 * TEST_MAX_PKT_SIZE`, for these problematic tests. Another option would be to make the original packets smaller (`TEST_MAX_PKT_SIZE  / 2`).

harman-marvell: Thanks for the review comment, I will send V2 with "params.pkt.len = 2 * TEST_MAX_PKT_SIZE" change, so as to create a bigger pool.

test/performance/odp_bench_packet.c
```diff
@@ -1839,7 +1839,7 @@ int main(int argc, char *argv[])
 	/* Create packet pool */
 	odp_pool_param_init(&params);
 	params.pkt.seg_len = PKT_POOL_SEG_LEN;
-	params.pkt.len     = TEST_MAX_PKT_SIZE;
+	params.pkt.len     = 2 * TEST_MAX_PKT_SIZE;
```
matiaselo, line 1842: Check on line: 1826:`	} else if (capa.pkt.max_len && capa.pkt.max_len < TEST_MAX_PKT_SIZE) {` should be updated to take into account this change. You could also add a comment why `params.pkt.len = 2 * TEST_MAX_PKT_SIZE` is required.

Otherwise looking good, you can add my reviewed-by tag on the same go.

harman-marvell: Thanks, I will send the V4 as per the comments.

matiaselo (approved):

PR #1088, opened 2020-11-03
[PATCH v3] test: cunit: add cunit run mode for immediate test suite exit on failure

ashwinyes:
In existing test framework, all the tests in a test suite are always
run irrespective of individual test failures.

This patchset adds an option to run the test suites in a mode where
a test suite gets killed on the first test failure. This mode can be
activated by setting the environment variable ODP_CUNIT_FAIL_IMMEDIATE.

In this mode, before the start of any test in a test suite, it checks
for errors in previously ran tests and immediately triggers an exit if
an error is detected.

### DIFF
--- a/test/common/odp_cunit_common.c
+++ b/test/common/odp_cunit_common.c
@@ -1,5 +1,6 @@
 /* Copyright (c) 2014-2018, Linaro Limited
  * Copyright (c) 2019, Nokia
+ * Copyright (c) 2021, Marvell
  * All rights reserved.
  *
  * SPDX-License-Identifier:     BSD-3-Clause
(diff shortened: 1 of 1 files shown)


test/common/odp_cunit_common.c
```diff
@@ -292,7 +326,9 @@ static int cunit_register_suites(odp_suiteinfo_t testsuites[])
 		if (sinfo->term_func)
 			term_func = sinfo->term_func;
 
-		suite = CU_add_suite(sinfo->name, _cunit_suite_init, term_func);
+		suite = CU_add_suite_with_setup_and_teardown(sinfo->name, _cunit_suite_init,
```
matiaselo, line 329: `CU_add_suite_with_setup_and_teardown()` is not available on CentOS 7, which causes the CI to fail.

ashwinyes: Okay. The CUnit version on Centos 7 could be older. Let me see if there is a way to workaround the version compatibiltiy.

matiaselo: Hi @ashwinyes, I just updated the CentOS 7 CI image and the build should now work. Please rebase this PR on top of the latest master.

test/common/odp_cunit_common.c
```diff
@@ -273,6 +273,40 @@ static int default_term_func(void)
 	return odp_cunit_print_inactive();
 }
 
+static void _cunit_test_setup_func(void)
```
matiaselo, line 276: Marvell copyright could be added to this file.

There is also a typo in the commit message: `inidividual `.

test/common/odp_cunit_common.c
```diff
@@ -273,6 +273,40 @@ static int default_term_func(void)
...
+{
+	CU_AllTestsCompleteMessageHandler all_test_comp_handler;
+	CU_SuiteCompleteMessageHandler suite_comp_handler;
+	CU_pFailureRecord pfailrec;
+	CU_pSuite psuite;
```
matiaselo, line 281: A bit nitpicky, but `psuite` could be renamed to `suite` to be more consistent with the rest of the file.

test/common/odp_cunit_common.c
```diff
@@ -273,6 +273,40 @@ static int default_term_func(void)
...
+	suite_comp_handler = CU_get_suite_complete_handler();
+	if (suite_comp_handler)
+		suite_comp_handler(psuite, pfailrec);
+
+	all_test_comp_handler = CU_get_all_test_complete_handler();
+	if (all_test_comp_handler)
+		all_test_comp_handler(pfailrec);
+
+	/* Set some CUnit error so that it aborts */
+	CU_set_error(CUE_NOTEST);
```
matiaselo, line 307: This is a CUnit internal function, which doesn't do anything else than prints an error message (a bit confusing one in this case) and exits. So, it could be better to just call `exit(EXIT_FAILURE)` instead.

matiaselo (approved):

matiaselo:
Looks good, please add my reviewed-by tag: `Reviewed-by: Matias Elo <matias.elo@nokia.com>`

PR #1089, opened 2020-11-03
[PATCH v4] linux-gen: modify structs for consistency with gcc docs and odp code

Malvika-Gupta:
As per the GCC documentation, the 'aligned' attribute should be added
after the 'struct' keyword. Consequently, the ODP_ALIGNED attribute has
been moved in various structs of the following files:
- include/odp/api/abi-default/atomic.h
- platform/linux-generic/include-abi/odp/api/abi/atomic.h
- platform/linux-generic/odp_queue_lf.c

From the first two files, the 'Enforce alignment' comment has also been
removed since the effect of the ODP_ALIGNED macro is easy to understand.
Also, for these files, the typedef declaration has been combined
with the struct definition which is consistent with most typedef structs
in ODP source code.

Suggested-by: Petri Savolainen <petri.savolainen@nokia.com>

### DIFF
--- a/include/odp/api/abi-default/atomic.h
+++ b/include/odp/api/abi-default/atomic.h
@@ -24,9 +24,9 @@ extern "C" {
  * @internal
  * Atomic 32-bit unsigned integer
  */
-struct odp_atomic_u32_s {
+typedef struct ODP_ALIGNED(sizeof(uint32_t)) odp_atomic_u32_s {
 	uint32_t v; /**< Actual storage for the atomic variable */
-} ODP_ALIGNED(sizeof(uint32_t)); /* Enforce alignment! */
+} odp_atomic_u32_t;
 
 #if __GCC_ATOMIC_LLONG_LOCK_FREE >= 2
 
@@ -34,9 +34,9 @@ struct odp_atomic_u32_s {
  * @internal
  * Atomic 64-bit unsigned integer
  */
-struct odp_atomic_u64_s {
+typedef struct ODP_ALIGNED(sizeof(uint64_t)) odp_atomic_u64_s {
 	uint64_t v; /**< Actual storage for the atomic variable */
-} ODP_ALIGNED(sizeof(uint64_t)); /* Enforce alignment! */
+} odp_atomic_u64_t;
 
 #else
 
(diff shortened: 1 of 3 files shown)


JereLeppanen:
This looks good, but the commit subject could be better. Since there is a change to an api header, it should probably be prefixed with "api:", and the subject itself could be more to the point. I would make it something like

`api, linux-gen: fix placement of ODP_ALIGNED in struct definitions`

Malvika-Gupta:
Thank you for reviewing, @JereLeppanen. Fixing placement of the ODP_ALIGNED is the change which is common for all the files. For the atomic files, there is a second change as well- combining the typedef declaration with the struct definition. Should I still change the subject line to 'fix placement of ODP_ALIGNED in struct definitions"?
 
Also, I wanted to ask you what is the correct format for the subject line, in terms of pointing the location of the change? When should we use linux-gen vs api/abi? I was unsure about this when I submitted the PR and I was hoping that one of the maintainers can clarify this. Thanks!

JereLeppanen:
> For the atomic files, there is a second change as well- combining the typedef declaration with the struct definition. Should I still change the subject line to 'fix placement of ODP_ALIGNED in struct definitions"?

I would say yes. The ODP_ALIGNED change is the main change here, and it may actually have some effect on behavior. The typedef change is only a syntactic cleanup.

> Also, I wanted to ask you what is the correct format for the subject line, in terms of pointing the location of the change? When should we use linux-gen vs api/abi?

At least any api changes should be clearly indicated on the subject line.

But please wait for comments from @psavol or @matiaselo on both of your questions.

Malvika-Gupta:
Thanks Jere. I will wait for Petri or Matias to give their comments

Malvika-Gupta:
@matiaselo @psavol can you please provide your comments for this PR? Thank you

matiaselo:
Hi @Malvika-Gupta,

The code changes look good. I would still split the commit in two and use a subject lines like Jere suggested. The first commit would modify the abi files and the second the implementation internal `odp_queue_lf.c` file.

E.g. 
abi: atomic: fix placement of ODP_ALIGNED in struct definitions
linux-gen: queue_lf: fix placement of ODP_ALIGNED in struct definitions

Malvika-Gupta:
Sounds good, Matias. I will send the new patches shortly.

matiaselo (approved):

PR #1090, opened 2020-11-05
[PATCH v5] api: add odp_event_is_valid()

psavol:
This is needed for debugging e.g. when it is suspected that scheduler returns bad event handles.

### DIFF
--- a/include/odp/api/spec/packet.h
+++ b/include/odp/api/spec/packet.h
@@ -2162,10 +2162,12 @@ uint32_t odp_packet_vector_size(odp_packet_vector_t pktv);
 void odp_packet_vector_size_set(odp_packet_vector_t pktv, uint32_t size);
 
 /**
- * Perform full packet vector validity check
+ * Check that packet vector is valid
  *
- * The operation may consume considerable number of cpu cycles depending on
- * the check level.
+ * This function can be used for debugging purposes to check if a packet vector handle represents
+ * a valid packet vector. The level of error checks depends on the implementation. Considerable
+ * number of cpu cycles may be consumed depending on the level. The call should not crash if
+ * the packet vector handle is corrupted.
  *
  * @param pktv Packet vector handle
  *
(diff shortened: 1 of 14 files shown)


include/odp/api/spec/packet.h
```diff
@@ -2055,10 +2055,12 @@ void odp_packet_print(odp_packet_t pkt);
...
 
 /**
- * Perform full packet validity check
+ * Check that packet is valid
  *
- * The operation may consume considerable number of cpu cycles depending on
- * the check level.
+ * This function can be used for debugging purposes to check if a packet handle represents
+ * a valid packet. The level of error checks depends on the implementation. Considerable number of
+ * cpu cycles may be consumed depending on the level. The call should not crash if the event handle
```
JereLeppanen, line 2062: Packet handle, not event handle.

JereLeppanen (commented):
odp_event_is_valid() checks something completely unspecified and depends on implementation. Is it obvious enough, or does it not matter because this is meant for debugging? It's not clear to me how an application could use this.

include/odp/api/spec/packet.h
psavol: Fixed in v2.

psavol:
> 
> 
> odp_event_is_valid() checks something completely unspecified and depends on implementation. Is it obvious enough, or does it not matter because this is meant for debugging? It's not clear to me how an application could use this.

Level of error checks is implementation dependent as it would be hard to specify which kind of checks can be done, how robust those are, how those impact performance, level may be selected through config file, etc. So, check can be as simple as handle != XXX_INVALID, but min_addr < handle < max_addr is already much useful to application developers as it would (likely) catch random, corrupted handle values. That would then give application better chance to print an error into log (and return) instead of e.g. crashing on the next API call.

platform/linux-generic/odp_packet.c
```diff
@@ -1693,7 +1693,10 @@ void odp_packet_print_data(odp_packet_t pkt, uint32_t offset,
 
 int odp_packet_is_valid(odp_packet_t pkt)
 {
-	if (odp_buffer_is_valid(packet_to_buffer(pkt)) == 0)
+	if (pkt == ODP_PACKET_INVALID)
+		return 0;
+
```
matiaselo, line 1698: Packet subtype validity could be checked here.

test/validation/api/event/event.c
```diff
@@ -33,6 +33,7 @@ static void event_test_free(void)
 
 	for (i = 0; i < EVENT_BURST; i++) {
 		buf = odp_buffer_alloc(pool);
+		CU_ASSERT(odp_event_is_valid(odp_buffer_to_event(buf)) == 1);
```
matiaselo, line 36: Similar checks could be added for `ODP_EVENT_CRYPTO_COMPL` and `ODP_EVENT_IPSEC_STATUS`events.

platform/linux-generic/odp_packet.c
psavol: done in v3

test/validation/api/event/event.c
psavol: done in v3

matiaselo (approved):

PR #1092, opened 2020-11-07
[PATCH v5] api: tm: tm spec changes 

nithind1988:
commit 1513b06d6d5dbee86b3f9b26d13d5b633daf5980
Date:   Wed Nov 4 22:49:11 2020 +0530

    validation: tm: add five level test case suite
    
    Current validation suite only validates 3 level TM systems.
    Refactor the test cases to use suite inited parameters instead
    of compile time and add suite init for 5 level TM system.
    
commit 6fc41eee7fea267b522ac2821b38f0fe7205255b
Date:   Wed Nov 4 22:41:56 2020 +0530

    test: cunit: add API to deactivate all test cases
    
    In test suite init, it is possible to detect that a given test suite
    is not applicable. Hence provide an API to deactivate all the
    test cases of a given test suite.
    
commit 7a7a767bc297873fc901c93e87c0541ceffddb41
Date:   Wed Nov 4 19:36:31 2020 +0530

    validation: tm: use tm start and stop API
    
    Add tm start API after creating a TM system and TM stop API before
    destroying it.
    
commit 7a7dc2878acfc5fe8fbc1f13a98c610fb43609b5
Date:   Wed Nov 4 19:23:55 2020 +0530

    example: tm: use TM start and stop
    
    Use TM start to start the TM system after creation and
    TM stop to stop it before destroying TM queues.
    
commit ee2422bcbe8b80c90d7d3e2c704584fef92719b4
Date:   Thu Nov 5 06:22:55 2020 +0530

    linux-gen: tm: fix issues related to name table cleanup
    
    Name table entry was not cleaned up properly in desroy path
    of tm profile/node/system. This patch fixes the same.
    
commit 8999e29d58434eb1d73ad67765a3a901f3fc7c26
Date:   Thu Nov 5 06:19:57 2020 +0530

    linux-gen: tm: express supported features in capa
    
    Update newly added capabilities fields based on available support.
    Also implement egress specific capability API which is in our case
    same as TM support is independent of egress kind.
    
commit 24a7f9318f2a5897f3d1adec43af697af358d088
Date:   Wed Nov 4 19:22:14 2020 +0530

    linux-gen: traffic_mngr: add implementation of TM start/stop
    
    Add implementation of TM start or stop API. Since linux generic platform
    does not have any setup and cleanup constraints, the TM start/stop
    are a no-op.
    
commit abe234464ef3a00215c1cb0850ea7fc63c23ef24
Date:   Wed Nov 4 20:11:25 2020 +0530

    api: tm: add capability to express support for dynamic update
    
    Not all platforms can support all dynamic TM system update like
    topology update, shaper update, sched update after TM is in start
    state.
    
    Dynamic TM system topology update API for TM node correspond to
    node create/connect/disconnect/destroy. Similarly for TM queue,
    queue create/connect/disconnect/destroy.
    
    Dynamic TM system sched/shaper update API for TM node's correspond
    to odp_tm_node_shaper_config() and odp_tm_node_sched_config()
    Similarly for TM queue, odp_tm_queue_shaper_config() and
    odp_tm_queue_sched_config(). Updating shaper params or sched
    params is also considered dynamic when nodes are already associated
    to that shaper or sched profile.
    
    This patch adds capability expressing the same. These updates
    can be done on any TM system when it is in stopped state.
    
commit 6e89babe70dee0f6cb43fb2ba285a82c1f0703f2
Date:   Wed Nov 4 20:06:49 2020 +0530

    api: tm: add missing capabilities and egress specific capability API
    
    Not all platforms support TM node query, node threshold, queue query,
    and queue threshold features. Since they are missing from existing
    global and level capabilities, adding the same to express no support.
    
    Also, in some platforms capabilities are specific to a egress kind
    like a PKTIO. Since capabilities provide a way to application to know
    things like how many levels can be supported etc, prior to TM creation,
    add a capability get API to retrieve egress function specific capabilities
    records.
    
commit b4a4e2b745a2e39f4dc67ea09c54c9725b778026
Date:   Fri Nov 6 20:48:35 2020 +0530

    api: tm: add option to enable packet mode in shaper
    
    Similar to using packet/frame count in scheduling
    algorithm and weight in units of number of packets,
    add option to enable packet mode in shaper
    by introducing a boolean flag in shaper profile params.
    When packet mode is set to TRUE, commit/peak rate is
    in PPS (packets per second) and commit/peak burst is in
    Packets.
    
commit ddcbd3e5c5e0c7137d6a4c63a8effd300dc5436d
Date:   Sat Nov 7 16:32:37 2020 +0530

    api: tm: add pktout tm queue mode and related API
    
    For Tx, Packet IO, with TM disabled has Direct mode for enqueueing packets
    and Queue mode for enqueueing packet or packet vector events with ordering
    maintained. For Packet IO with TM enabled, there is single mode
    which is equivalent to Direct mode for enqueuing packets. This patch
    adds queue mode with TM enabled as a mode to be able to enqueue packet or
    packet vector events to TM queues and maintain ordering.
    
    This patch also adds TM enqueue multi API equivalent to odp_tm_enq()
    but allowing multiple packets to be enqueued in single API call
    inline with pktout enqueue or send API's in non-TM mode.
    
commit 16db32e3a618413299e7b4f50f93c48bfb863a00
Date:   Wed Nov 4 19:21:28 2020 +0530

    api: tm: add interface to start and stop a TM system
    
    Not all TM systems are started after TM creation or topology
    setup. Platform needs to know when a TM system needs to
    start or stop in order to safely setup or cleanup TM system
    and drain pkts in it. This patch adds support for TM start/stop
    API. Platforms that don't have any constraint on setup or cleanup,
    can have start/stop as a no-op. But applications should call
    TM start after topology create.

commit 6fcbaedfd36ff29f3b8944c3ad46abeaa852237f
Date:   Wed Nov 4 21:57:28 2020 +0530

    api: tm: support node priority override
    
    Current TM spec associates a priority to a packet
    at TM queue and expects all the levels to honour that priority.
    In some platforms, this is not supportable and a packet's priority
    gets overridden by the node priority(with respect to parent),
    it is passing through. In other words, like assigning packet a
    priority based on the TM queue it is enqueued on, every node it
    is passing through overrides the packet priority with a fixed value.
    
    So add a capability expressing the limitation/feature of node
    priority override and also add node priority field in node
    params that corresponds to its priority among his direct
    siblings.
    
    This patch also adds "max_wfq_groups_per_node" capability that is
    expressed by a platform when node priority override is true. This
    field expresses maximum number of WFQ groups can be formed by
    a node under a given parent. WFQ group is a group of children
    having same priority and hence WFQ/WRR is performed among them
    in order to pick a packet for transmit.
    
### DIFF
--- a/include/odp/api/spec/packet_io.h
+++ b/include/odp/api/spec/packet_io.h
@@ -106,6 +106,10 @@ typedef enum odp_pktout_mode_t {
 	ODP_PKTOUT_MODE_QUEUE,
 	/** Packet output through traffic manager API */
 	ODP_PKTOUT_MODE_TM,
+	/** Packet output through traffic manager API where
+	 * TM queues act as event queues.
+	 */
+	ODP_PKTOUT_MODE_TM_QUEUE,
 	/** Application will never send to this interface */
 	ODP_PKTOUT_MODE_DISABLED
 } odp_pktout_mode_t;
@@ -595,7 +599,8 @@ typedef struct odp_pktio_config_t {
 	 *  through the IPSEC API.
 	 *
 	 *  Support of outbound IPSEC inline operation with traffic manager
-	 *  (ODP_PKTOUT_MODE_TM) can be queried with odp_ipsec_capability().
+	 *  (ODP_PKTOUT_MODE_TM or ODP_PKTOUT_MODE_TM_QUEUE) can be queried
+	 *  with odp_ipsec_capability().
 	 *
 	 * * 0: Disable outbound IPSEC inline operation (default)
 	 * * 1: Enable outbound IPSEC inline operation
@@ -865,8 +870,9 @@ typedef struct odp_pktio_capability_t {
  *
  * Packet output queue configuration must be setup with
  * odp_pktout_queue_config() before odp_pktio_start() is called. When packet
- * output mode is ODP_PKTOUT_MODE_DISABLED or ODP_PKTOUT_MODE_TM,
- * odp_pktout_queue_config() call is optional and will ignore all parameters.
+ * output mode is ODP_PKTOUT_MODE_DISABLED, ODP_PKTOUT_MODE_TM or
+ * ODP_PKTOUT_MODE_TM_QUEUE, odp_pktout_queue_config() call is optional
+ * and will ignore all parameters.
  *
  * Packet receive and transmit on the interface is enabled with a call to
  * odp_pktio_start(). If not specified otherwise, any interface level
--- a/include/odp/api/spec/traffic_mngr.h
+++ b/include/odp/api/spec/traffic_mngr.h
@@ -1655,6 +1655,20 @@ int odp_tm_queue_disconnect(odp_tm_queue_t tm_queue);
  */
 int odp_tm_enq(odp_tm_queue_t tm_queue, odp_packet_t pkt);
 
+/** The odp_tm_enq_multi() function is used to add packets to a given TM system.
+ * This API takes multiple packets to enqueue unlike odp_tm_enq() but
+ * rest of the behaviour is identical to that of odp_tm_enq().
+ *
+ * @param tm_queue  Specifies the tm_queue (and indirectly the TM system).
+ * @param packets   Array of packets to enqueue.
+ * @param num       Number of packets to send.
+ * @return          Returns >0 upon success indicating number of packets
+ *                  enqueued, <= 0 upon failure. One of the more common
+ *                  failure reasons is WRED drop.
+ */
+int odp_tm_enq_multi(odp_tm_queue_t tm_queue, const odp_packet_t packets[],
+		     int num);
+
 /** The odp_tm_enq_with_cnt() function behaves identically to odp_tm_enq(),
  * except that it also returns (an approximation to?) the current tm_queue
  * packet queue count.
@@ -1665,6 +1679,19 @@ int odp_tm_enq(odp_tm_queue_t tm_queue, odp_packet_t pkt);
  *                  this tm_queue upon success, < 0 upon failure.
  */
 int odp_tm_enq_with_cnt(odp_tm_queue_t tm_queue, odp_packet_t pkt);
+
+/** The odp_tm_queue_enq() function is used to add packets to a given TM system.
+ * This API needs to be used to enqueue events of type ODP_EVENT_PACKET_VECTOR
+ * or ODP_EVENT_PACKET to tm queue only when PKTOUT mode is
+ * ODP_PKTOUT_MODE_TM_QUEUE. Apart from this, rest of the behaviour is
+ * identical to that of odp_tm_enq().
+ *
+ * @param tm_queue  Specifies the tm_queue (and indirectly the TM system).
+ * @param ev        Event handle representing a packet or packet vector.
+ * @return          Returns 0 upon success, < 0 upon failure. One of the
+ *                  more common failure reasons is WRED drop.
+ */
+int odp_tm_queue_enq(odp_tm_queue_t tm_queue, odp_event_t ev);
 
 /* Dynamic state query functions */
 
--- a/include/odp/api/spec/traffic_mngr.h
+++ b/include/odp/api/spec/traffic_mngr.h
@@ -331,6 +331,26 @@ typedef struct {
 	/** The per_level array specifies the TM system capabilities that
 	 * can vary based upon the tm_node level. */
 	odp_tm_level_capabilities_t per_level[ODP_TM_MAX_LEVELS];
+
+	/** node_priority indicates that this TM system supports
+	 * only having single priority associated to fan-in nodes at any
+	 * level. When TRUE, Priority is fixed for a fanin node i.e all packets
+	 * fed by a particular fanin node to a tm node will have a fixed priority.
+	 * This is equivalent to the packet priority being overloaded by
+	 * fan-in node priority. When FALSE, every fan-in node can feed packets
+	 * to parent tm node with priorities ranging from
+	 * 0..ODP_TM_MAX_PRIORITIES - 1.
+	 */
+	odp_bool_t node_priority_override;
+
+	/** max_wfq_groups_per_node indicates maximum number of wfq groups supported
+	 * by a tm node at any level. This capability is valid only when
+	 * node_priority_override is TRUE. In a given topology only these many
+	 * groups of children can be formed of same node priority under a parent.
+	 * This value can range from 1..ODP_TM_MAX_PRIORITIES as there can be
+	 * only as many number of WFQ groups as MAX priority.
+	 */
+	uint8_t max_wfq_groups_per_node;
 } odp_tm_capabilities_t;
 
 /** Per Level Requirements
@@ -1233,6 +1253,12 @@ typedef struct {
 	 * greater levels may be connected to the fan-in of tm_node's with
 	 * numerically smaller levels. */
 	uint8_t level;
+
+	/** The strict priority level assigned to packets going through this
+	 * node. In other words all packets going through node
+	 * will be assigned this priority overriding previous priority.
+	 * in the range 0..max_priority. */
+	uint8_t priority;
 } odp_tm_node_params_t;
 
 /** odp_tm_node_params_init() must be called to initialize any
--- a/include/odp/api/spec/traffic_mngr.h
+++ b/include/odp/api/spec/traffic_mngr.h
@@ -794,20 +794,32 @@ typedef enum {
  */
 typedef struct {
 	/** The committed information rate for this shaper profile.  The units
-	 * for this integer are always in bits per second. */
-	uint64_t commit_bps;
+	 * for this integer is in bits per second when packet_mode is
+	 * not TRUE while packets per second when packet mode is TRUE.
+	 */
+	union {
+		uint64_t commit_bps; /**< Commit information rate in bps */
+		uint64_t commit_rate; /**< Commit information rate */
+	};
 
 	/** The peak information rate for this shaper profile.  The units for
-	 * this integer are always in bits per second. */
-	uint64_t peak_bps;
+	 * this integer is in bits per second when packet_mode is
+	 * not TRUE while in packets per second when packet mode is TRUE.
+	 */
+	union {
+		uint64_t peak_bps; /**< Peak information rate in bps */
+		uint64_t peak_rate; /**< Peak information rate */
+	};
 
 	/** The commit burst tolerance for this shaper profile.  The units for
-	 * this field are always bits.  This value sets an upper limit for the
+	 * this field is bits when packet_mode is not TRUE and packets when
+	 * packet_mode is TRUE.  This value sets an upper limit for the
 	 * size of the commitCnt. */
 	uint32_t commit_burst;
 
 	/** The peak burst tolerance for this shaper profile.  The units for
-	 * this field are always bits.  This value sets an upper limit for the
+	 * this field in bits when packet_mode is not TRUE and packets
+	 * when packet_mode is TRUE. This value sets an upper limit for the
 	 * size of the peakCnt. */
 	uint32_t peak_burst;
 
@@ -819,7 +831,9 @@ typedef struct {
 	 * to a value approximating the "time" (in units of bytes) taken by
 	 * the Ethernet preamble and Inter Frame Gap.  Traditionally this
 	 * would be the value 20 (8 + 12), but in same cases can be as low as
-	 * 9 (4 + 5). */
+	 * 9 (4 + 5).
+	 * This field is ignored when packet_mode is TRUE.
+	 */
 	int8_t shaper_len_adjust;
 
 	/** If dual_rate is TRUE it indicates the desire for the
@@ -828,6 +842,12 @@ typedef struct {
 	 * implementation specific, but in any case require a non-zero set of
 	 * both commit and peak parameters. */
 	odp_bool_t dual_rate;
+
+	/** If packet_mode is TRUE it indicates that shaper should work
+	 * in packet mode ignoring lengths of packet and hence shaping
+	 * traffic in packet's per second as opposed to bytes per second.
+	 */
+	odp_bool_t packet_mode;
 } odp_tm_shaper_params_t;
 
 /** odp_tm_shaper_params_init() must be called to initialize any
--- a/include/odp/api/spec/traffic_mngr.h
+++ b/include/odp/api/spec/traffic_mngr.h
@@ -189,6 +189,30 @@ extern "C" {
  * tree/hierarchy of nodes.
  */
 
+/** The tm_node_stats_type enumeration lists possible packet or octet
+ * statistics at a tm node.
+ */
+typedef enum odp_tm_node_stats_type_t {
+	/** Packets dropped by this node after scheduling/shaping at this node */
+	ODP_TM_NODE_STATS_PKTS_DROPPED,
+	/** Octets dropped after scheduling/shaping at this node */
+	ODP_TM_NODE_STATS_OCTETS_DROPPED,
+	/** Green packets that are sent through this tm node */
+	ODP_TM_NODE_STATS_GREEN_PKTS,
+	/** Green octets that are sent through this tm node */
+	ODP_TM_NODE_STATS_GREEN_OCTETS,
+	/** Yellow packets that are sent through this tm node */
+	ODP_TM_NODE_STATS_YELLOW_PKTS,
+	/** Yellow octets that are sent through this tm node */
+	ODP_TM_NODE_STATS_YELLOW_OCTETS,
+	/** Red packets that are sent through this tm node */
+	ODP_TM_NODE_STATS_RED_PKTS,
+	/** Red octets that are sent through this tm node */
+	ODP_TM_NODE_STATS_RED_OCTETS,
+	/** Node stats max */
+	ODP_TM_NODE_STATS_MAX,
+} odp_tm_node_stats_type_t;
+
 /** Per Level Capabilities
  *
  * The odp_tm_level_capabilities_t record is used to describe the capabilities
@@ -248,6 +272,15 @@ typedef struct {
 	/** tm_node_threshold_supported indicates that the tm_nodes at this
 	 * level support threshold profile set. */
 	odp_bool_t tm_node_threshold_supported;
+
+	/** tm_node_stats_supported indicates the statistics supported by
+	 * nodes at this level. When True, that stat counter is supported,
+	 * while False indicates it is not supported.
+	 *
+	 * @see odp_tm_node_stats_type_t
+	 * @see odp_tm_node_stats()
+	 */
+	odp_bool_t tm_node_stats_supported[ODP_TM_NODE_STATS_MAX];
 } odp_tm_level_capabilities_t;
 
 /** TM Capabilities Record.
@@ -1930,6 +1963,29 @@ typedef struct {
  */
 int odp_tm_node_fanin_info(odp_tm_node_t             tm_node,
 			   odp_tm_node_fanin_info_t *info);
+
+/** The odp_tm_node_stats_t record type  is used to return various stats
+ * supported by a given tm_node via the odp_tm_node_stats() function.
+ */
+typedef struct odp_tm_node_stats_t {
+	/** Node statistics. @see odp_tm_node_stats_type_t */
+	uint64_t stats[ODP_TM_NODE_STATS_MAX];
+} odp_tm_node_stats_t;
+
+/** Get tm_node stats
+ *
+ * The odp_tm_node_stats() function is used to extract runtime statistics
+ * associated with a given tm_node. The stats structure is written
+ * only on success.
+ *
+ * @param      tm_node    Specifies the tm_node to be queried.
+ * @param      clear      1: Clear stats before return.
+ *                        0: Don't clear stats.
+ * @param[out] node_stats A pointer to an odp_tm_node_stats_t record that is to
+ *                        be filled in by this call.
+ * @return                Returns < 0 upon failure, 0 upon success.
+ */
+int odp_tm_node_stats(odp_tm_node_t tm_node, int clear, odp_tm_node_stats_t *node_stats);
 
 /** The odp_tm_queue_info_t record type  is used to return various bits of
  * information about a given tm_queue via the odp_tm_queue_info() function.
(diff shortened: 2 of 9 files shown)


include/odp/api/spec/traffic_mngr.h
```diff
@@ -794,20 +794,26 @@ typedef enum {
...
+	 * not TRUE while packets per second when packet mode is TRUE.
+	 */
 	uint64_t commit_bps;
 
 	/** The peak information rate for this shaper profile.  The units for
-	 * this integer are always in bits per second. */
+	 * this integer is in bits per second when packet_mode is
+	 * not TRUE while in packets per second when packet mode is TRUE.
+	 */
 	uint64_t peak_bps;
```
JereLeppanen, line 911: Here we end up with *_bps parameters that are actually packets per second, not bits per second, if packet_mode is true, which is pretty horrible. At the very least the _bps postfix should be removed from these parameter names, but I'm not sure if we want to do that kind of incompatible change. @psavol should comment.

nithind1988: Ack, missed that. I can make it _rate and create a union of older deprecated name and new name if it is fine.

nithind1988:
Ping. Any thoughts on this pull request ?

include/odp/api/spec/traffic_mngr.h
```diff
@@ -794,20 +794,32 @@ typedef enum {
  */
 typedef struct {
 	/** The committed information rate for this shaper profile.  The units
-	 * for this integer are always in bits per second. */
-	uint64_t commit_bps;
+	 * for this integer is in bits per second when packet_mode is
+	 * not TRUE while packets per second when packet mode is TRUE.
+	 */
+	union {
+		uint64_t commit_bps; /**< Commit information rate in bps */
```
JannePeltonen: This one and also the peak_bps field should by marked deprecated. Or if backward compatibility for this is not needed, then the old field could just be dropped.

include/odp/api/spec/traffic_mngr.h
```diff
@@ -625,6 +625,34 @@ odp_tm_t odp_tm_find(const char            *name,
  */
 int odp_tm_capability(odp_tm_t odp_tm, odp_tm_capabilities_t *capabilities);
 
+/** Start a TM system.
+ *
+ * odp_tm_start() needs to be used to start an already created / found TM system
+ * and in stopped state. By default platforms can choose to be already in
+ * started state with immediately after created. Calling this function
+ * on an already started TM system will be ignored.
```
JannePeltonen, line 771: Maybe it would be better to require all implementations to behave similarly, even if that causes slight backward compatibility. So how about requiring odp_tm_start() to be called and having the TM always in stopped state before that? In any case, the expected behaviour in stopped state probably should be described better.

That said, I wonder if this TM specific start & stop is needed or whether the pktio start & stop would be sufficient for this?

include/odp/api/spec/traffic_mngr.h
nithind1988: I have already marked it as deprecated in last patch. I can remove it in last patch if you think it is not necessary to be backward compatible. Since it is being used in validation tests, I had to do it after changing the validation tests.

include/odp/api/spec/traffic_mngr.h
JannePeltonen: Yes. But I was thinking whether a TM specific start/stop is needed or if pktio start/stop could be used for that. I.e. if a system does not support dynamic TM update, then the pktio needs to be put to the stopped state for the update. But I suppose there are a couple of things where a separate TM start/stop is useful: pktio stop will stop pktin too, whereas TM stop can only stop the TX. And not all TM egresses are necessarily pktouts.

include/odp/api/spec/traffic_mngr.h
JannePeltonen: Sorry, I missed the deprecation in the last patch.

include/odp/api/spec/traffic_mngr.h
```diff
@@ -331,6 +331,26 @@ typedef struct {
...
+	/** node_priority indicates that this TM system supports
+	 * only having single priority associated to fan-in nodes at any
+	 * level. When TRUE, Priority is fixed for a fanin node i.e all packets
+	 * fed by a particular fanin node to a tm node will have a fixed priority.
+	 * This is equivalent to the packet priority being overloaded by
+	 * fan-in node priority. When FALSE, every fan-in node can feed packets
+	 * to parent tm node with priorities ranging from
+	 * 0..ODP_TM_MAX_PRIORITIES - 1.
+	 */
+	odp_bool_t node_priority_override;
```
JannePeltonen, line 381: This capability now makes the "node priority override" and the old behaviour mutually exclusive, but we can have implementations, like linux-gen, which can easily support both. I think it would be better to have separate capabilities for both types of priority handling (so an implementation could set them both) and have a global or per-node config setting for selecting which one to use.

include/odp/api/spec/traffic_mngr.h
```diff
@@ -331,6 +331,26 @@ typedef struct {
...
+
+	/** max_wfq_groups_per_node indicates maximum number of wfq groups supported
+	 * by a tm node at any level. This capability is valid only when
+	 * node_priority_override is TRUE. In a given topology only these many
+	 * groups of children can be formed of same node priority under a parent.
+	 * This value can range from 1..ODP_TM_MAX_PRIORITIES as there can be
+	 * only as many number of WFQ groups as MAX priority.
+	 */
+	uint8_t max_wfq_groups_per_node;
```
JannePeltonen, line 390: This is needed only if max_wfq_groups_per_node can be smaller than the max_priority of level capabilities. Do we see a need for that (i.e. not capping max_priority capability based on how many wfq groups can be supported)?

If we need this field, then maybe there should be a corresponding field in odp_tm_requirements_t or odp_tm_level_requirements_t, like there are for most capabilities.

Why is only for the priority_override case? If this capability value is added, why shouldn't it apply equally to the old style priority handling? If this capability field turns out to be unrelated to the priority_override setting, then I think it should be introduced in a separate patch.

There is no explanation of what WFQ groups is in the API. There is in the commit message but it should be in the API spec.

include/odp/api/spec/traffic_mngr.h
```diff
@@ -1233,6 +1253,12 @@ typedef struct {
 	 * greater levels may be connected to the fan-in of tm_node's with
 	 * numerically smaller levels. */
 	uint8_t level;
+
+	/** The strict priority level assigned to packets going through this
+	 * node. In other words all packets going through node
+	 * will be assigned this priority overriding previous priority.
+	 * in the range 0..max_priority. */
+	uint8_t priority;
```
JannePeltonen, line 1449: Looking at the text above one thinks that this priority will always set. The fact that the priority gets set only in the "override mode" is hidden in the capability description. The description of this field needs some info on when this value matter and the priority gets overridden.

include/odp/api/spec/traffic_mngr.h
```diff
@@ -625,6 +625,34 @@ odp_tm_t odp_tm_find(const char            *name,
...
+ * @return        0 upon success, < 0 upon failure.
+ */
+int odp_tm_start(odp_tm_t odp_tm);
+
+/** Stop a TM system.
+ *
+ * odp_tm_stop() can to be used to stop a TM system that is already started.
+ * Stopping a TM system will cause a TM system to stop / pause processing
+ * packets. This can be used to idle a TM system. Applications cannot call
+ * tm enqueue on a TM system which is in stopped state.
```
JannePeltonen, line 783: I wonder if it is a bit too strict a requirement to applications to not call tm_enqueue() in the stopped state. AFAICS, calling packet output is ok even when the pktio is in the stopped state. Perhaps (?) this could just say that packets sent in the stopped state may get dropped or may stay in queues until the TM goes to the started state. If we do disallow tm_enqueue() in stopped state, should we ban also odp_ipsec_out_inline() when TM pipelining could happen?

include/odp/api/spec/packet_io.h
```diff
@@ -106,6 +106,10 @@ typedef enum odp_pktout_mode_t {
 	ODP_PKTOUT_MODE_QUEUE,
 	/** Packet output through traffic manager API */
 	ODP_PKTOUT_MODE_TM,
+	/** Packet output through traffic manager API where
+	 * TM queues act as event queues.
+	 */
+	ODP_PKTOUT_MODE_TM_QUEUE,
```
JannePeltonen, line 112: I thought that the current behavior already was order-preserving as the linux-gen implementation appears to be like that. But maybe there are use case for non-order preserving enqueue too so that two modes here are useful.

include/odp/api/spec/traffic_mngr.h
```diff
@@ -1655,6 +1655,20 @@ int odp_tm_queue_disconnect(odp_tm_queue_t tm_queue);
...
+/** The odp_tm_enq_multi() function is used to add packets to a given TM system.
+ * This API takes multiple packets to enqueue unlike odp_tm_enq() but
+ * rest of the behaviour is identical to that of odp_tm_enq().
+ *
+ * @param tm_queue  Specifies the tm_queue (and indirectly the TM system).
+ * @param packets   Array of packets to enqueue.
+ * @param num       Number of packets to send.
+ * @return          Returns >0 upon success indicating number of packets
+ *                  enqueued, <= 0 upon failure. One of the more common
+ *                  failure reasons is WRED drop.
```
JannePeltonen, line 1827: I wonder how this is supposed to be used with WRED. How should the function behave when WRED wants to drop the 5th packet out of a 10 packet batch? Comsume all 10 packets, dropping the 5th one, and return 10? Consume all 10 packets, dropping the 5th one, and return 9? Consume 4 packets and return 4 and then return failure the next time the app calls the function (but maybe we probabilistic drop decision is not the same anymore)?

So this clearly needs more detailed specification on which packets get consumed, how (if?) WRED drop gets indicated etc.

include/odp/api/spec/traffic_mngr.h
```diff
@@ -1666,6 +1680,19 @@ int odp_tm_enq(odp_tm_queue_t tm_queue, odp_packet_t pkt);
...
+ * or ODP_EVENT_PACKET to tm queue only when PKTOUT mode is
+ * ODP_PKTOUT_MODE_TM_QUEUE. Apart from this, rest of the behaviour is
+ * identical to that of odp_tm_enq().
+ *
+ * @param tm_queue  Specifies the tm_queue (and indirectly the TM system).
+ * @param ev        Event handle representing a packet or packet vector.
+ * @return          Returns 0 upon success, < 0 upon failure. One of the
+ *                  more common failure reasons is WRED drop.
+ */
+int odp_tm_queue_enq(odp_tm_queue_t tm_queue, odp_event_t ev);
```
JannePeltonen, line 1854: This seems to be unrelated to everything else in this PR. Adding vector support in TM output should be in a separate PR, but I do not even see clear need for it, especially after the enqueue_multi gets added. Vectors are now only used in packet input to basically tie a set of packets into single event and thus single scheduling context. I do not see similar need in packet output.

include/odp/api/spec/traffic_mngr.h
```diff
@@ -244,6 +244,10 @@ typedef struct {
 	 * When true the min_weight and max_weight fields above specify
 	 * the legal range of such weights. */
 	odp_bool_t weights_supported;
+
+	/** tm_node_threshold_supported indicates that the tm_nodes at this
+	 * level support threshold profile set. */
+	odp_bool_t tm_node_threshold_supported;
```
JannePeltonen, line 274: Why do we need this capability field instead of requiring the threshold being always supported? As far as I can see, the threshold profile is basically just maximum queue length specification and thus seems mandatory everywhere where there can be queues.

That said, I do not quite understand why there is threshold config (and wred config) in every level as opposed to only in tm_queues. If tm_queues are the only place packets get queues, one would think that that is the only place where wred or tail drop gets applied so it is hard to see what the queue related parameters for tm nodes do. If there is some sort of virtual queuing in may levels so that the multilevel threshold and wred config would make sense, I think the API fails to describe it.

If this capability field makes sense and ends up getting added in the API, then I think corresponding field is needed in the level requirements struct.

I think we could drop _supported suffix from the newly added capability fields. It is a bit redundant since capability structures are about what is supported and what is not. This comment applies to all the added fields in this patch set.

include/odp/api/spec/traffic_mngr.h
```diff
@@ -351,6 +355,26 @@ typedef struct {
...
+	 * supported.
+	 */
+	odp_bool_t tm_queue_threshold_supported;
+
+	/** tm_queue_query_mask indicates supported types of TM queue query.
+	 * Types of TM queue query are same as query_flags that are passed to
+	 * odp_tm_queue_query(), odp_tm_priority_query() and
+	 * odp_tm_total_query(). When zero, none of the queue query API's are
+	 * supported. When non-zero, only the supported fields are filled
+	 * irrespective of the query_flags passed and rest of the fields are
```
JannePeltonen, line 404: Maybe not "irrespective of the query flags passed" but so that only the supported query flags are taken into account in the query functions.

include/odp/api/spec/traffic_mngr.h
```diff
@@ -351,6 +355,26 @@ typedef struct {
...
+	 * untouched.
+	 *
+	 * @see ODP_TM_QUERY_PKT_CNT, ODP_TM_QUERY_BYTE_CNT,
+	 * ODP_TM_QUERY_THRESHOLDS.
+	 */
+	uint8_t tm_queue_query_mask;
```
JannePeltonen, line 410: Somehow the name _mask does not sound good to me. I would name this tm_queue_query_flags.

JannePeltonen, line 410: Since query flags in the query functions is uint32_t, this should be too.

include/odp/api/spec/traffic_mngr.h
```diff
@@ -549,6 +573,41 @@ void odp_tm_egress_init(odp_tm_egress_t *egress);
...
 
+/** Query All TM Capabilities specific to an egress
+ *
+ * The odp_tm_egress_capabilities() function can be used to obtain the complete
+ * set of TM limits supported by this implementation for a given egress.
+ * The reason that this  returns a SET of capabilities and not just one, is
+ * because it is expected  that many HW based implementations may have one set
+ * of limits for the HW and also support a SW TM implementation with a
+ * (presumably larger) different set of limits.  There are also cases where
+ * there could be more than SW implementation (one supporting say tens of
```
JannePeltonen, line 663: Typo: "more that SW" -> "more than one SW"

include/odp/api/spec/traffic_mngr.h
```diff
@@ -549,6 +573,41 @@ void odp_tm_egress_init(odp_tm_egress_t *egress);
...
+ * thousands of tm_queues and a variant supporting tens of millions of
+ * tm_queues). This is needed as platforms might have different constraints
+ * for different egress type and there is no way to know them using
+ * odp_tm_capabilities() provided capabilities. The caller passes in an array
+ * of odp_tm_capabilities_t records  and the number of such records.  Then the
+ * first N of these records will be  filled in by the implementation and the
+ * number N will be returned.  In the event that N is larger than the
+ * capabilities_size, N will still be returned, but only capabilities_size
+ * records will be filled in.
```
JannePeltonen, line 672: If there are more than one TM implementation for a given egress and they have different capabilities, how is the one used selected? I do not think the requirement struct currently has a field for every capability so can there be a problem that at TM creation the implementation all implementation alternatives for a given egress have sufficient capabilities and the implementation just picks one of them but then later it turns out that it should have picked an other one?

It would improve readability if the description was not all in single paragraph.

include/odp/api/spec/traffic_mngr.h
nithind1988: Agreed, PKTIO start/stop enforcement will work for TM with PKTIO egresses but will stop both pktin and pktout.
Our platform cannot however support TM non pktio egress. So you can take a call on whether we can have TM start and TM stop or enforce odp_pktio_start()/stop().

include/odp/api/spec/traffic_mngr.h
nithind1988: Ack. I'll add it as mode in odp_tm_capabilities_t and odp_tm_requirements_t so that per-TM, user can mention the mode based on capability while doing odp_tm_create().

Per-node config will be complicated to visualize/design the topology I think.

include/odp/api/spec/traffic_mngr.h
nithind1988: Ack, will improve documentation.

include/odp/api/spec/traffic_mngr.h
nithind1988: I think at least in our HW, the behavior should be same as odp_pktio_stop(). PKTIO stop description says that new packets may not be accepted as well.

include/odp/api/spec/traffic_mngr.h
nithind1988: odp_queue_enq() already accepts ODP_EVENT_PACKET_VECTOR type of events even in linux-generic(pktout_enqueue()). So, this should be same, behavior I believe ?

include/odp/api/spec/traffic_mngr.h
nithind1988: Will fix documentation

include/odp/api/spec/traffic_mngr.h
nithind1988: Ack.

nithind1988: Ack.

include/odp/api/spec/packet_io.h
nithind1988: Yes, it is equivalent to Direct mode and Queue mode in non-TM PKTIO.

include/odp/api/spec/traffic_mngr.h
JannePeltonen: Sounds ok to me.

include/odp/api/spec/traffic_mngr.h
nithind1988: In our HW, we don't support WRED and we cannot support tail drop based on max_bytes on leaf/SQ.
On TM nodes' we don't support any kind of thresholds and as you said, I'm not sure if there are any HW out which support such behavior hence added it in capabilities.

I'll fix the name and add it in requirements struct as well.

include/odp/api/spec/traffic_mngr.h
JannePeltonen: I, with an application writer hat on, would think that the "may not be accepted" language in pktio means that I am allowed to call packet output functions but the packets may not go anywhere (they get dropped or I get an error). To say that tm_enqueue() cannot be called at all (or maybe the whole thing would crash?) would be a stronger requirement and might force an application writer do do additional synchronization between threads when a TM needs to be stopped to make sure that tm_enqueue() is never called when not allowed.

include/odp/api/spec/traffic_mngr.h
nithind1988: My guess is for WRED, stats should be in different mechanism and we should report all the 10 consumed if space is there even though some packets get dropped due to WRED.
However in our HW, we don't support WRED so, I cannot really comment here concretely

include/odp/api/spec/traffic_mngr.h
nithind1988: Ack, I get what you mean. I'll make the text same as PKTIO ?

include/odp/api/spec/traffic_mngr.h
JannePeltonen: As discussed, node_priority_override sounds a bit weird (we are not overriding the priority of a node). Maybe packet_priority_override, priority_override or priority_rewrite or something like that would be better?

include/odp/api/spec/traffic_mngr.h
JannePeltonen: As discussed, let's have the TM specific start & stop functions.

include/odp/api/spec/traffic_mngr.h
JannePeltonen: Note that TM output may happen also indirectly, through inline IPsec output. I think it would be good to specifiy that when TM is in the stopped state, packets sent to the TM (directly or indirectly) may get silently dropped or may get transmitted without TM treatment or the sending operation may fail.

include/odp/api/spec/packet_io.h
JannePeltonen: As proposed by Petri, Let's not have a new pktout mode. From pktout point-of-view it is sufficient to tell that TM will be used.

Whether input queue order is preserved in TM output (the same way as in ODP queue enqueue) can be configured through the TM API. We discussed that such config could be part of odp_tm_requirements_t passed to odp_tm_create(), but now I am wondering if it should actually be part of odp_tm_queue_param_t and thus a separate (possibly differing) setting for each input queue of a TM.

include/odp/api/spec/traffic_mngr.h
JannePeltonen: I think it would make sense to just consume all packets if (W)RED is used and then have (W)RED drop the ones it wants. I am not sure if tail drop should work the same way. Maybe.

include/odp/api/spec/traffic_mngr.h
JannePeltonen: As discussed, having a TM output function that takes events as input is unrelated to the rest of the changes. In particular, output with input-order-preservation does not require it. Let's not add it now. It will be possible to add it later in the API if needed. IPsec does not have such event-taking output function either.

include/odp/api/spec/traffic_mngr.h
JannePeltonen: It seems I misunderstood the meaning of max_wfq_groups_per_node. As discussed, the intention of the capability was to tell that only that many WFQ groups within a TM node can have more than one input among which WFQ is done. Whether we say that the rest of the WFQ groups can have only single input or that there are no more WFQ groups and the rest of the input do not go through WFQ schedulers at all may be just a matter of how to model and describe things in the API.

nithind1988:
Resolved most of the comments and sent the next version split into 4 pull requests as below.

https://github.com/OpenDataPlane/odp/pull/1234
https://github.com/OpenDataPlane/odp/pull/1235
https://github.com/OpenDataPlane/odp/pull/1236
https://github.com/OpenDataPlane/odp/pull/1237

JannePeltonen:
Can we close this PR as all the changes are covered in PRs 1234-1238?

PR #1093, opened 2020-11-11
[PATCH v1] api: ipsec: add flag to indicate inline enabled SA

anoobj:
Allow application to specify whether the SA created can be used for
inline IPsec operations. With this flag, application can indicate the
SAs that need to be inline processed. ESP packets for which the SA
has 'is_inline' set to 0, the packet will not be processed inline and
application can then submit the same to ODP for lookaside processing
using odp_ipsec_in() & odp_ipsec_in_enq().

The features supported in inline & non-inline(lookaside) path need not
be the same. So indicating the usage model will allow application to
create the SA with required flags and be assured that datapath APIs
won't suffer a failure due to lack of feature support.

For inline IPsec enabled inbound SAs, the lookup_mode must not be
ODP_IPSEC_LOOKUP_DISABLED, as with inline processing, lookup need to be
done before application receives the ESP packet.

Signed-off-by: Vidya Sagar <vvelumuri@marvell.com>

### DIFF
--- a/include/odp/api/spec/ipsec.h
+++ b/include/odp/api/spec/ipsec.h
@@ -692,6 +692,15 @@ typedef struct odp_ipsec_sa_param_t {
 	 */
 	uint32_t context_len;
 
+	/** Specify whether SA is to be used for inline IPsec
+	 *
+	 *  Allow inline IPSEC processing to be performed with the SA. For
+	 *  inbound SA, lookup_mode must not be ODP_IPSEC_LOOKUP_DISABLED
+	 *  as IPSEC processing need to be performed before application
+	 *  receives the packet for performing SA lookup.
+	 */
+	odp_bool_t is_inline;
+
 	/** IPSEC SA direction dependent parameters */
 	union {
 		/** Inbound specific parameters */


JannePeltonen:
> So indicating the usage model will allow application to create the SA with required flags and be assured that datapath APIs won't suffer a failure due to lack of feature support.

I am not sure I fully get this. I think the main point of introducing this flag is that it allows disabling inline processing for selected SAs while still allowing SA lookup offload in the async look-a-side operation. And in the output path it may allow potential optimizations since not all SAs in inline mode need to be prepared for inline processing.

include/odp/api/spec/ipsec.h
```diff
@@ -692,6 +692,15 @@ typedef struct odp_ipsec_sa_param_t {
...
 	uint32_t context_len;
 
+	/** Specify whether SA is to be used for inline IPsec
+	 *
+	 *  Allow inline IPSEC processing to be performed with the SA. For
+	 *  inbound SA, lookup_mode must not be ODP_IPSEC_LOOKUP_DISABLED
+	 *  as IPSEC processing need to be performed before application
+	 *  receives the packet for performing SA lookup.
+	 */
+	odp_bool_t is_inline;
```
JannePeltonen, line 702: I am not sure if this is clear enough on on what happens to received packets that match an SA that has is_inline false.

For lookup mode the API spec says: "In inline mode, a lookup miss directs the packet back to normal packet input interface processing".

I would say that in inline processing path it should be considered a lookup miss if a packet would match a non-inline SA. IOW, such SAs are not participating in the SA lookup in inline path regardless of the lookup mode configured for the SA. But I think the proposed API text could be interpreted also differently, which is a problem.

Then I think other parts of the API spec need some updates for clarity.

How about the API change below? It also fixes a bug in the API spec regarding lookup miss handling (packets get directed to normal input interface processing after lookup miss from the inline processing path, not always in inline mode where also async look-a-side operations can be used).

diff --git a/include/odp/api/spec/ipsec.h b/include/odp/api/spec/ipsec.h
index 48b046e43..9491e2489 100644
--- a/include/odp/api/spec/ipsec.h
+++ b/include/odp/api/spec/ipsec.h
@@ -607,10 +607,20 @@ typedef enum odp_ipsec_frag_mode_t {
  * Packet lookup mode
  *
  * Lookup mode controls how an SA participates in SA lookup offload.
- * Inbound operations perform SA lookup if application does not provide a SA as
- * a parameter. In inline mode, a lookup miss directs the packet back to normal
- * packet input interface processing. SA lookup failure status
- * (status.error.sa_lookup) is reported through odp_ipsec_packet_result_t.
+ * Inbound operations perform SA lookup if application does not provide
+ * a SA as a parameter.
+ *
+ * SA lookup failure status (status.error.sa_lookup) is reported through
+ * odp_ipsec_packet_result_t for other than inline processed packets.
+ *
+ * In inline mode, only those SAs that have inline processing enabled
+ * (inline_ena true in SA parameters) can be found by the lookup offload
+ * in the inline processing path. A lookup miss directs the packet back
+ * to normal packet input interface processing as if the packet was not
+ * an IPsec packet.
+ *
+ * Lookup mode must not be ODP_IPSEC_LOOKUP_DISABLED for SAs that
+ * have inline processing enabled.
  */
 typedef enum odp_ipsec_lookup_mode_t {
        /** Inbound SA lookup is disabled for the SA. */
@@ -678,6 +688,9 @@ typedef struct odp_ipsec_sa_param_t {
        /** SPI value */
        uint32_t spi;

+       /** Inline processing enabled */
+       odp_boot_t inline_ena;
+
        /** Destination queue for IPSEC events
         *
         *  Operations in asynchronous or inline mode enqueue resulting events
@@ -1456,10 +1469,10 @@ int odp_ipsec_out(const odp_packet_t pkt_in[], int num_in,
  * Logically, packet processing (e.g. sequence number check) happens in the
  * output order as defined above.
  *
- * The function may be used also in inline processing mode, e.g. for IPSEC
- * packets for which inline processing is not possible. Packets for the same SA
- * may be processed simultaneously in both modes (initiated by this function
- * and inline operation).
+ * The function may be used also in inline processing mode for all SAs,
+ * e.g. for IPSEC packets for which inline processing is not possible.
+ * Packets for the same SA may be processed simultaneously in both modes
+ * (initiated by this function and inline operation).
  *
  * @param          pkt      Packets to be processed
  * @param          num      Number of packets to be processed
@@ -1494,8 +1507,8 @@ int odp_ipsec_in_enq(const odp_packet_t pkt[], int num,
  * Logically, packet processing (e.g. sequence number assignment) happens in the
  * output order as defined above.
  *
- * The function may be used also in inline processing mode, e.g. for IPSEC
- * packets for which inline processing is not possible.
+ * The function may be used also in inline processing mode for all SAs,
+ * e.g. for IPSEC packets for which inline processing is not possible.
  *
  * @param          pkt      Packets to be processed
  * @param          num      Number of packets to be processed
@@ -1517,6 +1530,9 @@ int odp_ipsec_out_enq(const odp_packet_t pkt[], int num,
  * transformed packets to the specified output interface (or tm_queue), instead of
  * generating events for those.
  *
+ * This function may be called only for SAs that have inline processing
+ * enabled (inline_ena SA parameter set true).
+ *
  * Inline operation parameters are defined per packet. The array of parameters
  * must have 'num' elements and is pointed to by 'inline_param'.
  *

anoobj: So your suggestion is, all SAs (both lookaside & inline) would be processed by lookaside APIs (odp_ipsec_in/out) while inline APIs & processing would be reserved to the ones with inline_ena = true. From top level this looks okay, but I still think we are not addressing the real issue. Inline & lookaside can have different kind of feature sets (cipher support/udp_encap/checksum/fragmentation). Let's take the example of lookup_mode. 

ODP_IPSEC_LOOKUP_DSTADDR_SPI can be done in s/w quickly and so can be supported for lookaside operations. But the same getting supported in inline path would be more demanding on hardware. In addition, there is odp_ipsec_inbound_config_t which specifies lot of inline specific options including lookup range.

Also, here another assumption is that SA create will populate the hardware fields such that it can be used for inline and lookaside, which is also not true on all hardware.

So I still think we would need a flag to indicate inline only & lookaside only. But I do understand your preference to keep inline+lookaside mode (assuming that packet re-ordering is not a problem). May be what we need is a bit field stating what all modes can be supported with the SA.

```
diff --git a/include/odp/api/spec/ipsec.h b/include/odp/api/spec/ipsec.h
index 8766a4370..03846de6a 100644
--- a/include/odp/api/spec/ipsec.h
+++ b/include/odp/api/spec/ipsec.h
@@ -39,6 +39,12 @@ extern "C" {
  * Invalid IPSEC SA
  */

+#define ODP_IPSEC_OP_MODE_SYNC (1 << 0)
+
+#define ODP_IPSEC_OP_MODE_ASYNC (1 << 1)
+
+#define ODP_IPSEC_OP_MODE_INLINE (1 << 2)
+
 /**
  * IPSEC operation mode
  */
@@ -711,6 +717,8 @@ typedef struct odp_ipsec_sa_param_t {
         */
        uint32_t context_len

+       uint8_t op_mode;
+
        /** IPSEC SA direction dependent parameters */
        union {
                /** Inbound specific parameters */
```

This would mean, implementations have the option to allow only the combinations it supports. Like, if an implementation cannot do LOOKASIDE+INLINE, then SA create itself would fail for that implementation. But for an implementation like linux-generic, it would be supported.

JannePeltonen: > So your suggestion is, all SAs (both lookaside & inline) would be processed by lookaside APIs (odp_ipsec_in/out) while inline APIs & processing would be reserved to the ones with inline_ena = true.

I thought that was the intent of this PR. If the intent was to disallow look-a-side processing for inline-enabled SAs, then the spec test for the async lookaside operations would need to be updated to make it explicit that inline enabled SAs cannot be used there. When writing my version of the API change I added the word "all" there to remove ambiguity on whether inline enabled SAs can be used.

> Inline & lookaside can have different kind of feature sets (cipher support/udp_encap/checksum/fragmentation)

I would think it would not be unreasonable to require look-a-side support everything that inline can, even if the opposite were not true. Look-a-side can more easily use SW processing to fill possible gaps. OTOH, if look-a-side and inline use different HW (e.g. supporting different algorithms), then this can become difficult.

Another thing to ponder is to what extent SA creation can simply fail because of an unsupported feature and to what extent such things should be visible in IPsec capabilities. But I realise that simple capability bits cannot express what combinations can or can not be supported.

> So I still think we would need a flag to indicate inline only & lookaside only.

inline-only would also make inline implementation harder as it would require it to be able to process all packets (including fragmented packets). Or alternatively there needs to be some capability info telling the application what it can expect the inline SA to be able to do. In the current scheme inline can just not process a packet and the application can then fall back to look-a-side processing.

If we really want to have inline-only SAs, then having separate enables for all processing types would seem to make sense. Then we probably should also reconsider allowing sync processing at the same time as async or inline processing and change the global mode setting to a set of enable flags.

anoobj: > 
> 
> > So your suggestion is, all SAs (both lookaside & inline) would be processed by lookaside APIs (odp_ipsec_in/out) while inline APIs & processing would be reserved to the ones with inline_ena = true.
> 
> I thought that was the intent of this PR. If the intent was to disallow look-a-side processing for inline-enabled SAs, then the spec test for the async lookaside operations would need to be updated to make it explicit that inline enabled SAs cannot be used there. When writing my version of the API change I added the word "all" there to remove ambiguity on whether inline enabled SAs can be used.

Understood. I do have plans to update validation tests. Just that, since this is a fundamental shift in how we treat SAs, I wanted to first discuss it deeply before we start moving in the direction.

> 
> > Inline & lookaside can have different kind of feature sets (cipher support/udp_encap/checksum/fragmentation)
> 
> I would think it would not be unreasonable to require look-a-side support everything that inline can, even if the opposite were not true. Look-a-side can more easily use SW processing to fill possible gaps. OTOH, if look-a-side and inline use different HW (e.g. supporting different algorithms), then this can become difficult.

I agree to your thought process on inline being viewed as a mode with subset of features supported in look-aside. But such an assumption would mean, feature set has to be such that between inline & look-aside. Feature set being different is quite possible. Also, implementations may not be able to share IPsec context (internal IPsec state created during SA create)  between inline & look-aside.

> 
> Another thing to ponder is to what extent SA creation can simply fail because of an unsupported feature and to what extent such things should be visible in IPsec capabilities. But I realise that simple capability bits cannot express what combinations can or can not be supported.

That is also correct. I had the same question some time back as well. I think @psavol agreed that we cannot have all features documented as capabilities. Will be really challenging when we have variation in feature support (look-aside vs inline)

> 
> > So I still think we would need a flag to indicate inline only & lookaside only.
> 
> inline-only would also make inline implementation harder as it would require it to be able to process all packets (including fragmented packets). Or alternatively there needs to be some capability info telling the application what it can expect the inline SA to be able to do. In the current scheme inline can just not process a packet and the application can then fall back to look-a-side processing.

Unfortunately, for most use cases that is the sad reality. Even though ODP spec would allow applications to fall back to look-aside, applications may not choose to take that route because of packet ordering problems. So if application wants something to be used with inline, it has to be certain that all packets can be processed in inline mode itself. Any failure in the inline path would be treated as an error in that case. 

But I do agree it is up to the application to make that choice. Hence my suggestion to introduce flags to indicate all combinations. 

> 
> If we really want to have inline-only SAs, then having separate enables for all processing types would seem to make sense. Then we probably should also reconsider allowing sync processing at the same time as async or inline processing and change the global mode setting to a set of enable flags.

Agreed. Right now it's an enum with documentation which says INLINE means SYNC & ASYNC is also allowed. We can actually refine it further. If you can confirm you are okay with such a change, we can work on spec change. The spec change does look massive but I do believe that's the right way rather. Otherwise, it would be like ODP doesn't support inline only devices (like smart NICs).

PR #1094, opened 2020-11-11
[PATCH v3] Some IPsec fixes

JannePeltonen:


platform/linux-generic/odp_ipsec_sad.c
```diff
@@ -387,6 +387,16 @@ uint32_t _odp_ipsec_auth_digest_len(odp_auth_alg_t auth)
 	}
 }
 
+static uint32_t esp_block_len_to_mask(uint32_t block_len)
+{
+        /* ESP trailer should be 32-bit right aligned */
```
JereLeppanen, line 392: This line seems to be indented with spaces. Checkpatch should complain about this.

JereLeppanen (commented):
Just a minor comment about whitespace, otherwise looks OK to me. I'm not sure I would have switched from block_len (actually pad align) to a mask, but this is fine.

platform/linux-generic/odp_ipsec_sad.c
JannePeltonen: Fixed in v2. Also added a commit description (in addition to the subject line) in the fourth patch to prevent checkpatch from warning about it.

psavol (approved):

JannePeltonen:
v3: Added reviewed-by tags

PR #1095, opened 2020-11-12
[PATCH v4] api: ipsec: allow IPsec post-processing in odp_ipsec_result()

anoobj:
Allow implementations to perform additional post-processing in
odp_ipsec_result(). If applications consumes packet data without
checking IPsec result status, the behavior is undefined.

### DIFF
--- a/include/odp/api/spec/ipsec.h
+++ b/include/odp/api/spec/ipsec.h
@@ -63,6 +63,9 @@ typedef enum odp_ipsec_op_mode_t {
 	  * Packet input/output is connected directly to IPSEC inbound/outbound
 	  * processing. Application uses asynchronous or inline IPSEC
 	  * operations.
+	  *
+	  * Inline processed inbound packets are delivered to the application
+	  * in the same way as packets processed by odp_ipsec_in_enq().
 	  */
 	ODP_IPSEC_OP_MODE_INLINE,
 
@@ -1575,6 +1578,12 @@ int odp_ipsec_out(const odp_packet_t pkt_in[], int num_in,
  * may be processed simultaneously in both modes (initiated by this function
  * and inline operation).
  *
+ * Post-processing may be required after the reception of an IPsec packet
+ * event to complete IPsec processing for the packet. The post-processing
+ * happens in the odp_ipsec_result() function that must be called at least
+ * once before packet data or metadata (other than packet type and subtype)
+ * may be accessed.
+ *
  * @param          pkt      Packets to be processed
  * @param          num      Number of packets to be processed
  * @param          param    Inbound operation parameters
@@ -1610,6 +1619,12 @@ int odp_ipsec_in_enq(const odp_packet_t pkt[], int num,
  *
  * The function may be used also in inline processing mode, e.g. for IPSEC
  * packets for which inline processing is not possible.
+ *
+ * Post-processing may be required after the reception of an IPsec packet
+ * event to complete IPsec processing for the packet. The post-processing
+ * happens in the odp_ipsec_result() function that must be called at least
+ * once before packet data or metadata (other than packet type and subtype)
+ * may be accessed.
  *
  * @param          pkt      Packets to be processed
  * @param          num      Number of packets to be processed


JannePeltonen:
Now I started wondering why we want to allow post processing in odp_ipsec_result() instead of odp_ipsec_packet_from_event(). I do not see why not the latter. It would have the benefit of not having to define what packet metadata is invalid and in what way before calling_odp_ipsec_result(). As soon as the application has a packet handle, it would refer to a valid packet.

This would work for async and inline cases which return events. Sync case returns packets, but there we should not have a need to do post processing outside the actual sync call.

anoobj:
> Now I started wondering why we want to allow post processing in odp_ipsec_result() instead of odp_ipsec_packet_from_event(). I do not see why not the latter. It would have the benefit of not having to define what packet metadata is invalid and in what way before calling_odp_ipsec_result(). As soon as the application has a packet handle, it would refer to a valid packet.

For the regular packets, usage of odp_ipsec_packet_from_event() for post-processing is actually better. But this would get complicated when event is packet vector. For such a scenario, one event will translate to multiple packets each of which may or may not be IPsec.

For packet vector case, there is odp_packet_vector_tbl() API which gets the array of packets from the event. I see the following options,

1. For regular packets, let odp_ipsec_packet_from_event() do post-processing & and for packet vector, let odp_packet_vector_tbl() do the same. Cons: we would be doing IPsec post-processing in a non-IPsec API thereby incurring cost (for checks) in non-IPsec path.

2. For vector packet, introduce an API odp_ipsec_packet_vector_tbl() which would be identical to odp_packet_vector_tbl() but additionally performs post-processing of IPsec packets. Since one vector can have IPsec & non-IPsec, the API should be such that, non-IPsec packets would be untouched by the API. Cons: introduction of new API.

Another option is to introduce a new API for post-processing and make that the mandatory API. Like, 
odp_event_t
odp_ipsec_post_process(odp_event_t ev);

Returning event is to allow cases where platforms would need to update event handle.

JannePeltonen:
How about the API change below?

This would allow incomplete packets only with async (and inline, but I am not sure where to put the text for that as there is no much description of inline ipsec processing as there is no function that initiates it) mode and still require fully completed IPsec processing in sync mode.

This puts the description in the odp_ipsec_in_enq() and odp_ipsec_out_enq() documentation as the behaviour (the result) of those functions is affected.

```
diff --git a/include/odp/api/spec/ipsec.h b/include/odp/api/spec/ipsec.h
index 156d66019..c9219a3d0 100644
--- a/include/odp/api/spec/ipsec.h
+++ b/include/odp/api/spec/ipsec.h
@@ -1568,20 +1568,26 @@ int odp_ipsec_out(const odp_packet_t pkt_in[], int num_in,
  * multiple threads for the same SA(s).
  *
  * Logically, packet processing (e.g. sequence number check) happens in the
  * output order as defined above.
  *
  * The function may be used also in inline processing mode, e.g. for IPSEC
  * packets for which inline processing is not possible. Packets for the same SA
  * may be processed simultaneously in both modes (initiated by this function
  * and inline operation).
  *
+ * Post-processing may be required after the reception of an IPsec packet
+ * event to complete IPsec processing for the packet. The post-processing
+ * happens in the odp_ipsec_result() function that must be called at least
+ * once before packet data or metadata (other than packet type and subtype)
+ * may be accessed.
+ *
  * @param          pkt      Packets to be processed
  * @param          num      Number of packets to be processed
  * @param          param    Inbound operation parameters
  *
  * @return Number of input packets consumed (0 ... num)
  * @retval <0     On failure
  *
  * @see odp_ipsec_in(), odp_ipsec_result()
  */
 int odp_ipsec_in_enq(const odp_packet_t pkt[], int num,
@@ -1604,20 +1610,26 @@ int odp_ipsec_in_enq(const odp_packet_t pkt[], int num,
  * Packet order is also maintained when application otherwise guarantees
  * (e.g. using locks) that the operation is not called simultaneously from
  * multiple threads for the same SA(s).
  *
  * Logically, packet processing (e.g. sequence number assignment) happens in the
  * output order as defined above.
  *
  * The function may be used also in inline processing mode, e.g. for IPSEC
  * packets for which inline processing is not possible.
  *
+ * Post-processing may be required after the reception of an IPsec packet
+ * event to complete IPsec processing for the packet. The post-processing
+ * happens in the odp_ipsec_result() function that must be called at least
+ * once before packet data or metadata (other than packet type and subtype)
+ * may be accessed.
+ *
  * @param          pkt      Packets to be processed
  * @param          num      Number of packets to be processed
  * @param          param    Outbound operation parameters
  *
  * @return Number of input packets consumed (0 ... num)
  * @retval <0     On failure
  *
  * @see odp_ipsec_out(), odp_ipsec_result()
  */
 int odp_ipsec_out_enq(const odp_packet_t pkt[], int num,

```

JannePeltonen:
In addition to the changes I proposed earlier, I think it would be good to have this change to cover inline processing too:

```
diff --git a/include/odp/api/spec/ipsec.h b/include/odp/api/spec/ipsec.h
index c9219a3d0..0a45dd95d 100644
--- a/include/odp/api/spec/ipsec.h
+++ b/include/odp/api/spec/ipsec.h
@@ -63,6 +63,9 @@ typedef enum odp_ipsec_op_mode_t {
          * Packet input/output is connected directly to IPSEC inbound/outbound
          * processing. Application uses asynchronous or inline IPSEC
          * operations.
+         *
+         * Inline processed inbound packets are delivered to the application
+         * in the same way as packets processed by odp_ipsec_in_enq().
          */
        ODP_IPSEC_OP_MODE_INLINE,
```

Are you planning to submit an updated version or would it be more convenient if I just create a new PR with the changes I proposed?

anoobj:
> Are you planning to submit an updated version or would it be more convenient if I just create a new PR with the changes I proposed?

Please create a new PR. I'll close this once you propose the new PR.

anoobj:
> 
> 
> > Are you planning to submit an updated version or would it be more convenient if I just create a new PR with the changes I proposed?
> 
> Please create a new PR. I'll close this once you propose the new PR.

Replied too soon. I thought it was the other PR on inline processing flags. I can make the change. I'll revise the PR with the suggested text.

JannePeltonen (approved):

psavol (approved):

anoobj:
Rebased and updated review tags.

PR #1100, opened 2020-11-16
[PATCH v3] api: pool index and stats

jerinjacobk:
v3:
- Updated git commit comments

v2:
- Rebase to master
- Remove interface from odp_pool_index() API comments
- Introduced odp_pool_stats_opt_t to get supported stats counters and configure it

## api: pool: query pool index

Some of the ODP application can be better managed
if the application aware of the pool index.

The typical use case for this to create the basic building
block for effective lookup in application
as well as the debug support.

## api: pool: query pool statistics

Some of the ODP application can be better managed
if the application aware of pool stats like the number of entries
available and number of entries in local cache for the pool.

Introduce odp_pool_stats() to get pool statistics counters.

### DIFF
--- a/include/odp/api/spec/pool.h
+++ b/include/odp/api/spec/pool.h
@@ -542,6 +542,26 @@ uint64_t odp_pool_to_u64(odp_pool_t hdl);
 void odp_pool_param_init(odp_pool_param_t *param);
 
 /**
+ * Maximum pool index
+ *
+ * Return the maximum pool index. Pool indexes
+ * (e.g. returned by odp_pool_index()) range from zero to this maximum value.
+ *
+ * @return Maximum pool index
+ */
+unsigned int odp_pool_max_index(void);
+
+/**
+ * Get pool index
+ *
+ * @param pool      Pool handle
+ *
+ * @return        Pool index (0..odp_pool_max_index())
+ * @retval <0     On failure (e.g., handle not valid)
+ */
+int odp_pool_index(odp_pool_t pool);
+
+/**
  * @}
  */
 
--- a/include/odp/api/spec/pool.h
+++ b/include/odp/api/spec/pool.h
@@ -45,6 +45,57 @@ extern "C" {
 #define ODP_POOL_MAX_SUBPARAMS  7
 
 /**
+ * Pool statistics counter options bit field
+ *
+ * Pool statistics counter options listed in a bit field structure.
+ * This structure shall be used to get the supported counters for the pool
+ * and configure the statistics counters.
+ *
+ * @see odp_pool_capability_t::buf::stats_supported
+ * @see odp_pool_param_t::stats_config
+ *
+ */
+typedef union odp_pool_stats_opt_t {
+	/** Option flags */
+	struct {
+		/** Select the odp_pool_stats_t::available counter */
+		uint64_t available             : 1;
+
+		/** Select the odp_pool_stats_t::available_local_cache counter */
+		uint64_t available_local_cache : 1;
+
+		/** Select the odp_pool_stats_t::ops counter */
+		uint64_t ops                   : 1;
+	} bit;
+
+	/** All bits of the bit field structure
+	  *
+	  * This field can be used to set/clear all flags, or bitwise
+	  * operations over the entire structure. */
+	uint64_t all;
+} odp_pool_stats_opt_t;
+
+/**
+ * Pool statistics
+ */
+typedef struct odp_pool_stats_t {
+	/**
+	 * The number of available entries in the pool.
+	 */
+	uint64_t available;
+
+	/**
+	 * The number of available entries in the local cache of all the CPU of a pool.
+	 */
+	uint64_t available_local_cache;
+
+	/**
+	 * The number of free and alloc operations in the pool.
+	 */
+	uint64_t ops;
+} odp_pool_stats_t;
+
+/**
  * Pool capabilities
  */
 typedef struct odp_pool_capability_t {
@@ -76,6 +127,9 @@ typedef struct odp_pool_capability_t {
 
 		/** Maximum size of thread local cache */
 		uint32_t max_cache_size;
+
+		/** Supported statistics counters */
+		odp_pool_stats_opt_t stats_supported;
 	} buf;
 
 	/** Packet pool capabilities  */
@@ -161,6 +215,9 @@ typedef struct odp_pool_capability_t {
 
 		/** Maximum size of thread local cache */
 		uint32_t max_cache_size;
+
+		/** Supported statistics counters */
+		odp_pool_stats_opt_t stats_supported;
 	} pkt;
 
 	/** Timeout pool capabilities  */
@@ -179,6 +236,9 @@ typedef struct odp_pool_capability_t {
 
 		/** Maximum size of thread local cache */
 		uint32_t max_cache_size;
+
+		/** Supported statistics counters */
+		odp_pool_stats_opt_t stats_supported;
 	} tmo;
 
 	/** Vector pool capabilities */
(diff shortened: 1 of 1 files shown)


include/odp/api/spec/pool.h
```diff
@@ -498,6 +498,69 @@ uint64_t odp_pool_to_u64(odp_pool_t hdl);
  */
 void odp_pool_param_init(odp_pool_param_t *param);
 
+/**
+ * Maximum pool interface index
```
matiaselo, line 502: You can drop the `interface` word.

jerinjacobk: I will fix it in the next version.

include/odp/api/spec/pool.h
```diff
@@ -498,6 +498,26 @@ uint64_t odp_pool_to_u64(odp_pool_t hdl);
...
+ *
+ * Return the maximum pool interface index. Interface indexes
+ * (e.g. returned by odp_pool_index()) range from zero to this maximum value.
+ *
+ * @return Maximum pool interface index
+ */
+unsigned int odp_pool_max_index(void);
+
+/**
+ * Get pool interface index
```
matiaselo, line 512: Same goes here.
```
/**
 * Get pool index
 *
 * @param pool    Pool handle
 *
 * @return Pool index (0..odp_pool_max_index())
 * @retval <0 on failure (e.g., handle not valid)
 */
int odp_pool_index(odp_pool_t pool);
```

include/odp/api/spec/pool.h
```diff
@@ -518,6 +518,49 @@ unsigned int odp_pool_max_index(void);
...
+/**
+ * Pool statistics
+ */
+typedef struct odp_pool_stats_t {
+	/**
+	 * The number of available entries in the pool.
+	 * When the local cache is enabled, the count will account for the local cache
+	 * entries per CPU as well.
+	 */
+	uint64_t available;
```
psavol, line 530: Feels that only counting free or allocated events is enough, application can calculate the other figure. Maybe separate counters for global vs. thread local cache are needed since there will be delay when sampling those (sum might not match the pool size).

include/odp/api/spec/pool.h
```diff
@@ -518,6 +518,49 @@ unsigned int odp_pool_max_index(void);
  */
 int odp_pool_index(odp_pool_t pool);
 
+/**
+ * Pool statistics
+ */
```
psavol, line 523: Pool capability and parameter is needed to enable/disable statistics counting.

include/odp/api/spec/pool.h
jerinjacobk: Fixed in v2

include/odp/api/spec/pool.h
jerinjacobk: Fixed in v2 using by Introducin g odp_pool_stats_opt_t to get supported stats counters and configure it

matiaselo:
@jerinjacobk The API is starting to look good to me. Is it OK for you if I do the reference implementation and the validation tests? The API text needs a bit clean up here and there (no functional changes), but I can fix that while doing the reference implementation.

include/odp/api/spec/pool.h
```diff
@@ -44,6 +44,57 @@ extern "C" {
...
+
+	/**
+	 * The number of available entries in the local cache of all the CPU of a pool.
+	 */
+	uint64_t available_local_cache;
+
+	/**
+	 * The number of free and alloc operations in the pool.
+	 */
+	uint64_t ops;
```
matiaselo, line 95: This seems a bit vague. Should we instead have separate counters for allocs and frees?

jerinjacobk: On OcteonTx HW, we have only a single counter for both.(It is the total number of free and alloc operation happened during the life span of a pool). Any thoughts?

jerinjacobk:
> @jerinjacobk The API is starting to look good to me. Is it OK for you if I do the reference implementation and the validation 

Sure.

> tests? The API text needs a bit clean up here and there (no functional changes), but I can fix that while doing the reference implementation.

Sure. 

One thing, we realized while doing the implementation. Not all the counters can be reset e.s.p the level counters like "availability". I think, we can express the semantics either by:
1)  Add odp_pool_stats_opt_t based capability for expressing resettable counters
or
2) Application can read the counters after odp_pool_stats_reset() to see which counters actually reseted.

Please take care of this, while you are at taking the patch for implementation.

include/odp/api/spec/pool.h
matiaselo: OK. Thinking more about it, it could be useful to have separate counters for the global pool and local caches. 
E.g.
```
/**
 * Pool statistics
 */
typedef struct odp_pool_stats_t {
	/** The number of available entries in the pool */
	uint64_t available;

	/** The number of alloc operations from the pool */
	uint64_t alloc_ops;

	/** The number of free operations to the pool */
	uint64_t free_ops;

	/** The combined number of alloc and free operations */
	uint64_t combined_ops;

	/** Local cache statistics */
	struct {
		/** The number of available entries in the local caches of all threads using the
		 *  pool */
		uint64_t available;

		/** The number of alloc operations from the pool caches  */
		uint64_t alloc_ops;

		/** The number of free operations to the pool caches */
		uint64_t free_ops;

		/** The combined number of alloc and free operations */
		uint64_t combined_ops;

	} cache;

} odp_pool_stats_t;
```

matiaselo:
> One thing, we realized while doing the implementation. Not all the counters can be reset e.s.p the level counters like "availability". I think, we can express the semantics either by:
> 
> 1. Add odp_pool_stats_opt_t based capability for expressing resettable counters
>    or
> 2. Application can read the counters after odp_pool_stats_reset() to see which counters actually reseted.
> 
> Please take care of this, while you are at taking the patch for implementation.

A good point. I think we don't need a new capability for this. `odp_pool_stats_reset()` documentation just has to be reworded a bit.

jerinjacobk:
> > One thing, we realized while doing the implementation. Not all the counters can be reset e.s.p the level counters like "availability". I think, we can express the semantics either by:
> > 
> > 1. Add odp_pool_stats_opt_t based capability for expressing resettable counters
> >    or
> > 2. Application can read the counters after odp_pool_stats_reset() to see which counters actually reseted.
> > 
> > Please take care of this, while you are at taking the patch for implementation.
> 
> A good point. I think we don't need a new capability for this. `odp_pool_stats_reset()` documentation just has to be reworded a bit.

I agree. Adding @note in odp_pool_stats_reset() is enough.

include/odp/api/spec/pool.h
jerinjacobk: Looks good. I think, it is better to have cache_* variable instead of the new struct as opt_s is bit field.

matiaselo: True, this is the version I'm currently using for the implementation:
```
/**
 * Pool statistics
 */
typedef struct odp_pool_stats_t {
	/** The number of available entries in the pool */
	uint64_t available;

	/** The number of alloc operations from the pool */
	uint64_t alloc_ops;

	/** The number of free operations to the pool */
	uint64_t free_ops;

	/** The combined number of alloc and free operations */
	uint64_t combined_ops;

	/** The number of available entries in the local caches of all threads
	 *  using the pool */
	uint64_t cache_available;

	/** The number of alloc operations from the pool caches  */
	uint64_t cache_alloc_ops;

	/** The number of free operations to the pool caches */
	uint64_t cache_free_ops;

	/** The combined number of alloc and free operations */
	uint64_t cache_combined_ops;

} odp_pool_stats_t;
```

jerinjacobk: Looks good. The last comment is missing "from the pool caches"

jerinjacobk:
Close as it is duplicate now.

PR #1101, opened 2020-11-17
[PATCH v3] test: l2fwd: avoid tight loop in enqueue to pktout queue

nithind1988:
Currently in pktout queue mode, only negative return value
causes break from pkt enqueue loop unlike in other modes where
both negative and zero return will cause the packet to be dropped.

Some platforms always return zero which can either indicate no space
in Tx or when Tx is paused due to odp_pktio_stop()(for example in
exit sequence). In such cases, current code can become an infinite loop.

So this patch fixes the same by checking for <=0 return and
break from loop if < 0 or exit is already in progress.

### DIFF
--- a/test/performance/odp_l2fwd.c
+++ b/test/performance/odp_l2fwd.c
@@ -264,9 +264,9 @@ static inline int event_queue_send(odp_queue_t queue, odp_packet_t *pkt_tbl,
 	while (sent < pkts) {
 		ret = odp_queue_enq_multi(queue, &ev_tbl[sent], pkts - sent);
 
-		if (ret < 0) {
-			ODPH_ERR("Failed to send packet as events\n");
-			break;
+		if (odp_unlikely(ret <= 0)) {
+			if (ret < 0 || odp_atomic_load_u32(&gbl_args->exit_threads))
+				break;
 		}
 
 		sent += ret;


matiaselo (approved):

test/performance/odp_l2fwd.c
```diff
@@ -264,9 +264,9 @@ static inline int event_queue_send(odp_queue_t queue, odp_packet_t *pkt_tbl,
 	while (sent < pkts) {
 		ret = odp_queue_enq_multi(queue, &ev_tbl[sent], pkts - sent);
 
-		if (ret < 0) {
-			ODPH_ERR("Failed to send packet as events\n");
```
JereLeppanen, line 268: I don't suppose there's actually any reason to remove this error message. Maybe just keep it?

JereLeppanen (commented):
Instead of "avoid tight loop", the patch subject should probably be "avoid endless loop".

test/performance/odp_l2fwd.c
matiaselo: There is no error message in the `odp_pktout_send()` path and the TX drop counter is correctly incremented, so I think this message can be dropped.

JereLeppanen: Ah yes, I forgot about the counter.

nithind1988:
> 
> 
> Instead of "avoid tight loop", the patch subject should probably be "avoid endless loop".

Ack

PR #1102, opened 2020-11-17
[PATCH v10] api: ipsec: test sa update

asasidharan:
With IPSEC implementations, the protocol specific fields
like sequence number is not exposed to the application.
But for debugging as well as for validation, finer control
of such fields would be required. The proposed API,
odp_ipsec_test_sa_update(), would allow access to such
fields which are not expected to be exposed in the data
path.

### DIFF
--- a/include/odp/api/spec/ipsec.h
+++ b/include/odp/api/spec/ipsec.h
@@ -72,6 +72,46 @@ typedef enum odp_ipsec_op_mode_t {
 } odp_ipsec_op_mode_t;
 
 /**
+ * IPSEC TEST SA operation
+ */
+typedef enum odp_ipsec_test_sa_operation_t {
+	/** Update next sequence number
+	 *
+	 * The seq_num parameter is an outbound SA specific parameter.
+	 * Invoking the odp_ipsec_test_sa_update() API to update this
+	 * field on an inbound SA will cause the API to return failure.
+	 */
+	ODP_IPSEC_TEST_SA_UPDATE_SEQ_NUM = 0,
+
+	/** Update highest authenticated sequence number
+	 *
+	 * The antireplay_window_top parameter is inbound SA specific.
+	 * Invoking the odp_ipsec_test_sa_update() API to update this
+	 * field on an outbound SA will cause the API to return failure.
+	 */
+	ODP_IPSEC_TEST_SA_UPDATE_ANTIREPLAY_WINDOW_TOP
+
+} odp_ipsec_test_sa_operation_t;
+
+/**
+ * IPSEC TEST SA parameter
+ */
+typedef union odp_ipsec_test_sa_param_t {
+	/** Next sequence number
+	 *
+	 * @see ODP_IPSEC_TEST_SA_UPDATE_SEQ_NUM
+	 */
+	uint64_t seq_num;
+
+	/** Highest authenticated sequence number
+	 *
+	 * @see ODP_IPSEC_TEST_SA_UPDATE_ANTIREPLAY_WINDOW_TOP
+	 */
+	uint64_t antireplay_window_top;
+
+} odp_ipsec_test_sa_param_t;
+
+/**
  * Configuration options for IPSEC inbound processing
  */
 typedef struct odp_ipsec_inbound_config_t {
@@ -193,6 +233,28 @@ typedef struct odp_ipsec_outbound_config_t {
 } odp_ipsec_outbound_config_t;
 
 /**
+ * IPSEC TEST capability
+ */
+typedef struct odp_ipsec_test_capability_t {
+	/** Parameters supported for sa_update */
+	struct {
+		/** Next sequence number value
+		 *
+		 * @see ODP_IPSEC_TEST_SA_UPDATE_SEQ_NUM
+		 */
+		odp_bool_t seq_num;
+
+		/** Highest authenticated sequence number
+		 *
+		 * @see ODP_IPSEC_TEST_SA_UPDATE_ANTIREPLAY_WINDOW_TOP
+		 */
+		odp_bool_t antireplay_window_top;
+
+	} sa_operations;
+
+} odp_ipsec_test_capability_t;
+
+/**
  * IPSEC capability
  */
 typedef struct odp_ipsec_capability_t {
@@ -273,6 +335,11 @@ typedef struct odp_ipsec_capability_t {
 	 */
 	odp_support_t inline_ipsec_tm;
 
+	/** IPSEC TEST capabilities
+	 *
+	 * @see odp_ipsec_test_sa_update()
+	 */
+	odp_ipsec_test_capability_t test;
 } odp_ipsec_capability_t;
 
 /**
@@ -1714,6 +1781,36 @@ int odp_ipsec_result(odp_ipsec_packet_result_t *result, odp_packet_t packet);
  * @see odp_ipsec_sa_disable()
  */
 int odp_ipsec_status(odp_ipsec_status_t *status, odp_event_t event);
+
+/**
+ * IPSEC test API for modifying internal state of an SA.
+ *
+ * This function is not meant to be used by normal applications but by special
+ * test applications that test or debug the operation of the underlying ODP
+ * implementation. Calling this function may degrade the performance of the
+ * calling thread, other threads or the IPSEC implementation in general.
+ *
+ * Calling this function for an SA at the same time when the SA is used for
+ * processing traffic or when the SA is being modified through other parts
+ * of IPSEC API may result in undefined behaviour.
+ *
+ * SA state update through this function may not be supported by all ODP
+ * implementations, ODP instances or SA instances or at every moment. This
+ * function may return failure for unspecified reasons even when the capability
+ * call indicated support for updating a particular parameter and previous
+ * similar calls succeeded.
+ *
+ * @param          sa            IPSEC SA to be updated
+ * @param          op            Specifies operation to be performed
+ * @param          param         Pointer to IPSEC TEST SA param structure to be
+ *                               used for the operation
+ *
+ * @return 0      On success
+ * @retval <0     On failure
+ */
+int odp_ipsec_test_sa_update(odp_ipsec_sa_t sa,
+			     odp_ipsec_test_sa_operation_t op,
+			     const odp_ipsec_test_sa_param_t *param);
 
 /**
  * Update MTU for outbound IP fragmentation
--- a/test/validation/api/ipsec/ipsec.c
+++ b/test/validation/api/ipsec/ipsec.c
@@ -261,6 +261,18 @@ int ipsec_check_esp_chacha20_poly1305(void)
 				ODP_AUTH_ALG_CHACHA20_POLY1305, 0);
 }
 
+int ipsec_check_test_sa_update_seq_num(void)
+{
+	odp_ipsec_capability_t capa;
+
+	odp_ipsec_capability(&capa);
+
+	if (!capa.test.sa_operations.seq_num)
+		return ODP_TEST_INACTIVE;
+
+	return ODP_TEST_ACTIVE;
+}
+
 void ipsec_sa_param_fill(odp_ipsec_sa_param_t *param,
 			 odp_bool_t in,
 			 odp_bool_t ah,
@@ -734,6 +746,48 @@ static int ipsec_send_out_one(const ipsec_test_part *part,
 	return num_out;
 }
 
+int ipsec_test_sa_update_seq_num(odp_ipsec_sa_t sa, uint32_t seq_num)
+{
+	odp_ipsec_test_sa_operation_t sa_op;
+	odp_ipsec_test_sa_param_t sa_param;
+
+	sa_op = ODP_IPSEC_TEST_SA_UPDATE_SEQ_NUM;
+	sa_param.seq_num = seq_num;
+
+	return odp_ipsec_test_sa_update(sa, sa_op, &sa_param);
+}
+
+static void ipsec_pkt_seq_num_check(odp_packet_t pkt, uint32_t seq_num)
+{
+	uint32_t l3_off = odp_packet_l3_offset(pkt);
+	uint32_t l4_off;
+	odph_ipv4hdr_t ip;
+
+	CU_ASSERT_NOT_EQUAL_FATAL(ODP_PACKET_OFFSET_INVALID, l3_off);
+	CU_ASSERT_EQUAL_FATAL(0, odp_packet_copy_to_mem(pkt, l3_off, sizeof(ip),
+							&ip));
+
+	if (ODPH_IPV4HDR_VER(ip.ver_ihl) == ODPH_IPV4) {
+		l4_off = l3_off + (ODPH_IPV4HDR_IHL(ip.ver_ihl) * 4);
+
+		if (ip.proto == ODPH_IPPROTO_ESP) {
+			odph_esphdr_t esp;
+
+			odp_packet_copy_to_mem(pkt, l4_off, sizeof(esp), &esp);
+			CU_ASSERT_EQUAL(odp_be_to_cpu_32(esp.seq_no), seq_num);
+		} else if (ip.proto == ODPH_IPPROTO_AH) {
+			odph_ahhdr_t ah;
+
+			odp_packet_copy_to_mem(pkt, l4_off, sizeof(ah), &ah);
+			CU_ASSERT_EQUAL(odp_be_to_cpu_32(ah.seq_no), seq_num);
+		} else {
+			CU_FAIL("Unexpected IP Proto");
+		}
+	} else {
+		CU_FAIL("Unexpected IP Version");
+	}
+}
+
 static void ipsec_pkt_proto_err_set(odp_packet_t pkt)
 {
 	uint32_t l3_off = odp_packet_l3_offset(pkt);
@@ -897,6 +951,9 @@ void ipsec_check_out_in_one(const ipsec_test_part *part,
 		}
 		CU_ASSERT_FATAL(odp_packet_len(pkto[i]) <=
 				sizeof(pkt_in.data));
+
+		if (part->flags.test_sa_seq_num)
+			ipsec_pkt_seq_num_check(pkto[i], part->out[i].seq_num);
 
 		if (part->flags.stats == IPSEC_TEST_STATS_PROTO_ERR)
 			ipsec_pkt_proto_err_set(pkto[i]);
--- a/test/validation/api/ipsec/ipsec_test_out.c
+++ b/test/validation/api/ipsec/ipsec_test_out.c
@@ -460,6 +460,21 @@ static void test_out_in_common(ipsec_test_flags *flags,
 		test_ipsec_stats_zero_assert(&stats);
 	}
 
+	if (flags->test_sa_seq_num) {
+		int rc;
+
+		test.out[0].seq_num = 0x1235;
+		rc = ipsec_test_sa_update_seq_num(sa_out, test.out[0].seq_num);
+
+		/* Skip further checks related to this specific test if the
+		 * SA update call was not successful.
+		 */
+		if (rc < 0) {
+			printf("\t >> skipped");
+			test.flags.test_sa_seq_num = false;
+		}
+	}
+
 	ipsec_check_out_in_one(&test, sa_out, sa_in);
 
 	if (flags->stats == IPSEC_TEST_STATS_SUCCESS) {
@@ -1364,6 +1379,19 @@ static void test_sa_info(void)
 	ipsec_sa_destroy(sa_in);
 }
 
+static void test_test_sa_update_seq_num(void)
+{
+	ipsec_test_flags flags;
+
+	memset(&flags, 0, sizeof(flags));
+	flags.display_algo = true;
+	flags.test_sa_seq_num = true;
+
+	test_esp_out_in_all(&flags);
+
+	printf("\n  ");
+}
+
 static void ipsec_test_capability(void)
 {
 	odp_ipsec_capability_t capa;
@@ -1444,6 +1472,8 @@ odp_testinfo_t ipsec_out_suite[] = {
 				  ipsec_check_esp_null_sha256),
 	ODP_TEST_INFO_CONDITIONAL(test_sa_info,
 				  ipsec_check_esp_aes_cbc_128_sha1),
+	ODP_TEST_INFO_CONDITIONAL(test_test_sa_update_seq_num,
+				  ipsec_check_test_sa_update_seq_num),
 	ODP_TEST_INFO(test_esp_out_in_all_basic),
 	ODP_TEST_INFO_CONDITIONAL(test_esp_out_in_all_hdr_in_packet,
 				  is_out_mode_inline),
--- a/platform/linux-generic/odp_ipsec.c
+++ b/platform/linux-generic/odp_ipsec.c
@@ -136,6 +136,7 @@ int odp_ipsec_capability(odp_ipsec_capability_t *capa)
 	capa->max_queues = queue_capa.max_queues;
 	capa->inline_ipsec_tm = ODP_SUPPORT_NO;
 
+	capa->test.sa_operations.seq_num = 1;
 	return 0;
 }
 
@@ -2038,6 +2039,26 @@ err:
 	}
 
 	return in_pkt;
+}
+
+int odp_ipsec_test_sa_update(odp_ipsec_sa_t sa,
+			     odp_ipsec_test_sa_operation_t sa_op,
+			     const odp_ipsec_test_sa_param_t *sa_param)
+{
+	ipsec_sa_t *ipsec_sa;
+
+	ipsec_sa = _odp_ipsec_sa_entry_from_hdl(sa);
+	ODP_ASSERT(NULL != ipsec_sa);
+
+	switch (sa_op) {
+	case ODP_IPSEC_TEST_SA_UPDATE_SEQ_NUM:
+		odp_atomic_store_u64(&ipsec_sa->hot.out.seq, sa_param->seq_num);
+		break;
+	default:
+		return -1;
+	}
+
+	return 0;
 }
 
 int odp_ipsec_result(odp_ipsec_packet_result_t *result, odp_packet_t packet)
--- a/test/validation/api/ipsec/ipsec.h
+++ b/test/validation/api/ipsec/ipsec.h
@@ -59,6 +59,7 @@ typedef struct {
 	odp_bool_t lookup;
 	odp_bool_t ah;
 	odp_bool_t inline_hdr_in_packet;
+	odp_bool_t test_sa_seq_num;
 	enum ipsec_test_stats stats;
 } ipsec_test_flags;
 
@@ -73,6 +74,7 @@ typedef struct {
 		const ipsec_test_packet *pkt_res;
 		odp_proto_l3_type_t l3_type;
 		odp_proto_l4_type_t l4_type;
+		uint32_t seq_num;
 	} out[1];
 	struct {
 		odp_ipsec_op_status_t status;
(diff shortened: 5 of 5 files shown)


include/odp/api/spec/ipsec.h
```diff
@@ -1601,6 +1635,22 @@ int odp_ipsec_result(odp_ipsec_packet_result_t *result, odp_packet_t packet);
...
+ * is not available to the application in the datapath. The API allows
+ * application to debug and verify various features provided by the platform.
+ *
+ * @param          sa            IPSEC SA to be updated
+ * @param          sa_param      IPSEC SA param to update
+ *
+ * @return 0      On success
+ * @retval <0     On failure
+ */
+int odp_ipsec_debug_sa_update(odp_ipsec_sa_t sa, odp_ipsec_debug_sa_param_t sa_param);
```
psavol, line 1652: Any limitations on usage of this call? When this function can be called: when traffic is flowing, only before starting to use the SA, etc?

Capability needed to tell if implementation support IPsec debug functions.

asasidharan: Updated the following in V2:
- Clarified the usage of the API with the following statement:
```
+ * It is the responsiblity of the application to make sure that no traffic is   
+ * flowing through the SA when the API is called. Otherwise, this operation may 
+ * lead to undefined behavior.
```
- Added additional odp_ipsec_debug_capability_t in the existing odp_ipsec_capability_t structure.

JannePeltonen: I previously suggested having this debug stuff as an optional part of SA creation, but after the addition of the capabilities and the restriction on when the update function may be called, I think I am ok with the separate update function.

include/odp/api/spec/ipsec.h
```diff
@@ -191,6 +233,42 @@ typedef struct odp_ipsec_outbound_config_t {
...
+	odp_support_t sa_update;
+
+	/** Parameters supported for sa_update */
+	struct {
+		/** Inbound specific internal parameters */
+		union {
+			/** Mapping for individual bits */
+			struct {
+				/** Antireplay window top */
+				uint32_t antireplay_window_top  : 1;
```
JannePeltonen, line 250: I am not sure the capabilities should be defined as bit fields as it does not really save memory or cycles. Perhaps these could be just odp_bool_t (or odp_support_t?). Maybe Petri can comment?

include/odp/api/spec/ipsec.h
```diff
@@ -70,6 +70,48 @@ typedef enum odp_ipsec_op_mode_t {
...
+ * IPSEC DEBUG SA parameter type
+ */
+typedef enum odp_ipsec_debug_sa_param_type_t {
+	/** Sequence number */
+	ODP_IPSEC_DEBUG_SA_FIELD_SEQ_NUM = 0,
+
+	/* Antireplay window top */
+	ODP_IPSEC_DEBUG_SA_FIELD_ANTIREPLAY_WINDOW_TOP
+
+} odp_ipsec_debug_sa_param_type_t;
```
JannePeltonen, line 83: The names seem slightly awkward. I think the _FIELD part could be dropped.

> /** Sequence number */
I think this should be "Next sequence number".

> /* Antireplay window top */
This may also be a bit ambiguous. Maybe the comment could be "Highest authenticated sequence number"?

test/validation/api/ipsec/ipsec_test_out.c
```diff
@@ -1138,6 +1138,66 @@ static void test_out_ipv4_udp_esp_null_sha256(void)
 	ipsec_sa_destroy(sa);
 }
 
+static void test_debug_sa_update_seq_num(void)
```
JannePeltonen, line 1141: What is the purpose of this test?

According to the subject line of the commit ("add IPSEC SA sequence number test") this tests something about how SAs handle sequence numbers. Like maybe checking that sequence number starts from one and gets incremented by one for each packet. But it does not test that.

I did not understand the actual text of the commit message.

The name of the test function is ipsec_check_debug_sa_update_seq_num(), leading me to think that the test tests that odp_ipsec_debug_sa_update() somehow works. But it does not test that. odp_ipsec_debug_sa_update() could update the sequence number to some other value than requested or do nothing at all and the test would still pass.

So if the test is not about testing odp_ipsec_debug_sa_update() but about testing ODP IPsec functionality by using  dp_ipsec_debug_sa_update() as part of the test machinery, then the test function should be renamed.

But looking at the code, the only thing the test tests is that IPsec packet is succesfully received if the first received sequence number is 1235. That might test the anti-replay implementation just a little bit, except that the test does not set a non-zero value to antireplay_ws config parameter to ensure that anti-replay is enabled.

So maybe the only thing the test really tests is that the underlying ODP implementation has the SA debug API addition so that compilation does not fail? But I think a simpler test would suffice for that.

include/odp/api/spec/ipsec.h
psavol: After discussion in call, boolean makes most sense.

asasidharan: Changed antireplay_window_top and seq_num from bit fields to odp_bool_t in v3.

include/odp/api/spec/ipsec.h
asasidharan: Ack.

test/validation/api/ipsec/ipsec_test_out.c
asasidharan: > I did not understand the actual text of the commit message.

Ack. Updated the commit message to reflect the actual test being done in v3.


> odp_ipsec_debug_sa_update() could update the sequence number to some other value than requested or do nothing at all and the test would still pass.

Didn't realize this!. My original thought was to have a known vector (NULL cipher with SHA256 auth) which will have the sequence number as well embedded in the result packet. So the memcmp() operation will confirm that the sequence number is updated. But I ended up using AES-GCM here resulting in an incomplete test.

In v3, I have updated the test to check the sequence number present in the IPsec header of the resulting packet. Also I have tried to plug this into the common test framework so that this will be run for all the supported algorithm combinations.

Let me know if this looks good.

include/odp/api/spec/ipsec.h
psavol: As discussed on the call, may be "test" would be better name to use for this functionality than "debug", as user would use the API for better test coverage (and not always for debugging).

asasidharan: Ack. Updated the PR to use "test" instead of "debug" in v4.

include/odp/api/spec/ipsec.h
```diff
@@ -71,6 +71,48 @@ typedef enum odp_ipsec_op_mode_t {
...
+	/** Next sequence number value */
+	uint64_t seq_num;
+
+	/** Highest authenticated sequence number */
+	uint64_t antireplay_window_top;
+
+} odp_ipsec_test_sa_param_val_t;
+
+/**
+ * IPSEC TEST SA parameters to allow applications to override unexposed IPSEC
```
JannePeltonen, line 99: The term "override" sounds to me that the value given either permanently overrides the normal state parameter (in which case e.g. sequence number would stay constant) or once (after which the normal state parameter is used again). I think simple "set" would be better here and in other parts of the API description.

include/odp/api/spec/ipsec.h
```diff
@@ -71,6 +71,48 @@ typedef enum odp_ipsec_op_mode_t {
...
+ * IPSEC TEST SA parameter type
+ */
+typedef enum odp_ipsec_test_sa_param_type_t {
+	/** Next sequence number value */
+	ODP_IPSEC_TEST_SA_SEQ_NUM = 0,
+
+	/** Highest authenticated sequence number */
+	ODP_IPSEC_TEST_SA_ANTIREPLAY_WINDOW_TOP
+
+} odp_ipsec_test_sa_param_type_t;
```
JannePeltonen, line 84: I am not sure if calling this "type" (instead of "instance" or just parameter) is good. This way we cannot use the same "type" for different parameters, suggesting that it is not really a "type" that we are dealing with.

include/odp/api/spec/ipsec.h
```diff
@@ -192,6 +234,33 @@ typedef struct odp_ipsec_outbound_config_t {
 
 } odp_ipsec_outbound_config_t;
 
+/**
+ * IPSEC TEST capability
+ */
+typedef struct odp_ipsec_test_capability_t {
+	/** Support for updating internal parameters of the SA for test */
+	odp_support_t sa_update;
```
JannePeltonen, line 242: This looks redundant. It would seem to me that the individual parameter specific capabilities would be sufficient. If none of them is set, then update is not supported at all.

include/odp/api/spec/ipsec.h
```diff
@@ -192,6 +234,33 @@ typedef struct odp_ipsec_outbound_config_t {
...
+
+	/** Parameters supported for sa_update */
+	struct {
+		/** Inbound specific internal parameters */
+		struct {
+			/** Highest authenticated sequence number */
+			odp_bool_t antireplay_window_top;
```
JannePeltonen, line 249: I wonder if it is convenient to have this kind of array of booleans instead of e.g. returning an array of supported parameter "type" (or parameter ID) values. Maybe @psavol would like to comment.

If also the actual parameter struct for the update call were defined a bit differently, then the test functionality supported by this API could perhaps be extended by ODP implementation specific parameters without an API change.

include/odp/api/spec/ipsec.h
```diff
@@ -71,6 +71,48 @@ typedef enum odp_ipsec_op_mode_t {
...
+ * IPSEC TEST SA parameter value
+ */
+typedef union odp_ipsec_test_sa_param_val_t {
+	/** Next sequence number value */
+	uint64_t seq_num;
+
+	/** Highest authenticated sequence number */
+	uint64_t antireplay_window_top;
+
+} odp_ipsec_test_sa_param_val_t;
```
JannePeltonen, line 96: I wonder if the API would be more flexible if the above fields were both of the same "generic integer" type. This way most parameters would not need their own fields in this union. OTOH, once someone adds something like and IV or bytes to be used as random data or something like that, we would need an array of bytes. so new parameters may required new values in the union even if we combine the simple types now.

include/odp/api/spec/ipsec.h
```diff
@@ -1715,6 +1789,26 @@ int odp_ipsec_result(odp_ipsec_packet_result_t *result, odp_packet_t packet);
...
+ * IPSEC TEST API to set IPSEC SA parameter type field with new parameter value.
+ *
+ * This operation allows application to modify IPSEC SA parameters not otherwise
+ * exposed to the application. For IPSEC operations, fields like sequence number
+ * is not available to the application in the datapath. The API allows
+ * application to verify various features provided by the platform.
+ *
+ * It is the responsiblity of the application to make sure that no traffic is
+ * flowing through the SA when the API is called. Otherwise, this operation may
+ * lead to undefined behavior.
```
JannePeltonen, line 1802: How about using the following description here:

```
* IPsec test API for modifying internal state of an SA.
 *
 * This function is not meant to be used by normal applications but by special
 * test applications that test or debug the operation of the underlying ODP
 * implementation. Calling this function may degrade the performance of the
 * calling thread, other threads or the IPsec implementation in general.
 *
 * Calling this function for an SA at the same time when the SA is used for
 * processing traffic or when the SA is being modified through other parts
 * of IPsec API may result in undefined behaviour.
 *
 * SA state update through this function may not be supported by all ODP
 * implementations, ODP instances or SA instances or at every moment. This
 * function may fail for unspecified reasons even when the capability
 * call indicated support for updating a particular parameter and previous
 * similar calls succeeded.
 ```

File: test/validation/api/ipsec/ipsec.c [hunk 11]
```diff
@@ -854,6 +869,37 @@ static int ipsec_send_out_one(const ipsec_test_part *part,
...
 }
 
+void ipsec_test_sa_update_seq_num(odp_ipsec_sa_t sa, uint32_t seq_num)
+{
+	odp_ipsec_test_sa_param_t sa_param;
+
+	sa_param.type = ODP_IPSEC_TEST_SA_SEQ_NUM;
+	sa_param.val.seq_num = seq_num;
+
+	CU_ASSERT_EQUAL(0, odp_ipsec_test_sa_update(sa, sa_param));
```
JannePeltonen, line 879: If we allow this function to fail even when the capability call succeeds (so that the capa bit does not promise that update is possible for all SAs at all times), then it is not strictly correct to fail validation test if the call fails. Maybe we could live with the strict check now but it would be better to just notice if the update failed and then turn of the sequence number checking until the update call succeeds again.

include/odp/api/spec/ipsec.h
asasidharan: >If also the actual parameter struct for the update call were defined a bit differently,

I am not sure if I understood this correctly. Can you describe the layout you have in mind.

include/odp/api/spec/ipsec.h
asasidharan: > I wonder if the API would be more flexible if the above fields were both of the same "generic integer" type.

Did you mean something like:
union {
  uint64_t u64;
  uint8_t buffer[MAX_SIZE];
}

include/odp/api/spec/ipsec.h
asasidharan: Agreed. Updated v5 to use the term "set" instead of "override".

include/odp/api/spec/ipsec.h
asasidharan: Agreed. Updated v5 to use "instance" instead of "type".

include/odp/api/spec/ipsec.h
asasidharan: Agreed. Removed odp_support_t sa_update from capability with v5.

include/odp/api/spec/ipsec.h
asasidharan: Ack.

include/odp/api/spec/ipsec.h
psavol: Agree, this should indicate selection of what the operation does and what fields from the union (under) is used. Command, operation, select ...

test/validation/api/ipsec/ipsec.c
asasidharan: Agreed. Update the test to loop on the API call till either it succeeds or one minute timeout happens.

include/odp/api/spec/ipsec.h
JannePeltonen: After discussing with Petri, I think it is better to leave this as proposed, i.e. there is a separate union field for each possible parameter.

include/odp/api/spec/ipsec.h
JannePeltonen: Let's keep the capability fields as booleans, but let's also simplify the structure by moving all the fields to the top level. The inbound and outbound substructures do not seem to add any value and the field names have to be unique anyway because the parameter to be modified is specified by the same enum that must have unique corresponding values.

include/odp/api/spec/ipsec.h
```diff
@@ -1715,6 +1786,32 @@ int odp_ipsec_result(odp_ipsec_packet_result_t *result, odp_packet_t packet);
...
+ * call indicated support for updating a particular parameter and previous
+ * similar calls succeeded.
+ *
+ * @param          sa            IPSEC SA to be updated
+ * @param          sa_param      IPSEC SA param to update
+ *
+ * @return 0      On success
+ * @retval <0     On failure
+ */
+int odp_ipsec_test_sa_update(odp_ipsec_sa_t sa, odp_ipsec_test_sa_param_t sa_param);
```
JannePeltonen, line 1813: I think Petri's comment on this got somehow lost, so I am repeating it: The ID of the parameter that is modified could be a parameter of this function instead of contained in the sa_param parameter. Then odp_ipsec_sa_test_param_t could itself be the union of possible values.

test/validation/api/ipsec/ipsec.c
JannePeltonen: I think one minute timeout is excessive as it may repeat for multiple algorithms and then possible slow down CI a lot. Something like one second or less would be better if retrying made sense, but I do not think it does. If the call fails, it probably fails for a reason that does not go away by waiting it out (e.g. for this particular SA seqno cannot be set through this API).

My original point was that since the API says that the function may fail, it is not a violation of the API if the function fails and thus it is not strictly correct for the validation check to fail in that case. Ideally, the validation test would notice that the function failed and then make note of that so that it won't do the subsequent checks that check the outcome of the attempted test operation (that was not done in the first place because the call failed).

include/odp/api/spec/ipsec.h
```diff
@@ -1715,6 +1786,32 @@ int odp_ipsec_result(odp_ipsec_packet_result_t *result, odp_packet_t packet);
...
+ * implementation. Calling this function may degrade the performance of the
+ * calling thread, other threads or the IPSEC implementation in general.
+ *
+ * Calling this function for an SA at the same time when the SA is used for
+ * processing traffic or when the SA is being modified through other parts
+ * of IPSEC API may result in undefined behaviour.
+ *
+ * SA state update through this function may not be supported by all ODP
+ * implementations, ODP instances or SA instances or at every moment. This
+ * function may fail for unspecified reasons even when the capability
```
JannePeltonen, line 1803: After dicsussing with Petri, it would be good to replace "may fail" with "may return failure". The former may perhaps incorrectly imply that the function might fail in some fatal, unrecoverable way (like segfault).

asasidharan: Ack.

include/odp/api/spec/ipsec.h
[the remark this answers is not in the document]
asasidharan: Agreed. With V6, updated the API to have three parameters:
```
+int odp_ipsec_test_sa_update(odp_ipsec_sa_t sa,                                 
+                            odp_ipsec_test_sa_operation_t op,                   
+                            odp_ipsec_test_sa_param_t param); 
```

asasidharan: Agreed.

test/validation/api/ipsec/ipsec.c
asasidharan: Updated the test to skip further check related to seq_num when the API call fails. Also the test would print ">> skipped" for the test configuration which was skipped.

include/odp/api/spec/ipsec.h
asasidharan: Ack.

include/odp/api/spec/ipsec.h
asasidharan: Agreed. Made the capability structure to be flat in V6.

include/odp/api/spec/ipsec.h
```diff
@@ -1715,6 +1782,35 @@ int odp_ipsec_result(odp_ipsec_packet_result_t *result, odp_packet_t packet);
...
+ * @param          sa            IPSEC SA to be updated
+ * @param          op            Specifies operation to be performed
+ * @param          param         IPSEC SA param to be used for the operation
+ *
+ * @return 0      On success
+ * @retval <0     On failure
+ */
+int odp_ipsec_test_sa_update(odp_ipsec_sa_t sa,
+			     odp_ipsec_test_sa_operation_t op,
+			     odp_ipsec_test_sa_param_t param);
```
JannePeltonen, line 1812: I wonder if the param struct should be passed by reference rather than by value?

include/odp/api/spec/ipsec.h
```diff
@@ -192,6 +232,28 @@ typedef struct odp_ipsec_outbound_config_t {
...
+		 */
+		odp_bool_t seq_num;
+
+		/** Highest authenticated sequence number
+		 *
+		 * @see ODP_IPSEC_TEST_SA_UPDATE_ANTIREPLAY_WINDOW_TOP
+		 */
+		odp_bool_t antireplay_window_top;
+
+	} sa_update_params;
```
JannePeltonen, line 253: I think it would be better to name this sa_operations to match the name of the enum that got renamed odp_ipsec_test_sa_operation_t.

test/validation/api/ipsec/ipsec.c
```diff
@@ -351,6 +351,18 @@ int ipsec_check_esp_aes_gcm_256(void)
 				ODP_AUTH_ALG_AES_GCM, 0);
 }
 
+int ipsec_check_test_sa_update_seq_num(void)
```
JannePeltonen, line 354: This is a very minor thing, but it would be good to not place this function in the middle of the algorithm capability check functions but before or after them. It would keep the file slightly more easy to read as a whole.

test/validation/api/ipsec/ipsec.c
```diff
@@ -866,6 +878,38 @@ static int ipsec_send_out_one(const ipsec_test_part *part,
...
+	sa_op = ODP_IPSEC_TEST_SA_UPDATE_SEQ_NUM;
+	sa_param.seq_num = seq_num;
+
+	return odp_ipsec_test_sa_update(sa, sa_op, sa_param);
+}
+
+static void ipsec_pkt_seq_num_check(odp_packet_t pkt, uint32_t seq_num)
+{
+	uint32_t l3_off = odp_packet_l3_offset(pkt);
+	uint32_t l4_off = odp_packet_l4_offset(pkt);
```
JannePeltonen, line 895: This is incorrect. The value of l4 offset is undefined after outbound IPsec operation.

It would also be good to check whether the odp_packet_l3_offset() call succeeds.

test/validation/api/ipsec/ipsec.c
```diff
@@ -866,6 +878,38 @@ static int ipsec_send_out_one(const ipsec_test_part *part,
...
+	odph_ipv4hdr_t ip;
+
+	odp_packet_copy_to_mem(pkt, l3_off, sizeof(ip), &ip);
```
JannePeltonen, line 898: It would be good to check that the packet is an IPv4 packet. It would also be good to check the return value of the copy function.

test/validation/api/ipsec/ipsec.c
```diff
@@ -866,6 +878,38 @@ static int ipsec_send_out_one(const ipsec_test_part *part,
...
+
+	if (ip.proto == ODPH_IPPROTO_ESP) {
+		odph_esphdr_t esp;
+
+		odp_packet_copy_to_mem(pkt, l4_off, sizeof(esp), &esp);
+		CU_ASSERT_EQUAL(odp_be_to_cpu_32(esp.seq_no), seq_num);
+	} else {
```
JannePeltonen, line 905: It would be good to check that the packet indeed is AH at this point (and not e.g. UDP).

include/odp/api/spec/ipsec.h
asasidharan: Agreed. I will update this in v7.

include/odp/api/spec/ipsec.h
asasidharan: Agreed. Renamed to sa_operations in v7.

test/validation/api/ipsec/ipsec.c
asasidharan: Agreed. Moved the function after the algorithm check capability functions.

test/validation/api/ipsec/ipsec.c
asasidharan: Agreed. Updated the function to compute l4_off as l3_off + IPv4 Header length.

Also added check for  return value of odp_packet_l3_offset().

test/validation/api/ipsec/ipsec.c
asasidharan: Ack.

test/validation/api/ipsec/ipsec.c
asasidharan: Ack.

test/validation/api/ipsec/ipsec.c
```diff
@@ -734,6 +746,44 @@ static int ipsec_send_out_one(const ipsec_test_part *part,
...
+{
+	uint32_t l3_off = odp_packet_l3_offset(pkt);
+	uint32_t l4_off;
+	odph_ipv4hdr_t ip;
+
+	CU_ASSERT_NOT_EQUAL_FATAL(ODP_PACKET_OFFSET_INVALID, l3_off);
+	CU_ASSERT_EQUAL_FATAL(0, odp_packet_copy_to_mem(pkt, l3_off, sizeof(ip),
+							&ip));
+
+	if (ODPH_IPV4HDR_VER(ip.ver_ihl) == ODPH_IPV4) {
```
JannePeltonen, line 770: Now this function silently succeeds without checking the sequence number if IP version is not 4 or if IP protocol is not ESP or AH (e.g. IPsec with UDP encapsulation). In those case the function should fail (CU_FAIL() or similar).

asasidharan: Agreed. Fixed by invoking CU_FAIL() for unexpected IP version and IP proto values in v8.

JannePeltonen:
Sorry, one more change: the param parameter of odp_ipsec_test_sa_update() should be a const pointer:

```
int odp_ipsec_test_sa_update(odp_ipsec_sa_t sa,
                             odp_ipsec_test_sa_operation_t op,
                             const odp_ipsec_test_sa_param_t *param);
```

With that change,

asasidharan:
> 
> 
> Sorry, one more change: the param parameter of odp_ipsec_test_sa_update() should be a const pointer:
> 
> ```
> int odp_ipsec_test_sa_update(odp_ipsec_sa_t sa,
>                              odp_ipsec_test_sa_operation_t op,
>                              const odp_ipsec_test_sa_param_t *param);
> ```

Agreed. Updated v9 by making param a const pointer.

Also thank you for your valuable suggestions and comments which helped in shaping this PR. I have added Reviewed-by tag to the commits.

JannePeltonen (approved):

include/odp/api/spec/ipsec.h
```diff
@@ -71,6 +71,46 @@ typedef enum odp_ipsec_op_mode_t {
...
 } odp_ipsec_op_mode_t;
 
+/**
+ * IPSEC TEST SA operation
+ */
+typedef enum odp_ipsec_test_sa_operation_t {
+	/** Update next sequence number
+	 *
+	 * The seq_num parameter is an outbound SA specific parameter.
+	 * Invoking the odp_ipsec_sa_test_update() API to update this
```
psavol, line 81: typo: odp_ipsec_sa_test_update() -> ..._test_sa_...

include/odp/api/spec/ipsec.h
```diff
@@ -71,6 +71,46 @@ typedef enum odp_ipsec_op_mode_t {
...
+	 * field on an inbound SA will cause the API to return failure.
+	 */
+	ODP_IPSEC_TEST_SA_UPDATE_SEQ_NUM = 0,
+
+	/** Update highest authenticated sequence number
+	 *
+	 * The antireplay_window_top parameter is inbound SA specific.
+	 * Invoking the odp_ipsec_sa_test_update() API to update this
```
psavol, line 89: typo: odp_ipsec_sa_test_update() -> ..._test_sa_...

psavol:
API spec otherwise OK, but two typos in functions names. Add my reviewed-by to API patch for v10.

psavol:
Ping. This otherwise ready, but needs typo corrections and my reviewed-by to API patch.

asasidharan:
> 
> 
> Ping. This otherwise ready, but needs typo corrections and my reviewed-by to API patch.

@psavol  Thank you for pinging. I had missed to notice your previous comments until now.

With v10 fixed the typos and added  your reviewd-by tag to the API patch.

include/odp/api/spec/ipsec.h
asasidharan: Ack. Fixed with v10.

include/odp/api/spec/ipsec.h
asasidharan: Ack. Fixed with v10.

PR #1105, opened 2020-11-18
[PATCH v3] linux-gen: ipsec: discard replayed packets before lifetime update

JannePeltonen:
Discard packets that fail the anti-replay check before checking
SA lifetime and updating the packet and byte counters. Previously
such packets could have been used to prematurely age an SA, leading
to a possible denial of service.

### DIFF
--- a/platform/linux-generic/odp_ipsec.c
+++ b/platform/linux-generic/odp_ipsec.c
@@ -689,14 +689,14 @@ static ipsec_sa_t *ipsec_in_single(odp_packet_t pkt,
 		goto err;
 	}
 
-	if (_odp_ipsec_sa_stats_update(ipsec_sa,
-				       state.stats_length,
-				       status) < 0)
-		goto err;
-
 	if (_odp_ipsec_sa_replay_update(ipsec_sa,
 					state.in.seq_no,
 					status) < 0)
+		goto err;
+
+	if (_odp_ipsec_sa_stats_update(ipsec_sa,
+				       state.stats_length,
+				       status) < 0)
 		goto err;
 
 	state.ip = odp_packet_l3_ptr(pkt, NULL);


anoobj:
@JannePeltonen There are few other errors after SA lifetime is updated. Like the error in ipsec_in_esp_post(). Is it fine if those error packets consume SA lifetime?

JannePeltonen:
> There are few other errors after SA lifetime is updated. Like the error in ipsec_in_esp_post(). Is it fine if those error packets consume SA lifetime?

I my opinion that is the right thing to do. Limiting lifetime is about limiting the amount traffic for which the SA and the associated key material are used. Therefore it makes sense to take into account all packets that pass the authentication ant anti-replay checks even if there are some errors later. All those packets were sent by the original sender and consumed sequence numbers whereas the packets that failed the anti-replay check were just duplicates that did not come from the original sender.

anoobj:
> 
> 
> > There are few other errors after SA lifetime is updated. Like the error in ipsec_in_esp_post(). Is it fine if those error packets consume SA lifetime?
> 
> I my opinion that is the right thing to do. Limiting lifetime is about limiting the amount traffic for which the SA and the associated key material are used. Therefore it makes sense to take into account all packets that pass the authentication ant anti-replay checks even if there are some errors later. All those packets were sent by the original sender and consumed sequence numbers whereas the packets that failed the anti-replay check were just duplicates that did not come from the original sender.

Agreed. I think the error from ipsec_in_esp_post() is from ESP trailer checks. So agreed that these should be included in SA life calculations.

JannePeltonen:
v2: Added Reviewed-by tag.

anoobj:
Patch looks good to me.

Reviewed-by : Anoob Joseph<anoobj@marvell.com>

JannePeltonen:
v3: Rebased on latest master

anoobj (approved):

PR #1106, opened 2020-11-18
[PATCH v3] example: sysinfo: use new sysinfo APIs

psavol:
Use the new odp_sys_config_print() and odp_system_info()
calls.

### DIFF
--- a/example/sysinfo/odp_sysinfo.c
+++ b/example/sysinfo/odp_sysinfo.c
@@ -25,6 +25,90 @@ static const char *support_level(odp_support_t support)
 	}
 }
 
+static const char *cpu_arch_name(odp_system_info_t *sysinfo)
+{
+	odp_cpu_arch_t cpu_arch = sysinfo->cpu_arch;
+
+	switch (cpu_arch) {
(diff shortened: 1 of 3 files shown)


JereLeppanen:
To avoid every name appearing two times in a kind of error-prone fashion, could use something like

`#define CASE_CPU_ARCH(NAME) case ODP_CPU_ARCH_ ## NAME: return #NAME`

Either way though, the patch looks OK to me.

psavol:
v2: rebased and updated implementation. Now also CPU HW ISA version is printed. On ARMv8 devices ISA version is filled in if part number is recognized. Otherwise HW ISA version is unknown.

psavol:
> To avoid every name appearing two times in a kind of error-prone fashion, could use something like
> 
> `#define CASE_CPU_ARCH(NAME) case ODP_CPU_ARCH_ ## NAME: return #NAME`

I changed to print commonly used CPU arch and ISA names instead of enum names: e.g. ARMv8.1 instead of ARMV8_1.

matiaselo (approved):

PR #1109, opened 2020-11-21
[PATCH v7] linux-gen: rename some global symbols in the static library

JereLeppanen:
Rename some symbols that are global in the static library, in order to
avoid clashes with application code. The prefix "`_odp_`" is added to
the existing names.

This patch renames most of the global symbols, a few more still remain
to be renamed.

v2: Rebase.

v3: Fix checkpatch warnings.

v4: Squash.

v5: Rebase and remove conflicting renames.

v6: Use prefix "`_odp_`" instead of "`__odp_`".

v7: Rebase.


JereLeppanen:
> This patch renames most of the global symbols, a few more still remain
> to be renamed.

This is in order to avoid conflicts with the currently open PRs.

psavol (approved):

PR #1122, opened 2020-12-01
[PATCH v2] linux-gen: ipsec: report ICV errors as auth error

anoobj:
ICV errors need to be reported as auth errors and not alg errors. All
other crypto errors can be treated as alg errors.


JannePeltonen (approved):

PR #1123, opened 2020-12-02
[PATCH v2] linux-gen: ipsec: report all outb crypto failures as alg errors

anoobj:
All outbound crypto failures can be treated as alg errors. ICV errors
won't happen in outbound path.

### DIFF
--- a/platform/linux-generic/odp_ipsec.c
+++ b/platform/linux-generic/odp_ipsec.c
@@ -1521,18 +1521,7 @@ static ipsec_sa_t *ipsec_out_single(odp_packet_t pkt,
 	}
 
 	if (!crypto.ok) {
-		if ((crypto.cipher_status.alg_err !=
-		     ODP_CRYPTO_ALG_ERR_NONE) ||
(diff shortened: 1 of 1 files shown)


platform/linux-generic/odp_ipsec.c
```diff
@@ -1526,18 +1526,7 @@ static ipsec_sa_t *ipsec_out_single(odp_packet_t pkt,
 	}
 
 	if (!crypto.ok) {
```
anoobj, line 1528: @JannePeltonen Should we make line#1522 as,

`if ((rc < 0) || (!crypto.ok))`

and remove the separate code block line#1528-line#1531?

Only difference we have is the ODP_DBG print.

JannePeltonen: No, because crypto.ok check cannot be moved above the odp_crypto_result() call.

JannePeltonen:

Perhaps you could also rebase this when adding the tag even though there are no conflicts now.

JannePeltonen (approved):

PR #1125, opened 2020-12-02
[PATCH v2] linux-gen: ipsec: fix the amount of ESP padding in AES-GCM and -GMAC

JannePeltonen:
AES-GCM and AES-GMAC do not require additional padding on top of
what is required by ESP. Reduce padding to the minimum required by
ESP from the current padding of up to 15 bytes.


JereLeppanen (approved):
Looks correct, as far as I can tell.

PR #1126, opened 2020-12-03
[PATCH v5] test: sched_pktio: cleanup prefetched events before worker exit

nithind1988:
Cleanup prefetched events before worker exits. Also break before
schedule multi is called if exit is triggered.

### DIFF
--- a/test/performance/odp_sched_pktio.c
+++ b/test/performance/odp_sched_pktio.c
@@ -203,9 +203,6 @@ static int worker_thread_direct(void *arg)
 		odp_event_t ev[burst_size];
 		odp_packet_t pkt[burst_size];
 
-		num_pkt = odp_schedule_multi(&queue, ODP_SCHED_NO_WAIT,
-					     ev, burst_size);
-
 		polls++;
 
 		if (polls == CHECK_PERIOD) {
@@ -213,6 +210,9 @@ static int worker_thread_direct(void *arg)
 			if (test_global->stop_workers)
 				break;
 		}
+
+		num_pkt = odp_schedule_multi(&queue, ODP_SCHED_NO_WAIT,
+					     ev, burst_size);
 
 		if (num_pkt <= 0)
 			continue;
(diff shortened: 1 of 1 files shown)


test/performance/odp_sched_pktio.c
```diff
@@ -238,6 +238,26 @@ static int worker_thread_direct(void *arg)
...
+		if (ev == ODP_EVENT_INVALID)
+			break;
+		odp_event_free(ev);
+	}
+
+	/* Flush all writes */
+	odp_mb_full();
+
+	/* Restore scheduling state */
+	odp_schedule_resume();
```
JannePeltonen, line 259: So after this odp_schedule_resume() call there can again be local prefetched events and we are back to square one? I do not see how this patch fixes something, other than by chance through altering timing.

Maybe there should be thread barrier (like in l2fwd) to make sure everybody stops sending packets before anybody resumes scheduling again for the final flush (if it is needed anymore then). Or something like that.

nithind1988: Removed odp_schedule_resume() in V2. Let me know if it makes sense given that main thread is cleaning up all the other events in scheduler so I believe odp_l2fwd like per thread cleanup is not needed.

test/performance/odp_sched_pktio.c
```diff
@@ -238,6 +238,24 @@ static int worker_thread_direct(void *arg)
 			test_global->worker_stat[worker_id].rx_pkt += num_pkt;
 	}
 
+	/*
+	 * Free prefetched packets before exiting worker thread as
+	 * such packets can block main thread cleanup cause buffer leak.
```
JannePeltonen, line 243: There is a grammar issue in the sentence, making it hard to understand.

test/performance/odp_sched_pktio.c
```diff
@@ -238,6 +238,24 @@ static int worker_thread_direct(void *arg)
...
+		odp_event_t  ev;
+
+		ev = odp_schedule(NULL, ODP_SCHED_NO_WAIT);
+		if (ev == ODP_EVENT_INVALID)
+			break;
+		odp_event_free(ev);
+	}
+
+	/* Flush all writes */
+	odp_mb_full();
```
JannePeltonen, line 256: I wonder if and why this is needed. I would have thought thread exit followed by odph_odpthreads_join() ensures the needed memory synchronization between the threads.

test/performance/odp_sched_pktio.c
JannePeltonen: After the removal of the schedule_resume() this looks ok to me, apart from the minor unrelated comments I made.

JannePeltonen:
The commit message could say that this change is only for the case when pipelining or timeouts are not used. I suppose the pipeline and timeout variants of the worker function have the same problems.

test/performance/odp_sched_pktio.c
nithind1988: Ack, will fix in V2.

test/performance/odp_sched_pktio.c
nithind1988: Ack, Will remove this in V2. Join should be doing it.

JannePeltonen (approved):

matiaselo:
@nithind1988 Please add Janne's reviewed-by tag, so we can merge this PR.

PR #1127, opened 2020-12-03
[PATCH v5] api: ipsec: add print APIs

asasidharan:
Add API to print implementation defined information about specified SA.
When SA specified is invalid, implementation may choose to print global
IPSEC configuration and other related information.

Signed-off-by: Sachin Yaligar <syaligar@marvell.com>

### DIFF
--- a/include/odp/api/spec/ipsec.h
+++ b/include/odp/api/spec/ipsec.h
@@ -1627,6 +1627,23 @@ int odp_ipsec_sa_mtu_update(odp_ipsec_sa_t sa, uint32_t mtu);
 void *odp_ipsec_sa_context(odp_ipsec_sa_t sa);
 
 /**
+ * Print global IPSEC configuration info
+ *
+ * Print implementation-defined information about the global IPSEC
+ * configuration.
+ */
+void odp_ipsec_print(void);
+
+/**
+ * Print IPSEC SA info
+ *
+ * @param sa      SA handle
+ *
+ * Print implementation-defined IPSEC SA debug information to the ODP log.
+ */
+void odp_ipsec_sa_print(odp_ipsec_sa_t sa);
+
+/**
  * @}
  */
 
--- a/test/validation/api/ipsec/ipsec_test_in.c
+++ b/test/validation/api/ipsec/ipsec_test_in.c
@@ -1597,6 +1597,31 @@ static void test_in_ipv6_esp_udp_null_sha256_lookup(void)
 	ipsec_sa_destroy(sa);
 }
 
+static void test_ipsec_print(void)
+{
+	odp_ipsec_print();
+}
+
+static void test_ipsec_sa_print(void)
+{
+	odp_ipsec_sa_param_t param_in;
+	odp_ipsec_sa_t in_sa;
+
+	ipsec_sa_param_fill(&param_in,
+			    true, false, 123, NULL,
+			    ODP_CIPHER_ALG_AES_CBC, &key_a5_128,
+			    ODP_AUTH_ALG_SHA1_HMAC, &key_5a_160,
+			    NULL, NULL);
+
+	in_sa = odp_ipsec_sa_create(&param_in);
+
+	CU_ASSERT_NOT_EQUAL_FATAL(ODP_IPSEC_SA_INVALID, in_sa);
+
+	odp_ipsec_sa_print(in_sa);
+
+	ipsec_sa_destroy(in_sa);
+}
+
 static void ipsec_test_capability(void)
 {
 	odp_ipsec_capability_t capa;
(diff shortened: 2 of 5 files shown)


include/odp/api/spec/ipsec.h
```diff
@@ -1626,6 +1627,18 @@ int odp_ipsec_sa_mtu_update(odp_ipsec_sa_t sa, uint32_t mtu);
...
+ * Print IPSEC SA info
+ *
+ * @param sa      SA handle
+ * @param fp      Pointer to a file for output
+ *
+ * Print implementation-defined IPSEC SA debug information to the log file.
+ * The implementation may choose to print global IPSEC configuration when the
+ * SA handle passed is ODP_IPSEC_SA_INVALID.
+ */
+void odp_ipsec_sa_print(odp_ipsec_sa_t sa, FILE *fp);
```
psavol, line 1640: Other similar print functions of ODP API do not include FILE parameter. It should be remove also here and just state that function prints into ODP log (global init has log print function as parameter).

Also remove optional behavior of ODP_IPSEC_SA_INVALID. If IPSEC global config, etc debug information is needed, then it is much more clear to introduce another debug function for that purpose. So, this should just print SA specific info.

include/odp/api/spec/ipsec.h
```diff
@@ -18,6 +18,7 @@
 extern "C" {
 #endif
 
+#include <stdio.h>
```
psavol, line 21: This is not needed when FILE is not used.

include/odp/api/spec/ipsec.h
asasidharan: Agreed. In V2, removed the FILE parameter and introduced a separate odp_ipsec_print() API to print the global config information.

include/odp/api/spec/ipsec.h
asasidharan: Ack. Updated in V2.

anoobj:
@psavol @JannePeltonen Can this PR be merged? Do you have any further comments on this?

JannePeltonen:
> Can this PR be merged? Do you have any further comments on this?

I do not have any comments. But I think before this can be merged, API validation tests need to check that these functions actually exist and can be called and a minimal implementation in linux-gen to make the validation test pass.

anoobj:
> 
> > Can this PR be merged? Do you have any further comments on this?
> 
> I do not have any comments. But I think before this can be merged, API validation tests need to check that these functions actually exist and can be called and a minimal implementation in linux-gen to make the validation test pass.

Can you take that up? Because we may not know what makes sense for linux-gen implementation. Or we can add a dummy implementation that can be improved later on.

JannePeltonen:
> Or we can add a dummy implementation that can be improved later on.

A dummy implementation that does not print anything should be fine.

asasidharan:
Pushed V3 with a minimal linux-generic implementation and validation tests for the print APIs.

Once the stats PR is merged, we will remove this specific test case for sa print and plug into the framework.

test/validation/api/ipsec/ipsec_test_in.c
```diff
@@ -1597,6 +1597,42 @@ static void test_in_ipv6_esp_udp_null_sha256_lookup(void)
...
+
+	param_in.inbound.antireplay_ws = window_sz;
+
+	in_sa = odp_ipsec_sa_create(&param_in);
+
+	CU_ASSERT_NOT_EQUAL_FATAL(ODP_IPSEC_SA_INVALID, in_sa);
+
+	odp_ipsec_sa_print(in_sa);
+
+	odp_ipsec_sa_destroy(in_sa);
```
psavol, line 1633: odp_ipsec_sa_disable(sa) should be called before sa_destroy(). May be missing this call caused lots of CI jobs to fail.

test/validation/api/ipsec/ipsec_test_in.c
```diff
@@ -1597,6 +1597,42 @@ static void test_in_ipv6_esp_udp_null_sha256_lookup(void)
...
+	CU_ASSERT(odp_ipsec_capability(&capa) == 0);
+
+	memset(&in_tunnel, 0, sizeof(odp_ipsec_tunnel_param_t));
+
+	odp_ipsec_print();
+
+	ipsec_sa_param_fill(&param_in,
+			    true, false, 0x1234, &in_tunnel,
+			    ODP_CIPHER_ALG_AES_GCM, &key_mcgrew_gcm_3,
+			    ODP_AUTH_ALG_AES_GCM, NULL,
```
psavol, line 1622: A commonly supported algorithm should be selected, so that this test case will likely run. Would e.g. ODP_CIPHER_ALG_AES_CBC be more likely supported than GCM (not sure if there's a difference)?

JannePeltonen: Every IPsec implementation must support aes-cbc-128 and hmac-sha1-96, so using those might be slightly better, although aes-gcm is very common too.

test/validation/api/ipsec/ipsec_test_in.c
asasidharan: Thank you for pointing this out. In V4, I replaced the direct odp_ipsec_sa_destroy() call with ipsec_sa_destroy() function which takes care of the API sequence.

test/validation/api/ipsec/ipsec_test_in.c
asasidharan: In V4, I have updated SA print test function to use aes-cbc-128 cipher and sha1 auth.

Also I have cleaned up the test function.

psavol (approved):

psavol:
Looks good. Please send v5 with my review tag added:
```
```

asasidharan:
Thank you for your reviews @psavol and @JannePeltonen. 
Added Petri's "Reviewed-by" tag in v5.

PR #1129, opened 2020-12-03
[PATCH v3] IPsec capability fixes

JannePeltonen:
- Add IPsec capability printing in sysinfo example
- Remove unsupported algorithms from capabilities. After this validation test no longer complains about "Wrong auth capabilities" for AES-CCM.

The macro trick in the latter patch does not pass checkpatch but it does not seem easy to fix without making the code worse.

### DIFF
--- a/example/sysinfo/odp_sysinfo.c
+++ b/example/sysinfo/odp_sysinfo.c
@@ -323,12 +323,14 @@ int main(void)
 	odp_queue_capability_t queue_capa;
 	odp_timer_capability_t timer_capa;
 	odp_crypto_capability_t crypto_capa;
+	odp_ipsec_capability_t ipsec_capa;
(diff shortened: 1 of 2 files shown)


example/sysinfo/odp_sysinfo.c
```diff
@@ -407,6 +409,10 @@ int main(void)
 	if (crypto_ret < 0)
 		printf("crypto capability failed\n");
 
+	ipsec_ret = odp_ipsec_capability(&ipsec_capa);
+	if (crypto_ret < 0)
```
JereLeppanen, line 413: ipsec_ret, not crypto_ret.

JannePeltonen: Fixed in v2. Also removed a few unnecessary empty lines.

JereLeppanen (approved):

PR #1130, opened 2020-12-04
[PATCH v2] Add 64-bit RISC-V build test to CI

matiaselo:


JereLeppanen:
Looks OK. So the reason to not call the generic build.sh is to avoid configure --enable-dpdk?

matiaselo:
> Looks OK. So the reason to not call the generic build.sh is to avoid configure --enable-dpdk?

Yep, that was the reason.

JereLeppanen (approved):

PR #1132, opened 2020-12-04
validation: atomic: add synchronization to worker threads

Malvika-Gupta:
Currently in the atomic test cases, the worker threads are very likely to
run sequentially rather than parallelly since the tests are small and
no synchronization barriers are used. To enable the worker threads to
simultaneously access the shared variable, an odp_atomic_u32_t variable is
introduced. Once all the threads have been created, this variable is set
to 1 to allow threads concurrent access to the shared variable. When all
the threads exit, the variable is reset to 0. This code has been largely
inspired from DPDK.


JereLeppanen:
Wouldn't it be more natural to use odp_barrier rather than atomics?

JannePeltonen:
I think CNT should be changed to something like 10000 or greater if the purpose of this PR is to make it more likely that a non-atomic implementation of an atomic operation will be caught. Synchronizing the test threads seems somewhat futile if the actual test loop is run only few times.

Malvika-Gupta:
> Wouldn't it be more natural to use odp_barrier rather than atomics?

Hi @JereLeppanen , thank you for reviewing the PR. It had not occurred to me to use odp_barrier_wait; I saw the use of an u32 atomic variable to be simple enough for illustrating concurrent access by the threads. If you prefer I use odp_barrier in place of atomics, I can explore that option in v2. Thanks!

Malvika-Gupta:
> I think CNT should be changed to something like 10000 or greater if the purpose of this PR is to make it more likely that a non-atomic implementation of an atomic operation will be caught. Synchronizing the test threads seems somewhat futile if the actual test loop is run only few times.

Hi @JannePeltonen , thank you for reviewing the PR. I will modify CNT to 10000 in v2. Thanks!

JereLeppanen:
> If you prefer I use odp_barrier in place of atomics, I can explore that option in v2.

Please do, unless someone holds the opposite view.

Malvika-Gupta:
Hi @JereLeppanen and @JannePeltonen , I have updated the PR with your suggestions. Please review
Thank you!

JereLeppanen:
In the commit comment, "UINT_VAL_32" should actually be "U32_INIT_VAL". Other than that, this looks OK to me.

Malvika-Gupta:
> In the commit comment, "UINT_VAL_32" should actually be "U32_INIT_VAL". Other than that, this looks OK to me.

Thanks for reviewing @JereLeppanen . I have fixed it in v5.

JereLeppanen (approved):

psavol:
Atomic validation test fails in Shippable (runs ThunderX):

Test failed at line  539  in file  atomic.c :  odp_atomic_load_u32(&global_mem->a32u_max) > odp_atomic_load_u32(&global_mem->a32u_min)

psavol:
I think this test cannot work correctly:
static void test_atomic_max_min_32(void)
{
	test_atomic_max_32();
	test_atomic_min_32();
}

When multiple threads run it, some increment the same u32 word (in max_32()) while others decrement it (in min_32()). When min/max operations are executed in between, it's pretty random which value is used for each min/max call.

The test case should be splitted into separate min and max test cases.

JereLeppanen:
> The test case should be splitted into separate min and max test cases.

@Malvika-Gupta Do you want to do this, or should I? Either way, that fix should go in before the patch in this PR.

JereLeppanen:
So I said I would open another PR, but it turned out to be easiest to just modify Malvika's patch in this PR.

Even with CNT = 10000 the duration of individual tests were only a few milliseconds, so I set it to 100000. This brings the total time for this test up to about 8 seconds in my environment.

@psavol Please review, I just switched roles from a reviewer to co-author.

Malvika-Gupta:
Hi @JereLeppanen , I will also test this change in my Arm machines. I will let you know if I run into any issues.

JereLeppanen:
> I will also test this change in my Arm machines. I will let you know if I run into any issues.

Great, thanks Malvika.

Malvika-Gupta:
Hi Jere,

I ran the test case on my Arm ThunderX2 machine running with default number of threads (14). It took significantly longer to finish; elapsed time shows between 220-250 seconds. Reducing the number of threads to 6 or 8 still takes about 40 seconds.
However, on an x86 machine, I am seeing similar to what you have reported from your build environment. 

Also, @m-govind and I wanted to understand the scope of the change. Why the threads are being re-synchronized via odp_barrier_wait before each u32/u64 test case of the same type?

Thank you!

JereLeppanen:
> elapsed time shows between 220-250 seconds.

Ten times more than I expected. So it's probably best to just drop CNT back to 10000.

> Why the threads are being re-synchronized via odp_barrier_wait before each u32/u64 test case of the same type?

This is not actually necessary in all cases, I just wanted to keep the identical functions identical. Maybe this is what's confusing you.

JereLeppanen:
> > elapsed time shows between 220-250 seconds.
> Ten times more than I expected. So it's probably best to just drop CNT back to 10000.

Except if you mean elapsed time as reported by cunit in the test log, that's actually total cpu time for all threads, so 250 seconds with 14 threads would be about 250 / 14 = 18 seconds, which doesn't sound too bad. 40 seconds with 6 threads would be 40 / 6 = 7 seconds, which is probably tolerable.

JereLeppanen:
Updated the patch. I put the barriers in the bottom level test functions, and set CNT to 50000 so that the test will hopefully complete in under 10 seconds on arm.

Malvika-Gupta:
> Except if you mean elapsed time as reported by cunit in the test log, that's actually total cpu time for all threads, so 250 seconds with 14 threads would be about 250 / 14 = 18 seconds, which doesn't sound too bad. 40 seconds with 6 threads would be 40 / 6 = 7 seconds, which is probably tolerable.
I was referring to the elapsed time as reported by Cunit test log. But I agree with you that its probably better to reduce the number of iterations from 100000 to a number that may give better performance on Arm.

Malvika-Gupta:
> Updated the patch. I put the barriers in the bottom level test functions, and set CNT to 50000 so that the test will hopefully complete in under 10 seconds on arm.

Thanks Jere. I will test the new changes on Arm and let you know how it performs

Malvika-Gupta:
> Updated the patch. I put the barriers in the bottom level test functions, and set CNT to 50000 so that the test will hopefully complete in under 10 seconds on arm.

Jere, I have run the changes on Arm ThunderX2 with both 10000 and 50000 iterations for better comparison. The results are below:
**10000**
- With default 14 threads, the elapsed time as reported by Cunit test log is 27 seconds
- With 6-8 threads, the elapsed time becomes 4 seconds

**50000:** 
- With default 14 threads, the elapsed time as reported by Cunit test log is 120-130 seconds
- With 6-8 threads, the elapsed time becomes 10-20 seconds

JereLeppanen:
@Malvika-Gupta As expected. So do you think 50000 is OK or too heavy?

Malvika-Gupta:
> @Malvika-Gupta As expected. So do you think 50000 is OK or too heavy?

Jere, I think 50000 should be okay for Arm. We can go ahead with that. Thanks!

matiaselo (approved):

JereLeppanen:
v8: Rebase and add Matias' reviewed-by tag.

PR #1134, opened 2020-12-07
[PATCH v3] linux-gen: pktio: support transmitting packet vectors using odp_queue_enq()

matiaselo:
V2:
- Increase out_discards statistics counter when packets are dropped (Jere)

### DIFF
--- a/platform/linux-generic/odp_packet_io.c
+++ b/platform/linux-generic/odp_packet_io.c
@@ -320,6 +320,7 @@ static odp_pktio_t setup_pktio_entry(const char *name, odp_pool_t pool,
 	pktio_entry->s.handle = hdl;
 	pktio_entry->s.pktin_frame_offset = pktin_frame_offset;
 	odp_atomic_init_u64(&pktio_entry->s.stats_extra.in_discards, 0);
+	odp_atomic_init_u64(&pktio_entry->s.stats_extra.out_discards, 0);
 
 	/* Tx timestamping is disabled by default */
 	pktio_entry->s.enabled.tx_ts = 0;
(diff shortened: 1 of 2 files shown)


platform/linux-generic/odp_packet_io.c
```diff
@@ -932,34 +932,86 @@ static inline int pktin_recv_buf(pktio_entry_t *entry, int pktin_index,
...
+	num = odp_packet_vector_tbl(pktv, &pkt_tbl);
+	ODP_ASSERT(num > 0);
+	sent = odp_pktout_send(queue_fn->get_pktout(queue), pkt_tbl, num);
+
+	/* Return success if any packets were sent. Free the possible remaining
+	   packets in the vector. */
+	if (odp_unlikely(sent <= 0))
+		return -1;
+	else if (odp_unlikely(sent != num))
+		odp_packet_free_multi(&pkt_tbl[sent], num - sent);
```
JereLeppanen, line 950: In this case, aren't we dropping packets without any errors or counters visible to the application?

matiaselo: When using odp_queue_enq() to send packets there may be intermediate steps (e.g. internal reorder queue) before the packets are actually transmitted to the wire, so trusting to the function return value is alone is difficult. The correct error counters should be incremented on the pktio driver level. I'll go through the drivers and add counters where necessary.

matiaselo:
@JereLeppanen is V2 looking acceptable to you?

JereLeppanen:
@matiaselo Yes, this looks good, approving.

JereLeppanen (approved):

PR #1135, opened 2020-12-07
[PATCH v3] Fix warnings reported by static code checkers

matiaselo:
V2:
* Detect `print_packet()` failure (Jere)

### DIFF
--- a/example/packet/odp_packet_dump.c
+++ b/example/packet/odp_packet_dump.c
@@ -492,7 +492,10 @@ static int print_packet(test_global_t *global, odp_packet_t pkt,
 	nsec  = nsec - (sec * ODP_TIME_SEC_IN_NS);
 	pktio = odp_packet_input(pkt);
 
-	odp_pktio_info(pktio, &pktio_info);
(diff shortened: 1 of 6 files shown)


example/packet/odp_packet_dump.c
```diff
@@ -492,7 +492,10 @@ static int print_packet(test_global_t *global, odp_packet_t pkt,
 	nsec  = nsec - (sec * ODP_TIME_SEC_IN_NS);
 	pktio = odp_packet_input(pkt);
 
-	odp_pktio_info(pktio, &pktio_info);
+	if (odp_pktio_info(pktio, &pktio_info)) {
+		printf("Error: odp_pktio_info() failed\n");
+		return -1;
```
JereLeppanen, line 497: The caller is treating the return value as a boolean for "the packet was printed", so here we should probably return 0.

matiaselo: I'll add a new check to `receive_packets(()`, so errors are detected properly.

JereLeppanen (approved):

PR #1136, opened 2020-12-08
[PATCH v2] validation: ipsec: fix termination when inline processing is not supported

matiaselo:
When inline IPsec processing is not supported, suite_context.pktio is set
to ODP_PKTIO_INVALID by ipsec_suite_init(). Use odp_pktio_lookup() to get
the loop pktio device handle during test termination to properly close the
device.


JereLeppanen (approved):
Usage of suite_context.pktio could probably be cleaner overall, but this patch is fine.

matiaselo:
> Usage of suite_context.pktio could probably be cleaner overall, but this patch is fine.

Agreed, `suite_context.pktio` usage could be cleaned up in future PR. This minimal fix is required to enable merging latest changes to the ODP-DPDK repo.

PR #1137, opened 2020-12-08
[PATCH v5] api: ipsec: allow L2 hdr to be provided in packet for inline outbound operation

anoobj:


include/odp/api/spec/ipsec.h
```diff
@@ -272,6 +272,17 @@ typedef struct odp_ipsec_capability_t {
...
+	/** Insert outer header as part of inline outbound operation.
+	 *
+	 * * 0: Application is required to update packet with outer headers
+	 *      such that packet is ready for transmission after successful
+	 *      IPSEC transformation. IPSEC transformation would update
+	 *      fields starting from L3 offset.
+	 * * 1: Application would provide outer header as part of params to
+	 *      odp_ipsec_out_inline() API. (default)
+	 */
+	odp_bool_t outer_hdr_insert;
```
nithind1988, line 284: Instead of even having capability, changing the semantics  always for odp_ipsec_out_inline() make sure l2 data is intact in packet, is better for applications as otherwise, applications need to check for this capability (atleast upstream examples) which impacts fast path.

anoobj: L2 data immediately preceding L3 may not be true for all platform implementations. For example, for purely s/w implementations, the IPsec tunnel header additions etc would mean packet would grow into headroom, anticipating which would be very cumbersome for the application.

I had described this in the commit description of the API change also. Please take a look at it.

JannePeltonen:
I think we should retain the current L2 header prepending semantics in all cases, at least as an option. The reason for the API not requiring L2 headers be in the packet in front of the L3 data is that typically L2 headers are not part of the original packet (that may be locally originated or forwarded) but come from somewhere else (e.g. forwarding cache, ARP table), in which case it can be more convenient and possibly faster not to require the application to copy L2 headers in the packet at some point.

Now, if we want to allow L2 headers to be already in the packet as a performance optimization for some platforms (how much does it speed things up?), then would the following API change cut it?

```
@@ -1189,7 +1189,12 @@ typedef struct odp_ipsec_out_inline_param_t {
        struct {
                /** Points to first byte of outer headers to be copied in
                 *  front of the outgoing IPSEC packet. Implementation copies
-                *  the headers during odp_ipsec_out_inline() call. */
+                *  the headers during odp_ipsec_out_inline() call.
+                *
+                *  Null value indicates that the outer headers of len bytes
+                *  are in the packet data, starting at L2 offset. In this
+                *  case L2 offset must be set.
+                */
                const uint8_t *ptr;
```

I am not sure that having ptr set to null needs to be defined as capability. Maybe it is simpler to just require all implementations to support it.

Perhaps we should also make it more clear in the API that the "outer headers" in this context basically mean L2 headers (or possibly L2 + some tunneling headers), excluding the outer IP header produced by an IPsec tunnel transformation. But this is a separate issue from what this PR is about.

anoobj:
> I am not sure that having ptr set to null needs to be defined as capability. Maybe it is simpler to just require all implementations to support it.

Unfortunately, not all implementations can support this model. For example, for purely s/w implementations, it is possible that the IPsec additions would grow into original L2 space. 

Implementation #1
..............................[L2][L3 - payload]
..............................[L2][IPsec tunnel][encrypted payload]

Implementation #2 
..............................[L2][L3 - payload]
...[L2][IPsec tunnel][encrypted payload]

For implementation #2, application cannot predict where eventual L2 is going to be.

Or are you fine with implementations, such as #2, to copy L2 to desired location and remove handling of L2 in ODP implementation altogether? In that case, we don't need a capability.

JannePeltonen:
> Unfortunately, not all implementations can support this model.

All implementations can support all models since this just a question about parameter passing to a SW function. Efficiency may vary.

> Or are you fine with implementations, such as 2, to copy L2 to desired location

Yes. If the L2 header is already in the packet, then someone has to copy it away anyway.

> and remove handling of L2 in ODP implementation altogether?

I am not sure what this refers to.

nithind1988:
> I think we should retain the current L2 header prepending semantics in all cases, at least as an option. The reason for the API not requiring L2 headers be in the packet in front of the L3 data is that typically L2 headers are not part of the original packet (that may be locally originated or forwarded) but come from somewhere else (e.g. forwarding cache, ARP table), in which case it can be more convenient and possibly faster not to require the application to copy L2 headers in the packet at some point.

Agreed, having it as an option is fine with NULL indicating that it is already part of packet. But if checks are still costly.

Just want to add that, since pointer can be from any memory(not necessarily from ODP memory), I think platform cannot offload it to HW for copying due to DMA mapping constraints. So they might endup copying in the implementation.

Additionally, application might be having const l2 lengths, which helps in their copy code getting optimized out by compiler 
vs, a general len parameter passed to ODP which cannot be optimized that much.

anoobj:
@JannePeltonen @nithind1988 Agreed. I'll update the PR with the text change the Janne has suggested.

anoobj:
Changed PR to just update text to allow for the proposed case.

JannePeltonen:
This is ok for me. But would this slight rewording of the text make it better:

Null value indicates that the outer headers are in the packet data, starting at L2 offset and ending at the byte before L3 offset.  In this case L2 offset must be set be set and less than L3 offset and field 'len' will be ignored.

anoobj:
> This is ok for me. But would this slight rewording of the text make it better:
> 
> Null value indicates that the outer headers are in the packet data, starting at L2 offset and ending at the byte before L3 offset. In this case L2 offset must be set be set and less than L3 offset and field 'len' will be ignored.

Agreed. Made this change as part of v3.

JannePeltonen:
Are you planning to make the necessary linux-gen update for this to make this ok for merging or should we do it?

It would also be good to have a validation test case for this, but maybe that can be added later and not in this PR.

anoobj:
> Are you planning to make the necessary linux-gen update for this to make this ok for merging or should we do it?

You can take that up. No prob from my side. Should I add your reviewed-by to this patch? Hope we can mark this as "Spec OK" then.

> It would also be good to have a validation test case for this, but maybe that can be added later and not in this PR.

Same thoughts.

JannePeltonen:
> You can take that up

ok.

> Should I add your reviewed-by to this patch? 

Yes, you can add it.

> Hope we can mark this as "Spec OK" then.

I asked if Petri wants to take a look, so let's wait with that a little bit.

include/odp/api/spec/ipsec.h
```diff
@@ -1189,7 +1189,13 @@ typedef struct odp_ipsec_out_inline_param_t {
 	struct {
 		/** Points to first byte of outer headers to be copied in
 		 *  front of the outgoing IPSEC packet. Implementation copies
-		 *  the headers during odp_ipsec_out_inline() call. */
+		 *  the headers during odp_ipsec_out_inline() call.
+		 *
+		 *  Null value indicates that the outer headers are in the
+		 *  packet data, starting at L2 offset and ending at the byte
+		 *  before L3 offset. In this case L2 offset must be set and
+		 *  less than L3 offset and field 'len' will be ignored.
```
psavol, line 1197: After thinking this again with Janne, it could be better to require that application sets len also when ptr ==NULL. This way we have backwards compatible option for "no outer header defined" case (ptr==NULL && len==0) and also later we can allow shorter outer headers than l3-l2_offset. Anyway, we could start with the strict requirement that outer header is just before l3_offset. So, the last sentence could be modified something like this:

Null value indicates that the outer headers are in the
packet data, starting at L2 offset and ending at the byte
before L3 offset. In this case, value of 'len' field must be 
greater than zero and set to L3 offset minus L2 offset.

anoobj: Agreed. Unrelated to the PR, one minor clarification. What should be the effect of 

"no outer header defined" case (ptr==NULL && len==0)

Should that be reported as an error?

JannePeltonen: That would be against the API spec so an application is not allowed to make such call. If it does, the behavior is undefined. In addition to the popular choice of starting nethack, the implementation could e.g. process the packet as if no header prepending was requested, but an application must not rely on that (ptr!=NULL, len==0 would be the only API-defined way to request zero-lenght header).

JannePeltonen:
This PR is now superseded by PR 1145.

JannePeltonen:
Continued in PR 1145.

PR #1121, opened 2020-12-01
[PATCH v7] api: new_cls_terms

kkirankk:
## api: cls: introduce IGMP group addr term

Introduce IGMP group address term.

Signed-off-by: Jerin Jacob <jerinj@marvell.com>
Signed-off-by: Kiran Kumar K <kirankumark@marvell.com>
Reviewed-by: Janne Peltonen <janne.peltonen@nokia.com>

## api: cls: introduce ICMP terms

Introduce ICMP identifier, code and type terms.

Signed-off-by: Jerin Jacob <jerinj@marvell.com>
Signed-off-by: Kiran Kumar K <kirankumark@marvell.com>
Reviewed-by: Janne Peltonen <janne.peltonen@nokia.com>

## api: cls: introduce SCTP terms

Introduce SCTP source and destination port terms.

Signed-off-by: Jerin Jacob <jerinj@marvell.com>
Signed-off-by: Kiran Kumar K <kirankumark@marvell.com>
Reviewed-by: Janne Peltonen <janne.peltonen@nokia.com>

## api: cls: introduce GTPU tunnel endpoint id term

Introduce GTPU tunnel endpoint identifier term.

Signed-off-by: Jerin Jacob <jerinj@marvell.com>
Signed-off-by: Kiran Kumar K <kirankumark@marvell.com>
Reviewed-by: Janne Peltonen <janne.peltonen@nokia.com>

## helper: add odph_gtp header description

Adds GTP header description struct odph_gtphdr_t in helper directory.
This structure is used for accessing GTPU header information from the
packet.

Signed-off-by: Kiran Kumar K <kirankumark@marvell.com>
Signed-off-by: Satheesh Paul <psatheesh@marvell.com>
Reviewed-by: Janne Peltonen <janne.peltonen@nokia.com>

## helper: add odph_igmp header description

Adds IGMP header description struct odph_igmphdr_t in helper directory.
This structure is used for accessing IGMP header information from the
packet.

Signed-off-by: Kiran Kumar K <kirankumark@marvell.com>
Signed-off-by: Satheesh Paul <psatheesh@marvell.com>
Reviewed-by: Janne Peltonen <janne.peltonen@nokia.com>

## validation: cls: add validation test cases for pmr types sctp

Adding validation test cases to test PMR types SCTP SPORT, SCTP DPORT.

Signed-off-by: Kiran Kumar K <kirankumark@marvell.com>
Reviewed-by: Janne Peltonen <janne.peltonen@nokia.com>

[...]

PR #1139, opened 2020-12-10
[PATCH v3] Classifier validation test cleanup

JannePeltonen:
Factor out repetitive test mechanics code from individual PMR test cases so that the test cases become more readable and maintainable.

This PR introduces 1 commit on top of PR #1121 and needs to be rebased after PR 1121 has been merged.

~~This PR introduces 3 commits on top of PR #1121 and needs to be rebased after PR 1121 is done.~~

~~The first new commit (f73f0312f) could be put in before PR 1121 if PR 1121 would then absorb the last two commits (i.e. reshaping the corresponding validation commits). Or if PR 1121 goes in its current form, then the two cleanup commits in this PR can be squashed and PR 1121 could still do the changes of the middle commit (9b8a734db).~~

### DIFF
--- a/test/validation/api/classification/odp_classification_test_pmr.c
+++ b/test/validation/api/classification/odp_classification_test_pmr.c
@@ -369,13 +369,22 @@ static void _classification_test_pmr_term_tcp_dport(int num_pkt)
 	odp_pktio_close(pktio);
 }
 
-static void classification_test_pmr_term_tcp_sport(void)
+typedef enum match_t {
+	MATCH,
+	NO_MATCH
+} match_t;
+
+/*
+ * Test that PMR created using the given parameters matches or does not match
+ * given packet. The packet, that gets consumed, must have been created using
+ * create_packet() so that it contains the testing sequence number.
+ *
+ * Ethernet addresses of the packet will be overwritten.
+ */
+static void test_pmr(const odp_pmr_param_t *pmr_param, odp_packet_t pkt,
+		     match_t match)
 {
-	odp_packet_t pkt;
-	odph_tcphdr_t *tcp;
 	uint32_t seqno;
-	uint16_t val;
-	uint16_t mask;
 	int retval;
 	odp_pktio_t pktio;
 	odp_queue_t queue;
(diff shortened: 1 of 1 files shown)


test/validation/api/classification/odp_classification_test_pmr.c
```diff
@@ -1580,9 +1579,6 @@ static void classification_test_pmr_term_sctp(bool is_dport)
 	CU_ASSERT(retqueue == queue);
 	recvpool = odp_packet_pool(pkt);
 	CU_ASSERT(recvpool == pool);
-	sctp = (odph_sctphdr_t *)odp_packet_l4_ptr(pkt, NULL);
-	CU_ASSERT(sctp->tag == 0x11223344);
-	odp_packet_free(pkt);
```
JannePeltonen: I think you are referring to this line. You are right, odp_packet_free() is needed. I somehow screwed this up when splitting this to a separate commit. Thanks.

Note that my suggestion was to consider this commit as a review comment to your PR so that you could add these changes and then I could just delete this commit from this PR.

test/validation/api/classification/odp_classification_test_pmr.c
```diff
@@ -369,13 +369,22 @@ static void _classification_test_pmr_term_tcp_dport(int num_pkt)
...
+} match_t;
+
+/*
+ * Test that PMR created using the given parameters matches or does not match
+ * given packet. The packet, that gets consumed, must have been created using
+ * create_packet() so that it contains the testing sequence number.
+ *
+ * Ethernet addresses of the packet will be overwritten.
+ */
+static void test_pmr(const odp_pmr_param_t *pmr_param, odp_packet_t pkt,
```
kkirankk, line 384: This function looks good, but just a suggestion, Instead of creating and destroying PMR/COS sequence for every packet,
Can we take array of packets and result as input?
May be like this,
struct test_packet {
odp_packet_t pkt;
match_t match;
}test_packet;

Then the function will be something like this.
static void test_pmr(pmr_param, test_packet *tpkt[], num_packets)

kkirankk (commented):
Looks good to me.

kkirankk (commented):
look good to me.

test/validation/api/classification/odp_classification_test_pmr.c
JannePeltonen: If there ever is a performance issue or other reason to avoid creating and destroying everything for each packet, I think it might be better to have a setup and cleanup function and have a simple single-packet interface for the actual test in between. But anyway, that more improvements can be done later if seen useful. I think this code ok for now and not much different from how things were (there are a bit more creation and destroying going on, but it should not really matter).

JannePeltonen:
v3: rebased and added reviewed-by tag

PR #1141, opened 2020-12-14
[PATCH v2] Fix strict priority scheduler group create

psavol:
Threads were not informed about new group create or destroy. Threads were informed only when group join/leave was used. Added new test case to check that newly created group schedules events (without group join).


matiaselo (approved):
Copyright in `platform/linux-generic/odp_schedule_sp.c` should be updated but otherwise looking good.

PR #1143, opened 2020-12-18
[PATCH v8] example/l2fwd: add argument to get packet length

ksatha:
Present l2fwd consider packet length 1536 for pool creation, it limits
the packet length to X * 1536 when platform supports X number of segments.
This new command line argument along with pool capabilities will create
packet pool to support packets up to 64K MTU.

Signed-off-by: Satha Rao <skoteshwar@marvell.com>

### DIFF
--- a/test/performance/odp_l2fwd.c
+++ b/test/performance/odp_l2fwd.c
@@ -105,6 +105,8 @@ typedef struct {
 	uint64_t vec_tmo_ns;    /* Vector formation timeout in ns */
 	uint32_t vec_size;      /* Vector size */
 	int verbose;		/* Verbose output */
+	uint32_t packet_len;	/* Maximum packet length supported */
+	uint32_t seg_len;	/* Pool segment length */
 	int promisc_mode;       /* Promiscuous mode enabled */
 } appl_args_t;
 
@@ -1507,10 +1509,13 @@ static void usage(char *progname)
 	       "  -x, --vec_size <num>    Vector size (default %i).\n"
 	       "  -z, --vec_tmo_ns <ns>   Vector timeout in ns (default %llu ns).\n"
 	       "  -P, --promisc_mode      Enable promiscuous mode.\n"
+	       "  -l, --packet_len <len>  Maximum length of packets supported (default %d).\n"
+	       "  -L, --seg_len <len>     Packet pool segment length\n"
+	       "                          (default equal to packet length).\n"
 	       "  -v, --verbose           Verbose output.\n"
 	       "  -h, --help              Display help and exit.\n\n"
 	       "\n", NO_PATH(progname), NO_PATH(progname), MAX_PKTIOS, DEFAULT_VEC_SIZE,
-		DEFAULT_VEC_TMO
+		DEFAULT_VEC_TMO, POOL_PKT_LEN
 	    );
 }
 
@@ -1551,12 +1556,14 @@ static void parse_args(int argc, char *argv[], appl_args_t *appl_args)
 		{"vec_tmo_ns", required_argument, NULL, 'z'},
 		{"vector_mode", no_argument, NULL, 'u'},
 		{"promisc_mode", no_argument, NULL, 'P'},
+		{"packet_len", required_argument, NULL, 'l'},
+		{"seg_len", required_argument, NULL, 'L'},
 		{"verbose", no_argument, NULL, 'v'},
 		{"help", no_argument, NULL, 'h'},
 		{NULL, 0, NULL, 0}
 	};
 
-	static const char *shortopts = "+c:t:a:i:m:o:r:d:s:e:k:g:b:p:y:n:w:x:z:uPvh";
+	static const char *shortopts = "+c:t:a:i:m:o:r:d:s:e:k:g:b:p:y:n:l:L:w:x:z:uPvh";
 
 	appl_args->time = 0; /* loop forever if time to run is 0 */
 	appl_args->accuracy = 1; /* get and print pps stats second */
@@ -1571,6 +1578,8 @@ static void parse_args(int argc, char *argv[], appl_args_t *appl_args)
 	appl_args->chksum = 0; /* don't use checksum offload by default */
 	appl_args->pool_per_if = 0;
 	appl_args->num_pkt = 0;
+	appl_args->packet_len = POOL_PKT_LEN;
+	appl_args->seg_len = UINT32_MAX;
 	appl_args->promisc_mode = 0;
 	appl_args->vector_mode = 0;
 	appl_args->num_vec = 0;
@@ -1715,6 +1724,12 @@ static void parse_args(int argc, char *argv[], appl_args_t *appl_args)
 			break;
 		case 'n':
 			appl_args->num_pkt = atoi(optarg);
+			break;
+		case 'l':
+			appl_args->packet_len = atoi(optarg);
+			break;
+		case 'L':
+			appl_args->seg_len = atoi(optarg);
 			break;
 		case 'P':
 			appl_args->promisc_mode = 1;
@@ -1928,7 +1943,7 @@ int main(int argc, char *argv[])
 	odp_pool_t pool, vec_pool;
 	odp_init_t init;
 	odp_pool_capability_t pool_capa;
-	uint32_t pkt_len, num_pkt;
+	uint32_t pkt_len, num_pkt, seg_len;
 
 	/* Let helper collect its own arguments (e.g. --odph_proc) */
 	argc = odph_parse_options(argc, argv);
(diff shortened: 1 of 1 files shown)


test/performance/odp_l2fwd.c
```diff
@@ -105,6 +108,7 @@ typedef struct {
 	uint64_t vec_tmo_ns;    /* Vector formation timeout in ns */
 	uint32_t vec_size;      /* Vector size */
 	int verbose;		/* Verbose output */
+	int packet_len;		/* Maximum pacekt length supported */
```
JereLeppanen, line 111: Typo: "pacekt".

test/performance/odp_l2fwd.c
```diff
@@ -1507,6 +1511,7 @@ static void usage(char *progname)
 	       "  -x, --vec_size <num>    Vector size (default %i).\n"
 	       "  -z, --vec_tmo_ns <ns>   Vector timeout in ns (default %llu ns).\n"
 	       "  -P, --promisc_mode      Enable promiscuous mode.\n"
+	       "  -l, --packet_len <len>  Maximum length of packets supported.\n"
```
JereLeppanen, line 1514: Default value could be printed here.

test/performance/odp_l2fwd.c
```diff
@@ -2035,13 +2045,23 @@ int main(int argc, char *argv[])
...
+	pkt_len = gbl_args->appl.packet_len;
 
 	if (pool_capa.pkt.max_len && pkt_len > pool_capa.pkt.max_len) {
 		pkt_len = pool_capa.pkt.max_len;
 		printf("\nWarning: packet length reduced to %u\n\n", pkt_len);
 	}
 
+	seg_len = pkt_len > MIN_SEG_LEN ? MIN_SEG_LEN : pkt_len;
+	if (seg_len < (pkt_len / pool_capa.pkt.max_segs_per_pkt))
+		seg_len = pkt_len / pool_capa.pkt.max_segs_per_pkt;
```
JereLeppanen, line 2057: Instead of this logic, would it be better to make segment length also a command line option?

JereLeppanen (commented):
Thanks, this looks useful.

The patch subject line prefix should be "test: l2fwd:".

ksatha:
> Thanks, this looks useful.
> 
> The patch subject line prefix should be "test: l2fwd:".

Thanks for reviewing. Will change subject in next revision.

test/performance/odp_l2fwd.c
ksatha: ACK

test/performance/odp_l2fwd.c
ksatha: ACK

test/performance/odp_l2fwd.c
ksatha: Adding segment length option will be good suggestion, will submit revised patch with new argument.
And I will move this logic to under no argument case, because if user requested bigger packet length(say 16K) and platform supports limited segments(assume 8), keeping default segment length (1536) will be problem here, so we need to tune segment length based on new packet length.

test/performance/odp_l2fwd.c
```diff
@@ -105,6 +108,8 @@ typedef struct {
 	uint64_t vec_tmo_ns;    /* Vector formation timeout in ns */
 	uint32_t vec_size;      /* Vector size */
 	int verbose;		/* Verbose output */
+	int packet_len;		/* Maximum packet length supported */
+	uint32_t seg_len;	/* Pool segment length */
```
JereLeppanen, line 112: Minor thing, but these should be of the same type, it looks funny if they're not.

test/performance/odp_l2fwd.c
```diff
@@ -1507,6 +1512,8 @@ static void usage(char *progname)
 	       "  -x, --vec_size <num>    Vector size (default %i).\n"
 	       "  -z, --vec_tmo_ns <ns>   Vector timeout in ns (default %llu ns).\n"
 	       "  -P, --promisc_mode      Enable promiscuous mode.\n"
+	       "  -l, --packet_len <len>  Maximum length of packets supported(default length 1536).\n"
+	       "  -L, --seg_len <len>     Packet pool segment length(default length 1536).\n"
```
JereLeppanen, line 1516: The defaults here should be handled the same way as in the other options.

test/performance/odp_l2fwd.c
```diff
@@ -2035,13 +2052,30 @@ int main(int argc, char *argv[])
...
+	pkt_len = gbl_args->appl.packet_len;
 
 	if (pool_capa.pkt.max_len && pkt_len > pool_capa.pkt.max_len) {
 		pkt_len = pool_capa.pkt.max_len;
 		printf("\nWarning: packet length reduced to %u\n\n", pkt_len);
 	}
 
+	seg_len = gbl_args->appl.seg_len;
+
+	/* Check weather we have sufficient segments to support requested packet
```
JereLeppanen, line 2064: Typo: "weather" -> "whether".

test/performance/odp_l2fwd.c
```diff
@@ -2035,13 +2052,30 @@ int main(int argc, char *argv[])
...
+	 * length, if not adjust to bigger segment size */
+	if ((seg_len * pool_capa.pkt.max_segs_per_pkt) < pkt_len)
+		seg_len = pkt_len / pool_capa.pkt.max_segs_per_pkt;
```
JereLeppanen, line 2067: It would be clearer to use the same calculation for the condition, i.e. if (seg_len < pkt_len / pool_capa.pkt.max_segs_per_pkt).

JereLeppanen (commented):
I added a few mostly minor comments, please have a look.

Also please remove the merge commit and rebase.

test/performance/odp_l2fwd.c
ksatha: ACK

test/performance/odp_l2fwd.c
ksatha: Will change if condition. Regarding warning at line number 2076 we are printing both requested and configured values for user reference

test/performance/odp_l2fwd.c
ksatha: ACK

ksatha:
> I added a few mostly minor comments, please have a look.
> 
> Also please remove the merge commit and rebase.

Thanks for suggestions. Updated patch pushed with fixes can you please review

test/performance/odp_l2fwd.c
```diff
@@ -1571,6 +1580,8 @@ static void parse_args(int argc, char *argv[], appl_args_t *appl_args)
 	appl_args->chksum = 0; /* don't use checksum offload by default */
 	appl_args->pool_per_if = 0;
 	appl_args->num_pkt = 0;
+	appl_args->packet_len = POOL_PKT_LEN;
+	appl_args->seg_len = POOL_SEG_LEN;
```
JereLeppanen, line 1584: Sorry, I should have realized earlier that segment length should probably default to packet length, if the segment length argument is not given.

test/performance/odp_l2fwd.c
```diff
@@ -1507,10 +1512,12 @@ static void usage(char *progname)
 	       "  -x, --vec_size <num>    Vector size (default %i).\n"
 	       "  -z, --vec_tmo_ns <ns>   Vector timeout in ns (default %llu ns).\n"
 	       "  -P, --promisc_mode      Enable promiscuous mode.\n"
+	       "  -l, --packet_len <len>  Maximum length of packets supported(default %d).\n"
+	       "  -L, --seg_len <len>     Packet pool segment length(default %d).\n"
```
JereLeppanen, line 1516: One more nitpick here: space before "(default", just like the other defaults, above.

test/performance/odp_l2fwd.c
ksatha: ACK

test/performance/odp_l2fwd.c
ksatha: ACK

test/performance/odp_l2fwd.c
```diff
@@ -1571,6 +1577,8 @@ static void parse_args(int argc, char *argv[], appl_args_t *appl_args)
 	appl_args->chksum = 0; /* don't use checksum offload by default */
 	appl_args->pool_per_if = 0;
 	appl_args->num_pkt = 0;
+	appl_args->packet_len = POOL_PKT_LEN;
+	appl_args->seg_len = POOL_PKT_LEN;
```
JereLeppanen, line 1581: This makes segment length and packet length default to the same value, which is different from segment length defaulting to packet length. To elaborate, I meant that once we've figured out what packet length is used, by processing command line parameters and checking capabilities, segment length should then default to that resulting packet length.

ksatha: Hi JereLeppanen, Thanks for detailed explanation, will integrate this logic in next version

test/performance/odp_l2fwd.c
```diff
@@ -1507,10 +1509,12 @@ static void usage(char *progname)
 	       "  -x, --vec_size <num>    Vector size (default %i).\n"
 	       "  -z, --vec_tmo_ns <ns>   Vector timeout in ns (default %llu ns).\n"
 	       "  -P, --promisc_mode      Enable promiscuous mode.\n"
+	       "  -l, --packet_len <len>  Maximum length of packets supported (default %d).\n"
+	       "  -L, --seg_len <len>     Packet pool segment length (default %d).\n"
```
JereLeppanen, line 1513: Now this is no longer the default. Maybe something like "(default equal to packet length)".

ksatha: ACK

JereLeppanen (approved):
OK, looks good. Thanks for bearing with me.

ksatha:
> OK, looks good. Thanks for bearing with me.

Thanks for your review and valuable suggestions.

matiaselo:
@ksatha Please add Jere's reviewed-by tag to the commit, so it can be merged.

ksatha:
> @ksatha Please add Jere's reviewed-by tag to the commit, so it can be merged.

Thanks, updated commit message.

PR #1144, opened 2020-12-18
[PATCH v2] Fix max sched groups

psavol:
Fix max scheduler group capability in basic and scalable scheduler. Add validation test case to create max number of groups.


matiaselo (approved):
Tested with PR #1072.

PR #1146, opened 2020-12-22
[PATCH v4] helper: ipsec: add capability check function

asasidharan:
Add IPSEC capability check function in helper to check whether the specified
IPSEC configuration is supported.

### DIFF
--- a/helper/include/odp/helper/ipsec.h
+++ b/helper/include/odp/helper/ipsec.h
@@ -71,6 +71,29 @@ ODP_STATIC_ASSERT(sizeof(odph_ahhdr_t) == ODPH_AHHDR_LEN,
 		  "ODPH_AHHDR_T__SIZE_ERROR");
 
 /**
+ * Check IPSEC algorithm support
+ *
+ * Based on the capabilities exposed by the ODP implementation, check whether
+ * the specified IPSEC algorithm configuration is supported by the
+ * implementation. The caller provides the IPSEC capability structure as an
+ * argument to the helper function.
+ *
+ * @param      capa            IPSEC capability structure
+ * @param      cipher_alg      Cipher algorithm
+ * @param      cipher_key_len  Length of cipher key in bytes
+ * @param      auth_alg        Authentication algorithm
+ * @param      auth_key_len    Length of authentication key in bytes
+ *
+ * @retval 0  on success
+ * @retval <0 on failure
+ */
+int odph_ipsec_alg_check(odp_ipsec_capability_t capa,
+			 odp_cipher_alg_t cipher_alg,
+			 uint32_t cipher_key_len,
+			 odp_auth_alg_t auth_alg,
+			 uint32_t auth_key_len);
+
+/**
  * @}
  */
 
--- a/helper/ipsec.c
+++ b/helper/ipsec.c
@@ -1,0 +1,163 @@
+/* Copyright (c) 2017-2018, Linaro Limited
+ * Copyright (c) 2020 Marvell
+ * All rights reserved.
+ *
+ * SPDX-License-Identifier:	 BSD-3-Clause
+ */
+
+#include <odp/helper/ipsec.h>
+#include <odp/helper/odph_debug.h>
+
+int odph_ipsec_alg_check(odp_ipsec_capability_t capa,
+			 odp_cipher_alg_t cipher_alg,
+			 uint32_t cipher_key_len,
+			 odp_auth_alg_t auth_alg,
+			 uint32_t auth_key_len)
+{
+	int i, num, max_capa;
+	odp_bool_t found;
+
+	/* Check whether requested cipher algorithm is supported */
+	switch (cipher_alg) {
+	case ODP_CIPHER_ALG_NULL:
+		if (!capa.ciphers.bit.null)
+			return -1;
+		break;
+	case ODP_CIPHER_ALG_DES:
+		if (!capa.ciphers.bit.des)
+			return -1;
+		break;
+	case ODP_CIPHER_ALG_3DES_CBC:
+		if (!capa.ciphers.bit.trides_cbc)
+			return -1;
+		break;
+	case ODP_CIPHER_ALG_AES_CBC:
+		if (!capa.ciphers.bit.aes_cbc)
+			return -1;
+		break;
+	case ODP_CIPHER_ALG_AES_CTR:
+		if (!capa.ciphers.bit.aes_ctr)
+			return -1;
+		break;
+	case ODP_CIPHER_ALG_AES_GCM:
+		if (!capa.ciphers.bit.aes_gcm)
+			return -1;
+		break;
+	case ODP_CIPHER_ALG_AES_CCM:
+		if (!capa.ciphers.bit.aes_ccm)
+			return -1;
+		break;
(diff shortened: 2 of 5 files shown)


test/performance/odp_ipsec.c
```diff
@@ -784,20 +784,181 @@ run_measure_one_async(ipsec_args_t *cargs,
...
+		break;
+	default:
+		printf("    Unsupported authentication algorithm\n");
+		return -1;
+	}
+
+	num = odp_ipsec_cipher_capability(config->crypto.cipher_alg,
+					  cipher_capa, MAX_ALG_CAPA);
+	if (num <= 0) {
+		printf("    Wrong cipher capabilities\n");
```
JannePeltonen, line 896: I think this is a bit unclear message, even though the same message is in the validation test. Something like "Could not get cipher capabilities" or "Cipher capability call failed" would be more clear, I think.

JannePeltonen:
This looks good to me, except for the minor error message comment and the fact that this code is a duplicate of the same code in the validation test. There is a potential of having more of this duplication if more of the ipsec tests and examples start doing proper capability checks. So maybe the right thing to do would be to add IPsec capability check function in the helper library and just use it from here and other code?

@psavol, @matiaselo, I wonder what is the correct prefix for the commit subject line. This one has "performance: ipsec" but looking at the repository, maybe the prefix should be "test: ipsec" or "test: performance"?

test/performance/odp_ipsec.c
asasidharan: Agreed. Will update this in the next version.

asasidharan:
> So maybe the right thing to do would be to add IPsec capability check function in the helper library and just use it from here and other code?

If @psavol  is also aligned with this, I can work on adding the IPsec capability check function  in the helper library.

helper/ipsec.c
```diff
@@ -0,0 +1,157 @@
...
+				odp_auth_alg_t auth_alg,
+				uint32_t auth_key_len)
+{
+	odp_ipsec_capability_t capa;
+	odp_ipsec_cipher_capability_t cipher_capa[MAX_ALG_CAPA];
+	odp_ipsec_auth_capability_t   auth_capa[MAX_ALG_CAPA];
+	int i, num;
+	odp_bool_t found;
+
+	if (odp_ipsec_capability(&capa) < 0)
```
JannePeltonen, line 20: I do not this this AH capa checks belongs here. There are plenty of various capability flags in the capability structure and it does not make sense to check them all here (requiring a huge number of parameters too). An application can as easily check the relevant capability flags itself.

The crypto algorithm check is the useful part that improves code reuse and that is what this function should concentrate on. The name should then be changed accordingly (odp_ipsec_alg_check?).

Maybe the function should take odp_ipsec_capability_t as a parameter?

helper/include/odp/helper/ipsec.h
```diff
@@ -70,6 +70,29 @@ typedef struct ODP_PACKED {
 ODP_STATIC_ASSERT(sizeof(odph_ahhdr_t) == ODPH_AHHDR_LEN,
 		  "ODPH_AHHDR_T__SIZE_ERROR");
 
+#define MAX_ALG_CAPA 32  /**< Maximum algorithm capabilities */
```
JannePeltonen, line 73: This is a bit problematic. Some day 32 may not be enough and it might be hard to spot the limitation here. I think at minimum the function should print an error message if the array turns out to be too small. Using a dynamically sized array (c.g. odp_ipsec.c:odp_ipsec_cipher_capability) would be another alternative.

helper/ipsec.c
```diff
@@ -0,0 +1,157 @@
...
+	num = odp_ipsec_cipher_capability(cipher_alg, cipher_capa, MAX_ALG_CAPA);
+	if (num <= 0) {
+		ODPH_ERR("Could not get cipher capabilities\n");
+		return -1;
+	}
+
+	/* Search for the test case */
+	found = false;
+	for (i = 0; i < num; i++) {
+		if (cipher_capa[i].key_len == cipher_key_len) {
```
JannePeltonen, line 125: This does out-of-bounds array access if the number of cipher capabilities exceed the size of the array. Note that odp_ipsec_cipher_capability() returns the total number of capabilities there are, even if it is larger than the size of the supplied array.

helper/ipsec.c
```diff
@@ -0,0 +1,157 @@
...
+			found = 1;
+			break;
+		}
+	}
+
+	if (!found) {
+		ODPH_ERR("Unsupported key length\n");
```
JannePeltonen, line 132: I do not think this should print an error when it turns out that a queried algorithm+keylen is not supported. It is an application matter whether an error needs to be logged or not.

helper/ipsec.c
```diff
@@ -0,0 +1,157 @@
...
+		return -1;
+	}
+
+	num = odp_ipsec_cipher_capability(cipher_alg, cipher_capa, MAX_ALG_CAPA);
+	if (num <= 0) {
+		ODPH_ERR("Could not get cipher capabilities\n");
+		return -1;
+	}
+
+	/* Search for the test case */
```
JannePeltonen, line 122: These comments do not make much sense here. This is not a test specific piece of code anymore and I think the comments were bad even in the original context.

helper/ipsec.c
```diff
@@ -0,0 +1,157 @@
+/* SPDX-License-Identifier: BSD-3-Clause
+ * Copyright(C) 2020 Marvell.
```
JannePeltonen, line 2: This should include:
Copyright (c) 2017-2018, Linaro Limited

We do not want to remove copyright attribution of existing code just because we move it to a different file.

helper/ipsec.c
asasidharan: Ack.
Updated the helper function to just check algorithm specific capabilities.
Also added odp_ipsec_capability_t as a new parameter that needs to be passed to the function.

helper/include/odp/helper/ipsec.h
asasidharan: Ack. Updating to use dynamically sized array similar to odp_ipsec.c:odp_ipsec_cipher_capability.

helper/ipsec.c
asasidharan: Ack. Updated the function to use dynamic sized array.

helper/ipsec.c
asasidharan: Made this to debug print currently as there are multiple points where the function can return error. Or would you prefer to simply remove the prints altogether?

helper/ipsec.c
asasidharan: Agreed. Updated the comments.

helper/ipsec.c
asasidharan: Ack. Keeping "Copyright (c) 2017-2018, Linaro Limited" in the new file.

JannePeltonen:
Looks good, but please change the subject line of the last commit to "test: ipsec: add capability checks"

asasidharan:
Thank you @JannePeltonen for reviewing the PR.

Updated v4 with reviewed-by tag and last commit's subject line updated as suggested.

JannePeltonen (approved):

PR #1149, opened 2021-01-05
[PATCH v2] linux-gen: packet: fix packet print

psavol:
Packet print function used buffer_snprint, which assumed
that event type is buffer and printed an error on packets.
Re-implemented buffer and packet print functions to not
use buffer_snprint.

### DIFF
--- a/platform/linux-generic/odp_packet.c
+++ b/platform/linux-generic/odp_packet.c
@@ -1567,10 +1567,11 @@ void odp_packet_print(odp_packet_t pkt)
 	int len = 0;
 	int n = max_len - 1;
 	odp_packet_hdr_t *hdr = packet_hdr(pkt);
-	odp_buffer_t buf      = packet_to_buffer(pkt);
 
-	len += snprintf(&str[len], n - len, "Packet ");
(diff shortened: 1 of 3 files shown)


platform/linux-generic/odp_packet.c
```diff
@@ -1567,10 +1567,11 @@ void odp_packet_print(odp_packet_t pkt)
 	int len = 0;
 	int n = max_len - 1;
 	odp_packet_hdr_t *hdr = packet_hdr(pkt);
-	odp_buffer_t buf      = packet_to_buffer(pkt);
 
-	len += snprintf(&str[len], n - len, "Packet ");
-	len += odp_buffer_snprint(&str[len], n - len, buf);
+	len += snprintf(&str[len], n - len, "Packet\n------\n");
```
JereLeppanen, line 1571: odp_buffer_print() doesn't print similar underlining. Is this intentional?

psavol: No, I'll add it there as well. I added it into this function as the other packet print had it.

JereLeppanen (approved):

PR #1151, opened 2021-01-05
[PATCH v5] api: sched: pause and scheduling group configuration improvements

matiaselo:
V4:
- Rebase
- Reviewed-by tags

### DIFF
--- a/include/odp/api/spec/schedule.h
+++ b/include/odp/api/spec/schedule.h
@@ -340,10 +340,10 @@ int odp_schedule_capability(odp_schedule_capability_t *capa);
  * Creates a schedule group with the thread mask. Only threads in the
  * mask will receive events from a queue that belongs to the schedule group.
  * Thread masks of various schedule groups may overlap. There are predefined
- * groups such as ODP_SCHED_GROUP_ALL and ODP_SCHED_GROUP_WORKER, which are
- * always present and automatically updated. The use of group name is optional.
- * Unique names are not required. However, odp_schedule_group_lookup() returns
- * only a single matching group.
+ * groups, such as ODP_SCHED_GROUP_ALL and ODP_SCHED_GROUP_WORKER, which are
+ * present by default and are automatically updated. The use of group name is
+ * optional. Unique names are not required. However, odp_schedule_group_lookup()
+ * returns only a single matching group.
  *
  * @param name    Name of the schedule group or NULL. Maximum string length is
  *                ODP_SCHED_GROUP_NAME_LEN.
--- a/include/odp/api/spec/schedule.h
+++ b/include/odp/api/spec/schedule.h
@@ -157,13 +157,27 @@ int odp_schedule_multi_no_wait(odp_queue_t *from, odp_event_t events[],
 /**
  * Pause scheduling
  *
- * Pause global scheduling for this thread. After this call, all schedule calls
- * will return only locally pre-scheduled events (if any). User can exit the
- * schedule loop only after the schedule function indicates that there's no more
- * (pre-scheduled) events.
+ * Pause global scheduling for this thread. After this call, only
+ * ODP_SCHED_NO_WAIT schedule calls are allowed and these calls will return only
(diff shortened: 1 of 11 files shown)


shijith (commented):
Tested-by: Shijith Thotton <sthotton@marvell.com>

include/odp/api/spec/schedule.h
```diff
@@ -157,13 +157,14 @@ int odp_schedule_multi_no_wait(odp_queue_t *from, odp_event_t events[],
...
- * Must be used with odp_schedule() and odp_schedule_multi() before exiting (or
- * stalling) the schedule loop.
+ * Pause global scheduling for this thread. After this call, only
+ * ODP_SCHED_NO_WAIT schedule calls are allowed and these calls will return only
+ * locally pre-scheduled events (if any). User can exit the schedule loop only
+ * after the schedule function indicates that there's no more (pre-scheduled)
+ * events.
+ *
+ * Must be used with odp_schedule(), odp_schedule_multi(), or
+ * odp_schedule_multi_no_wait() before exiting (or stalling) the schedule loop.
```
psavol, line 167: This refers actually to the main schedule loop (not to sched calls after pause), so the list should cover all odp_schedule calls. It could be better change the text to refer to schedule loop and list e.g. odp_schedule() as an example. Also an example (code) of the required call pattern could be added ( schedule loop -> pause -> schedule no wait until no events -> exit schedule loop).

include/odp/api/spec/schedule_types.h
```diff
@@ -260,6 +260,18 @@ typedef struct odp_schedule_config_t {
 	 */
 	uint32_t max_flow_id;
 
+	/** Predefined scheduling groups */
+	struct {
```
psavol, line 266: Documentation could be a bit more explicit that this config is used to enable/disable predefined groups (e.g. 1 == enable, 0 == disable) and that default value is "enabled" for all.

include/odp/api/spec/schedule_types.h
```diff
@@ -253,8 +255,8 @@ typedef struct odp_schedule_config_t {
 	 *  scheduling of the event and synchronization is maintained per flow
 	 *  within each queue.
 	 *
-	 *  Depeding on implementation, there may be much more flows supported
-	 *  than queues, as flows are lightweight entities.
+	 *  Depending on the implementation, there may be much more flows
+	 *  supported than queues, as flows are lightweight entities.
 	 *
```
psavol, line 260: this typo correction could be moved to the group config patch.

include/odp/api/spec/schedule_types.h
```diff
@@ -190,7 +190,9 @@ typedef struct odp_schedule_capability_t {
 	/** Maximum number of ordered locks per queue */
 	uint32_t max_ordered_locks;
 
-	/** Maximum number of scheduling groups */
+	/** Maximum number of scheduling groups. The value includes the enabled
+	 *  predefined scheduling groups (ODP_SCHED_GROUP_ALL,
+	 *  ODP_SCHED_GROUP_WORKER, and ODP_SCHED_GROUP_CONTROL). */
```
psavol, line 195: Could mention that by default application may create in maximum max_groups - 3 groups.

include/odp/api/spec/schedule.h
matiaselo: Ok, fixed in V2.

include/odp/api/spec/schedule_types.h
matiaselo: Fixed in V2.

include/odp/api/spec/schedule_types.h
matiaselo: Fixed in V2.

include/odp/api/spec/schedule_types.h
matiaselo: Fixed in V2.

psavol (approved):

PR #1154, opened 2021-01-07
[PATCH v11] api: pktio: enable setting input and output maximum frame lengths

matiaselo:
Add new function odp_pktio_maxlen_set() for setting the maximum frame
lengths for packet input and output. The required capabilities have been
added to odp_pktio_capability_t.

V2:
- Renamed odp_pktio_set_op_t.op.pktio_maxlen to maxlen (Jere)

V3:
- Renamed odp_pktio_maxlen_set() arguments (Petri)
- Renamed odp_pktio_capability_t.maxlen member and improved documentation (Petri)

V4:
- Squashed commits

V5:
- Added implementation


include/odp/api/spec/packet_io.h
```diff
@@ -610,6 +610,8 @@ typedef union odp_pktio_set_op_t {
 		uint32_t mac_addr : 1;
 		/** Per port header offset(skip)set */
 		uint32_t skip_offset : 1;
+		/** Maximum frame length */
+		uint32_t pktio_maxlen : 1;
```
JereLeppanen, line 614: "pktio" seems redundant here since this is inside odp_pktio_set_op_t. Should this just be the same as (or in some way aligned with) the name of the capability struct?

matiaselo: The target was to align with the matching function names. I'll drop the redundant `pktio` prefix.

JereLeppanen (commented):
In the commit comment, "The required capabilities have been
added" sounds like they were added in an earlier commit. Something like "Add the corresponding capabilities" might be better.

Otherwise looks good. Please add my reviewed-by tag.

nithind1988:

include/odp/api/spec/packet_io.h
```diff
@@ -673,6 +675,25 @@ typedef struct odp_pktio_capability_t {
...
 
+	/** Supported frame lengths for odp_pktio_maxlen_set()
+	 *
+	 * A frame length value of zero indicates an unsupported operation. If
+	 * the 'common_len' flag is set, the same maximum frame length is used
+	 * for both input and output directions. In this case the min/max values
+	 * for input and output are identical. */
+	struct {
+		/** Common maximum frame length for both packet input and output */
+		odp_bool_t common_len;
```
psavol, line 686: The flag documentation from odp_pktio_maxlen_set() could be copied here:

... when flag is set, the same maximum frame length value has to be used for both input and output directions

include/odp/api/spec/packet_io.h
```diff
@@ -673,6 +675,25 @@ typedef struct odp_pktio_capability_t {
...
+		/** Minimum supported frame length at packet input */
+		uint32_t min_input_len;
+		/** Maximum supported frame length at packet input */
+		uint32_t max_input_len;
+		/** Minimum supported frame length at packet output */
+		uint32_t min_output_len;
+		/** Maximum supported frame length at packet output */
+		uint32_t max_output_len;
+	} max_frame_len;
```
psavol, line 695: Naming and documentation could be clearer that these are min/max valid values for odp_pktio_maxlen_set() params (and not the current link min/max frame sizes).

matiaselo:
> In the commit comment, "The required capabilities have been
> added" sounds like they were added in an earlier commit. Something like "Add the corresponding capabilities" might be better.
> 
> Otherwise looks good. Please add my reviewed-by tag.

Ack

include/odp/api/spec/packet_io.h
matiaselo: Ack

include/odp/api/spec/packet_io.h
matiaselo: Ack

psavol (approved):

PR #1156, opened 2021-01-12
validation: cls: fix GTP test by changing seqno location

satheeshpaul:
One GTP-u PMR test creates a GTP packet but then changes the UDP port
which makes cls_pkt_get_seq() treat the packet as a plain UDP packet,
causing it to read the test packet sequence number from a wrong location.

Fix the problem and simplify the sequence number functions by storing the
sequence number in the last bytes of the packet regardless of packet type.

### DIFF
--- a/test/validation/api/classification/odp_classification_common.c
+++ b/test/validation/api/classification/odp_classification_common.c
@@ -91,98 +91,60 @@ int stop_pktio(odp_pktio_t pktio)
 	return 0;
 }
 
+static uint32_t seqno_offset(odp_packet_t pkt)
+{
+	uint32_t l3_offset = odp_packet_l3_offset(pkt);
+	int rc;
+	uint16_t len = 0;
+
+	CU_ASSERT_FATAL(l3_offset != ODP_PACKET_OFFSET_INVALID);
+
+	if (odp_packet_has_ipv4(pkt)) {
+		odph_ipv4hdr_t ip;
+
+		rc = odp_packet_copy_to_mem(pkt, l3_offset, sizeof(ip), &ip);
+		CU_ASSERT_FATAL(rc == 0);
+		len = odp_be_to_cpu_16(ip.tot_len);
+	} else if (odp_packet_has_ipv6(pkt)) {
+		odph_ipv6hdr_t ip;
+
+		rc = odp_packet_copy_to_mem(pkt, l3_offset, sizeof(ip), &ip);
+		CU_ASSERT_FATAL(rc == 0);
+		len = sizeof(ip) + odp_be_to_cpu_16(ip.payload_len);
+	} else {
+		CU_FAIL_FATAL("Unexcpected packet type");
+	}
+
+	return l3_offset + len - sizeof(cls_test_packet_t);
+}
+
 int cls_pkt_set_seq(odp_packet_t pkt)
 {
-	static uint32_t seq;
 	cls_test_packet_t data;
+	static uint32_t seq;
 	uint32_t offset;
-	odph_ipv4hdr_t *ip;
-	odph_tcphdr_t *tcp;
-	odph_udphdr_t *udp;
-	uint16_t port = 0;
-	uint32_t hlen = 0;
 	int status;
 
 	data.magic = DATA_MAGIC;
 	data.seq = ++seq;
 
-	ip = (odph_ipv4hdr_t *)odp_packet_l3_ptr(pkt, NULL);
-	offset = odp_packet_l4_offset(pkt);
-	CU_ASSERT_FATAL(offset != ODP_PACKET_OFFSET_INVALID);
+	offset = seqno_offset(pkt);
 
-	if (ip->proto == ODPH_IPPROTO_IGMP) {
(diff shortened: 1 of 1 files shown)


JannePeltonen:
While this change fixes the bug, it also alters the test packet so that the TEID no longer matches the magic value and the PMR match would fail because of that even if the tested implementation did not check the UDP port number at all. This is not good since the intention of the test is to check that mismatching UDP port causes a match failure.

Because of this I suggest a different fix: Change the code so that the sequence number is stored in the last 4 bytes of the packet instead of immediately after the headers. Then the currently faulty code would work as such as it would no longer be a problem to change a GTP packet to a mere UDP packet. It would also make the sequence number setting and getting functions quite a bit simpler as they would not need the packet type specific cruft in them.

test/validation/api/classification/odp_classification_common.c
```diff
@@ -93,97 +93,60 @@ int stop_pktio(odp_pktio_t pktio)
...
+	uint8_t *buf;
 	int status;
 
+	buf = odp_packet_data(pkt);
+
+	if (odp_packet_has_ipv4(pkt)) {
+		l3_offset = odp_packet_l3_offset(pkt);
+		ipv4hdr = (odph_ipv4hdr_t *)(buf + l3_offset);
+		l3_len = odp_be_to_cpu_16(ipv4hdr->tot_len);
+		seqno_offset = l3_offset + l3_len - sizeof(data);
```
JannePeltonen, line 110: Why not simply: seqno_offset = odp_packet_len(pkt) - sizeof(data);

Then the pair of functions would work also with non-IP packets and with
segmented packets.

JannePeltonen:
The commit message needs to be tuned. This version lacks information of the true motivation of this change, i.e. fixing one of the GTP PMR tests. The subject line also implies that there was something to be fixed in the seqno functions, even though that was not quite the case.

How about something like this:

```
validation: cls: fix GTP test by changing seqno location

One GTP-u PMR test creates a GTP packet but then changes the UDP port
which makes cls_pkt_get_seq() treat the packet as a plain UDP packet,
causing it to read the test packet sequence number from a wrong location.

Fix the problem and simplify the sequence number functions by storing the
sequence number in the last bytes of the packet regardless of packet type.
```

test/validation/api/classification/odp_classification_common.c
satheeshpaul: On Octeon, there is a configuration (NIX_AF_SMQX_CFG[MINLEN]) which tells the hardware to add padding if the packet received is below the threshold. So, when the testcase creates a 50b packet (ex:UDP dport testcase)  and sends it, on the receive side, packet gets padded and becomes 60b. The receive side code then looks for the seqno at a different offset and quite a few tests fail because of this. Hence, the logic is made to rely on L3 lengths.

JannePeltonen: Ok, but it would still be good to write the code so that it does not assume single-segment packets. Like this:

```
static uint32_t seqno_offset(odp_packet_t pkt)
{
        uint32_t l3_offset = odp_packet_l3_offset(pkt);
        int rc;
        uint16_t len = 0;

        CU_ASSERT_FATAL(l3_offset != ODP_PACKET_OFFSET_INVALID);

        if (odp_packet_has_ipv4(pkt)) {
                odph_ipv4hdr_t ip;

                rc = odp_packet_copy_to_mem(pkt, l3_offset, sizeof(ip), &ip);
                CU_ASSERT_FATAL(rc == 0);
                len = odp_be_to_cpu_16(ip.tot_len);
        } else if (odp_packet_has_ipv6(pkt)) {
                odph_ipv6hdr_t ip;

                rc = odp_packet_copy_to_mem(pkt, l3_offset, sizeof(ip), &ip);
                CU_ASSERT_FATAL(rc == 0);
                len = sizeof(ip) + odp_be_to_cpu_16(ip.payload_len);
        } else
                CU_FAIL_FATAL("Unexcpected packet type");

        return l3_offset + len - sizeof(cls_test_packet_t);
}
```

test/validation/api/classification/odp_classification_common.c
```diff
@@ -91,98 +91,58 @@ int stop_pktio(odp_pktio_t pktio)
...
-	} else {
-		tcp = (odph_tcphdr_t *)odp_packet_l4_ptr(pkt, NULL);
-		odp_packet_copy_to_mem(pkt, offset + tcp->hl * 4,
-				       sizeof(data), &data);
-	}
+	uint32_t offset;
+
+	offset = seqno_offset(pkt);
+
+	odp_packet_copy_to_mem(pkt, offset, sizeof(data), &data);
```
JannePeltonen, line 145: I think the success of the copy_to_mem operation should be checked. If it fails, the magic number may still match because the now uninitialized data may still hold the magic number from a previous invocation.

satheeshpaul: Ack

JannePeltonen (approved):

matiaselo:
Hi @satheeshpaul, please add Janne's reviewed-by tag so we can merge this PR.

PR #1159, opened 2021-01-19
[PATCH v3] example: ipsec: fix verification of output stream

asasidharan:
Report failure in verifying output stream in case the received
packets are not IPsec processed as configured.
Without this, the verify succeeds even when the input stream
is directly forwarded to the output interface of the stream entry.

### DIFF
--- a/example/ipsec_crypto/odp_ipsec_stream.c
+++ b/example/ipsec_crypto/odp_ipsec_stream.c
@@ -425,6 +425,13 @@ odp_bool_t verify_ipv4_packet(stream_db_entry_t *stream,
 	/* Find IPsec headers if any and compare against entry */
 	hdr_len = locate_ipsec_headers(ip, &ah, &esp);
 
+	/* Verify if the packet is IPsec encapsulated or is cleartext as
+	 * expected
+	 */
+	if (((stream->output.entry && (!ah && !esp))) ||
+	    (stream->input.entry && (ah || esp)))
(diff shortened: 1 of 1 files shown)


example/ipsec_crypto/odp_ipsec_stream.c
```diff
@@ -425,6 +425,11 @@ odp_bool_t verify_ipv4_packet(stream_db_entry_t *stream,
 	/* Find IPsec headers if any and compare against entry */
 	hdr_len = locate_ipsec_headers(ip, &ah, &esp);
 
+	/* Report failure if the stream is not IPsec protected as configured */
```
JannePeltonen, line 428: The change looks otherwise ok to me but the comment does not seem to quite match with the test (the input part). Maybe reword "...if the packet is not IPsec encapsulated or cleartext as expected"?

asasidharan: Agreed. Updating the comment as:
```
+       /* Verify if the packet is IPsec encapsulated or is cleartext as         
+        * expected                                                              
+        */
```
Let me know if this text looks OK.

JannePeltonen (approved):

psavol:
Needs reviewed-by tag.

asasidharan:
@JannePeltonen Thank you for reviewing. @psavol Updated v3 with Reviewed-by tag.

PR #1160, opened 2021-01-20
[PATCH v1] validation: pktio: skip enabling inline IPsec

vvelumuri:
Inline IPsec functionality is not required for pktio tests.
Disable features that are not required.

### DIFF
--- a/test/validation/api/pktio/pktio.c
+++ b/test/validation/api/pktio/pktio.c
@@ -1564,6 +1564,10 @@ static void pktio_test_pktio_config(void)
 	CU_ASSERT_FATAL(odp_pktio_capability(pktio, &capa) == 0);
 
 	config = capa.config;
+
+	config.inbound_ipsec = 0;
+	config.outbound_ipsec = 0;
+
 	CU_ASSERT(odp_pktio_config(pktio, &config) == 0);
 
 	CU_ASSERT_FATAL(odp_pktio_close(pktio) == 0);


JannePeltonen:
I do not see the reason for this change and the commit message does not make sense to me (what does "required" mean and why would it be good to narrow the test because of what is "required"?).

pktio_test_pktio_config() tests that it is possible to enable all the settings that the capability function advertises as supported. That includes inline IPsec. After this change an implementation could advertise inline ipsec capability but not accept it in the config and the test would no longer notice it.

anoobj:
> I do not see the reason for this change and the commit message does not make sense to me (what does "required" mean and why would it be good to narrow the test because of what is "required"?).

Currently, all validations tests are testing with all features enabled. It is happening because we use config from capabilities. Issue with inline IPsec is, it could have additional dependencies. Like, IPsec configuration also has an enable flag for inline IPsec (op_mode). For pktio tests, that is not enabled. So what is the expectation when pktio config tries to have inline IPsec enabled? 

> pktio_test_pktio_config() tests that it is possible to enable all the settings that the capability function advertises as supported. That includes inline IPsec. After this change an implementation could advertise inline ipsec capability but not accept it in the config and the test would no longer notice it.

But then, there are IPsec tests specifically for this. IMO, isn't it better to have some tests that test on different feature set to allow for more test coverage? Implementation can have different code paths based on the features enabled.

Currently, the patch looks like disabling features because we are passing config from capabilities directly to pktio_config. If it wasn't that way, the application is likely to just enable what it needs (which won't include inline IPsec).

JannePeltonen:
The test that is being modified by the proposed commit is the one that tests that odp_pktio_config() accepts the settings advertised by odp_pktio_capability(). It does not test anything else. The commit message and your comment imply that it would be about something else. But it is only about testing odp_pktio_config() with respect to odp_pktio_capability(). Why would it be good to reduce the test coverage of that test? Or if inline IPsec should be disabled in that test, then why, by the same argument, every other feature should not be disabled too?

anoobj:
> The test that is being modified by the proposed commit is the one that tests that odp_pktio_config() accepts the settings advertised by odp_pktio_capability(). It does not test anything else. The commit message and your comment imply that it would be about something else. But it is only about testing odp_pktio_config() with respect to odp_pktio_capability(). Why would it be good to reduce the test coverage of that test? Or if inline IPsec should be disabled in that test, then why, by the same argument, every other feature should not be disabled too?

May be let me explain the broader issue here. When inline IPsec is enabled in pktio, it kind of has some dependency on IPsec configuration. IOW, if inline IPsec is enabled in pktio but is not enabled in IPsec config, inline IPsec config would fail. Should this lead to pktio_config returning an error or should the implementation just log an error and return success?

Even though it is not evident from the spec, for inline IPsec, there are two settings. One in pktio & another in IPsec. The one in pktio is more like a flag and IPsec settings add more details. Now, IPsec is a global configuration and it doesn't say inline IPsec need to be enabled on so and so pktio etc. So either we need to handle these in the implementation (which may not be perfect, as implementation wouldn't know if it's ipsec_inline_out test or a pktio test) or we need to have application only enable the features it needs.

But then, I agree that what the current test is doing is right. I hadn't looked at the specific test case when I commented earlier. But in the broader sense, we are trying to get feedback on what makes sense for such cases? In this case, we can have the test succeed by just ignoring the internal error altogether, but is that what we would want? In that case, inline IPsec wouldn't work even if pktio_config returns success.

JannePeltonen:
> But then, I agree that what the current test is doing is right.

So I propose closing this PR.

> IOW, if inline IPsec is enabled in pktio but is not enabled in IPsec config, inline IPsec config would fail

I suppose you meant: "non-inline IPsec config would fail".

I guess the API spec is not unambiguous on this but I would have expected async or sync mode IPsec config to succeed even if some pktios have inbound inline ipsec enabled.

anoobj:
> I guess the API spec is not unambiguous on this but I would have expected async or sync mode IPsec config to succeed even if some pktios have inbound inline ipsec enabled.

My question is about pktios having inline IPsec enabled but IPsec config op_mode says, let's say, SYNC. What should be the behavior in that case? I understand that pktio_config API should return success, but we cannot have the actual feature (inline processing of packets) happen, right? Is that the API expectation, when inline IPsec is enabled in pktio?

JannePeltonen:
I think the situation is different between inbound and outbound processing.

My interpretation of the outbound case is that it is ok to enable outbound_ipsec in a pktio but not enable it in ipsec config. An application cannot then do inline IPsec output but look-a-side IPsec should work, as it should even when inline output is enabled.

But for the inbound case I think the API leaves some ambiguity. One reasonable interpretation of the API text would be that ipsec config without inbound inline enabled can fail if any open pktio has inbound inline ipsec enabled and vice versa. But another, also reasonable, interpretation would be that it is ok to have inbound inline IPsec enabled in pktio but not in IPsec config and in that case inline ipsec processing would just not happen (as if the pktio did not have inline IPsec enabled).

There is also another question: What is supposed to happen to received packets when a pktio has inline IPsec enabled but IPsec config has not (yet?) been done at all. I think the least-surprise answer would be that incoming IPsec packets are processed normally as if the pktio did not have inline IPsec enabled. But perhaps some implementations would like to drop incoming IPsec packets in that case (even though once inline IPsec gets configured through the IPsec API, packets that do not match any SA are supposed to be received through the normal non-inline-IPsec path).

I am inclined to think that having packets processed normally through the non-inline-ipsec path when IPsec has not been configured at all or has not been configured in inline mode would be preferable to other choises, but I do not think the API really mandates that.

I am not sure the ambiguity in the API spec regarding conflicting pktio and IPsec config is a big problem because any reasonable application could just avoid using such config. But the question about the case when IPsec has not yet been configured may be relevant also for well behaving applications.

vvelumuri:
created a new request:  https://github.com/OpenDataPlane/odp/pull/1165

PR #1162, opened 2021-01-20
[PATCH v9] linux-gen: improve software checksum implementation

JereLeppanen:
```
linux-gen: improve software checksum implementation
    
    Improve software checksum implementation and put it in a separate
    header file so that it may be easily called from multiple places.
    
linux-gen: packet: use new software checksum functions
    
    Use new software checksum functions and clean up the related code.
    
```

v1:
- Draft.
- Probably need to revise usage of __ARM_FEATURE_UNALIGNED macro.

v2:
- Include "introduce _ODP_UNALIGNED" commit.
- Use _ODP_UNALIGNED instead of __ARM_FEATURE_UNALIGNED.

v3:
- Add arch/x86/odp_cpu.h in linux-generic/Makefile.am.
- Rebase.

v4:
- Add "linux-gen: include odp/api/cpu.h where odp_cpu_pause() is called".

v5:
- Add _ODP_UNALIGNED also for arch/arm.
- Draft -> open.

v6:
- De-doxygenify comments.
- Add validation tests for pseudorandom data and very long data.
- Rebase.

v7:
- Validation tests: Fix pointer casts and pointer mask types.

v8:
- Remove doze().
- Constant values as suggested by Petri.
- Test data lengths from 1 to over 9000 in chksum_ones_complement_pseudorandom().

v9:
- Add reviewed-by tags.

### DIFF
--- a/platform/linux-generic/arch/aarch64/odp_cpu_idling.h
+++ b/platform/linux-generic/arch/aarch64/odp_cpu_idling.h
@@ -30,12 +30,6 @@ static inline int wfe(void)
 	return 1;
 }
 
-static inline void doze(void)
-{
-	/* When using WFE do not stall the pipeline using other means */
-	odp_cpu_pause();
-}
-
 #define monitor128(addr, mo) lld((addr), (mo))
 #define monitor64(addr, mo) ll64((addr), (mo))
 #define monitor32(addr, mo) ll32((addr), (mo))
--- a/platform/linux-generic/include/odp_chksum_internal.h
+++ b/platform/linux-generic/include/odp_chksum_internal.h
@@ -1,0 +1,189 @@
+/* Copyright (c) 2020, Nokia
+ * All rights reserved.
+ *
+ * SPDX-License-Identifier:	BSD-3-Clause
+ */
+
+#ifndef ODP_CHKSUM_INTERNAL_H_
+#define ODP_CHKSUM_INTERNAL_H_
+
+#ifdef __cplusplus
+extern "C" {
+#endif
+
+#include <odp/api/hints.h>
+#include <odp/api/byteorder.h>
+#include <odp_cpu.h>
+#include <stdint.h>
+
+/*
+ * Compute the final Internet checksum (RFC 1071) based on a partial
+ * sum. A partial sum can be obtained e.g. by calling
+ * chksum_partial().
+ */
+static inline uint16_t chksum_finalize(uint64_t sum)
+{
+	sum = (sum >> 32) + (sum & 0xffffffff);
+	sum = (sum >> 16) + (sum & 0xffff);
+	/*
+	 * The final & 0xffff is intentionally omitted, the extra bits
+	 * are discarded by the implicit cast to the return type.
+	 */
+	return (sum >> 16) + sum;
+}
+
+/*
+ * Compute a partial checksum. Several partial checksums may be summed
+ * together. The final checksum may be obtained by calling
+ * chksum_finalize(). Parameter offset is the offset of this segment
+ * of data from the start of IP header.
+ *
+ * This implementation
+ *
+ * - Accepts unaligned data.
+ *
+ * - Accepts data at any byte offset from the start of IP header,
+ *   including odd offsets.
+ *
+ * - Uses unaligned memory access only if available.
+ *
+ * - Is optimized (for skylake, cn96, a53) by trial and error.
+ *
+ * The following did not improve performance (in synthetic tests):
+ *
+ * - 2 or 4 sub-sums in the main loop (to break dependency chains).
+ *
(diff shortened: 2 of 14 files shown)


platform/linux-generic/include/odp_chksum_internal.h
```diff
@@ -0,0 +1,201 @@
+/* Copyright (c) 2020, Nokia
+ * All rights reserved.
+ *
+ * SPDX-License-Identifier:	BSD-3-Clause
+ */
+
+/**
+ * @file
+ *
```
psavol, line 9: Only API files (and maybe some examples) generate doxygen documentation. Implementation internal files do not need doxygen tags (and /**). You can test with "make doxygen-doc" that there's no warnings after removing tags.

psavol:
Looks good.

Our validation test for odp_chksum_ones_comp16() is 4 byte aligned (validation/api/chksum). Test cases of e.g. 2 byte aligned data could be added, as it's pretty common alignment for an incoming IP packet (over ethernet).  

static uint8_t ip_hdr_test_vect[NUM_IP_HDR][IP_HDR_LEN] ODP_ALIGNED(4)

platform/linux-generic/include/odp_chksum_internal.h
JereLeppanen: That's not a reason to not use doxygen comments, though. I usually use doxygen comments just to conform to a known format. Whether some documentation is actually generated from it is a different matter. Also, there are plenty of doxygen comments in linux-gen.

So the question isn't whether it's needed, but whether it's allowed.

JereLeppanen:
> Our validation test for odp_chksum_ones_comp16() is 4 byte aligned (validation/api/chksum). Test cases of e.g. 2 byte aligned data could be added

I can have a look at that.

platform/linux-generic/include/odp_chksum_internal.h
psavol: I agree that the format is known, but in general we should avoid adding unnecessary lines or code that is not tested by CI. When our build does not run doxygen on these internal files, tags will be broken during future edits of these files. When people tend to create new files by copying the old ones, broken tags will spread further, and in the end there would be many files that look like having proper doxygen documentation, but if you run doxygen on those the end result would be  errors and potentially no useful documentation.

Usually, I remove extra doxygen tags from internal files while editing those.

platform/linux-generic/arch/aarch64/odp_cpu_idling.h
```diff
@@ -19,6 +19,8 @@
 
 #else /* CONFIG_WFE */
 
+#include <odp/api/cpu.h>
+
```
psavol, line 23: It's better to remove doze() altogether, since it's always implemented as odp_cpu_pause(). Doze is called only in two files: odp_queue_scalable.c and odp_schedule_scalable.c

test/validation/api/chksum/chksum.c
```diff
@@ -313,10 +313,87 @@ static void chksum_ones_complement_udp_long(void)
...
+#define CNG (z = 6906969069LL * z + 1234567)
+#define KISS (MWC + XSH + CNG)
+
+/*
+ * Test relatively short pseudorandom data. This test is trying to maximize the
+ * number of checksums computed.
+ */
+static void chksum_ones_complement_pseudorandom(void)
+{
+	const int size = (1 << 15);
```
psavol, line 361: I'd change this to e.g. 32 * 1024 for better code readability.

test/validation/api/chksum/chksum.c
```diff
@@ -313,10 +313,87 @@ static void chksum_ones_complement_udp_long(void)
...
+	const int max_len = 64;
```
psavol, line 362: This is rather short length for typical UDP/TCP checksums. If implementation has different code paths for e.g. short and long packets, the long packet code would be not tested. Maybe the test case could have couple of loops for different max_len values: e.g. 1M x 64B, 100k x 640B, 10k x 9000B

It's OK for this kind of validation test to run for some number of seconds.

test/validation/api/chksum/chksum.c
```diff
@@ -389,11 +389,43 @@ static void chksum_ones_complement_pseudorandom(void)
 	free(buf);
 }
 
+/*
+ * Test with very long data with most of the bits set. The idea is to
+ * maximize the number of carries.
+ */
+static void chksum_ones_complement_very_long(void)
+{
+	const int size = (1 << 16);
```
psavol, line 398: I'd prefer: size = 64 * 1024

test/validation/api/chksum/chksum.c
JereLeppanen: I ended up with a different solution, but this now tests a wide range of data lengths. Please check.

psavol (approved):

PR #1163, opened 2021-01-21
[PATCH v1] linux-gen: introduce _ODP_UNALIGNED macro

JereLeppanen:
The _ODP_UNALIGNED pre-processor macro is defined if the CPU features
efficient unaligned memory access.

### DIFF
--- a/platform/linux-generic/arch/aarch64/odp_cpu.h
+++ b/platform/linux-generic/arch/aarch64/odp_cpu.h
@@ -59,4 +59,8 @@ do {							     \
 #include "odp_atomic.h"
 #include "odp_cpu_idling.h"
 
+#ifdef __ARM_FEATURE_UNALIGNED
+#define _ODP_UNALIGNED
+#endif
+
 #endif  /* PLATFORM_LINUXGENERIC_ARCH_ARM_ODP_CPU_H */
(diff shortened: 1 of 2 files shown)


matiaselo:
Looks like a reasonable addition. The new header `platform/linux-generic/arch/x86/odp_cpu.h` is missing from `platform/linux-generic/Makefile.am`. This commit would be easier to review and test as a part of PR #1162, where this change is actually needed.

platform/linux-generic/arch/aarch64/odp_cpu.h
```diff
@@ -59,4 +59,8 @@ do {							     \
 #include "odp_atomic.h"
 #include "odp_cpu_idling.h"
 
+#ifdef __ARM_FEATURE_UNALIGNED
+#define _ODP_UNALIGNED
```
psavol, line 63: This should be defined with value 0 or 1. We try to avoid "#ifdef FOO" and use if(FOO) instead. This way both if and else branches are always compiled and are less likely to break build later on. Optimizer removes the unused branch, so there should not be performance impact.

JereLeppanen:
I've included this patch in PR #1162, with changes according to the comments made here. Closing this PR.

PR #1164, opened 2021-01-25
[PATCH v6] example: ipsec_crypto: use comma as separator in args

asasidharan:
As part of -i option, interface name could be in DBDF format to
represent the device. Application will throw error during argument
parsing as DBDF format itself contains colon(:). So updating parsing
logic to use comma(,) as a separator instead of colon(:).


JannePeltonen:
What does DBDF stand for?

Changing the separator from colon to comma looks reasonable to me even though there is no guarantee that a pktio name would not contain a comma. Comma works better with MAC addresses and IPv6 addresses.

The example command in README should be updated to use comma in place of colon.

I would be good to make the same change in ipsec_offload example too. The parsing code seems to be almost the same.

I would be nice to change the MAC address presentation to use colons now that colon is no longer reserved for field separation.

asasidharan:
> 
> 
> What does DBDF stand for?

DBDF stands for Domain, Bus, Device and Function for a PCIe device (eg. 0000:00:10.0). Some platforms like Marvell's cn10k uses PCIe DBDF to represent pktio interface name. Hence the need for this change.

> 
> The example command in README should be updated to use comma in place of colon.

Ack.

> I would be nice to change the MAC address presentation to use colons now that colon is no longer reserved for field separation.

Ack.

> 
> I would be good to make the same change in ipsec_offload example too. The parsing code seems to be almost the same.
> 

I haven't updated this PR yet for ipsec_offload example application. The changes for ipsec_offload example application seems to be straight forward, but I was not able to successfully run the application on linux-generic platform even without any change.

I was not able to find a separate README file for this application to see if there is any additional dependencies to get the ipsec_offload application running.
I could see that the run script uses dpni.0 and dpni.1 as the pktio interfaces. Does that mean we have dependency on some other framework for this application to run properly?

On a similar note, I observed an issue while running the ipsec_api example application using the script for ESP inbound standalone test. I had opened a github issue for this since I was not sure where I should describe the same: https://github.com/OpenDataPlane/odp/issues/1167.

example/ipsec_offload/run_left.sh
```diff
@@ -2,13 +2,15 @@
 #
 
 ./odp_ipsec_offload -f 64 -i dpni.1,dpni.2 \
```
psavol, line 4: As discussed, interface names should be generic (loop or virtual interfaces). Let's check first if the application is duplicate of other ipsec test applications.

matiaselo: The ipsec_offload application code seems to be mostly duplicate from ipsec_api and the app hasn't event started properly for the past two years (odp_schedule_config() is called too late). So, I thing it's safe to remove the whole application.

JannePeltonen: I looked at the ipsec_offload example and agree that we should just remove it. So let's not bother to change its parameter format as part of this PR.

JannePeltonen:
usage() functions of ipsec_api and ipsec_crypto still show the old colon based format.

There are several comments that still mention the old format (grep "String is of the format" to find them).

example/ipsec_offload/run_left.sh
asasidharan: Ack. Dropping the changes to ipsec_offload application from this PR.

asasidharan:
> 
> 
> usage() functions of ipsec_api and ipsec_crypto still show the old colon based format.
> 
> There are several comments that still mention the old format (grep "String is of the format" to find them).

Thanks for pointing this out, missed to notice those. Updated the usage and the comments to reflect the updated format with v5.

JannePeltonen (approved):

asasidharan:
Updated v6 with Reviewed-by tag. Thank you @JannePeltonen , @psavol  and @matiaselo for reviewing this PR.

> I looked at the ipsec_offload example and agree that we should just remove it.

Also wanted to confirm on how this should be done. Will you be creating a separate pull request for removing the ipsec_offload application? Or do you want me to do that?

psavol:
> 
> 
> Updated v6 with Reviewed-by tag. Thank you @JannePeltonen , @psavol and @matiaselo for reviewing this PR.
> 
> > I looked at the ipsec_offload example and agree that we should just remove it.
> 
> Also wanted to confirm on how this should be done. Will you be creating a separate pull request for removing the ipsec_offload application? Or do you want me to do that?

Please send a new PR for removal of ipsec_offload. I'll merge this one.

PR #1165, opened 2021-01-27
[PATCH v4] api: pktio: allow inline inbound to be enabled only if IPsec config is done

anoobj:


JannePeltonen:
If we add this restriction, it would still be ok to do the following steps:

- open a pktio with inline IPsec disabled
- receive packets through the pktio
- configure IPsec with inbound inline enabled
- close the pktio
- open the pktio with inline IPsec enabled

anoobj:
> 
> 
> If we add this restriction, it would still be ok to do the following steps:
> 
>     * open a pktio with inline IPsec disabled
> 
>     * receive packets through the pktio
> 
>     * configure IPsec with inbound inline enabled
> 
>     * close the pktio
> 
>     * open the pktio with inline IPsec enabled
> 
> 
> Would this be ok in your implementation?

Yes. That should be fine.

JannePeltonen:
Looks ok to me.

The API text appears also in doc/users-guide/users-guide-pktio.adoc. It would be good to change the text there too. It probably needs to be a separate commit to keep API commits clean from other stuff.

JannePeltonen (approved):

anoobj:
@JannePeltonen There was a typo in the commit header of the last patch. v4 is to address that.

JannePeltonen:
> There was a typo in the commit header of the last patch. v4 is to address that.

v4 looks ok.

PR #1166, opened 2021-01-27
validation: pktio: fix segment length pool param

harman-kalra:
During pool creation segment length (param.pkt.seg_len) gets
initialised with max segment length possible (which could go
up to max packet length supported in case of single segmented
buffer), while packet length (param.pkt.len) value gets
initialised with PKT_BUF_SIZE. With this initialisation
seg length becomes much greater than packet length i.e.
(param.pkt.seg_len > param.pkt.len).

With given pool params, expectation is to support param.pkt.num
packets of param.pkt.len with first buffer segment size as
param.pkt.seg_len, which is not feasible.


matiaselo (approved):
Looks good. Please change the commit subject line to:
`validation: pktio: fix segment length pool param`

The first commit message paragraph is also missing final period.

PR #1170, opened 2021-02-01
[PATCH v9] helper: add CLI helper API and implementation

JereLeppanen:
```
helper: add CLI helper API and implementation
    
    CLI helper API allows starting and stopping ODP CLI server, which may
    be connected to using a telnet client. CLI commands may be used to get
    information from an ODP instance, for debugging purposes.
    
    This implementation depends on libcli. If libcli is not available, CLI
    helper is automatically excluded.
    
helper: test: implement CLI helper test
    
    Test CLI helper by starting and stopping the ODP CLI server.
    
example: add CLI helper example
    
    This example shows how to start and stop ODP CLI using the CLI helper
    API functions.
    
```
v2:
- Fix missing initializer.
- Add configure option --with-libcli-path.

v3:
- Fix missing initializer even more.

v4:
- Add LIBCLI_LIBS to Libs.private in helper/libodphelper.pc.in.
- example: Stay open indefinitely by default.
- example: Add odp_cli_run.sh script.
- example: Print "CLI server started" message.
- example: Rename option "num" to "time".
- example: Use odph_options().
- example: Use ODPH_ERR() and exit(EXIT_FAILURE).
- Define ODPH_IPV4ADDRSTR_LEN and ODPH_IPV6ADDRSTR_LEN in ip.h.
- Document default values in odph_cli_param_init().
- Document thread creation and termination.
- Minor CLI helper API documentation cleanups.
- Use shm.

v5:
- example: Add odp_cli_run.sh to dist.

v6:
- Create autoheader_external for helper.
- Include cli.h conditionally in odph_api.h.
- Move documentation of default values from *_param_init() to *_param_t, according to convention.

v7:
- Fix various linking issues with --with-libcli-path option.
- Fail is --with-libcli-path option given and libcli not found.
- Add help string "<name>" to commands that have that parameter.
- Change odph_cli_param_t.address to const char *.
- Rebase.

v8:
- odp_cli_run.sh: Run cli server for two seconds.
- configure: Move libcli logic to a separate libcli.m4 file.
- example/cli: Add signal handler for SIGINT.
- helper/cli.c: Check return value of odp_shm_lookup() calls.
- Increment libodphelper version.
- example/cli: Print a message before stopping cli server.
- helper/cli.c: Free shm in odph_cli_stop().

### DIFF
--- a/example/cli/odp_cli.c
+++ b/example/cli/odp_cli.c
@@ -1,0 +1,176 @@
+/* Copyright (c) 2021, Nokia
+ * All rights reserved.
+ *
+ * SPDX-License-Identifier:     BSD-3-Clause
+ */
+
+/*
+ * ODP CLI Helper Example
+ *
+ * This example shows how to start and stop ODP CLI using the CLI helper
+ * API functions. This example application can also be used to try out
+ * the CLI by connecting to a running application with a telnet client.
+ */
+
+#include <odp_api.h>
+#include <odp/helper/odph_api.h>
+
+#include <stdio.h>
+#include <stdint.h>
+#include <signal.h>
+
+typedef struct {
+	int time;
+	char *addr;
+	uint16_t port;
+} options_t;
+
+static void usage(const char *prog)
+{
+	printf("\n"
+	       "Usage: %s [options]\n"
+	       "\n"
+	       "OPTIONS:\n"
+	       "  -t, --time <sec>        Keep CLI open for <sec> seconds. (default -1 (infinite))\n"
+	       "  -a, --address <addr>    Bind listening socket to IP address <addr>.\n"
+	       "  -p, --port <port>       Bind listening socket to port <port>.\n"
+	       "\n"
+	       "ODP helper defaults are used for address and port, if the options are\n"
+	       "not given.\n"
+	       "\n",
+	       prog);
+}
+
+static void parse_args(int argc, char *argv[], options_t *opt)
+{
+	static const struct option longopts[] = {
+		{ "time", required_argument, NULL, 't' },
+		{ "address", required_argument, NULL, 'a' },
+		{ "port", required_argument, NULL, 'p' },
+		{ "help", no_argument, NULL, 'h' },
+		{ NULL, 0, NULL, 0 }
+	};
+
+	static const char *shortopts = "+t:a:p:h";
+
+	while (1) {
+		int c = getopt_long(argc, argv, shortopts, longopts, NULL);
+
+		if (c == -1)
+			break; /* No more options */
+
+		switch (c) {
+		case 't':
+			opt->time = atoi(optarg);
+			break;
+		case 'a':
+			opt->addr = optarg;
+			break;
+		case 'p':
+			opt->port = atoi(optarg);
+			break;
+		default:
+			usage(argv[0]);
+			exit(EXIT_SUCCESS);
+			break;
+		}
+	}
+
+	optind = 1; /* reset 'extern optind' from the getopt lib */
+}
+
+static volatile int shutdown_sig;
+
+static void sig_handler(int signo)
+{
+	(void)signo;
+
+	shutdown_sig = 1;
+}
+
+int main(int argc, char *argv[])
+{
+	signal(SIGINT, sig_handler);
+
+	odph_helper_options_t helper_options;
+
+	/* Let helper collect its own arguments (e.g. --odph_proc) */
+	argc = odph_parse_options(argc, argv);
+	if (odph_options(&helper_options)) {
+		ODPH_ERR("Error: reading ODP helper options failed.\n");
+		exit(EXIT_FAILURE);
+	}
+
+	odp_init_t init;
+
+	odp_init_param_init(&init);
+	init.mem_model = helper_options.mem_model;
+
+	options_t opt = {
+		.time = -1,
+		.addr = NULL,
+		.port = 0,
+	};
+
+	parse_args(argc, argv, &opt);
+
+	/* Initialize ODP. */
+
+	odp_instance_t inst;
+
+	if (odp_init_global(&inst, &init, NULL)) {
+		ODPH_ERR("Global init failed.\n");
+		exit(EXIT_FAILURE);
+	}
+
+	if (odp_init_local(inst, ODP_THREAD_CONTROL)) {
+		ODPH_ERR("Local init failed.\n");
+		exit(EXIT_FAILURE);
+	}
+
+	/* Prepare CLI parameters. */
+
+	odph_cli_param_t cli_param;
+
+	odph_cli_param_init(&cli_param);
+
+	if (opt.addr)
+		cli_param.address = opt.addr;
+
+	if (opt.port)
+		cli_param.port = opt.port;
+
+	/* Start CLI server. */
+	if (odph_cli_start(inst, &cli_param)) {
+		ODPH_ERR("CLI start failed.\n");
+		exit(EXIT_FAILURE);
+	}
+
+	printf("CLI server started on %s:%d\n", cli_param.address, cli_param.port);
+
+	/* Wait for the given number of seconds. */
+	for (int i = 0; (opt.time < 0 || i < opt.time) && !shutdown_sig; i++)
+		odp_time_wait_ns(ODP_TIME_SEC_IN_NS);
+
+	printf("Stopping CLI server.\n");
+
+	/* Stop CLI server. */
+	if (odph_cli_stop()) {
+		ODPH_ERR("CLI stop failed.\n");
+		exit(EXIT_FAILURE);
+	}
+
+	/* Terminate ODP. */
+
+	if (odp_term_local()) {
+		ODPH_ERR("Local term failed.\n");
+		exit(EXIT_FAILURE);
+	}
+
+	if (odp_term_global(inst)) {
+		ODPH_ERR("Global term failed.\n");
+		exit(EXIT_FAILURE);
+	}
+
+	return 0;
+}
--- a/helper/cli.c
+++ b/helper/cli.c
@@ -1,0 +1,499 @@
+/* Copyright (c) 2021, Nokia
+ * All rights reserved.
+ *
+ * SPDX-License-Identifier:     BSD-3-Clause
+ */
+
+#include <odp/helper/cli.h>
+#include <odp_api.h>
+#include <odp/helper/odph_api.h>
+#include <libcli.h>
+#include <sys/types.h>
+#include <sys/socket.h>
+#include <arpa/inet.h>
+#include <netinet/in.h>
+#include <unistd.h>
+#include <errno.h>
+#include <poll.h>
+
+/* Socketpair socket roles. */
+enum {
+	SP_READ = 0,
+	SP_WRITE = 1,
+};
+
+typedef struct {
+	volatile int cli_fd;
+	/* Server thread will exit if this is false. */
+	volatile int run;
+	/* Socketpair descriptors. */
+	int sp[2];
+	int listen_fd;
+	/* This lock guards cli_fd and run, which must be accessed atomically. */
+	odp_spinlock_t lock;
+	odph_thread_t thr_server;
+} cli_shm_t;
+
+static const char *shm_name = "_odp_cli";
+
+static const odph_cli_param_t param_default = {
+	.address = "127.0.0.1",
+	.port = 55555,
+};
+
+void odph_cli_param_init(odph_cli_param_t *param)
+{
+	*param = param_default;
+}
+
+static int cmd_show_cpu(struct cli_def *cli, const char *command ODP_UNUSED,
+			char *argv[] ODP_UNUSED, int argc ODP_UNUSED)
+{
+	for (int c = 0; c < odp_cpu_count(); c++) {
+		cli_print(cli, "% 4d: %s %.03f / %.03f GHz", c,
+			  odp_cpu_model_str_id(c),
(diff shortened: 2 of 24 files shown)


helper/m4/configure.m4
```diff
@@ -17,5 +17,18 @@ AC_ARG_ENABLE([helper-linux],
...
+#########################################################################
+# If libcli is available, enable CLI helper
+#########################################################################
+helper_cli=no
+LIBCLI_LIBS=""
+AC_CHECK_HEADER(libcli.h,
+    [AC_CHECK_LIB(cli, cli_init,
+                  [helper_cli=yes; LIBCLI_LIBS="-lcli"],
+                  [])],
+    [])
```
matiaselo, line 29: For cross-compilation it would be useful to have a configure option for passing a custom libcli install path (e.g. `--with-libcli-path`).

helper/m4/configure.m4
```diff
@@ -17,5 +17,47 @@ AC_ARG_ENABLE([helper-linux],
...
+#########################################################################
+# If libcli is available, enable CLI helper
+#########################################################################
+helper_cli=no
+AC_CHECK_HEADER(libcli.h,
+    [AC_CHECK_LIB(cli, cli_init, [helper_cli=yes], [])],
+    [])
+
+if test "$helper_cli" = "yes"; then
+    LIBCLI_LIBS="$LIBCLI_LIBS -lcli"
```
matiaselo, line 47: I think `LIBCLI_LIBS` should be added to Libs.private in `helper/libodphelper.pc.in`. You can check `platform/linux-generic/libodp-linux.pc.in` for examples.

example/cli/odp_cli.c
```diff
@@ -0,0 +1,145 @@
...
+
+	optind = 1; /* reset 'extern optind' from the getopt lib */
+}
+
+int main(int argc, char *argv[])
+{
+	odp_instance_t inst;
+	options_t opt;
+	int i;
+
```
matiaselo, line 87: Since `odph_cli_start()` creates a new worker, it could be beneficial to support also process mode.

E.g. copy-paste from odp_l2fwd:
```
	odph_helper_options_t helper_options;

	/* Let helper collect its own arguments (e.g. --odph_proc) */
	argc = odph_parse_options(argc, argv);
	if (odph_options(&helper_options)) {
		ODPH_ERR("Error: reading ODP helper options failed.\n");
		exit(EXIT_FAILURE);
	}

	odp_init_param_init(&init);

	init.mem_model = helper_options.mem_model;

	if (odp_init_global(&instance, &init, NULL)) {
...
```

example/cli/odp_cli.c
```diff
@@ -0,0 +1,145 @@
...
+	memset(&opt, 0, sizeof(opt));
+	parse_args(argc, argv, &opt);
+
+	/* Initialize ODP. */
+
+	if (odp_init_global(&inst, NULL, NULL)) {
+		printf("Global init failed.\n");
+		return -1;
```
matiaselo, line 95: Please use `ODPH_ERR()` macro for error messages and exit with `exit(EXIT_FAILURE)`.

helper/include/odp/helper/cli.h
```diff
@@ -0,0 +1,84 @@
...
+
+/**
+ * @addtogroup odph_cli ODPH CLI
+ * @{
+ */
+
+/** ODP CLI server parameters. */
+typedef struct {
+	/** A character string containing an IP address. */
+	char address[INET6_ADDRSTRLEN];
```
matiaselo, line 39: I'd like to avoid having to include the `netinet/in.h` header. Instead, this be `const char *` or `ODPH_IPV6ADDRSTR_LEN` could be defined in `helper/include/odp/helper/ip.h.

helper/include/odp/helper/cli.h
```diff
@@ -0,0 +1,84 @@
...
+	/** TCP port. */
+	uint16_t port;
```
matiaselo, line 41: The default values for address and port could be documented here?

helper/include/odp/helper/cli.h
```diff
@@ -0,0 +1,84 @@
...
+ *
+ * @param[out] param Pointer to parameter structure.
+ */
+void odph_cli_param_init(odph_cli_param_t *param);
+
+/**
+ * Start CLI server.
+ *
+ * Upon successful return from this function, the CLI server will be
+ * accepting client connections.
```
matiaselo, line 55: It would be good to clarify that this function spawns a new thread of type `ODP_THREAD_CONTROL` using `odp_cpumask_default_control()`.

helper/include/odp/helper/cli.h
```diff
@@ -0,0 +1,84 @@
...
+ * @retval <0 Failure.
+ */
+int odph_cli_start(const odp_instance_t instance,
+		   const odph_cli_param_t *param);
+
+/**
+ * Stop CLI server.
+ *
+ * Stop accepting new client connections and disconnect currently connected
+ * client.
```
matiaselo, line 69: Related to the previous comment, the documentation could state that the control thread created in `odph_cli_start()` is terminated here.

helper/include/odp/helper/cli.h
```diff
@@ -0,0 +1,84 @@
...
+/** ODP CLI server parameters. */
+typedef struct {
+	/** A character string containing an IP address. */
+	char address[INET6_ADDRSTRLEN];
+	/** TCP port. */
+	uint16_t port;
+} odph_cli_param_t;
+
+/**
+ * Initialize CLI server parameter structure.
```
matiaselo, line 45: To be consistent with the curren *_init() APIs:
```
/**
 * Initialize CLI server params
 *
 * Initialize an odph_cli_param_t to its default values for all fields.
```

helper/include/odp/helper/cli.h
```diff
@@ -0,0 +1,84 @@
+/* Copyright (c) 2021, Nokia
+ * All rights reserved.
+ *
+ * SPDX-License-Identifier:     BSD-3-Clause
+ */
+
+/**
+ * @file
+ *
+ * ODP CLI helper API
```
matiaselo, line 10: This header should be included in `helper/include/odp/helper/odph_api.h`. You can wrap the include with e.g. `#ifdef _ODPH_CLI` (see `_ODP_PKTIO_NETMAP` for example).

helper/cli.c
```diff
@@ -0,0 +1,462 @@
...
+/* Socketpair descriptors. */
+static int sp[2];
+static int listen_fd;
+static volatile int cli_fd;
+/* Server thread will exit if this is false. */
+static volatile int run;
+/* This lock guards cli_fd and run, which must be accessed atomically. */
+static odp_spinlock_t lock;
+static odph_thread_t thr_server;
+static odph_cli_param_t param;
```
matiaselo, line 33: Please reserve a single shm block and put all global variables inside it.

example/cli/odp_cli.c
```diff
@@ -0,0 +1,145 @@
...
+	uint16_t port;
+} options_t;
+
+static void usage(const char *prog)
+{
+	printf("\n"
+	       "Usage: %s [options]\n"
+	       "\n"
+	       "OPTIONS:\n"
+	       "  -n, --num <sec>         Keep CLI open for <sec> seconds. (default 0)\n"
```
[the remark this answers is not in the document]
JereLeppanen: I wanted to avoid creating yet another useless script, but OK, I'll make the change.

example/cli/odp_cli.c
```diff
@@ -0,0 +1,145 @@
...
+	       "ODP helper defaults are used for address and port, if the options are\n"
+	       "not given.\n"
+	       "\n",
+	       prog);
+}
+
+static void parse_args(int argc, char *argv[], options_t *opt)
+{
+	static const struct option longopts[] = {
+		{ "num", required_argument, NULL, 'n' },
```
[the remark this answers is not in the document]
JereLeppanen: I'll change it. I copied this from example/hello.

example/cli/odp_cli.c
JereLeppanen: Again, copied this from example/hello. I'll change it.

helper/include/odp/helper/cli.h
JereLeppanen: _ODP_PKTIO_NETMAP is probably not a good template here, since it's in an ODP autoheader. Helper doesn't have an autoheader. Should we create an autoheader for ODP helper?

helper/include/odp/helper/cli.h
JereLeppanen: I'm putting the defines (for both v4 and v6) in ip.h. Trying to avoid passing pointers, although it probably doesn't matter much.

helper/include/odp/helper/cli.h
JereLeppanen: Moved these from odph_cli_param_init() to here. Seems to be a convention.

helper/cli.c
```diff
@@ -0,0 +1,484 @@
...
+
+	c = cli_register_command(cli, NULL, "call", NULL,
+				 PRIVILEGE_UNPRIVILEGED, MODE_EXEC,
+				 "Call ODP API function.");
+	cli_register_command(cli, c, "odp_ipsec_print",
+			     cmd_call_odp_ipsec_print,
+			     PRIVILEGE_UNPRIVILEGED, MODE_EXEC, NULL);
+	cli_register_command(cli, c, "odp_pktio_print",
+			     cmd_call_odp_pktio_print,
+			     PRIVILEGE_UNPRIVILEGED, MODE_EXEC, NULL);
```
matiaselo, line 245: Help strings are missing from the API calls. Also, in the help string it would be good to mention when an object name argument is required.

helper/include/odp/helper/cli.h
```diff
@@ -0,0 +1,92 @@
...
+ * @{
+ */
+
+/** ODP CLI server parameters */
+typedef struct {
+	/**
+	 * A character string containing an IP address. Default is
+	 * "127.0.0.1".
+	 */
+	char address[ODPH_IPV6ADDRSTR_LEN];
```
psavol, line 42: This could be "const char *address" instead of table. It would be easier for application to give pointer to string than copy it. Helper needs to copy the address anyway (if it needs to save it for later use). This API spec can require that max string length is ODPH_IPV6ADDRSTR_LEN for IPv6 and ODPH_IPV4ADDRSTR_LEN for IPv4.

helper/cli.c
JereLeppanen: The help string is not exactly the correct place for parameter help, but since there is no place for that in libcli 1.9 (while apparently there is a place in libcli 1.10), I will add the help string "<name>" to all the commands that have that parameter. It doesn't look right, but perhaps it's better than nothing.

With the help strings in place, due to a bug(?) in some versions of libcli, the user may see different error messages ("incomplete command" vs. "name not found") when typing "<command> ?", but I don't think this is a problem.

helper/include/odp/helper/cli.h
JereLeppanen: I thought it might be a bad idea to return a pointer to string in odph_cli_param_init(), but okay. At least currently cli helper doesn't need a copy of the string, so with the parameter changed to a pointer, there's no need for any length information anywhere, and no need to introduce the ODPH_IPV?ADDRSTR_LEN defines either. So I'm dropping that commit.

helper/cli.c
```diff
@@ -0,0 +1,482 @@
...
+	cli_register_command(cli, c, "odp_sys_info_print",
+			     cmd_call_odp_sys_info_print,
+			     PRIVILEGE_UNPRIVILEGED, MODE_EXEC, NULL);
+
+	return cli;
+}
+
+static int cli_server(void *arg ODP_UNUSED)
+{
+	cli_shm_t *shm = (cli_shm_t *)odp_shm_addr(odp_shm_lookup(shm_name));
```
matiaselo, line 269: The return value of `odp_shm_lookup()` should be checked first. if I remember correctly, Coverity will complain otherwise about unchecked return value.

helper/cli.c
```diff
@@ -0,0 +1,482 @@
...
+
+int odph_cli_start(const odp_instance_t instance,
+		   const odph_cli_param_t *param_in)
+{
+	if (odp_shm_lookup(shm_name) != ODP_SHM_INVALID) {
+		ODPH_ERR("Error: cli server already running (shm %s exists)\n", shm_name);
+		return -1;
+	}
+
+	cli_shm_t *shm = (cli_shm_t *)odp_shm_addr(odp_shm_reserve(shm_name,
```
matiaselo, line 346: The return value of `odp_shm_reserve()` should be checked first.

helper/cli.c
```diff
@@ -0,0 +1,482 @@
...
+	close(shm->sp[SP_WRITE]);
+	close(shm->listen_fd);
+	close(shm->cli_fd);
+	shm->run = 0;
+	return -1;
+}
+
+int odph_cli_stop(void)
+{
+	cli_shm_t *shm = (cli_shm_t *)odp_shm_addr(odp_shm_lookup(shm_name));
```
matiaselo, line 444: The return value of `odp_shm_lookup()` should be checked first.

matiaselo (commented):
Some minor comments but otherwise the PR looking good. I also tested cross-compilation on arm64 and everything worked as expected.

matiaselo:
Also, the helper version number should be incremented to 1.0.6 in a separate final commit in this PR.

matiaselo (approved):

PR #1174, opened 2021-02-08
[PATCH v2] example: debug: add debug print calls

psavol:
Add calls to odp_pktio_print(), odp_packet_print_data(), odp_ipsec_print() and odp_sys_config_print().

### DIFF
--- a/example/debug/odp_debug.c
+++ b/example/debug/odp_debug.c
@@ -1,20 +1,24 @@
-/* Copyright (c) 2020, Nokia
+/* Copyright (c) 2020-2021, Nokia
  * All rights reserved.
  *
  * SPDX-License-Identifier:     BSD-3-Clause
  */
 
+#include <stdlib.h>
 #include <stdio.h>
 #include <string.h>
 #include <getopt.h>
 
 #include <odp_api.h>
+#include <odp/helper/odph_api.h>
 
 typedef struct test_global_t {
+	int system;
 	int shm;
-	int shm_all;
 	int pool;
 	int queue;
+	int pktio;
+	int ipsec;
 
 } test_global_t;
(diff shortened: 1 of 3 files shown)


example/debug/odp_debug.c
```diff
@@ -285,49 +345,67 @@ int main(int argc, char *argv[])
...
+		global->ipsec   = 1;
 	} else {
 		if (parse_options(argc, argv, global))
 			return -1;
 	}
 
 	if (odp_init_global(&inst, NULL, NULL)) {
-		printf("Global init failed.\n");
+		ODPH_ERR("Global init failed.\n");
 		return -1;
```
matiaselo, line 361: While your are cleaning the error messages, you could also use standard return codes. E.g. `exit(EXIT_FAILURE)`.

example/debug/odp_debug.c
```diff
@@ -260,21 +271,70 @@ static int queue_debug(void)
...
+	}
+
+	return 0;
+}
+
+static int ipsec_debug(void)
+{
+	printf("\n");
+	odp_ipsec_print();
+
```
matiaselo, line 334: When have also`odp_ipsec_sa_print()`, but support for this can be added in a separate commit.

matiaselo (approved):

example/debug/odp_debug.c
psavol: Yes, left that for the next round. It needs considerable number of lines (to check capabilities) to successfully create a SA of any supported type.

example/debug/odp_debug.c
psavol: OK. I'll do that as well.

PR #1176, opened 2021-02-11
[PATCH v6] validation: ipsec: add test routine to verify udp_encap

anoobj:
Add test routine to verify udp_encap with tunnel mode
for IPv4 and IPv6 for various algorithms

Signed-off-by: Narayana Prasad <pathreya@marvell.com>

### DIFF
--- a/test/validation/api/ipsec/ipsec.c
+++ b/test/validation/api/ipsec/ipsec.c
@@ -18,7 +18,6 @@
 struct suite_context_s suite_context;
 
 #define PKT_POOL_NUM  64
-#define PKT_POOL_LEN  (1 * 1024)
 
 #define PACKET_USER_PTR	((void *)0x1212fefe)
 #define IPSEC_SA_CTX	((void *)0xfefefafa)
@@ -817,6 +816,35 @@ static void ipsec_pkt_auth_err_set(odp_packet_t pkt)
 	odp_packet_copy_from_mem(pkt, len - sizeof(data), sizeof(data), &data);
 }
 
+static void ipsec_pkt_v4_check_udp_encap(odp_packet_t pkt)
+{
+	uint32_t l3_off = odp_packet_l3_offset(pkt);
+	odph_ipv4hdr_t ip;
+
+	odp_packet_copy_to_mem(pkt, l3_off, sizeof(ip), &ip);
+
+	CU_ASSERT(ip.proto == ODPH_IPPROTO_UDP);
+}
+
+static void ipsec_pkt_v6_check_udp_encap(odp_packet_t pkt)
+{
+	uint32_t l3_off = odp_packet_l3_offset(pkt);
+	odph_ipv6hdr_ext_t ext;
+	odph_ipv6hdr_t ip;
+	uint8_t next_hdr;
+
+	odp_packet_copy_to_mem(pkt, l3_off, sizeof(ip), &ip);
+
+	next_hdr = ip.next_hdr;
+
+	if (ip.next_hdr == ODPH_IPPROTO_HOPOPTS) {
+		odp_packet_copy_to_mem(pkt, l3_off + sizeof(ip), sizeof(ext), &ext);
+		next_hdr = ext.next_hdr;
+	}
+
+	CU_ASSERT(next_hdr == ODPH_IPPROTO_UDP);
+}
+
 void ipsec_check_in_one(const ipsec_test_part *part, odp_ipsec_sa_t sa)
 {
 	int num_out = part->num_pkt;
@@ -961,6 +989,14 @@ void ipsec_check_out_in_one(const ipsec_test_part *part,
 		if (part->flags.stats == IPSEC_TEST_STATS_AUTH_ERR)
 			ipsec_pkt_auth_err_set(pkto[i]);
 
+		if (part->flags.udp_encap) {
+			if ((!part->flags.tunnel && part->flags.v6) ||
+			    (part->flags.tunnel && part->flags.tunnel_is_v6))
+				ipsec_pkt_v6_check_udp_encap(pkto[i]);
+			else
+				ipsec_pkt_v4_check_udp_encap(pkto[i]);
+		}
+
 		pkt_in.len = odp_packet_len(pkto[i]);
 		pkt_in.l2_offset = odp_packet_l2_offset(pkto[i]);
 		pkt_in.l3_offset = odp_packet_l3_offset(pkto[i]);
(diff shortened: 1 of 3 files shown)


test/validation/api/ipsec/ipsec.h
```diff
@@ -41,7 +41,7 @@ typedef struct {
 	uint32_t l2_offset;
 	uint32_t l3_offset;
 	uint32_t l4_offset;
-	uint8_t data[256];
+	uint8_t data[384];
```
anoobj, line 44: @JannePeltonen This field was updated to handle packet size requirement for IPv6 + UDP encap.

For reassembly UT we need to have larger packet sizes and so we need to either update this to large value (like 5800 for example) or replicate this struct as a ipsec_reass_test_packet. Replicating the struct would mean lot of functions also would get replicated. So I would like to update this field for that case.

This change would increase only the standard vectors size. Buffer size used for packet pool creation would be unaffected for the existing UTs.

Can you confirm if updating this field is fine.

JannePeltonen: I do not see a problem in increasing the size of the array.

However, it is hard to verify that 384 or any other value is enough. The compiler will complain if the initializers in the test vectors exceed the size, but ipsec_check_out_in_one() writes to the array without doing any length check. I would suggest defining the array size through a macro and checking packet lenght against it in ipsec_check_out_in_one() before the odp_packet_copy_to_mem() call.

test/validation/api/ipsec/ipsec_test_out.c
```diff
@@ -407,22 +419,44 @@ static void test_out_in_common(ipsec_test_flags *flags,
 	    (auth == ODP_AUTH_ALG_NULL))
 		return;
 
+	if (flags->v4_tunnel) {
+		memset(&tunnel, 0, sizeof(odp_ipsec_tunnel_param_t));
```
JannePeltonen, line 423: This leaves the TTL field zero which is bad. And same for hop limit in the IPv6 case.

JannePeltonen: Sorry, hop limit was set, just TTL was not.

test/validation/api/ipsec/ipsec_test_out.c
```diff
@@ -407,22 +419,44 @@ static void test_out_in_common(ipsec_test_flags *flags,
...
 
 	ipsec_sa_param_fill(&param,
-			    true, flags->ah, 123, NULL,
+			    true, flags->ah, 123, tun_ptr,
 			    cipher, cipher_key,
 			    auth, auth_key,
 			    cipher_key_extra, auth_key_extra);
 
+	if (flags->udp_encap)
+		param.opt.udp_encap = 1;
```
JannePeltonen, line 458: Just setting this parameter and then checking that IPsec transformations still work does not test that UDP encapsulation is even done. An implementation could simply ignore the parameter and perform plain non-UDP encapsulated transformations and this validation test would not notice anything wrong.

So, I think at minimum some additional checks are needed to see that UDP encapsulation was actually done, even if its correctness is not fully tested.

test/validation/api/ipsec/ipsec_test_out.c
```diff
@@ -1392,6 +1426,25 @@ static void test_ipsec_stats(void)
...
+	memset(&flags, 0, sizeof(flags));
+	flags.udp_encap = 1;
+
+	printf("\n        IPv4 Tunnel");
+	flags.v4_tunnel = 1;
+	test_esp_out_in_all(&flags);
+	flags.v4_tunnel = 0;
+
+	printf("\n        IPv6 Tunnel");
+	flags.v6_tunnel = 1;
```
JannePeltonen, line 1442: Why is this test limited to tunnel mode? Transport mode could be tested too.

test/validation/api/ipsec/ipsec.h
anoobj: It makes sense to introduce macros to indicate the max size of the array. 

> The compiler will complain if the initializers in the test vectors exceed the size, but ipsec_check_out_in_one() writes
> to the array without doing any length check.

Would this be sufficient? 
https://github.com/OpenDataPlane/odp/blob/master/test/validation/api/ipsec/ipsec.c#L898

The sizeof(pkt_in.data) would be same as the size of the array, right?

test/validation/api/ipsec/ipsec_test_out.c
anoobj: Will add TTL. Will set it 64 like in other instances. Thanks for pointing out.

test/validation/api/ipsec/ipsec.h
JannePeltonen: Yes, that is sufficient, so you can ignore my comment about there not being a check. I just missed it because it is not close to the copy operation for which it is needed.

test/validation/api/ipsec/ipsec_test_out.c
anoobj: I'll add an assert for ipv4.proto & ipv6.next_hdr. Hope that would suffice.

JannePeltonen: That's enough for me now. It can be improved later if needed but this change already improves test coverage.

test/validation/api/ipsec/ipsec_test_out.c
anoobj: Agreed. I tried enabling ipv4 & ipv6 transport mode. ipv6.next_hdr is coming as 0 in case of transport mode with linux-generic. Let me try to debug this.

test/validation/api/ipsec/ipsec_test_out.c
```diff
@@ -446,6 +482,16 @@ static void test_out_in_common(ipsec_test_flags *flags,
 		},
 	};
 
+	if (flags->v6 && !flags->tunnel) {
```
JannePeltonen, line 485: The v6 flag has a problematic double meaning. In tunnel mode it tells whether the tunnel is IPv6 or IPv4 and in transport mode it tells if the payload packet is IPv6 or IPv4. Now if I want to add a test for tunneled IPv6 packet, the logic related to the flag breaks apart. It would be better to just define two different flags for the two different things already now. And once you do that, you could even add a tests for IPv6 payloads in UDP encapsulated tunnels.

test/validation/api/ipsec/ipsec.c
```diff
@@ -763,6 +763,35 @@ static void ipsec_pkt_auth_err_set(odp_packet_t pkt)
...
+	odph_ipv6hdr_t ip;
+	uint8_t next_hdr;
+
+	odp_packet_copy_to_mem(pkt, l3_off, sizeof(ip), &ip);
+
+	next_hdr = ip.next_hdr;
+
+	if (ip.next_hdr == ODPH_IPPROTO_HOPOPTS) {
+		odp_packet_copy_to_mem(pkt, l3_off + sizeof(ip), sizeof(ext), &ext);
+		next_hdr = ext.next_hdr;
```
JannePeltonen, line 789: I wonder why this handling of hop-by-hop options is here? AFAICS, we do not have that kind of packets in the tests so this branch would never be executed. Why not simply remove it and fail the validation test if there are any extension headers?

anoobj: When I tried IPv6 with transport mode, the test was failing. Hence added this. Basically, the test vector used has hop-by-hop options and in transport mode, ESP additions would be after all IPv6 headers.

test/validation/api/ipsec/ipsec_test_out.c
anoobj: So I'll update flags as,

```
odp_bool_t v6;
odp_bool_t tunnel;
odp_bool_t v6_tunnel;
```

The first one will say test packet's IP version. 

Or should I go back to having v4_tunnel & v6_tunnel? If both are not set, then that would indicate transport. What is your preference?

JannePeltonen: Either way is ok for me. Maybe the v6_tunnel in the example above could be something like tunnel_is_v6 to make it clearer that tunnel is not just a shorter version of v4_tunnel but an overall tunneling flag.

test/validation/api/ipsec/ipsec.c
JannePeltonen: You are right. Now I can see the hop-by-hop options in the test vectors.

test/validation/api/ipsec/ipsec_test_out.c
anoobj: I tried enabling all combinations,

```
  Test: test_udp_encap ...
        IPv4 Transport
        IPv6 Transport
        IPv4-in-IPv4 Tunnel
        IPv4-in-IPv6 Tunnel
        IPv6-in-IPv4 Tunnel
        IPv6-in-IPv6 Tunnel
  passed
```

JannePeltonen (approved):

PR #1181, opened 2021-02-12
[PATCH v2] linux-gen: fix static code analyzer warnings

matiaselo:


platform/linux-generic/odp_classification.c
```diff
@@ -691,8 +691,7 @@ static int pmr_create_term(pmr_term_value_t *value,
 		return -1;
 	}
 
-	memset(&value->match.value, 0, MAX_PMR_TERM_SIZE);
-	memset(&value->match.mask, 0, MAX_PMR_TERM_SIZE);
+	memset(&value->match, 0, sizeof(value->match));
```
JereLeppanen, line 694: So here we don't expect in the future to have anything in value->match that we don't want to set to zero?

matiaselo: The following two lines will set all `match` contents. So it should be easy to notice if a new member, which we don't want to zero, would be added to `match`. However, this is very unlikely.

JereLeppanen (approved):

PR #1182, opened 2021-02-12
[PATCH v2] api: packet: introduce max flow hash value constant

harman-kalra:
## api: packet: introduce max flow hash pktio capability

Introducing max flow hash value as pktio capability, which
defines the maximum flow hash that can be set by the
application. This capability is required as some platforms
may have limitation on maximum flow hash value that can be
set.

## linux-gen: pktio: max flow hash value

Defining maximum flow hash pktio capability for linux
generic platform.

## validation: pktio: use maximum flow hash capability

Some platforms may have limitation on size of flow hash value
that can be set, using max_flow_hash pktio capability to set
platform defined max flow hash value.

### DIFF
--- a/include/odp/api/spec/packet.h
+++ b/include/odp/api/spec/packet.h
@@ -1913,7 +1913,8 @@ uint32_t odp_packet_flow_hash(odp_packet_t pkt);
  *
  * Store the packet flow hash for the packet and sets the flow hash flag. This
  * enables (but does not require!) application to reflect packet header
- * changes in the hash.
+ * changes in the hash. Maximum hash value that can be set is defined by
+ * pktio capability 'max_flow_hash'.
  *
  * @param      pkt              Packet handle
  * @param      flow_hash        Hash value to set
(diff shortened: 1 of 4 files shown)


include/odp/api/abi-default/packet.h
```diff
@@ -34,6 +34,7 @@ typedef _odp_abi_packet_vector_t *odp_packet_vector_t;
 #define ODP_PACKET_SEG_INVALID    ((odp_packet_seg_t)0)
 #define ODP_PACKET_OFFSET_INVALID 0xffff
 #define ODP_PACKET_VECTOR_INVALID   ((odp_packet_vector_t)0)
+#define ODP_PACKET_MAX_FLOW_HASH  UINT32_MAX
```
psavol, line 37: max flow hash value could be a pktio capability. pktio max hash >= flow aware scheduler max flow hash capa. This would be also the max flow hash value that application can set into packet.

harman-kalra: Thanks for reviewing, I will get back with V2.

PR #1183, opened 2021-02-12
[PATCH v1] api: packet: updating documentation for packet input set

harman-kalra:
## api: packet: updating documentation for packet input set

Adding a note to the documentation of odp_packet_input_set() API,
to be sure that PktIO handle is in configured state before passing
it as an argument to the API.

## validation: packet: configure pktio before packet input set

PktIO handle remains unconfigured after odp_pktio_open() and
to take it into configured state, pktio must be configured for
input and output.
odp_packet_input_set() may fail in some platform if pktio is in
unconfigured state.

### DIFF
--- a/include/odp/api/spec/packet.h
+++ b/include/odp/api/spec/packet.h
@@ -1539,6 +1539,8 @@ odp_pktio_t odp_packet_input(odp_packet_t pkt);
  * @param pktio Handle to a valid packet input interface or ODP_PKTIO_INVALID.
  *              ODP_PKTIO_INVALID indicates that the packet was not received by
  *              any packet IO interface.
+ *
+ * @note odp_packet_input_set() expects PktIO handle to be in configured state.
  */
 void odp_packet_input_set(odp_packet_t pkt, odp_pktio_t pktio);
 
--- a/test/validation/api/packet/packet.c
+++ b/test/validation/api/packet/packet.c
@@ -1442,6 +1442,9 @@ static void packet_test_meta_data_copy(void)
 	pktio = odp_pktio_open("loop", pool, NULL);
 	CU_ASSERT_FATAL(pktio != ODP_PKTIO_INVALID);
 
+	CU_ASSERT(odp_pktin_queue_config(pktio, NULL) == 0);
+	CU_ASSERT(odp_pktout_queue_config(pktio, NULL) == 0);
+
 	t1 = odp_time_global();
 
 	pkt = odp_packet_alloc(pool, packet_len);


include/odp/api/spec/packet.h
```diff
@@ -1539,6 +1539,8 @@ odp_pktio_t odp_packet_input(odp_packet_t pkt);
  * @param pktio Handle to a valid packet input interface or ODP_PKTIO_INVALID.
  *              ODP_PKTIO_INVALID indicates that the packet was not received by
  *              any packet IO interface.
+ *
+ * @note odp_packet_input_set() expects PktIO handle to be in configured state.
```
JannePeltonen, line 1543: I do not think "configured state" is well enough defined. Does it mean that pktin/pktout queue config has been done or only that odp_pktio_config() has been called (but calling odp_pktio_config() is optional in the API)? Also, it is not the pktio handle that is in "configured state" but the pktio itself.

The commit message should be improved to describe not only the change itself but the reason behind the change as it is not obvious. The subject line of the commit message should be made more descriptive.

psavol: Is there any APIs that give pktio as input, and would need the same config state requirements.

nithind1988: Since odp_packet_input(), odp_packet_input_index(), odp_packet_input_set() are used in fast path API's, we are only using internal link identifiers to identify PKTIO. In our HW, this link identifier/channel is allocated only when we provide pktin queue count and pktout queue count. Since even spec says that for a PKTIO, odp_pktin_queue_config() and odp_pktout_queue_config() are mandatory API's for PKTIO to be functional(unless pktin/pktout are disabled), we are allocating resources after these two calls. 

We agree that "configured state" is not mentioned in packet_io.h but only mentioned in users-guide-pktio.adoc.
All the other users of pktio handle in other subsystems like odp_cls are non-fast path. There is another place in ipsec inline enq param where pktio handle is provided but it can only be enqueued when PKTIO is running.

Since it is a fast path impacting change, can we go with this and change the note to ?

@note odp_packet_input_set() for performance reasons can only be called with pktio handle only after all the mandatory and must be called Packet IO config API's such as odp_pktin_queue_config() and/or odp_pktout_queue_config() are successfully complete.

JannePeltonen: > we are only using internal link identifiers to identify PKTIO. In our HW, this link identifier/channel is allocated only when we provide pktin queue count and pktout queue count. 

odp_pktin_queue_config() can be called multiple times to reconfigure the queues (e.g. to change queue count etc). If you cannot handle it, then further API adjustments would be needed. If you can handle it, I wonder why we need any API change at all. I would think that you could, e.g. have your implementation do the the necessary allocations at pktio open time as if the application called odp_pktin_queue_config() at the same time (even though in reality it did not) and handle the actual queue config the same way as you would handle queue reconfig. Maybe I am missing something?

Note that with the current API, and also after your proposed API change, an application can expect to be able to set "packet input" metadata in packet, then reconfigure the pktio, and get the correct pktio handle from the packet in a later odp_packet_input() call.

JannePeltonen: > Since even spec says that for a PKTIO, odp_pktin_queue_config() and odp_pktout_queue_config() are mandatory API's for PKTIO to be functional(unless pktin/pktout are disabled)

odp_pktout_queue_config() call is optional even when the pktout is not disabled. The API spec says: "When packet output mode is ODP_PKTOUT_MODE_DISABLED or ODP_PKTOUT_MODE_TM, odp_pktout_queue_config() call is optional and will ignore all parameters."

nithind1988: Ok. Actually we proposed this spec on lines with that it is logical to have such restrictions across subsystems such as packet, pktio and handle sharing b/w them.
But with our current Kernel/Mbox interface, we have to do a bit of major rework to support this to avoid issues with things like FLR, etc. 
We will close this PR for now and until we rework our code to accomodate this, we will have this as limitation in our implementation.

PR #1188, opened 2021-02-16
[PATCH v6] api: ipsec: define more default values & improve SA info API

JannePeltonen:
Define default values for more fields of odp_ipsec_sa_param_t so that applications do not always have to explicitly initialize as many fields after the odp_ipsec_sa_param_init() call.

The current validation tests do not actually set all the parameters that they should. It can cause test failures in implementations that set the implementation-defined values differently from linux-gen. This API change makes the needed fix there smaller.

### DIFF
--- a/include/odp/api/spec/ipsec.h
+++ b/include/odp/api/spec/ipsec.h
@@ -522,13 +522,13 @@ typedef struct odp_ipsec_ipv4_param_t {
 	/** IPv4 destination address (NETWORK ENDIAN) */
 	void *dst_addr;
 
-	/** IPv4 Differentiated Services Code Point */
+	/** IPv4 Differentiated Services Code Point. The default value is 0. */
 	uint8_t dscp;
 
-	/** IPv4 Don't Fragment bit */
+	/** IPv4 Don't Fragment bit. The default value is 0. */
 	uint8_t df;
 
-	/** IPv4 Time To Live */
+	/** IPv4 Time To Live. The default value is 255. */
 	uint8_t ttl;
 
 } odp_ipsec_ipv4_param_t;
@@ -541,13 +541,13 @@ typedef struct odp_ipsec_ipv6_param_t {
 	/** IPv6 destination address (NETWORK ENDIAN) */
 	void *dst_addr;
 
-	/** IPv6 flow label */
+	/** IPv6 flow label. The default value is 0. */
 	uint32_t flabel;
 
-	/** IPv6 Differentiated Services Code Point */
+	/** IPv6 Differentiated Services Code Point. The default value is 0. */
 	uint8_t dscp;
 
-	/** IPv6 hop limit */
+	/** IPv6 hop limit. The default value is 255. */
 	uint8_t hlimit;
 
 } odp_ipsec_ipv6_param_t;
@@ -561,11 +561,11 @@ typedef struct odp_ipsec_ipv6_param_t {
  * pointers and copied byte-by-byte from memory to the packet.
  */
 typedef struct odp_ipsec_tunnel_param_t {
-	/** Tunnel type: IPv4 or IPv6 */
+	/** Tunnel type: IPv4 or IPv6. The default is IPv4. */
 	odp_ipsec_tunnel_type_t type;
 
-	/** Variant mappings for tunnel parameters */
-	union {
+	/** Tunnel type specific parameters */
+	struct {
 		/** IPv4 header parameters */
 		odp_ipsec_ipv4_param_t ipv4;
 
@@ -581,7 +581,7 @@ typedef struct odp_ipsec_sa_opt_t {
 	/** Extended Sequence Numbers (ESN)
 	  *
 	  * * 1: Use extended (64 bit) sequence numbers
-	  * * 0: Use normal sequence numbers
+	  * * 0: Use normal sequence numbers (the default value)
 	  */
 	uint32_t esn : 1;
 
@@ -589,7 +589,7 @@ typedef struct odp_ipsec_sa_opt_t {
 	  *
 	  * * 1: Do UDP encapsulation/decapsulation so that IPSEC packets can
 	  *      traverse through NAT boxes.
-	  * * 0: No UDP encapsulation
+	  * * 0: No UDP encapsulation (the default value)
 	  */
 	uint32_t udp_encap : 1;
 
@@ -599,7 +599,7 @@ typedef struct odp_ipsec_sa_opt_t {
 	  *      the outer IP header in encapsulation, and vice versa in
 	  *      decapsulation.
 	  * * 0: Use values from odp_ipsec_tunnel_param_t in encapsulation and
-	  *      do not change DSCP field in decapsulation.
+	  *      do not change DSCP field in decapsulation (the default value).
 	  */
 	uint32_t copy_dscp : 1;
 
@@ -607,7 +607,7 @@ typedef struct odp_ipsec_sa_opt_t {
 	  *
 	  * * 1: Copy IPv6 flow label from inner IPv6 header to the
 	  *      outer IPv6 header.
-	  * * 0: Use value from odp_ipsec_tunnel_param_t
+	  * * 0: Use value from odp_ipsec_tunnel_param_t (the default value)
 	  */
 	uint32_t copy_flabel : 1;
 
@@ -615,7 +615,7 @@ typedef struct odp_ipsec_sa_opt_t {
 	  *
 	  * * 1: Copy the DF bit from the inner IPv4 header to the outer
 	  *      IPv4 header.
-	  * * 0: Use value from odp_ipsec_tunnel_param_t
+	  * * 0: Use value from odp_ipsec_tunnel_param_t (the default value)
 	  */
 	uint32_t copy_df : 1;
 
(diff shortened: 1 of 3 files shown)


JannePeltonen:
v2:
- Change some of the unions inside odp_ipsec_sa_param_t to structs so that it is possible to define (nonzero) defaults that the init function can set. This may also help in backward compatibility of future API changes as new fields with non-zero defaults can be added in all parts of the parameter structure.
- Add linux-gen implementation

JereLeppanen:
Typo in the commit comment of the first commit: "paramteter".

include/odp/api/spec/ipsec.h
```diff
@@ -712,10 +714,12 @@ typedef struct odp_ipsec_sa_param_t {
 	uint32_t context_len;
 
 	/** IPSEC SA direction dependent parameters */
-	union {
+	struct {
```
anoobj, line 717: Is this intentional?

test/validation/api/ipsec/ipsec_test_out.c
```diff
@@ -1371,6 +1371,53 @@ static void ipsec_test_capability(void)
...
+	CU_ASSERT(config.inbound.lookup.spi_overlap == 0);
+	CU_ASSERT(config.inbound.retain_outer == ODP_PROTO_LAYER_NONE);
+	CU_ASSERT(config.inbound.parse_level == ODP_PROTO_LAYER_NONE);
+	CU_ASSERT(config.inbound.chksums.all_chksum == 0);
+	CU_ASSERT(config.outbound.all_chksum == 0);
+	CU_ASSERT(!config.stats_en);
+
+	odp_ipsec_sa_param_init(&sa_param);
+	CU_ASSERT(sa_param.proto == ODP_IPSEC_ESP);
+	CU_ASSERT(sa_param.crypto.cipher_alg == ODP_CIPHER_ALG_NULL);
```
anoobj, line 1394: Just wondering what should be the behavior if the implementation doesn't support NULL algos.

include/odp/api/spec/ipsec.h
JannePeltonen: Yes it is, as explained in the github comment and the commit message.

Using a union has the drawback that default values cannot be defined for the union members as odp_ipsec_sa_param_init() cannot write them without affecting all the union members at the same time. And not being able to define defaults makes the _param_init() pattern used in ODP less useful. Changing the union to a struct fixes this.

Increased memory usage is a drawback but maybe it does not matter too much since this structure is needed only at SA creation time for parameter passing. In my x86 build the size seems to increase from 208 bytes to 272 bytes.

So, does the change make sense or should we just stick with the unions (and define fewer default values)?

test/validation/api/ipsec/ipsec_test_out.c
JannePeltonen: TL; DR: We can leave things as they are or consider changing the default authentication algorithm to SHA-1.

odp_ipsec_sa_param_init() is supposed to set the default values as specified in the API spec regardless of the actual algorithm support in the implementation. As you must have noticed, this PR does not change the default algorithms but just adds a test for what was already required.

If an implementation does not support null encryption or null authentication, we have a situation where the default parameters set by odp_ipsec_sa_param_init() cannot be passed to odp_ipsec_sa_create() and the application has to set the parameters to suitable values. But that is not much different from parameters which do not have default value defined.

That said, null encryption in ESP is likely supported by many implementations because it is mandatory in IPsec and also tends to be somewhat easier to implement than most other ciphers.

Null authentication is no longer mandatory in IPsec and even if it is supported, it may not be combined with null encryption in real usage (but in performance tuning such combo can be useful and ODP API allows it).

Maybe SHA-1, which is mandatory in IPsec and thus likely supported by most implementations, would be a better default value for the authentication algorithm than the current null authentication.

include/odp/api/spec/ipsec.h
anoobj: The only drawback is, sa_info is supposed to return this struct and the memory requirement would just increase. Or if we can state that odp_ipsec_sa_param_t.inbound/outbound would be ignored based on odp_ipsec_sa_param_t.dir, then it should be fine.

IOW, when dir = inbound, sa_info.param.outbound would have default values (& the other way round also). Would that be fine?

JannePeltonen: I think the SA info API is a bit problematic as the current wording suggests that the returned param struct is a exact (?) copy of the one used in SA creation. But we do not want that as we do not want to force implementations to store the struct or the values of the ignored fields that cannot be recreated from internal SA data.

This problem exists already before the union-to-struct change, but the change makes it words.

I think I would like to propose the following change to the sa_info API, regardless of what we do with this PR:

```
 typedef struct odp_ipsec_sa_info_t {
-       /** Copy of IPSEC Security Association (SA) parameters */
+       /** IPsec SA parameters
+        *
+        * This is not necessarily an exact copy of the actual parameter
+        * structure used in SA creation. The fields that were relevant
+        * for the SA in the creation phase will have the same values,
+        * but other fields, such as tunnel parameters for a transport
+        * mode SA, will have undefined values.
+        */
        odp_ipsec_sa_param_t param;
```

anoobj: I agree to the suggestion on sa_info. With that change, move to struct is not an issue. Agreed that separate structs allows for specifying defaults better.

JannePeltonen:
v3:
- Added SA info API change to relax the requirements toward the returned parameter struct
- Fixed the typo in a commit message
- Rebased

include/odp/api/spec/ipsec.h
```diff
@@ -765,7 +769,9 @@ typedef struct odp_ipsec_sa_param_t {
 			/** Parameters for tunnel mode */
 			odp_ipsec_tunnel_param_t tunnel;
 
-			/** Fragmentation mode */
+			/** Fragmentation mode
+			 *  The default value is ODP_IPSEC_RRAG_DISABLED.
```
psavol, line 773: Typo: ODP_IPSEC_RRAG_DISABLED -> FRAG

psavol:
First commit log entry has typo: ... remember the content of also those fields thae -> that

test/validation/api/ipsec/ipsec_test_out.c
anoobj: > If an implementation does not support null encryption or null authentication, we have a situation where the default 
> parameters set by odp_ipsec_sa_param_init() cannot be passed to odp_ipsec_sa_create() and the application has to set
> the parameters to suitable values.

This is what I had in mind. If application is expected to check crypto capabilities before doing odp_ipsec_sa_create(), then we can keep existing behavior. RFC compliance would be the rationale to state algos as mandatory. But implementations need not be RFC compliant. 

Other option is to leave default values for algos as implementation defined. Would that be fine? But then testing such a field would be very complicated.

JannePeltonen:
v4: Fixed the two typos.

test/validation/api/ipsec/ipsec_test_out.c
JannePeltonen: > If application is expected to check crypto capabilities before doing odp_ipsec_sa_create()

Yes, it is expected to do that or accept that the creation may fail.

> Other option is to leave default values for algos as implementation defined. Would that be fine?

I think that would be ok too, but I do not think it would be any more useful to an application than the null defaults. You cannot just let the implementation choose the crypto algorithms for an SA and you would have to give keys of correct length too.

Note that this PR is not changing the default values for the algorithms but only adds a validation check for what was already specified in the API.

anoobj: > You cannot just let the implementation choose the crypto algorithms for an SA and you would have to give keys of
> correct length too

Yeah. In that sense, it is better to retain the current behavior (with NULLs).

> Note that this PR is not changing the default values for the algorithms but only adds a validation check for what was
> already specified in the API.

Agreed. I was just trying to understand the spec requirements in case NULL algos are not supported.

include/odp/api/spec/ipsec.h
```diff
@@ -491,11 +491,11 @@ typedef struct odp_ipsec_ipv6_param_t {
...
  */
 typedef struct odp_ipsec_tunnel_param_t {
-	/** Tunnel type: IPv4 or IPv6 */
+	/** Tunnel type: IPv4 or IPv6. The default is IPv4. */
 	odp_ipsec_tunnel_type_t type;
 
-	/** Variant mappings for tunnel parameters */
-	union {
+	/** Tunnel type specific parameters */
+	struct {
```
anoobj, line 498: Missed this earlier. Is this change required? Or are we replacing all unions with struct in general? We are not specifying default values for these, right?

psavol: Defaults are defined at odp_ipsec_ipv4_param_t  and odp_ipsec_ipv6_param_t definition e.g. ttl and hlimit.

anoobj: Missed it. Thanks!

psavol (approved):

anoobj:

JannePeltonen:
v5: Added review tags

PR #1195, opened 2021-02-19
[PATCH v2] test: performance: add new atomic operation benchmark application

matiaselo:
Add new performance test application for atomic operations. The test
repeats each atomic operation 'num_round' times and measures the maximum
rate of operations. Each worker thread can have a private atomic variable
on a separate cache line ('-p 1'), or all threads can share a common
variable (default).

V2:
- Validate load test output (Petri)


test/performance/odp_atomic_perf.c
```diff
@@ -0,0 +1,1183 @@
...
+
+	*result = ret;
+}
+
+static inline int validate_atomic_init_val_u32(void *val, uint32_t num_round ODP_UNUSED,
+					       uint32_t worker ODP_UNUSED, int private ODP_UNUSED)
+{
+	odp_atomic_u32_t *atomic_val = val;
+
+	return (odp_atomic_load_u32(atomic_val) != INIT_VAL);
```
psavol, line 124: Load test could check that thread output equals num_rounds * INIT_VAL

PR #1196, opened 2021-02-19
[PATCH v2] don't run tests in parallel

JereLeppanen:
```
ci: use multiple jobs for distcheck
    
    Use $(nproc) jobs to run distcheck in CI.
    
    Signed-off-by: Jere Leppä?nen <jere.leppanen@nokia.com>

don't run tests in parallel
    
    Explicitly prevent tests from being run in parallel when the -j option
    is used with make check.
    
    Using the .NOTPARALLEL special target is the only easy way to ensure
    that tests are run one by one. Unfortunately this also prevents
    parallel builds for the tests.
    
    Signed-off-by: Jere Leppä?nen <jere.leppanen@nokia.com>
```
v2:
- Add "build:" tag to subject line of first commit.
- Rebase.
- Add reviewed-by tags.

### DIFF
--- a/example/packet/Makefile.am
+++ b/example/packet/Makefile.am
@@ -35,3 +35,5 @@ clean-local:
 			rm -f $(builddir)/$$f; \
 		done \
 	fi
+
+.NOTPARALLEL:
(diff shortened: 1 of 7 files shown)


matiaselo (approved):
Needs rebase and commit message tweak. Otherwise looking good.

PR #1197, opened 2021-02-23
[PATCH v3] Increase the maximum number of IPsec SAs and crypto sessions

JannePeltonen:
- linux-gen: ipsec: increase the maximum number of SAs
- linux-gen: ipsec: do not alloc per-thread memory for too many threads
- validation: ipsec: test maximum number of SAs
- validation: ipsec: enable antireplay service and lifetimes
- linux-gen: crypto: increase the maximum number of sessions to 4000

Also, update copyrights in the affected files not only for this change but for earlier changes for which the update was not done.

This PR depends (logically, not textually) on PR 1188 (api: ipsec: define more default values & improve SA info API).

### DIFF
--- a/platform/linux-generic/odp_crypto_openssl.c
+++ b/platform/linux-generic/odp_crypto_openssl.c
@@ -37,7 +37,7 @@
 #define _ODP_HAVE_CHACHA20_POLY1305 0
 #endif
 
-#define MAX_SESSIONS 32
+#define MAX_SESSIONS 4000
 #define AES_BLOCK_SIZE 16
 #define AES_KEY_LENGTH 16
 
--- a/test/validation/api/ipsec/ipsec.c
+++ b/test/validation/api/ipsec/ipsec.c
@@ -1,5 +1,5 @@
 /* Copyright (c) 2017-2018, Linaro Limited
- * Copyright (c) 2019-2020, Nokia
+ * Copyright (c) 2018-2021, Nokia
  * Copyright (c) 2020, Marvell
  * All rights reserved.
  *
@@ -16,6 +16,7 @@
 #include "test_vectors.h"
 
 struct suite_context_s suite_context;
+static odp_ipsec_capability_t capa;
 
 #define PKT_POOL_NUM  64
 
@@ -123,11 +124,6 @@ int ipsec_check(odp_bool_t ah,
 		odp_auth_alg_t auth,
 		uint32_t auth_bits)
 {
-	odp_ipsec_capability_t capa;
-
-	if (odp_ipsec_capability(&capa) < 0)
-		return ODP_TEST_INACTIVE;
-
 	if ((ODP_IPSEC_OP_MODE_SYNC == suite_context.inbound_op_mode &&
 	     ODP_SUPPORT_NO == capa.op_mode_sync) ||
 	    (ODP_IPSEC_OP_MODE_SYNC == suite_context.outbound_op_mode &&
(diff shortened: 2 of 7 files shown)


test/validation/api/ipsec/ipsec_test_out.c
```diff
@@ -1392,6 +1392,126 @@ static void test_ipsec_stats(void)
...
 }
 
+static void test_max_num_sa(void)
+{
+	odp_ipsec_capability_t capa;
+	uint32_t sa_pairs;
+	odp_bool_t odd = false;
+	uint32_t n;
+	uint8_t cipher_key_data[128 / 8]; /* 128 bit key for AES */
+	uint8_t auth_key_data[160 / 8];   /* 192 bit key for SHA-1 */
```
JereLeppanen, line 1402: 160, not 192, in the comment.

test/validation/api/ipsec/ipsec.c
```diff
@@ -309,6 +306,11 @@ void ipsec_sa_param_fill(odp_ipsec_sa_param_t *param,
 
 	if (auth_key_extra)
 		param->crypto.auth_key_extra = *auth_key_extra;
+
+	param->lifetime.soft_limit.bytes = 900 * 1000;
+	param->lifetime.hard_limit.bytes = 1000 * 1000;
+	param->lifetime.soft_limit.packets = 9000 * 1000;
+	param->lifetime.hard_limit.packets = 10000 * 1000;
```
JereLeppanen, line 313: Are these limits exceeded in any test?

platform/linux-generic/odp_crypto_openssl.c
```diff
@@ -37,7 +37,7 @@
 #define _ODP_HAVE_CHACHA20_POLY1305 0
 #endif
 
-#define MAX_SESSIONS 32
+#define MAX_SESSIONS 4000
```
JereLeppanen, line 40: How much memory does this take?

JannePeltonen: On my x86_64 system with OpenSSL 1.1.1f, roughly this much more, if I looked right:
- 2.3 MB for the process
- 0.15 MB for each thread of the process
- 3,3 MB + some amount of malloc overhead for each ODP thread (unless crypto has been disabled in the runtime config)

This is a bit unrelated but the usage of malloc inside the openssl library is actually a bit problematic since it won't work in the process mode where each ODP thread is an OS process. Suitable hooking could fix that with some effort, though.

test/validation/api/ipsec/ipsec.c
JannePeltonen: Not even close. I just wanted to put nonzero values there to force the implementation to perform lifetime checking just to get more of the code run, even if not really tested, in the tests. Real life time tests seem to be missing.

The limits are largish (compared to the handful of packets the tests actually send) to prevent us getting
an unexpected expiration indication if an implementation (like the linux-gen reference implementation) optimizes lifetime checking in ways that make it somewhat inexact.

test/validation/api/ipsec/ipsec_test_out.c
JannePeltonen: ack.

test/validation/api/ipsec/ipsec.c
JereLeppanen: Some code comment might be good here because the numbers look like they might have some meaning, but they're really just large values.

JannePeltonen:
v2:
- fixed the key length error in comment
- added comment about the life time values
- rebased on latest master

JereLeppanen (approved):

PR #1205, opened 2021-02-26
api: pktio: add pktout aging drop support

nithind1988:
commit 132546529758b059d83bfaba24186b1a54b5d5e5
Author: Jerin Jacob <jerinj@marvell.com>
Date:   Thu Feb 11 13:43:24 2021 +0530

    api: pktio: add pktout aging drop support
    
    Add pktout support to age packets stuck in packet output queue and drop
    them after given timeout that is set per packet.
    
Example usage of Tx packet Aging feature:
```
    # Enable Aging on PKTIO
    config.pktout.bit.aging_ena = 1;
	
    # Set aging Timeout to 1000 usec
    odp_packet_aging_tmo_set(pkt, 1000);
    odp_packet_aging_drop_set(pkt, true);

    # Tx packet on a Packet out queue
    odp_pktout_send(pktout_queue, &pkt, 1);
```


include/odp/api/spec/packet.h
```diff
@@ -2140,6 +2140,50 @@ uint32_t odp_packet_payload_offset(odp_packet_t pkt);
...
+ * Set a packet aging timeout in usec which limits the age of packet in pktout queue or
+ * traffic scheduler/shapers. Packet exceeding its timeout will be dropped when packet aging
+ * drop is enabled.
+ *
+ * @param pkt     Packet handle
+ * @param tmo_us  Timeout to age in usec
+ *
+ * @see odp_pktio_capability_t::max_tx_aging_tmo_us
+ */
+void odp_packet_aging_tmo_set(odp_packet_t pkt, unsigned int tmo_us);
```
JannePeltonen, line 2166: Unit of the timeout should be nanoseconds as that is the time unit generally used elsewhere in ODP. And then the type should be uint64_t to make sure there are enough bits in it.

include/odp/api/spec/packet.h
```diff
@@ -2140,6 +2140,50 @@ uint32_t odp_packet_payload_offset(odp_packet_t pkt);
...
+ * @param enable  True indicates aging to be enabled on the packet.
+ */
+void odp_packet_aging_drop_set(odp_packet_t pkt, odp_bool_t enable);
+
+/**
+ * Set Tx packet aging timeout
+ *
+ * Set a packet aging timeout in usec which limits the age of packet in pktout queue or
+ * traffic scheduler/shapers. Packet exceeding its timeout will be dropped when packet aging
+ * drop is enabled.
```
JannePeltonen, line 2159: Maybe the exact point of packet drop should be specified (or alternatively, specified as implementation-specific). In particular, if a packet ages out when in TM queue, is the packet dropped from the queue right then and there, or does it still go through the schedulers and shapers in TM (affecting other traffic) and dropped at a later point?

include/odp/api/spec/packet.h
nithind1988: Ack, will fix it in v2.

include/odp/api/spec/packet.h
nithind1988: As discussed, I think we can leave this as implementation defined behavior as to when will it be dropped and what part of SoC bandwidth is consumed even if is about to be dropped.

include/odp/api/spec/packet.h
```diff
@@ -2140,6 +2140,50 @@ uint32_t odp_packet_payload_offset(odp_packet_t pkt);
...
+/**
+ * Enable or disable Tx packet drop based on age
+ *
+ * Enable or disable Tx packet drop based on packet age. When enabled, packet will be dropped
+ * if it is in Tx pktout queue for longer than per-packet timeout.
+ *
+ * @param pkt     Packet handle
+ * @param enable  True indicates aging to be enabled on the packet.
+ */
+void odp_packet_aging_drop_set(odp_packet_t pkt, odp_bool_t enable);
```
psavol, line 2152: Is separate drop enable/disable calls needed (odp_packet_aging_drop_set() and odp_packet_has_aging_drop())? Zero nanoseconds is not a useful value for aging, so we could specify that nsec==0 means that aging is disabled.

nithind1988: Ack, will make it to two API's(one for set and one for get) where 0 nsec means disable otherwise enable.

include/odp/api/spec/packet.h
```diff
@@ -2140,6 +2140,32 @@ uint32_t odp_packet_payload_offset(odp_packet_t pkt);
...
+ * if it is in Tx pktout queue or traffic shapers/schedulers for longer than timeout set.
+ *
+ * @param pkt        Packet handle
+ * @param tmo_ns     Timeout to age in nsec
+ *                >0 Aging drop is enabled with specified timeout.
+ *                =0 Aging drop is disabled.
+ *
+ * @see odp_pktio_capability_t::max_tx_aging_tmo_ns
+ */
+void odp_packet_aging_drop_set(odp_packet_t pkt, const uint64_t tmo_ns);
```
psavol, line 2156: const is not needed for tmo_ns (it's not a pointer).

I'd revise tmo_ns param documentation like this:
 param tmo_ns     Packet aging drop timeout in nsec. When 0, aging drop is disabled (default).
 
 The body text could say that that: !0 enables aging, 0 disables aging and aging is disabled by default.

include/odp/api/spec/packet.h
```diff
@@ -2140,6 +2140,32 @@ uint32_t odp_packet_payload_offset(odp_packet_t pkt);
...
+/**
+ * Check if packet has Tx aging drop enabled
+ *
+ * @param pkt Packet handle
+ *
+ * @return Aging drop timeout if enabled.
+ * @retval >0  Aging drop timeout in nano seconds and implies aging drop is enabled.
+ * @retval =0  If Aging drop is disabled.
+ */
+uint64_t odp_packet_has_aging_drop(odp_packet_t pkt);
```
psavol, line 2167: "retval =0", should be just "retval 0".

packet_has_xxx calls return flags (0 vs !0). This is not a flag, but a value. So, this should be just:
uint64_t odp_packet_aging_drop(odp_packet_t pkt);

Actually, the metadata value these read/write is not "aging drop", but "aging timeout in nsec". So, I think better name for both functions would be:
odp_packet_aging_tmo()
odp_packet_aging_tmo_set()

include/odp/api/spec/packet_io.h
```diff
@@ -489,6 +489,9 @@ typedef union odp_pktout_config_opt_t {
 		 */
 		uint64_t no_packet_refs  : 1;
 
+		/** Enable packet aging and drop */
+		uint64_t aging_ena  : 1;
```
psavol, line 493: Like with tx_compl, this could mention that:
0: application will not request packet aging (default)
1: application will request packet aging

include/odp/api/spec/packet_io.h
```diff
@@ -845,6 +848,12 @@ typedef struct odp_pktio_capability_t {
 		uint32_t max_output;
 	} maxlen;
 
+	/**
+	 * Max Tx aging timeout in nano seconds supported when packet aging
+	 * feature is supported.
+	 */
+	uint64_t max_tx_aging_tmo_ns;
```
psavol, line 855: This misses to define when aging is / is not supported. E.g.:
0: aging is not supported
>0: maximum aging timeout supported in nanoseconds

include/odp/api/spec/packet.h
nithind1988: Ack. Will change it in v5.

include/odp/api/spec/packet_io.h
nithind1988: Ack.

include/odp/api/spec/packet_io.h
nithind1988: Ack.

include/odp/api/spec/packet.h
nithind1988: Ack.

psavol:
API in v5 looks good to me.

include/odp/api/spec/packet_io.h
```diff
@@ -489,6 +489,13 @@ typedef union odp_pktout_config_opt_t {
 		 */
 		uint64_t no_packet_refs  : 1;
 
+		/** Enable packet aging and drop
+		 *
+		 * 0: application will not request packet aging (default)
+		 * 1: application will request packet aging
```
JannePeltonen, line 495: This could be: 1: application may request packet aging

Other than that, the patch looks ok to me.

test/validation/api/pktio/pktio.c
```diff
@@ -3468,6 +3468,109 @@ static void pktio_test_recv_maxlen_set(void)
 	packet_len = PKT_LEN_NORMAL;
 }
 
+static int pktio_check_pktout_aging_tmo(void)
+{
+	odp_pktio_param_t pktio_param;
+	odp_pktio_capability_t capa;
+	odp_pktio_t pktio;
+	int idx = 0;
```
psavol, line 3476: This variable is not needed. Other pktio_check functions use iface_name[0] directly.

test/validation/api/pktio/pktio.c
```diff
@@ -3468,6 +3468,109 @@ static void pktio_test_recv_maxlen_set(void)
...
+	pktio_param.out_mode = ODP_PKTOUT_MODE_DIRECT;
+
+	pktio = odp_pktio_open(iface_name[idx], pool[idx], &pktio_param);
+	if (pktio == ODP_PKTIO_INVALID)
+		return ODP_TEST_INACTIVE;
+
+	ret = odp_pktio_capability(pktio, &capa);
+	(void)odp_pktio_close(pktio);
+
+	if (ret < 0 || !capa.config.pktout.bit.aging_ena)
```
psavol, line 3490: This should check max_tx_aging_tmo_ns capability instead. That's documented to be 0 when not supported. We should remove .config from capa at some point.

test/validation/api/pktio/pktio.c
```diff
@@ -3468,6 +3468,109 @@ static void pktio_test_recv_maxlen_set(void)
...
+		return ODP_TEST_INACTIVE;
+
+	return ODP_TEST_ACTIVE;
+}
+
+static void pktio_test_pktout_aging_tmo(void)
+{
+	odp_pktio_t pktio[MAX_NUM_IFACES] = {0};
```
psavol, line 3498: odp_pktio_t pktio[MAX_NUM_IFACES] = {ODP_PKTIO_INVALID};

test/validation/api/pktio/pktio.c
```diff
@@ -3468,6 +3468,109 @@ static void pktio_test_recv_maxlen_set(void)
...
+	ret = odp_pktout_queue(pktio_tx, &pktout_queue, 1);
+	CU_ASSERT_FATAL(ret > 0);
+
+	/* Prepare packets with aging */
+	for (i = 0; i < TX_BATCH_LEN; i++) {
+		/* Aging disabled by default */
+		CU_ASSERT(odp_packet_aging_tmo(pkt_tbl[i]) == 0);
+
+		odp_packet_aging_tmo_set(pkt_tbl[i], pktio_capa.max_tx_aging_tmo_ns - 1);
+		CU_ASSERT(odp_packet_aging_tmo(pkt_tbl[i]) != 0);
```
psavol, line 3550: We don't have resolution capa in API, but I think you could test with couple of values to see that the returned value correlate those e.g.:
odp_packet_aging_tmo_set(pkt_tbl[i], pktio_capa.max_tx_aging_tmo_ns - 1);
tmo_0 = odp_packet_aging_tmo(pkt_tbl[i]);
odp_packet_aging_tmo_set(pkt_tbl[i], pktio_capa.max_tx_aging_tmo_ns/2);
tmo_1 = odp_packet_aging_tmo(pkt_tbl[i]);
CU_ASSERT(tmo_0 > tmo_1);

nithind1988:
Addressed all comments in v8

psavol (approved):

PR #1207, opened 2021-02-26
[PATCH v18] api: packet_io: introduce generic protocol stats framework with Tx per-packet stats offload

nithind1988:
commit 192b077eb422c0064c04e6715226e579765cc086
Author: Jerin Jacob <jerinj@marvell.com>
Date:   Mon Feb 15 15:37:43 2021 +0530

    api: packet: add proto stats meta data support
    
    Add metadata operations needed on per-packet basis for Packet IO
    protocol stats feature on Tx
    
commit 8b2a99509e3eac33f3a435072ba7562dbfd585f0
Author: Jerin Jacob <jerinj@marvell.com>
Date:   Mon Feb 15 15:35:30 2021 +0530

    api: pktio: add support for Tx proto stats
    
    Add support for Packet IO protocol stats feature on Tx. A packet is
    associated with a proto stats object and when packet reaches
    terminal state in pktout queue, those stats are updated. Proto stats
    needs to be requested per-packet with by default it being disabled.
    
commit c90b3abebef6ba77a10bb98777e8e45e12eca4a2
Author: Jerin Jacob <jerinj@marvell.com>
Date:   Mon Feb 15 15:29:42 2021 +0530

    api: proto_stats: introduce generic protocol stats framework
    
    Add generic protocol stats framework where protocol stats object
    supports basic operations like creation, lookup, access and destroy
    of stats object is supported. Each object supports a set of
    stats based on object params. Layout of the stats is implementation
    defined.
    
    This framework is added so that in platforms where the layout of
    stats is platform dependent and different for various HW blocks / modes
    it is easier to integrate. This generic framework is also useful
    when we have to associate stats objects to fast path with minimal
    overhead as it is as close to HW implementation as possible.
    
Example usage of Tx protocol stats feature:

```
   # Enable Tx protocol stats feature on PKTIO
   config.pktout.bit.proto_stats = 1;
	
   # Create a stats object to count pkts and drop count
   params.mode = ODP_PROTO_STATS_MODE_TX_PKT;
   odp_proto_stats_t stat = odp_proto_stats_create("Tx IP packets", &params);

   # Associate a stats object to a packet and enable proto stats update for this packet
   odp_packet_proto_stats_set(pkt, stat);
   
   # Send the packet
   odp_pktout_send(pktout_queue, &pkt, 1) ;

   # Access stats
   uint64_t *stat_base = odp_proto_stats(stat, &max_index);
   uint8_t tx_pkt_sent_index = odp_proto_stats_index(stat, ODP_PROTO_STATS_ID_TX_PKT);
   uint8_t tx_pkt_drop_index = odp_proto_stats_index(stat, ODP_PROTO_STATS_ID_TX_PKT_DROP);

   stat_base[tx_pkt_sent_index];
   stat_base[tx_pkt_drop_index];
   
```

### DIFF
--- a/include/odp/api/spec/packet.h
+++ b/include/odp/api/spec/packet.h
@@ -19,6 +19,7 @@
 extern "C" {
 #endif
 
+#include <odp/api/proto_stats.h>
 #include <odp/api/time.h>
 #include <odp/api/packet_types.h>
 
@@ -2029,6 +2030,61 @@ int odp_packet_tx_compl_request(odp_packet_t pkt, const odp_packet_tx_compl_opt_
  * @retval 0         TX completion event is not requested
  */
 int odp_packet_has_tx_compl_request(odp_packet_t pkt);
+
+/**
+ * Packet proto stats options.
+ */
+typedef struct odp_packet_proto_stats_opt_t {
+	/** Packet proto stats object handle
+	 *
+	 * Stats in the packet proto stats object will be updated.
+	 */
+	odp_proto_stats_t stat;
+
+	/** Octet counter 0 adjust */
+	int32_t oct_count0_adj;
+
+	/** Octet counter 1 adjust */
+	int32_t oct_count1_adj;
+} odp_packet_proto_stats_opt_t;
+
+/**
+ * Request packet proto stats.
+ *
+ * The statistics enabled in the proto stats object are updated for the packet in
+ * packet output when the packet is transmitted or dropped. The statistics update
+ * is done as the last step of output processing after possible packet
+ * transformations (e.g. fragmentation, IPsec) that may occur. If a packet is
+ * fragmented or segmented to multiple packets as part of output processing, all
+ * the resulting packets inherit the proto stats object association from the
+ * original packet.
+ *
+ * The relevant octet counts will be updated with the actual packet length at
+ * the time of transmission or drop plus the respective adjustment value passed
+ * in the 'opt' parameter. The octet counts thus include possible additional
+ * headers added by ODP during packet output (e.g. ESP header added by inline
+ * outbound IPsec processing). Ethernet padding and FCS are not included in the
+ * octet counts. The adjustment value is added only if the respective capability
+ * field is true and otherwise ignored.
+ *
+ * @param pkt   Packet handle
+ * @param opt   Proto stats options. If NULL, then proto stats update is
+ *              disabled for this packet.
+ *
+ * @see odp_proto_stats_capability_t::tx
+ */
+void odp_packet_proto_stats_request(odp_packet_t pkt, odp_packet_proto_stats_opt_t *opt);
+
+/**
+ * Get proto stats object.
+ *
+ * Get the proto stats object associated with the given packet.
+ *
+ * @param pkt Packet handle
+ *
+ * @return Proto stats object handle or ODP_PROTO_STATS_INVALID if not set.
+ */
+odp_proto_stats_t odp_packet_proto_stats(odp_packet_t pkt);
 
 /*
  *
--- a/include/odp/api/spec/proto_stats.h
+++ b/include/odp/api/spec/proto_stats.h
@@ -25,14 +25,67 @@ extern "C" {
  * Invalid proto stats handle
  */
 
+/** ODP proto stats counters
+ *
+ * Statistics that can be enabled in proto stats object. For Tx stats counters,
+ * Pktout config `odp_pktout_config_opt_t::bit::proto_stats_ena` needs to be
+ * enabled.
+ *
+ * Tx packet and octet sent/drop statistics might include packets sent/dropped via
+ * Traffic Manager or Tx packet Aging or due to any other Tx errors. It is
+ * implementation specific as to what all Tx sent/drop events are accounted for.
+ */
+typedef union odp_proto_stats_counters_t {
+	/** Option flags */
+	struct {
+		/** Tx packet sent count */
+		uint64_t tx_pkts : 1;
+
+		/** Tx packet drop count */
+		uint64_t tx_pkt_drops : 1;
+
+		/** Tx packet sent Octet counter 0 */
+		uint64_t tx_oct_count0 : 1;
+
+		/** Tx packet drop Octet counter 0 */
+		uint64_t tx_oct_count0_drops : 1;
+
+		/** Tx packet sent octet counter 1 */
+		uint64_t tx_oct_count1 : 1;
+
+		/** Tx packet drop octet counter 1 */
+		uint64_t tx_oct_count1_drops : 1;
+	} bit;
+
+	/** All bits of the bit field structure
+	 *
+	 * This field can be used to set/clear all flags, or bitwise
+	 * operations over the entire structure.
+	 */
+	uint64_t all_bits;
+} odp_proto_stats_counters_t;
+
 /** ODP proto stats params */
 typedef struct odp_proto_stats_param_t {
+	/** Stats counters to enable */
+	odp_proto_stats_counters_t counters;
 } odp_proto_stats_param_t;
 
 /**
  * Proto stats capabilities
  */
 typedef struct odp_proto_stats_capability_t {
+	/** Tx capabilities */
+	struct {
+		/** Stats counters supported */
+		odp_proto_stats_counters_t counters;
+
+		/** Packet adjust support for Octet counter 0 */
+		odp_bool_t oct_count0_adj;
+
+		/** Packet adjust support for Octet counter 1 */
+		odp_bool_t oct_count1_adj;
+	} tx;
 } odp_proto_stats_capability_t;
 
 /**
@@ -92,6 +145,13 @@ odp_proto_stats_t odp_proto_stats_lookup(const char *name);
  *
  * Destroy a proto stats object already created.
  *
+ * Before destroying proto stats object having tx statistics enabled,
+ * for all PKTIO devices to which packets were Tx'ed earlier with
+ * this proto stats object, odp_pktio_stop() must be called. Additionally,
+ * existing packets that refer to the proto stats object being destroyed
+ * must not be sent at the same time as or after the proto stats object
+ * destruction.
+ *
  * @param stat Proto stats handle
  *
  * @retval 0 on success
@@ -101,6 +161,23 @@ int odp_proto_stats_destroy(odp_proto_stats_t stat);
 
 /** ODP proto stats counters */
 typedef struct odp_proto_stats_data_t {
+	/** Packet sent count */
+	uint64_t tx_pkts;
+
+	/** Packet drop count */
+	uint64_t tx_pkt_drops;
+
+	/** Packet sent Octet counter 0 */
+	uint64_t tx_oct_count0;
+
+	/** Packet drop Octet counter 0 */
+	uint64_t tx_oct_count0_drops;
+
+	/** Packet sent octet counter 1 */
+	uint64_t tx_oct_count1;
+
+	/** Packet drop octet counter 1 */
+	uint64_t tx_oct_count1_drops;
 } odp_proto_stats_data_t;
 
 /**
--- a/include/odp/api/spec/proto_stats.h
+++ b/include/odp/api/spec/proto_stats.h
@@ -1,0 +1,139 @@
+/* SPDX-License-Identifier: BSD-3-Clause
+ * Copyright(C) 2021 Marvell.
+ */
+
+/**
+ * @file
+ *
+ * ODP Proto Stats
+ */
+
+#ifndef ODP_API_SPEC_PROTO_STATS_H_
+#define ODP_API_SPEC_PROTO_STATS_H_
+#include <odp/visibility_begin.h>
+
+#ifdef __cplusplus
+extern "C" {
+#endif
+
+/** @addtogroup odp_proto_stats
+ *  @{
+ */
+
+/**
+ * @def ODP_PROTO_STATS_INVALID
+ * Invalid proto stats handle
+ */
+
+/** ODP proto stats params */
+typedef struct odp_proto_stats_param_t {
+} odp_proto_stats_param_t;
+
+/**
+ * Proto stats capabilities
+ */
+typedef struct odp_proto_stats_capability_t {
+} odp_proto_stats_capability_t;
+
+/**
+ * Initialize proto stats parameters
+ *
+ * Initialize an odp_proto_stats_param_t to its default values.
+ * By default all the statistics are disabled.
+ *
+ * @param param   Proto stats parameter pointer.
+ */
+void odp_proto_stats_param_init(odp_proto_stats_param_t *param);
+
+/**
+ * Get proto stats capability
+ *
+ * Get supported protocol statistics and metadata for a PKTIO.
+ *
+ * @param      pktio  Packet IO handle
+ * @param[out] capa   Pointer where capabilities are updated
+ *
+ * @retval 0 on success
+ * @retval <0 on failure
+ */
+int odp_proto_stats_capability(odp_pktio_t pktio, odp_proto_stats_capability_t *capa);
+
+/**
+ * Create a proto stats object
+ *
+ * Create a proto stats object with given name and parameters.
+ * A proto stats object can be created with any set of statistics but only the
+ * statistics that are supported by a PKTIO are updated in a proto stats object
+ * for that PKTIO associated packets. Same proto stats object can be used with
+ * any PKTIO.
+ *
+ * @param name  Object name
+ * @param param Proto stats parameters
+ *
+ * @return Proto stats object handle
+ * @retval ODP_PROTO_STATS_INVALID on failure
+ */
+odp_proto_stats_t odp_proto_stats_create(const char *name, const odp_proto_stats_param_t *param);
+
+/**
+ * Lookup a proto stats object by name
+ *
+ * Lookup an already created proto stats object by name.
+ *
+ * @param name Proto stats object name
+ *
+ * @return Proto stats object handle
+ * @retval ODP_PROTO_STATS_INVALID on failure
+ */
+odp_proto_stats_t odp_proto_stats_lookup(const char *name);
+
+/**
+ * Destroy a proto stats object
+ *
+ * Destroy a proto stats object already created.
+ *
+ * @param stat Proto stats handle
+ *
+ * @retval 0 on success
+ * @retval <0 on failure
+ */
+int odp_proto_stats_destroy(odp_proto_stats_t stat);
+
+/** ODP proto stats counters */
+typedef struct odp_proto_stats_data_t {
+} odp_proto_stats_data_t;
+
+/**
+ * Get all proto stats counters
+ *
+ * Get current values of all counters of the proto stats object.
+ * The values of counters that are not enabled in the proto stats object are undefined.
+ *
+ * @param      stat   Proto stats object handle
+ * @param[out] data   Pointer to a caller allocated structure where the statistics will
+ *                    be written to.
+ *
+ * @retval =0 on success
+ * @retval <0 on failure
+ */
+int odp_proto_stats(odp_proto_stats_t stat, odp_proto_stats_data_t *data);
+
+/**
+ * Print proto stats object info to ODP log.
+ *
+ * Print implementation-defined proto stats debug information to ODP log.
+ *
+ * @param stat Proto stats object handle
+ */
+void odp_proto_stats_print(odp_proto_stats_t stat);
+
+/**
+ * @}
+ */
+
+#ifdef __cplusplus
+}
+#endif
+
+#include <odp/visibility_end.h>
+#endif
--- a/test/validation/api/pktio/pktio.c
+++ b/test/validation/api/pktio/pktio.c
@@ -2578,6 +2578,186 @@ static void pktio_test_extra_stats(void)
 	CU_ASSERT(odp_pktio_close(pktio) == 0);
 }
 
+static int pktio_check_proto_statistics_counters(void)
+{
+	odp_proto_stats_capability_t capa;
+	odp_pktio_param_t pktio_param;
+	odp_pktio_t pktio;
+	int ret;
+
+	odp_pktio_param_init(&pktio_param);
+	pktio_param.in_mode = ODP_PKTIN_MODE_SCHED;
+
+	pktio = odp_pktio_open(iface_name[0], pool[0], &pktio_param);
+	if (pktio == ODP_PKTIO_INVALID)
+		return ODP_TEST_INACTIVE;
+
+	ret = odp_proto_stats_capability(pktio, &capa);
+	(void)odp_pktio_close(pktio);
+
+	if (ret < 0 || capa.tx.counters.all_bits == 0)
+		return ODP_TEST_INACTIVE;
+
+	return ODP_TEST_ACTIVE;
+}
+
+static void validate_proto_stats(odp_proto_stats_t stat, odp_packet_proto_stats_opt_t opt,
+				 odp_proto_stats_capability_t capa, uint64_t pkts)
+{
+	odp_proto_stats_data_t data;
+	int ret;
+
+	ret = odp_proto_stats(stat, &data);
+	CU_ASSERT(ret == 0);
+
+	CU_ASSERT(!(capa.tx.counters.bit.tx_pkt_drops && (data.tx_pkt_drops > 0)));
+	CU_ASSERT(!(capa.tx.counters.bit.tx_oct_count0_drops && (data.tx_oct_count0_drops > 0)));
+	CU_ASSERT(!(capa.tx.counters.bit.tx_oct_count1_drops && (data.tx_oct_count1_drops > 0)));
+	CU_ASSERT(!(capa.tx.counters.bit.tx_pkts && (data.tx_pkts != pkts)));
+
+	if (capa.tx.counters.bit.tx_oct_count0) {
+		int64_t counted_bytes = PKT_LEN_NORMAL;
+
+		if (capa.tx.oct_count0_adj)
+			counted_bytes += opt.oct_count0_adj;
+		CU_ASSERT(data.tx_oct_count0 == counted_bytes * pkts);
+	}
+
+	if (capa.tx.counters.bit.tx_oct_count1) {
+		int64_t counted_bytes = PKT_LEN_NORMAL;
+
+		if (capa.tx.oct_count1_adj)
+			counted_bytes += opt.oct_count1_adj;
+		CU_ASSERT(data.tx_oct_count1 == counted_bytes * pkts);
+	}
+}
+
+static void pktio_test_proto_statistics_counters(void)
+{
+	odp_pktio_t pktio_rx, pktio_tx;
+	odp_pktio_t pktio[MAX_NUM_IFACES] = {
+		ODP_PKTIO_INVALID, ODP_PKTIO_INVALID
+	};
+	odp_packet_t pkt;
+	const uint32_t num_pkts = 10;
+	odp_packet_t tx_pkt[num_pkts];
+	uint32_t pkt_seq[num_pkts];
+	odp_event_t ev;
+	int i, pkts, tx_pkts, ret, alloc = 0;
+	odp_pktout_queue_t pktout;
+	uint64_t wait = odp_schedule_wait_time(ODP_TIME_MSEC_IN_NS);
+	uint64_t flow0_pkts = 0, flow1_pkts = 0;
+	odp_proto_stats_capability_t capa;
+	odp_packet_proto_stats_opt_t opt0;
+	odp_packet_proto_stats_opt_t opt1;
+	odp_proto_stats_param_t param;
+	odp_pktio_config_t config;
+	odp_proto_stats_t stat0;
+	odp_proto_stats_t stat1;
+
+	memset(&pktout, 0, sizeof(pktout));
+
+	for (i = 0; i < num_ifaces; i++) {
+		pktio[i] = create_pktio(i, ODP_PKTIN_MODE_SCHED,
+					ODP_PKTOUT_MODE_DIRECT);
+
+		CU_ASSERT_FATAL(pktio[i] != ODP_PKTIO_INVALID);
+	}
+	pktio_tx = pktio[0];
+	pktio_rx = (num_ifaces > 1) ? pktio[1] : pktio_tx;
+
+	/* Enable protocol stats on Tx interface */
+	odp_pktio_config_init(&config);
+	config.pktout.bit.proto_stats_ena = 1;
+	ret = odp_pktio_config(pktio_tx, &config);
+	CU_ASSERT(ret == 0);
+
+	CU_ASSERT_FATAL(odp_pktout_queue(pktio_tx, &pktout, 1) == 1);
+
+	ret = odp_pktio_start(pktio_tx);
+	CU_ASSERT(ret == 0);
+	if (num_ifaces > 1) {
+		ret = odp_pktio_start(pktio_rx);
+		CU_ASSERT(ret == 0);
+	}
+
+	odp_proto_stats_param_init(&param);
+	odp_proto_stats_capability(pktio_tx, &capa);
+	CU_ASSERT(capa.tx.counters.all_bits != 0);
+	param.counters.all_bits = capa.tx.counters.all_bits;
+	/* Create statistics object with all supported counters */
+	stat0 = odp_proto_stats_create("flow0_stat", &param);
+	CU_ASSERT_FATAL(stat0 != ODP_PROTO_STATS_INVALID);
+	stat1 = odp_proto_stats_create("flow1_stat", &param);
+	CU_ASSERT_FATAL(stat1 != ODP_PROTO_STATS_INVALID);
+
+	/* Flow-0 options */
+	opt0.stat = stat0;
+	opt0.oct_count0_adj = 0;
+	/* oct1 contains byte count of packets excluding Ethernet header */
+	opt0.oct_count1_adj = -14;
+
+	/* Flow-1 options */
+	opt1.stat = stat1;
+	opt1.oct_count0_adj = -8;
+	opt1.oct_count1_adj = 14;
+
+	alloc = create_packets(tx_pkt, pkt_seq, num_pkts, pktio_tx, pktio_rx);
+
+	/* Attach statistics object to all Tx packets */
+	for (pkts = 0; pkts < alloc; pkts++) {
+		if ((pkts % 2) == 0) {
+			odp_packet_proto_stats_request(tx_pkt[pkts], &opt0);
+			flow0_pkts++;
+		} else {
+			odp_packet_proto_stats_request(tx_pkt[pkts], &opt1);
+			flow1_pkts++;
+		}
+	}
+
+	/* send */
+	for (pkts = 0; pkts != alloc; ) {
+		ret = odp_pktout_send(pktout, &tx_pkt[pkts], alloc - pkts);
+		if (ret < 0) {
+			CU_FAIL("unable to send packet\n");
+			break;
+		}
+		pkts += ret;
+	}
+	tx_pkts = pkts;
+
+	/* get */
+	for (i = 0, pkts = 0; i < (int)num_pkts && pkts != tx_pkts; i++) {
+		ev = odp_schedule(NULL, wait);
+		if (ev != ODP_EVENT_INVALID) {
+			if (odp_event_type(ev) == ODP_EVENT_PACKET) {
+				pkt = odp_packet_from_event(ev);
+				if (pktio_pkt_seq(pkt) != TEST_SEQ_INVALID)
+					pkts++;
+			}
+			odp_event_free(ev);
+		}
+	}
+
+	CU_ASSERT(pkts == tx_pkts);
+
+	/* Validate Flow-0 packet statistics */
+	validate_proto_stats(stat0, opt0, capa, flow0_pkts);
+
+	/* Validate Flow-1 packet statistics */
+	validate_proto_stats(stat1, opt1, capa, flow1_pkts);
+
+	for (i = 0; i < num_ifaces; i++) {
+		CU_ASSERT(odp_pktio_stop(pktio[i]) == 0);
+		flush_input_queue(pktio[i], ODP_PKTIN_MODE_SCHED);
+		CU_ASSERT(odp_pktio_close(pktio[i]) == 0);
+	}
+
+	/* Destroy proto statistics object */
+	CU_ASSERT(odp_proto_stats_destroy(stat0) == 0);
+	CU_ASSERT(odp_proto_stats_destroy(stat1) == 0);
+}
+
 static int pktio_check_start_stop(void)
 {
 	if (getenv("ODP_PKTIO_TEST_DISABLE_START_STOP"))
@@ -4617,6 +4797,8 @@ odp_testinfo_t pktio_suite_unsegmented[] = {
 	ODP_TEST_INFO_CONDITIONAL(pktio_test_event_queue_statistics_counters,
 				  pktio_check_event_queue_statistics_counters),
 	ODP_TEST_INFO(pktio_test_extra_stats),
+	ODP_TEST_INFO_CONDITIONAL(pktio_test_proto_statistics_counters,
+				  pktio_check_proto_statistics_counters),
 	ODP_TEST_INFO_CONDITIONAL(pktio_test_pktin_ts,
 				  pktio_check_pktin_ts),
 	ODP_TEST_INFO_CONDITIONAL(pktio_test_pktout_ts,
--- a/include/Makefile.am
+++ b/include/Makefile.am
@@ -37,6 +37,7 @@ odpapiinclude_HEADERS = \
 	odp/api/packet_io_stats.h \
 	odp/api/protocols.h \
 	odp/api/pool.h \
+	odp/api/proto_stats.h \
 	odp/api/queue.h \
 	odp/api/queue_types.h \
 	odp/api/random.h \
@@ -92,6 +93,7 @@ odpapispecinclude_HEADERS = \
 		  odp/api/spec/protocols.h \
 		  odp/api/spec/pool.h \
 		  odp/api/spec/pool_types.h \
+		  odp/api/spec/proto_stats.h \
 		  odp/api/spec/queue.h \
 		  odp/api/spec/queue_types.h \
 		  odp/api/spec/random.h \
@@ -144,6 +146,7 @@ odpapiabidefaultinclude_HEADERS = \
 	odp/api/abi-default/packet_types.h \
 	odp/api/abi-default/packet_flags.h \
 	odp/api/abi-default/packet_io.h \
+	odp/api/abi-default/proto_stats.h \
 	odp/api/abi-default/pool.h \
 	odp/api/abi-default/queue.h \
 	odp/api/abi-default/queue_types.h \
@@ -193,6 +196,7 @@ odpapiabiarchinclude_HEADERS = \
 	odp/arch/arm32-linux/odp/api/abi/packet_flags.h \
 	odp/arch/arm32-linux/odp/api/abi/packet_io.h \
 	odp/arch/arm32-linux/odp/api/abi/pool.h \
+	odp/arch/arm32-linux/odp/api/abi/proto_stats.h \
 	odp/arch/arm32-linux/odp/api/abi/queue.h \
 	odp/arch/arm32-linux/odp/api/abi/queue_types.h \
 	odp/arch/arm32-linux/odp/api/abi/rwlock.h \
@@ -237,6 +241,7 @@ odpapiabiarchinclude_HEADERS = \
 	odp/arch/arm64-linux/odp/api/abi/packet_flags.h \
 	odp/arch/arm64-linux/odp/api/abi/packet_io.h \
 	odp/arch/arm64-linux/odp/api/abi/pool.h \
+	odp/arch/arm64-linux/odp/api/abi/proto_stats.h \
 	odp/arch/arm64-linux/odp/api/abi/queue.h \
 	odp/arch/arm64-linux/odp/api/abi/queue_types.h \
 	odp/arch/arm64-linux/odp/api/abi/rwlock.h \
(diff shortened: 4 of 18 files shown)


include/odp/api/spec/proto_stats.h
```diff
@@ -0,0 +1,151 @@
...
+ *
+ * Get pointer to array of 64-bit elements where stats are updated. Index for the array is based
+ * on type of proto stat object and stat id.
+ *
+ * @param stat      Proto stat object handle
+ * @param max_index Pointer to location where max index allowed to base array is returned.
+ *
+ * @return Pointer to base of stats array
+ */
+uint64_t *odp_proto_stats(odp_proto_stats_t stat, uint16_t *max_index);
```
JannePeltonen, line 187: Wouldn't it be better to have a normal stats reading function for each counter rather than exposing this sort of magically updating memory region? How is an application supposed to read the memory location? Using atomic accesses? With what kind of memory ordering?

include/odp/api/spec/proto_stats.h
```diff
@@ -0,0 +1,151 @@
...
+ * Destroy a proto stats object
+ *
+ * Destroy a proto stats object already created.
+ *
+ * @param stat Proto stats handle
+ *
+ * @retval 0 on success
+ * @retval <0 on failure
+ */
+int odp_proto_stats_destroy(odp_proto_stats_t stat);
```
JannePeltonen, line 174: When is it safe to call the destroy? Any time? When no packets that refer to the proto stat are no longer in flight (but how would one know it?)?

include/odp/api/spec/packet.h
```diff
@@ -2140,6 +2141,66 @@ uint32_t odp_packet_payload_offset(odp_packet_t pkt);
...
+ * Set packet specific metadata for proto stats
+ * Per-Packet metadata needs to be updated for modes ODP_PROTO_STATS_MODE_TX_*
+ *
+ * @param pkt   Packet handle
+ * @param op    Proto stats meta data op
+ * @param data  Pointer to meta data specific to op
+ *
+ * @see odp_proto_stats_mode_t
+ */
+void odp_packet_proto_stats_ops_set(odp_packet_t pkt, odp_packet_proto_stats_op_t op, void *data);
```
JannePeltonen, line 2193: How is one supposed to use the data pointer? The object pointed by it does not seem to be defined anywhere.

nithind1988: I missed to document this more in enum's of ``odp_packet_proto_stats_op_t ``
When ``op`` is ODP_PACKET_PROTO_STATS_OP_**S16**_INNER_ADJ data is supposed to be pointer to **S16** i.e int16_t.
When ``op`` is ODP_PACKET_PROTO_STATS_OP_**BOOL**_OUTER_ENABLE data is supposed to be pointer to bool.

include/odp/api/spec/proto_stats.h
```diff
@@ -27,12 +27,73 @@ extern "C" {
...
+	 */
+	ODP_PROTO_STATS_MODE_TX_OCT_DROP_ONLY,
+	/** Tx packet and octet drop mode
+	 * Packet drop count and Octet drop count updated only when packet is dropped.
+	 *
+	 * @see ODP_PROTO_STATS_ID_TX_PKT_DROP
+	 * @see ODP_PROTO_STATS_ID_TX_INNER_IP_DROP_OCT
+	 * @see ODP_PROTO_STATS_ID_TX_OUTER_IP_DROP_OCT
+	 */
+	ODP_PROTO_STATS_MODE_TX_PKT_OCT_DROP_ONLY,
```
JannePeltonen, line 104: This mode struct seems problematic as the number of combinations grows exponentially as the number of available counters grow. Wouldn't it be better to just drop the mode structure and specify the exact counters of interest when the stats object is created? Expressing the capabilities would then not be so straightforward but I am not sure it is necessary to show there which combos are supported as opposed to just failing the stats object create for an unsupported combo.

include/odp/api/spec/packet.h
JannePeltonen: I am not sure this is really in line with the rest of ODP API that does not use void pointers like that but instead uses specified types (c.f. odp_ipsec_test_sa_param_t, odp_pmr_param_t, odp_tm_egress_t, odp_threshold_t).

nithind1988: Ack, will change to individual calls to set the content.

include/odp/api/spec/proto_stats.h
nithind1988: Will change to mask based stats creation in V2.

include/odp/api/spec/proto_stats.h
```diff
@@ -27,12 +27,73 @@ extern "C" {
...
+	/** Packet drop count */
+	ODP_PROTO_STATS_ID_TX_PKT_DROP,
+	/** Inner packet octet count */
+	ODP_PROTO_STATS_ID_TX_INNER_IP_OCT,
+	/** Inner packet octet drop count */
+	ODP_PROTO_STATS_ID_TX_INNER_IP_DROP_OCT,
+	/** Outer packet octet count */
+	ODP_PROTO_STATS_ID_TX_OUTER_IP_OCT,
+	/** Outer packet octet drop count */
+	ODP_PROTO_STATS_ID_TX_OUTER_IP_DROP_OCT,
```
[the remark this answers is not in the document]
nithind1988: Ack, will change inner and outer to counter1 and counter2. In future other platforms can support more counters by adding counter3.

include/odp/api/spec/proto_stats.h
nithind1988: Fixed in V2.

include/odp/api/spec/proto_stats.h
nithind1988: Our thought is to preallocate stats objects and use it. Hence there won't be any segfaults and application can free destroy odp_proto_stats_t anytime. But in order to get the correct updates, they need to stop pktio which ensures all the packets are flushed out.

JannePeltonen:
The commit message of the first commit talks about the layout of the statistics, but the layout is not relevant anymore since the counters are not accessed directly. I would just remove the "Layout of the stats..." sentence and everything after it.

include/odp/api/spec/proto_stats.h
```diff
@@ -0,0 +1,163 @@
...
+
+/** ODP proto stats params */
+typedef struct odp_proto_stats_param_t {
+	/** Stats id mask
+	 *
+	 * Each bit corresponds to one ODP_PROTO_STATS_ID_* and
+	 * bit position is same as ODP_PROTO_STATS_ID_*.
+	 */
+	uint64_t stats_mask;
+} odp_proto_stats_param_t;
```
JannePeltonen, line 68: This should say that the bits set in stats_mask define which statistics are enabled. Now it explains which bit is which but. not what the bits do.

include/odp/api/spec/proto_stats.h
```diff
@@ -0,0 +1,163 @@
...
+ * Proto stats capabilities
+ */
+typedef struct odp_proto_stats_capa_t {
+	/** Supported proto stats id mask
+	 *
+	 * A bit is set at position ODP_PROTO_STATS_ID_*,
+	 * when that proto stats id is supported.
+	 */
+	uint64_t stats_mask_supported;
+} odp_proto_stats_capa_t;
```
JannePeltonen, line 129: This should be named odp_proto_stats_capability_t for similarity with the rest of the ODP API.

Similarly, odp_proto_stats_capa() should be odp_proto_stats_capability().

include/odp/api/spec/proto_stats.h
```diff
@@ -0,0 +1,163 @@
...
+
+/**
+ * Initialize proto stats parameters
+ *
+ * Initialize an odp_proto_stats_param_t to its default values.
+ *
+ * @param param   Proto stats parameter pointer.
+ */
+void odp_proto_stats_init(odp_proto_stats_param_t *param);
```
JannePeltonen, line 138: This should be named odp_proto_stats_param_init() as it is initializing the parameter structure and not the proto stats object.

JannePeltonen: The default values of the parameter structure have not been speficied here nor at the structure definition. I would suggest adding a statement that by default all statistics are disabled.

include/odp/api/spec/proto_stats.h
```diff
@@ -0,0 +1,163 @@
...
+ *
+ * @param capa Pointer where capabilities are updated.
+ */
+void odp_proto_stats_capa(odp_proto_stats_capa_t *capa);
+
+/**
+ * Create a proto stats object
+ *
+ * Create a proto stats object with given name and mode.
+ * Semantics of each mode is associated with that mode.
```
JannePeltonen, line 153: There are no modes anymore in this version of the API draft, so this text should be updated. Maybe: ...name and parameters. Or just : ...name.

include/odp/api/spec/proto_stats.h
```diff
@@ -0,0 +1,163 @@
...
+odp_proto_stats_t odp_proto_stats_create(const char *name, const odp_proto_stats_param_t *param);
+
+/**
+ * Lookup a proto stats object by name
+ *
+ * Lookup an already created proto stats object by name.
+ *
+ * @param name Proto stats object name
+ *
+ * @return Proto stats object handle
```
JannePeltonen, line 169: And return ODP_PROTO_STATS_INVALID in case of lookup failure.

include/odp/api/spec/proto_stats.h
```diff
@@ -0,0 +1,163 @@
...
+	/** Array of counters */
+	uint64_t counter[ODP_PROTO_STATS_ID_MAX];
+} odp_proto_stats_data_t;
+
+/**
+ * Get all proto stats counters
+ *
+ * Get current values of all counters of the proto stats object. Only the counters
+ * enabled on the proto stats object will have valid value in odp_proto_stats_data_t
+ * and rest of the counters will be uninitialized.
```
JannePeltonen, line 209: I do not think there is a need to promise not to write in the counter fields that are not enabled. I would rephrase: The values of counters that are not enabled in the proto stats object are undefined.

include/odp/api/spec/proto_stats.h
```diff
@@ -0,0 +1,163 @@
...
+ *
+ * @param      stat   Proto stats object handle
+ * @param[out] data   Pointer to structure where all the stats counters
+ *                    are returned.
```
JannePeltonen, line 213: I would rephrase: Pointer to a caller allocated structure where the statistic will be written to.

include/odp/api/spec/proto_stats.h
```diff
@@ -27,6 +27,26 @@ extern "C" {
 
 /** ODP proto stats id */
 typedef enum odp_proto_stats_id_t {
+	/** Tx proto stats id ODP_PROTO_STATS_ID_TX_*
+	 *
+	 * - All the below modes are associated to packet stats and accounting
+	 *   on Tx of a packet via Packet IO. Packet IO pktout config
+	 *   `odp_pktout_config_opt_t::bit::proto_stats` needs to be enabled for
+	 *   the offload to work.
+	 */
```
JannePeltonen, line 42: Since we do not have modes anymore, I would rephrase this simply as: Statistics that can be enabled in proto stats objects.

include/odp/api/spec/packet.h
```diff
@@ -2140,6 +2141,57 @@ uint32_t odp_packet_payload_offset(odp_packet_t pkt);
  */
 int odp_packet_payload_offset_set(odp_packet_t pkt, uint32_t offset);
 
+/**
+ * Set proto stats object.
+ *
+ * Associate a proto stats object to a packet. This enables updating protocol
+ * stats of the packet in the given object after packet is enqueued to queue
+ * and reaches a terminal state i.e either transmit or drop.
```
JannePeltonen, line 2149: The "enqueued to queue" part is misleading since the stats are not updated in all ODP queues but only when the packet is leaving the system through a pktout.

It would also be good the specify what happens when a packet gets fragmented (e.g. in inline IPsec output) or fragmented/segmented (through the LSO mechanism).

I would rephrase:

Associate a packet with a proto stats object. The statistics enabled in the proto stats object are updated for the packet in packet output when the packet is transmitted or dropped. The statistics update is done as the last step of output processing after possible packet transformations (e.g. fragmentation, IPsec) that may occur. If a packet is fragmented or segmented to multiple packets as part of output processing, all the resulting packets inherit the proto stats object association from the original packet.

include/odp/api/spec/packet.h
```diff
@@ -2140,6 +2141,57 @@ uint32_t odp_packet_payload_offset(odp_packet_t pkt);
...
+ *                 0: Octet counter update is enabled.
+ *                 If per-packet octet counter enable/disable is not supported,
+ *                 this field is ignored.
+ *
+ * @see odp_proto_stats_capa_t::tx::oct_count0_adj
+ * @see odp_proto_stats_capa_t::tx::oct_count1_adj
+ * @see odp_proto_stats_capa_t::tx::oct_count0_ctrl
+ * @see odp_proto_stats_capa_t::tx::oct_count1_ctrl
+ */
+void odp_packet_proto_stats_oct_set(odp_packet_t pkt, uint8_t counter, int adjust, int disable);
```
JannePeltonen, line 2193: The disable parameter seems to make the API a bit baroque. It seems to me that it is not really needed, since instead of marking some packets with the disable flag, an application could simply create two proto stats objects, one with the counter enabled and one with the counter disabled. when reading the counters, the application would read both proto stats objects and add their counters. The end result would be the same, without having to have the disable flag here and in the capabilities and the resulting API complexity.

Does this API have the problem of requiring too many function calls in common cases? If I use the two available counters, I likely need three calls (stats proto set and the adjustment value set for both counters) for every packet. Maybe there should be a function that could do all those in one go. If that function took its parameters in a struct, then it would be straightforward to have a version that sets the metadata for multiple packets: e.g. odp_packet_proto_stats_set_multi(odp_packet_t pkt[], odp_packet_proto_stats_param_t param[], int num) or something like that.

include/odp/api/spec/packet.h
```diff
@@ -2140,6 +2141,57 @@ uint32_t odp_packet_payload_offset(odp_packet_t pkt);
...
+ * updating octet counter in proto stats object. By default adjust value will be
+ * zero and octet counter update is enabled. In other words,
+ *
+ * if (sent)
+ *    counter[ODP_PROTO_STATS_ID_TX_OCT_COUNTx] += `odp_packet_len() + adjust`
+ * else if (drop)
+ *    counter[ODP_PROTO_STATS_ID_TX_OCT_COUNTx_DROP] += `odp_packet_len() + adjust`
+ *
+ * @param pkt      Packet handle
+ * @param counter  Non-zero counter index when more than 1 octet counter is supported.
```
JannePeltonen, line 2181: This would mean that if only one counter is supported, the index value is ignored. I think it would be better to require the index value always match the counter ID, regardless of how many counters are supported.

include/odp/api/spec/proto_stats.h
```diff
@@ -33,6 +33,12 @@ typedef enum odp_proto_stats_id_t {
 	 *   on Tx of a packet via Packet IO. Packet IO pktout config
 	 *   `odp_pktout_config_opt_t::bit::proto_stats` needs to be enabled for
 	 *   the offload to work.
+	 * - For each of the below mode, proto stats object needs to be associated
+	 *   on per-packet basis using `odp_packet_proto_stats_set(pkt, stat)`
+	 *   for stats update to be enabled for that packet.
+	 * - Default behaviour is stats update is disabled.
+	 *
+	 * @see odp_packet_proto_stats_set()
```
JannePeltonen, line 41: I am not sure this text update is needed at all.

include/odp/api/spec/proto_stats.h
```diff
@@ -71,6 +77,55 @@ typedef struct odp_proto_stats_capa_t {
 	 * when that proto stats id is supported.
 	 */
 	uint64_t stats_mask_supported;
+
+	/** ODP_PROTO_STATS_ID_TX_* flags */
+	struct {
+		/** Packet adjust support for ODP_PROTO_STATS_ID_TX_OCT_COUNT0
+		 *
+		 * By default octet counter is updated using `odp_packet_len()`.  When adjust
```
JannePeltonen, line 85: > By default octet counter is updated using `odp_packet_len()`

...at the time of the final packet transmit or drop.

I mean, is it clear that the octet count is the final octet count (after e.g. IPsec transformation) and not odp_packet_len() of some earlier moment?

include/odp/api/spec/proto_stats.h
```diff
@@ -71,6 +77,55 @@ typedef struct odp_proto_stats_capa_t {
...
+		 *
+		 * Normally when proto stats object is created with
+		 * ODP_PROTO_STATS_ID_TX_OCT_COUNT0, octet counter 0 updates are enabled for packets
+		 * associated with stats object. This capability says if additionally octet
+		 * counter 0 update alone can be disabled for this packet while keeping all other
+		 * updates enabled.
+		 *
+		 * @see odp_packet_proto_stats_oct_set()
+		 */
+		odp_bool_t oct_count0_ctrl;
```
JannePeltonen, line 104: These would not be needed if we got rid of the disable flag as suggested.

include/odp/api/spec/proto_stats.h
nithind1988: Ack, I fixed the comment in V3. It is the final packet transmit length after transformations.

include/odp/api/spec/packet.h
nithind1988: Removed disable option in v4. Also made multiple packet API's to single as odp_packet_proto_stats_request() similar to LSO/Tx completion. which takes odp_packet_proto_stats_opt_t containing a both object handle and adjust values for octet counter 0/1.

include/odp/api/spec/proto_stats.h
JannePeltonen: If I create a proto stats object, associate a packet with it, then delete the proto stats object and create a new one, it is possible that the packet that still refers to the old, now stale, proto stats object might end up updating the newly added proto stats object and screw up its statistics? I would suppose preventing this would require some extra logic in the ODP implementation.

include/odp/api/spec/proto_stats.h
```diff
@@ -0,0 +1,166 @@
...
+ * By default all the statistics are disabled.
+ *
+ * @param param   Proto stats parameter pointer.
+ */
+void odp_proto_stats_param_init(odp_proto_stats_param_t *param);
+
+/**
+ * Get proto stats capability
+ *
+ * Get proto stats capability of supported modes.
```
JannePeltonen, line 117: The "of supported modes" part seems to be a leftover from earlier versions. It does not mean anything anymore and could be removed.

include/odp/api/spec/packet.h
```diff
@@ -2290,6 +2291,59 @@ int odp_packet_tx_compl_request(odp_packet_t pkt, const odp_packet_tx_compl_opt_
...
+ * the resulting packets inherit the proto stats object association from the
+ * original packet.
+ *
+ * if (sent)
+ *    counter[ODP_PROTO_STATS_ID_TX_OCT_COUNTx] += `odp_packet_len() + adjust`
+ * else if (drop)
+ *    counter[ODP_PROTO_STATS_ID_TX_OCT_COUNTx_DROP] += `odp_packet_len() + adjust`
+ *
+ * @param pkt   Packet handle
+ * @param opt   Proto stats options. If NULL is set, then proto stats update is
```
JannePeltonen, line 2328: "If NULL is set, then..." should be "If NULL, then ..."

include/odp/api/spec/packet.h
```diff
@@ -2290,6 +2291,59 @@ int odp_packet_tx_compl_request(odp_packet_t pkt, const odp_packet_tx_compl_opt_
...
+ * is done as the last step of output processing after possible packet
+ * transformations (e.g. fragmentation, IPsec) that may occur. If a packet is
+ * fragmented or segmented to multiple packets as part of output processing, all
+ * the resulting packets inherit the proto stats object association from the
+ * original packet.
+ *
+ * if (sent)
+ *    counter[ODP_PROTO_STATS_ID_TX_OCT_COUNTx] += `odp_packet_len() + adjust`
+ * else if (drop)
+ *    counter[ODP_PROTO_STATS_ID_TX_OCT_COUNTx_DROP] += `odp_packet_len() + adjust`
```
JannePeltonen, line 2325: I think odp_packet_len() here is misleading since one can understand it to mean odp_packet_len(pkt) at the time of the call.

Maybe the pseudocode could be replaces with a text like this:

"The relevant octet counts will be updated with the actual packet lenght at the time of transmission or drop plus the respective adjustment value passed in the 'opt' parameter. The octet counts thus include possible additional headers added by ODP during packet output (e.g. ESP headr added by inline outbound IPsec processing). Ethernet padding and FCS are not included in the octet counts. The adjustment value is added only if the respective capability field is true and otherwise ignored."

include/odp/api/spec/proto_stats.h
nithind1988: Yes. If we have to handle synchronization in ODP implementation, it would be requiring costly  logic either in implementation and impacts fast path perf.

JannePeltonen: Such synchronization would seem to be needed with the current API draft since the API seems to allow calling the destroy at any time. If this is not ok, then you probably want specify in the API when the proto stats object may be destroyed.

nithind1988: Ack, I'll update the documentation.

include/odp/api/spec/proto_stats.h
```diff
@@ -104,6 +123,12 @@ odp_proto_stats_t odp_proto_stats_lookup(const char *name);
  *
  * Destroy a proto stats object already created.
  *
+ * - Before destroying proto stats object having ODP_PROTO_STATS_ID_TX_*
+ *   enabled, all PKTIO devices to which packets were Tx'ed earlier with
+ *   this proto stats object need to be stopped. This is to make sure that
+ *   all the packets in Tx are flushed and there are no inflight packets
+ *   referring to this object.
```
JannePeltonen, line 139: I am nos sure if this is adequate. What about packets inside TM or IPsec? Stopping a pktio device does not make sure that there are no packets in flight inside TM or inline outbound IPsec.

nithind1988: For TM/Outbound IPSec, isn't it equivalent to stopping the PKTIO while TM/Outbound IPSec is running ?

nithind1988: I mean, platform should already be handling by dropping existing packets and/or reporting error on enqueuing.

include/odp/api/spec/proto_stats.h
```diff
@@ -27,6 +27,25 @@ extern "C" {
...
+	 *
+	 * - Statistics that can be enabled in proto stats objects for Tx of a packet via
+	 *   Packet IO. Packet IO pktout config `odp_pktout_config_opt_t::bit::proto_stats_ena`
+	 *   needs to be enabled forthe offload to work.
+	 */
+
+	/** Packet sent count */
+	ODP_PROTO_STATS_ID_TX_PKT,
+	/** Packet drop count */
+	ODP_PROTO_STATS_ID_TX_PKT_DROP,
```
JannePeltonen, line 40: I started wondering if the API is too ambiguous regarding what packet drops update the counter. My (maybe unfounded) assumption has been that the drop counter gets incremented if the packet gets dropped in the output path for any reason, including TM and IPsec. In other words, whenever the sent counter is not increased, the drop count is. But I am not sure if the API actually says that or whether that is even the intention.

Even if we wanted to count all drops regardless of where they occur, I am not sure if inline outbound IPsec really drops many packets as opposed to returning them to the application through an error event.

include/odp/api/spec/proto_stats.h
JannePeltonen: I am not sure how exactly odp_pktio_stop() and TM & IPsec are supposed to interact. For odp_pktio_stop() the spec says: "Packets already stored for transmit are not freed. A following odp_packet_start() call restarts packet receive and transmit.". So if the in-flight packets are not freed (but in IPsec and TM cases they actually might, in linux-gen), then calling odp_pktio_stop() does not flush the Tx as the proposed API text now says.

include/odp/api/spec/proto_stats.h
nithind1988: I agree. In our platform, ODP_PROTO_STATS_ID_TX_PKT_DROP will be counting only packets either dropped due to TM shaping/ratelimiting or packet Aging. It doesn't count drops due to Outbound IPSec processing or other cases.

But can't it be left this way as it is really platform architecture dependent ?

nithind1988: Updated the text to clearly state that what all kind of Tx pkt statistics are counted is implementation defined in next version. I hope it is fine.

include/odp/api/spec/proto_stats.h
nithind1988: Modified the API text to mandate odp_pktio_stop() before destroying. It can be implementation defined as they will have to make sure proto stats object destroy is safe to be called.

include/odp/api/spec/proto_stats.h
```diff
@@ -104,6 +128,10 @@ odp_proto_stats_t odp_proto_stats_lookup(const char *name);
  *
  * Destroy a proto stats object already created.
  *
+ * - Before destroying proto stats object having ODP_PROTO_STATS_ID_TX_*
+ *   enabled, for all PKTIO devices to which packets were Tx'ed earlier with
+ *   this proto stats object, odp_pktio_stop() must be called.
```
JannePeltonen, line 142: This sounds ok, but maybe this should specify an additional restriction that existing packets that refer to the proto stats object being destroyed, must not be sent.

JannePeltonen: I meant, ...must not be sent at the same time as or after the proto stats object destruction.

JannePeltonen:

include/odp/api/spec/proto_stats.h
```diff
@@ -0,0 +1,205 @@
...
+ *
+ * @retval =0 on success
+ * @retval <0 on failure
+ */
+int odp_proto_stats_all(odp_proto_stats_t stat, odp_proto_stats_data_t *data);
+
+/**
+ * Print proto stats object info on console.
+ *
+ * Print implementation-defined proto stats debug information to the console.
```
psavol, line 190: These debug print functions print "to the ODP log", not to the console. Application may override the log function, and that function may not output to the console, but into file, etc.

include/odp/api/spec/proto_stats.h
```diff
@@ -0,0 +1,205 @@
...
+ * @retval ODP_PROTO_STATS_INVALID on failure
+ */
+odp_proto_stats_t odp_proto_stats_lookup(const char *name);
+
+/**
+ * Destroy a proto stats object
+ *
+ * Destroy a proto stats object already created.
+ *
+ * - Before destroying proto stats object having ODP_PROTO_STATS_ID_TX_*
```
psavol, line 139: Please remove the bullet "-" and the extra indentation.

include/odp/api/spec/proto_stats.h
```diff
@@ -0,0 +1,205 @@
...
+typedef enum odp_proto_stats_id_t {
+	/** Tx proto stats id ODP_PROTO_STATS_ID_TX_*
+	 *
+	 * - Statistics that can be enabled in proto stats objects for Tx of a packet via
+	 *   Packet IO. Packet IO pktout config `odp_pktout_config_opt_t::bit::proto_stats_ena`
+	 *   needs to be enabled for the offload to work.
+	 *
+	 * - Tx packet and octet sent/drop statistics might include packets sent/dropped via
+	 *   Traffic Manager or Tx packet Aging or due to any other Tx errors. It is
+	 *   implementation specific as to what all Tx sent/drop events are accounted for.
```
psavol, line 38: I think text is readable without the bullets "-". Not sure how those display on doxygen documentation. Could remove those.

include/odp/api/spec/proto_stats.h
```diff
@@ -0,0 +1,205 @@
...
+ * The values of counters that are not enabled in the proto stats object are undefined.
+ *
+ * @param      stat   Proto stats object handle
+ * @param[out] data   Pointer to a caller allocated structure where the statistics will
+ *                    be written to.
+ *
+ * @retval =0 on success
+ * @retval <0 on failure
+ */
+int odp_proto_stats_all(odp_proto_stats_t stat, odp_proto_stats_data_t *data);
```
psavol, line 185: In order to match with other stats APIs, this could be named odp_proto_stats(). The single counter read function could be then odp_proto_stats_counter().

include/odp/api/spec/proto_stats.h
```diff
@@ -0,0 +1,205 @@
...
+ * @return Stats counter value
+ * @retval Unknown value if stats id is not valid for the object.
+ */
+uint64_t odp_proto_stats(odp_proto_stats_t stat, odp_proto_stats_id_t id);
+
+/** ODP proto stats counters */
+typedef struct odp_proto_stats_data_t {
+	/** Array of counters */
+	uint64_t counter[ODP_PROTO_STATS_ID_MAX];
+} odp_proto_stats_data_t;
```
psavol, line 170: In order to match with other stats APIs, this struct could list each counter by its name. Application would not need to use counter ID enum for reading counters. Currently, enum fixes the order of the counters anyway.

include/odp/api/spec/proto_stats.h
```diff
@@ -0,0 +1,205 @@
...
+typedef struct odp_proto_stats_param_t {
+	/** Stats id mask
+	 *
+	 * Each bit corresponds to one ODP_PROTO_STATS_ID_* and
+	 * bit position is same as ODP_PROTO_STATS_ID_*.
+	 * Bits set in this mask are the stat counters to be enabled
+	 * in the created proto stats object.
+	 */
+	uint64_t stats_mask;
+} odp_proto_stats_param_t;
```
psavol, line 67: It would match better other stats APIs, mask can be build with a bit field or if enum values would be directly the bit positions:
mask.all = 0;
mask.tx_pkt =1;
mask.tx_pkt_drop = 1;
OR
mask = ODP_PROTO_STATS_ID_TX_PKT | ODP_PROTO_STATS_ID_TX_PKT_DROP;

CURRENTLY:
mask =  1 << ODP_PROTO_STATS_ID_TX_PKT | 1 << ODP_PROTO_STATS_ID_TX_PKT_DROP;

I think the current proposal is less user friendly than the other two. The bit field would match other APIs, but may not produce simplest implementation of the single counter read function.  On the other hand, how often a single counter would be read - could we just remove the single counter read function.

nithind1988: Checked with Transport team. Read all seems to be sufficient. I'll change the mask to bitfield as you mentioned above.

include/odp/api/spec/proto_stats.h
```diff
@@ -25,14 +25,67 @@ extern "C" {
...
+ *
+ * Tx packet and octet sent/drop statistics might include packets sent/dropped via
+ * Traffic Manager or Tx packet Aging or due to any other Tx errors. It is
+ * implementation specific as to what all Tx sent/drop events are accounted for.
+ */
+typedef union odp_proto_stats_tx_opt_t {
+	/** Option flags */
+	struct {
+		/** Packet sent count */
+		int64_t pkts : 1;
```
psavol, line 42: type should be uint64_t

include/odp/api/spec/packet.h
```diff
@@ -2290,6 +2291,61 @@ int odp_packet_tx_compl_request(odp_packet_t pkt, const odp_packet_tx_compl_opt_
...
+ */
+typedef struct odp_packet_proto_stats_opt_t {
+	/** Packet proto stats object handle
+	 *
+	 * Stats in the packet proto stats object will be updated.
+	 */
+	odp_proto_stats_t stat;
+
+	/** Octet counter 0 adjust */
+	int oct_count0_adj;
```
psavol, line 2305: Other packet API offsets and lengths are uint32_t type. Similarly, this could be int32_t. In theory, "int" could be 16 bits, but in practice it is 32 bits. So, using int32_t would results the same size, but would document it explicitly.

include/odp/api/spec/proto_stats.h
```diff
@@ -84,6 +137,13 @@ odp_proto_stats_t odp_proto_stats_lookup(const char *name);
  *
  * Destroy a proto stats object already created.
  *
+ * Before destroying proto stats object having ODP_PROTO_STATS_ID_TX_*
+ * enabled, for all PKTIO devices to which packets were Tx'ed earlier with
```
psavol, line 141: ODP_PROTO_STATS_ID_TX_ enums are not there anymore.

include/odp/api/spec/proto_stats.h
```diff
@@ -25,14 +25,67 @@ extern "C" {
...
+		uint64_t oct_count1_drop : 1;
+	} bit;
+
+	/** All bits of the bit field structure
+	 *
+	 * This field can be used to set/clear all flags, or bitwise
+	 * operations over the entire structure.
+	 */
+	uint64_t all_bits;
+} odp_proto_stats_tx_opt_t;
```
psavol, line 66: I'd prefer that the options/capa would have same naming convention (/hierarchy) that the counter data struct. Since odp_proto_stats_data_t is flat, this could be simply e.g. odp_proto_stats_counters_t (or odp_proto_stats_counter_flags_t ,etc) and contain all counter flags flat (tx_pkts, tx_pkt_drops, ...).

If this struct will just list all counters (and no other options), then "opt" may not be optimal name for it.

include/odp/api/spec/proto_stats.h
```diff
@@ -0,0 +1,131 @@
...
+void odp_proto_stats_param_init(odp_proto_stats_param_t *param);
+
+/**
+ * Get proto stats capability
+ *
+ * Get supported protocol statistics and metadata.
+ *
+ * @param capa Pointer where capabilities are updated.
+ */
+void odp_proto_stats_capability(odp_proto_stats_capability_t *capa);
```
psavol, line 108: Is there need to link proto stats support to a pktio interface? Can all TX interfaces do the counting.

Normal pktio stats are calculated per interface, and the capa is per interface:
int odp_pktio_capability(odp_pktio_t pktio, odp_pktio_capability_t *capa);

nithind1988: PKTIO already has a capabilitiy and config in odp_pktout_config_opt_t::proto_stats_ena which expresses capability of that pktio interface to support protocol stats.

JannePeltonen: I think Petri's comment is valid since the capabilities may be different in different interfaces, not simply so that all interfaces that support some proto stats, support the exact same set of stats.

So, I think it would make sense to have the capability call take pktio as a parameter also in this case.

Then maybe we should also say that one can request any of the stats for a packet regardless of the capabilities, but the ones not supported by the particular pktio will not be updated. This way an app would not have to carefully set the statistics metadata in packets in an interface specific manner.

include/odp/api/spec/proto_stats.h
```diff
@@ -0,0 +1,136 @@
...
+
+/**
+ * Get proto stats capability
+ *
+ * Get supported protocol statistics and metadata for a PKTIO.
+ *
+ * @param      pktio  Packet IO handle
+ * @param[out] capa   Pointer where capabilities are updated
+ */
+void odp_proto_stats_capability(odp_pktio_t pktio, odp_proto_stats_capability_t *capa);
```
psavol, line 109: Return value of success (0) vs fail (<0) is needed now, since application may pass a bad pktio handle.

include/odp/api/spec/packet.h
```diff
@@ -2290,6 +2291,61 @@ int odp_packet_tx_compl_request(odp_packet_t pkt, const odp_packet_tx_compl_opt_
...
+ * is done as the last step of output processing after possible packet
+ * transformations (e.g. fragmentation, IPsec) that may occur. If a packet is
+ * fragmented or segmented to multiple packets as part of output processing, all
+ * the resulting packets inherit the proto stats object association from the
+ * original packet.
+ *
+ * The relevant octet counts will be updated with the actual packet length at
+ * the time of transmission or drop plus the respective adjustment value passed
+ * in the 'opt' parameter. The octet counts thus include possible additional
+ * headers added by ODP during packet output (e.g. ESP headr added by inline
```
psavol, line 2325: Typo: ESP header

psavol:
v10 looks good. Only two comments: missing capa function return value and a typo.

nithind1988:
@JannePeltonen @psavol I hope these changes are final. We will start our platform implementation and validation test suite based on this spec.
cc: @ksatha

psavol:
API spec in v11 is OK

test/validation/api/pktio/pktio.c
```diff
@@ -2578,6 +2578,152 @@ static void pktio_test_extra_stats(void)
...
+	memset(&pktout, 0, sizeof(pktout));
+	odp_pktio_config_init(&config);
+	config.pktout.bit.proto_stats_ena = 1;
+
+	for (i = 0; i < num_ifaces; i++) {
+		pktio[i] = create_pktio(i, ODP_PKTIN_MODE_SCHED,
+					ODP_PKTOUT_MODE_DIRECT);
+
+		CU_ASSERT_FATAL(pktio[i] != ODP_PKTIO_INVALID);
+		odp_pktio_config(pktio[i], &config);
```
JannePeltonen, line 2636: This sets proto_stats_ena also for pktio[1] even though it is needed only for pktio[0] and the capability is checked only for pktio[0].

test/validation/api/pktio/pktio.c
```diff
@@ -2578,6 +2578,152 @@ static void pktio_test_extra_stats(void)
...
+		opt.stat = stat;
+		opt.oct_count0_adj = 0;
+		/* oct1 contains byte count of packets excluding Ethernet header */
+		opt.oct_count1_adj = -14;
+		odp_packet_proto_stats_request(tx_pkt[pkts], &opt);
+	}
+
+	/* send */
+	for (pkts = 0; pkts != alloc; ) {
+		ret = odp_pktout_send(pktout, &tx_pkt[pkts], alloc - pkts);
```
JannePeltonen, line 2671: This code makes the assumption that there is enough packet buffering in the pktout, pktin and scheduler combined that it is possible to send 1000 packets before trying to receive the first one of them, without any packet loss. I think this is a bit fragile and should not be done, even though some existing code seems to do the same thing.

I do not see the need for 1000 packets though. I think 10 should be fine, and making the above assumption regarding 10 packets rather than 1000 would be much more reasonable.

test/validation/api/pktio/pktio.c
```diff
@@ -2578,6 +2578,152 @@ static void pktio_test_extra_stats(void)
...
+
+	CU_ASSERT(pkts == tx_pkts);
+
+	ret = odp_proto_stats(stat, &data);
+	CU_ASSERT(ret == 0);
+	CU_ASSERT(!(capa.tx.counters.bit.tx_pkts && (data.tx_pkts != (uint64_t)pkts)));
+	CU_ASSERT(!(capa.tx.counters.bit.tx_pkt_drops && (data.tx_pkt_drops > 0)));
+	CU_ASSERT(!(capa.tx.counters.bit.tx_oct_count0_drops && (data.tx_oct_count0_drops > 0)));
+	CU_ASSERT(!(capa.tx.counters.bit.tx_oct_count1_drops && (data.tx_oct_count1_drops > 0)));
+	if (capa.tx.oct_count0_adj) {
```
JannePeltonen, line 2701: I think these tx octest counter tests are not very readable. Something like this would seem better, wouldn't it:
```
        if (capa.tx.counters.bit.tx_oct_count0) {
                uint64_t counted_bytes = PKT_LEN_NORMAL;

                if (capa.tx.oct_count0_adj)
                        counted_bytes += opt.oct_count0_adj;
                CU_ASSERT(data.tx_oct_count0 != counted_bytes * (uint64_t)pkts);
        }
```

test/validation/api/pktio/pktio.c
```diff
@@ -2578,6 +2578,152 @@ static void pktio_test_extra_stats(void)
...
+		ret = odp_pktio_start(pktio_rx);
+		CU_ASSERT(ret == 0);
+	}
+
+	odp_proto_stats_param_init(&param);
+	odp_proto_stats_capability(pktio_tx, &capa);
+	CU_ASSERT(capa.tx.counters.all_bits != 0);
+	param.counters.all_bits = capa.tx.counters.all_bits;
+	/* Create statistics object with all supported counters */
+	stat = odp_proto_stats_create("flow0_stat", &param);
```
JannePeltonen, line 2655: I think the test would be much better if it created two proto stats objects instead of one and associated packets with both objects. That way the per-packet association with proto stats objects would get tested better.

test/validation/api/pktio/pktio.c
ksatha: Will remove for pktio[1]

test/validation/api/pktio/pktio.c
ksatha: Thanks for your suggestions, considering below comment will change test application to send only 10 packets, and create two objects (1) one object associated with even packets (0, 2, 4, 6,8) referred as flow-0 and (2) other object associated with odd packets (1, 3, 5, 7, 9) referred as flow-1. Will validate flow-0 and flow-1 statistics accordingly.

test/validation/api/pktio/pktio.c
ksatha: ACK

test/validation/api/pktio/pktio.c
ksatha: Thanks for suggestion, will address in next version.

JannePeltonen:
The capability API lacks info about the maximum number of protocol stat objects that can be created. Maybe it would be good to have it. But let's not put it in this PR since this has been in the API-spec-ok state for some time. Such capability field can be added through a separate PR if it is deemed useful.

test/validation/api/pktio/pktio.c
```diff
@@ -2578,6 +2578,202 @@ static void pktio_test_extra_stats(void)
 	CU_ASSERT(odp_pktio_close(pktio) == 0);
 }
 
+/* Per packet protocol statistics are platform specific,
+ * skip pktio tests when underline platform not supports
```
JannePeltonen, line 2582: This comment has a few language problems. I suggest just removing it as the code is quite obvious anyway.

test/validation/api/pktio/pktio.c
```diff
@@ -2578,6 +2578,202 @@ static void pktio_test_extra_stats(void)
...
+
+static void pktio_test_proto_statistics_counters(void)
+{
+	odp_pktio_t pktio_rx, pktio_tx;
+	odp_pktio_t pktio[MAX_NUM_IFACES] = {
+		ODP_PKTIO_INVALID, ODP_PKTIO_INVALID
+	};
+	odp_packet_t pkt;
+	odp_packet_t tx_pkt[1000];
+	uint32_t pkt_seq[1000];
```
JannePeltonen, line 2615: You forgot to change the size of these arrays when changing the number of packets to 10. That also shows that the magic number for the number of packets is not maintainable. I suggest you define something like this and use it instead of "10" in the code:

const uint64_t num_pkts = 10;

test/validation/api/pktio/pktio.c
```diff
@@ -2578,6 +2578,202 @@ static void pktio_test_extra_stats(void)
...
+	odp_event_t ev;
+	int i, pkts, tx_pkts, ret, alloc = 0;
+	odp_pktout_queue_t pktout;
+	uint64_t wait = odp_schedule_wait_time(ODP_TIME_MSEC_IN_NS);
+	int64_t flow0_pkts = 0, flow1_pkts = 0;
```
JannePeltonen, line 2620: What is the point of defining these as signed types and then casting them to unsigned types later? Just define these as unsigned types and get rid of the casts.

test/validation/api/pktio/pktio.c
```diff
@@ -2578,6 +2578,202 @@ static void pktio_test_extra_stats(void)
...
+	config.pktout.bit.proto_stats_ena = 1;
+
+	for (i = 0; i < num_ifaces; i++) {
+		pktio[i] = create_pktio(i, ODP_PKTIN_MODE_SCHED,
+					ODP_PKTOUT_MODE_DIRECT);
+
+		CU_ASSERT_FATAL(pktio[i] != ODP_PKTIO_INVALID);
+		/* Enable protocol stats on interface 0 (Tx interface) */
+		if (!i)
+			odp_pktio_config(pktio[i], &config);
```
JannePeltonen, line 2642: There is no point in putting this inside the loop when it is done only for the first item. And the return value should be checked. So code like this after the loop and after assigning pktio_tx and pktio_rx would be better:

```
                /* Enable protocol stats Tx interface */
                odp_pktio_config_init(&config);
                config.pktout.bit.proto_stats_ena = 1;
                ret = odp_pktio_config(pktio_tx, &config);
                CU_ASSERT(ret == 0);
```

test/validation/api/pktio/pktio.c
```diff
@@ -2578,6 +2578,202 @@ static void pktio_test_extra_stats(void)
...
+
+	if (capa.tx.counters.bit.tx_oct_count1) {
+		int64_t counted_bytes = PKT_LEN_NORMAL;
+
+		if (capa.tx.oct_count1_adj)
+			counted_bytes += opt0.oct_count1_adj;
+		CU_ASSERT(data0.tx_oct_count1 == counted_bytes * (uint64_t)flow0_pkts);
+	}
+
+	/* Validate Flow-1 packet statistics */
```
JannePeltonen, line 2741: Please do not repeat the same code twice but put the common code in a function and call it for each flow.

ksatha: Thanks for suggestion, will change in next version

test/validation/api/pktio/pktio.c
ksatha: Thanks for suggestion, will change in next version

test/validation/api/pktio/pktio.c
ksatha: ACK

test/validation/api/pktio/pktio.c
ksatha: ACK

test/validation/api/pktio/pktio.c
ksatha: ACK

JannePeltonen (approved):
Looks ok. My final suggestion is to squash the two linux-gen implementation patches as I do not see any benefit in keeping them separate (they even have the same commit message). The prefix can then be simply "linux-gen:".

PR #1208, opened 2021-03-01
[PATCH v6] example: debug: fix timer pool configuration

PavanNikhilesh:
Currently, the max and min timeout are being configured with the
minimum resolution supported by the timer implementation, instead
retrieve the resolution and min timeout supported for the given
max timeout and configure them accordingly.

### DIFF
--- a/example/debug/odp_debug.c
+++ b/example/debug/odp_debug.c
@@ -346,6 +346,7 @@ static int timer_debug(void)
 	odp_pool_t pool;
 	odp_pool_param_t pool_param;
 	odp_timeout_t timeout;
+	odp_timer_res_capability_t timer_res_capa;
 	odp_timer_capability_t timer_capa;
 	odp_timer_pool_t timer_pool;
 	odp_timer_pool_param_t timer_param;
@@ -382,8 +383,15 @@ static int timer_debug(void)
 	if (timer_capa.max_tmo.max_tmo < max_tmo)
 		max_tmo = timer_capa.max_tmo.max_tmo;
 
-	if (timer_capa.max_tmo.res_ns > res)
-		res = timer_capa.max_tmo.res_ns;
+	memset(&timer_res_capa, 0, sizeof(odp_timer_res_capability_t));
+	timer_res_capa.max_tmo = max_tmo;
+	if (odp_timer_res_capability(ODP_CLOCK_CPU, &timer_res_capa)) {
+		ODPH_ERR("Timer resolution capability failed\n");
+		return -1;
+	}
+
+	if (timer_res_capa.res_ns > res)
+		res = timer_res_capa.res_ns;
 
 	memset(&timer_param, 0, sizeof(timer_param));
 	timer_param.res_ns  = res;


example/debug/odp_debug.c
```diff
@@ -378,16 +379,19 @@ static int timer_debug(void)
...
 	}
+	memset(&timer_res_capa, 0, sizeof(odp_timer_res_capability_t));
+	timer_res_capa.res_ns = res;
+	if (odp_timer_res_capability(ODP_CLOCK_CPU, &timer_res_capa)) {
+		ODPH_ERR("Timer resolution capability failed\n");
+		return -1;
+	}
 
-	if (timer_capa.max_tmo.max_tmo < max_tmo)
-		max_tmo = timer_capa.max_tmo.max_tmo;
```
psavol, line 383: Actually, we don't care about resolution, but maximum tmo being long enough that we can print debug info before timer expires. So, this (max_tmo=1sec or max_tmo=capa.max_tmo) should be there first, and then get resolution and min_tmo with odp_timer_res_capability().

PavanNikhilesh: Ack, I will send the next version.

psavol: Max tmo capa may be less than 1sec, so we need to first check that. After that res_capa will succeed and we'll use res and min tmo from its output.

```
if (timer_capa.max_tmo.max_tmo < max_tmo)
	max_tmo = timer_capa.max_tmo.max_tmo;

memset(&timer_res_capa, 0, sizeof(odp_timer_res_capability_t));
timer_res_capa.max_tmo = max_tmo;
if (odp_timer_res_capability(ODP_CLOCK_CPU, &timer_res_capa)) {
	ODPH_ERR("Timer resolution capability failed\n");
	return -1;
}
```

example/debug/odp_debug.c
```diff
@@ -382,12 +383,19 @@ static int timer_debug(void)
...
-		res = timer_capa.max_tmo.res_ns;
+	memset(&timer_res_capa, 0, sizeof(odp_timer_res_capability_t));
+	timer_res_capa.max_tmo = max_tmo;
+	if (odp_timer_res_capability(ODP_CLOCK_CPU, &timer_res_capa)) {
+		ODPH_ERR("Timer resolution capability failed\n");
+		return -1;
+	}
+
+	if (timer_res_capa.res_ns < res)
+		res = timer_res_capa.res_ns;
```
psavol, line 394: Requested resolution is too high, if it is smaller than capa.res_ns. So, the comparison should be other way around

```
if (timer_res_capa.res_ns > res)
		res = timer_res_capa.res_ns;
```

example/debug/odp_debug.c
```diff
@@ -382,12 +383,19 @@ static int timer_debug(void)
...
+		return -1;
+	}
+
+	if (timer_res_capa.res_ns > res)
+		res = timer_res_capa.res_ns;
 
 	memset(&timer_param, 0, sizeof(timer_param));
 	timer_param.res_ns  = res;
-	timer_param.min_tmo = max_tmo / 10;
+	timer_param.min_tmo = res;
```
psavol, line 398: I think you had earlier timer_res_capa.min_tmo here. Value of res may cause problems here as timer_res_capa.min_tmo value may be larger than timer_res_capa.res_ns. So, it would be safer to use timer_res_capa.min_tmo, or something around max_tmo (but less than max_tmo/2 which we are using for timeouts).

MatiasElo: Ping @PavanNikhilesh

psavol (approved):

PR #1212, opened 2021-03-02
validation: pktio: evaluate bad checksum

gmuthukr:
Add tests for validating bad checksum in udp and sctp packets.

Signed-off-by: Gowrishankar Muthukrishnan <gmuthukrishn@marvell.com>
Change-Id: I875f82c0a454be339ea44e710701199c0f230e5b

### DIFF
--- a/test/validation/api/pktio/pktio.c
+++ b/test/validation/api/pktio/pktio.c
@@ -2767,6 +2767,28 @@ static void pktio_test_chksum_in_udp(void)
 			  pktio_test_chksum_in_udp_test);
 }
 
+static void pktio_test_chksum_bad_in_udp_prep(odp_packet_t pkt)
+{
+	odph_udphdr_t *hdr;
+
+	/* Insert incorrect checksum */
+	pktio_test_chksum_in_udp_prep(pkt);
+	hdr = odp_packet_l4_ptr(pkt, NULL);
+	hdr->chksum ^= 1;
+}
+
+static void pktio_test_chksum_bad_in_udp_test(odp_packet_t pkt)
+{
+	CU_ASSERT(odp_packet_l4_chksum_status(pkt) == ODP_PACKET_CHKSUM_BAD);
+}
+
+static void pktio_test_chksum_bad_in_udp(void)
+{
+	pktio_test_chksum(pktio_test_chksum_in_udp_config,
+			  pktio_test_chksum_bad_in_udp_prep,
+			  pktio_test_chksum_bad_in_udp_test);
+}
+
 static int pktio_check_chksum_in_sctp(void)
 {
 	odp_pktio_t pktio;
@@ -2824,6 +2846,28 @@ static void pktio_test_chksum_in_sctp(void)
 	pktio_test_chksum_sctp(pktio_test_chksum_in_sctp_config,
 			       pktio_test_chksum_in_sctp_prep,
 			       pktio_test_chksum_in_sctp_test);
+}
+
+static void pktio_test_chksum_bad_in_sctp_prep(odp_packet_t pkt)
+{
+	odph_sctphdr_t *hdr;
+
+	/* Insert incorrect checksum */
+	pktio_test_chksum_in_sctp_prep(pkt);
+	hdr = odp_packet_l4_ptr(pkt, NULL);
+	hdr->chksum ^= 1;
+}
+
+static void pktio_test_chksum_bad_in_sctp_test(odp_packet_t pkt)
+{
+	CU_ASSERT(odp_packet_l4_chksum_status(pkt) == ODP_PACKET_CHKSUM_BAD);
+}
+
+static void pktio_test_chksum_bad_in_sctp(void)
+{
+	pktio_test_chksum_sctp(pktio_test_chksum_in_sctp_config,
+			       pktio_test_chksum_bad_in_sctp_prep,
+			       pktio_test_chksum_bad_in_sctp_test);
 }
 
 static int pktio_check_chksum_out_ipv4(void)
@@ -3632,7 +3676,11 @@ odp_testinfo_t pktio_suite_unsegmented[] = {
 				  pktio_check_chksum_in_ipv4),
 	ODP_TEST_INFO_CONDITIONAL(pktio_test_chksum_in_udp,
 				  pktio_check_chksum_in_udp),
+	ODP_TEST_INFO_CONDITIONAL(pktio_test_chksum_bad_in_udp,
+				  pktio_check_chksum_in_udp),
 	ODP_TEST_INFO_CONDITIONAL(pktio_test_chksum_in_sctp,
+				  pktio_check_chksum_in_sctp),
+	ODP_TEST_INFO_CONDITIONAL(pktio_test_chksum_bad_in_sctp,
 				  pktio_check_chksum_in_sctp),
 	ODP_TEST_INFO_CONDITIONAL(pktio_test_chksum_out_ipv4_no_ovr,
 				  pktio_check_chksum_out_ipv4),


gmuthukr:
> Looks good, but would you be willing to do the same also for pktio_test_chksum_in_ipv4?

Could you please check v5 ? Added chksum bad test for ipv4 as well. Thanks.

gmuthukr:
> > Looks good, but would you be willing to do the same also for pktio_test_chksum_in_ipv4?
> 
> Could you please check v5 ? Added chksum bad test for ipv4 as well. Thanks.

@JereLeppanen could you please review v6 as I have also fixed minor offset problem for platform linux_gen (and hence v5 broken).

test/validation/api/pktio/pktio.c
```diff
@@ -2709,6 +2709,36 @@ static void pktio_test_chksum_in_ipv4(void)
...
+
+	/* Insert incorrect checksum */
+	offset = odp_packet_l3_offset(pkt);
+	odp_packet_copy_to_mem(pkt,
+					offset + ODPH_IPV4HDR_CSUM_OFFSET,
+					2, &chksum);
+	chksum ^= 1;
+	odp_packet_copy_from_mem(pkt,
+					offset + ODPH_IPV4HDR_CSUM_OFFSET,
+					2, &chksum);
```
JereLeppanen, line 2727: Indentation seems to be a little off here.

JereLeppanen (changes requested):

platform/linux-generic/odp_packet.c
```diff
@@ -1953,6 +1953,7 @@ static inline uint8_t parse_ipv4(packet_parser_t *prs, const uint8_t **parseptr,
 		if (chksum_finalize(chksum_partial(ipv4, ihl * 4, 0)) != 0xffff) {
 			prs->flags.ip_err = 1;
 			prs->flags.l3_chksum_err = 1;
+			*offset += ihl * 4;
```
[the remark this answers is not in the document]
gmuthukr: Patch v5 which included bad ckcum check for IPv4 header broke pktio validation test in linux_gen platform. Upon inspecting logs, there was failure at test/validation/api/pktio/pktio.c:229 as head.magic != TEST_SEQ_MAGIC. It was due to incorrect L4 offset used to retrieve packet data for this check, which arises from odp_packet.c:2181 where parse_ipv4() did not set offset pointer (for L4) when L3 checksum was incorrect in the received packet and the former pointer value will eventually be stored in packet_parser->l4_offset (and consequently, this will be used to retrieve packet payload following L4).

I agree when checksum fails, ihl could be corrupt, we anyway will have odp_packet_l3_chksum_status() returning ODP_PACKET_CHKSUM_BAD which is actually the test here. So, this patch is to enable the test do its validation part successfully.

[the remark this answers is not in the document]
JereLeppanen: Good explanation, thank you. So of course in case of incorrect L3 checksum, the L4 checks shouldn't be done. I'll comment the other commit. Also, I think in that case the l4_offset should be invalid, but currently it isn't. So instead of this change, the caller (packet_parse_common_l3_l4()) should set l4_offset only if parse_ipv4() returns success.

test/validation/api/pktio/pktio.c
```diff
@@ -2709,6 +2709,36 @@ static void pktio_test_chksum_in_ipv4(void)
...
+}
+
+static void pktio_test_chksum_bad_in_ipv4_test(odp_packet_t pkt)
+{
+	CU_ASSERT(odp_packet_l3_chksum_status(pkt) == ODP_PACKET_CHKSUM_BAD);
+}
+
+static void pktio_test_chksum_bad_in_ipv4(void)
+{
+	pktio_test_chksum(pktio_test_chksum_in_ipv4_config,
```
JereLeppanen, line 2737: So pktio_test_chksum() function can't be used as-is for this test, because it does L4 checks, and these packets have an incorrect L3 checksum. It's a little more work to implement this one. It's up to you if you want to do it - if not, then please just remove this test.

gmuthukr: I think pktio_test_chksum is correct on its way for how it is now testing L3 checksum, if it is EXPECTED that, the implementation platform (linux_gen) should still return packet parser header with L4 offset set correctly and when the L3 proto is found to be a known protocol as well. It is just L3 header that got corrupted and we have L3_chksum_status API to validate this from application point of view. So, is the expectation on linux_gen correct as above ?. If so, we should fix this platform and hence I proposed fix to return right offset. Would you agree ?

JereLeppanen: Even if L4 offset is set in case L3 checksum is bad, it's not clear (to me, at least) what the application or this validation test should or shouldn't do with it. In the general case, if L3 checksum is bad, then L4 offset could be wrong. In this particular test, we know that only the checksum is bad and the rest of ipv4 header is correct. But ODP implementation doesn't know that. Maybe @psavol can chime in here.

psavol: API does not specify in which packet error cases implementation sets L4 offset with a correct value vs. offset_invalid. However, implementation should not set offset with a wrong value. Application should not assume that a packet has no errors, if L4 offset is set. It should use odp_packet_has_error(), etc instead. I.e. a packet may have errors (e.g. bad address) that does not prevent parser detecting where e.g. L4 starts.

gmuthukr: So, an additional assert for no other error  i,e odp_packet_has_error == 0 along with odp_packet_l4_chksum_status check in test side is sufficient I believe. I would not want any other portion of test frame work (like get_packets) to check for error as tests may intentionally test error packets in other places. Sound good ?

gmuthukr: @psavol @JereLeppanen ??.

psavol: A test function for pktio_test_chksum() - such as pktio_test_chksum_bad_in_ipv4_test() could check that:
- chksum_status is ODP_PACKET_CHKSUM_BAD
- odp_packet_has_error() returns !=0
- odp_packet_has_l3_error() returns !=0 (an IP error)

But it should not check e.g. L4 offset, which it does not currently do.

gmuthukr: > So, an additional assert for no other error i,e odp_packet_has_error == 0 along with odp_packet_l4_chksum_status check in test side is sufficient I believe. I would not want any other portion of test frame work (like get_packets) to check for error as tests may intentionally test error packets in other places. Sound good ?

I meant odp_packet_l3_chksum_status here. Apologies.

gmuthukr: > A test function for pktio_test_chksum() - such as pktio_test_chksum_bad_in_ipv4_test() could check that:
> 
> * chksum_status is ODP_PACKET_CHKSUM_BAD
> * odp_packet_has_error() returns !=0
> * odp_packet_has_l3_error() returns !=0 (an IP error)
> 
> But it should not check e.g. L4 offset, which it does not currently do.

Right. I did not mean l4_chksum_status in ipv4 test and corrected my previous reply. Now, I am sending newer patch with above additional checks in test validation part. Could you please check. Thanks.

gmuthukr: @psavol I have added error and [l3/l4]_error checks in all checksum tests except in sctp, where l4_error() always == 0 meaning even if bad checksum is injected, this api never reported error. Up on checking, I think packet_l4_chksum() would have not set p.flags.l4_chksum_err correctly. I did not want to break CI for this known failure if I am correct and hence, skipped additional tests for sctp (bad checksum). Once I find root cause for it, i can send separate patch. Or we can tag it known failure (CU_PASS ??) in pktio_test_chksum_bad_in_sctp_test(), after adding l4_error check. What would you suggest ?. Thanks.

test/validation/api/pktio/pktio.c
psavol: Yes, indention should be aligned with parenthesis. Actually, this checksum modification could be done the same way as is done in UDP/STCP prep functions (without copies).

hdr = odp_packet_l3_ptr(pkt, NULL);
hdr->chksum ^= 1;

Copy is needed is packet is segmented in the middle of the field. I guess this test application makes sure, or can be sure that all headers fit into the first segment of the packet.

test/validation/api/pktio/pktio.c
psavol: Looks good to me, except that pktio_test_chksum_bad_in_ipv4_prep() could be done the same way as pktio_test_chksum_bad_in_udp_prep().

STCP fix and test case improvement can be done in another PR.

platform/linux-generic/odp_packet.c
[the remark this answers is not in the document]
psavol: When L3 checksum failed, implementation cannot be sure about correct L4 offset. So, it's better to set it to ODP_PACKET_OFFSET_INVALID. API does not guarantee to application that, but it's safer to set it invalid in a software parser while it does not harm performance, etc. With HW parsers the offset may not be set to invalid without performance cost, so API leaves it open.

test/validation/api/pktio/pktio.c
gmuthukr: Thanks @psavol and @JereLeppanen . Sent v8 addressing your valuable comments. Please check once.

JereLeppanen: I think L4 offset is still being checked in the call chain pktio_test_chksum -> wait_for_packets -> wait_for_packets_hdr -> pktio_pkt_seq_hdr -> odp_packet_l4_offset. This does not cause a test failure because the offset is currently being set despite of incorrect L3 checksum. But as Petri stated, L4 checksum should not be checked in this case.

Speaking of which, since it's not required to set L4 offset when L3 checksum is bad, please revise the first commit in this PR accordingly, or simply drop the commit.

platform/linux-generic/odp_packet.c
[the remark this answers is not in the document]
psavol: Set to invalid offset, instead of leaving it to L3 offset or incrementing to L4 offset.
`*offset =ODP_PACKET_OFFSET_INVALID;`

test/validation/api/pktio/pktio.c
psavol: True. wait_for_packets_hdr()/pktio_pkt_seq_hdr() needs to skip sequence number check when bad checksum is tested and there is a checksum error. The tricky part is that pktio_pkt_seq_hdr() is there to filter out packets originated from outside of the test case (e.g. on some pktios Linux IP stack sends periodic multicasts which this function filters out with sequence number checking).

Maybe some field in IP header could be used for filtering (in this bad checksum case), instead of magic number in UDP payload.

platform/linux-generic/odp_packet.c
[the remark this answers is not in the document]
gmuthukr: @psavol Apologies for a quite longer delay in response. I have followed your and @JereLeppanen suggestions and sent v9 patch. Please check if the commits are good. Appreciating both of your time.

test/validation/api/pktio/pktio.c
gmuthukr: Ack

test/validation/api/pktio/pktio.c
gmuthukr: I have modified the approach slightly to check return code (bad checksum can be ignored as the test injects them). Also, assumption is that, bad checksum can not be anything else bad than test is setting. Custom packet field may inject any side effect in case hw parser manipulates it (just a thought). Please see if this is fine.

gmuthukr:
@JereLeppanen @psavol could you please check if the change is fine, as I am not sure if you get notified unless "changes requested" on me were not cleared (as they got outdated). =

test/validation/api/pktio/pktio.c
```diff
@@ -23,7 +23,8 @@
 
 #define USE_MTU                0
 #define MAX_NUM_IFACES         2
-#define TEST_SEQ_INVALID       ((uint32_t)~0)
+#define TEST_SEQ_IGNORE        ODP_PACKET_OFFSET_INVALID
+#define TEST_SEQ_INVALID       ((uint32_t)~1)
```
JereLeppanen, line 27: Changing this value doesn't actually do anything, or does it?

JereLeppanen (commented):
Please remove the merge commit and rebase.

test/validation/api/pktio/pktio.c
```diff
@@ -740,6 +741,11 @@ static int wait_for_packets_hdr(pktio_info_t *pktio_rx, odp_packet_t pkt_tbl[],
 			continue;
 
 		for (i = 0; i < n; ++i) {
+			if (seq_tbl[i] == TEST_SEQ_IGNORE) {
+				pkt_tbl[num_rx++] = pkt_tmp[i];
+				continue;
+			}
```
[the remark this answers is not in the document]
gmuthukr: I think, non-test packet may appear for any other test as well. For that case, harness itself need changes to get rid of seq table way of remembering packets. I agree on your point for overall effectiveness, but the objective of this patch is only to improve current testing based on existing methods. IMHO we can address the incompetence in seq_table logic through totally new way for all tests including checksum.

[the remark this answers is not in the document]
JereLeppanen: This is probably the only validation test where we use pktios on interfaces that are visible to Linux, which is why we may receive some non-test packets during the test. That's why this filtering mechanism is needed specifically in this test.

After a short discussion with the team, I think we should probably just check for the magic numbers in the packets even in the bad checksum tests, to filter out non-test packets. So basically check the magic numbers in the payload first and drop the packets that don't match, then proceed with the other checks. How does that sound?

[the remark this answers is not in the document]
gmuthukr: Ack @JereLeppanen . I have updated the patch to check magic number too. But the strategy is to invalidate only packets after the first packet in test suite and compare seq number derived from invalid packet with first one relatively.
This approach is only for IP bad checksum test.

Reason is that, as the implementation sets l4 offset invalid as a consequence of invalid IP checksum, odp_packet_l4_offset() can not help us derive L4 payload (magic). So, first packet is good packet and used as remembrance for relative comparison in terms of seq number with following packets. This way, all aspects suggested are covered.

Please let me know if it sounds good.

gmuthukr:
@JereLeppanen could this version be merged if there are no more changes needed. Thanks.

test/validation/api/pktio/pktio.c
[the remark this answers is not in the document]
JereLeppanen: Sorry about the slow replies. The latest version looks better. But it looks like TEST_SEQ_RELATIVE doesn't really work. Instead, pktio_pkt_seq_hdr() should be called with the nonzero l4_offset if we are running a bad checksum test. So this should really be a parameter to wait_for_packets_hdr(), passed down from the test case.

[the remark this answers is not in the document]
gmuthukr: Hi @JereLeppanen,
pktio_pkt_seq_hdr() is already called with nonzero offset at line 747, for TEST_SEQ_RELATIVE packets (i.e bad IPv4 chksum packet).

  l4_offset = ODPH_ETHHDR_LEN + ODPH_IPV4HDR_LEN;

Also to note, only pktio_test_chksum_bad_in_ipv4_prep() sets TEST_SEQ_RELATIVE and hence, bad IPv4 chksum packets are correctly tested by above offset.

I have tested and did not find any failure. Could you please let me know if you see tests break ?
Thanks for your time.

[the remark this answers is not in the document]
JereLeppanen: I haven't tried to run this, but looks like there are at least two problems in wait_for_packets_hdr(). One is that in the TEST_SEQ_RELATIVE case, even if pktio_pkt_seq_hdr() returns TEST_SEQ_INVALID, the code still expects the next sequence number to be seq_tbl[0] + i. This should perhaps be seq_tbl[0] + num_rx. The other problem in that same case is that the packet is not freed.

But all that is beside the point, which is that none of this is needed. TEST_SEQ_RELATIVE is set for all or none in seq_tbl[], so an array is really not needed. As far as I can tell, all that is needed in wait_for_packets_hdr() is function parameter to indicate which l4_offset to use.

PR #1218, opened 2021-03-05
[PATCH v11] linux-gen: queue_lf: move aarch64 specific code to arch files

Malvika-Gupta:
odp_queue_lf.c and odp_bitset.h contain aarch64 specific code which causes the build to fail on arm4 systems when the --disable-host-optimization flag is enabled. These patches separate out the default and aarch64 specific code to the arch-specific files and resolve the build issues.

### DIFF
--- a/platform/linux-generic/arch/aarch64/odp_atomic.h
+++ b/platform/linux-generic/arch/aarch64/odp_atomic.h
@@ -217,4 +217,28 @@ static inline __int128 __lockfree_load_16(__int128 *var, int mo)
 	return old;
 }
 
+typedef unsigned __int128 _u128_t;
+
+static inline _u128_t lockfree_load_u128(_u128_t *atomic)
+{
+	return __lockfree_load_16((__int128 *)atomic, __ATOMIC_RELAXED);
+}
+
+static inline int lockfree_cas_acq_rel_u128(_u128_t *atomic,
+					    _u128_t old_val,
+					    _u128_t new_val)
+{
+	return __lockfree_compare_exchange_16((__int128 *)atomic,
+					      (__int128 *)&old_val,
+					      new_val,
+					      0,
+					      __ATOMIC_ACQ_REL,
+					      __ATOMIC_RELAXED);
+}
+
+static inline int lockfree_check_u128(void)
+{
+	return 1;
+}
+
 #endif  /* PLATFORM_LINUXGENERIC_ARCH_ARM_ODP_ATOMIC_H */
--- a/platform/linux-generic/arch/default/odp_atomic.h
+++ b/platform/linux-generic/arch/default/odp_atomic.h
@@ -1,0 +1,36 @@
+/* Copyright (c) 2021, Arm Limited
+ * All rights reserved.
+ *
+ * SPDX-License-Identifier:     BSD-3-Clause
+ */
+
+#ifndef ODP_DEFAULT_ATOMIC_H_
+#define ODP_DEFAULT_ATOMIC_H_
+
+#ifdef __SIZEOF_INT128__
+
+typedef unsigned __int128 _u128_t;
+
+static inline _u128_t lockfree_load_u128(_u128_t *atomic)
+{
+	return __atomic_load_n(atomic, __ATOMIC_RELAXED);
+}
+
+static inline int lockfree_cas_acq_rel_u128(_u128_t *atomic,
+					    _u128_t old_val,
+					    _u128_t new_val)
+{
+	return __atomic_compare_exchange_n(atomic, &old_val, new_val,
+					   0 /* strong */,
+					   __ATOMIC_ACQ_REL,
+					   __ATOMIC_RELAXED);
+}
+
+static inline int lockfree_check_u128(void)
+{
+	return __atomic_is_lock_free(16, NULL);
+}
+
+#endif
+
+#endif
--- a/platform/linux-generic/odp_queue_lf.c
+++ b/platform/linux-generic/odp_queue_lf.c
@@ -23,61 +23,12 @@
 
 typedef unsigned __int128 u128_t;
 
-static inline void atomic_zero_u128(u128_t *atomic)
+static inline void lockfree_zero_u128(u128_t *atomic)
 {
 	__atomic_store_n(atomic, 0, __ATOMIC_RELAXED);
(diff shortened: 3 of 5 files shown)


platform/linux-generic/odp_queue_lf.c
```diff
@@ -28,57 +28,8 @@ static inline void atomic_zero_u128(u128_t *atomic)
...
 }
 
-#if defined(__aarch64__)
-/* ARMv8 has atomic load-acq/store-rel instructions for a pair of
- * 64bit of data. GCC atomic built-in for 128bits does not utilize these
- * instructions but uses locks instead. Override GCC built-in for ARMv8.
- */
 #include <odp_cpu.h>
 
-static inline int atomic_cas_acq_rel_u128(u128_t *atomic, u128_t old_val,
```
matiaselo, line 38: Instead of duplicating functionality, couldn't the new 128-bit atomic API functions be used here?

Malvika-Gupta: I will look into it. Thanks!

Malvika-Gupta:
Hi @psavol and @matiaselo, as we discussed on Monday, I have separated out the commits into 2 PRs. Some code such as that in the `platform/linux-generic/Makefile.am` and `arch/default/odp_cpu.h` is duplicated in both. 
Also, I looked into replacing the existing code with 128-bit atomic functions. But unlike the other u32 and u64 bit atomic functions, they are not defined in `/odp/ap/abi/plat/atomic_inlines.h` but in platform/linux-generic/odp_atomic.c. So I am not sure how I will be able to use them unless those 128-bit APIs are moved first. 
I look forward to your comments. Thank you!

matiaselo:
> Also, I looked into replacing the existing code with 128-bit atomic functions. But unlike the other u32 and u64 bit atomic functions, they are not defined in `/odp/api/plat/atomic_inlines.h` but in `platform/linux-generic/odp_atomic.c`. So I am not sure how I will be able to use them unless those 128-bit APIs are first moved.

Hi @Malvika-Gupta,

There shouldn't be anything blocking you to use the API functions. This is already done all over the implementation. You just have to include the API header (e.g. `#include <odp/api/atomic.h>`).

platform/linux-generic/odp_queue_lf.c
```diff
@@ -355,11 +268,14 @@ void _odp_queue_lf_term_global(void)
 static void init_queue(queue_lf_t *queue_lf)
 {
 	int i;
+	odp_u128_t u128_t;
```
matiaselo, line 271: `_t` postfix is usually used for data types, so this variable naming is a bit confusing.

platform/linux-generic/arch/aarch64/odp_atomic.h
```diff
@@ -267,4 +267,9 @@ static inline __int128 __lockfree_load_16(__int128 *var, int mo)
 
 #endif
 
+static inline int atomic_is_lockfree_u128(void)
```
matiaselo, line 270: Couldn't you use `odp_atomic_lock_free_u128()` instead? You just have to check the individual bits: load, cas

platform/linux-generic/odp_queue_lf.c
```diff
@@ -21,97 +21,10 @@
 
 #ifdef __SIZEOF_INT128__
 
-typedef unsigned __int128 u128_t;
```
matiaselo, line 24: Copyright message should be updated.

Malvika-Gupta: Will do. Thanks!

platform/linux-generic/arch/aarch64/odp_atomic.h
Malvika-Gupta: Okay, I will use that and remove this API. Thanks!

platform/linux-generic/odp_queue_lf.c
Malvika-Gupta: I will change to a more suitable name. Thanks!

Malvika-Gupta:
Hi @matiaselo, I have made the above changes. Please review. Thanks!

matiaselo (changes requested):
Hi @Malvika-Gupta,

Unfortunately I just figured out there is a blocker for this approach. The lock-free queues have to be supported also on ARMv8 platforms and with the current 128-bit atomics implementation this is not possible. So I would suggest going back to the original approach in V1 and archiving this version for later use. I'll start working on the atomics code reorganization, so we can add 128-bit atomics support for additional platforms. Sorry for the extra work.

Malvika-Gupta:
Hi @matiaselo I have reverted the patch set to V1. Please review. Thank you!

platform/linux-generic/arch/default/odp_atomic.h
```diff
@@ -0,0 +1,36 @@
...
+	return __atomic_compare_exchange_n(atomic, &old_val, new_val,
+					   0 /* strong */,
+					   __ATOMIC_ACQ_REL,
+					   __ATOMIC_RELAXED);
+}
+
+static inline int atomic_is_lockfree_u128(void)
+{
+	return __atomic_is_lock_free(16, NULL);
+}
```
matiaselo, line 32: These generic variants can now be implemented using standard  ODP API 128-bit atomics.

matiaselo: On a second thought, this is simpler as the types don't have to be changed. Looks like the PR only needs a rebase and it's ready to be merged.

psavol: I'd suggest to rename these functions while moving into header files. Otherwise, we loose a bit the context why queue_lf.c uses these functions and not the new 128bit API. Functions could be renamed like this:
atomic_load_u128 -> lockfree_load_u128
atomic_cas_acq_rel_u128 -> lockfree_cas_acq_rel_u128
atomic_is_lockfree_u128 -> lockfree_check_u128

Malvika-Gupta:
Hi @psavol and @matiaselo, I have renamed the APIs in the header files and consequently had to replicate that in odp_queue_lf.c file for the locked version of the APIs. Please review. Thanks!

matiaselo (approved):

PR #1221, opened 2021-03-09
api: classification: print debug info for all classifier rules

JereLeppanen:
Add odp_cls_print_all(), which prints implementation specific debug information for all classifier rules.

Example print:
```
Classifier info
---------------

DefaultCos(0)
    queue: DefaultQueue
    rules: PMR_SIP_ADDR: 0a0a0a00 ffffff00 -> CoSqueue1(1)
           PMR_SIP_ADDR: c0a80000 ffff0000 -> CoSqueue2(2)
           PMR_CUSTOM_FRAME: offset:48 0123456789abcdef ffffffffffffffff -> mark:7 CoSqueue2(3)
CoSqueue1(1)
    queue: queue1
CoSqueue2(2)
    queue: queue2
CoSqueue2(3)
    queue: queue2
```


platform/linux-generic/odp_classification.c
```diff
@@ -1782,3 +1782,95 @@ uint64_t odp_pmr_to_u64(odp_pmr_t hdl)
...
+
+			print_cos_ident(&cos->linked_cos[j]->s);
+		}
+		UNLOCK(&pmr->lock);
+	}
+}
+
+void odp_cls_print_all(void)
+{
+	ODP_PRINT("Classifier info\n"
```
psavol, line 1865: Info prints usually start with a new line. That way those are easier to spot in the middle of a log, especially if the previous print didn't end with a new line.

platform/linux-generic/odp_classification.c
```diff
@@ -1782,3 +1782,95 @@ uint64_t odp_pmr_to_u64(odp_pmr_t hdl)
...
+
+static
+void print_queue_ident(odp_queue_t q)
+{
+	odp_queue_info_t info;
+
+	if (odp_queue_info(q, &info))
+		ODP_PRINT("%" PRIx64, odp_queue_to_u64(q));
+	else
+		ODP_PRINT("%s", info.name);
```
psavol, line 1800: Name is optional for both CoS and queue. In both cases, handle pointer could be printed when the name is missing.

platform/linux-generic/odp_classification.c
```diff
@@ -1782,3 +1782,95 @@ uint64_t odp_pmr_to_u64(odp_pmr_t hdl)
...
+
+static
+void cls_print_cos(struct cos_s *cos)
+{
+	print_cos_ident(cos);
+	ODP_PRINT("    queue: ");
+	print_queue_ident(cos->queue);
+
+	uint32_t num_rule = odp_atomic_load_u32(&cos->num_rule);
+	bool first = true;
```
psavol, line 1822: Usually, variables should be introduced in the beginning of a code block. E.g. a variable size array may be an exception to that (when size  is not know in the beginning of the block). Checkpatch does not seem to complain any more about this, but at least to me code is more readable when variables are introduced only on certain places (code block starts).

platform/linux-generic/odp_classification.c
```diff
@@ -1782,3 +1782,95 @@ uint64_t odp_pmr_to_u64(odp_pmr_t hdl)
...
+	ODP_PRINT("\n");
+}
+
+static
+void print_hex(const void *vp, int len)
+{
+	const uint8_t *p = vp;
+
+	for (int i = 0; i < len; i++)
+		ODP_PRINT("%02x", (int)*p++ & 0xff);
```
psavol, line 1811: Bitwise operations with signed values likely produce new static code checker warnings. Is cast to int needed? May be cast to uint32_t instead of int, and do bitwise operation with 0xffu.

platform/linux-generic/odp_classification.c
JereLeppanen: For cos, handle to u64 is just index+1, so it looks a little goofy:
```
1(0)
    queue: 7f7df0d98180
    rules: PMR_SIP_ADDR: 0a0a0a00 ffffff00 -> 2(1)
2(1)
    queue: 7f7df0d98300
```

platform/linux-generic/odp_classification.c
JereLeppanen: I'll move these.

For the record, I disagree with this rule. Checkpatch doesn't require this, but it does require a blank line after declarations, which anyway makes it impossible to write naturally flowing code.

platform/linux-generic/odp_classification.c
psavol: _odp_cos_entry_from_idx() returns CoS pointer, if pointer is preferred over handle.

However, maybe CoS index and (optional) name are enough to identify CoS line in the table. CoS line would be either e.g. "cos: (index)" or "cos: name (index)". I think it's easy to follow "(index)" tags in the table. CoS names don't necessarily mention "cos".

```
Classifier info
---------------
cos: DefaultCos(0)
    queue: DefaultQueue
    rules: PMR_SIP_ADDR: 0a0a0a00 ffffff00 -> (1)
           PMR_SIP_ADDR: c0a80000 ffff0000 -> (2)
           PMR_CUSTOM_FRAME: offset:48 0123456789abcdef ffffffffffffffff -> mark:7 CoSqueue2(3)
cos: (1)
    queue: queue1
cos: (2)
    queue: queue2
cos: CoSqueue2(3)
    queue: queue2
   
```

JereLeppanen: Done. I also changed the number in paranthesis from internal cos_s.index to odp_cos_to_u64(). Please check.

psavol (approved):

jerinjacobk (approved):

PR #1224, opened 2021-03-11
[PATCH v2] configure: allow user to override *FLAGS

JereLeppanen:
```
Place user provided options at the end of CFLAGS, CXXFLAGS and
LDFLAGS, so that they override other options.

This way we also don't have to refrain from using the -march option
just because user has provided some CFLAGS.

```
v2:
- Add reviewed-by tag.
- Rebase.

### DIFF
--- a/configure.ac
+++ b/configure.ac
@@ -240,11 +240,10 @@ AC_ARG_ENABLE([abi-compat],
 	ODP_LIBSO_VERSION=0:0:0
 
 	# do not use -march=native with clang (due to possible failures on
-	# clang optimizations) or when user have defined CFLAGS (may contain
-	# another -march option).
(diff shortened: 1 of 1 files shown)


JereLeppanen:
With regard to a discussion with @psavol and others: When a single-value option is given multiple times on gcc command line, the last one takes effect. This behavior is undocumented, but I think it's necessary and widely depended upon behavior. For example, automake documentation explains how CFLAGS overrides other *_CFLAGS variables by being placed last on the compiler command line.

psavol (approved):
Tested that e.g. -O3 -O0 produces worse performance that -O3 alone. Also compiler allowed to specify two different -march options.

PR #1227, opened 2021-03-17
[PATCH v2] 128-bit atomics implementation optimization

matiaselo:
V2:
- Fixed `odp_atomic_lock_free_u128()` capabilities (Petri)


platform/linux-generic/arch/aarch64/odp_atomic.c
```diff
@@ -37,11 +37,18 @@ int odp_atomic_lock_free_u128(odp_atomic_op_t *atomic_op)
 	}
 
 	return 2;
-#else
+#elif defined(__SIZEOF_INT128__)
+	if (__atomic_is_lock_free(16, NULL)) {
+		if (atomic_op) {
+			atomic_op->all_bits = ~((uint32_t)0);
+			atomic_op->op.init  = 0;
```
psavol, line 44: Since API has only three 128bit operations (+ init), it's better to set only those three here.
atomic_op->all_bits = 0;
atomic_op->op.load  = 1;
atomic_op->op.store = 1;
atomic_op->op.cas  = 1;

platform/linux-generic/arch/default/odp_atomic.c
```diff
@@ -28,6 +28,15 @@ int odp_atomic_lock_free_u64(odp_atomic_op_t *atomic_op)
 
 int odp_atomic_lock_free_u128(odp_atomic_op_t *atomic_op)
 {
+#ifdef __SIZEOF_INT128__
+	if (__atomic_is_lock_free(16, NULL)) {
+		if (atomic_op) {
+			atomic_op->all_bits = ~((uint32_t)0);
+			atomic_op->op.init  = 0;
```
psavol, line 35: Same here, only three 128b operations.

psavol (approved):
Two comments about lock-free capa flags. Otherwise OK.

PR #1232, opened 2021-03-25
[PATCH v4] linux-gen: hash: use x86 crc32 instructions

JereLeppanen:
```
linux-gen: hash: use x86 crc32 instructions
    
    Use x86 SSE 4.2 crc32 instructions via gcc builtin functions, if
    available. These instructions support crc32c only.
    
test: perf: implement crc throughput test
    
    This test measures megabytes per second throughput of the
    odp_hash_crc32() and odp_hash_crc32c() API functions.
    
```

### DIFF
--- a/test/performance/odp_crc.c
+++ b/test/performance/odp_crc.c
@@ -1,0 +1,300 @@
+/* Copyright (c) 2021, Nokia
+ *
+ * All rights reserved.
+ *
+ * SPDX-License-Identifier:     BSD-3-Clause
+ */
+
+#include <stdio.h>
+#include <string.h>
+#include <stdint.h>
+#include <inttypes.h>
+#include <getopt.h>
+
+#include <odp_api.h>
+#include <odp/helper/odph_api.h>
+
+#define KB 1024ull
+#define MB (1024ull * 1024ull)
+
+/* Command line options */
+typedef struct {
+	uint32_t size;
+	uint32_t rounds;
+	uint32_t offset;
+	uint32_t test;
+} options_t;
+
+static options_t options;
+static const options_t options_def = {
+	.size = 16,
+	.rounds = 10000,
+	.offset = 0,
+	.test = 0,
+};
+
+static void print_usage(void)
+{
+	printf("\n"
+	       "CRC performance test\n"
+	       "\n"
+	       "Usage: odp_crc_perf [options]\n"
+	       "\n"
+	       "  -s, --size    Size of buffer in KB (default %u)\n"
+	       "  -r, --rounds  Number of test rounds (default %u)\n"
+	       "                Rounded down to nearest multiple of 8\n"
+	       "  -o, --offset  Offset of data (default %u)\n"
+	       "  -t, --test    Which API to test (default %u)\n"
+	       "                0: both\n"
+	       "                1: odp_hash_crc32c\n"
+	       "                2: odp_hash_crc32\n"
+	       "  -h, --help    This help\n"
+	       "\n",
+	       options_def.size, options_def.rounds, options_def.offset,
+	       options.test);
+}
+
+static int parse_options(int argc, char *argv[])
+{
+	int opt;
+	int long_index;
+	int ret = 0;
+
+	static const struct option longopts[] = {
+		{ "size", required_argument, NULL, 's' },
+		{ "rounds", required_argument, NULL, 'r' },
+		{ "offset", required_argument, NULL, 'o' },
+		{ "test", required_argument, NULL, 't' },
+		{ "help", no_argument, NULL, 'h' },
+		{ NULL, 0, NULL, 0 }
+	};
+
+	static const char *shortopts = "+s:r:o:t:h";
+
+	options = options_def;
+
+	while (1) {
+		opt = getopt_long(argc, argv, shortopts, longopts, &long_index);
+
+		if (opt == -1)
+			break;
+
+		switch (opt) {
+		case 's':
+			options.size = atol(optarg);
+			break;
+		case 'r':
+			options.rounds = atol(optarg);
+			break;
+		case 'o':
+			options.offset = atol(optarg);
+			break;
+		case 't':
+			options.test = atol(optarg);
+			break;
+		case 'h':
+			/* fall through */
+		default:
+			print_usage();
+			ret = -1;
+			break;
+		}
+	}
+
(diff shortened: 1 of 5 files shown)


test/performance/odp_crc.c
```diff
@@ -0,0 +1,238 @@
...
+}
+
+typedef uint32_t (*test_fn_t)(const void *data, uint32_t data_len,
+			      uint32_t init_val);
+
+static uint64_t measure(test_fn_t fn, uint32_t size, uint32_t offset,
+			uint32_t num)
+{
+	void *p = data + offset;
+	volatile uint32_t crc = 1;
```
psavol, line 113: Volatile cannot be used here, since it degrades performance. It tells compiler that something outside of this function may also read/write the variable.

Instead of volatile, output the crc value in the end of the function and e.g. print it out with measurement results. This way compiler cannot optimize crc calculation away, but can perform other optimizations. When I removed volatile, result of the 32B crc32c test case improved 25%.

static uint64_t measure(test_fn_t fn, uint32_t size, uint32_t offset, uint32_t num, uint32_t *crc_out)

test/performance/odp_crc.c
```diff
@@ -0,0 +1,238 @@
...
+	odp_time_t start = odp_time_local();
+
+	for (uint32_t i = 0; i < num; i++) {
+		crc ^= fn(p, size, crc);
```
psavol, line 117: Since this measures absolute performance (and only two functions under test.), I'd have two separate measurement functions and not use function pointers (as those may prevent code inlining).

Quick test with direct crc32c calls seems to give also a small performance improvement, but not as clear as removal of the volatile.

test/performance/odp_crc.c
```diff
@@ -0,0 +1,238 @@
...
+		nsec = measure(fn, size, offset, num);
+
+		if (nsec / 1000000 > options.time / 2)
+			break;
+
+		num *= 2;
+	}
+
+	num = (double)num * (double)options.time / 1000.0 /
+	      ((double)nsec / (double)ODP_TIME_SEC_IN_NS);
```
psavol, line 148: This time measurement makes test runs not repeatable. The same parameters may result different number of rounds (and data size).

Instead of time, user could pass test length in MB. That would tell number of rounds for each data size, and have about the same execution time on each case. The default length could be pretty small (e.g. 1MB -> 32x32kB), so that user (or CI) don't think that test hang when CRC implementation is slow (default SW).

test/performance/odp_crc.c
```diff
@@ -0,0 +1,238 @@
...
+static const int bufsize = 32 * 1024;
+
+static void prepare_data(void)
+{
+	uint64_t seed = 1;
+	const unsigned long page = 4096;
+
+	/* One extra page for alignment, another one for offset. */
+	buf = (uint8_t *)malloc(bufsize + page * 2);
+	data = (uint8_t *)(((uintptr_t)buf + (page - 1)) & ~(page - 1));
```
psavol, line 93: ODPH_ERR() and exit if malloc() fails.

Usage of SHM would be better than malloc, if bufsize would be increased later on (SHM uses likely hugepages for large allocs).

test/performance/odp_crc.c
```diff
@@ -0,0 +1,238 @@
...
+	if (options.time < 1) {
+		ODPH_ERR("Invalid time: %" PRIu32 "\n", options.time);
+		return -1;
+	}
+
+	return ret;
+}
+
+static uint8_t *buf, *data;
+static const int bufsize = 32 * 1024;
```
matiaselo, line 84: `buf` and `data` could be normal function arguments instead of global variables.

I'd also prefer `bufsize` being either a define or an application option.

test/performance/odp_crc.c
JereLeppanen: Fixed in v2 (without the print).

test/performance/odp_crc.c
JereLeppanen: Done in v2, although the inlining prevention was intentional. With inlining, test results may change for example due to inlining criteria changing in a new compiler version.

test/performance/odp_crc.c
JereLeppanen: Done in v2, although this means that the test doesn't produce any useful results with the default parameters. And also that for example on x86 the crc32 test takes over 15 times longer than the crc32c test, which seems completely impractical. Should I do something about that?

test/performance/odp_crc.c
JereLeppanen: > ODPH_ERR() and exit if malloc() fails.

Done in v2.

I vote that we switch to shm later if needed. Either way, above 32K buffer size this increasingly becomes a memory read speed test instead of crc speed test.

test/performance/odp_crc.c
JereLeppanen: In v2 we no longer test multiple buffer sizes, but this is simply a command line parameter, and the default is 32K. I hope this is what you meant.

test/performance/odp_crc.c
psavol: The default size (-s) could be e.g. 16kB, so that it surely fits into L1 cache. With that -m could be 32MB, which results about 0.1sec in my system with the SW based CRC. Which CRC to test could be also an option, so that user can pick values based on the algorithm under test. Default selection could be both (so that CI runs both).

Now, when there's single size in test, -m option could be simply number of rounds (-r). It's may be more intuitive what is tested: e.g. 10000x16kB.

test/performance/odp_crc.c
psavol: It's OK to use malloc for now.

test/performance/odp_crc.c
JereLeppanen: Done in v3. The default is now 10000x16KB.

test/performance/odp_crc.c
```diff
@@ -0,0 +1,297 @@
...
+	uint32_t offset;
+	uint32_t test;
+} options_t;
+
+static options_t options;
+static const options_t options_def = {
+	.size = 16,
+	.rounds = 10000,
+	.offset = 0,
+};
```
psavol, line 33: Default value of "test" could be explicitly set here also (.test = 0).

psavol (approved):
One minor comment. Add my reviewed-by on the next version.

PR #1233, opened 2021-03-27
[PATCH v2] IPsec validation test cleanup

JannePeltonen:
This commit has several patches that clean up the IPsec validation test code and also fix a couple of issues in there. There is also one small change in linux-gen, made possible by a fix in the validation code.

The main change is to decouple the combined output+input tests from the lower level generic test code that will now invoke and check one IPsec packet operation at a time. Combining several operations in one test is now done at the test case level.

@anoobj could you perhaps review these changes and consider rebasing your reassembly validation code on top of this if things look ok?

### DIFF
--- a/test/validation/api/ipsec/ipsec.c
+++ b/test/validation/api/ipsec/ipsec.c
@@ -638,7 +638,7 @@ static int ipsec_send_out_one(const ipsec_test_part *part,
 	} else {
 		struct odp_ipsec_out_inline_param_t inline_param;
 		uint32_t hdr_len;
-		uint8_t hdr[32];
+		odph_ethhdr_t hdr;
 		odp_queue_t queue = ODP_QUEUE_INVALID;
 
 		if (NULL != part->out[0].pkt_res) {
@@ -649,16 +649,16 @@ static int ipsec_send_out_one(const ipsec_test_part *part,
 			 */
 			hdr_len = part->out[0].pkt_res->l3_offset;
 			CU_ASSERT_FATAL(hdr_len <= sizeof(hdr));
-			memcpy(hdr, part->out[0].pkt_res->data, hdr_len);
-		} else if (part->pkt_in->l3_offset !=
-			   ODP_PACKET_OFFSET_INVALID) {
-			hdr_len = part->pkt_in->l3_offset;
-			CU_ASSERT_FATAL(hdr_len <= sizeof(hdr));
-			memcpy(hdr, part->pkt_in->data, hdr_len);
+			memcpy(&hdr, part->out[0].pkt_res->data, hdr_len);
 		} else {
-			/* Dummy header */
 			hdr_len = 14;
-			memset(hdr, 0xff, hdr_len);
+			memset(&hdr, 0xff, hdr_len);
+
+			if (part->out[0].l3_type == ODP_PROTO_L3_TYPE_IPV6) {
+				hdr.type = odp_cpu_to_be_16(ODPH_ETHTYPE_IPV6);
+			} else {
+				hdr.type = odp_cpu_to_be_16(ODPH_ETHTYPE_IPV4);
+			}
 		}
 
 		if (part->flags.inline_hdr_in_packet) {
@@ -681,14 +681,14 @@ static int ipsec_send_out_one(const ipsec_test_part *part,
 			CU_ASSERT_FATAL(ret >= 0);
 			odp_packet_l2_offset_set(pkt, new_l2_offset);
 			odp_packet_l3_offset_set(pkt, new_l3_offset);
-			odp_packet_copy_from_mem(pkt, new_l2_offset, hdr_len, hdr);
+			odp_packet_copy_from_mem(pkt, new_l2_offset, hdr_len, &hdr);
 			if (l4_offset != ODP_PACKET_OFFSET_INVALID)
 				odp_packet_l4_offset_set(pkt, new_l3_offset +
 							 l4_offset - l3_offset);
 
 			inline_param.outer_hdr.ptr = NULL;
 		} else {
-			inline_param.outer_hdr.ptr = hdr;
+			inline_param.outer_hdr.ptr = (void *)&hdr;
 		}
 
 		inline_param.pktio = suite_context.pktio;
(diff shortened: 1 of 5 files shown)


test/validation/api/ipsec/ipsec.h
```diff
@@ -82,12 +86,6 @@ typedef struct {
 		odp_proto_l4_type_t l4_type;
 		uint32_t seq_num;
 	} out[1];
```
anoobj, line 81: While we are at it, do you think this out[1] should be continued? If the test_part is supposed to indicate one packet and the details associated with it, num_pkt and this [1] looks redundant to me.

With reassembly spec change, I did submit a change in increasing the array size, but now that I redid some of the code, I'm starting to think that these logic can stay in the caller and the test part can be independent. 

I would assume the following with reassembly,

Field, pkt_in can be NULL. In that case no packet to be send. Intention of test_part would be to hold the resultant output packet.
Field ipsec_test_part.out[1].pkt_res can be NULL, then no packet to be received. Intention of test_part to just transmit out the packet.

Of course, I won't be making these assumptions in the code. I intend to move all logic to callers rather than make decisions based on ipsec_test_part. But the above would be the overall angle that I would take. Just wanted to confirm that with reassembly one input packet may not have an output (yet). And one output packet (one among the partial reassembly case) may not have a corresponding input packet.

If I'm not clear, we can defer this discussion when I revisit reassembly.

test/validation/api/ipsec/ipsec.c
```diff
@@ -987,53 +1007,21 @@ void ipsec_check_out_in_one(const ipsec_test_part *part,
...
 		ipsec_test_part part_in = *part;
 		ipsec_test_packet pkt_in;
-		odp_ipsec_packet_result_t result;
-
-		if (ODP_PACKET_INVALID == pkto[i]) {
-			CU_FAIL("ODP_PACKET_INVALID received");
-			continue;
-		}
 
-		if (ODP_EVENT_PACKET_IPSEC !=
```
anoobj, line 974: Is this intentional? If yes, (assuming that you are removing an existing check), should we add this in commit description?

JannePeltonen: The removed code was duplicate of that in ipsec_check_out(), which this function now calls. The old code called ipsec_send_out_one() which did not do result processing at all. Does this answer your question?

anoobj: Yeah. Thanks for the explanation.

test/validation/api/ipsec/ipsec.c
```diff
@@ -935,21 +935,18 @@ static void parse_ip(odp_packet_t pkt)
...
+		    odp_packet_t *pkto)
 {
-	int num_out = part->num_pkt;
-	odp_packet_t pkto[num_out];
 	int i;
+	int num_out;
 
 	num_out = ipsec_send_out_one(part, sa, pkto);
 
 	for (i = 0; i < num_out; i++) {
```
anoobj, line 877: Do we expect part->num_pkt to be more than 1? If not, this is still ipsec_check_out_one, right? Or may be, we should have this function take an argument 'num_part' and have this function work on more than one part?

test/validation/api/ipsec/ipsec.h
JannePeltonen: > While we are at it, do you think this out[1] should be continued? If the test_part is supposed to indicate one packet and the details associated with it, num_pkt and this [1] looks redundant to me.

ipsec_test_part is about one ipsec *operation*. One IPsec operation for one packet can have output of multiple packets (in case of fragmentation offload) and therefore ipsec_test_part can hold one input packet and multiple output packets. The multiple-output case is not currently used but I did not want to strip it away.

test/validation/api/ipsec/ipsec.c
JannePeltonen: As I wrote in the other response, ipsec_test_part is about one operation and one operation can have multiple packets as output and therefore part->num_pkt could, in princple, be greater than one. Currently there are not such test cases though.

test/validation/api/ipsec/ipsec.h
anoobj: @JannePeltonen Okay. Then I will retain increasing the size of this array as required for reassembly case.

test/validation/api/ipsec/ipsec.c
anoobj: @JannePeltonen Agreed.

test/validation/api/ipsec/ipsec_test_out.c
```diff
@@ -471,7 +471,7 @@ static void test_out_in_common(ipsec_test_flags *flags,
 			  .status.error.all = 0,
 			  .l3_type = ODP_PROTO_L3_TYPE_IPV4,
 			  .l4_type = ODP_PROTO_L4_TYPE_ICMPV4,
```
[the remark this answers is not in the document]
JannePeltonen: Yes it is. It is fixed later in this patch:

472a9dd8c validation: ipsec: always check packet type after ipsec output

That patch fixes also other similar problems and adds actual checking of the l3/l4 types so that the values start to matter (which is why I did not do the fix yet in this patch).

test/validation/api/ipsec/ipsec_test_out.c
```diff
@@ -417,27 +417,27 @@ static void ipsec_pkt_auth_err_set(odp_packet_t pkt)
 	odp_packet_copy_from_mem(pkt, len - sizeof(data), sizeof(data), &data);
 }
 
-static void ipsec_check_out_in_one(const ipsec_test_part *part,
+static void ipsec_check_out_in_one(const ipsec_test_part *part_out,
+				   ipsec_test_part part_in,
```
anoobj, line 435: Why did you choose to pass by value here? Why not pass the pointer? No strong opinion here. Just trying to understand the rationale.

JannePeltonen: Maybe I just did not want to change the existing lines that referred to part_in. But it is a bit silly to pass it by value when part_out is passed by reference. I'll change it in the next version.

anoobj:
@JannePeltonen The PR looks quite good to me. I have some more ideas regarding code refactoring which I'll take up as part of reassembly. You can add my reviewed by tag. There was a minor comment I had posted (regarding pass by value vs pass pointer). I'm okay with either approach.

JannePeltonen:
v2:
- Rebased on latest master
- Changed the second parameter of ipsec_check_out_in_one to be passed by reference instead of by value
- Added reviewed-by tags

@anoobj could you check that the parameter change is ok and this can now be merged.

test/validation/api/ipsec/ipsec_test_out.c
```diff
@@ -417,27 +417,28 @@ static void ipsec_pkt_auth_err_set(odp_packet_t pkt)
...
+	int num_out = part_outbound->num_pkt;
 	odp_packet_t pkto[num_out];
 	int i;
 
-	num_out = ipsec_check_out(part, sa, pkto);
+	num_out = ipsec_check_out(part_outbound, sa, pkto);
 
 	for (i = 0; i < num_out; i++) {
-		ipsec_test_part part_in = *part;
+		ipsec_test_part part_in = *part_inbound;
```
anoobj, line 447: We can remove this if we pass part_inbound not as const, right?

JannePeltonen: Yes, but I do not think it is good to make it non-const and have this function modify the parameter structure owned by the caller for no good reason. IOW, from the caller point-of-view it makes more sense to have a const pointer parameter since the function is not supposed to really return anything through it.

anoobj: Agreed. Thanks for the confirmation.

anoobj:
@JannePeltonen @psavol This change looks good to me.

PR #1234, opened 2021-03-29
[PATCH v4] traffic_mngr: linux-gen and validation compliance fixes.

nithind1988:
commit 03db2a3cb430937694c752581185020d2b9049ff
Date:   Sat Mar 27 14:40:44 2021 +0530

    linux-generic: pktio: add internal API to configure pktout for TM
    
    Add internal routine to support configuring pktout when in TM mode.
    
commit 9127eb2fd70eca739c7213ae1f188a353dd10234
Date:   Thu Nov 5 06:22:55 2020 +0530

    linux-gen: tm: fix issues related to name table cleanup
    
    Name table entry was not cleaned up properly in destroy path
    of tm profile/node/system. This patch fixes the same.
    
commit fd3d1acb96103fd74809121788eafc1935af9007
Date:   Sat Mar 27 14:17:29 2021 +0530

    validation: traffic_mngr: use pktout mode as TM for Tx
    
    Use ODP_PKTOUT_MODE_TM for pktio that is used for Tx. This is
    as per TM spec. Also change suite init/term to handle cases where
    TM is not supported by platform which can only be known when
    odp_pktio_open() with out mode as TM fails.
    
commit 08071d0787a7027cc9e6602a5cff786ff77032fd
Date:   Wed Nov 4 22:41:56 2020 +0530

    test: cunit: add API to deactivate all test cases
    
    In test suite init, it is possible to detect that a given test suite
    is not applicable. Hence provide an API to deactivate all the
    test cases of a given test suite.
    
### DIFF
--- a/platform/linux-generic/odp_packet_io.c
+++ b/platform/linux-generic/odp_packet_io.c
@@ -2065,6 +2065,77 @@ int odp_pktin_queue_config(odp_pktio_t pktio,
 	return 0;
 }
 
+int _odp_pktio_pktout_tm_config(odp_pktio_t pktio_hdl,
+				odp_pktout_queue_t *queue, bool reconf)
+{
+	odp_pktout_queue_param_t param;
+	bool pktio_started = false;
+	odp_pktout_mode_t mode;
+	pktio_entry_t *entry;
+	unsigned int i;
+	int rc = 0;
+
+	odp_pktout_queue_param_init(&param);
+	param.num_queues = 1;
+
+	entry = get_pktio_entry(pktio_hdl);
+	if (entry == NULL) {
+		ODP_DBG("pktio entry %d does not exist\n", pktio_hdl);
+		return -1;
+	}
+
+	rc = -ENOTSUP;
+	mode = entry->s.param.out_mode;
+	/* Don't proceed further if mode is not TM */
+	if (mode != ODP_PKTOUT_MODE_TM)
+		return rc;
+
+	/* Don't reconfigure unless requested */
+	if (entry->s.num_out_queue && !reconf) {
+		*queue = entry->s.out_queue[0].pktout;
+		return 0;
+	}
+
+	if (entry->s.state == PKTIO_STATE_STARTED) {
+		pktio_started = true;
+		rc = odp_pktio_stop(pktio_hdl);
+		if (rc) {
+			ODP_ERR("Unable to stop pktio, rc=%d\n", rc);
+			return rc;
+		}
+	}
+
+	/* If re-configuring, destroy old queues */
+	if (entry->s.num_out_queue) {
+		destroy_out_queues(entry, entry->s.num_out_queue);
+		entry->s.num_out_queue = 0;
+	}
+
+	init_out_queues(entry);
+	for (i = 0; i < param.num_queues; i++) {
+		entry->s.out_queue[i].pktout.index = i;
+		entry->s.out_queue[i].pktout.pktio = pktio_hdl;
+	}
+
+	entry->s.num_out_queue = param.num_queues;
+
+	rc = 0;
+	if (entry->s.ops->output_queues_config) {
+		rc = entry->s.ops->output_queues_config(entry, &param);
+		if (rc)
+			ODP_ERR("Unable to setup output queues, rc=%d\n", rc);
+	}
+
+	/* Return pktout queue on success */
+	if (!rc)
+		*queue = entry->s.out_queue[0].pktout;
+
+	/* Take pktio back to its previous state */
+	if (pktio_started)
+		rc |= odp_pktio_start(pktio_hdl);
+	return rc;
+}
+
 int odp_pktout_queue_config(odp_pktio_t pktio,
 			    const odp_pktout_queue_param_t *param)
 {
--- a/platform/linux-generic/odp_traffic_mngr.c
+++ b/platform/linux-generic/odp_traffic_mngr.c
@@ -3065,6 +3065,7 @@ int odp_tm_destroy(odp_tm_t odp_tm)
 	_odp_queue_pool_destroy(tm_system->_odp_int_queue_pool);
 	_odp_timer_wheel_destroy(tm_system->_odp_int_timer_wheel);
 
+	_odp_int_name_tbl_delete(tm_system->name_tbl_id);
 	tm_system_free(tm_system);
 	return 0;
 }
@@ -3573,7 +3574,7 @@ int odp_tm_wred_destroy(odp_tm_wred_t wred_profile)
 		return -1;
 
 	return tm_common_profile_destroy(wred_profile,
-					 ODP_INVALID_NAME);
+					 wred_params->name_tbl_id);
 }
 
 int odp_tm_wred_params_read(odp_tm_wred_t wred_profile,
--- a/platform/linux-generic/odp_traffic_mngr.c
+++ b/platform/linux-generic/odp_traffic_mngr.c
@@ -22,6 +22,7 @@
 #include <odp/api/std_types.h>
 #include <protocols/eth.h>
 #include <protocols/ip.h>
+#include <odp_packet_io_internal.h>
 #include <odp_traffic_mngr_internal.h>
 #include <odp/api/plat/packet_inlines.h>
 #include <odp/api/plat/byteorder_inlines.h>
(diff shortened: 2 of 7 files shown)


platform/linux-generic/odp_traffic_mngr.c
```diff
@@ -2918,9 +2919,11 @@ odp_tm_t odp_tm_create(const char            *name,
 	/* If we are using pktio output (usual case) get the first associated
 	 * pktout_queue for this pktio and fail if there isn't one.
 	 */
-	if (egress->egress_kind == ODP_TM_EGRESS_PKT_IO &&
-	    odp_pktout_queue(egress->pktio, &pktout, 1) != 1)
-		return ODP_TM_INVALID;
+	if (egress->egress_kind == ODP_TM_EGRESS_PKT_IO) {
+		rc = _odp_pktio_pktout_tm_config(egress->pktio, &pktout);
```
JannePeltonen, line 2923: I wonder if the TM API allows having more than one TM system attached to the same pktio egress. I could not immediately see anything in the current linux-gen implementation that would not work in that case (but I may have missed it). After this change there can be only one TM system for one pktio. I wonder if there should be an API change or API clarification about it.

The commit message of this patch needs improvement. It should describe better what is done and why (pktio output done by TM implementation is done in a way that requires an output queue to exists, but the ODP application cannot configure it when output mode is TM). The current message also says that this adds an internal API, but this patch also adds usage for that API.

JannePeltonen (commented):
This patch (linux-gen: tm: fix issues related to name table cleanup) looks fine.

test/validation/api/traffic_mngr/traffic_mngr.c
```diff
@@ -2094,9 +2123,24 @@ static int traffic_mngr_suite_init(void)
...
+	/* Positive return indicates, that pktio open failed with out mode as TM
+	 * but succeeded with direct mode.
+	 */
+	if (ret > 0)
+		goto skip_tests;
+
+	return 0;
+skip_tests:
+	/* Mark all tests as inactive under this suite */
+	odp_cunit_set_inactive();
```
JannePeltonen, line 2142: I am not sure we really need this new API in the common test framework. There are at least two ways this code could effectively disable the test without having the new API:
1) Do the pktio open etc before suite init so that suite init and test case registration is not even done if the pktio is not suitable
2) Use ODP_TEST_INFO_CONDITIONAL and the check function to disable every test at registration time if the init phase has failed.

test/validation/api/traffic_mngr/traffic_mngr.c
```diff
@@ -2107,14 +2151,20 @@ static int traffic_mngr_suite_term(void)
 	/* Close the pktios and associated packet pools. */
 	free_rcvd_pkts();
 	for (iface = 0; iface < num_ifaces; iface++) {
-		if (odp_pktio_stop(pktios[iface]) != 0)
-			return -1;
+		/* Skip pktios not initialized */
+		if (pktios[iface] != ODP_PKTIO_INVALID) {
+			if (odp_pktio_stop(pktios[iface]) != 0)
```
JannePeltonen, line 2156: This may call odp_pktio_stop() for pktios that have not been started, which is against the API spec. Maybe it should be tracked separately whether the pktio has been started.

test/validation/api/traffic_mngr/traffic_mngr.c
```diff
@@ -510,15 +540,13 @@ static int open_pktios(void)
...
-		pktios[iface] = pktio;
-
 		if (odp_pktin_queue(pktio, &pktins[iface], 1) != 1) {
 			odp_pktio_close(pktio);
 			ODPH_ERR("odp_pktio_open() failed: no pktin queue\n");
 			return -1;
 		}
 
-		if (odp_pktout_queue(pktio, &pktouts[iface], 1) != 1) {
+		if (iface && odp_pktout_queue(pktio, &pktouts[iface], 1) != 1) {
```
JannePeltonen, line 549: This could be just removed, along with the odp_pktout_queue_config() that is done a bit earlier. The API says: "When packet output mode is ODP_PKTOUT_MODE_DISABLED or ODP_PKTOUT_MODE_TM, odp_pktout_queue_config() call is optional and will ignore all parameters."

test/validation/api/traffic_mngr/traffic_mngr.c
```diff
@@ -494,10 +497,37 @@ static int open_pktios(void)
...
+						       &pktio_param);
+
+				/* Return +ve to indicate no TM support */
+				if (pktio != ODP_PKTIO_INVALID) {
+					odp_pktio_close(pktio);
+					return 1;
+				}
+			}
+		} else {
+			pktio_param.out_mode = ODP_PKTOUT_MODE_DIRECT;
```
JannePeltonen, line 524: This could as well be ODP_PKTOUT_MODE_DISABLED.

test/validation/api/traffic_mngr/traffic_mngr.c
nithind1988: Method '1' is not sufficient as in changes in #1237 for every suite I validate whether platform supports enough levels, modes etc,  If I do it in main before cunit registration, then it will be not clean I guess.

Method '2' can be done but just that every test case will become ODP_TEST_INFO_CONDITIONAL() though the it is a global inactive decision.

platform/linux-generic/odp_traffic_mngr.c
nithind1988: Ack, I broke that functionality. I believe API does allow multiple TM systems with same egress. Our platform however doesn't support it. Should I change the logic to do pktout queue setup in _odp_pktio_pktout_tm_config() only once ?

nithind1988:
@JannePeltonen Any comments on this updated v2 ?

test/validation/api/traffic_mngr/traffic_mngr.c
JannePeltonen: Ok, let's have this as proposed.

platform/linux-generic/odp_packet_io.c
```diff
@@ -2065,6 +2065,77 @@ int odp_pktin_queue_config(odp_pktio_t pktio,
...
+	}
+
+	rc = -ENOTSUP;
+	mode = entry->s.param.out_mode;
+	/* Don't proceed further if mode is not TM */
+	if (mode != ODP_PKTOUT_MODE_TM)
+		return rc;
+
+	/* Don't reconfigure unless requested */
+	if (entry->s.num_out_queue && !reconf) {
```
JannePeltonen, line 2094: Here we have a race between multiple threads creating TM systems at the same time on top of the same pktio egress. I guess a new lock around this queue creation code here would take care of it.

test/validation/api/traffic_mngr/traffic_mngr.c
```diff
@@ -317,8 +319,8 @@ static uint32_t num_ifaces;
 static odp_pool_t pools[MAX_NUM_IFACES] = {ODP_POOL_INVALID, ODP_POOL_INVALID};
 
 static odp_pktio_t pktios[MAX_NUM_IFACES];
+static bool pktio_started[MAX_NUM_IFACES];
```
JannePeltonen, line 322: Maybe this should be odp_bool_t instead of bool, as ODP seems to avoid using the latter (but I am not sure why).

test/validation/api/traffic_mngr/traffic_mngr.c
```diff
@@ -494,36 +497,54 @@ static int open_pktios(void)
...
+			/* On failure check if pktio can be opened in non-TM mode.
+			 * If non-TM mode works, then we can assume that PKTIO
+			 * does not support TM
+			 */
+			if (pktio == ODP_PKTIO_INVALID) {
+				pktio_param.out_mode = ODP_PKTOUT_MODE_DIRECT;
+				pktio = odp_pktio_open(iface_name[iface], pkt_pool,
+						       &pktio_param);
+
+				/* Return +ve to indicate no TM support */
```
JannePeltonen, line 517: It took me a moment to understand what '+ve' stands for. Maybe it could be replaces with '>0'.

platform/linux-generic/odp_packet_io.c
JannePeltonen: The existing tm_glb->create_lock in the caller can probably handle this if the pktout config is moved below it, right?

nithind1988: Ya, thats true as for PKTIO in TM mode, everyone can only configure pktout using _odp_pktio_pktout_tm_config() via tm_create(). Will move it down in v3

nithind1988:
Addressed comments from Janne

JannePeltonen (approved):

nithind1988:
Added Reviewed-by: Janne Peltonen janne.peltonen@nokia.com

PR #1236, opened 2021-03-29
[PATCH v12] api: tm: add start and stop tm API

nithind1988:
Depends on series in order:
1. https://github.com/OpenDataPlane/odp/pull/1234
2. https://github.com/OpenDataPlane/odp/pull/1235

commit 358c96f4d80dc71f30b71270bf3525c936525aa7
Date:   Wed Nov 4 19:36:31 2020 +0530

    validation: tm: use tm start and stop API
    
    Add tm start API after creating a TM system and TM stop API before
    destroying it.
    
commit 3f46857434bb0922b1cc4dd80d01accdd7586642
Date:   Wed Nov 4 19:23:55 2020 +0530

    example: tm: use TM start and stop
    
    Use TM start to start the TM system after creation and
    TM stop to stop it before destroying TM queues.
    
commit d12ca8902200c5779f3b264e016b6167573fb9f5
Date:   Wed Nov 4 19:22:14 2020 +0530

    linux-gen: tm: add implementation of TM start/stop
    
    Add implementation of TM start or stop API. Since linux generic platform
    does not have any setup and cleanup constraints, the TM start/stop
    are a no-op.
    
commit cae6e0c352188dd456caba7790ceb2fa3d5a75e7
Date:   Wed Nov 4 19:21:28 2020 +0530

    api: tm: add interface to start and stop a TM system
    
    Some platforms needs to know when a TM system needs to
    start or stop in order to safely setup or cleanup TM system
    and drain pkts in it. Hence this patch adds support for
    TM start/stop API.
    
### DIFF
--- a/example/traffic_mgmt/odp_traffic_mgmt.c
+++ b/example/traffic_mgmt/odp_traffic_mgmt.c
@@ -826,6 +826,13 @@ int main(int argc, char *argv[])
 
 	create_and_config_tm();
 
+	/* Start TM */
+	rc = odp_tm_start(odp_tm_test);
+	if (rc != 0) {
+		printf("Error: odp_tm_start() failed, rc=%d\n", rc);
+		return -1;
+	}
+
 	odp_random_data(random_buf, RANDOM_BUF_LEN, 1);
 	next_rand_byte = 0;
 
@@ -840,6 +847,13 @@ int main(int argc, char *argv[])
 	       pkts_into_tm, pkts_from_tm);
 
 	odp_tm_stats_print(odp_tm_test);
+
+	/* Stop TM */
+	rc = odp_tm_stop(odp_tm_test);
+	if (rc != 0) {
+		printf("Error: odp_tm_stop() failed, rc = %d\n", rc);
+		return -1;
+	}
 
 	rc = destroy_tm_queues();
 	if (rc != 0) {
--- a/include/odp/api/spec/traffic_mngr.h
+++ b/include/odp/api/spec/traffic_mngr.h
@@ -605,6 +605,42 @@ odp_tm_t odp_tm_find(const char            *name,
  */
 int odp_tm_capability(odp_tm_t odp_tm, odp_tm_capabilities_t *capabilities);
 
+/**
+ * Start a TM system
+ *
+ * odp_tm_start() needs to be used to start an already created or found TM
+ * system. By default, all the TM systems are in stopped state.
+ *
+ * @param tm  TM system to be started
+ *
+ * @retval 0  on success
+ * @retval <0 on failure
+ */
+int odp_tm_start(odp_tm_t tm);
+
+/**
+ * Stop a TM system
+ *
+ * odp_tm_stop() can to used to stop a TM system that is already started for the
+ * purpose of reconfiguration that cannot be done dynamically.
+ *
+ * When TM is in the stopped state,
+ * - New packets must not be sent to the TM either directly by TM API or indirectly
+ *   via IPsec outbound inline API. New packets can only be enqueued after
+ *   starting the TM system using odp_tm_start().
+ * - Packets already inflight inside TM or IPSec for transmit may get silently dropped
+ *   or may get transmitted with unspecified TM treatment.
+ *
+ * A following call to odp_tm_start() restarts TM system and its scheduling/shaping
+ * on existing and new packets.
+ *
+ * @param tm  TM system to be stopped
+ *
+ * @retval 0  on success
+ * @retval <0 on failure
+ */
+int odp_tm_stop(odp_tm_t tm);
+
 /** Destroy a TM system.
  *
  * odp_tm_destroy() may be used to destroy TM systems created via
--- a/include/odp/api/spec/traffic_mngr.h
+++ b/include/odp/api/spec/traffic_mngr.h
@@ -331,6 +331,52 @@ typedef struct {
 	/** The per_level array specifies the TM system capabilities that
 	 * can vary based upon the tm_node level. */
 	odp_tm_level_capabilities_t per_level[ODP_TM_MAX_LEVELS];
+
+	/** dynamic_topology_update indicates support for TM system dynamic
+	 * topology update. A dynamic topology update is defined as update to
+	 * a TM system topology while TM system is not in stopped state.
+	 * When TRUE, application can update topology dynamically
+	 * without bringing the TM system to stopped state. When FALSE,
+	 * application has to call odp_tm_stop() before updating the
+	 * topology and odp_tm_start() after completing the update.
+	 */
+	odp_bool_t dynamic_topology_update;
+
+	/** dynamic_shaper_update indicates support for TM system's dynamic
+	 * shaper profile changes. When TRUE, application can update shaper
+	 * profile of a TM queue or TM node dynamically.
+	 * When FALSE, it implies that TM system should be brought to
+	 * stopped state before changing the shaper profile or updating
+	 * the parameters of the shaper profile of any TM node or TM queue.
+	 */
+	odp_bool_t dynamic_shaper_update;
+
+	/** dynamic_sched_update indicates support for TM system's dynamic
+	 * sched profile changes. When TRUE, application can update sched
+	 * profile of a TM queue or TM node dynamically.
+	 * When FALSE, it implies that TM system should be brought to
+	 * stopped state before changing the sched profile or updating
+	 * the parameters of the sched profile of any TM node or TM queue.
+	 */
+	odp_bool_t dynamic_sched_update;
+
+	/** dynamic_wred_update indicates support for TM system's dynamic
+	 * wred profile changes. When TRUE, application can update wred
+	 * profile of a TM queue or TM node dynamically.
+	 * When FALSE, it implies that TM system should be brought to
+	 * stopped state before changing the wred profile or updating
+	 * the parameters of the wred profile of any TM node or TM queue.
+	 */
+	odp_bool_t dynamic_wred_update;
+
+	/** dynamic_threshold_update indicates support for TM system's dynamic
+	 * threshold profile changes. When TRUE, application can update
+	 * threshold profile of a TM queue or TM node dynamically.
+	 * When FALSE, it implies that TM system should be brought to
+	 * stopped state before changing the threshold profile or updating
+	 * the parameters of the threshold profile of any TM node or TM queue.
+	 */
+	odp_bool_t dynamic_threshold_update;
 } odp_tm_capabilities_t;
 
 /** Per Level Requirements
@@ -638,6 +684,12 @@ int odp_tm_start(odp_tm_t tm);
  *
  * @retval 0  on success
  * @retval <0 on failure
+ *
+ * @see odp_tm_capabilities_t::dynamic_topology_update
+ * @see odp_tm_capabilities_t::dynamic_shaper_update
+ * @see odp_tm_capabilities_t::dynamic_sched_update
+ * @see odp_tm_capabilities_t::dynamic_wred_update
+ * @see odp_tm_capabilities_t::dynamic_threshold_update
  */
 int odp_tm_stop(odp_tm_t tm);
 
--- a/test/validation/api/traffic_mngr/traffic_mngr.c
+++ b/test/validation/api/traffic_mngr/traffic_mngr.c
@@ -283,6 +283,11 @@ static uint32_t        num_odp_tm_systems;
 
 static odp_tm_capabilities_t tm_capabilities;
 
+static bool dynamic_shaper_update = true;
+static bool dynamic_sched_update = true;
+static bool dynamic_threshold_update = true;
+static bool dynamic_wred_update = true;
+
 static odp_tm_shaper_t    shaper_profiles[NUM_SHAPER_PROFILES];
 static odp_tm_sched_t     sched_profiles[NUM_SCHED_PROFILES];
 static odp_tm_threshold_t threshold_profiles[NUM_THRESHOLD_PROFILES];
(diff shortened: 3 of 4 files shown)


include/odp/api/spec/traffic_mngr.h
```diff
@@ -605,6 +605,36 @@ odp_tm_t odp_tm_find(const char            *name,
  */
 int odp_tm_capability(odp_tm_t odp_tm, odp_tm_capabilities_t *capabilities);
 
+/** Start a TM system.
+ *
+ * odp_tm_start() needs to be used to start an already created / found TM system
+ * By default, all the TM systems are in stopped state.
+ * Calling this function on an already started TM system will be ignored.
```
JannePeltonen, line 612: Do we need to allow this? Why not "This function may not be called for a TM system that is already in the started state."?

include/odp/api/spec/traffic_mngr.h
```diff
@@ -605,6 +605,36 @@ odp_tm_t odp_tm_find(const char            *name,
...
+/** Stop a TM system.
+ *
+ * odp_tm_stop() can to be used to stop a TM system that is already started.
+ * Stopping a TM system will cause it to stop / pause processing packets and
+ * become idle.
+ *
+ * When TM is in the stopped state, new packets sent to the TM (directly or
+ * indirectly) may get silently dropped or may get transmitted without TM
+ * treatment or the sending operation may fail. Packets already stored for
+ * transmit are not freed.
```
JannePeltonen, line 628: > Packets already stored for transmit are not freed.

What exactly does that mean ("stored for transmit" does not appear to be a well defined concept in the API) and why is the statement needed in the API? If a packet is dropped, then I would expect it is also freed.

JannePeltonen:
I think the point of this start and stop was to limit dynamic updates of the TM config to the stopped state unless dynamic updates at any time are supported by the underlying implementation. This PR does not now properly integrate this start and stop to the rest of the API:

- Capabilities that tell whether dynamic updates are supported or whether a TM system must be stopped first are missing. There is a capability adding API patch in PR 1237. It does not belong there but in this PR.
- Maybe the update functions or other part of the existing API needs a mention that a dynamic update without stopping a TM is allowed only if the relevant capability is present.

nithind1988:
Addressed all comments in v2

include/odp/api/spec/traffic_mngr.h
```diff
@@ -331,6 +331,52 @@ typedef struct {
...
+	 * topology and odp_tm_start() after completing the update.
+	 */
+	odp_bool_t dynamic_topology_update;
+
+	/** dynamic_shaper_update indicates support for TM system's dynamic
+	 * shaper profile changes. When TRUE, application can update shaper
+	 * profile of a TM queue or TM node dynamically.
+	 * When FALSE, it implies that TM system should be brought to
+	 * stopped state before doing any TM node or TM queue shaper profile
+	 * change.
```
JannePeltonen, line 350: I wonder if it is clear that the lack of dynamic update capability means that the shaper of a node 0or queue cannot be touched in either of the two ways that are possible (odp_tm_{queue,node}_shaper_config() and odp_tm_shaper_params_update()). Maybe the end of the last sentence could be rephrased like this "...before changing the shaper profile or updating the parameters of the shaper profile of any TM node or TM queue."

And the same comment for the other similar capability fields.

example/traffic_mgmt/odp_traffic_mgmt.c
```diff
@@ -753,6 +755,8 @@ static int destroy_tm_queues(void)
 	int class;
 	int ret;
 
+	odp_tm_stop(odp_tm_test);
```
JannePeltonen, line 758: I think it would be better to raise odp_tm_start() and odp_tm_stop() calls to the top level so that they stand out better. The stop is anyway not really logically part of queue destroying so it seems a bit out of place here.

test/validation/api/traffic_mngr/traffic_mngr.c
```diff
@@ -1664,6 +1664,14 @@ static int create_tm_system(void)
 		return -1;
 	}
 
+	/* Start TM system */
+	CU_ASSERT((rc = odp_tm_start(odp_tm)) == 0);
```
JannePeltonen, line 1668: The validation test now incorrectly assumes that dynamic updates are always supported as the TM is not stopped before dynamic updates in the test cases such as traffic_mngr_test_shaper().

The test should do stop-start around modifications either always or depending on the dynamic update capability.

JannePeltonen:
It appears that the TM stopping added in set_shaper() in the validation test code can break the logic of its callers. Several test functions call set_shaper to modify shaping rate after sending packets to TM but before receiving them back. Since stopping the TM may cause the in-flight packets to be dropped, the test functions may not receive the packets they expect.

The following functions seem to be affected:

test_sched_queue_priority()
test_sched_node_priority()
test_sched_wfq()
test_byte_wred()
test_pkt_wred()
test_query_functions()

The other commits looked ok to me.

include/odp/api/spec/traffic_mngr.h
```diff
@@ -605,6 +605,36 @@ odp_tm_t odp_tm_find(const char            *name,
...
 
+/** Start a TM system.
+ *
+ * odp_tm_start() needs to be used to start an already created / found TM system
+ * By default, all the TM systems are in stopped state.
+ *
+ * @param odp_tm  The odp_tm_t value of the TM system to be started.
+ * @return        0 upon success, < 0 upon failure.
+ */
+int odp_tm_start(odp_tm_t odp_tm);
```
psavol, line 662: Documentation format could be update to match other APIs. TM API is not the best reference on how to document APIs.

- The first sentence misses a period (".") at the end: "... already created / found TM system."
- The first parameter could be simply "tm" and documented as TM system (handle):  "@param tm  TM system to be started"
- Simple return codes are usually documented with retval (return is used for more complex documentation):
```
@retval 0 on success
@retval <0 on failure

File: include/odp/api/spec/traffic_mngr.h [hunk 7]
```diff
@@ -605,6 +605,36 @@ odp_tm_t odp_tm_find(const char            *name,
...
+ * - Packets already in queue for transmit may get silently dropped or may get
+ *   transmitted without TM treatment.
+ *
+ * A following call to odp_tm_start() restarts TM system and its scheduling/shaping
+ * on existing and new packets.
+ *
+ * @param odp_tm  The odp_tm_t value of the TM system to be stopped.
+ * @return        0 upon success, < 0 upon failure.
+ */
+int odp_tm_stop(odp_tm_t odp_tm);
```
psavol, line 688: Also here:
- odp_tm -> tm
- use retval

nithind1988:
> It appears that the TM stopping added in set_shaper() in the validation test code can break the logic of its callers. Several test functions call set_shaper to modify shaping rate after sending packets to TM but before receiving them back. Since stopping the TM may cause the in-flight packets to be dropped, the test functions may not receive the packets they expect.
> 
> The following functions seem to be affected:
> 
> test_sched_queue_priority()
> test_sched_node_priority()
> test_sched_wfq()
> test_byte_wred()
> test_pkt_wred()
> test_query_functions()
> 
> The other commits looked ok to me.

Since there is no other alternative as we cannot even put the odp_tm_start() and odp_tm_stop() outside set_schaper() to resolve this issue, I'll disable these tests if dynamic shaper update is not supported. Is it fine ?

nithind1988:
> 
> 
> > It appears that the TM stopping added in set_shaper() in the validation test code can break the logic of its callers. Several test functions call set_shaper to modify shaping rate after sending packets to TM but before receiving them back. Since stopping the TM may cause the in-flight packets to be dropped, the test functions may not receive the packets they expect.
> > The following functions seem to be affected:
> > test_sched_queue_priority()
> > test_sched_node_priority()
> > test_sched_wfq()
> > test_byte_wred()
> > test_pkt_wred()
> > test_query_functions()
> > The other commits looked ok to me.
> 
> Since there is no other alternative as we cannot even put the odp_tm_start() and odp_tm_stop() outside set_schaper() to resolve this issue, I'll disable these tests if dynamic shaper update is not supported. Is it fine ?

Pushed v5 with changes to disable those tests when dynamic shaper update is not supported.

JannePeltonen:
> >  Since there is no other alternative as we cannot even put the odp_tm_start() and odp_tm_stop() outside set_schaper() to resolve this issue, I'll disable these tests if dynamic shaper update is not supported. Is it fine ?

> Pushed v5 with changes to disable those tests when dynamic shaper update is not supported.

It is not ok to simply disable a whole bunch of tests like that. It is also not true that there is no other alternative. That would be case when the feature being tested is not present (e.g. after introduction of a capability flag for a previously mandatory feature), but that is not the case here.

Looking at the code, the offending set_shaper() calls are for disabling the shaper to speed up flushing the TM. In those cases where the packets are still received and verified after the set shaper call, the new TM start and stop cannot be done. But since the set_shaper call is just for speeding up the test, one could simply remove it, at the cost of longer test execution time. But a test that takes long is still much better than a test that is not run at all.

I just tried the above quickly and the test duration increased to about 260 seconds from the few seconds that it was before. I would think that is still ok, but maybe more insightful changes in the validation code could bring the test time back down. I had to still disable shaper in test_sched_wfq() and test_pkt_wred() (but after packet reception this time) for subsequent tests to pass.

So, please change the validation test code as needed so that the tests do not have to be unnecessarily disabled.

nithind1988:
>So, please change the validation test code as needed so that the tests do not have to be unnecessarily disabled.
Fixed this in V7. Now if dynamic shaper update is supported, the behavior is as existing. If it is not supported, then disable shaper will not be done at places where we don't expect packet drops.

include/odp/api/spec/traffic_mngr.h
```diff
@@ -605,6 +605,42 @@ odp_tm_t odp_tm_find(const char            *name,
...
+ * @retval <0 on failure
+ */
+int odp_tm_start(odp_tm_t tm);
+
+/**
+ * Stop a TM system
+ *
+ * odp_tm_stop() can to be used to stop a TM system that is already started.
+ * Stopping a TM system will cause it to stop / pause processing packets and
+ * become idle.
```
JannePeltonen, line 672: The API text here may contradict with the later text or at least be confusing. Here it is said that the TM systems stops processing packets and becomes idle, but the description below says that the TM may continue transmitting packets (but without QoS treatment).

Then it looks like the linux-gen implementation, which supports dynamic updates and implements odp_tm_stop() as a no-op, may be in violation of this API spec. Linux gen TM in a stopped state will continue processing packets with QoS treatment, although that is not mentioned as a possibility in the text. Also, the TM does not become idle if it is fed enough packets (e.g. by another thread), so the promise about becoming idle is not kept.

I am not sure to what extent this problem is about the API text being a bit vague and not fully communicating the real intent and to what extent the intended API needs rethinking.

One possibility would be to define odp_tm_stop() like this:

```
 * Stop a TM system
 *
 * odp_tm_stop() can to be used to stop a TM system that is already started.
 * When a TM system is in a stopped state, its configuration can be freely
 * modified but it is undefined whether the TM system is processing packets
 * normally, processing packets without TM treatment, dropping packets silently
 * or returning an error to possible enqueue operation.
 *
 * Stopping a TM system is needed for changing TM configuration for which
 * dynamic updates are not supported. It is undefined whether TM configuration
 * changed done in the stopped stae take effect immediately or when the
 * TM system is started.
 *
 * A TM system must be stopped before it is destroyed.
```

And the description of odp_tm_stop() could be amended with this: "TM configuration changes for which dynamic updating is not supported may not take effect until the TM is started."

I am not sure if the above API is any good but I think it probably matches the intended behaviour as I have understood it. It also makes no claims regarding what odp_tm_is_idle() returns for a stopped TM.

The above API looks a bit ugly to me, partly because it leaves so much unspecified and partly because it calls the functions odp_tm_start() and odp_tm_stop() even though there is no promise that a TM in a stopped state is really stopped in any way. So odp_tm_stop() is actually more like odp_tm_prepare_for_config_change() and odp_tm_start() is really odp_tm_config_changes_done().

test/validation/api/traffic_mngr/traffic_mngr.c
```diff
@@ -2711,8 +2783,12 @@ static int test_sched_queue_priority(const char *shaper_name,
 
 	busy_wait(100 * ODP_TIME_MSEC_IN_NS);
 
-	/* Disable the shaper, so as to get the pkts out quicker. */
-	set_shaper(node_name, shaper_name, 0, 0);
+	/* Disable the shaper, so as to get the pkts out quicker.
+	 * We cannot do this if dynamic shaper update is not supported
+	 * without it, set_shaper() can cause packet drops due to start/stop.
```
JannePeltonen, line 2788: There is a grammar issue with this sentence. Perhaps it should read: "We cannot do this if dynamic shaper update is not supported. Without dynamic update support set_shaper() can cause packet drops due to start/stop."

JannePeltonen:
v7 basically looks ok, except for the API as I commented. Depending on if and how the API gets changed, the rest of the patches may or may not need updating.

include/odp/api/spec/traffic_mngr.h
nithind1988: I also feel the above text you proposed has lot of undefined behavior.

With existing text, I think since there is an option of processing traffic without TM treatment when in stopped state provided to platforms, I can remove "become idle" part of the text from API to resolve the issue.

And for linux-gen, I can set and clear a flag in odp_tm_start() and odp_tm_stop() and change the odp_tm_enq() to fail if TM is in stopped state if it is fine.

nithind1988: Or, remove the API text that says "New packets will be transmitted without TM treatment" as after fixing linux-generic, packets will not be transmitted. Our platform also doesn't support transmission in stopped state

JannePeltonen: I had a chat with @psavol about this and now we both think that the API should say that when a TM is in the stopped state, an ODP application must not enqueue new packets to it either directly through odp_tm_enq() and odp_tm_enq_with_cnt() or indirectly through odp_ipsec_out_inline(). Thus the application must call odp_tm_start() before calling those functions and must stop calling those functions (for the relevant TM) before calling odp_tm_stop(),

The fate of packets that are in flight, either inside TM or inside IPsec, could be left unspecified: they might get sent out with unspecified TM treatment or they might get dropped.

How does that sound?

JannePeltonen: I am not sure what you mean with linux-generic issue. If you are referring to the fact that the linux-generic code in this PR does not drop packets at tm_enqueue() when the TM is in the stopped state, there is no problem with that with the revised API since an application would not anymore be allowed to call tm_enqueue() when the TM is in the stopped state.

nithind1988: Yes, I meant the same. Apologies for not describing it better.

Ack. I'll leave the linux-generic code for odp_tm_start() and odp_tm_stop() as it is.

nithind1988:
Fixed all comments in v8

include/odp/api/spec/traffic_mngr.h
JannePeltonen: The description of odp_tm_stop() still promises too much as it talks about processing packets and the TM becoming idle although the TM may still continue processing in-flight packets. I would propose changing the first paragraph to something like this:

"odp_tm_stop() can to used to stop a TM system that is already started for the purpose of reconfiguration that cannot be done dynamically."

nithind1988: Ack.

JannePeltonen (approved):

nithind1988:
Added Reviewed by tag from Janne

psavol (approved):
Please add my review tag to the API patches.

nithind1988:
Added Reviewed-by: Petri Savolainen <petri.savolainen@nokia.com> to API patches

nithind1988:
Rebased to latest master

PR #1237, opened 2021-03-29
[PATCH v9] api: tm: add support for non-global packet priority mode in TM system

nithind1988:
**Depends on series in order:**
1. https://github.com/OpenDataPlane/odp/pull/1234
2. https://github.com/OpenDataPlane/odp/pull/1235
3. https://github.com/OpenDataPlane/odp/pull/1236

**Commits to be ignored for this pull request:**
* f7ee3dc validation: tm: use tm start and stop API - Nithin Dabilpuram, 5 months ago
* db7b267 validation: tm: use commit_rate and peak_rate in shaper params - Nithin Dabilpuram, 3 months ago
* c7bbc8d validation: traffic_mngr: use pktout mode as TM for Tx - Nithin Dabilpuram, 2 days ago

**Commits for review:**
commit 4625b916486048466c88beade7c8b85057cde432
Date:   Tue Jul 6 22:19:04 2021 +0530

    validation: tm: add five level test case suite
    
    Current validation suite only validates 3 level TM systems.
    Refactor the test cases to use suite inited parameters instead
    of compile time and add suite init for 5 level TM system.
    
commit 831754cb21fcb5b8029f5d0e88254c8acea7dbf6
Date:   Tue Jul 6 22:21:37 2021 +0530

    validation: tm: use pkt priority mode supported by platform
    
    Use pkt priority mode supported by a given platform for
    TM systems.
    
commit 28c3e865d793f9abfe142cb9452ed470983c915e
Date:   Tue Jun 29 15:35:32 2021 +0530

    linux-gen: tm: express support pkt priority mode
    
    Linux generic only supports Packet Priority mode preserve.
    
commit b869de28d3d13828d1bb1759bb55ce69d7892cb6
Date:   Wed Nov 4 21:57:28 2020 +0530

    api: tm: support different packet priority modes
    
    Current TM spec associates a priority to a packet at TM queue and
    expects all the levels to honour that priority globally in the TM
    system. In some platforms, this is not supportable and a packet's
    priority gets overridden when it passes through a fan-in node, it
    is passing through. In other words, like assigning packet a
    priority based on the TM queue it is enqueued on, every node it
    is passing through overrides the packet priority with a fixed value.
    
    So this patch adds flexibility for platforms to expose pkt priority
    modes supported and also choose the mode per TM system among supported
    modes at time of odp_tm_create().
    
    This patch also adds "max_schedulers_per_node" capability that is
    expressed by a platform. This field expresses maximum number of schedulers
    available which can schedule among multiple equal priority fan-ins
    as some platforms have limitation on how many such schedulers can be
    configured.
    
### DIFF
--- a/include/odp/api/spec/traffic_mngr.h
+++ b/include/odp/api/spec/traffic_mngr.h
@@ -502,6 +502,16 @@ typedef struct {
 	 * odp_tm_requirements_t.
 	 */
 	odp_bool_t pkt_prio_modes[ODP_TM_PKT_PRIO_MODE_MAX];
+
+	/** Maximum number of schedulers supported by a TM node at any level.
+	 * A TM node contains a WFQ/WRR scheduler for each packet priority level
+	 * for which the node has more than one possible input. TM topology and
+	 * priority configuration must be made so that the resulting number of
+	 * WFQ/WRR schedulers does not exceed this capability in any TM node.
+	 *
+	 * The value can vary between 0 and ODP_TM_MAX_PRIORITIES.
+	 */
+	uint8_t max_schedulers_per_node;
 } odp_tm_capabilities_t;
 
 /** Per Level Requirements
--- a/include/odp/api/spec/traffic_mngr.h
+++ b/include/odp/api/spec/traffic_mngr.h
@@ -312,6 +312,33 @@ typedef struct {
 	odp_bool_t tm_node_threshold;
 } odp_tm_level_capabilities_t;
 
+/** The tm_pkt_prio_mode_t enumeration type is used to indicate different
+ * modes a tm system supports with respect to assigning priority to a packet
+ * and propagating it across TM system. All the nodes in a TM system can
+ * function only on single mode specified at time of odp_tm_create().
+ */
+typedef enum odp_tm_pkt_prio_mode {
+	/** Indicates Packet priority preserve mode. In this mode, a packet gets
+	 * its priority based on a TM queue it gets enqueued to and then it
+	 * carries the same priority along with it as long as it is in the TM
+	 * system. At every TM node in the topology, that specific pkt is
+	 * scheduled as per that priority.
+	 */
+	ODP_TM_PKT_PRIO_MODE_PRESERVE,
+
+	/** Indicates Packet priority overwrite mode. In this mode, a packet
+	 * gets a new priority every time it passes through a TM queue or a
+	 * TM node. All the packets fed by a fan-in node will get the same
+	 * priority and that will be valid until overwritten again by another TM
+	 * node. This priority is part of the TM fan-in node parameters and is
+	 * fixed at node creation time.
+	 */
+	ODP_TM_PKT_PRIO_MODE_OVERWRITE,
+
+	/** Max enum of Packet priority mode */
+	ODP_TM_PKT_PRIO_MODE_MAX,
+} odp_tm_pkt_prio_mode_t;
+
 /** TM Capabilities Record.
  *
  * The odp_tm_capabilities_t record type is used to describe the feature set
@@ -468,6 +495,13 @@ typedef struct {
 	 * ODP_TM_QUERY_THRESHOLDS.
 	 */
 	uint32_t tm_queue_query_flags;
+
+	/** Indicates the packet priority modes supported by TM systems on a
+	 * platform. A platform can support multiple packet priority modes. The
+	 * actual mode a TM system runs with is defined by
+	 * odp_tm_requirements_t.
+	 */
+	odp_bool_t pkt_prio_modes[ODP_TM_PKT_PRIO_MODE_MAX];
 } odp_tm_capabilities_t;
 
 /** Per Level Requirements
@@ -587,6 +621,12 @@ typedef struct {
 	 * the application expects to use this color in conjunction with one or
 	 * more of the marking APIs. */
 	odp_bool_t marking_colors_needed[ODP_NUM_PACKET_COLORS];
+
+	/** Packet priority mode.
+	 * TM capabilities indicate which modes are supported.
+	 * The default value is ODP_TM_PKT_PRIO_MODE_PRESERVE.
+	 */
+	odp_tm_pkt_prio_mode_t pkt_prio_mode;
 
 	/** The per_level array specifies the TM system requirements that
 	 * can vary based upon the tm_node level. */
@@ -1473,6 +1513,22 @@ typedef struct {
 	 * greater levels may be connected to the fan-in of tm_node's with
 	 * numerically smaller levels. */
 	uint8_t level;
+
+	/** New strict priority level assigned to packets going through this
+	 * node when packet priority mode is ODP_TM_PKT_PRIO_MODE_OVERWRITE.
+	 * In other packet priority modes this field is ignored. The new
+	 * priority does not affect packet processing in this node but in
+	 * its destination node.
+	 *
+	 * The value must be in the range 0..ODP_TM_MAX_PRIORITIES-1.
+	 * Additionally, the total number of possible priorities seen by
+	 * the destination node must not exceed the max priority configured
+	 * for the destination node.
+	 *
+	 * @see odp_tm_pkt_prio_mode_t
+	 * @see odp_tm_level_requirements_t::max_priority
+	 */
+	uint8_t priority;
 } odp_tm_node_params_t;
 
 /** odp_tm_node_params_init() must be called to initialize any
--- a/platform/linux-generic/odp_traffic_mngr.c
+++ b/platform/linux-generic/odp_traffic_mngr.c
@@ -2548,6 +2548,8 @@ odp_bool_t odp_tm_is_idle(odp_tm_t odp_tm)
 void odp_tm_requirements_init(odp_tm_requirements_t *requirements)
 {
 	memset(requirements, 0, sizeof(odp_tm_requirements_t));
+
+	requirements->pkt_prio_mode = ODP_TM_PKT_PRIO_MODE_PRESERVE;
 }
 
 void odp_tm_egress_init(odp_tm_egress_t *egress)
@@ -2588,6 +2590,9 @@ static int tm_capabilities(odp_tm_capabilities_t capabilities[],
 	cap_ptr->dynamic_sched_update     = true;
 	cap_ptr->dynamic_wred_update      = true;
 	cap_ptr->dynamic_threshold_update = true;
+
+	/* We only support pkt priority mode preserve */
+	cap_ptr->pkt_prio_modes[ODP_TM_PKT_PRIO_MODE_PRESERVE] = true;
 
 	for (color = 0; color < ODP_NUM_PACKET_COLORS; color++)
 		cap_ptr->marking_colors_supported[color] = true;
@@ -2690,6 +2695,8 @@ static void tm_system_capabilities_set(odp_tm_capabilities_t *cap_ptr,
 	cap_ptr->dynamic_sched_update     = true;
 	cap_ptr->dynamic_wred_update      = true;
 	cap_ptr->dynamic_threshold_update = true;
+
+	cap_ptr->pkt_prio_modes[ODP_TM_PKT_PRIO_MODE_PRESERVE] = true;
 
 	for (color = 0; color < ODP_NUM_PACKET_COLORS; color++)
 		cap_ptr->marking_colors_supported[color] =
@@ -2998,6 +3005,11 @@ odp_tm_t odp_tm_create(const char            *name,
 		return ODP_TM_INVALID;
 	}
 
+	/* We only support global pkt priority mode */
+	if (requirements->pkt_prio_mode != ODP_TM_PKT_PRIO_MODE_PRESERVE) {
+		ODP_ERR("Unsupported Packet priority mode\n");
+		return ODP_TM_INVALID;
+	}
 	odp_ticketlock_lock(&tm_glb->create_lock);
 
 	/* If we are using pktio output (usual case) get the first associated
--- a/platform/linux-generic/odp_traffic_mngr.c
+++ b/platform/linux-generic/odp_traffic_mngr.c
@@ -2584,6 +2584,7 @@ static int tm_capabilities(odp_tm_capabilities_t capabilities[],
 	cap_ptr->tm_queue_query_flags          = (ODP_TM_QUERY_PKT_CNT |
 						  ODP_TM_QUERY_BYTE_CNT |
 						  ODP_TM_QUERY_THRESHOLDS);
+	cap_ptr->max_schedulers_per_node       = ODP_TM_MAX_PRIORITIES;
 
 	cap_ptr->dynamic_topology_update  = true;
 	cap_ptr->dynamic_shaper_update    = true;
@@ -2689,6 +2690,7 @@ static void tm_system_capabilities_set(odp_tm_capabilities_t *cap_ptr,
 	cap_ptr->tm_queue_query_flags          = (ODP_TM_QUERY_PKT_CNT |
 						  ODP_TM_QUERY_BYTE_CNT |
 						  ODP_TM_QUERY_THRESHOLDS);
+	cap_ptr->max_schedulers_per_node       = ODP_TM_MAX_PRIORITIES;
 
 	cap_ptr->dynamic_topology_update  = true;
 	cap_ptr->dynamic_shaper_update    = true;
--- a/test/validation/api/traffic_mngr/traffic_mngr.c
+++ b/test/validation/api/traffic_mngr/traffic_mngr.c
@@ -388,6 +388,7 @@ static int test_overall_capabilities(void)
 	odp_tm_capabilities_t        capabilities_array[2];
 	odp_tm_capabilities_t       *cap_ptr;
 	odp_tm_egress_t              egress;
+	odp_bool_t                  *prio_modes;
 	uint32_t                     num_records, idx, num_levels, level;
 	int                          rc;
 
@@ -443,6 +444,11 @@ static int test_overall_capabilities(void)
 				return -1;
 			}
 		}
+
+		/* At least one pkt priority mode needs to be supported */
+		prio_modes = cap_ptr->pkt_prio_modes;
+		CU_ASSERT((prio_modes[ODP_TM_PKT_PRIO_MODE_PRESERVE] != 0) ||
+			  (prio_modes[ODP_TM_PKT_PRIO_MODE_OVERWRITE] != 0))
 	}
 
 	return 0;
@@ -1436,6 +1442,9 @@ static tm_node_desc_t *create_tm_node(odp_tm_t        odp_tm,
 
 	node_params.max_fanin = FANIN_RATIO;
 	node_params.level = level;
+	/* This is ignored when pkt priority mode is not overwrite */
+	node_params.priority = 0;
+
 	if (parent_node_desc == NULL)
 		snprintf(node_name, sizeof(node_name), "node_%" PRIu32,
 			 node_idx + 1);
@@ -1679,6 +1688,16 @@ set_reqs_based_on_capas(odp_tm_requirements_t *req)
 
 	if (tm_capabilities.tm_queue_shaper_supported)
 		req->tm_queue_shaper_needed = true;
+
+	/* We can use any packet priority mode since it does not affect
+	 * our tests. Our scheduler test tests scheduling only in a node
+	 * directly connected to TM queues and such nodes see the original
+	 * packet priority before it could have been overwritten by any node.
+	 */
+	req->pkt_prio_mode = ODP_TM_PKT_PRIO_MODE_PRESERVE;
+	if (!tm_capabilities.pkt_prio_modes[ODP_TM_PKT_PRIO_MODE_PRESERVE])
+		req->pkt_prio_mode = ODP_TM_PKT_PRIO_MODE_OVERWRITE;
+
 }
 
 static int create_tm_system(void)
--- a/test/validation/api/traffic_mngr/traffic_mngr.c
+++ b/test/validation/api/traffic_mngr/traffic_mngr.c
@@ -2267,6 +2267,14 @@ static int traffic_mngr_suite_init(void)
 	if (j != NUM_LEVELS)
 		goto skip_tests;
 
+	if (egress_capa.pkt_prio_modes[ODP_TM_PKT_PRIO_MODE_PRESERVE] &&
+	    egress_capa.max_schedulers_per_node < NUM_QUEUES_PER_NODE)
+		goto skip_tests;
+
+	if (!egress_capa.pkt_prio_modes[ODP_TM_PKT_PRIO_MODE_PRESERVE] &&
+	    egress_capa.max_schedulers_per_node < 1)
+		goto skip_tests;
+
 	/* Init tm capabilities with matching egress capa until tm is created */
 	tm_capabilities = egress_capa;
 


JannePeltonen:
Generic comments on the patch set:

4ee4e22fe validation: tm: add five level test case suite

This patch contains multiple changes. The code changes would be clearer and easier to follow if unrelated things were put in different patches.

The different changes that this patch contains, includes:

- Checking newly added capabilities of old functionality and making the respective tests conditional on them
- Refactoring and adding 5-level test case

----

f7ee3dc0f validation: tm: use tm start and stop API
db7b26779 validation: tm: use commit_rate and peak_rate in shaper params
c7bbc8d74 validation: traffic_mngr: use pktout mode as TM for Tx
1ab3219a0 api: tm: add capability to express support for dynamic update

These patches do not belong in this PR but they are patches of other PRs. If there is a real dependency (I am not sure there is) on these patches, then I think they should come first in the series and not interleaved with the actual patches of this PR.

But maybe this gets resolved if the other PRs get merged first and this one then rebased.

----

There are no validation tests for the new node stats query API nor for the non-global packet priority mode. I am not sure if there should be.

include/odp/api/spec/traffic_mngr.h
```diff
@@ -246,6 +246,29 @@ typedef struct {
...
+	/** Indicates Packet priority global mode. In this mode, a packet gets its priority
+	 * based on a TM queue it gets enqueued to and then it carries the same priority along
+	 * with it as long as it is in the TM system. At every TM node in the topology,
+	 * that specific pkt is scheduled as per that priority.
+	 */
+	ODP_TM_PKT_PRIO_MODE_GLOBAL,
+	/** Indicates Packet priority override mode. In this mode, a packet gets a new priority
+	 * every time it passes through a TM queue or a TM node. All the packets fed by a fan-in
+	 * node will get the same priority and that will be valid until it reaches the next
+	 * TM node. This priority is part of the TM fan-in node parameters and is fixed
```
JannePeltonen, line 301: > will be valid until it reaches the next TM node.

Here "next TM node" does not mean the next TM node after the one which overrides the priority (since that is the one that looks at the priority) but the one after that. Maybe this should be: "... valid until overridden again by another TM node".

Btw, the term override does not sound quite right to me but I am not sure what I would propose instead.

include/odp/api/spec/traffic_mngr.h
```diff
@@ -331,6 +354,31 @@ typedef struct {
...
+	 * the WFQ groups can only have one fan-in. This also implies that rest of
+	 * the WFQ groups with single fan-in act as static priority fan-in.
+	 * In ODP_TM_PKT_PRIO_MODE_GLOBAL mode, since packet priority is global,
+	 * every node should support as many number of WFQ groups as number of
+	 * priorities available. Hence this field is not applicable.
+	 *
+	 * This value can range from 1..ODP_TM_MAX_PRIORITIES as there can be
+	 * only as many number of WFQ groups as number of priorities.
+	 */
+	uint8_t max_wfq_groups_per_node;
```
JannePeltonen, line 418: I do not think there is a need for this new WFQ group concept. I suppose this could be simply max_num_schedulers_per_node.

> This capability indicates that at a given node, at max only so many wfq groups can have multiple fan-in's. Rest of the WFQ groups can only have one fan-in.

This is a bit inconsistent since max_wfq_groups does not here actually mean the maximum number of WFQ groups but the maximum number of WFQ groups that have more than one fan-in.

> every node should support as many number of WFQ groups as number of

This should be "...every node supports as many WFQ groups as the number of ..."

include/odp/api/spec/traffic_mngr.h
```diff
@@ -443,6 +491,13 @@ typedef struct {
 	 * more of the marking API's. */
 	odp_bool_t marking_colors_needed[ODP_NUM_PACKET_COLORS];
 
+	/** The pkt_prio_mode indicates the packet priority mode, this TM system
+	 * needs to run with. A TM system may or may not support a given mode.
+	 * TM capabilities will indicate support for different modes by the
+	 * platform. Default value of this mode is platform defined.
+	 */
+	odp_tm_pkt_prio_mode_t pkt_prio_mode;
```
JannePeltonen, line 608: > Default value of this mode is platform defined.

Specifying a concrete default value would be more useful to an application. I do not see any benefit in a platform defined default here.

include/odp/api/spec/traffic_mngr.h
```diff
@@ -1464,6 +1464,12 @@ typedef struct {
 	 * have the same single strict priority level and this level must be
 	 * in the range 0..max_priority. */
 	uint8_t priority;
+
+	/** The order_keep when TRUE indicates that when events are enqueued
+	 * into this tm queue while holding an ordered queue synchronization
+	 * context maintain the original event order of the source queue.
+	 */
+	odp_bool_t order_keep;
```
JannePeltonen, line 1616: This could be reworded a bit since it is packets that are enqueued here, not events. How about this;

/*
 * Maintain original packet order of the source queue when enqueuing
 * packets to this queue while holding ordered or atomic queue
 * synchronization context.
 */

odp_boot_t ordered_enqueue;

include/odp/api/spec/traffic_mngr.h
```diff
@@ -1665,6 +1671,20 @@ int odp_tm_queue_disconnect(odp_tm_queue_t tm_queue);
...
+ *
+ * @param tm_queue  Specifies the tm_queue (and indirectly the TM system).
+ * @param packets   Array of packets to enqueue.
+ * @param num       Number of packets to send.
+ * @return          Returns >0 upon success indicating number of packets
+ *                  enqueued, <= 0 upon failure. One of the more common
+ *                  failure reasons is WRED drop.
+ */
+int odp_tm_enq_multi(odp_tm_queue_t tm_queue, const odp_packet_t packets[],
+		     int num);
```
JannePeltonen, line 1830: As commented previously, returning failure because of WRED drop is not going to work well. If the N:th packet would cause WRED drop, then it does not make sense for the function to return failure (indicating that no packets were enqueued). Just consuming the first N-1 packets would not be good either because it would require multiple retries to get a batch of packets processed (enqueued or WRED dropped). Maybe the function should just consume all the packets, not indicating if some of them got WRED dropped.

include/odp/api/spec/traffic_mngr.h
```diff
@@ -379,6 +383,24 @@ typedef struct {
 	 * only as many number of WFQ groups as number of priorities.
 	 */
 	uint8_t max_wfq_groups_per_node;
+
+	/** tm_queue_threshold indicates support for threshold profile on a
+	 * TM queue. When TRUE, users can set/clear/update threshold profile
+	 * on a TM queue. When false clear none of it is supported.
```
JannePeltonen, line 422: Typo: "false clear"

include/odp/api/spec/traffic_mngr.h
```diff
@@ -584,6 +614,41 @@ void odp_tm_egress_init(odp_tm_egress_t *egress);
...
 
+/** Query All TM Capabilities specific to an egress
+ *
+ * The odp_tm_egress_capabilities() function can be used to obtain the complete
+ * set of TM limits supported by this implementation for a given egress.
+ * The reason that this  returns a SET of capabilities and not just one, is
+ * because it is expected  that many HW based implementations may have one set
+ * of limits for the HW and also support a SW TM implementation with a
+ * (presumably larger) different set of limits.  There are also cases where
+ * there could be more than SW implementation (one supporting say tens of
```
JannePeltonen, line 704: Typo: "more than SW" -> "more than one SW"

Paragraph breaks would make this more readable.

There are double spaces inside sentences in a few places.

include/odp/api/spec/traffic_mngr.h
```diff
@@ -584,6 +614,41 @@ void odp_tm_egress_init(odp_tm_egress_t *egress);
...
+ * thousands of tm_queues and a variant supporting tens of millions of
+ * tm_queues). This is needed as platforms might have different constraints
+ * for different egress type and there is no way to know them using
+ * odp_tm_capabilities() provided capabilities. The caller passes in an array
+ * of odp_tm_capabilities_t records  and the number of such records.  Then the
+ * first N of these records will be  filled in by the implementation and the
+ * number N will be returned.  In the event that N is larger than the
+ * capabilities_size, N will still be returned, but only capabilities_size
+ * records will be filled in.
```
JannePeltonen, line 713: > Then the first N of these records will be  filled in by the implementation and the number N will be returned.  In the event that N is larger than the capabilities_size, N will still be returned, but only capabilities_size records will be filled in.

This is not quite right and needs to be fixed (N cannot be both the total number of capabilities and the number of records filled in the array). See e.g. odp_crypto_cipher_capability() for alternate wording.

include/odp/api/spec/traffic_mngr.h
```diff
@@ -379,6 +383,24 @@ typedef struct {
...
+	 * Types of TM queue query are same as query_flags that are passed to
+	 * odp_tm_queue_query(), odp_tm_priority_query() and
+	 * odp_tm_total_query(). When zero, none of the queue query API's are
+	 * supported. When non-zero, only the only supported types of passed
+	 * query_flags are taken into account and corresponding fields updated.
+	 *
+	 * @see ODP_TM_QUERY_PKT_CNT, ODP_TM_QUERY_BYTE_CNT,
+	 * ODP_TM_QUERY_THRESHOLDS.
+	 */
+	uint32_t tm_queue_query_flags;
```
JannePeltonen, line 436: The existing API has this text in the query functions: "The query_flags indicate whether or not packet counts, byte counts or both are being requested.  It is an error to request neither. The implementation may still return both sets of counts regardless of query_flags if the cost of returning all the counts is comparable to the cost of checking the query_flags."

Now that there could be implementations that do not support the counts, this text probably should be changed. e.g. "The query flags indicate which information is begin requested. The implementation may choose to return additional information that was not requested"

include/odp/api/spec/traffic_mngr.h
```diff
@@ -1918,6 +1951,29 @@ typedef struct {
...
+
+/** Get tm_node stats
+ *
+ * The odp_tm_node_stats() function is used to extract runtime statistics
+ * associated with a given tm_node. The stats structure is written
+ * only on success.
+ *
+ * @param      tm_node    Specifies the tm_node to be queried.
+ * @param      clear      1: Clear stats before return.
+ *                        0: Don't clear stats.
```
JannePeltonen, line 1970: Do we really need the clearing? It easily causes synchronization overhead even when the application never actually requests clearing (the piece that writes the statistics has be be prepared for it).

platform/linux-generic/odp_traffic_mngr.c
```diff
@@ -2560,6 +2560,18 @@ int odp_tm_capabilities(odp_tm_capabilities_t capabilities[] ODP_UNUSED,
 	cap_ptr->vlan_marking_supported        = true;
 	cap_ptr->ecn_marking_supported         = true;
 	cap_ptr->drop_prec_marking_supported   = true;
+	cap_ptr->tm_queue_threshold            = true;
```
JannePeltonen, line 2563: Not really related to the changes at hand, but this function incorrectly marks capabilities parameter with ODP_UNUSED. Now that you are changing this function anyway, you could fix it at the same time if is is ok with you,

test/validation/api/traffic_mngr/traffic_mngr.c
```diff
@@ -1367,20 +1363,19 @@ static int destroy_tm_queue(odp_tm_queue_t tm_queue)
...
 }
 
-static tm_node_desc_t *create_tm_node(odp_tm_t        odp_tm,
-				      uint32_t        level,
-				      uint32_t        num_levels,
-				      uint32_t        node_idx,
-				      tm_node_desc_t *parent_node_desc)
+static tm_node_desc_t *
+create_tm_node(odp_tm_t odp_tm, uint32_t level, uint32_t num_levels,
+	       uint32_t node_idx, tm_node_desc_t *parent_node_desc)
```
JannePeltonen, line 1368: Please avoid making this kind of cosmetic formatting changes in the same patch as actual functional changes. I would also argue that this change is not an improvement and should not be done at all.

test/validation/api/traffic_mngr/traffic_mngr.c
```diff
@@ -1367,20 +1363,19 @@ static int destroy_tm_queue(odp_tm_queue_t tm_queue)
...
-	int                   rc;
+	odp_tm_wred_t green_profile, yellow_profile, red_profile;
+	uint32_t node_desc_size, queue_desc_size, priority;
+	odp_tm_node_t tm_node, parent_node;
+	odp_tm_node_params_t node_params;
+	tm_queue_desc_t *queue_desc;
+	char node_name[TM_NAME_LEN];
+	tm_node_desc_t *node_desc;
+	uint16_t queue;
+	int rc;
```
JannePeltonen, line 1378: These lines have a lot of unnecessary formatting and ordering changes that should not be done in this patch. The only real change seems to be the addition of the definition of queue, so adding single line should suffice.

test/validation/api/traffic_mngr/traffic_mngr.c
```diff
@@ -2802,30 +3012,35 @@ static int test_sched_node_priority(const char *shaper_name,
...
 	tm_node_desc_t *node_desc;
-	queue_array_t  *queue_array;
-	pkt_info_t      pkt_info;
-	uint32_t        total_num_queues, max_queues, num_queues, pkt_cnt;
-	uint32_t        pkts_in_order, base_idx, queue_idx, idx, unique_id;
-	uint32_t        pkt_len, base_pkt_len, total_pkt_cnt, pkts_sent;
-	int             priority;
+	pkt_info_t pkt_info;
+	int priority;
+	uint16_t sz;
```
JannePeltonen, line 3024: There are unnecessary formatting and ordering changes here. Only one existing line needs to be modified and two new lines (for queues_set and sz definitions) need to be added.

include/odp/api/spec/traffic_mngr.h
nithind1988: How about ODP_TM_PKT_PRIO_MODE_PER_LVL  ?

include/odp/api/spec/traffic_mngr.h
nithind1988: Will remove the test indicating RED drop would return as failure

include/odp/api/spec/traffic_mngr.h
nithind1988: Ack, will remove.

include/odp/api/spec/traffic_mngr.h
JannePeltonen: This is still open. I think "more than SW" should be "more than one".

include/odp/api/spec/traffic_mngr.h
JannePeltonen: I think _PER_LVL is confusing. Also, the _GLOBAL does not sound quite right to me.

How about _PRIO_PRESERVE and_PRIO_OVERWRITE? Or _PRIO_FIXED and _PRIO_NODE?

include/odp/api/spec/traffic_mngr.h
JannePeltonen: I think it would be good to specify the default value and make it be the "global" mode (or whatever it ends up getting called). This way odp_tm_requirements_init() will use the default and things are more backward compatible.

nithind1988: But what about platforms that cannot support Global mode ? I guess default should be supported.

include/odp/api/spec/traffic_mngr.h
nithind1988: PRESERVE and OVERWRITE sounds fine to me. I'll make the change in v3.

include/odp/api/spec/traffic_mngr.h
JannePeltonen: I would be good if this specified a default value (that is written by ofp_tm_queue_params_init()). I think the default value should be true, since that corresponds to the current behaviour.

The commit message says this about the current implementation: "For Packet IO with TM enabled, there is single mode which is equivalent to Direct mode for enqueuing packets." I do not agree with that since the current linux-gen implementation does ensure the ordering. The API is not explicit about it but the word "enqueue" hints at it. And linux-gen ODP was (is?) a reference implementation of the API.

So, I think this patch is not introducing ordered TM enqueue, but the possibility for unordered enqueue(). The change itself is good and useful.

include/odp/api/spec/traffic_mngr.h
JannePeltonen: I think this is still slightly problematic.

The description says: "[the] rest of the behaviour is identical to that of odp_tm_enq()." And the text for odp_tm_enq() says "One of the more common failure reasons is WRED drop."

I wonder if odp_tm_enq() should no longer be allowed to return failure in WRED drop case, but if that is not changed, it might be good to say in the description of odp_tm_enq_multi() that WRED drops do not cause failures of the call. Also, "number of packets enqueued" could be changed to "number of packets consumed" so that it does not imply that all the packets ended up in the queue.

include/odp/api/spec/traffic_mngr.h
JannePeltonen: This has not been fixed. The quoted text still appears unchanged in the code.

nithind1988: I changed it to for new API. Do you mean for existing API ?
". If number of such records implementation supports is larger than
 * caller requested number, then only caller requested number of records are
 * written and return value is max number of records implementation supports.
 * Caller then may again call with larger number of records to be returned."

include/odp/api/spec/traffic_mngr.h
```diff
@@ -385,6 +408,31 @@ typedef struct {
...
+	 * is required. Rest of the fan-in nodes should have unique priority.
+	 *
+	 * In ODP_TM_PKT_PRIO_MODE_GLOBAL mode, since packet priority is global,
+	 * every node should support as many number of schedulers as number of
+	 * priorities available. Hence this field is not applicable.
+	 *
+	 * This value can range from 1..ODP_TM_MAX_PRIORITIES as there can be
+	 * only as many number of WFQ groups as number of priorities.
+	 */
+	uint8_t max_num_schedulers_per_node;
```
JannePeltonen, line 472: I wonder if we really need this capability field. The number of schedulers needed is no greater than the number of prioriries, so we could just rely on the maximum number of priorities for this instead of the separate capa field. odp_tm_level_capabilities_t already contains max_priority. This new capability field would seem to add value only in the case where there is only one fan-in with certain priorities (so that schedulers is not needed for them) and the max number of schedulers would would be enough only for the priorities that do have more than one fan-in. 

I do not quite understand the rationale for making this capability applicable only for the "override" mode. The "every node should support as many number of schedulers as number of priorities available" could as well be said about the "global" mode as well. In single-level TM system the user must not connect too many sets of queues of the different priorities to a node to not exceed the max scheduler count in the "override" mode. But the same can be done in the "global" mode too.

The text discusses DWRR (as if this capa would concern "only" DWRR schedulers and not other scheduler types such as (W)RR or WFQ), even though DWRR is not mentioned anywhere else in the API. The reference to DWRR should be removed, just "scheduler" should be enough.

Even though checkpatch no longer complains about lines longer than 80 characters but less than 100 characters, I think it would make sense to stick with the same max line length in this description text as the existing API text in this file. It just looks weird to have very different line lengths in the comment sections, sometimes for the same API element. This comment applies to other relevant changes too.

include/odp/api/spec/traffic_mngr.h
```diff
@@ -638,6 +668,41 @@ void odp_tm_egress_init(odp_tm_egress_t *egress);
...
+ *                               in the capabilities array.
+ * @param      egress            Only capabilities compatible with this egress
+ *                               are returned.
+ * @return                       Returns < 0 upon failure. Returns N > 0,
+ *                               where N is the maximum number of different
+ *                               odp_tm_capabilities_t records that the
+ *                               implementations supports. *NOTE* that this
+ *                               number can be > capabilities_size!
+ */
+int odp_tm_egress_capabilities(odp_tm_capabilities_t capabilities[],
```
JannePeltonen, line 735: Now that there is this egress specific capability call, do we need odp_tm_capabilities() anymore? If it stays in the API, what does it return? If it returns something that is not supported by all egresses, then it cannot be relied on and it appears rather useless. So if it stays, maybe it should return the intersection of the capabilities of all egresses (i.e. the capabilities that are supported by all egresses), but even then I wonder if it would be of much use to an application.

include/odp/api/spec/traffic_mngr.h
```diff
@@ -2039,12 +2104,9 @@ typedef struct {
 } odp_tm_query_info_t;
 
 /** The odp_tm_queue_query() function can be used to check a single tm_queue's
- * queue utilization.  The query_flags indicate whether or not packet counts,
- * byte counts or both are being requested.  It is an error to request
- * neither.  The implementation may still return both sets of counts
- * regardless of query_flags if the cost of returning all the counts is
- * comparable to the cost of checking the query_flags. The info structure is
- * written only on success.
+ * queue utilization.  The query flags indicate which information is begin requested.
```
JannePeltonen, line 2161: typo: begin -> being. This appears in two other places too.

include/odp/api/spec/traffic_mngr.h
JannePeltonen: I do not think default needs to be supported. Current implementations will support it, so an unchanged application on top of such implementation would continue to work even after this API change if the default is specified like that. Future ODP implementations may not support it but the application is expected to check the capability field anyway. Also, if the default is not specified, an application cannot rely on the default getting initialized to some sane supported value anyway.

nithind1988: Ok. If default need not be supported, its fine with me. I'll make existing mode as default.

include/odp/api/spec/traffic_mngr.h
nithind1988: >  This new capability field would seem to add value only in the case where there is only one fan-in with certain priorities (so that schedulers is not needed for them) and the max number of schedulers would would be enough only for the priorities that do have more than one fan-in.

Yes, that is the case in our hardware. We only have only one priority with multiple fan-in node and hence DWRR on them.
We support 10 priorities out of which only one priority can be having multiple fan-in's (those nodes overwriting priority value to 'X'). Rest of the 9 priorities can only have one fan-in nodes atmost.

> In single-level TM system the user must not connect too many sets of queues of the different priorities to a node to not exceed the max scheduler count in the "override" mode. But the same can be done in the "global" mode too.
But even in single-level TM system in global mode, multiple fan-ins can have same priorities. Say for example, if implementation supports 3 priorities (0..2) then there could be
SET1: Priority 0: (Fanin0, 1, 2.,..)
SET2: Priority 1: (fanin 3, 4, 5, 6, 7)
SET 3: Priority 2 : (fanin 9, 10, 11)

Now there are 12 queues connected to same Parent node with max three priorities. Now parent node has to run schedulers on each of the SET separately. In our case there can only one such SET where scheduler(DWRR or WRR or RR) is needed.
Rest of the SET's can have only one fan-in => no need to schedule.,

> The text discusses DWRR (as if this capa would concern "only" DWRR schedulers and not other scheduler types such as (W)RR or WFQ), even though DWRR is not mentioned anywhere else in the API. The reference to DWRR should be removed, just "scheduler" should be enough.
Will change it to scheduler.

>Even though checkpatch no longer complains about lines longer than 80 characters but less than 100 characters, I think it would make sense to stick with the same max line length in this description text as the existing API text in this file. It just looks weird to have very different line lengths in the comment sections, sometimes for the same API element. This comment applies to other relevant changes too.

Ack.

include/odp/api/spec/traffic_mngr.h
nithind1988: We are fine to deprecate older call.

include/odp/api/spec/traffic_mngr.h
JannePeltonen: I got confused and looked at odp_tm_capabilities() which contains the above text and which this patch set has not changes. As that too is a bit inaccurate, it could also be fixed.

include/odp/api/spec/traffic_mngr.h
JannePeltonen: > Yes, that is the case in our hardware. We can only have only one priority with multiple fan-in node and hence DWRR on them.

Ok. let's have this capa flag then.

> But even in single-level TM system in global mode, multiple fan-ins can have same priorities

Yes. And the exact same thing in non-global mode. I do not see why the capa should not apply to the global mode too. Let's say, there is a HW like yours (with limited number of schedulers) but without the priority overwriting. Then the capa would tell how I am allowed to configure things in global mode. IOW, an application can avoid TM config that exceeds the number of schedulers needed regardless of the priority mode. With single level TM the way to ensure that is the same but in multi-level config it is a but different. But to me, the capability itself does not seem mode specific at all.

nithind1988: Ok, I'll remove the applicability string and for linux generic, set it as same as number of priorities supported.

include/odp/api/spec/traffic_mngr.h
nithind1988: Ok. I'll change just odp_tm_enq_multi() to say that WRED drops do not cause failures of the call.

include/odp/api/spec/traffic_mngr.h
nithind1988: Ack.

nithind1988:
Addressed all comments in v3

include/odp/api/spec/traffic_mngr.h
JannePeltonen: I think the behaviour of the non-egress specific capability function should be clarified (i.e. it returns capabilities that are valid for all egresses) even if the function gets deprecated.

JannePeltonen:
General comment:

It would be good to have separate things in separate patches in such a way that for separate API patches there are separate implementation patches (and preferably separate validation test patches).

This way it would be easier to see if the implementation for a particular API patch is present and complete. Now the implementations of unordered enqueue and odp_tm_enq_multi() are missing.

include/odp/api/spec/traffic_mngr.h
```diff
@@ -385,6 +410,29 @@ typedef struct {
...
+	 * This capability indicates that at a given node, max schedulers
+	 * available. This implies at a given node there cannot be more
+	 * than max_schedulers_per_node priorities that have packets from
+	 * multiple fan-in nodes and hence need scheduling.
+	 * Rest of the fan-in nodes should have unique priority packets.
+	 *
+	 * This value can range from 0..ODP_TM_MAX_PRIORITIES as there can be
+	 * only as one scheduler is assigned per static priority.
+	 */
+	uint8_t max_schedulers_per_node;
```
JannePeltonen, line 472: The description here is not very clear and has grammar problems. I propose the following text instead:

```
Maximum number of schedulers supported by a TM node at any level.

A TM node contains a WFQ/WRR scheduler for each packet priority level
for which the node has more than one possible input. TM topology and
priority configuration must be made so that the resulting number of
WFQ/WRR schedulers does not exceed this capability in any TM node.

The value can vary between 0 and ODP_TM_MAX_PRIORITIES.
```

include/odp/api/spec/traffic_mngr.h
```diff
@@ -497,6 +545,13 @@ typedef struct {
 	 * more of the marking API's. */
 	odp_bool_t marking_colors_needed[ODP_NUM_PACKET_COLORS];
 
+	/** The pkt_prio_mode indicates the packet priority mode, this TM system
+	 * needs to run with. A TM system may or may not support a given mode.
+	 * TM capabilities will indicate support for different modes by the
+	 * platform. Default value is ODP_TM_PKT_PRIO_MODE_PRESERVE.
+	 */
```
JannePeltonen, line 615: I would like to propose the following shorter text:

Packet priority mode. TM capabilities indicate which modes are supported.
The default value is ODP_TM_PKT_PRIO_MODE_PRESERVE.

include/odp/api/spec/traffic_mngr.h
```diff
@@ -1351,6 +1406,15 @@ typedef struct {
...
 	uint8_t level;
+
+	/** The strict priority level assigned to packets going through this
+	 * fan-in node. In other words all packets going through node
+	 * will be assigned this priority overwriting previous priority.
+	 * in the range 0..max_priority. This field is only valid if packet
+	 * priority mode is ODP_TM_PKT_PRIO_MODE_OVERWRITE.
+	 * @see odp_tm_pkt_prio_mode_t
+	 */
+	uint8_t priority;
```
JannePeltonen, line 1519: As far as I can see, the max_priority configuration does not restrict the priority value but the number of distinct priorities a node has to be able to handle. So the allowed range should be from zero to ODP_TM_MAX_PRIORITIES - 1.

I would like to propose the following text:

```
New strict priority level assigned to packets going through this node
when packet priority mode is ODP_TM_PKT_PRIO_MODE_OVERWRITE. In other
packet priority modes this field is ignored. The new priority does not
affect packet processing in this node but in its destination node.

The value must be in the range 0..ODP_TM_MAX_PRIORITIES-1. Additionally,
the total number of possible priorities seen by the destination node
must not exceed the max priority configured for the destination node.

@see odp_tm_pkt_prio_mode_t, odp_tm_level_requirements_t::max_priority
```

include/odp/api/spec/traffic_mngr.h
```diff
@@ -1783,6 +1789,21 @@ int odp_tm_queue_disconnect(odp_tm_queue_t tm_queue);
  */
 int odp_tm_enq(odp_tm_queue_t tm_queue, odp_packet_t pkt);
 
+/** The odp_tm_enq_multi() function is used to add packets to a given TM system.
+ * This API takes multiple packets to enqueue unlike odp_tm_enq() but
+ * rest of the behaviour is identical to that of odp_tm_enq().
+ * WRED drops donot cause failure of the call.
```
JannePeltonen, line 1897: I would like to propose this wording:

The odp_tm_enq_multi() function is used to add packets to a given TM system. This function enqueues multiple packets but is otherwise similar to odp_tm_enq(). Packets dropped by WRED or other queue management action do not cause this function to return a failure. Such packets get consumed just like the packets that are not dropped.

include/odp/api/spec/traffic_mngr.h
```diff
@@ -189,6 +189,30 @@ extern "C" {
...
+	/** Green octets that are sent through this tm node */
+	ODP_TM_NODE_STATS_GREEN_OCTETS,
+	/** Yellow packets that are sent through this tm node */
+	ODP_TM_NODE_STATS_YELLOW_PKTS,
+	/** Yellow octets that are sent through this tm node */
+	ODP_TM_NODE_STATS_YELLOW_OCTETS,
+	/** Red packets that are sent through this tm node */
+	ODP_TM_NODE_STATS_RED_PKTS,
+	/** Red octets that are sent through this tm node */
+	ODP_TM_NODE_STATS_RED_OCTETS,
```
JannePeltonen, line 211: Just to check if we have the same understanding: The color referred to here is the odp_packet_color(pkt) of the packet as it entered the TM system and the color never changes within the TM system. Right?

include/odp/api/spec/traffic_mngr.h
```diff
@@ -189,6 +189,30 @@ extern "C" {
...
+	/** Node stats max */
+	ODP_TM_NODE_STATS_MAX,
+} odp_tm_node_stats_type_t;
```
JannePeltonen, line 214: As discussed on tuesday, this way of defining the stats is different from other stats and the stats proposed in PR 1305. I actually like this way since it can make statistics handling in the application more streamlined, but we should not have two different ways of doing statistics within TM.

So, this PR and PR 1305 should converge to the same overall scheme of providing statistics and maybe also the actual counters should be the same in queue level (PR 1305) and in node level (this PR).

include/odp/api/spec/traffic_mngr.h
```diff
@@ -252,6 +276,15 @@ typedef struct {
...
 	odp_bool_t tm_node_threshold;
+
+	/** tm_node_stats indicates the statistics supported by nodes at this level.
+	 * When True, that stat counter is supported, while False indicates it is not
+	 * supported.
+	 *
+	 * @see odp_tm_node_stats_type_t
+	 * @see odp_tm_node_stats()
+	 */
+	odp_bool_t tm_node_stats[ODP_TM_NODE_STATS_MAX];
```
JannePeltonen, line 287: I think that in addition to having capabilities, maybe it should be configurable (in the per-level requirements or node parameters) which counters, if any, need to be counted.

platform/linux-generic/odp_traffic_mngr.c
```diff
@@ -3960,6 +4001,8 @@ int odp_tm_node_context_set(odp_tm_node_t tm_node, void *user_context)
 void odp_tm_queue_params_init(odp_tm_queue_params_t *params)
 {
 	memset(params, 0, sizeof(odp_tm_queue_params_t));
+
+	params->ordered_enqueue = true;
```
JannePeltonen, line 4005: This is only part of the implementation. Now that support for unordered enqueue was added, the implementation should look at the ordered_enqueue parameter and skip the ordering step (i.e. the _odp_sched_fn->order_lock() and the unlock call) in tm_enqueue().

include/odp/api/spec/traffic_mngr.h
```diff
@@ -250,6 +250,31 @@ typedef struct {
...
+ * function only on single mode specified at time of odp_tm_create().
+ */
+typedef enum odp_tm_pkt_prio_mode {
+	/** Indicates Packet priority preserve mode. In this mode, a packet gets
+	 * its priority based on a TM queue it gets enqueued to and then it
+	 * carries the same priority along with it as long as it is in the TM
+	 * system. At every TM node in the topology, that specific pkt is
+	 * scheduled as per that priority.
+	 */
+	ODP_TM_PKT_PRIO_MODE_PRESERVE,
```
JannePeltonen, line 302: It would be good to have an empty line before the description of the next field starts (here and elsewhere).

include/odp/api/spec/traffic_mngr.h
```diff
@@ -385,6 +410,29 @@ typedef struct {
 	 * the parameters of the threshold profile of any TM node or TM queue.
 	 */
 	odp_bool_t dynamic_threshold_update;
+
+	/** Indicates the packet priority modes supported by a TM systems on a
+	 * platform. A platform can support multiple packet priority modes. The
+	 * actual mode, a TM system runs with is defined by
```
JannePeltonen, line 453: I think the comma should be removed.

test/validation/api/traffic_mngr/traffic_mngr.c
```diff
@@ -1951,7 +1991,12 @@ static int destroy_all_shaper_profiles(void)
 	uint32_t        idx;
 	int             rc;
 
-	for (idx = 0; idx < NUM_SHAPER_PROFILES; idx++) {
+	if (num_shaper_profiles >= MAX_SHAPER_PROFILES) {
```
JannePeltonen, line 2040: These checks that have been added in various destroy functions look a bit out-of-place. The destroy function could still destroy the objects it can, so this check does not seem to be needed for the destroy function. This looks more like some internal consistency checking that would be better done at creation/allocation time than way later here in the destroy functions.

platform/linux-generic/odp_traffic_mngr.c
nithind1988: Ok. I can try making those two locks conditional on this flag of unordered enqueue. Is that sufficient ?
If not, then can you or matias do the change as I don't have much internal understanding of linux-generic.

Another approach could be that, since maintaining ordering even when ordered_enqueue is false is not functionally incorrect but just has performance implications, we can leave the implementation as it is and you guys can optimize it later ?

nithind1988:
> It would be good to have separate things in separate patches in such a way that for separate API patches there are separate implementation patches (and preferably separate validation test patches).
> 
> This way it would be easier to see if the implementation for a particular API patch is present and complete. Now the implementations of unordered enqueue and odp_tm_enq_multi() are missing.

Ok, I'll split the linux-gen implementation to multiple patches. 
With regards to validation test suite, I have already split as you suggested earlier to two parts first one being addition of 5 level test suite and second one being using capabilities to conditionally disable each test case.
 Splitting second validation patch further will make them too small I believe.

include/odp/api/spec/traffic_mngr.h
nithind1988: Moved this patch to other pull request. Will provide new version there.

include/odp/api/spec/traffic_mngr.h
nithind1988: Ok, In our HW, it will be a no-op as the stats are counted in HW and no option to enable/disable it. We are fine to have such capability if needed.

nithind1988:
Addressed all comments and moved node stats API change to new pull request

include/odp/api/spec/traffic_mngr.h
nithind1988: We don't currently support pre-colored packets on Tx. And all our shapers are based on RFC2698 like meter were packet color is modified and action of drop or forward to next shaper is decided and hence packet color internally changes as it progresses.

JannePeltonen: It seems that is not according to the API. As far as I can see, the API does not say that odp_packet_color() of a packet can change inside a TM system, and looking at the linux-gen implementation, the color indeed does not change (AFAICS) and e.g. egress VLAN marking is done based on the original, unchanged packet color that the packet had when it entered the TM system.

I am not sure if this makes a lot of sense, but that is how both the API spec and the linux-gen implementation look like now. Maybe API changes are needed.

nithind1988: We also currently don't support any marking feature in our implementation such as VLAN, IP ECN or IP DSCP. 
One thing is spec doesn't and might not be able to say how single rate or dual rate shaping is done in a particular HW. 
And RFC2698 is one method of doing so and then coloring a packet based on the result.

However unless packet marking is supported which updates the drop precedence based on final color, I guess coloring packets is not of much use.

>egress VLAN marking is done based on the original, unchanged packet color that the packet had when it entered the TM system.

Below is the text from doc/users-guide/users-guide-tm.adoc. which talks about final packet color.
I couldn't find anything about odp_packet_color() implication in traffic_mngr.h

> ==== Packet Marking
> 
> The Packet Marking API is used to mark the packet based upon the final packet
> color assigned to it when it reaches the egress node.
> This is an optional feature and if available on the platform is used to reflect
> the packet color on IPv4/IPv6 DiffServ filed in accordance with https://www.ietf.org/rfc/rfc2474.txt[RFC 2474].
...
...

JannePeltonen:
It is good that the statistics are now in a separate PR.

This PR still contains a few unrelated API changes:
- Unordered enqueue & multi packet enqueue
- More capability bits and egress specific capabilities
- Overwrite packet priority mode

Each of these have their respective API changes, implementation changes and validation test changes.

If the above topics were in their own PRs, then they would not necessarily have to wait that all the patches become ready and could potentially get merged sooner.

It is ok for me to keep the changes in this PR, but even then I think the changes should not be interleaved. So, the first three commits could be
- Unordered enqueue & multi packet enqueue API
- Linux-gen implementation of it
- Validation test of it *)

Then the next set of commits could add the new capabilities, their implementation and their validation test changes. And the final set of commits could add the priority mode thing.

This way the resulting change history would be much more clean as closely related changes would be next to each other and not mixed with unrelated changes.

*) Validation tests for the new API elements are missing. The validation tests should at least refer to the new field and new function so that the tests would fail to build if the implementation is missing.

include/odp/api/spec/traffic_mngr.h
```diff
@@ -633,6 +665,41 @@ void odp_tm_egress_init(odp_tm_egress_t *egress);
...
+ *                               in the capabilities array.
+ * @param      egress            Only capabilities compatible with this egress
+ *                               are returned.
+ * @return                       Returns < 0 upon failure. Returns N > 0,
+ *                               where N is the maximum number of different
+ *                               odp_tm_capabilities_t records that the
+ *                               implementations supports. *NOTE* that this
+ *                               number can be > capabilities_size!
+ */
+int odp_tm_egress_capabilities(odp_tm_capabilities_t capabilities[],
```
JannePeltonen, line 702: We discussed with @MatiasElo  and @psavol  that in order to simplify the TM API a little bit, we could remove the
possibility of TM capability functions (the old and and this new one) returning more than one capability structure. Currently we are not aware of any ODP implementation that would benefit from exposing multiple capability structures.

What would you think about such a change?

include/odp/api/spec/traffic_mngr.h
```diff
@@ -1584,6 +1584,12 @@ typedef struct {
 	 * have the same single strict priority level and this level must be
 	 * in the range 0..max_priority. */
 	uint8_t priority;
+
```
JannePeltonen, line 1657: The subject line of the commit message of this patch could be changed to be more descriptive so that it mentions also the addition of the multi packet enqueue function: 
Subject line now: api: tm: add option to do an unordered enqueue to TM queue
This would be better: api: tm: add unordered enqueue and multi-packet enqueue

include/odp/api/spec/traffic_mngr.h
```diff
@@ -611,14 +641,16 @@ void odp_tm_egress_init(odp_tm_egress_t *egress);
  * a SET of capabilities and not just one, is because it is expected that
  * many HW based implementations may have one set of limits for the HW and
  * also support a SW TM implementation with a (presumably larger) different
- * set of limits.  There are also cases where there could be more than
+ * set of limits.  There are also cases where there could be more than one
  * SW implementation (one supporting say tens of thousands of tm_queues and
- * a variant supporting tens of millions of tm_queues).
+ * a variant supporting tens of millions of tm_queues). It returns capabilities
+ * that are valid for all egresses.
```
JannePeltonen, line 650: > "It returns capabilities that are valid for all egresses."

There is a faulty pronoun reference here (i.e. it is not clear at all what "it" refers to). The sentence is also a bit hidden in there. I suggest removing it and rephrasing the first sentence (the one that starts "The odp_tm_capabilities()") like this:

This function returns the complete set of TM capabilities for all egresses.

You could also rephrase the corresponding sentence of the description of odp_tm_egress_capabilities() in the same way.

include/odp/api/spec/traffic_mngr.h
nithind1988: Sounds fine to me. Our platform also doesn't support multiple capabilities per pktio. There is only single capability per pktio.

include/odp/api/spec/traffic_mngr.h
```diff
@@ -468,6 +495,23 @@ typedef struct {
 	 * ODP_TM_QUERY_THRESHOLDS.
 	 */
 	uint32_t tm_queue_query_flags;
+
+	/** Indicates the packet priority modes supported by a TM systems on a
```
JannePeltonen, line 499: typo: "a TM systems" -> "TM systems"

test/validation/api/traffic_mngr/traffic_mngr.c
```diff
@@ -1679,6 +1685,17 @@ set_reqs_based_on_capas(odp_tm_requirements_t *req)
 
 	if (tm_capabilities.tm_queue_shaper_supported)
 		req->tm_queue_shaper_needed = true;
+
+	/* Since packet priority mode effects only scheduler tests
+	 * and in our scheduler tests, we only test with set queues under
+	 * same parent node, we are fine with any packet priority mode.
+	 * This is because we are concerned with packet priority obeyed
+	 * at immediate parent node only.
```
JannePeltonen, line 1718: I do not understand the explanation in the comment. The priority mode does not matter if the test is testing a scheduler in TM nodes that are directly connected to TM queues and not in TM nodes fed by other TM nodes that could change the priority on the way.

However, it looks to me that the test is testing scheduler in a TM node fed by other TM nodes. "node_1_3" is the one being tested, the node has child nodes which in turn have queues. The tests use one queue of each child node of node_1_3. If priority mode is overwrite, then the child nodes overwrite packet priority before packets reach the node being tested.

test/validation/api/traffic_mngr/traffic_mngr.c
```diff
@@ -1679,6 +1685,17 @@ set_reqs_based_on_capas(odp_tm_requirements_t *req)
...
+	 */
+	req->pkt_prio_mode = ODP_TM_PKT_PRIO_MODE_PRESERVE;
+	if (!tm_capabilities.pkt_prio_modes[ODP_TM_PKT_PRIO_MODE_PRESERVE])
+		req->pkt_prio_mode = ODP_TM_PKT_PRIO_MODE_OVERWRITE;
```
JannePeltonen, line 1722: Node priority is left unset. The code here enables the overwrite mode which requires that odp_node_param_t::priority is set to a valid value at TM node creation. The priority parameter is left unset (and is also not guaranteed to be initialized by odp_tm_node_params_init()).

So, create_tm_node() should initialize the priority parameter to a valid value. In addition to fixing the bug it makes validation test build fail if the parameter is entirely missing.

test/validation/api/traffic_mngr/traffic_mngr.c
```diff
@@ -443,6 +444,11 @@ static int test_overall_capabilities(void)
 				return -1;
 			}
 		}
+
```
JannePeltonen, line 437: Checking of max_schedulers_per_node capability is not added by this patch or by any other patch, even though the TM configuration used in the validation tests seems to require that max_schedulers_per_node is at least NUM_PRIORITIES (which is now 4).

Maybe it would be good to separate the addition of the max_schedulers_per_node capability into its own API patch and its own implementation and validation patches. Then the API change would not be hidden in an unrelated commit and it would be obvious that validation test changes are missing.

test/validation/api/traffic_mngr/traffic_mngr.c
```diff
@@ -22,18 +22,13 @@
 
 #define MAX_NUM_IFACES           2
 #define MAX_TM_SYSTEMS           3
-#define NUM_LEVELS               3
```
JannePeltonen, line 25: I have not yet properly reviewed this last patch (add five level test suite). Since the patch seems unrelated to the API changes in this PR, you might be able to get the API changes merged quicker if this patch patch was moved to a separate PR so that the API change related patches would not have to wait until it is ready.

Could you also split the refactoring and the actual addition of new tests to separate patches. That would make it easier to see what code change is related to what and it would help in case later someone has to investigate a problem related either to the new test cases or to the refactoring.

test/validation/api/traffic_mngr/traffic_mngr.c
nithind1988: Added it in v7 to check if max_schedulers_per_node  <= ODP_TM_MAX_PRIORITIES.

nithind1988: The reason, I didn't add check for it to be NUM_PRIORITIES is because, in packet priority mode overwrite, since current tests are having node priority 0, it works in platforms that support even max_schedulers per node = 1.

test/validation/api/traffic_mngr/traffic_mngr.c
nithind1988: In traffic_mngr_test_scheduler(), only test that is enabled is test_sched_queue_priority() with node "node_1_1_3".
So the test case is currently testing priority b/w child nodes of same parent node.  When the test cases are enabled, we can split it to not run on overwrite mode ?

JannePeltonen: I missed the early return statement. It is correct that the potentially problematic scheduler tests are not run.

I think the comment is still a bit unclear. How about this:
```
        /* We can use any packet priority mode since it does not affect
         * our tests. Our scheduler test tests scheduling only in a node
         * directly connected to TM queues and such nodes see the original
         * packet priority before it could have been overwritten by any node.
         */
```

test/validation/api/traffic_mngr/traffic_mngr.c
JannePeltonen: > Added it in v7 to check if max_schedulers_per_node <= ODP_TM_MAX_PRIORITIES.

I do not see why we want to check that max_scheduelrs_per_node capability is *less* than some value. My point is that the tests require certain number of schedulers per node and should check the capability against that number.

> it works in platforms that support even max_schedulers per node = 1.

In some platforms max_schedulers_per_node could be zero. So you should check that the capability is at least one before configuring the TM topology (the API says that the TM config must be such that the scheduler limit is not exceeded, even if you are not going to send single packet through the TM system).

TM priority and topology configuration in the test setup is such that NUM_QUEUES_PER_NODE (which now equals NUM_PRIORITIES) queues of distinct priorities are connected to the level 2 nodes and multiple level 2 node connect to the same level 1 node. Thus in the "preserve" packet priority mode the level 1 nodes need NUM_QUEUES_PER_NODE schedulers. Creating such a config is against the API rules if max_schedulers_per_node is not at least NUM_QUEUES_PER_NODE.

JannePeltonen (approved):

nithind1988:
Added reviewed by tag of Janne

PR #1239, opened 2021-03-30
[PATCH v4] api: timer tick info

psavol:
Add clock source enumeration options. Add tick info which can document precisely tick period in nsec, hz or input clock cycles.

### DIFF
--- a/include/odp/api/spec/timer.h
+++ b/include/odp/api/spec/timer.h
@@ -375,6 +375,54 @@ uint64_t odp_timer_ns_to_tick(odp_timer_pool_t timer_pool, uint64_t ns);
 uint64_t odp_timer_current_tick(odp_timer_pool_t timer_pool);
 
 /**
+ * Timer tick information
+ */
+typedef struct odp_timer_tick_info_t {
+	/**
+	 * Timer tick frequency in hertz
+	 *
+	 * Timer tick frequency expressed as a fractional number. The integer part contains
+	 * full hertz. The fraction part (numerator / denominator) contains parts of
+	 * a hertz to be added with the integer.
+	 *
+	 * For example, a timer tick frequency of 333 333 and 1/3 Hz could be presented with
+	 * these values: integer = 333 333, numer = 1, denom = 3. Implementation may choose numer
+	 * and denom values freely.
+	 */
+	odp_fract_u64_t freq;
+
+	/**
+	 * One timer tick in nanoseconds
+	 *
+	 * Nanoseconds per tick is expressed as a fractional number. The integer part contains
+	 * full nanoseconds. The fraction part (numerator / denominator) contains parts of
+	 * a nanosecond to be added with the integer.
+	 *
+	 * For example, a timer tick period of 3.125 nanoseconds (320MHz) could be presented with
+	 * these values: integer = 3, numer = 125 000 000, denom = 1 000 000 000. Implementation
+	 * may choose numer and denom values freely.
+	 */
+	odp_fract_u64_t nsec;
+
+	/**
+	 * One timer tick in source clock cycles
+	 *
+	 * The clock cycle count is expressed as a fractional number. The integer part contains
+	 * full clock cycles. The fraction part (numerator / denominator) contains parts of
+	 * a clock cycle to be added with the integer.
+	 *
+	 * For example, a timer tick period of 42 and 1/3 source clock cycles could be presented
+	 * with these values: integer = 42, numer = 1, denom = 3. Implementation may choose numer
+	 * and denom values freely.
+	 *
+	 * The value is zero, when there is no direct connection between tick and the source
+	 * clock signal.
+	 */
+	odp_fract_u64_t clk_cycle;
+
+} odp_timer_tick_info_t;
+
+/**
  * ODP timer pool information and configuration
  */
 typedef struct {
@@ -389,6 +437,9 @@ typedef struct {
 
 	/** Name of timer pool */
 	const char *name;
+
+	/** Timer pool tick information */
+	odp_timer_tick_info_t tick_info;
 
 } odp_timer_pool_info_t;
 
--- a/include/odp/api/spec/std_types.h
+++ b/include/odp/api/spec/std_types.h
@@ -1,5 +1,6 @@
 /* Copyright (c) 2013-2018, Linaro Limited
  * Copyright (c) 2021, ARM Limited
+ * Copyright (c) 2021, Nokia
  * All rights reserved.
  *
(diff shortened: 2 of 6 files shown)


jerinjacobk (approved):

include/odp/api/spec/timer.h
```diff
@@ -357,6 +357,54 @@ uint64_t odp_timer_ns_to_tick(odp_timer_pool_t timer_pool, uint64_t ns);
...
+	 *
+	 * Timer tick frequency expressed as a fractional number. The integer part contains
+	 * full hertz. The fraction part (numerator / denominator) contains parts of
+	 * a hertz to be added with the integer.
+	 *
+	 * For example, a timer tick frequency of 333 333 and 1/3 Hz could be presented with
+	 * these values: integer = 333 333, numer = 1, denom = 3. Implementation may choose numer
+	 * and denom values freely.
+	 */
+	odp_fract_u64_t freq;
```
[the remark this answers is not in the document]
psavol: Fraction allows to express exact relation of tick and input clock (or time or frequency). E.g. if timer HW divides input clock by 256 and then multiplies it by 3, a tick can be expressed as 256/3 input clock cycles exactly (vs 85.3333333).

include/odp/api/spec/timer.h
```diff
@@ -357,6 +357,54 @@ uint64_t odp_timer_ns_to_tick(odp_timer_pool_t timer_pool, uint64_t ns);
...
+	 * The clock cycle count is expressed as a fractional number. The integer part contains
+	 * full clock cycles. The fraction part (numerator / denominator) contains parts of
+	 * a clock cycle to be added with the integer.
+	 *
+	 * For example, a timer tick period of 42 and 1/3 source clock cycles could be presented
+	 * with these values: integer = 42, numer = 1, denom = 3. Implementation may choose numer
+	 * and denom values freely.
+	 *
+	 * The value is zero, when there is no direct connection between tick and the source
+	 * clock signal.
```
JereLeppanen, line 402: Is this an example or the only acceptable reason? When exactly is an implementation allowed to set this to zero?

psavol: Basically, when there is no fixed ratio between source clock cycle and tick cycle, or there is no src clk signal defined. E.g. linux-gen uses nsec values as tick, and checks time with posix timer, or polls system time / CPU register. 

May be the the sentence could continue: " ... or when there is not signal defined."

include/odp/api/spec/timer.h
```diff
@@ -37,15 +37,32 @@ extern "C" {
 
 /**
  * Clock sources for timers in timer pool.
+ *
+ * ODP_CLOCK_CPU is the default clock source and it is supported always. It is implementation
+ * defined which other clock sources are supported. See implementation documentation for
```
yuri-tolstov, line 42: The notion of external clock, probably, needs to be defined more explicitly. For example the clock may come from the GPIO pin or some SOC internal source (PLL). Are they both external in this case?

include/odp/api/spec/timer.h
```diff
@@ -357,6 +357,54 @@ uint64_t odp_timer_ns_to_tick(odp_timer_pool_t timer_pool, uint64_t ns);
...
 uint64_t odp_timer_current_tick(odp_timer_pool_t timer_pool);
 
+/**
+ * Timer tick information
+ */
+typedef struct odp_timer_tick_info_t {
+	/**
+	 * Timer tick frequency in hertz
+	 *
+	 * Timer tick frequency expressed as a fractional number. The integer part contains
```
yuri-tolstov, line 367: I am still not clear why fractional description should be used and not floating-point. The number should show the exact frequency (or period) of clock, which feeds the timer hardware, and floating-point can express that. If the role of numerator and denominator is to show the relationship between input and timer clocks, then two numbers may not be enough, because, in general, such relations may be more complex.

psavol: This is not frequency of the clock signal, but frequency of the timer tick (in the API). For example, if tick_info tells that timer tick is 100MHz, application can request timeout every 100 ticks and get exact 1MHz timeout frequency. 

Fractional presentation is exact on the relation between nsec and hz, e.g. when tick period is 1200nsec the  frequency is 833 333 1/3 Hz. It's up to implementation if it rounds that freq to 6 digit granularity 833 333 333333/1000000 (833333.333333), or if it outputs the exact ratio (1/3). With float in API, implementation could not output the exact ratio.

Implementation can express relation between tick (in API) and source clock signal accurately, as long as it a series of multipliers and dividers: (2 * 128) / 3 = 85 1/3

include/odp/api/spec/timer.h
```diff
@@ -37,15 +37,32 @@ extern "C" {
...
+ * possible clock sources and their enumeration to these values.
  */
 typedef enum {
-	/** Use CPU clock as clock source for timers */
+	/** CPU clock as clock source for timers (default) */
 	ODP_CLOCK_CPU,
-	/** Use external clock as clock source for timers */
-	ODP_CLOCK_EXT
-	/* Platform dependent which other clock sources exist */
+
```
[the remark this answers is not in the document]
psavol: Timer API functions have this enumeration as input parameter. It is backwards compatible to add new  enumerations vs e.g. introduce a new parameter (int ext_clock_index).

Implementation manual could tell that e.g. EXT_0 is GPIO X, EXT_1 is driven by PTP clock, etc. We can add (now or) later additional clock sources per type if that is needed (ODP_CLOCK_GPIO_0, ODP_CLOCK_PTP, ...). But even then these EXT_0 ... _3 could remain.

Clock source enumeration needs some fixed format, so that validation test can check also those.

include/odp/api/spec/timer.h
psavol: Basically, ODP_CLOCK_CPU is the default (may be internal or external), EXT_0 ... 3  are implementation defined where those connect.

We could also use term ODP_CLOCK_SRC_0 ... _3, and leave EXT if it's confusing. Timer API does not have many references to EXT, so the naming could be changed.

include/odp/api/spec/timer.h
JereLeppanen: It's probably fine even without that addition. I was just thinking that the reason why an implementation has set this to zero is not really validatable by the caller. But I suppose that's not a problem.

psavol:
v2: rebased

JereLeppanen (approved):

jerinjacobk (approved):

jerinjacobk (approved):

PR #1242, opened 2021-03-30
[PATCH v1] api: add print functions with callback parameter

JereLeppanen:
```
Add new print API functions, which are the same as the existing print
functions, except for an added callback parameter. This callback
function is called to do the actual output.

```

### DIFF
--- a/include/odp/api/spec/errno.h
+++ b/include/odp/api/spec/errno.h
@@ -12,6 +12,9 @@
 
 #ifndef ODP_API_SPEC_ERRNO_H_
 #define ODP_API_SPEC_ERRNO_H_
+
+#include <odp/api/init.h>
+
 #include <odp/visibility_begin.h>
 
 #ifdef __cplusplus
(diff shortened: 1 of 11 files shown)


JereLeppanen:
I've copied documentation from the existing print functions, but would it better to remove the descriptions and use `@see` references instead?

Is odp_log_func_t enough, or might some context pointer be needed now or in the future?

include/odp/api/spec/errno.h
```diff
@@ -65,6 +68,17 @@ void odp_errno_zero(void);
 */
 void odp_errno_print(const char *str);
 
+/**
+* Print ODP errno
+*
+* Interprets the value of ODP errno as an error message, and prints it,
+* optionally preceding it with the custom message specified in str.
```
matiaselo, line 75: Instead of duplicating the documentation, perhaps it would make sense to write something like this:
`Like odp_errno_print(), but uses a callback function for output.`

Same goes for all the alt functions.

JereLeppanen: Should it actually be "Provides the same information and format as odp_errno_print(), ...", or is that an unnecessary requirement?

JereLeppanen:
PR #1252 selected instead of this one.

PR #1243, opened 2021-03-31
github_ci: fix coverity dir

jlinkes:
Coverity expects that the directory where it's storing the results of
the build is named "cov-int". Rename the directory to address this.

Signed-off-by: Juraj Linkeš <juraj.linkes@pantheon.tech>


matiaselo (approved):
Please change the commit subject line to e.g. `scripts: coverity: fix directory name` and add my reviewed-by tag. Otherwise looking good.

PR #1247, opened 2021-04-01
[PATCH v3] linux-gen: detect cpu cache line size during configure

matiaselo:
Detect CPU cache line size automatically on aarch64 systems when abi-compatibility is disabled. Automatic cache line size detection is disabled when cross-compiling.

V2:
- Validation test fix

### DIFF
--- a/platform/linux-generic/m4/odp_cpu.m4
+++ b/platform/linux-generic/m4/odp_cpu.m4
@@ -1,0 +1,35 @@
+##########################################################################
+# Set ODP_CACHE_LINE_SIZE define
+##########################################################################
+# Currently used only for aarch64
+if test "${ARCH_DIR}" = "aarch64"; then
+	cache_line_size=64
+	# Use default cache size if cross-compiling
+	if test $build = $host; then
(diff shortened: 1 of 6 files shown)


platform/linux-generic/m4/odp_cpu.m4
```diff
@@ -0,0 +1,35 @@
...
+		done < /proc/cpuinfo
+
+		# Cavium
+		if test "$cpu_implementer" == "0x43"; then
+			# ThunderX2 (0x0af) 64B, others 128B
+			if test "$cpu_part" == "0x0af"; then
+				cache_line_size=64;
+			else
+				cache_line_size=128;
+			fi
```
matiaselo, line 30: Hi @jerinjacobk, is this assumption correct? There is a similar check in VPP build.

jerinjacobk: Yes. 

Cpu part is 0xaf for ThunderX2

Not sure it comes as 0x0af or 0xaf.

psavol (approved):

PR #1248, opened 2021-04-07
[PATCH v2] linux-gen: fix Coverity warnings

psavol:
Fix 6 warnings reported by Coverity

### DIFF
--- a/platform/linux-generic/odp_init.c
+++ b/platform/linux-generic/odp_init.c
@@ -173,9 +173,6 @@ static int term_global(enum init_stage stage)
 		}
 		/* Fall through */
 
-	case RANDOM_INIT:
-		/* Fall through */
-
 	case TIMER_INIT:
 		if (_odp_timer_term_global()) {
 			ODP_ERR("ODP timer term failed.\n");
@@ -283,8 +280,8 @@ static int term_global(enum init_stage stage)
 		}
 		/* Fall through */
 
-	case NO_INIT:
-		;
+	default:
+		break;
 	}
 
 	return rc;
@@ -410,9 +407,6 @@ int odp_init_global(odp_instance_t *instance,
 		goto init_failed;
 	}
 	stage = TIMER_INIT;
-
-	/* No init neeeded */
-	stage = RANDOM_INIT;
 
 	if (_odp_crypto_init_global()) {
 		ODP_ERR("ODP crypto init failed.\n");
--- a/platform/linux-generic/odp_traffic_mngr.c
+++ b/platform/linux-generic/odp_traffic_mngr.c
@@ -2246,12 +2246,14 @@ static void tm_send_pkt(tm_system_t *tm_system, uint32_t max_sends)
 			tm_egress_marking(tm_system, odp_pkt);
 
 		tm_system->egress_pkt_desc = EMPTY_PKT_DESC;
-		if (tm_system->egress.egress_kind == ODP_TM_EGRESS_PKT_IO)
-			odp_pktout_send(tm_system->pktout, &odp_pkt, 1);
-		else if (tm_system->egress.egress_kind == ODP_TM_EGRESS_FN)
+		if (tm_system->egress.egress_kind == ODP_TM_EGRESS_PKT_IO) {
+			if (odp_pktout_send(tm_system->pktout, &odp_pkt, 1) != 1)
+				odp_packet_free(odp_pkt);
+		} else if (tm_system->egress.egress_kind == ODP_TM_EGRESS_FN) {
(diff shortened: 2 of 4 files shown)


platform/linux-generic/odp_init.c
```diff
@@ -173,9 +173,6 @@ static int term_global(enum init_stage stage)
 		}
 		/* Fall through */
 
-	case RANDOM_INIT:
-		/* Fall through */
-
```
JannePeltonen, line 178: I am not sure this is a good change as I think it makes the code slightly less robust for future changes. The RANDOM_INIT still exists in the enum.

platform/linux-generic/odp_init.c
```diff
@@ -411,9 +408,6 @@ int odp_init_global(odp_instance_t *instance,
 	}
 	stage = TIMER_INIT;
 
-	/* No init neeeded */
-	stage = RANDOM_INIT;
```
JannePeltonen, line 415: If this consecutive assignment is triggering some warnings, then how about commenting it out here but not touching the switch case in term_global()?

platform/linux-generic/odp_init.c
psavol: RANDOM_INIT is used in local init/term side.

Local init/term uses only part of the enums. After this change global init/term would also have only those enums that actually do something.

platform/linux-generic/odp_traffic_mngr.c
```diff
@@ -2246,12 +2246,14 @@ static void tm_send_pkt(tm_system_t *tm_system, uint32_t max_sends)
 			tm_egress_marking(tm_system, odp_pkt);
 
 		tm_system->egress_pkt_desc = EMPTY_PKT_DESC;
-		if (tm_system->egress.egress_kind == ODP_TM_EGRESS_PKT_IO)
-			odp_pktout_send(tm_system->pktout, &odp_pkt, 1);
-		else if (tm_system->egress.egress_kind == ODP_TM_EGRESS_FN)
+		if (tm_system->egress.egress_kind == ODP_TM_EGRESS_PKT_IO) {
+			if (odp_pktout_send(tm_system->pktout, &odp_pkt, 1) != 1)
+				odp_packet_free(odp_pkt);
```
JannePeltonen, line 2251: This looks ok as it fixes the packet leak and thus makes the code more correct than it was.

That said, this tm_send_pkt() looks a bit fishy to me as it does not seem to be designed to handle the case when the pktout cannot absorb all the packets sent through it. I would think TM should then slow down sending and let the configured schedulers and queue management policies do their thing and not simply drop packets here (or within the odp_pktout_send). But that does not seem to be the case and it seems to me that TM works only when a shaper is limiting the output rate and not the pktout itself.

Maybe the TM code here assumed that odp_pktout_send() will not normally fail and will block when there congestion in the output link. But I am not sure if TM would work correctly even if odp_pktout_send() worked that way.

So, I am ok with this fix as it fixes the packet leak and does not break anything, although I am not sure if the action taken in the failure case should be something else (retry?) than freeing the packet if we want to fix other things. Assuming there are other things to fix and I am not just confused. The potential other issues can be fixed separately.

platform/linux-generic/odp_init.c
JannePeltonen: Yes, but since it is in the enum and the enum seems to imply certain ordering (even though it is the code that really does), it seems a bit questionable to me to change the behaviour of term_global() if called with RANDOM_INIT. If someone adds a RANDOM_INIT stage in init_global() in the place it is now (implied by the ordering in the enum) and does not need any term handling and does not add the switch case back, we may have a bug.

Or to put it in other way, I do not see how this change makes the code any better, other than maybe making coverity happy. But in that case we could just suppress the warning.

psavol: Init and term function sequences need to match. If random init global would be added later, it would need to be added into correct position relative to other inits. The term the stage would be added into the matching position. These positions are not dictated by the enum, but which code block uses which other code blocks.

So, I think it's better to remove this dead code, which may be placed wrongly anyway. Similarly, local init/term does not have extra stages.

JannePeltonen: > The term the stage would be added into the matching position.

Only if the term would be needed. There could be init without any associated term.

> These positions are not dictated by the enum, but which code block uses which other code blocks.

Yes, which is why I added: "(even though it is the code that really does)".

> So, I think it's better to remove this dead code, which may be placed wrongly anyway. Similarly, local init/term does not have extra stages.

That is a valid point too. We can leave the code as it is in the patch.

JannePeltonen (approved):

PR #1250, opened 2021-04-08
[PATCH v1] pktin/pool capability flag to indicate vector size should be power of 2

shijith:


include/odp/api/spec/packet_io.h
```diff
@@ -790,6 +790,9 @@ typedef struct odp_pktin_vector_capability_t {
 	/** Packet input vector availability */
 	odp_support_t supported;
 
+	/** odp_pktin_vector_config_t::max_size should be power of two. */
+	odp_bool_t size_is_pow2;
```
JannePeltonen, line 794: I wonder if this is needed or whether it would be better to adjust the specification of odp_pktin_vector_config_t::max_size so that the pktin shall deliver received packets _no later than_ when the timeout occurs or when it has accumulated max_size packets. In other words, allow the pktin to produce vectors shorter than max_size even when the timeout does not occur. The application anyway has to be able to handle such shorter vectors for the timeout case.

include/odp/api/spec/pool.h
```diff
@@ -271,6 +271,9 @@ typedef struct odp_pool_capability_t {
 
 	/** Vector pool capabilities */
 	struct {
+		/** odp_pool_param_t::vector::max_size should be power of two. */
+		odp_bool_t size_is_pow2;
```
JannePeltonen, line 275: I do not see why this is needed. I would think the implementation could simply internally round up the size to the next power of two.

PR #1251, opened 2021-04-14
[PATCH v3] validation: crc: add CRC32 test cases

psavol:
Additional test cases for CRC32 validation. Use Ethernet frames with CRC as test vectors. Run test cases also with odd data alignment.


test/validation/api/hash/hash.c
```diff
@@ -238,6 +240,85 @@ static void hash_test_crc32(void)
...
+
+	ret = odp_hash_crc32(test_packet_ipv4_udp_503_crc,
+			     sizeof(test_packet_ipv4_udp_503_crc), CRC32_INIT);
+	CU_ASSERT(ret == verify);
+}
+
+static void hash_test_ethernet_crc32_odd_align(void)
+{
+	uint32_t ret, size;
+	uint32_t verify = 0xdebb20e3;
```
JereLeppanen, line 280: This could be defined in just one place. Also, could be const.

test/validation/api/hash/hash.c
```diff
@@ -238,6 +240,85 @@ static void hash_test_crc32(void)
...
+	uint32_t max_size = sizeof(test_packet_ipv4_udp_503_crc);
```
JereLeppanen, line 281: Const.

JereLeppanen (approved):
Just a couple of minor comments, you can decide what to do about them. Either way, approved.

test/validation/api/hash/hash.c
psavol: Fixed in v2. Changed it to a define.

test/validation/api/hash/hash.c
psavol: fixed in v2

PR #1252, opened 2021-04-15
[PATCH v8] api: introduce per-thread log function

JereLeppanen:
```
api: introduce per-thread log function
    
    The odp_log_thread_fn_set() API allows the caller to set a per-thread
    log function. This log function is called to do output when the thread
    calls other API functions.
    
linux-gen: implement odp_thread_set_log_fn()
    
    Implement the odp_thread_set_log_fn() API function.
    
helper: cli: direct prints to the CLI user instead of the log
    
    Use the odp_log_thread_fn_set() API to direct prints from API calls to
    the CLI user.
    
validation: init: test thread-specific log function
    
    Test that a thread-specific log function is called if and only if it
    is set.
    
```
This PR is alternative to PR #1242.

v3:
- Call _ODP_LOG_FN in debug_internal.h.
- Add validation test.
- Add reviewed-by tags.

v4:
- helper/cli.c: Remove _GNU_SOURCE, strdup() and *asprintf().
- api: Improve description of odp_log_thread_fn_set().
- test/validation/api/init/.gitignore: Add init_log_thread.
- Set _odp_this_thread to NULL in term_local.
- Use the term "thread specific" consistently (instead of "per thread").

v5:
- Add commit "helper: test: linux: process: add missing odp_term_local() calls"
- Add commit "linux-gen: init: fix odp_term_local() return value"

v6:
- Add commit "linux-gen: thread: in global termination, check number of active threads"

v7:
- odp_thread.c: ODP_DBG() -> ODP_ERR()

v8:
- Free cli_log_fn buffer when client disconnects.
- Call odp_log_thread_fn_set() right before and after cli_loop().
- Use "CLI client" instead of "CLI user".

### DIFF
--- a/include/odp/api/spec/init.h
+++ b/include/odp/api/spec/init.h
@@ -333,6 +333,20 @@ int odp_term_local(void);
 int odp_term_global(odp_instance_t instance);
 
 /**
+ * Set thread specific log function
+ *
+ * By default, all ODP log writes use the global log function, which may be set
+ * as part of odp_init_t. Using this operation, an alternative ODP log function
+ * may be set for the calling thread. When set, ODP uses the thread specific log
+ * function for all log writes originating from ODP API calls made by the
+ * calling thread. Setting the log function to NULL causes the calling thread to
+ * use the global log function.
+ *
+ * @param func Log function
+ */
+void odp_log_thread_fn_set(odp_log_func_t func);
+
+/**
  * @}
  */
 
--- a/include/odp/api/spec/init.h
+++ b/include/odp/api/spec/init.h
@@ -20,7 +20,7 @@ extern "C" {
 #include <odp/api/std_types.h>
 #include <odp/api/hints.h>
 #include <odp/api/feature.h>
-#include <odp/api/thread.h>
+#include <odp/api/spec/thread_types.h>
 #include <odp/api/cpumask.h>
 
 /** @defgroup odp_initialization ODP INITIALIZATION
--- a/platform/linux-generic/odp_thread.c
+++ b/platform/linux-generic/odp_thread.c
@@ -88,7 +88,14 @@ int _odp_thread_init_global(void)
 
 int _odp_thread_term_global(void)
 {
-	int ret;
+	int ret, num;
+
+	odp_spinlock_lock(&thread_globals->lock);
+	num = thread_globals->num;
+	odp_spinlock_unlock(&thread_globals->lock);
+
+	if (num)
+		ODP_ERR("%u threads have not called odp_term_local().\n", num);
 
 	ret = odp_shm_free(odp_shm_lookup("_odp_thread_globals"));
 	if (ret < 0)
--- a/platform/linux-generic/odp_thread.c
+++ b/platform/linux-generic/odp_thread.c
@@ -231,6 +231,8 @@ int _odp_thread_term_local(void)
 	if (type == ODP_THREAD_CONTROL && group_control)
 		_odp_sched_fn->thr_rem(ODP_SCHED_GROUP_CONTROL, id);
 
+	_odp_this_thread = NULL;
+
 	odp_spinlock_lock(&thread_globals->lock);
 	num = free_id(id);
 	odp_spinlock_unlock(&thread_globals->lock);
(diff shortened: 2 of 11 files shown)


jerinjacobk (approved):

include/odp/api/spec/init.h
```diff
@@ -332,6 +332,16 @@ int odp_term_local(void);
  */
 int odp_term_global(odp_instance_t instance);
 
+/**
+ * Set log function
+ *
+ * Sets the log function for the calling thread. Setting the log function to
+ * NULL causes the calling thread to use the global log function.
```
psavol, line 339: I think the spec text could be more explicit that the intention of this call is to override the global log function (and its usage is optional). May be something like this:

"Set per thread log function

By default, all ODP log writes use the global log function, which may be set as part of odp_init_t. Using this operation, an alternative ODP log function may be set for the calling thread. When set, ODP uses the thread specific log function for all log writes originating from ODP API calls made by the calling thread. Setting the log function to NULL ..."

test/validation/api/init/init_main.c
```diff
@@ -5,11 +5,14 @@
  * SPDX-License-Identifier:     BSD-3-Clause
  */
 
+#define _GNU_SOURCE
```
psavol, line 8: Why this needs to be added?

include/odp/api/spec/init.h
JereLeppanen: Fixed in v4 (with "Set thread specific log function").

test/validation/api/init/init_main.c
JereLeppanen: It doesn't. Left behind by mistake. Removed in v4.

platform/linux-generic/odp_thread.c
```diff
@@ -88,7 +88,14 @@ int _odp_thread_init_global(void)
...
 {
-	int ret;
+	int ret, num;
+
+	odp_spinlock_lock(&thread_globals->lock);
+	num = thread_globals->num;
+	odp_spinlock_unlock(&thread_globals->lock);
+
+	if (num)
+		ODP_DBG("%u threads have not called odp_term_local().\n", num);
```
psavol, line 98: This could be ODP_ERR() since ODP_DBG prints out only with a debug build. I think we'd like to see this by default on all application logs if they violate the proper termination sequence.

psavol (approved):
One minor comment: ODP_DBG() -> ODP_ERR()

platform/linux-generic/odp_thread.c
JereLeppanen: I was just about to ask about this. Yes, ODP_ERR is better here.

include/odp/api/spec/init.h
```diff
@@ -332,6 +332,20 @@ int odp_term_local(void);
...
+ * By default, all ODP log writes use the global log function, which may be set
+ * as part of odp_init_t. Using this operation, an alternative ODP log function
+ * may be set for the calling thread. When set, ODP uses the thread specific log
+ * function for all log writes originating from ODP API calls made by the
+ * calling thread. Setting the log function to NULL causes the calling thread to
+ * use the global log function.
+ *
+ * @param func Log function
+ */
+void odp_log_thread_fn_set(odp_log_func_t func);
```
yuri-tolstov, line 347: Considering that this is API call, I think, it needs to be consistent with the API naming convention -- the function needs to show to which ODP module it belongs. Plus, to be consistent with the existing odp_log_func_t type, probably, the better function name would be -- odp_thread_log_func_set()? This is, of course, if we assume that it is a part of the thread module. If this is a part of the "initialization" module, where, it seems, more freedom exists with the naming (e.g. odp_init_..., odp_term_..., etc, then some rules for naming in that module needs to be specified in the API. This simplifies the work with the ODP documentation.

psavol: It belongs to "ODP log" module, which is currently embedded into init.h. Log related definitions include: odp_log_level_t, odp_log_func_t and odp_override_log(). All these could be moved into a new odp_log.h file at some point. odp_override_log() could be renamed to odp_log_override(). It's likely not used much as it's just another way to set global log function (instead of setting it through init param). It could be also deprecated.

PR #1258, opened 2021-04-22
[PATCH v2] test: performance: ipsec: test app for multi-core ipsec performance

vijayinavolu:
Adding new application odp_ipsec_ordered to test ipsec outbound
processing with multiple worker cores. The current IPsec performance
test application is not suitable to test ordering operations overhead
hence a new app is written with the following capabilities. Main core
creates and supplies packets with a certain level of flow control to
worker cores via the configured scheduler. Scheduler mode and workers
count is configurable. Performance data is collected only for the IPsec
calls and summary is displayed per thread per algo for each packet size

Based on the feedback this app can be evolved more to add any additional
capabilities

### DIFF
--- a/test/performance/odp_ipsec_ordered.c
+++ b/test/performance/odp_ipsec_ordered.c
@@ -1,0 +1,1196 @@
+/* Copyright (c) 2021, Marvell
+ * All rights reserved.
+ *
+ * SPDX-License-Identifier:     BSD-3-Clause
+ */
+
+#ifndef _GNU_SOURCE
+#define _GNU_SOURCE
+#endif /* _GNU_SOURCE */
+
+#include <stdlib.h>
+#include <string.h>
+#include <getopt.h>
+#include <unistd.h>
+#include <stdio.h>
+#include <sys/time.h>
+#include <sys/resource.h>
+
+#include <odp_api.h>
+#include <odp/helper/odph_api.h>
+
+#define app_err(fmt, ...) \
+	fprintf(stderr, "%s:%d:%s(): Error: " fmt, __FILE__, \
+			__LINE__, __func__, ##__VA_ARGS__)
+
+/** @def POOL_NUM_PKT
+ * Number of packets in the pool
+ */
+#define POOL_NUM_PKT  4096
+
+static uint8_t test_salt[16] = "0123456789abcdef";
+
+static uint8_t test_key16[16] = { 0x01, 0x02, 0x03, 0x04, 0x05,
+	0x06, 0x07, 0x08, 0x09, 0x0a,
+	0x0b, 0x0c, 0x0d, 0x0e, 0x0f,
+	0x10,
+};
+
+static uint8_t test_key20[20] = { 0x01, 0x02, 0x03, 0x04, 0x05,
+	0x06, 0x07, 0x08, 0x09, 0x0a,
+	0x0b, 0x0c, 0x0d, 0x0e, 0x0f,
+	0x10, 0x11, 0x12, 0x13, 0x14,
+};
+
+static uint8_t test_key24[24] = { 0x01, 0x02, 0x03, 0x04, 0x05,
+	0x06, 0x07, 0x08, 0x09, 0x0a,
+	0x0b, 0x0c, 0x0d, 0x0e, 0x0f,
+	0x10, 0x11, 0x12, 0x13, 0x14,
+	0x15, 0x16, 0x17, 0x18
+};
+
+static uint8_t test_key32[32] = { 0x01, 0x02, 0x03, 0x04, 0x05,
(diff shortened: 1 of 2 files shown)


JannePeltonen:
This seems to be heavily based on the existing odp_ipsec performance test application, which is (c) Linaro.

To avoid duplication of code, maybe there should be just one test application that can do different kind of tests or some way of code sharing between the different applications.

This application performs two system calls (getrusage) for every processed packet. This is not good and can skew the results badly. If you really want to measure just the the IPsec call and not the whole loop, using just gettimeofday() (or clock_gettime() for better resolution) or other mechanism implemented using vDSO without actual kernel entry should work much better. But maybe the code should just use ODP APIs (odp_time_local() etc) for it.

Since ODP worker threads are supposed to be run on dedicated cores, I am not sure there is even point in trying to read the kernel maintained run time statistics. They may not even work that well if the worker threads are well isolated from the kernel (using full tickless mode etc).

There is also some false cache line sharing going on between the worker threads through gbl_args->result as the elements for different threads are not in different cache lines.

I am not sure measuring the duration of just odp_ipsec_out() instead of the whole loop tells the whole story.Some of the ordering overhead (like releasing the ordered context as part of the next schedule call) happens outside it.

test/performance/odp_ipsec_ordered.c
```diff
@@ -0,0 +1,1279 @@
...
+ * Get diff of elapsed time between two time snap records
+ */
+static unsigned long long
+get_elapsed_usec(time_record_t *start, time_record_t *end)
+{
+	unsigned long long s;
+	unsigned long long e;
+
+	s = (start->tv.tv_sec * 1000000) + (start->tv.tv_usec);
+	e = (end->tv.tv_sec * 1000000) + (end->tv.tv_usec);
```
JannePeltonen, line 554: These can overflow in 32 bit systems. You could fix that (for reasonable test durations) and also avoid the use of long long by first subtracting the seconds from each other and the microseconds from each other and then combining.

test/performance/odp_ipsec_ordered.c
```diff
@@ -0,0 +1,1279 @@
...
+
+	/* Scheduler Init */
+	if (odp_schedule_capability(&schedule_capa)) {
+		printf("Error: Schedule capa failed.\n");
+		return -1;
+	}
+	if ((cargs->sched_sync == ODP_SCHED_SYNC_ORDERED) &&
+	    (schedule_capa.max_ordered_locks < 1)) {
+		ODPH_ERR("Error: Ordered locks not available.\n");
+		exit(EXIT_FAILURE);
```
JannePeltonen, line 1179: This application is not using ordered locks, so there is no point in requiring them here and in the scheduler config.

vijayinavolu:
> This seems to be heavily based on the existing odp_ipsec performance test application, which is (c) Linaro.
> 
> To avoid duplication of code, maybe there should be just one test application that can do different kind of tests or some way of code sharing between the different applications.
> 
> This application performs two system calls (getrusage) for every processed packet. This is not good and can skew the results badly. If you really want to measure just the the IPsec call and not the whole loop, using just gettimeofday() (or clock_gettime() for better resolution) or other mechanism implemented using vDSO without actual kernel entry should work much better. But maybe the code should just use ODP APIs (odp_time_local() etc) for it.
> 
> Since ODP worker threads are supposed to be run on dedicated cores, I am not sure there is even point in trying to read the kernel maintained run time statistics. They may not even work that well if the worker threads are well isolated from the kernel (using full tickless mode etc).
> 
> There is also some false cache line sharing going on between the worker threads through gbl_args->result as the elements for different threads are not in different cache lines.
> 
> I am not sure measuring the duration of just odp_ipsec_out() instead of the whole loop tells the whole story.Some of the ordering overhead (like releasing the ordered context as part of the next schedule call) happens outside it.




This app is derived from the original single core app and it seems better to keep it separate for few reasons

The multi core IPsec app only supports sync mode so combining this single core app which does both sync and async might become tricky to get meaningful measurements in common format

The original app uses getrusage based measurements and changing that to something else might make it difficult for any past and future comparisons to use the app

Changing the measurement from per IPsec call to entire run like the original app may not scale well in this multicore app as the number of workers are increased 

The current examples which are present already have duplicated functions between them which are slightly modified for each test application, this new app also does the same

test/performance/odp_ipsec_ordered.c
vijayinavolu: It will be fixed in next version

test/performance/odp_ipsec_ordered.c
vijayinavolu: Order lock checks are now removed next version

JannePeltonen:

PR #1259, opened 2021-04-22
[PATCH v1] linux-gen: ipsec: Enable Order Locks for ipsec processing

vijayinavolu:
Enables ordered processing for IPsec anti replay check and sequence
number generation. It uses a separate lock index for inbound and
outbound which are defined locally for IPsec

Linux generic scheduler currently  supports two lock indices per queue by
default which are going to be used by IPsec with this patch

### DIFF
--- a/platform/linux-generic/odp_ipsec.c
+++ b/platform/linux-generic/odp_ipsec.c
@@ -103,6 +103,9 @@ static int set_ipsec_crypto_capa(odp_ipsec_capability_t *capa)
 	return 0;
 }
 
+#define	IPSEC_IN_ORDER_LOCK_INDEX 	0
+#define	IPSEC_OUT_ORDER_LOCK_INDEX 	1
+
 int odp_ipsec_capability(odp_ipsec_capability_t *capa)
 {
 	int rc;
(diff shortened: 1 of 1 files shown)


JannePeltonen:
This breaks IPsec applications that try to use ordered locks. The ordered lock API toward the application does not have such a concept that some of the "available" ordered locks are already taken by the implementation. Even if there was such a concept, it would be difficult to ensure the ordering between the ordered locks in the application and the ordered locks in the implementation is correct.

If it turns out that this sort of ordered lock is really needed inside the IPsec implementation, then maybe the IPsec API should make it explicit and application configurable.

But I am not yet sure the implementation really needs to use the ordered locks.

The IPsec API does not guarantee ordering between synchronous calls made between different threads, so the current implementation seems correct to me. Thus it seems wrong that this PR affects simnilarly both the synchronous and asynchronous case.

The asynchronous interface is supposed to maintain the original input queue order also across threads, i.e. it is supposed to have the same rdering semantics as queue enqueuing (but unfortunately the current implementation does not do it). But maybe instead of ordered locks an implementation should just do essentially queue enqueuing and dequeuing with the ordering and new scheduling context creation that they imply. The implementation could hopefully optimize things internally so that the enqueue-dequeue would be less costly than when done through the API functions.

JannePeltonen:
Can we close this PR now since PR 1284, that was merged as e94c28af5, introduces a simple ordering mehcanism (although disabled by default) that has essentially the same effect as this would have had but without using the odered-lock API?

Further improvements can be done through new PRs.

vijayinavolu:
Closing this initial PR as its no longer required

PR #1260, opened 2021-04-23
[PATCH v16] api: external memory pool

psavol:
This is still a draft. Some API details need to be still decided: e.g. populate done flag (or no indication from application). Also did not yet add packet head to handle conversion. Validation tests need to be extended with packet API and possibly pktio tests.


psavol:
v5: rebased, added capability fields: type and odp_trailer_size

include/odp/api/spec/pool_ext.h
```diff
@@ -0,0 +1,318 @@
...
+			  * each buffer start address must be a multiple of 0x900
+			  * (e.g. 0x12000900, 0x12001200, 0x12004800, etc). */
+			uint16_t buf_size_aligned : 1;
+		};
+
+		/** Maximum headroom parameter value
+		 *
+		 *  The packet pool headroom parameter can be configured in maximum to this value.
+		 */
+		uint32_t max_headroom;
```
ashwinyes, line 112: Could you please explain on what is the difference between max_headroom and max_headroom_size .

include/odp/api/spec/pool_ext.h
```diff
@@ -0,0 +1,318 @@
...
+ *                      |                               |     |                      |
+ *                      +-------------------------------+    --                     --
+ *
+ * @endcode
+ *
+ * @param pool    External memory pool
+ * @param buf     Buffer pointers to be populated into the pool
+ * @param num     Number of buffer pointers
+ * @param flags   Flags reserved for future use.
+ *                TODO: will additional information be needed per buffer? E.g.
```
jerinjacobk, line 297: Please add one DONE flag. To denote the completion of populate.
i.e after DONE odp_packet_alloc() etc can be called.

include/odp/api/spec/packet.h
```diff
@@ -475,6 +475,22 @@ int odp_packet_reass_partial_state(odp_packet_t pkt, odp_packet_t frags[],
  */
 void *odp_packet_head(odp_packet_t pkt);
 
+/**
+ * Packet handle from packet head address
+ *
+ * Returns packet handle based on its head pointer value. Application must have ownership of
+ * the packet when requesting this conversion. Returned packet handle value (for the same pointer
+ * value) may change between packet ownership changes (e.g. due to handle generation counting).
+ * Application must not store and use handle values from previous ownerships of the packet.
```
jerinjacobk, line 484: Thinking, more advanced use cases like VPP, avoid such limitations and since APP header is exposed to ODP spec in prev ODP API patch.
I think, it is better to have the following APIs instead of odp_packet_from_head()

1. Introduce new object similar to odp_packet_seg_t whose scope is get split and combine packet and have basic operation on the object like get/set data pointer and size etc LIKE
2. int odp_packet_app_seg_split(odp_packet_t pkt, odp_packet_app_seg_t *app_segs, int *num_segs);
3. odp_packet_t  odp_packet_app_seg_combine( odp_packet_app_seg_t *app_segs, int num_segs);
4. void* odp_packet_app_seg_header(odp_packet_app_seg_t  seg);
5.  odp_packet_app_seg_t  odp_packet_app_seg_from_header(odp_pool_t pool, void *app_header);
6. int odp_packet_app_seg_len(odp_packet_app_seg_t  seg);
7. void* odp_packet_app_seg_data(odp_packet_app_seg_t  seg);

include/odp/api/spec/pool_ext.h
psavol: Application requests minimum headroom size that packets have after alloc or packet input. This max_headroom capa is the maximum value for that (min) headroom parameter. Implementation may need to round up headroom size e.g. on packet input to align packet headers optimally. Max_headroom_size capa tells to application how much it needs to reserve for  headroom bytes "in the worst case", so that it is able to select large enough buffer size that fits e.g. MTU frame size in a single buffer.

include/odp/api/spec/pool_ext.h
psavol: done in v8

include/odp/api/spec/packet.h
psavol: done in v8

include/odp/api/spec/packet.h
```diff
@@ -1155,6 +1165,135 @@ int odp_packet_concat(odp_packet_t *dst, odp_packet_t src);
...
+ * @return Number of handles written (equals the number of segments in the packet)
+ * @retval 0 on failure
+ */
+uint32_t odp_packet_disassemble(odp_packet_t pkt, odp_packet_buf_t pkt_buf[], uint32_t num);
+
+/**
+ * Reassemble packet from packet buffers
+ *
+ * Forms a new packet from packet buffers of a previous odp_packet_disassemble() call(s). Packet
+ * buffers from different dissabled packets may be used, but all buffers must be from packets of
```
ashwinyes, line 1275: disassembled

include/odp/api/spec/packet.h
```diff
@@ -1155,6 +1165,135 @@ int odp_packet_concat(odp_packet_t *dst, odp_packet_t src);
...
+ * Packet buffer data set
+ *
+ * Update packet data start offset and length in the packet buffer. Valid offset values range
+ * from 0 to buf_size - 1. Valid length values range from 1 to buf_size - data_offset.
+ *
+ * @param      pkt_buf      Packet buffer
+ * @param      data_offset  Packet buffer data offset in bytes (from the buffer head pointer)
+ * @param      data_len     Packet buffer data length in bytes
+ */
+void odp_packet_buf_data_set(odp_packet_buf_t pkt_buf, uint32_t data_offset, uint32_t data_len);
```
ashwinyes, line 1229: This should be returning success/failure. 

The api can internally verify the values passed and can fail if invalid values are given,

psavol: We did leave return value away, since this is a fast path API and the call should succeed always when application provides valid input values. Implementation may report error to log if some value is invalid.

include/odp/api/abi-default/packet.h
```diff
@@ -19,6 +19,9 @@ typedef struct { char dummy; /**< @internal Dummy */ } _odp_abi_packet_t;
 /** @internal Dummy  type for strong typing */
 typedef struct { char dummy; /**< *internal Dummy */ } _odp_abi_packet_seg_t;
 
+/** @internal Dummy  type for strong typing */
+typedef struct { char dummy; /**< *internal Dummy */ } _odp_abi_packet_buf_t;
+
```
yuri-tolstov, line 24: The difference between packet buffer and packet segment, probably, needs to be clarified (explicitly specified) to avoid confusion in implementation.

psavol: Fixed, documentation added into odp/api/spec/packet.h.

include/odp/api/spec/packet.h
psavol: Fixed in v10

include/odp/api/spec/pool_ext_types.h
```diff
@@ -0,0 +1,221 @@
...
+		uint32_t max_uarea_size;
+
+	} pkt;
+
+} odp_pool_ext_capability_t;
+
+/**
+ * External memory pool parameters
+ */
+typedef struct odp_pool_ext_param_t {
```
ashwinyes, line 143: @psavol odp_pool_info_t has odp_pool_param_t

So what should be the behavior of odp_pool_info() API for external pools ?

include/odp/api/spec/pool_ext_types.h
```diff
@@ -0,0 +1,221 @@
...
+		 *  These flags specify additional alignment requirements for packet buffers.
+		 *  If not stated otherwise, min_buf_align and min_head_align alignment
+		 *  requirements apply also.
+		 */
+		struct {
+			/** Packet buffer pointers must be aligned to the buffer size.
+			  * For example, if the buffer size would be 2304 bytes (0x900),
+			  * each buffer start address must be a multiple of 0x900
+			  * (e.g. 0x12000900, 0x12001200, 0x12004800, etc). */
+			uint16_t buf_size_aligned : 1;
```
ashwinyes, line 113: Please add a 'contiguous' capability bit as well. 

This capability if set will enforce the buffers to be contiguous in memory.

test/validation/api/pool/pool.c
```diff
@@ -1450,6 +1450,126 @@ static void test_packet_pool_ext_alloc_seg(void)
...
+
+		/* Disassemble packet */
+		num_pkt_buf = odp_packet_disassemble(pkt, pkt_buf, num_seg);
+		CU_ASSERT_FATAL(num_pkt_buf == num_seg);
+
+		CU_ASSERT(odp_packet_buf_head(pkt_buf[0]) == pkt_head);
+		CU_ASSERT(odp_packet_buf_data_offset(pkt_buf[0]) == hr);
+
+		for (j = 0; j < num_seg; j++) {
+			CU_ASSERT(odp_packet_buf_size(pkt_buf[j]) == buf_size);
```
ashwinyes, line 1534: Is this check correct ?

Here we are comparing param.pkt.buf_size and odp_packet_buf_size() which are not same according to spec.

test/validation/api/pool/pool.c
```diff
@@ -1121,6 +1129,327 @@ static void pool_test_timeout_pool_statistics(void)
...
+		buf_index = find_buf(pkt[i], buf, num_buf, head_offset);
+		CU_ASSERT(buf_index < num_buf);
+		hr = (uintptr_t)odp_packet_data(pkt[i]) - (uintptr_t)odp_packet_head(pkt[i]);
+		CU_ASSERT(hr == odp_packet_headroom(pkt[i]));
+		CU_ASSERT(num_seg == odp_packet_num_segs(pkt[i]));
+		CU_ASSERT(odp_packet_data(pkt[i]) == odp_packet_data_seg_len(pkt[i], &seg_len));
+		CU_ASSERT(odp_packet_seg_len(pkt[i]) == seg_len);
+
+		if (num_seg == 1) {
+			tr = buf_size - head_offset - hr - pkt_len;
```
ashwinyes, line 1405: Need to reduce trailer size as well to calculate tr.

test/validation/api/pool/pool.c
```diff
@@ -1121,6 +1129,327 @@ static void pool_test_timeout_pool_statistics(void)
...
+		} else {
+			odp_packet_seg_t seg = odp_packet_last_seg(pkt[i]);
+			uint32_t last_seg_len = odp_packet_seg_data_len(pkt[i], seg);
+			uint32_t max_tr = buf_size - head_offset - last_seg_len;
+
+			CU_ASSERT(odp_packet_tailroom(pkt[i]) < max_tr);
+			CU_ASSERT(pkt_len == (odp_packet_seg_len(pkt[i]) + last_seg_len));
+		}
+
+		CU_ASSERT(odp_packet_buf_len(pkt[i]) == num_seg * (buf_size - head_offset));
```
ashwinyes, line 1417: Need to reduce trailer size as well to calculate buf len.

include/odp/api/spec/pool_ext_types.h
```diff
@@ -0,0 +1,221 @@
...
+		};
+
+		/** Maximum headroom parameter value
+		 *
+		 *  The packet pool headroom parameter may not exceed this value.
+		 */
+		uint32_t max_headroom;
+
```
ashwinyes, line 121: I think we require a min_tailroom capability as well. Because application needs to know that to reserve the space for it.

test/validation/api/pool/pool.c
psavol: True, there's a bug in implementation and test case. I'll fix it.

test/validation/api/pool/pool.c
psavol: Fixed in v11

test/validation/api/pool/pool.c
psavol: Fixed in v11

test/validation/api/pool/pool.c
psavol: Fixed in v11

include/odp/api/spec/pool_ext_types.h
psavol: I don't tailroom capa is needed with ext mem pools.  Application decides the total buffer size (includes all headers, trailer, head-/tailroom and data) and tailroom is the unused buffer space after headroom and payload data (and before trailer).

Application needs to know trailer size, but that's already in the capa.

include/odp/api/spec/pool_ext_types.h
psavol: Added in v12

include/odp/api/spec/pool_ext_types.h
ashwinyes: Okay. Then min_tailroom is 0 for external pools.

include/odp/api/spec/pool_ext_types.h
```diff
@@ -0,0 +1,231 @@
...
+			 *  For example, if the buffer size would be 2304 bytes (0x900),
+			 *  each buffer start address must be a multiple of 0x900
+			 *  (e.g. 0x12000900, 0x12001200, 0x12004800, etc). */
+			uint16_t buf_size_aligned : 1;
+
+			/** Packet buffers are contiguous in memory
+			 *
+			 *  When set, packet buffers populated into a pool must form a contiguous
+			 *  memory area (holes are not allowed between buffers). */
+			uint16_t contiguous_mem : 1;
```
psavol, line 122: I'll remove this flag, and instead add a spec text telling that two pools must not have overlapping memory allocations.

include/odp/api/spec/pool_ext_types.h
```diff
@@ -0,0 +1,231 @@
...
+		 *  memory area (after buffer populate call).
+		 */
+		uint32_t odp_trailer_size;
+
+		/** Minimum packet buffer pointer alignment in bytes
+		 *
+		 *  Packet buffer pointers populated into a pool must be evenly divisible with
+		 *  this value. The value is 1 when there is no alignment requirement.
+		 */
+		uint32_t min_buf_align;
```
psavol, line 93: I'll add another alignment capability about minimum alignment of the start of the memory area used for an external memory pool. This can e.g. ensure that memory of two pools do not get mapped to the same page.

ashwinyes: Instead of an alignment capability, I suggest to add capability as "sys_page_sharing".

This capability when **not** set, puts a restriction that the memory area used for two different pools must not have common shared sys pages. 

Memory area of a pool is defined as the memory between the beginning of lowest addressed buffer and the end of the highest addressed buffer populated into the pool.

sys page is defined as any memory area whose address is aligned to and having size of odp_sys_page_size().

This will also avoid the need for having a spec text regarding overlapping memory alloc as it is implied by this.

include/odp/api/spec/pool_ext_types.h
psavol: Removed in v13. Added note into populate function documentation.

include/odp/api/spec/pool_ext_types.h
psavol: Added in v13.

include/odp/api/spec/pool.h
```diff
@@ -173,6 +173,107 @@ int odp_pool_stats(odp_pool_t pool, odp_pool_stats_t *stats);
...
+ * Populate can be called multiple times to add memory buffers into the pool. Application must
+ * populate the pool with the exact number of buffers specified in pool parameters. The pool is
+ * ready to be used for allocations only after all populate calls have returned successfully.
+ * Application marks the last populate call with ODP_POOL_POPULATE_DONE flag.
+ *
+ * Depending on pool usage (and ODP implementation), the memory may need to be accessible by
+ * HW accelerators. Application may use e.g. odp_shm_reserve() with ODP_SHM_HW_ACCESS flag to
+ * ensure HW access. The memory area used for one pool (from the lowest addressed buffer to
+ * the end of the highest addressed buffer) must not overlap with the memory area used for
+ * any other pool.
```
ashwinyes, line 229: This still doesn't imply that sys pages must not be shared between pools. 

Could you add a capability "sys_page_sharing" described below (or add spec text which implies this restriction.)

"sys_page_sharing" capability when **not** set, puts a restriction that the memory area used for two different pools must not have common shared sys pages.

Memory area of a pool is defined as the memory between the beginning of lowest addressed buffer and the end of the highest addressed buffer populated into the pool.

sys page is defined as any memory area whose address is aligned to and having size of odp_sys_page_size().

include/odp/api/spec/pool_ext_types.h
ashwinyes: Please see my comment https://github.com/OpenDataPlane/odp/pull/1260#discussion_r661171604

include/odp/api/spec/pool.h
psavol: odp_sys_page_size() does not return e.g. "System MMU"  page size, but just regular page size of the OS (e.g. 4 or 64kB on Linux). A problem with "page sharing" flag is to define which page size it refers to: normal page, some huge page size, SMMU page or something else.

We could extend the above spec text with: "...must not overlap with memory area or memory pages used for any other pool." The same problem of what page size is meant would remain. We could also think that application should know memory mapping rules for this pool memory in this case, when it wants to do the memory management itself.

Memory area alignment requirement would allow implementation to specify the correct page size (from SMMU).

ashwinyes: >> Memory area alignment requirement would allow implementation to specify the correct page size (from SMMU).

Memory alignment would apply to the lowest addressed buffer being populated. Right? This in combination with **buf_size_aligned** flag (which is the case in Marvell implementation) would lead to lot of memory wastage.

Eg: Lets say we want a buffer size of 37KB. But if our system page size is 64KB, then we will have to do either of the following 2 things.
1) Find a page whose address is aligned to 37KB and 64KB to use as the first buffer. For this user will have to allocate a memory chunk of at least 37*64 KB to find such an address.
2) Give buffer size as 64KB but use only 37KB out of it.

Just another thought. May be add two additional parameters mem_area_start and mem_area_size which defines the range of buffer addresses that will be populated into the pool. Then add the 'Memory area alignment' capability which will apply to both mem_area_start and mem_area_size instead of the lowest addressed buffer.

include/odp/api/spec/pool.h
```diff
@@ -173,6 +173,107 @@ int odp_pool_stats(odp_pool_t pool, odp_pool_stats_t *stats);
...
+ *         buf[N] --->  +-------------------------------+    --                     --
+ *                      |                               |     |                      |
+ *                      | ODP header (optional)         |      > odp_header_size     |
+ *                      |                               |     |                      |
+ *                      +-------------------------------+    --                      |
+ *                      |                               |     |                      |
+ *                      | Application header (optional) |      > app_header_size     |
+ *                      |                               |     |                       > buf_size
+ *                      +-------------------------------+    --                      |
+ * odp_packet_head()--> |                               |                            |
```
ashwinyes, line 250: Minor comment. odp_packet_head() should be in line 249.

include/odp/api/spec/pool.h
psavol: Actually, I did not mean alignment of the first buffer, but alignment of the memory area used for all the buffers: "minimum alignment of the start of the memory area used for an external memory pool".

Spec text could be formulated like this:
"The memory area used for one pool, staring from (or before) the lowest addressed buffer and 
extending to the end (or after) of the highest addressed buffer, must not overlap with the memory 
area or memory pages used for any other pool."

In your example, min_mem_align would be 64kB, and application would use that e.g. in odp_shm_reserve() call or reserve that much extra for start alignment.

include/odp/api/spec/pool.h
psavol: To me +-----------+ line does not contain any bytes, it's just a line in the drawing dividing e.g. header fields, or header from payload. See e.g. how RFCs document headers in ascii graphics https://datatracker.ietf.org/doc/html/rfc791

In the figure, odp_packet_head() points now to the first byte after application header.

ashwinyes: Okay.

In that case does **buf[N] --->** also need to change ?

include/odp/api/spec/pool.h
ashwinyes: Okay.

If we define memory area as above and say that min_mem_align should apply to memory area, then I think it should be fine.

include/odp/api/spec/pool.h
psavol: Yes, I'll move that.

include/odp/api/spec/pool.h
psavol: Fixed in v14 with new memory area definition text, and added min_mem_align capa.

ashwinyes:
@psavol v14 looks good to go

platform/linux-generic/odp_pool.c
```diff
@@ -1537,3 +1552,242 @@ int odp_buffer_is_valid(odp_buffer_t buf)
...
+}
+
+static int check_pool_ext_param(const odp_pool_ext_param_t *param)
+{
+	uint32_t head_offset = sizeof(odp_packet_hdr_t) + param->pkt.app_header_size;
+
+	if (param->type != ODP_POOL_PACKET)
+		return -1;
+
+	if (param->cache_size > CONFIG_POOL_CACHE_MAX_SIZE) {
```
MatiasElo, line 1613: Could you check the parameters against `odp_pool_ext_capability_t` (like it is done in `check_params()`)? This would ease code maintenance if/when the capability values are modified.

test/validation/api/pool/pool.c
```diff
@@ -1175,11 +1534,26 @@ odp_testinfo_t pool_suite[] = {
...
 
+odp_testinfo_t pool_ext_suite[] = {
+	ODP_TEST_INFO(test_packet_pool_ext_capa),
+	ODP_TEST_INFO(test_packet_pool_ext_create),
+	ODP_TEST_INFO(test_packet_pool_ext_lookup),
+	ODP_TEST_INFO(test_packet_pool_ext_info),
+	ODP_TEST_INFO(test_packet_pool_ext_populate),
+	ODP_TEST_INFO(test_packet_pool_ext_alloc),
+	ODP_TEST_INFO(test_packet_pool_ext_alloc_max),
+	ODP_TEST_INFO(test_packet_pool_ext_alloc_seg),
```
MatiasElo, line 1666: External memory pool tests should be conditional.

test/validation/api/pool/pool.c
```diff
@@ -1121,6 +1130,356 @@ static void pool_test_timeout_pool_statistics(void)
...
+	uint32_t hr, tr, uarea_size, max_payload, buf_data_size;
+	int num_seg;
+	void *buf[EXT_NUM_BUF];
+	odp_packet_t pkt[EXT_NUM_BUF];
+	uint32_t seg_len = 0;
+
+	CU_ASSERT_FATAL(odp_pool_ext_capability(ODP_POOL_PACKET, &capa) == 0);
+
+	if (capa.pkt.max_segs_per_pkt < 2)
+		return;
```
MatiasElo, line 1385: It would be easier to notice from the test output log that segmentation is not supported if `test_packet_pool_ext_alloc_seg()` test would be conditional. Also, now it seems `PKT_LEN_NORMAL` and `PKT_LEN_MAX` tests are also skipped. Is this correct?

test/validation/api/pool/pool.c
```diff
@@ -1121,6 +1130,356 @@ static void pool_test_timeout_pool_statistics(void)
...
+		pkt_len = (buf_data_size - headroom) / 2;
+	} else if (len_test == PKT_LEN_MAX) {
+		pkt_len = max_payload;
+	} else {
+		 /* length that results 2 segments */
+		pkt_len = max_payload + (buf_size / 2);
+		num_seg = 2;
+		num_pkt = num_buf / num_seg;
+	}
+
```
MatiasElo, line 1417: It would be good to check that the ODP implementation doesn't touch to the application header area.

include/odp/api/spec/pool.h
```diff
@@ -173,6 +173,108 @@ int odp_pool_stats(odp_pool_t pool, odp_pool_stats_t *stats);
...
+ * Populate external memory pool with buffer memory
+ *
+ * Populate can be called multiple times to add memory buffers into the pool. Application must
+ * populate the pool with the exact number of buffers specified in pool parameters. The pool is
+ * ready to be used for allocations only after all populate calls have returned successfully.
+ * Application marks the last populate call with ODP_POOL_POPULATE_DONE flag.
+ *
+ * Depending on pool usage (and ODP implementation), the memory may need to be accessible by
+ * HW accelerators. Application may use e.g. odp_shm_reserve() with ODP_SHM_HW_ACCESS flag to
+ * ensure HW access. The memory area used for one pool, staring from (or before) the lowest
```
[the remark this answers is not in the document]
psavol: fixed in v15

include/odp/api/spec/pool_types.h
```diff
@@ -506,6 +506,203 @@ typedef struct odp_pool_param_t {
...
+		 *  Application must reserve this many bytes from the end of a packet buffer
+		 *  for ODP implementation usage. When the value is zero, ODP implementation does
+		 *  not need trailer space to be reserved for it. Application will not modify this
+		 *  memory area (after buffer populate call).
+		 */
+		uint32_t odp_trailer_size;
+
+		/** Minimum packet pool memory area alignment in bytes
+		 *
+		 *  The memory area used for a packet pool, staring from (or before) the lowest
```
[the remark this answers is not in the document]
psavol: fixed in v15

platform/linux-generic/odp_packet.c
```diff
@@ -2950,3 +2950,154 @@ odp_packet_reass_partial_state(odp_packet_t pkt, odp_packet_t frags[],
 	(void)res;
 	return -ENOTSUP;
 }
+
+void *odp_packet_buf_head(odp_packet_buf_t pkt_buf)
+{
+	odp_packet_hdr_t *pkt_hdr = (odp_packet_hdr_t *)(uintptr_t)pkt_buf;
```
MatiasElo, line 2956: Using a wrapper function for these casts would make the code cleaner.
E.g. `packet_buf_hdr()`

platform/linux-generic/odp_packet.c
```diff
@@ -2950,3 +2950,154 @@ odp_packet_reass_partial_state(odp_packet_t pkt, odp_packet_t frags[],
...
+
+	return (uintptr_t)pkt_hdr->seg_data - (uintptr_t)odp_packet_buf_head(pkt_buf);
+}
+
+uint32_t odp_packet_buf_data_len(odp_packet_buf_t pkt_buf)
+{
+	odp_packet_hdr_t *pkt_hdr = (odp_packet_hdr_t *)(uintptr_t)pkt_buf;
+
+	return pkt_hdr->seg_len;
+}
```
MatiasElo, line 2989: Some of these smaller functions would be rather easy to inline but this can be done in a separate PR.

test/validation/api/pool/pool.c
```diff
@@ -1480,6 +1480,127 @@ static void test_packet_pool_ext_alloc_seg(void)
...
+	uint32_t pkt_len, head_offset, trailer_size, headroom, max_headroom;
+	uint32_t hr, max_payload, buf_data_size;
+	uint32_t num_seg;
+	void *buf[EXT_NUM_BUF];
+	odp_packet_t pkt_tbl[EXT_NUM_BUF];
+
+	CU_ASSERT_FATAL(odp_pool_ext_capability(ODP_POOL_PACKET, &capa) == 0);
+
+	if (capa.pkt.max_segs_per_pkt < 2)
+		return;
```
MatiasElo, line 1499: The test could be made conditional based on this.

test/validation/api/pool/pool.c
psavol: Fixed in v15. All test cases, except the first capa test, are conditional.

test/validation/api/pool/pool.c
```diff
@@ -19,6 +21,13 @@
 #define CACHE_SIZE 32
 #define MAX_NUM_DEFAULT (10 * 1024 * 1024)
 
+#define EXT_NUM_BUF    10
+#define EXT_BUF_SIZE   2048
+#define EXT_BUF_ALIGN  64
+#define EXT_HDR_SIZE   128
```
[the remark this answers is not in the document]
psavol: Fixed in v15

platform/linux-generic/odp_pool.c
psavol: Fixed in v15

test/validation/api/pool/pool.c
psavol: Fixed in v15. Segmentation testing is conditional.

test/validation/api/pool/pool.c
psavol: Fixed in v15

test/validation/api/pool/pool.c
```diff
@@ -1121,6 +1130,356 @@ static void pool_test_timeout_pool_statistics(void)
...
+
+	if (buf_size > capa.pkt.max_buf_size)
+		buf_size = capa.pkt.max_buf_size;
+
+	if (uarea_size > capa.pkt.max_uarea_size)
+		uarea_size = capa.pkt.max_uarea_size;
+
+	if (headroom > capa.pkt.max_headroom)
+		headroom = capa.pkt.max_headroom;
+
```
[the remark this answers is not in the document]
psavol: It's better to leave cache size to default on basic test cases. New param init test case reads cache_size param.

test/validation/api/pool/pool.c
```diff
@@ -1121,6 +1130,356 @@ static void pool_test_timeout_pool_statistics(void)
...
+	odp_pool_type_t type = ODP_POOL_PACKET;
+	uint32_t num_buf = EXT_NUM_BUF;
+	uint32_t buf_size = EXT_BUF_SIZE;
+	uint32_t uarea_size = EXT_UAREA_SIZE;
+	uint32_t headroom = EXT_HEADROOM;
+	uint32_t app_hdr_size = EXT_HDR_SIZE;
+
+	CU_ASSERT_FATAL(odp_pool_ext_capability(type, &capa) == 0);
+
+	odp_pool_ext_param_init(type, param);
```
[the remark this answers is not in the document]
psavol: Fixed in v15. Added new test case for it, and changed API to define 0 as default value for app_hdr_size and uarea_size. I didn't change headroom documentation as it would differ from normal pool (implementation) then.

test/validation/api/pool/pool.c
```diff
@@ -1121,6 +1130,356 @@ static void pool_test_timeout_pool_statistics(void)
...
+static void test_packet_pool_ext_capa(void)
+{
+	odp_pool_ext_capability_t capa;
+	odp_pool_type_t type = ODP_POOL_PACKET;
+
+	CU_ASSERT_FATAL(odp_pool_ext_capability(type, &capa) == 0);
+
+	CU_ASSERT(capa.type == type);
+	CU_ASSERT(capa.max_pools > 0);
+	CU_ASSERT(capa.pkt.max_num_buf > 0);
```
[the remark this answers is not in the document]
psavol: Fixed in v15

test/validation/api/pool/pool.c
psavol: added in v15

platform/linux-generic/odp_packet.c
psavol: using packet_buf_to_hdr() in v15

platform/linux-generic/odp_packet.c
```diff
@@ -2950,3 +2950,154 @@ odp_packet_reass_partial_state(odp_packet_t pkt, odp_packet_t frags[],
...
 	return -ENOTSUP;
 }
+
+void *odp_packet_buf_head(odp_packet_buf_t pkt_buf)
+{
+	odp_packet_hdr_t *pkt_hdr = (odp_packet_hdr_t *)(uintptr_t)pkt_buf;
+	pool_t *pool = pkt_hdr->buf_hdr.pool_ptr;
+	uint32_t head_offset = sizeof(odp_packet_hdr_t) + pool->ext_param.pkt.app_header_size;
+
+	if (pool->pool_ext == 0) {
```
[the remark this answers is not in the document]
psavol: Fixed in v15

psavol:
v15: Fixed review comments by Matias. Also, added default values (0) to odp_pool_ext_param_t app_header_size and uarea_size field specification.

MatiasElo (approved):

PR #1261, opened 2021-04-26
[PATCH v8] linux-gen: ipsec: add anti-replay larger window sizes support.

mrchalla:
Patch series adds support for larger anti-replay window sizes.
The maximum supported AR window size is limitted to 4096, where as
the minimum supported AR window size is remained the same 32.

Will send separate PR for performance related reference odp_ipsec application
changes.

### DIFF
--- a/platform/linux-generic/include/odp_ipsec_internal.h
+++ b/platform/linux-generic/include/odp_ipsec_internal.h
@@ -84,8 +84,27 @@ int _odp_ipsec_status_send(odp_queue_t queue,
 
 #define IPSEC_MAX_SALT_LEN	4    /**< Maximum salt length in bytes */
 
-/* 32 is minimum required by the standard. We do not support more */
-#define IPSEC_ANTIREPLAY_WS	32
+/* The minimum supported AR window size */
+#define IPSEC_AR_WIN_SIZE_MIN	32
+
+/* The maximum supported AR window size */
+#define IPSEC_AR_WIN_SIZE_MAX	4096
+
+/* For a 64-bit bucket size */
+#define IPSEC_AR_WIN_BUCKET_BITS	6
+#define IPSEC_AR_WIN_BUCKET_SIZE	(1 << IPSEC_AR_WIN_BUCKET_BITS)
+#define IPSEC_AR_WIN_BITLOC_MASK	(IPSEC_AR_WIN_BUCKET_SIZE - 1)
+
+/*
+ * We need one extra bucket in addition to the buckets that contain
+ * part of the window.
+ */
+#define IPSEC_AR_WIN_NUM_BUCKETS(window_size)	\
+	(((window_size) - 1) / IPSEC_AR_WIN_BUCKET_SIZE + 2)
+
+/* Maximum number of buckets */
+#define IPSEC_AR_WIN_BUCKET_MAX		\
+	IPSEC_AR_WIN_NUM_BUCKETS(IPSEC_AR_WIN_SIZE_MAX)
 
 struct ipsec_sa_s {
 	odp_atomic_u32_t state ODP_ALIGNED_CACHE;
@@ -101,7 +120,14 @@ struct ipsec_sa_s {
 
 		union {
 			struct {
-				odp_atomic_u64_t antireplay;
+				/* AR window lock */
+				odp_spinlock_t lock;
+
+				/* AR window top sequence number */
+				odp_atomic_u64_t wintop_seq;
+
+				/* AR window bucket array */
+				uint64_t bucket_arr[IPSEC_AR_WIN_BUCKET_MAX];
 			} in;
 
 			struct {
@@ -162,6 +188,16 @@ struct ipsec_sa_s {
 	union {
 		struct {
 			odp_ipsec_ip_version_t lookup_ver;
+
+			/* Anti-replay window management. */
+			struct {
+				/* Number of buckets for AR window */
+				uint16_t num_buckets;
+
+				/* AR window size  */
+				uint32_t win_size;
+			} ar;
+
 			union {
 				odp_u32be_t	lookup_dst_ipv4;
 				uint8_t lookup_dst_ipv6[_ODP_IPV6ADDR_LEN];
--- a/platform/linux-generic/odp_ipsec_sad.c
+++ b/platform/linux-generic/odp_ipsec_sad.c
@@ -21,6 +21,7 @@
 #include <odp/api/plat/cpu_inlines.h>
 
 #include <string.h>
+#include <inttypes.h>
 
 /*
  * SA state consists of state value in the high order bits of ipsec_sa_t::state
@@ -446,6 +447,37 @@ static uint32_t esp_block_len_to_mask(uint32_t block_len)
 	return block_len - 1;
 }
 
+/* AR window management initialization */
+static int ipsec_antireplay_init(ipsec_sa_t *ipsec_sa,
+				 const odp_ipsec_sa_param_t *param)
+{
+	uint16_t num_bkts = 0;
+
+	if (param->inbound.antireplay_ws > IPSEC_AR_WIN_SIZE_MAX) {
+		ODP_ERR("Anti-replay window size %" PRIu32 " is not supported.\n",
+			param->inbound.antireplay_ws);
+		return -1;
+	}
+
+	ipsec_sa->antireplay = (param->inbound.antireplay_ws != 0);
+	if (!ipsec_sa->antireplay)
+		return 0;
+
+	ipsec_sa->in.ar.win_size = param->inbound.antireplay_ws;
+	/* Window size should be at least IPSEC_AR_WIN_SIZE_MIN */
+	if (ipsec_sa->in.ar.win_size < IPSEC_AR_WIN_SIZE_MIN)
+		ipsec_sa->in.ar.win_size = IPSEC_AR_WIN_SIZE_MIN;
+
+	num_bkts = IPSEC_AR_WIN_NUM_BUCKETS(ipsec_sa->in.ar.win_size);
+	ipsec_sa->in.ar.num_buckets = num_bkts;
+	odp_atomic_init_u64(&ipsec_sa->hot.in.wintop_seq, 0);
+	memset(ipsec_sa->hot.in.bucket_arr, 0, sizeof(uint64_t) * num_bkts);
+
+	odp_spinlock_init(&ipsec_sa->hot.in.lock);
+
+	return 0;
+}
+
 odp_ipsec_sa_t odp_ipsec_sa_create(const odp_ipsec_sa_param_t *param)
 {
 	ipsec_sa_t *ipsec_sa;
@@ -488,10 +520,8 @@ odp_ipsec_sa_t odp_ipsec_sa_create(const odp_ipsec_sa_param_t *param)
 				       sizeof(ipsec_sa->in.lookup_dst_ipv6));
 		}
 
-		if (param->inbound.antireplay_ws > IPSEC_ANTIREPLAY_WS)
+		if (ipsec_antireplay_init(ipsec_sa, param))
 			goto error;
-		ipsec_sa->antireplay = (param->inbound.antireplay_ws != 0);
-		odp_atomic_init_u64(&ipsec_sa->hot.in.antireplay, 0);
 	} else {
 		ipsec_sa->lookup_mode = ODP_IPSEC_LOOKUP_DISABLED;
 		odp_atomic_init_u64(&ipsec_sa->hot.out.seq, 1);
@@ -909,10 +939,9 @@ int _odp_ipsec_sa_lifetime_update(ipsec_sa_t *ipsec_sa, uint32_t len,
 
 static uint64_t ipsec_sa_antireplay_max_seq(ipsec_sa_t *ipsec_sa)
 {
-	uint64_t state, max_seq;
+	uint64_t max_seq = 0;
 
-	state = odp_atomic_load_u64(&ipsec_sa->hot.in.antireplay);
-	max_seq = state & 0xffffffff;
+	max_seq = odp_atomic_load_u64(&ipsec_sa->hot.in.wintop_seq) & 0xffffffff;
 
 	return max_seq;
 }
@@ -921,9 +950,8 @@ int _odp_ipsec_sa_replay_precheck(ipsec_sa_t *ipsec_sa, uint32_t seq,
 				  odp_ipsec_op_status_t *status)
 {
 	/* Try to be as quick as possible, we will discard packets later */
-	if (ipsec_sa->antireplay &&
-	    seq + IPSEC_ANTIREPLAY_WS <=
-	    (odp_atomic_load_u64(&ipsec_sa->hot.in.antireplay) & 0xffffffff)) {
+	if (ipsec_sa->antireplay && ((seq + ipsec_sa->in.ar.win_size) <=
+	    (odp_atomic_load_u64(&ipsec_sa->hot.in.wintop_seq) & 0xffffffff))) {
 		status->error.antireplay = 1;
 		return -1;
 	}
@@ -931,22 +959,73 @@ int _odp_ipsec_sa_replay_precheck(ipsec_sa_t *ipsec_sa, uint32_t seq,
 	return 0;
 }
 
-int _odp_ipsec_sa_replay_update(ipsec_sa_t *ipsec_sa, uint32_t seq,
-				odp_ipsec_op_status_t *status)
+static inline int ipsec_wslarge_replay_update(ipsec_sa_t *ipsec_sa, uint32_t seq,
+					      odp_ipsec_op_status_t *status)
 {
-	int cas = 0;
+	uint32_t bucket, wintop_bucket, new_bucket;
+	uint32_t bkt_diff, bkt_cnt, top_seq;
+	uint64_t bit = 0;
+
+	odp_spinlock_lock(&ipsec_sa->hot.in.lock);
+
+	top_seq = odp_atomic_load_u64(&ipsec_sa->hot.in.wintop_seq);
+	if ((seq + ipsec_sa->in.ar.win_size) <= top_seq)
+		goto ar_err;
+
+	bucket = (seq >> IPSEC_AR_WIN_BUCKET_BITS);
+
+	/* Check if the seq is within the range */
+	if (seq > top_seq) {
+		wintop_bucket =	top_seq >> IPSEC_AR_WIN_BUCKET_BITS;
+		bkt_diff = bucket - wintop_bucket;
+
+		/* Seq is way after the range of AR window size */
+		if (bkt_diff > ipsec_sa->in.ar.num_buckets)
+			bkt_diff = ipsec_sa->in.ar.num_buckets;
+
+		for (bkt_cnt = 0; bkt_cnt < bkt_diff; bkt_cnt++) {
+			new_bucket = (bkt_cnt + wintop_bucket + 1) %
+				     ipsec_sa->in.ar.num_buckets;
+			ipsec_sa->hot.in.bucket_arr[new_bucket] = 0;
+		}
+
+		/* AR window top sequence number */
+		odp_atomic_store_u64(&ipsec_sa->hot.in.wintop_seq, seq);
+	}
+
+	bucket %= ipsec_sa->in.ar.num_buckets;
+	bit = (uint64_t)1 << (seq & IPSEC_AR_WIN_BITLOC_MASK);
+
+	/* Already seen the packet, discard it */
+	if (ipsec_sa->hot.in.bucket_arr[bucket] & bit)
+		goto ar_err;
+
+	/* Packet is new, mark it as seen */
+	ipsec_sa->hot.in.bucket_arr[bucket] |= bit;
+	odp_spinlock_unlock(&ipsec_sa->hot.in.lock);
+
+	return 0;
+ar_err:
+	status->error.antireplay = 1;
+	odp_spinlock_unlock(&ipsec_sa->hot.in.lock);
+	return -1;
+}
+
+static inline int ipsec_ws32_replay_update(ipsec_sa_t *ipsec_sa, uint32_t seq,
+					   odp_ipsec_op_status_t *status)
+{
 	uint64_t state, new_state;
+	int cas = 0;
 
-	state = odp_atomic_load_u64(&ipsec_sa->hot.in.antireplay);
-
+	state = odp_atomic_load_u64(&ipsec_sa->hot.in.wintop_seq);
 	while (0 == cas) {
 		uint32_t max_seq = state & 0xffffffff;
 		uint32_t mask = state >> 32;
 
-		if (seq + IPSEC_ANTIREPLAY_WS <= max_seq) {
+		if (seq + IPSEC_AR_WIN_SIZE_MIN <= max_seq) {
 			status->error.antireplay = 1;
 			return -1;
-		} else if (seq >= max_seq + IPSEC_ANTIREPLAY_WS) {
+		} else if (seq >= max_seq + IPSEC_AR_WIN_SIZE_MIN) {
 			mask = 1;
 			max_seq = seq;
 		} else if (seq > max_seq) {
@@ -961,12 +1040,24 @@ int _odp_ipsec_sa_replay_update(ipsec_sa_t *ipsec_sa, uint32_t seq,
 		}
 
 		new_state = (((uint64_t)mask) << 32) | max_seq;
-
-		cas = odp_atomic_cas_acq_rel_u64(&ipsec_sa->hot.in.antireplay,
+		cas = odp_atomic_cas_acq_rel_u64(&ipsec_sa->hot.in.wintop_seq,
 						 &state, new_state);
 	}
-
 	return 0;
+}
+
+int _odp_ipsec_sa_replay_update(ipsec_sa_t *ipsec_sa, uint32_t seq,
+				odp_ipsec_op_status_t *status)
+{
+	int ret;
+
+	/* Window update for ws equal to 32 */
+	if (ipsec_sa->in.ar.win_size == IPSEC_AR_WIN_SIZE_MIN)
+		ret = ipsec_ws32_replay_update(ipsec_sa, seq, status);
+	else
+		ret = ipsec_wslarge_replay_update(ipsec_sa, seq, status);
+
+	return ret;
 }
 
 uint16_t _odp_ipsec_sa_alloc_ipv4_id(ipsec_sa_t *ipsec_sa)
@@ -1084,7 +1175,7 @@ static void ipsec_in_sa_info(ipsec_sa_t *ipsec_sa, odp_ipsec_sa_info_t *sa_info)
 	sa_info->param.inbound.lookup_param.dst_addr = dst;
 
 	if (ipsec_sa->antireplay) {
-		sa_info->inbound.antireplay_ws = IPSEC_ANTIREPLAY_WS;
+		sa_info->inbound.antireplay_ws = ipsec_sa->in.ar.win_size;
 		sa_info->inbound.antireplay_window_top =
 			ipsec_sa_antireplay_max_seq(ipsec_sa);
 	}
--- a/platform/linux-generic/odp_ipsec.c
+++ b/platform/linux-generic/odp_ipsec.c
@@ -158,7 +158,7 @@ int odp_ipsec_capability(odp_ipsec_capability_t *capa)
 
 	capa->max_num_sa = _odp_ipsec_max_num_sa();
 
-	capa->max_antireplay_ws = IPSEC_ANTIREPLAY_WS;
+	capa->max_antireplay_ws = IPSEC_AR_WIN_SIZE_MAX;
 
 	rc = set_ipsec_crypto_capa(capa);
 	if (rc < 0)


JannePeltonen:
General comments:

The anti-replay check does not quite work but may let replayed packets through. _odp_ipsec_sa_replay_update() does not drop a packet if its sequence number is below the anti-replay window but instead it calculates a bucket and bit location that actually corresponds to a later sequence number. If that later genuine packet has not arrived yet, the replayed packet is accepted and later the genuine packet gets dropped.

The code goes through the trouble of rounding the window to a power of two, for apparently no reason. Everything would work as well without the rounding, as far as I can see. However, the number of buckets is never a power of two, resulting in non-constant integer divisions. It would make more sense to round the number of buckets than the size of the window.

The code seems inefficient in the typical case where the received sequence number is higher than the top of the window but still within the same bucket. The only thing really needed would be adjustment of the window top, but now the next bucket is also cleared, which is an unnecessary write to a potentially different cache line that may be owned by another core.

The existing anti-replay implementation for 32 packet window may be somewhat faster than this one since it performs essentially just an atomic read and an atomic CAS instead of the spinlocking and all the logic that this version has. Maybe it would make sense to retain that code and use the simpler mechanism when window size is 32.

platform/linux-generic/include/odp_ipsec_internal.h
```diff
@@ -83,8 +83,22 @@ int _odp_ipsec_status_send(odp_queue_t queue,
 
 #define IPSEC_MAX_SALT_LEN	4    /**< Maximum salt length in bytes */
 
-/* 32 is minimum required by the standard. We do not support more */
-#define IPSEC_ANTIREPLAY_WS	32
+/* The minimum supported AR window size, we do not support less */
+#define IPSEC_ARWIN_SIZE_MIN	32
```
JannePeltonen, line 87: I am not sure if it is immediately clear what ARWIN means. I wonder if AR_WIN or just WIN would be better.

platform/linux-generic/include/odp_ipsec_internal.h
```diff
@@ -113,6 +128,21 @@ struct ipsec_sa_s {
...
+
+		/* AR window size  */
+		uint32_t win_size;
+
+		/* Number of buckets in AR window */
+		uint16_t nb_bucket;
+
+		/* AR window lock */
+		odp_spinlock_t lock;
+	} ar;
```
JannePeltonen, line 144: This (or at least the spin lock and the buckets) should be placed in the hot.in structure as these are frequently modified by multiple threads. See the comment in the code.

platform/linux-generic/odp_ipsec_sad.c
```diff
@@ -423,6 +423,42 @@ static uint32_t esp_block_len_to_mask(uint32_t block_len)
 	return block_len - 1;
 }
 
+/* Return supported power-of-two AR window size */
+static uint32_t ipsec_antireplay_ws(uint32_t win_size)
```
JannePeltonen, line 427: This function (which does not work for all possible values of win_size) seems unnecessary as the ROUNDUP_POWER2_U32() macro of odp_align_internal.h could be used.

platform/linux-generic/odp_ipsec_sad.c
```diff
@@ -423,6 +423,42 @@ static uint32_t esp_block_len_to_mask(uint32_t block_len)
...
+		ipsec_antireplay_ws(param->inbound.antireplay_ws);
+
+	num_bkts = (ipsec_sa->ar.win_size / IPSEC_ARWIN_BUCKET_SIZE) + 1;
+	if (num_bkts < IPSEC_ARWIN_BUCKET_MIN)
+		num_bkts = IPSEC_ARWIN_BUCKET_MIN;
+
+	ipsec_sa->ar.nb_bucket = num_bkts;
+	ipsec_sa->hot.in.wintop_seq = 0;
+	memset(ipsec_sa->ar.bucket_arr, 0,
+	       sizeof(uint64_t) * IPSEC_ARWIN_BUCKET_MAX);
```
JannePeltonen, line 457: This could clear only the part of the array actually used.

platform/linux-generic/odp_ipsec_sad.c
```diff
@@ -461,10 +497,17 @@ odp_ipsec_sa_t odp_ipsec_sa_create(const odp_ipsec_sa_param_t *param)
...
+		     (param->inbound.antireplay_ws > IPSEC_ARWIN_SIZE_MAX)) {
+			ODP_ERR("AR window size[%u] is not supported\n",
+				param->inbound.antireplay_ws);
 			goto error;
+		}
+
 		ipsec_sa->antireplay = (param->inbound.antireplay_ws != 0);
-		odp_atomic_init_u64(&ipsec_sa->hot.in.antireplay, 0);
+		if (ipsec_sa->antireplay)
+			ipsec_antireplay_init(ipsec_sa, param);
```
JannePeltonen, line 510: Now that ipsec_antireplay_init() is being introduced, all anti-replay related initialization could move there, not just part of it. So the checking of the window size parameter as well as the conditional setting of the ipsec_sa->antireplay flag could move there.

I also do not see why SA creation should fail when the window parameter is below the min size. It may actually be against the API spec to not accept small window values. The implementation could just round the window size up as it does in many other cases.

platform/linux-generic/odp_ipsec_sad.c
```diff
@@ -893,9 +935,8 @@ int _odp_ipsec_sa_replay_precheck(ipsec_sa_t *ipsec_sa, uint32_t seq,
 				  odp_ipsec_op_status_t *status)
 {
 	/* Try to be as quick as possible, we will discard packets later */
-	if (ipsec_sa->antireplay &&
-	    seq + IPSEC_ANTIREPLAY_WS <=
-	    (odp_atomic_load_u64(&ipsec_sa->hot.in.antireplay) & 0xffffffff)) {
+	if ((ipsec_sa->antireplay) && ((seq + ipsec_sa->ar.win_size) <=
+	    ipsec_sa->hot.in.wintop_seq)) {
```
JannePeltonen, line 939: Reading ipsec_sa->hot.in.wintop_seq is a data race since window update running in another thread may be writing it at the same time. Using an atomic variable and atomic read would avoid the data race.

platform/linux-generic/include/odp_ipsec_internal.h
mrchalla: ACK. Will change ARWIN to AR_WIN for better readability.

platform/linux-generic/include/odp_ipsec_internal.h
mrchalla: ACK. Will move full or part of this to hot.in.

platform/linux-generic/odp_ipsec_sad.c
mrchalla: We tried to avoid this function for rounding up window size to the next power of two, but couldn't find an ODP API that you suggested or similar to this, will check this API and try to avoid using this function.

platform/linux-generic/odp_ipsec_sad.c
mrchalla: ACK, will make a change to only clear the required number of buckets(ipsec_sa->ar.nb_bucket) for that particular window size.

platform/linux-generic/odp_ipsec_sad.c
mrchalla: ACK.
1. Will move all the anti-replay related initialization inside "ipsec_antireplay_init()" function.
2. Will also handle the case of window size < 32, will round it to 32 instead of returning fail.

platform/linux-generic/odp_ipsec_sad.c
mrchalla: ACK, will check atomic and also will check the usage of existing spinlock.

mrchalla:
Please find below response to some of your questions.

>  _odp_ipsec_sa_replay_update() does not drop a packet if its sequence number is below the anti-replay window but instead it calculates a bucket and bit location that actually corresponds to a later sequence number. 

[Mahipal] : These packets won't reach till "_odp_ipsec_sa_replay_update()" function, instead these will be dropped by "_odp_ipsec_sa_replay_precheck()"  function early

> The anti-replay check does not quite work but may let replayed packets through.

[Mahipal] : Ran validation tests for verifying the detection and dropping of replayed packets.
1. the existing "ipsec_sync" validation test covers the AR packet replay test case, which is passed. 
2. Also the submitted reference test PR [ https://github.com/OpenDataPlane/odp/pull/1264 ] covered a test case for replayed packets. Just sharing the log for window size of 4096.
---------------------------------------------------
AR_WS: 4096
TEST1: seq[8192] = WS * 2
TEST2: seq[4097] = WS + 1
TEST3: seq[4096] = WS
TEST4: seq[6144] = WS + WS/2
TEST5: replay pkt with seq[6144] = WS + WS/2
----------------------------------------------------
3. Also please suggest us some test cases that you feel should be covered.

> The code seems inefficient in the typical case where the received sequence number is higher than the top of the window but still within the same bucket. The only thing really needed would be adjustment of the window top, but now the next bucket is also cleared, which is an unnecessary write to a potentially different cache line that may be owned by another core.

[Mahipal] : For the same bucket case, only top of window will be updated and the corresponding bit is set in the bucket( if the packet is not a replayed packet). New bucket will not be cleared in this case.

The new bucket will be cleared if the received packet sequence number is higher than the top of the window and if it doesn't fall in the current bucket.

This test case is verified with existing "ipsec_sync"  and inbound benchmarking reference "odp_ipsec" test, will submit you the corresponding logs. 

So , in this case, the code seems to do exactly what is expected to be done.

> The existing anti-replay implementation for 32 packet window may be somewhat faster than this one since it performs essentially just an atomic read and an atomic CAS instead of the spinlocking and all the logic that this version has. Maybe it would make sense to retain that code and use the simpler mechanism when window size is 32.

[Mahipal] : the benchmarking data of the existing AR window size 32 and new AR implementation window size 32 are approximately same(very close and comparable). 

Existing ws32  |  New ws32
--------------- | --------------------
  65104           |   65178
192960           | 193603
431853           | 432750
631173           | 631647
652869           | 653146
-------------------------------------
We also shared more benchmarking data in the email, the corresponding reference PR for benchmarking is "https://github.com/OpenDataPlane/odp/pull/1262"

JannePeltonen:
> > _odp_ipsec_sa_replay_update() does not drop a packet if its sequence number is below the anti-replay window but instead it calculates a bucket and bit location that actually corresponds to a later sequence number.
> [Mahipal] : These packets won't reach till "_odp_ipsec_sa_replay_update()" function, instead these will be dropped by "_odp_ipsec_sa_replay_precheck()" function early

That is flawed reasoning. Because this is multithreaded code, _odp_ipsec_sa_replay_precheck() does not necessarily see the same anti-replay state as _odp_ipsec_sa_replay_update(). A packet may be within the window when inspected by the precheck phase but it may fall behind and be below the window when it reaches _odp_ipsec_sa_replay_update(), with the consequences I mentioned. The only way you could deduce something about _odp_ipsec_sa_replay_update() based on what happens in _odp_ipsec_sa_replay_precheck() would be to have some sort or locking preventing any other thread touching the SA in between.

> > The anti-replay check does not quite work but may let replayed packets through.
> [Mahipal] : Ran validation tests for verifying the detection and dropping of replayed packets.

That does not prove much. The tests cannot have 100 % coverage, especially for this short of timing dependent issues.

JannePeltonen:
> [Mahipal] : For the same bucket case, only top of window will be updated and the corresponding bit is set in the bucket( if the packet is not a replayed packet). New bucket will not be cleared in this case.

Sorry, I misread the code. But there seems to be a bug in this line (which would be cleared if written using a simple if-statement):

bkt_diff = (bkt_diff > 1) ? (bkt_diff - 1) : bkt_diff;

If the bucket difference is 2 or higher, then the top most bucket will not be cleared as it should, causing false positives in the anti-replay check. Or am I still misreading something?

JannePeltonen:
> [Mahipal] : the benchmarking data of the existing AR window size 32 and new AR implementation window size 32 are approximately same(very close and comparable).

That only means there is no difference in that particular benchmark.

JannePeltonen:
This PR now seems to have a new version that
- Does not fix or change anything about v2
- Adds a new commit that contains broken ESN support

The ESN support does not work correctly because it does not include the high order bits of the ESN in the ICV calculation. Please check RFC 4303, sections 2 and 2.2.1 and table 1.

mrchalla:
Thanks Janne.

> Does not fix or change anything about v2
Yes, AR related changes are not done, working on them, will submit this week.

> This PR now seems to have a new version that
Since the ESN support has dependency on AR feature, submitted ESN support in the same PR as a new commit.
If this is not the correct way to submit the ESN support, please suggest a correct way to submit the next version of ESN. 

> Adds a new commit that contains broken ESN support
> The ESN support does not work correctly because it does not include the high order bits of the ESN in the ICV calculation. Please check RFC 4303, sections 2 and 2.2.1 and table 1.

ACK, will check fix it in the next version.

mrchalla:
> > > _odp_ipsec_sa_replay_update() does not drop a packet if its sequence number is below the anti-replay window but instead it calculates a bucket and bit location that actually corresponds to a later sequence number.
> 
> > [Mahipal] : These packets won't reach till "_odp_ipsec_sa_replay_update()" function, instead these will be dropped by "_odp_ipsec_sa_replay_precheck()" function early
> 
> That is flawed reasoning. Because this is multithreaded code, _odp_ipsec_sa_replay_precheck() does not necessarily see the same anti-replay state as _odp_ipsec_sa_replay_update(). A packet may be within the window when inspected by the precheck phase but it may fall behind and be below the window when it reaches _odp_ipsec_sa_replay_update(), with the consequences I mentioned. The only way you could deduce something about _odp_ipsec_sa_replay_update() based on what happens in _odp_ipsec_sa_replay_precheck() would be to have some sort or locking preventing any other thread touching the SA in between.

[Mahipal]: Ack, understood the issue you mentioned. Will make appropriate changes ( most likely it may make sense to have something similar to _odp_ipsec_sa_replay_precheck().

> 
> > > The anti-replay check does not quite work but may let replayed packets through.
> 
> > [Mahipal] : Ran validation tests for verifying the detection and dropping of replayed packets.
> 
> That does not prove much. The tests cannot have 100 % coverage, especially for this short of timing dependent issues.

mrchalla:
> > [Mahipal] : For the same bucket case, only top of window will be updated and the corresponding bit is set in the bucket( if the packet is not a replayed packet). New bucket will not be cleared in this case.
> 
> Sorry, I misread the code. But there seems to be a bug in this line (which would be cleared if written using a simple if-statement):
> 
> bkt_diff = (bkt_diff > 1) ? (bkt_diff - 1) : bkt_diff;
> 
> If the bucket difference is 2 or higher, then the top most bucket will not be cleared as it should, causing false positives in the anti-replay check. Or am I still misreading something?

[Mahipal]: Okay, let me try to make it more clear.
if ( bkt_diff  > 1)
     bkt_diff = bkt_diff - 1;
This is required to limit the new bucket clearing till the top most bucket because the new_bucket finding logic contains "+1".  
```
for (bkt_cnt = 0; bkt_cnt < bkt_diff; bkt_cnt++) {
                        new_bucket = (bkt_cnt + wintop_bucket + 1) %
                                ipsec_sa->ar.nb_bucket;
                        ipsec_sa->ar.bucket_arr[new_bucket] = 0;
                }
```

mrchalla:
> The existing anti-replay implementation for 32 packet window may be somewhat faster than this one since it performs essentially just an atomic read and an atomic CAS instead of the spinlocking and all the logic that this version has. Maybe it would make sense to retain that code and use the simpler mechanism when window size is 32. 

Okay, we agree to this proposal of retaining the existing anti-replay code for window size 32.  The new anti-replay code will only be used for rest of the larger window size from 64 to 4096.

JannePeltonen:
> if ( bkt_diff > 1)
> bkt_diff = bkt_diff - 1;
> This is required to limit the new bucket clearing till the top most bucket because the new_bucket finding logic contains "+1".

I do not understand what you are saying. If bucket-wintop_bucket is 2, you need two new buckets that you need to clear. But the subtraction makes bkt_diff 1, so the clearing is done for one bucket only. Am I missing something?

> bkt_diff = (bkt_diff > 1) ? (bkt_diff - 1) : bkt_diff;

Please get rid of the conditional operator. It just make the code less readable than a simple if statement, for no reason. This would be much better (if it were needed, but to me this whole adjustment is just a bug):

if (bkt_diff > 1)
        bkt_diff--;

mrchalla:
> > if ( bkt_diff > 1)
> > bkt_diff = bkt_diff - 1;
> > This is required to limit the new bucket clearing till the top most bucket because the new_bucket finding logic contains "+1".
> 
> I do not understand what you are saying. If bucket-wintop_bucket is 2, you need two new buckets that you need to clear. But the subtraction makes bkt_diff 1, so the clearing is done for one bucket only. Am I missing something?
> 
> > bkt_diff = (bkt_diff > 1) ? (bkt_diff - 1) : bkt_diff;
> 
> Please get rid of the conditional operator. It just make the code less readable than a simple if statement, for no reason. This would be much better (if it were needed, but to me this whole adjustment is just a bug):
> 
> if (bkt_diff > 1)
> bkt_diff--;

We will run more relevant tests to see if there is a bug in this logic. If its required to be retained this logic, will remove the conditional operator logic and use the simple if () statements.

platform/linux-generic/odp_ipsec_sad.c
```diff
@@ -423,6 +423,39 @@ static uint32_t esp_block_len_to_mask(uint32_t block_len)
...
+	if (param->inbound.antireplay_ws > IPSEC_AR_WIN_SIZE_MAX) {
+		ODP_ERR("AR_WS[%u] not supported\n", param->inbound.antireplay_ws);
+		return -1;
+	}
+
+	ipsec_sa->antireplay = (param->inbound.antireplay_ws != 0);
+	if (!ipsec_sa->antireplay)
+		return 0;
+
+	ipsec_sa->hot.in.win_size = ROUNDUP_POWER2_U32(param->inbound.antireplay_ws);
```
JannePeltonen, line 441: What is the purpose of this rounding? As I commented about an earlier version of this PR, as far as I can see everything would work fine without this rounding. However, with suitable rounding the number of buckets could be made a power of two so that figuring out the right bucket would not need an integer division but just a shift.

platform/linux-generic/odp_ipsec_sad.c
```diff
@@ -906,41 +935,91 @@ int _odp_ipsec_sa_replay_precheck(ipsec_sa_t *ipsec_sa, uint32_t seq,
...
+				ipsec_sa->hot.in.nb_bucket;
+			ipsec_sa->hot.in.bucket_arr[new_bucket] = 0;
 		}
 
-		new_state = (((uint64_t)mask) << 32) | max_seq;
+		/* AR window top sequence number */
+		odp_atomic_store_u64(&ipsec_sa->hot.in.wintop_seq, seq);
+	}
+
+	bucket %= ipsec_sa->hot.in.nb_bucket;
```
JannePeltonen, line 1009: This could be replaced with a shift if nb_bucket was always a power of to.

platform/linux-generic/odp_ipsec_sad.c
```diff
@@ -906,41 +935,91 @@ int _odp_ipsec_sa_replay_precheck(ipsec_sa_t *ipsec_sa, uint32_t seq,
...
 
 	if (!ipsec_sa->antireplay)
 		return 0;
 
-	state = odp_atomic_load_u64(&ipsec_sa->hot.in.antireplay);
+	if ((seq + ipsec_sa->hot.in.win_size) <=
+	    (odp_atomic_load_u64(&ipsec_sa->hot.in.wintop_seq) & 0xffffffff)) {
+		status->error.antireplay = 1;
+		return -1;
+	}
```
JannePeltonen, line 949: The checking here does not fix the race condition that I commented about. A packet may pass this test but fall below the window before the actual window update part later in this function. Also, this check is done also for the 32-packet window code path even though it is not needed in that case (because the check is done in the existing code).

platform/linux-generic/odp_ipsec_sad.c
```diff
@@ -906,41 +935,91 @@ int _odp_ipsec_sa_replay_precheck(ipsec_sa_t *ipsec_sa, uint32_t seq,
...
+
+	/* Window update for ws equal to 32 */
+	if (ipsec_sa->hot.in.win_size == IPSEC_AR_WIN_SIZE_MIN) {
```
JannePeltonen, line 952: I think it would be good to separate these two totally different anti-replay codes in to their own functions to improve the clarity of the code.

mrchalla: Ack, will separate out into two different functions, thinking of the below two

-  ipsec_ws32_replay_update()  -- for WS 32 case
-  ipsec_largews_replay_update()  -- for all other window sizes till 4096.

platform/linux-generic/odp_ipsec_sad.c
mrchalla: > What is the purpose of this rounding?

As part of the NOKIA requirement, only the AR window sizes 32, 64, 128, 256, 512, 1024, 2048 and 4096 are supposed to be implemented, hence rounding the AR window sizes to the next power of 2 and figuring out the minimum required  bit buckets from the rounded window sizes. 
 
>However, with suitable rounding the number of buckets could be made a power of two so that figuring out the right bucket would not need an integer division but just a shift.

Yes agreed, integer division would be avoided, but unfortunately could not make number of buckets as power of two, because for each WS one extra bucket is required to avoid the loss of very first bucket bits inside the window(to avoid history loss) during the bucket recycling process. 

===================
WS       |   Minimum buckets
==== ==============
64        |        2 
128      |        3 
256      |        5 
512      |        9 
1026    |       17
2048    |       33
4096    |       65
==================

One solution to solve the integer division issue is to make the effective window size implemented as application WS minus 64, this way the buckets can be rounded to the power of two. With this solution 
============================================
Application WS    |    Implemented WS     |       buckets
============================================
64                         |        64                        |           2     
128                       |        64                        |           2 
256                       |      192   ( 256 -64)     |           4
512                       |       448  (512 -64)      |           8
1024                     |       960  (1024 -64)    |          16 
2048                     |      1984 ( 2048 -64)   |          32 
4096                     |       4032( 4086 -64)   |          64
=============================================


> as far as I can see everything would work fine without this rounding.

Yes, we agree that this implementation would work for all the window sizes till 4096. So if you like to make it generic to work for all the window sizes, please confirm so that we will work on any required changes.

platform/linux-generic/odp_ipsec_sad.c
mrchalla: Correct, this check is not required for WS 32 case, will correct it.

platform/linux-generic/odp_ipsec_sad.c
JannePeltonen: > As part of the NOKIA requirement, only the AR window sizes 32, 64, 128, 256, 512, 1024, 2048 and 4096 are supposed to be implemented, hence rounding the AR window sizes to the next power of 2

Discussion of possible requirements agreed between you and Nokia does not belong here.

That said, if there is a requirement to support certain window sizes, it does not mean that other sizes must not be supported. Linux-generic ODP can support things that are not included in some random requirement list somewhere. In this case there is no point having additional code (the rounding stuff) whose only purpose is to limit the functionality if the implementation does not take any advantage of the limitation and would work equally well without the limitation.

mrchalla: > > As part of the NOKIA requirement, only the AR window sizes 32, 64, 128, 256, 512, 1024, 2048 and 4096 are supposed to be implemented, hence rounding the AR window sizes to the next power of 2
> 
> Discussion of possible requirements agreed between you and Nokia does not belong here.
> 
> That said, if there is a requirement to support certain window sizes, it does not mean that other sizes must not be supported. Linux-generic ODP can support things that are not included in some random requirement list somewhere. In this case there is no point having additional code (the rounding stuff) whose only purpose is to limit the functionality if the implementation does not take any advantage of the limitation and would work equally well without the limitation.

Okay, will remove the window size rounding, this means that
 

- All the application window sizes ranging from 33 to 4096 will be considered as they are by the implementation, there will be no adjustments done by the implementation.
- All the application window sizes ranging from 1 to 32 are adjusted to window size 32 by the implementation.

mrchalla:
Updated the PR.

- Added separate functions for window size 32 and for rest for the window sizes
- Removed window size rounding, now its generic to work for all the window sizes.
- Rounded bit buckets to power of two, integer division is avoided.

mrchalla:
> Updated the PR.
> 
> * Added separate functions for window size 32 and for rest for the window sizes
> * Removed window size rounding, now its generic to work for all the window sizes.
> * Rounded bit buckets to power of two, integer division is avoided.

platform/linux-generic/include/odp_ipsec_internal.h
```diff
@@ -101,7 +115,20 @@ struct ipsec_sa_s {
...
 			struct {
-				odp_atomic_u64_t antireplay;
+				/* AR window lock */
+				odp_spinlock_t lock;
+
+				/* Number of buckets in AR window */
+				uint16_t nb_bucket;
+
+				/* AR window size  */
+				uint32_t win_size;
```
JannePeltonen, line 125: It would be better to move nb_bucket and win_size out of the hot section as they stay constant and the hot section may be frequently written to by multiple threads.

platform/linux-generic/include/odp_ipsec_internal.h
```diff
@@ -84,8 +84,22 @@ int _odp_ipsec_status_send(odp_queue_t queue,
 
 #define IPSEC_MAX_SALT_LEN	4    /**< Maximum salt length in bytes */
 
-/* 32 is minimum required by the standard. We do not support more */
-#define IPSEC_ANTIREPLAY_WS	32
+/* The minimum supported AR window size, we do not support less */
+#define IPSEC_AR_WIN_SIZE_MIN	32
+
+/* The maximum supported AR window size, we do not support more */
```
JannePeltonen, line 90: The "we do not support more" and "we do not support less" parts are redundant (these are already described and named as maximum and minimum) and could be removed.

platform/linux-generic/include/odp_ipsec_internal.h
```diff
@@ -84,8 +84,22 @@ int _odp_ipsec_status_send(odp_queue_t queue,
...
+/* For a 64-bit bucket size */
+#define IPSEC_AR_WIN_BUCKET_BITS		6
+#define IPSEC_AR_WIN_BUCKET_SIZE		(1 << IPSEC_AR_WIN_BUCKET_BITS)
+#define IPSEC_AR_WIN_BITLOC_MASK		(IPSEC_AR_WIN_BUCKET_SIZE - 1)
+
+/* Maximum number of buckets */
+#define IPSEC_AR_WIN_BUCKET_MAX		\
+	(IPSEC_AR_WIN_SIZE_MAX / IPSEC_AR_WIN_BUCKET_SIZE + 1)
+/* Minimum number of buckets */
+#define IPSEC_AR_WIN_BUCKET_MIN		2
```
JannePeltonen, line 102: This should not be needed.

platform/linux-generic/odp_ipsec_sad.c
```diff
@@ -446,6 +446,39 @@ static uint32_t esp_block_len_to_mask(uint32_t block_len)
...
+	uint16_t num_bkts = 0;
+
+	if (param->inbound.antireplay_ws > IPSEC_AR_WIN_SIZE_MAX) {
+		ODP_ERR("AR_WS[%u] not supported\n", param->inbound.antireplay_ws);
+		return -1;
+	}
+
+	ipsec_sa->antireplay = (param->inbound.antireplay_ws != 0);
+	if (!ipsec_sa->antireplay)
+		return 0;
```
JannePeltonen, line 462: This check is unnecessary. It has already been done by the caller.

platform/linux-generic/odp_ipsec_sad.c
```diff
@@ -956,14 +1036,29 @@ int _odp_ipsec_sa_replay_update(ipsec_sa_t *ipsec_sa, uint32_t seq,
...
 	return 0;
 }
 
+int _odp_ipsec_sa_replay_update(ipsec_sa_t *ipsec_sa, uint32_t seq,
+				odp_ipsec_op_status_t *status)
+{
+	int ret;
+
+	if (!ipsec_sa->antireplay)
+		return 0;
```
JannePeltonen, line 1051: This check is unnecessary. It has already been done by the caller.

platform/linux-generic/odp_ipsec_sad.c
```diff
@@ -916,32 +946,82 @@ int _odp_ipsec_sa_replay_precheck(ipsec_sa_t *ipsec_sa, uint32_t seq,
...
+	odp_spinlock_lock(&ipsec_sa->hot.in.lock);
+	if ((seq + ipsec_sa->hot.in.win_size) <=
+	    (odp_atomic_load_u64(&ipsec_sa->hot.in.wintop_seq) & 0xffffffff))
+		goto ar_err;
+
+	bucket = (seq >> IPSEC_AR_WIN_BUCKET_BITS);
+
+	/* Check if the seq is within the range */
+	if (seq > odp_atomic_load_u64(&ipsec_sa->hot.in.wintop_seq)) {
+		wintop_bucket =	odp_atomic_load_u64(&ipsec_sa->hot.in.wintop_seq) >>
```
JannePeltonen, line 974: It would be nicer to do the atomic load only once.

platform/linux-generic/odp_ipsec_sad.c
```diff
@@ -916,32 +946,82 @@ int _odp_ipsec_sa_replay_precheck(ipsec_sa_t *ipsec_sa, uint32_t seq,
...
+			bkt_diff = ipsec_sa->hot.in.nb_bucket;
+
+		for (bkt_cnt = 0; bkt_cnt < bkt_diff; bkt_cnt++) {
+			new_bucket = (bkt_cnt + wintop_bucket + 1) &
+				     (ipsec_sa->hot.in.nb_bucket - 1);
+			ipsec_sa->hot.in.bucket_arr[new_bucket] = 0;
+		}
+
+		/* AR window top sequence number */
+		odp_atomic_store_u64(&ipsec_sa->hot.in.wintop_seq, seq);
```
JannePeltonen, line 989: Maybe the logic of this function should be driven even more by bucket index than the sequence number so that wintop_seq (or top_bucket or whatever it would be called), would always correspond to the topmost sequence number of the topmost bucket that is in use. In other words, once a new bucket is taken into use, we could just start using all the bits as if the topmost sequence number was already received.

Currently, if packets are processed in the sequence number order (as is quite likely), every packet would have to update wintop_seq (which may not be cheap due to sharing of the cache line with other cores) and run through the logic associated with wintop increase. So instead of doing that 64 times per bucket, we could do it just once per bucket and effectively increase wintop by 64 when a new bucket is taken into use (but then maybe we could just count buckets instead of sequence numbers also for wintop).

platform/linux-generic/odp_ipsec_sad.c
```diff
@@ -446,6 +446,39 @@ static uint32_t esp_block_len_to_mask(uint32_t block_len)
...
+	ipsec_sa->antireplay = (param->inbound.antireplay_ws != 0);
+	if (!ipsec_sa->antireplay)
+		return 0;
+
+	ipsec_sa->hot.in.win_size = param->inbound.antireplay_ws;
+	/* Window size should be at least IPSEC_AR_WIN_SIZE_MIN */
+	if (ipsec_sa->hot.in.win_size < IPSEC_AR_WIN_SIZE_MIN)
+		ipsec_sa->hot.in.win_size = IPSEC_AR_WIN_SIZE_MIN;
+
+	num_bkts = ROUNDUP_POWER2_U32(ipsec_sa->hot.in.win_size) / IPSEC_AR_WIN_BUCKET_SIZE;
```
JannePeltonen, line 469: This is broken in a way that makes it possible for replayed packets to pass the anti-replay check.

There should be at least one extra bucket so that when the window is not aligned with the buckets, the top and bottom bucket do not end up being the same. Now they do, and thus the clearing of the new top bucket also clears the bits that are still used for the bottom of the window.

JannePeltonen, line 469: Since num_bucket gets rounded up like that, the actual anti-replay window could perhaps also be rounded up to the maximum size allowed by the rounded num_bucket value. I do not see a whole lot of benefit in restricting the window to be smaller than what is anyway naturally handled by the checking code without extra effort.

JannePeltonen:
The commit message should be fixed so that it focuses on this commit and makes sense as part of the commit history of the repository. It should not be about the whole patch series or what review comments were taken into account since the previous draft.

platform/linux-generic/odp_ipsec_sad.c
mrchalla: > Since num_bucket gets rounded up like that, the actual anti-replay window could perhaps also be rounded up to the maximum size allowed by the rounded num_bucket value. I do not see a whole lot of benefit in restricting the window to be smaller than what is anyway naturally handled by the checking code without extra effort.

The anti-replay window would not be rounded. 
The rounding of the buckets is as you suggested to avoid the integer division.

mrchalla:
> The commit message should be fixed so that it focuses on this commit and makes sense as part of the commit history of the repository. It should not be about the whole patch series or what review comments were taken into account since the previous draft.

Yes, I will fix this.

platform/linux-generic/odp_ipsec_sad.c
JannePeltonen: I am not saying that the window is rounded but that the code could be changed so that it gets rounded (not to power of two but the largest window allowed by the number of buckets). In other words, the size of the window (and thus packet reordering tolerance) could be rounded up without it causing additional processing cost.

If the window is not rounded, then (depending on window size) you could have up to half of the buckets unused for keeping track of which packets were received even though the status bits are already there and the code traverses and touches the full bucket array anyway.

mrchalla: > I am not saying that the window is rounded but that the code could be changed so that it gets rounded (not to power of two but the largest window allowed by the number of buckets). In other words, the size of the window (and thus packet reordering tolerance) could be rounded up without it causing additional processing cost.

Yes, understood, can be done. This will also avoid the corner case problem of top new bucket clearing the bottom bucket. 

I am trying to give a example to align my understanding with you.

Lets say, user passed window size of 130, and the buckets rounded value is 4 in this case. So we can safely increase the window size internally to 192(64 * num_buckets - 64 = 64*4 -64 =  192).

Hope we are aligned now.

platform/linux-generic/include/odp_ipsec_internal.h
mrchalla: ack.

platform/linux-generic/include/odp_ipsec_internal.h
mrchalla: ack, will remove these comments.

platform/linux-generic/include/odp_ipsec_internal.h
mrchalla: Ack, can be avoided.

platform/linux-generic/odp_ipsec_sad.c
JannePeltonen: > This will also avoid the corner case problem of top new bucket clearing the bottom bucket.

No, that is a separate issue. It is a bug (and not a corner case problem but a fatal bug) in calculation of the needed number of buckets. One the number of minimum number of buckets for a requested window size is known (and hopefully calculated correctly), we can increase the actual window size to what can be handled using that number of buckets.

> Lets say, user passed window size of 130, and the buckets rounded value is 4 in this case. So we can safely increase the window size internally to 192(64 * num_buckets - 64 = 64*4 -64 = 192).

Yes.

mrchalla: > > This will also avoid the corner case problem of top new bucket clearing the bottom bucket.
> 
> No, that is a separate issue. It is a bug (and not a corner case problem but a fatal bug) in calculation of the needed number of buckets. One the number of minimum number of buckets for a requested window size is known (and hopefully calculated correctly), we can increase the actual window size to what can be handled using that number of buckets.

Okay, understood. So this means that in the case of the application passed window sizes from 4033 to 4096, the supported window size internally is 4032(4096 - 64), and the number of rounded buckets are 64( the maximum supported by the implementation).

JannePeltonen: An ODP implementation is allowed to increase the actual window size from what was requested by the application but it is not allowed to decrease it. So, if the implementation advertises max window of 4096 packets in IPsec capabilities, it is not allowed to round it down to something smaller. Thus 64 buckets of 64 packets each would not be enough.

mrchalla: So the minimum number of buckets required in this case are 64+1=65. After rounding(bucket rounding is to avoid the integer division during the window update processing) it would be increased to 128. So is it okay to increase the max supported buckets to 128.

JannePeltonen: There is a tradeoff between the potential speedup of getting rid of the division and the drawbacks (reserving more memory & touching more memory) of rounding the number of buckets up. With relatively small window sized the rounding probably makes sense, but with large window sizes it may be counter productive. It is hard to tell which approach is better without suitable measurements.

Maybe a better approach would be to try to make the code so that rounding is not needed but integer division is still avoided. For example, the bucket for a packet that falls within the current window would not be calculated as seqno % nb_bucket but as (seqno - bottom_seqno) >> BUCKET_BITS, where bottom_seqno is updated when the buckets are moved. And similarly for the other cases.

mrchalla: We will consider your suggestion and check this solution for the integer division optimization. So there are now two optimizations that are to be checked.

- Integer division optimization( with out bucket rounding).
- top sequence number update once per bucket( not per every packet).

As these optimizations will take some more time, would you please allow me to take these two optimizations as part of a separate PR, so that I will update this PR with all your other comments/suggestions. Please suggest.

platform/linux-generic/odp_ipsec_sad.c
mrchalla: ack

platform/linux-generic/odp_ipsec_sad.c
mrchalla: Will check and remove it.

platform/linux-generic/odp_ipsec_sad.c
JannePeltonen: Feel free to leave the optimizations out of the next version of this PR. Let's check after that if we want to include some of them.

mrchalla: Thanks Janne.

mrchalla:
Updated the PR with all the comments/suggestions except the optimizations.  

- Now the implementation doesn't support either window rounding or buckets rounding.
- All the application passed window sizes ( from 32 to 4096) are supported as they are in the implementation.
- For each window size the minimum required buckets are used in the implementation(not rounded to the next power of 2).

JannePeltonen:
It seems that the commit message has included my reviewed-by tag since version 5 of this PR even though the code has not passed my review. Please do not add the reviewed-by tag before the review has successfully completed.

The commit message text could also be improved and the grammar fixed, like this:

```
linux-gen: ipsec: add support for larger anti-replay windows

Add support for anti-replay windows of up to 4096 packets. Minimum window
size remains at 32 and uses the existing implementation. New code is used
for larger windows.
```

platform/linux-generic/include/odp_ipsec_internal.h
```diff
@@ -84,8 +84,20 @@ int _odp_ipsec_status_send(odp_queue_t queue,
...
+#define IPSEC_AR_WIN_SIZE_MAX	4096
+
+/* For a 64-bit bucket size */
+#define IPSEC_AR_WIN_BUCKET_BITS	6
+#define IPSEC_AR_WIN_BUCKET_SIZE	(1 << IPSEC_AR_WIN_BUCKET_BITS)
+#define IPSEC_AR_WIN_BITLOC_MASK	(IPSEC_AR_WIN_BUCKET_SIZE - 1)
+
+/* Maximum number of buckets */
+#define IPSEC_AR_WIN_BUCKET_MAX		\
+	(IPSEC_AR_WIN_SIZE_MAX / IPSEC_AR_WIN_BUCKET_SIZE + 1)
```
JannePeltonen, line 100: This assumes that IPSEC_AR_WIN_SIZE_MAX is divisible by IPSEC_AR_WIN_BUCKET_SIZE and can break if someone later does not notice it and changes the max window size.

I would propose defining the following macro and using it to determine the number of buckets for both the maximum number of buckets and the actual number of buckets used for an SA.

```
/*
 * We need one extra bucket in addition to the buckets that contain
 * part of the window.
 */
#define IPSEC_AR_WIN_NUM_BUCKETS(window_size) \
        (((window_size) - 1) / IPSEC_AR_WIN_BUCKET_SIZE + 2)
```

platform/linux-generic/include/odp_ipsec_internal.h
```diff
@@ -114,6 +133,15 @@ struct ipsec_sa_s {
...
 	} hot;
 
+	/* Anti-replay window management. */
+	struct {
+		/* Number of buckets in AR window */
+		uint16_t nb_bucket;
+
+		/* AR window size  */
+		uint32_t win_size;
+	} ar;
```
JannePeltonen, line 143: Since these are needed only for inbound SAs, they could be moved to ipsec_sa_t::in (i.e. in the same struct where lookup_ver is).

I think num_buckets would be a better name than nb_bucket. The comment should not say "in window" since not all the buckets are necessarily inside the window (when the window boundary coincides with a bucket boundary, there is
at least one bucket that is not "in the window"). But "for AR window" would sound ok to me.

platform/linux-generic/odp_ipsec_sad.c
```diff
@@ -446,6 +446,39 @@ static uint32_t esp_block_len_to_mask(uint32_t block_len)
...
+		return 0;
+
+	ipsec_sa->ar.win_size = param->inbound.antireplay_ws;
+	/* Window size should be at least IPSEC_AR_WIN_SIZE_MIN */
+	if (ipsec_sa->ar.win_size < IPSEC_AR_WIN_SIZE_MIN)
+		ipsec_sa->ar.win_size = IPSEC_AR_WIN_SIZE_MIN;
+
+	num_bkts = (ipsec_sa->ar.win_size / IPSEC_AR_WIN_BUCKET_SIZE) + 1;
+	if (ipsec_sa->ar.win_size % IPSEC_AR_WIN_BUCKET_SIZE)
+		num_bkts++;
```
JannePeltonen, line 471: These lines could be replaced with the same macro that I proposed in an earlier comment.

platform/linux-generic/odp_ipsec_sad.c
```diff
@@ -916,32 +946,81 @@ int _odp_ipsec_sa_replay_precheck(ipsec_sa_t *ipsec_sa, uint32_t seq,
...
-int _odp_ipsec_sa_replay_update(ipsec_sa_t *ipsec_sa, uint32_t seq,
-				odp_ipsec_op_status_t *status)
+static inline int ipsec_wslarge_replay_update(ipsec_sa_t *ipsec_sa, uint32_t seq,
+					      odp_ipsec_op_status_t *status)
 {
-	int cas = 0;
-	uint64_t state, new_state;
+	uint32_t bucket, wintop_bucket, new_bucket;
+	uint32_t bkt_diff, bkt_cnt;
+	uint64_t top_seq = 0, bit = 0;
```
JannePeltonen, line 963: Declaring top_seq uint64_t is a bit inconsistent since the code otherwise assumes 32-bit sequence numbers and uses 32 bit variables. So I would either change these to uint32_t too, or change the code to handle 64 bit sequence numbers so that it would not need to be changed when ESN support is added.

platform/linux-generic/odp_ipsec_sad.c
```diff
@@ -916,32 +946,81 @@ int _odp_ipsec_sa_replay_precheck(ipsec_sa_t *ipsec_sa, uint32_t seq,
...
+
+	odp_spinlock_lock(&ipsec_sa->hot.in.lock);
+
+	top_seq = odp_atomic_load_u64(&ipsec_sa->hot.in.wintop_seq);
+	if ((seq + ipsec_sa->ar.win_size) <= (top_seq & 0xffffffff))
```
JannePeltonen, line 968: The anding with UINT32_MAX does not seem to serve any purpose, so it could be removed.

mrchalla:
> It seems that the commit message has included my reviewed-by tag since version 5 of this PR even though the code has not passed my review. Please do not add the reviewed-by tag before the review has successfully completed.
> 
> The commit message text could also be improved and the grammar fixed, like this:
> 
> ```
> linux-gen: ipsec: add support for larger anti-replay windows
> 
> Add support for anti-replay windows of up to 4096 packets. Minimum window
> size remains at 32 and uses the existing implementation. New code is used
> for larger windows.
> ```

Ack.

platform/linux-generic/include/odp_ipsec_internal.h
mrchalla: The macro "IPSEC_AR_WIN_BUCKET_MAX" is defined for bucket array size initialization "uint64_t bucket_arr[IPSEC_AR_WIN_BUCKET_MAX];". So I will define it as 

`#define IPSEC_AR_WIN_BUCKET_MAX    65`

platform/linux-generic/include/odp_ipsec_internal.h
mrchalla: Ack, will make these changes.

platform/linux-generic/odp_ipsec_sad.c
```diff
@@ -446,6 +446,39 @@ static uint32_t esp_block_len_to_mask(uint32_t block_len)
...
 }
 
+/* AR window management initialization */
+static int ipsec_antireplay_init(ipsec_sa_t *ipsec_sa,
+				 const odp_ipsec_sa_param_t *param)
+{
+	uint16_t num_bkts = 0;
+
+	if (param->inbound.antireplay_ws > IPSEC_AR_WIN_SIZE_MAX) {
+		ODP_ERR("AR_WS[%u] not supported\n", param->inbound.antireplay_ws);
```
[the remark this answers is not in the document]
mrchalla: Ack.

platform/linux-generic/odp_ipsec_sad.c
mrchalla: Yes, will use the new proposed macro here.

platform/linux-generic/odp_ipsec_sad.c
mrchalla: Agreed, presently since the sequence numbers are 32 bit, will change this variable to uint32_t.

platform/linux-generic/odp_ipsec_sad.c
mrchalla: Agreed, will remove it also the variable will be changed to 32-bit

platform/linux-generic/include/odp_ipsec_internal.h
JannePeltonen: > So I will define it as #define IPSEC_AR_WIN_BUCKET_MAX 65

Why would you want to do that instead of letting the macro calculate the number of buckets? The magic number 65 would have to be changed when the max window size is changes and getting it wrong may open up a security problem in the implementation. Since the correct number of buckets can be, and needs to be, calculated from the window size, it is better to let the computer, i.e. the macro, do the calculation and thus also make the calculation explicit, instead of hiding it behind the magic number and counting on that the next person changing the code figures it all out.

So, please define IPSEC_AR_WIN_BUCKET_MAX in terms of IPSEC_AR_WIN_SIZE_MAX and IPSEC_AR_WIN_BUCKET_SIZE which it depends on and not through a magic number that comes out of the sky.

mrchalla: Understood it now,  visualized your first comment a bit differently, sorry for that.
Now, the newly proposed macro is used in both the cases as per your first comment.

1) The maximum number of buckets macro is defined as.

```
#define IPSEC_AR_WIN_BUCKET_MAX         \
        IPSEC_AR_WIN_NUM_BUCKETS(IPSEC_AR_WIN_SIZE_MAX)
```

2) The actual number of buckets used for a given window size is determined as

`num_bkts = IPSEC_AR_WIN_NUM_BUCKETS(ipsec_sa->in.ar.win_size);`

mrchalla:
Hi @JannePeltonen, updated the PR with your suggestions.

JannePeltonen (approved):

Please rebase.

mrchalla:
> Reviewed-by: Janne Peltonen [janne.peltonen@nokia.com](mailto:janne.peltonen@nokia.com)
> 
> Please rebase.

Thank you Janne. I will rebase it ASAP.

PR #1263, opened 2021-04-28
[PATCH v5] api: pktio: statistics: add in_packets/out_packets counters

MatiasElo:
V4:
- Clarified the term 'successful' (Yuri)
- Mention Ethernet in `in_ucast_pkts` and `out_ucast_pkts` documentation (Yuri)
- Add mention about pause frames (Yuri)

### DIFF
--- a/include/odp/api/spec/packet_io_stats.h
+++ b/include/odp/api/spec/packet_io_stats.h
@@ -19,6 +19,8 @@
 extern "C" {
 #endif
 
+#include <odp/api/deprecated.h>
+
 /** @addtogroup odp_packet_io
  *  @{
  */
@@ -71,8 +73,10 @@ typedef struct odp_pktio_stats_t {
 	 * because of an unknown or unsupported protocol.  For any interface
 	 * that does not support protocol multiplexing, this counter will always
 	 * be 0. See ifInUnknownProtos in RFC 2863, RFC 3635.
+	 *
+	 * @deprecated This counter has been deprecated.
 	 */
-	uint64_t in_unknown_protos;
+	uint64_t ODP_DEPRECATE(in_unknown_protos);
 
 	/**
 	 * The number of octets transmitted in valid MAC frames on this
--- a/include/odp/api/spec/packet_io_stats.h
+++ b/include/odp/api/spec/packet_io_stats.h
@@ -26,42 +26,28 @@ extern "C" {
  */
 
 /**
- * Packet IO statistics
+ * Packet IO statistics counters
  *
- * Packet IO statistics counters follow RFCs for Management Information Base
- * (MIB)for use with network management protocols in the Internet community:
- * https://tools.ietf.org/html/rfc3635
- * https://tools.ietf.org/html/rfc2863
- * https://tools.ietf.org/html/rfc2819
+ * In the counter definitions the term successfully refers to packets which were
+ * not discarded or detected to contain errors by the packet IO interface. In
+ * case of Ethernet, it's implementation specific whether valid pause frames are
+ * included in the counters or not.
  */
 typedef struct odp_pktio_stats_t {
-	/**
-	 * The number of octets in valid MAC frames received on this interface,
-	 * including the MAC header and FCS. See ifHCInOctets counter
-	 * description in RFC 3635 for details.
-	 */
+	/** Number of octets in successfully received packets. In case of
+	 *  Ethernet, packet size includes MAC header and FCS. */
 	uint64_t in_octets;
 
-	/**
-	 * The number of packets, delivered by this sub-layer to a higher
-	 * (sub-)layer, which were not addressed to a multicast or broadcast
-	 * address at this sub-layer. See ifHCInUcastPkts in RFC 2863, RFC 3635.
-	 */
+	/** Number of successfully received Ethernet packets with a unicast
+	 *  destination MAC address. */
 	uint64_t in_ucast_pkts;
 
-	/**
-	 * The number of inbound packets which were chosen to be discarded
-	 * even though no errors had been detected to prevent their being
-	 * deliverable to a higher-layer protocol.  One possible reason for
-	 * discarding such a packet could be to free up buffer space.
-	 * See ifInDiscards in RFC 2863.
-	 */
+	/** Number of inbound packets which were discarded due to a lack of free
+	 *  resources (e.g. buffers) or other reasons than packet errors. */
 	uint64_t in_discards;
 
-	/**
-	 * The sum for this interface of AlignmentErrors, FCSErrors,
-	 * FrameTooLongs, InternalMacReceiveErrors. See ifInErrors in RFC 3635.
-	 */
+	/** Number of inbound packets with errors. Depending on packet input
+	 *  configuration, packets with errors may be dropped or not. */
 	uint64_t in_errors;
 
 	/**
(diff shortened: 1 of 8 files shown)


include/odp/api/spec/packet_io_stats.h
```diff
@@ -26,42 +26,23 @@ extern "C" {
...
+	/** Number of inbound packets which were discarded, even though no
+	 *  errors had been detected. This can be caused by a lack of free
+	 *  resources (e.g. buffers). */
 	uint64_t in_discards;
 
-	/**
-	 * The sum for this interface of AlignmentErrors, FCSErrors,
-	 * FrameTooLongs, InternalMacReceiveErrors. See ifInErrors in RFC 3635.
-	 */
+	/** Number of inbound packets with errors. */
```
psavol, line 48: Clarification may be needed: Depending on packet input configuration, packets with errors may be dropped or not.

include/odp/api/spec/packet_io_stats.h
```diff
@@ -26,42 +26,23 @@ extern "C" {
...
-	 * The number of inbound packets which were chosen to be discarded
-	 * even though no errors had been detected to prevent their being
-	 * deliverable to a higher-layer protocol.  One possible reason for
-	 * discarding such a packet could be to free up buffer space.
-	 * See ifInDiscards in RFC 2863.
-	 */
+	/** Number of inbound packets which were discarded, even though no
+	 *  errors had been detected. This can be caused by a lack of free
+	 *  resources (e.g. buffers). */
 	uint64_t in_discards;
```
psavol, line 46: Clarification about counter not including packets dropped due to a packet error:
 Number of inbound packets which were discarded due to a lack of free resources (e.g. buffers) or other reasons than packet errors.

include/odp/api/spec/packet_io_stats.h
```diff
@@ -26,42 +26,23 @@ extern "C" {
...
+ * Packet IO statistics counters
  */
 typedef struct odp_pktio_stats_t {
-	/**
-	 * The number of octets in valid MAC frames received on this interface,
-	 * including the MAC header and FCS. See ifHCInOctets counter
-	 * description in RFC 3635 for details.
-	 */
+	/** Number of octets in successfully received packets (includes MAC
+	 *  headers and FCSs). */
```
psavol, line 33: Preferring singular: (includes MAC header and FCS)

platform/linux-generic/pktio/loop.c
```diff
@@ -191,7 +191,7 @@ static int loopback_recv(pktio_entry_t *pktio_entry, int index ODP_UNUSED,
 	}
 
 	pktio_entry->s.stats.in_errors += failed;
-	pktio_entry->s.stats.in_ucast_pkts += num_rx - failed;
```
JereLeppanen, line 194: Don't *_ucast_pkts still need to be calculated?

include/odp/api/spec/packet_io_stats.h
matiaselo: Fixed in V2

include/odp/api/spec/packet_io_stats.h
matiaselo: Fixed in V2

include/odp/api/spec/packet_io_stats.h
matiaselo: Fixed in V2

platform/linux-generic/pktio/loop.c
matiaselo: The previous calculation of unicast packets was a bit faulty. The destination MACs were never checked. I'd prefer leaving `in_ucast_pkts` and `out_ucast_pkts` unsupported on loop interface to avoid extra overhead.

JereLeppanen: And pcap as well, right? Okay, sounds reasonable to me.

matiaselo: Yep, same thing.

psavol (approved):

MatiasElo:
Hi @jerinjacobk, is this API change OK for you?

jerinjacobk:
> Hi @jerinjacobk, is this API change OK for you?

Looks good to me.

include/odp/api/spec/packet_io_stats.h
```diff
@@ -26,42 +26,23 @@ extern "C" {
...
  */
 typedef struct odp_pktio_stats_t {
-	/**
-	 * The number of octets in valid MAC frames received on this interface,
-	 * including the MAC header and FCS. See ifHCInOctets counter
-	 * description in RFC 3635 for details.
-	 */
+	/** Number of octets in successfully received packets (includes MAC
+	 *  header and FCS). */
 	uint64_t in_octets;
```
yuri-tolstov, line 34: I think the notion of "successfully" needs to be clarified. E.g. if statistics is gathered on the MAC/hardware level, then successfully received (by hardware) packets may include good and bad packets, those which will be sent to application and dropped. So, successfully may be specified as "non-dropped good and bad", or "non-dropped good (no errors) only", "all, including dropped,  good or bad", etc.

include/odp/api/spec/packet_io_stats.h
```diff
@@ -26,42 +26,23 @@ extern "C" {
...
 
-	/**
-	 * The number of packets, delivered by this sub-layer to a higher
-	 * (sub-)layer, which were not addressed to a multicast or broadcast
-	 * address at this sub-layer. See ifHCInUcastPkts in RFC 2863, RFC 3635.
-	 */
+	/** Number of successfully received packets with a unicast destination
+	 *  MAC address. */
```
yuri-tolstov, line 40: It seems the specification implies the Ethernet packet centric API. In general, there may be other (including custom) packet formats, which may not fit into this definition. Probably, this is not a problem. Just wanted to have clarity on this.

yuri-tolstov, line 40: Do we want to distinguish ucast vs bcast/mcast? From the existing counters, it may be not straightforward to calculate the number of bcast/mcast packets.

include/odp/api/spec/packet_io_stats.h
yuri-tolstov, line 34: From my experience, sometimes control (pause) packets may distort picture of packet rate observed by the hardware, due to they take certain bandwidth. I assume, they are not calculated in the total number of in_packets/in_octets?

yuri-tolstov:
General note: The pktio statistics says: "Counters not supported by the interface are set to zero.". From my point of view, this may case confusions for users, because from the existing API itself it is not possible to distinguish, if counter is really 0 or not implemented. Probably, this may be resolved by adding the related things in the pktio capability structure.

MatiasElo:
> General note: The pktio statistics says: "Counters not supported by the interface are set to zero.". From my point of view, this may case confusions for users, because from the existing API itself it is not possible to distinguish, if counter is really 0 or not implemented. Probably, this may be resolved by adding the related things in the pktio capability structure.

I'd agree that the cleanest solution would be to add a bitfield of supported counters to pktio capability in the same manner as for pool statistics. Now the question is, should this be done now or in a follow-up PR?

include/odp/api/spec/packet_io_stats.h
MatiasElo: I'll add a mention of Ethernet here to make the spec clearer.

MatiasElo: Since we already have a counter for unicast packets, adding counters for multicast and broadcast packets would be quite logical.

include/odp/api/spec/packet_io_stats.h
MatiasElo: Updated in V4

include/odp/api/spec/packet_io_stats.h
MatiasElo: The number of multicast/broadcast packets can now be calculated from: `in_packets` - `in_ucast_pkts`

We can add the specific counter in a separate PR.

include/odp/api/spec/packet_io_stats.h
MatiasElo: Update in V4

MatiasElo:
@yuri-tolstov @jerinjacobk Is the updated API OK for you?

I'm planning to submit a second PR later on, which will add APIs for capabilities, platform specific counters, and multicast/broadcast counters. The main target for this PR was to add the missing `in_packets/out_packets` counters.

yuri-tolstov:
Matias, the updated API looks fine.

PR #1264, opened 2021-04-28
test: validation: ipsec: add antireplay tests for larger window sizes

mrchalla:
A reference validation test for exercising more antireplay tests
required for validating the AR larger window sizes(PR#1261).

### DIFF
--- a/test/validation/api/ipsec/ipsec.c
+++ b/test/validation/api/ipsec/ipsec.c
@@ -937,6 +937,73 @@ void ipsec_check_out_one(const ipsec_test_part *part, odp_ipsec_sa_t sa)
 		odp_packet_free(pkto[i]);
 }
 
+static void ipsec_pkt_proto_err_set(odp_packet_t pkt)
+{
+	uint32_t l3_off = odp_packet_l3_offset(pkt);
+	odph_ipv4hdr_t ip;
+
+	/* Simulate proto error by corrupting protocol field */
+
+	odp_packet_copy_to_mem(pkt, l3_off, sizeof(ip), &ip);
+
+	if (ip.proto == ODPH_IPPROTO_ESP)
+		ip.proto = ODPH_IPPROTO_AH;
+	else
+		ip.proto = ODPH_IPPROTO_ESP;
+
+	odp_packet_copy_from_mem(pkt, l3_off, sizeof(ip), &ip);
+}
+
+static void ipsec_pkt_auth_err_set(odp_packet_t pkt)
+{
+	uint32_t data, len;
+
+	/* Simulate auth error by corrupting ICV */
+
+	len = odp_packet_len(pkt);
+	odp_packet_copy_to_mem(pkt, len - sizeof(data), sizeof(data), &data);
+	data = ~data;
+	odp_packet_copy_from_mem(pkt, len - sizeof(data), sizeof(data), &data);
+}
+
+void ipsec_check_out_in_one(const ipsec_test_part *part_outbound,
+			    const ipsec_test_part *part_inbound,
+			    odp_ipsec_sa_t sa,
+			    odp_ipsec_sa_t sa_in,
+			    const ipsec_test_flags *flags)
+{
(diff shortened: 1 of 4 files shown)


test/validation/api/ipsec/ipsec.c
```diff
@@ -937,6 +937,73 @@ void ipsec_check_out_one(const ipsec_test_part *part, odp_ipsec_sa_t sa)
 		odp_packet_free(pkto[i]);
 }
 
+static void ipsec_pkt_proto_err_set(odp_packet_t pkt)
```
JannePeltonen, line 940: It would be good to have a separate patch for moving the existing code around. Then the actual change is not mixed with noise and it easier to read the commit later. That said, I am not sure this code should be moved here (as opposed to staying in where it is or moving to a new .c file).

test/validation/api/ipsec/ipsec_test_in.c
```diff
@@ -1727,6 +1727,137 @@ static void test_ipsec_sa_print(void)
...
+	CU_ASSERT_NOT_EQUAL_FATAL(ODP_IPSEC_SA_INVALID, in_sa);
+
+	memset(&sa_info, 0, sizeof(odp_ipsec_sa_info_t));
+	CU_ASSERT_EQUAL_FATAL(0, odp_ipsec_sa_info(in_sa, &sa_info));
+
+	window_sz = sa_info.inbound.antireplay_ws;
+
+	/* 1. Advance the TOP of the window to WS * 2 */
+	sa_param.seq_num =  window_sz * 2;
+	CU_ASSERT_EQUAL(0, odp_ipsec_test_sa_update(out_sa,
```
JannePeltonen, line 1804: Use of odp_ipsec_test_sa_update() is not very nice since it is an optional part of the API and may not be supported by all implementations. Failing a validation test because odp_ipsec_test_sa_update() fails is, strictly speaking, not correct, although with current implementations it would work.

I do not think the use of odp_ipsec_test_sa_update() is necessary here. The test case could just process enough packets using the outbound SA to get the sequence number where it wants it and then after that do the actual test.

test/validation/api/ipsec/ipsec.c
mrchalla: > It would be good to have a separate patch for moving the existing code around. Then the actual change is not mixed with noise and it easier to read the commit later.

Ack.  
 
> I am not sure this code should be moved here (as opposed to staying in where it is or moving to a new .c file).

As we are discussing the code organization in a separate email thread, once decision is made, will plan to make the changes as per that decision.

test/validation/api/ipsec/ipsec_test_in.c
mrchalla: > Use of odp_ipsec_test_sa_update() is not very nice since it is an optional part of the API and may not be supported by all implementations

We have just used this available API for updating the required seq number. Does "optional part of the API" mean, will it be deprecated soon.

> I do not think the use of odp_ipsec_test_sa_update() is necessary here. The test case could just process enough packets using the outbound SA to get the sequence number where it wants it and then after that do the actual test.

Sure, we will explore this thought.

JannePeltonen: > We have just used this available API for updating the required seq number. Does "optional part of the API" mean, will it be deprecated soon.

Optionality means that an ODP implementation is not required to support the API. There is capability info for the sequence number update support and in addition to that the API says that the function is not required to always succeed.

Since this is API validation test code, it is not correct to fail the test (i.e. essentially declare the implementation under test as API-incompliant) just because an optional testing feature is not supported. At minimum the validation test should check the relevant capability and be prepared for the SA update function to fail. Since this would mean that the test case could not be run with all ODP implementations, it would be better if the test case did not rely on the optional feature. Tests with smallish sequence numbers do not really need the SA update feature, but potential future ESN tests probably need it.

JannePeltonen:

PR #1266, opened 2021-04-29
[PATCH v3] validation: packet: fix packet head/tail pull lengths

ashwinyes:
According to the spec packet pull head/tail may pull up to segment
length minus 1 bytes. Fix the tests to align to the spec.


MatiasElo (approved):
The commit message could be cleaned a bit:
```
According to the spec packet pull head/tail may pull up to segment length
minus 1 bytes. Fix the tests to align to the spec.
```

Otherwise looking good.

PR #1267, opened 2021-04-29
[PATCH v4] validation: packet: fix verification for trunc/extend tests

ashwinyes:
In packet test cases for headroom/tailroom, extend head/tail
APIs are always called with extend length greater than
headroom/tailroom. Also trunc head/tail APIs are always
called with truncate length greater than or equal to segment
length.

In these cases, the spec allows the implementation to change
the layout of the packet completely if required. So properties
of the output packet like segment lengths, headroom, tailroom
etc. can be entirely different from those of the input packet
and should not be compared with each other.

Also, comparison of older pointers from input packet with new
pointers from output packet should be done only if the return
value of these APIs is 0 as otherwise the older pointers are to
be considered invalid.

### DIFF
--- a/test/validation/api/packet/packet.c
+++ b/test/validation/api/packet/packet.c
@@ -929,13 +929,7 @@ static void _verify_headroom_shift(odp_packet_t *pkt,
 	CU_ASSERT_PTR_NOT_NULL(data);
 	if (extended) {
 		CU_ASSERT(rc >= 0);
-		if (shift >= 0) {
-			CU_ASSERT(odp_packet_seg_len(*pkt) == shift - room);
-		} else {
-			CU_ASSERT(odp_packet_headroom(*pkt) >=
-				  (uint32_t)abs(shift) - seg_data_len);
-		}
-		CU_ASSERT(odp_packet_head(*pkt) != head_orig);
+		CU_ASSERT(odp_packet_seg_len(*pkt) == seg_len);
 	} else {
 		CU_ASSERT(odp_packet_headroom(*pkt) == room - shift);
 		CU_ASSERT(odp_packet_seg_len(*pkt) == seg_data_len + shift);
@@ -1032,15 +1026,13 @@ static void _verify_tailroom_shift(odp_packet_t *pkt,
 	CU_ASSERT_PTR_NOT_NULL(tail);
 	if (extended) {
 		CU_ASSERT(rc >= 0);
-		CU_ASSERT(odp_packet_last_seg(*pkt) != seg);
-		seg = odp_packet_last_seg(*pkt);
-		if (shift > 0) {
-			CU_ASSERT(odp_packet_seg_data_len(*pkt, seg) ==
-				  shift - room);
+
+		if (shift >= 0) {
+			if (rc == 0)
+				CU_ASSERT(tail == tail_orig);
 		} else {
-			CU_ASSERT(odp_packet_tailroom(*pkt) >=
-				  (uint32_t)abs(shift) - seg_data_len);
-			CU_ASSERT(seg_len == odp_packet_tailroom(*pkt));
+			CU_ASSERT(odp_packet_tail(*pkt) == tail);
+			CU_ASSERT(odp_packet_tailroom(*pkt) == seg_len);
 		}
 	} else {
 		CU_ASSERT(odp_packet_seg_data_len(*pkt, seg) ==
@@ -1049,19 +1041,15 @@ static void _verify_tailroom_shift(odp_packet_t *pkt,
 		if (room == 0 || (room - shift) == 0)
 			return;
 		if (shift >= 0) {
-			CU_ASSERT(odp_packet_tail(*pkt) ==
-				  tail_orig + shift);
+			CU_ASSERT(odp_packet_tail(*pkt) == tail_orig + shift);
+			CU_ASSERT(tail == tail_orig);
 		} else {
+			CU_ASSERT(odp_packet_tail(*pkt) == tail);
 			CU_ASSERT(tail == tail_orig + shift);
 		}
 	}
 
 	CU_ASSERT(odp_packet_len(*pkt) == pkt_data_len + shift);
-	if (shift >= 0) {
-		CU_ASSERT(tail == tail_orig);
-	} else {
-		CU_ASSERT(odp_packet_tail(*pkt) == tail);
-	}
 }
 
 static void packet_test_tailroom(void)


test/validation/api/packet/packet.c
```diff
@@ -929,13 +929,16 @@ static void _verify_headroom_shift(odp_packet_t *pkt,
...
-				  (uint32_t)abs(shift) - seg_data_len);
+		if (rc == 0) {
+			if (shift >= 0) {
+				CU_ASSERT(odp_packet_seg_len(*pkt) ==
+					  shift - room);
+			} else {
+				CU_ASSERT(odp_packet_headroom(*pkt) >=
+					  (uint32_t)abs(shift) - seg_data_len);
+			}
+			CU_ASSERT(odp_packet_head(*pkt) != head_orig);
```
JannePeltonen, line 940: So what exactly is the problem that you are fixing by not executing these lines if the packet changed in the extend or trunc call? These operations are not using the old handle but the new handle returned by the extend/trunk call.

ashwinyes: Spec says. 
```
"A successful operation overwrites the packet handle with a new handle, which
application must use as the reference to the packet instead of the old
handle. Depending on the implementation, the old and new handles may be
equal.
The operation return value indicates if any packet data or metadata (e.g.
user_area) were moved in memory during the operation. If some memory areas
were moved, application must use new packet/segment handles to update
data pointers. Otherwise, all old pointers remain valid.".

``` 
I believe by this it means the packet pointed to by the new handle can have an entirely different layout from the original and it is left to the implementation. So we can't compare any property of the original packet with the new packet.

Assuming my understanding of the spec is correct, 

1. seg_data_len is a property of the original packet. 
If suppose the new implementation simply adds a new segment at the beginning of the packet with the required extend length, it means the new packet can have different segmentation and new headroom need not be equal to (uint32_t)abs(shift) - seg_data_len.

2. head_orig is a property of the original packet. 
If suppose the implementation simply memmove the existing data ahead in the  segment to create more space at the beginning which means we cannot guarantee CU_ASSERT(odp_packet_head(*pkt) != head_orig);

JannePeltonen: > I believe by this it means the packet pointed to by the new handle can have an entirely different layout from the original and it is left to the implementation. So we can't compare any property of the original packet with the new packet.

That is what I think too *). The problem with the code is not that it does not realize that the handle might change but the issue that the packet layout may change. The commit message could be made more clear what the real problem is.

*) There are, of course, some properties that can be compared and that maybe should be checked in the validation code. The old packet data should still be there (even if maybe in a new memory location), packet length should change exactly the requested amount, etc.

ashwinyes: Right. Anything relating to the original packet layout in the memory should not be compared.

Packet length needs to be verified to be orig_pkt_len + extend_len which it is doing and I have not removed that check from the existing code.

Data within the packet also must remain intact. But its not verifying the same in the existing code. So I didn't add that additional verification step.

I hope the change is fine. I will modify the commit message to be more descriptive explaining the issue at hand.

test/validation/api/packet/packet.c
```diff
@@ -929,13 +929,16 @@ static void _verify_headroom_shift(odp_packet_t *pkt,
...
 		CU_ASSERT(rc >= 0);
-		if (shift >= 0) {
-			CU_ASSERT(odp_packet_seg_len(*pkt) == shift - room);
-		} else {
-			CU_ASSERT(odp_packet_headroom(*pkt) >=
-				  (uint32_t)abs(shift) - seg_data_len);
+		if (rc == 0) {
+			if (shift >= 0) {
+				CU_ASSERT(odp_packet_seg_len(*pkt) ==
+					  shift - room);
```
JannePeltonen, line 935: rc should never be zero, so the assert that checks that it is >=0 could just check that rc is strictly positive. The preceding code calls extend_head only when there is not enough headroom in which case either odp_packet_head() must change or data must move. And similarly for the truncation case. We could then get rid of all the code that was made conditional on rc being zero as it is obviously dead code.

Btw, there is a bug in the condition which decides whether pull can be done instead of trunc. Only seg_data_len - 1 bytes may be pulled, not all the data bytes in the segment.

> CU_ASSERT(odp_packet_seg_len(*pkt) == shift - room);

This assert here assumes that a new segment was added in the packet and contains the data bytes that did not fit in the old head room. But if segmentation changes, we do not come here. And in this function, segmentation always changes when extended is true, so this is dead code now.

If the preceding code was different and we could come here without segmentation changing, then this assert would not make any sense at all.

There are similar issues in the tailroom shift code.

ashwinyes: >rc should never be zero

Agree, (rc == 0) only if extend_head becomes equivalent to push_head which will never happen here. Will remove the dead code.

> Btw, there is a bug in the condition which decides whether pull can be done instead of trunc. Only seg_data_len - 1 bytes may be pulled, not all the data bytes in the segment.

This is addressed in #1266

ashwinyes: @JannePeltonen  On another thought, just for extend tail case, I think rc can be 0. Because we can achieve extend tail by conserving all existing pointers same except odp_packet_tail().

odp_packet_tail() cannot be conserved in anyway in any case when doing extend tail.

JannePeltonen: > On another thought, just for extend tail case, I think rc can be 0. Because we can achieve extend tail beyond tailroom by conserving all existing pointers except odp_packet_tail().

I agree. The implementation could just make the segment larger if it had means to do so. But in that case the assert in the rc==0 case is incorrect as it assumes that a new segment was created.

Actually, I think that even trunc_tail might return 0 even if it has to remove the last segment. If the various packet pointers are not within the truncated range, they do not need to change. And if they are in the truncated range, then they should not be dereferenced anymore anyway, even in the pull_tail case, I suppose. I am not sure if the API is clear enough on the exact semantics.

ashwinyes: Thinking in that lines, even extend/trunc head can be done with by making the segment larger. Only odp_packet_head() pointer needs to change there.

JannePeltonen: > Only odp_packet_head() pointer needs to change there.

But zero return code means that all pointers stay valid.

ashwinyes: You are right. 

I was trying to compare/correlate  trunc/extend head to trunc/extend tail semantics.  Like if trunc/extend tail can return 0 with change in odp_packet_tail(), then trunc/extend head should be allowed to return 0 with change in odp_packet_head(). 

But I guess the odp_packet_data() is the pointer which is allowed to change for trunc/extend head rather than odp_packet_head().

JannePeltonen: > rc should never be zero

This is not actually true. We both misunderstood the API. The should probably say that all pointers to packet data bytes remain valid (meaning, still pointing to the same data bytes), not just that "all pointers remain valid" which can be interpreted, as we did, to mean about all pointers (segment pointer, head pointer, tail pointer etc) that appear in the API.

Linux-gen implementation actually never returns a positive return code from the extend/trunc head functions, which is why the latest version of this patch is now failing in CI tests.

ashwinyes: @JannePeltonen  Sorry for the delay in updating this PR. I missed your latest comment on this PR.

> (meaning, still pointing to the same data bytes)

In that case , irrespective of the return value rc, pointers like segment pointer, head pointer, tail pointer etc. are to be considered invalid and must not be compared. Right ?

But internal data pointers like L2 ptr, L3 ptr etc. can be compared if rc == 0.

Is my understanding correct ? 


> Linux-gen implementation actually never returns a positive return code from the extend/trunc head functions, which is why the latest version of this patch is now failing in CI tests.

As an experiment, I tried to fix linux-gen to return rc > 0 in case of change in segmentation, but ran into more failures in concat/split test cases. So didn't go down that path further.

JannePeltonen: > In that case , irrespective of the return value rc, pointers like segment pointer, head pointer, tail pointer etc. are to be considered invalid and must not be compared. Right ?

I think so, although the API is perhaps not crystal clear and easy to read.

JannePeltonen (approved):

PR #1268, opened 2021-04-29
[PATCH v9] helper: cli: add user command functions

JereLeppanen:
```
helper: cli: try to avoid closing reused file descriptors
    
    The only way to stop cli_loop() is to close the socket. But after
    closing, cli_loop() still uses or closes the file descriptor, which is
    a problem if the file descriptor has already been reused.
    
    Try to avoid this problem by using dup2() to switch to a higher file
    descriptor number. Also avoid double closes outside cli_loop() with a
    couple of small tweaks.
    
helper: cli: add user command functions
    
    Allow user to register CLI commands with a name, function, and an
    optional help text.
    
helper: increment library version to 1.0.7
    
    Increment helper library version number due to changes in CLI helper
    API.
    
```

v2:
- Multiple individually registered commands.

v3:
- Fix according to comments from Petri.
- Add explanation of argc and argv in documentation of user command function type.
- Tweak of odph_cli_register_command() documentation.

v4:
- Add commit "helper: cli: try to avoid closing reused file descriptors".
- Add commit "helper: increment library version to 1.0.7".
- helper api: Remove mention about prints to console, now that cli helper directs prints to the cli client.
- helper api: odph_cli_register_command() "should be called after odph_cli_init() and before odph_cli_start()".
- Add implementation, update test and example.

v5:
- odph_cli_log(): Remove unused "level" parameter.

v6:
- Fix clang compilation error.

v7:
- Default case.
- Fix helper API version.

v8:
- Fix checkpatch error.

v9:
- Add reviewed-by tags.

### DIFF
--- a/configure.ac
+++ b/configure.ac
@@ -21,8 +21,8 @@ AC_SUBST(ODP_VERSION_API_MINOR)
 # Helper library version
 ##########################################################################
 m4_define([odph_version_generation], [1])
-m4_define([odph_version_major], [0])
-m4_define([odph_version_minor], [6])
+m4_define([odph_version_major], [1])
+m4_define([odph_version_minor], [0])
 
 m4_define([odph_version],
     [odph_version_generation.odph_version_major.odph_version_minor])
--- a/helper/cli.c
+++ b/helper/cli.c
@@ -16,12 +16,22 @@
 #include <errno.h>
 #include <poll.h>
 #include <stdio.h>
+#include <strings.h>
 
 /* Socketpair socket roles. */
 enum {
 	SP_READ = 0,
 	SP_WRITE = 1,
 };
+
+#define MAX_NAME_LEN 20
+#define MAX_HELP_LEN 100
+
+typedef struct {
+	odph_cli_user_cmd_func_t fn;
+	char name[MAX_NAME_LEN];
+	char help[MAX_HELP_LEN];
+} user_cmd_t;
 
 typedef struct {
 	volatile int cli_fd;
@@ -30,9 +40,15 @@ typedef struct {
 	/* Socketpair descriptors. */
 	int sp[2];
 	int listen_fd;
-	/* This lock guards cli_fd and run, which must be accessed atomically. */
+	/* Guards cli_fd and run, which must be accessed atomically. */
 	odp_spinlock_t lock;
+	odp_spinlock_t api_lock;
 	odph_thread_t thr_server;
+	odp_instance_t instance;
+	struct sockaddr_in addr;
+	uint32_t max_user_commands;
+	uint32_t num_user_commands;
+	user_cmd_t user_cmd[0];
 } cli_shm_t;
 
 static const char *shm_name = "_odp_cli";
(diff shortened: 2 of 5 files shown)


helper/include/odp/helper/cli.h
```diff
@@ -42,6 +45,12 @@ typedef struct {
 	const char *address;
 	/** TCP port. Default is 55555. */
 	uint16_t port;
+	/**
+	 * User defined command function. If set, this function is called when
+	 * the CLI client invokes the command "user". If not set (set to NULL),
+	 * then the CLI command "user" does not exist. Default is NULL.
+	 */
+	odph_cli_user_cmd_func_t user_cmd_fn;
```
MatiasElo, line 53: I'm thinking it would make sense to allow a user to add multiple custom commands.

helper/include/odp/helper/cli.h
```diff
@@ -33,6 +33,9 @@ extern "C" {
  * @{
  */
 
+/** User defined command function type. */
+typedef void (*odph_cli_user_cmd_func_t)(int argc, char *argv[]);
```
MatiasElo, line 37: In addition to the function pointer a user could also pass the CLI command sting.

JereLeppanen: The idea here is that the application can implement whatever command hierarchy it wants with argv[]. The user function can for example treat argv[0] as the name of a command. In comparison to this, having multiple functions and strings would just impose unnecessary limitations.

JereLeppanen: @psavol pointed out in a discussion that independent commands are likely needed by for example EM and application. I'll add support for multiple independently registered user commands.

yuri-tolstov: Jere, if I understand your motivation correctly that you are trying to add an extension to the existing CLI framework? In form of one function and one ("user") command? If existing CLI is missing such capabilities -- to add user command(s)? If so, then, it is probably better (more generic and practical) to make such extension by adding a table of commands (instead of one command) or add multiple commands (one by one) to the command list.

JereLeppanen: Changed in v2.

helper/include/odp/helper/cli.h
```diff
@@ -54,20 +59,55 @@ typedef struct {
...
  * type ODP_THREAD_CONTROL using odp_cpumask_default_control().
  *
- * @param instance ODP instance
- * @param param CLI server parameters to use
  * @retval 0 Success
  * @retval <0 Failure
  */
-int odph_cli_start(const odp_instance_t instance,
-		   const odph_cli_param_t *param);
+int odph_cli_start();
```
psavol, line 110: Prototype needs to be int odph_cli_start(void)

Documentation could repeat shortly here the correct start up order: odph_cli_init() -> optionally add commands -> start()

helper/include/odp/helper/cli.h
```diff
@@ -54,20 +59,55 @@ typedef struct {
...
+ *
+ * This function initializes the CLI helper. It must be called before
+ * odph_cli_register_command() and odph_cli_start().
+ *
+ * @param instance ODP instance
+ * @param param CLI server parameters to use
+ * @retval 0 Success
+ * @retval <0 Failure
+ */
+int odph_cli_init(const odp_instance_t instance, const odph_cli_param_t *param);
```
psavol, line 73: To me "const" is not necessary for instance, as application cannot modify its content anyway.

Similarly, init_local() does not have const:
int odp_init_local(odp_instance_t instance, odp_thread_type_t thr_type)

helper/include/odp/helper/cli.h
JereLeppanen: Fixed in v2.

helper/cli.c
```diff
@@ -16,37 +16,162 @@
...
+	shm->addr.sin_port = htons(param->port);
+
+	switch (inet_pton(AF_INET, param->address, &shm->addr.sin_addr)) {
+	case -1:
+		ODPH_ERR("Error: inet_pton(): %s\n", strerror(errno));
+		return -1;
+	case 0:
+		ODPH_ERR("Error: inet_pton(): illegal address format\n");
+		return -1;
+	}
```
psavol, line 118: Some compilers or static code checkers may complain about a switch statement that does not have a default case. So, I'd suggest to add default, or change this into two if-clauses.

configure.ac
```diff
@@ -22,7 +22,7 @@ AC_SUBST(ODP_VERSION_API_MINOR)
 ##########################################################################
 m4_define([odph_version_generation], [1])
 m4_define([odph_version_major], [0])
-m4_define([odph_version_minor], [6])
+m4_define([odph_version_minor], [7])
```
psavol, line 25: Since changes are not backwards compatible, the new version number should be 1.1.0. Minor number increment indicate backwards compatible changes.

helper/cli.c
JereLeppanen: Added default case in v7.

configure.ac
JereLeppanen: Fixed in v7.

helper/cli.c
psavol: Checkpatch requires break also:

```
default:
        break;
```

JereLeppanen: Fixed in v8.

psavol (approved):

PR #1270, opened 2021-04-30
[PATCH v8] github_ci: enable native arm64 ci jobs

dean-arnold:
To enable the arm64 CI, ci-pipeline-arm64.yaml file has been added
with approximately 40+ jobs. Some of the CI scripts were also
modified to add an ARCH environment variable for allowing native
arm64 builds for the jobs.

Jobs have been restricted to only run when the repo is owned by
OpenDataPlane. This is to prevent the CI triggering on forks of the
project where Graviton2 self-hosted runners are not available.

Signed-off-by: Dean Arnold <dean.arnold@arm.com>

### DIFF
--- a/.github/workflows/ci-pipeline-arm64.yml
+++ b/.github/workflows/ci-pipeline-arm64.yml
@@ -1,0 +1,194 @@
+name: Arm64 CI
+
+# github.repository has been used to ensure CI is only run on the repo where
+# self-hosted runners are installed. This will prevent [self-hosted, ARM64] CI failing on forks
+
+on: [push, pull_request]
+env:
+  ARCH: arm64
+  CC: gcc
+  CONTAINER_NAMESPACE: ghcr.io/opendataplane/odp-docker-images
+  OS: ubuntu_20.04
+
+jobs:
+
+  Build:
+    if: ${{ github.repository == 'OpenDataPlane/odp' }}
+    runs-on: [self-hosted, ARM64]
+    strategy:
+      fail-fast: false
+      matrix:
+        cc: [gcc, clang]
+        conf: ['', 'CFLAGS=-O3', 'CFLAGS=-O1', 'CFLAGS=-O0 --enable-debug=full', '--enable-lto', '--enable-lto --disable-abi-compat', '--enable-pcapng-support']
+        exclude:
+          - cc: clang
+            conf: '--enable-lto'
+          - cc: clang
+            conf: '--enable-lto --disable-abi-compat'
+    steps:
+      - uses: AutoModality/action-clean@v1.1.0
+      - uses: actions/checkout@v2
+      - run: sudo docker run -i -v `pwd`:/odp --privileged --shm-size 8g -e CC="${{matrix.cc}}"
+               -e CONF="${{matrix.conf}}" $CONTAINER_NAMESPACE/odp-ci-${OS}-${ARCH}-native /odp/scripts/ci/build_${ARCH}.sh
+
+  Build_OS:
+    if: ${{ github.repository == 'OpenDataPlane/odp' }}
+    runs-on: [self-hosted, ARM64]
+    strategy:
+      fail-fast: false
+      matrix:
+        cc: [gcc, clang]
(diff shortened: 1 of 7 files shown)


m-govind:
Thanks Dean !!

.github/workflows/ci-pipeline-arm64.yml
```diff
@@ -0,0 +1,194 @@
+name: Arm64-CI
```
MatiasElo, line 1: May be nitpicking, but please remove the dash (`Arm64 CI`), so the workflow naming is consistent in actions tab.

.github/workflows/ci-pipeline-arm64.yml
```diff
@@ -0,0 +1,194 @@
...
+        exclude:
+          - cc: clang
+            conf: '--enable-lto'
+          - cc: clang
+            conf: '--enable-lto --disable-abi-compat'
+    steps:
+      - uses: AutoModality/action-clean@v1.1.0
+      - uses: actions/checkout@v2
+      - run: sudo docker run -i -v `pwd`:/odp --privileged --shm-size 8g -e CC="${{matrix.cc}}"
+               -e CONF="${{matrix.conf}}" $CONTAINER_NAMESPACE/odp-ci-${OS}-${ARCH}-native /odp/scripts/ci/build_${ARCH}.sh
```
MatiasElo, line 32: `build_arm64.sh` ends up calling `build.sh`, which still assumes an x86 host. It always sets `--build=x86_64-linux-gnu` to ODP configure and the dynamically linked test application is not run if `TARGET_ARCH` is set.

MatiasElo (commented):
Mainly minor tweaks required. Great work!

.github/workflows/ci-pipeline-arm64.yml
```diff
@@ -0,0 +1,194 @@
+name: Arm64-CI
+
+# github.repository has been used to ensure CI is only run on the repo where
+# self-hosted runners are installed. This will prevent Arm64 CI failing on forks
```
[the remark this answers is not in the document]
MatiasElo: This comment seems deprecated now.

.github/workflows/ci-pipeline-arm64.yml
```diff
@@ -0,0 +1,182 @@
...
+env:
+  ARCH: arm64
+  CC: gcc
+  CONTAINER_NAMESPACE: ghcr.io/opendataplane/odp-docker-images
+  OS: ubuntu_20.04
+
+jobs:
+
+  Build:
+    runs-on: self-hosted
```
MatiasElo, line 16: One could potentially have runners with multiple architectures, so the following would be better:
`runs-on: [self-hosted, ARM64]`

Same comment for all `runs-on` lines in this file.

scripts/ci/build.sh
```diff
@@ -4,7 +4,7 @@ set -e
 cd "$(dirname "$0")"/../..
 ./bootstrap
 ./configure \
-	--host=${TARGET_ARCH} --build=x86_64-linux-gnu \
+	--host=${TARGET_ARCH} --build=${BUILD_ARCH:-x86_64-linux-gnu} \
```
MatiasElo, line 7: The `if [ -z "$TARGET_ARCH" ]` line below still prevents running the dynamically linked example application on arm64 hosts.

You could add a ${BUILD_ARCH} = ${TARGET_ARCH} comparison to fix this.

MatiasElo:
Couple remaining comments. While you fix those please squash the changes into a single commit. You can also add my reviewed-by tag to the commit.

.github/workflows/ci-pipeline-arm64.yml
dean-arnold: As discussed on our call, I have added the repository checks back in now. I tried removing them to see jobs were skipped in the event of no self-hosted runners being present, but unfortunately this just leads to timeout and failure.

MatiasElo (approved):

MatiasElo:
The PR is now ready. Thank you Dean for the great work on the new ARM CI!

We'll first have to merge couple API PRs before this one goes in.

PR #1271, opened 2021-05-03
[PATCH v3] github_ci: run x86_64 tests with debug build too

JannePeltonen:
Run x86_64 tests with debug build too to catch issues that e.g. trigger
asserts in the ODP implementation. Add debug build flag in the existing
--enable-deprecated test configuration to avoid increasing the number
of test runs.

The coverage test does already use debug build, but this change makes
the debug build explicitly part of the common test run and makes
possible error reports hopefully more clear.


MatiasElo (approved):

JannePeltonen:
v2: Rebased and added testing with debug build in the ARM workflow too, as discussed.

PR #1272, opened 2021-05-05
[PATCH v3] api: ipsec: include success bytes in stats

asasidharan:
Update IPsec stats with "success bytes" representing total bytes of data processed in success cases.

### DIFF
--- a/include/odp/api/spec/ipsec.h
+++ b/include/odp/api/spec/ipsec.h
@@ -937,6 +937,15 @@ typedef struct odp_ipsec_stats_t {
 
 	/** Number of packets with hard lifetime(packets) expired */
 	uint64_t hard_exp_pkts_err;
+
+	/** Total bytes of packet data processed by IPsec SA in success cases
+	 *
+	 * The range of packet bytes included in the success_bytes count is
+	 * implementation defined but includes at least the bytes input for
+	 * encryption or bytes output after decryption in ESP or the bytes
+	 * authenticated in AH.
+	 */
+	uint64_t success_bytes;
 } odp_ipsec_stats_t;
 
 /**
--- a/test/validation/api/ipsec/ipsec_test_out.c
+++ b/test/validation/api/ipsec/ipsec_test_out.c
@@ -361,17 +361,19 @@ static void test_ipsec_stats_zero_assert(odp_ipsec_stats_t *stats)
 	CU_ASSERT_EQUAL(stats->mtu_err, 0);
 	CU_ASSERT_EQUAL(stats->hard_exp_bytes_err, 0);
 	CU_ASSERT_EQUAL(stats->hard_exp_pkts_err, 0);
+	CU_ASSERT_EQUAL(stats->success_bytes, 0);
 }
 
 static void test_ipsec_stats_test_assert(odp_ipsec_stats_t *stats,
-					 enum ipsec_test_stats test)
+					 enum ipsec_test_stats test,
+					 uint64_t succ_bytes)
 {
 	if (test == IPSEC_TEST_STATS_SUCCESS) {
-		/* Braces needed by CU macro */
 		CU_ASSERT_EQUAL(stats->success, 1);
+		CU_ASSERT(stats->success_bytes >= succ_bytes);
 	} else {
-		/* Braces needed by CU macro */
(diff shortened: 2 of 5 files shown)


include/odp/api/spec/ipsec.h
```diff
@@ -937,6 +937,9 @@ typedef struct odp_ipsec_stats_t {
 
 	/** Number of packets with hard lifetime(packets) expired */
 	uint64_t hard_exp_pkts_err;
+
+	/** Bytes of data processed successfully */
+	uint64_t success_bytes;
```
JannePeltonen, line 942: I wonder if this is too vague. The "of data" part sounds like it really means something, although I am not sure what that would be.

Maybe this could say: "The packet bytes included in the count are the same as in the byte based lifetime calculation." But it would not be much better since lifetime config does not specify the exact bytes either (and maybe rightly so). And maybe some implementations cannot easily make these two byte counts the same.

Or perhaps this could say: "The range of packet bytes included in the count is implementation defined but includes at least the bytes encrypted or decrypted in ESP or the bytes authenticated in AH". This would appear to match the linux-gen implementation.

Or maybe simply: "The range of packet bytes included in the count is implementation defined."

Depending on if and how the API text gets adjusted, the validation test could be made to check more than just that the count is greater than zero.

JannePeltonen:
Apart from the API issue I commented, this PR looks ok to me.

include/odp/api/spec/ipsec.h
asasidharan: Let me know if the following text looks fine:

/** Total bytes of packet data processed by IPsec SA  in success cases
 *
 * The range of packet bytes included in the success_bytes count is implementation defined but includes
 * at least the bytes input for encryption or bytes output after decryption in ESP or the bytes authenticated in AH.
 */

Will update the validation test to check accordingly in the next version.

JannePeltonen: How about: ...at least the encrypted bytes in ESP and the authenticated bytes in AH.

asasidharan: The wording "at least encrypted bytes" would mean that implementation would need to account for padded bytes. This may not be supported on all platforms.

My intention was to specify the minimum value as the data range in the plain packet being encrypted in case of encryption. In case of decryption, the final output data range which will be placed in the resulting packet. This would keep the padding out of the equation.

JannePeltonen: I see. What about TFC padding in that case? Undefined whether it gets included at all)? Maybe that would be fine and if you think so too, I am ok with the wording you suggested. In that case you can already add my review tag in the next version.

asasidharan: I think better to leave the TFC padding part unspecified, as it would allow flexibility to the implementations. And we are only looking at defining the minimum value to be counted.

> I am ok with the wording you suggested. In that case you can already add my review tag in the next version.

@JannePeltonen shall I add reviewed-by tag to linux-generic commit as well?

JannePeltonen: > shall I add reviewed-by tag to linux-generic commit as well?

Yes, you can do so.

asasidharan:
Updated v2 of the PR with:
* updated API text for success_bytes.
* added reviewed-by tags to API and linux-generic commits
* updated validation test to check if the success bytes is getting incremented properly.

test/validation/api/ipsec/ipsec_test_out.c
```diff
@@ -588,6 +591,24 @@ static void test_out_in_common(const ipsec_test_flags *flags,
 		test_ipsec_stats_zero_assert(&stats);
 		CU_ASSERT_EQUAL(odp_ipsec_stats(sa_in, &stats), 0);
 		test_ipsec_stats_zero_assert(&stats);
+
+		/* Minimum bytes to be counted for stats.success_bytes */
+		if (!flags->ah) {
+			succ_bytes = test_out.pkt_in[0].len -
```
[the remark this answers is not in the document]
asasidharan: Agreed. Moved the succ_bytes calculation as suggested in v3.
Also added reviewed-by tag to this commit.

JannePeltonen (approved):

PR #1275, opened 2021-05-06
[PATCH v2] helper: ipsec: fix capability API usage

asasidharan:
Currently, the alg check function uses odp_crypto_*_capability
API to get the number of capability entries for a specific algorithm.
This is incorrect. Replacing the use of crypto capability APIs with
the correct IPsec APIs.


JannePeltonen (approved):

asasidharan:
Added Reviewed-by tag in v2.

PR #1276, opened 2021-05-06
[PATCH v3] linux-gen: fix print formats and enable compile-time checking

JereLeppanen:
```
linux-gen: fix print formats
    
    Fix mismatches between print formats and parameters in ODP_PRINT(),
    ODP_DBG(), etc. calls.
    
linux-gen: debug: check print formats and parameters
    
    Format strings and parameters in ODP_LOG(), ODP_PRINT(), etc. calls
    have not been cross-checked during compilation for some time due to
    commit b776ca47c0d8 ("linux-generic: debug: use global fn ptr for
    logging fn"). Enable this checking.
    
```

### DIFF
--- a/platform/linux-generic/odp_comp.c
+++ b/platform/linux-generic/odp_comp.c
@@ -133,7 +133,7 @@ static void process_input(odp_packet_t pkt_out,
 	do {
 		out_data =
 		    odp_packet_offset(pkt_out, start, &out_len, &cur_seg);
-		ODP_DBG("out_data 0x%x seg_data_ptr 0x%x out_len %d seg 0x%x\n",
+		ODP_DBG("out_data %p seg_data_ptr %p out_len %d seg %p\n",
 			out_data, odp_packet_seg_data(pkt_out, cur_seg),
 			out_len, cur_seg);
 
@@ -258,7 +258,7 @@ static int deflate_comp(odp_packet_t pkt_in,
 					 read,
 					 &in_len,
 					 &in_seg);
-		ODP_DBG("data 0x%x in_len %d seg 0x%x len %d\n",
+		ODP_DBG("data %p in_len %d seg %p len %d\n",
 			data, in_len, in_seg, len);
 
 		if (in_len > len)
--- a/platform/linux-generic/odp_packet_io.c
+++ b/platform/linux-generic/odp_packet_io.c
@@ -377,7 +377,7 @@ static const char *driver_name(odp_pktio_t hdl)
 
 	entry = get_pktio_entry(hdl);
 	if (entry == NULL) {
-		ODP_ERR("pktio entry %d does not exist\n", hdl);
+		ODP_ERR("pktio entry %" PRIuPTR " does not exist\n", (uintptr_t)hdl);
 		return "bad handle";
 	}
 
@@ -1253,7 +1253,7 @@ static inline uint32_t pktio_maxlen(odp_pktio_t hdl)
 
 	entry = get_pktio_entry(hdl);
 	if (entry == NULL) {
-		ODP_DBG("pktio entry %d does not exist\n", hdl);
+		ODP_DBG("pktio entry %" PRIuPTR " does not exist\n", (uintptr_t)hdl);
 		return 0;
 	}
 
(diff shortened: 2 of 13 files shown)


platform/linux-generic/odp_comp.c
```diff
@@ -133,7 +133,7 @@ static void process_input(odp_packet_t pkt_out,
 	do {
 		out_data =
 		    odp_packet_offset(pkt_out, start, &out_len, &cur_seg);
-		ODP_DBG("out_data 0x%x seg_data_ptr 0x%x out_len %d seg 0x%x\n",
+		ODP_DBG("out_data %p seg_data_ptr %p out_len %d seg %p\n",
```
JannePeltonen, line 136: I wonder if we want to use %p format for handles. Of course, with the format warning enaled, we will get a warning from this format if segment handle changes to a non-pointer. Or maybe this should not even try to print the handle directly, but use odp_packet_seg_to_u64()?

platform/linux-generic/odp_packet_io.c
```diff
@@ -377,7 +377,7 @@ static const char *driver_name(odp_pktio_t hdl)
 
 	entry = get_pktio_entry(hdl);
 	if (entry == NULL) {
-		ODP_ERR("pktio entry %d does not exist\n", hdl);
+		ODP_ERR("pktio entry %" PRIuPTR " does not exist\n", (uintptr_t)hdl);
```
JannePeltonen, line 380: Would it be better to use odp_pktio_to_u64() to get a printable value? OTOH, this is odp_packet_io.c which presumably knows very well what a pktio handle really is.

(And the same comment to other similar changes, e.g. in odp_timer.c)

platform/linux-generic/odp_traffic_mngr.c
```diff
@@ -4685,10 +4685,10 @@ void odp_tm_stats_print(odp_tm_t odp_tm)
 
 	ODP_PRINT("odp_tm_stats_print - tm_system=0x%" PRIX64 " tm_idx=%u\n",
 		  odp_tm, tm_system->tm_idx);
-	ODP_PRINT("  input_work_queue size=%u current cnt=%u peak cnt=%u\n",
-		  INPUT_WORK_RING_SIZE, input_work_queue->queue_cnt,
+	ODP_PRINT("  input_work_queue size=%u current cnt=%" PRIu64 " peak cnt=%u\n",
+		  INPUT_WORK_RING_SIZE, odp_atomic_load_u64(&input_work_queue->queue_cnt),
```
JannePeltonen, line 4689: I think the format for peak_cnt should also be fixed to use PRIu32.

platform/linux-generic/pktio/dpdk.c
```diff
@@ -1286,7 +1286,7 @@ static int dpdk_pktio_init(void)
 	masklen = odp_cpumask_to_str(&mask, mask_str, ODP_CPUMASK_STR_SIZE);
 
 	if (masklen < 0) {
-		ODP_ERR("CPU mask error: d\n", masklen);
+		ODP_ERR("CPU mask error: %d\n", masklen);
```
JannePeltonen, line 1289: I suppose this should use PRId32.

platform/linux-generic/odp_comp.c
JereLeppanen: We should probably just use to_u64() everywhere for all handles. I didn't want to do this for the handles that happen to be visible in this commit, but I could try to go through them systematically in a separate commit.

JannePeltonen: I am ok with not changing to use to_u64() in this commit. We anyway get the warning if the format gets out of sync with the handle type.

platform/linux-generic/odp_traffic_mngr.c
JereLeppanen: Fixed in v2.

platform/linux-generic/pktio/dpdk.c
JereLeppanen: Fixed in v2.

JannePeltonen (approved):

JereLeppanen:
v3: Add reviewed-by tags.

PR #1277, opened 2021-05-07
[PATCH v3] linux-gen: shm: fix unlink in _odp_ishm_cleanup_files()

chrhong:
In _odp_ishm_cleanup_files(), when previous shm files still remain, unlink failed
due to wrong path given. Use correct path, which includes the instance uid.

Signed-off-by: Christian Hong <hongguochun@hotmail.com>

### DIFF
--- a/platform/linux-generic/odp_ishm.c
+++ b/platform/linux-generic/odp_ishm.c
@@ -1645,7 +1645,7 @@ int _odp_ishm_cleanup_files(const char *dirpath)
 				return -1;
 			}
 			snprintf(fullpath, PATH_MAX, "%s/%s",
-				 dirpath, e->d_name);
+				 userdir, e->d_name);
 			ODP_DBG("deleting obsolete file: %s\n", fullpath);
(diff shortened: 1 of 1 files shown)


platform/linux-generic/odp_ishm.c
```diff
@@ -1645,7 +1645,7 @@ int _odp_ishm_cleanup_files(const char *dirpath)
 				return -1;
 			}
 			snprintf(fullpath, PATH_MAX, "%s/%s",
-				 dirpath, e->d_name);
+				 userdir, e->d_name);
```
MatiasElo, line 1648: A good catch. The following line above: `fullpath = malloc(d_len + f_len + 2);` also needs to be fixed. Currently, `d_len` refers to `dirpath`.

chrhong: A good review.

MatiasElo: Hi @chrhong, any progress with signing the  CLA?

PR #1278, opened 2021-05-07
[PATCH v2] CI clean-up

MatiasElo:


JereLeppanen (approved):
Looks good. Github UI incorrectly shows a removed line in ci-pipeline.yml:

`-        run: find . -name "*.trs" ...`

But the actual commit is correct.

PR #1279, opened 2021-05-07
[PATCH v2] linux-gen: ipsec: a couple of improvements and bug fixes

JannePeltonen:
- Keep free SAs in a freelist
- Simplify SA states
- Remove unnecessary synchronization
- Fix a parameter handling bug in SA creation


JannePeltonen:
v2: Fixed the typo and added review tags

PR #1283, opened 2021-05-12
[PATCH v5] Ipsec example padding fix

JannePeltonen:
example: ipsec: fix content of ESP padding bytes

    Fix ESP padding so that it is filled with an increasing counter value
    as specified by the ESP RFC. Incorrect (zeroed) padding caused the ESP
    packets to be rejected by the IPsec offload of the linux-gen ODP.

    Fixes: https://github.com/OpenDataPlane/odp/issues/1167

### DIFF
--- a/example/ipsec_api/odp_ipsec.c
+++ b/example/ipsec_api/odp_ipsec.c
@@ -640,11 +640,20 @@ pkt_disposition_e do_ipsec_in_classify(odp_packet_t *ppkt)
 		return PKT_POSTED;
 	} else {
 		int out = 1;
+		odp_ipsec_packet_result_t result;
 
 		rc = odp_ipsec_in(ppkt, 1, ppkt, &out, &in_param);
 		if (rc <= 0)
 			return PKT_DROP;
 
+		if (odp_ipsec_result(&result, *ppkt) < 0) {
+			ODPH_DBG("odp_ipsec_result() failed\n");
+			return PKT_DROP;
+		}
+		if (result.status.error.all != 0) {
+			ODPH_DBG("Error in inbound IPsec processing\n");
+			return PKT_DROP;
+		}
 		return PKT_CONTINUE;
 	}
 }
@@ -704,11 +713,20 @@ pkt_disposition_e do_ipsec_out_classify(odp_packet_t *ppkt, pkt_ctx_t *ctx)
 		return PKT_POSTED;
 	} else {
 		int out = 1;
+		odp_ipsec_packet_result_t result;
 
 		rc = odp_ipsec_out(ppkt, 1, ppkt, &out, &out_param);
 		if (rc <= 0)
 			return PKT_DROP;
 
+		if (odp_ipsec_result(&result, *ppkt) < 0) {
+			ODPH_DBG("odp_ipsec_result() failed\n");
+			return PKT_DROP;
+		}
+		if (result.status.error.all != 0) {
+			ODPH_DBG("Error in outbound IPsec processing\n");
+			return PKT_DROP;
+		}
 		return PKT_CONTINUE;
 	}
 }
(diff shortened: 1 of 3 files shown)


example/ipsec_api/odp_ipsec.c
```diff
@@ -640,11 +640,20 @@ pkt_disposition_e do_ipsec_in_classify(odp_packet_t *ppkt)
 		return PKT_POSTED;
 	} else {
 		int out = 1;
+		odp_ipsec_packet_result_t result;
 
 		rc = odp_ipsec_in(ppkt, 1, ppkt, &out, &in_param);
 		if (rc <= 0)
 			return PKT_DROP;
 
+		if (odp_ipsec_result(&result, *ppkt) < 0) {
```
asasidharan, line 649: So even in sync mode, we need to call odp_ipsec_result() to check if the IPsec processing was successful or not. Seeing this, I was wondering if the odp_ipsec_result() API should be made as a mandatory call in sync mode as well. What is your thoughts on that?

I recall seeing some related discussion in https://github.com/OpenDataPlane/odp/pull/1095.

JannePeltonen: I do not quite see the need for making odp_ipsec_result() mandatory in sync mode.

For async mode the rationale was that for a HW offloaded implementation some SW post-processing may be needed and making odp_ipsec_result() mandatory avoids having to check for the post-processing need in odp_schedule() (when vectors are used and there is no separate event-to-packet conversion).

In sync mode the implementation should be able to easily perform the needed postprocessing within the synchronous call.

example/ipsec_crypto/odp_ipsec_stream.c
```diff
@@ -322,8 +322,8 @@ odp_packet_t create_ipv4_packet(stream_db_entry_t *stream,
 
 		encrypt_len = ESP_ENCODE_LEN(payload_len + sizeof(*esp_t),
 					     entry->esp.block_len);
-		memset(data, 0, encrypt_len - payload_len);
-		data += encrypt_len - payload_len;
+		for (int n = 0; n < encrypt_len - payload_len; n++)
+			*data++ = n + 1;
```
asasidharan, line 326: While verifying this fix on our platform, observed that for tunnel mode, the total_len field of the inner IP is not set. It always comes as zero which seems incorrect.

Can we fix this as well with something like:
```
--- a/example/ipsec_crypto/odp_ipsec_stream.c                                    
+++ b/example/ipsec_crypto/odp_ipsec_stream.c                                    
@@ -272,6 +272,9 @@ odp_packet_t create_ipv4_packet(stream_db_entry_t *stream,   
                inner_ip = (odph_ipv4hdr_t *)data;                               
                memset((char *)inner_ip, 0, sizeof(*inner_ip));                  
                inner_ip->ver_ihl = 0x45;                                        
+               inner_ip->tot_len = odp_cpu_to_be_16(stream->length +            
+                                                    sizeof(*icmp) +             
+                                                    sizeof(*inner_ip));         
                inner_ip->proto = ODPH_IPPROTO_ICMPV4;                           
                inner_ip->id = odp_cpu_to_be_16(stream->id);                     
                inner_ip->ttl = 64;  ****
```

example/ipsec_api/odp_ipsec.c
asasidharan: Ack.

example/ipsec_crypto/odp_ipsec_stream.c
JannePeltonen: Yes, let's fix that too. I submitted a new commit with a fix that is slightly (sizeof stream_pkt_hdr_t, to be exact) different from above. Please check if it is ok now.

JannePeltonen:
v2: added a new commit: example: ipsec: fix length field of inner IP header in tunnel mode

example/ipsec_crypto/odp_ipsec_stream.c
asasidharan: Yes. The fix looks good.

asasidharan: > v2: added a new commit: example: ipsec: fix length field of inner IP header in tunnel mode

The PR looks good to me. You can add my reviewed-by if required.

JannePeltonen:
v3: added reviewed-by tag

MatiasElo (approved):

PR #1284, opened 2021-05-12
[PATCH v4] linux-gen: ipsec: add configurable ordering mode

JannePeltonen:
Add new config file parameters to control how original input queue
order is maintained during asynchronous IPsec processing.

When asynchronous IPsec operations are started within an ordered
scheduling context, packet order is maintained in IPsec completion
queues but the order in which the sequence number gets assigned or
the anti-replay check done is not guaranteed. This commit adds a config
option to force the stateful part of IPsec processing to be done in the
correct order, at the cost of reduced parallelism.

The ordering mechanism added in this commit uses the order_lock method
of the internal scheduling API to stop the calling thread until it
is its turn (i.e the preceding scheduling contexts associated with the
same input queue have been released). Future commits may add more
elaborate mechanisms.

The default behaviour does not change. By default no ordering is done.

### DIFF
--- a/config/odp-linux-generic.conf
+++ b/config/odp-linux-generic.conf
@@ -16,7 +16,7 @@
 
 # Mandatory fields
 odp_implementation = "linux-generic"
-config_file_version = "0.1.15"
+config_file_version = "0.1.16"
 
 # System options
 system: {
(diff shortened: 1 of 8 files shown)


JannePeltonen:
v2: Added config file version bumps that were missing from the fist version.

config/odp-linux-generic.conf
```diff
@@ -234,3 +234,51 @@ timer: {
...
+	#    Lack of ordering means that outbound IPsec packets, although
+	#    remaining in the correct order, may have their sequence numbers
+	#    assigned out of order. This can cause unexpected packet loss if
+	#    the anti-replay window of the receiving end is not large enough
+	#    to cover the possible misordering.
+	#
+	#    Similarly, since anti-replay check is not done in the reception
+	#    order, the anti-replay check sees additional packet misordering
+	#    on top of the true misordering of the received packets. This a
+	#    larger anti-replay window is required to avoid packet loss.
```
JereLeppanen, line 267: Last sentence needs fixing (grammar).

JannePeltonen:
v3: Fixed the botched comment sentence in the config file.

JereLeppanen (approved):

JannePeltonen:
v4: rebased and added review-tag

PR #1285, opened 2021-05-14
[PATCH v3] validation: pktin event queue

psavol:
Add test where (in addition to packet input) application sends events into pktin event queue. Scalable scheduler needs bug fixes (or further development) to support this. CI is disabled to run checks on scalable until it is fixed.

### DIFF
--- a/.github/workflows/ci-pipeline-arm64.yml
+++ b/.github/workflows/ci-pipeline-arm64.yml
@@ -135,23 +135,29 @@ jobs:
         if: ${{ failure() }}
         run: find . -name "*.trs" | xargs grep -l '^.test-result. FAIL' | while read trs ; do echo FAILURE detected at $trs; cat ${trs%%.trs}.log ; done
 
-
-  Run_scheduler:
+  Run_scheduler_sp:
     if: ${{ github.repository == 'OpenDataPlane/odp' }}
     runs-on: [self-hosted, ARM64]
-    strategy:
-      fail-fast: false
-      matrix:
-        scheduler: ['sp', 'scalable']
     steps:
       - uses: AutoModality/action-clean@v1.1.0
       - uses: actions/checkout@v2
       - run: sudo docker run -i -v `pwd`:/odp --privileged --shm-size 8g -e CC="${CC}" -e ARCH="${ARCH}"
-               -e CONF="${CONF}" -e ODP_SCHEDULER=${{matrix.scheduler}} $CONTAINER_NAMESPACE/odp-ci-${OS}-${ARCH}-native /odp/scripts/ci/check.sh
+               -e CONF="${CONF}" -e ODP_SCHEDULER=sp $CONTAINER_NAMESPACE/odp-ci-${OS}-${ARCH}-native /odp/scripts/ci/check.sh
       - name: Failure log
         if: ${{ failure() }}
         run: find . -name "*.trs" | xargs grep -l '^.test-result. FAIL' | while read trs ; do echo FAILURE detected at $trs; cat ${trs%%.trs}.log ; done
 
+  Run_scheduler_scalable:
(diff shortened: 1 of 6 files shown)


jerinjacobk (commented):
We will check the PR in Marvell ODP and provide feedback. Please wait for feedback to merging this.

PavanNikhilesh:
Petri, 
In our odp implementation when pktin is configured as plain queue, only HW is allowed to enqueue to it, can we have a capability check to avoid this test case?

JannePeltonen:
> In our odp implementation when pktin is configured as plain queue, only HW is allowed to enqueue to it.

Can you change your implementation that also SW could enqueue to such a queue?

jerinjacobk:
> > In our odp implementation when pktin is configured as plain queue, only HW is allowed to enqueue to it.
> 
> Can you change your implementation so that also SW could enqueue to such a queue?

Our HW does not support SW to enqueue to plain queue (only HW can enqueue it), In the event mode, HW support enqueue by SW. We think, emulation on the same with SW ring is complex and affect the direct queue performance not prime use case with our HW. The application can use odp_stash API to enable such use cases.

psavol:
v2: Plain queue test case removed. This new test case does not run with scalable scheduler, all other tests are run as before.

test/validation/api/pktio/pktio.c
```diff
@@ -3810,6 +3810,169 @@ static void pktio_test_pktout_aging_tmo(void)
...
+	buf_pool = odp_pool_create("buffer pool", &pool_param);
+	CU_ASSERT_FATAL(buf_pool != ODP_POOL_INVALID);
+
+	buf = odp_buffer_alloc(buf_pool);
+	CU_ASSERT_FATAL(buf != ODP_BUFFER_INVALID);
+
+	odp_pktin_queue_param_init(&in_queue_param);
+	in_queue_param.num_queues  = 1;
+	in_queue_param.hash_enable = 0;
+	in_queue_param.classifier_enable = 0;
```
MatiasElo, line 3852: Both are disabled by default.

test/validation/api/pktio/pktio.c
```diff
@@ -3810,6 +3810,169 @@ static void pktio_test_pktout_aging_tmo(void)
...
+	int num_buf = 0;
+	int num_bad = 0;
+	odp_pktio_t pktio[MAX_NUM_IFACES] = {0};
+	uint64_t wait_sec = odp_schedule_wait_time(ODP_TIME_SEC_IN_NS);
+
+	CU_ASSERT_FATAL(num_ifaces >= 1);
+
+	odp_pool_param_init(&pool_param);
+	pool_param.type = ODP_POOL_BUFFER;
+	pool_param.buf.num = 10;
```
MatiasElo, line 3840: TX_BATCH_LEN could be used also here, so that the pool is always large enough.

psavol: Fixed in v3.

test/validation/api/pktio/pktio.c
psavol: I put those here to demonstrate to the reader that we have a single (atomic) queue without any flow detection.

.github/workflows/ci-pipeline-arm64.yml
```diff
@@ -135,23 +135,29 @@ jobs:
         if: ${{ failure() }}
         run: find . -name "*.trs" | xargs grep -l '^.test-result. FAIL' | while read trs ; do echo FAILURE detected at $trs; cat ${trs%%.trs}.log ; done
 
-
-  Run_scheduler:
+  Run_scheduler_sp:
```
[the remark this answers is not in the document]
psavol: fixed in v3

MatiasElo (approved):

PR #1287, opened 2021-05-17
[PATCH v2] validation: scheduler: add new atomicity test

MatiasElo:
Add a new atomicity validation test, which checks that only a single thread
is able to process events from an atomic queue at the same time.

V2:
- Added comment about releasing atomic context and prescheduled events (Janne)

### DIFF
--- a/test/validation/api/scheduler/scheduler.c
+++ b/test/validation/api/scheduler/scheduler.c
@@ -59,6 +59,7 @@
 #define WAIT_1MS_RETRIES 1000
 
 #define SCHED_AND_PLAIN_ROUNDS 10000
+#define ATOMICITY_ROUNDS 100
 
 /* Test global variables */
 typedef struct {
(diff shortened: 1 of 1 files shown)


test/validation/api/scheduler/scheduler.c
```diff
@@ -2484,6 +2489,120 @@ static void scheduler_test_ordered_and_plain(void)
...
+		CU_ASSERT_FATAL(odp_atomic_cas_acq_rel_u32(state, &old, 0));
+
+		CU_ASSERT_FATAL(odp_queue_enq(from, ev) == 0);
+
+		num_processed++;
+	}
+
+	odp_schedule_pause();
+	while ((ev = odp_schedule(NULL, ODP_SCHED_NO_WAIT)) != ODP_EVENT_INVALID)
+		CU_ASSERT_FATAL(odp_queue_enq(atomic_queue, ev) == 0);
```
JannePeltonen, line 2535: I was about to write a comment here that this loop here seems redundant because of the draining that is done later, but then I realized that because we are using a non-parallel queue we have to pause the scheduler and get rid of possible prefetched events and our scheduling context in order to not block other threads and end up in a deadlock with the barrier. A short comment in the code might be helpful if this is not something that should be obvious anyway. 

Otherwise this commit looks good to me.

MatiasElo: Okay, I'll add a comment in V2.

PR #1288, opened 2021-05-19
[PATCH v10] api: pktio: add statistics capabilities and new counters

MatiasElo:
- Add capabilities for statistics counters
- Add statistics counter for broadcast and multicast packets
- Add input/output queue specific statistics counters

V2:
- Added new `odp_pktio_queue_t` union for passing queue handles to `odp_pktin_queue_stats()` and `odp_pktout_queue_stats()` functions instead of using queue indices 

V3:
- Added reference implementation and validation tests

V4:
- Added separate functions for direct and event packet input/output queues

V5:
- Extend `odp_pktio_stats_reset()` documentation to state that resetting other than interface level statistics is implementation defined

### DIFF
--- a/include/odp/api/spec/packet_io_stats.h
+++ b/include/odp/api/spec/packet_io_stats.h
@@ -87,6 +87,57 @@ typedef struct odp_pktio_stats_t {
 } odp_pktio_stats_t;
 
 /**
+ * Packet IO statistics capabilities
+ */
+typedef struct odp_pktio_stats_capability_t {
+	/** Interface level capabilities */
+	struct {
+		/** Supported counters */
+		union {
+			/** Statistics counters in a bit field structure */
+			struct {
+				/** @see odp_pktio_stats_t::in_octets */
+				uint64_t in_octets          : 1;
+
+				/** @see odp_pktio_stats_t::in_packets */
+				uint64_t in_packets         : 1;
+
+				/** @see odp_pktio_stats_t::in_ucast_pkts */
+				uint64_t in_ucast_pkts      : 1;
+
+				/** @see odp_pktio_stats_t::in_discards */
+				uint64_t in_discards        : 1;
+
+				/** @see odp_pktio_stats_t::in_errors */
+				uint64_t in_errors          : 1;
+
+				/** @see odp_pktio_stats_t::out_octets */
+				uint64_t out_octets         : 1;
+
+				/** @see odp_pktio_stats_t::out_packets */
+				uint64_t out_packets        : 1;
+
+				/** @see odp_pktio_stats_t::out_ucast_pkts */
+				uint64_t out_ucast_pkts     : 1;
+
+				/** @see odp_pktio_stats_t::out_discards */
+				uint64_t out_discards       : 1;
+
+				/** @see odp_pktio_stats_t::out_errors */
+				uint64_t out_errors         : 1;
+			} counter;
+
+			/** All bits of the bit field structure
+			 *
+			 *  This field can be used to set/clear all flags, or
+			 *  for bitwise operations over the entire structure. */
+			uint64_t all_counters;
+		};
+	} pktio;
+
+} odp_pktio_stats_capability_t;
+
+/**
  * Get statistics for pktio handle
  *
  * Counters not supported by the interface are set to zero.
--- a/include/odp/api/spec/packet_io_stats.h
+++ b/include/odp/api/spec/packet_io_stats.h
@@ -35,7 +35,7 @@ extern "C" {
  */
 typedef struct odp_pktio_stats_t {
 	/** Number of octets in successfully received packets. In case of
-	 *  Ethernet, packet size includes MAC header and FCS. */
+	 *  Ethernet, packet size includes MAC header. */
 	uint64_t in_octets;
 
 	/** Number of successfully received packets. */
@@ -68,7 +68,7 @@ typedef struct odp_pktio_stats_t {
 	uint64_t ODP_DEPRECATE(in_unknown_protos);
 
 	/** Number of octets in successfully transmitted packets. In case of
-	 *  Ethernet, packet size includes MAC header and FCS. */
+	 *  Ethernet, packet size includes MAC header. */
 	uint64_t out_octets;
 
 	/** Number of successfully transmitted packets. */
--- a/include/odp/api/spec/packet_io_stats.h
+++ b/include/odp/api/spec/packet_io_stats.h
@@ -20,6 +20,7 @@ extern "C" {
 #endif
 
 #include <odp/api/deprecated.h>
+#include <odp/api/queue.h>
 
 /** @addtogroup odp_packet_io
  *  @{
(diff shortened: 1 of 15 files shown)


include/odp/api/spec/packet_io_stats.h
```diff
@@ -86,6 +86,57 @@ typedef struct odp_pktio_stats_t {
 	uint64_t out_errors;
 } odp_pktio_stats_t;
 
+/**
+ * Packet IO statistics capabilities
+ */
+typedef struct odp_pktio_stats_capability_t {
```
jerinjacobk, line 150: How about not duplicating this structure and reuse odp_pktio_stats_t

MatiasElo: The reasoning for adding a separate struct was to enable adding additional capabilities in the future, which may not be counters. E.g. are pause frames included in the counters.

psavol: Benefits of the bit field are saved memory space (8 vs 80bytes) and easy comparison of multiple flags (by using all_counters).

The stats capa structure allows to combine capabilities for all pktio statistics, and also potential relation between those (e.g. "both interface and queue level stats supported, but not simultaneously").

psavol:
API spec looks good to me.

include/odp/api/spec/packet_io_stats.h
```diff
@@ -163,8 +206,75 @@ typedef struct odp_pktio_stats_capability_t {
...
+/**
+ * Union of all queue handle types used for packet input/output
+ */
+typedef union {
+	/** Event queue handle for packet input or output */
+	odp_queue_t event_queue;
+	/** Direct packet input queue handle */
+	odp_pktin_queue_t pktin_queue;
+	/** Direct packet output queue handle */
+	odp_pktout_queue_t pktout_queue;
```
jerinjacobk, line 274: Need TM queue and CoS handle also. Right?

MatiasElo: That was not the intention. Concept-wise these are so different compared to pktio that I would prefer adding separate statistics functions for these two modules.

jerinjacobk: Ok. Makes sense to add as separate functions. Will it be in this patch or another patch in series?

MatiasElo: They'll be in a separate PR.

include/odp/api/spec/packet_io_stats.h
```diff
@@ -102,6 +103,48 @@ typedef struct odp_pktio_stats_t {
...
 
+/**
+ * Packet IO input queue specific statistics counters
+ *
+ * Statistics counters for an individual packet input queue. Refer to packet IO
+ * level statistics odp_pktio_stats_t for counter definitions.
+ */
+typedef struct odp_pktin_queue_stats_t {
+	/** @see odp_pktio_stats_t::in_octets */
+	uint64_t in_octets;
```
yuri-tolstov, line 114: Counters are already in the input (queue) namespace, so, I think, no reason to add "in_" prefix. The same applies to output queue.

include/odp/api/spec/packet_io_stats.h
yuri-tolstov, line 274: What is use case(s) for this unified queue descriptor? It cannot be used without specified queue type. Plus, it mixes bidirectional and unidirectional queues. So, all in all, it may be confusing during its usage.

include/odp/api/spec/packet_io_stats.h
```diff
@@ -178,6 +288,44 @@ typedef struct odp_pktio_stats_capability_t {
...
+ *
+ * @param       pktio	 Packet IO handle
+ * @param       queue	 Input queue handle
+ * @param[out]  stats	 Output buffer for counters
+ *
+ * @retval  0 on success
+ * @retval <0 on failure
+ */
+int odp_pktin_queue_stats(odp_pktio_t pktio, odp_pktio_queue_t queue,
+			  odp_pktin_queue_stats_t *stats);
```
yuri-tolstov, line 308: In this case, it would be more logical to pass "odp_pktin_queue_t queue". Unless, the entire PKTIO API is realigned with using unified queue descriptor, so the usage of descriptors is uniform. The same applies to the pktout queue statistics call.

include/odp/api/spec/packet_io_stats.h
MatiasElo: The original reasoning was to make the naming consistent with `odp_pktio_stats_t`, but now when we'll likely have also `odp_cls_queue_stats_t ` and `odp_tm_queue_stats_t ` stats without the prefixes, I'd tend to agree with you.

include/odp/api/spec/packet_io_stats.h
MatiasElo: The unified queue type was added to avoid having to duplicate the same functions for different packet input/output modes (`ODP_PKTIN_MODE_DIRECT`/`ODP_PKTIN_MODE_QUEUE` and `ODP_PKTOUT_MODE_DIRECT`/`ODP_PKTOUT_MODE_QUEUE`). The implementation knows which queue type to use based on `odp_pktin_queue_stats()`/`odp_pktout_queue_stats()`  functions' `pktio` argument.

include/odp/api/spec/packet_io_stats.h
MatiasElo: We'd need a separate function for packet input mode `ODP_PKTIN_MODE_SCHED` and `ODP_PKTIN_MODE_QUEUE`.

include/odp/api/spec/packet_io_stats.h
MatiasElo: Fixed in V4.

include/odp/api/spec/packet_io_stats.h
MatiasElo: Fixed in V4.

platform/linux-generic/pktio/stats/ethtool_stats.c
```diff
@@ -131,9 +131,15 @@ static int ethtool_stats(int fd, struct ifreq *ifr, odp_pktio_stats_t *stats)
 		} else if (!strcmp(cnt, "rx_packets")) {
 			stats->in_packets = val;
 			cnts++;
-		} else if (!strcmp(cnt, "rx_ucast_packets")) {
+		} else if (!strcmp(cnt, "rx_unicast_packets")) {
```
psavol, line 134: Extra stats PR touches this same line. Better leave it as is here to avoid merge conflict.

platform/linux-generic/pktio/stats/ethtool_stats.c
```diff
@@ -131,9 +131,15 @@ static int ethtool_stats(int fd, struct ifreq *ifr, odp_pktio_stats_t *stats)
...
 			stats->in_ucast_pkts = val;
 			cnts++;
+		} else if (!strcmp(cnt, "rx_broadcast_packets")) {
+			stats->in_bcast_pkts = val;
+			cnts++;
+		} else if (!strcmp(cnt, "rx_multicast_packets")) {
+			stats->in_mcast_pkts = val;
+			cnts++;
```
psavol, line 142: It's likely that broad/multicast have alternative names like other counters.

platform/linux-generic/pktio/stats/ethtool_stats.c
MatiasElo: Fixed in V7.

platform/linux-generic/pktio/stats/ethtool_stats.c
MatiasElo: Fixed in V7.

jerinjacobk (approved):

jerinjacobk (approved):

PR #1293, opened 2021-05-26
[PATCH v4] api: time: add strict time stamp read functions

psavol:
I named these functions "strict" due to they maintain program order more strictly than the basic functions. Other terms like precise/exact/etc would likely promise too much about when a time stamp is actually taken. This flavor of calls use more barriers/etc instructions to maintain program order than the basic calls (and thus hurt performance more), but may not still be able to guarantee 100% program order (if ISA does not allow it, or if it is impractical).

### DIFF
--- a/include/odp/api/spec/time.h
+++ b/include/odp/api/spec/time.h
@@ -1,5 +1,5 @@
 /* Copyright (c) 2013-2018, Linaro Limited
- * Copyright (c) 2020, Nokia
+ * Copyright (c) 2020-2021, Nokia
  * All rights reserved.
  *
  * SPDX-License-Identifier:     BSD-3-Clause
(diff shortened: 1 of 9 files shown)


JereLeppanen (approved):

include/odp/api/spec/time.h
```diff
@@ -65,6 +65,36 @@ extern "C" {
...
+ *
+ * @return Local time stamp
+ */
+odp_time_t odp_time_local_strict(void);
+
+/**
+ * Current local time in nanoseconds (strict)
+ *
+ * Like odp_time_local_ns(), but reads the time stamp value more strictly (see
+ * odp_time_local_strict() documentation) in the program order.
```
MatiasElo, line 92: Or: `Like odp_time_local_strict(), but the time stamp value is converted into nanoseconds.`

include/odp/api/spec/time.h
```diff
@@ -78,22 +108,33 @@ odp_time_t odp_time_local(void);
...
+ *
+ * @return Global time stamp
+ */
+odp_time_t odp_time_global_strict(void);
+
+/**
+ * Current global time in nanoseconds (strict)
+ *
+ * Like odp_time_global_ns(), but reads the time stamp value more strictly (see
+ * odp_time_local_strict() documentation) in the program order.
```
MatiasElo, line 133: Or: `Like odp_time_global_strict(), but the time stamp value is converted into nanoseconds.`

platform/linux-generic/include/odp/api/plat/time_inlines.h
```diff
@@ -46,6 +46,18 @@ static inline odp_time_t _odp_time_cur(void)
 	return _odp_timespec_cur();
 }
 
+static inline odp_time_t _odp_time_cur_strict(void)
```
MatiasElo, line 49: Copyright can be updated.

include/odp/api/spec/time.h
```diff
@@ -65,6 +65,36 @@ extern "C" {
...
 odp_time_t odp_time_local(void);
 
+/**
+ * Current local time in nanoseconds
+ *
+ * Like odp_time_local(), but the time stamp value is converted into nanoseconds.
+ *
+ * @return Local time stamp in nanoseconds
+ */
+uint64_t odp_time_local_ns(void);
```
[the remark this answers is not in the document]
psavol: Fixed in v3.

include/odp/api/spec/time.h
psavol: Fixed in v3.

include/odp/api/spec/time.h
psavol: Fixed in v3.

platform/linux-generic/include/odp/api/plat/time_inlines.h
psavol: Fixed in v3.

MatiasElo (approved):

jerinjacobk (approved):
I kind of like precise. But strict is fine too.

PR #1294, opened 2021-05-26
[PATCH v4] api: timer: add clock src enumerations

psavol:
I moved clock source enumeration patches into this PR from the tick info PR. I think it makes sense to add rename CPU to DEFAULT  clock source at the same time.

### DIFF
--- a/example/timer/odp_timer_accuracy.c
+++ b/example/timer/odp_timer_accuracy.c
@@ -106,8 +106,8 @@ static void print_usage(void)
 	       "  -e, --early_retry <num> When timer restart fails due to ODP_TIMER_TOOEARLY, retry this many times\n"
 	       "                          with expiration time incremented by the period. Default: 0\n"
 	       "  -s, --clk_src           Clock source select (default 0):\n"
-	       "                            0: ODP_CLOCK_CPU\n"
-	       "                            1: ODP_CLOCK_EXT\n"
+	       "                            0: ODP_CLOCK_DEFAULT\n"
+	       "                            1: ODP_CLOCK_SRC_1, ...\n"
 	       "  -i, --init              Set global init parameters. Default: init params not set.\n"
 	       "  -h, --help              Display help and exit.\n\n");
 }
@@ -144,7 +144,7 @@ static int parse_options(int argc, char *argv[], test_global_t *test_global)
 	test_global->opt.burst     = 1;
 	test_global->opt.burst_gap = 0;
 	test_global->opt.mode      = 0;
-	test_global->opt.clk_src   = 0;
+	test_global->opt.clk_src   = ODP_CLOCK_DEFAULT;
 	test_global->opt.init      = 0;
 	test_global->opt.output    = 0;
 	test_global->opt.early_retry = 0;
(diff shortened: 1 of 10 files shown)


include/odp/api/spec/timer.h
```diff
@@ -36,16 +36,45 @@ extern "C" {
...
+	/** Number of clock source enumerations */
+	ODP_CLOCK_NUM_SRC
+
 } odp_timer_clk_src_t;
 
+/** The default clock source */
+#define ODP_CLOCK_DEFAULT ODP_CLOCK_SRC_0
+
+/** For backwards compatibility, ODP_CLOCK_CPU is synonym for ODP_CLOCK_DEFAULT */
+#define ODP_CLOCK_CPU ODP_CLOCK_DEFAULT
```
JannePeltonen, line 73: ODP_CLOCK_CPU seems to be still used in examples and in test/performance after all commits of this PR. And ODP_CLOCK_EXT is still used in example/timer.

psavol: I'll update all test apps once the API changes has been approved.

include/odp/api/spec/timer.h
```diff
@@ -36,16 +36,45 @@ extern "C" {
  */
 
 /**
- * Clock sources for timers in timer pool.
+ * Clock sources for timer pools
+ *
+ * ODP_CLOCK_DEFAULT is the default clock source and it is supported always. It is implementation
+ * defined which other clock sources are supported. See implementation documentation how the
+ * supported clock sources are mapped into these enumerations.
```
MatiasElo, line 43: Maybe better: `See from implementation documentation how the...` or `See implementation documentation for information on how the...`

include/odp/api/spec/timer.h
MatiasElo: Since this define is required only for backward compatibility you could also state that it will be deprecated in the future. Just to encourage people to use the new enumerations.

include/odp/api/spec/timer.h
psavol: Fixed with "See from .." in v2

example/timer/odp_timer_accuracy.c
```diff
@@ -291,9 +291,9 @@ static int start_timers(test_global_t *test_global)
 	test_global->timeout_pool = pool;
 
 	if (test_global->opt.clk_src == 0)
-		clk_src = ODP_CLOCK_CPU;
+		clk_src = ODP_CLOCK_DEFAULT;
 	else
-		clk_src = ODP_CLOCK_EXT;
+		clk_src = test_global->opt.clk_src;
```
JannePeltonen, line 296: How about replacing this whole if-else with simply "clk_src = test_global->opt.clk_src"? The API promises that ODP_CLOCK_DEFAULT is zero.

Other than the above, this patch set looks good to me.

psavol: I was thinking that it's more explicit to the user that "-s 0" option is DEFAULT, and for other values it is implementation specific where they connect to. But maybe it's enough that the help text documents it.

psavol: Fixed in v3.

JannePeltonen:

MatiasElo (approved):

jerinjacobk (approved):

PR #1296, opened 2021-05-26
linux-gen: ipsec: add esn feature support

mrchalla:
IPsec ESN feature support is added to the linux-gen platform
implementation.

The ESN feature is added on top of the submitted anti-replay
feature implementation, its PR is
[https://github.com/OpenDataPlane/odp/pull/1261].

Signed-off-by: Sachin Yaligar <syaligar@marvell.com>


JannePeltonen:
This does not work as it is missing a crucial part of ESN implementation. I already commented this in PR 1261:

```
The ESN support does not work correctly because it does not include the high order bits of the ESN
in the ICV calculation. Please check RFC 4303, sections 2 and 2.2.1 and table 1.
```

Another problem in this PR is that the commit contains not only the ESN changes but antireplay changes which are supposed to be in a separate commit in a separate PR. Please do not squash separate commit like that. If you need (temporarily) to have another commit in this PR because of some dependency, then keep it has a separate commit.

The commit message should also be fixed. We do not discuss github PR submissions in commit messages.

mrchalla:
> This does not work as it is missing a crucial part of ESN implementation. I already commented this in PR 1261:
> 
> ```
> The ESN support does not work correctly because it does not include the high order bits of the ESN
> in the ICV calculation. Please check RFC 4303, sections 2 and 2.2.1 and table 1.
> ```


Ack. seems, I missed the ICV part. 


> If you need (temporarily) to have another commit in this PR because of some dependency, then keep it has a separate commit.

Got it thanks, will fix it.


> 
> The commit message should also be fixed. We do not discuss github PR submissions in commit messages.

Thanks again, will remove it.

mrchalla:
Updated the PR.

- Added ESN higher bits to IPsec outbound and inbound ICV computation.
- This version only supports the ESP protocol and  the support for AH protocol will be added in the future version of this PR.

Please review and provide your suggestions.

mrchalla:
Hi @JannePeltonen, please provide the review comments/suggestions, I will try to incorporate these comments in the next version of the PR with also AH mode support, which I am still working.

mrchalla:
Updated the PR with
- Inbound and outbound ICV computation for AH mode protocol support with ESN.

mrchalla:
Hi @JannePeltonen, 
Would you please review this PR.

mrchalla:
Hi @JannePeltonen , Kindly review this PR.

JannePeltonen:
ESN implementation for AES-GMAC and AES-CCM is missing. At minimum, SA creation with ESN enabled for those algorithms should not be allowed.

platform/linux-generic/include/odp_ipsec_internal.h
```diff
@@ -158,6 +158,8 @@ struct ipsec_sa_s {
 	void		*context;
 	odp_queue_t	queue;
 
+	/* ESN higher 32 bits length */
+	uint32_t	seqh_len;
```
JannePeltonen, line 162: I think the comment is slightly misleading as the "length" of the 32 high bits of the sequence number is of course 32 bits, i.e. 4 bytes.

I would actually suggest removing this field and adding a "combined_mode" or "aead" flag instead. The code could then check the flag when it needs to know whether to put the high part of the seqno after the next header field or not. And for the number of bytes the code could just use the constant 4 (defined appropriately).

platform/linux-generic/odp_ipsec.c
```diff
@@ -375,6 +375,34 @@ typedef struct {
...
+/*
+ * Computes 64-bit seq number according to RFC4303 A2
+ */
+static inline uint64_t ipsec_compute_esn(ipsec_sa_t *ipsec_sa, uint32_t seq)
+{
+	uint32_t wintop_h, wintop_l, winbot_l, ws;
+	uint64_t seq64 = 0;
+
+	wintop_h = odp_atomic_load_u64(&ipsec_sa->hot.in.wintop_seq) >> 32;
+	wintop_l = odp_atomic_load_u64(&ipsec_sa->hot.in.wintop_seq) & 0xffffffff;
```
JannePeltonen, line 387: You should not split the load to loads like that. The two loads could observe different values of the variable (i.e. a write by another thread occuring in between the loads).

platform/linux-generic/odp_ipsec.c
```diff
@@ -565,18 +593,40 @@ static int ipsec_in_esp(odp_packet_t *pkt,
 	param->auth_iv_ptr = state->iv;
 
 	state->esp.aad.spi = esp.spi;
-	state->esp.aad.seq_no = esp.seq_no;
 	state->in.seq_no = odp_be_to_cpu_32(esp.seq_no);
+	if (ipsec_sa->esn)
+		ipsec_sa->in.seq64 = ipsec_compute_esn(ipsec_sa, state->in.seq_no);
```
JannePeltonen, line 598: This is broken with multiple threads processing IPsec packets. You cannot store temporary per-thread per-packet data in a shared variable like that.

platform/linux-generic/odp_ipsec.c
```diff
@@ -565,18 +593,40 @@ static int ipsec_in_esp(odp_packet_t *pkt,
...
+
+	/* AAD with ESN */
+	if (ipsec_sa->esn && (!ipsec_sa->seqh_len))
+		state->esp.aad.seq_no64 = odp_cpu_to_be_64(ipsec_sa->in.seq64);
+	else if (!ipsec_sa->esn)
+		state->esp.aad.seq_no = esp.seq_no;
```
JannePeltonen, line 604: To me this construct is hard to read. Instead of this:
```
if (a & !b)
        x();
else if (!a)
        y();
```
...one could write this:
```
if (a) {
        if (!b)
                x();
} else
        y();
```

platform/linux-generic/odp_ipsec.c
```diff
@@ -565,18 +593,40 @@ static int ipsec_in_esp(odp_packet_t *pkt,
...
 
 	param->aad_ptr = (uint8_t *)&state->esp.aad;
 
+	/* Inbound ICV computation includes ESN higher 32 bits for individual algo's */
```
JannePeltonen, line 608: This is true also for combined mode algorithms (it is just done in a different way) so the comment may be a bit misleading.

platform/linux-generic/odp_ipsec.c
```diff
@@ -565,18 +593,40 @@ static int ipsec_in_esp(odp_packet_t *pkt,
...
+	if (ipsec_sa->seqh_len) {
+		uint32_t inb_seqh = (ipsec_sa->in.seq64 >> 32);
+		uint32_t icv_offset =  odp_packet_len(*pkt) - ipsec_sa->icv_len;
+
+		if (odp_packet_add_data(pkt, icv_offset, ipsec_sa->seqh_len) < 0) {
```
JannePeltonen, line 613: odp_packet_add_data() actually copies the whole packet (both in linux-gen and odp-dpdk). We do not want to do such a slow operation here.

platform/linux-generic/odp_ipsec.c
```diff
@@ -565,18 +593,40 @@ static int ipsec_in_esp(odp_packet_t *pkt,
...
 	param->auth_range.length = state->ip_tot_len -
 				  state->ip_hdr_len -
-				  ipsec_sa->icv_len;
+				  ipsec_sa->icv_len +
+				  ipsec_sa->seqh_len;
 	param->hash_result_offset = state->ip_offset +
 				   state->ip_tot_len -
-				   ipsec_sa->icv_len;
+				   ipsec_sa->icv_len +
+				   ipsec_sa->seqh_len;
```
JannePeltonen, line 629: Please keep the summation order the same as the order of the fields in the packet in both the offset and the length calculation).

platform/linux-generic/odp_ipsec.c
```diff
@@ -834,11 +901,12 @@ static ipsec_sa_t *ipsec_in_single(odp_packet_t pkt,
 		if ((crypto.cipher_status.alg_err ==
 		     ODP_CRYPTO_ALG_ERR_ICV_CHECK) ||
 		    (crypto.auth_status.alg_err ==
-		     ODP_CRYPTO_ALG_ERR_ICV_CHECK))
+		     ODP_CRYPTO_ALG_ERR_ICV_CHECK)) {
+			ODP_ERR("ICV_CHECK failed\n");
```
JannePeltonen, line 905: I do not think ODP_ERR() should be called here. In a sense, this is not an error (like e.g. some allocation failure) but normal operation.

Outputting an error log here would allow an attacker to fill the error log by sending us bogus or replayed IPsec packets. I think this should be at most ODP_DEBUG.

platform/linux-generic/odp_ipsec.c
```diff
@@ -870,6 +938,9 @@ static ipsec_sa_t *ipsec_in_single(odp_packet_t pkt,
 		goto post_lifetime_err_cnt_update;
 	}
 
+	/* Truncate the ESN higher 32 bits */
```
JannePeltonen, line 941: The comment is confusing since there is no truncation going on here.

platform/linux-generic/odp_ipsec.c
```diff
@@ -1498,6 +1604,16 @@ static void ipsec_out_ah_post(ipsec_state_t *state, odp_packet_t pkt)
 		ipv6hdr->ver_tc_flow = state->ah_ipv6.ver_tc_flow;
 		ipv6hdr->hop_limit = state->ah_ipv6.hop_limit;
 	}
+
+	/* As ICV is already computed, truncate the ESN higher 32 bits. */
+	if (ipsec_sa->seqh_len) {
+		if (odp_packet_trunc_tail(&pkt, ipsec_sa->seqh_len, NULL, NULL) < 0) {
```
JannePeltonen, line 1610: This may change the packet handle but the changed handle is not propagated back to the caller who would continue operating with a stale handle.

platform/linux-generic/odp_ipsec_sad.c
```diff
@@ -956,7 +974,8 @@ int _odp_ipsec_sa_replay_precheck(ipsec_sa_t *ipsec_sa, uint32_t seq,
 				  odp_ipsec_op_status_t *status)
 {
 	/* Try to be as quick as possible, we will discard packets later */
-	if (ipsec_sa->antireplay && ((seq + ipsec_sa->in.ar.win_size) <=
+	if ((!ipsec_sa->esn) && ipsec_sa->antireplay &&
+	    ((seq + ipsec_sa->in.ar.win_size) <=
```
JannePeltonen, line 978: This change breaks the anti-replay checking logic. After the change, replayed packets whose sequence number is below the window may get accepted.

platform/linux-generic/odp_ipsec_sad.c
```diff
@@ -965,17 +984,18 @@ int _odp_ipsec_sa_replay_precheck(ipsec_sa_t *ipsec_sa, uint32_t seq,
...
-static inline int ipsec_wslarge_replay_update(ipsec_sa_t *ipsec_sa, uint32_t seq,
+static inline int ipsec_wslarge_replay_update(ipsec_sa_t *ipsec_sa, uint64_t seq,
 					      odp_ipsec_op_status_t *status)
 {
-	uint32_t bucket, wintop_bucket, new_bucket;
-	uint32_t bkt_diff, bkt_cnt, top_seq;
-	uint64_t bit = 0;
+	uint64_t bucket, wintop_bucket, new_bucket;
+	uint64_t bit = 0, top_seq;
+	uint32_t bkt_diff, bkt_cnt;
```
JannePeltonen, line 992: The fact the bkt_diff stays uint32_t but get stored the difference between bucket and wintop_bucket looks a bit dubious to me.

At first look it seems like there could be an overflow and then breakage of the subsequent logic. But I suppose maybe such large differences cannot occur here because such packets would be rejected by the authentication check (because we only try with one higher value of the high order part of the ESN). Is that the thinking behind this?

To me this is non-obvious and a bit fragile, so I would suggest just using a 64 bit variable for the bkt_diff too. Or am I missing something?

mrchalla:
> At minimum, SA creation with ESN enabled for those algorithms should not be allowed.
Ack, will check and add these changes.

platform/linux-generic/include/odp_ipsec_internal.h
mrchalla: >I would actually suggest removing this field and adding a "combined_mode" or "aead" flag instead.
Agree with your suggestion.

  >And for the number of bytes the code could just use the constant 4 (defined appropriately).
Sure, will use an appropriate macro.

platform/linux-generic/odp_ipsec.c
mrchalla: Ack, will use the load once and use this loaded value to get "wintop_h" and "wintop_l".

platform/linux-generic/odp_ipsec.c
mrchalla: Agreed, will fix it.

platform/linux-generic/odp_ipsec.c
mrchalla: ack, will make it readable.

platform/linux-generic/odp_ipsec.c
mrchalla: ack, will correct this comment.

platform/linux-generic/odp_ipsec.c
mrchalla: Ack, will fix it.

platform/linux-generic/odp_ipsec.c
mrchalla: ack, ODP_ERR() is not required, will remove it.

platform/linux-generic/odp_ipsec.c
mrchalla: Ack, will fix it.

platform/linux-generic/odp_ipsec_sad.c
mrchalla: >After the change, replayed packets whose sequence number is below the window may get accepted.
Yes, if ESN is enabled. In this case these packets will be part of the next 32-bit AR window( not the current window). As per the ESN, these packets are supposed to be allowed.

platform/linux-generic/odp_ipsec_sad.c
mrchalla: Yes, that's the thought. But will change the bkt_diff as 64-bit variable.

platform/linux-generic/odp_ipsec_sad.c
JannePeltonen: Sorry, I pasted my comment to a wrong piece of code. My comments was for similar change in ipsec_wslarge_replay_update(). There the change lets through old packets that fall below the window, breaking anti-replay protection.

mrchalla:
Updated the PR with your review suggestions.

platform/linux-generic/odp_ipsec_sad.c
mrchalla: Ack, fixed it.

platform/linux-generic/odp_ipsec.c
mrchalla: This API is used some other places in the same way. It seems that the new and old handles are the same.

platform/linux-generic/include/odp_ipsec_internal.h
mrchalla: As per your suggestion added a new "combined_mode". But realized that removal of "seqh_len" field will introduce extra "if" checks in the data path, so had to keep it.

platform/linux-generic/odp_ipsec.c
mrchalla: I am working on replacing this "odp_packet_add_data()" API  with some other odp_packet_* APIs, will push these changes as part of the next version of this PR.

mrchalla:
updated the PR to v6:

- odp_packet_add_data() API is replaced with odp_packet_extend_tail() / odp_packet_data_move APIs, with these the same objective is achieved.

mrchalla:
Hi @JannePeltonen, Please review this PR.

platform/linux-generic/include/odp_ipsec_internal.h
JannePeltonen: > removal of "seqh_len" field will introduce extra "if" checks in the data path, so had to keep it.

The thing is that passing segh_len instead of the constant 4 to things like odp_packet_copy_from_mem() prevents optimizations by the compiler. Since the compiler cannot see that the value is always 4 (if you additionally skip the calls if it is zero), it must create a loop for the memcpy etc. So my suggestion really was about having  an SA flag that tells whether the high order bits of the sequence number are to be put in the packet after the ESP trailer for the purpose of ICV computation. Then the code that already checks the flag could use a compile time constant for the "length".

In a couple of places having segh_len still around may make sense to get rid of an unpredictable branch. But maybe even in those places you could just compute the "length" based on the flag value. (i.e. segh_len = flag * 4 or segh_len = (1- flag) * 4, depending on the definition of the flag). Such computation should be put in a function.

platform/linux-generic/odp_ipsec.c
JannePeltonen: You have not fixed the problem.

You cannot store temporary per-thread per-packet data to a variable (ipsec SA field) that is shared between threads. Multiple threads can be processing the SA at the same time and they would step on each other's toes.

You are still writing the calculated full sequence number of the currently processed packet in the SA. Adding spinlocking around the write does not fix the problem. Another thread may overwrite the field at any time once you release the spinlock.

The correct solution should be quite obvious. Take a look at how similar per-packet data is handled in the existing code.

platform/linux-generic/odp_ipsec.c
JannePeltonen: Now the comment says:
+       /* Inbound ICV computation includes ESN higher 32 bits. */

It is a true statement, but it should be better related to the subsequent code which is for non-combined mode algorithms only. How about this: "Insert high-order bits of ESN before the ICV for ICV check with non-combined mode algorithms."

platform/linux-generic/odp_ipsec.c
mrchalla: Thanks for correcting this. 
It seems the "ipsec_state_t" is used for the temporary storage of per-thread per-packet data, so I will move the "uint64_t seq64" from the SA data structure to "ipsec_state_t".

JannePeltonen: Yes, that will fix the problem.

platform/linux-generic/include/odp_ipsec_internal.h
mrchalla: Thanks for the giving deeper insight, I will try to modify this as below:

1. For the constant 4, will use this macro in places like "odp_packet_copy_from_mem()".
/* ESN higher bits length in bytes */
#define IPSEC_SEQH_LEN      4

2. For "length" at other places will use below function to compute its value:
static inline uint32_t ipsec_get_seqh_len(ipsec_sa_t *ipsec_sa)
{
        if (!ipsec_sa->icv_len || ipsec_sa->combined_mode) {
                return 0;
        else
               return IPSEC_SEQH_LEN;
}

3. Will remove the "seqh_len".

platform/linux-generic/odp_ipsec.c
mrchalla: ack, will change it.

platform/linux-generic/include/odp_ipsec_internal.h
JannePeltonen: That looks a bit awkward and probably results in the branches that you wanted to avoid. How about defining insert_seq_hi flag (or whatever it would be named) in IPsec SA so that the flag would tell exactly when high part of the seqno is inserted after the next header field. Then you would not need to special case the null algorithm using the icv lenght. And this function would not need any conditionals but could simply return ipsec_sa->insert_seq_hi * IPSEC_SEQ_HI_LEN.

mrchalla: Agreed.
1. /* ESN higher bits length in bytes */
#define IPSEC_SEQH_LEN 4
2. Will rename the "seqh_len" to  this new flag "insert_seq_hi".
3. Will set this new flag during SA creation.
```
        ipsec_sa->insert_seq_hi = (ipsec_sa->esn) ? 1 : 0;
        if (!ipsec_sa->icv_len || ipsec_sa->combined_mode)
                ipsec_sa->insert_seq_hi = 0;

```

4. New function does not have conditions.

```
static inline uint32_t ipsec_get_seqh_len(ipsec_sa_t *ipsec_sa)
{
        return (ipsec_sa->insert_seq_hi * IPSEC_SEQ_HI_LEN);
}

```

mrchalla:
Updated the PR to v7:

1. Addressed "seqh_len" part, replaced it with "insert_seq_hi" flag.
2. Added IPSEC_SEQ_HI_LEN for constant 4
3. Added a function to compute the seqh_len value using the "insert_seq_hi" flag and IPSEC_SEQ_HI_LEN.
4. Addressed other comments/suggestions.

platform/linux-generic/include/odp_ipsec_internal.h
```diff
@@ -158,6 +160,8 @@ struct ipsec_sa_s {
 	void		*context;
 	odp_queue_t	queue;
 
+	uint32_t	combined_mode;
+	uint32_t	insert_seq_hi;
```
JannePeltonen, line 164: Since these are just simple flags, there is no need to reserve a full uint32_t for each of them, especially since there is an unnamed structure of single bit flags just a little bit below this. You could put these in there as single bit flags, just like you did for the esn flag.

platform/linux-generic/include/odp_ipsec_internal.h
```diff
@@ -264,8 +269,14 @@ typedef struct odp_ipsec_sa_lookup_s {
...
 typedef struct ODP_PACKED {
-	odp_u32be_t spi;     /**< Security Parameter Index */
-	odp_u32be_t seq_no;  /**< Sequence Number */
+	/**< Security Parameter Index */
+	odp_u32be_t spi;
+
+	/**< Sequence Number */
+	union {
+		odp_u32be_t seq_no;
+		odp_u64be_t seq_no64;
```
JannePeltonen, line 278: There is no need to have both seq_no and seq64 fiels in the state. There is only one sequence number and in the 32 bit case the high order bits would just be zero. So please, have just single uint64_t seq_no here (it is a bug that the existing seq_no is odp_u32be_t instead of uint32_t, but it does not affect compiled code).

platform/linux-generic/odp_ipsec.c
```diff
@@ -376,6 +377,40 @@ typedef struct {
...
+	if (wintop_l >= (ws - 1)) {
+		if (seq < winbot_l)
+			wintop_h++;
+	/* case B: window spans two sequence number subspaces */
+	} else {
+		if (seq >= winbot_l)
+			wintop_h--;
+	}
+
+	seq64 = (((uint64_t)wintop_h << 32) | seq);
```
JannePeltonen, line 405: The outer parentheses are not needed here.

platform/linux-generic/odp_ipsec.c
```diff
@@ -376,6 +377,40 @@ typedef struct {
...
+	return seq64;
+}
+
+static inline uint32_t ipsec_get_seqh_len(ipsec_sa_t *ipsec_sa)
+{
+	return (ipsec_sa->insert_seq_hi * IPSEC_SEQ_HI_LEN);
```
JannePeltonen, line 411: The parentheses are not needed here.

platform/linux-generic/odp_ipsec.c
```diff
@@ -566,17 +601,44 @@ static int ipsec_in_esp(odp_packet_t *pkt,
 	param->auth_iv_ptr = state->iv;
 
 	state->esp.aad.spi = esp.spi;
-	state->esp.aad.seq_no = esp.seq_no;
 	state->in.seq_no = odp_be_to_cpu_32(esp.seq_no);
 
+	/* AAD with ESN */
+	if (ipsec_sa->esn) {
+		state->in.seq64 = ipsec_compute_esn(ipsec_sa, state->in.seq_no);
+		if (ipsec_sa->combined_mode)
```
JannePeltonen, line 609: The check for combined mode is not really necessary here. It was not done for the non-ESN case before and still isn't. There is not much harm in doing a possibly unnecessary write to the state structure that is written here anyway, but the branch for the combined_mode test is potentially unpredictable. And the same comment for the outbound case.

platform/linux-generic/odp_ipsec.c
```diff
@@ -566,17 +601,44 @@ static int ipsec_in_esp(odp_packet_t *pkt,
...
+	} else {
+		state->esp.aad.seq_no = esp.seq_no;
+		state->in.seq64 = state->in.seq_no;
+	}
 	param->aad_ptr = (uint8_t *)&state->esp.aad;
 
+	/* Insert high-order bits of ESN before the ICV for ICV check
+	 * with non-combined mode algorithms.
+	 */
+	if (ipsec_sa->esn && !ipsec_sa->combined_mode) {
```
JannePeltonen, line 620: The insert_seq_hi flag should be used here instead of this more complicated logic. And the same comment in the 4 other occurrences of the same test.

platform/linux-generic/odp_ipsec.c
```diff
@@ -1354,6 +1443,17 @@ static int ipsec_out_esp(odp_packet_t *pkt,
...
 
+	/* Outbound ICV computation includes ESN higher 32 bits as part of ESP
+	 * implicit trailer for individual algo's.
+	 */
+	if (ipsec_sa->esn && !ipsec_sa->combined_mode) {
+		uint32_t outb_seqh = (seq_no >> 32);
+
+		odp_packet_copy_from_mem(*pkt,
+					 esptrl_offset + _ODP_ESPTRL_LEN,
+					 IPSEC_SEQ_HI_LEN, &outb_seqh);
```
JannePeltonen, line 1454: It seems this may attempt to copy beyond the end of the packet if ICV length is less than 4 (0 for null authentication). Checkin the insert_seq_hi flag as I suggested would help a bit. But this is buggy otherwise too, see below.

platform/linux-generic/odp_ipsec.c
```diff
@@ -1368,7 +1468,8 @@ static int ipsec_out_esp(odp_packet_t *pkt,
 
 	param->auth_range.offset = ipsec_offset;
 	param->auth_range.length = state->ip_tot_len -
-				  state->ip_hdr_len -
+				  state->ip_hdr_len +
+				  ipsec_get_seqh_len(ipsec_sa) -
 				  ipsec_sa->icv_len;
 	param->hash_result_offset = state->ip_offset +
```
JannePeltonen, line 1474: This make hash result overlap with the ESN field. The code in this function is missing 1) making room in the packet for the 4 high order bytes of ESN for ICV calculation and 2) adjusting hash result offset when ESN is in use. Also, ipsec_out_esp_post() should have code for removing the high order ESN bytes and shifting the ICV to its correct place.

platform/linux-generic/odp_ipsec_sad.c
```diff
@@ -700,10 +706,16 @@ odp_ipsec_sa_t odp_ipsec_sa_create(const odp_ipsec_sa_param_t *param)
 	case ODP_AUTH_ALG_AES128_GCM:
 #endif
 	case ODP_AUTH_ALG_AES_GCM:
-		crypto_param.auth_aad_len = sizeof(ipsec_aad_t);
+		if (ipsec_sa->esn) {
+			crypto_param.auth_aad_len = 12; /* With ESN */
```
JannePeltonen, line 710: I think the "With ESN" comments in this function are unnecessary clutter since the very previous line obviously checks if ESN is used.

mrchalla:
Thanks for the review comments, I will address them and submit the next version.

platform/linux-generic/include/odp_ipsec_internal.h
mrchalla: Ack, will move.

platform/linux-generic/include/odp_ipsec_internal.h
mrchalla: Will check the feasibility.

platform/linux-generic/odp_ipsec.c
mrchalla: Ack, will remove it.

platform/linux-generic/odp_ipsec.c
mrchalla: Ack, will remove it.

platform/linux-generic/odp_ipsec.c
mrchalla: I will check the feasibility of not using "combined_mode" flag here and in outbound case as well.

platform/linux-generic/odp_ipsec.c
mrchalla: Ack, thought about it, but didn't check this. I will check the feasibility, it seems to be possible.

platform/linux-generic/odp_ipsec.c
mrchalla: Let me clarify here what we are doing:
1. We are intentionally making use of the ICV bytes place for inserting the ESN higher 4B.
2. The hash result(calculated ICV) is made to overwrite the existing ESN inserted higher 4B.
3. This way we avoided creating extra room for ESN higher 4B.

And here for "ODP_AUTH_ALG_NULL" case the "ipsec_get_seqh_len( )" will always return zero bytes.

So it seems to me that the code block functions correctly.

platform/linux-generic/odp_ipsec.c
mrchalla: Yes, checking with "insert_seq_hi" will be correct which is zero for "ODP_AUTH_ALG_NULL" case( before, this code block used to check with "seqh_len" which was zero for "ODP_AUTH_ALG_NULL"), I feel this should be sufficient here.

platform/linux-generic/odp_ipsec_sad.c
mrchalla: Ack, will remove this comment

platform/linux-generic/include/odp_ipsec_internal.h
mrchalla: yes, changed the "odp_u32be_t seq_no" to "uint64_t seq_no" and removed "odp_u64be_t seq_no64".

platform/linux-generic/odp_ipsec.c
mrchalla: Yes, everything seems fine with out the "combined_mode" flag check.
After removing these checks, it seems that there is no significance of having "combined_mode" flag, only the "insert_seq_hi" flag seems sufficient, so removed this "combined_mode" as well.

platform/linux-generic/odp_ipsec.c
mrchalla: yes, everything is fine with "insert_seq_hi" flag.

mrchalla:
Updated the PR to v8:

1. All the suggestions are addressed.
2. Removed the "combined_mode" flag as the "insert_seq_hi" flag alone sufficient.

platform/linux-generic/odp_ipsec.c
JannePeltonen: Yes, I can see what you are trying to do, but the spec for hash_result_offset says this: "If this offset lies within specified 'auth_range', implementation will mute this field before calculating the hash result." Thus your high part of ESN will get cleared before the hash gets calculated.

Now that I look at the openssl based crypto implementation in linux-gen ODP, it appears that the hash field is cleared only when checking the hash but not when generating it. That seems to be a violation of the API spec. ODP DPDK does clear the field in both cases.

mrchalla: > Now that I look at the openssl based crypto implementation in linux-gen ODP, it appears that the hash field is cleared only when checking the hash but not when generating it. That seems to be a violation of the API spec. ODP DPDK does clear the field in both cases.

Okay, as discussed in todays call, I will change this as per the current API spec.

platform/linux-generic/include/odp_ipsec_internal.h
JannePeltonen: I wrote: "There is no need to have both seq_no and seq64 fiels in the state."

So I was talking about the sequence numbers in the ipsec_state_t, not in ipsec_aad_t, but unfortunately I placed the comment in the wrong place in the github interface. Sorry about that. Now in the latest revision the state struct still has two sequence number fields and the change in ipsec_aad_t introduces a byteorder bug and should not have been done.

platform/linux-generic/include/odp_ipsec_internal.h
```diff
@@ -176,6 +178,10 @@ struct ipsec_sa_s {
 			unsigned	aes_ctr_iv : 1;
 			unsigned	udp_encap : 1;
 
+			/* Only for ESN */
```
JannePeltonen, line 181: Please remove this comment and the preceding blank line. We do not have comments like "only for UDP encapsulation" for other flags either as such things are quite obvious. And the comment is not even accurate since the flags are used also without ESN (to see that ESN is off and to see that zero bytes of ESN gets inserted in the implicit trailer.

platform/linux-generic/include/odp_ipsec_internal.h
mrchalla: > I wrote: "There is no need to have both seq_no and seq64 fiels in the state."
So I was talking about the sequence numbers in the ipsec_state_t, not in ipsec_aad_t, but unfortunately I placed the comment in the wrong place in the github interface. Sorry about that. 

Thought it "in the state" was a typo, so made changes in "ipsec_aad_t".  
Next
1) will revert the ipsec_aad_t" changes.
2) In "ipsec_state_t" will just use the "uint64_t seq_no" and will delete the seq_no64.

platform/linux-generic/odp_ipsec.c
```diff
@@ -876,6 +956,8 @@ static ipsec_sa_t *ipsec_in_single(odp_packet_t pkt,
 		goto post_lifetime_err_cnt_update;
 	}
 
+	state.in.trl_len += ipsec_get_seqh_len(ipsec_sa);
```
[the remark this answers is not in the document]
mrchalla: > This corrupts the IP length field of the received packets with individual algorithms. Apparently the patch has not been tested.

All the ESN and non-ESN tests are passed with this change.  I will if something is missing there.

 

> but a couple of lines later it is also used for subtracting the trimmed length from the IP packet length. And since the seq_hi field was not present in the actual packet received from the network, it is not included in the IP length and thus should not be subtracted from it.

you meant this line 
`state.ip_tot_len -= state.in.trl_len;`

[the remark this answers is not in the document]
mrchalla: Agreed, its correct that the ESN higher bytes are just has to be truncated from the tail, not supposed to be subtracted from the IP length, changes done as below:
1) Removed this line

> state.in.trl_len += ipsec_get_seqh_len(ipsec_sa);

2) Added the ESN higher bits part in "odp_packet_trunc_tail( )".

> odp_packet_trunc_tail(&pkt, state.in.trl_len + ipsec_get_seqh_len(ipsec_sa), NULL, NULL)

platform/linux-generic/include/odp_ipsec_internal.h
mrchalla: > will revert the ipsec_aad_t" changes.
> In "ipsec_state_t" will just use the "uint64_t seq_no" and will delete the seq_no64.
> 

These above mentioned changes are made.

JannePeltonen:
Sorry, I must have accidentally pushed a wrong button.

It seems that github does not allow me to reopen the PR since your esn_odp branch is no longer the same. I think we have to options: 1) You change your esn_odp branch back to that it was in v11 (i.e. the head would be 0e7fb48fc), we reopen the PR and then you force push v12. 2) You just create a new PR for v12 and we add appropriate references to both PRs.

mrchalla:
> You just create a new PR for v12 and we add appropriate references to both PRs.

Okay, I will try to create a new PR for V12. Thanks

mrchalla:
> > You just create a new PR for v12 and we add appropriate references to both PRs.
> 
> Okay, I will try to create a new PR for V12. Thanks

New PR created(PR 1406). Kindly add the appropriate references. Thanks.

PR #1297, opened 2021-05-27
[PATCH v3] api: timer: update timer set return codes

psavol:


include/odp/api/spec/timer.h
```diff
@@ -70,31 +70,32 @@ typedef enum {
...
+/** Timer expiration time is too near to the current time. */
+#define ODP_TIMER_TOOEARLY ODP_TIMER_TOO_NEAR
+
+/** Timer expiration time is too far from the current time. */
+#define ODP_TIMER_TOOLATE  ODP_TIMER_TOO_FAR
+
+/** Timer set operation failed because no event specified and no event
+ *  present in the timer (timer inactive/expired). */
+#define ODP_TIMER_NOEVENT ODP_TIMER_FAIL
+
```
MatiasElo, line 98: Should a comment be added that these defines are for backward compatibility and will be deprecated at some point?

example/timer/odp_timer_test.c
```diff
@@ -56,12 +56,12 @@ static const char *timerset2str(odp_timer_set_t val)
 	switch (val) {
 	case ODP_TIMER_SUCCESS:
 		return "success";
-	case ODP_TIMER_TOOEARLY:
```
MatiasElo, line 59: `ODP_TIMER_TOOEARLY` is still used in odp_timer_accuracy example.

include/odp/api/spec/timer.h
```diff
@@ -70,31 +70,32 @@ typedef enum {
...
 
+/** Timer expiration time is too near to the current time. */
+#define ODP_TIMER_TOOEARLY ODP_TIMER_TOO_NEAR
+
+/** Timer expiration time is too far from the current time. */
+#define ODP_TIMER_TOOLATE  ODP_TIMER_TOO_FAR
+
+/** Timer set operation failed because no event specified and no event
+ *  present in the timer (timer inactive/expired). */
+#define ODP_TIMER_NOEVENT ODP_TIMER_FAIL
```
JereLeppanen, line 97: Since this is the same as ODP_TIMER_FAIL, shouldn't the documentation also be the same?

example/timer/odp_timer_test.c
psavol: Fixed in v2

include/odp/api/spec/timer.h
psavol: Fixed in v2.

JereLeppanen (approved):

MatiasElo (approved):

jerinjacobk (approved):

PR #1304, opened 2021-06-04
[PATCH v6] api: cls: add queue specific statistics counters

MatiasElo:
Add classifier queue specific statistics counters (odp_cls_queue_stats_t)
and matching capabilities (odp_cls_stats_capability_t). Statistics counters
can be reset with odp_cls_queue_stats_reset() function and read
with odp_cls_queue_stats().

V2:
- Clarified `odp_cls_queue_stats_reset()` and `odp_cls_queue_stats()` argument documentation

V3:
- Removed `odp_cls_queue_stats_reset()` and clarified `odp_cls_queue_stats()` documentation

V4:
- Added implementation and validation tests
- Removed FCS from `odp_cls_queue_stats_t.octets` specification (Sunil)

### DIFF
--- a/include/odp/api/spec/classification.h
+++ b/include/odp/api/spec/classification.h
@@ -128,8 +128,9 @@ typedef union odp_cls_pmr_terms_t {
  * threshold values. RED is enabled when 'red_enable' boolean is true and
  * the resource usage is equal to or greater than the minimum threshold value.
  * Resource usage could be defined either as the percentage of pool being full
- * or the number of packets/bytes occupied in the queue depening on the platform
- * capabilities.
+ * or the number of packets/bytes occupied in the queue depending on the
+ * platform capabilities.
+ *
  * When RED is enabled for a particular flow then further incoming packets are
  * assigned a drop probability based on the size of the pool/queue.
  *
@@ -196,7 +197,7 @@ typedef struct odp_cls_capability_t {
 	/** Maximum number of CoS supported */
 	unsigned int max_cos;
 
-	/** Maximun number of queue supported per CoS
+	/** Maximun number of queues supported per CoS
 	 * if the value is 1, then hashing is not supported*/
 	unsigned int max_hash_queues;
 
@@ -227,8 +228,8 @@ typedef struct odp_cls_capability_t {
  * class of service packet drop policies
  */
 typedef enum {
-	ODP_COS_DROP_POOL,    /**< Follow buffer pool drop policy */
-	ODP_COS_DROP_NEVER,    /**< Never drop, ignoring buffer pool policy */
+	ODP_COS_DROP_POOL,      /**< Follow buffer pool drop policy */
+	ODP_COS_DROP_NEVER,     /**< Never drop, ignoring buffer pool policy */
 } odp_cls_drop_t;
 
 /**
@@ -264,12 +265,12 @@ typedef struct odp_cls_cos_param {
 	 * and application need not configure any queue to the class of service.
 	 * When hashing is disabled application has to configure the queue to
 	 * the class of service.
-	 * Depening on the implementation this number might be rounded-off to
+	 * Depending on the implementation this number might be rounded-off to
 	 * nearest supported value (e.g power of 2)
 	 */
 	uint32_t num_queue;
 
-	/** Variant mapping for queue hash configurataion */
+	/** Variant mapping for queue hash configuration */
 	union {
 		/** Mapping used when num_queue = 1, hashing is disabled in
 		 * this case and application has to configure this queue and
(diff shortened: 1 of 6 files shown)


include/odp/api/spec/classification.h
```diff
@@ -178,6 +179,60 @@ typedef struct odp_bp_param_t {
...
+
+			/** All bits of the bit field structure
+			 *
+			 *  This field can be used to set/clear all flags, or
+			 *  for bitwise operations over the entire structure. */
+			uint64_t all_counters;
+		};
+	} queue;
+
+} odp_cls_stats_capability_t;
```
yuri-tolstov, line 234: Are you planning to use other statistics elements in this structure? If not, then, probably, no need for the internal "queue" structure.

MatiasElo: Yes, additional statistics could be added later (e.g. CoS).

include/odp/api/spec/classification.h
```diff
@@ -478,6 +536,37 @@ int odp_cos_with_l3_qos(odp_pktio_t pktio_in,
...
+ *
+ * Reset all statistics counters to zero.
+ *
+ * @param cos          CoS handle
+ * @param queue        CoS event queue handle
+ *
+ * @retval  0 on success
+ * @retval <0 on failure
+ */
+int odp_cls_queue_stats_reset(odp_cos_t cos, odp_queue_t queue);
```
MatiasElo, line 550: Do we need the `cos` argument?

include/odp/api/spec/classification.h
```diff
@@ -478,6 +536,39 @@ int odp_cos_with_l3_qos(odp_pktio_t pktio_in,
 			odp_cos_t cos_table[],
 			odp_bool_t l3_preference);
 
+/**
+ * Reset statistics for a CoS queue
+ *
+ * Reset all statistics counters to zero. The queue must be assigned to the
+ * given CoS. Event queue handles can be requested with odp_cos_queue() and
```
nithind1988, line 543: "Queue must be assigned to the given cos" is probably not always true right ? Same odp_queue_t can be given to multiple CoS objects.

include/odp/api/spec/classification.h
nithind1988: In our HW, we would need cos argument as a Scheduler queues are independent entities and can be associated to multiple CoS objects.

I guess even odp_cls_cos_create() can take same odp_queue_t as input (num_queue = 1) multiple times.

MatiasElo: Ok

include/odp/api/spec/classification.h
MatiasElo: Removed in V3

include/odp/api/spec/classification.h
```diff
@@ -178,6 +179,60 @@ typedef struct odp_bp_param_t {
 	odp_threshold_t threshold;
 } odp_bp_param_t;
 
+/**
+ * Classifier queue specific statistics counters
+ */
+typedef struct odp_cls_queue_stats_t {
+	/** Number of octets in packets delivered to the queue. In case of
+	 *  Ethernet, packet size includes MAC header and FCS. */
```
sunilkori, line 187: @MatiasElo Hi, This is Sunil Kumar Kori from Marvell. Sorry for commenting from my personal account as my official account is not working for some reason. 

As per spec, octets contains FCS also. So I just wanted to understand the use case of adding it. Because in general cases FCS are striped  at H/W and remaining packet size is updated as a stats.

MatiasElo: Hi @sunilkori, I added this text to be compatible with the original `odp_pktio_stats_t.in_octets` spec. If this proves to be problematic we'll have to also modify the old API as these counter definitions should match.

An easy implementation fix would  be to just add  `odp_cls_queue_stats_t.packets * 4` to the value returned by HW.

sunilkori: That I understand and easily doable . My only point is to understand the usage of this. But if there is no specific use case then I believe, changing ODP spec is better.

MatiasElo: Fixed in V4

nithind1988 (approved):
Looks good to me.

platform/linux-generic/odp_classification.c
```diff
@@ -1770,6 +1780,32 @@ uint64_t odp_pmr_to_u64(odp_pmr_t hdl)
...
+	queue_idx = _odp_cos_queue_idx(cos, queue);
+	if (odp_unlikely(queue_idx < 0)) {
+		ODP_ERR("Invalid odp_queue_t handle\n");
+		return -1;
+	}
+
+	if (stats)
+		memset(stats, 0, sizeof(odp_cls_queue_stats_t));
+
+	stats->discards = odp_atomic_load_u64(&cos->s.stats[queue_idx].discards);
```
psavol, line 1803: May be it better to return failure if stats == NULL.

Now if (stats) check applies only to memset, but not actual stat writes.

MatiasElo: Fixed in V5.

psavol (approved):

PR #1305, opened 2021-06-04
[PATCH v6] api: tm: add queue specific statistics counters

MatiasElo:
Add traffic manager queue specific statistics counters
(odp_tm_queue_stats_t) and matching capabilities
(odp_tm_stats_capability_t). Statistics counters can be reset with
odp_tm_queue_stats_reset() function and read with odp_tm_queue_stats().

V2:
- Removed `odp_tm_queue_stats_reset()`

V3:
- Renamed `odp_tm_stats_capability_t` -> `odp_tm_queue_stats_capability_t`
- Added implementation and validation test

V4:
- Added `odp_tm_queue_stats_t.discard_octets` and `odp_tm_queue_stats_t.discard_packets`

V5:
- Renamed `odp_tm_queue_stats_t.discard_packets` to `odp_tm_queue_stats_t.discards`

### DIFF
--- a/include/odp/api/spec/traffic_mngr.h
+++ b/include/odp/api/spec/traffic_mngr.h
@@ -37,7 +37,7 @@ extern "C" {
  * based systems or one or more hybrid systems - where because of
  * hardware constraints some of the packet scheduling is done in hardware
  * and some is done in software.  In addition, there may also be additional
- * API's beyond those described here for (a) controlling advanced capabilities
+ * APIs beyond those described here for (a) controlling advanced capabilities
  * supported by specific hardware, software or hybrid subsystems or (b)
  * dealing with constraints and limitations of specific implementations.
  */
@@ -283,7 +283,7 @@ typedef struct {
 	 * proper TM shaping.  Note that TM Shaping is NOT the same thing as
 	 * Ingress Metering/Policing as specified by RFC 2697 (A Single Rate
 	 * Three Color Marker) or RFC 2698 (A Two Rate Three Color Marker).
-	 * These RFC's can be used for a Diffserv traffic conditioner, or
+	 * These RFCs can be used for a Diffserv traffic conditioner, or
 	 * other ingress policing.  They make no mention of and have no
 	 * algorithms for delaying packets - which is what TM shapers are
 	 * expected to do. */
@@ -437,7 +437,7 @@ typedef struct {
 	 * disciplines. */
 	odp_bool_t fair_queuing_needed;
 
-	/** weights_needd indicates that the tm_node schedulers at this
+	/** weights_needed indicates that the tm_node schedulers at this
 	 * level are expected have different weights for their different
 	 * fanins.  When true the min_weight and max_weight fields above
 	 * specify the used range of such weights. */
@@ -494,7 +494,7 @@ typedef struct {
 	 * the application will not enable this color for vlan marking,
 	 * ecn marking nor drop precedence marking.  A value of TRUE means that
 	 * the application expects to use this color in conjunction with one or
-	 * more of the marking API's. */
+	 * more of the marking APIs. */
 	odp_bool_t marking_colors_needed[ODP_NUM_PACKET_COLORS];
 
 	/** The per_level array specifies the TM system requirements that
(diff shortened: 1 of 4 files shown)


JannePeltonen:
This probably needs to be harmonized with the statistics proposed in PR 1237.

Maybe the statistics to be enabled should be configurable so that they do not need to be tracked when not needed.

include/odp/api/spec/traffic_mngr.h
```diff
@@ -385,6 +440,10 @@ typedef struct {
 	 * the parameters of the threshold profile of any TM node or TM queue.
 	 */
 	odp_bool_t dynamic_threshold_update;
+
+	/** Statistics counters capabilities */
+	odp_tm_stats_capability_t stats;
```
nithind1988, line 445: For node stats, capabilities should be per level. So Maybe name it explicitly as odp_tm_queue_stats_capability ?

MatiasElo: Makes sense if we don't require additional TM level statistics capabilities. Do you know if this is the case?

MatiasElo: Fixed in V3

nithind1988: Missed to respond to your comment. If you mean other global TM statistics apart from queue stats, then I think yes, atleast our HW doesn't support such stats which are global and not queue related.

include/odp/api/spec/traffic_mngr.h
```diff
@@ -189,6 +190,56 @@ extern "C" {
...
+	/** Number of octets in successfully transmitted packets. In case of
+	 *  Ethernet, packet size includes MAC header. */
+	uint64_t octets;
+
+	/** Number of successfully transmitted packets. */
+	uint64_t packets;
+
+	/** Number of packets discarded due to other reasons (e.g. aging) than
+	 *  errors. */
+	uint64_t discards;
```
psavol, line 206: Discarded octets counter will be added.

MatiasElo: Added in V4

nithind1988 (approved):
Looks good to me.

include/odp/api/spec/traffic_mngr.h
```diff
@@ -1455,14 +1459,14 @@ int odp_tm_node_wred_config(odp_tm_node_t tm_node,
 /** odp_tm_node_lookup() can be used to find the tm_node object created with
  * the specified name.
  *
- * @param odp_tm  Odp_tm is used to identify the TM system into which this
- *                odp_tm_node object is created.
+ * @param tm      Handle of the TM system into which this odp_tm_node object is
+ *                created.
```
psavol, line 1525: Original text has copy-paste error from create call - this is lookup.

include/odp/api/spec/traffic_mngr.h
```diff
@@ -189,6 +190,63 @@ extern "C" {
...
+	/** Number of successfully transmitted packets. */
+	uint64_t packets;
+
+	/** Number of octets in packets discarded due to other reasons (e.g.
+	 * aging) than errors. */
+	uint64_t discard_octets;
+
+	/** Number of packets discarded due to other reasons (e.g. aging) than
+	 *  errors. */
+	uint64_t discard_packets;
```
psavol, line 210: This could be simply "discards" as in other APIs. If other APIs add discard_octets later, we have the same logic in all APIs (discards + discard_octets)

include/odp/api/spec/traffic_mngr.h
```diff
@@ -189,6 +190,63 @@ extern "C" {
...
+
+	/** Number of packets which transmission errors. */
+	uint64_t errors;
```
psavol, line 213: type: ... with ....

include/odp/api/spec/traffic_mngr.h
MatiasElo: Fixed in V5

include/odp/api/spec/traffic_mngr.h
MatiasElo: Fixed in V5

include/odp/api/spec/traffic_mngr.h
MatiasElo: Fixed in V5

psavol (approved):

PR #1306, opened 2021-06-08
[PATCH v6] example: ipsec_api: enhance application tests

asasidharan:
Enable support for SHA256 auth algorithm in AH mode  and make this the default algorithm for AH tests.
Enable live and router mode tests by default and skip those tests for linux-generic platform.


example/ipsec_crypto/odp_ipsec_stream.c
```diff
@@ -359,15 +360,32 @@ odp_packet_t create_ipv4_packet(stream_db_entry_t *stream,
...
+		else if (entry->ah.alg == ODP_AUTH_ALG_SHA1_HMAC)
+			HMAC(EVP_sha1(),
+			     entry->ah.key.data,
+			     entry->ah.key.length,
+			     (uint8_t *)ip,
+			     auth_len,
+			     hash,
+			     NULL);
+		else if (entry->ah.alg == ODP_AUTH_ALG_SHA256_HMAC)
+			HMAC(EVP_sha256(),
```
JannePeltonen, line 380: Instead of expanding the existing HMAC call to this if-construct here and in verify_ipv4_packet(), the digest argument could be parameterized and stored in ipsec_cache_entry_t in create_ipsec_cache_entry(). It would make the code shorter and simpler, especially if more algorithms get added later.

Except for the above, this PR looks good to me.

asasidharan: Ack. Updated v4 to save the digest method in ipsec_cache_entry_t.

Please note that I have added a new commit to the PR which fixes a bug in IPsec Live and router mode scripts.

JannePeltonen: Actually, saving the digest method like that makes odp_ipsec_sa_db.c depend on OpenSSL which breaks the build in ipsec_api example when openssl is not enabled. I think direct OpenSSL dependencies need to be kept in odp_ipsec_stream.[ch].

It seems the digest method could be saved in stream_db_entry_t and initialized in resolve_stream_db(). Or if you would rather go back to the previous version where the digest method is not stored, then I would suggest defining a function that returns the method based on the ODP auth algorithm so that the somewhat ugly if construct would not have to be repeated.

asasidharan: Thanks for pointing this out. I guess CI currently doesn't cover the openssl disabled case. Since the CI passed, I was under the impression that the changes didn't break any cases.

With v5, I have updated the change to save the digest method in stream_db_entry_t as suggested. Let me know if it is fine now.

example/ipsec_crypto/odp_ipsec_stream.c
```diff
@@ -139,6 +139,27 @@ int create_stream_db_entry(char *input)
 	return 0;
 }
 
+static const void *get_evp_md(odp_auth_alg_t auth)
```
JannePeltonen, line 142: I think this should return const EVP_MD *. Then there would be no need for any casting in this function or elsewhere.

JannePeltonen:
Looks good apart from the type cast issue that I commented. You can add my review tag while fixing that.

I think it would be better if the patch that disables the live and router mode tests in linux-gen came before the patch that enables the tests in CI. That way CI would not be broken after any of these commits. So maybe you could swap their order.

asasidharan:
> I think it would be better if the patch that disables the live and router mode tests in linux-gen came before the patch that enables the tests in CI. That way CI would not be broken after any of these commits. So maybe you could swap their order.

Ack.

With v6:

- using "const EVP_MD *" in place of "const void *".
- Reordered the CI enable commit for IPsec live and router mode tests.
- Added Reviewed-by tags.

example/ipsec_crypto/odp_ipsec_stream.c
asasidharan: Ack. Fixed in v6.

JannePeltonen (approved):

PR #1307, opened 2021-06-08
[PATCH v3] api: pktio: add implementation specific extra statistics counters

MatiasElo:
Add new functions for reading ODP implementation specific custom packet IO
statistics counters.

V2:
- Removed mention of `odp_pktio_stats_reset()`. PR #1288 will extend `odp_pktio_stats_reset()` documentation.

### DIFF
--- a/include/odp/api/spec/packet_io_stats.h
+++ b/include/odp/api/spec/packet_io_stats.h
@@ -26,6 +26,12 @@ extern "C" {
  */
 
 /**
+ * @def ODP_PKTIO_STATS_EXTRA_NAME_LEN
+ * Maximum packet IO extra statistics counter name length in chars including
+ * null char
+ */
+
+/**
  * Packet IO statistics counters
  *
  * In the counter definitions the term successfully refers to packets which were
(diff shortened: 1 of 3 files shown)


include/odp/api/spec/packet_io_stats.h
```diff
@@ -111,6 +126,69 @@ int odp_pktio_stats(odp_pktio_t pktio, odp_pktio_stats_t *stats);
  */
 int odp_pktio_stats_reset(odp_pktio_t pktio);
 
+/**
+ * Get extra statistics counter information for pktio handle
```
psavol, line 130: ... information for a packet IO interface

Spec text is more readable when the target of an operation is the interface, rather than the handle pointing to the interface.

include/odp/api/spec/packet_io_stats.h
```diff
@@ -111,6 +126,69 @@ int odp_pktio_stats(odp_pktio_t pktio, odp_pktio_stats_t *stats);
...
+ *
+ * @param       pktio    Packet IO handle
+ * @param[out]  info     Array of extra statistics info structs for output
+ * @param       num      Maximum number of info structs to output
+ *
+ * @return Number of extra statistics
+ * @retval <0 on failure
+ */
+int odp_pktio_extra_stat_info(odp_pktio_t pktio,
+			      odp_pktio_extra_stat_info_t info[], uint32_t num);
```
psavol, line 151: Application needs to compare return value and num. Compiler complains about comparison between signed and unsigned integers. Should we change type of num to int.

E.g. queue enq multi has "int num"
int odp_queue_enq_multi(odp_queue_t queue, const odp_event_t events[], int num)

include/odp/api/spec/packet_io_stats.h
MatiasElo: Fixed in V3.

psavol:
API looks good in v3

PR #1311, opened 2021-06-14
[PATCH v12] api: ipsec: add destination queue capability

asasidharan:
In some ODP implementations, plain queues might be using software
primitives such as a simple ring and the ODP IPsec hardware
can enqueue completion events only into schedule queues.
Introducing capability queue_type_sched and queue_type_plain
which define the queue types that can be the destination of the
IPsec events.

### DIFF
--- a/example/ipsec_api/odp_ipsec.c
+++ b/example/ipsec_api/odp_ipsec.c
@@ -921,6 +921,7 @@ main(int argc, char *argv[])
 	odp_shm_t shm;
 	odp_cpumask_t cpumask;
 	char cpumaskstr[ODP_CPUMASK_STR_SIZE];
+	odp_ipsec_capability_t ipsec_capa;
 	odp_pool_param_t params;
 	odp_instance_t instance;
 	odp_init_t init_param;
@@ -961,6 +962,23 @@ main(int argc, char *argv[])
 	if (odp_init_local(instance, ODP_THREAD_CONTROL)) {
 		ODPH_ERR("Error: ODP local init failed.\n");
 		exit(EXIT_FAILURE);
+	}
+
+	if (odp_ipsec_capability(&ipsec_capa)) {
+		ODPH_ERR("Error: IPsec capability request failed.\n");
+		exit(EXIT_FAILURE);
+	}
+
+	if (queue_create == polled_odp_queue_create) {
+		if (!ipsec_capa.queue_type_plain) {
+			ODPH_ERR("Error: Plain type dest queue not supported.\n");
+			exit(EXIT_FAILURE);
+		}
+	} else {
+		if (!ipsec_capa.queue_type_sched) {
+			ODPH_ERR("Error: scheduled type dest queue not supported.\n");
+			exit(EXIT_FAILURE);
+		}
 	}
 
 	/* Reserve memory for args from shared mem */
--- a/include/odp/api/spec/ipsec.h
+++ b/include/odp/api/spec/ipsec.h
@@ -338,6 +338,28 @@ typedef struct odp_ipsec_capability_t {
 	 */
 	uint32_t max_cls_cos;
 
+	/**
+	 * Scheduled queue support
+	 *
+	 * 0: Scheduled queues are not supported either as IPsec SA destination
+	 *    queues or as IPsec default queue
+	 * 1: Scheduled queues are supported as both IPsec SA destination queues
+	 *    and IPsec default queue
+	 * @see odp_ipsec_sa_param_t
+	 */
+	odp_bool_t queue_type_sched;
+
+	/**
+	 * Plain queue support
+	 *
+	 * 0: Plain queues are not supported either as IPsec SA destination
+	 *    queues or as IPsec default queue
+	 * 1: Plain queues are supported as both IPsec SA destination queues and
+	 *    IPsec default queue
+	 * @see odp_ipsec_sa_param_t
+	 */
+	odp_bool_t queue_type_plain;
+
 	/** Maximum number of different destination queues. The same queue may
 	 *  be used for many SAs. */
 	uint32_t max_queues;
--- a/test/performance/odp_ipsec.c
+++ b/test/performance/odp_ipsec.c
@@ -1030,6 +1030,7 @@ int main(int argc, char *argv[])
 	odph_odpthread_t thr[num_workers];
 	odp_instance_t instance;
 	odp_init_t init_param;
+	odp_ipsec_capability_t ipsec_capa;
 	odp_pool_capability_t capa;
 	odp_ipsec_config_t config;
 	uint32_t max_seg_len;
@@ -1091,6 +1092,21 @@ int main(int argc, char *argv[])
 		exit(EXIT_FAILURE);
 	}
 	odp_pool_print(pool);
+
+	if (odp_ipsec_capability(&ipsec_capa) < 0) {
+		app_err("IPSEC capability call failed.\n");
+		exit(EXIT_FAILURE);
+	}
+
+	if (cargs.schedule && !ipsec_capa.queue_type_sched) {
+		app_err("Scheduled type destination queue not supported.\n");
+		exit(EXIT_FAILURE);
+	}
+
+	if (cargs.poll && !ipsec_capa.queue_type_plain) {
+		app_err("Plain type destination queue not supported.\n");
+		exit(EXIT_FAILURE);
+	}
 
 	odp_ipsec_config_init(&config);
 	config.max_num_sa = 2;
--- a/test/validation/api/ipsec/ipsec.c
+++ b/test/validation/api/ipsec/ipsec.c
@@ -516,6 +516,28 @@ static int send_pkts(const ipsec_test_part part[], int num_part)
 	return num_part;
 }
 
+/* Receive async inbound packet */
+static odp_event_t recv_pkt_async_inbound(odp_ipsec_op_status_t status)
+{
+	odp_event_t event;
+	odp_queue_t queue;
+
+	/*
+	 * In case of SA lookup failure, the event is enqueued to the default
+	 * queue specified during odp_ipsec_config()
+	 */
+	if (status.error.sa_lookup == 0)
+		queue = suite_context.queue;
+	else
+		queue = suite_context.default_queue;
+
+	do {
+		event = odp_queue_deq(queue);
+	} while (event == ODP_EVENT_INVALID);
+
+	return event;
+}
+
 /* Receive inline processed packets */
 static int recv_pkts_inline(const ipsec_test_part *part,
 			    odp_packet_t *pkto)
@@ -625,9 +647,7 @@ static int ipsec_process_in(const ipsec_test_part *part,
 			odp_event_t event;
 			odp_event_subtype_t subtype;
 
-			do {
-				event = odp_queue_deq(suite_context.queue);
-			} while (event == ODP_EVENT_INVALID);
+			event = recv_pkt_async_inbound(part->out[i].status);
 
 			CU_ASSERT(odp_event_is_valid(event) == 1);
 			CU_ASSERT_EQUAL(ODP_EVENT_PACKET,
@@ -1060,6 +1080,7 @@ int ipsec_init(odp_instance_t *inst, odp_ipsec_op_mode_t mode)
 	odp_pool_t pool;
 	odp_queue_t out_queue;
 	odp_pool_capability_t pool_capa;
+	odp_queue_t default_queue;
 	odp_pktio_t pktio;
 	odp_init_t init_param;
 	odph_helper_options_t helper_options;
@@ -1122,6 +1143,12 @@ int ipsec_init(odp_instance_t *inst, odp_ipsec_op_mode_t mode)
 			fprintf(stderr, "IPsec outq creation failed.\n");
 			return -1;
 		}
+
+		default_queue = odp_queue_create("ipsec-default", NULL);
+		if (ODP_QUEUE_INVALID == default_queue) {
+			fprintf(stderr, "IPsec defaultq creation failed.\n");
+			return -1;
+		}
 	}
 
 	if (mode == ODP_IPSEC_OP_MODE_INLINE) {
@@ -1166,7 +1193,7 @@ int ipsec_config(odp_instance_t ODP_UNUSED inst)
 	ipsec_config.inbound_mode = suite_context.inbound_op_mode;
 	ipsec_config.outbound_mode = suite_context.outbound_op_mode;
 	ipsec_config.outbound.all_chksum = ~0;
-	ipsec_config.inbound.default_queue = suite_context.queue;
+	ipsec_config.inbound.default_queue = suite_context.default_queue;
 	ipsec_config.inbound.parse_level = ODP_PROTO_LAYER_ALL;
 	ipsec_config.inbound.chksums.all_chksum = ~0;
 	ipsec_config.stats_en = true;
@@ -1243,6 +1270,7 @@ int ipsec_term(odp_instance_t inst)
 {
 	odp_pool_t pool = suite_context.pool;
 	odp_queue_t out_queue = suite_context.queue;
+	odp_queue_t default_queue = suite_context.default_queue;
 	/* suite_context.pktio is set to ODP_PKTIO_INVALID by ipsec_suite_init()
 	   if inline processing is not supported. */
 	odp_pktio_t pktio = odp_pktio_lookup("loop");
@@ -1255,6 +1283,11 @@ int ipsec_term(odp_instance_t inst)
 	if (ODP_QUEUE_INVALID != out_queue) {
 		if (odp_queue_destroy(out_queue))
 			fprintf(stderr, "IPsec outq destroy failed.\n");
+	}
+
+	if (ODP_QUEUE_INVALID != default_queue) {
+		if (odp_queue_destroy(default_queue))
+			fprintf(stderr, "IPsec defaultq destroy failed.\n");
 	}
 
 	if (ODP_POOL_INVALID != pool) {
--- a/test/validation/api/ipsec/ipsec.c
+++ b/test/validation/api/ipsec/ipsec.c
@@ -100,6 +100,14 @@ static int pktio_start(odp_pktio_t pktio, odp_bool_t in, odp_bool_t out)
 	return 1;
 }
 
+static odp_event_t recv_event(odp_queue_t queue)
+{
+	if (odp_queue_type(queue) == ODP_QUEUE_TYPE_PLAIN)
+		return odp_queue_deq(queue);
+	else
+		return odp_schedule(NULL, ODP_SCHED_NO_WAIT);
+}
+
 static void pktio_stop(odp_pktio_t pktio)
 {
 	odp_queue_t queue = ODP_QUEUE_INVALID;
@@ -110,7 +118,7 @@ static void pktio_stop(odp_pktio_t pktio)
 		fprintf(stderr, "IPsec pktio stop failed.\n");
 
 	while (1) {
-		odp_event_t ev = odp_queue_deq(queue);
+		odp_event_t ev = recv_event(queue);
 
 		if (ev != ODP_EVENT_INVALID)
 			odp_event_free(ev);
@@ -138,6 +146,17 @@ int ipsec_check(odp_bool_t ah,
 	    (ODP_IPSEC_OP_MODE_INLINE == suite_context.outbound_op_mode &&
 	     ODP_SUPPORT_NO == capa.op_mode_inline_out))
 		return ODP_TEST_INACTIVE;
+
+	if (!(ODP_IPSEC_OP_MODE_SYNC == suite_context.inbound_op_mode &&
+	      ODP_IPSEC_OP_MODE_SYNC == suite_context.outbound_op_mode) &&
+	    ODP_QUEUE_INVALID != suite_context.queue) {
+		if (suite_context.q_type == ODP_QUEUE_TYPE_PLAIN &&
+		    !capa.queue_type_plain)
+			return ODP_TEST_INACTIVE;
+		if (suite_context.q_type == ODP_QUEUE_TYPE_SCHED &&
+		    !capa.queue_type_sched)
+			return ODP_TEST_INACTIVE;
+	}
 
 	/* suite_context.pktio is set to ODP_PKTIO_INVALID in ipsec_suite_init()
 	 * if the pktio device doesn't support inline IPsec processing. */
@@ -357,7 +376,7 @@ void ipsec_sa_destroy(odp_ipsec_sa_t sa)
 
 	if (ODP_QUEUE_INVALID != suite_context.queue) {
 		do {
-			event = odp_queue_deq(suite_context.queue);
+			event = recv_event(suite_context.queue);
 		} while (event == ODP_EVENT_INVALID);
 
 		CU_ASSERT(odp_event_is_valid(event) == 1);
@@ -532,7 +551,7 @@ static odp_event_t recv_pkt_async_inbound(odp_ipsec_op_status_t status)
 		queue = suite_context.default_queue;
 
 	do {
-		event = odp_queue_deq(queue);
+		event = recv_event(queue);
 	} while (event == ODP_EVENT_INVALID);
 
 	return event;
@@ -554,7 +573,7 @@ static int recv_pkts_inline(const ipsec_test_part *part,
 		odp_event_t ev;
 		odp_event_subtype_t subtype;
 
-		ev = odp_queue_deq(queue);
+		ev = recv_event(queue);
 		if (ODP_EVENT_INVALID != ev) {
 			CU_ASSERT(odp_event_is_valid(ev) == 1);
 			CU_ASSERT_EQUAL(ODP_EVENT_PACKET,
@@ -569,7 +588,7 @@ static int recv_pkts_inline(const ipsec_test_part *part,
 			continue;
 		}
 
-		ev = odp_queue_deq(suite_context.queue);
+		ev = recv_event(suite_context.queue);
 		if (ODP_EVENT_INVALID != ev) {
 			odp_packet_t pkt;
 			int num_pkts = 0;
@@ -701,7 +720,7 @@ static int ipsec_send_out_one(const ipsec_test_part *part,
 			odp_event_subtype_t subtype;
 
 			do {
-				event = odp_queue_deq(suite_context.queue);
+				event = recv_event(suite_context.queue);
 			} while (event == ODP_EVENT_INVALID);
 
 			CU_ASSERT(odp_event_is_valid(event) == 1);
@@ -784,7 +803,7 @@ static int ipsec_send_out_one(const ipsec_test_part *part,
 			odp_event_t ev;
 			odp_event_subtype_t subtype;
 
-			ev = odp_queue_deq(queue);
+			ev = recv_event(queue);
 			if (ODP_EVENT_INVALID != ev) {
 				CU_ASSERT(odp_event_is_valid(ev) == 1);
 				CU_ASSERT_EQUAL(ODP_EVENT_PACKET,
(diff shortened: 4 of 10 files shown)


include/odp/api/spec/ipsec.h
```diff
@@ -335,6 +335,28 @@ typedef struct odp_ipsec_capability_t {
 	 */
 	uint32_t max_cls_cos;
 
+	/**
+	 * Scheduled destination queue support
+	 *
+	 * This defines whether scheduled queues are supported as destination
+	 * queue.
```
JannePeltonen, line 342: This API change looks ok to me. I would just remove the "this defines..." sentences since they do not seem to add anything to what is otherwise anyway said.

example/ipsec_api/odp_ipsec.c
```diff
@@ -963,6 +964,21 @@ main(int argc, char *argv[])
...
+		ODPH_ERR("Error: IPsec capability request failed.\n");
+		exit(EXIT_FAILURE);
+	}
+
+	if ((NULL == getenv("ODP_IPSEC_USE_POLL_QUEUES")) &&
+	    (ipsec_capa.queue_type_sched == 0)) {
+		ODPH_ERR("Error: scheduled type dest queue not supported.\n");
+		exit(EXIT_FAILURE);
+	} else if ((NULL != getenv("ODP_IPSEC_USE_POLL_QUEUES")) &&
+		    ipsec_capa.queue_type_plain == 0) {
```
JannePeltonen, line 977: How about:

```
if (queue_create == polled_odp_queue_create) {
	if (!ipsec_capa.queue_type_plain) {
		blahblah
	}
} else {
	if (!ipsec_capa.queue_type_sched) {
		blahblah
	}
}
```

test/performance/odp_ipsec.c
```diff
@@ -809,6 +809,16 @@ run_measure_one_config(ipsec_args_t *cargs,
 		return -1;
 	}
 
+	if (cargs->schedule && capa.queue_type_sched == 0) {
```
JannePeltonen, line 812: capa.queue_type_sched is odp_boolean_t so I think '!capa.queue_type_sched' (or 'capa.queue_type_sched == false') would be better.

test/validation/api/ipsec/ipsec.c
```diff
@@ -1054,7 +1054,7 @@ int ipsec_out_term(void)
 	return ipsec_suite_term();
 }
 
-int ipsec_init(odp_instance_t *inst, odp_ipsec_op_mode_t mode)
+int ipsec_init(odp_instance_t *inst)
```
JannePeltonen, line 1157: I would have wished for a better commit message for this change. Yes, the mode argument can be removed if we ensure this and that as the message says, but why do we want to do it, what is the purpose of the change? Why is it better to pass mode through suite context instead of the mode argument?

test/validation/api/ipsec/ipsec.c
```diff
@@ -1115,16 +1115,19 @@ int ipsec_init(odp_instance_t *inst, odp_ipsec_op_mode_t mode)
 		fprintf(stderr, "Packet pool creation failed.\n");
 		return -1;
 	}
-	if (mode == ODP_IPSEC_OP_MODE_ASYNC ||
-	    mode == ODP_IPSEC_OP_MODE_INLINE) {
+
+	if (suite_context.inbound_op_mode == ODP_IPSEC_OP_MODE_ASYNC ||
+	    suite_context.outbound_op_mode == ODP_IPSEC_OP_MODE_ASYNC ||
+	    suite_context.outbound_op_mode == ODP_IPSEC_OP_MODE_INLINE) {
```
JannePeltonen, line 1226: If inbound_op_mode is inline and outbound_op_mode is sync, this function fails to create the default queue.

test/validation/api/ipsec/ipsec.c
```diff
@@ -1119,9 +1222,9 @@ int ipsec_init(odp_instance_t *inst)
 	if (suite_context.inbound_op_mode == ODP_IPSEC_OP_MODE_ASYNC ||
 	    suite_context.outbound_op_mode == ODP_IPSEC_OP_MODE_ASYNC ||
 	    suite_context.outbound_op_mode == ODP_IPSEC_OP_MODE_INLINE) {
-		out_queue = odp_queue_create("ipsec-out", NULL);
-		if (ODP_QUEUE_INVALID == out_queue) {
-			fprintf(stderr, "IPsec outq creation failed.\n");
+		default_queue = odp_queue_create("ipsec-default", NULL);
```
JannePeltonen, line 1227: If plain queues are not supported as destination queues, then I think it means the default queue cannot be a plain queue either.

I also do not see why the default queue was made a separate from the destination queue (previously they were the same). Maybe that improves the test if events are dequeued only from the expected queue. But that seems like an unrelated change that should be in a different patch.

include/odp/api/spec/ipsec.h
asasidharan: Ack. Will remove these sentences in next version.

example/ipsec_api/odp_ipsec.c
asasidharan: Agreed. Will update so in next version.

test/performance/odp_ipsec.c
asasidharan: Initially I kept it  'capa.queue_type_sched == false' but it gave checkpatch error so kept it this way. I will update this to '!capa.queue_type_sched'.

test/validation/api/ipsec/ipsec.c
asasidharan: I feel that the mode argument is creating an unnecessary copy of information which could be directly inferred from the suite_context if we make sure that suite_context is properly initialized. My thinking was that this change would bring some improvement in the code readability.

test/validation/api/ipsec/ipsec.c
asasidharan: Now when I look at this again, from the spec what I understand is that, default_queue will be used only where inbound lookup failure happens. And that is dependent only on the inbound_op_mode. So I think the following if conditional would be enough for the default_queue creation:

```
if (suite_context.inbound_op_mode == ODP_IPSEC_OP_MODE_ASYNC ||
   suite_context.inbound_op_mode == ODP_IPSEC_OP_MODE_INLINE) {
   ...
}
```

Please correct me if I am wrong.

test/validation/api/ipsec/ipsec.c
asasidharan: > If plain queues are not supported as destination queues, then I think it means the default queue cannot be a plain queue either.

Updated the validation test to create default queue based on the new capability. On a related note, updated the API spec to say that the new queue type capability applies to both IPsec SA destination queue as well as the global IPsec default queue.

> I also do not see why the default queue was made a separate from the destination queue (previously they were the same). Maybe that improves the test if events are dequeued only from the expected queue. But that seems like an unrelated change that should be in a different patch.

Ack. Created a separate commit to introduce separate default queue into the test.

asasidharan:
With v3:

- Updated API text to say that the queue capability applies for both IPsec SA destination queue as well as IPsec default queue.
- Updated capability checks in example and perf applications as suggested
- Updated commit message for "validation: ipsec: remove mode arg from ipsec_init"
- Created a separate commit to introduce separate default queue into the test

test/validation/api/ipsec/ipsec.c
JannePeltonen: I think that is correct. The default queue is used when SA is not known (if SA was known, then the SA specific queue would be used), and in outbound processing the SA is always known.

test/validation/api/ipsec/ipsec.c
JannePeltonen: I am still not quite sure if removing the argument and relying on the global variable (and it being initialized early enough) is an improvement. But I do not feel too strongly about it. Actually, it seems the relevant piece of code gets moved from this function elsewehere in another patch and if the patches are rearranged as I suggest in another commit, then this patch mostly disappears.

include/odp/api/spec/ipsec.h
```diff
@@ -337,6 +337,28 @@ typedef struct odp_ipsec_capability_t {
 	 */
 	uint32_t max_cls_cos;
 
+	/**
+	 * Scheduled queue support
+	 *
+	 * 0: Scheduled queues are not supported as IPsec SA destination queues
+	 *    or IPsec default queue
+	 * 1: Scheduled queues are supported as IPsec SA destination queues or
```
JannePeltonen, line 345: The 'or' should be 'and'. If this capability field is true, then scheduled queues are supported both as destination queues and as the default queue. And the same comment for the plain queue capability below.

test/validation/api/ipsec/ipsec.c
JannePeltonen: I think the patches should be ordered differently. The set of patches that are directly related to the new queue type capability API change should come first. The other three patches (mode arg removal, separate default queue, time out) should come last. This way separate topics do not get mixed, making both the changes and the commit history more readable.

The other patches could have even been in a different PR (and we do not interleave commits of different PRs either, when merging), but it is ok to have them in this PR as well, after the API change related patches.

asasidharan: > The other patches could have even been in a different PR (and we do not interleave commits of different PRs either, when merging), but it is ok to have them in this PR as well, after the API change related patches.

@JannePeltonen Bringing the default queue change after the API spec related changes looks challenging.

Once the queue type capability change is introduced in the validation test, I am forced to create and destroy the ipsec destination queue (suite_context.queue) per suite. But the "default_queue" passed to odp_ipsec_config() is global and cannot change per suite. So I am not able to pass the destination queue as default queue leading to segfault for lookup failure cases.

So in my opinion, the default queue related changes on validation test should come before the API spec related changes.

If you are OK with

1. moving the default queue related commits before API spec change commits in the same PR, I can do that way.
2. Or if you still prefer to create a separate PR for the default queue related changes, I can do that way.

JannePeltonen: Both are fine. No need to create a new PR if you prefer that as I think this whole PR probably gets into a mergeable state soon.

The point was just to keep the API change unrelated improvements not interleaved with the dest_queue capability related stuff as I think that makes the individual patches easier to read. Whether they come before or after does not matter that much, so pick the order that you think makes sense.

include/odp/api/spec/ipsec.h
asasidharan: In v4, updated the text as follows:

```
+        * 0: Scheduled queues are not supported either as IPsec SA destination
+        *    queues or as IPsec default queue
+        * 1: Scheduled queues are supported as both IPsec SA destination queues
+        *    and IPsec default queue
```

Let me know if this looks OK.

test/validation/api/ipsec/ipsec.c
asasidharan: Ack. With v4, I have changed the order of the commits so that the API change related changes are not interleaved.

asasidharan:
With v4:

- Improved API text
- Reordered commits so that the API related commits are not interleaved
- Added more detail to commit description for "validation: ipsec: set 1min time out for event receive "

test/validation/api/ipsec/ipsec.c
JannePeltonen: I am still not convinced about the usefulness of this patch. To me, the code gets more hairy, as instead of a function argument that tells the function what to do, dependencies to global variables get created. And subsequent patches remove much of what got changed by this patch.

The condition here has not been fixed. If inbound op mode is inline and outbound op mode is sync, this piece of code fails to create the destination queue (not the default queue as I said in the older comment). But maybe the easiest fix is to remove this patch?

test/validation/api/ipsec/ipsec.c
```diff
@@ -516,6 +516,30 @@ static int send_pkts(const ipsec_test_part part[], int num_part)
...
+		return event;
+	}
+
+	/*
+	 * In case of SA lookup failure, the event is enqueued to the default
+	 * queue specified during odp_ipsec_config()
+	 */
+	do {
+		event = odp_queue_deq(suite_context.default_queue);
+	} while (event == ODP_EVENT_INVALID);
```
JannePeltonen: It would be better to use a temporary variable for the queue to be used so that the dequeuing loop would not need to be repeated. I think it would also help making the later patches of this PR just a little bit cleaner.

test/validation/api/ipsec/ipsec.c
```diff
@@ -534,7 +543,10 @@ static odp_event_t recv_pkt_async_inbound(odp_ipsec_op_status_t status)
 	 * queue specified during odp_ipsec_config()
 	 */
 	do {
-		event = odp_queue_deq(suite_context.default_queue);
+		if (odp_queue_type(suite_context.default_queue) == ODP_QUEUE_TYPE_PLAIN)
+			event = odp_queue_deq(suite_context.default_queue);
+		else
+			event = odp_schedule(NULL, ODP_SCHED_NO_WAIT);
```
JannePeltonen: This no longer checks that the event comes through the default queue, which is what the second patch added and the purpose of the second patch. Similarly, the reception through the destination queue does not check in scheduled mode that the queue indeed was the destination queue and not the default queue.

I propose using a common function for the event reception and passing the expected queue as a parameter. The function can check the queue type and use the appropriate dequeuing mechanism and in case of scheduled mode it can check that the input queue reported by odp_schedule() is the expected queue.

test/validation/api/ipsec/ipsec.c
```diff
@@ -525,29 +523,41 @@ static int send_pkts(const ipsec_test_part part[], int num_part)
 	return num_part;
 }
 
+static odp_event_t plain_queue_deq(odp_queue_t queue, uint64_t wait_ns)
+{
+	odp_time_t cur, wait, next;
+	odp_event_t event;
+
+	if (wait_ns == 0)
+		return odp_queue_deq(queue);
```
JannePeltonen, line 532: This if statement seems unnecessary to me as the timeout loop could handle zero timeouts too (if you change the > to >=).

test/validation/api/ipsec/ipsec.c
```diff
@@ -1111,14 +1119,17 @@ static odp_queue_t plain_queue_create(const char *name)
...
 }
 
-static odp_event_t sched_dest_queue_deq(void)
+static odp_event_t sched_dest_queue_deq(uint64_t wait)
 {
-	return odp_schedule(NULL, ODP_SCHED_NO_WAIT);
+	if (wait == 0)
+		return odp_schedule(NULL, ODP_SCHED_NO_WAIT);
+
+	return odp_schedule(NULL, wait);
```
JannePeltonen, line 1127: This should use odp_schedule_wait_time() to convert nanoseconds to a value suitable for the wait parameter.

test/validation/api/ipsec/ipsec.c
asasidharan: Ack. Added queue variable in v5.

test/validation/api/ipsec/ipsec.c
asasidharan: Ack. Introduced new function recv_event(). Added check in it to test the queue from which the event was received for sched mode.

test/validation/api/ipsec/ipsec.c
asasidharan: Ack.

test/validation/api/ipsec/ipsec.c
asasidharan: Ack.

asasidharan:
With v5:

- Removed separate commit removing mode arg from ipsec_init()
- Added check for sched mode whether the lookup failure packet is actually received via default queue configured.

Updates with v6:

- Updated the change to check for destination queue also for scheduled mode. This was missed in v5.
- Fixed checkpatch error.

test/validation/api/ipsec/ipsec.c
```diff
@@ -1136,22 +1233,9 @@ int ipsec_init(odp_instance_t *inst, odp_ipsec_op_mode_t mode)
...
-		default_queue = odp_queue_create("ipsec-default", NULL);
-		if (ODP_QUEUE_INVALID == default_queue) {
-			fprintf(stderr, "IPsec defaultq creation failed.\n");
-			return -1;
-		}
-	}
 
-	if (mode == ODP_IPSEC_OP_MODE_INLINE) {
+	if (suite_context.inbound_op_mode == ODP_IPSEC_OP_MODE_INLINE ||
+	    suite_context.outbound_op_mode == ODP_IPSEC_OP_MODE_INLINE) {
```
JannePeltonen, line 1253: The mode argument removal is still here even though the patch was removed. Since the op mode fields of suite_context are not initialized before ipsec_init call, this code no longer works.

What I meant with the removal of the mode-removal patch was that no only the patch is removed but also the changes in it (i.e. not squashed to some other patch).

To fix the situation, I propose resurrecting the mode argument and undoing the changes related to it.

asasidharan: Ack. Couldn't verify the tests locally before pushing due to some environment issues. Will bring back the mode arg related changes with next revision.

asasidharan:
Brought back the mode argument related changes in v7.

test/validation/api/ipsec/ipsec.c
```diff
@@ -1257,6 +1285,11 @@ int ipsec_term(odp_instance_t inst)
 			fprintf(stderr, "IPsec outq destroy failed.\n");
 	}
 
+	if (ODP_QUEUE_INVALID != default_queue) {
```
JannePeltonen, line 1395: In sync mode default_queue is not necessarily initialized to ODP_QUEUE_INVALID, depending on whether the macro maps to zeroes in the the tested ODP implementation or not.

Thus explicit initialization is needed. For the default queue it is done at the start of ipsec_init() (was added in e13682688).

test/validation/api/ipsec/ipsec.c
```diff
@@ -139,6 +139,15 @@ int ipsec_check(odp_bool_t ah,
 	     ODP_SUPPORT_NO == capa.op_mode_inline_out))
 		return ODP_TEST_INACTIVE;
 
+	if (suite_context.queue != ODP_QUEUE_INVALID) {
+		if (suite_context.q_type == ODP_QUEUE_TYPE_PLAIN &&
+		    !capa.queue_type_plain)
+			return ODP_TEST_INACTIVE;
```
JannePeltonen, line 145: This accidentally inactivates all tests for an implementation that supports only sync mode and does not support queues of any type.

test/validation/api/ipsec/ipsec.h
```diff
@@ -42,6 +45,8 @@ struct suite_context_s {
 	odp_queue_t default_queue;
 	odp_queue_t queue;
 	odp_pktio_t pktio;
+	odp_queue_type_t q_type;
+	odp_event_t (*dest_queue_deq)(void);
```
JannePeltonen: I started wondering if the dest_queue_deq function pointer and the corresponding functions are very useful, as opposed to having a single function that checks the dest_queue type, especially since such function basically already exists in the form of recv_event.

test/validation/api/ipsec/ipsec.c
```diff
@@ -365,9 +365,7 @@ void ipsec_sa_destroy(odp_ipsec_sa_t sa)
 	CU_ASSERT_EQUAL(ODP_IPSEC_OK, odp_ipsec_sa_disable(sa));
 
 	if (ODP_QUEUE_INVALID != suite_context.queue) {
-		do {
-			event = suite_context.dest_queue_deq();
-		} while (event == ODP_EVENT_INVALID);
+		event = suite_context.dest_queue_deq(ODP_TIME_MIN_IN_NS);
```
JannePeltonen, line 368: One minute timeout is quite long, especially in the CI which could suffer from it very many times after a faulty commit. I think one second would be plenty, except perhaps in some negative reassembly tests (?).

It would be good to define the timeout as a macro or function so that it could be tweaked later without having to change the code in several places.

MatiasElo:
Hi @asasidharan,

Can you give an estimate when the next version is coming? We'd like to tag a new release and this PR could still fit into it.

asasidharan:
> 
> 
> Hi @asasidharan,
> 
> Can you give an estimate when the next version is coming? We'd like to tag a new release and this PR could still fit into it.

As some of the validation test changes needs to be verified on our platform as well, I would need couple more days to send a next version. Most likely, I can send it by EOW. If it is too late, may be we can skip this one from the upcoming release.

MatiasElo:
> > Hi @asasidharan,
> > Can you give an estimate when the next version is coming? We'd like to tag a new release and this PR could still fit into it.
> 
> As some of the validation test changes needs to be verified on our platform as well, I would need couple more days to send a next version. Most likely, I can send it by EOW. If it is too late, may be we can skip this one from the upcoming release.

Thanks for the info. We are not in a terrible hurry to tag the next version, but I'd like to do it this week. Let's see how it goes.

asasidharan:
@MatiasElo I am not yet ready with the next version as I ran into some issue while verifying the validation changes on our platform. I think I will need couple of more days on this. Apologies for the delay in updating you.

test/validation/api/ipsec/ipsec.c
JannePeltonen: The commit message still talks about 1 minute timeout.

test/performance/odp_ipsec.c
```diff
@@ -809,6 +809,16 @@ run_measure_one_config(ipsec_args_t *cargs,
 		return -1;
 	}
 
+	if (cargs->schedule && !capa.queue_type_sched) {
+		app_err("Scheduled type destination queue not supported.\n");
+		return -1;
+	}
+
+	if (cargs->poll && !capa.queue_type_plain) {
```
JannePeltonen, line 817: These checks are a bit too late since they are done after odp_ipsec_config() which sets the default queue.

test/validation/api/ipsec/ipsec.h
```diff
@@ -30,6 +30,9 @@ int ipsec_in_inline_init(void);
 int ipsec_out_inline_init(void);
 
 int ipsec_suite_init(void);
+int ipsec_suite_plain_init(void);
+int ipsec_suite_sched_init(void);
```
JannePeltonen, line 34: Now with the new suite init functions, maybe ipsec_suite_init() could be renamed something more specific (ipsec_suite_sync_init?).


The suite init functions do not seem to reinitialize suite_context to default values. It seems to me that e.g. suite_context.queue could be left to a stale handle if sync test suite is run after an async or inline suite. Currently it does not happen because those suites are run in different processes, but for robustness I think it would be better
not to rely on that but have suite_init initialize everything in suite context.

test/validation/api/ipsec/ipsec.c
```diff
@@ -29,16 +29,28 @@ static odp_pktio_t pktio_create(odp_pool_t pool)
...
+		return ODP_PKTIO_INVALID;
+	}
+
 	odp_pktio_param_init(&pktio_param);
 	pktio_param.in_mode = ODP_PKTIN_MODE_QUEUE;
 
+	if (suite_context.inbound_op_mode == ODP_IPSEC_OP_MODE_INLINE ||
+	    suite_context.inbound_op_mode == ODP_IPSEC_OP_MODE_INLINE) {
+		if (!ipsec_capa.queue_type_plain && ipsec_capa.queue_type_sched)
+			pktio_param.in_mode = ODP_PKTIN_MODE_SCHED;
```
JannePeltonen, line 52: This cannot be right as the two or'ed expressions are the same.

Actually, I do not see why the pktio queue type should be affected by the IPsec capability that concerns IPsec completion queues.

test/validation/api/ipsec/ipsec.c
asasidharan: Missed to update the commit message. Will fix this in v9.

test/performance/odp_ipsec.c
asasidharan: Ack. Will move the checks to main() itself.

test/validation/api/ipsec/ipsec.c
asasidharan: This was a typo. I meant to set in_mode to sched if we are running inline mode test. Thank you spotting this.

On our platform, with inline IPsec, we can submit only via scheduled queues and that would be the pktin queues configured for the packet_io device.

asasidharan:
With v8:

- Fixed the commit message for "validation: ipsec: set 1sec time out for event receive"
- Moved the queue capability check for IPsec performance app to the main()
- Introduced d ipsec_suite_sync_init() for suite initialization in sync mode
- Fixed the condition setting pktin queue mode in pktio_create()

test/validation/api/ipsec/ipsec.h
JannePeltonen: ok, I see. It is a bit messy since there are both global (with the lifetime of the process) and suite specific stuff in suite_context. But it is not the fault of this patch, so this is fine.

test/validation/api/ipsec/ipsec.c
JannePeltonen: > On our platform, with inline IPsec, we can submit only via scheduled queues and that would be the pktin queues configured for the packet_io device.

API validation tests are written against the API spec, not any particular implementation.

Setting pktio queue type in this patch is wrong. This patch is about checking _ipsec_ destination queue capability, which does not have anything (in the API spec) to do with pktio queue type.

As discussed, your implementation is currently not API compliant with respect to the destination queue in inline mode. For this we need an API spec change that allows inline processed packets to be delivered through pktio queues. And for that API change we need to change the validation test accordingly. But at this point, this pktio queue type change looks totally unexpected to anyone who does not know the backround and what future API changes might be coming.

Btw, are you saying that in your implementation there it not only the limitation that inline processed packets can only be delivered through pktio queueus, but also the limitation that said pktio queues cannot be plain queues? But you do support plain pktio queues without IPsec do you?

JannePeltonen:
Other patches than "validation: ipsec: check dest queue capability" look ok and can be considered reviewed by me.

test/validation/api/ipsec/ipsec.h
asasidharan: Ack.

test/validation/api/ipsec/ipsec.c
asasidharan: Removed the pktin queue configuration introduced inside pktio_create() in v9. Updated the pktio_create() function to check whether inline mode is supported using the pktio capability.

asasidharan:
With v10:

- Removed the pktin queue config introduced in v9.
- Added Reviewed-by tag to all commits except "validation: ipsec: check dest queue capability".

asasidharan:
With v11:

- Removed redundant IPsec inline capability check introduced in pktio_create().
- Reverted back to the original logic in verifying IPsec inline functionality.

test/validation/api/ipsec/ipsec.c
```diff
@@ -118,7 +146,7 @@ static void pktio_stop(odp_pktio_t pktio)
 		fprintf(stderr, "IPsec pktio stop failed.\n");
 
 	while (1) {
-		odp_event_t ev = recv_event(queue);
+		odp_event_t ev = recv_event(queue, EVENT_WAIT_TIME);
```
JannePeltonen, line 149: I noticed that this now always takes a second, adding two seconds in the execution time of the inline in and out tests programs. recv_event(queue, 0) would not have changed the old behaviour. I wonder if the change is intentional? Maybe in some environments a non-zero wait time ensures that possible in-flight packets have time to end up in the queue, making non-zero wait time here reasonable?

JannePeltonen:
The validation patch looks ok to me and can be considered reviewed (please add the tag). So the whole patch set is now reviewed, except that in the last patch I noticed one minor thing that makes the tests take a few seconds longer. The code is not wrong, but I wonder if it would be good to change it to make the tests a little bit faster.

test/validation/api/ipsec/ipsec.c
asasidharan: The non-zero wait time was not intentional here. Thanks for spotting this, will make it to zero wait time and update the PR.

JannePeltonen (approved):

asasidharan:
Thank you @JannePeltonen for you time in reviewing the PR.

With v12:
- Corrected the recv_event() wait time to zero.
- Added Reviewed-by tag to "validation: ipsec: check dest queue capability" commit.

PR #1313, opened 2021-06-17
[PATCH v3] validation: ipsec: populate packet with layer info

asasidharan:
Populate the test packets with layer info

Change-Id: Ib1bfd36ddf03bb34a8661f1e02806e4dbfb19c69

### DIFF
--- a/test/validation/api/ipsec/ipsec.c
+++ b/test/validation/api/ipsec/ipsec.c
@@ -24,6 +24,8 @@ static odp_ipsec_capability_t capa;
 #define PACKET_USER_PTR	((void *)0x1212fefe)
 #define IPSEC_SA_CTX	((void *)0xfefefafa)
 
+static void parse_ip(odp_packet_t pkt);
+
 static odp_pktio_t pktio_create(odp_pool_t pool)
 {
 	odp_pktio_t pktio;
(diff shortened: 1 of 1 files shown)


JannePeltonen:
What is the purpose of this change? Please provide some context and rationale in the commit message.

asasidharan:
> 
> What is the purpose of this change?

Inline outbound processing would require parsed IP packet to help with the result packet length calculations. Otherwise per-packet re-parsing needs to be done by implementation.

JannePeltonen:
> Inline outbound processing would require parsed IP packet to help with the result packet length calculations.

The current ODP API does not require parsed IP packet for outbound IPsec operations. Quote from the API spec:

```
 * is ignored):
 * - L3 offset: Offset to the first byte of the (outmost) IP header
 * - L4 offset: Offset to the L4 header if L4 checksum offload is requested
 *
 * Additionally, input IP packet length (odp_packet_len() minus
 * odp_packet_l3_offset()) must match values in protocol headers.
```

API validation tests are for testing that an ODP implementation complies with the API spec. Since the current test code seems to use the API correctly, a test failure must be fixed in the ODP implementation being tested or through an ODP API specification change, not by simply changing the validation test to be easier for the implementation.

IOW, even if the current API might cause additional packet parsing in your implementation, you cannot ignore the API spec and change the validation test to make your implementation look compliant when it is not. Applications have to be able to rely on the behaviour specified in the API.

JannePeltonen:

PR #1317, opened 2021-06-18
[PATCH v4] linux-gen: pktio: use names other than index also as dpdk ports

vijayinavolu:
dpdk_open() only allows netdev string to be a port number
This is to allow passing all other supported DPDK ports as well.

### DIFF
--- a/platform/linux-generic/pktio/dpdk.c
+++ b/platform/linux-generic/pktio/dpdk.c
@@ -1676,6 +1676,7 @@ static int dpdk_open(odp_pktio_t id ODP_UNUSED,
 	uint32_t mtu;
 	int i;
 	pool_t *pool_entry;
+	uint16_t port_id;
 
 	if (disable_pktio)
 		return -1;
@@ -1684,8 +1685,15 @@ static int dpdk_open(odp_pktio_t id ODP_UNUSED,
 		return -1;
 	pool_entry = pool_entry_from_hdl(pool);
 
-	if (!dpdk_netdev_is_valid(netdev)) {
-		ODP_ERR("Invalid dpdk netdev: %s\n", netdev);
+	/* Init pktio entry */
+	memset(pkt_dpdk, 0, sizeof(*pkt_dpdk));
+
+	if (!rte_eth_dev_get_port_by_name(netdev, &port_id))
+		pkt_dpdk->port_id = port_id;
+	else if (dpdk_netdev_is_valid(netdev))
+		pkt_dpdk->port_id = atoi(netdev);
+	else {
+		ODP_ERR("Invalid DPDK interface name: %s\n", netdev);
 		return -1;
 	}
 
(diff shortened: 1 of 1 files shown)


vijayinavolu:
Related odp-dpdk PR
https://github.com/OpenDataPlane/odp-dpdk/pull/135

platform/linux-generic/pktio/dpdk.c
```diff
@@ -1684,8 +1685,12 @@ static int dpdk_open(odp_pktio_t id ODP_UNUSED,
 		return -1;
 	pool_entry = pool_entry_from_hdl(pool);
 
-	if (!dpdk_netdev_is_valid(netdev)) {
-		ODP_ERR("Invalid dpdk netdev: %s\n", netdev);
+	if (!rte_eth_dev_get_port_by_name(netdev, &port_id))
+		pkt_dpdk->port_id = port_id;
+	else if (dpdk_netdev_is_valid(netdev))
+		pkt_dpdk->port_id = atoi(netdev);
```
MatiasElo, line 1691: There is a `memset()` below which initializes `pkt_dpdk `.

platform/linux-generic/pktio/dpdk.c
```diff
@@ -1684,8 +1685,12 @@ static int dpdk_open(odp_pktio_t id ODP_UNUSED,
...
+	else {
+		ODP_ERR("Invalid interface name!: %s\n", netdev);
```
MatiasElo, line 1693: In linux-generic implementation a user may want to use also other interface types than DPDK, so maybe it  would be better to print:
`ODP_DBG("Invalid DPDK interface name: %s\n", netdev);`

platform/linux-generic/pktio/dpdk.c
vijayinavolu: moved the memset up

platform/linux-generic/pktio/dpdk.c
vijayinavolu: Updated the print statement

platform/linux-generic/pktio/dpdk.c
```diff
@@ -1684,8 +1685,15 @@ static int dpdk_open(odp_pktio_t id ODP_UNUSED,
...
-		ODP_ERR("Invalid dpdk netdev: %s\n", netdev);
+	/* Init pktio entry */
+	memset(pkt_dpdk, 0, sizeof(*pkt_dpdk));
+
+	if (!rte_eth_dev_get_port_by_name(netdev, &port_id))
+		pkt_dpdk->port_id = port_id;
+	else if (dpdk_netdev_is_valid(netdev))
+		pkt_dpdk->port_id = atoi(netdev);
+	else {
+		ODP_ERR("Invalid DPDK interface name!: %s\n", netdev);
```
MatiasElo, line 1696: You can still remove the exclamation mark but otherwise the PR is looking good.

MatiasElo (approved):

platform/linux-generic/pktio/dpdk.c
vijayinavolu: Changed the log message and added reviewed-by tag

PR #1318, opened 2021-06-21
[PATCH v2] api: ipsec, pktio: specify more default values for init functions

JannePeltonen:
This PR specifies more default values for init functions, improves validation testing of the defaults and fixes one bug in default value setting.

Contained patches (reverse order):

validation: pktio: check default values of odp_pktio_config_t
api: pktio: make loopback mode disabled by default
validation: ipsec: check reassembly related default values
api: ipsec: specify default values for reassembly enables
linux-gen: reass: fix the default value of max_num_frags

### DIFF
--- a/test/validation/api/ipsec/ipsec_test_out.c
+++ b/test/validation/api/ipsec/ipsec_test_out.c
@@ -1570,6 +1570,12 @@ static void ipsec_test_default_values(void)
 	CU_ASSERT(config.inbound.retain_outer == ODP_PROTO_LAYER_NONE);
 	CU_ASSERT(config.inbound.parse_level == ODP_PROTO_LAYER_NONE);
 	CU_ASSERT(config.inbound.chksums.all_chksum == 0);
+	CU_ASSERT(!config.inbound.reassembly.en_ipv4);
+	CU_ASSERT(!config.inbound.reassembly.en_ipv6);
+	CU_ASSERT(config.inbound.reassembly.max_wait_time == 0);
+	CU_ASSERT(config.inbound.reassembly.max_num_frags == 2);
+	CU_ASSERT(!config.inbound.reass_async);
+	CU_ASSERT(!config.inbound.reass_inline);
 	CU_ASSERT(config.outbound.all_chksum == 0);
 	CU_ASSERT(!config.stats_en);
 
@@ -1592,6 +1598,7 @@ static void ipsec_test_default_values(void)
 	CU_ASSERT(sa_param.inbound.lookup_mode == ODP_IPSEC_LOOKUP_DISABLED);
 	CU_ASSERT(sa_param.inbound.antireplay_ws == 0);
 	CU_ASSERT(sa_param.inbound.pipeline == ODP_IPSEC_PIPELINE_NONE);
+	CU_ASSERT(!sa_param.inbound.reassembly_en);
 	CU_ASSERT(sa_param.outbound.tunnel.type == ODP_IPSEC_TUNNEL_IPV4);
 	CU_ASSERT(sa_param.outbound.tunnel.ipv4.dscp == 0);
 	CU_ASSERT(sa_param.outbound.tunnel.ipv4.df == 0);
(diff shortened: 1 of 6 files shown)


psavol (approved):

test/validation/api/ipsec/ipsec_test_out.c
```diff
@@ -1570,6 +1570,12 @@ static void ipsec_test_default_values(void)
 	CU_ASSERT(config.inbound.retain_outer == ODP_PROTO_LAYER_NONE);
 	CU_ASSERT(config.inbound.parse_level == ODP_PROTO_LAYER_NONE);
 	CU_ASSERT(config.inbound.chksums.all_chksum == 0);
+	CU_ASSERT(!config.inbound.reassembly.en_ipv4);
+	CU_ASSERT(!config.inbound.reassembly.en_ipv6);
+	CU_ASSERT(config.inbound.reassembly.max_wait_time == 0);
+	CU_ASSERT(config.inbound.reassembly.max_num_frags == 2);
```
anoobj, line 1576: @JannePeltonen On Marvell hardware, the field max_num_frags has to be 4. It cannot support 2. I remember we had discussed this earlier during the spec proposal and I think we had ignored it since there was no test case actually exercising the same.

Now that you are adding test case, we may need to reconsider the options we have,
1. Amend the spec so that max_num_frags would be implementation defined
2. Marvell can expose capability as 4, default config as 2, but if an application chooses to use default config, we will have to return error (as 2 is unsupported by hardware)

JannePeltonen: I am not sure if there is a big problem, even with that HW restriction.

The default value set in the init functions can obviously be anything.

IMHO, odp_reass_config_t::max_num_frags can be allowed to be set to 2 even with the HW restriction. If the HW is always able to handle at least 4 fragments, then it can handle 2 and 3 as well. I my opinion the point of the config option is to tell the implementation how many fragments, at most, the application expects to receive for each fragmented. I do not think the intent should be to tell the implementation that it should not attempt to reassemble more fragments into one packet. Maybe we could amend the description of the config setting so that it explicitly allows reassembling more than the minimum required number of fragments.

odp_packet_reass_partial_state() promises not to return more than odp_reass_config_t::max_num_frags through the array. I suppose the promise could be kept by generating not one but two incomplete reassembly events, with two fragments in each, when reassembly with 4 fragments accumulated in the HW fails.

If the above is not feasible, then maybe we could try to change odp_packet_reass_partial_state() so that it could somehow return more than odp_reass_config_t::max_num_frags.

anoobj: @JannePeltonen I think the application semantics will be simpler and easier with the current spec (ie, odp_packet_reass_partial_state() does not return more than odp_reass_config_t::max_num_frags). 

> odp_packet_reass_partial_state() promises not to return more than odp_reass_config_t::max_num_frags through the array.

If we take this approach then odp_reass_config::max_num_frags would lose it's relevance. If this value doesn't mean anything, we can as well remove it.

Since the current reassembly UT overrides these config values, I don't see any problem now. If you think this behavior is fine, then it's all good.

This patch looks good otherwise.

JannePeltonen: > >     odp_packet_reass_partial_state() promises not to return more than odp_reass_config_t::max_num_frags through the array.

> If we take this approach then odp_reass_config::max_num_frags would lose it's relevance

In this approach (i.e. the current API) odp_reass_config_t::max_num_frags would indicate the max number of fragments that can be returned by the odp_packet_reass_partial_state(), i.e. it would be the minimum array size the application must use.

But maybe you were referring to the approach where more than max_num_frags could be returned after a failed reassembly. In that case max_num_frags config setting would still be useful as an application could tell how many fragments per packets it expects and then the implementation could allocated reassembly resources appropriately and not for the absolute maximum number of fragments the implementation supports.

anoobj: @JannePeltonen Yes. You are right. My concern was about the alternate approach (more than max_num_frags could be returned after a failed reassembly).

Thanks for the clarification.

PR #1319, opened 2021-06-22
[PATCH v3] api: tm: increase sched weight parameter size

nithind1988:
Increase sched weight parameter size to accommodate larger
sched weights. This is needed for platform that support
weights larger than 255.

### DIFF
--- a/platform/linux-generic/odp_traffic_mngr.c
+++ b/platform/linux-generic/odp_traffic_mngr.c
@@ -2604,7 +2604,8 @@ static void tm_system_capabilities_set(odp_tm_capabilities_t *cap_ptr,
 	odp_bool_t                   dual_slope;
 	uint32_t                     num_levels, level_idx, max_nodes;
 	uint32_t                     max_queues, max_fanin;
-	uint8_t                      max_priority, min_weight, max_weight;
+	uint32_t                     min_weight, max_weight;
+	uint8_t                      max_priority;
 
 	num_levels = MAX(MIN(req_ptr->num_levels, ODP_TM_MAX_LEVELS), 1);
 	memset(cap_ptr, 0, sizeof(odp_tm_capabilities_t));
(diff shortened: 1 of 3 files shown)


platform/linux-generic/odp_traffic_mngr.c
```diff
@@ -2646,9 +2646,9 @@ static void tm_system_capabilities_set(odp_tm_capabilities_t *cap_ptr,
 		max_priority     = MIN(per_level_req->max_priority,
 				       ODP_TM_MAX_PRIORITIES - 1);
 		min_weight       = MAX(per_level_req->min_weight,
-				       ODP_TM_MIN_SCHED_WEIGHT);
+				       (uint32_t)ODP_TM_MIN_SCHED_WEIGHT);
 		max_weight       = MIN(per_level_req->max_weight,
-				       ODP_TM_MAX_SCHED_WEIGHT);
+				       (uint32_t)ODP_TM_MAX_SCHED_WEIGHT);
```
JannePeltonen, line 2651: These casts are a bit ugly. I wonder if ODP_TM_MAX_SCHED_WEIGHT should be defines as unsigned in the ABI instead, i.e.: #define ODP_TM_MAX_SCHED_WEIGHT  255U

It seems one issue here is that ODP API specifies various constants that the ABI headers should define but it does not not specify their type. It seems that most or all of such literals are signed currently.

Note also that there is a convention that linux-gen changes are not done in the same patch as API changes so that API patches stay free from code that is specific to any particular implementation.

JannePeltonen:
The API change looks good to me but I am not sure what is the right way of getting rid of the signed/unsigned comparison warning.

You might want to consider changing the type of min_weight and max_weight in tm_system_capabilities_set() so that the function does not break if someone increases ODP_TM_MAX_SCHED_WEIGHT and does not notice the uint8_t there.

platform/linux-generic/odp_traffic_mngr.c
nithind1988: Ack, I'll change to 255U. After changing that, we atleast for now won't get signed/unsigned comparison warning.\

I'm also not sure how we can enforce types for constants

JannePeltonen (approved):

PR #1323, opened 2021-06-24
[PATCH v2] ipsec: add support for AES-CMAC-96 auth algorithm

JannePeltonen:
ODP IPsec API supports AES-CMAC-96 but the algorithm is missing from validation test and the capability checker in helper. This PR adds minimal IPsec validation test (the digest algorithm itself is already validated in crypto validation tests) and makes the helper aware of the algorithm.

This PR also adds AES-CMAC-96 support in linux-gen IPsec implementation. The change is a one-liner as the crypto subsystem supports the algorithm and the IPsec implementation just needs to enable its use.

### DIFF
--- a/helper/ipsec.c
+++ b/helper/ipsec.c
@@ -98,6 +98,10 @@ int odph_ipsec_alg_check(odp_ipsec_capability_t capa,
 		if (!capa.auths.bit.aes_ccm)
 			return -1;
 		break;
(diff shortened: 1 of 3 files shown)


JereLeppanen (approved):

JannePeltonen:
v2: added reviewed-by tag

PR #1325, opened 2021-06-29
[PATCH v11] api: tm: add node stats capability and query API 

nithind1988:
api: tm: add node stats capability and query API

Add capability to advertise capability of nodes at a given level
to provide statistics.

Also add a new API to query for packet sent count in green, yellow
and red color and packet drop count stats, along with corresponding
octet count statistics. TM node stats query will return only platform
supported statistics.

linux-gen: tm: add node stats API stub

Add node stats API stub.

### DIFF
--- a/include/odp/api/spec/traffic_mngr.h
+++ b/include/odp/api/spec/traffic_mngr.h
@@ -1,7 +1,7 @@
 /* SPDX-License-Identifier: BSD-3-Clause
  * Copyright (c) 2015-2018 Linaro Limited
  * Copyright (c) 2021-2022 Nokia
- * Copyright (c) 2022 Marvell
+ * Copyright (c) 2022-2025 Marvell
  */
 
 #ifndef ODP_API_SPEC_TRAFFIC_MNGR_H_
@@ -234,6 +234,51 @@ typedef struct odp_tm_queue_stats_capability_t {
 	};
 } odp_tm_queue_stats_capability_t;
 
+/** TM node specific statistics counters */
+typedef struct odp_tm_node_stats_t {
+	/** Number of discarded packets. */
+	uint64_t discards;
+
+	/** Number of octets in discarded packets. */
+	uint64_t discard_octets;
+
+	/** Number of successfully processed packets. */
+	uint64_t packets;
+
+	/** Number of octets in successfully processed packets. */
+	uint64_t octets;
+} odp_tm_node_stats_t;
+
+/**
+ * TM node statistics capabilities
+ */
+typedef struct odp_tm_node_stats_capability_t {
+	/** Supported counters */
+	union {
+		/** Statistics counters in a bit field struct */
+		struct {
+			/** See odp_tm_node_stats_t::discards */
+			uint64_t discards       : 1;
+
+			/** See odp_tm_node_stats_t::discard_octets */
+			uint64_t discard_octets : 1;
+
+			/** See odp_tm_node_stats_t::packets */
+			uint64_t packets        : 1;
+
+			/** See odp_tm_node_stats_t::octets */
+			uint64_t octets         : 1;
+
+		} counter;
+
+		/** All bits of the bit field structure
+		 *
+		 *  This field can be used to set/clear all flags, or
+		 *  for bitwise operations over the entire structure. */
+		uint64_t all_counters;
+	};
+} odp_tm_node_stats_capability_t;
+
 /** Per Level Capabilities
  *
  * The odp_tm_level_capabilities_t record is used to describe the capabilities
@@ -357,6 +402,8 @@ typedef struct {
 
 	} tm_node_threshold;
 
+	/** Node statistics counter capabilities */
+	odp_tm_node_stats_capability_t node_stats;
 } odp_tm_level_capabilities_t;
 
 /** The tm_pkt_prio_mode_t enumeration type is used to indicate different
@@ -2143,6 +2190,23 @@ typedef struct {
  */
 int odp_tm_node_fanin_info(odp_tm_node_t             tm_node,
 			   odp_tm_node_fanin_info_t *info);
+
+/** Get statistics for a TM node
+ *
+ * Read statistics counters associated with the given TM node. Counters not
+ * supported by the node are set to zero by the implementation.
+ *
+ * Depending on the implementation, certain statistics counters may be updated
+ * only on specific nodes in the TM node tree (e.g. by the final node in the
+ * tree even if the packet was marked to be dropped by an earlier node).
+ *
+ * @param      tm_node TM node handle
+ * @param[out] stats   Statistics structure for output
+ *
+ * @retval  0 on success
+ * @retval <0 on failure
+ */
+int odp_tm_node_stats(odp_tm_node_t tm_node, odp_tm_node_stats_t *stats);
 
 /** The odp_tm_queue_info_t record type  is used to return various bits of
  * information about a given tm_queue via the odp_tm_queue_info() function.
--- a/test/validation/api/traffic_mngr/traffic_mngr.c
+++ b/test/validation/api/traffic_mngr/traffic_mngr.c
@@ -1,6 +1,6 @@
 /* SPDX-License-Identifier: BSD-3-Clause
  * Copyright (c) 2015-2018 Linaro Limited
- * Copyright (c) 2022 Marvell
+ * Copyright (c) 2022-2025 Marvell
  * Copyright (c) 2022 Nokia
  */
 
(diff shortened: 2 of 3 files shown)


include/odp/api/spec/traffic_mngr.h
```diff
@@ -189,6 +189,47 @@ extern "C" {
  * tree/hierarchy of nodes.
  */
 
+/**
+ * TM node stats capability
```
MatiasElo, line 193: To make the statistics capabilities documentation uniform:
`TM node statistics capabilities`

include/odp/api/spec/traffic_mngr.h
```diff
@@ -189,6 +189,47 @@ extern "C" {
...
+ */
+typedef struct odp_tm_node_stats_capability_t {
+	/** Supported counters */
+	union {
+		/** Statistics counters supported in bit field struct */
+		struct {
+			/** @see odp_tm_node_stats_t::dropped_pkts */
+			uint64_t dropped_pkts;
```
MatiasElo, line 201: We use `discard` in all other packet statistics. Also, I think we can spare a few letters and just use e.g. `discarded_packets`.

include/odp/api/spec/traffic_mngr.h
```diff
@@ -189,6 +189,47 @@ extern "C" {
...
+
+			/** @see odp_tm_node_stats_t::dropped_octets */
+			uint64_t dropped_octets;
+
+			/** @see odp_tm_node_stats_t::green_pkts */
+			uint64_t green_pkts;
```
MatiasElo, line 207: Can be `green_packets`. Same for the following fields.

include/odp/api/spec/traffic_mngr.h
```diff
@@ -248,6 +289,9 @@ typedef struct {
 	 * When true the min_weight and max_weight fields above specify
 	 * the legal range of such weights. */
 	odp_bool_t weights_supported;
+
+	/** Node statistics counter capabilities */
+	odp_tm_node_stats_capability_t stats;
```
MatiasElo, line 294: Could we potentially need other statistics capabilities at this level later on? In that case it could perhaps be better to name this `node_stats`.

include/odp/api/spec/traffic_mngr.h
```diff
@@ -1841,6 +1885,48 @@ typedef struct {
...
+	uint64_t yellow_pkts;
+
+	/** Yellow octets that are sent through this tm node */
+	uint64_t yellow_octets;
+
+	/** Red packets that are sent through this tm node */
+	uint64_t red_pkts;
+
+	/** Red octets that are sent through this tm node */
+	uint64_t red_octets;
```
MatiasElo, line 1914: Previous comments about renaming the struct members in `odp_tm_node_stats_capability_t `.

include/odp/api/spec/traffic_mngr.h
```diff
@@ -1841,6 +1885,48 @@ typedef struct {
 int odp_tm_node_fanin_info(odp_tm_node_t             tm_node,
 			   odp_tm_node_fanin_info_t *info);
 
+/** The odp_tm_node_stats_t record type  is used to return various stats
+ * supported by a given tm_node via the odp_tm_node_stats() function.
+ */
```
MatiasElo, line 1890: Could be just: `TM node specific statistics counters`

include/odp/api/spec/traffic_mngr.h
```diff
@@ -1841,6 +1885,48 @@ typedef struct {
...
+
+	/** Red packets that are sent through this tm node */
+	uint64_t red_pkts;
+
+	/** Red octets that are sent through this tm node */
+	uint64_t red_octets;
+} odp_tm_node_stats_t;
+
+/** Get tm_node stats
+ *
```
MatiasElo, line 1918:
```
/**
 * Get statistics for a TM node
```

include/odp/api/spec/traffic_mngr.h
```diff
@@ -1841,6 +1885,48 @@ typedef struct {
...
+ * The odp_tm_node_stats() function is used to extract runtime statistics
+ * associated with a given tm_node. The stats structure is written
+ * only on success.
+ *
```
MatiasElo, line 1922: Should also include: `Counters not supported by the node are set to zero.`

include/odp/api/spec/traffic_mngr.h
```diff
@@ -1841,6 +1885,48 @@ typedef struct {
...
+ * @param      tm_node    Specifies the tm_node to be queried.
+ * @param[out] node_stats A pointer to an odp_tm_node_stats_t record that is to
+ *                        be filled in by this call.
+ * @return                Returns < 0 upon failure, 0 upon success.
```
MatiasElo, line 1926:
```
 * @param      tm_node  TM node handle
 * @param[out] stats    Output buffer for counters
 *
 * @retval  0 on success
 * @retval <0 on failure
```

MatiasElo (commented):
Some minor comments to sync this PR with #1305.

include/odp/api/spec/traffic_mngr.h
JannePeltonen, line 201: It seems these capability fields were supposed to be bit fields, but now they are not defined as such.

nithind1988: I missed it mistakenly. Will fix it in v2.

JannePeltonen: I wonder if "discarded" is really better than "dropped" in the TM API. The term drop is often used in traffic management context and also in the ODP TM API spec, which discusses drop probabilities when (W)RED is used, RED versus tail drop algorithm and so forth.

include/odp/api/spec/traffic_mngr.h
```diff
@@ -1841,6 +1885,48 @@ typedef struct {
...
+ */
+typedef struct odp_tm_node_stats_t {
+	/** Packets dropped by this node post scheduling/shaping at this node */
+	uint64_t dropped_pkts;
+
+	/** Octets dropped after scheduling/shaping at this node */
+	uint64_t dropped_octets;
+
+	/** Green packets that are sent through this tm node */
+	uint64_t green_pkts;
```
JannePeltonen, line 1899: I wonder if we need to or want to add these color specific counters now instead of single packet and byte counter for non-dropped packets of any color.

I am asking this because packet color related aspects of the TM API seem a bit incomplete to me and the API does not seem to allow the color (odp_packet_color(pkt)) of a packet to change when inside the TM system.

include/odp/api/spec/traffic_mngr.h
MatiasElo: We ended up naming these counters `discard_octets`and `discard_packets` in #1305.

include/odp/api/spec/traffic_mngr.h
```diff
@@ -2052,6 +2124,20 @@ typedef struct {
...
 			   odp_tm_node_fanin_info_t *info);
 
+/** Get statistics for a TM node
+ *
+ * The odp_tm_node_stats() function is used to extract runtime statistics
+ * associated with a given tm_node. The stats structure is written
+ * only on success. Counters not supported by the node are set to zero.
+ *
+ * @param      tm_node TM node handle
+ * @param[out] stats   Output buffer handle for counters
```
MatiasElo, line 2134: ` * @param[out] stats    Statistics structure for output`

include/odp/api/spec/traffic_mngr.h
```diff
@@ -247,6 +247,75 @@ typedef struct odp_tm_queue_stats_capability_t {
...
+
+	/** Red packets that are sent through this tm node */
+	uint64_t red_packets;
+
+	/** Red octets that are sent through this tm node */
+	uint64_t red_octets;
+} odp_tm_node_stats_t;
+
+/**
+ * TM node statistics capability
```
MatiasElo, line 278: `capabilities`

include/odp/api/spec/traffic_mngr.h
```diff
@@ -247,6 +247,75 @@ typedef struct odp_tm_queue_stats_capability_t {
...
+ */
+typedef struct odp_tm_node_stats_capability_t {
+	/** Supported counters */
+	union {
+		/** Statistics counters supported in bit field struct */
```
MatiasElo, line 283: `supported` can be dropped.

include/odp/api/spec/traffic_mngr.h
```diff
@@ -247,6 +247,75 @@ typedef struct odp_tm_queue_stats_capability_t {
...
 
+/** TM node specific statistics counters */
+typedef struct odp_tm_node_stats_t {
+	/** Packets dropped by this node post scheduling/shaping at this node */
+	uint64_t discards;
+
+	/** Octets dropped after scheduling/shaping at this node */
+	uint64_t discard_octets;
+
+	/** Green packets that are sent through this tm node */
```
MatiasElo, line 258: E.g. `Number of successfully processed green packets.` Same goes for all the colors.

include/odp/api/spec/traffic_mngr.h
```diff
@@ -247,6 +247,75 @@ typedef struct odp_tm_queue_stats_capability_t {
...
+	uint64_t green_packets;
+
+	/** Green octets that are sent through this tm node */
```
MatiasElo, line 261: E.g. `Number of octets in successfully processed green packets.`

include/odp/api/spec/traffic_mngr.h
```diff
@@ -247,6 +247,75 @@ typedef struct odp_tm_queue_stats_capability_t {
 	};
 } odp_tm_queue_stats_capability_t;
 
+/** TM node specific statistics counters */
+typedef struct odp_tm_node_stats_t {
+	/** Packets dropped by this node post scheduling/shaping at this node */
+	uint64_t discards;
+
+	/** Octets dropped after scheduling/shaping at this node */
```
MatiasElo, line 255: And `Number of octets in packets discarded after scheduling/shaping.`

include/odp/api/spec/traffic_mngr.h
```diff
@@ -247,6 +247,75 @@ typedef struct odp_tm_queue_stats_capability_t {
 	};
 } odp_tm_queue_stats_capability_t;
 
+/** TM node specific statistics counters */
+typedef struct odp_tm_node_stats_t {
+	/** Packets dropped by this node post scheduling/shaping at this node */
```
MatiasElo, line 252: E.g. `Number of packets discarded after scheduling/shaping.`

include/odp/api/spec/traffic_mngr.h
```diff
@@ -247,6 +247,75 @@ typedef struct odp_tm_queue_stats_capability_t {
 	};
 } odp_tm_queue_stats_capability_t;
 
+/** TM node specific statistics counters */
```
MatiasElo, line 250: I'd like to avoid having to repeat `sent through this tm node` for each struct member. If required, an additional explainer sentence could be added here. See comments below.

MatiasElo, line 250: Marvell copyright can be added to the file.

include/odp/api/spec/traffic_mngr.h
nithind1988: Discussed with marvell transport team. Will drop the color specific stats from this PR and just add packets/octets counters.

include/odp/api/spec/traffic_mngr.h
```diff
@@ -2052,6 +2124,20 @@ typedef struct {
 int odp_tm_node_fanin_info(odp_tm_node_t             tm_node,
 			   odp_tm_node_fanin_info_t *info);
 
+/** Get statistics for a TM node
+ *
+ * The odp_tm_node_stats() function is used to extract runtime statistics
+ * associated with a given tm_node. The stats structure is written
+ * only on success. Counters not supported by the node are set to zero.
```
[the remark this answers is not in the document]
nithind1988: Ack, removed it since all the other non-tm stats doesn't have any such clause.

include/odp/api/spec/traffic_mngr.h
```diff
@@ -247,6 +247,51 @@ typedef struct odp_tm_queue_stats_capability_t {
...
+ */
+typedef struct odp_tm_node_stats_capability_t {
+	/** Supported counters */
+	union {
+		/** Statistics counters in a bit field struct */
+		struct {
+			/** @see odp_tm_node_stats_t::discards */
+			uint64_t discards       : 1;
+
+			/** @see odp_tm_node_stats_t::discard_octets*/
```
MatiasElo, line 276: Space missing at the end.

include/odp/api/spec/traffic_mngr.h
```diff
@@ -247,6 +247,51 @@ typedef struct odp_tm_queue_stats_capability_t {
...
 
+/** TM node specific statistics counters */
+typedef struct odp_tm_node_stats_t {
+	/** Number of packets discarded after scheduling/shaping. */
+	uint64_t discards;
+
+	/** Number of octets in packets discarded after scheduling/shaping. */
+	uint64_t discard_octets;
+
+	/** Number of successfully processed packets */
```
MatiasElo, line 258: Missing period.

include/odp/api/spec/traffic_mngr.h
```diff
@@ -247,6 +247,51 @@ typedef struct odp_tm_queue_stats_capability_t {
...
+	uint64_t packets;
+
+	/** Number of Octets in successfully processed packets */
```
MatiasElo, line 261: `octets` and missing period.

include/odp/api/spec/traffic_mngr.h
MatiasElo: The commit message should be updated to match the current API.

MatiasElo (commented):
Except for some minor style tweaks and missing copyright the API changes are looking OK to me.

test/validation/api/traffic_mngr/traffic_mngr.c
```diff
@@ -4309,6 +4309,116 @@ static void traffic_mngr_test_queue_stats(void)
...
+	odp_tm_queue_t tm_queue;
+	odp_tm_capabilities_t capa;
+	pkt_info_t pkt_info;
+	uint32_t pkts_sent;
+	uint32_t num_pkts = MIN(50, MAX_PKTS);
+	uint32_t pkt_len = 256;
+	int i;
+
+	memset(stats_start, 0, sizeof(stats_start));
+	memset(&stats_stop, 0, sizeof(stats_stop));
```
MatiasElo, line 4344: Maybe it would be better to remove these. They may hide a scenario where the implementation doesn't set unsupported fields to zero.

test/validation/api/traffic_mngr/traffic_mngr.c
```diff
@@ -4309,6 +4309,116 @@ static void traffic_mngr_test_queue_stats(void)
...
+
+		if (stats_capa.counter.packets)
+			CU_ASSERT(stats_stop.packets >= stats_start[i].packets + num_rcv_pkts);
+		if (stats_capa.counter.octets)
+			CU_ASSERT(stats_stop.octets >= stats_start[i].octets +
+				  (num_rcv_pkts * pkt_len));
+		CU_ASSERT((stats_stop.discards - stats_start[i].discards) == 0);
+		CU_ASSERT((stats_stop.discard_octets - stats_start[i].discard_octets) == 0);
+
+		printf("\nLevel %u TM node statistics\n-------------------\n", i);
```
MatiasElo, line 4401: Nitpicking but type should be `%d`.

include/odp/api/spec/traffic_mngr.h
```diff
@@ -234,6 +234,51 @@ typedef struct odp_tm_queue_stats_capability_t {
 	};
 } odp_tm_queue_stats_capability_t;
 
+/** TM node specific statistics counters */
+typedef struct odp_tm_node_stats_t {
+	/** Number of packets discarded after scheduling/shaping. */
+	uint64_t discards;
```
MatiasElo, line 240: We were discussing internally if we should specify more precisely at which node the packets were discarded if there are multiple connected TM nodes. @JannePeltonen can probably formulate the thinking better.

JannePeltonen: Unfortunately the TM api appears (at least to me) somwhat ambiguous regarding the conceptual traffic management model being configured. One way to think about the model is such that all queueing happens in TM queues and the node hierarchy performs only scheduling and does not internally drop any packets. That is also what the (probably outdated) TM user guide suggests. In that model discard or drop counters would seem to make sense only in TM queues, where they already exist.

The introduction of rate limiting as an alternative to rate shaping changed the model such that now packet drop can conceptually happen within the node hierarcy too. Also the WRED config in the nodes imply that there are packet drop capable queues (virtual or real) between TM nodes but it is unclear to me how WRED in an internal node (as opposed to a TM queue) is supposed to behave.

When a packet get dropped by a hierarchically configured TM system, in which node the discard counter gets incremented? To put it in other way, how does the discard counter map to the currently underspecified conceptual model of the TM system? I suppose specifying how exactly the discard counter behaves (and which nodes count what) when there is scheduling, shaping, rate limiting, WRED and whatnot involved in a multilevel tree structure, requires that the underlying conceptual TM model behind the TM API gets clarified.

include/odp/api/spec/traffic_mngr.h
```diff
@@ -2144,6 +2191,20 @@ typedef struct {
 int odp_tm_node_fanin_info(odp_tm_node_t             tm_node,
 			   odp_tm_node_fanin_info_t *info);
 
+/** Get statistics for a TM node
+ *
+ * The odp_tm_node_stats() function is used to extract runtime statistics
+ * associated with a given tm_node. Counters not supported by the node are set
+ * to zero.
```
MatiasElo, line 2198: To better cover possible implementation specifics the text could be:

```
 * Read statistics counters associated with the given TM node. Counters not
 * supported by the node are set to zero by the implementation.
 *
 * Depending on the implementation, certain statistics counters may be updated
 * only on specific nodes in the TM node tree (e.g. by the final node in the
 * tree even if the packet was marked to be dropped by an earlier node).
```

include/odp/api/spec/traffic_mngr.h
MatiasElo: We discussed about this and came up with the following solution. Documentation for the discards fields could be simplified a bit:
```
	/** Number of discarded packets. */
 	uint64_t discards;
 
	/** Number of octets in discarded packets. */
 	uint64_t discard_octets;

```
And then `odp_tm_node_stats()` function could document the rest. See comment there.

MatiasElo (commented):
Final tweaks to the API. There are also couple remaining comments for the implementation and validation test.

MatiasElo:
Hi @nithind1988, we came up with a possible solution for the remaining API questions. Please take a look.

psavol (approved):
Reviewed API spec commit

MatiasElo (approved):
Copyrights should be changed to: `Copyright (c) 2022-2025 Marvell` but otherwise OK.

PR #1331, opened 2021-07-06
[PATCH v3] api: packet/queue: add separate header for type definitions

MatiasElo:
Enable easier function inlining by moving type definitions into a separate
header file.


yuri-tolstov:
Hi Matias. Now things becomes confusing and API integrity, kind of, broken. Broken, because, the API packet*.h and queue*.h organization is different ("asymmetric") from other modules. Or you are planning to make this split for all modules? Confusing, because, the role of multiple packet*.h files is not clear. For example, in the include/odp/api/abi-default/ directory there are 3 packet files, two of which are empty (packet_flags.h and packet.h). What is the role of packet.h? Why packet_flags.h is separate? Shouldn't its contents to be a part of packet_types.h? Having too many not fully structured (put in hierarchy) files, make the use of API being more difficult.

MatiasElo:
> Hi Matias. Now things becomes confusing and API integrity, kind of, broken. Broken, because, the API packet*.h and queue*.h organization is different ("asymmetric") from other modules. Or you are planning to make this split for all modules? Confusing, because, the role of multiple packet*.h files is not clear. For example, in the include/odp/api/abi-default/ directory there are 3 packet files, two of which are empty (packet_flags.h and packet.h). What is the role of packet.h? Why packet_flags.h is separate? Shouldn't its contents to be a part of packet_types.h? Having too many not fully structured (put in hierarchy) files, make the use of API being more difficult.

Hi Yuri,

Yes, we are planning to use this same style for all modules. There should be one <module_name>.h file and one <module_name>_types.h file for each ODP module going forward. Since this change will require modifying multiple files (as you can see from this PR) I'd prefer doing it piece by piece to keep the PR sizes somewhat manageable.

I left `include/odp/api/abi-default/packet.h` there as a placeholder for possible ABI function implementations. However, it and `packet_flags.h` could also be removed to reduce file clutter.

I've no recollection of why a separate packet_flags module was originally added. Maybe @psavol remembers?

yuri-tolstov:
Thank you, Matias, for clarification. IMO, the less number of files in the API spec with well defined hierarchy and dependencies, the more clean API, easier its use, and simpler platform implementation.

psavol (approved):

PR #1333, opened 2021-07-07
[PATCH v5] api: tm: add unordered enqueue and multi-packet enqueue

nithind1988:
commit 07d45a29fdfcf90b9b3be2c1eaf00d7f81d54834
Date:   Tue Jul 6 23:41:45 2021 +0530

    validation: tm: set ordered enqueue and use enq multi API
    
    Set new field ordered_enqueue in tm_queue_params and also use
    odp_tm_enq_multi() API in scheduler test as burst Tx function.
    
commit f2a318af9c78c955c18acbf331d9c05f0a3f46ce
Date:   Tue Jun 29 16:50:42 2021 +0530

    linux-gen: tm: support unordered enqueue and enq multi
    
    Add support for unordered enqueue and enq multi API.
    
commit 6c9fbeca05726e3646d2f4fccb1d5cef297ab799
Date:   Sat Nov 7 16:32:37 2020 +0530

    api: tm: add unordered enqueue and multi-packet enqueue
    
    For Tx, Packet IO, with TM disabled has Direct mode for enqueueing packets
    and Queue mode for enqueueing packet or packet vector events with ordering
    maintained. For Packet IO with TM enabled, there is only single mode
    which is equivalent to Queue mode for enqueuing packets with ordering
    maintained.
    
    For benefit of disabling ordering when not needed such as a case of
    Direct mode, this patch adds per TM queue parameter to be provided at
    TM queue create time indicating whether order needs to be preserved or
    not.
    
    This patch also adds TM enqueue multi API equivalent to odp_tm_enq()
    but allowing multiple packets to be enqueued in single API call
    inline with pktout enqueue or send API's in non-TM mode.
    
This was reviewed and originally part of pull request https://github.com/OpenDataPlane/odp/pull/1237

### DIFF
--- a/platform/linux-generic/odp_traffic_mngr.c
+++ b/platform/linux-generic/odp_traffic_mngr.c
@@ -2024,14 +2024,18 @@ static int tm_enqueue(tm_system_t *tm_system,
 		drop = random_early_discard(tm_system, tm_queue_obj,
 					    initial_tm_wred_node, pkt_color);
 		if (drop)
-			return -1;
+			return -2;
 	}
 
 	work_item.queue_num = tm_queue_obj->queue_num;
 	work_item.pkt = pkt;
-	_odp_sched_fn->order_lock();
+	if (tm_queue_obj->ordered_enqueue)
+		_odp_sched_fn->order_lock();
+
 	rc = input_work_queue_append(tm_system, &work_item);
-	_odp_sched_fn->order_unlock();
+
+	if (tm_queue_obj->ordered_enqueue)
+		_odp_sched_fn->order_unlock();
 
 	if (rc < 0) {
 		ODP_DBG("%s work queue full\n", __func__);
@@ -3961,6 +3965,8 @@ int odp_tm_node_context_set(odp_tm_node_t tm_node, void *user_context)
 void odp_tm_queue_params_init(odp_tm_queue_params_t *params)
 {
 	memset(params, 0, sizeof(odp_tm_queue_params_t));
+
+	params->ordered_enqueue = true;
 }
 
 odp_tm_queue_t odp_tm_queue_create(odp_tm_t odp_tm,
@@ -3997,6 +4003,7 @@ odp_tm_queue_t odp_tm_queue_create(odp_tm_t odp_tm,
 		memset(queue_obj, 0, sizeof(tm_queue_obj_t));
 		queue_obj->user_context = params->user_context;
 		queue_obj->priority = params->priority;
+		queue_obj->ordered_enqueue = params->ordered_enqueue;
 		queue_obj->tm_idx = tm_system->tm_idx;
 		queue_obj->queue_num = (uint32_t)_odp_int_pkt_queue;
 		queue_obj->_odp_int_pkt_queue = _odp_int_pkt_queue;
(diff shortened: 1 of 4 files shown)


test/validation/api/traffic_mngr/traffic_mngr.c
```diff
@@ -1098,6 +1098,46 @@ static int make_pkts(uint32_t    num_pkts,
...
+		xmt_pkt_desc->xmt_idx = xmt_pkt_idx + i;
+		xmt_pkt_desc->xmt_time = odp_time_local();
+		xmt_pkt_desc->tm_queue = tm_queue;
+		pkts_sent++;
+	}
+
+	/* Free rejected pkts */
+	if (rc < num_pkts) {
+		i = rc > 0 ? rc : 0;
+		for (i = rc; i < num_pkts; i++) {
```
JannePeltonen, line 1127: Here in the for-statement i gets assigned again, potentially to a negative value. Just removing that should work, but maybe you could just rely on the fact that i has the correct value after the previous for loop and this loop is just a continuation of that. So maybe you could do this:

````

+               pkts_sent++;
+       }
+       /* Free rejected pkts */
+       for (; i < num_pkts; i++) {
+               xmt_pkt_desc = &xmt_pkt_descs[xmt_pkt_idx + i];
````

Otherwise this PR looks ok. I think you could add my review tag already in the next version.

File: test/validation/api/traffic_mngr/traffic_mngr.c [hunk 2]
```diff
@@ -1098,6 +1098,45 @@ static int make_pkts(uint32_t    num_pkts,
...
+	for (i = 0; i < rc; i++) {
+		xmt_pkt_desc = &xmt_pkt_descs[xmt_pkt_idx + i];
+		xmt_pkt_desc->xmt_idx = xmt_pkt_idx + i;
+		xmt_pkt_desc->xmt_time = odp_time_local();
+		xmt_pkt_desc->tm_queue = tm_queue;
+		pkts_sent++;
+	}
+
+	/* Free rejected pkts */
+	if (rc < num_pkts) {
```
JannePeltonen, line 1125: You could still remove this rc < num_pkts checking here since the loop body will not be executed anyway if rc >= num_pkts. If you want to check that rc is not larger than num_pkts, then that should be a CU_ASSERT right after the odp_tm_enq_multi() call.

JannePeltonen (approved):

platform/linux-generic/odp_traffic_mngr.c
```diff
@@ -4438,6 +4445,37 @@ int odp_tm_enq_with_cnt(odp_tm_queue_t tm_queue, odp_packet_t pkt)
...
+
+	for (i = 0; i < num; i++) {
+		rc = tm_enqueue(tm_system, tm_queue_obj, packets[i]);
+		if (rc < 0)
+			break;
+	}
+
+	/* Free rejected buffers */
+	for (; i < num; i++)
+		odp_packet_free(packets[i]);
```
JannePeltonen, line 4474: @psavol noticed that there is something wrong here. An indeed there is. If tm_enqueue() decides to drop a packet because of WRED, it does not mean that all subsequent packets would be dropped. So freeing the rest of the packets in this loop is likely to drop packets that would go through without the probability based WRED algorithm dropping them. In other words, this effectively breaks the logic of the WRED dropper.

In case tm_enqueue() returns a failure because of WRED drop, the correct thing to do would be to free only that packet and to continue calling tm_enqueue() for the rest of the packets.

One problem is that tm_enqueue() does not tell if it is dropping a packet because of WRED or for other reasons. The simple thing to do would be to continue as if all failures were because of WRED (which is likely). Another possibility  would be to distinguish WRED drops from other failures in the return code of tm_enqueue(). It seems that in case of WRED drop, tm_enqueue() could easily return e.g. -2 instead of -1 and tm_enq_multi() could then check that. In case of other tm_enqueue() failures than WRED, tm_enq_multi() should stop calling tm_enqueue() and then return the number of packets consumed that far without freeing the rest.

nithind1988: Ack, fixed it by changing return value of red drop to -2 and freeing pkt and continuing if it is red drop

JannePeltonen: Now that WRED drops are distinguished from other errors, the function should not consume all packets if it encounters an other error than WRED drop. So instead of freeing the rest of the packets and returning num, the function should not free the packets but return the number of packets consumed before the error (i.e. the value of i).

This also means that tail drop (in addition to some real internal errors) does not cause packets to be consumed but the application must do the actual drop in that case. In some cases that might be slightly inconvenient but it allows an application to do something else than drop the remaining packets if there is no RED configured and the queue becomes full.

JannePeltonen:
Looks good to me. Marking this again ready for merging.

PR #1334, opened 2021-07-07
[PATCH v8] api:tm: add missing capabilities and egress specific capability API

nithind1988:
commit df6ce5a7a00d0fc61690fedba0001fa0139345f2
Date:   Mon Jun 7 18:45:39 2021 +0530

    api: tm: deprecate tm capabilities API
    
    Deprecate odp_tm_capabilities() API as odp_tm_egress_capabilities()
    provides similar functionality with additional input of egress
    so that egress specific capabilities can be returned. Egress specific
    capabilities are more useful as each TM system should have an egress.
    
commit 06dda1fa4a53132a04813999f482015ec85b4b73
Date:   Tue Jul 6 22:12:58 2021 +0530

    validation: tm: use egress capabability before running tests
    
    Use egress capabilities to identify if different features are
    supported before running the test for features like, threshold,
    pkt marking, wred etc. Also mark test as inactive if required
    number of levels, queues or nodes are not supported for a test suite.
    
    This patch also splits marking tests to subtests for better control
    and replaces odp_tm_capabilities() calls with odp_tm_egress_capabilities()
    
commit 754052d51b64cfe2d9525decde492a7b305e6b75
Date:   Tue Jun 29 15:50:38 2021 +0530

    linux-gen: tm: express new capa and support egress capa API
    
    Express new capabilities for queue/node query/threshold.
    Also implement egress capability API.
    
commit 3b08ed7b87e48b9e0872df86c477a33f2167a83d
Date:   Wed Nov 4 20:06:49 2020 +0530

    api: tm: add missing capabilities and egress specific capability API
    
    Not all platforms support TM node query, node threshold, queue query,
    and queue threshold features. Since they are missing from existing
    global and level capabilities, adding the same to express no support.
    
    Also, in some platforms capabilities are specific to a egress kind
    like a PKTIO. Since capabilities provide a way to application to know
    things like how many levels can be supported etc, prior to TM creation,
    add a capability get API to retrieve egress function specific capabilities
    records.
    
This was originally part of https://github.com/OpenDataPlane/odp/pull/1237

### DIFF
--- a/include/odp/api/spec/traffic_mngr.h
+++ b/include/odp/api/spec/traffic_mngr.h
@@ -306,6 +306,10 @@ typedef struct {
 	 * When true the min_weight and max_weight fields above specify
 	 * the legal range of such weights. */
 	odp_bool_t weights_supported;
+
+	/** tm_node_threshold indicates that the tm_nodes at this
+	 * level support threshold profiles. */
+	odp_bool_t tm_node_threshold;
 } odp_tm_level_capabilities_t;
 
 /** TM Capabilities Record.
@@ -447,6 +451,23 @@ typedef struct {
 	/** TM queue statistics counter capabilities */
 	odp_tm_queue_stats_capability_t queue_stats;
 
+	/** tm_queue_threshold indicates support for threshold profile on a
+	 * TM queue. When TRUE, users can set/clear/update threshold profile
+	 * on a TM queue. When false none of it is supported.
+	 */
+	odp_bool_t tm_queue_threshold;
+
+	/** tm_queue_query_flags indicates supported types of TM queue query.
+	 * Types of TM queue query are same as query_flags that are passed to
+	 * odp_tm_queue_query(), odp_tm_priority_query() and
+	 * odp_tm_total_query(). When zero, none of the queue query API's are
+	 * supported. When non-zero, only the only supported types of passed
+	 * query_flags are taken into account and corresponding fields updated.
+	 *
+	 * @see ODP_TM_QUERY_PKT_CNT, ODP_TM_QUERY_BYTE_CNT,
+	 * ODP_TM_QUERY_THRESHOLDS.
+	 */
+	uint32_t tm_queue_query_flags;
 } odp_tm_capabilities_t;
 
 /** Per Level Requirements
@@ -504,6 +525,10 @@ typedef struct {
 	 * fanins.  When true the min_weight and max_weight fields above
 	 * specify the used range of such weights. */
 	odp_bool_t weights_needed;
+
+	/** tm_node_threshold_needed indicates that the tm_nodes at this
+	 * level may use threshold profile support */
+	odp_bool_t tm_node_threshold_needed;
 } odp_tm_level_requirements_t;
 
 /** TM Requirements Record.
@@ -532,6 +557,10 @@ typedef struct {
 	 * expected to use the dual slope WRED capability.  This field is
 	 * ignored if tm_queue_wred_needed above is false. */
 	odp_bool_t tm_queue_dual_slope_needed;
+
+	/** tm_queue_threshold_needed indicates that the tm_queues are
+	 * expected to use threshold profile support */
+	odp_bool_t tm_queue_threshold_needed;
 
 	/** vlan_marking_needed indicates that the ODP application expects
 	 * to use some form of VLAN egress marking using the
@@ -618,19 +647,20 @@ void odp_tm_egress_init(odp_tm_egress_t *egress);
 
 /** Query All TM Capabilities
  *
- * The odp_tm_capabilities() function can be used to obtain the complete set of
- * TM limits supported by this implementation.  The reason that this returns
- * a SET of capabilities and not just one, is because it is expected that
- * many HW based implementations may have one set of limits for the HW and
- * also support a SW TM implementation with a (presumably larger) different
- * set of limits.  There are also cases where there could be more than
- * SW implementation (one supporting say tens of thousands of tm_queues and
- * a variant supporting tens of millions of tm_queues).
+ * This function returns the set of TM capabilities that are common for all
+ * egresses. The reason that this returns a SET of capabilities and not just
+ * one, is because it is expected that many HW based implementations may have
+ * one set of limits for the HW and also support a SW TM implementation with a
+ * (presumably larger) different set of limits. There are also cases where
+ * there could be more than one SW implementation (one supporting say tens of
+ * thousands of tm_queues and a variant supporting tens of millions of
+ * tm_queues). It returns capabilities that are valid for all egresses.
  * The caller passes in an array of odp_tm_capabilities_t records and the
- * number of such records.  Then the first N of these records will be filled
- * in by the implementation and the number N will be returned.  In the event
- * that N is larger than the capabilities_size, N will still be returned,
- * but only capabilities_size records will be filled in.
+ * maximum number of such records to output. If number of such records
+ * implementation supports is larger than caller requested number, then
+ * only caller requested number of records are written and return value is
+ * max number of records implementation supports.
+ * Caller then may again call with larger number of records to be returned.
  *
  * @param[out] capabilities      An array of odp_tm_capabilities_t records to
  *                               be filled in.
@@ -644,6 +674,30 @@ void odp_tm_egress_init(odp_tm_egress_t *egress);
  */
 int odp_tm_capabilities(odp_tm_capabilities_t capabilities[],
 			uint32_t              capabilities_size);
+
+/** Query TM Capabilities specific to an egress
+ *
+ * The function returns the set of TM limits supported by this implementation
+ * for a given egress. Unlike odp_tm_capability() which return's capabilities
+ * of already created TM system which are limited by its requirements, this
+ * function returns maximum TM system limits.
+ *
+ * Lack of TM support in the given egress does not cause this
+ * function to return a failure. Lack of TM support is indicated
+ * by zero max_tm_queues capability.
+ *
+ * If the pktio of an egress of the pktio kind has not been opened
+ * in the ODP_PKTOUT_MODE_TM pktout mode, the capabilities will
+ * indicate that TM is not supported.
+ *
+ * @param[out] capabilities      odp_tm_capabilities_t record to be filled in.
+ * @param      egress            Only capabilities compatible with this egress
+ *                               are returned.
+ * @retval 0  on success
+ * @retval <0 on failure
+ */
+int odp_tm_egress_capabilities(odp_tm_capabilities_t *capabilities,
+			       const odp_tm_egress_t *egress);
 
 /** Create/instantiate a TM Packet Scheduling system.
  *
@@ -2044,12 +2098,10 @@ typedef struct {
 } odp_tm_query_info_t;
 
 /** The odp_tm_queue_query() function can be used to check a single tm_queue's
- * queue utilization.  The query_flags indicate whether or not packet counts,
- * byte counts or both are being requested.  It is an error to request
- * neither.  The implementation may still return both sets of counts
- * regardless of query_flags if the cost of returning all the counts is
- * comparable to the cost of checking the query_flags. The info structure is
- * written only on success.
+ * queue utilization.  The query flags indicate which information is being
+ * requested.
+ * The implementation may choose to return additional information that was not
+ * requested. The info structure is written only on success.
  *
  * @param      tm_queue     Specifies the tm_queue (and indirectly the
  *                          TM system).
@@ -2064,12 +2116,10 @@ int odp_tm_queue_query(odp_tm_queue_t       tm_queue,
 		       odp_tm_query_info_t *info);
 
 /** The odp_tm_priority_query() function can be used to check the queue
- * utilization of all tm_queue's with the given priority.  The query_flags
- * indicate whether or not packet counts, byte counts or both are being
- * requested.  It is an error to request neither.  The implementation may
- * still return both sets of counts regardless of query_flags if the cost of
- * returning all the counts is comparable to the cost of checking the
- * query_flags. The info structure is written only on success.
+ * utilization of all tm_queue's with the given priority.  The query flags
+ * indicate which information is being requested. The implementation may
+ * choose to return additional information that was not requested.
+ * The info structure is written only on success.
  *
  * @param      odp_tm       Specifies the TM system.
  * @param      priority     Supplies the strict priority level used to specify
@@ -2087,11 +2137,9 @@ int odp_tm_priority_query(odp_tm_t             odp_tm,
 
 /** The odp_tm_total_query() function can be used to check the queue
  * utilization of all tm_queue's in a single TM system.  The query_flags
- * indicate whether or not packet counts, byte counts or both are being
- * requested.  It is an error to request neither.  The implementation may
- * still return both sets of counts regardless of query_flags if the cost of
- * returning all the counts is comparable to the cost of checking the
- * query_flags. The info structure is written only on success.
+ * indicate which information is being requested. The implementation may
+ * choose to return additional information that was not requested.
+ * The info structure is written only on success.
  *
  * @param      odp_tm       Specifies the TM system.
  * @param      query_flags  A set of flag bits indicating which counters are
--- a/include/odp/api/spec/traffic_mngr.h
+++ b/include/odp/api/spec/traffic_mngr.h
@@ -647,6 +647,9 @@ void odp_tm_egress_init(odp_tm_egress_t *egress);
 
 /** Query All TM Capabilities
  *
+ * @deprecated Use odp_tm_egress_capabilities() instead that also additionally
+ * takes egress as input to provide capabilities specific to a given egress.
+ *
  * This function returns the set of TM capabilities that are common for all
  * egresses. The reason that this returns a SET of capabilities and not just
  * one, is because it is expected that many HW based implementations may have
@@ -672,8 +675,8 @@ void odp_tm_egress_init(odp_tm_egress_t *egress);
  *                               implementations supports. *NOTE* that this
  *                               number can be > capabilities_size!
  */
-int odp_tm_capabilities(odp_tm_capabilities_t capabilities[],
-			uint32_t              capabilities_size);
+int ODP_DEPRECATE(odp_tm_capabilities)(odp_tm_capabilities_t capabilities[],
+				       uint32_t              capabilities_size);
 
 /** Query TM Capabilities specific to an egress
  *
--- a/platform/linux-generic/odp_traffic_mngr.c
+++ b/platform/linux-generic/odp_traffic_mngr.c
@@ -2555,8 +2555,8 @@ void odp_tm_egress_init(odp_tm_egress_t *egress)
 	memset(egress, 0, sizeof(odp_tm_egress_t));
 }
 
-int odp_tm_capabilities(odp_tm_capabilities_t capabilities[],
-			uint32_t              capabilities_size)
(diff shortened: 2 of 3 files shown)


include/odp/api/spec/traffic_mngr.h
```diff
@@ -583,6 +614,40 @@ void odp_tm_egress_init(odp_tm_egress_t *egress);
...
+ * @param      egress            Only capabilities compatible with this egress
+ *                               are returned.
+ * @return                       Returns < 0 upon failure. Returns N > 0,
+ *                               where N is the maximum number of different
+ *                               odp_tm_capabilities_t records that the
+ *                               implementations supports. *NOTE* that this
+ *                               number can be > capabilities_size!
+ */
+int odp_tm_egress_capabilities(odp_tm_capabilities_t capabilities[],
+			       uint32_t              capabilities_size,
```
JannePeltonen, line 651: Maybe we should return just single capability structure for an egress as discussed, and since this is a new API, we could do it right from the start. The old API can be handled separately.

include/odp/api/spec/traffic_mngr.h
```diff
@@ -556,19 +586,20 @@ void odp_tm_egress_init(odp_tm_egress_t *egress);
...
  *
- * The odp_tm_capabilities() function can be used to obtain the complete set of
- * TM limits supported by this implementation.  The reason that this returns
- * a SET of capabilities and not just one, is because it is expected that
- * many HW based implementations may have one set of limits for the HW and
- * also support a SW TM implementation with a (presumably larger) different
- * set of limits.  There are also cases where there could be more than
- * SW implementation (one supporting say tens of thousands of tm_queues and
- * a variant supporting tens of millions of tm_queues).
+ * This function returns the complete set of TM capabilities for all egresses.
```
JannePeltonen, line 592: Now that I read this I think the wording I suggested may give the wrong impression that the array parameter is there so that the function can return a capa struct for every egress. So I would like to propose another rewording for the opening sentence: "This function returns the set of TM capabilities that are common for all egresses."

include/odp/api/spec/traffic_mngr.h
```diff
@@ -583,6 +614,22 @@ void odp_tm_egress_init(odp_tm_egress_t *egress);
...
+ * function returns maximum TM system limits.
+ *
+ * @param[out] capabilities      odp_tm_capabilities_t record to be filled in.
+ * @param      egress            Only capabilities compatible with this egress
+ *                               are returned.
+ * @retval 0  on success
+ * @retval <0 on failure
+ */
+int odp_tm_egress_capabilities(odp_tm_capabilities_t *capabilities,
+			       odp_tm_egress_t       *egress);
```
JannePeltonen, line 634: In ODP it is customary that in case that a feature is not supported, the corresponding capability function does not return a failure but successfully returns a capability structure that indicates that the feature is not supported. Failure of a capability function should not normally happen and indicates some sort of internal error. When reading the validation test I realized that you have a different idea of how this function works.

Another issue is that it may be unclear how this function behaves when called with a pktio that has not been opened in the TM output mode. 

To clarify the semantics, maybe the following should be added in the description:

```
Lack of TM support in the given egress does not cause this
function to return a failure. Lack of TM support is indicated
by zero max_tm_queues capability.

If the pktio of an egress of the pktio kind has not been opened
in the ODP_PKTOUT_MODE_TM pktout mode, the capabilities will
indicate that TM is not supported.
```

Another possibility would be to return a failure if TM is not supported, but then maybe the failure should be distinguished from other kind of failures by having different return values. But it seems that typically capability functions in ODP do not fail.

JannePeltonen, line 634: The egress argument should be a const pointer.

platform/linux-generic/odp_traffic_mngr.c
```diff
@@ -2542,8 +2542,8 @@ void odp_tm_egress_init(odp_tm_egress_t *egress)
 	memset(egress, 0, sizeof(odp_tm_egress_t));
 }
 
-int odp_tm_capabilities(odp_tm_capabilities_t capabilities[] ODP_UNUSED,
-			uint32_t              capabilities_size)
+static int tm_capabilities(odp_tm_capabilities_t capabilities[],
```
JannePeltonen, line 2545: Changing this function to an internal one and creating a new odp_tm_capabilities function that just calls this seems totally unnecessary to me. I do not see why this could not be left as it is (apart from removing the incorrect ODP_UNUSED attribute).

test/validation/api/traffic_mngr/traffic_mngr.c
```diff
@@ -20,7 +20,6 @@
```
JannePeltonen, line 20: There is a type (capabability) in the subject line of the commit message. How about:
validation: tm: check egress capababilities before running tests

test/validation/api/traffic_mngr/traffic_mngr.c
```diff
@@ -386,35 +385,32 @@ static odp_bool_t approx_eq64(uint64_t val, uint64_t correct)
...
+	odp_tm_egress_t              egress;
 	uint32_t                     num_records, idx, num_levels, level;
 	int                          rc;
 
-	rc = odp_tm_capabilities(capabilities_array, MAX_CAPABILITIES);
+	odp_tm_egress_init(&egress);
+	egress.egress_kind = ODP_TM_EGRESS_PKT_IO;
+	egress.pktio = xmt_pktio;
+
+	rc = odp_tm_egress_capabilities(&capabilities_array[0], &egress);
```
JannePeltonen, line 398: It should be checked that the return value is zero (positive return values are not possible according to the API spec and negatives indicate errors).

Btw, the CU_ASSERT in the existing code below seems to be incorrect (I think the intention is to trigger the assert in the failure case but now it is just dead code).

test/validation/api/traffic_mngr/traffic_mngr.c
```diff
@@ -386,35 +385,32 @@ static odp_bool_t approx_eq64(uint64_t val, uint64_t correct)
...
-	if (MAX_CAPABILITIES < rc)
-		ODPH_DBG("There were more than %u capabilities (%u)\n",
-			 MAX_CAPABILITIES, rc);
-	else
-		num_records = rc;
+	odp_tm_egress_init(&egress);
+	egress.egress_kind = ODP_TM_EGRESS_FN;
+	rc = odp_tm_egress_capabilities(&capabilities_array[1], &egress);
+	if (!rc)
+		num_records++;
```
JannePeltonen, line 413: This seems to assume that lack of TM support causes the capability function to fail instead of returning capabilities that indicate no support. That was a bit of a surprise to me. See the comment in the API patch.

test/validation/api/traffic_mngr/traffic_mngr.c
```diff
@@ -386,35 +385,32 @@ static odp_bool_t approx_eq64(uint64_t val, uint64_t correct)
...
-		CU_ASSERT(rc != 0);
-		return -1;
-	}
-
-	/* Now test the return code to see if there were more capabilities
-	 * records than the call above allowed for.  This is not an error,
-	 * just an interesting fact.
+	/* Get capabilities for egress kind function, it is not mandatory
+	 * to support the same as the current test suite only tests with
+	 * egress packet IO.
```
JannePeltonen, line 407: > it is not mandatory to support the same as the current test suite only tests with egress packet IO.

The reason that the test suite does not test other kind of egresses may be that there is simply a gap in testing.

test/validation/api/traffic_mngr/traffic_mngr.c
```diff
@@ -1610,12 +1606,45 @@ static uint32_t find_child_queues(uint8_t         tm_system_idx,
...
+		req->vlan_marking_needed = true;
+	if (tm_capabilities.tm_queue_threshold)
+		req->tm_queue_threshold_needed = true;
+
+	for (j = 0; j < tm_capabilities.max_levels; j++) {
+		if (tm_capabilities.per_level[j].tm_node_threshold)
+			req->per_level[j].tm_node_threshold_needed = true;
+	}
+
+	/* Mark colors as supported even if one of them supports marking */
```
JannePeltonen, line 1636: I did not understand this comment (especially the "even if one of them supports marking" part). Maybe you could clarify it or just remove it.

platform/linux-generic/odp_traffic_mngr.c
nithind1988: You mean, I should use odp_tm_capabilities() as it is in odp_tm_egress_capabilities() ?

include/odp/api/spec/traffic_mngr.h
nithind1988: Ack, I forgot about that. Will fix the usage in test and comment.

test/validation/api/traffic_mngr/traffic_mngr.c
nithind1988: It was a type on comment. I actually meant
 /* Mark colors as needed if at least one of the marking feature is needed */

platform/linux-generic/odp_traffic_mngr.c
JannePeltonen: Yes, why not?

nithind1988: Ack, just wanted to confirm.

nithind1988: I realized later that in the last patch, we are deprecating odp_tm_capabilities(). So using it in odp_tm_egress_capabilities() as ODP_DEPRECATED(odp_tm_capabilities) then is right I guess ?

JannePeltonen: I was thinking about the patch that introduced the change and forgot the deprecation in the later patch. So yes, the new internal function is needed. Having the change now in the last patch is logical.

nithind1988: Ack.

platform/linux-generic/odp_traffic_mngr.c
```diff
@@ -2589,19 +2593,33 @@ int odp_tm_capabilities(odp_tm_capabilities_t capabilities[] ODP_UNUSED,
...
 }
 
+int odp_tm_egress_capabilities(odp_tm_capabilities_t *capabilities,
+			       const odp_tm_egress_t *egress)
+{
+	int ret;
+
+	(void)egress;
+
+	ret = odp_tm_capabilities(capabilities, 1);
```
JannePeltonen: Now that the API says that pktio type egresses without TM enabled cause capability to indicate no TM support, the implementation should be changed accordingly. So you need to check if get_pktio_entry()->s.param.out_mode is ODP_PKTOUT_MODE_TM before calling the common capability function.

test/validation/api/traffic_mngr/traffic_mngr.c
JannePeltonen: I managed to copy paste the original typo in my comment (sorry about that) and now it got copy pasted in the latest comment, so the subject line still contains "capababilities".

test/validation/api/traffic_mngr/traffic_mngr.c
```diff
@@ -386,35 +385,36 @@ static odp_bool_t approx_eq64(uint64_t val, uint64_t correct)
...
 	int                          rc;
 
-	rc = odp_tm_capabilities(capabilities_array, MAX_CAPABILITIES);
-	if (rc < 0) {
-		CU_ASSERT(rc < 0);
+	odp_tm_egress_init(&egress);
+	egress.egress_kind = ODP_TM_EGRESS_PKT_IO;
+	egress.pktio = xmt_pktio;
+
+	/* This egress function should already be supported as it is
```
JannePeltonen, line 398: Maybe you mean "egress kind" instead of "egress function"? It would also be ok to remove the comment.

test/validation/api/traffic_mngr/traffic_mngr.c
nithind1988: Sorry, I also missed to read it while copy pasting.

test/validation/api/traffic_mngr/traffic_mngr.c
```diff
@@ -386,35 +385,36 @@ static odp_bool_t approx_eq64(uint64_t val, uint64_t correct)
...
+	 * only then the test suite runs.
+	 */
+	rc = odp_tm_egress_capabilities(&capabilities_array[0], &egress);
+	CU_ASSERT(rc == 0);
+	if (rc < 0)
```
JannePeltonen, line 403: This condition (and the same for the other call below) could be "rc !=0" as we do not want to continue if the capa function returns an unexpected return value. Or the assert could be CU_ASSERT_FATAL().

test/validation/api/traffic_mngr/traffic_mngr.c
```diff
@@ -1627,14 +1662,10 @@ static int create_tm_system(void)
...
-	requirements.tm_queue_shaper_needed     = true;
-	requirements.tm_queue_wred_needed       = true;
-	requirements.tm_queue_dual_slope_needed = true;
-	requirements.vlan_marking_needed        = false;
-	requirements.ecn_marking_needed         = true;
-	requirements.drop_prec_marking_needed   = true;
-	for (color = 0; color < ODP_NUM_PACKET_COLORS; color++)
-		requirements.marking_colors_needed[color] = true;
+	/* We are not using TM queue shaping */
+	requirements.tm_queue_shaper_needed     = false;
```
JannePeltonen, line 1666: This appears incorrect since shaper_profiles[0] is set to a TM queue in create_tm_queue(). But maybe the old code is a bit obfuscated and the shaper profile is actually ODP_TM_INVALID at that point so that the shaper is actually disabled?

But even then, the validation test could still require the queue level shaper if it is supported. Node thresholds do not seem to be used either but your code enables them anyway, which may slightly improve test coverage.

The enabling could be added in the get_supported_marking_wred_threshold() function that could also be renamed at the same time (set_reqs_based_on_capas()?).

test/validation/api/traffic_mngr/traffic_mngr.c
```diff
@@ -2156,26 +2190,52 @@ static int traffic_mngr_suite_init(void)
...
+
+	/* Get TM capabilities */
+	ret = odp_tm_egress_capabilities(&egress_capa, &egress);
+	if (ret) {
+		ODPH_ERR("Failed to retrieve tm capabilities");
+		return ret;
+	}
+
+	/* Check for sufficient TM queues */
+	if (egress_capa.max_tm_queues < NUM_TM_QUEUES)
```
JannePeltonen, line 2205: This is inconsistent with create_tm_system() that sets requirements.max_tm_queues to NUM_TM_QUEUES + 1. I do not know why it does it and I think it is bad that NUM_TM_QUEUES does not really mean the total number of TM queues (or if it does then the +1 is there for no reason), so maybe the best fix is not to change this to NUM_TM_QUEUES + 1 but something else.

test/validation/api/traffic_mngr/traffic_mngr.c
```diff
@@ -2156,26 +2190,52 @@ static int traffic_mngr_suite_init(void)
...
+	if (j != NUM_LEVELS)
+		goto skip_tests;
+
+	/* Init tm capabilities with matching egress capa until tm is created */
+	tm_capabilities = egress_capa;
+
 	/* Fetch initial dynamic update capabilities, it will be updated
 	 * later after TM system is created.
 	 */
-	ret = odp_tm_capabilities(capabilities_array, MAX_CAPABILITIES);
```
JannePeltonen, line 2162: Now that this line is removed, I think the comment above the line should also be removed as the comment becomes unaccurate and not helpful.

test/validation/api/traffic_mngr/traffic_mngr.c
```diff
@@ -4067,6 +4127,8 @@ static void traffic_mngr_test_tm_create(void)
 
 static void traffic_mngr_test_shaper(void)
 {
+	CU_ASSERT_FATAL(odp_tm_systems[0] != ODP_TM_INVALID);
```
JannePeltonen, line 4130: I do not quite see why we need these CUnit asserts in all these test functions. It seems a bit out of place as I think the CUnit assert should be where the TM system is created and the rest of the tests should not even run if things failed already at TM creation.

test/validation/api/traffic_mngr/traffic_mngr.c
nithind1988: Ack, will set it if supported based on caps

test/validation/api/traffic_mngr/traffic_mngr.c
nithind1988: Reduced it to NUM_TM_QUEUES at requirements init.

test/validation/api/traffic_mngr/traffic_mngr.c
```diff
@@ -386,35 +385,31 @@ static odp_bool_t approx_eq64(uint64_t val, uint64_t correct)
...
-	 * to make sure that there is at least ONE capabilities record
-	 * returned */
-	if (rc == 0) {
-		CU_ASSERT(rc != 0);
+	/* Get capabilities for egress kind function. */
+	odp_tm_egress_init(&egress);
+	egress.egress_kind = ODP_TM_EGRESS_FN;
+	rc = odp_tm_egress_capabilities(&capabilities_array[1], &egress);
+	CU_ASSERT(rc == 0);
+	if (rc < 0)
```
JannePeltonen, line 407: This could also be changed the same way as the similar condition little earlier, i.e. we want to stop the test if the capabilities call returns unexpected return code, even if it is positive.

Otherwise this PR looks ok to. With the above change, Reviewed-by: Janne Peltonen <janne.peltonen@nokia.com>

JannePeltonen (approved):

MatiasElo:
Hi @nithind1988, this PR requires a rebase due to some merge conflicts.

nithind1988:
Hi Matias,

Ack, will rebase it and send a new version.


Thanks
Nithin

From: Matias Elo ***@***.***>
Sent: Monday, July 12, 2021 12:30 PM
To: OpenDataPlane/odp ***@***.***>
Cc: Nithin Kumar Dabilpuram ***@***.***>; Mention ***@***.***>
Subject: [EXT] Re: [OpenDataPlane/odp] [PATCH v7] api:tm: add missing capabilities and egress specific capability API (#1334)

External Email
________________________________

Hi @nithind1988<https://urldefense.proofpoint.com/v2/url?u=https-3A__github.com_nithind1988&d=DwMCaQ&c=nKjWec2b6R0mOyPaz7xtfQ&r=FZ_tPCbgFOh18zwRPO9H0yDx8VW38vuapifdDfc8SFQ&m=whs0o79RcLwTlmYcHmDkD5O1AElh6wzZ4tP_m2fON5I&s=iAp1yNamnUHgbmmkiRTyv_3RizR8kHryoFGfFMiK50w&e=>, this PR requires a rebase due to some merge conflicts.

test/validation/api/traffic_mngr/traffic_mngr.c
```diff
@@ -4157,12 +4215,7 @@ static void traffic_mngr_test_thresholds(void)
 
 static int traffic_mngr_check_queue_stats(void)
 {
-	odp_tm_capabilities_t capa;
-
-	if (odp_tm_capabilities(&capa, 1) < 1)
-		return ODP_TEST_INACTIVE;
-
-	if (capa.queue_stats.all_counters == 0)
+	if (tm_capabilities.queue_stats.all_counters == 0)
```
nithind1988, line 4218: Rebased and modified this to use tm_capabilities as odp_tm_capabilities() is deprecated

PR #1336, opened 2021-07-08
[PATCH v3] validation: crypto: remove test descriptions

Malvika-Gupta:
The test case for AES_GCM algorithm incorrectly uses 3DES_CBC algorithm in
the description. Also, many test case descriptions had missing full stops
at the end of the sentence.

Reviewed-by: Govindarajan <govindrajan.mohandoss@arm.com>

### DIFF
--- a/test/validation/api/crypto/odp_crypto_test_inp.c
+++ b/test/validation/api/crypto/odp_crypto_test_inp.c
@@ -1035,10 +1035,6 @@ static int check_alg_3des_cbc(void)
 	return check_alg_support(ODP_CIPHER_ALG_3DES_CBC, ODP_AUTH_ALG_NULL);
 }
 
-/* This test verifies the correctness of encode (plaintext -> ciphertext)
- * operation for 3DES_CBC algorithm. IV for the operation is the session IV.
- * In addition the test verifies if the implementation can use the
- * packet buffer as completion event buffer.*/
 static void crypto_test_enc_alg_3des_cbc(void)
 {
 	check_alg(ODP_CRYPTO_OP_ENCODE,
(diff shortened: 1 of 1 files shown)


JannePeltonen:
I would just remove the comments rather than trying to maintain them. As all the functions call check_alg() anyway it is quite obvious what they test so the comments look to me more as noise than something helpful. If something does needs to be explained, then it probably should be done in a comment about check_alg() and not as repeated comments in all the callers.

Malvika-Gupta:
> I would just remove the comments rather than trying to maintain them. As all the functions call check_alg() anyway it is quite obvious what they test so the comments look to me more as noise than something helpful. If something does needs to be explained, then it probably should be done in a comment about check_alg() and not as repeated comments in all the callers.

Thanks @JannePeltonen for your comments. I will amend the PR to instead remove all the individual test comments and add a single comment before check_alg() instead.

JannePeltonen:
> I don't think a comment needs to be added before check_alg() as it is quite self-explanatory. Please let me know if this works

I do not see a need to add a comment there. The patch looks good to me, except that the commit message says this: " Since all test cases call check_alg(), a comment has instead been added before it explaining its function."  Please remote that.

Malvika-Gupta:
> > I don't think a comment needs to be added before check_alg() as it is quite self-explanatory. Please let me know if this works
> 
> I do not see a need to add a comment there. The patch looks good to me, except that the commit message says this: " Since all test cases call check_alg(), a comment has instead been added before it explaining its function." Please remove that.

Oops, my bad! Thanks for pointing it out, @Janne. I have removed that in the commit message and pushed the changes again.  Please take a look. Thanks!

JannePeltonen (approved):

PR #1341, opened 2021-07-14
[PATCH v12] api: tm: add spec to express shaper burst limits

nithind1988:
commit d2988a3b6bde68dc7bc863ecf3edf15f1291351a
Date:   Wed Jul 14 20:18:06 2021 +0530

    abi: tm: provide shaper burst limits
    
    Provide shaper min burst and max burst limits.
    
commit b44421e0bd4e8fd05dac955fa333c71985a84fbb
Date:   Wed Jul 14 20:17:52 2021 +0530

    api: tm: add spec to express shaper burst limits
    
    Add macro's that needs to be defined by platform to express
    shaper rate burst min and max in bytes. This is similar to existing
    shaper rate bandwidth min and max macros.
    
### DIFF
--- a/example/traffic_mgmt/odp_traffic_mgmt.c
+++ b/example/traffic_mgmt/odp_traffic_mgmt.c
@@ -1,6 +1,7 @@
 /* Copyright 2015 EZchip Semiconductor Ltd. All Rights Reserved.
  *
  * Copyright (c) 2015-2018, Linaro Limited
+ * Copyright (c) 2022, Marvell
  * All rights reserved.
  *
  * SPDX-License-Identifier: BSD-3-Clause
@@ -293,6 +294,40 @@ static uint8_t  g_print_tm_stats   = TRUE;
 
 static void tester_egress_fcn(odp_packet_t odp_pkt);
 
+static uint64_t tm_shaper_min_rate;
+static uint64_t tm_shaper_max_rate;
+static uint32_t tm_shaper_min_burst;
+static uint32_t tm_shaper_max_burst;
+
+static uint64_t
+clamp_rate(uint64_t rate)
+{
+	uint64_t val = MIN(MAX(rate, tm_shaper_min_rate), tm_shaper_max_rate);
+
+	/* PIR can be zero just with CIR valid and vice versa */
+	if (!rate)
+		return 0;
+
+	if (val != rate)
+		printf("INFO: Clamped shaper rate from %" PRIu64 " bps"
+		       " to %" PRIu64 " bps\n", rate, val);
+	return val;
+}
+
+static uint32_t
+clamp_burst(uint32_t burst)
+{
+	uint32_t val = MIN(MAX(burst, tm_shaper_min_burst), tm_shaper_max_burst);
+
+	if (!burst)
+		return 0;
+
+	if (val != burst)
+		printf("INFO: Clamped shaper burst from %" PRIu32 "bits to %" PRIu32 "bits\n",
+		       burst, val);
+	return val;
+}
+
 /* Returns the number of errors encountered. */
 
 static uint32_t create_profile_set(profile_params_set_t *profile_params_set,
@@ -317,10 +352,14 @@ static uint32_t create_profile_set(profile_params_set_t *profile_params_set,
 
 	odp_tm_shaper_params_init(&shaper_params);
 	shaper                          = &profile_params_set->shaper_params;
-	shaper_params.commit_rate       = shaper->commit_rate  * shaper_scale;
-	shaper_params.peak_rate         = shaper->peak_rate    * shaper_scale;
-	shaper_params.commit_burst      = shaper->commit_burst * shaper_scale;
-	shaper_params.peak_burst        = shaper->peak_burst   * shaper_scale;
+	shaper_params.commit_rate       = clamp_rate(shaper->commit_rate *
+						     shaper_scale);
+	shaper_params.peak_rate         = clamp_rate(shaper->peak_rate *
+						     shaper_scale);
+	shaper_params.commit_burst      = clamp_burst(shaper->commit_burst *
+						      shaper_scale);
+	shaper_params.peak_burst        = clamp_burst(shaper->peak_burst *
+						      shaper_scale);
 	shaper_params.dual_rate         = shaper->dual_rate;
 	shaper_params.shaper_len_adjust = shaper->shaper_len_adjust;
 	profile_set->shaper_profile     = odp_tm_shaper_create(name,
@@ -550,6 +589,7 @@ static int create_and_config_tm(void)
 {
 	odp_tm_level_requirements_t *per_level;
 	odp_tm_requirements_t        requirements;
+	odp_tm_capabilities_t        tm_capa;
 	odp_tm_egress_t              egress;
 	uint32_t                     level, err_cnt;
 
@@ -579,6 +619,44 @@ static int create_and_config_tm(void)
 	egress.egress_fcn  = tester_egress_fcn;
 
 	odp_tm_test = odp_tm_create("TM test", &requirements, &egress);
+	if (odp_tm_test == ODP_TM_INVALID) {
+		printf("Error: failed to create TM\n");
+		return -1;
+	}
+
+	if (odp_tm_capability(odp_tm_test, &tm_capa) != 0) {
+		printf("Error: failed to get tm capability");
+		return -1;
+	}
+
+	tm_shaper_min_rate = tm_capa.per_level[0].min_rate;
+	tm_shaper_max_rate = tm_capa.per_level[0].max_rate;
+	tm_shaper_min_burst = tm_capa.per_level[0].min_burst;
+	tm_shaper_max_burst = tm_capa.per_level[0].max_burst;
+
+	for (level = 1; level < tm_capa.max_levels; level++) {
+		odp_tm_level_capabilities_t *per_level =
+			&tm_capa.per_level[level];
+
+		if (per_level->min_rate > tm_shaper_min_rate)
+			tm_shaper_min_rate = per_level->min_rate;
+
+		if (per_level->min_burst > tm_shaper_min_burst)
+			tm_shaper_min_burst = per_level->min_burst;
+
+		if (per_level->max_rate < tm_shaper_max_rate)
+			tm_shaper_max_rate = per_level->max_rate;
+
+		if (per_level->max_burst < tm_shaper_max_burst)
+			tm_shaper_max_burst = per_level->max_burst;
+	}
+
+	if (tm_shaper_min_rate > tm_shaper_max_rate ||
+	    tm_shaper_min_burst > tm_shaper_max_burst) {
+		printf("Error: No shaper rate supported by all TM levels");
+		return -1;
+	}
+
 	err_cnt     = init_profile_sets();
 	if (err_cnt != 0)
 		printf("%s init_profile_sets encountered %" PRIu32 " errors\n",
@@ -873,7 +951,9 @@ int main(int argc, char *argv[])
 	if (process_cmd_line_options(argc, argv) < 0)
 		return -1;
 
-	create_and_config_tm();
+	rc = create_and_config_tm();
+	if (rc != 0)
+		return rc;
 
 	/* Start TM */
 	rc = odp_tm_start(odp_tm_test);
--- a/example/traffic_mgmt/odp_traffic_mgmt.c
+++ b/example/traffic_mgmt/odp_traffic_mgmt.c
@@ -107,7 +107,7 @@ static profile_params_set_t COS0_PROFILE_PARAMS = {
 	.shaper_params = {
 		.commit_rate = 1 * MBPS, .commit_burst      = 100000,
 		.peak_rate   = 4 * MBPS, .peak_burst        = 200000,
-		.dual_rate  = FALSE,     .shaper_len_adjust = 20
+		.dual_rate  = TRUE,      .shaper_len_adjust = 20
 	},
 
 	.threshold_params = {
@@ -149,7 +149,7 @@ static profile_params_set_t COS1_PROFILE_PARAMS = {
 	.shaper_params = {
 		.commit_rate = 500  * KBPS, .commit_burst      = 50000,
 		.peak_rate   = 1500 * KBPS, .peak_burst        = 150000,
-		.dual_rate  = FALSE,        .shaper_len_adjust = 20
+		.dual_rate  = TRUE,         .shaper_len_adjust = 20
 	},
 
 	.threshold_params = {
@@ -191,7 +191,7 @@ static profile_params_set_t COS2_PROFILE_PARAMS = {
 	.shaper_params = {
 		.commit_rate = 200 * KBPS, .commit_burst      = 20000,
 		.peak_rate   = 400 * KBPS, .peak_burst        = 40000,
-		.dual_rate  = FALSE,       .shaper_len_adjust = 20
+		.dual_rate  = TRUE,        .shaper_len_adjust = 20
 	},
 
 	.threshold_params = {
@@ -304,7 +304,6 @@ clamp_rate(uint64_t rate)
 {
 	uint64_t val = MIN(MAX(rate, tm_shaper_min_rate), tm_shaper_max_rate);
 
-	/* PIR can be zero just with CIR valid and vice versa */
 	if (!rate)
 		return 0;
 
--- a/include/odp/api/spec/traffic_mngr.h
+++ b/include/odp/api/spec/traffic_mngr.h
@@ -1,5 +1,6 @@
 /* Copyright (c) 2015-2018, Linaro Limited
  * Copyright (c) 2021, Nokia
+ * Copyright (c) 2022, Marvell
  * All rights reserved.
  *
  * SPDX-License-Identifier: BSD-3-Clause
@@ -277,6 +278,54 @@ typedef struct {
 	 * below is true, in which case it specifies the largest value
 	 * of the weights allowed at this level. */
 	uint32_t max_weight;
+
+	/** Minimum allowed value for odp_tm_shaper_params_t::commit_burst and
+	 * odp_tm_shaper_params_t::peak_burst when
+	 * odp_tm_shaper_params_t::packet_mode is true.
+	 */
+	uint32_t min_burst_packets;
+
+	/** Maximum allowed value fand odp_tm_shaper_params_t::commit_burst and
+	 * odp_tm_shaper_params_t::peak_burst when
+	 * odp_tm_shaper_params_t::packet_mode is true.
+	 */
+	uint32_t max_burst_packets;
+
+	/** Minimum allowed value for odp_tm_shaper_params_t::commit_rate and
+	 * odp_tm_shaper_params_t::peak_rate when
+	 * odp_tm_shaper_params_t::packet_mode is true.
+	 */
+	uint64_t min_rate_packets;
+
+	/** Maximum allowed value for odp_tm_shaper_params_t::commit_rate and
+	 * odp_tm_shaper_params_t::peak_rate when
+	 * odp_tm_shaper_params_t::packet_mode is true.
+	 */
+	uint64_t max_rate_packets;
+
+	/** Minimum allowed value for odp_tm_shaper_params_t::commit_burst and
+	 * odp_tm_shaper_params_t::peak_burst when
+	 * odp_tm_shaper_params_t::packet_mode is false.
+	 */
+	uint32_t min_burst;
+
+	/** Maximum allowed value for odp_tm_shaper_params_t::commit_burst and
+	 * odp_tm_shaper_params_t::peak_burst when
+	 * odp_tm_shaper_params_t::packet_mode is false.
+	 */
+	uint32_t max_burst;
+
+	/** Minimum allowed value for odp_tm_shaper_params_t::commit_rate and
+	 * odp_tm_shaper_params_t::peak_rate when
+	 * odp_tm_shaper_params_t::packet_mode is false.
+	 */
+	uint64_t min_rate;
+
+	/** Maximum allowed value for odp_tm_shaper_params_t::commit_rate and
+	 * odp_tm_shaper_params_t::peak_rate when
+	 * odp_tm_shaper_params_t::packet_mode is false.
+	 */
+	uint64_t max_rate;
 
 	/** tm_node_shaper_supported indicates that the tm_nodes at this level
 	 * all support TM shaping, */
--- a/include/odp/api/spec/traffic_mngr.h
+++ b/include/odp/api/spec/traffic_mngr.h
@@ -1063,6 +1063,7 @@ typedef struct {
 	/** The peak information rate for this shaper profile.  The units for
 	 * this integer is in bits per second when packet_mode is
 	 * not TRUE while in packets per second when packet mode is TRUE.
+	 * This field is ignored when dual_rate is FALSE.
 	 */
 	union {
 		/**< @deprecated Use peak_rate instead */
@@ -1079,7 +1080,9 @@ typedef struct {
 	/** The peak burst tolerance for this shaper profile.  The units for
 	 * this field in bits when packet_mode is not TRUE and packets
 	 * when packet_mode is TRUE. This value sets an upper limit for the
-	 * size of the peakCnt. */
+	 * size of the peakCnt.
+	 * This field is ignored when dual_rate is FALSE.
+	 */
 	uint32_t peak_burst;
 
 	/** The shaper_len_adjust is a value between -128 and 127 which is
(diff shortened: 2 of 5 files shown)


include/odp/api/spec/traffic_mngr.h
```diff
@@ -108,6 +108,18 @@ extern "C" {
  * be set to. It is in units of 1000 bytes/second.
  */
 
+/**
+ * @def ODP_TM_MIN_SHAPER_BURST
```
JannePeltonen, line 112: This makes the limit part of the ABI, which is problematic. I think these kind of things should be in the capabilities so that different implementations of the same ABI can have their own limits.

There are many simiilar defines with the same problem in the traffic manager API. It would be good to move them too to capabilities.

include/odp/api/spec/traffic_mngr.h
```diff
@@ -278,6 +278,24 @@ typedef struct {
...
 	uint32_t max_weight;
 
+	/** min_burst only has significane when the nodes at this level support
+	 * TM shaping, in which case it specifies the smallest value of
+	 * the shaper burst allowed at this level. Unit is in bits.
+	 *
+	 * @see odp_shaper_profile_params_t::commit_burst
+	 * @see odp_shaper_profile_params_t::peak_burst
+	 */
+	uint32_t min_burst;
```
JannePeltonen, line 288: Now that packet mode was added, there probably should be similar capability fields for the max and min burst sized in packets. IOW, the capability in the units of bits does not mean anything in packet mode.

So, maybe there should be separate capability fields for bit and packet mode, e.g. something like this:

```
/** Minimum allowed value for odp_tm_shaper_params_t::commit_burst when
 *  odp_tm_shaper_params_t::packet_mode is true.
 */
uint32_t min_burst_packets;

/** Minimum allowed value for odp_tm_shaper_params_t::commit_burst when
 *  odp_tm_shaper_params_t::packet_mode is false.
 */
```

Note that the type in the comment should be odp_tm_shaper_params_t.

include/odp/api/spec/traffic_mngr.h
```diff
@@ -278,6 +278,54 @@ typedef struct {
...
+	 * odp_tm_shaper_params_t::peak_rate when
+	 * odp_tm_shaper_params_t::packet_mode is false.
+	 */
+	uint64_t min_rate;
+
+	/** Maximum allowed value for odp_tm_shaper_params_t::commit_rate or
+	 * odp_tm_shaper_params_t::peak_rate when
+	 * odp_tm_shaper_params_t::packet_mode is false.
+	 */
+	uint64_t max_rate;
```
JannePeltonen, line 327: min_rate and max_rate duplicate ODP_TM_MIN_SHAPER_BW and ODP_TM_MAX_SHAPER_BW, raising the question which ones should be used by an application. I suppose the old defines could be deprecated or removed altogether.

include/odp/api/spec/traffic_mngr.h
```diff
@@ -278,6 +278,54 @@ typedef struct {
 	 * of the weights allowed at this level. */
 	uint32_t max_weight;
 
+	/** Minimum allowed value for odp_tm_shaper_params_t::commit_burst or
```
JannePeltonen, line 281: I guess in all these descriptions the 'or' should be 'and'.

platform/linux-generic/odp_traffic_mngr.c
```diff
@@ -2607,6 +2607,10 @@ static int tm_capabilities(odp_tm_capabilities_t capabilities[],
 		per_level_cap->max_priority       = ODP_TM_MAX_PRIORITIES - 1;
 		per_level_cap->min_weight         = ODP_TM_MIN_SCHED_WEIGHT;
 		per_level_cap->max_weight         = ODP_TM_MAX_SCHED_WEIGHT;
+		per_level_cap->min_burst          = 1U;
+		per_level_cap->max_burst          = 0xFFFFFFFFU;
```
JannePeltonen, line 2611: I think UINT32_MAX would be better here than the literal which has more F's than I can easily count (which is about 4).

platform/linux-generic/odp_traffic_mngr.c
```diff
@@ -2607,6 +2607,10 @@ static int tm_capabilities(odp_tm_capabilities_t capabilities[],
 		per_level_cap->max_priority       = ODP_TM_MAX_PRIORITIES - 1;
 		per_level_cap->min_weight         = ODP_TM_MIN_SCHED_WEIGHT;
 		per_level_cap->max_weight         = ODP_TM_MAX_SCHED_WEIGHT;
+		per_level_cap->min_burst          = 1U;
```
JannePeltonen, line 2610: I wonder why min burst cannot be zero? Especially since this is a shaper and not policer?

test/validation/api/traffic_mngr/traffic_mngr.c
```diff
@@ -443,6 +443,20 @@ static int test_overall_capabilities(void)
 				CU_ASSERT(per_level->max_priority != 0);
 				return -1;
 			}
+
+			if (per_level->tm_node_shaper_supported) {
+				CU_ASSERT(per_level->max_burst > 0);
+				CU_ASSERT(per_level->min_burst > 0);
```
JannePeltonen, line 449: I put a comment in the wrong place, but this is where I wonder why burst limit of zero is not ok.

test/validation/api/traffic_mngr/traffic_mngr.c
```diff
@@ -443,6 +443,20 @@ static int test_overall_capabilities(void)
 				CU_ASSERT(per_level->max_priority != 0);
 				return -1;
 			}
+
```
JannePeltonen, line 446: Validation test should check capabilities before configuring rates and burst
sizes in traffic_mngr_test_shaper_profile() and in set_shaper().

Also, the traffic manager example app should check the new capabilities before configuring things.

platform/linux-generic/odp_traffic_mngr.c
nithind1988: Made it zero for linux-gen

platform/linux-generic/odp_traffic_mngr.c
```diff
@@ -48,6 +48,9 @@ static const pkt_desc_t EMPTY_PKT_DESC = { .word = 0 };
 #define MAX_PRIORITIES ODP_TM_MAX_PRIORITIES
 #define NUM_SHAPER_COLORS ODP_NUM_SHAPER_COLORS
 
+#define TM_MIN_SHAPER_BW  1
+#define TM_MAX_SHAPER_BW  12500000UL
```
JannePeltonen, line 52: There should be a comment about the units of these, especially if the unit is non-obvious. Now these are in the units of 1000 bytes/s but I am not sure why that is better than bits/s which is the units of the corresponding capabilities and parameters. I think it would be clearer to have these in bits/s (e.g. #define TM_MAX_SHAPER_BW (100ULL * 1000 * 1000 * 1000).

platform/linux-generic/odp_traffic_mngr.c
```diff
@@ -2607,6 +2610,10 @@ static int tm_capabilities(odp_tm_capabilities_t capabilities[],
 		per_level_cap->max_priority       = ODP_TM_MAX_PRIORITIES - 1;
 		per_level_cap->min_weight         = ODP_TM_MIN_SCHED_WEIGHT;
 		per_level_cap->max_weight         = ODP_TM_MAX_SCHED_WEIGHT;
+		per_level_cap->min_burst          = 0;
+		per_level_cap->max_burst          = UINT32_MAX;
+		per_level_cap->min_rate           = TM_MIN_SHAPER_BW * 8U * 1000U;
+		per_level_cap->max_rate           = TM_MAX_SHAPER_BW * 8U * 1000U;
```
JannePeltonen, line 2616: This overflows in a system with 32 bit long ints. ULL suffix should make the literals large enough.

test/validation/api/traffic_mngr/traffic_mngr.c
```diff
@@ -1773,6 +1790,27 @@ static int create_tm_system(void)
...
+		if (per_level->min_burst > tm_shaper_min_burst)
+			tm_shaper_min_burst = per_level->min_burst;
+
+		if (per_level->max_rate < tm_shaper_max_rate)
+			tm_shaper_max_rate = per_level->max_rate;
+
+		if (per_level->max_burst < tm_shaper_max_burst)
+			tm_shaper_max_burst = per_level->max_burst;
+	}
```
JannePeltonen, line 1812: If different levels have non-overlapping allowed ranges of the parameters, after this loop it could be that min_rate > max_rate. I suggest checking for that and giving up the test in that case.

platform/linux-generic/odp_traffic_mngr.c
nithind1988: Ack, I just moved the macros from ODP_TM_MIN_SHAPER_BW to this TM_MIN_SHAPER_BW and the unit of ODP_TM_MIN_SHAPER_BW was in 1000 bytes/sec.

I'll fix the unit to bits/sec

test/validation/api/traffic_mngr/traffic_mngr.c
nithind1988: Ack.

test/validation/api/traffic_mngr/traffic_mngr.c
```diff
@@ -2616,13 +2666,24 @@ static void traffic_mngr_test_wred_profile(void)
...
 		      const uint64_t commit_bps,
-		      const uint64_t commit_burst_in_bits)
+		      const uint64_t commit_burst_in_bits,
+		      bool clamp)
 {
 	odp_tm_shaper_params_t shaper_params;
 	odp_tm_shaper_t        shaper_profile;
 	odp_tm_node_t          tm_node;
+	uint64_t               clamp_bps;
+	uint64_t               clamp_burst_in_bits;
```
JannePeltonen, line 2676: These are unnecessary and cause additional changes. Since the point is to sometimes clamp the values of the pass-by-value parameters, just assign to the parameters, i.e. this is the only thing needed:
```
        if (clamp) {
                commit_bps = clamp_rate(commit_bps);
                commit_burst_in_bits = clamp_burst(commit_burst_in_bits);
        }
```

Actually, making it conditional whether to clamp the values or not is rather dubious. If the values are already in the range (as is the case if/when this gets called by test_shaper_bw() or test_query_functions()), then the claming does not change anything, so it can as well be done. If the values are outside the range, then not clamping them means that unsupported values gets passed to ODP, violating the API spec.

So one starts to wonder why there is the clamp parameter if not clamping the values is not actually ever needed. It turns out that set_shaper() with zero rate and burst parameters are used for disabling the shaper.

The problem (and a pre-existing API spec violation in the validation test) is that set_shaper() just passes the zero values to ODP as a way to disable shaper, even if zero values are outside the valid range and even though ODP API does not specify that as a method of disabling a shaper. 

The proper way to disable a shaper is to call odp_tm_node_shaper_config() or odp_tm_queue_shaper_config() with the shaper profile parameter set to ODP_TM_INVALID.

I propose this bug to be fixed in a separate patch before the validation test changes for the API change at hand. The fix patch can be part of the same patch series in this PR. After the fix, the clamp parameter is no longer needed.

test/validation/api/traffic_mngr/traffic_mngr.c
```diff
@@ -2686,6 +2747,10 @@ static int traffic_mngr_check_shaper(void)
 		return ODP_TEST_INACTIVE;
 	}
 
+	/* This test needs 1 Mbps, 4Mbps, 10 Mpbs, 40Mbps, 100 Mbps */
+	if ((tm_shaper_min_rate > MBPS * 1) || (tm_shaper_max_rate < 100 * MBPS))
```
JannePeltonen, line 2751: I do not think it is good to disable the whole test if only one of the rates falls outside the supported range. The shaper test could just run the subtests that use a supported rate and the whole test could be disabled only if none of the rates are supported.

Space is missing from 4 Mbps and 40 Mbps.

test/validation/api/traffic_mngr/traffic_mngr.c
nithind1988: >These are unnecessary and cause additional changes. Since the point is to sometimes clamp the values of the pass-by-value parameters, just assign to the parameters, i.e. this is the only thing needed:
Actually commit_bps and commit_burst_bps is const. 

>The proper way to disable a shaper is to call odp_tm_node_shaper_config() or odp_tm_queue_shaper_config() with the shaper profile parameter set to ODP_TM_INVALID.


Yes, the main reason was to avoid it for shaper rate zero case. 
I'll change the shaper disable logic.

JannePeltonen: > Actually commit_bps and commit_burst_bps is const.

Indeed, I did not notice. But since they are not pointers, they can simply be changed to non-consts to avoid having to introduce new local variables as a sort-of workaround to change the values that are declared consts.

test/validation/api/traffic_mngr/traffic_mngr.c
JannePeltonen: Space is still missing from 4 Mbps.

nithind1988: Sorry, somehow I missed that.

test/validation/api/traffic_mngr/traffic_mngr.c
```diff
@@ -2371,10 +2426,10 @@ static void traffic_mngr_test_shaper_profile(void)
...
 		snprintf(shaper_name, sizeof(shaper_name),
 			 "shaper_profile_%" PRIu32, idx);
-		shaper_params.commit_rate  = idx * MIN_COMMIT_BW;
-		shaper_params.peak_rate    = idx * MIN_PEAK_BW;
-		shaper_params.commit_burst = idx * MIN_COMMIT_BURST;
-		shaper_params.peak_burst   = idx * MIN_PEAK_BURST;
+		shaper_params.commit_rate  = clamp_rate(idx * MIN_COMMIT_BW);
+		shaper_params.peak_rate    = clamp_rate(idx * MIN_PEAK_BW);
+		shaper_params.commit_burst = clamp_burst(idx * MIN_COMMIT_BURST);
+		shaper_params.peak_burst   = clamp_burst(idx * MIN_PEAK_BURST);
```
JannePeltonen, line 2432: The corresponding clamping is not done in check_shaper_profile(), causing validation tests to fail if the clamping actually did something.

test/validation/api/traffic_mngr/traffic_mngr.c
```diff
@@ -4378,6 +4465,10 @@ static int traffic_mngr_check_query(void)
 	if ((tm_capabilities.tm_queue_query_flags & query_flags) != query_flags)
 		return ODP_TEST_INACTIVE;
 
+	/* This test needs 64Kbps */
+	if (tm_shaper_min_rate > 64 * 1000)
```
JannePeltonen, line 4469: The check below the comment does not check that 64 Kbps (please use space before the Kbps unit) is actually supported, since max rate could be lower than that. I first thought that it is a bug, but it seems to me that the called test function does not really care if the rate is smaller. If so, then the comment should be fixed to match the code. It could be something like this: "This test needs a rate of at most 64 Kbps".

test/validation/api/traffic_mngr/traffic_mngr.c
```diff
@@ -4180,26 +4248,45 @@ static void traffic_mngr_test_tm_create(void)
...
-						"node_1_1_1",
-						3,
-						40  * MBPS)));
-	CU_ASSERT(!odp_cunit_ret(test_shaper_bw("bw100",
-						"node_1_1_2",
-						0,
-						100 * MBPS)));
+	if ((tm_shaper_min_rate <= 1 * MBPS) &&
+	    (tm_shaper_max_rate >= 1 * MBPS)) {
+		CU_ASSERT(!odp_cunit_ret(test_shaper_bw("bw1",
```
JannePeltonen, line 4253: The rate checks is ok, but I wonder how much test_shaper_bw() depends on burst size being 10000 and not some clamped value?

test/validation/api/traffic_mngr/traffic_mngr.c
```diff
@@ -2615,14 +2670,17 @@ static void traffic_mngr_test_wred_profile(void)
...
+		      uint64_t      commit_bps,
+		      uint64_t      commit_burst_in_bits)
 {
 	odp_tm_shaper_params_t shaper_params;
 	odp_tm_shaper_t        shaper_profile;
 	odp_tm_node_t          tm_node;
 	int rc;
 
+	commit_bps = clamp_rate(commit_bps);
+	commit_burst_in_bits = clamp_burst(commit_burst_in_bits);
```
JannePeltonen, line 2682: It seems that some of the call sites of set_shaper() may not be prepared to have the rate and burst parameters getting clamped. I am looking at these:

test_sched_queue_priority()
test_sched_node_priority()
test_sched_wfq()
test_byte_wred()
test_pkt_wred()
test_query_functions()

They all use specific burst sizes and I wonder if it is ok to just change it somewhat arbitrarily without possibly breaking the tests. Also, the rate gets clamped, but the functions later use the unclamped rate when receiving packets. receive_pkts calculates RX loop timeout based on the rate. I think there might be too early timeouts if the rate gets clamped downward too much since then the RX loop does not know to be patient enough.

example/traffic_mgmt/odp_traffic_mgmt.c
```diff
@@ -272,10 +289,14 @@ static uint32_t create_profile_set(profile_params_set_t *profile_params_set,
 
 	odp_tm_shaper_params_init(&shaper_params);
 	shaper                          = &profile_params_set->shaper_params;
-	shaper_params.commit_rate       = shaper->commit_rate  * shaper_scale;
-	shaper_params.peak_rate         = shaper->peak_rate    * shaper_scale;
-	shaper_params.commit_burst      = shaper->commit_burst * shaper_scale;
-	shaper_params.peak_burst        = shaper->peak_burst   * shaper_scale;
+	shaper_params.commit_rate       = clamp_rate(shaper->commit_rate *
```
JannePeltonen, line 292: I wonder if it is good to silently change the values in this example program. Maybe some printout should be done if clamping happens.

example/traffic_mgmt/odp_traffic_mgmt.c
```diff
@@ -534,6 +556,44 @@ static int create_and_config_tm(void)
...
+		if (per_level->max_rate < tm_shaper_max_rate)
+			tm_shaper_max_rate = per_level->max_rate;
+
+		if (per_level->max_burst < tm_shaper_max_burst)
+			tm_shaper_max_burst = per_level->max_burst;
+	}
+
+	if (tm_shaper_min_rate > tm_shaper_max_rate ||
+	    tm_shaper_min_burst > tm_shaper_max_burst) {
+		printf("Error: shaper min and max rates are non overlapping");
```
JannePeltonen, line 593: I think this message is a bit unclear (I would expect max and min rates to not "overlap" (be the same)). Maybe something like this would be better: "Error: No shaper rate supported by all TM levels"

example/traffic_mgmt/odp_traffic_mgmt.c
```diff
@@ -293,6 +293,40 @@ static uint8_t  g_print_tm_stats   = TRUE;
...
+clamp_rate(uint64_t rate)
+{
+	uint64_t val = MIN(MAX(rate, tm_shaper_min_rate), tm_shaper_max_rate);
+
+	/* PIR can be zero just with CIR valid and vice versa */
+	if (!rate)
+		return 0;
+
+	if (val != rate)
+		printf("INFO: Clamped shaper rate from %lu bps to %lu bps\n",
```
JannePeltonen, line 311: The format specifiers in this and other printfs do not match the data types, which causes CI tests to fail. When the argument type is uint32_t, use PRIu32 and when the type is uint64_t, use PRIu64.

test/validation/api/traffic_mngr/traffic_mngr.c
nithind1988: Currently I'm checking if min rate and max rate allows 64Kbps to be configured. Does the comment sound correct now ?

JannePeltonen: The comment does not match the code. If the comment says "needs a rate of at most 64 kbps", then a smaller rate should be ok too. But it isn't, so the earlier comment was more correct. Either remove the comment or change it to "this test uses 64 Kbps rate and a 1000 bit burst size".

example/traffic_mgmt/odp_traffic_mgmt.c
```diff
@@ -293,6 +293,40 @@ static uint8_t  g_print_tm_stats   = TRUE;
...
+static uint32_t tm_shaper_max_burst;
+
+static uint64_t
+clamp_rate(uint64_t rate)
+{
+	uint64_t val = MIN(MAX(rate, tm_shaper_min_rate), tm_shaper_max_rate);
+
+	/* PIR can be zero just with CIR valid and vice versa */
+	if (!rate)
+		return 0;
```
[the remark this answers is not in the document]
JannePeltonen: Since neither this patch nor any preceding patch has changed, then the problem has has been truly resolved. In general we should not have the correctness of a patch depend on later patches. However, since the issue is related to an existing problem and not introduced by this PR, we can live with it temporarily. But still the "vice versa" part of the comment is not true and AFAICS not needed here and can be removed from the comment. I would also propose adding a remark in the commit message for the benefit of whoever might bump into this commit later:

"Clamp shaper profile rate and burst to limits from capability. Continue the practice of using zero peak rate and zero peak burst to indicate single-rate shaping even though it is not strictly compliant with the API."

include/odp/api/spec/traffic_mngr.h
```diff
@@ -1062,6 +1062,7 @@ typedef struct {
 	/** The peak information rate for this shaper profile.  The units for
 	 * this integer is in bits per second when packet_mode is
 	 * not TRUE while in packets per second when packet mode is TRUE.
+	 * This field is ignored when dual_rate is FALSE.
```
JannePeltonen, line 1065: This patch looks good but I would like to improve the commit message. My suggestion:

```
api: tm: ignore peak rate and peak burst when dual rate is false

Clarify that peak rate and peak burst is ignored when dual rate
is set to false, indicating that the single rate is commit rate.
```

test/validation/api/traffic_mngr/traffic_mngr.c
```diff
@@ -2430,6 +2430,7 @@ static void traffic_mngr_test_shaper_profile(void)
 		shaper_params.peak_rate    = clamp_rate(idx * MIN_PEAK_BW);
 		shaper_params.commit_burst = clamp_burst(idx * MIN_COMMIT_BURST);
 		shaper_params.peak_burst   = clamp_burst(idx * MIN_PEAK_BURST);
+		shaper_params.dual_rate    = true;
```
JannePeltonen, line 2433: There is no need to add this line here. The dual_rate field is already set before this loop. The existing assignment
should simply be modified to set the field to true instead of false.

Then I would like to improve the commit message of this commit too. e.g.:
```
validation: tm: set dual_rate flag in shaper profile test

Set dual_rate to true in the shaper profile test which checks
that shaper profile can be created with the desired parameters,
including peak_rate and peak_burst.
```

JannePeltonen:
I made a few rather minor comments. The dual_rate API changes can be split to another PR if you want, but I think they are close to ready and can stay in this PR too. The linux-gen implementation patch for checking if dual_rate is false is the biggest thing missing but it should be quite a simple addition.

example/traffic_mgmt/odp_traffic_mgmt.c
```diff
@@ -106,7 +106,7 @@ static profile_params_set_t COS0_PROFILE_PARAMS = {
 	.shaper_params = {
 		.commit_rate = 1 * MBPS, .commit_burst      = 100000,
 		.peak_rate   = 4 * MBPS, .peak_burst        = 200000,
-		.dual_rate  = FALSE,     .shaper_len_adjust = 20
+		.dual_rate  = TRUE,      .shaper_len_adjust = 20
```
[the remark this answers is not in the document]
nithind1988: I removed the comment and left the check to avoid the clamp logs.

nithind1988:
>I made a few rather minor comments. The dual_rate API changes can be split to another PR if you want, but I think they are close to ready and can stay in this PR too. The linux-gen implementation patch for checking if dual_rate is false is the biggest thing missing but it should be quite a simple addition.

Updated linux gen implementation

JannePeltonen:
Maybe you would like to also update the copyright statements in the relevant files.

psavol:
Nithin, we are waiting this PR to finish. So, that we can starts merging the new API version.

MatiasElo:
Hi @nithind1988, we've started merging API changes for the next release and we'd like to have this PR included. Please check the final comments from Janne.

JannePeltonen (approved):

PR #1342, opened 2021-07-14
[PATCH v2] example: tm: add support for pktio loop and pkt prio mode overwrite

nithind1988:
Add support to use pktio loop when egress function is not available.
Also support TM systems which only support pkt prio mode overwrite.


JannePeltonen:
Is this PR still relevant? If yes, it should be rebased and I think the two changes (loop pktio and prio mode overwrite) should be separated into two patches.

JannePeltonen:

PR #1345, opened 2021-07-21
[PATCH v3] api: packet: move odp_packet_proto_stats_opt_t definition

MatiasElo:
Packet type definitions belong to packet_types.h header.

V2:
- Added separate header for proto_stats  type definitions

### DIFF
--- a/include/Makefile.am
+++ b/include/Makefile.am
@@ -38,6 +38,7 @@ odpapiinclude_HEADERS = \
 	odp/api/protocols.h \
 	odp/api/pool.h \
 	odp/api/proto_stats.h \
+	odp/api/proto_stats_types.h \
 	odp/api/queue.h \
 	odp/api/queue_types.h \
 	odp/api/random.h \
@@ -94,6 +95,7 @@ odpapispecinclude_HEADERS = \
 		  odp/api/spec/pool.h \
 		  odp/api/spec/pool_types.h \
 		  odp/api/spec/proto_stats.h \
+		  odp/api/spec/proto_stats_types.h \
 		  odp/api/spec/queue.h \
 		  odp/api/spec/queue_types.h \
 		  odp/api/spec/random.h \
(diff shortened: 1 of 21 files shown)


JannePeltonen:
Just recording here the offline discussion that this patch makes packet_types.h include odp_proto_stats.h which does not contain only type definitions. Ideally odp_proto_stats.h header should be split to two headers (one for type definitions and one for other things like function declarations) as has recently been done for other headers. That would allow ODP implementations to e.g. define ODP API functions as inline functions. We concluded that we can postpone such splitting of odp_proto_stats.h until the need arises.

JannePeltonen (approved):

nithind1988:
> 
> 
> Just recording here the offline discussion that this patch makes packet_types.h include odp_proto_stats.h which does not contain only type definitions. Ideally odp_proto_stats.h header should be split to two headers (one for type definitions and one for other things like function declarations) as has recently been done for other headers. That would allow ODP implementations to e.g. define ODP API functions as inline functions. We concluded that we can postpone such splitting of odp_proto_stats.h until the need arises.

Missed this as the splitting types and function declarations came later.

But even with this scheme, there shouldn't an issue in defining the functions for proto stats inline right ?

JannePeltonen:
> But even with this scheme, there shouldn't an issue in defining the functions for proto stats inline right ?

I think not for odp_packet_proto_stats_request() and odp_packet_proto_stats(), but I guess there could be a problem for odp_proto_stats() and other functions declared in proto_stats.h. Maybe @MatiasElo can comment better.

MatiasElo:
> > But even with this scheme, there shouldn't an issue in defining the functions for proto stats inline right ?
> 
> I think not for odp_packet_proto_stats_request() and odp_packet_proto_stats(), but I guess there could be a problem for odp_proto_stats() and other functions declared in proto_stats.h. Maybe @MatiasElo can comment better.

Yes, the functions which use types defined in `proto_stats.h` are problematic for inlining. `odp_proto_stats()` is a good example. I can take a look how much work it would take to split the header. Or how critical is function inlining for you @nithind1988 at this point?

nithind1988:
Thanks for the changes. For us, atleast odp_packet_proto_stats_request() needs to be inline as it is the only API used in fast path.

JannePeltonen:

PR #1347, opened 2021-07-21
[PATCH v4] test: l2fwd: use max flow id to configure scheduler

shijith:
Multiple flows are disabled in default scheduler configuration. Using
max supported flows will improve the throughput of l2fwd.

Signed-off-by: Shijith Thotton <sthotton@marvell.com>

### DIFF
--- a/test/performance/odp_l2fwd.c
+++ b/test/performance/odp_l2fwd.c
@@ -1,6 +1,6 @@
 /* Copyright (c) 2014-2018, Linaro Limited
  * Copyright (c) 2019-2021, Nokia
- * Copyright (c) 2020, Marvell
+ * Copyright (c) 2020-2021, Marvell
  * All rights reserved.
  *
  * SPDX-License-Identifier:     BSD-3-Clause
@@ -108,6 +108,7 @@ typedef struct {
 	uint32_t packet_len;	/* Maximum packet length supported */
 	uint32_t seg_len;	/* Pool segment length */
 	int promisc_mode;       /* Promiscuous mode enabled */
+	int flow_aware;         /* Flow aware scheduling enabled */
 	int mtu;                /* Interface MTU */
 } appl_args_t;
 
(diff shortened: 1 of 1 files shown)


test/performance/odp_l2fwd.c
```diff
@@ -2210,7 +2212,13 @@ int main(int argc, char *argv[])
...
 	bind_workers();
 
-	odp_schedule_config(NULL);
+	odp_schedule_config_init(&sched_config);
+	if (odp_schedule_capability(&sched_capa)) {
+		ODPH_ERR("Error: schedule capability failed\n");
+		exit(EXIT_FAILURE);
+	}
+	sched_config.max_flow_id = sched_capa.max_flow_id;
+	odp_schedule_config(&sched_config);
```
MatiasElo, line 2221: A user may also be interested about performance without flow aware scheduling. This change would also break a rather long performance measurement record we have with this application. So, it would be better to put this behind a configuration option (e.g. `-f, --flow_aware`) and keep the default value as disabled.

MatiasElo (changes requested):

test/performance/odp_l2fwd.c
shijith: Sure. I will post the next version with the change.

test/performance/odp_l2fwd.c
```diff
@@ -108,6 +108,7 @@ typedef struct {
 	uint32_t packet_len;	/* Maximum packet length supported */
 	uint32_t seg_len;	/* Pool segment length */
 	int promisc_mode;       /* Promiscuous mode enabled */
+	int flow_aware;         /* Flow aware scheduling enabled */
```
MatiasElo, line 111: Marvell copyright year can be updated for this file.

test/performance/odp_l2fwd.c
```diff
@@ -2210,7 +2221,16 @@ int main(int argc, char *argv[])
...
 
-	odp_schedule_config(NULL);
+	odp_schedule_config_init(&sched_config);
+	if (odp_schedule_capability(&sched_capa)) {
+		ODPH_ERR("Error: schedule capability failed\n");
+		exit(EXIT_FAILURE);
+	}
+
+	if (gbl_args->appl.flow_aware)
+		sched_config.max_flow_id = sched_capa.max_flow_id;
```
MatiasElo, line 2231: Perhaps it would be good to check here if flow aware mode is supported and throw an error + exit if not. Currently, a user has no way of knowing if the platform actually supports flow aware mode.

MatiasElo (commented):
Two minor comments but otherwise looking good. Please add my reviewed-by tag in the next version.

test/performance/odp_l2fwd.c
shijith: Okay.

test/performance/odp_l2fwd.c
shijith: I will add it in the next version.

MatiasElo (approved):

PR #1349, opened 2021-07-22
[PATCH v3] api: ipsec: provide original ESP/AH packet length in IPsec packet result

JannePeltonen:
Provide the original ESP/AH packet length before inbound IPsec processing
in odp_ipsec_packet_result_t. An application may use it for statistics
purposes.

This can be useful since post-IPsec reassemly offload makes it difficult
to deduce the original octet count that went into the creation of the
result packet. Even without reassembly offload, possible extra padding
cannot be deduced from the result packet.

### DIFF
--- a/test/validation/api/ipsec/ipsec.c
+++ b/test/validation/api/ipsec/ipsec.c
@@ -878,6 +878,11 @@ static void verify_in(const ipsec_test_part *part,
 			if (ODP_IPSEC_SA_INVALID != sa)
 				CU_ASSERT_EQUAL(IPSEC_SA_CTX,
 						odp_ipsec_sa_context(sa));
+			if (suite_context.inbound_op_mode != ODP_IPSEC_OP_MODE_SYNC) {
+				uint32_t len = part->pkt_in->len - part->pkt_in->l3_offset;
+
+				CU_ASSERT(result.orig_ip_len == len);
+			}
 		}
 		ipsec_check_packet(part->out[i].pkt_res,
 				   pkto[i],
--- a/include/odp/api/spec/ipsec.h
+++ b/include/odp/api/spec/ipsec.h
@@ -1532,6 +1532,17 @@ typedef struct odp_ipsec_packet_result_t {
 		uint32_t len;
 	} outer_hdr;
 
+	/** Total IP length of the original ESP or AH packet before IPsec
+	 *  decapsulation. This is valid only for inbound inline and async
+	 *  processed packets. Zero value means that the length information
+	 *  is not available.
+	 *
+	 *  If the result packet was reassembled from multiple IPsec
+	 *  protected packets, this is the sum of the lengths of all the
+	 *  involved IPsec packets.
+	 */
+	uint32_t orig_ip_len;
+
 } odp_ipsec_packet_result_t;
 
 /**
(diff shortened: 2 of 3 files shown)


anoobj:
@JannePeltonen This proposal looks good to me.

JannePeltonen:
v2: Added implementation and validation test, rebased.

psavol (approved):

test/validation/api/ipsec/ipsec.c
```diff
@@ -878,6 +878,11 @@ static void verify_in(const ipsec_test_part *part,
 			if (ODP_IPSEC_SA_INVALID != sa)
 				CU_ASSERT_EQUAL(IPSEC_SA_CTX,
 						odp_ipsec_sa_context(sa));
+			if (suite_context.inbound_op_mode != ODP_IPSEC_OP_MODE_SYNC) {
+				uint32_t len = part->pkt_in->len - part->pkt_in->l3_offset;
+
+				CU_ASSERT(result.orig_ip_len == len);
```
anoobj, line 884: Wouldn't this ASSERT make result.orig_ip_len mandatory?

API text makes it optional.

> Zero value means that the length information is not available.

anoobj: @JannePeltonen There is one more issue here. In case of reassembly test cases, part->pkt_in would be just one fragment (in fact last fragment). So this would be comparing IP packet length of one fragment against IP packet length of all fragments, which is not the intention,

There is no quick way of fixing this. I have added a fix, but that would involve a new member ``ipsec_test_part.orig_ip_len`` and I would populate this field in case of reassembly tests.

And the change in verify_in would be like,
`+                               uint32_t len;                                                                                                                                                                                                                                                                                                                                           +                               if (part->orig_ip_len)                                                                                                                                                             +                                       len = part->orig_ip_len;                                                                                                                                                   +                               else                                                                                                                                                                               +                                       len = part->pkt_in->len - part->pkt_in->l3_offset; `

Another option is to set part->orig_ip_len in all cases, which would lead to lot of code duplication.

I can push a patch if you agree that we have a problem and you agree with either solution.

Apologies that I didn't review this patch earlier.

anoobj: > The new orig_ip_len field of test_part could have some comment that says that it is nonzero only if the expected len differs from that of the input test packet.

Agreed. And as I worked on it further, it looked like making orig_ip_len as part of ipsec_test_part.out[i] seemed like a better approach. For incomplete frags cases, there can be more than 1 frag and single value won't work for all frags. Shall I propose so?

> I would think that the value of the expected orig_ip_len could be calculated on the fly in test_multi_out_in() from the test fragments so that it would not need to be hardcoded in the reassembly test cases.

That is the intention. We are on the same page.

JannePeltonen: > it looked like making orig_ip_len as part of ipsec_test_part.out[i] seemed like a better approach

That sounds right, even though the test code does not appear to be ready to handle more than one fragment in the incomplete case (either that or I got confused reading the code). But that is another issue and not a problem until such a test case gets added.

anoobj: Agreed. I'll push a PR with the discussed changes.

PR #1253, opened 2021-04-20
[PATCH v49] linux-gen: crypto: Integration of ARMv8 crypto library

m-govind:
Implementation of AES-CBC using ARMv8 crypto API.

Signed-off-by: Govindarajan Mohandoss <govindarajan.mohandoss@arm.com>

PR #1354, opened 2021-08-06
[PATCH v10] crypto: add AES-CBC + HMAC-SHA into ARM crypto

Malvika-Gupta:
Armv8 crypto library implementation supports AES-CBC+HMAC-SHA1 and AES-CBC+HMAC-SHA256 chained operations. This PR includes all the changes required in the ODP crypto and ODP IPSec applications to support these operations.

Note:
The first commit of this PR is identical to the first commit of the AES-GCM PR #1253. It is expected that the PR #1253 will be merged before this PR and once that happens, this commit will be dropped.


DEPENDENCIES
```diff
@@ -98,11 +98,36 @@ Prerequisites for building the OpenDataPlane (ODP) API
      --with-openssl-path=/home/${USER}/src/install-openssl-aarch64
    $ make
```
psavol, line 100:
```
Signed-off-by: Govindarajan <govindarajan.mohandoss@arm.com>
               Malvika Gupta <Malvika.Gupta@arm.com>
               Matias Elo <matias.elo@nokia.com>
```

There needs to be "Signed-off-by:" (or reviewed-by, etc) before each name in commit log message.

JannePeltonen:
The proposed API changes do not make much sense to me. I do not think the API should be changed at all.

The patches introduce crypto and IPsec capability flags that tell if "chained mode" is supported, but do not define what "chained mode" means.

In the context of crypto, "chained mode" can make one think of the cipher block chaining (CBC) mode and its variant, but that is not what the flag is really about. It appears that the chained mode flag tells if both ciphering and authentication can be done by single ODP crypto operation. But that is already supported by the ODP crypto API which can do either ciphering or authentication or both within a single operation. I also think that it is reasonable and desirable to expect all ODP crypto implementations to be able to support the current API.

Not setting the "chained mode" flag in the OpenSSL based crypto implementation does not make any sense to me either.

In the context of IPsec the "chained mode" flag is even harder to understand. All IPsec implementations have to be able to support ciphering and authentication within the same SA. And it is unclear what the term "chained mode" even means in IPsec (one can think about SA chaining, which is a totally different thing).

m-govind:
> The proposed API changes do not make much sense to me. I do not think the API should be changed at all.
> 
> The patches introduce crypto and IPsec capability flags that tell if "chained mode" is supported, but do not define what "chained mode" means.
> 
> In the context of crypto, "chained mode" can make one think of the cipher block chaining (CBC) mode and its variant, but that is not what the flag is really about. It appears that the chained mode flag tells if both ciphering and authentication can be done by single ODP crypto operation. But that is already supported by the ODP crypto API which can do either ciphering or authentication or both within a single operation. I also think that it is reasonable and desirable to expect all ODP crypto implementations to be able to support the current API.
> 
> Not setting the "chained mode" flag in the OpenSSL based crypto implementation does not make any sense to me either.
> 
> In the context of IPsec the "chained mode" flag is even harder to understand. All IPsec implementations have to be able to support ciphering and authentication within the same SA. And it is unclear what the term "chained mode" even means in IPsec (one can think about SA chaining, which is a totally different thing).

Hi Janne,
  I agree that the flag "chained mode" is a misnomer. Either we have to come up with a new name or define a new API. 
  I would like to brief on the problem that we want to solve. ARMv8 crypto library supports "AES-CBC + HMAC_SHAx" 
  together only. Application cannot do AES-CBC only or HMAC_SHAx only operations using ARMv8 crypto. Whereas with
  Open SSL, it is possible to do any cipher only or auth only operations. 
  ODP Crypto test suite validates cipher only and auth only cases with NIST test vector as golden reference to validate
  the output. It is not possible to run these test cases with ARMv8 crypto for AES-CBC/HMAC_SHAx algos because there is
  no golden reference available from NIST test vector for combined cipher and auth operations. We need a flag to block
  these test cases when ARMv8 crypto is used. Also, we can use the same flag to add new test case to validate
  ARMv8 supported "AES-CBC + HMAC_SHAx" case. Open SSL output can be used as a golden reference to validate the
  ARMv8 crypto output for "AES-CBC + HMAC_SHAx" case.
  If this is not acceptable, please suggest a way to integrate it.

Thanks
Govind

m-govind:
> So this PR is not about adding a capability field that tells whether chained operations are supported (since they are supported in all implementations, also in the ARM crypto lib). Therefore the code that clears the "chaining supported" flag for OpenSSL and prevents chaining tests to be run with OpenSSL are wrong.
> 
> What you really seem to want is a capability field that tells if a cipher or auth algorithm can be used alone, in a non-chained fashion. This would enable crypto implementations that would e.g. support SHA-1 but only together with a non-null cipher and never alone. Such an API seems weird to me and makes me wonder if the crypto implementation (which consists of the library and the ODP specific glue code) should be improved to match ODP API instead of vice versa.
> 
> > ARMv8 crypto library supports "AES-CBC + HMAC_SHAx" together only.
> 
> It seems to me that you could easily support AES-CBC without authentication by simply ignoring the digest. It also seems that you could do SHAx without cipher by doing decryption and throwing away the decrypted data and just keeping the digest. This way no API changes would be needed.

We discussed this topic already in "https://github.com/OpenDataPlane/odp/pull/1253" and came to similar conclusion.
We tried that option to make ARM crypto work the same way as Open SSL and ended up facing other issues. Making
changes in ARM crypto library is not feasible as of now. We have to try different options to accept ARM crypto
library as is.

Note:
As per our Open SSL team, AES-CBC/HMAC_SHAx related optimizations are already done in Open SSL library. 
Once we break the chaining part from ARM crypto library, it will be same as Open SSL.
In general, ARM crypto library can be used as a reference code, to improve the performance of IPSec use case by doing
the algo work for AES-CBC and HMAC_SHAx together, using ARM crypto assembly instructions with supported block alignments.

psavol:
There has been some crypto API and test changes in master, so this PR needs a rebase.

Malvika-Gupta:
Hi @JannePeltonen, I have rebased the patches and they are now ready for review. Thanks!

platform/linux-generic/arch/aarch64/odp_crypto_armv8.c
```diff
@@ -449,6 +476,261 @@ static int process_aes_gcm_param(odp_crypto_generic_session_t *session)
...
+	uint32_t in_pos = param->cipher_range.offset;
+	uint32_t hash_pos = param->auth_range.offset;
+	uint8_t key_expanded[256] = {0};
+	uint8_t hash_buf[hash_len];
+	uint8_t digest[ARM_CRYPTO_MAX_DIGEST_LENGTH];
+	int rc;
+
+	/* Fail early if cipher_range is too large */
+	if (odp_unlikely(in_len > ARM_CRYPTO_MAX_DATA_LENGTH)) {
+		ODP_DBG("ARM Crypto: Packet size too large for requested operation\n");
```
JannePeltonen, line 497: It is not the packet size that is too large but the cipher range, as also mentioned in the comment. The "for requested operation" part of the message seems unnecessary. Maybe have it simply: "Cipher range too large".

platform/linux-generic/arch/aarch64/odp_crypto_armv8.c
```diff
@@ -449,6 +476,261 @@ static int process_aes_gcm_param(odp_crypto_generic_session_t *session)
...
+	else if (session->p.cipher_iv.data)
+		iv_ptr = session->cipher.iv_data;
+	else
+		goto err;
+#else
+	iv_ptr = param->cipher_iv_ptr;
+	ODP_ASSERT(session->p.cipher_iv_len == 0 || iv_ptr != NULL);
+#endif
+
+	armv8_expandkeys_enc_aes_cbc_128(key_expanded, session->cipher.key_data);
```
JannePeltonen, line 513: Why are you doing the key expansion separately for every packet instead of doing it only once per session?

platform/linux-generic/arch/aarch64/odp_crypto_armv8.c
```diff
@@ -449,6 +476,261 @@ static int process_aes_gcm_param(odp_crypto_generic_session_t *session)
...
+	arg.cipher.key = key_expanded;
+	arg.cipher.iv = iv_ptr;
+
+	arg.digest.hmac.key = session->auth.key;
+	arg.digest.hmac.i_key_pad = session->auth.i_key_pad;
+	arg.digest.hmac.o_key_pad = session->auth.o_key_pad;
+
+	uint32_t seg_len = 0;
+	uint8_t *data = odp_packet_offset(pkt, in_pos, &seg_len, NULL);
```
JannePeltonen, line 522: Since seg_len is not used, it could be removed (there is the same issue in existing code).

platform/linux-generic/arch/aarch64/odp_crypto_armv8.c
```diff
@@ -449,6 +476,261 @@ static int process_aes_gcm_param(odp_crypto_generic_session_t *session)
...
+		odp_packet_copy_to_mem(pkt, in_pos, in_len, local.buffer);
+
+		data = local.buffer;
+	}
+
+	seg_len = 0;
+	uint8_t *hash = odp_packet_offset(pkt, hash_pos, &seg_len, NULL);
+
+	if (odp_unlikely(odp_packet_is_segmented(pkt)) ||
+	    odp_unlikely(odp_packet_tailroom(pkt) < OOB_WRITE_LEN)) {
```
JannePeltonen, line 538: This just repeats the same exact tests that were done a few lines earlier. The bodies of the two if-statements should be combined.

platform/linux-generic/arch/aarch64/odp_crypto_armv8.c
```diff
@@ -449,6 +476,261 @@ static int process_aes_gcm_param(odp_crypto_generic_session_t *session)
...
+		break;
+	case ODP_AUTH_ALG_SHA256_HMAC:
+		rc = armv8_enc_aes_cbc_sha256_128(data, local.buffer, in_len,
+						  hash, digest, hash_len,
+						  &arg);
+
+		if (odp_unlikely(rc)) {
+			ODP_DBG("ARM Crypto: AES-CBC Encoding and HMAC-SHA256 "
+				"digest generation failed\n");
+			goto err;
```
JannePeltonen, line 570: Now the essentially same error checking code and printing is repeated in all the switch cases. The common code could be moved out.

platform/linux-generic/arch/aarch64/odp_crypto_armv8.c
```diff
@@ -449,6 +476,261 @@ static int process_aes_gcm_param(odp_crypto_generic_session_t *session)
...
+		/* Packet is segmented or it may not be safe to read and write
+		 * beyond the end of packet data. Copy the hash range to a
+		 * contiguous buffer. */
+		odp_packet_copy_to_mem(pkt, hash_pos, hash_len, hash_buf);
+
+		hash = hash_buf;
+	}
+
+	if (odp_unlikely(session->p.hash_result_in_auth_range))
+		_odp_packet_set_data(pkt, param->hash_result_offset, 0, session->p.auth_digest_len);
```
JannePeltonen, line 548: At this point the auth range of packet data may already have been copied into a temporary buffer, in which case the zeroing here does not have any effect. So this code zeroes the hash location or does not zero it, depending on outside factors.

Actually, the hash location should not be zeroed in encode sessions (see the API spec) so this code´should just be removed.

In decode sessions the zeroing should be done, but the code there has the same problem that the packet data may already have been copied into a temporary buffer.

platform/linux-generic/arch/aarch64/odp_crypto_armv8.c
```diff
@@ -449,6 +476,261 @@ static int process_aes_gcm_param(odp_crypto_generic_session_t *session)
...
+
+	switch (session->p.auth_alg) {
+	case ODP_AUTH_ALG_SHA1_HMAC:
+		rc = armv8_enc_aes_cbc_sha1_128(data, local.buffer, in_len,
```
JannePeltonen, line 552: The output buffer for the cipher is always 'local.buffer' which may be separate from 'data'. i.e. the code is not doing in-place processing even when it would be possible.

platform/linux-generic/arch/aarch64/odp_crypto_armv8.c
```diff
@@ -449,6 +476,261 @@ static int process_aes_gcm_param(odp_crypto_generic_session_t *session)
...
+#if ODP_DEPRECATED_API
+	if (param->cipher_iv_ptr)
+		iv_ptr = param->cipher_iv_ptr;
+	else if (session->p.cipher_iv.data)
+		iv_ptr = session->cipher.iv_data;
+	else
+		goto err;
+#else
+	iv_ptr = param->cipher_iv_ptr;
+	ODP_ASSERT(session->p.cipher_iv_len == 0 || iv_ptr != NULL);
```
JannePeltonen, line 510: cipher_iv_len should never be zero here (same issue in existing code).

platform/linux-generic/arch/aarch64/odp_crypto_armv8.c
```diff
@@ -449,6 +476,261 @@ static int process_aes_gcm_param(odp_crypto_generic_session_t *session)
...
+
+	switch (session->p.auth_alg) {
+	case ODP_AUTH_ALG_SHA1_HMAC:
+		rc = armv8_dec_aes_cbc_sha1_128(data, local.buffer, in_len,
+						hash, digest, hash_len,
+						&arg);
+
+		if (odp_unlikely(rc)) {
+			ODP_DBG("ARM Crypto: AES-CBC Decoding and HMAC-SHA1 "
+				"digest generation failed\n");
```
JannePeltonen, line 669: Why does the error message say "digest _generation_ failed"? Is the function simply generating the HMAC, not checking it? If that was the case, then HMAC would not checked at all, making the whole SHA computation pointless.

Or maybe the function actually performs the checking and just the error message is misleading? Which way is it?

platform/linux-generic/arch/aarch64/odp_crypto_armv8.c
```diff
@@ -631,13 +956,29 @@ odp_crypto_session_create(const odp_crypto_session_param_t *param,
 		}
 		break;
 	}
+	case ODP_CIPHER_ALG_AES_CBC:
+		/* Only AES-CBC and HMAC-SHA chained operation is supported */
+		if (param->auth_alg != ODP_AUTH_ALG_SHA1_HMAC &&
+		    param->auth_alg != ODP_AUTH_ALG_SHA256_HMAC) {
+			rc = -2;
```
JannePeltonen, line 963: It is not clear to my why this needs to be the case. Maybe you could do an individual operation by just dropping the part of the result that you do not need?

The comment could be more helpful in explaining why individual operations are not supported. I can see from the code easily that only "chained" operation is supported, so the current comment does not add much. It would be much more useful to know _why_ cipher or auth alone cannot be supported.

platform/linux-generic/arch/aarch64/odp_crypto_armv8.c
```diff
@@ -449,6 +476,261 @@ static int process_aes_gcm_param(odp_crypto_generic_session_t *session)
...
+}
+
+static int process_cipher_auth_hmac_param(odp_crypto_generic_session_t *session)
+{
+	/* Verify cipher Key len is valid */
+	if (16 != session->p.cipher_key.length)
+		return -1;
+
+	/* Verify cipher IV len is correct */
+	if (ARM_CRYPTO_MAX_IV_LENGTH != session->p.cipher_iv_len)
```
JannePeltonen, line 706: If someone adds other algorithms and changes the MAX_IV_LENGTH, then this code breaks. The parameter should not be compared against maximum IV length but the IV length actually used by this algorithm.

platform/linux-generic/arch/aarch64/odp_crypto_armv8.c
```diff
@@ -449,6 +476,261 @@ static int process_aes_gcm_param(odp_crypto_generic_session_t *session)
...
+
+	/* Verify auth IV len is correct */
+	if (0 != session->p.auth_iv_len)
+		return -1;
+
+	/* Zero memory under key */
+	if (session->p.auth_alg == ODP_AUTH_ALG_SHA1_HMAC)
+		memset(session->auth.key, 0, sizeof(session->auth.key));
+	else if (session->p.auth_alg == ODP_AUTH_ALG_SHA256_HMAC)
+		memset(session->auth.key, 0, sizeof(session->auth.key));
```
JannePeltonen, line 717: Are these memsets needed?

platform/linux-generic/arch/aarch64/odp_crypto_armv8.c
```diff
@@ -449,6 +476,261 @@ static int process_aes_gcm_param(odp_crypto_generic_session_t *session)
...
+
+	/* Convert keys */
```
JannePeltonen, line 719: This is not converting anything but just copying the keys.

Actually, I think the comments in this function do not add much value and could be just removed.

platform/linux-generic/arch/aarch64/odp_crypto_armv8.c
```diff
@@ -449,6 +476,261 @@ static int process_aes_gcm_param(odp_crypto_generic_session_t *session)
...
+	memcpy(session->cipher.key_data, session->p.cipher_key.data,
+	       session->p.cipher_key.length);
+	memcpy(session->auth.key, session->p.auth_key.data,
+	       session->p.auth_key.length);
+
+	/* Set function */
+	if (ODP_CRYPTO_OP_ENCODE == session->p.op)
+		session->func = aes_cbc_encrypt_hmac_gen;
```
JannePeltonen, line 727: This fails to take the auth_cipher_text session parameter into account and may end up performing MAC calculation and ciphering in a wrong order.

Session creation should fail if processing order that is not supported is requested. But if the underlying library can perform ciphering and authentication alone, then it should be possible to support both orders (by invoking the algorithms separately).

platform/linux-generic/arch/aarch64/odp_crypto_armv8.c
```diff
@@ -658,13 +999,78 @@ odp_crypto_session_create(const odp_crypto_session_param_t *param,
...
 		}
 		break;
+	case ODP_AUTH_ALG_SHA1_HMAC:
+		/* Only AES-CBC and HMAC-SHA chained operation is supported */
+		if (param->cipher_alg != ODP_CIPHER_ALG_AES_CBC) {
+			rc = -2;
+			break;
+		}
+
+		if (param->cipher_alg == ODP_CIPHER_ALG_AES_CBC) {
```
JannePeltonen, line 1009: This condition is always true at this point, so testing it is unnecessary. The code in the body could be put in a function to remove clutter from the session creation function and to allow reuse between SHA-1 and SHA-2.

test/validation/api/crypto/odp_crypto_test_inp.c
```diff
@@ -527,6 +527,17 @@ static void alg_test_execute(const alg_test_param_t *param)
 		     param->op == ODP_CRYPTO_OP_ENCODE))
 			continue;
 
+		/*
+		 * Checking against wrong digest is meaningless for AES-CBC +
+		 * HMAC-SHA chained operation. As per Armv8 crypto library
+		 * specification, a new digest is generated for decryption too.
+		 */
```
JannePeltonen, line 534: > "Checking against wrong digest is meaningless for AES-CBC + HMAC-SHA
> chained operation."

Why would it be meaningless? There would be no point in doing SHA if the HMAC was not checked. And the validation test should check that an ODP implementation notices an incorrect MAC.

> As per Armv8 crypto library specification, a new digest is generated
> for decryption too.

I could not find such specification (can you provide a pointer?), but that is irrelevant since this code is ODP API validation test that does not have anything directly to do with the ARM crypto library. An ODP implementation must adhere to the ODP API spec and hide possible peculiarities of the underlying crypto library used in the implementation.

include/odp/api/abi-default/ipsec.h
```diff
@@ -29,6 +29,7 @@ typedef struct { char dummy; /**< @internal Dummy */ } _odp_abi_ipsec_sa_t;
 typedef _odp_abi_ipsec_sa_t *odp_ipsec_sa_t;
 
 #define ODP_IPSEC_SA_INVALID ((odp_ipsec_sa_t)0)
+#define ODP_IPSEC_SA_INVALID_COMBO ((odp_ipsec_sa_t)-1)
```
JannePeltonen, line 32: This is an ODP API change that needs to be done also in the API spec headers, not only in the ABI files. And the commit title should be prefixed with "api: crypto:" for such changes.

platform/linux-generic/odp_ipsec_sad.c
```diff
@@ -745,8 +745,18 @@ odp_ipsec_sa_t odp_ipsec_sa_create(const odp_ipsec_sa_param_t *param)
 	}
 
 	if (odp_crypto_session_create(&crypto_param, &ipsec_sa->session,
-				      &ses_create_rc))
+				      &ses_create_rc)) {
```
JannePeltonen, line 748: This changes linux-gen implementation too even though the commit message claims this is a validation test change. It would anyway be better to separate validation test changes from implementation changes, when possible.

Malvika-Gupta:
Abandoning this PR to make way for integrating IPSec Mbuf library later

PR #1355, opened 2021-08-11
helper: threads: parameter initialization functions and new thread parameters

JereLeppanen:
This series adds new helper API functions odph_thread_common_param_init() and odph_thread_param_init(), and two new thread creation parameters.


helper/threads.c
```diff
@@ -390,10 +431,10 @@ int odph_odpthreads_create(odph_odpthread_t *thread_tbl,
 		start_args->instance   = thr_params->instance;
 
 		if (helper_options.mem_model == ODP_MEM_MODEL_THREAD) {
-			if (create_pthread(&thread_tbl[i], cpu))
+			if (create_pthread(&thread_tbl[i], cpu, start_args->thr_params.stack_size))
 				break;
 		 } else {
-			if (create_process(&thread_tbl[i], cpu))
+			if (create_process(&thread_tbl[i], cpu, start_args->thr_params.stack_size))
```
JannePeltonen, line 437: This is a backward incompatible change since the caller may not have memzeroed the thread parameter struct but just explicitly initialized the fields that existed in a previous version, leaving stack_size to an uninitialized value.

Since this function is deprecated, maybe its behaviour should not be changed in a backward incompatible manner. Then old code calling the function would not be at the risk of breaking. OTOH, if the calling code could be changed (e.g. to use the new parameter), then I suppose it should just starting using odph_thread_create() instead.

I think it would be good to stop using this function in ODP source tree, especially in the examples.

Otherwise the patches look good to me.

JereLeppanen: > This is a backward incompatible change since the caller may not have memzeroed the thread parameter struct but just explicitly initialized the fields that existed in a previous version, leaving stack_size to an uninitialized value.

Yes, and the same applies to the other added parameter in this series as well.

> Since this function is deprecated, maybe its behaviour should not be changed in a backward incompatible manner.

Agree. But in that case, either the documentation for this function should state that stack_size is ignored, or stack_size should be added only to odph_thread_param_t, not odph_odpthread_params_t. The latter is currently just a typedef on the former. I'll come up with something.

> I think it would be good to stop using this function in ODP source tree, especially in the examples.

There are quite a few occurrences, but I can do that too.

JereLeppanen: Fixed these in v2.

helper/threads.c
```diff
@@ -118,6 +122,25 @@ static int create_process(odph_thread_t *thread, int cpu)
...
+			ODPH_ERR("getrlimit() failed: %s\n", strerror(errno));
+			return -3;
+		}
+
+		rlimit.rlim_cur = stack_size;
+
+		if (rlimit.rlim_cur > rlimit.rlim_max)
+			rlimit.rlim_max = rlimit.rlim_cur;
+
+		if (setrlimit(RLIMIT_STACK, &rlimit)) {
```
MatiasElo, line 138: An unprivileged user is only able to set `rlim_cur` (up to current `rlim_max`). Maybe it would be good to add a specific error print for this case (EPERM).

helper/test/odpthreads.c
```diff
@@ -143,15 +144,19 @@ int main(int argc, char *argv[])
 	printf("new cpu mask:               %s\n", cpumaskstr);
 	printf("new num worker threads:     %i\n\n", num_workers);
 
-	memset(&thr_params, 0, sizeof(thr_params));
-	thr_params.start    = worker_fn;
-	thr_params.arg      = NULL;
-	thr_params.thr_type = ODP_THREAD_WORKER;
-	thr_params.instance = odp_instance;
+	odph_thread_common_param_init(&thr_common);
```
MatiasElo, line 147: It would be good to add a test case for the new `sync_timeout` field.

helper/test/odpthreads.c
```diff
@@ -143,15 +144,19 @@ int main(int argc, char *argv[])
...
+	thr_common.instance = odp_instance;
+	thr_common.cpumask = &cpu_mask;
+	thr_common.share_param = 1;
 
-	odph_odpthreads_create(&thread_tbl[0], &cpu_mask, &thr_params);
+	odph_thread_param_init(&thr_param);
```
MatiasElo, line 152: It would be good to add a test case for the new `stack_size` field.

example/ipfragreass/odp_ipfragreass.c
```diff
@@ -230,8 +230,9 @@ int main(void)
 	odp_pool_t fragment_pool;
 	odp_shm_t shm;
 	odp_cpumask_t cpumask;
-	odph_odpthread_t threads[MAX_WORKERS];
-	odph_odpthread_params_t thread_params;
```
MatiasElo, line 234: Should we deprecate these types and related functions in the helper API?

helper/threads.c
JereLeppanen: This prints something like: "setrlimit() failed: Operation not permitted". Isn't that enough?

example/ipfragreass/odp_ipfragreass.c
JereLeppanen: I added the `@deprecated` tag to the documentation of these types. The functions are already tagged. We'll think about actual deprecation later.

helper/threads.c
JereLeppanen: After another look, I think it's best if we don't try to change rlim_max here, it should come from the system administrator (or equivalent). So I removed this in v3.

helper/test/odpthreads.c
```diff
@@ -12,6 +12,11 @@
 
 #include <unistd.h>
 #include <stdlib.h>
+#include <sys/time.h>
```
MatiasElo, line 15: Nokia's copyright can be added to the file.

helper/test/odpthreads.c
```diff
@@ -160,6 +168,39 @@ int main(int argc, char *argv[])
 	if (ret < 0)
 		exit(EXIT_FAILURE);
 
+	/* Test threads with non-default stack size and sync timeout. */
+
+	pthread_attr_init(&attr);
+
+	if (pthread_attr_getstacksize(&attr, &stack_size)) {
+		printf("pthread_attr_getstacksize() failed\n");
```
MatiasElo, line 176: Please use `ODPH_ERR()`

helper/test/odpthreads.c
```diff
@@ -160,6 +168,39 @@ int main(int argc, char *argv[])
...
+		return -1;
+	}
+
+	printf("\n");
+	printf("pthread default stack size is %zu\n", stack_size);
+
+	if (getrlimit(RLIMIT_STACK, &rlimit)) {
+		printf("getrlimit() failed: %s\n", strerror(errno));
```
MatiasElo, line 184: `ODPH_ERR()`

helper/test/odpthreads.c
```diff
@@ -160,6 +168,39 @@ int main(int argc, char *argv[])
...
+
+	pthread_attr_init(&attr);
+
+	if (pthread_attr_getstacksize(&attr, &stack_size)) {
+		printf("pthread_attr_getstacksize() failed\n");
+		return -1;
+	}
+
+	printf("\n");
+	printf("pthread default stack size is %zu\n", stack_size);
```
MatiasElo, line 181: The print values could be aligned for better readability. Now the output is a bit messy:
```
default cpu mask:           0xFFFF00000000
default num worker threads: 16
the first CPU:              32
new cpu mask:               0xFFFF00000000
new num worker threads:     16

Worker thread on CPU 34
Worker thread on CPU 35
Worker thread on CPU 41
Worker thread on CPU 42
Worker thread on CPU 32
Worker thread on CPU 37
Worker thread on CPU 40
Worker thread on CPU 38
Worker thread on CPU 36
Worker thread on CPU 33
Worker thread on CPU 39
Worker thread on CPU 43
Worker thread on CPU 44
Worker thread on CPU 45
Worker thread on CPU 47
Worker thread on CPU 46

pthread default stack size is 8388608
stack size soft limit (rlim_cur) is 8388608
use stack size 8384512
use sync timeout 5000000000
```

MatiasElo (commented):
Some minor comments but otherwise looking good. You can add my reviewed-by tags to the next version.

MatiasElo (approved):

PR #1357, opened 2021-08-13
[PATCH v11] linux-gen: atomic: emit STADD explicitly for __atomic_fetch_add operation

Malvika-Gupta:
Following performance gains were observed for the atomic add/sub/inc/dec APIs in the atomic perf test:
```
         _______________________________________
        |   odp_atomic_add_u32   |     7-8%   |
        |   odp_atomic_sub_u32    |     7-8%   |
        |   odp_atomic_inc_u32     |     5-6%   |
        |   odp_atomic_dec_u32    |     4-5%   |
        |   odp_atomic_add_u64    |     4-5%   |
        |   odp_atomic_sub_u64    |     5-6%   |
        |   odp_atomic_inc_u64     |     5-6%   |
        |   odp_atomic_dec_u64    |     4-5%   |
        |___________________________|__________ |

```
For the pool performance test, no significant performance improvement was observed with this change.

### DIFF
--- a/.github/workflows/ci-pipeline-arm64.yml
+++ b/.github/workflows/ci-pipeline-arm64.yml
@@ -108,6 +108,23 @@ jobs:
         if: ${{ failure() }}
         run: find . -name "*.trs" | xargs grep -l '^.test-result. FAIL' | while read trs ; do echo FAILURE detected at $trs; cat ${trs%%.trs}.log ; done
 
+  Run_CFLAGS:
+    if: ${{ github.repository == 'OpenDataPlane/odp' }}
+    runs-on: [self-hosted, ARM64]
+    strategy:
+      fail-fast: false
+      matrix:
+        cc: [gcc, clang]
+        cflags: ['-march=armv8.2-a -O2', '-march=armv8-a+lse -O2']
+    steps:
+      - uses: AutoModality/action-clean@v1.1.0
+      - uses: actions/checkout@v2
+      - run: sudo docker run -i -v `pwd`:/odp --privileged --shm-size 8g -e CC="${{matrix.cc}}" -e ARCH="${ARCH}"
+               -e CXX=g++-10 -e CFLAGS="${{matrix.cflags}}" $CONTAINER_NAMESPACE/odp-ci-${OS}-${ARCH}-native /odp/scripts/ci/check.sh
+      - name: Failure log
+        if: ${{ failure() }}
+        run: find . -name "*.trs" | xargs grep -l '^.test-result. FAIL' | while read trs ; do echo FAILURE detected at $trs; cat ${trs%%.trs}.log ; done
+
 
   Run_OS:
     if: ${{ github.repository == 'OpenDataPlane/odp' }}
--- a/.github/workflows/ci-pipeline.yml
+++ b/.github/workflows/ci-pipeline.yml
@@ -100,6 +100,18 @@ jobs:
         run: sudo docker run -i -v `pwd`:/odp --privileged --shm-size 8g -e CC="${{matrix.cc}}"
                -e CONF="${CONF}" $CONTAINER_NAMESPACE/odp-ci-${OS}-${ARCH} /odp/scripts/ci/build_${ARCH}.sh
 
+      - name: CFLAGS="-march=armv8.2-a"
+        env:
+          CONF: "CFLAGS=-march=armv8.2-a"
+        run: sudo docker run -i -v `pwd`:/odp --privileged --shm-size 8g -e CC="${{matrix.cc}}"
+               -e CONF="${CONF}" $CONTAINER_NAMESPACE/odp-ci-${OS}-${ARCH} /odp/scripts/ci/build_${ARCH}.sh
+
+      - name: CFLAGS="-march=armv8-a+lse"
+        env:
+          CONF: "CFLAGS=-march=armv8-a+lse"
+        run: sudo docker run -i -v `pwd`:/odp --privileged --shm-size 8g -e CC="${{matrix.cc}}"
+               -e CONF="${CONF}" $CONTAINER_NAMESPACE/odp-ci-${OS}-${ARCH} /odp/scripts/ci/build_${ARCH}.sh
+
       - name: Ubuntu 20.04
         env:
           CONF: "--enable-dpdk-shared"
--- a/platform/linux-generic/arch/aarch64/odp/api/abi/atomic_inlines.h
+++ b/platform/linux-generic/arch/aarch64/odp/api/abi/atomic_inlines.h
@@ -152,7 +152,60 @@ static inline int _odp_atomic_cas_acq_rel_u128(odp_atomic_u128_t *atom, odp_u128
 
 	return 0;
 }
-#else /* _ODP_LOCK_FREE_128BIT_ATOMICS */
+
+static inline void _odp_atomic_add_u32(odp_atomic_u32_t *atom, uint32_t val)
+{
+	__asm__ volatile("stadd   %w[val], %[atom]"
+			 : [atom] "+Q" (atom->v)
+			 : [val] "r" (val));
+}
+
+static inline void _odp_atomic_sub_u32(odp_atomic_u32_t *atom, uint32_t val)
+{
+	int32_t neg_val = -val;
+
+	__asm__ volatile("stadd   %w[neg_val], %[atom]"
+			 : [atom] "+Q" (atom->v)
+			 : [neg_val] "r" (neg_val));
+}
+
+static inline void _odp_atomic_inc_u32(odp_atomic_u32_t *atom)
+{
+	_odp_atomic_add_u32(atom, 1);
+}
+
+static inline void _odp_atomic_dec_u32(odp_atomic_u32_t *atom)
+{
+	_odp_atomic_sub_u32(atom, 1);
+}
+
+static inline void _odp_atomic_add_u64(odp_atomic_u64_t *atom, uint64_t val)
+{
+	__asm__ volatile("stadd   %[val], %[atom]"
+			 : [atom] "+Q" (atom->v)
+			 : [val] "r" (val));
+}
+
+static inline void _odp_atomic_sub_u64(odp_atomic_u64_t *atom, uint64_t val)
+{
+	int64_t neg_val = -val;
+
+	__asm__ volatile("stadd   %[neg_val], %[atom]"
+			 : [atom] "+Q" (atom->v)
+			 : [neg_val] "r" (neg_val));
+}
+
+static inline void _odp_atomic_inc_u64(odp_atomic_u64_t *atom)
+{
+	_odp_atomic_add_u64(atom, 1);
+}
+
+static inline void _odp_atomic_dec_u64(odp_atomic_u64_t *atom)
+{
+	_odp_atomic_sub_u64(atom, 1);
+}
+
+#else /* !_ODP_LOCK_FREE_128BIT_ATOMICS */
 
 /* Use generic implementation */
 #include <odp/api/abi/atomic_generic.h>
--- a/platform/linux-generic/arch/default/odp/api/abi/atomic_generic.h
+++ b/platform/linux-generic/arch/default/odp/api/abi/atomic_generic.h
@@ -10,6 +10,46 @@
 
 #include <odp/api/atomic.h>
 
+static inline void _odp_atomic_add_u32(odp_atomic_u32_t *atom, uint32_t val)
+{
+	(void)__atomic_fetch_add(&atom->v, val, __ATOMIC_RELAXED);
+}
+
+static inline void _odp_atomic_sub_u32(odp_atomic_u32_t *atom, uint32_t val)
+{
+	(void)__atomic_fetch_sub(&atom->v, val, __ATOMIC_RELAXED);
+}
+
+static inline void _odp_atomic_inc_u32(odp_atomic_u32_t *atom)
+{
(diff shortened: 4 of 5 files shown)


.github/workflows/ci-pipeline.yml
```diff
@@ -100,6 +100,18 @@ jobs:
...
 
+      - name: CFLAGS="-march=armv8.2-a"
+        env:
+          CONF: "CFLAGS='-march=armv8.2-a'"
+        run: sudo docker run -i -v `pwd`:/odp --privileged --shm-size 8g -e CC="${{matrix.cc}}"
+               -e CONF="${CONF}" $CONTAINER_NAMESPACE/odp-ci-${OS}-${ARCH} /odp/scripts/ci/build_${ARCH}.sh
+
+      - name: CFLAGS="-march=armv8-a+lse"
+        env:
+          CONF: "CFLAGS='-march=armv8-a+lse'"
```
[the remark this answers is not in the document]
Malvika-Gupta: Thanks Matias! I will make the changes and push a new version

m-govind:
> Following performance gains were observed for the atomic add/sub/inc/dec APIs in the atomic perf test:
> 
> ```
>          _______________________________________
>         |   odp_atomic_add_u32   |     7-8%   |
>         |   odp_atomic_sub_u32    |     7-8%   |
>         |   odp_atomic_inc_u32     |     5-6%   |
>         |   odp_atomic_dec_u32    |     4-5%   |
>         |   odp_atomic_add_u64    |     4-5%   |
>         |   odp_atomic_sub_u64    |     5-6%   |
>         |   odp_atomic_inc_u64     |     5-6%   |
>         |   odp_atomic_dec_u64    |     4-5%   |
>         |___________________________|__________ |
> ```
> 
> For the pool performance test, no significant performance improvement was observed with this change.

Hi Petri, Let us know whether these performance numbers are good. As Malvika mentioned, we don't see an improvement with pool performance tests.

platform/linux-generic/arch/aarch64/odp/api/abi/atomic_inlines.h
```diff
@@ -152,7 +152,62 @@ static inline int _odp_atomic_cas_acq_rel_u128(odp_atomic_u128_t *atom, odp_u128
...
+			 : "memory");
+}
+
+static inline void _odp_atomic_sub_u32(odp_atomic_u32_t *atom, uint32_t val)
+{
+	__asm__ volatile("neg     %w0, %w0\n"
+			 "stadd   %w0, [%1]"
+			 :
+			 : "r" (val), "r" (atom)
+			 : "memory");
```
MatiasElo, line 170: This code doesn't seem to work if optimization is enabled (`-O2`).

E.g. atomic validation test
```
Num of threads used = 12
	ODP API version: 1.31.0
	ODP implementation name:    odp-linux
	ODP implementation version: odp-linux 1.31.0-0 (v1.31.0) 1.31.0.0


     CUnit - A unit testing framework for C - Version 2.1-3
     http://cunit.sourceforge.net/


Suite: atomic
  Test: atomic_test_atomic_inc_dec ...FAILED
    1. atomic.c:696  - U32_INIT_VAL == odp_atomic_load_u32(&global_mem->a32u)
    2. atomic.c:697  - U64_INIT_VAL == odp_atomic_load_u64(&global_mem->a64u)
  Test: atomic_test_atomic_add_sub ...threads.c:212:wait_pthread():Bad exit status cpu #12 0x2
error: odph_thread_join() failed.
FAILED
    1. atomic.c:696  - U32_INIT_VAL == odp_atomic_load_u32(&global_mem->a32u)
    2. atomic.c:697  - U64_INIT_VAL == odp_atomic_load_u64(&global_mem->a64u)
  Test: atomic_test_atomic_fetch_inc_dec ...threads.c:212:wait_pthread():Bad exit status cpu #12 0x4
error: odph_thread_join() failed.
passed
  Test: atomic_test_atomic_fetch_add_sub ...threads.c:212:wait_pthread():Bad exit status cpu #12 0x4
error: odph_thread_join() failed.
passed
  Test: atomic_test_atomic_max_min ...threads.c:212:wait_pthread():Bad exit status cpu #12 0x4
error: odph_thread_join() failed.
passed
  Test: atomic_test_atomic_cas_inc_dec ...threads.c:212:wait_pthread():Bad exit status cpu #12 0x4
error: odph_thread_join() failed.
passed
  Test: atomic_test_atomic_xchg ...^C
```
I'm not that familiar with assembly coding, but I guess you shouldn't write to an input register.

Malvika-Gupta:
Hi @MatiasElo, I have fixed the issue with -O2 optimization and checked that it works in my local environment. Please review. Thanks!

platform/linux-generic/arch/aarch64/odp/api/abi/atomic_inlines.h
```diff
@@ -152,7 +152,62 @@ static inline int _odp_atomic_cas_acq_rel_u128(odp_atomic_u128_t *atom, odp_u128
...
+}
+
+static inline void _odp_atomic_sub_u32(odp_atomic_u32_t *atom, uint32_t val)
+{
+	__asm__ volatile("neg     %w0, %w0\n"
+			 "stadd   %w0, [%1]"
+			 : "+r" (val)
+			 : "r" (atom)
+			 : "memory");
```
MatiasElo, line 170: We would prefer avoiding the `neg` call if possible (seems to hinder inlining). Would the following code work?
```
	int32_t neg_val = -val;

	__asm__ volatile("stadd   %w0, [%1]"
			 :
			 : "r" (neg_val), "r" (atom)
			 : "memory");
```

.github/workflows/ci-pipeline-arm64.yml
```diff
@@ -98,7 +98,8 @@ jobs:
         conf: ['', '--enable-abi-compat', '--enable-deprecated --enable-debug=full',
                '--enable-dpdk-zero-copy --disable-static-applications',
                '--disable-host-optimization', '--disable-host-optimization --enable-abi-compat',
-               '--without-openssl --without-pcap']
+               '--without-openssl --without-pcap', 'CFLAGS="-march=armv8.2-a"',
+               'CFLAGS="-march=armv8-a+lse"']
```
MatiasElo, line 102: It's probably a good idea to add also `-O2` flag, since it previously revealed an implementation bug.

MatiasElo:
Hi @Malvika-Gupta,

>Also, I see that the armv8-a+lse jobs are even not getting triggered for x86 CI but are successfully being deployed for the arm64 CI. Can you please look into why that is happening?

The x86 jobs are actually triggered, they are just a bit hidden.
Screen Shot 2021-09-16 at 17 41 15

platform/linux-generic/arch/aarch64/odp/api/abi/atomic_inlines.h
Malvika-Gupta: Hi Matias, I have changed the code to above in v4. It works as expected and all test cases pass with -O2 flag as well. Please review.

.github/workflows/ci-pipeline-arm64.yml
Malvika-Gupta: I have made this change in v4. Thanks!

Malvika-Gupta:
> Hi @Malvika-Gupta,
> 
> > Also, I see that the armv8-a+lse jobs are even not getting triggered for x86 CI but are successfully being deployed for the arm64 CI. Can you please look into why that is happening?
> 
> The x86 jobs are actually triggered, they are just a bit hidden.
> Screen Shot 2021-09-16 at 17 41 15

Thanks for clarifying this, Matias!

.github/workflows/ci-pipeline-arm64.yml
```diff
@@ -98,7 +98,8 @@ jobs:
         conf: ['', '--enable-abi-compat', '--enable-deprecated --enable-debug=full',
                '--enable-dpdk-zero-copy --disable-static-applications',
                '--disable-host-optimization', '--disable-host-optimization --enable-abi-compat',
-               '--without-openssl --without-pcap']
+               '--without-openssl --without-pcap', 'CFLAGS="-O2 -march=armv8.2-a"',
+               'CFLAGS="-O2 -march=armv8-a+lse"']
```
MatiasElo, line 102: It looks like the additional quotation marks are not handled correctly, which causes the CI failures. We may need to do some trial and error to figure out a syntax that works.

platform/linux-generic/arch/aarch64/odp/api/abi/atomic_inlines.h
```diff
@@ -152,7 +152,64 @@ static inline int _odp_atomic_cas_acq_rel_u128(odp_atomic_u128_t *atom, odp_u128
 
 	return 0;
 }
-#else /* _ODP_LOCK_FREE_128BIT_ATOMICS */
+
+static inline void _odp_atomic_add_u32(odp_atomic_u32_t *atom, uint32_t val)
```
MatiasElo, line 156: The commit subject line could be tweaked a bit:
`linux-gen: atomic: use STADD explicitly for aarch64 atomic add operations`

MatiasElo (commented):
The arm64 CI still needs to be fixed but otherwise looking OK.

platform/linux-generic/arch/aarch64/odp/api/abi/atomic_inlines.h
Malvika-Gupta: Sure, I will make the change. Thanks Matias!

.github/workflows/ci-pipeline-arm64.yml
Malvika-Gupta: I will push another version with some changes and we can see if it works. Unfortunately, there is no way for me to run the arm64 jobs in my GitHub account so I'll keep submitting new versions until we figure it out.

Malvika-Gupta:
Hi @MatiasElo I have tried a bunch of different combinations to fix the arm64 CI job. One of them didn't make it past docker creation and failed there itself. With the following options, docker creation was successful but the job doesn't move past the configuration:
```
- 'CFLAGS=-O2 -march=armv8-a+lse'
- 'CFLAGS=-O2 -march="armv8-a+lse"'
- 'CFLAGS=-O2 "-march=armv8-a+lse"'
```
I did observe something rather odd. For all the above scenarios, the error was `**configure: error: unrecognized option: `-march=armv8-a+lse'**`. If I reverse the order of the arguments passed to CFLAGS, i.e., 'CFLAGS=-O2 -march=armv8-a+lse', then the error becomes `**configure: error: unrecognized option: `-O2'**`. This seems to imply that there is an issue with CFLAGS accepting two arguments during configuration. I looked at all the previously added jobs and none of them pass two arguments to CFLAGS. Maybe we need to modify something in `configure.ac` for this job to succeed?

MatiasElo:
Well that was more complicated than I originally though. I finally figured out a working syntax: https://github.com/OpenDataPlane/odp/pull/1374/commits/f4ddd28b96b9dd6e7a0a1ed7362065e5fe96099b

platform/linux-generic/arch/aarch64/odp/api/abi/atomic_inlines.h
```diff
@@ -152,7 +152,64 @@ static inline int _odp_atomic_cas_acq_rel_u128(odp_atomic_u128_t *atom, odp_u128
...
+			 :
+			 : "r" (val), "r" (atom)
+			 : "memory");
+}
+
+static inline void _odp_atomic_sub_u32(odp_atomic_u32_t *atom, uint32_t val)
+{
+	int32_t neg_val = -val;
+
+	__asm__ volatile("stadd   %w0, [%1]"
```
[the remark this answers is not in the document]
MatiasElo: I tested the current implementation (`__atomic_fetch_add()`) with up to GCC 11.1.0 and the produced code used always `ldadd`.

MatiasElo (approved):

platform/linux-generic/arch/aarch64/odp/api/abi/atomic_inlines.h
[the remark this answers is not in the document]
Malvika-Gupta: We brought this up in our discussion with Arm compiler's team last week and there is an issue in letting the compiler generate an STADD instead of LDADD from `atomic_fetch_add()` directly when the return value is not used. Essentially, if we perform an `atomic_fetch_add` before an `atomic_thread_fence (acquire)`, then the add operation should not be allowed to pass the fence. But, because we've  let the LDADD decay to an STADD for an unused return value, it can now pass the one-way barrier. You go look at this thread for a more detailed explanation: https://gcc.gnu.org/legacy-ml/gcc-patches/2018-10/msg02042.html

.github/workflows/ci-pipeline-arm64.yml
MatiasElo, line 102: `-O2` flag, which revealed earlier an implementation bug, should also be used. It could be enabled with a following syntax: https://github.com/OpenDataPlane/odp/commit/f4ddd28b96b9dd6e7a0a1ed7362065e5fe96099b

MatiasElo (changes requested):
The implementation looks good, but the CI tests still need to include `-O2` flag.

.github/workflows/ci-pipeline-arm64.yml
Malvika-Gupta: Sorry, my bad. I applied the older patch when upstreaming. I have updated the PR. Thanks Matias!

MatiasElo (approved):

PR #1359, opened 2021-08-18
[PATCH v7] Helper cli

Lifang-Zhang:
Add new API function odph_cli_log_va() and a new member "hostname" to odph_cli_param_t to allow users to provide hostname.

### DIFF
--- a/helper/cli.c
+++ b/helper/cli.c
@@ -528,6 +528,15 @@ int odph_cli_log(const char *fmt, ...)
 	return r;
 }
 
+ODP_PRINTF_FORMAT(1, 0)
+int odph_cli_log_va(const char *fmt, va_list in_args)
+{
+	int r;
+
+	r = cli_log_va(ODP_LOG_PRINT, fmt, in_args);
+	return r;
+}
+
 static int msg_recv(int fd)
 {
 	uint32_t msg;
--- a/helper/cli.c
+++ b/helper/cli.c
@@ -56,6 +56,7 @@ static const odph_cli_param_t param_default = {
 	.address = "127.0.0.1",
 	.port = 55555,
 	.max_user_commands = 50,
+	.hostname = "ODP",
 };
 
 void odph_cli_param_init(odph_cli_param_t *param)
@@ -371,7 +372,7 @@ static struct cli_def *create_cli(cli_shm_t *shm)
 
 	cli = cli_init();
 	cli_set_banner(cli, NULL);
-	cli_set_hostname(cli, "ODP");
+	cli_set_hostname(cli, shm->cli_param.hostname);
 
 	c = cli_register_command(cli, NULL, "call", NULL,
 				 PRIVILEGE_UNPRIVILEGED, MODE_EXEC,
--- a/helper/cli.c
+++ b/helper/cli.c
@@ -46,7 +46,6 @@ typedef struct {
 	odp_instance_t instance;
 	odph_cli_param_t cli_param;
 	struct sockaddr_in addr;
-	uint32_t max_user_commands;
 	uint32_t num_user_commands;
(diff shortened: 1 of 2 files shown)


JereLeppanen:
How about we change this helper so that the caller, not the helper, creates the CLI server thread? In that case we don't need any of the init and term callbacks. If that sounds good, I could probably make such a change pretty quickly.

Lifang-Zhang:
> How about we change this helper so that the caller, not the helper, creates the CLI server thread? In that case we don't need any of the init and term callbacks. If that sounds good, I could probably make such a change pretty quickly.

That would be perfect since that is actually what we want. I will delete the last few commits after your change :)

JereLeppanen:
> That would be perfect since that is actually what we want. I will delete the last few commits after your change :)

I've posted the changes in PR #1369, which is now ready for review. You can probably rebase on top of that.

helper/cli.c
```diff
@@ -44,6 +44,7 @@ typedef struct {
 	odp_spinlock_t lock;
 	odp_spinlock_t api_lock;
 	odp_instance_t instance;
+	odph_cli_param_t cli_param;
```
JereLeppanen, line 47: A typo in the commit comment: "affacting".

helper/include/odp/helper/cli.h
```diff
@@ -131,6 +132,25 @@ int odph_cli_run(void);
...
+/**
+ * Log to CLI
+ *
+ * Mainly for it to be used by app specific log functions in the form
+ * of LOG(level, fmt, ...) where level is often an enum type and differs
+ * in applications. For example, a function can be defined in the app specific
+ * LOG form (app_log_level_t level, const char *fmt, ...), inside which
+ * level is made unused and odph_cli_log_va is called. This new function can
+ * then be used to replace the app specific LOG function in order to print to
+ * CLI client instead of local terminal.
```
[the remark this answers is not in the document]
Lifang-Zhang: Now fix them.

helper/include/odp/helper/cli.h
```diff
@@ -131,6 +132,20 @@ int odph_cli_run(void);
...
+ * Similar to odph_cli_log(), except that this one takes its arguments as
+ * a va_list.
+ *
+ * @param fmt printf-style message format
+ * @param in_args variadic arguments
+ * @return On success, the number of characters printed or buffered, without
+ *         accounting for any line feed conversions. If an error is encountered,
+ *         a negative value is returned.
+ */
+int odph_cli_log_va(const char *fmt, va_list in_args);
```
MatiasElo, line 149: Nitpicking, but it would be more logical that `odph_cli_log()` is defined before `odph_cli_log_va()`.

helper/cli.c
MatiasElo, line 47: Space missing in Signed-off-by tag between name and email.

helper/cli.c
```diff
@@ -56,6 +56,7 @@ static const odph_cli_param_t param_default = {
 	.address = "127.0.0.1",
 	.port = 55555,
 	.max_user_commands = 50,
+	.hostname = "ODP",
```
MatiasElo, line 59: Commit message: `Host name` -> `Hostname`

MatiasElo (commented):
Couple minor comments but otherwise looking good. You can add my reviewed-by tags in the next version. @JereLeppanen have all your comments been resolved?

helper/cli.c
Lifang-Zhang: Now have them fixed :)

MatiasElo (approved):

JereLeppanen:
Yes, my comments have also been resolved.

JereLeppanen (approved):

PR #1360, opened 2021-08-18
linux-gen: ipsec: fix for flow & tos values in ipv6hdr

vijayinavolu:
These values are not updated for ipv6 tunnel correctly due
to endianness issue


platform/linux-generic/odp_ipsec.c
```diff
@@ -953,13 +953,14 @@ static int ipsec_out_tunnel_parse_ipv6(ipsec_state_t *state,
 				       ipsec_sa_t *ipsec_sa)
 {
 	_odp_ipv6hdr_t *ipv6hdr = state->ip;
+	uint32_t ver_tc_flow = odp_cpu_to_be_32(ipv6hdr->ver_tc_flow);
```
JannePeltonen, line 956: This should use odp_be_to_cpu_32(). Even though the effect is the same, that is the logical direction of the conversion here.

JannePeltonen:
Please rebase on the latest master.

It seems there is an existing bug in ipsec_out_tunnel_ipv6() where where the whole tos octet is copied to the outer header instead of just the DSCP field. IOW, also the ECN bits get copied which seems to violate the API spec. Would you like to fix that problem too?

platform/linux-generic/odp_ipsec.c
vijayinavolu: Will update in next version

vijayinavolu:
> Please rebase on the latest master.
> 
> It seems there is an existing bug in ipsec_out_tunnel_ipv6() where where the whole tos octet is copied to the outer header instead of just the DSCP field. IOW, also the ECN bits get copied which seems to violate the API spec. Would you like to fix that problem too?

Sure, will fix this in next version

JannePeltonen:
The subject line of the commit message could be improved a little, e.g.: 
linux-gen: ipsec: fix flow label & DSCP in IPv6 tunnel encapsulation

> As per the spec (below) ECN bits should be always copied from inner to outer header irrespective of the odp ipsec dscp copy configuration, so this might not be an issue ?

You are correct, there is no bug there. Sorry for the false alarm.

PR #1367, opened 2021-08-27
[PATCH v2] crypto: fix processing of multi-segment packets

JannePeltonen:
Fix processing of multi-segment packets and do validation testing with multi-segment packets too.

### DIFF
--- a/platform/linux-generic/odp_crypto_openssl.c
+++ b/platform/linux-generic/odp_crypto_openssl.c
@@ -1,4 +1,5 @@
 /* Copyright (c) 2014-2018, Linaro Limited
+ * Copyright (c) 2021, Nokia
  * All rights reserved.
  *
  * SPDX-License-Identifier:     BSD-3-Clause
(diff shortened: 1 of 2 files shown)


JereLeppanen (commented):
Looks good to me. Just one question about a context line.

platform/linux-generic/odp_crypto_openssl.c
```diff
@@ -438,8 +438,12 @@ void packet_aes_xcbc_mac(odp_packet_t pkt,
 		len -= datalen;
 		if (eoff != 0) {
 			if (eoff + datalen > AES_BLOCK_SIZE) {
```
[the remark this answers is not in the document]
JannePeltonen: No. The reason is that in AES-XCBC-MAC (see. e.g. RFC 3566 section 4 for a nice description) the last block, even if full, requires special handling. In the case of a full block there is additional XORing with 'K2' before the encryption operation. The code makes sure that in all cases the last block. whether partial or full, is handled at the end outside the loop.

JereLeppanen (approved):

test/validation/api/crypto/odp_crypto_test_inp.c
```diff
@@ -444,9 +477,9 @@ static void alg_test_execute(const alg_test_param_t *param)
 	uint8_t *cipher_iv = param->override_iv ? ref->cipher_iv : NULL;
 	uint8_t *auth_iv   = param->override_iv ? ref->auth_iv : NULL;
 
-	cipher_range.offset = 0;
+	cipher_range.offset = param->header_len;
```
asasidharan, line 480: For bit mode algorithms, as per the current spec, cipher_range.offset and auth_range.offset should be in bits not bytes.

Apologies for the delayed response. We tried verifying this on our platform and ran into failures with bit mode algorithms and had been debugging the same.

JannePeltonen: Thanks for testing and investigating this. The interesting thing is that validation tests with linux-gen pass now but with the fix they fail. It seems the linux-gen implementation of bit mode crypto and the implementation of EIA2 CMAC are buggy in that they assume the offset is in bytes.

JannePeltonen: I made a new PR (1380) to fix both the validation test and linux-gen implementation. Please take a look if it fixes the validation problem with your ODP.

asasidharan: Thank you @JannePeltonen for the new PR to fix this. Will verify with the new PR and update you there.

PR #1371, opened 2021-09-13
[PATCH v3] Fix parsing of segmented packets

JannePeltonen:
Various fixes for packet parsing, mostly for segmented packets that get parsed when requested through odp_packet_parse() or after going through loop.

### DIFF
--- a/platform/linux-generic/odp_packet.c
+++ b/platform/linux-generic/odp_packet.c
@@ -1922,6 +1922,16 @@ static inline uint16_t parse_eth(packet_parser_t *prs, const uint8_t **parseptr,
 		*parseptr += sizeof(_odp_vlanhdr_t);
 	}
 
+	/*
+	 * The packet was too short for what we parsed. We just give up
+	 * entirely without trying to parse what fits in the packet.
+	 */
+	if (odp_unlikely(*offset > frame_len)) {
+		input_flags.all = 0;
+		input_flags.l2  = 1;
+		ethtype = 0;
+	}
+
 error:
 	prs->input_flags.all |= input_flags.all;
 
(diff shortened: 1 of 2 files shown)


platform/linux-generic/odp_packet.c
```diff
@@ -1847,7 +1848,11 @@ static uint32_t packet_sum_crc32c(odp_packet_hdr_t *pkt_hdr,
 	return sum;
 }
 
-/** Parser helper function for Ethernet packets */
+#define PARSE_ETH_BYTES (sizeof(_odp_ethhdr_t) + 8 + 2 * sizeof(_odp_vlanhdr_t))
```
JereLeppanen, line 1851: Especially the `+8` here could use a comment, even if it's not unclear where it comes from.

platform/linux-generic/odp_packet.c
```diff
@@ -2095,8 +2113,11 @@ static inline void parse_tcp(packet_parser_t *prs, const uint8_t **parseptr,
 	*parseptr += len;
 }
 
+#define PARSE_UDP_BYTES (sizeof(_odp_udphdr_t) + 4)
```
JereLeppanen, line 2116: Same here for the `+4`.

platform/linux-generic/odp_packet.c
```diff
@@ -2607,6 +2635,8 @@ int odp_packet_parse(odp_packet_t pkt, uint32_t offset,
 	int ret;
 	uint16_t ethtype;
 	uint64_t l4_part_sum = 0;
+	const uint32_t min_seglen = PARSE_ETH_BYTES + PARSE_L3_L4_BYTES;
```
JereLeppanen, line 2638: Quite a lot of macroing to come up with one constant value. The alternative would be to just put the value here and comment where it comes from, but I'm sure you already considered that. Anyway, this is fine.

JannePeltonen: Yes, it is not very nice, but I did not see a better way. I think a single constant would be easily forgotten when things change. With the macros defined near the relevant functions I hope that the constraints stay up to date better.

JannePeltonen:
v2: Added comments as requested and rebased.

JereLeppanen (approved):

PR #1373, opened 2021-09-17
[PATCH v3] linux-gen: cpu: refactor inlining code

MatiasElo:
Refactor CPU ABI and implementation files to follow common project inlining
style. odp_cpu_pause() function is now defined only once per architecture.
After this clean-up it's simpler for an implementation to add inlined
versions of CPU module functions.

V3:
- Rebase


platform/linux-generic/arch/mips64/odp/api/abi/cpu_inlines.h
```diff
@@ -0,0 +1,56 @@
...
+	__asm__ __volatile__ ("nop");
+}
+
+uint64_t _odp_cpu_cycles(void)
+{
+	#define CVMX_TMP_STR(x) CVMX_TMP_STR2(x)
+	#define CVMX_TMP_STR2(x) #x
+	uint64_t cycle;
+
+	__asm__ __volatile__ ("rdhwr %[rt],$" CVMX_TMP_STR(31) :
```
JereLeppanen, line 31: CVMX_TMP_STR seems unnecessary. This could be just `"rdhwr %[rt],$31"`. And a comment explaining the magic number.

I suppose $31 is specific to Octeon, since mips64 says that $31 is "reserved for implementation-dependent use", while $2 is cycle counter.

MatiasElo: In this PR I'm only moving the current implementation as-is (e.g. this function was originally in `platform/linux-generic/arch/mips64/odp_cpu_cycles.c`). This function could be optimized in a separate PR.

However, before we spend more time on this we should decide if we still need to support mips64. Currently, the architecture is not tested in the CI at all.

JereLeppanen: Ok, agree. I got confused for a moment there, because github hides removed files.

JereLeppanen (approved):

PR #1375, opened 2021-09-20
[PATCH v2] api: ipsec: allow losing completion events if destination queue is full

JannePeltonen:
The first patch changes API so that packet completion events can be lost if the destination queue is full. The patch does not change whether SA disable completion events can be lost (the API is not explicit about it but the spirit is that those events cannot be lost, although in current linux-gen it can happen).

The second patch fixes existing bugs related to enqueue failures.


JannePeltonen:
v2: rebased

JannePeltonen:
Closing this PR as inactive and outdated. And maybe allowing crypto queues to drop events is not so good anyway.

PR #1376, opened 2021-09-20
[PATCH v2] api: ipsec: do not promise original packet back after all errors

JannePeltonen:
Specify that the output packet of an errored IPsec operation is not
the unmodified input packet but a packet with undefined content. This
is already the case in the linux-gen implementation and is basically
required by implementations that use in-place crypto operations which
modify the packet (with combined mode algorithms) before checking for
authentication error.

Still promise the original input packet back after sa_lookup and mtu
errors since in those cases the application is expected to continue
processing the packet.


psavol (approved):

anoobj:
Minor nit, "mtu" should be capitalized (it is done that way in the rest of the file).

PR #1379, opened 2021-09-23
[PATCH v3] github_ci: update gh pages build workflow

MatiasElo:
Use Ubuntu 20.04 to build documentation and bump ghaction-github-pages
version to v2.5.0.

### DIFF
--- a/.github/workflows/gh-pages.yml
+++ b/.github/workflows/gh-pages.yml
@@ -7,7 +7,7 @@ on:
 
 jobs:
   Documentation:
-    runs-on: ubuntu-18.04
+    runs-on: ubuntu-20.04
     steps:
     - uses: actions/checkout@v2
     - name: Install dependencies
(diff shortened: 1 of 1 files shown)


.github/workflows/gh-pages.yml
```diff
@@ -7,7 +7,7 @@ on:
 
 jobs:
   Documentation:
-    runs-on: ubuntu-18.04
+    runs-on: ubuntu-20.04
```
JereLeppanen, line 10: Is there a specific reason for this change? If yes, then it could be mentioned in the commit comment.

JereLeppanen (approved):

.github/workflows/gh-pages.yml
MatiasElo: The reasoning was to install a newer version of Doxygen to reveal possible new warnings. I'll update the commit log.

PR #1380, opened 2021-09-23
[PATCH v2] crypto: fix crypto and auth range offset in bit mode

JannePeltonen:
In bit mode the crypto and auth ranges are supposed to be given in bits,
not in bytes. Fix the linux-gen implementation that interpreted crypto and
auth offset in bytes in AES EEA2 and AES EIA2, respectively. Fix crypto
validation test to provide the offsets in bits as specified by the API.

### DIFF
--- a/platform/linux-generic/odp_crypto_openssl.c
+++ b/platform/linux-generic/odp_crypto_openssl.c
@@ -700,7 +700,7 @@ int packet_cmac_eia2(odp_packet_t pkt,
 {
 	CMAC_CTX *ctx = local.cmac_ctx[session->idx];
 	void *iv_ptr;
-	uint32_t offset = param->auth_range.offset;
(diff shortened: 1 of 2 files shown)


MatiasElo (approved):

anoobj:
Shouldn't this be 2 separate patches? One for linux-gen and one for validation test?

Apart from that,

JannePeltonen:
> Shouldn't this be 2 separate patches? One for linux-gen and one for validation test?

That would cause the intermediate revision to fail validation tests and we do not want that.

JannePeltonen:
@anoobj, can I add your reviewed-by tag to this patch which is not split?

anoobj:
@JannePeltonen You can add my reviewed-by.

PR #1381, opened 2021-09-23
[PATCH v5] api: shm: introduce flags capability and ODP_SHM_NO_HP flag

MatiasElo:
V2:
- Removed `odp_shm_addr_pa()` API
- `ODP_SHM_NO_HP` documentation clarification
- State that some flag combinations may not be supported

V3:
- Rebase

### DIFF
--- a/include/odp/api/spec/shared_memory.h
+++ b/include/odp/api/spec/shared_memory.h
@@ -97,6 +97,15 @@ extern "C" {
 #define ODP_SHM_HW_ACCESS	0x20
 
 /**
+ * Don't use huge pages
+ *
+ * When set, this flag guarantees that the memory reserved by odp_shm_reserve()
+ * is not allocated from huge pages. This flag must not be combined with
+ * ODP_SHM_HP.
+ */
+#define ODP_SHM_NO_HP		0x40
+
+/**
  * Shared memory block info
  */
 typedef struct odp_shm_info_t {
--- a/include/odp/api/spec/shared_memory.h
+++ b/include/odp/api/spec/shared_memory.h
@@ -1,5 +1,5 @@
-/* Copyright (c) 2019, Nokia
- * Copyright (c) 2013-2018, Linaro Limited
+/* Copyright (c) 2013-2018, Linaro Limited
+ * Copyright (c) 2019-2021, Nokia
  * All rights reserved.
  *
  * SPDX-License-Identifier:     BSD-3-Clause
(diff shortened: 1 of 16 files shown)


include/odp/api/spec/shared_memory.h
```diff
@@ -96,6 +96,14 @@ extern "C" {
  */
 #define ODP_SHM_HW_ACCESS	0x20
 
+/**
+ * Don't use huge pages
+ *
+ * When set, this flag guarantees that the memory reserved by odp_shm_reserve()
+ * is not allocated from huge pages.
```
psavol, line 103: Text could mention that this cannot be combined with ODP_SHM_HP

include/odp/api/spec/shared_memory.h
```diff
@@ -243,6 +243,31 @@ odp_shm_t odp_shm_import(const char *remote_name,
...
+ * @param[out] size   Array of physical memory segment lengths for output or
+ *                    NULL
+ * @param      num    Maximum number of physical addresses and memory segment
+ *                    lengths to output
+ *
+ * @return Number of physical addresses
+ * @retval 0 on failure
+ */
+uint32_t odp_shm_addr_pa(odp_shm_t shm, void *pa[], uint64_t size[],
+			 uint32_t num);
```
psavol, line 269: Pointer and size could be combined into a struct (odp_shm_addr_info_t). That would allow backward compatible additions in the future. We have odp_shm_info() already, but that does not seem flexible enough for variable number of PA addresses per block.

uint32_t odp_shm_addr_info(odp_shm_t shm, odp_shm_addr_info_t[], uint32_t num);

MatiasElo: Makes sense. We could add e.g. iova address there in the future.

include/odp/api/spec/shared_memory.h
MatiasElo: That should be quite self-evident but an explanation won't hurt.

MatiasElo: Fixed in V2.

include/odp/api/spec/shared_memory.h
MatiasElo: `odp_shm_addr_pa()` removed in V2.

psavol (approved):

JereLeppanen (approved):

jerinjacobk (approved):

jerinjacobk (approved):

PR #1385, opened 2021-10-05
linux-gen: atomic: explicitly use staddl for add/sub release ops

MatiasElo:
GCC generates LDADDL instructions when using __atomic_fetch_add() on
aarch64. Explicitly use STADDL to implement odp_atomic_add_rel_uXY() and
odp_atomic_sub_rel_uXY() functions for better performance.


m-govind:
Hi Matias,
  Do you seen an improvement in performance ?

m-govind:
> > Hi Matias,
> > Do you seen an improvement in performance ?
> 
> Hi Govind,
> 
> Although the performance gain is not as big as with the relaxed functions it's still significant on CN96xx. For example `odp_atomic_perf` test's throughput for `odp_atomic_add_rel_u64()` increased from 28.2M ops/sec to 38.7M ops/sec per core.

Thanks Matias.

PR #1386, opened 2021-10-05
[PATCH v2] validation: ipsec: allow orig_ip_len to be unavailable

JannePeltonen:
The API specifies that zero odp_ipsec_packet_result_t::orig_ip_len means
that the length is not available. Check orig_ip_len against the original
packet length only when it is available.


anoobj:

psavol (approved):

PR #1400, opened 2021-10-18
[PATCH v6] linux-gen: sysinfo: dump aarch64 cpu hardware flags

Malvika-Gupta:
CPU specific hardware/software features are exposed by the kernel to the
userspace through a set of flags called hwcaps, exposed in the auxiliary
vector. Userspace software can test for features by acquiring the AT_HWCAP
and AT_HWCAP2 entry of the auxiliary vector.
These hwcaps are defined in terms of architected system register fields
and those exposed in AT_HWCAP and AT_HWCAP2 typically have feature flags
associated with them. The mapped feature flags and their description
are printed as a part of the odp_sys_info_print() API.

Reviewed-by: Govindarajan <govindarajan.mohandoss@arm.com>

### DIFF
--- a/platform/linux-generic/arch/aarch64/cpu_flags.c
+++ b/platform/linux-generic/arch/aarch64/cpu_flags.c
@@ -1,0 +1,155 @@
+/* Copyright (c) 2018, Linaro Limited
+ * Copyright (c) 2020-2021, Nokia
+ * All rights reserved.
+ *
+ * SPDX-License-Identifier:     BSD-3-Clause
+ */
+
+#include <string.h>
+#include <stdlib.h>
+
+#include <odp/api/hints.h>
+#include <odp_debug_internal.h>
+#include "cpu_flags.h"
+
+static void _odp_sys_info_print_acle_flags(void)
+{
+	const char *ndef = "n/a";
+
+	/* Avoid compiler warning about unused variable */
+	(void)ndef;
+
+	/* See ARM C Language Extensions documentation for details */
+	ODP_PRINT("ARM FEATURES:\n");
+
+	ODP_PRINT("  __ARM_ARCH              ");
+#ifdef __ARM_ARCH
+	ODP_PRINT("%i\n", __ARM_ARCH);
+#else
+	ODP_PRINT("%s\n", ndef);
+#endif
+
+	ODP_PRINT("  __ARM_ARCH_ISA_A64      ");
+#ifdef __ARM_ARCH_ISA_A64
+	ODP_PRINT("%i\n", __ARM_ARCH_ISA_A64);
+#else
+	ODP_PRINT("%s\n", ndef);
+#endif
+
+	ODP_PRINT("  __ARM_FEATURE_ATOMICS   ");
+#ifdef __ARM_FEATURE_ATOMICS
+	ODP_PRINT("%i\n", __ARM_FEATURE_ATOMICS);
+#else
+	ODP_PRINT("%s\n", ndef);
+#endif
+
+	ODP_PRINT("  __ARM_FEATURE_CRC32     ");
+#ifdef __ARM_FEATURE_CRC32
+	ODP_PRINT("%i\n", __ARM_FEATURE_CRC32);
+#else
+	ODP_PRINT("%s\n", ndef);
+#endif
+
+	ODP_PRINT("  __ARM_FEATURE_CRYPTO    ");
+#ifdef __ARM_FEATURE_CRYPTO
+	ODP_PRINT("%i\n", __ARM_FEATURE_CRYPTO);
+#else
+	ODP_PRINT("%s\n", ndef);
+#endif
+
+	ODP_PRINT("  __ARM_FEATURE_DOTPROD   ");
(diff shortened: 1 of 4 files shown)


platform/linux-generic/arch/aarch64/cpu_flags.c
```diff
@@ -8,11 +8,738 @@
...
+		.feat_flag = "FEAT_FP",
+		.hwcap_field = HWCAP_FP,
+		.valid = 1,
+	},
+#else
+	{
+		.feat_flag = "FEAT_FP",
+		.valid = 0,
+	},
+#endif
```
MatiasElo, line 37: These ifdefs could be simplified a bit as members without initializers are automatically zeroed. For example:
```
	{
		/* Floating-point Extensions */
		.feat_flag = "FEAT_FP",
#ifdef HWCAP_FP
		.hwcap_field = HWCAP_FP,
		.valid = 1,
#endif
	},
```

platform/linux-generic/arch/aarch64/cpu_flags.c
```diff
@@ -8,11 +8,738 @@
...
+
+#ifdef HWCAP_ASIMD
```
MatiasElo, line 39: As with compile time flags, it would be easier to check the output if the feature flags would be in alphabetical order.

platform/linux-generic/arch/aarch64/cpu_flags.c
```diff
@@ -395,7 +1122,118 @@ static void _odp_sys_info_print_acle_flags(void)
...
+static void _odp_sys_info_print_hwcap_flags(void)
+{
+	unsigned long hwcaps, hwcaps2;
+	unsigned int size, size2;
+
+	/* Check if the hwcap_flags[] is an empty array. If true, the function
+	 * will return. This is unlikely behaviour. */
+	size = sizeof(hwcap_flags) / sizeof(hwcap_feat_flag_t);
+	if (size == 0)
+		return;
```
MatiasElo, line 1154: I guess  this return and the comment above can now be removed, since we also have the `ARM FEATURES NOT SUPPORTED BY KERNEL:` print and the array is never empty.

platform/linux-generic/arch/aarch64/cpu_flags.c
```diff
@@ -395,7 +1122,118 @@ static void _odp_sys_info_print_acle_flags(void)
...
+			if (hwcaps & 0x01)
+				ODP_PRINT("%s ", hwcap_flags[i].feat_flag);
+			hwcaps = hwcaps >> 1;
+		}
+	}
+
+	/* Check if the hwcap2_flags[] is an empty array. If true, the function
+	 * will print out the unsupported flags for AT_HWCAP and exit. Otherwise,
+	 * print supported flags for AT_HWCAP2.
+	 * CentOS/RedHat systems do not support AT_HWCAP2 flag. */
```
MatiasElo, line 1176: The second sentence is deprecated and I'd assume that `AT_HWCAP2` is supported by newer RedHat releases. Maybe it's easiest to remove the whole comment.

platform/linux-generic/arch/aarch64/cpu_flags.c
Malvika-Gupta: I had written the implementation to be consistent with the Linux hwcap.h file: https://github.com/torvalds/linux/blob/master/arch/arm64/include/uapi/asm/hwcap.h. It makes a lot more sense this way because the array indexing matches exactly with macro definitions, i.e, 5th flag, `HWCAP_PMULL` is at location `hwcap_flags[4] `. 
Let me know what you think. Thanks!

platform/linux-generic/arch/aarch64/cpu_flags.c
Malvika-Gupta: I missed removing that in the code. Thanks Matias!

platform/linux-generic/arch/aarch64/cpu_flags.c
Malvika-Gupta: Looks good. I will make the change. Thanks Matias!

platform/linux-generic/arch/aarch64/cpu_flags.c
Malvika-Gupta: I will remove it. Thanks!

platform/linux-generic/arch/aarch64/cpu_flags.c
Malvika-Gupta: Also, I just saw that the output of `lscpu` is also in the same order as given in the `hwcap.h` file

MatiasElo: Okay, that makes sense.

MatiasElo (approved):

MatiasElo:
Hi @Malvika-Gupta, please add my reviewed-by tags to the commets, so this PR can be merged.

Malvika-Gupta:
> Hi @Malvika-Gupta, please add my reviewed-by tags to the commets, so this PR can be merged.

Added the tags, @MatiasElo. Thanks!

PR #1408, opened 2021-11-02
[PATCH v2] api: ipsec: improve inline IPsec config through inline SA scope

JannePeltonen:
This is WIP for discussion and an alternative to an approach where
the concept of "ipsec instance" is added.

Add a concept of inline SA scope that defines one or more pktios where an
inline SA can be used. The scope can also include look-a-side processing
for cases where an SA state can be shared between inline processing and
async look-a-side processing.

Inline SA scope is added to SA creation parameters. Inline SA scopes
of different SAs can overlap.

Add new capability functions that take inline SA scope as a parameter.

Add new pktio specific inline IPsec config function.

### DIFF
--- a/include/odp/api/spec/ipsec.h
+++ b/include/odp/api/spec/ipsec.h
@@ -123,11 +123,20 @@ typedef struct odp_ipsec_inbound_config_t {
 	 *
 	 *  When inbound SA lookup fails in the asynchronous mode,
 	 *  resulting IPSEC events are enqueued into this queue.
+	 *
+	 *  Ignored by odp_ipsec_inline_config().
 	 */
 	odp_queue_t default_queue;
 
 	/** Constraints for SPI values used with inbound SA lookup. Minimal
-	 *  SPI range and unique values may improve performance. */
+	 *  SPI range and unique values may improve performance.
+	 *
+	 *  In odp_ipsec_config() these fields specify the constraints
+	 *  to lookups done by look-a-side processing.
+	 *
+	 *  In odp_ipsec_inline_config() these fields specify the constraints
+	 *  to lookups done by inline processing in the particular pktio.
+	 */
 	struct {
 		/** Minimum SPI value for SA lookup. Default value is 0. */
 		uint32_t min_spi;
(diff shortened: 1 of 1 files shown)


JannePeltonen:
v2: fixed a couple of typos.

include/odp/api/spec/ipsec.h
```diff
@@ -1164,6 +1324,30 @@ void odp_ipsec_config_init(odp_ipsec_config_t *config);
...
+ *                 done
+ * @param config   Pointer to IPSEC configuration structure
+ *
+ * @retval 0 on success
+ * @retval <0 on failure
+ *
+ * @see odp_ipsec_inline_capability(), odp_ipsec_config_init()
+ */
+int odp_ipsec_inline_config(odp_pktio_t pktio,
+			    const odp_ipsec_config_t *config);
```
nithind1988, line 1349: Some of the contents in odp_ipsec_config_t such as mode is not probably needed.
Can the required inline related config be part of odp_pktio_config_t  instead of this new API ?

include/odp/api/spec/ipsec.h
```diff
@@ -1097,6 +1195,18 @@ typedef struct odp_ipsec_sa_info_t {
...
+/**
+ * Query capabilities for a given inline SA scope
+ *
+ * @param      scope  Inline SA scope
+ * @param[out] capa   Pointer to capability structure for output
+ *
+ * @retval 0 on success
+ * @retval <0 on failure
+ */
+int odp_ipsec_inline_capability(const odp_ipsec_inline_sa_scope_t *scope,
```
nithind1988, line 1207: This capability can be part of odp_pktio_capability_t ?

include/odp/api/spec/ipsec.h
```diff
@@ -870,6 +942,22 @@ typedef struct odp_ipsec_sa_param_t {
...
+		 * depends on the scope parameter.
+		 *
+		 * Default value is false.
+		 */
+		odp_bool_t inline_enable;
+
+		/** Scope where this SA can be used. Ignored if inline
+		 * processing is not enabled. */
+		odp_ipsec_inline_sa_scope_t inline_scope;
+	} inline_param;
```
nithind1988, line 959: This method of a associating PKTIO with Inline SA looks fine to me.

include/odp/api/spec/ipsec.h
JannePeltonen: This capability is associates with the SA scope, which consists of one _or more_ pktios and possibly look-a-side too. The same pktio could be used within different scopes. For instance pktio A could support HMAC-MD5 but pktio B not, then the capabilities for the scope that includes both pktio A and B cannot include HMAC-MD5.

Tying inline capabilities directly to pktios would not allow one to express what is supported when an SA state can be shared between multiple pktios (or a pktio and look-a-side).

TuomasTaipale:
Closed due to inactivity.

PR #1413, opened 2021-11-05
[PATCH v4] api: init: add function for reading current instance handle

MatiasElo:
Add odp_instance() function for reading the current ODP instance handle.

Includes additional minor API changes to fix Doxygen output.

V2:
- Use `memcpm()` in validation test (Petri)

V3:
- Added output parameter for the instance handle

### DIFF
--- a/include/odp/api/spec/init.h
+++ b/include/odp/api/spec/init.h
@@ -1,5 +1,5 @@
 /* Copyright (c) 2013-2018, Linaro Limited
- * Copyright (c) 2019, Nokia
+ * Copyright (c) 2019-2021, Nokia
  * All rights reserved.
  *
  * SPDX-License-Identifier:     BSD-3-Clause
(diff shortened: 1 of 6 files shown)


test/validation/api/init/init_main.c
```diff
@@ -61,6 +62,9 @@ static void init_test_defaults(void)
 	ret = odp_init_local(instance, ODP_THREAD_WORKER);
 	CU_ASSERT_FATAL(ret == 0);
 
+	current_instance = odp_instance();
+	CU_ASSERT(current_instance == instance);
```
psavol, line 66: Instance handle type is implementation specific, and could be also a struct or string (e.g. file path/name). So, better to use memcmp() to check equality.

MatiasElo: Fixed in V2.

psavol (approved):

include/odp/api/spec/init.h
```diff
@@ -355,6 +355,17 @@ int odp_term_global(odp_instance_t instance);
...
+/**
+ * Get instance handle
+ *
+ * Returns the calling thread's ODP instance handle.
+ *
+ * @return Instance handle
+ *
+ * @see odp_init_global(), odp_init_local()
+ */
+odp_instance_t odp_instance(void);
```
psavol, line 367: Agreed to change instance from return value to an output param. So, that global_init() and this call can output the same content into instance object.

MatiasElo: Fixed in V3.

jerinjacobk (approved):

PR #1416, opened 2021-11-05
[PATCH v4] api: add packet vectors to ipsec and clarify packet vector API

JannePeltonen:
The first API patch tweaks the existing pktin packet vector spec so that when packet vectors are enabled, plain packet events will not be generated by the pktin. The second API patch adds packet vector support to packet completion events in the IPsec API.

This PR also adds more default value checks for the pktio related init functions (except lso). There seem to be an inconsistency or an ambiquity in the ODP API regrading the default value of queue mode, which I will try to handle separately.

### DIFF
--- a/include/odp/api/spec/ipsec.h
+++ b/include/odp/api/spec/ipsec.h
@@ -364,6 +364,9 @@ typedef struct odp_ipsec_capability_t {
 	 *  be used for many SAs. */
 	uint32_t max_queues;
 
+	/** Support for returning completion packets as vectors */
+	odp_pktin_vector_capability_t vector;
+
 	/** Maximum anti-replay window size. */
 	uint32_t max_antireplay_ws;
 
@@ -454,6 +457,16 @@ typedef struct odp_ipsec_config_t {
 	 *  @see odp_ipsec_stats(), odp_ipsec_stats_multi()
 	 */
 	odp_bool_t stats_en;
+
+	/**
+	 * Packet vector configuration for async and inline operations
+	 *
+	 * This packet vector configuration affects packets delivered to
+	 * the application through the default queue and the SA destination
+	 * queues. It does not affect packets delivered through pktio
+	 * input queues.
+	 */
+	odp_pktin_vector_config_t vector;
 
 } odp_ipsec_config_t;
 
--- a/include/odp/api/spec/packet_io.h
+++ b/include/odp/api/spec/packet_io.h
@@ -181,7 +181,9 @@ typedef struct odp_pktin_vector_config_t {
 	 *
 	 * When true, packet input vector is enabled and configured with vector
 	 * config parameters. Otherwise, packet input vector configuration
-	 * parameters are ignored.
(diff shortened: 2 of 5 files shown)


include/odp/api/spec/ipsec.h
```diff
@@ -455,6 +458,16 @@ typedef struct odp_ipsec_config_t {
 	 */
 	odp_bool_t stats_en;
 
+	/**
+	 * Packet vector configuration for async and inline operations
```
nithind1988, line 462: >inline operations

For inline, having this config here will conflict with the config in odp_pktin_queue_param_t right ?
I think the one in odp_pktin_queue_param_t together with odp_pktio_config_t::inbound_ipsec should suffice for 
IPSec inline vector support ?

JannePeltonen: Are you thinking about the case where inline processed inbound packets are delivered to the application through the pktin queue? The current API does not allow it as the packets are always delivered through the IPsec destination queues. Even if delivery through pktin was supported, I do not see a conflict since the API text here says that the config applies to the IPsec destination queues and the default queue.

Are you still planning to propose an API change that would allow an implementation to deliver inbound inline completion packets through pktin queues?

include/odp/api/spec/packet_io.h
```diff
@@ -178,7 +178,8 @@ typedef struct odp_pktin_vector_config_t {
 	 *
 	 * When true, packet input vector is enabled and configured with vector
 	 * config parameters. Otherwise, packet input vector configuration
-	 * parameters are ignored.
+	 * parameters are ignored. When vectors are enabled, all packets
+	 * will be delivered as vectors. Default value is false.
```
psavol, line 182: Agreed to document that even when vectors are enabled packets may be also received

JannePeltonen:
v2:
- API: even if packet vectors are enabled, plain packet events can come
- API: packet vectors are disabled by default
- validation checks always accept plain packet events
- l2fwd not modified, will take care of it in a separate PR
- pktio default value checks moved to a different PR

anoobj:
@JannePeltonen The proposal looks good. Are you planning to add some validation test changes also?

JannePeltonen:
> Are you planning to add some validation test changes also?

I was not planning to, at least not in this PR. The current validation tests continue to work without changes, but having additional tests for the vectors would of course be good. But that pretty much requires an implementation of the ipsec vectors in linux-gen, which is also not included in this PR.

psavol (approved):

include/odp/api/spec/ipsec.h
nithind1988: I believe in our case, such a spec change is must as inline decrypted packets will be treated as plain packets and will reach PKTIO PKTIN queue or specific plain packet matching CoS but not destination queue provided in IPSec SA.

anoobj:

JannePeltonen:
v4: rebased and added a reviewed-by tag.

PR #1418, opened 2021-11-08
[PATCH v5] api: pool: clarify odp_pool_param_t.pkt.seg_len documentation

MatiasElo:
Clarify odp_pool_param_t.pkt.seg_len documentation to state that 'seg_len'
is valid only for new packets and that this data starts from
odp_packet_data() (i.e. headroom is not included in this value).


psavol (approved):

ashwinyes:

PR #1419, opened 2021-11-08
[PATCH v5] linux-gen: random: implement crypto random and true random data on x86

JereLeppanen:
```
linux-gen: random: implement crypto random and true random data on x86
    
    Implement crypto random data using rdrand, and true random data using
    rdseed, via compiler builtin functions.
    
    Add architecture-specific file for crypto random and true random
    functions.
    
configure: add option to disable openssl random
    
    Add configure option --disable-openssl-rand, which disables the use of
    OpenSSL RAND_bytes() to generate basic random and crypto random data.
    
test: perf: implement random data throughput test
    
    This test measures megabytes per second throughput of the random data
    generation API functions.
    
linux-gen: README: add note about random data on x86
    
    Add a note about usage of rdrand and rdseed.
    
```

v2:
- Revise comments and add the same information in linux-gen README.

v3:
- Use rdrand for crypto and rdseed for true random.
- Add configure option --disable-openssl-rand.
- Simplify and clean up *_max_kind() functions and kind parameters.
- Revise README.

v4:
- Janne's comments.

v5:
- Fix checkpatch errors.
- Add review tags.

### DIFF
--- a/platform/linux-generic/odp_random_openssl.c
+++ b/platform/linux-generic/odp_random_openssl.c
@@ -7,7 +7,6 @@
 
 #include <odp_posix_extensions.h>
 #include <stdint.h>
-#include <odp/api/random.h>
 #include <odp/autoheader_internal.h>
 #include <odp_init_internal.h>
 #include <odp_random_openssl_internal.h>
@@ -15,37 +14,17 @@
 #if _ODP_OPENSSL
 #include <openssl/rand.h>
 
-odp_random_kind_t _odp_random_openssl_max_kind(void)
-{
-	return ODP_RANDOM_CRYPTO;
-}
-
-int32_t _odp_random_openssl_data(uint8_t *buf, uint32_t len,
-				 odp_random_kind_t kind)
+int32_t _odp_random_openssl_data(uint8_t *buf, uint32_t len)
 {
 	int rc;
 
-	switch (kind) {
-	case ODP_RANDOM_BASIC:
-	case ODP_RANDOM_CRYPTO:
-		rc = RAND_bytes(buf, len);
-		return (1 == rc) ? (int)len /*success*/: -1 /*failure*/;
-
-	case ODP_RANDOM_TRUE:
-	default:
-		return -1;
-	}
+	rc = RAND_bytes(buf, len);
+	return (1 == rc) ? (int)len /*success*/: -1 /*failure*/;
 }
 #else
 /* Dummy functions for building without OpenSSL support */
-odp_random_kind_t _odp_random_openssl_max_kind(void)
-{
-	return ODP_RANDOM_BASIC;
-}
-
 int32_t _odp_random_openssl_data(uint8_t *buf ODP_UNUSED,
-				 uint32_t len ODP_UNUSED,
-				 odp_random_kind_t kind ODP_UNUSED)
+				 uint32_t len ODP_UNUSED)
 {
 	return -1;
 }
--- a/platform/linux-generic/odp_random_openssl.c
+++ b/platform/linux-generic/odp_random_openssl.c
@@ -11,7 +11,7 @@
 #include <odp_init_internal.h>
 #include <odp_random_openssl_internal.h>
 
-#if _ODP_OPENSSL
+#if _ODP_OPENSSL_RAND
 #include <openssl/rand.h>
 
 int32_t _odp_random_openssl_data(uint8_t *buf, uint32_t len)
@@ -28,7 +28,7 @@ int32_t _odp_random_openssl_data(uint8_t *buf ODP_UNUSED,
 {
 	return -1;
 }
-#endif /* _ODP_OPENSSL */
+#endif /* _ODP_OPENSSL_RAND */
 
 int _odp_random_openssl_init_local(void)
 {
--- a/test/performance/odp_random.c
+++ b/test/performance/odp_random.c
@@ -1,0 +1,377 @@
+/* Copyright (c) 2021, Nokia
+ *
+ * All rights reserved.
+ *
+ * SPDX-License-Identifier:     BSD-3-Clause
+ */
+
+#include <stdio.h>
+#include <string.h>
+#include <stdint.h>
+#include <inttypes.h>
+#include <getopt.h>
+
+#include <odp_api.h>
+#include <odp/helper/odph_api.h>
+
+#define MB (1024ull * 1024ull)
+
+/* Command line options */
+typedef struct {
+	int num_threads;
+	uint32_t size;
+	uint32_t rounds;
+} options_t;
+
+static options_t options;
+static const options_t options_def = {
+	.num_threads = 1,
+	.size = 256,
+	.rounds = 100000,
+};
+
+static void print_usage(void)
+{
+	printf("\n"
+	       "random data performance test\n"
+	       "\n"
+	       "Usage: odp_random [options]\n"
+	       "\n"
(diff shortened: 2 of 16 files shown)


Malvika-Gupta:
Hi @JereLeppanen, Thanks for opening this PR; I am also working on adding the true RNG for aarch64 and will use your PR as a base. Since ODP_RANDOM_TRUE implementation was not supported, validation test suite for random APIs have no coverage for it. Although we cannot verify a true random number, we can still add a test to check if the hardware was able to generate a TRN in a reasonable period of time. What do you think about adding such a test? I was thinking of including that in my PR.
Thanks!

JereLeppanen:
Yes, I was thinking the same, but didn't get around to it yet. random_test_get_size() already does this for ODP_RANDOM_BASIC, so all that is needed is to parametrize the random type and test all types up to odp_random_max_kind(). If you want to do this in your PR, please go ahead, it'll be appreciated.

Malvika-Gupta:
> Yes, I was thinking the same, but didn't get around to it yet. random_test_get_size() already does this for ODP_RANDOM_BASIC, so all that is needed is to parametrize the random type and test all types up to odp_random_max_kind(). If you want to do this in your PR, please go ahead, it'll be appreciated.

Thanks Jere, I will add it as a part of my PR

platform/linux-generic/arch/x86/odp/api/abi/random.h
```diff
@@ -0,0 +1,107 @@
...
+{
+	return 1;
+}
+
+#ifdef __RDSEED__
+#define _rd64 __builtin_ia32_rdseed_di_step
+#define _rd32 __builtin_ia32_rdseed_si_step
+#define _rd16 __builtin_ia32_rdseed_hi_step
+#else
+#define _rd64 __builtin_ia32_rdrand64_step
```
JannePeltonen, line 51: To me rdrand does not appear suitable for ODP_RANDOM_TRUE. The Intel document describes how the numbers returned by rdseed have "multiplicative brute-force prediction resistance", which is what I would expect from ODP_RANDOM_TRUE. The document also describes (section 5.2.6) how to derive seeds from rdrand output, which is something that the user of ODP_RANDOM_TRUE data should be be expected to do (as opposed to just using the data directly).

So, I would provide ODP_RANDOM_TRUE only if rdseed is available and use rdrand for ODP_RANDOM_CRYPTO.

platform/linux-generic/odp_random_openssl.c
```diff
@@ -32,6 +36,8 @@ int32_t _odp_random_openssl_data(uint8_t *buf, uint32_t len,
 		return (1 == rc) ? (int)len /*success*/: -1 /*failure*/;
 
 	case ODP_RANDOM_TRUE:
+		return _odp_random_true_data(buf, len);
```
JannePeltonen, line 39: It is a bit inconvenient that this new random data is tied to OpenSSL and is not available if OpenSSL is not compiled in (which could easily be the case in ODP DPDK which can use DPDK crypto instead of OpenSSL crypto).

Maybe things could work so by default ODP_RANDOM_CRYPTO is provided by rdrand and ODP_RANDOM_TRUE by rdseed, and if OpenSSL is used, then ODP_RANDOM_CRYPTO comes from OpenSSL. Or maybe it should be configurable.

JannePeltonen: > Maybe things could work so that by default ODP_RANDOM_CRYPTO is provided by rdrand

It appears that rdrand has as low bandwidth as rdseed in some CPUs so maybe it is not good for ODP_RANDOM_CRYPTO, at least without some configurability.

platform/linux-generic/arch/x86/odp/api/abi/random.h
JereLeppanen: Fixed in v3.

platform/linux-generic/odp_random_openssl.c
JereLeppanen: Configure option added in v3.

platform/linux-generic/arch/default/odp/api/abi/random.h
```diff
@@ -0,0 +1,41 @@
+/* Copyright (c) 2021, Nokia
+ * All rights reserved.
+ *
+ * SPDX-License-Identifier:     BSD-3-Clause
+ */
+
+#ifndef ODP_API_ABI_RANDOM_H_
+#define ODP_API_ABI_RANDOM_H_
```
JannePeltonen, line 8: It seems to me that these new random.h files in the default and x86 archs should not be in the abi directory as they are not part of the abi and not needed for the abi. I think they could as well be in the top level default and x86 arch directories. Or maybe I am missing something?

test/performance/odp_random.c
```diff
@@ -0,0 +1,376 @@
...
+
+	return shm;
+}
+
+static int test_random(void *p)
+{
+	(void)p;
+
+	uint8_t *buf, *data;
+	const unsigned long page = 4 * KB;
```
JannePeltonen, line 125: Maybe this should be ODP_PAGE_SIZE?

test/performance/odp_random.c
```diff
@@ -0,0 +1,376 @@
...
+	odp_time_t start;
+	uint64_t nsec;
+	test_shm_t *shm = shm_lookup();
+
+	if (!shm) {
+		ODPH_ERR("Failed to look up shm %s\n", shm_name);
+		exit(EXIT_FAILURE);
+	}
+
+	/* One extra page for alignment, another one for offset. */
```
JannePeltonen, line 135: > another one for offset

What does this mean? Why is the second extra page needed?

test/performance/odp_random.c
```diff
@@ -0,0 +1,376 @@
...
+	}
+
+	nsec = odp_time_diff_ns(odp_time_local(), start);
+	shm->nsec[odp_thread_id()] = nsec;
+	free(buf);
+
+	return 0;
+}
+
+static int test_random_test(void *p)
```
JannePeltonen, line 172: This is an almost exact copy of test_random(). It would be better to have single function that would take a parameter that tells if the normal or test random data function is to be tested.

JereLeppanen: Should the conditional branch then be in the inner loop, the outer loop, or outside the loops? In the latter two cases, a third of the function would be repeated anyway.

JannePeltonen: In the inner loop (a conditional or call through a function pointer or whatever the exact mechanism is). I would guess the overhead is negligible, but if you are worried about it, then you could force the common function to be always inlined.

JereLeppanen: I'm not, but @psavol was worried about using a function pointer in some other performance test, probably crc. I'm also not worried about repeating a few lines of test code, so anything's fine with me.

test/performance/odp_random.c
JereLeppanen: Fixed in v4.

test/performance/odp_random.c
JereLeppanen: It's not, this is just a copy-paste error. Fixed in v4.

test/performance/odp_random.c
JereLeppanen: I tried the conditional, and in my environment in the worst case (1 byte length test data) the measured throughput decreases by 3.4%. If that's acceptable, I can push the change. I left the code duplication as is for now.

JannePeltonen (approved):

PR #1420, opened 2021-11-09
[PATCH v4] test: sched_perf: add option to use packet pool

MatiasElo:
Add command line option '-p, --pool_type' for using a packet pool instead
of buffer pool for allocating test events.

### DIFF
--- a/test/performance/odp_sched_perf.c
+++ b/test/performance/odp_sched_perf.c
@@ -38,9 +38,11 @@ typedef struct test_options_t {
 	int      num_group;
 	uint32_t num_join;
 	uint32_t max_burst;
+	odp_pool_type_t pool_type;
 	int      queue_type;
 	int      forward;
 	int      fairness;
+	uint32_t event_size;
 	uint32_t queue_size;
 	uint32_t tot_queue;
 	uint32_t tot_event;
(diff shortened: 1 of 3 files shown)


test/performance/odp_sched_perf.c
```diff
@@ -356,12 +372,83 @@ static int set_num_cpu(test_global_t *global)
...
+	pool = odp_pool_create("sched perf", &pool_param);
+
+	if (pool == ODP_POOL_INVALID)
+		ODPH_ERR("Error: buffer pool create failed\n");
+
+	return pool;
+}
+
+static odp_pool_t create_packet_pool(odp_pool_capability_t *pool_capa, uint32_t tot_event,
+				     uint32_t event_size)
```
JereLeppanen, line 412: This is practically the same as create_buffer_pool(). Should these two be combined into one function that takes pool type as a parameter?

MatiasElo: Fixed in V2.

JereLeppanen (approved):

test/performance/odp_sched_perf.c
```diff
@@ -665,16 +697,30 @@ static int create_queues(test_global_t *global)
...
+
+				if (pkt == ODP_PACKET_INVALID) {
+					ODPH_ERR("Error: alloc failed %u/%u\n", i, j);
+					return -1;
+				}
+
+				if (odp_queue_enq(queue, odp_packet_to_event(pkt))) {
+					ODPH_ERR("Error: enqueue failed %u/%u\n", i, j);
+					return -1;
+				}
```
psavol, line 723: There could be single enqueue call (outside of if-else):
```
   ev = odp_packet_to_event(pkt);
}

if (odp_queue_enq(queue, ev)) {
...

```

MatiasElo: Fixed in V3.

PR #1426, opened 2021-11-16
[PATCH v3] api: crypto session and IPsec SA creation parameters can be fully freed afterwards

JannePeltonen:
Clarify the API that session/SA creation parameters, including memory buffers pointed to be pointers in the parameter structure, can be freed after the creation call. This has been the intention of the API already. If an ODP implementation needs to remember some parameters after session creation, it must to copy them. Linux-gen already works this way.

Made crypto API validation test clear the parameters, including key data, after session creation to better catch non-compliant implementations. The PR does not have a similar patch for IPsec API validation tests as it is not as trivial to add there.

### DIFF
--- a/include/odp/api/spec/crypto.h
+++ b/include/odp/api/spec/crypto.h
@@ -1053,6 +1053,9 @@ int odp_crypto_auth_capability(odp_auth_alg_t auth,
  * default values. If call ends up with an error no new session will be
  * created.
  *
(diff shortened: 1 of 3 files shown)


anoobj:

psavol (approved):

JannePeltonen:
v3: rebased and added a reviewed-by tag.

PR #1428, opened 2021-11-19
[PATCH v3] api: clarify default values that were underspecified

JannePeltonen:
Specify that odp_pktin_queue_config() ignores the queue type passed in the parameter struct but deduces the queue type from the pktio input mode. This is the current behaviour in linux-gen and the expected behaviour in the API validation tests.

Specify a few more default values for classifier config structures. The added defaults were already assumed by the API validation tests.

Add validation tests to check the defaults.

This PR assumes that PR 1416 gets merged first (for the vector.enable default value spec).


psavol (approved):

nithind1988:
Reviewed-by: Nithin Dabilpuram<ndabilpuram@marvell.com>

JannePeltonen:
v3: rebased and added a reviewed-by tag.

PR #1429, opened 2021-11-24
[PATCH v3] l2fwd fixes related to packet vectors

JannePeltonen:
- Never send vectors but just plain packets, even in vector mode. Sending vectors is not supported by ODP API.
- Accept plain packet events in addition to vectors in vector mode. ODP api does not promise that only vectors are received in vector mode.
- Reduce duplication of code


JereLeppanen (approved):

MatiasElo (approved):
No measurable performance impact when comparing to my measurement history.

PR #1430, opened 2021-11-25
[PATCH v5] api: timer: add timer pool param init

psavol:
Add timer pool parameter init function which replaces memset(0) as param init. This allows adding new parameters with non-zero default values.


include/odp/api/timer_types.h
```diff
@@ -0,0 +1,32 @@
...
+#define ODP_API_TIMER_TYPES_H_
+
+#ifdef __cplusplus
+extern "C" {
+#endif
+
+#include <odp/api/std_types.h>
+#include <odp/api/abi/event.h>
+#include <odp/api/abi/pool.h>
+#include <odp/api/abi/queue_types.h>
```
MatiasElo, line 23: Could we drop these includes from this level? It would be easier to maintain the required includes in the headers were they are actually required.

include/odp/api/spec/timer.h
```diff
@@ -63,10 +63,20 @@ int odp_timer_capability(odp_timer_clk_src_t clk_src, odp_timer_capability_t *ca
 int odp_timer_res_capability(odp_timer_clk_src_t clk_src,
 			     odp_timer_res_capability_t *res_capa);
 
+/**
+ * Initialize timer pool parameters
+ *
+ * Initialize an odp_timer_pool_param_t to its default values for all fields.
+ *
+ * @param param   Pointer to the odp_timer_pool_param_t structure to be initialized
```
MatiasElo, line 71: Could be marked with `[out]`.

test/validation/api/timer/timer.c
```diff
@@ -408,7 +408,7 @@ static void timer_pool_create_destroy(void)
 	queue = odp_queue_create("timer_queue", &queue_param);
 	CU_ASSERT_FATAL(queue != ODP_QUEUE_INVALID);
 
-	memset(&tparam, 0, sizeof(odp_timer_pool_param_t));
+	odp_timer_pool_param_init(&tparam);
```
MatiasElo, line 411: The validation test changes could be split into a separate commit for clearer git history. It would also be good to add a simple test for checking `odp_timer_pool_param_t` default values.

include/odp/api/timer_types.h
psavol: Fixed in v2

test/validation/api/timer/timer.c
psavol: Both fixed in v2.

include/odp/api/spec/timer.h
psavol: Fixed in v2

MatiasElo (approved):

PR #1431, opened 2021-11-29
[PATCH v1] validation: ipsec: continue tests for pktio create failure

asasidharan:
As the tests would be skipped based on the pktio handle,
return success even when pktio creation fails.


JannePeltonen:
What is the motivation for this change? Is there some problem that this helps to fix? Loop pktio creation should not fail, so the current code looks ok to me.

JannePeltonen:
Is this PR still relevant or can we close it?

asasidharan:
PR is no more relevant.

PR #1435, opened 2021-11-30
[PATCH v5] api: crypto: change hash generation and verification details

JannePeltonen:
Make hash generation and verification spec match linux-gen implementation.
In particular: After verification, the hash field in the packet is left
to undefined value, not the calculated hash value and in generation the
hash field is not cleared before the hash calculation.

### DIFF
--- a/include/odp/api/spec/crypto.h
+++ b/include/odp/api/spec/crypto.h
@@ -1,4 +1,5 @@
 /* Copyright (c) 2014-2018, Linaro Limited
+ * Copyright (c) 2021-2022, Nokia
  * All rights reserved.
  *
  * SPDX-License-Identifier:     BSD-3-Clause
@@ -566,6 +567,16 @@ typedef struct odp_crypto_session_param_t {
 	 */
 	odp_bool_t auth_cipher_text;
 
+	/** Hash result location may overlap authentication range
+	 *
+	 *  This flag indicates that the hash result location may (but is
+	 *  not required to) overlap authentication range. Setting this
+	 *  flag may reduce performance.
+	 *
+	 *  Default value is false.
+	 */
+	odp_bool_t hash_result_in_auth_range;
+
 	/** Preferred sync vs. async for odp_crypto_operation()
 	 *
 	 *  The default value is ODP_CRYPTO_SYNC.
@@ -710,11 +721,17 @@ typedef struct odp_crypto_op_param_t {
 
 	/** Offset from start of packet for hash result
 	 *
-	 *  Specifies the offset where the hash result is to be stored. In case
-	 *  of decode sessions, input hash values will be read from this offset,
-	 *  and overwritten with hash results. If this offset lies within
-	 *  specified 'auth_range', implementation will mute this field before
-	 *  calculating the hash result.
+	 *  In case of decode sessions, the expected hash will be read from
+	 *  this offset and compared with the calculated hash. After the
+	 *  operation the hash bytes will have undefined values.
+	 *
+	 *  In case of encode sessions the calculated hash will be stored in
+	 *  this offset.
+	 *
+	 *  If the hash_result_in_auth_range session parameter is true,
+	 *  the hash result location may overlap auth_range. In that case
+	 *  the result location will be zeroed in decode sessions before
+	 *  hash calculation. Zeroing is not done in encode sessions.
 	 */
 	uint32_t hash_result_offset;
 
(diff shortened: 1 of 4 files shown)


psavol (approved):

anoobj:

JannePeltonen:
v2:
- reworded the API text a bit
- added a hint "hash_result_not_in_auth_range" to enable optimizations in the ODP implementation
- made IPsec implementation set the hint and crypto implementation skip hash result field clearing if the hint is set

There was discussion about getting rid of the possibility of having hash_result_offset within auth_range (and thus the clearing in that case too). Doing that would make it a bit inconvenient for some applications to adapt since they would then have to provide the expected hash outside the auth range but it still would have to be part of packet data and not e.g. in a separate buffer. So I did not make that change this time.

JannePeltonen:
@psavol , @anoobj , could you please recheck this as this has changed a bit.

include/odp/api/spec/crypto.h
```diff
@@ -710,11 +720,17 @@ typedef struct odp_crypto_op_param_t {
...
+	 *  this offset and compared with the calculated hash. After the
+	 *  operation the hash bytes will have undefined values.
+	 *
+	 *  In case of encode sessions the calculated hash will be stored in
+	 *  this offset.
+	 *
+	 *  If the hash_result_not_in_auth_range session parameter is false,
+	 *  the hash result location may overlap auth_range. In that case the
+	 *  result location will be zeroed in decode sessions before hash
+	 *  calculation. Zeroing is not done in encode sessions.
```
psavol, line 733: Since hash_result_not_in_auth_range  applies only for decode sessions, I'd merge this last paragraph into decode and encode paragraphs above. So, that e.g. a person creating an encode session would need to care only on the paragraph starting with "In case of encode sessions..."

include/odp/api/spec/crypto.h
```diff
@@ -566,6 +566,16 @@ typedef struct odp_crypto_session_param_t {
...
 
+	/** Hash result location will not overlap authentication range
+	 *
+	 *  This flag can be set to indicate that the hash result location
+	 *  will not overlap authentication range. Setting this flag may
+	 *  improve performance.
+	 *
+	 *  Default value is false.
+	 */
+	odp_bool_t hash_result_not_in_auth_range;
```
psavol, line 577: This is a bit complex parameter name.  E.g. hash_result_overlap would be shorter and without negation in the name.

How often result would overlap the range? If it's rare, we could select default value not to overlap - although it's backwards incompatible.

JannePeltonen: Hash result field outside of the authenticated range is the more common case, AFAIK.

How about hash_result_in_auth_range? I prefer "in_auth_range" over "overlap" as it is more descriptive?

My logic in the negative form was that if the parameter is set, then it hints that the result field is not in the auth range and otherwise it does not hint anything (may or may not be in the range). But maybe it is better just to describe the parameter well enough that the naming does not matter.

include/odp/api/spec/crypto.h
JannePeltonen: Actually the paragraph, except the middle sentence, applies to encode sessions too. In encode sessions hash result may be in auth range but only if the ...not_in_auth_range parameter is not set. But in that case the zeroing is not done in encode sessions.

JannePeltonen:
v3:
- changed the new session flag to hash_result_in_auth_range
- updated ipsec_crypto example now that the default changed to "not in auth range"
- updated a few copyrights

platform/linux-generic/odp_crypto_openssl.c
```diff
@@ -509,8 +509,9 @@ auth_xcbcmac_check(odp_packet_t pkt,
 	odp_packet_copy_to_mem(pkt, param->hash_result_offset,
 			       bytes, hash_in);
 
-	_odp_packet_set_data(pkt, param->hash_result_offset,
-			     0, bytes);
+	if (odp_unlikely(session->p.hash_result_in_auth_range))
+		_odp_packet_set_data(pkt, param->hash_result_offset,
+				     0, bytes);
```
psavol, line 514: These _odp_packet_set_data() calls in this patch should fit into a single line (<100 chars)

psavol:
API text OK in v3

platform/linux-generic/odp_crypto_openssl.c
JannePeltonen: fixed in v4

JannePeltonen:
v4: minor code formatting changes

anoobj:

JannePeltonen:
v5: rebased, added reviewed-by tags

PR #1437, opened 2021-12-01
[PATCH v5] test: performance: add lock performance test

JereLeppanen:
```
Test lock performance by measuring the time it takes to do a number of
rounds, where a round is a set of counter and lock operations specific
to the lock type. At the end, the resulting counter values are
compared against expected values.

```

v2:
- Number of CPUs default is 10 instead of all.

v3:
- Remove volatile.

v4:
- The rest of Petri's comments.

v5:
- Add review tags.

### DIFF
--- a/test/performance/odp_lock_perf.c
+++ b/test/performance/odp_lock_perf.c
@@ -1,0 +1,663 @@
+/* Copyright (c) 2021, Nokia
+ *
+ * All rights reserved.
+ *
+ * SPDX-License-Identifier:     BSD-3-Clause
+ */
+
+#include <stdio.h>
+#include <string.h>
+#include <stdint.h>
+#include <inttypes.h>
+#include <stdlib.h>
+#include <getopt.h>
+
+#include <odp_api.h>
+#include <odp/helper/odph_api.h>
+
+/* Max number of workers if num_cpu=0 */
+#define DEFAULT_MAX_WORKERS 10
+
+/* Max number of counters */
+#define MAX_COUNTERS 8
+
+#define TEST_INFO(name, test, validate) { name, test, validate }
+
+typedef enum place_t {
+	PLACE_PACK,
+	PLACE_SEPARATE,
+	PLACE_ALL_SEPARATE,
+} place_t;
+
+/* Command line options */
+typedef struct test_options_t {
+	uint32_t num_cpu;
+	uint32_t type;
+	uint64_t num_round;
+	uint32_t num_counter;
+	place_t place;
+} test_options_t;
+
+/* command line options default values */
+static test_options_t test_options_def = {
+	.num_cpu = 0,
+	.type = 0,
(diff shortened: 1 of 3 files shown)


test/performance/odp_lock_perf.c
```diff
@@ -0,0 +1,670 @@
...
+	test_options_t test_options;
+	uint32_t cur_type;
+	odp_barrier_t barrier;
+	odp_cpumask_t cpumask;
+	odph_thread_t thread_tbl[ODP_THREAD_COUNT_MAX];
+	test_thread_ctx_t thread_ctx[ODP_THREAD_COUNT_MAX];
+	struct {
+		struct ODP_ALIGNED_CACHE {
+			odp_spinlock_t lock;
+			volatile uint64_t counter[MAX_COUNTERS];
```
psavol, line 78: Volatile should not be needed. Counters are read/written while holding the lock.

test/performance/odp_lock_perf.c
```diff
@@ -0,0 +1,670 @@
...
+} test_case_t;
+
+static test_global_t *test_global;
+
+static inline void test_spinlock(test_global_t *g, volatile uint64_t **counter,
+				 uint32_t num_counter)
+{
+	odp_spinlock_t *lock = &g->item.spinlock.lock;
+
+	for (uint64_t i = 0; i < (uint64_t)g->test_options.num_round; i++) {
```
psavol, line 118: num_round option could be uint64_t. So that test run time is not limited to 4 billion rounds.

test/performance/odp_lock_perf.c
```diff
@@ -0,0 +1,670 @@
...
+	odp_ticketlock_init(&g->item.ticketlock.lock);
+
+	return 0;
+}
+
+static void fill_counter_ptrs(test_global_t *g, volatile uint64_t **counter_out)
+{
+	test_options_t *test_options = &g->test_options;
+
+	memset(counter_out, 0, sizeof(uint64_t) * MAX_COUNTERS);
```
psavol, line 383: sizeof should be sizeof(uint64_t *) - i.e. memset() array of pointers

test/performance/odp_lock_perf.c
```diff
@@ -0,0 +1,670 @@
...
+	printf("          1        2        3        4        5        6        7        8        9       10");
+
+	for (i = 0; i < ODP_THREAD_COUNT_MAX; i++) {
+		if (global->thread_ctx[i].nsec) {
+			if ((num % 10) == 0)
+				printf("\n   ");
+
+			printf("%8.3f ",
+			       num_round /
+				       (global->thread_ctx[i].nsec / 1000.0));
```
psavol, line 525: If this didn't fit into single (100 chars) line, it would be more readable to use a temp variable for the calculation and then print the variable.

psavol (approved):

PR #1440, opened 2021-12-07
[PATCH v3] validation: atomic: some improvements

JereLeppanen:
```
validation: atomic: break validation function into multiple functions
    
    Break the validation function into multiple separate validation
    functions. This way existing code (the single validation function)
    doesn't need to be changed when adding new tests.

validation: atomic: validate max and min values
    
    Validate the resulting max and min values, instead of just max > min.
    
validation: atomic: add increment/add and decrement/sub tests
    
    Test increments and adds separately from decrements and subs. This is
    in addition to the existing tests, which test that a counter reaches
    its starting value after matching increments and decrements, and after
    matching adds and subs.
    
validation: atomic: test xchg operations with random data
    
    Add a new test, which uses random bytes with xchg operations and
    counts the sums of the input and output values of xchg operations. The
    difference of those sums must be equal to the initial value of the
    atomic variable.

validation: atomic: validate resulting value from cas operations
    
    Validate the resulting value from cas operations, instead of just a
    magic number set at the end of the test.
```

v2:
- Increase test rounds.
- Remove unused global and thread specific data.
- Tuomas' comments.

### DIFF
--- a/test/validation/api/atomic/atomic.c
+++ b/test/validation/api/atomic/atomic.c
@@ -41,54 +41,22 @@ typedef struct {
 	odp_atomic_u32_t a32u_xchg;
 
 	uint32_t g_num_threads;
-	uint32_t g_iterations;
-	uint32_t g_verbose;
 
 	odp_barrier_t global_barrier;
 } global_shared_mem_t;
-
-/* Per-thread memory */
-typedef struct {
-	global_shared_mem_t *global_mem;
-
-	int thread_id;
-	int thread_core;
-
-	volatile_u64_t delay_counter;
-} per_thread_mem_t;
 
 static odp_shm_t global_shm;
 static global_shared_mem_t *global_mem;
 
 /* Initialise per-thread memory */
-static per_thread_mem_t *thread_init(void)
+static void thread_init(void)
 {
 	global_shared_mem_t *global_mem;
-	per_thread_mem_t *per_thread_mem;
 	odp_shm_t global_shm;
-	uint32_t per_thread_mem_len;
-
-	per_thread_mem_len = sizeof(per_thread_mem_t);
-	per_thread_mem = malloc(per_thread_mem_len);
-	memset(per_thread_mem, 0, per_thread_mem_len);
-
-	per_thread_mem->delay_counter = 1;
-
-	per_thread_mem->thread_id = odp_thread_id();
-	per_thread_mem->thread_core = odp_cpu_id();
 
 	global_shm = odp_shm_lookup(GLOBAL_SHM_NAME);
 	global_mem = odp_shm_addr(global_shm);
 	CU_ASSERT_PTR_NOT_NULL(global_mem);
-
-	per_thread_mem->global_mem = global_mem;
-
-	return per_thread_mem;
-}
-
-static void thread_finalize(per_thread_mem_t *per_thread_mem)
-{
-	free(per_thread_mem);
 }
 
 static void test_atomic_inc_32(void)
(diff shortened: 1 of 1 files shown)


test/validation/api/atomic/atomic.c
```diff
@@ -375,46 +371,48 @@ static void test_atomic_xchg_32(void)
...
 	odp_atomic_u64_t *a64u_xchg = &global_mem->a64u_xchg;
+	uint8_t buf[CNT];
+	uint64_t seed = odp_thread_id();
+	uint64_t count_old = 0, count_new = 0;
+
+	odp_random_test_data(buf, CNT, &seed);
 
 	odp_barrier_wait(&global_mem->global_barrier);
 
 	for (i = 0; i < CNT; i++) {
```
TuomasTaipale, line 407: Should the break condition be `i <= CNT` like in `test_atomic_xchg_32()` above?

test/validation/api/atomic/atomic.c
```diff
@@ -26,10 +26,6 @@
```
TuomasTaipale, line 26: Copyright note could be updated.

test/validation/api/atomic/atomic.c
```diff
@@ -691,39 +721,94 @@ static void test_atomic_store(void)
...
-	}
+static void test_atomic_validate_inc_add(void)
+{
+	test_atomic_validate_init_val_128();
 
-	if (check & CHECK_MAX_MIN) {
-		CU_ASSERT(odp_atomic_load_u32(&global_mem->a32u_max) >
-			  odp_atomic_load_u32(&global_mem->a32u_min));
+	/* Two increment tests, one cas increment test and two add tests. */
+	uint64_t total_count = CNT * (3 + 2 * ADD_SUB_CNT) * global_mem->g_num_threads;
```
TuomasTaipale, line 749: `total_count` could be `const`.

test/validation/api/atomic/atomic.c
```diff
@@ -691,39 +721,94 @@ static void test_atomic_store(void)
...
+	CU_ASSERT(U32_INIT_VAL - total_count == odp_atomic_load_u32(&global_mem->a32u));
+	CU_ASSERT(U64_INIT_VAL - total_count == odp_atomic_load_u64(&global_mem->a64u));
+}
+
+static void test_atomic_validate_cas(void)
+{
+	test_atomic_validate_init_val_32_64();
+
+	odp_u128_t a128u = odp_atomic_load_u128(&global_mem->a128u);
+	uint64_t iterations = 0;
```
TuomasTaipale, line 771: Could be already initialized with `a128u.u64[0] - a128u.u64[1]`.

test/validation/api/atomic/atomic.c
```diff
@@ -863,6 +948,32 @@ static int test_atomic_fetch_add_sub_thread(void *arg UNUSED)
 	return CU_get_number_of_failures();
 }
 
+static int test_atomic_inc_add_thread(void *arg UNUSED)
+{
+	per_thread_mem_t *per_thread_mem;
```
TuomasTaipale, line 953: Could be already initialized with `thread_init()`.

TuomasTaipale (changes requested):

test/validation/api/atomic/atomic.c
JereLeppanen: Yes, very good catch. However, the whole setting to zero at the end is unnecessary, so I removed the whole thing and the condition is now i < CNT in both tests.

TuomasTaipale (approved):

psavol (approved):

PR #1442, opened 2021-12-08
[PATCH v4] example: classifier: fix potential memory leaks

TuomasTaipale:
Fix memory leak that would occur if pattern matching rule policy
parsing is stopped due to an error.

Signed-off-by: Tuomas Taipale <tuomas.taipale@nokia.com>

### DIFF
--- a/example/classifier/odp_classifier.c
+++ b/example/classifier/odp_classifier.c
@@ -1,6 +1,6 @@
 /* Copyright (c) 2015-2018, Linaro Limited
- * Copyright (c) 2019-2020, Nokia
- * Copyright (C) 2020, Marvell
+ * Copyright (c) 2019-2021, Nokia
+ * Copyright (c) 2020, Marvell
  * All rights reserved.
  *
  * SPDX-License-Identifier:     BSD-3-Clause
@@ -892,6 +892,10 @@ static int parse_pmr_policy(appl_args_t *appl_args, char *optarg)
 	len = strlen(optarg);
 	len++;
 	pmr_str = malloc(len);
+	if (pmr_str == NULL) {
+		ODPH_ERR("Memory allocation failed\n");
+		return -1;
+	}
 	strcpy(pmr_str, optarg);
 
 	/* PMR TERM */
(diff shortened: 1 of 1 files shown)


example/classifier/odp_classifier.c
```diff
@@ -892,6 +892,8 @@ static int parse_pmr_policy(appl_args_t *appl_args, char *optarg)
 	len = strlen(optarg);
 	len++;
 	pmr_str = malloc(len);
+	if (pmr_str == NULL)
+		return -1;
```
MatiasElo, line 896: An error message would be useful. E.g. `ODPH_ERR("malloc() failed\n");`. Also, the copyright year can be updated for this file.

example/classifier/odp_classifier.c
```diff
@@ -993,6 +995,7 @@ static int parse_pmr_policy(appl_args_t *appl_args, char *optarg)
 		mask_sz = parse_custom(token, stats[policy_count].rule.mask_be, ODPH_ETHADDR_LEN);
 		if (mask_sz != ODPH_ETHADDR_LEN) {
 			ODPH_ERR("Invalid mask. Provide mask without 0x prefix.\n");
+			free(pmr_str);
```
MatiasElo, line 998: Instead of adding several `free()` calls it could be cleaner to add `goto err;` to each error branch and handle the return value and `free()` in one place. The `exit(EXIT_FAILURE);` lines could also be replaced with the same pattern.

example/classifier/odp_classifier.c
```diff
@@ -1075,6 +1086,8 @@ static int parse_policy_ci_pass_count(appl_args_t *appl_args, char *optarg)
 	len = strlen(optarg);
 	len++;
 	count_str = malloc(len);
+	if (count_str == NULL)
+		return -1;
```
MatiasElo, line 1090: Error message could also be added here.

example/classifier/odp_classifier.c
TuomasTaipale: Log added in v2.

example/classifier/odp_classifier.c
TuomasTaipale: Log added in v2.

example/classifier/odp_classifier.c
TuomasTaipale: Error handling with `goto` added.

MatiasElo (approved):

PR #1443, opened 2021-12-08
[PATCH v2] test fixes

gmuthukr:


test/validation/api/crypto/test_vectors_len.h
```diff
@@ -8,7 +8,7 @@
```
JannePeltonen, line 8: > ci: skip_checkformat skip_all
> Change-Id: I360b7f7ec9af1833ca13b4662286bd9a88f61d54

Please remove these from all commit messages.

test/validation/api/crypto/test_vectors_len.h
```diff
@@ -8,7 +8,7 @@
 
 /* Maximum */
 #define MAX_KEY_LEN         64
-#define MAX_IV_LEN          16
+#define MAX_IV_LEN          32
```
JannePeltonen, line 11: Why is this change needed? I cannot see it. The commit message could describe what exactly is the problem being fixed.

example/classifier/odp_classifier.c
```diff
@@ -302,7 +302,7 @@ static odp_pktio_t create_pktio(const char *dev, odp_pool_t pool)
 static int pktio_receive_thread(void *arg)
 {
 	int thr;
-	odp_pktout_queue_t pktout;
+	odp_pktout_queue_t pktout[1] = {0};
```
JannePeltonen, line 305: This is not very nice.

This is almost the same as if you simply added {0} initializer without making the variable an array and without making all the other changes that the change of the type requires.

The problem is that neither code is strictly correct since the ODP application code is not supposed to know what odp_pktout_queue_t actually is and you cannot have an initializer that would work with all possible types.

I wonder your use of the array is a trick to allow both structure types and simple integer and pointer types as odp_pktout_queue_t since the initializer would be valid for both cases (the braces of a structure initializer are optional when inside an array initializer). But such a trick is quite non-obvious and might confuse the reader. This is an example application after all.

I am not sure what the right workaround for the LTO warning is, but I feel that making the code too weird or not stricly ODP API compliant may not be the right way to go. The code could use memset, with a comment that it is only for suppressing bogus warnings, but maybe it would be better if ODP API provided ODP_PKTOUT_QUEUE_INVALID for this kind of initialization purposes.

test/validation/api/init/init_main.c
```diff
@@ -57,12 +57,14 @@ static void init_test_defaults(void)
 
 	odp_init_param_init(&param);
 
+	memset(&instance, 0, sizeof(instance));
 	ret = odp_init_global(&instance, &param, NULL);
```
MatiasElo, line 61: The contents of `odp_instance_t` should be initialized by the `odp_init_global()` implementation.

test/validation/api/init/init_main.c
```diff
@@ -57,12 +57,14 @@ static void init_test_defaults(void)
...
 	CU_ASSERT_FATAL(ret == 0);
 
 	ret = odp_init_local(instance, ODP_THREAD_WORKER);
 	CU_ASSERT_FATAL(ret == 0);
 
+	memset(&current_instance, 0, sizeof(current_instance));
 	CU_ASSERT_FATAL(odp_instance(&current_instance) == 0);
```
MatiasElo, line 68: Same as above.

test/validation/api/init/init_main.c
JannePeltonen: But maybe this code should not fully trust it since this is API validation test. The commit comment should clarify the purpose of this change since now we can only guess.

The memsets would help if odp_instance_t was actually a structure with internal padding because of the memcmp, but maybe that is a bit far-fetched.

MatiasElo: Yeah, I was assuming this fix was done to fix LTO build, but that was just a guess since the commit message is insufficient.

test/validation/api/crypto/test_vectors_len.h
gmuthukr: Ack

test/validation/api/crypto/test_vectors_len.h
gmuthukr: odp/test/validation/api/crypto/test_vectors.h:2465:10: error: excess elements in array initializer [-Werror]
 2465 |          0xff, 0x3f, 0x3f, 0x3f, 0x3f, 0x3f, 0x3f, 0x3f,   
          |          ^~~~                                               

As some test crypto_iv and auth_iv arrays used in test are of size more than MAX_IV_LEN, we need this fix. I have amended commit with better description.

JannePeltonen: > As some test crypto_iv and auth_iv arrays used in test are of size more than MAX_IV_LEN

Which ones? I do not see any that is longer than 16 and if there were, compilation would fail to the same error that you mentioned.

test/validation/api/crypto/test_vectors_len.h
JannePeltonen: This has not been fixed in v2.

example/classifier/odp_classifier.c
JannePeltonen: This has not been fixed in v2. If the initialization is really needed, I would suggest using memset and leaving the rest of the code alone. But is the initialization really needed instead of the underlying ODP implementation needing a fix?

test/validation/api/crypto/test_vectors_len.h
gmuthukr: I am dropping this patch as it was only required for internal patch which is yet to be in upstream. Thanks for the verification.

test/validation/api/crypto/test_vectors_len.h
gmuthukr: This patch was needed for some change we are yet to upstream, sorry about this pop up early. We will revisit it later, for now dropping this change.

example/classifier/odp_classifier.c
gmuthukr: I had relook at our changes and fixed implementation rather with some minor changes. I am revoking this proposal. Thanks for your time reviewing these changes.

PR #1446, opened 2021-12-10
[PATCH v3] example: ipsec_crypto: fix outbound ESP processing to not reuse IVs

JannePeltonen:
Reusing the same ESP IV within the same SA should not be done. Fix the code to not reuse pre-generated IVs but generate a new random IV for every outbound ESP packet.

Somewhat related to this, I am working on a crypto API change proposal to remove per-session IVs altogether (i.e. require an IV for every packet in the per-crypto-operation parameters), as the per-session IVs are rarely, if ever, used as opposed to per-packet IVs. I will submit a separate PR on that later.

Another change in this PR is simple removal of unnecessary tunnel header IP ID randomization after ID wrap-around.

### DIFF
--- a/example/ipsec_crypto/odp_ipsec_cache.c
+++ b/example/ipsec_crypto/odp_ipsec_cache.c
@@ -56,6 +56,7 @@ int create_ipsec_cache_entry(sa_db_entry_t *cipher_sa,
 	odp_crypto_ses_create_err_t ses_create_rc;
 	odp_crypto_session_t session;
 	sa_mode_t mode = IPSEC_SA_MODE_TRANSPORT;
+	uint8_t unused_iv[cipher_sa ? cipher_sa->iv_len : 1];
 
 	/* Verify we have a good entry */
 	entry = &ipsec_cache->array[ipsec_cache->index];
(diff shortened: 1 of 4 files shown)


example/ipsec_crypto/odp_ipsec_cache.c
```diff
@@ -95,7 +95,7 @@ int create_ipsec_cache_entry(sa_db_entry_t *cipher_sa,
 		params.cipher_alg  = cipher_sa->alg.u.cipher;
 		params.cipher_key.data  = cipher_sa->key.data;
 		params.cipher_key.length  = cipher_sa->key.length;
-		params.cipher_iv.data = entry->state.iv;
+		params.cipher_iv.data = NULL;
 		params.cipher_iv.length = cipher_sa->iv_len;
```
TuomasTaipale, line 99: If `data` above is `NULL`, then this probably should be set to 0.

JannePeltonen: I was getting ahead of myself related to the coming (proposed) deprecation of cipher_iv.data in the parameters. In the current version of the API the data cannot be omitted (even if it gets never used and even though linux-gen accepts a null pointer).

Fixed the problem in v2 by setting the IV data pointer point to a dummy array to make the code comply with the API spec even though the IV data in the session will not be used.

TuomasTaipale (approved):

PR #1448, opened 2021-12-15
[PATCH v3] linux-gen: shm: stop listing deprecated ODP_SHM_SW_ONLY flag as supported

MatiasElo:
Stop listing deprecated SHM flag ODP_SHM_SW_ONLY as supported as it is no
longer used by any example applications or tests.

V2:
- Enable usage of `ODP_SHM_SW_ONLY` flag with `--enable-deprecated` configuration option (Janne)

### DIFF
--- a/platform/linux-generic/odp_shared_memory.c
+++ b/platform/linux-generic/odp_shared_memory.c
@@ -8,6 +8,7 @@
 #include <odp_config_internal.h>
 #include <odp_debug_internal.h>
 #include <odp/api/debug.h>
+#include <odp/api/deprecated.h>
 #include <odp/api/std_types.h>
 #include <odp/api/shared_memory.h>
 #include <odp/api/plat/strong_types.h>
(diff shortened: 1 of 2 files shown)


TuomasTaipale (approved):

JereLeppanen (approved):

platform/linux-generic/odp_shared_memory.c
```diff
@@ -17,7 +17,7 @@
 #include <string.h>
 
 /* Supported ODP_SHM_* flags */
-#define SUPPORTED_SHM_FLAGS (ODP_SHM_SW_ONLY | ODP_SHM_PROC | ODP_SHM_SINGLE_VA | ODP_SHM_EXPORT | \
+#define SUPPORTED_SHM_FLAGS (ODP_SHM_PROC | ODP_SHM_SINGLE_VA | ODP_SHM_EXPORT | \
```
JannePeltonen, line 20: If an out-of-tree application uses the flag, this change may break it even if deprecated APIs are enabled. Maybe linux-gen should not remove support for a deprecated API without a good reason. Here ODP_SHM_SW_ONLY could be included in the SHM capabilities if ODP is compiled with the deprecated APIs enabled.

MatiasElo: Fixed in V2.

PR #1449, opened 2021-12-16
[PATCH v7] random data improvements and tests

JereLeppanen:
```
linux-gen: random: improve ODP_RANDOM_BASIC data and random test data
    
    Improve the quality of data generated by and the speed of
    odp_random_test_data() and _odp_random_std_data() (used when OpenSSL
    random is not enabled).
    
validation: random: add test for buffer overflow and alignment problems
    
    Add a new test, which checks that odp_random_data() and
    odp_random_test_data() do not modify bytes outside the given buffer.
    
validation: random: test randomness
    
    Test that random data looks random by adding four new tests:
    
    random_test_frequency checks that random data is uniformly
    distributed.
    
    random_test_independence checks that bytes are independent.
    
    random_test_runs checks that the number of runs (uninterrupted
    sequence of identical bits) is what is expected of random data.
    
    random_test_matrix_rank checks that the ranks of 32x32 binary matrices
    correspond to expected probabilities.
    
```
v2:
- Tuomas' comments.

v3:
- Rebase.
- Remove API change (separate it into its own PR).

v4:
- Add review tags.

v6:
- Rebase.
- Remove commit "validation: chksum: use odp_random_test_data()".
- Petri's comments.
- Add review tags.

### DIFF
--- a/platform/linux-generic/odp_random_std.c
+++ b/platform/linux-generic/odp_random_std.c
@@ -4,52 +4,88 @@
  * SPDX-License-Identifier:     BSD-3-Clause
  */
 
-#include <odp_posix_extensions.h>
-#include <stdint.h>
-#include <stdlib.h>
 #include <odp/api/byteorder.h>
 #include <odp/api/cpu.h>
 #include <odp/api/debug.h>
 #include <odp_init_internal.h>
 #include <odp_random_std_internal.h>
+#include <odp_cpu.h>
 
+#include <stdint.h>
 #include <time.h>
 
-/* Assume at least two rand bytes are available and RAND_MAX is power of two - 1 */
-ODP_STATIC_ASSERT(RAND_MAX >= UINT16_MAX, "RAND_MAX too small");
-ODP_STATIC_ASSERT((RAND_MAX & (RAND_MAX + 1ULL))  ==  0, "RAND_MAX not power of two - 1");
-
-static int32_t _random_data(uint8_t *buf, uint32_t len, uint32_t *seed)
+/*
+ * Xorshift64*, adapted from [1], and modified to return only the high 32 bits.
+ *
+ * [1] An experimental exploration of Marsaglia's xorshift generators, scrambled
+ *     Sebastiano Vigna, July 2016.
+ *     http://vigna.di.unimi.it/ftp/papers/xorshift.pdf
+ */
+static inline uint32_t xorshift64s32(uint64_t *x)
 {
-	union {
-		uint32_t rand_word;
-		uint8_t rand_byte[4];
-	} u;
-	uint32_t i = 0, j, k;
+	/* The variable x should be initialized to a nonzero seed. [1] */
+	if (!*x)
+		/*
+		 * 2^64 / phi. As far away as possible from any small integer
+		 * fractions, which the caller might be likely to use for the
+		 * next seed after 0.
+		 */
+		*x = 11400714819323198485ull;
 
-	while (i < len) {
-		u.rand_word = rand_r(seed);
+	*x ^= *x >> 12; /* a */
+	*x ^= *x << 25; /* b */
+	*x ^= *x >> 27; /* c */
+	return (*x * 2685821657736338717ull) >> 32;
+}
 
-		/* Use two least significant bytes */
-		j = ODP_LITTLE_ENDIAN ? 0 : 2;
-		for (k = 0; k < 2 && i < len; i++, j++, k++)
-			*buf++ = u.rand_byte[j];
+static int32_t _random_data(uint8_t *buf, uint32_t len, uint64_t *seed)
+{
+	const uint32_t ret = len;
+
+	if (!_ODP_UNALIGNED && ((uintptr_t)buf & 3) && len) {
+		uint32_t r = xorshift64s32(seed);
+
+		if ((uintptr_t)buf & 1) {
+			*(uint8_t *)(uintptr_t)buf = r & 0xff;
+			r >>= 8;
+			buf += 1;
+			len -= 1;
+		}
+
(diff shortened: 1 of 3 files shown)


platform/linux-generic/odp_random_std.c
```diff
@@ -4,52 +4,88 @@
...
+			*(uint16_t *)(uintptr_t)buf = r & 0xffff;
+			r >>= 16;
+			buf += 2;
+		}
+
+		if (len & 1)
+			*(uint8_t *)(uintptr_t)buf = r & 0xff;
+	}
+
+	return ret;
```
TuomasTaipale, line 80: Is `memcpy()` too slow here (would handle alignment issues and there would be less code)?

test/validation/api/random/random.c
```diff
@@ -80,10 +80,370 @@ static void random_test_repeat(void)
...
+ * Perform randomness tests on this and higher types of random data. To test
+ * also odp_random_test_data(), set MIN_KIND to -1.
+ */
+#define MIN_KIND ODP_RANDOM_CRYPTO
+
+/*
+ * Alpha for P-value tests. This does not affect the tests that use a
+ * precomputed critical value.
+ */
+const double alpha = 0.00000001;
```
TuomasTaipale, line 105: This could be `static`.

test/validation/api/random/random.c
```diff
@@ -80,10 +80,370 @@ static void random_test_repeat(void)
...
+			r |= buf[bit / 8];
+			bit += 8;
+			i += 8;
+			continue;
+		}
+		/* Single bit. */
+		r <<= 1;
+		r |= (buf[bit / 8] >> (7 - (bit & 7))) & 1;
+		bit++;
+		i++;
```
TuomasTaipale, line 160: Pre-increment is typically slightly faster as there's no need for the implicit temporary variable.

test/validation/api/random/random.c
```diff
@@ -80,10 +80,370 @@ static void random_test_repeat(void)
...
+				chisq += dif * dif / expected;
+			}
+
+			printf("%s ; bits %d ; chisq %g ; df %u ; crit %g ; ",
+			       kname(kind), bits, chisq, cells - 1, crit);
+
+			if (chisq < crit)
+				printf("pass\n");
+			else
+				printf("FAIL\n");
```
TuomasTaipale, line 223: A bit less code:
```c
printf("%s\n", chisq < crit ? "pass" : "FAIL");
```

platform/linux-generic/odp_random_std.c
JereLeppanen: Unfortunately memcpy() is quite a lot slower. In the worst case, around 20% lower throughput in the performance test (in my environment).

test/validation/api/random/random.c
JereLeppanen: When it's a basic type, the unused temporary is usually optimized away completely, and the resulting code from pre-increment and post-increment is exactly the same. At least gcc does this even with -O0. The convention seems to be to use post-increment unless there's a specific reason to use pre-increment. I can get behind that, because pre-increment looks ugly.

test/validation/api/random/random.c
JereLeppanen: I put this in a function to boot, so that the strings appear in just one place.

test/validation/api/random/random.c
TuomasTaipale: OK. Well, it's trusting that compiler makes it fast vs. making it fast by default. In any case, doesn't probably matter here.

TuomasTaipale (approved):

test/validation/api/random/random.c
```diff
@@ -80,10 +80,357 @@ static void random_test_repeat(void)
...
+	const double critical[] = {
+		32.8413, 40.1300, 50.8129, 68.0293,
+		97.0285, 147.463, 237.614, 402.685,
+		711.187, 1297.50, 2426.64, 4623.37,
+		8929.74, 17419.3, 34224.0, 67587.1,
+	};
+
+	printf("\n\n");
+
+	for (int kind = MIN_KIND; kind <= (int)odp_random_max_kind(); kind++) {
```
psavol, line 205: I'd prefer random kinds to be tested in separate test cases. Implementation may have different implementation for different kinds, and some of those could fail randomness test more often than others. As separate cases, it would be easier to track which kinds were tested (which kind failed a case) when those are separate cases.

platform/linux-generic/odp_random_std.c
```diff
@@ -4,52 +4,88 @@
...
+			*(uint8_t *)(uintptr_t)buf = r & 0xff;
+			r >>= 8;
+			buf += 1;
+			len -= 1;
+		}
+
+		if ((uintptr_t)buf & 2) {
+			*(uint16_t *)(uintptr_t)buf = r & 0xffff;
+			buf += 2;
+			len -= 2;
```
psavol, line 58: Len underflows here if application requests a small number of bytes (e.g. len = 1).

A test case could be added to test various small lengths (and unaligned buf pointers).

test/validation/api/random/random.c
```diff
@@ -80,10 +80,84 @@ static void random_test_repeat(void)
...
+				(int32_t)len);
+	}
+}
+
+static void random_test_align_and_overflow(odp_random_kind_t kind)
+{
+	uint8_t ODP_ALIGNED_CACHE buf[64];
+
+	for (int align = 8; align < 16; align++) {
+		for (int len = 0; len < 8; len++) {
```
psavol, line 109: API does not specify min/max len, but to me it would be an application error if it requests 0 bytes. So, I start validation test from len = 1.

psavol (approved):

PR #1450, opened 2021-12-17
[PATCH v8] implement true random number generation for aarch64

Malvika-Gupta:


platform/linux-generic/arch/aarch64/odp_random.h
```diff
@@ -0,0 +1,74 @@
...
+		return RNDR_OUTPUT_LEN;
+	}
+#endif
+
+	return _odp_random_true_data_generic(buf, len);
+}
+
+static inline int32_t _odp_random_crypto_data(uint8_t *buf, uint32_t len)
+{
+	return _odp_random_crypto_data_generic(buf, len);
```
MatiasElo, line 67: From `odp_random_max_kind()` spec: `Implementations support the returned max kind and all kinds weaker than it.` So if `ODP_RANDOM_TRUE` is supported, `ODP_RANDOM_CRYPTO` should also be supported. The same `_odp_rndr()` function could be used for the implementation.

platform/linux-generic/arch/aarch64/odp_random.h
```diff
@@ -0,0 +1,74 @@
...
+}
+
+static inline int32_t _odp_random_true_data(uint8_t *buf, uint32_t len)
+{
+#ifdef __ARM_FEATURE_RNG
+	if (_ODP_OPENSSL_RAND) {
+		while (!_odp_rndr((unsigned long *)buf))
+			;
+
+		return RNDR_OUTPUT_LEN;
```
MatiasElo, line 58: The function should at least try to generate `len` bytes of random data.

platform/linux-generic/arch/aarch64/odp_random.h
```diff
@@ -0,0 +1,74 @@
...
+
+#define RNDR_OUTPUT_LEN      8
+
+odp_random_kind_t _odp_random_max_kind_generic(void);
+int32_t _odp_random_true_data_generic(uint8_t *buf, uint32_t len);
+int32_t _odp_random_crypto_data_generic(uint8_t *buf, uint32_t len);
+
+static inline int _odp_random_max_kind(void)
+{
+	if (_ODP_OPENSSL_RAND) {
```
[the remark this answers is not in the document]
Malvika-Gupta: Hi @MatiasElo 
Armv8.5-a RNG extensions support 2 registers, `RNDR & RNDRRS`, both of which return TRUE random numbers according to the [specification](https://developer.arm.com/documentation/ddi0595/2021-06/AArch64-Registers/RNDR--Random-Number). There is no crypto random number generator equivalent in Arm ISA as far as I know. 
As you point out in a later comment, ODP spec states that `Implementations support the returned max kind and all kinds weaker than it.` If OpenSSL is disabled in the current setup for architecture specific random implementation, the highest supported random kind will become `ODP_RANDOM_BASIC`. `ODP_RANDOM_CRYPTO` cannot be supported without OpenSSL and `ODP_RANDOM_TRU`E may be supported but the ODP spec imposes a limitation since `ODP_RANDOM_CRYPTO`. 
So, what changes should be done to support Armv8.5-a RNG? I see two options at the moment:
1. Change the ODP to remove the limitation for max kind to support all weaker kinds
2. Change the current setup for random to architecture specific random implementation and OpenSSL concurrently

Let me know what you think.

platform/linux-generic/arch/aarch64/odp_random.h
```diff
@@ -0,0 +1,74 @@
...
+			 :
+			 : "cc");
+
+	return pass;
+}
+
+static inline int32_t _odp_random_true_data(uint8_t *buf, uint32_t len)
+{
+#ifdef __ARM_FEATURE_RNG
+	if (_ODP_OPENSSL_RAND) {
```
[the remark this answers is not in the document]
Malvika-Gupta: Please check the comment above

platform/linux-generic/arch/aarch64/odp_random.h
Malvika-Gupta: Please check the comment above

platform/linux-generic/arch/aarch64/odp_random.h
Malvika-Gupta: RNDR registers gives 64-bit random data so we had fixed the return length to be the same. I am not exactly clear on what change should I do here. Should I add a check for len and RNDR_OUTPUT_LEN or should the change be something else?

platform/linux-generic/arch/aarch64/odp_random.h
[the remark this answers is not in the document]
MatiasElo: My understanding of the spec is that it's okay to use the higher lever random implementation also for the lower level(s). It's not a problem if the returned data is "too random".

platform/linux-generic/arch/aarch64/odp_random.h
MatiasElo: An application may request for example only 1 byte of random data, so you cannot write directly to the output `buf`. To make application usage easier you could add a loop here and write 64 bits at a time until `len` bytes have been written.

platform/linux-generic/arch/aarch64/odp_random.h
JereLeppanen: Or, similarly to what we decided on x86 with RDRAND and RDSEED, since RNDRRS is seeded every time from TRNG and RNDR only occasionally, RNDR should probably be used only for ODP_RANDOM_CRYPTO, and RNDRRS for ODP_RANDOM_TRUE.

Malvika-Gupta: Hi @JereLeppanen, this thought had occurred to me to as well and I did a bit of research on it to see if RNDR ~= RDRAND and RNDRRS ~= RDSEED for the purpose of this PR. I found this discussion on the LKML which exactly discusses if this association is valid: https://lore.kernel.org/linux-arm-kernel/6e75d7b9-1c30-adab-bb74-1aaaa4e98ad4@linaro.org/
If you still think that we should proceed with RNDR for `ODP_RANDOM_CRYPTO`, and RNDRRS for `ODP_RANDOM_TRUE`, I'll make the changes in v2

platform/linux-generic/arch/aarch64/odp_random.h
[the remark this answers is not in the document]
Malvika-Gupta: So, according to the spec, I can still use the Armv8.5-a RNDR extension to generate a random number but the return type for `_odp_random_max_kind()` would be `ODP_RANDOM_BASIC` when using AArch64 specific implementation since we can upgrade a random data request. Is my understanding correct?
@MatiasElo, can you please check the comment by myself @JereLeppanen below? There are different approaches to resolve this and depending on how we decide to proceed with the implementation, I'll make the changes. Thanks!

platform/linux-generic/arch/aarch64/odp_random.h
Malvika-Gupta: Okay, I'll make this change in v2. Thanks!

platform/linux-generic/arch/aarch64/odp_random.h
JereLeppanen: Yes, I saw that, and it looks like that discussion was never fully resolved, at least not explicitly. As far as I can tell, the "RNDR ~= RDRAND and RNDRRS ~= RDSEED" comment in that thread is correct.

Malvika-Gupta: Okay, let me make these changes in v2. Thanks Jere!

platform/linux-generic/arch/aarch64/odp_random.h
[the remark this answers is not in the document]
MatiasElo: I agree with Jere. So, `RNDR` would be used for `ODP_RANDOM_CRYPTO` and `RNDRRS` for `ODP_RANDOM_TRUE`.

test/validation/api/random/random.c
```diff
@@ -18,22 +20,48 @@ static void random_test_get_size(void)
...
+	uint32_t bytes;
 	uint8_t buf[32];
-
-	do {
-		ret = odp_random_data(buf + bytes, sizeof(buf) - bytes,
-				      ODP_RANDOM_BASIC);
-		bytes += ret;
-		if (ret < 0 || bytes >= sizeof(buf))
+	odp_random_kind_t kind;
+	int max_kind = odp_random_max_kind();
```
MatiasElo, line 27: Instead of testing all random kinds inside the same function, random kind could be made a function argument like in the new tests Jere added:
```
	ODP_TEST_INFO_CONDITIONAL(random_test_align_and_overflow_basic, check_kind_basic),
	ODP_TEST_INFO_CONDITIONAL(random_test_align_and_overflow_crypto, check_kind_crypto),
	ODP_TEST_INFO_CONDITIONAL(random_test_align_and_overflow_true, check_kind_true),
```

This way it's easier to notice skipped/failed tests.

MatiasElo:
Hi @Malvika-Gupta,

Do you happen to know which GCC version is required to use this? I tried 10.3 and 11.1 and neither defined `__ARM_FEATURE_RNG` with `-march=armv8.5-a`.

Malvika-Gupta:
> Hi @Malvika-Gupta,
> 
> Do you happen to know which GCC version is required to use this? I tried 10.3 and 11.1 and neither defined `__ARM_FEATURE_RNG` with `-march=armv8.5-a`.
> 
> Edit. The solution was to use `-march=armv8.5-a+rng`.

Yes, that's correct. RNG is an optional extension to the Armv8.5-a so you have to explicitly add it as a configuration option with `-march`

platform/linux-generic/arch/aarch64/odp_random.h
```diff
@@ -0,0 +1,105 @@
+/* Copyright (c) ARM Limited
```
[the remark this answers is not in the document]
Malvika-Gupta: Missed that. I will add it in v3. Thanks!

platform/linux-generic/arch/aarch64/odp_random.h
```diff
@@ -0,0 +1,105 @@
...
+#ifdef __cplusplus
+extern "C" {
+#endif
+
+#include <odp/api/spec/random.h>
+#include <odp/autoheader_internal.h>
+
+#include <stdint.h>
+
+#define RNDR_OUTPUT_LEN      8
```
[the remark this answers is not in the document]
Malvika-Gupta: I will remove it. Thanks!

platform/linux-generic/arch/aarch64/odp_random.h
```diff
@@ -0,0 +1,105 @@
...
+			 : "=r" (*v), [pass] "=r" (pass)
+			 :
+			 : "cc");
+
+	return pass;
+}
+
+static inline int32_t _odp_random_crypto_data(uint8_t *buf, uint32_t len)
+{
+	for (uint32_t i = 0; i < len / 8; i++) {
```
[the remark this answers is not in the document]
Malvika-Gupta: I saw how this is being handled in x86 files; if the `len` is smaller than 64-bits, then for the remainder bits are handled by another built-in function which generates either 32/16/8 bits of random data. But, if Arm has support for generating only 64-bit random number. 
I can generate an error for the non 8-byte lengths and redirect it to use the generic version instead. Or, we can round up `len` to a factor of 8, generate 8 bytes of random data till that upper limit and return the truncated `len` bytes only.
I am not sure how should this is to be handled so please suggest any alternate approaches. Thanks Matias!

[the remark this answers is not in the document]
JannePeltonen: _odp_random_crypto_data() should return the requested number of random bytes, so make it behave so. Returning an error if len is not divisible by 8 is not the right approach.

If your underlying primitive provides random data 8 bytes at a time, then you just have to throw some of the bytes away if len is not divisible by 8.

MatiasElo:
FYI the validation tests are currently passing for `ODP_RANDOM_CRYPTO`, but `ODP_RANDOM_TRUE` tests are taking extremely long. We'll have to consult Jere about this when he returns from his vacation next week.

platform/linux-generic/arch/aarch64/odp_random.h
```diff
@@ -0,0 +1,172 @@
...
+		uint64_t temp;
+
+		while (!_odp_rndr(&temp))
+			;
+		*((uint64_t *)buf) = temp;
```
JereLeppanen, line 76: buf may be unaligned. One possibility would be to test for __ARM_FEATURE_UNALIGNED in addition to __ARM_FEATURE_RNG, i.e. `#if defined(__ARM_FEATURE_RNG) && defined(__ARM_FEATURE_UNALIGNED)` at the top of the file. Maybe @psavol can comment if that's sufficient here.

psavol: I guess __ARM_FEATURE_UNALIGNED should be supported on ARMv8 always. So, it would be sufficient to just stop the build if RNG is supported, but UNALIGNED not.

```
#ifdef __ARM_FEATURE_RNG

#if __ARM_FEATURE_UNALIGNED != 1
#error Unaligned access not supported
#endif
```

Malvika-Gupta: Sounds good, I will make the change. Thanks!

MatiasElo:
Hi,

The issues I commented about earlier have now been fixed, but the issues pointed out by Jere (alignment and non-optimal implementation) are not.

Malvika-Gupta:
> Hi,
> 
> The issues I commented about earlier have now been fixed, but the issues pointed out by Jere (alignment and non-optimal implementation) are not.

I had made the changes but just realized that I forgot to commit them before pushing so accidently pushed the previous patch set again. I will update it. Thanks for bringing this to my notice, Matias!

platform/linux-generic/arch/aarch64/odp_random.h
```diff
@@ -0,0 +1,158 @@
...
+	if (len & 7) {
+		uint64_t temp;
+
+		while (!_odp_rndr(&temp))
+			;
+
+		if (len & 4) {
+			*(uint32_t *)(uintptr_t)buf = temp & 0xffffffff;
+			buf += 4;
+		}
```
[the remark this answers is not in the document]
Malvika-Gupta: Makes sense. I will make this change. Thanks Petri!

platform/linux-generic/arch/aarch64/odp_random.h
```diff
@@ -0,0 +1,162 @@
...
+#if __ARM_FEATURE_UNALIGNED != 1
+#error Unaligned access not supported
+#endif
+
+static inline int _odp_random_max_kind(void)
+{
+	return ODP_RANDOM_TRUE;
+}
+
+static inline int _odp_rndr(unsigned long *v)
```
psavol, line 34: Type of the pointer should be uint64_t. The instruction outputs 64bits, but in armv8 long type may be either 64 or 32 bits depending if LP64 or LLP64 model is used. Maybe long is in practice always 64bits, but usage of it adds confusion about output data size/align. Remember to change also (unsigned long *) casts to uint64_t cast under.

_odp_rndrrs() prototype needs the same change.

platform/linux-generic/arch/aarch64/odp_random.h
```diff
@@ -0,0 +1,162 @@
...
+			 :
+			 : "cc");
+
+	return pass;
+}
+
+static inline int32_t _odp_random_crypto_data(uint8_t *buf, uint32_t len)
+{
+	for (uint32_t i = 0; i < len / 8; i++) {
+		while (!_odp_rndr((unsigned long *)buf))
```
psavol, line 76: It would be safer to use uint64_t temp variable also in this case, and then store temp into *buf (possibly unaligned).

Performance should be the same, as MRS moves the random value first into a register anyway and then register value is stored into memory (*buf). Compiler likely handles temp as a register.

In the current form, the code looks like as if MRS would store the value directly into the memory, which it does not do:
"MRS: Move System Register allows the PE to read an AArch64 System register into a general-purpose register."

Also, the cast is not needed with temp.

platform/linux-generic/arch/aarch64/odp_random.h
Malvika-Gupta: This point was something even me and Govind had discussed and we had concluded after some research that `unsigned long` will most likely end up being 64-bits. But, to avoid any ambiguity, I will make the change to use `uint64_t` instead.

platform/linux-generic/arch/aarch64/odp_random.h
```diff
@@ -0,0 +1,170 @@
...
+		}
+
+		if (len & 2) {
+			*(uint16_t *)(uintptr_t)buf = temp & 0xffff;
+			temp >>= 16;
+			buf += 2;
+		}
+
+		if (len & 1)
+			*(uint8_t *)(uintptr_t)buf = temp & 0xff;
```
psavol, line 104: No need to cast as buf is uint8_t pointer.

platform/linux-generic/arch/aarch64/odp_random.h
```diff
@@ -0,0 +1,170 @@
...
+			 :
+			 : "cc");
+
+	return pass;
+}
+
+static inline int32_t _odp_random_crypto_data(uint8_t *buf, uint32_t len)
+{
+	for (uint32_t i = 0; i < len / 8; i++) {
+		uint64_t temp;
```
psavol, line 76: Please introduce temp variable once in the beginning of the function. Code is cleaner when the same variable is not introduced twice and couple of lines are saved.

psavol:
Reviewed v7. Only two minor comments (extra cast and duplicate variable introduction), please fix those for both functions (crypto_data and true_data). Otherwise OK. Please, add my review-by to the next version.

platform/linux-generic/arch/aarch64/odp_random.h
Malvika-Gupta: Okay, I will make the change. Thanks!

platform/linux-generic/arch/aarch64/odp_random.h
Malvika-Gupta: Sounds good, I will make the change. Thanks!

platform/linux-generic/arch/aarch64/odp_random.h
```diff
@@ -0,0 +1,170 @@
...
+#include <stdint.h>
+
+odp_random_kind_t _odp_random_max_kind_generic(void);
+int32_t _odp_random_true_data_generic(uint8_t *buf, uint32_t len);
+int32_t _odp_random_crypto_data_generic(uint8_t *buf, uint32_t len);
+
+#ifdef __ARM_FEATURE_RNG
+
+#if __ARM_FEATURE_UNALIGNED != 1
+#error Unaligned access not supported
```
JereLeppanen, line 26: This error message may be a little confusing. Maybe somethign like "This implementation requires unaligned access".

Malvika-Gupta: I will make the change. Thanks!

Malvika-Gupta:
Hi @psavol I have addressed all the pending comments and added the Reviewed-by tags too. Please check. Thanks!

psavol (approved):

JereLeppanen (approved):

MatiasElo (approved):

PR #1453, opened 2021-12-21
[PATCH v2] linux-gen: sched: update random data table

psavol:
Random data table had same values multiple times. Update table
with suffled values from 0 to 127. This way random data has even
distribution of values, which results even distribution of
spread modulos.


MatiasElo (approved):
One typo in commit message: `suffled` -> `shuffled`.

PR #1454, opened 2021-12-21
[PATCH v7] api: crypto: deprecate IV data in session parameters

JannePeltonen:
Reusing the same initialization vector (IV) with the same key is almost always wrong and results in weakened security. Therefore the ability of specifying IV in crypto session creation and reusing it in multiple crypto operations is of questionable value. It complicates both the API and implementations and may encourage incorrect IV reuse (see PR 1446).

This PR deprecates odp_crypto_iv_t type (which holds IV length and IV data) and the cipher_iv and auth_iv parameters of odp_crypto_session_param_t. New session parameters, cipher_iv_len and auth_iv_len, are introduced for setting IV lengths in the session. From now on, IV data must always be passed through the existing cipher_iv_ptr and auth_iv_ptr parameters of odp_crypto_op_param_t and odp_crypto_packet_op_param_t.

### DIFF
--- a/include/odp/api/spec/crypto.h
+++ b/include/odp/api/spec/crypto.h
@@ -525,6 +525,8 @@ typedef struct odp_crypto_key {
 
 /**
  * Crypto API IV structure
+ *
+ * @deprecated
  */
 typedef struct odp_crypto_iv {
 	/** IV data
@@ -532,7 +534,7 @@ typedef struct odp_crypto_iv {
 	 *  Ignored when length is zero. Null value indicates that an
 	 *  IV will be provided for each packet through the crypto
 	 *  operation parameters. In that case the per-operation
-	 *  IV override parameter must always point to a valid IV.
+	 *  IV parameter must always point to a valid IV.
 	 *
 	 *  Default value is NULL.
 	 */
@@ -541,7 +543,7 @@ typedef struct odp_crypto_iv {
 	/** IV length in bytes. Default value is zero. */
 	uint32_t length;
 
-} odp_crypto_iv_t;
+} ODP_DEPRECATE(odp_crypto_iv_t);
 
 /**
  * Crypto API data range specifier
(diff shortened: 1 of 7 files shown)


JannePeltonen:
v2:
- Rebased
- Got rid of doxygen warnings
- Fixed the description of odp_crypto_packet_op_param_t::cipher_iv_ptr
- Added a patch that updates crypto user guide

JannePeltonen:
v3: rebased

anoobj:
@JannePeltonen There is a typo in the patch,

validation: crypto: use the new way of setting crypto IVs 

This series looks good. Won't this PR be split into multiple PRs to allow for slow deprecation? Can you describe the plan?

JannePeltonen:
v4: fixed the typo and rebased

JannePeltonen:
> This series looks good. Won't this PR be split into multiple PRs to allow for slow deprecation? Can you describe the plan?

I am not planning for any such split. I might have mentioned at some point that the changes done in this PR are split to rather many patches in order to do the deprecation gradually within the set of patches so that every revision of the tree still compiles and passes the CI tests. But once this PR gets merged, the IV-in-session stuff deprecation is complete.

Or do you see a need for a more gradual approach (meaning that the actual deprecation would not be done yet but at a later date)? I would think it is not needed since any possible usage of the per-session-IV api is basically a security bug that should be fixed ASAP anyway.

anoobj:
> Or do you see a need for a more gradual approach (meaning that the actual deprecation would not be done yet but at a later date)?

No. I was just wondering if you had something like that planned.

JannePeltonen:
v5: added reviewed-by tags.

include/odp/api/spec/crypto.h
```diff
@@ -1,4 +1,5 @@
 /* Copyright (c) 2014-2018, Linaro Limited
+ * Copyright (c) 2021, Nokia
```
psavol, line 2: Update year to 2022

include/odp/api/spec/crypto.h
```diff
@@ -611,6 +611,17 @@ typedef struct odp_crypto_session_param_t {
 
 		/** Cipher Initialization Vector (IV) */
 		odp_crypto_iv_t cipher_iv;
+
+		/** Cipher IV length */
+		struct {
+			/** Unused padding field */
+			uint8_t *dummy_padding_0;
```
psavol, line 627: Comment could mention that intention is to remove this field when iv and cipher_iv above are removed.

JannePeltonen: Agreed not to change the comment since the last API patch makes the field part of the deprecated API (i.e. not normally visible).

include/odp/api/spec/crypto.h
JannePeltonen: Keeping it as 2021 since the code in this PR was published in 2021 and not changed in 2022.

JannePeltonen:
v6: rebased and added more reviewed-by tags

psavol (approved):

JannePeltonen:
v7: rebased

PR #1459, opened 2021-12-29
[PATCH v5] test: performance: add DMA performance test

TuomasTaipale:
ODP DMA API performance is tested with a scenario where a configurable
amount of input segments are transferred to a single large output
segment configurable amount of times. Both synchronous and asynchronous
transfers can be tested.

Current implementation supports packet-type transfers, plain buffer
based support will be added later.

v2:
- fix miscellaneous style issues

v3:
- fix transfer and completion parameter init usage

v4:
- year bump
- check event completion pool capability
- allocate individual event for every event-completion transfer

v5:
- add reviewed-by tag

### DIFF
--- a/test/performance/odp_dma_perf.c
+++ b/test/performance/odp_dma_perf.c
@@ -1,0 +1,889 @@
+/* Copyright (c) 2021-2022, Nokia
+ *
+ * All rights reserved.
+ *
+ * SPDX-License-Identifier:     BSD-3-Clause
+ */
+
+#ifndef _GNU_SOURCE
+#define _GNU_SOURCE
+#endif
+
+#include <stdio.h>
+#include <stdlib.h>
+#include <string.h>
+#include <inttypes.h>
+
+#include <odp_api.h>
+#include <odp/helper/odph_api.h>
+
+#define DEFAULT_SEG_SIZE 1024U
+#define ROUNDS 1000000
+#define COMPL_DELIMITER ","
+/* For now, a static maximum amount of input segments */
+#define MAX_NUM_IN_SEGS 64
+
+#define TRS_TYPE_SYNC 0
+#define TRS_TYPE_ASYNC 1
+
+#define GRN_ALL 0
+#define GRN_IND 1
+
+#define PKT_TYPE 0
+#define BUF_TYPE 1
+
+#define COMPL_MODE_POLL 0
+#define COMPL_MODE_EVENT 1
+
+#define GIGAS 1000000000
+#define MEGAS 1000000
+#define KILOS 1000
+
+typedef struct test_config_t {
+	int trs_type;
+	int trs_grn;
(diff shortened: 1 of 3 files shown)


test/performance/odp_dma_perf.c
```diff
@@ -0,0 +1,876 @@
...
+	if (config->trs_grn == GRN_IND && capa.max_segs < 2U) {
+		ODPH_ERR("Unsupported total amount of segments.\n");
+		return -1;
+	}
+
+	if (config->num_in_seg * config->seg_size > capa.max_seg_len) {
+		ODPH_ERR("Unsupported total segment size.\n");
+		return -1;
+	}
+
```
MatiasElo, line 352: `odp_dma_capability_t.pool.max_num` could also be checked (see comment below about completion events).

test/performance/odp_dma_perf.c
```diff
@@ -0,0 +1,876 @@
...
+			compl_params[i].transfer_id =
+				odp_dma_transfer_id_alloc(config->dma_config.handle);
+
+			if (compl_params[i].transfer_id == ODP_DMA_TRANSFER_ID_INVALID) {
+				ODPH_ERR("Error creating DMA transfer ID.\n");
+				return -1;
+			}
+		}
+
+		compl_params[i].event = odp_dma_compl_to_event(config->dma_config.compl_ev);
```
MatiasElo, line 632: If I understood the code correctly, it seems that multiple simultaneous async transfers may now share the same completion events (e.g. `odp_dma_perf -t 1 -g 1 -i 4 -m 1,1,1,1`). This is not allowed by the API.

TuomasTaipale: Yes, this is the case. Where is it stated that this is not allowed? (The command runs and completes fine.)

MatiasElo: This could perhaps be stated more clearly also in the DMA API, but the general ODP principle is that once an event is passed to scheduling, the caller loses access to the event handle. Only after receiving the event again, for example from `odp_schedule()`, can it be used again. Also, the same event can currently only exist once in scheduling/queues.

test/performance/odp_dma_perf.c
```diff
@@ -0,0 +1,876 @@
+/* Copyright (c) 2021, Nokia
```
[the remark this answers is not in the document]
TuomasTaipale: Fixed in v4.

test/performance/odp_dma_perf.c
TuomasTaipale: Fixed in v4.

test/performance/odp_dma_perf.c
TuomasTaipale: OK, fixed in v4.

MatiasElo (approved):

PR #1461, opened 2022-01-03
[PATCH v3] example: timer_accuracy: close output file on failure

MatiasElo:
V2:
- Updated copyright year
- Print error string in case of `fopen()` failure

### DIFF
--- a/example/timer/odp_timer_accuracy.c
+++ b/example/timer/odp_timer_accuracy.c
@@ -1,10 +1,11 @@
 /* Copyright (c) 2018, Linaro Limited
- * Copyright (c) 2019-2021, Nokia
+ * Copyright (c) 2019-2022, Nokia
  * All rights reserved.
  *
  * SPDX-License-Identifier:     BSD-3-Clause
  */
 
+#include <errno.h>
 #include <stdio.h>
 #include <string.h>
 #include <inttypes.h>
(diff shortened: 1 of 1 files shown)


example/timer/odp_timer_accuracy.c
```diff
@@ -717,15 +717,6 @@ int main(int argc, char *argv[])
 	if (parse_options(argc, argv, &test_global))
```
TuomasTaipale, line 717: 2022 could be updated in the copyright.

example/timer/odp_timer_accuracy.c
```diff
@@ -767,11 +759,20 @@ int main(int argc, char *argv[])
 		void *addr;
 		uint64_t size = test_global.tot_timers * sizeof(test_log_t);
 
+		test_global.file = fopen(test_global.filename, "w");
+		if (test_global.file == NULL) {
+			printf("Failed to open output file: %s\n",
```
TuomasTaipale, line 764: The actual error could also be printed (e.g. `perror()` or `strerror()`).

example/timer/odp_timer_accuracy.c
MatiasElo: Fixed in V2.

example/timer/odp_timer_accuracy.c
MatiasElo: Fixed in V2.

TuomasTaipale (approved):

PR #1462, opened 2022-01-03
[PATCH v4] validation: packet: add new test for user area testing

ashwinyes:
## linux-gen: pool: reduce max uarea size to 2048 bytes

Failures were seen in CI when a packet pool was being created with
the existing maximum uarea size of 10MB.

So reduce maximum user area size to 2048 bytes.

## validation: packet: add new test for user area testing

Add a new test which creates multiple pools with different user area
sizes and validates that memset of the user area does not affect the
packet validity.

### DIFF
--- a/test/validation/api/packet/packet.c
+++ b/test/validation/api/packet/packet.c
@@ -3271,6 +3271,52 @@ static void packet_test_max_pools(void)
 		CU_ASSERT(odp_pool_destroy(pool[i]) == 0);
 }
 
+static void packet_test_user_area(void)
+{
+	odp_pool_param_t param;
+	odp_packet_t pkt;
+	odp_pool_t pool;
+
+	memcpy(&param, &default_param, sizeof(odp_pool_param_t));
+
+	param.pkt.uarea_size = 0;
+	pool = odp_pool_create("zero_uarea", &param);
+	CU_ASSERT_FATAL(pool != ODP_POOL_INVALID);
+	pkt = odp_packet_alloc(pool, param.pkt.len);
+	CU_ASSERT_FATAL(pkt != ODP_PACKET_INVALID);
+	CU_ASSERT(odp_packet_user_area(pkt) == NULL);
+	CU_ASSERT(odp_packet_user_area_size(pkt) == 0);
+	odp_packet_free(pkt);
+	CU_ASSERT(odp_pool_destroy(pool) == 0);
+
+	param.pkt.uarea_size = 1;
+	pool = odp_pool_create("one_uarea", &param);
+	CU_ASSERT_FATAL(pool != ODP_POOL_INVALID);
+	pkt = odp_packet_alloc(pool, param.pkt.len);
+	CU_ASSERT_FATAL(pkt != ODP_PACKET_INVALID);
+	CU_ASSERT_FATAL(odp_packet_user_area(pkt) != NULL);
+	CU_ASSERT(odp_packet_user_area_size(pkt) == 1);
+	*(char *)odp_packet_user_area(pkt) = 0;
+	CU_ASSERT_FATAL(odp_packet_is_valid(pkt) == 1);
+	odp_packet_free(pkt);
+	CU_ASSERT(odp_pool_destroy(pool) == 0);
+
+	if (pool_capa.pkt.max_uarea_size)
+		param.pkt.uarea_size = pool_capa.pkt.max_uarea_size;
+	else
+		param.pkt.uarea_size = 512;
+	pool = odp_pool_create("max_uarea", &param);
+	CU_ASSERT_FATAL(pool != ODP_POOL_INVALID);
+	pkt = odp_packet_alloc(pool, param.pkt.len);
+	CU_ASSERT_FATAL(pkt != ODP_PACKET_INVALID);
+	CU_ASSERT_FATAL(odp_packet_user_area(pkt) != NULL);
+	CU_ASSERT(odp_packet_user_area_size(pkt) == param.pkt.uarea_size);
+	memset(odp_packet_user_area(pkt), 0, param.pkt.uarea_size);
+	CU_ASSERT_FATAL(odp_packet_is_valid(pkt) == 1);
+	odp_packet_free(pkt);
+	CU_ASSERT(odp_pool_destroy(pool) == 0);
+}
+
 static int packet_parse_suite_init(void)
 {
 	int num_test_pkt, i;
@@ -4260,6 +4306,7 @@ odp_testinfo_t packet_suite[] = {
 	ODP_TEST_INFO(packet_test_offset),
 	ODP_TEST_INFO(packet_test_ref),
 	ODP_TEST_INFO(packet_test_max_pools),
+	ODP_TEST_INFO(packet_test_user_area),
 	ODP_TEST_INFO_NULL,
 };
 
--- a/platform/linux-generic/odp_pool.c
+++ b/platform/linux-generic/odp_pool.c
@@ -47,6 +47,9 @@
 /* Define a practical limit for contiguous memory allocations */
 #define MAX_SIZE   (10 * 1024 * 1024)
 
+/* Maximum packet user area size */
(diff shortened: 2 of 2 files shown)


test/validation/api/packet/packet.c
```diff
@@ -3271,6 +3271,53 @@ static void packet_test_max_pools(void)
...
+
+	memcpy(&param, &default_param, sizeof(odp_pool_param_t));
+
+	param.pkt.uarea_size = 0;
+	pool = odp_pool_create("zero_uarea", &param);
+	CU_ASSERT_FATAL(pool != ODP_POOL_INVALID);
+	pkt = odp_packet_alloc(pool, param.pkt.len);
+	CU_ASSERT_FATAL(pkt != ODP_PACKET_INVALID);
+	CU_ASSERT_FATAL(odp_packet_user_area(pkt) == NULL);
+	CU_ASSERT_FATAL(odp_packet_user_area_size(pkt) == 0);
```
MatiasElo, line 3288: These checks don't have to be fatal. Same comment also applies to several checks below.

test/validation/api/packet/packet.c
```diff
@@ -3271,6 +3271,53 @@ static void packet_test_max_pools(void)
...
+	CU_ASSERT_FATAL(odp_packet_user_area(pkt) != NULL);
+	CU_ASSERT_FATAL(odp_packet_user_area_size(pkt) == pool_capa.pkt.max_uarea_size);
+	memset(odp_packet_user_area(pkt), 0, pool_capa.pkt.max_uarea_size);
+	CU_ASSERT_FATAL(odp_packet_is_valid(pkt) == 1);
+	odp_packet_free(pkt);
+	CU_ASSERT(odp_pool_destroy(pool) == 0);
+
+	param.pkt.uarea_size = pool_capa.pkt.max_uarea_size + 1;
+	pool = odp_pool_create("invalid_uarea", &param);
+	CU_ASSERT_FATAL(pool == ODP_POOL_INVALID);
```
MatiasElo, line 3318: Using invalid parameter value leads to undefined behavior, so this should not be tested.

test/validation/api/packet/packet.c
```diff
@@ -3271,6 +3271,53 @@ static void packet_test_max_pools(void)
...
+	CU_ASSERT(odp_pool_destroy(pool) == 0);
+
+	param.pkt.uarea_size = 1;
+	pool = odp_pool_create("one_uarea", &param);
+	CU_ASSERT_FATAL(pool != ODP_POOL_INVALID);
+	pkt = odp_packet_alloc(pool, param.pkt.len);
+	CU_ASSERT_FATAL(pkt != ODP_PACKET_INVALID);
+	CU_ASSERT_FATAL(odp_packet_user_area(pkt) != NULL);
+	CU_ASSERT_FATAL(odp_packet_user_area_size(pkt) == 1);
+	memset(odp_packet_user_area(pkt), 0, 1);
```
MatiasElo, line 3299: Please make checkpatch happy :)

test/validation/api/packet/packet.c
```diff
@@ -3271,6 +3271,53 @@ static void packet_test_max_pools(void)
...
+	CU_ASSERT_FATAL(odp_packet_is_valid(pkt) == 1);
+	odp_packet_free(pkt);
+	CU_ASSERT(odp_pool_destroy(pool) == 0);
+
+	param.pkt.uarea_size = pool_capa.pkt.max_uarea_size;
```
MatiasElo, line 3304: Linux-generic sets `pool_capa.pkt.max_uarea_size` to 10MB, which likely causes the CI failures. We should probably set this to a more reasonable value (e.g. 512B).

MatiasElo: `pool_capa.pkt.max_uarea_size` may also be zero ("limited only by the available memory size"). In this case you should default to some reasonable value (e.g. 512B).

ashwinyes: This is meant to test the max user area setting is indeed supported and doesn't cause issues. Basically in Marvell implementation, user area is part of packet buffer, so memset of max uarea size should not be overwriting the packet data. This was the intention of this test. Setting to some fixed value other than max uarea size defeats the intention.

For linux generic, if CI fails with 10MB of user area, isn't the linux generic implementation broken ?

I take your suggestion for the zero case. If zero is given, we should use some fixed sane value.

test/validation/api/packet/packet.c
ashwinyes: Will fix.

test/validation/api/packet/packet.c
ashwinyes: Will Fix.

test/validation/api/packet/packet.c
ashwinyes: Will fix.

test/validation/api/packet/packet.c
MatiasElo: > This is meant to test the max user area setting is indeed supported and doesn't cause issues. Basically in Marvell implementation, user area is part of packet buffer, so memset of max uarea size should not be overwriting the packet data. This was the intention of this test. Setting to some fixed value other than max uarea size defeats the intention. 

This is fine. The problem was only the case where `max_uarea_size` was zero, but this is now fixed in V2.

> For linux generic, if CI fails with 10MB of user area, isn't the linux generic implementation broken ?

It works according to the spec ("...limited only by the available memory size for the pool."). So, either we must limit the test to some reasonable value (e.g. 1024B) or lower the `max_uarea_size` of linux-generic. I'd prefer limiting the linux-generic `max_uarea_size`, but this may have effects to applications. What do you think @psavol?

MatiasElo: We talked with Petri about this and ended up deciding that setting `max_uarea_size` to 2048 bytes would be the best solution. So please add another commit (before this one) to this PR, which sets pool and ext pool `max_uarea_size` to 2048.
E.g.
```
diff --git a/platform/linux-generic/odp_pool.c b/platform/linux-generic/odp_pool.c
index ca696fc96..9ef178e5b 100644
--- a/platform/linux-generic/odp_pool.c
+++ b/platform/linux-generic/odp_pool.c
@@ -47,6 +47,9 @@
 /* Define a practical limit for contiguous memory allocations */
 #define MAX_SIZE   (10 * 1024 * 1024)

+/* Maximum packet user area size */
+#define MAX_UAREA_SIZE 2048
+
 ODP_STATIC_ASSERT(CONFIG_PACKET_SEG_LEN_MIN >= 256,
                  "ODP Segment size must be a minimum of 256 bytes");

@@ -1388,7 +1391,7 @@ int odp_pool_capability(odp_pool_capability_t *capa)
        capa->pkt.max_segs_per_pkt = PKT_MAX_SEGS;
        capa->pkt.min_seg_len      = CONFIG_PACKET_SEG_LEN_MIN;
        capa->pkt.max_seg_len      = max_seg_len;
-       capa->pkt.max_uarea_size   = MAX_SIZE;
+       capa->pkt.max_uarea_size   = MAX_UAREA_SIZE;
        capa->pkt.min_cache_size   = 0;
        capa->pkt.max_cache_size   = CONFIG_POOL_CACHE_MAX_SIZE;
        capa->pkt.stats.all = supported_stats.all;
@@ -1682,7 +1685,7 @@ int odp_pool_ext_capability(odp_pool_type_t type, odp_pool_ext_capability_t *cap
        capa->pkt.max_headroom        = CONFIG_PACKET_HEADROOM;
        capa->pkt.max_headroom_size   = CONFIG_PACKET_HEADROOM;
        capa->pkt.max_segs_per_pkt    = PKT_MAX_SEGS;
-       capa->pkt.max_uarea_size      = MAX_SIZE;
+       capa->pkt.max_uarea_size      = MAX_UAREA_SIZE;

        return 0;
 }
```

ashwinyes: Sent v3 with the suggested fixes.

MatiasElo (approved):
Looks good.

PR #1464, opened 2022-01-04
timer: fix timer pool destroy race

psavol:
Fix a race condition between timer pool destroy and timer pool scan. When a timer pool is destroyed, other threads may be still scanning it.


platform/linux-generic/odp_timer.c
```diff
@@ -298,31 +308,59 @@ static odp_timer_pool_t timer_pool_new(const char *name,
 		return ODP_TIMER_POOL_INVALID;
 	}
 
+	/* Find timer pool that has not been used for a while, or is used least recently.
+	 * This ensures that inline scan of an old timer pool has completed and it's memory
```
MatiasElo, line 312: Typo: `it's`. Copyright year can also be updated.

platform/linux-generic/odp_timer.c
```diff
@@ -395,20 +436,25 @@ static odp_timer_pool_t timer_pool_new(const char *name,
...
 	}
 
+	/* Update the highest index for inline timer scan */
+	if (tp_idx > timer_global->highest_tp_idx)
+		timer_global->highest_tp_idx = tp_idx;
+
 	odp_ticketlock_unlock(&timer_global->lock);
 
-	if (!odp_global_rw->inline_timers) {
-		if (tp->param.clk_src == ODP_CLOCK_DEFAULT)
```
MatiasElo, line 401: The `odp_timer_pool_param_t.clk_src` is never checked in `odp_timer_pool_create()`. Using a wrong clock source doesn't seem to have any effect, but it would still be good to detect it.

test/validation/api/timer/timer.c
```diff
@@ -478,6 +480,77 @@ static void timer_pool_create_destroy(void)
...
+	tp_param.min_tmo    = tmo_ns / 2;
+	tp_param.max_tmo    = tmo_ns;
+	tp_param.num_timers = 1;
+
+	for (i = 0; i < num; i++) {
+		tp[i] = odp_timer_pool_create("test_max", &tp_param);
+		if (tp[i] == ODP_TIMER_POOL_INVALID)
+			ODPH_ERR("Timer pool create failed: %u / %u\n", i, num);
+
+		CU_ASSERT_FATAL(tp[i] != ODP_TIMER_POOL_INVALID);
```
TuomasTaipale, line 531: Could these be combined to a `CU_FAIL_FATAL()` (no formatting though then)?

platform/linux-generic/odp_timer.c
psavol: Fixed in v2

platform/linux-generic/odp_timer.c
psavol: Fixed in v2

test/validation/api/timer/timer.c
psavol: The error print is there for helping to understand which iteration failed, so format is needed. Typically, failure happens close to the max capa.

Usually, I don't use CU_FAIL, since it buffers the output and prints at the end of the test (if test did not crash).

TuomasTaipale (approved):

MatiasElo (approved):

PR #1465, opened 2022-01-10
[PATCH v2] test: performance: dma_perf: add plain memory-type transfer support

TuomasTaipale:
### test: performance: dma_perf: add plain memory-type transfer support

In addition to packet-type data, shared memory backed plain memory-type
data can now also be configured to be used as the test data.

### test: performance: dma_perf: improve error handling

Instead of always exiting when an error occurs, try cleaning up first
when possible.

v2:
- improve function names

### DIFF
--- a/test/performance/odp_dma_perf.c
+++ b/test/performance/odp_dma_perf.c
@@ -363,9 +363,12 @@ static int check_capabilities(test_config_t *config)
 	return 0;
 }
 
-static int configure_packet_pool(test_config_t *config)
+static int configure_packets(test_config_t *config)
 {
 	odp_pool_param_t param;
+
+	for (int i = 0; i < config->num_in_seg + 1; ++i)
+		config->seg_config.pkts[i] = ODP_PACKET_INVALID;
 
 	odp_pool_param_init(&param);
 	param.type = ODP_POOL_PACKET;
@@ -421,7 +424,7 @@ static int populate_packets(test_config_t *config)
 
 static int setup_packet_segments(test_config_t *config)
 {
-	return configure_packet_pool(config) ||
+	return configure_packets(config) ||
 	       allocate_packets(config) ||
 	       populate_packets(config);
 }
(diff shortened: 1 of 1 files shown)


test/performance/odp_dma_perf.c
```diff
@@ -150,6 +162,7 @@ static void print_usage(void)
 	       "  -T, --in_seg_type         Input segment data type. Packet by default.\n"
 	       "                            Types:\n"
 	       "                                0: packet\n"
+	       "                                1: buffer\n"
```
MatiasElo, line 165: Perhaps buffer is a bit misleading name. We may actually end up adding support for performing DMA operations with buffers (`odp_buffer_t`) in the future. For example `memory` or `memory address` would be a clearer name and it would better match the API spec. The same comment applies to function naming below.

test/performance/odp_dma_perf.c
```diff
@@ -130,6 +141,7 @@ static void print_usage(void)
 	       "    odp_dma_perf\n"
 	       "    odp_dma_perf -t 0 -g 1 -i 2\n"
 	       "    odp_dma_perf -t 1 -g 1 -i 4 -m 0,0,0,0\n"
+	       "    odp_dma_perf -t 1 -g 1 -i 7 -m 0,0,0,0,0,0,1 -T 1 -r 1000 -s 2048\n"
```
MatiasElo, line 144: It would be good to run the application with both data types in the CI. `test/performance/odp_sched_perf_run.sh` is one example of how this can be done.

test/performance/odp_dma_perf.c
TuomasTaipale: Yes, those are better names, will be fixed in v2.

test/performance/odp_dma_perf.c
TuomasTaipale: Sure. Could be a separate PR though.

MatiasElo (approved):

PR #1466, opened 2022-01-10
[PATCH v5] api: crypto: API clarifications and new session creation error codes

JannePeltonen:
api: crypto: add session creation error codes for unsupported combos

    Add crypto session creation error codes for cases where the requested
    combination of algorithms, order of algorithms or other combination
    of creation parameters is not supported. The new errors are for cases
    where a limitation of an ODP implementation cannot be expressed
    through crypto capabilities.

    api: crypto: clarify auth_range parameter with AEAD algorithms

    AEAD algorithms authenticate the plaintext that is to be encrypted
    plus additional authenticated data (AAD). Clarify that in case of AEAD
    algorithms the auth_range parameter is not used, except with AES-GMAC.

    AES-GMAC, as defined, does not take any plaintext to be encrypted
    as input but only AAD. In ODP API the aad_ptr parameter is not used
    but the AAD is provided through auth_range, allowing the data be located
    as non-contiguous pieces in multiple packet segments. Clarify the API
    text to not sound as if ODP did not use AAD as defined in the GMAC
    algorithm specification.

### DIFF
--- a/example/ipsec_crypto/odp_ipsec_cache.c
+++ b/example/ipsec_crypto/odp_ipsec_cache.c
@@ -119,7 +119,7 @@ int create_ipsec_cache_entry(sa_db_entry_t *cipher_sa,
 	/* Synchronous session create for now */
 	if (odp_crypto_session_create(&params, &session, &ses_create_rc))
 		return -1;
-	if (ODP_CRYPTO_SES_CREATE_ERR_NONE != ses_create_rc)
+	if (ODP_CRYPTO_SES_ERR_NONE != ses_create_rc)
 		return -1;
 
 	/* Copy remainder */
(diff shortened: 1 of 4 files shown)


JannePeltonen:
v2: rebased and added validation test patch to skip tests with unsupported algorithm combinations.

JannePeltonen:
v3: Shortened the existing and new crypto session error names as suggested by @psavol.

anoobj:
This series looks good to me.

psavol (approved):

JannePeltonen:
v5: rebased, removed copyright update from API (conflicts with an earlier PR) and updated copyright in the validation test file.

PR #1467, opened 2022-01-10
[PATCH v2] validation: pool: check odp_pool_param_init() output

MatiasElo:
Test that default odp_pool_param_t values are set correctly by
odp_pool_param_init() function.

### DIFF
--- a/test/validation/api/pool/pool.c
+++ b/test/validation/api/pool/pool.c
@@ -1,6 +1,6 @@
 /* Copyright (c) 2014-2018, Linaro Limited
  * Copyright (c) 2020, Marvell
- * Copyright (c) 2020-2021, Nokia
+ * Copyright (c) 2020-2022, Nokia
  * All rights reserved.
  *
  * SPDX-License-Identifier:	BSD-3-Clause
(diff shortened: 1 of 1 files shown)


test/validation/api/pool/pool.c
```diff
@@ -42,6 +42,27 @@ static odp_pool_capability_t global_pool_capa;
 static odp_pool_param_t default_pool_param;
 static odp_pool_ext_capability_t global_pool_ext_capa;
 
+static void pool_test_param_init(void)
+{
+	odp_pool_param_t param;
```
TuomasTaipale, line 47: `odp_pool_param_t` spec documents default values for a bunch of different fields. Should all those default values (which `odp_pool_param_init()` should init to) be verified here?

JannePeltonen: Maybe the API is slightly ambiguous here but the way I read it there are "default values" to be set by the init function and then there are defaults that odp_pool_create() will use in various situations, e.g. when the parameter value is zero. In the latter case the "default value" set by the init function is not necessarily defined.

At quick look, the patch seems to verify all fields that have well enough defined default values that can be verified. Is there a specific field which could still be verified but which I do not see?

MatiasElo: Yep, as Janne stated I only added tests to the fields which have defined default values.

psavol (approved):

PR #1469, opened 2022-01-12
[PATCH v2] linux-gen: sysinfo: fix odp_sys_huge_page_size_all() return value order

MatiasElo:
odp_sys_huge_page_size_all() function implementation uses qsort() to sort
the system huge page sizes into ascending order. qsort() comparison
function returns an integer and with large page sizes (e.g. 16G) the return
value could overflow with the previous implementation.

### DIFF
--- a/platform/linux-generic/odp_system_info.c
+++ b/platform/linux-generic/odp_system_info.c
@@ -471,7 +471,14 @@ uint64_t odp_sys_huge_page_size(void)
 
 static int pagesz_compare(const void *pagesz1, const void *pagesz2)
 {
-	return (*(const uint64_t *)pagesz1 - *(const uint64_t *)pagesz2);
+	const uint64_t val1 = *(const uint64_t *)pagesz1;
+	const uint64_t val2 = *(const uint64_t *)pagesz2;
+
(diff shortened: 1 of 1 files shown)


psavol (approved):

platform/linux-generic/odp_system_info.c
```diff
@@ -471,7 +471,14 @@ uint64_t odp_sys_huge_page_size(void)
 
 static int pagesz_compare(const void *pagesz1, const void *pagesz2)
 {
-	return (*(const uint64_t *)pagesz1 - *(const uint64_t *)pagesz2);
+	uint64_t val1 = *(const uint64_t *)pagesz1;
+	uint64_t val2 = *(const uint64_t *)pagesz2;
+
+	if (val1 < val2)
+		return -1;
+	else if (val1 > val2)
```
TuomasTaipale, line 479: `else` is redundant. `val1`/`2` could be `const`.

MatiasElo: Fixed in V2.

TuomasTaipale (approved):

PR #1470, opened 2022-01-12
[PATCH v3] test: random: use less random data

JereLeppanen:
```
validation: random: use less random data
    
    True random generators may be slow, so use much less random data in
    the randomness tests so that the tests don't take too much time.
    
    random_test_independence_true test is removed, while some other tests
    are scaled down. With this change, the randomness tests use about 51
    MB of crypto random data and about 527 KB of true random data.
    
test: perf: random: divide test rounds by 100 for ODP_RANDOM_TRUE
    
    True random data may be slow, so divide the test length by 100 for
    ODP_RANDOM_TRUE.
    
```
v2:
- Rebase.
- Matias' comments.

v3:
- Add review tags,

### DIFF
--- a/test/validation/api/random/random.c
+++ b/test/validation/api/random/random.c
@@ -1,4 +1,5 @@
 /* Copyright (c) 2015-2018, Linaro Limited
+ * Copyright (c) 2021-2022, Nokia
  * All rights reserved.
  *
  * SPDX-License-Identifier:     BSD-3-Clause
@@ -206,7 +207,7 @@ static const char *res_str(int pass)
 static void random_test_frequency(odp_random_kind_t kind)
 {
 	/* Mean number of hits per cell. */
-	const uint32_t expected = 100;
+	const uint32_t expected = 50;
 
 	/* From LibreOffice CHISQ.INV.RT(0.00000001; df). */
 	const double critical[] = {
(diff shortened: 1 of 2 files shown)


test/validation/api/random/random.c
```diff
@@ -218,10 +218,10 @@ static void random_test_frequency(odp_random_kind_t kind)
 
 	printf("\n\n");
 
-	for (int bits = 1; bits <= 16; bits++) {
+	for (int bits = 1; bits <= 8; bits++) {
 		const uint32_t cells = 1 << bits;
 		const uint64_t num = expected * cells;
-		uint64_t f[256 * 256] = { 0 };
+		uint64_t f[1 << 8] = { 0 };
```
MatiasElo, line 224: Please use normal number presentation for easier readability.

MatiasElo (commented):
One minor comment and copyright can be updated to both files.

test/validation/api/random/random.c
JereLeppanen: The reason why I put it that way was to highlight that it's the same 8 as in the loop condition `bits <= 8`. But I'll change it.

MatiasElo (approved):

PR #1473, opened 2022-01-13
[PATCH v3] validation: atomic: fix validation in case of 32 bit overflow and underflow

JereLeppanen:
```
validation: atomic: limit random buffer size
    
    Set the size of the buffer holding random numbers in the xchg tests to
    64K, so that we don't run out of memory if the test length is
    increased.
    
validation: atomic: fix validation in case of 32 bit overflow and underflow
    
    If test length is increased, 32 bit counters will overflow or
    underflow, and min and max variables will be saturated. Fix validation
    for these cases.
    
```
v2:
- Rebase.
- Tuomas' comments.

v3:
- Rebase.
- Add review tags.

### DIFF
--- a/test/validation/api/atomic/atomic.c
+++ b/test/validation/api/atomic/atomic.c
@@ -1,5 +1,5 @@
 /* Copyright (c) 2014-2018, Linaro Limited
- * Copyright (c) 2021 Nokia
+ * Copyright (c) 2021-2022 Nokia
  * All rights reserved.
  *
  * SPDX-License-Identifier:	 BSD-3-Clause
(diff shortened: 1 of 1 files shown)


test/validation/api/atomic/atomic.c
```diff
@@ -18,15 +18,18 @@
 #define ADD_SUB_CNT		5
```
TuomasTaipale, line 19: Copyright year could be updated.

Also, old code, but `VERBOSE` and `MAX_ITERATIONS` seem to be unused and could be removed.

TuomasTaipale (approved):

MatiasElo (approved):

PR #1477, opened 2022-01-20
[PATCH v6] linux-gen: packet: clean up packet metadata copy operations

MatiasElo:


platform/linux-generic/include/odp_packet_internal.h
```diff
@@ -248,6 +251,76 @@ static inline void packet_init(odp_packet_hdr_t *pkt_hdr, uint32_t len)
...
+	if (src_hdr->p.flags.payload_off)
+		dst_hdr->payload_offset = src_hdr->payload_offset;
+
+	dst_hdr->p = src_hdr->p;
+
+	if (uarea_copy) {
+		pool_t *src_pool = src_hdr->event_hdr.pool_ptr;
+		pool_t *dst_pool = dst_hdr->event_hdr.pool_ptr;
+		uint32_t src_uarea_size = src_pool->param_uarea_size;
+		uint32_t dst_uarea_size = dst_pool->param_uarea_size;
```
TuomasTaipale, line 298: These could be `const`.

platform/linux-generic/include/odp_packet_internal.h
```diff
@@ -248,6 +251,76 @@ static inline void packet_init(odp_packet_hdr_t *pkt_hdr, uint32_t len)
...
+
+		if (dst_hdr->uarea_addr != NULL && src_hdr->uarea_addr != NULL) {
+			memcpy(dst_hdr->uarea_addr, src_hdr->uarea_addr,
+			       dst_uarea_size <= src_uarea_size ? dst_uarea_size :
```
TuomasTaipale, line 302: There seems to be a `MIN()` macro available in `odp_macros_internal.h`, that could be used?

platform/linux-generic/include/odp_packet_internal.h
MatiasElo: Fixed in V2.

platform/linux-generic/include/odp_packet_internal.h
MatiasElo: Fixed in V2.

platform/linux-generic/odp_packet.c
```diff
@@ -1380,7 +1342,7 @@ odp_packet_t odp_packet_copy(odp_packet_t pkt, odp_pool_t pool)
 	odp_packet_t newpkt = odp_packet_alloc(pool, pktlen);
 
 	if (newpkt != ODP_PACKET_INVALID) {
-		if (_odp_packet_copy_md_to_packet(pkt, newpkt) ||
+		if (_odp_packet_copy_md(srchdr, packet_hdr(newpkt), 1) ||
```
psavol, line 1345: copy_md() return value can be removed if caller checks uarea sizes beforehand. Optimally uarea sizes would be checked from src vs dst pool, so that packet_alloc() is avoided if uarea sizes don't match.

platform/linux-generic/odp_packet.c
```diff
@@ -166,44 +166,6 @@ static inline void push_tail(odp_packet_hdr_t *pkt_hdr, uint32_t len)
...
-static inline void packet_seg_copy_md(odp_packet_hdr_t *dst,
-				      odp_packet_hdr_t *src)
-{
-	dst->p = src->p;
-
-	/* lengths are not copied:
-	 *   .frame_len
-	 *   .headroom
-	 *   .tailroom
-	 */
```
psavol, line 180: The new _odp_packet_copy_md() function should also document which fields will not be copied: those that relate to data buffering and are set already by a previous alloc/copy call.

platform/linux-generic/odp_packet.c
MatiasElo: Fixed in V2.

JereLeppanen (approved):

platform/linux-generic/include/odp_packet_internal.h
```diff
@@ -248,6 +251,106 @@ static inline void packet_init(odp_packet_hdr_t *pkt_hdr, uint32_t len)
...
+ *
+ * @param      src_hdr     Source packet header
+ * @param      dst_hdr     Destination packet header
+ * @param      uarea_copy  Copy user area data. If false, user area pointers
+ *                         are swapped between the packet headers (allowed
+ *                         only when packets are from the same pool).
+ */
+static inline void _odp_packet_copy_md(odp_packet_hdr_t *src_hdr,
+				       odp_packet_hdr_t *dst_hdr,
+				       odp_bool_t uarea_copy)
```
psavol, line 297: Usually, copy functions have "dst" on left, and "src" on right (analog to dst = src):
  void * memcpy ( void * destination, const void * source, size_t num );

_odp_packet_copy_md_possible() would need the same change.

platform/linux-generic/include/odp_packet_internal.h
```diff
@@ -248,6 +251,106 @@ static inline void packet_init(odp_packet_hdr_t *pkt_hdr, uint32_t len)
...
+	src_hdr = pool_entry_from_hdl(src_pool);
+	dst_hdr = pool_entry_from_hdl(dst_pool);
+
+	if (dst_hdr->param_uarea_size < src_hdr->param_uarea_size)
+		return -1;
+
+	return 1;
+}
+
+/**
```
psavol, line 282: Doxygen style comments not needed as this is internal function (and not seen/checked by Doxygen)

MatiasElo: Fixed in V6.

platform/linux-generic/include/odp_packet_internal.h
MatiasElo: Fixed in V6.

psavol (approved):

PR #1481, opened 2022-01-25
[PATCH v7] test: performance: ipsec: introduce burst processing

anoobj:
Introduce an option to prepare and submit multiple packets for IPsec
processing.

ODP allows multi APIs with packet alloc, free & IPsec processing. Usage
of these APIs would help in improving performance with burst cases. New
option '--burst' or '-b' can be used to enable burst processing. When
specified without any arguments, burst size would be set as 32. User can
control burst size by specifying the optional argument.

Usage: 'odp_ipsec --burst=512'

### DIFF
--- a/test/performance/odp_ipsec.c
+++ b/test/performance/odp_ipsec.c
@@ -1,4 +1,5 @@
 /* Copyright (c) 2018, Linaro Limited
+ * Copyright (c) 2022, Nokia
  * All rights reserved.
  *
  * SPDX-License-Identifier:	BSD-3-Clause
@@ -18,6 +19,7 @@
 
 #include <odp_api.h>
 #include <odp/helper/odph_api.h>
+#include <inttypes.h>
 
 /** @def POOL_NUM_PKT
  * Number of packets in the pool
@@ -608,6 +610,17 @@ make_packet(odp_pool_t pkt_pool, unsigned int payload_length)
 	return pkt;
 }
 
+static inline void check_ipsec_result(odp_packet_t ipsec_pkt)
+{
+	odp_ipsec_packet_result_t result;
+
+	if (odp_unlikely(odp_ipsec_result(&result, ipsec_pkt)))
+		ODPH_ERR("odp_ipsec_result() failed\n");
+	else if (odp_unlikely(result.status.error.all))
+		ODPH_ERR("IPsec processing error: %" PRIu32 "\n",
+			 result.status.error.all);
+}
+
 /**
  * Run measurement iterations for given config and payload size.
  * Result of run returned in 'result' out parameter.
@@ -665,13 +678,8 @@ run_measure_one(ipsec_args_t *cargs,
 				odp_packet_free(pkt);
 				break;
 			}
-			if (odp_packet_has_error(out_pkt)) {
-				odp_ipsec_packet_result_t result;
+			check_ipsec_result(out_pkt);
 
-				odp_ipsec_result(&result, out_pkt);
-				ODPH_ERR("Received error packet: %d\n",
-					 result.status.error.all);
-			}
 			packets_sent += rc;
 			packets_received += num_out;
 			if (cargs->debug_packets)
@@ -756,10 +764,9 @@ run_measure_one_async(ipsec_args_t *cargs,
 
 		while (ev != ODP_EVENT_INVALID) {
 			odp_packet_t out_pkt;
-			odp_ipsec_packet_result_t result;
 
 			out_pkt = odp_ipsec_packet_from_event(ev);
-			odp_ipsec_result(&result, out_pkt);
+			check_ipsec_result(out_pkt);
 
 			if (cargs->debug_packets)
 				odp_packet_print_data(out_pkt, 0,
--- a/test/performance/odp_ipsec.c
+++ b/test/performance/odp_ipsec.c
@@ -1,4 +1,5 @@
 /* Copyright (c) 2018, Linaro Limited
+ * Copyright (c) 2022, Marvell
  * Copyright (c) 2022, Nokia
  * All rights reserved.
  *
@@ -24,7 +25,7 @@
 /** @def POOL_NUM_PKT
  * Number of packets in the pool
  */
-#define POOL_NUM_PKT  64
+#define POOL_NUM_PKT  4096
 
 static uint8_t test_salt[16] = "0123456789abcdef";
 
@@ -99,11 +100,11 @@ typedef struct {
 	int in_flight;
 
 	/**
-	 * Number of iteration to repeat crypto operation to get good
-	 * average number. Specified through -i or --terations option.
+	 * Number of packets to be IPsec processed to get good average number.
+	 * Specified through -c or --count option.
 	 * Default is 10000.
 	 */
-	int iteration_count;
+	int packet_count;
 
 	/**
 	 * Payload size to test. If 0 set of predefined payload sizes
@@ -141,6 +142,13 @@ typedef struct {
 	 * Specified through -u argument.
 	 */
 	int ah;
+
+	/*
+	 * Burst size.
+	 * Prepare and submit as many packets for IPsec processing in each
+	 * iteration of the loop.
+	 */
+	int burst_size;
 } ipsec_args_t;
 
 /*
@@ -518,7 +526,7 @@ print_result(ipsec_args_t *cargs,
 
 	throughput = (1000000.0 / result->elapsed) * payload_length / 1024;
 	printf("%30.30s %15d %15d %15.3f %15.3f %15.3f %15d\n",
-	       config->name, cargs->iteration_count, payload_length,
+	       config->name, cargs->packet_count, payload_length,
 	       result->elapsed, result->rusage_self, result->rusage_thread,
 	       throughput);
 }
@@ -586,28 +594,30 @@ static uint8_t test_data[] = {
 	0x08, 0x00, 0xfb, 0x37, 0x12, 0x34, 0x00, 0x00
 };
 
-static odp_packet_t
-make_packet(odp_pool_t pkt_pool, unsigned int payload_length)
+static int
+make_packet_multi(odp_pool_t pkt_pool, unsigned int payload_length,
+		  odp_packet_t pkt[], int num)
 {
-	odp_packet_t pkt;
+	int i, ret;
 
-	if (payload_length < sizeof(test_data))
-		return ODP_PACKET_INVALID;
-
-	pkt = odp_packet_alloc(pkt_pool, payload_length);
-	if (pkt == ODP_PACKET_INVALID) {
-		ODPH_ERR("failed to allocate buffer\n");
-		return pkt;
+	ret = odp_packet_alloc_multi(pkt_pool, payload_length, pkt, num);
+	if (ret != num) {
+		ODPH_ERR("Could not allocate buffer\n");
+		if (ret > 0)
+			odp_packet_free_sp(pkt, ret);
+		return -1;
 	}
 
-	odp_packet_copy_from_mem(pkt, 0, sizeof(test_data), test_data);
-	odp_packet_l3_offset_set(pkt, 0);
+	for (i = 0; i < num; i++) {
+		odp_packet_copy_from_mem(pkt[i], 0, sizeof(test_data), test_data);
+		odp_packet_l3_offset_set(pkt[i], 0);
 
-	uint8_t *mem = odp_packet_data(pkt);
-	((odph_ipv4hdr_t *)mem)->tot_len = odp_cpu_to_be_16(payload_length);
-	memset(mem + sizeof(test_data), 1, payload_length - sizeof(test_data));
+		uint8_t *mem = odp_packet_data(pkt[i]);
+		((odph_ipv4hdr_t *)mem)->tot_len = odp_cpu_to_be_16(payload_length);
+		memset(mem + sizeof(test_data), 1, payload_length - sizeof(test_data));
+	}
 
-	return pkt;
+	return 0;
 }
 
 static inline void check_ipsec_result(odp_packet_t ipsec_pkt)
@@ -632,16 +642,20 @@ run_measure_one(ipsec_args_t *cargs,
 		time_record_t *start,
 		time_record_t *end)
 {
+	int in_flight, pkts_allowed, num_out, num_pkts, rc = 0;
+	const int max_in_flight = cargs->in_flight;
+	const int burst_size = cargs->burst_size;
 	odp_ipsec_out_param_t param;
 	odp_pool_t pkt_pool;
-	odp_packet_t pkt = ODP_PACKET_INVALID;
-	int rc = 0;
 
 	pkt_pool = odp_pool_lookup("packet_pool");
 	if (pkt_pool == ODP_POOL_INVALID) {
 		ODPH_ERR("pkt_pool not found\n");
 		return -1;
 	}
+
+	if (payload_length < sizeof(test_data))
+		return -1;
 
 	int packets_sent = 0;
 	int packets_received = 0;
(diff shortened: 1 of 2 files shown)


test/performance/odp_ipsec.c
```diff
@@ -139,6 +139,13 @@ typedef struct {
 	 * Specified through -u argument.
 	 */
 	int ah;
+
+	/*
+	 * Burst size.
+	 * Prepare and submit as many packets for IPsec processing in each
+	 * iteration of the loop.
+	 */
+	int burst_sz;
```
JannePeltonen, line 148: Maybe burst_size would be better name than burst_sz in all places where burst_sz is used?

test/performance/odp_ipsec.c
```diff
@@ -584,28 +591,29 @@ static uint8_t test_data[] = {
...
-		return ODP_PACKET_INVALID;
-
-	pkt = odp_packet_alloc(pkt_pool, payload_length);
-	if (pkt == ODP_PACKET_INVALID) {
-		ODPH_ERR("failed to allocate buffer\n");
-		return pkt;
+	ret = odp_packet_alloc_multi(pkt_pool, payload_length, pkt, num);
+	if (ret != num) {
+		ODPH_ERR("Could not allocate buffer\n");
+		odp_packet_free_sp(pkt, ret);
```
JannePeltonen, line 603: ret may be negative here. This should be called only if it is positive.

test/performance/odp_ipsec.c
```diff
@@ -643,41 +655,58 @@ run_measure_one(ipsec_args_t *cargs,
 
 	while ((packets_sent < cargs->iteration_count) ||
 	       (packets_received < cargs->iteration_count)) {
-		if ((packets_sent < cargs->iteration_count) &&
+		int num_out = cargs->iteration_count - packets_sent;
```
JannePeltonen, line 658: Looking at the (packets_sent < cargs->iteration_count) condition, now the iteration_count is really packet count and not the number of iterations or number of times the IPsec enqueue API gets called. Maybe the description of -i in usage() should be updated?

test/performance/odp_ipsec.c
```diff
@@ -643,41 +655,58 @@ run_measure_one(ipsec_args_t *cargs,
...
 				break;
 			}
-			if (odp_packet_has_error(out_pkt)) {
-				odp_ipsec_packet_result_t result;
 
-				odp_ipsec_result(&result, out_pkt);
-				ODPH_ERR("Received error packet: %d\n",
-					 result.status.error.all);
+			for (i = 0; i < num_out; i++) {
+				if (odp_packet_has_error(out_pkt[i])) {
```
JannePeltonen, line 692: I think checking odp_packet_has_error() here and not checking the IPsec packet result here and in the async function is wrong. But maybe that needs to be fixed in a different patch.

test/performance/odp_ipsec.c
```diff
@@ -726,52 +760,63 @@ run_measure_one_async(ipsec_args_t *cargs,
...
+		 i = 0;
+
 		if (cargs->schedule)
 			ev = odp_schedule(NULL,
 					  ODP_SCHED_NO_WAIT);
 		else
 			ev = odp_queue_deq(out_queue);
 
 		while (ev != ODP_EVENT_INVALID) {
-			odp_packet_t out_pkt;
```
JannePeltonen, line 758: Maybe this async test loop should use odp_schedule_multi() now?

test/performance/odp_ipsec.c
JannePeltonen: I made PR 1506 to fix this.

test/performance/odp_ipsec.c
JannePeltonen: PR 1506 takes odp_schedule_multi and odp_queue_deq_multi into use here. Please take a look.

test/performance/odp_ipsec.c
anoobj: Agreed. Will make this change in v2.

test/performance/odp_ipsec.c
anoobj: Agreed. Will make this change in v2.

test/performance/odp_ipsec.c
anoobj: Agreed. Do you suggest renaming the variable as well? (like, iteration_count -> packet_count?)

Or would updating description of -i option good enough?

JannePeltonen: Its probably good to rename the variable too for clarity.

test/performance/odp_ipsec.c
```diff
@@ -643,41 +655,58 @@ run_measure_one(ipsec_args_t *cargs,
...
 	       (packets_received < cargs->iteration_count)) {
-		if ((packets_sent < cargs->iteration_count) &&
+		int num_out = cargs->iteration_count - packets_sent;
+
+		num_out = num_out > burst_sz ? burst_sz : num_out;
+
+		if (num_out &&
 		    (packets_sent - packets_received <
 		     cargs->in_flight)) {
-			odp_packet_t out_pkt;
```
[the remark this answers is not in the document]
anoobj: I think it's better to have this fixed in code. I'll address this in v2.

anoobj:
@JannePeltonen Do you have any comments on v2? If not I can push next version with your patch from PR 1506

test/performance/odp_ipsec.c
```diff
@@ -971,8 +1031,18 @@ static void parse_args(int argc, char *argv[], ipsec_args_t *cargs)
 		case 'd':
 			cargs->debug_packets = 1;
 			break;
-		case 'i':
-			cargs->iteration_count = atoi(optarg);
+		case 'c':
```
JannePeltonen, line 1034: Now that the argument letter changes, odp_ipsec_run.sh needs an update too.

test/performance/odp_ipsec.c
```diff
@@ -641,43 +655,64 @@ run_measure_one(ipsec_args_t *cargs,
...
+
+		/* Enqueue up to (max in flight - current in flight) */
+		in_flight = packets_sent - packets_received;
+		pkts_allowed = max_in_flight - in_flight;
+		num_out = num_out > pkts_allowed ? pkts_allowed : num_out;
 
-			pkt = make_packet(pkt_pool, payload_length);
-			if (ODP_PACKET_INVALID == pkt)
+		if (odp_likely(num_out)) {
+			odp_packet_t out_pkt[POOL_NUM_PKT];
```
JannePeltonen, line 671: This could be out_pkt[num_out].

test/performance/odp_ipsec.c
```diff
@@ -693,11 +728,14 @@ run_measure_one_async(ipsec_args_t *cargs,
...
+	int in_flight, pkts_allowed, i, num_pkts, rc = 0;
+	const int max_in_flight = cargs->in_flight;
+	const int burst_size = cargs->burst_size;
 	odp_ipsec_out_param_t param;
 	odp_pool_t pkt_pool;
 	odp_queue_t out_queue;
-	odp_packet_t pkt = ODP_PACKET_INVALID;
-	int rc = 0;
+	odp_packet_t pkt_out[POOL_NUM_PKT];
+	odp_packet_t pkt[POOL_NUM_PKT];
```
JannePeltonen, line 738: These could also be smaller. If moved down in the loop, then num_pkts as the size would suffice, it seems.

JannePeltonen: ...or max_in_flight.

test/performance/odp_ipsec.c
```diff
@@ -641,43 +655,64 @@ run_measure_one(ipsec_args_t *cargs,
...
+				for (i = 0; i < num_out; i++)
+					odp_packet_print_data(pkt[i], 0,
+							      odp_packet_len(pkt[i]));
+			}
 
-			rc = odp_ipsec_out(&pkt, 1,
-					   &out_pkt, &num_out,
+			i = num_out;
+
+			rc = odp_ipsec_out(pkt, num_out,
```
JannePeltonen, line 688: If this does not consume all packets, then the leftover packet are leaked.

test/performance/odp_ipsec.c
```diff
@@ -722,56 +763,72 @@ run_measure_one_async(ipsec_args_t *cargs,
...
-				odp_packet_print_data(pkt, 0,
-						      odp_packet_len(pkt));
+			if (cargs->debug_packets) {
+				for (i = 0; i < num_pkts; i++)
+					odp_packet_print_data(pkt[i], 0,
+							      odp_packet_len(pkt[i]));
+			}
 
-			rc = odp_ipsec_out_enq(&pkt, 1,
+			rc = odp_ipsec_out_enq(pkt, num_pkts,
```
JannePeltonen, line 793: There is a packet leak also here if the enqueue does not consume all input packets.

test/performance/odp_ipsec.c
anoobj: Agreed. Will update in v3.

test/performance/odp_ipsec.c
anoobj: Agreed. Will fix in v3.

test/performance/odp_ipsec.c
anoobj: Agreed. Will fix in v3.

test/performance/odp_ipsec.c
anoobj: Agreed. Will move it deeper and use max_in_flight & num_pkts.

test/performance/odp_ipsec.c
anoobj: Updated in v3.

anoobj:
@JannePeltonen Do you have any further comments?

test/performance/odp_ipsec.c
```diff
@@ -632,17 +642,22 @@ run_measure_one(ipsec_args_t *cargs,
 		time_record_t *start,
 		time_record_t *end)
 {
+	int in_flight, pkts_allowed, num_out, num_enq, rc = 0;
+	const int max_in_flight = cargs->in_flight;
+	const int burst_size = cargs->burst_size;
+	odp_packet_t pkt[POOL_NUM_PKT];
```
JannePeltonen, line 648: This could be made much smaller if moved down, i.e. similar change could be done here as was now done in the async case.

test/performance/odp_ipsec.c
```diff
@@ -654,38 +669,57 @@ run_measure_one(ipsec_args_t *cargs,
...
+				for (i = 0; i < num_enq; i++)
+					odp_packet_print_data(pkt[i], 0,
+							      odp_packet_len(pkt[i]));
+			}
+
+			num_out = num_enq;
 
-			rc = odp_ipsec_out(&pkt, 1,
-					   &out_pkt, &num_out,
+			rc = odp_ipsec_out(pkt, num_enq,
```
JannePeltonen, line 702: Leftover packets are leaked if this does not consume all packets.

test/performance/odp_ipsec.c
```diff
@@ -654,38 +669,57 @@ run_measure_one(ipsec_args_t *cargs,
...
-				odp_packet_print_data(out_pkt, 0,
-						      odp_packet_len(out_pkt));
-			odp_packet_free(out_pkt);
+			if (cargs->debug_packets) {
+				for (i = 0; i < num_out; i++)
+					odp_packet_print_data(out_pkt[i], 0,
+							      odp_packet_len(out_pkt[i]));
+			}
+
+			odp_packet_free_sp(out_pkt, num_enq);
```
JannePeltonen, line 722: This can try to free invalid or stale packet handles. The number of packets in the array is not num_enq but num_out, which may be less than num_enq.

test/performance/odp_ipsec.c
```diff
@@ -730,55 +768,76 @@ run_measure_one_async(ipsec_args_t *cargs,
...
-			if (ODP_PACKET_INVALID == pkt)
+		num_pkts = cargs->packet_count - packets_sent;
+
+		/* Enqueue up to burst size */
+		num_pkts = num_pkts > burst_size ? burst_size : num_pkts;
+
+		/* Enqueue up to (max in flight - current in flight) */
+		in_flight = packets_sent - packets_received;
+		pkts_allowed = max_in_flight - in_flight;
+		num_pkts = num_pkts > pkts_allowed ? pkts_allowed : num_pkts;
```
JannePeltonen, line 784: I think this should be:

num_pkts = num_pkts > pkts_allowed ? 0 : num_pkts;

This would make the burst size always (except for the last burst) the same what was specified in the -b option, which is what one would expect. Now the burst size can get smaller if IPsec processing is slower than the test loop (as is expected) and then using a smaller burst size can further slow down the processing so that the loop ends up using small burst size for pretty much the whole test.

And the same comment for the sync case.

With this change (always using the requested burst size), I think the program should complain and exit if the in_flight parameter is less than the burst size parameter.

test/performance/odp_ipsec.c
```diff
@@ -632,17 +642,22 @@ run_measure_one(ipsec_args_t *cargs,
 		time_record_t *start,
 		time_record_t *end)
 {
+	int in_flight, pkts_allowed, num_out, num_enq, rc = 0;
```
JannePeltonen, line 645: I think num_enq is not a very good name here in the sync case which does not really enqueue packets. The async case seems to use num_pkts which could be used here too (and it would have the added benefit of making the two functions more similar in variable naming).

test/performance/odp_ipsec_run.sh
```diff
@@ -7,9 +7,9 @@
 
 TEST_DIR="${TEST_DIR:-$(dirname $0)}"
 
-# Run with a small number of iterations in make check
+# Run with small number of packets in make check
```
JannePeltonen, line 10: There is no reason to remove the indefinite article.

test/performance/odp_ipsec.c
anoobj: If out_pkt is replaced with pkt, that should be enough, right?

test/performance/odp_ipsec.c
anoobj: Agreed. Will change in v4.

test/performance/odp_ipsec.c
JannePeltonen: Yes.

test/performance/odp_ipsec.c
anoobj: No strong opinion on variable name. Will rename as part of v4.

test/performance/odp_ipsec.c
anoobj: In that case, are you okay with making packet_count a multiple of burst_size? Then we can improve/simplify the run_measure_*() loops.

test/performance/odp_ipsec.c
anoobj: I'll make this change as part of v4.

test/performance/odp_ipsec.c
JannePeltonen: Maybe making packet_count a multiple of burst size does not quite enable the simplification that you may be thinking, because the ipsec out/enq function can still consume only part of the burst, making the count of processed packets not divisible by the burst size.

But simplification is good, and I think the test loop could just stop when the remaining packet count falls below the burst count (or below zero or whatever comes naturally). IOW, the packet count would not be an exact number anymore (because who would actually care?), but just a limit to stop the test at that point.

Actually, I think it would be even better if the test termination would not be based on packet count but elapsed time, as that would make it easier to run the test loop in multiple threads simultaneously without syncing about the number of processed packets. That is what I have been doing with my private test apps (i.e. time based termination & multiple threads running a test loop).

But since this test app does not support multiple threads, packet count limit is ok. I just do not see any need for it to be exact. So a loop that always sends a full burst or nothing and then terminates roughly, but not precisely, after the requested number of packets should be fine. And it would simpler and clearer than having extra logic for handling a partial burst at the end.

anoobj: @JannePeltonen Thanks for your inputs. I hadn't thought about partial enqueue of the burst. So, agreed that such a change won't lead to the simplification I had in mind. So, I'll stick to your earlier suggestion.

test/performance/odp_ipsec.c
JannePeltonen: Sorry, I misread the question when I answered. Replacing num_enq with out_pkt would have fixed the problem. Replacing out_pkt with pkt is wrong.

odp_ipsec_out() consumes some or all of the input packets and produces output packets. The output packets are not necessarily the same as the input packets although they can be. So after odp_ipsec_out() call you have a zero or more packets left in the input array and you need to free them in order not to leak them. You also have zero or more packets in the output array, but they are new packets and cannot be freed using the old input array handles that are now stale (except that in some implementations, like linux-gen, they may sometimes (but not always), be the same handles in which case the current code accidentally works).

test/performance/odp_ipsec.c
JannePeltonen: > With this change (always using the requested burst size), I think the program should complain and exit if the in_flight parameter is less than the burst size parameter.

I think something like this should be done or otherwise the program gets stuck when run with smaller -f than -b.

anoobj: Agreed. I'll have this addressed in v5.

test/performance/odp_ipsec.c
anoobj: Agreed. Will have this changed in v5.

test/performance/odp_ipsec_run.sh
JannePeltonen: It disappeared again in v5.

test/performance/odp_ipsec.c
```diff
@@ -654,38 +668,62 @@ run_measure_one(ipsec_args_t *cargs,
...
-						      odp_packet_len(out_pkt));
-			odp_packet_free(out_pkt);
+			if (cargs->debug_packets) {
+				for (i = 0; i < num_out; i++)
+					odp_packet_print_data(out_pkt[i], 0,
+							      odp_packet_len(out_pkt[i]));
+			}
+
+			if (odp_unlikely(num_out != num_pkts))
+				odp_packet_free_sp(&pkt[num_out], num_pkts - num_out);
```
JannePeltonen, line 725: This code makes the assumption that the number of consumed packets (rc) is the same as the number of output packets (num_out), which won't be true as soon as fragmentation offload gets enabled. It would be more robust to check if the number of consumed packets (i.e. rc) is not the same as the number of input packets and also use the number of consumed packet instead of the number of output packets in the freeing statement. The number of output packets is not directly relevant to freeing the unconsumed packets.

test/performance/odp_ipsec.c
```diff
@@ -1008,6 +1086,11 @@ static void parse_args(int argc, char *argv[], ipsec_args_t *cargs)
 		}
 	}
 
+	if (cargs->in_flight < cargs->burst_size) {
+		printf("-f (flight) should be greater than -b (burst)\n");
```
JannePeltonen, line 1090: Burst size should not exceed the maximum number of in-flight packets, but it may equal it.

test/performance/odp_ipsec.c
anoobj: @JannePeltonen So I'll replace num_out with rc in this code snippet and that should be good enough, right?

test/performance/odp_ipsec.c
anoobj: @JannePeltonen  Shall I update the text as,

`printf("-f (flight) must be greater than or equal to -b (burst)\n");`

Or did you mean use the above text verbatim?

test/performance/odp_ipsec.c
JannePeltonen: Yes, that should fix it. There is already that kind of code in the async case, so this just becomes more similar.

test/performance/odp_ipsec.c
JannePeltonen: Actually, your proposed update looks ok whereas copying my comment verbatim would make the error message more complicated than it needs to be, I think.

JannePeltonen (approved):
Looks good.

PR #1482, opened 2022-01-25
[PATCH v2] api: tm: define default values

JannePeltonen:
api: tm: define default values for many parameters

    Define the default values for many parameters to be set by various
    init functions. Current API validation tests already rely on certain
    defaults (at least packet_mode being false) which are set by the
    linux-gen implementation but which are not specified in the API.

    validation: tm: test default values

    Test that init functions initialize parameters to their default values.


nithind1988:

JannePeltonen:
v2: rebased and added reviewed-by tags.

PR #1483, opened 2022-01-27
[PATCH v6] api: timer: expiration mode parameter

psavol:
New timer pool parameter which allows application to force timers not to expire before target time.

### DIFF
--- a/include/odp/api/spec/timer_types.h
+++ b/include/odp/api/spec/timer_types.h
@@ -203,9 +203,46 @@ typedef enum {
 #define ODP_CLOCK_EXT ODP_CLOCK_SRC_1
 
 /**
+ * Timer expiration mode
+ *
+ * Expiration mode selects how timer expiration tick is interpreted. In ODP_TIMER_EXP_RELAXED mode,
+ * timer implementation may round the requested time up or down within resolution limits, and
+ * therefore application may receive timeouts slightly before or after the requested time.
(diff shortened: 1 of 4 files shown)


include/odp/api/spec/timer_types.h
```diff
@@ -202,10 +202,42 @@ typedef enum {
...
+ * (ODP_TIMER_EXP_DEFAULT), a timer implementation may round the requested expiration time up or
+ * down within resolution limits, and therefore application may receive timeouts slightly before or
+ * after the requested time. In ODP_TIMER_EXP_AFTER mode, timers are limited to expire exactly on
+ * the requested time or after it, but never before the time.
+ */
+typedef enum {
+	/** Expiration mode default
+	 *
+	 *  Timers may expire before or after the specified time. */
+	ODP_TIMER_EXP_DEFAULT = 0,
```
TuomasTaipale, line 218: Using "default" in the name might be confusing (e.g. interpreted as pointing to a higher level constant), maybe `ODP_TIMER_EXP_RELAXED`?

psavol: Fixed in v2. Decided to change EXP_AFTER as the default value, since a validation test expected that already.

TuomasTaipale (approved):

psavol:
v4: rebased

jerinjacobk (approved):

PR #1484, opened 2022-01-27
[PATCH v3] Speed up make check

JannePeltonen:
Many tests run unnecessarily long during make check, especially on slower platforms. Often just a small number of iterations or short test time is sufficient for checking that things work and longer run time just wastes cycles without much benefit.

Reduce the run time of various tests for which it was easy to do.

### DIFF
--- a/example/classifier/odp_classifier.c
+++ b/example/classifier/odp_classifier.c
@@ -187,6 +187,8 @@ static inline void print_cls_statistics(appl_args_t *args)
 		infinite = 1;
 
 	for (; timeout > 0 || infinite; timeout--) {
+		sleep(1);
+		printf("\r");
 		for (i = 0; i < args->policy_count; i++) {
 			printf("%-6" PRIu64 " ",
 			       odp_atomic_load_u64(&args->stats[i]
@@ -198,13 +200,10 @@ static inline void print_cls_statistics(appl_args_t *args)
 
 		printf("%-" PRIu64, odp_atomic_load_u64(&args->
 							total_packets));
+		fflush(stdout);
 
 		if (args->shutdown_sig)
 			break;
-
-		sleep(1);
-		printf("\r");
-		fflush(stdout);
 	}
 
 	printf("\n");
--- a/platform/linux-generic/test/pktio_ipc/pktio_ipc_run.sh
+++ b/platform/linux-generic/test/pktio_ipc/pktio_ipc_run.sh
@@ -17,9 +17,9 @@ PATH=$(dirname $0):$PATH
 PATH=$(dirname $0)/../../../../platform/linux-generic/test/pktio_ipc:$PATH
 PATH=.:$PATH
 
-RUNTIME1=10
-RUNTIME2=5
-TIMEOUT=13
+RUNTIME1=3
+RUNTIME2=1
+TIMEOUT=3
 if [ "${TEST}" = "coverage" ]; then
 	RUNTIME1=30
 	RUNTIME2=15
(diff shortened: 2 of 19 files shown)


example/classifier/odp_classifier.c
```diff
@@ -187,6 +187,9 @@ static inline void print_cls_statistics(appl_args_t *args)
 		infinite = 1;
 
 	for (; timeout > 0 || infinite; timeout--) {
+		sleep(1);
+		printf("\r");
+		fflush(stdout);
```
TuomasTaipale, line 192: The flushing could be moved to just before the `shutdown_sig` check.

test/performance/Makefile.am
```diff
@@ -25,7 +25,10 @@ COMPILE_ONLY = odp_dma_perf \
 	       odp_scheduling \
 	       odp_timer_perf
 
-TESTSCRIPTS = odp_dma_perf_run.sh \
+TESTSCRIPTS = odp_cpu_bench_run.sh \
+	      odp_crypto_run.sh \
+              odp_dma_perf_run.sh \
```
TuomasTaipale, line 30: Seems to be all whitespaces in this line.

example/classifier/odp_classifier.c
JannePeltonen: Indeed. Will move it.

test/performance/Makefile.am
JannePeltonen: will fix

JannePeltonen:
v2: rebased and fixed the commented issues

TuomasTaipale (approved):

platform/linux-generic/test/pktio_ipc/pktio_ipc_run.sh
```diff
@@ -17,9 +17,9 @@ PATH=$(dirname $0):$PATH
 PATH=$(dirname $0)/../../../../platform/linux-generic/test/pktio_ipc:$PATH
 PATH=.:$PATH
 
-RUNTIME1=10
-RUNTIME2=5
-TIMEOUT=13
+RUNTIME1=3
+RUNTIME2=1
+TIMEOUT=3
```
JereLeppanen, line 22: I wonder if this timeout is too short for some software emulator, as usual.

test/performance/odp_pktio_perf.c
```diff
@@ -52,6 +52,12 @@
  * received by this time will be assumed to have been lost. */
 #define SHUTDOWN_DELAY_NS (ODP_TIME_MSEC_IN_NS * 100)
 
+/* Number of duration units in a second. */
+#define T_SCALE 10
```
JereLeppanen, line 56: How about milliseconds?

platform/linux-generic/test/pktio_ipc/pktio_ipc_run.sh
JannePeltonen: Maybe, who knows. We will see and can fix it then. The test script should actually be improved so that it does not take longer than necessary when things complete fast but would have more generous timeout for when things are slow (e.g. it should not sleep for a fixed time but poll for the process exit periodically). But I did not bother to do that but just this simple change that seems to be ok in the CI and my own environments.

test/performance/odp_pktio_perf.c
JannePeltonen: I am not sure how that would be better other than milliseconds being more typical a unit than the tenths of seconds used now. One benefit of not using finer time scale than necessary is that duration can be longer before it overflows.

JereLeppanen: > more typical a unit

That is definitely reason enough for me. If 20+ days is not a sufficient duration, then make it 64 bit. But I leave it to you to decide.

JereLeppanen (approved):

JannePeltonen:
v3: rebased and added review tags

PR #1487, opened 2022-02-01
[PATCH v2] linux-gen: tm: synchronize with TM thread when configuring shaper

JannePeltonen:
Use the signal_request/check_request mechanism to make sure the TM
system thread is not accessing shaper parameters at the same time
when they are being changed by another thread.

This should prevent the crash that sometimes occurs when running TM
validation tests that disable shapers on the fly, changing the shaper
parameter pointer in shaper objects NULL just when the TM system thread
is about to dereference it.

This patch does not fix other issues in the synchronization mechanism:
The mechanism is not always used when needed, it uses relaxed atomics
that do not guarantee memory ordering and it does not work with
multiple TM system threads.

### DIFF
--- a/platform/linux-generic/odp_traffic_mngr.c
+++ b/platform/linux-generic/odp_traffic_mngr.c
@@ -3918,6 +3918,7 @@ int odp_tm_node_shaper_config(odp_tm_node_t tm_node,
 {
 	tm_node_obj_t *tm_node_obj;
 	tm_system_t *tm_system;
+	odp_bool_t sync_needed;
 
 	tm_node_obj = GET_TM_NODE_OBJ(tm_node);
 	if (!tm_node_obj)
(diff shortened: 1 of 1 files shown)


platform/linux-generic/odp_traffic_mngr.c
```diff
@@ -151,6 +151,10 @@ typedef struct {
 static tm_global_t *tm_glb;
 
 /* Forward function declarations. */
+
+static void signal_request(void);
+static void signal_request_done(void);
```
psavol, line 156: Forward declarations are not needed as functions are defined before they are called.

platform/linux-generic/odp_traffic_mngr.c
```diff
@@ -3928,8 +3932,18 @@ int odp_tm_node_shaper_config(odp_tm_node_t tm_node,
 		return -1;
 
 	odp_ticketlock_lock(&tm_glb->profile_lock);
+	if (!tm_glb->main_loop_running) {
+		tm_shaper_config_set(tm_system, shaper_profile,
+				     &tm_node_obj->shaper_obj);
+		goto out;
+	}
```
psavol, line 3939: Goto is not needed. Simple If-else should be fine.

Even simpler would be to read main_loop_running once, and then based on that call signal_req/_done.

psavol (approved):

PR #1490, opened 2022-02-02
[PATCH v2] fixes for compiling with GCC-11

JannePeltonen:
A few minor fixes to make GCC 11.2 happy (on aarch64).


MatiasElo:
We should also add GCC-10 build tests to the CI (doesn't have to be done in this PR).

psavol (approved):

PR #1494, opened 2022-02-04
[PATCH v8] linux-gen: pktio: implement tx completion event

TuomasTaipale:
Implement event completion for transmitted and dropped packets.

v2:
- address Petri's and Matias' comments

v5:
- rebase
- address Petri's comments

v6:
- address Petri's comments

v8:
- rebase
- address Matias' comment
- add reviewed-by tags

### DIFF
--- a/config/odp-linux-generic.conf
+++ b/config/odp-linux-generic.conf
@@ -16,7 +16,7 @@
 
 # Mandatory fields
 odp_implementation = "linux-generic"
-config_file_version = "0.1.19"
+config_file_version = "0.1.20"
 
 # System options
 system: {
@@ -115,6 +115,10 @@ pktio: {
 	# (zero-copy DPDK, loop and ipc) that do not copy data ignore this
 	# option.
 	pktin_frame_offset = 0
+
+	# Pool size allocated for potential completion events for transmitted and
+	# dropped packets. Separate pool for different packet IO instances.
+	tx_compl_pool_size = 1024
 }
 
 # DPDK pktio options
--- a/platform/linux-generic/include/odp_packet_internal.h
+++ b/platform/linux-generic/include/odp_packet_internal.h
@@ -118,7 +118,8 @@ typedef struct ODP_ALIGNED_CACHE odp_packet_hdr_t {
 	/* Timestamp value */
 	odp_time_t timestamp;
 
-	/* Classifier destination queue */
+	/* Used as classifier destination queue, in IPsec inline input processing and as Tx
+	 * completion event queue. */
 	odp_queue_t dst_queue;
 
 	/* Reference count */
--- a/platform/linux-generic/include/odp_packet_io_internal.h
+++ b/platform/linux-generic/include/odp_packet_io_internal.h
@@ -90,6 +90,8 @@ struct pktio_entry {
 				uint8_t cls : 1;
 				/* Tx timestamp */
 				uint8_t tx_ts : 1;
+				/* Tx completion events */
+				uint8_t tx_compl : 1;
 			};
 		};
 	} enabled;
@@ -135,6 +137,9 @@ struct pktio_entry {
 	odp_pool_t pool;
 	odp_pktio_param_t param;
 	odp_pktio_capability_t capa;	/**< Packet IO capabilities */
+
+	/* Pool for Tx completion events */
+	odp_pool_t tx_compl_pool;
 
 	/* Storage for queue handles
 	 * Multi-queue support is pktio driver specific */
@@ -182,6 +187,8 @@ typedef struct {
 	struct {
 		/* Frame start offset from base pointer at packet input */
 		uint16_t pktin_frame_offset;
+		/* Pool size for potential completion events */
+		uint32_t tx_compl_pool_size;
 	} config;
 
 	pktio_entry_t entries[ODP_CONFIG_PKTIO_ENTRIES];
@@ -278,6 +285,11 @@ static inline int _odp_pktio_tx_ts_enabled(pktio_entry_t *entry)
 	return entry->s.enabled.tx_ts;
 }
 
+static inline int _odp_pktio_tx_compl_enabled(const pktio_entry_t *entry)
+{
+	return entry->s.enabled.tx_compl;
+}
+
 static inline void _odp_pktio_tx_ts_set(pktio_entry_t *entry)
 {
 	odp_time_t ts_val = odp_time_global();
@@ -331,6 +343,9 @@ int _odp_lso_num_packets(odp_packet_t packet, const odp_packet_lso_opt_t *lso_op
 int _odp_lso_create_packets(odp_packet_t packet, const odp_packet_lso_opt_t *lso_opt,
 			    uint32_t payload_len, uint32_t left_over_len,
 			    odp_packet_t pkt_out[], int num_pkt);
+
+void _odp_pktio_allocate_and_send_tx_compl_events(const pktio_entry_t *entry,
+						  const odp_packet_t packets[], int num);
 
 #ifdef __cplusplus
 }
--- a/platform/linux-generic/include/odp_packet_io_internal.h
+++ b/platform/linux-generic/include/odp_packet_io_internal.h
@@ -244,6 +244,10 @@ typedef struct pktio_if_ops {
 				    const odp_pktout_queue_param_t *p);
 } pktio_if_ops_t;
 
+typedef struct {
+	const void *user_ptr;
+} _odp_pktio_tx_compl_t;
+
 extern void *_odp_pktio_entry_ptr[];
 
 static inline pktio_entry_t *get_pktio_entry(odp_pktio_t pktio)
--- a/platform/linux-generic/odp_event.c
+++ b/platform/linux-generic/odp_event.c
@@ -76,6 +76,9 @@ void odp_event_free(odp_event_t event)
 	case ODP_EVENT_IPSEC_STATUS:
 		_odp_ipsec_status_free(_odp_ipsec_status_from_event(event));
 		break;
+	case ODP_EVENT_PACKET_TX_COMPL:
+		odp_packet_tx_compl_free(odp_packet_tx_compl_from_event(event));
+		break;
 	case ODP_EVENT_DMA_COMPL:
 		odp_dma_compl_free(odp_dma_compl_from_event(event));
 		break;
@@ -124,6 +127,8 @@ int odp_event_is_valid(odp_event_t event)
 	case ODP_EVENT_PACKET_VECTOR:
 		/* Fall through */
 	case ODP_EVENT_DMA_COMPL:
+		/* Fall through */
+	case ODP_EVENT_PACKET_TX_COMPL:
 		break;
 	default:
 		return 0;
--- a/platform/linux-generic/odp_packet.c
+++ b/platform/linux-generic/odp_packet.c
@@ -22,6 +22,7 @@
 #include <odp_event_internal.h>
 #include <odp_macros_internal.h>
 #include <odp_packet_internal.h>
+#include <odp_packet_io_internal.h>
 
 /* Inlined API functions */
 #include <odp/api/plat/byteorder_inlines.h>
@@ -2890,43 +2891,41 @@ uint64_t odp_packet_aging_tmo(odp_packet_t pkt)
 
 int odp_packet_tx_compl_request(odp_packet_t pkt, const odp_packet_tx_compl_opt_t *opt)
 {
-	(void)pkt;
-	(void)opt;
+	odp_packet_hdr_t *pkt_hdr = packet_hdr(pkt);
 
-	return -1;
-}
-
-int odp_packet_has_tx_compl_request(odp_packet_t pkt)
-{
-	(void)pkt;
+	pkt_hdr->p.flags.tx_compl = opt->mode == ODP_PACKET_TX_COMPL_ALL ? 1 : 0;
+	pkt_hdr->dst_queue = opt->queue;
 
 	return 0;
 }
 
-odp_packet_tx_compl_t odp_packet_tx_compl_from_event(odp_event_t ev)
+int odp_packet_has_tx_compl_request(odp_packet_t pkt)
 {
-	(void)ev;
+	odp_packet_hdr_t *pkt_hdr = packet_hdr(pkt);
 
-	return ODP_PACKET_TX_COMPL_INVALID;
-}
-
-odp_event_t odp_packet_tx_compl_to_event(odp_packet_tx_compl_t tx_compl)
-{
-	(void)tx_compl;
-
-	return ODP_EVENT_INVALID;
+	return pkt_hdr->p.flags.tx_compl;
 }
 
 void odp_packet_tx_compl_free(odp_packet_tx_compl_t tx_compl)
 {
-	(void)tx_compl;
+	if (odp_unlikely(tx_compl == ODP_PACKET_TX_COMPL_INVALID)) {
+		ODP_ERR("Bad TX completion event handle\n");
+		return;
+	}
+
+	odp_buffer_free((odp_buffer_t)tx_compl);
 }
 
 void *odp_packet_tx_compl_user_ptr(odp_packet_tx_compl_t tx_compl)
 {
-	(void)tx_compl;
+	if (odp_unlikely(tx_compl == ODP_PACKET_TX_COMPL_INVALID)) {
+		ODP_ERR("Bad TX completion event handle\n");
+		return NULL;
+	}
 
-	return NULL;
+	_odp_pktio_tx_compl_t *data = odp_buffer_addr((odp_buffer_t)tx_compl);
+
+	return (void *)(uintptr_t)data->user_ptr;
 }
 
 odp_packet_reass_status_t
--- a/platform/linux-generic/odp_packet_io.c
+++ b/platform/linux-generic/odp_packet_io.c
@@ -50,6 +50,12 @@
 /* Max wait time supported to avoid potential overflow */
 #define MAX_WAIT_TIME (UINT64_MAX / 1024)
 
+typedef struct {
+	const void *user_ptr;
+	odp_queue_t queue;
+	uint16_t idx;
+} tx_compl_info_t;
+
 /* Global variables */
 static pktio_global_t *pktio_global;
 
@@ -85,6 +91,20 @@ static int read_config_file(pktio_global_t *pktio_glb)
 	}
 
 	pktio_glb->config.pktin_frame_offset = val;
+	ODP_PRINT("  %s: %i\n", str, val);
+
+	str = "pktio.tx_compl_pool_size";
+	if (!_odp_libconfig_lookup_int(str, &val)) {
+		ODP_ERR("Config option '%s' not found.\n", str);
+		return -1;
+	}
+
+	if (val < 0) {
+		ODP_ERR("Bad value %s = %i\n", str, val);
+		return -1;
+	}
+
+	pktio_glb->config.tx_compl_pool_size = val;
 	ODP_PRINT("  %s: %i\n", str, val);
 
 	ODP_PRINT("\n");
@@ -246,6 +266,8 @@ static void init_pktio_entry(pktio_entry_t *entry)
 
 	/* Clear all flags */
 	entry->s.enabled.all_flags = 0;
+
+	entry->s.tx_compl_pool = ODP_POOL_INVALID;
 
 	odp_atomic_init_u64(&entry->s.stats_extra.in_discards, 0);
 	odp_atomic_init_u64(&entry->s.stats_extra.out_discards, 0);
@@ -493,6 +515,14 @@ int odp_pktio_close(odp_pktio_t hdl)
 	entry->s.num_in_queue  = 0;
 	entry->s.num_out_queue = 0;
 
+	if (entry->s.tx_compl_pool != ODP_POOL_INVALID) {
+		if (odp_pool_destroy(entry->s.tx_compl_pool)) {
+			unlock_entry(entry);
+			ODP_ERR("Unable to destroy Tx event completion pool\n");
+			return -1;
+		}
+	}
+
 	odp_spinlock_lock(&pktio_global->lock);
 	res = _pktio_close(entry);
 	odp_spinlock_unlock(&pktio_global->lock);
@@ -502,6 +532,30 @@ int odp_pktio_close(odp_pktio_t hdl)
 	unlock_entry(entry);
 
 	ODP_DBG("interface: %s\n", entry->s.name);
+
+	return 0;
+}
+
+static int configure_tx_event_compl(pktio_entry_t *entry)
+{
+	odp_pool_param_t params;
+	const char *name_base = "_odp_pktio_tx_compl_pool_";
+	char pool_name[ODP_POOL_NAME_LEN];
+
+	if (entry->s.tx_compl_pool != ODP_POOL_INVALID)
+		return 0;
+
+	snprintf(pool_name, sizeof(pool_name), "%s%d", name_base,
+		 odp_pktio_index(entry->s.handle));
+	odp_pool_param_init(&params);
+
+	params.type = ODP_POOL_BUFFER;
+	params.buf.num = pktio_global->config.tx_compl_pool_size;
+	params.buf.size = sizeof(_odp_pktio_tx_compl_t);
+	entry->s.tx_compl_pool = odp_pool_create(pool_name, &params);
+
+	if (entry->s.tx_compl_pool == ODP_POOL_INVALID)
+		return -1;
 
 	return 0;
 }
@@ -558,6 +612,14 @@ int odp_pktio_config(odp_pktio_t hdl, const odp_pktio_config_t *config)
 	entry->s.in_chksums.chksum.sctp = config->pktin.bit.sctp_chksum;
 
 	entry->s.enabled.tx_ts = config->pktout.bit.ts_ena;
+	entry->s.enabled.tx_compl = config->pktout.bit.tx_compl_ena;
+
+	if (entry->s.enabled.tx_compl)
+		if (configure_tx_event_compl(entry)) {
+			unlock_entry(entry);
+			ODP_ERR("Unable to configure Tx event completion\n");
+			return -1;
+		}
 
 	if (entry->s.ops->config)
 		res = entry->s.ops->config(entry, config);
@@ -944,6 +1006,11 @@ static inline int packet_vector_send(odp_pktout_queue_t pktout_queue, odp_event_
 		ODP_ASSERT(entry != NULL);
 
 		odp_atomic_add_u64(&entry->s.stats_extra.out_discards, discards);
+
+		if (odp_unlikely(_odp_pktio_tx_compl_enabled(entry)))
+			_odp_pktio_allocate_and_send_tx_compl_events(entry, &pkt_tbl[sent],
+								     discards);
+
 		odp_packet_free_multi(&pkt_tbl[sent], discards);
 	}
 
@@ -1816,6 +1883,11 @@ int odp_pktio_capability(odp_pktio_t pktio, odp_pktio_capability_t *capa)
 		capa->lso.proto.ipv4             = 1;
 		capa->lso.proto.custom           = 1;
 		capa->lso.mod_op.add_segment_num = 1;
+
+		capa->config.pktout.bit.tx_compl_ena = 1;
+		capa->tx_compl.queue_type_sched = 1;
+		capa->tx_compl.queue_type_plain = 1;
+		capa->tx_compl.mode_all = 1;
 	}
 
 	/* Packet vector generation is common for all pktio types */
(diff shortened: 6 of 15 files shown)


platform/linux-generic/include/odp_packet_internal.h
```diff
@@ -148,6 +148,12 @@ typedef struct ODP_ALIGNED_CACHE odp_packet_hdr_t {
 	/* LSO profile index */
 	uint8_t lso_profile_idx;
 
+	/* Tx completion event queue */
+	odp_queue_t tx_compl_queue;
+
+	/* Tx completion event mode */
+	odp_packet_tx_compl_mode_t tx_compl_mode;
```
psavol, line 155: Packet header struct should be kept as compact (small and packed) as possible. tx_compl_mode enum should not be needed, tx_compl flag in output flags should be enough. dst_queue could be reused for TX compl, since it's currently used only on packet input side (by classifier).

test/validation/api/pktio/pktio.c
```diff
@@ -3316,7 +3316,7 @@ static void pktio_test_pktout_compl(bool use_plain_queue)
 		opt.queue = compl_queue[i];
 		opt.mode = ODP_PACKET_TX_COMPL_ALL;
 		odp_packet_tx_compl_request(pkt_tbl[i], &opt);
-
+		CU_ASSERT(odp_packet_has_tx_compl_request(pkt_tbl[i]) == 1);
```
psavol, line 3319: Validation test changes should not be in the same commit with linux-gen implementation changes. It's easier to merge test changes to other implementations, when impl and test changes are separate.

platform/linux-generic/odp_packet.c
```diff
@@ -2887,43 +2887,40 @@ uint64_t odp_packet_aging_tmo(odp_packet_t pkt)
...
 
 void odp_packet_tx_compl_free(odp_packet_tx_compl_t tx_compl)
 {
-	(void)tx_compl;
+	if (odp_unlikely(tx_compl == ODP_PACKET_TX_COMPL_INVALID)) {
+		ODP_ERR("Bad TX completion event handle\n");
+		return;
+	}
+
+	odp_packet_free(odp_packet_from_event(odp_packet_tx_compl_to_event(tx_compl)));
```
psavol, line 2913: At this point event type should be TX_COMPL (could be checked in debug mode) and the type should be reset back to PACKET before the free call, so that the next alloc returns PACKET and not TX_COMPL event.

platform/linux-generic/odp_packet_io.c
```diff
@@ -581,10 +609,19 @@ int odp_pktio_config(odp_pktio_t hdl, const odp_pktio_config_t *config)
 	entry->s.in_chksums.chksum.sctp = config->pktin.bit.sctp_chksum;
 
 	entry->s.enabled.tx_ts = config->pktout.bit.ts_ena;
+	entry->s.enabled.tx_compl = config->pktout.bit.tx_compl_ena;
+
+	if (entry->s.enabled.tx_compl)
+		if (configure_tx_event_compl(entry)) {
+			ODP_ERR("Unable to configure Tx event completion\n");
+			res = -1;
+			goto out_unlock;
```
psavol, line 618: Since code with goto does not save lines and it does print while holding the lock. It would be better to do the same sequence as if-statement above does:

unlock_entry(entry);
ODP_ERR();
return -1;

platform/linux-generic/odp_packet_io.c
```diff
@@ -967,7 +1006,22 @@ static inline int packet_vector_send(odp_pktout_queue_t pktout_queue, odp_event_
 		ODP_ASSERT(entry != NULL);
 
 		odp_atomic_add_u64(&entry->s.stats_extra.out_discards, discards);
-		odp_packet_free_multi(&pkt_tbl[sent], discards);
+
+		if (odp_unlikely(_odp_pktio_tx_compl_enabled(entry)))
+			for (i = sent; i < num; i++) {
```
psavol, line 1011: If-statement has multiple lines, but misses bracets {}.  Code is more readable and robust to maintain, when for and if-else statements have their own bracets.

platform/linux-generic/include/odp_packet_io_internal.h
```diff
@@ -278,6 +283,39 @@ static inline void _odp_pktio_tx_ts_set(pktio_entry_t *entry)
 	odp_atomic_store_u64(&entry->s.tx_ts, ts_val.u64);
 }
 
+static inline int _odp_pktio_tx_compl_enabled(pktio_entry_t *entry)
+{
+	return entry->s.enabled.tx_compl;
+}
+
+static inline void _odp_pktio_send_tx_compl_ev(pktio_entry_t *entry, odp_queue_t queue,
+					       const void *user_ptr, odp_packet_t pkt)
```
psavol, line 292: Usage of this function would be simpler if it would be implemented in odp_packet_io.c. Prototype would contain only pktio entry and packet.

_odp_pktio_send_tx_compl_ev(pktio_entry_t *entry, odp_packet_t pkt)

platform/linux-generic/pktio/loop.c
```diff
@@ -331,11 +339,21 @@ static int loopback_send(pktio_entry_t *pktio_entry, int index ODP_UNUSED,
...
 		}
+
+		if (odp_unlikely(tx_compl_enabled && pkt_hdr->p.flags.tx_compl &&
+				 pkt_hdr->tx_compl_mode == ODP_PACKET_TX_COMPL_ALL)) {
+			compl_info[i].tx_compl_q = pkt_hdr->tx_compl_queue;
+			compl_info[i].user_ptr = pkt_hdr->user_ptr;
+		} else {
+			compl_info[i].tx_compl_q = ODP_QUEUE_INVALID;
+			compl_info[i].user_ptr = NULL;
+		}
```
psavol, line 356: Could these writes be avoided - so that compiler is happy and minimal amount of if statements / data initialization is added to the likely branch (not tx compl).

E.g. instead of compl_info[i].tx_compl_q signaling need to send event. There could be uint8_t send_compl[num] array of indexes that need tx compl. Another variable for number of packet/indexes (init to zero). In the common case number of indexes would remain zero, also when tx_compl_enabled.

platform/linux-generic/pktio/dpdk.c
```diff
@@ -986,21 +991,34 @@ static inline int pkt_to_mbuf_zero(pktio_entry_t *pktio_entry,
...
 				*tx_ts_idx = i + 1;
 		}
+
+		if (odp_unlikely(tx_compl_enabled && pkt_hdr->p.flags.tx_compl &&
+				 pkt_hdr->tx_compl_mode == ODP_PACKET_TX_COMPL_ALL)) {
+			pkt_info[i].tx_compl_q = pkt_hdr->tx_compl_queue;
+			pkt_info[i].user_ptr = pkt_hdr->user_ptr;
+		} else {
+			pkt_info[i].tx_compl_q = ODP_QUEUE_INVALID;
+			pkt_info[i].user_ptr = NULL;
```
MatiasElo, line 1020: If there would be an additional output argument (e.g. postprocess), where you would return the number of compl events/copies, this branch could be skipped. All writing could be done in the unlikely branch.

platform/linux-generic/pktio/dpdk.c
```diff
@@ -986,21 +991,34 @@ static inline int pkt_to_mbuf_zero(pktio_entry_t *pktio_entry,
 			if (odp_unlikely(chksum_enabled))
 				pkt_set_ol_tx(pktout_cfg, pktout_capa, pkt_hdr,
 					      mbuf, odp_packet_data(pkt));
+
+			pkt_info[i].is_copied = false;
```
MatiasElo, line 995: See comment below about additional argument.

platform/linux-generic/pktio/dpdk.c
```diff
@@ -2090,23 +2129,22 @@ static int dpdk_send(pktio_entry_t *pktio_entry, int index,
...
-				odp_packet_hdr_t *pkt_hdr = packet_hdr(pkt);
-
-				if (pkt_hdr->seg_count > 1) {
-					if (odp_likely(i < tx_pkts))
-						odp_packet_free(pkt);
-					else
-						rte_pktmbuf_free(tx_mbufs[i]);
-					freed++;
-				}
+		for (i = 0; i < mbufs; i++) {
```
MatiasElo, line 2132: This adds additional loop to the fast path, which could be avoided by returning the number of packets requiring postprocessing from `pkt_to_mbuf_zero()`.

platform/linux-generic/pktio/socket.c
```diff
@@ -468,9 +471,22 @@ static int sock_mmsg_send(pktio_entry_t *pktio_entry, int index ODP_UNUSED,
 	if (odp_unlikely(tx_ts_idx && i >= tx_ts_idx))
 		_odp_pktio_tx_ts_set(pktio_entry);
 
-	odp_ticketlock_unlock(&pkt_sock->tx_lock);
+	if (odp_unlikely(tx_compl_enabled))
+		for (j = 0; j < i; j++) {
```
psavol, line 475: This would simplify a lot with another _odp_pktio_send_tx_compl call: send tx compl or free pkt if no compl needed.

platform/linux-generic/pktio/dpdk.c
```diff
@@ -969,11 +974,11 @@ static inline int pkt_to_mbuf_zero(pktio_entry_t *pktio_entry,
 	uint8_t chksum_enabled = pktio_entry->s.enabled.chksum_insert;
 	uint8_t tx_ts_enabled = _odp_pktio_tx_ts_enabled(pktio_entry);
 	int i;
-	*copy_count = 0;
 
 	for (i = 0; i < num; i++) {
 		odp_packet_t pkt = pkt_table[i];
 		odp_packet_hdr_t *pkt_hdr = packet_hdr(pkt);
+
```
MatiasElo, line 981: Extra newline.

platform/linux-generic/include/odp_packet_internal.h
TuomasTaipale: `dst_queue` it is, enum ditched as well in v2.

test/validation/api/pktio/pktio.c
TuomasTaipale: Moved to separate commit in v2.

platform/linux-generic/odp_packet.c
TuomasTaipale: Done in v2.

platform/linux-generic/odp_packet_io.c
TuomasTaipale: Done in v2.

platform/linux-generic/odp_packet_io.c
TuomasTaipale: Moved to own function in v2.

platform/linux-generic/include/odp_packet_io_internal.h
TuomasTaipale: Yes, good idea, done in v2.

platform/linux-generic/pktio/loop.c
TuomasTaipale: Should be more optimized in v2.

platform/linux-generic/pktio/socket.c
TuomasTaipale: Done in v2.

platform/linux-generic/pktio/dpdk.c
TuomasTaipale: Should be more optimized in v2.

platform/linux-generic/pktio/dpdk.c
TuomasTaipale: Fixed in v2.

test/validation/api/pktio/pktio.c
```diff
@@ -3316,7 +3316,6 @@ static void pktio_test_pktout_compl(bool use_plain_queue)
 		opt.queue = compl_queue[i];
 		opt.mode = ODP_PACKET_TX_COMPL_ALL;
 		odp_packet_tx_compl_request(pkt_tbl[i], &opt);
-
```
psavol, line 3319: This patch (linux-gen: packet: implement tx completion event APIs) still this (accidental) modification of test/validation/api/pktio/pktio.c file.

test/validation/api/pktio/pktio.c
```diff
@@ -3285,7 +3285,7 @@ static void pktio_test_pktout_compl(bool use_plain_queue)
 			} else {
 				CU_ASSERT_FATAL(pktio_capa.tx_compl.queue_type_sched != 0);
 			}
-
+			CU_ASSERT_FATAL(pktio_capa.config.pktout.bit.tx_compl_ena == 1);
```
psavol, line 3288: Actually, we should remove .config field from pktio capa, as it makes automatically everything optional.

It should be enough for application to check only pktio_capa.tx_compl.mode_all. So, this check should not be added.

test/validation/api/pktio/pktio.c
```diff
@@ -3316,6 +3316,7 @@ static void pktio_test_pktout_compl(bool use_plain_queue)
 		opt.queue = compl_queue[i];
 		opt.mode = ODP_PACKET_TX_COMPL_ALL;
 		odp_packet_tx_compl_request(pkt_tbl[i], &opt);
+		CU_ASSERT(odp_packet_has_tx_compl_request(pkt_tbl[i]) == 1);
```
psavol, line 3319: By the spec value is non-zero, so check should be != 0.

test/validation/api/pktio/pktio.c
```diff
@@ -3421,6 +3435,7 @@ static int pktio_check_pktout_compl(bool plain)
 	(void)odp_pktio_close(pktio);
 
 	if (ret < 0 || !capa.tx_compl.mode_all ||
+	    !capa.config.pktout.bit.tx_compl_ena ||
```
psavol, line 3438: capa.tx_compl.mode_all check is enough

platform/linux-generic/odp_packet_io.c
```diff
@@ -529,6 +536,27 @@ int odp_pktio_close(odp_pktio_t hdl)
...
+
+	if (odp_pool_info(entry->s.pool, &info))
+		return -1;
+
+	odp_pool_param_init(&params);
+
+	params.type = ODP_POOL_PACKET;
+	params.pkt.num = info.pkt.max_num;
+	params.pkt.len = 1;
+	entry->s.tx_compl_pool = odp_pool_create("_odp_pktio_tx_compl_pool", &params);
```
psavol, line 552: Pool name could include pktio index. It would help debugging when pools have unique names.

test/validation/api/pktio/pktio.c
TuomasTaipale: OK, currently, pktio implementation specific capability is signaled with the bit (there's a hefty amount of other features as well using the `config` fields for capa). Anyways, I can switch to use the mode field in pktios.

platform/linux-generic/odp_packet_io.c
```diff
@@ -967,7 +1003,11 @@ static inline int packet_vector_send(odp_pktout_queue_t pktout_queue, odp_event_
 		ODP_ASSERT(entry != NULL);
 
 		odp_atomic_add_u64(&entry->s.stats_extra.out_discards, discards);
-		odp_packet_free_multi(&pkt_tbl[sent], discards);
+
+		if (odp_unlikely(_odp_pktio_tx_compl_enabled(entry)))
+			_odp_pktio_send_tx_compl_ev_pkt_multi(pkt_tbl, sent, num);
+		else
+			odp_packet_free_multi(&pkt_tbl[sent], discards);
```
psavol, line 1010: Code would be more readable if _odp_pktio_send_tx_compl_ev_pkt_multi() parameter would be just pkt[] and number of packets. Here it would be easy to see that tx_compl_ev_pkt_multi() and packet_free_multi() handle the same set of packets.

platform/linux-generic/pktio/loop.c
psavol: When thinking this again. Could all/most tx compls be handled at the top level (in odp_pktout_send()). That would need extra information from driver telling if "sent" packets were freed or not, and modification to drivers generally not to free sent packets. Steps on top level would be (if tx compl enabled):
1. Record which packets request comp event (and their dest queue/ctx ptr)
2. When send returns, check if sent packets had requests: reuse sent packet if it's still available, or alloc new event if packet not available

TuomasTaipale: They certainly could. The downside is a slower still slowpath, if this is acceptable then sure.

platform/linux-generic/odp_packet_io.c
```diff
@@ -306,6 +329,7 @@ static odp_pktio_t setup_pktio_entry(const char *name, odp_pool_t pool,
 	snprintf(pktio_entry->s.full_name, sizeof(pktio_entry->s.full_name), "%s", name);
 	pktio_entry->s.pool = pool;
 	memcpy(&pktio_entry->s.param, param, sizeof(odp_pktio_param_t));
+	pktio_entry->s.tx_compl_pool = ODP_POOL_INVALID;
```
psavol, line 332: This should be done now inside init_pktio_entry(), since the value does not require anything from the parameters.

config/odp-linux-generic.conf
```diff
@@ -115,6 +115,10 @@ pktio: {
 	# (zero-copy DPDK, loop and ipc) that do not copy data ignore this
 	# option.
 	pktin_frame_offset = 0
+
+	# Pool size allocated for potential completion events for transmitted and
+	# dropped packets. Separate pool for different packet IO instances.
+	num_tx_compl_evs = 1024
```
psavol, line 121: This could be "tx_compl_pool_size", which would high light that it's not total number of events, but number of events per pool/interface.

platform/linux-generic/odp_packet_io.c
```diff
@@ -506,6 +535,28 @@ int odp_pktio_close(odp_pktio_t hdl)
...
+	char pool_name[name_len];
+
+	snprintf(pool_name, name_len, "%s%d", name_base, odp_pktio_index(entry->s.handle));
+	odp_pool_param_init(&params);
+
+	params.type = ODP_POOL_BUFFER;
+	params.buf.num = pktio_global->config.num_tx_compl_evs;
+	/* Room for odp_packet_hdr_t::user_ptr */
+	params.buf.size = sizeof(void *);
+	entry->s.tx_compl_pool = odp_pool_create(pool_name, &params);
```
psavol, line 552: For better code readability, I'd add a tx_compl struct type containing void *user_ptr, and then use sizeof() that here (and cast buffer addr into that when accessing user_ptr elsewhere).

Packet interface could be reconfigured (... start() -> stop() -> config() -> start() ...), so here we would create the pool only if it's not yet created. Destroy can be in close().

platform/linux-generic/odp_packet_io.c
```diff
@@ -1816,6 +1901,11 @@ int odp_pktio_capability(odp_pktio_t pktio, odp_pktio_capability_t *capa)
 		capa->lso.proto.ipv4             = 1;
 		capa->lso.proto.custom           = 1;
 		capa->lso.mod_op.add_segment_num = 1;
+
+		capa->config.pktout.bit.tx_compl_ena = 1;
```
psavol, line 1905: capa.config should not be filled in, as we are planning to remove it. mode_all under should tell to the application that tx_compl is supported.

platform/linux-generic/odp_packet_io.c
```diff
@@ -2851,11 +2941,53 @@ uint64_t odp_pktin_wait_time(uint64_t nsec)
...
+	odp_packet_hdr_t *pkt_hdr;
+	odp_event_t ev;
+	int i;
+
+	for (i = 0; i < num; i++) {
+		pkt = packets[i];
+		pkt_hdr = packet_hdr(pkt);
+
+		if (odp_unlikely(pkt_hdr->p.flags.tx_compl)) {
+			ev = _odp_pktio_allocate_tx_compl_event(entry, odp_packet_user_ptr(pkt));
```
psavol, line 2957: Single odp_buffer_alloc_multi() is much more efficient than odp_buffer_alloc() in a loop.

So, first packets would be looped through to find those (couple of) packets that need tx_compl. Then alloc multi call, which determines up to which index we can send.

It's possible that we'll have other per packet checks done at top level send call, so this for-each-packet loop could be raised up into the top level.

platform/linux-generic/odp_packet_io.c
TuomasTaipale: For now, this needs to be here as application signals the activation of the feature with the bit when calling `odp_pktio_config()` and config and capa needs to match.

platform/linux-generic/odp_packet_io.c
TuomasTaipale: As you mentioned, there are likely just a few completion event requiring packets so probably not much difference allocating the events for those few already in the loop.

config/odp-linux-generic.conf
TuomasTaipale: Renamed in v5.

platform/linux-generic/odp_packet_io.c
TuomasTaipale: Using multi allocation in v5.

platform/linux-generic/odp_packet_io.c
```diff
@@ -2851,11 +2944,37 @@ uint64_t odp_pktin_wait_time(uint64_t nsec)
...
+}
+
+static void send_tx_compl_events(tx_compl_info_t *info, uint16_t num, odp_buffer_t bufs[],
+				 int num_sent)
+{
+	for (int i = 0; i < num; i++) {
+		if (info[i].idx < num_sent)
+			_odp_pktio_send_tx_compl_event(bufs[i], info[i].user_ptr, info[i].queue);
+		else
+			odp_buffer_free(bufs[i]);
```
psavol, line 2965: When we hit else there won't be anymore compl events to send (all following idx will be larger). We can call buffer_free_multi and break.

odp_buffer_free_multi(&bufs[i], num - i);
break;

platform/linux-generic/odp_packet_io.c
```diff
@@ -2869,7 +2988,27 @@ int odp_pktout_send(odp_pktout_queue_t queue, const odp_packet_t packets[],
...
 		_odp_dump_pcapng_pkts(entry, queue.index, packets, num);
 
-	return entry->s.ops->send(entry, queue.index, packets, num);
+	if (odp_unlikely(_odp_pktio_tx_compl_enabled(entry))) {
+		for (int i = 0; i < num; i++)
+			check_tx_compl_ev(packet_hdr(packets[i]), i, tx_compl_info, &num_tx_cevs);
+
+		if (num_tx_cevs > 0) {
+			const int num_alloc = odp_buffer_alloc_multi(entry->s.tx_compl_pool, bufs,
+								     num_tx_cevs);
```
psavol, line 2997: By the spec alloc_multi may return <0 (on failure). Seems that implementation does not do that, but crashes if pool is invalid.

Maybe a comment could be added saying that num_alloc >= 0 is assumed. I think LTO and static checkers would warn if we change alloc implementation later to return -1 (on invalid pool). You could try that with a LTO build.

platform/linux-generic/odp_packet_io.c
TuomasTaipale: Done in v6.

platform/linux-generic/odp_packet_io.c
TuomasTaipale: Ended up handling < 0 case in v6.

platform/linux-generic/odp_packet_io.c
```diff
@@ -2869,7 +2990,30 @@ int odp_pktout_send(odp_pktout_queue_t queue, const odp_packet_t packets[],
 	if (_ODP_PCAPNG)
 		_odp_dump_pcapng_pkts(entry, queue.index, packets, num);
 
-	return entry->s.ops->send(entry, queue.index, packets, num);
+	if (odp_unlikely(_odp_pktio_tx_compl_enabled(entry))) {
+		for (int i = 0; i < num; i++)
+			check_tx_compl_ev(packet_hdr(packets[i]), i, tx_compl_info, &num_tx_cevs);
+
+		if (num_tx_cevs > 0) {
```
psavol, line 2997: This could be also odp_unlikely(num_tx_cevs), as tx_compl requests should be rare.

platform/linux-generic/odp_packet_io.c
```diff
@@ -2869,7 +2990,30 @@ int odp_pktout_send(odp_pktout_queue_t queue, const odp_packet_t packets[],
...
+			int num_alloc = odp_buffer_alloc_multi(entry->s.tx_compl_pool, bufs,
+							       num_tx_cevs);
+
+			if (num_alloc < num_tx_cevs) {
+				if (num_alloc < 0)
```
psavol, line 3002: Also, these two are unlikely branches.

platform/linux-generic/odp_traffic_mngr.c
```diff
@@ -2250,6 +2254,16 @@ static void tm_send_pkt(tm_system_t *tm_system, uint32_t max_sends)
...
+				pktio_entry = get_pktio_entry(tm_system->pktout.pktio);
+				pkt_hdr = packet_hdr(odp_pkt);
+				if (odp_unlikely(_odp_pktio_tx_compl_enabled(pktio_entry) &&
+						 pkt_hdr->p.flags.tx_compl)) {
+					buf = odp_buffer_alloc(pktio_entry->s.tx_compl_pool);
+					if (buf != ODP_BUFFER_INVALID)
+						_odp_pktio_send_tx_compl_event(buf,
+									       pkt_hdr->user_ptr,
+									       pkt_hdr->dst_queue);
+				}
```
MatiasElo, line 2266: Instead of using `_odp_pktio_send_tx_compl_event()`, `allocate_and_send_tx_compl_events()` could be used, which minimizes the amount of code required in the TM module.

platform/linux-generic/odp_packet_io.c
TuomasTaipale: Done in v7.

platform/linux-generic/odp_packet_io.c
TuomasTaipale: Done in v7.

test/validation/api/pktio/pktio.c
```diff
@@ -3394,13 +3394,26 @@ static void pktio_test_pktout_compl(bool use_plain_queue)
...
 	for (i = 0; i < TX_BATCH_LEN; i++)
 		odp_queue_destroy(compl_queue[i]);
 
 	for (i = 0; i < num_ifaces; i++) {
 		CU_ASSERT_FATAL(odp_pktio_stop(pktio[i]) == 0);
 		CU_ASSERT_FATAL(odp_pktio_close(pktio[i]) == 0);
 	}
+
+	odp_schedule_resume();
 }
```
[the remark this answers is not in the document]
TuomasTaipale: Done in v7.

platform/linux-generic/odp_event.c
```diff
@@ -76,6 +76,9 @@ void odp_event_free(odp_event_t event)
 	case ODP_EVENT_IPSEC_STATUS:
 		_odp_ipsec_status_free(_odp_ipsec_status_from_event(event));
 		break;
+	case ODP_EVENT_PACKET_TX_COMPL:
```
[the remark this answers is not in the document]
TuomasTaipale: Done in v7.

platform/linux-generic/odp_packet_io.c
```diff
@@ -493,6 +518,11 @@ int odp_pktio_close(odp_pktio_t hdl)
 	entry->s.num_in_queue  = 0;
 	entry->s.num_out_queue = 0;
 
+	if (entry->s.tx_compl_pool != ODP_POOL_INVALID) {
+		res = odp_pool_destroy(entry->s.tx_compl_pool);
+		ODP_ASSERT(res == 0);
```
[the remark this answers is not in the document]
TuomasTaipale: Done in v7.

platform/linux-generic/odp_packet_io.c
```diff
@@ -506,6 +536,30 @@ int odp_pktio_close(odp_pktio_t hdl)
 	return 0;
 }
 
+static int configure_tx_event_compl(pktio_entry_t *entry)
+{
+	odp_pool_param_t params;
+	const char *name_base = "_odp_pktio_tx_compl_pool_";
+	const size_t name_len = strlen(name_base) + strlen(XSTR(ODP_CONFIG_PKTIO_ENTRIES)) + 1;
+	char pool_name[name_len];
```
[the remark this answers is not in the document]
TuomasTaipale: Simpler, sure. Done in v7.

platform/linux-generic/odp_traffic_mngr.c
TuomasTaipale: Done in v7.

platform/linux-generic/include/odp_packet_io_internal.h
```diff
@@ -328,6 +344,11 @@ int _odp_lso_create_packets(odp_packet_t packet, const odp_packet_lso_opt_t *lso
 			    uint32_t payload_len, uint32_t left_over_len,
 			    odp_packet_t pkt_out[], int num_pkt);
 
+void _odp_pktio_send_tx_compl_event(odp_buffer_t buf, const void *user_ptr, odp_queue_t queue);
```
MatiasElo, line 347: This function is no longer used by other modules, so it can be a static function in `odp_packet_io.c`.

MatiasElo (approved):
One minor comment but otherwise looking OK.

psavol (approved):

PR #1495, opened 2022-02-04
api: LSO with TM

psavol:
Enable usage of LSO with TM. Implementation is just a skeleton currently.


platform/linux-generic/odp_packet_io.c
```diff
@@ -3242,66 +3250,95 @@ static int pktout_send_lso(odp_pktout_queue_t queue, odp_packet_t packet,
...
-		num_free = num_pkt;
+
 		if (ret > 0) {
 			first_free = ret;
 			num_free = num_pkt - ret;
 		}
+
+		odp_packet_free_multi(&pkt_out[first_free], num_free);
+		return -1;
 	}
```
TuomasTaipale, line 3336: Perhaps this could be simplified to:
```C
if (ret < num_pkt) {
    const int cnt = ret >= 0 ? ret : 0; // Could also use MAX() macro defined in odp_macros_internal.h
    ODP_DBG("Packet send failed %i\n", ret);
    odp_packet_free_multi(&pkt_out[cnt], num_pkt - cnt);
    return -1;
}
```

platform/linux-generic/odp_traffic_mngr.c
```diff
@@ -4549,6 +4549,75 @@ int odp_tm_enq_multi(odp_tm_queue_t tm_queue, const odp_packet_t packets[],
...
+			goto error;
+		}
+
+		/* Create packets */
+		odp_packet_t pkt_out[num_pkt];
+
+		ret = _odp_lso_create_packets(pkt, opt_ptr, payload_len, left_over_len, pkt_out,
+					      num_pkt);
+
+		if (odp_unlikely(ret))
```
TuomasTaipale, line 4599: `_odp_lso_create_packets()` could be called inside the `if` making `ret` unnecessary.

platform/linux-generic/odp_traffic_mngr.c
```diff
@@ -4549,6 +4549,75 @@ int odp_tm_enq_multi(odp_tm_queue_t tm_queue, const odp_packet_t packets[],
...
+			goto error;
+
+		/* Enqueue resulting packets */
+		if (odp_unlikely(odp_tm_enq_multi(tm_queue, pkt_out, num_pkt)) != num_pkt) {
```
TuomasTaipale, line 4603: Should you have the `!= num_pkt` inside `odp_unlikely()`?

psavol: Yes, fixed in v3.

platform/linux-generic/odp_traffic_mngr.c
psavol: I had it that way, but a single line containing unlikely() and 6 parameters it's less readable than using the temp variable:
`
if (odp_unlikely(_odp_lso_create_packets(pkt, opt_ptr, payload_len, left_over_len,
							 pkt_out, num_pkt)))
						goto error;
`

TuomasTaipale (approved):

platform/linux-generic/odp_traffic_mngr.c
```diff
@@ -4549,6 +4549,75 @@ int odp_tm_enq_multi(odp_tm_queue_t tm_queue, const odp_packet_t packets[],
...
+
+		ret = _odp_lso_create_packets(pkt, opt_ptr, payload_len, left_over_len, pkt_out,
+					      num_pkt);
+
+		if (odp_unlikely(ret))
+			goto error;
+
+		/* Enqueue resulting packets */
+		if (odp_unlikely(odp_tm_enq_multi(tm_queue, pkt_out, num_pkt) != num_pkt)) {
+			ODP_DBG("TM enqueue failed on packet %i\n", i);
```
JannePeltonen, line 4604: This seems to leak some of the pkt_out packets in the error case.

include/odp/api/spec/traffic_mngr.h
```diff
@@ -2001,6 +2001,25 @@ int odp_tm_enq(odp_tm_queue_t tm_queue, odp_packet_t pkt);
...
+ * to the resulting packets.
+ *
+ * @param tm_queue  Specifies the tm_queue (and indirectly the TM system)
+ * @param packets   Array of packets to enqueue with segmentation offload
+ * @param num       Number of packets in the array
+ * @param lso_opt   LSO options to be used for all packets. When NULL, LSO options are
+ *                  read from each packet (see odp_packet_lso_request()).
+ *
+ * @return Number of packets successfully processed and consumed (0 ... num)
+ * @retval <0 on failure
```
JannePeltonen, line 2018: I think this should be "<= 0 on failure" to match odp_tm_enq_multi(). Otherwise the API looks ok to me.

psavol: Changed return value of enq_multi() in v5.

JannePeltonen (approved):

nithind1988:

test/validation/api/traffic_mngr/traffic_mngr.c
```diff
@@ -1220,6 +1250,64 @@ static uint32_t send_pkts(odp_tm_queue_t tm_queue, uint32_t num_pkts)
...
+
+	lso_opt.lso_profile     = profile;
+	lso_opt.payload_offset  = offset;
+	lso_opt.max_payload_len = max_len;
+
+	pkts_sent = 0;
+
+	for (i = 0; i < num_pkts; i++) {
+		pkt = xmt_pkts[i];
+		xmt_pkt_desc = &xmt_pkt_descs[i];
```
JannePeltonen, line 1281: The indexing here could start from num_pkts_sent instead of zero. It would make this function similar to the other send fumctions in that it can be called more than once after init_xmt_pkts().

psavol: Fixed in v7

PR #1498, opened 2022-02-09
[PATCH v15] linux-gen: pktio: classification result handling and parse cleanups

JereLeppanen:
Make sure that when classification result is to drop the packet, all pktios do so cleanly without incrementing error counters.

This is preparation for #1499.

v2:
- loop: Revert change to "failed" based on classification result.
- Separate patch for in_errors fix.

v3:
- Merge #1505.

v4:
- Add commit `linux-gen: pktio: loop: use in_discards counter`
- Add commit `linux-gen: pktio: loop: fix counters`
- Add commit `linux-gen: drop error packets according to pktin configuration`
- Petri's comments.

v5:
- Move parse function to new odp_parse.c file.

v8:
- Fix CI errors.

v9:
- Matias' comments.
- Add commit `linux-gen: parse: fix _odp_packet_parse_common() contiguous data requirement`
- Add commit `linux-gen: parse: document _odp_packet_parse_common() return values`

v10:
- Fix ABI compatibility compilation.

v11-v13:
- Petri's comments.

v14:
- Rebase.
- Remove commit `linux-gen: drop error packets according to pktin configuration`

v15:
- Rebase.
- Add review tags.


platform/linux-generic/pktio/loop.c
```diff
@@ -162,7 +162,6 @@ static int loopback_recv(pktio_entry_t *pktio_entry, int index ODP_UNUSED,
 						       pkt_len, seg_len,
 						       &new_pool, pkt_hdr, true);
 			if (ret) {
-				failed++;
```
psavol, line 165: It would good to separation parse and cls as part of this PR. classify_packet() could fail for other reasons than "drop CoS", but in practice it does not do that today as it does not check return value of parse. Also, packet_parse_reset() is now called twice on this CLS path.

It would be cleaner do parse call outside of classify_packet(). That way it would be clean to increment this counter when there are parse errors, but not when a "drop CoS" is hit.

JereLeppanen: These are addressed in #1499 and #1505.

platform/linux-generic/pktio/pcap.c
```diff
@@ -294,9 +294,18 @@ static int pcapif_recv_pkt(pktio_entry_t *pktio_entry, int index ODP_UNUSED,
 		if (pktio_cls_enabled(pktio_entry)) {
 			odp_packet_t new_pkt;
 
+			packet_parse_reset(pkt_hdr, 1);
+			packet_set_len(pkt_hdr, pkt_len);
```
psavol, line 299: Seems that this set length call is not needed, as alloc and pull_head should have set it correctly.

platform/linux-generic/include/odp_parse_internal.h
```diff
@@ -0,0 +1,555 @@
...
+	return ethtype;
+}
+
+#define PARSE_IPV4_BYTES (0xfU * 4) /* max IPv4 header length with options */
+/**
+ * Parser helper function for IPv4
+ *
+ * Requires up to PARSE_IPV4_BYTES bytes of contiguous packet data.
+ */
+static inline uint8_t parse_ipv4(packet_parser_t *prs, const uint8_t **parseptr,
```
MatiasElo, line 160: Maybe it would be better to move these rather large functions into a source file (e.g. odp_parse.c) and expose only the necessary functions to other modules (with _odp prefixes).

JereLeppanen: Yes, if the effect on inlining is acceptable.

platform/linux-generic/pktio/netmap.c
```diff
@@ -847,17 +848,17 @@ static inline int netmap_pkt_to_odp(pktio_entry_t *pktio_entry,
...
-		pkt = pkt_tbl[i];
+		pkt = pkt_tbl[num_rx];
 		pkt_hdr = packet_hdr(pkt);
 		pull_tail(pkt_hdr, max_len - len);
 		if (frame_offset)
 			pull_head(pkt_hdr, frame_offset);
 
 		if (odp_packet_copy_from_mem(pkt, 0, len, slot.buf) != 0)
-			goto fail;
+			continue;
```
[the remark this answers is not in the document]
JereLeppanen: v4.

platform/linux-generic/pktio/pcap.c
JereLeppanen: In v4, removed these and a packet_parse_reset() call in sock_mmsg_recv().

platform/linux-generic/include/odp_parse_internal.h
JereLeppanen: v5.

platform/linux-generic/include/odp_parse_internal.h
```diff
@@ -0,0 +1,122 @@
...
+ * Parse common packet headers up to given layer
+ *
+ * The function expects at least PACKET_PARSE_SEG_LEN bytes of data to be
+ * available from the ptr. Also parse metadata must be already initialized.
+ */
+static inline int _odp_packet_parse_common(packet_parser_t *prs,
+					   const uint8_t *ptr,
+					   uint32_t frame_len, uint32_t seg_len,
+					   int layer,
+					   odp_proto_chksums_t chksums)
```
MatiasElo: It would be helpful to document the return values.

MatiasElo (commented):
Couple minor comments but otherwise looking OK.

platform/linux-generic/pktio/socket.c
```diff
@@ -274,9 +274,16 @@ static int sock_mmsg_recv(pktio_entry_t *pktio_entry, int index ODP_UNUSED,
 			if (msgvec[i].msg_hdr.msg_iov->iov_len < pkt_len)
 				seg_len = msgvec[i].msg_hdr.msg_iov->iov_len;
 
-			if (_odp_cls_classify_packet(pktio_entry, base, pkt_len,
-						     seg_len, &pool, pkt_hdr,
-						     true)) {
+			packet_set_len(pkt_hdr, pkt_len);
```
psavol, line 278: It seems that odp_packet_trunc_tail() should be done instead of packet_set_len(), otherwise, e.g. tailroom will not be updated. So, trunc_tail() call (from couple of line below) could be moved up to before pktio_cls_enabled() line. It would need to be called anyway.

platform/linux-generic/pktio/loop.c
```diff
@@ -114,7 +114,7 @@ static int loopback_recv(pktio_entry_t *pktio_entry, int index ODP_UNUSED,
 	odp_time_t ts_val;
 	odp_time_t *ts = NULL;
 	int num_rx = 0;
-	int failed = 0;
+	int packets = 0, octets = 0, errors = 0;
```
psavol, line 117: Since all APIs use uint32_t or uint64_t for byte counts, it would be good to use e.g. uint32_t here for octets count. By the C spec, int could be in theory 16 bits.

platform/linux-generic/odp_parse.c
```diff
@@ -0,0 +1,460 @@
+/* Copyright (c) 2013-2018, Linaro Limited
+ * Copyright (c) 2019-2022, Nokia
+ * All rights reserved.
+ *
```
psavol, line 4: _odp_packet_parse_layer() should move into the new odp_parse.c also.

platform/linux-generic/include/odp_packet_internal.h
```diff
@@ -475,6 +468,9 @@ int _odp_packet_tcp_chksum_insert(odp_packet_t pkt);
 int _odp_packet_udp_chksum_insert(odp_packet_t pkt);
 int _odp_packet_sctp_chksum_insert(odp_packet_t pkt);
 
+int packet_l4_chksum(odp_packet_hdr_t *pkt_hdr, odp_proto_chksums_t chksums,
```
psavol: Function name should be "_odp_" prefixed when it's visible in symbol table (moved from static to non-static)

platform/linux-generic/include/odp_parse_internal.h
```diff
@@ -0,0 +1,104 @@
...
+#define PARSE_UDP_BYTES (sizeof(_odp_udphdr_t) + 4)
+#define PARSE_SCTP_BYTES (sizeof(_odp_sctphdr_t))
+
+#define PARSE_L3_L4_BYTES (MAX(PARSE_IPV4_BYTES, PARSE_IPV6_BYTES) + \
+			   MAX3(PARSE_TCP_BYTES, PARSE_UDP_BYTES, PARSE_SCTP_BYTES))
+
+uint16_t parse_eth(packet_parser_t *prs, const uint8_t **parseptr,
+		   uint32_t *offset, uint32_t frame_len);
+
+int packet_parse_common_l3_l4(packet_parser_t *prs, const uint8_t *parseptr,
```
psavol: Also, these two functions would need _odp prefix

MatiasElo (approved):

psavol (approved):

PR #1499, opened 2022-02-09
[PATCH v13] api: cls: add cos action and cos specific counters

JereLeppanen:
Add an action parameter to CoS parameters (odp_cls_cos_param_t). Action may be to accept or drop the packet classified to the CoS.

Add CoS specific statistics counters (odp_cls_cos_stats_t) and matching capabilities (in odp_cls_stats_capability_t). Statistics counters can be read with odp_cls_cos_stats().

Branched from #1498.

v1 - v4:
- Add commit linux-gen: cls: return odp_cos_action_t from _odp_cls_classify_packet()
- loop: Fix in_errors.

v5:
- Rebase on #1498.
- Rename ACTION_DROP to ACTION_ENQUEUE and tweak documentation accordingly.
- Add mention of final matching CoS.

v6:
- Add capability for max number of CoS with statistics.
- Add validation test for max number of CoS.
- Add validation test for max number of CoS with statistics.

v7:
- Rebase.

v8:
- Petri's comments.

v9 - v11:
- Petri's comments.

v12:
- Rebase.
- Add review tags.

### DIFF
--- a/include/odp/api/spec/classification.h
+++ b/include/odp/api/spec/classification.h
@@ -36,9 +36,7 @@ extern "C" {
 
 /**
  * @def ODP_COS_INVALID
- * This value is returned from odp_cls_cos_create() on failure,
- * May also be used as a sink class of service that
- * results in packets being discarded.
+ * This value is returned from odp_cls_cos_create() on failure.
  */
 
 /**
@@ -315,10 +313,42 @@ typedef enum {
 } odp_cos_hdr_flow_fields_t;
 
 /**
+ * Enumeration of actions for CoS.
+ */
+typedef enum {
+	/**
+	 * Enqueue packet
+	 *
+	 * Packets that arrive in the CoS are enqueued to a destination queue.
+	 */
+	ODP_COS_ACTION_ENQUEUE,
+
+	/**
+	 * Drop packet
+	 *
+	 * Packets that arrive in the CoS are dropped. Packets are freed into
+	 * their originating pool.
+	 */
+	ODP_COS_ACTION_DROP,
+} odp_cos_action_t;
+
+/**
  * Class of service parameters
  * Used to communicate class of service creation options
  */
 typedef struct odp_cls_cos_param {
+	/** Action to take. When action is ODP_COS_ACTION_DROP, all the other
+	 * parameters are ignored.
+	 *
+	 * The final match in the CoS chain defines the action for a packet.
+	 * I.e. packet is dropped only when the CoS of the last matching rule
+	 * has drop action. Actions in the previous CoSes in the chain are
+	 * ignored.
+	 *
+	 * Default is ODP_COS_ACTION_ENQUEUE.
+	 */
+	odp_cos_action_t action;
+
 	/** Number of queues to be linked to this CoS.
 	 * If the number is greater than 1 then hashing is enabled.
 	 * If number is equal to 1 then hashing is disabled.
@@ -391,6 +421,9 @@ int odp_cls_capability(odp_cls_capability_t *capability);
 /**
  * Create a class-of-service
  *
+ * Depending on the action parameter, packets to the CoS are either enqueued to
+ * a destination queue, or dropped.
+ *
  * The use of class-of-service name is optional. Unique names are not required.
  * Use odp_cls_cos_param_init() to initialize parameters into their default
  * values.
@@ -401,10 +434,6 @@ int odp_cls_capability(odp_cls_capability_t *capability);
  *
  * @retval Class-of-service handle
  * @retval ODP_COS_INVALID on failure.
- *
- * @note ODP_QUEUE_INVALID and ODP_POOL_INVALID are valid values for queue
- * and pool associated with a class of service. When either of these values
- * is configured as INVALID packets assigned to the CoS get dropped.
  */
 odp_cos_t odp_cls_cos_create(const char *name,
 			     const odp_cls_cos_param_t *param);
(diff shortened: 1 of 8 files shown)


include/odp/api/spec/classification.h
```diff
@@ -346,6 +395,12 @@ typedef struct odp_cls_cos_param {
 	 */
 	odp_cos_action_t action;
 
+	/** Enable statistics. If true, counters are incremented when packets
+	 * are classified to the CoS. Default is false. @see
+	 * odp_cls_cos_stats().
+	 */
+	odp_bool_t stats_enable;
```
psavol, line 402: If implementation has limited number of CoS that can have stats enabled, how to indicate that implementation is out-of-stats resources? Maybe new odp_cls_cos_create() return value.

include/odp/api/spec/classification.h
```diff
@@ -317,11 +315,37 @@ typedef enum {
...
+	ODP_COS_ACTION_DROP,
+} odp_cos_action_t;
+
 /**
  * Class of service parameters
  * Used to communicate class of service creation options
  */
 typedef struct odp_cls_cos_param {
+	/** Action to take. When action is ODP_COS_ACTION_DROP, all the other
+	 * parameters are ignored. Default is ODP_COS_ACTION_ACCEPT.
```
psavol, line 394: Another naming option for accept CoS could be ACTION_ENQUEUE. It would allow us to add other accept type actions later.

include/odp/api/spec/classification.h
```diff
@@ -317,11 +315,37 @@ typedef enum {
...
+ * Enumeration of actions for CoS.
+ */
+typedef enum {
+	/**
+	 * Accept packet
+	 *
+	 * Packets that arrive in the CoS are accepted and their processing
+	 * continues towards other chained CoSes or the destination queue.
+	 */
+	ODP_COS_ACTION_ACCEPT,
```
psavol, line 377: Another naming option for accept CoS could be ACTION_ENQUEUE. It would allow us to add other accept type actions later.

In general, action description would need to say in someway that: the last "matching" CoS defines the action for a packet. I.e. packet is dropped only when the CoS of the last matching rule has drop action (there may be other CoSes with drop action on the way).

include/odp/api/spec/classification.h
JereLeppanen: Capability added in v6.

include/odp/api/spec/classification.h
JereLeppanen: ACTION_ENQUEUE in v5.

test/validation/api/classification/odp_classification_basic.c
```diff
@@ -52,6 +52,43 @@ static void classification_test_create_cos(void)
...
+
+	odp_cos_t cos[num];
+
+	for (int i = 0; i < num; i++) {
+		odp_cls_cos_param_init(&cls_param);
+		cls_param.action = ODP_COS_ACTION_DROP;
+		cls_param.stats_enable = stats;
+
+		cos[i] = odp_cls_cos_create(NULL, &cls_param);
+		CU_ASSERT(cos[i] != ODP_COS_INVALID);
```
psavol, line 75: For easier debugging, I'd suggest to break the loop after the first create fails and print values of i and num with ODPH_ERR(). Destroy loop would then run up to num created.

test/validation/api/classification/odp_classification_basic.c
```diff
@@ -52,6 +52,43 @@ static void classification_test_create_cos(void)
...
+	}
+
+	for (int i = 0; i < num; i++)
+		odp_cos_destroy(cos[i]);
```
psavol, line 79: Return value needs to be checked

include/odp/api/spec/classification.h
```diff
@@ -253,6 +302,10 @@ typedef struct odp_cls_capability_t {
 	/** Maximum number of CoS supported */
 	unsigned int max_cos;
 
+	/** Maximum number of CoSes that can have statistics enabled at the same
+	 * time. */
+	uint32_t max_cos_stats;
```
psavol, line 307: This could note in some form that: CoS level statistics are not supported when zero

psavol (approved):

include/odp/api/spec/classification.h
```diff
@@ -176,6 +176,27 @@ typedef struct odp_bp_param_t {
 	odp_threshold_t threshold;
 } odp_bp_param_t;
 
+/**
+ * Classifier CoS specific statistics counters
+ *
+ * Counters are incremented per packet classified to the CoS.
```
psavol, line 182: This sentence needs clarification: if all (stats enabled) CoS'es on the way from default CoS to the "resulting" CoS increment their counters for a packet, or if only the "resulting" CoS increments its counters.

include/odp/api/spec/classification.h
```diff
@@ -176,6 +176,27 @@ typedef struct odp_bp_param_t {
...
+typedef struct odp_cls_cos_stats_t {
+	/** Number of octets in classified packets. In case of Ethernet, packet
+	 * size includes MAC header. */
+	uint64_t octets;
+
+	/** Number of classified packets. */
+	uint64_t packets;
+
+	/** Number of packets discarded due to a drop action. */
+	uint64_t discards;
```
psavol, line 193: Another option would be to increment packets/octets normally also in case of drop CoS. Application knows which CoSes are setup with drop and enqueue actions.

include/odp/api/spec/classification.h
JereLeppanen: Chose the former alternative in v9.

include/odp/api/spec/classification.h
JereLeppanen: In v9.

psavol (approved):

PR #1502, opened 2022-02-11
[PATCH v4] example: sysinfo: update capa prints

psavol:


example/sysinfo/odp_sysinfo.c
```diff
@@ -465,62 +464,42 @@ int main(void)
...
 	printf("  POOL\n");
 	printf("    max_pools:            %u\n", pool_capa.max_pools);
 	printf("    buf.max_pools:        %u\n", pool_capa.buf.max_pools);
-	printf("    buf.max_align:        %" PRIu32 " B\n",
-	       pool_capa.buf.max_align);
-	printf("    buf.max_size:         %" PRIu32 " kB\n",
-	       pool_capa.buf.max_size / KB);
-	printf("    buf.max_num:          %" PRIu32 "\n",
-	       pool_capa.buf.max_num);
+	printf("    buf.max_align:        %u B\n", pool_capa.buf.max_align);
```
JannePeltonen, line 493: > Shorten printf lines by not using PRIu32 as uint32_t is in practice equivalent to unsigned int.

I think this is quite questionable, especially in an example application. If you keep the change, I would suggest adding a string comparison that PRIu32 is the same as "u" and failing in an error if they are not.

psavol: Agree that it's not 100% type safe, but such prints have not caused problems in practice. It would be a problem for a system where "unsigned int" is smaller than 32bits. If such systems still exist, may those have other problems as well (to run modern Linux, tools, etc). Also, our CI catches real size errors (e.g. %u with uint64_t).

example/sysinfo/odp_sysinfo.c
```diff
@@ -399,6 +401,11 @@ int main(void)
 		return -1;
 	}
 
+	if (odp_dma_capability(&dma_capa)) {
```
MatiasElo, line 404: For completeness `odp_cls_capability()` and `odp_comp_capability()` could also be added.

psavol: Added in v2

JannePeltonen (approved):
Looks ok to me. I suggested adding a couple more prints, but the changes in this PR are valid also without them.

example/sysinfo/odp_sysinfo.c
```diff
@@ -477,14 +508,61 @@ int main(void)
...
+	printf("\n");
+	printf("  COMPRESSION\n");
+	printf("    max_sessions:         %u\n", comp_capa.max_sessions);
+	printf("    compl_algos:          0x%x\n", comp_capa.comp_algos.all_bits);
+	printf("    hash_algos:           0x%x\n", comp_capa.hash_algos.all_bits);
+	printf("    sync support:         %i\n", comp_capa.sync);
+	printf("    async support:        %i\n", comp_capa.async);
+
+	printf("\n");
+	printf("  DMA\n");
```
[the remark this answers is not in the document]
psavol: Added in v4.

example/sysinfo/odp_sysinfo.c
```diff
@@ -498,21 +576,41 @@ int main(void)
...
-	       timer_capa.highest_res_ns);
+	printf("  STASH\n");
+	printf("    max_stashes_any_type: %u\n", stash_capa.max_stashes_any_type);
+	printf("    max_stashes:          %u\n", stash_capa.max_stashes);
+	printf("    max_num_obj:          %" PRIu64 "\n", stash_capa.max_num_obj);
+	printf("    max_obj_size:         %u\n", stash_capa.max_obj_size);
+	printf("    max_cache_size:       %u\n", stash_capa.max_cache_size);
 
+	printf("\n");
+	printf("  TIMER (ODP_CLOCK_DEFAULT)\n");
```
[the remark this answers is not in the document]
psavol: Added in v4. Also added those for crypto and IPSEC.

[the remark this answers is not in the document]
JannePeltonen: v4 looks good to me.

PR #1504, opened 2022-02-11
[PATCH v3] validation: crypto: test hash result within authenticated range

JannePeltonen:
Test that hash result field that overlaps the authenticated range is handled according to the latest API spec: The field must be zeroed before hash verification in decode sessions and left alone before hash generation in encode sessions.

### DIFF
--- a/test/validation/api/crypto/odp_crypto_test_inp.c
+++ b/test/validation/api/crypto/odp_crypto_test_inp.c
@@ -283,6 +283,7 @@ static int alg_packet_op(odp_packet_t pkt,
 			 unsigned int hash_result_offset)
 {
 	int rc;
+	odp_event_t event;
 	odp_crypto_packet_result_t result;
 	odp_crypto_packet_op_param_t op_params;
 	odp_event_subtype_t subtype;
@@ -303,10 +304,31 @@ static int alg_packet_op(odp_packet_t pkt,
 
 	op_params.hash_result_offset = hash_result_offset;
 
-	rc = odp_crypto_op(&pkt, &out_pkt, &op_params, 1);
-	if (rc <= 0) {
-		CU_FAIL("Failed odp_crypto_packet_op()");
-		return rc;
+	if (suite_context.op_mode == ODP_CRYPTO_SYNC) {
+		rc = odp_crypto_op(&pkt, &out_pkt, &op_params, 1);
+		if (rc <= 0) {
+			CU_FAIL("Failed odp_crypto_packet_op()");
+			return rc;
+		}
+	} else {
+		rc = odp_crypto_op_enq(&pkt, &pkt, &op_params, 1);
+		if (rc <= 0) {
+			CU_FAIL("Failed odp_crypto_op_enq()");
+			return rc;
+		}
+
+		/* Get crypto completion event from compl_queue. */
+		CU_ASSERT_FATAL(NULL != suite_context.compl_queue_deq);
+		do {
+			event = suite_context.compl_queue_deq();
+		} while (event == ODP_EVENT_INVALID);
+
+		CU_ASSERT(ODP_EVENT_PACKET == odp_event_type(event));
+		CU_ASSERT(ODP_EVENT_PACKET_CRYPTO == odp_event_subtype(event));
+		CU_ASSERT(ODP_EVENT_PACKET == odp_event_types(event, &subtype));
+		CU_ASSERT(ODP_EVENT_PACKET_CRYPTO == subtype);
+
+		pkt = odp_crypto_packet_from_event(event);
 	}
 
 	CU_ASSERT(out_pkt == pkt);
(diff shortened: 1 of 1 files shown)


test/validation/api/crypto/odp_crypto_test_inp.c
```diff
@@ -407,6 +407,40 @@ static int alg_packet_op_enq(odp_packet_t pkt,
...
+		     uint8_t *cipher_iv,
+		     uint8_t *auth_iv,
+		     odp_packet_data_range_t *cipher_range,
+		     odp_packet_data_range_t *auth_range,
+		     uint8_t *aad,
+		     unsigned int hash_result_offset)
+{
+	int rc;
+
+	if (!suite_context.packet)
```
TuomasTaipale, line 372: The data that is used to choose which operation to call seems to be known already beforehand. Perhaps `alg_test_param_t` could include a function pointer to the relevant operation.

test/validation/api/crypto/odp_crypto_test_inp.c
```diff
@@ -688,14 +713,20 @@ static void alg_test(odp_crypto_op_t op,
 		     odp_bool_t bit_mode)
 {
 	unsigned int initial_num_failures = CU_get_number_of_failures();
+	hash_test_mode_t hash_mode;
```
TuomasTaipale, line 716: Could be defined already here:
```C
const hash_test_mode_t hash_mode = digest_offset < ref->length ? HASH_OVERLAP : HASH_NO_OVERLAP;
```

test/validation/api/crypto/odp_crypto_test_inp.c
```diff
@@ -1095,12 +1132,187 @@ static void test_capability(void)
...
+	const uint32_t auth_bytes = 100;
+	uint32_t enc_digest_offset = auth_bytes;
+	odp_packet_data_range_t cipher_range = {.offset = 0, .length = 0};
+	odp_packet_data_range_t auth_range = {.offset = 0};
+
+	ref->auth_key_length = capa->key_len;
+	ref->auth_iv_length = capa->iv_len;
+	ref->digest_length = capa->digest_len;
+	if (capa->bit_mode)
+		ref->length = auth_bytes * 8;
```
TuomasTaipale, line 1163:
```C
ref->length = capa->bit_mode ? auth_bytes * 8 : auth_bytes;
```

test/validation/api/crypto/odp_crypto_test_inp.c
```diff
@@ -1095,12 +1132,187 @@ static void test_capability(void)
...
+	else
+		ref->length = auth_bytes;
+	auth_range.length = ref->length;
+
+	for (n = 0; n < ref->auth_key_length; n++)
+		ref->auth_key[n] = n;
+	for (n = 0; n < ref->auth_iv_length; n++)
+		ref->auth_iv[n] = n;
+	for (n = 0; n < ref->length; n++)
```
TuomasTaipale, line 1172: A fill-like helper function could be useful for these.

test/validation/api/crypto/odp_crypto_test_inp.c
```diff
@@ -1095,12 +1132,187 @@ static void test_capability(void)
...
+	CU_ASSERT(rc == 0);
+
+	session = session_create(ODP_CRYPTO_OP_ENCODE, ODP_CIPHER_ALG_NULL,
+				 auth, ref, PACKET_IV, HASH_NO_OVERLAP);
+
+	if (crypto_op(pkt, &ok, session,
+		      ref->cipher_iv, ref->auth_iv,
+		      &cipher_range, &auth_range,
+		      ref->aad, enc_digest_offset))
+		return;
```
TuomasTaipale, line 1190: Should there be an assert in the failure branch as well?

test/validation/api/crypto/odp_crypto_test_inp.c
JannePeltonen: Will change in v2.

test/validation/api/crypto/odp_crypto_test_inp.c
JannePeltonen: Will change in v2.

test/validation/api/crypto/odp_crypto_test_inp.c
JannePeltonen: Will add in v2.

test/validation/api/crypto/odp_crypto_test_inp.c
JannePeltonen: There are more specific asserts down in alg_op() and alg_packet_op() so an assert has already fired in the failure case here.

JannePeltonen:
v2: Made changes based on review comments and added a few length checks.

test/validation/api/crypto/odp_crypto_test_inp.c
TuomasTaipale: The benefit would be that test cases would use the operation function directly instead of a wrapper that fans out to other functions that are chosen based on information that is set elsewhere.

TuomasTaipale (approved):

JannePeltonen:
v3: rebased, added review tags

PR #1505, opened 2022-02-15
[PATCH v2] linux-gen: packet parse related cleanups

JereLeppanen:
Move parse call out of _odp_cls_classify_packet() and move parse code to its own header file.

Branched from #1498.

v2:
- Rebase on #1498.


JereLeppanen:
Merged into #1498.

PR #1506, opened 2022-02-15
[PATCH v5] test: performance: ipsec: fix a minor bug and use schedule_multi

JannePeltonen:
This PR has the following patches. @anoobj , maybe you could do the review since you are adding the burst support.

    test: performance: ipsec: fix IPsec processing error checking

    Do not use odp_packet_has_error() for checking whether outbound
    IPsec processing succeeded. The flag is not supposed to indicate
    any IPsec errors other than inner packet parse errors in IPsec
    decapsulation.

    Do not ignore the result status returned by odp_ipsec_result(),
    but print an error message if the odp_ipsec_result() call fails
    or indicates an error.

    test: performance: ipsec: use schedule_multi and queue_deq_multi

    Dequeue/schedule multiple IPsec-processed packets at a time by using
    odp_queue_deq_multi() and odp_schedule_multi().

### DIFF
--- a/test/performance/odp_ipsec.c
+++ b/test/performance/odp_ipsec.c
@@ -594,6 +594,14 @@ static uint8_t test_data[] = {
 	0x08, 0x00, 0xfb, 0x37, 0x12, 0x34, 0x00, 0x00
 };
 
+static inline void debug_packets(int debug, odp_packet_t *pkt, int num_pkts)
+{
+	if (odp_likely(!debug))
+		return;
+	for (int i = 0; i < num_pkts; i++)
+		odp_packet_print_data(pkt[i], 0, odp_packet_len(pkt[i]));
+}
+
 static int
 make_packet_multi(odp_pool_t pkt_pool, unsigned int payload_length,
 		  odp_packet_t pkt[], int num)
@@ -645,6 +653,7 @@ run_measure_one(ipsec_args_t *cargs,
 	int in_flight, pkts_allowed, num_out, num_pkts, rc = 0;
 	const int max_in_flight = cargs->in_flight;
 	const int burst_size = cargs->burst_size;
+	const int debug = cargs->debug_packets;
 	odp_ipsec_out_param_t param;
 	odp_pool_t pkt_pool;
 
@@ -693,12 +702,7 @@ run_measure_one(ipsec_args_t *cargs,
 							   num_pkts)))
 				return -1;
 
-			if (cargs->debug_packets) {
-				for (i = 0; i < num_pkts; i++)
-					odp_packet_print_data(pkt[i], 0,
-							      odp_packet_len(pkt[i]));
-			}
-
+			debug_packets(debug, pkt, num_pkts);
 			num_out = num_pkts;
 
 			rc = odp_ipsec_out(pkt, num_pkts,
@@ -715,11 +719,7 @@ run_measure_one(ipsec_args_t *cargs,
 
 			packets_sent += rc;
 			packets_received += num_out;
-			if (cargs->debug_packets) {
-				for (i = 0; i < num_out; i++)
-					odp_packet_print_data(out_pkt[i], 0,
-							      odp_packet_len(out_pkt[i]));
-			}
+			debug_packets(debug, out_pkt, num_out);
 
 			if (odp_unlikely(rc != num_pkts))
 				odp_packet_free_sp(&pkt[rc], num_pkts - rc);
@@ -742,6 +742,7 @@ run_measure_one_async(ipsec_args_t *cargs,
 	int in_flight, pkts_allowed, i, num_pkts, rc = 0;
 	const int max_in_flight = cargs->in_flight;
 	const int burst_size = cargs->burst_size;
+	const int debug = cargs->debug_packets;
 	odp_ipsec_out_param_t param;
 	odp_pool_t pkt_pool;
 	odp_queue_t out_queue;
@@ -798,11 +799,7 @@ run_measure_one_async(ipsec_args_t *cargs,
 							   num_pkts)))
 				return -1;
 
-			if (cargs->debug_packets) {
-				for (i = 0; i < num_pkts; i++)
-					odp_packet_print_data(pkt[i], 0,
-							      odp_packet_len(pkt[i]));
-			}
+			debug_packets(debug, pkt, num_pkts);
 
 			rc = odp_ipsec_out_enq(pkt, num_pkts, &param);
 			if (odp_unlikely(rc <= 0)) {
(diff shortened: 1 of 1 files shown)


test/performance/odp_ipsec.c
```diff
@@ -665,13 +678,8 @@ run_measure_one(ipsec_args_t *cargs,
 				odp_packet_free(pkt);
 				break;
 			}
-			if (odp_packet_has_error(out_pkt)) {
-				odp_ipsec_packet_result_t result;
+			check_ipsec_result(out_pkt);
```
anoobj, line 683: I had noticed this discrepancy when I added the burst support. The way I understood, odp_ipsec_result() call in ASYNC path is mandatory to have post-processing done. For SYNC, it's not mandatory unless to verify that IPsec operation has been completed successfully.

The thing I wasn't sure about is, do we expect performance application also to call odp_ipsec_result() and check the results? If yes, why not make odp_ipsec_result() mandatory for both SYNC & ASYNC?

Usage of odp_packet_has_error() is wrong anyway. So I agree with the overall fix.

JannePeltonen: Since odp_ipsec_result() is practically always used, it makes sense for this test app to use it too even when not mandatory. Even in the async case the call is not mandatory overall but just needs to be done before packet metadata or packet data is accessed. I am not sure adding that requirement to the sync case would have much benefit, but it would add an additional restriction.

anoobj: > I am not sure adding that requirement to the sync case would have much benefit, but it would add an additional restriction.
I don't think it's a restriction if we agree that the API is used "practically always". Currently, with SYNC APIs, implementation would require executing post-processing from the odp_ipsec_in/out APIs and when odp_ipsec_result() is called, it would need additional logic to skip post-processing.

My suggestion is to avoid extra checks in the data path.

JannePeltonen: Ok, it could save one check there. But it would require an API change which would have to be handled separately. It is not really related to this PR.

anoobj: Agreed.

JannePeltonen:
@anoobj , can we finish this review? I think it would be best to merge this PR and then rebase the burst more PR on top.

anoobj:
@JannePeltonen The first patch looks okay to me. The second patch is rather questionable and I'm still checking if it's better introduced as part of burst processing. Needless to say, both patches would cause conflicts to burst patch. And considering these are relatively minor changes, my preference would be to converge on that first.

JannePeltonen:
> The second patch is rather questionable

Maybe you could elaborate on that, perhaps in the form of review comments?

anoobj:
> Maybe you could elaborate on that, perhaps in the form of review comments?

The code has checks in datapath which could be avoided. Current code and the proposed code is identical, but moving out some code could mean later we may have trouble if we decide to optimize threads based on control path settings (basically cargs->schedule and cargs->debug_packets). 

The code is also making assumptions that one event would result in one packet. So some of this would also need to be changed when we add vector packet handling (which is also in pipeline).

In the async routine, burst patch proposes batching up of packets from multiple events and free all of them with single API call. With this patch, packets from one dequeue_burst() call would be freed together. So for implementations returning single event at a time, current code would effectively just use single packet free in the loop.

Overall, I think the second patch would cause conflicts to the burst patch. Do you mind waiting for couple of days to see if we can converge on burst patch?

JannePeltonen:
I think the issues you listed are not really related to this patch itself but to envisioned future changes. There cannot be multiple packets in one event in the current code. Future patches that add vector support change that and they have to include the necessary changes in event handling.

But anyway, I made this PR because I thought it would be helpful in resolving the comments I made in the burst mode PR. We can abandon this PR if it does not fit in your planned sequence of changes. Maybe you could just pick the first PR from here and put it as the first patch in your burst mode PR with your reviewed-by tag added and then take care of schedule_multi through a new patch or include the changes in another patch it if makes more sense that way.

anoobj:
@JannePeltonen Works for me. I'll add these changes to PR 1481. Can you review v2 of 1481?

JannePeltonen:
v5:
- rebased, so the first patch, which was already merged to master, went away.
- The first patch in v5 changes repeated schedule/dequeue calls to schedule_multi/dequeue_multi calls, without changing other logic
- The second patch keeps more packets in flight by scheudling/dequeuing only when needed
- The third patch is just a very small cleanup

@anoobj , maybe you could take a look.

test/performance/odp_ipsec.c
```diff
@@ -818,32 +843,26 @@ run_measure_one_async(ipsec_args_t *cargs,
...
 
-		if (i)
-			odp_packet_free_sp(pkt_out, i);
+				if (cargs->debug_packets)
+					odp_packet_print_data(out_pkt, 0,
+							      odp_packet_len(out_pkt));
+				out_pkts[n] = out_pkt;
+			}
+			packets_received += num;
+			odp_packet_free_sp(out_pkts, num);
```
anoobj, line 864: Current code aggregates packets from multiple "dequeue" calls and frees in one shot. Proposed code frees packets from single "dequeue_burst" call.

Also, I didn't understand why length of pkt_out or out_pkt array was changed from max_in_flight to MAX_DEQUEUE_BURST.

test/performance/odp_ipsec.c
```diff
@@ -596,6 +596,14 @@ static uint8_t test_data[] = {
 	0x08, 0x00, 0xfb, 0x37, 0x12, 0x34, 0x00, 0x00
 };
 
+static inline void debug_packets(int debug, odp_packet_t *pkt, int num_pkts)
+{
+	if (odp_likely(!debug))
+		return;
```
anoobj, line 602: Is there an extra blank line here?

JannePeltonen: I do not see any blank lines there. Or did you mean that one should be added?

test/performance/odp_ipsec.c
anoobj: > If there are multiple packets available in the queue, presumably they can be all got through single dequeue_multi/schedule_multi call.

Actually this is the issue. On MRVL platforms, schedule_multi() calls would only return one packet. And freeing single packet is inefficient compared to freeing in a burst. 

If we make the array large enough to accommodate max_in_flight, this issue can be easily worked around. Since this is IPsec perf application, I guess it's better to minimize costs from non-IPsec paths.

JannePeltonen:
v3: reshuffled the patched a bit and made packet freeing happen for multiple packets at a time even if dequeue_multi/schedule_multi return packets one-by-one.

test/performance/odp_ipsec.c
JannePeltonen: > Actually this is the issue. On MRVL platforms, schedule_multi() calls would only return one packet. And freeing single packet is inefficient compared to freeing in a burst.

Ok, changed the code to be more like it was, collecting to-be-freed packets from multiple dequeue_multi operations.

anoobj:
Thanks for taking in the suggestion.

PR #1507, opened 2022-02-16
[PATCH v2] test: cunit: reduce bogus warnings after fatal assertion failures

JannePeltonen:
Replace various fatal assertion macros with our own versions that
indicate to the compiler and static analyzers when they do not return.


MatiasElo (approved):
Nokia's copyright could be added to the file.

JannePeltonen:
v2: rebased, added copyright and reviewed-by tag

PR #1509, opened 2022-02-17
[PATCH v2] example: ipsec_crypto: add capability checking and bit mode support

JannePeltonen:
example: ipsec_crypto: fix crypto capability checking

    Check that the used algorithm as well as key length etc are supported
    before trying to create a crypto session. Give up if the used ODP
    implementation supports the algorithm in bit mode since the code expects
    byte mode and would pass incorrect auth_range to an implementation which
    works in bit mode.

    example: ipsec_crypto: support bit mode algorithm implementations too

    Do not give up if the underlying ODP supports an algorithm in bit mode
    but scale the range parameter as appropriate.

### DIFF
--- a/example/ipsec_api/odp_ipsec.c
+++ b/example/ipsec_api/odp_ipsec.c
@@ -1396,3 +1396,10 @@ static void usage(char *progname)
 	       "\n", NO_PATH(progname), NO_PATH(progname)
 		);
 }
+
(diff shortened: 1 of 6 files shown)


JereLeppanen (approved):

JannePeltonen:
v2: rebased

PR #1510, opened 2022-02-18
[PATCH v14] validation: ipsec: add sa expiry test

asasidharan:
Add IPsec SA expiry tests in out suite.

### DIFF
--- a/platform/linux-generic/include/odp_ipsec_internal.h
+++ b/platform/linux-generic/include/odp_ipsec_internal.h
@@ -272,6 +272,14 @@ struct ipsec_sa_s {
 			} out;
 		};
 	} sa_info;
+
+	/*
+	 * Flag to check if the SA soft expiry status event was already
+	 * sent. This field is applicable only for the soft expiry status
+	 * event that gets generated for IPsec SAs configured in inline
+	 * outbound mode.
+	 */
+	odp_atomic_u32_t soft_expiry_notified;
 };
 
 /**
--- a/platform/linux-generic/odp_ipsec.c
+++ b/platform/linux-generic/odp_ipsec.c
@@ -2477,6 +2477,25 @@ static void ipsec_out_inline_finalize(odp_packet_t pkt_in[],
 		ipsec_inline_op_t *op = &ops[i];
 		odp_packet_t *pkt = &pkt_in[i];
 
+		if (op->op.status.warn.soft_exp_packets || op->op.status.warn.soft_exp_bytes) {
+			if (!odp_atomic_load_u32(&op->op.sa->soft_expiry_notified)) {
+				int rc;
+
+				/*
+				 * Another thread may have sent the notification by now but we do
+				 * not care since sending duplicate expiry notifications is allowed.
+				 */
+				rc = _odp_ipsec_status_send(op->op.sa->queue,
+							    ODP_IPSEC_STATUS_WARN,
+							    op->op.sa->ipsec_sa_hdl,
+							    0, op->op.status.warn);
+				if (rc == 0)
+					odp_atomic_store_u32(&op->op.sa->soft_expiry_notified, 1);
+				else
+					_ODP_DBG("IPsec status event submission failed\n");
+			}
+		}
+
 		if (odp_unlikely(op->op.status.error.all))
 			goto handle_err;
 
--- a/platform/linux-generic/odp_ipsec_sad.c
+++ b/platform/linux-generic/odp_ipsec_sad.c
@@ -562,6 +562,7 @@ odp_ipsec_sa_t odp_ipsec_sa_create(const odp_ipsec_sa_param_t *param)
 	odp_atomic_init_u64(&ipsec_sa->stats.hard_exp_pkts_err, 0);
 	odp_atomic_init_u64(&ipsec_sa->stats.post_lifetime_err_pkts, 0);
 	odp_atomic_init_u64(&ipsec_sa->stats.post_lifetime_err_bytes, 0);
+	odp_atomic_init_u32(&ipsec_sa->soft_expiry_notified, 0);
 
 	if (ODP_IPSEC_MODE_TUNNEL == ipsec_sa->mode &&
 	    ODP_IPSEC_DIR_OUTBOUND == param->dir) {
--- a/test/validation/api/ipsec/ipsec.c
+++ b/test/validation/api/ipsec/ipsec.c
@@ -27,9 +27,12 @@ static struct buffered_event_s sched_ev_buffer[EVENT_BUFFER_SIZE];
 struct suite_context_s suite_context;
 static odp_ipsec_capability_t capa;
 static int sched_ev_buffer_tail;
+odp_bool_t sa_expiry_notified;
+
 
 #define PKT_POOL_NUM  64
 #define EVENT_WAIT_TIME ODP_TIME_SEC_IN_NS
+#define STATUS_EVENT_WAIT_TIME ODP_TIME_MSEC_IN_NS
 #define SCHED_EVENT_RETRY_COUNT 2
 
 #define PACKET_USER_PTR	((void *)0x1212fefe)
@@ -437,6 +440,50 @@ void ipsec_sa_param_fill(odp_ipsec_sa_param_t *param,
 	param->lifetime.hard_limit.packets = 10000 * 1000;
 }
 
+static void ipsec_status_event_handle(odp_event_t ev_status,
+				      odp_ipsec_sa_t sa,
+				      enum ipsec_test_sa_expiry sa_expiry)
+{
+	odp_ipsec_status_t status = {
+		.id = 0,
+		.sa = ODP_IPSEC_SA_INVALID,
+		.result = 0,
+		.warn.all = 0,
+	};
+
+	CU_ASSERT_FATAL(ODP_EVENT_INVALID != ev_status);
+	CU_ASSERT_EQUAL(1, odp_event_is_valid(ev_status));
+	CU_ASSERT_EQUAL_FATAL(ODP_EVENT_IPSEC_STATUS, odp_event_type(ev_status));
+
+	CU_ASSERT_EQUAL(0, odp_ipsec_status(&status, ev_status));
+	CU_ASSERT_EQUAL(ODP_IPSEC_STATUS_WARN, status.id);
+	CU_ASSERT_EQUAL(sa, status.sa);
+	CU_ASSERT_EQUAL(0, status.result);
+
+	if (IPSEC_TEST_EXPIRY_IGNORED != sa_expiry) {
+		if (IPSEC_TEST_EXPIRY_SOFT_PKT == sa_expiry) {
+			CU_ASSERT_EQUAL(1, status.warn.soft_exp_packets);
+			sa_expiry_notified = true;
+		} else if (IPSEC_TEST_EXPIRY_SOFT_BYTE == sa_expiry) {
+			CU_ASSERT_EQUAL(1, status.warn.soft_exp_bytes);
+			sa_expiry_notified = true;
+		}
+	}
+
+	odp_event_free(ev_status);
+}
+
+void ipsec_status_event_get(odp_ipsec_sa_t sa,
+			    enum ipsec_test_sa_expiry sa_expiry)
+{
+	uint64_t wait_time = (sa_expiry == IPSEC_TEST_EXPIRY_IGNORED) ? 0 : STATUS_EVENT_WAIT_TIME;
+	odp_event_t ev;
+
+	ev = recv_event(suite_context.queue, wait_time);
+	if (ODP_EVENT_INVALID != ev)
+		ipsec_status_event_handle(ev, sa, sa_expiry);
+}
+
 void ipsec_sa_destroy(odp_ipsec_sa_t sa)
 {
 	odp_event_t event;
@@ -765,6 +812,45 @@ static int ipsec_process_in(const ipsec_test_part *part,
 	return num_out;
 }
 
+static int ipsec_check_sa_expiry(enum ipsec_test_sa_expiry sa_expiry,
+				 odp_ipsec_packet_result_t *result)
+{
+	if (sa_expiry == IPSEC_TEST_EXPIRY_IGNORED)
+		return 0;
+
+	if (!sa_expiry_notified) {
+		if (sa_expiry == IPSEC_TEST_EXPIRY_SOFT_PKT) {
+			if (result->status.warn.soft_exp_packets)
+				sa_expiry_notified = true;
+		} else if (sa_expiry == IPSEC_TEST_EXPIRY_SOFT_BYTE) {
+			if (result->status.warn.soft_exp_bytes)
+				sa_expiry_notified = true;
+		} else if (sa_expiry == IPSEC_TEST_EXPIRY_HARD_PKT) {
+			if (result->status.error.hard_exp_packets)
+				sa_expiry_notified = true;
+
+			return -1;
+		} else if (sa_expiry == IPSEC_TEST_EXPIRY_HARD_BYTE) {
+			if (result->status.error.hard_exp_bytes)
+				sa_expiry_notified = true;
+
+			return -1;
+		}
+	} else {
+		if (sa_expiry == IPSEC_TEST_EXPIRY_HARD_PKT) {
+			CU_ASSERT(result->status.error.hard_exp_packets);
+
+			return -1;
+		} else if (sa_expiry == IPSEC_TEST_EXPIRY_HARD_BYTE) {
+			CU_ASSERT(result->status.error.hard_exp_bytes);
+
+			return -1;
+		}
+	}
+
+	return 0;
+}
+
 static int ipsec_send_out_one(const ipsec_test_part *part,
 			      odp_ipsec_sa_t sa,
 			      odp_packet_t *pkto)
@@ -891,18 +977,39 @@ static int ipsec_send_out_one(const ipsec_test_part *part,
 
 				pkto[i] = odp_packet_from_event(ev);
 				CU_ASSERT_FATAL(pkto[i] != ODP_PACKET_INVALID);
+
+				if (part->out[i].sa_expiry != IPSEC_TEST_EXPIRY_NONE)
+					ipsec_status_event_get(sa, part->out[i].sa_expiry);
+
 				i++;
 				continue;
 			}
 
 			ev = recv_event(suite_context.queue, 0);
 			if (ODP_EVENT_INVALID != ev) {
+				odp_event_type_t ev_type;
+
 				CU_ASSERT(odp_event_is_valid(ev) == 1);
+				ev_type = odp_event_types(ev, &subtype);
+
+				if ((ODP_EVENT_IPSEC_STATUS == ev_type) &&
+				    part->out[i].sa_expiry != IPSEC_TEST_EXPIRY_NONE) {
+					ipsec_status_event_handle(ev, sa, part->out[i].sa_expiry);
+					continue;
+				}
+
 				CU_ASSERT_EQUAL(ODP_EVENT_PACKET,
-						odp_event_types(ev, &subtype));
+						ev_type);
 				CU_ASSERT_EQUAL(ODP_EVENT_PACKET_IPSEC,
 						subtype);
-				CU_ASSERT(part->out[i].status.error.all);
+
+				/* In the case of SA hard expiry tests, hard expiry error bits are
+				 * expected to be set. The exact error bits expected to be set based
+				 * on sa_expiry is checked eventually in ipsec_check_sa_expiry()
+				 * from the caller of this function.
+				 */
+				if (part->out[i].sa_expiry == IPSEC_TEST_EXPIRY_NONE)
+					CU_ASSERT(part->out[i].status.error.all);
 
 				pkto[i] = odp_ipsec_packet_from_event(ev);
 				CU_ASSERT_FATAL(pkto[i] != ODP_PACKET_INVALID);
@@ -991,6 +1098,8 @@ static void verify_in(const ipsec_test_part *part,
 					ODP_IPSEC_OP_MODE_INLINE,
 					result.flag.inline_mode);
 			CU_ASSERT_EQUAL(sa, result.sa);
+			CU_ASSERT_EQUAL(part->out[i].status.warn.all,
+					result.status.warn.all);
 			if (ODP_IPSEC_SA_INVALID != sa)
 				CU_ASSERT_EQUAL(IPSEC_SA_CTX,
 						odp_ipsec_sa_context(sa));
@@ -1066,6 +1175,11 @@ int ipsec_check_out(const ipsec_test_part *part, odp_ipsec_sa_t sa,
 		} else {
 			/* IPsec packet */
 			CU_ASSERT_EQUAL(0, odp_ipsec_result(&result, pkto[i]));
+
+			if (part->out[i].sa_expiry != IPSEC_TEST_EXPIRY_NONE)
+				if (ipsec_check_sa_expiry(part->out[i].sa_expiry, &result) != 0)
+					return num_out;
+
 			CU_ASSERT_EQUAL(part->out[i].status.error.all,
 					result.status.error.all);
 			if (0 == result.status.error.all)
--- a/test/validation/api/ipsec/ipsec.c
+++ b/test/validation/api/ipsec/ipsec.c
@@ -16,11 +16,21 @@
 #include "test_vectors.h"
 #include "reass_test_vectors.h"
 
+#define EVENT_BUFFER_SIZE 3
+
+struct buffered_event_s {
+	odp_queue_t from;
+	odp_event_t event;
+};
+
+static struct buffered_event_s sched_ev_buffer[EVENT_BUFFER_SIZE];
 struct suite_context_s suite_context;
 static odp_ipsec_capability_t capa;
+static int sched_ev_buffer_tail;
 
 #define PKT_POOL_NUM  64
 #define EVENT_WAIT_TIME ODP_TIME_SEC_IN_NS
+#define SCHED_EVENT_RETRY_COUNT 2
 
 #define PACKET_USER_PTR	((void *)0x1212fefe)
 #define IPSEC_SA_CTX	((void *)0xfefefafa)
@@ -101,11 +111,68 @@ static int pktio_start(odp_pktio_t pktio, odp_bool_t in, odp_bool_t out)
 	return 1;
 }
 
-static odp_event_t sched_queue_deq(uint64_t wait_ns)
+static int sched_event_buffer_add(odp_queue_t from, odp_event_t event)
+{
+	if (sched_ev_buffer_tail + 1 == EVENT_BUFFER_SIZE)
+		return -ENOMEM;
+
+	sched_ev_buffer[sched_ev_buffer_tail].from = from;
+	sched_ev_buffer[sched_ev_buffer_tail].event = event;
+	sched_ev_buffer_tail++;
+
+	return 0;
+}
+
+static odp_event_t sched_event_buffer_get(odp_queue_t from)
+{
+	odp_event_t ev;
+	int i, j;
+
+	if (odp_queue_type(from) == ODP_QUEUE_TYPE_PLAIN)
+		return ODP_EVENT_INVALID;
+
+	/* Look for a matching entry */
+	for (i = 0; i < sched_ev_buffer_tail; i++)
+		if (sched_ev_buffer[i].from == from)
+			break;
+
+	/* Remove entry from buffer */
+	if (i != sched_ev_buffer_tail) {
+		ev = sched_ev_buffer[i].event;
+
+		for (j = 1; i + j < sched_ev_buffer_tail; j++)
+			sched_ev_buffer[i + j - 1] = sched_ev_buffer[i + j];
+
+		sched_ev_buffer_tail--;
+	} else {
+		ev = ODP_EVENT_INVALID;
+	}
+
+	return ev;
+}
+
+static odp_event_t sched_queue_deq(odp_queue_t queue, uint64_t wait_ns)
 {
 	uint64_t wait = odp_schedule_wait_time(wait_ns);
+	odp_event_t ev = ODP_EVENT_INVALID;
+	odp_queue_t from;
+	int retry = 0;
 
-	return odp_schedule(NULL, wait);
+	/* Check if buffered events are available */
+	ev = sched_event_buffer_get(queue);
+	if (ODP_EVENT_INVALID != ev)
+		return ev;
+
+	do {
+		ev = odp_schedule(&from, wait);
+
+		if ((ev != ODP_EVENT_INVALID) && (from != queue)) {
+			CU_ASSERT_FATAL(0 == sched_event_buffer_add(from, ev));
+			ev = ODP_EVENT_INVALID;
+		}
+	} while (ev == ODP_EVENT_INVALID && (++retry < SCHED_EVENT_RETRY_COUNT));
+
+	return ev;
 }
 
 static odp_event_t plain_queue_deq(odp_queue_t queue, uint64_t wait_ns)
@@ -131,7 +198,7 @@ static odp_event_t recv_event(odp_queue_t queue, uint64_t wait_ns)
 	if (odp_queue_type(queue) == ODP_QUEUE_TYPE_PLAIN)
 		event = plain_queue_deq(queue, wait_ns);
 	else
-		event = sched_queue_deq(wait_ns);
+		event = sched_queue_deq(queue, wait_ns);
 
 	return event;
 }
--- a/test/validation/api/ipsec/ipsec.h
+++ b/test/validation/api/ipsec/ipsec.h
@@ -72,6 +72,15 @@ enum ipsec_test_stats {
 	IPSEC_TEST_STATS_AUTH_ERR,
 };
 
+enum ipsec_test_sa_expiry {
+	IPSEC_TEST_EXPIRY_NONE = 0,
+	IPSEC_TEST_EXPIRY_IGNORED,
+	IPSEC_TEST_EXPIRY_SOFT_BYTE,
+	IPSEC_TEST_EXPIRY_SOFT_PKT,
+	IPSEC_TEST_EXPIRY_HARD_BYTE,
+	IPSEC_TEST_EXPIRY_HARD_PKT,
+};
+
 typedef struct {
 	odp_bool_t lookup;
 	odp_bool_t inline_hdr_in_packet;
@@ -101,8 +110,11 @@ typedef struct {
 		 * differs from that of input test packet (pkt_in).
 		 */
 		uint32_t orig_ip_len;
+		enum ipsec_test_sa_expiry sa_expiry;
 	} out[MAX_FRAGS];
 } ipsec_test_part;
+
+extern odp_bool_t sa_expiry_notified;
 
 void ipsec_sa_param_fill(odp_ipsec_sa_param_t *param,
 			 odp_ipsec_dir_t dir,
@@ -151,5 +163,7 @@ int ipsec_check_test_sa_update_seq_num(void);
 int ipsec_check_esp_aes_gcm_128_reass_ipv4(void);
 int ipsec_check_esp_aes_gcm_128_reass_ipv6(void);
 int ipsec_check_esp_null_aes_xcbc(void);
+void ipsec_status_event_get(odp_ipsec_sa_t sa,
+			    enum ipsec_test_sa_expiry sa_expiry);
 
 #endif
--- a/test/validation/api/ipsec/ipsec_test_out.c
+++ b/test/validation/api/ipsec/ipsec_test_out.c
@@ -1593,6 +1593,138 @@ static void test_test_sa_update_seq_num(void)
 	printf("\n  ");
 }
 
+#define SOFT_LIMIT_PKT_CNT 1024
+#define HARD_LIMIT_PKT_CNT 2048
+#define DELTA_PKT_CNT 320
+
+static void test_out_ipv4_esp_sa_expiry(enum ipsec_test_sa_expiry expiry)
+{
+	int byte_count_per_packet = pkt_ipv4_icmp_0.len - pkt_ipv4_icmp_0.l3_offset;
+	uint32_t src = IPV4ADDR(10, 0, 11, 2);
+	uint32_t dst = IPV4ADDR(10, 0, 22, 2);
+	odp_ipsec_tunnel_param_t out_tunnel;
+	odp_ipsec_sa_param_t param_out;
+	int i, inc, limit, delta;
+	uint64_t soft_limit_byte;
+	uint64_t hard_limit_byte;
+	uint64_t soft_limit_pkt;
+	uint64_t hard_limit_pkt;
+	odp_ipsec_sa_t out_sa;
+
+	switch (expiry)	{
+	case IPSEC_TEST_EXPIRY_SOFT_PKT:
+		soft_limit_pkt = SOFT_LIMIT_PKT_CNT;
+		hard_limit_pkt = HARD_LIMIT_PKT_CNT;
+		soft_limit_byte = 0;
+		hard_limit_byte = 0;
+		delta = DELTA_PKT_CNT;
+		limit = soft_limit_pkt;
+		inc = 1;
+		break;
+	case IPSEC_TEST_EXPIRY_HARD_PKT:
+		soft_limit_pkt = SOFT_LIMIT_PKT_CNT;
+		hard_limit_pkt = HARD_LIMIT_PKT_CNT;
+		soft_limit_byte = 0;
+		hard_limit_byte = 0;
+		delta = DELTA_PKT_CNT;
+		limit = hard_limit_pkt;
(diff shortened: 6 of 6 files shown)


asasidharan:
> FAIL: validation/api/pktio/pktio_run_netmap.sh

@MatiasElo, there seems to be a failure in one of the CI tests that looks unrelated to this PR.  Could you please help me re-trigger the CI?

asasidharan:
> 
> Hi @asasidharan, the netmap test randomly fails to create interfaces, which causes the faulure. I restarted the CI run.

Ack.

platform/linux-generic/odp_ipsec_sad.c
```diff
@@ -90,6 +90,13 @@ typedef struct sa_thread_local_s {
 	 * counter(s).
 	 */
 	odp_ipsec_op_status_t lifetime_status;
+
+	/*
+	 * Flag to check if the SA expiry status event was already
+	 * sent. This is applicable only for IPsec SA in inline outbound
+	 * mode.
+	 */
+	odp_bool_t status_notify;
```
JannePeltonen, line 99: Why is this thread local? There is no need to send a separate status event for each thread that happens to notice soft limit crossing and making this thread local will consume more memory than having it as an atomic in the sa structure.

platform/linux-generic/odp_ipsec_sad.c
```diff
@@ -969,6 +977,17 @@ int _odp_ipsec_sa_lifetime_update(ipsec_sa_t *ipsec_sa, uint32_t len,
...
 	status->all |= sa_tl->lifetime_status.all;
 
+	if (is_out_inline && (sa_tl->lifetime_status.warn.soft_exp_packets ||
+			      sa_tl->lifetime_status.warn.soft_exp_bytes)) {
+		if (!sa_tl->status_notify) {
+			_odp_ipsec_status_send(ipsec_sa->queue,
+					       ODP_IPSEC_STATUS_WARN,
+					       ipsec_sa->ipsec_sa_hdl,
+					       0, sa_tl->lifetime_status.warn);
+			sa_tl->status_notify = 1;
```
JannePeltonen, line 987: This flag is set even if the status message could not be sent due to event allocation failure or full event queue. The return value of _odp_ipsec_status_send() should be checked.

I think it would be better to put this added section in a separate function that is called by odp_ipsec_out_inline(). Then the non-inline path and the RX path does not need to be cluttered with the "is_out_inline" checks and code which is specific to the out-inline case.

JannePeltonen (changes requested):
A couple of comments on the linux-gen changes. I did not look at the validation test patch yet.

platform/linux-generic/odp_ipsec_sad.c
asasidharan: Ack. Made status_notify as an atomic u32 in the SA structure with v5.

platform/linux-generic/odp_ipsec_sad.c
asasidharan: Agreed. With v5, status event is now generated directly from the odp_ipsec_out_inline().

asasidharan:
@MatiasElo Seems like CI has failed for 3 checks, but I am not able to see any logs for the failing checks. I can see just a message saying "This check failed". Could you please help me understand why these checks failed?

image

MatiasElo:
The checks failed due to GitHub Actions internal errors. You can see the error messages in the Summary view.
Screen Shot 2022-03-30 at 10 02 30

Usually these errors are GitHub internal and unrelated to the tests. I restarted the failed three tests.

asasidharan:
> Usually these errors are GitHub internal and unrelated to the tests. I restarted the failed three tests.

@MatiasElo Got it. Thank you for restarting the tests.

asasidharan:
@JannePeltonen I have made changes to the linux-generic implementation as suggested. Could you review the PR again?

asasidharan:
@JannePeltonen Ping.

test/validation/api/ipsec/ipsec.c
```diff
@@ -815,10 +858,26 @@ static int ipsec_send_out_one(const ipsec_test_part *part,
...
 				CU_ASSERT_EQUAL(ODP_EVENT_PACKET,
 						odp_event_types(ev, &subtype));
 				CU_ASSERT_EQUAL(ODP_EVENT_PACKET_BASIC,
 						subtype);
+
+				if (part->flags.test_sa_expiry) {
+					if (part->out[i].sa_expiry == IPSEC_TEST_EXPIRY_HARD_PKT ||
+					    part->out[i].sa_expiry == IPSEC_TEST_EXPIRY_HARD_BYTE) {
+						suite_context.status_notify = true;
+						return num_out;
```
JannePeltonen, line 878: I do not understand this. Whenever hard lifetime expiry is being tested, any succesfully outbound inline IPsec processing causes the "status notify" flag to be set. So the flag would get set even if the expiration was never reported.

The return here means that some of the later asserts (about sa, sa, ctx, user_ptr) are skipped for no good reason.

The return also leaves the pkto output parameter unset, crashing the caller in some cases (e.g. if the ODP under test is not reporting hard lifetime expiry).

JannePeltonen (changes requested):

test/validation/api/ipsec/ipsec.c
```diff
@@ -370,6 +370,49 @@ void ipsec_sa_param_fill(odp_ipsec_sa_param_t *param,
...
+				      enum ipsec_test_sa_expiry sa_expiry)
+{
+	odp_ipsec_status_t status = {
+		.id = 0,
+		.sa = ODP_IPSEC_SA_INVALID,
+		.result = 0,
+		.warn.all = 0,
+	};
+
+	CU_ASSERT(ODP_EVENT_INVALID != ev_status);
```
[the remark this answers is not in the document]
asasidharan: Ack. Updated with v7.

test/validation/api/ipsec/ipsec.c
```diff
@@ -815,10 +858,26 @@ static int ipsec_send_out_one(const ipsec_test_part *part,
 			ev = recv_event(queue, 0);
 			if (ODP_EVENT_INVALID != ev) {
 				CU_ASSERT(odp_event_is_valid(ev) == 1);
+
+				if (ODP_EVENT_IPSEC_STATUS == odp_event_type(ev) &&
```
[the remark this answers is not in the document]
asasidharan: This is needed in the case where the pktio queue is running in scheduled mode and the IPsec destination queue is as well configured for scheduled mode. In this case, the recv_event() call can return an event from either the pktio queue or the IPsec SA destination queue. Hence the handling had to be placed here.

test/validation/api/ipsec/ipsec.c
```diff
@@ -844,6 +913,15 @@ static int ipsec_send_out_one(const ipsec_test_part *part,
...
 			}
 		}
+
+		if ((ODP_QUEUE_TYPE_PLAIN == suite_context.q_type) &&
+		    part->flags.test_sa_expiry)
+			for (i = 0; i < num_out; i++) {
+				if (suite_context.status_notify)
+					break;
+
+				ipsec_status_event_get(sa, part->out[i].sa_expiry, false);
```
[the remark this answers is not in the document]
asasidharan: Ack. This was added initially when I saw that with linux-gen, the test was receiving the status event after some delay. But with the updated changes, I no longer see such an issue. Removing this part of the change with v7.

asasidharan:
v8: fixed compilation issue introduced with new global variable sa_expiry_notified.

asasidharan:
@JannePeltonen ping!

test/validation/api/ipsec/ipsec.c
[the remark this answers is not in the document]
JannePeltonen: This is getting a bit messy. We are seemingly handling an event received from loop pktio, so one would think the event should never be an IPsec status event here. Except that in scheduled mode recv_event() receives from any queue, which is not that obvious.

As a result we have two almost identical branches here with weird checks for scheduled mode and whatnot. I think some refactoring of this code and recv_event() is needed to make the code more readable and maintainable.

One possibility that comes to mind (not sure if it is ok, did not really think it through) could be to have recv_event() always (unless it timeouts) return an event from the queue passed to it, even in scheduled mode. In scheduled mode recv_event() could internally cache received events that came through an unexpected queue. I think it would be sufficient to cache just one event per queue. With this the calling code would be simple and could handle one queue at a time regardless of mode (those "if scheduled mode" checks etc would not be needed).

platform/linux-generic/include/odp_ipsec_internal.h
```diff
@@ -272,6 +272,13 @@ struct ipsec_sa_s {
 			} out;
 		};
 	} sa_info;
+
+	/*
+	 * Flag to check if the SA expiry status event was already
+	 * sent. This is applicable only for IPsec SA in inline outbound
+	 * mode.
+	 */
+	odp_atomic_u32_t status_notify;
```
JannePeltonen, line 281: This should be renamed to be more descriptive, e.g. expiry_notified or expiry_msg_sent.

It would also be good to move this before stats in the SA structure as this field may get referenced for every packet unlike sa_info and stats.

platform/linux-generic/odp_ipsec.c
```diff
@@ -2250,6 +2250,21 @@ int odp_ipsec_out_inline(const odp_packet_t pkt_in[], int num_in,
 		ipsec_sa = ipsec_out_single(pkt, sa, &pkt, opt, true, &status);
 		ODP_ASSERT(NULL != ipsec_sa);
 
+		if (status.warn.soft_exp_packets || status.warn.soft_exp_bytes) {
+			if (!odp_atomic_load_u32(&ipsec_sa->status_notify)) {
```
JannePeltonen, line 2254: It would be good to add a comment that the race condition here is intentional, e.g.:

```
	/*
	 * Another thread may have sent the notification by now but we do
	 * not care since sending duplicate expiry notifications is allowed.
	 */
	rc = _odp_ipsec_status_send(ipsec_sa->queue,
```

test/validation/api/ipsec/ipsec.c
```diff
@@ -998,6 +1070,26 @@ int ipsec_check_out(const ipsec_test_part *part, odp_ipsec_sa_t sa,
...
 			CU_ASSERT_EQUAL(0, odp_ipsec_result(&result, pkto[i]));
+			if (part->flags.test_sa_expiry && !sa_expiry_notified) {
+				if (part->out[i].sa_expiry == IPSEC_TEST_EXPIRY_SOFT_PKT) {
+					if (result.status.warn.soft_exp_packets)
+						sa_expiry_notified = true;
+				} else if (part->out[i].sa_expiry == IPSEC_TEST_EXPIRY_SOFT_BYTE) {
+					if (result.status.warn.soft_exp_bytes)
+						sa_expiry_notified = true;
+				} else if (part->out[i].sa_expiry == IPSEC_TEST_EXPIRY_HARD_PKT) {
+					if (result.status.error.hard_exp_packets) {
```
JannePeltonen, line 1081: It would be good to check that after certain point the hard expiry error is in every packet. Now the test would pass even if after hard limit expiration (not after the first error packet, since that may become a bit earlier) some later packets would still be processed without errors, which would be a serious problem.

test/validation/api/ipsec/ipsec.h
```diff
@@ -101,9 +110,12 @@ typedef struct {
 		 * differs from that of input test packet (pkt_in).
 		 */
 		uint32_t orig_ip_len;
+		enum ipsec_test_sa_expiry sa_expiry;
```
JannePeltonen, line 113: This is part of the "expected output" in ipsec_test_part, but it is not handled exactly like that. Maybe have "IGNORED" value in addition to the "NONE" value and then in the test code always check the expiry status unless the expected status is "IGNORE"?

test/validation/api/ipsec/ipsec_test_out.c
```diff
@@ -1591,6 +1591,150 @@ static void test_test_sa_update_seq_num(void)
 	printf("\n  ");
 }
 
+#define SOFT_LIMIT_PKT_CNT 1024
+#define HARD_LIMIT_PKT_CNT 2048
+#define DELTA_PKT_CNT 320
+
+static void test_out_ipv4_esp_sa_expiry(enum ipsec_test_sa_expiry expiry)
+{
+	int test_packet_len = pkt_ipv4_icmp_0.len - pkt_ipv4_icmp_0.l3_offset;
```
JannePeltonen, line 1600: Maybe rename this to e.g. byte_count_per_packet as it is not really test packet length?

test/validation/api/ipsec/ipsec_test_out.c
```diff
@@ -1591,6 +1591,150 @@ static void test_test_sa_update_seq_num(void)
...
+		return;
+	}
+
+	memset(&out_tunnel, 0, sizeof(odp_ipsec_tunnel_param_t));
+
+	out_tunnel.type = ODP_IPSEC_TUNNEL_IPV4;
+	out_tunnel.ipv4.src_addr = &src;
+	out_tunnel.ipv4.dst_addr = &dst;
+
+	CU_ASSERT(odp_ipsec_capability(&capa) == 0);
```
JannePeltonen, line 1660: This is unnecessary.
c.f. 044c92df7 validation: ipsec: remove unneeded capability calls

test/validation/api/ipsec/ipsec_test_out.c
```diff
@@ -1591,6 +1591,150 @@ static void test_test_sa_update_seq_num(void)
...
+			  .l4_type = ODP_PROTO_L4_TYPE_ESP,
+			},
+		},
+	};
+
+	if ((suite_context.outbound_op_mode == ODP_IPSEC_OP_MODE_INLINE) &&
+	    (expiry == IPSEC_TEST_EXPIRY_HARD_PKT ||
+	     expiry == IPSEC_TEST_EXPIRY_HARD_BYTE)) {
+		test_out.flags.test_sa_expiry = true;
+		test_out.out[0].sa_expiry = IPSEC_TEST_EXPIRY_NONE;
```
JannePeltonen, line 1692: Why is this special handling for hard limits in inline mode needed here? To make sure status events are received if they come even before limit-delta packets? This seems a bit hacky and out-of-place here.

test/validation/api/ipsec/ipsec_test_out.c
```diff
@@ -1591,6 +1591,150 @@ static void test_test_sa_update_seq_num(void)
...
+	test_out.out[0].sa_expiry = expiry;
+
+	for (; i <= limit; i += inc) {
+		ipsec_check_out_one(&test_out, out_sa);
+		if (sa_expiry_notified)
+			break;
+	}
+
+	if (suite_context.outbound_op_mode == ODP_IPSEC_OP_MODE_INLINE)
+		ipsec_status_event_get(out_sa, expiry, true);
```
JannePeltonen, line 1709: Why is this needed? For handling status events that ipsec_check_out_one() failed to handle? Maybe ipsec_check_out_one() should then be improved instead?

test/validation/api/ipsec/ipsec_test_out.c
```diff
@@ -1898,6 +2042,10 @@ odp_testinfo_t ipsec_out_suite[] = {
 				  ipsec_check_esp_null_aes_xcbc),
 	ODP_TEST_INFO_CONDITIONAL(test_sa_info,
 				  ipsec_check_esp_aes_cbc_128_sha1),
+	ODP_TEST_INFO_CONDITIONAL(test_out_ipv4_esp_sa_pkt_expiry,
+				  ipsec_check_esp_aes_cbc_128_sha1),
+	ODP_TEST_INFO_CONDITIONAL(test_out_ipv4_esp_sa_byte_expiry,
```
JannePeltonen, line 2047: It would be good to have SA expiry tests for inbound SAs too, but of course it does not have to be part of this PR. Any plans for it?

test/validation/api/ipsec/ipsec.c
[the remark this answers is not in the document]
asasidharan: Ack. Added logic to cache at most three scheduled events along with the queue handle via which it was received. With the updated test logic, it runs fine on both linux-gen and cn10k platform.

platform/linux-generic/odp_ipsec.c
asasidharan: Ack. Updated with v9.

test/validation/api/ipsec/ipsec.c
asasidharan: Ack. Updated the test to continue sending test packets even after hitting first error packet  and check for the corresponding error bits in the result.

test/validation/api/ipsec/ipsec.h
asasidharan: Ack. Further looking at the test, realized that the NONE is not really needed. Replaced the NONE value with IGNORED and updated the test code as suggested.

test/validation/api/ipsec/ipsec_test_out.c
asasidharan: Ack. Renamed the variable it to byte_count_per_packet.

test/validation/api/ipsec/ipsec_test_out.c
asasidharan: Ack. Removed the capability call.

test/validation/api/ipsec/ipsec_test_out.c
asasidharan: When testing for hard expiry, before reaching the hard limit, the first loop of test packets would cross the soft expiry limit. With this special handling, the test expects the soft expiry warning event to be received anytime during the first loop itself.

In other words, so that the status event is not stuck for ever in case of hard expiry tests, this special handling was added.

test/validation/api/ipsec/ipsec_test_out.c
asasidharan: Ack. For soft expiry status event test, with 0 wait time for receiving IPsec status event, observation was that the status event was received a little after the entire test packets were run. By having a 1 msec wait time for each IPsec status event receive calls, the test seems to run fine. So removed this handling with v9.

test/validation/api/ipsec/ipsec_test_out.c
asasidharan: Will take inbound test as separate PR.

asasidharan:
@JannePeltonen I have updated the PR with suggested changes. Could you please review the updated PR?

anoobj:
@JannePeltonen @psavol This PR has been pending for quite some time. Can this be reviewed?

test/validation/api/ipsec/ipsec.c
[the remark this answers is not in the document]
JannePeltonen: The code changes that make recv_event() take packets only from the specified queue also in scheduled mode could be put in a separate patch that precedes the patch that adds the new test. That would make the changes easier to read.

test/validation/api/ipsec/ipsec.h
JannePeltonen: I think the logic and the meaning of the expiry field is still a little bit confusing. I think it would be better if IGNORE would attempt to read status message if they are available and then ignore them, rather than let them pile in the queue from which they get read later when the status changed from IGNORE. Then NONE or similar could still be used for checking that expiry did not happen. The test would then send a bunch of packets with expiry causing a valitation error, then another bunch of packets where expiry is allowed to happen but not required, and finally a third bunch of packets where expiry must happen.

test/validation/api/ipsec/ipsec.c
```diff
@@ -16,11 +16,25 @@
 #include "test_vectors.h"
 #include "reass_test_vectors.h"
 
+#define MAX_EVENT_CACHE 3
```
JannePeltonen, line 19: EVENT_CACHE_SIZE or MAX_CACHED_EVENTS would be easier to understand.

test/validation/api/ipsec/ipsec.c
```diff
@@ -16,11 +16,25 @@
...
+
+struct sched_event_cache_s {
```
JannePeltonen, line 21: The type name could be cached_event_s or something like that since the type represents one cached event instead of the whole cache and the subsequent array is an array of cached events, not an array of event caches.

test/validation/api/ipsec/ipsec.c
```diff
@@ -101,11 +115,85 @@ static int pktio_start(odp_pktio_t pktio, odp_bool_t in, odp_bool_t out)
...
+static odp_event_t sched_event_cache_get(odp_queue_t from)
+{
+	odp_event_t ev;
+	int i, j;
+
+	if (odp_queue_type(from) == ODP_QUEUE_TYPE_PLAIN)
+		return ODP_EVENT_INVALID;
+
+	odp_spinlock_lock(&sched_ev_cache_lock);
+	if (sched_ev_cache_tail == 0) {
```
JannePeltonen, line 146: This appears redundant.

test/validation/api/ipsec/ipsec.c
```diff
@@ -101,11 +115,85 @@ static int pktio_start(odp_pktio_t pktio, odp_bool_t in, odp_bool_t out)
...
+		if (sched_ev_cache[i].from == from)
+			break;
+
+	/* Remove entry from cache */
+	if (i != sched_ev_cache_tail) {
+		ev = sched_ev_cache[i].event;
+
+		for (j = 1; i + j < sched_ev_cache_tail; j++) {
+			sched_ev_cache[i + j - 1].from = sched_ev_cache[i + j].from;
+			sched_ev_cache[i + j - 1].event = sched_ev_cache[i + j].event;
```
JannePeltonen, line 163: Struct assignment would be better and more robust, i.e. sched_ev_cache[i + j - 1] = sched_ev_cache[i + j];

test/validation/api/ipsec/ipsec.c
```diff
@@ -370,6 +458,48 @@ void ipsec_sa_param_fill(odp_ipsec_sa_param_t *param,
...
+	odp_ipsec_status_t status = {
+		.id = 0,
+		.sa = ODP_IPSEC_SA_INVALID,
+		.result = 0,
+		.warn.all = 0,
+	};
+
+	CU_ASSERT_FATAL(ODP_EVENT_INVALID != ev_status);
+	CU_ASSERT_EQUAL(1, odp_event_is_valid(ev_status));
+	CU_ASSERT_EQUAL(ODP_EVENT_IPSEC_STATUS, odp_event_type(ev_status));
```
JannePeltonen, line 474: This should probably be a fatal assert since odp_ipsec_status() must not be called for wrong types of events.

test/validation/api/ipsec/ipsec.c
```diff
@@ -370,6 +458,48 @@ void ipsec_sa_param_fill(odp_ipsec_sa_param_t *param,
...
+		sa_expiry_notified = true;
+	}
+
+	odp_event_free(ev_status);
+}
+
+void ipsec_status_event_get(odp_ipsec_sa_t sa,
+			    enum ipsec_test_sa_expiry sa_expiry)
+{
+	uint64_t wait_time = sa_expiry_notified ? 0 : STATUS_EVENT_WAIT_TIME;
```
JannePeltonen, line 495: The zero wait time has the potential issue that if events do come, they may not get received here but be left in the scheduler and get received at a later point.

test/validation/api/ipsec/ipsec.c
```diff
@@ -815,26 +947,46 @@ static int ipsec_send_out_one(const ipsec_test_part *part,
...
+					CU_ASSERT_EQUAL(ODP_EVENT_PACKET_BASIC,
+							subtype);
+					CU_ASSERT(!part->out[i].status.error.all);
+				}
 
 				pkto[i] = odp_packet_from_event(ev);
 				CU_ASSERT_FATAL(pkto[i] != ODP_PACKET_INVALID);
+
+				if (part->out[i].sa_expiry != IPSEC_TEST_EXPIRY_IGNORED)
+					ipsec_status_event_get(sa, part->out[i].sa_expiry);
```
JannePeltonen, line 964: If status events come when they are ignored, they stay in the queue until read at some later point. This can be problematic. I think this is related to the question on what IGNORE really means.

test/validation/api/ipsec/ipsec.c
```diff
@@ -815,26 +947,46 @@ static int ipsec_send_out_one(const ipsec_test_part *part,
...
 			if (ODP_EVENT_INVALID != ev) {
 				CU_ASSERT(odp_event_is_valid(ev) == 1);
+
 				CU_ASSERT_EQUAL(ODP_EVENT_PACKET,
 						odp_event_types(ev, &subtype));
-				CU_ASSERT_EQUAL(ODP_EVENT_PACKET_BASIC,
-						subtype);
-				CU_ASSERT(!part->out[i].status.error.all);
+
+				if (ODP_EVENT_PACKET_IPSEC != subtype) {
```
JannePeltonen, line 954: This condition should always be true. At this point we are reading the pktio queue of the of the loopback interface to which we are sending inline-processed packets. So the only thing we are supposed to receive here are plain packets.

test/validation/api/ipsec/ipsec.c
```diff
@@ -815,26 +947,46 @@ static int ipsec_send_out_one(const ipsec_test_part *part,
...
+
 				CU_ASSERT_EQUAL(ODP_EVENT_PACKET,
-						odp_event_types(ev, &subtype));
+						ev_type);
 				CU_ASSERT_EQUAL(ODP_EVENT_PACKET_IPSEC,
 						subtype);
-				CU_ASSERT(part->out[i].status.error.all);
+				if (part->out[i].sa_expiry != IPSEC_TEST_EXPIRY_HARD_PKT &&
+				    part->out[i].sa_expiry != IPSEC_TEST_EXPIRY_HARD_BYTE)
+					CU_ASSERT(part->out[i].status.error.all);
```
JannePeltonen, line 989: This can be a bit hard to understand. It is related to the somewhat unclear maning of the sa_expiry field. Maybe it would be clearer if the expected status is checked to contain an error always except when sa_expiry is IGNORE, in which case both hard expiry and success are ok (with a bit tweaked meaning of IGNORE and with the expiry test setting the expected error flag when it in fact expects an error.

test/validation/api/ipsec/ipsec.c
```diff
@@ -998,6 +1153,39 @@ int ipsec_check_out(const ipsec_test_part *part, odp_ipsec_sa_t sa,
 		} else {
 			/* IPsec packet */
 			CU_ASSERT_EQUAL(0, odp_ipsec_result(&result, pkto[i]));
+
+			if (part->out[i].sa_expiry != IPSEC_TEST_EXPIRY_IGNORED &&
+			    !sa_expiry_notified) {
```
JannePeltonen, line 1158: I think in other cases than inline-out-with-soft-expiry this should always check the warning/error bits as all packets should have them set. Now I think the validation test would pass if hard expiry gets reported for some packets but not all packets after the SA expiry.

test/validation/api/ipsec/ipsec_test_out.c
JannePeltonen: I think this should be handled more "automatically" by the generic code. I think now the test case code and the generic code are not well separated regarding the expiry checking.

test/validation/api/ipsec/ipsec.c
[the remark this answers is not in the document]
asasidharan: Ack. Made recv_event() change as a separate commit.

test/validation/api/ipsec/ipsec.h
asasidharan: I think now I understood what you meant by IGNORE: 
1. For both lookaside and inline cases, we would not ASSERT or return any errors based on the status/result returned.
2. For inline outbound status events, we will consume the status event so that it doesn't pile up.

But I am not quite sure if I understood the meaning for NONE.
I was thinking of having NONE for disabling all SA expiry related logic in the common code. The assumption is that SA expiry is not expected to happen during those tests with NONE as sa_expiry, and it is safe to disable the SA expiry test logic in the common code.

I have updated the tests with these two semantics for IGNORE and NONE. Let me know if this is in alignment with your thoughts.

> The test would then send a bunch of packets with expiry causing a valitation error, then another bunch of packets where expiry is allowed to happen but not required, and finally a third bunch of packets where expiry must happen.

I have updated the test code such that the second loop of test packets would stop once sa_expiry_notified is set. And added another loop where the test packets are sent until it crosses limit + delta. As suggested, this will allow us to check that hard expiry is set for every subsequent packet after sa_expiry_notified is set.

test/validation/api/ipsec/ipsec_test_out.c
asasidharan: Ack. Realized that we could make use of IGNORE here to make the handling "automatic". Let me know if the test case code now looks OK.

test/validation/api/ipsec/ipsec.c
asasidharan: Ack. Renamed to EVENT_CACHE_SIZE.

test/validation/api/ipsec/ipsec.c
asasidharan: Ack. Renamed to cached_event_s.

test/validation/api/ipsec/ipsec.c
```diff
@@ -101,11 +115,85 @@ static int pktio_start(odp_pktio_t pktio, odp_bool_t in, odp_bool_t out)
 	return 1;
 }
 
-static odp_event_t sched_queue_deq(uint64_t wait_ns)
+static int sched_event_cache_add(odp_queue_t from, odp_event_t event)
+{
+	odp_spinlock_lock(&sched_ev_cache_lock);
```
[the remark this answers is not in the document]
asasidharan: Ack. Removed the use of spinlock.

test/validation/api/ipsec/ipsec.c
asasidharan: Ack. Removed the redundant check.

test/validation/api/ipsec/ipsec.c
asasidharan: Ack. Using struct assignment now.

test/validation/api/ipsec/ipsec.c
asasidharan: Ack. Made it CU_ASSERT_EQUAL_FATAL().

test/validation/api/ipsec/ipsec.c
asasidharan: Ack. Only observation with same timeout for all the events was that the test was taking slightly longer time to run. Keeping the same timeout for all the cases.

test/validation/api/ipsec/ipsec.c
asasidharan: Ack. Using NONE to disable the SA expiry related logic in common code and IGNORE now just ignores the values present in the status result.

test/validation/api/ipsec/ipsec.c
asasidharan: Ack. Reverting this.

test/validation/api/ipsec/ipsec.c
asasidharan: Agreed. I think in this function (ipsec_send_out_one()) if the SA expiry is NONE, we can assert on status.error.all.
If sa_expiry is set to any value other than NONE, the result status structure is subsequently getting checked in ipsec_check_out() via ipsec_check_sa_expiry().

I had introduced the new function ipsec_check_sa_expiry() to check the status fields for better code indentation.

test/validation/api/ipsec/ipsec.c
asasidharan: I didn't quite understand the first part of the question. Did you mean to keep a CU_ASSERT for warn bits similar to the error bits like:
                        CU_ASSERT_EQUAL(part->out[i].status.warn.all,
                                        result.status.warn.all)

> Now I think the validation test would pass if hard expiry gets reported for some packets but not all packets after the SA expiry.
In the "else if {}"part, we are having CU_ASSERT on hard_exp_[bytes|packets] for every subsequent packet once sa_expiry_notified is set for the first time in the "if {}" part of the if - else if.

As mentioned earlier, this part of the code is now moved to ipsec_check_sa_expiry() function for better code indentation.

anoobj:
@JannePeltonen @psavol Can this patch be reviewed?

platform/linux-generic/include/odp_ipsec_internal.h
```diff
@@ -272,6 +272,13 @@ struct ipsec_sa_s {
 			} out;
 		};
 	} sa_info;
+
+	/*
+	 * Flag to check if the SA expiry status event was already
+	 * sent. This is applicable only for IPsec SA in inline outbound
+	 * mode.
+	 */
+	odp_atomic_u32_t expiry_notified;
```
JannePeltonen, line 281: It would be good mention in the comment that this is only for the soft expiry status event. Perhaps the field could also be renamed to soft_expiry_notified?

test/validation/api/ipsec/ipsec.c
```diff
@@ -16,11 +16,21 @@
 #include "test_vectors.h"
 #include "reass_test_vectors.h"
 
+#define EVENT_CACHE_SIZE 3
```
JannePeltonen, line 19: The commit message and the subject line could be impoved since "event caching" does not tell much. How about something like this:

```
validation: ipsec: get event only from the specified queue

When expecting an event from a certain queue, get an event only from that
queue even when the queue is a scheduled queue. Add a queue parameter to
sched_queue_deq() and return an  event only when it came from the specified
queue. If the scheduler returns an event from another queue, buffer it
locally for possible later dequeue from that queue.
```

I also think buffering would be more accurate a term than caching. Perhaps you would like to rename the relevant identifiers in the code?

test/validation/api/ipsec/ipsec.c
```diff
@@ -437,6 +440,49 @@ void ipsec_sa_param_fill(odp_ipsec_sa_param_t *param,
...
+
+	odp_event_free(ev_status);
+}
+
+void ipsec_status_event_get(odp_ipsec_sa_t sa,
+			    enum ipsec_test_sa_expiry sa_expiry)
+{
+	odp_event_t ev;
+
+	ev = recv_event(suite_context.queue, STATUS_EVENT_WAIT_TIME);
```
JannePeltonen, line 481: The wait here is slowing down the inline outbound test from less than 1 second to more than 20 seconds, which is going to slow down the CI too.

I think instead of lowering the wait time, the code could perhaps try to be a bit smarter about when a nonzero wait time is needed. One think that could help is to not wait if sa_expiry is IPSEC_TEST_EXPIRY_IGNORED, but I am not sure if that is the best or only thing to do.

test/validation/api/ipsec/ipsec.c
```diff
@@ -765,6 +813,44 @@ static int ipsec_process_in(const ipsec_test_part *part,
 	return num_out;
 }
 
+static int ipsec_check_sa_expiry(enum ipsec_test_sa_expiry sa_expiry,
+				 odp_ipsec_packet_result_t *result)
+{
+	if (sa_expiry != IPSEC_TEST_EXPIRY_IGNORED &&
```
JannePeltonen, line 819: The test whether sa_expiry equals IPSEC_TEST_EXPIRY_IGNORED could be done in the beginning to avoid repeating it later and cluttering the code. i.e.:

```
if (sa_expiry == IPSEC_TEST_EXPIRY_IGNORED)
        return 0;
```

test/validation/api/ipsec/ipsec.c
```diff
@@ -642,9 +688,11 @@ static int recv_pkts_inline(const ipsec_test_part *part,
 			CU_ASSERT(odp_event_is_valid(ev) == 1);
 			CU_ASSERT_EQUAL(ODP_EVENT_PACKET,
 					odp_event_types(ev, &subtype));
-			CU_ASSERT_EQUAL(ODP_EVENT_PACKET_BASIC,
-					subtype);
-			CU_ASSERT(part->out[i].status.error.sa_lookup);
+			if (ODP_EVENT_PACKET_IPSEC != subtype) {
```
JannePeltonen, line 691: This change looks wrong since we should never get ipsec subtype events from a pktio queue. If a packet was not ipsec-processed, it comes through the pktio queue as plain packet and if it was ipsec-processed it does not come through the pktio queue at all but through the SA queue.

This change also does not seem to be related to the topic of the PR.

test/validation/api/ipsec/ipsec.c
```diff
@@ -891,18 +977,33 @@ static int ipsec_send_out_one(const ipsec_test_part *part,
...
+					continue;
+				}
+
 				CU_ASSERT_EQUAL(ODP_EVENT_PACKET,
-						odp_event_types(ev, &subtype));
+						ev_type);
 				CU_ASSERT_EQUAL(ODP_EVENT_PACKET_IPSEC,
 						subtype);
-				CU_ASSERT(part->out[i].status.error.all);
+				if (part->out[i].sa_expiry == IPSEC_TEST_EXPIRY_NONE)
```
JannePeltonen, line 1005: I do not see the reason for this change. We should not get an event in the SA queue unless there was an error (and thus some error flag must be set) or the event is a status event, but status event have already been handled. So this change looks unnecessary to me. If it is not, maybe it deserves a clarifying comment?

JannePeltonen (changes requested):

platform/linux-generic/include/odp_ipsec_internal.h
asasidharan: Ack. Will update with the next version.

test/validation/api/ipsec/ipsec.c
asasidharan: Ack. Will rename as suggested.

test/validation/api/ipsec/ipsec.c
asasidharan: In the initial proposal it was with zero wait time to mitigate this issue. It was updated as per your feedback. Are you suggesting to revert to earlier proposal?

test/validation/api/ipsec/ipsec.c
asasidharan: Ack. Will update so with the next version.

test/validation/api/ipsec/ipsec.c
JannePeltonen: No I am not. This is not a binary decision but there are many different ways to do the logic.

test/validation/api/ipsec/ipsec.c
asasidharan: When expiry is being tested (i.e, sa_expiry != IPSEC_TEST_EXPIRY_NONE), eventually ipsec_check_sa_expiry() gets invoked from ipsec_check_out(), the caller of this function (ipsec_send_out_one). In ipsec_check_sa_expiry(), the error bits are getting checked as expected. I can add comment as below to clarify this:

```
+
+                               /* In the case of SA hard expiry tests, hard expiry error bits are
+                                * expected to be set. The exact error bits expected to be set based
+                                * on sa_expiry is checked eventually in ipsec_check_sa_expiry()
+                                * from the caller of this function.
+                                */
```

Let me know if this works.

test/validation/api/ipsec/ipsec.c
asasidharan: Ack. Reverted the change with v11.

asasidharan:
With v11:

- linux-gen: renamed expiry_notified field to soft_expiry_notified
- Updated commit message for 2nd commit and replaced usage of "cache" with "buffer"
- Added comment to clarify the checking of error bits in IPsec result in case of SA expiry tests.

test/validation/api/ipsec/ipsec.c
asasidharan: Updated the change to use zero wait time when sa_expiry is  IPSEC_TEST_EXPIRY_IGNORED in v12.

asasidharan:
With v12 updated the test to use zero wait time while receiving status event when sa_expiry is IPSEC_TEST_EXPIRY_IGNORED.
Please review the updated PR.

anoobj:
@JannePeltonen @psavol Can this PR be reviewed and closed before next release?

test/validation/api/ipsec/ipsec.c
JannePeltonen: I am not sure the comment really helps to understand why the assertion should be skipped when the expiry test is ongoing. The real reason seems to be that the expected error (part->out[i].status.error.all) is not set to match what is really expected (and in some cases the expected result is that an expiry status may or may not be set).

This is actually part of the bigger issue that the added expiry check does not quite follow the earlier pattern that test cases set the test packet and the expected output in the test_part and then let ipsec.c just do the processing and checking without knowing avout the test case. This patch blurs that a bit by having expiry test specific logic inside ipsec.c. OTOH, part of that is a bit unavoidable since SA expiry reporting (especially with soft limits) is not exact.

JannePeltonen (approved):

JannePeltonen:
I think the validation test code could be made cleaner as I commented and that might also help avoiding the current slowdown caused by waiting the status events. But since this PR has been going on for a long time and makes things better, even if not perfect, by adding missing validation tests and by fixing a bug in the implementation, I think it would probably be best to leave the improvements for later PRs and merge this in this form now. Thus I approved the PR now.

asasidharan:
@JannePeltonen Thank you for your time in review the PR. I have added your "Reviewed-by" tag and rebased to the latest commit on master branch with v14.

PR #1512, opened 2022-02-18
[PATCH v3] test: performance: crypto: use non-empty AAD and fix capability checking

JannePeltonen:
test: performance: crypto: use AAD of 8 bytes

    AEAD algorithms are typically used with non-empty AAD, so use
    AAD of 8 bytes in the tests when applicable.

    test: performance: crypto: fix capability checking and enable bit mode

    Check that an algorithm is supported and supports thew desired key
    lengths etc before trying to test it. Do not assume that algorithms
    work in byte mode but used them in bit mode if algorithm capabilities
    indicate bit mode operation.

### DIFF
--- a/test/performance/odp_crypto.c
+++ b/test/performance/odp_crypto.c
@@ -1,4 +1,5 @@
 /* Copyright (c) 2015-2018, Linaro Limited
+ * Copyright (c) 2022, Nokia
  * All rights reserved.
  *
  * SPDX-License-Identifier:	BSD-3-Clause
@@ -24,6 +25,9 @@
  */
 #define POOL_NUM_PKT  64
 
+#define AAD_LEN 8 /* typical AAD length used in IPsec when ESN is not in use */
+
+static uint8_t test_aad[AAD_LEN] = "01234567";
 static uint8_t test_iv[16] = "0123456789abcdef";
 
 static uint8_t test_key16[16] = { 0x01, 0x02, 0x03, 0x04, 0x05,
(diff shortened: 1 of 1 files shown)


test/performance/odp_crypto.c
```diff
@@ -80,6 +80,8 @@ static uint8_t test_key64[64] = { 0x01, 0x02, 0x03, 0x04, 0x05,
 typedef struct {
```
JereLeppanen, line 80: A couple of typos in the commit description: "thew" and "used".

test/performance/odp_crypto.c
```diff
@@ -80,6 +80,8 @@ static uint8_t test_key64[64] = { 0x01, 0x02, 0x03, 0x04, 0x05,
...
 	const char *name;		      /**< Algorithm name */
 	odp_crypto_session_param_t session;   /**< Prefilled crypto session params */
+	int cipher_in_bit_mode;               /* probed at run time */
+	int auth_in_bit_mode;                 /* probed at run time */
```
JereLeppanen, line 84: Use uniform comment style for all members of a struct, if possible.

JereLeppanen (approved):

test/performance/odp_crypto.c
JannePeltonen: Fixed in v2.

test/performance/odp_crypto.c
JannePeltonen: Fixed in v2. I also made the comments hopefully a bit more descriptive.

JannePeltonen:
v2: rebased and handled the review comments.

JannePeltonen:
v3: added reviewed-by tag

PR #1513, opened 2022-02-18
[PATCH v3] validation: crypto: fix checking of AAD length capability

JannePeltonen:
Checking of AAD length capability is entirely missing. Add a check that
the AAD length used in a test vector is actually supported before
executing the test.

### DIFF
--- a/test/validation/api/crypto/odp_crypto_test_inp.c
+++ b/test/validation/api/crypto/odp_crypto_test_inp.c
@@ -777,6 +777,19 @@ static void alg_test(odp_crypto_op_t op,
 	CU_ASSERT(!rc);
 }
 
+static odp_bool_t aad_len_ok(const odp_crypto_auth_capability_t *capa, uint32_t len)
+{
+	if (len < capa->aad_len.min || len > capa->aad_len.max)
+		return false;
+
+	if (len == capa->aad_len.min)
+		return true;
+	if (capa->aad_len.inc == 0)
+		return false;
+
+	return ((len - capa->aad_len.min) % capa->aad_len.inc) == 0;
+}
+
 static void check_alg(odp_crypto_op_t op,
 		      odp_cipher_alg_t cipher_alg,
 		      odp_auth_alg_t auth_alg,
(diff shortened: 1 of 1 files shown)


test/validation/api/crypto/odp_crypto_test_inp.c
```diff
@@ -777,6 +777,19 @@ static void alg_test(odp_crypto_op_t op,
...
 
+static odp_bool_t aad_len_ok(const odp_crypto_auth_capability_t *capa, uint32_t len)
+{
+	if (len < capa->aad_len.min || len > capa->aad_len.max)
+		return false;
+
+	if (len == capa->aad_len.min)
+		return true;
+	if (capa->aad_len.inc == 0)
+		return false;
```
JereLeppanen, line 788: This requires that inc is nonzero, but API does not. Is this ok?

JannePeltonen: It does not require that inc is nonzero in all cases. If the desired AAD length is the same as the minimum advertized AAD length, then inc is ignored.

My interpretation of the API is that when inc is zero, then only the min AAD length is supported. So if max differs from inc, it is not supported if inc is zero. I think another possible interpretation is that min and max are always supported, regardless of inc, and then other values in between may be supported, depending on the value of inc. I chose the former in the added check. I am not sure if it is right, but it is more conservative.

Maybe the validation check should sanity-check the capability values by cheking that the difference of max and min is either zero or divisible by inc?

JereLeppanen: Shouldn't validation tests use the most liberal interpretation of the API? Which would be your latter interpretation. Although another possible interpretation might even be that inc=0 is the same as inc=1. A bit of a stretch, I know.

Either way is actually fine, I think, this is just academic nitpicking.

JannePeltonen: > Shouldn't validation tests use the most liberal interpretation of the API?

I am not sure. Maybe the most liberal interpretation is not actually correct? A stricter validation test can enforce the stricter interpretation. If it becomes a problem, it can be easily relaxed. With a liberal test we might not notice at all if an implementation behaved weirdly, even if it does it unintentionally.

JereLeppanen (approved):

PR #1514, opened 2022-02-22
[PATCH v6] linux-gen: implement tx packet aging

TuomasTaipale:
Add aging support for packets enqeued for transmission to linux-gen. This ends up affecting traffic manager only as in practice there is no long output queuing elsewhere.

v2:
- address Jere's comments

v3:
- address Matias' comment

v4:
- rebase on-top-of #1495 
- add reviewed-by tags

v5:
- rebase

### DIFF
--- a/platform/linux-generic/include/odp/api/plat/packet_inline_types.h
+++ b/platform/linux-generic/include/odp/api/plat/packet_inline_types.h
@@ -115,7 +115,7 @@ typedef union {
 	uint32_t all_flags;
 
 	struct {
-		uint32_t reserved1:      7;
+		uint32_t reserved1:      6;
 
 	/*
 	 * Init flags
@@ -133,6 +133,7 @@ typedef union {
 		uint32_t l4_chksum:      1; /* L4 chksum override */
 		uint32_t ts_set:         1; /* Set Tx timestamp */
 		uint32_t tx_compl:       1; /* Tx completion event requested */
+		uint32_t tx_aging:       1; /* Packet aging at Tx requested */
 		uint32_t shaper_len_adj: 8; /* Adjustment for traffic mgr */
 
 	/*
@@ -150,8 +151,8 @@ typedef union {
 
 	/* Flag groups */
 	struct {
-		uint32_t reserved2:      7;
-		uint32_t other:         17; /* All other flags */
+		uint32_t reserved2:      6;
+		uint32_t other:         18; /* All other flags */
 		uint32_t error:          8; /* All error flags */
 	} all;
 
--- a/platform/linux-generic/include/odp_packet_internal.h
+++ b/platform/linux-generic/include/odp_packet_internal.h
@@ -146,6 +146,10 @@ typedef struct ODP_ALIGNED_CACHE odp_packet_hdr_t {
 	/* Max payload size in a LSO segment */
 	uint16_t lso_max_payload;
 
+	/* Packet aging drop timeout before enqueue. Once enqueued holds the maximum age (time of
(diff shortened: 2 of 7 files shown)


platform/linux-generic/include/odp/api/plat/packet_inline_types.h
```diff
@@ -115,7 +115,7 @@ typedef union {
 	uint32_t all_flags;
```
JereLeppanen, line 115: Commit subject and description: use singular API, avoid plural. Actually mentioning API at all may be redundant, since practically any implementation implements some part of the API.

platform/linux-generic/odp_packet_io.c
```diff
@@ -50,6 +50,9 @@
 /* Max wait time supported to avoid potential overflow */
 #define MAX_WAIT_TIME (UINT64_MAX / 1024)
 
+/* One hour maximum aging timeout */
+#define MAX_TX_AGING_TMO_NS 3600000000000ULL
```
JereLeppanen, line 54: If there is some reason for limiting the timeout to this value, it could be mentioned here.

platform/linux-generic/odp_traffic_mngr.c
```diff
@@ -1994,6 +1994,12 @@ static void tm_queue_cnts_decrement(tm_system_t *tm_system,
 	odp_atomic_sub_u64(&queue_cnts->byte_cnt, frame_len);
```
JereLeppanen, line 1994: Commit description: "determines" -> "considers"?

JereLeppanen (approved):
Looks ok to me, I only have a couple of very minor comments.

platform/linux-generic/include/odp_packet_internal.h
```diff
@@ -149,6 +149,10 @@ typedef struct ODP_ALIGNED_CACHE odp_packet_hdr_t {
 	/* LSO profile index */
 	uint8_t lso_profile_idx;
 
+	/* Packet aging drop timeout before enqueue. Once enqueued holds the maximum age (time of
+	 * request + requested drop timeout). */
+	uint64_t tx_aging_ns;
```
MatiasElo, line 154: I'd swap this and `lso_profile_idx`, so the gap in struct would be in a clearer location.

platform/linux-generic/include/odp/api/plat/packet_inline_types.h
TuomasTaipale: Sure, dropped the plural. Keeping the API word still as this is a multi-commit feature and in this commit just the actual interface level functions are modified (with one big squashed commit it would be a bit pointless yes).

platform/linux-generic/odp_packet_io.c
TuomasTaipale: No reason really, just some value -> opened up the fudge factorness with a comment.

platform/linux-generic/odp_traffic_mngr.c
TuomasTaipale: Sure, "considers" it is.

platform/linux-generic/include/odp_packet_internal.h
TuomasTaipale: OK, swapped in v3. Didn't do any calculations, hopefully alignment is better now.

MatiasElo (approved):

TuomasTaipale:
Update: rebased on top of LSO with TM changes and modified packet aging traffic manager test case accordingly.

PR #1515, opened 2022-03-03
[PATCH v4] linux-gen: queue: add missing include directive

JereLeppanen:
linux-gen: queue: add missing include directive
    
Include odp/api/event.h in queue_inline_types.h, since it uses odp_event_t. This omission also prevented the proto_stats.h API header from being compiled on its own.
    
v3:
- Include order.

v4:
- Rebase.
- Add review tags.


MatiasElo (approved):
Headers could be included in alphabetical order but OK either way.

PR #1518, opened 2022-03-08
[PATCH v6] crypto: miscellaneous ARM crypto optimizations

JannePeltonen:
linux-gen: crypto: arm: avoid unneeded packet data copy
linux-gen: crypto: arm: use constant length memcpy for AAD
linux-gen: crypto: arm: optimize authentication tag handling
linux-gen: crypto: arm: combine cipher and auth into one function

### DIFF
--- a/platform/linux-generic/arch/aarch64/odp_crypto_armv8.c
+++ b/platform/linux-generic/arch/aarch64/odp_crypto_armv8.c
@@ -1,5 +1,6 @@
 /* Copyright (c) 2014-2018, Linaro Limited
  * Copyright (c) 2021, ARM Limited
+ * Copyright (c) 2022, Nokia
  * All rights reserved.
  *
  * SPDX-License-Identifier:     BSD-3-Clause
@@ -37,9 +38,9 @@
 
 /*
  * ARM crypto library may read up to 15 bytes past the end of input
- * data and AAD.
+ * data and AAD and write up to 15 bytes past the end of output data.
  */
-#define OOB_READ_LEN 15
+#define OOB_WRITE_LEN 16 /* rounded up to 16 bytes for efficiency */
 
 /*
  * Data buffer size must be a multiple of 16, because the ARM crypto
@@ -201,6 +202,8 @@ odp_crypto_alg_err_t aes_gcm_encrypt(odp_packet_t pkt,
 	uint32_t in_pos = param->cipher_range.offset;
 	uint32_t in_len = param->cipher_range.length;
 	int ret = 0;
+	odp_bool_t continuous_data;
+	uint16_t saved_tail[OOB_WRITE_LEN];
 
 	/* Fail early if cipher_range is too large */
 	if (in_len > ARM_CRYPTO_MAX_DATA_LENGTH) {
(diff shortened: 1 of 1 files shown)


platform/linux-generic/arch/aarch64/odp_crypto_armv8.c
```diff
@@ -239,28 +242,34 @@ odp_crypto_alg_err_t aes_gcm_encrypt(odp_packet_t pkt,
 	uint8_t *data = odp_packet_offset(pkt, in_pos, &seg_len, NULL);
 
 	if (odp_unlikely(odp_packet_is_segmented(pkt)) ||
```
TuomasTaipale, line 269: Might have been nice to get this and the if-esle block at line 269 to separate functions as they seem to repeat in encrypt and decrypt, but I guess this becomes a bit cumbersome due to the tag handling asymmetry introduced in a following patch.

platform/linux-generic/arch/aarch64/odp_crypto_armv8.c
```diff
@@ -236,7 +236,10 @@ odp_crypto_alg_err_t aes_gcm_encrypt(odp_packet_t pkt,
 	 * read it in 16 byte chunks. */
 	uint8_t aad[ARM_CRYPTO_MAX_AAD_LENGTH];
 
-	memcpy(aad, param->aad_ptr, session->p.auth_aad_len);
+	if (session->p.auth_aad_len == 8)
+		memcpy(aad, param->aad_ptr, 8);
```
TuomasTaipale, line 262: Alternatively:
```C
memcpy(aad, param->aad_ptr, session->p.auth_aad_len == 8 ? 8 : 12);
```

platform/linux-generic/arch/aarch64/odp_crypto_armv8.c
```diff
@@ -148,6 +144,8 @@ typedef struct crypto_local_t {
 
 static __thread crypto_local_t local;
 
+static odp_crypto_packet_result_t *get_op_result_from_packet(odp_packet_t pkt);
```
TuomasTaipale, line 147: Definition of `get_op_result_from_packet()` could be just moved before `set_crypto_op_result()` as it is now used in that function for the first time.

platform/linux-generic/arch/aarch64/odp_crypto_armv8.c
```diff
@@ -178,18 +176,41 @@ void free_session(odp_crypto_generic_session_t *session)
...
+{
+	odp_crypto_packet_result_t *op_result;
+
+	op_result = get_op_result_from_packet(pkt);
+	op_result->cipher_status.alg_err = cipher_err;
+	op_result->cipher_status.hw_err = ODP_CRYPTO_HW_ERR_NONE;
+	op_result->auth_status.alg_err = auth_err;
+	op_result->auth_status.hw_err = ODP_CRYPTO_HW_ERR_NONE;
+	op_result->ok = (cipher_err == ODP_CRYPTO_ALG_ERR_NONE &&
+			 auth_err   == ODP_CRYPTO_ALG_ERR_NONE);
```
TuomasTaipale, line 191: Extra whitespaces after `auth_err`.

platform/linux-generic/arch/aarch64/odp_crypto_armv8.c
```diff
@@ -178,18 +176,41 @@ void free_session(odp_crypto_generic_session_t *session)
...
+}
+
+static inline void set_crypto_op_result_ok(odp_packet_t pkt)
```
TuomasTaipale, line 194: `set_crypto_op_result_success`?

platform/linux-generic/arch/aarch64/odp_crypto_armv8.c
```diff
@@ -405,9 +434,9 @@ static int process_aes_gcm_param(odp_crypto_generic_session_t *session)
 
 	/* Set function */
 	if (ODP_CRYPTO_OP_ENCODE == session->p.op)
-		session->cipher.func = aes_gcm_encrypt;
+		session->func = aes_gcm_encrypt;
```
TuomasTaipale, line 437: Alternatively:
```c
session->func = ODP_CRYPTO_OP_ENCODE == session->p.op ? aes_gcm_encrypt : aes_gcm_decrypt;
```

platform/linux-generic/arch/aarch64/odp_crypto_armv8.c
JannePeltonen: Will move it.

platform/linux-generic/arch/aarch64/odp_crypto_armv8.c
JannePeltonen: The extra white space was intentional to line up the two lines of the expression nicely.

platform/linux-generic/arch/aarch64/odp_crypto_armv8.c
JannePeltonen: I am not sure if that is any better. The result structure uses "ok" rather than "success" so "ok" here matches that better, I think.

platform/linux-generic/arch/aarch64/odp_crypto_armv8.c
JannePeltonen: I do not see the benefit of that here. I think it it would be slightly less readable and would also be a slightly bigger change.

platform/linux-generic/arch/aarch64/odp_crypto_armv8.c
TuomasTaipale: OK, angle here was that we have more "success"/"failure" than "ok"/"nok" in code, but perhaps "ok" fits here better.

platform/linux-generic/arch/aarch64/odp_crypto_armv8.c
TuomasTaipale: OK, this could be in the commit (cbcad2c) log.

JannePeltonen: Improved the log in v4 and also moved the optimized copy to a function and added a comment in the code.

JannePeltonen:
v4: Made the AAD copying patch cleaner.

JannePeltonen:
v5: fixed a typo in the previous version

TuomasTaipale (approved):

PR #1522, opened 2022-03-11
[PATCH v10] api: crypto: deprecate odp_crypto_operation()

JannePeltonen:
Depracate odp_crypto_operation() and the associated data structures and the completion event.

The odp_crypto_operation() function is a bit problematic since it can be synchronous or asynchronous on per-packet basis without application having control on it. Uncontrolled sync versus async behaviour can have adverse effects on packet ordering and application design in general. The function is also somewhat redundant with the newer odp_crypto_op() and odp_crypto_op_enq() functions that provide explicit control on the crypto operation completion behavior.

### DIFF
--- a/include/odp/api/spec/crypto.h
+++ b/include/odp/api/spec/crypto.h
@@ -41,7 +41,7 @@ extern "C" {
 
 /**
  * @typedef odp_crypto_compl_t
-* Crypto API completion event (platform dependent).
+*  @deprecated Crypto API completion event (platform dependent).
 */
 
 /**
@@ -558,8 +558,10 @@ typedef struct odp_crypto_session_param_t {
 	/** Preferred sync vs. async for odp_crypto_operation()
 	 *
 	 *  The default value is ODP_CRYPTO_SYNC.
+	 *
+	 *  @deprecated Used only with deprecated odp_crypto_operation()
 	 */
-	odp_crypto_op_mode_t pref_mode;
+	odp_crypto_op_mode_t ODP_DEPRECATE(pref_mode);
 
 	/** Operation mode when using packet interface: sync or async
 	 *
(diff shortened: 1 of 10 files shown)


JannePeltonen:
v2: fixed compliation with --disable-openssl and fixed a doxygen issue.

JannePeltonen:
v3: added a patch to update user guide

JannePeltonen:
v4: Removed (= made conditional on ODP_DEPRECATED_API) the removed crypto completion event type in ABI headers and the remaining references to it.

JannePeltonen:
v5: fixed a silly typo

JannePeltonen:
v6: rebased after the event header split in master

include/odp/api/spec/crypto.h
```diff
@@ -1140,9 +1144,13 @@ int odp_crypto_session_create(const odp_crypto_session_param_t *param,
  */
 int odp_crypto_session_destroy(odp_crypto_session_t session);
 
+#if ODP_DEPRECATED_API
+
 /**
  * Return crypto completion handle that is associated with event
  *
+ * @deprecated
```
psavol, line 1152: Doxygen documentation looks nicer when some text follows the "@deprecated" tag.

include/odp/api/spec/crypto.h
```diff
@@ -559,7 +559,7 @@ typedef struct odp_crypto_session_param_t {
 	 *
 	 *  The default value is ODP_CRYPTO_SYNC.
 	 */
-	odp_crypto_op_mode_t pref_mode;
+	odp_crypto_op_mode_t ODP_DEPRECATE(pref_mode);
```
psavol, line 562: compl_queue documentation (under) refers to odp_crypto_operation(). That reference could be removed.

include/odp/api/spec/crypto.h
JannePeltonen: Added text in v7.

include/odp/api/spec/crypto.h
JannePeltonen: I reworded the text but did not completely remove the reference since the usage with odp_crypto_operation() is still part of the deprecated variant of the API.

psavol (approved):

JannePeltonen:
v8: added reveiw tags and added @deprecated to the comment for the deprecated pref_mode session parameter.

anoobj:

PR #1526, opened 2022-03-18
[PATCH v2] api: event: split header file

psavol:
Split event.h API file into separate files for types and functions. No API spec change.

### DIFF
--- a/include/odp/api/spec/crypto.h
+++ b/include/odp/api/spec/crypto.h
@@ -16,14 +16,13 @@
 #include <odp/visibility_begin.h>
 
 #include <odp/api/deprecated.h>
+#include <odp/api/packet_types.h>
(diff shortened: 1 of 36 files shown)


include/odp/api/spec/crypto.h
```diff
@@ -18,13 +18,12 @@
 #include <odp/api/deprecated.h>
 #include <odp/api/pool_types.h>
 #include <odp/api/std_types.h>
+#include <odp/api/packet_types.h>
```
MatiasElo, line 21: Alphabetical order

MatiasElo (approved):

include/odp/api/spec/crypto.h
psavol: fixed in v2

PR #1531, opened 2022-03-23
validation: buffer: test with cache enabled/disabled

ashwinyes:
Buffer pools can be created with and without local cache.
Add tests to validate both cases.


test/validation/api/buffer/buffer.c
```diff
@@ -370,6 +376,56 @@ static void buffer_test_pool_max_pools(void)
...
 }
 
+static void buffer_test_pool_alloc_free(void)
+{
+	odp_pool_param_t param;
+
+	memcpy(&param, &default_param, sizeof(odp_pool_param_t));
+	test_pool_alloc_free(&param);
+	set_cache_size(&param);
+	test_pool_alloc_free(&param);
```
MatiasElo, line 386: The default cache size is implementation specific and for example with linux-generic it's 256. So, perhaps it makes sense to add three test cases: min, default, and max.

test/validation/api/buffer/buffer.c
```diff
@@ -370,6 +376,56 @@ static void buffer_test_pool_max_pools(void)
...
+}
+
+static void buffer_test_pool_alloc_free_multi(void)
+{
+	odp_pool_param_t param;
+
+	memcpy(&param, &default_param, sizeof(odp_pool_param_t));
+	test_pool_alloc_free_multi(&param);
+	set_cache_size(&param);
+	test_pool_alloc_free_multi(&param);
```
MatiasElo, line 396: Running the same function multiple times within a test case can make debugging failures a bit difficult as there may be no way of knowing which parameters were used based on the test output alone. So perhaps it would make sense to separate these into multiple tests.
E.g.
```
buffer_test_pool_alloc_free_multi
buffer_test_pool_alloc_free_multi_min_cache
buffer_test_pool_alloc_free_multi_max_cache
```
Same goes for all the test cases.

ashwinyes:
@MatiasElo Ping

MatiasElo (approved):

PR #1536, opened 2022-03-24
[PATCH v2] linux-gen: crypto: fix configure portability issue

MatiasElo:
OpenSSL and ARMv8 crypto implementations are platform specific, so wrap the
AM_CONDITIONALs to avoid configure failures on other platforms:
	configure: error: conditional "WITH_OPENSSL_CRYPTO" was never defined.

Suggested-by: Nithin Dabilpuram <ndabilpuram@marvell.com>


MatiasElo:
V2:
- Improved commit message

JannePeltonen (approved):

PR #1537, opened 2022-03-24
api: pktio: add capability for classifier CoS pool

JereLeppanen:
Add pktio capability for allocating received packets from the pool indicated by CoS.

### DIFF
--- a/include/odp/api/spec/classification.h
+++ b/include/odp/api/spec/classification.h
@@ -351,7 +351,11 @@ typedef struct odp_cls_cos_param {
 			odp_pktin_hash_proto_t hash_proto;
 		};
 	};
-	/** Pool associated with CoS */
+	/** Pool associated with CoS
+	 *
+	 * If the pktio does not support moving packets to another pool, this
+	 * parameter is ignored. @see odp_pktio_capability_t::cls_cos_pool
+	 */
 	odp_pool_t pool;
 
 	/** Drop policy associated with CoS */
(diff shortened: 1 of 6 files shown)


include/odp/api/spec/classification.h
```diff
@@ -351,7 +351,11 @@ typedef struct odp_cls_cos_param {
 			odp_pktin_hash_proto_t hash_proto;
 		};
 	};
-	/** Pool associated with CoS */
+	/** Pool associated with CoS
+	 *
+	 * If the pktio does not support moving packets to another pool, this
+	 * parameter is ignored. @see odp_pktio_capability_t::cls_cos_pool
+	 */
```
psavol, line 358: Pool parameter documentation could be improved:
- default value is POOL_INVALID
- when POOL_INVALID, pool configured to packet IO is used
- when !=POOL_INVALID, this pool is used
- new capa tells if this feature is supported. cos_create() returns failure if pool is set, but feature not supported. But we could leave it undefined (as usual) what happens if application acts against capa.

include/odp/api/spec/packet_io_types.h
```diff
@@ -917,6 +917,11 @@ typedef struct odp_pktio_capability_t {
 	/** Statistics counters capabilities */
 	odp_pktio_stats_capability_t stats;
 
+	/** This defines whether the pktio supports moving received packets to
+	 * the pool indicated by Class of Service. @see
+	 * odp_cls_cos_param_t::pool */
+	odp_bool_t cls_cos_pool;
+
```
psavol, line 924: This would need to be part of odp_cls_capability_t.

It's another topic to decide if per pktio interface cls capa is needed. Maybe it is and then we could add odp_cls_capability_t into odp_pktio_capability_t.

JereLeppanen:
As Petri commented, in order to do this properly, we would need pktio specific cls capabilities. Right now it's best to just add the missing implementation. The capabilites can be added later if they're really needed.

PR #1538, opened 2022-03-25
api: convert unsigned int to uint32_t

psavol:
Convert remaining 'unsigned int' types in API spec to uint32_t. Pool and packet IO maximum index functions return still 'unsigned int' since those are paired with index functions which return 'int'.


include/odp/api/spec/packet_io_types.h
```diff
@@ -257,7 +257,7 @@ typedef struct odp_pktin_queue_param_t {
 	  * More than one input queues require flow hashing configured.
 	  * The maximum value is defined by pktio capability 'max_input_queues'.
 	  * Queue type is defined by the input mode. The default value is 1. */
-	unsigned int num_queues;
+	uint32_t num_queues;
```
MatiasElo, line 260: `doc/users-guide/users-guide-pktio.adoc` has still code fragments with the old `unsigned int` types.

MatiasElo (approved):

include/odp/api/spec/packet_io_types.h
psavol: User guide includes many copies of API structures. Agreed to minimize amount of copy-paste code in user guide in another PR.

jerinjacobk (approved):

PR #1542, opened 2022-03-28
[PATCH v1] validation: classification: increase wait time to receive packets

sunilkori:
As on ASIM, there is time delay between sending and receiving packets
on loopback interface. Once application sends packet and tries to
receive packets immediatly then application might get lesser packets
on ASIM.

This patch increases wait time so that all sent packets can be enqueue'd
to Rx queues and application can receive al the packets.

### DIFF
--- a/test/validation/api/classification/odp_classification_tests.c
+++ b/test/validation/api/classification/odp_classification_tests.c
@@ -354,7 +354,7 @@ void test_cls_pmr_chain(odp_bool_t enable_pktv)
 
 	enqueue_pktio_interface(pkt, pktio_loop);
 
-	pkt = receive_packet(&queue, ODP_TIME_SEC_IN_NS, enable_pktv);
+	pkt = receive_packet(&queue, 60 * ODP_TIME_SEC_IN_NS, enable_pktv);
 	CU_ASSERT_FATAL(pkt != ODP_PACKET_INVALID);
 	CU_ASSERT(queue == queue_list[CLS_PMR_CHAIN_DST]);
 	CU_ASSERT(seqno == cls_pkt_get_seq(pkt));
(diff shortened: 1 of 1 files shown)


test/validation/api/classification/odp_classification_tests.c
```diff
@@ -354,7 +354,7 @@ void test_cls_pmr_chain(odp_bool_t enable_pktv)
 
 	enqueue_pktio_interface(pkt, pktio_loop);
 
-	pkt = receive_packet(&queue, ODP_TIME_SEC_IN_NS, enable_pktv);
```
JannePeltonen, line 357: I am not sure this is the right thing to do.

I think the existing 1 second timeout is quite generous for any real ODP implementation. Increasing the timeout to one minute can make 'make check' and CI tests take a long time if something goes wrong, so maybe it should not be done unconditionally like this.

I am also afraid that if we tweak the code to handle special requirements of a simulation environment, the changes do not end here but changes would be needed in many other timeouts all over the code.

Perhaps a better way to handle the issues of a simulation environment would to scale the clock so that the clock runs slow enough that timeouts are not exeeded. This could be done in the ODP implementation without changing the API validation tests.

If that is not possible and if API validation test changes are really needed in this ODP repo, I think it would be good if the changes were somehow configurable e.g. through a compile time flag so that for normal ODP implementations the timeouts would not change.

sunilkori: I do agree with you and changes are relevant for Marvell simulation environment only.

JannePeltonen: Is this PR still relevant or should it be closed?

sunilkori: @JannePeltonen 

Sorry for the late reply. It should be closed as it is not applicable.

PR #1543, opened 2022-03-29
test: random perf latency mode

psavol:
Added new -m/-d options into random number performance test application, which are used to test latency of random number generation.


test/performance/odp_random.c
```diff
@@ -14,8 +14,32 @@
...
+
 #define MB (1024ull * 1024ull)
 
+typedef struct test_global_t test_global_t;
+
+typedef struct thread_arg_t {
+	test_global_t *global;
+	int thread_idx;
+	uint8_t *data;
+
```
TuomasTaipale, line 27: Extra newline.

test/performance/odp_random.c
```diff
@@ -97,198 +121,145 @@ static int parse_options(int argc, char *argv[])
...
+	int thread_idx = thread_arg->thread_idx;
+	uint8_t *data = thread_arg->data;
+	uint32_t size = options.size;
+	uint32_t rounds = global->rounds;
+
+	/* One warm up round */
+	if ((int)type == PSEUDO_RANDOM)
+		pseudo_random_data_loop(1, data, size);
+	else
+		random_data_loop(type, 1, data, size);
```
TuomasTaipale, line 221: This seems to repeat, could be in a separate function or alternatively `*_data_loop` signatures could be unified and called via a pointer.

test/performance/odp_random.c
```diff
@@ -200,6 +237,97 @@ static int test_random_perf(void *ptr)
...
+
+	for (i = 0; i < rounds; i++) {
+		uint32_t pos = 0;
+
+		if (delay)
+			odp_time_wait_ns(delay);
+
+		if ((int)type == PSEUDO_RANDOM) {
+			t1 = odp_time_local_strict();
+			while (pos < size) {
```
TuomasTaipale, line 264: These inner blocks seem very similar to the `*_data_loop` functions, reuse those?

TuomasTaipale (commented):
Typo in 5bae514 "Added also a *wram* up".

test/performance/odp_random.c
psavol: I prefer this style where the name of the type/struct is easier to find after the list of struct members.

TuomasTaipale: OK, maybe same treatment to `options_t`.

test/performance/odp_random.c
psavol: The difference is time measurement: for latency it is measured for every round, for performance it is measured once for all rounds (to avoid measurement overhead).

test/performance/odp_random.c
psavol: I think a new function is not necessary in this case, as it would save so few lines (add 6 new lines, remove 9 lines).

TuomasTaipale: OK. In general, it's a bit cumbersome to require the type check everytime there's a call to get random data.

psavol:
> Typo in [5bae514](https://github.com/OpenDataPlane/odp/commit/5bae514601098623c7c1c7bf596b0c4f7289ef7e) "... Added also a _wram_ up ...".

Fixed in v2

test/performance/odp_random.c
psavol: Done in v2.

test/performance/odp_random.c
psavol: Fixed in v2. Moved pseudo_random_data_loop() inside random_data_loop() as it had all needed arguments already.

PR #1545, opened 2022-03-30
[PATCH v3] test: performance: ipsec: add vector packet support

anoobj:
Receive IPsec processed packets as a vector packet on supported
platforms.

ODP allows IPsec processed packets to be received as a vector packet
by aggregating events with similar ordering context. New option
'--vector' or '-v' can be used to enable vector aggregation post IPsec
processing. When specified without any arguments, vector size would be
set as 32. User can control vector size by specifying the optional
argument.

Usage: 'odp_ipsec --vector=32'

### DIFF
--- a/test/performance/odp_ipsec.c
+++ b/test/performance/odp_ipsec.c
@@ -151,6 +151,12 @@ typedef struct {
 	 * iteration of the loop.
 	 */
 	int burst_size;
+
+	/*
+	 * Use vector packet completion from IPsec APIs.
+	 * Specified through -v or --vector argument.
+	 */
+	uint32_t vec_pkt_size;
 } ipsec_args_t;
 
 /*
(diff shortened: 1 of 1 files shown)


test/performance/odp_ipsec.c
```diff
@@ -25,7 +25,7 @@
 /** @def POOL_NUM_PKT
  * Number of packets in the pool
  */
-#define POOL_NUM_PKT  4096
+#define POOL_NUM_PKT  8192
```
JannePeltonen, line 28: This POOL_NUM_PKT change does not appear to have anything to do with the vector support and the rest of the patch, so I think it should be taken out of this patch for clarity, and made through a separate patch if there is a reason for it.

Otherwise this patch looks ok. Reviewed-by: Janne Peltonen <janne.peltonen@nokia.com>

JannePeltonen (approved):

PR #1548, opened 2022-04-07
[PATCH v6] linux-gen: scheduler: improve priority scheduling

psavol:
Currently, scheduler selects group first (round robin), and then finds highest priority queue from that. This leads to bad response time to high priority events when there are multiple groups, but only few of those have high priority queues.

This PR changes scheduler to check first highest priority level on all groups. This causes more work for each schedule call. A new mask of active groups per priority is introduced to optimize the additional checking.

### DIFF
--- a/platform/linux-generic/odp_schedule_basic.c
+++ b/platform/linux-generic/odp_schedule_basic.c
@@ -1202,173 +1202,162 @@ static inline int poll_pktin(uint32_t qi, int direct_recv,
 	return ret;
 }
 
-static inline int do_schedule_grp(odp_queue_t *out_queue, odp_event_t out_ev[], uint32_t max_num,
-				  int grp, int first_spr, int balance)
+static inline int schedule_grp_prio(odp_queue_t *out_queue, odp_event_t out_ev[], uint32_t max_num,
+				    int grp, int prio, int first_spr, int balance)
 {
-	int prio, spr, new_spr, i, ret;
+	int spr, new_spr, i, ret;
 	uint32_t qi;
-	uint16_t burst_def;
 	int num_spread = sched->config.num_spread;
 	uint32_t ring_mask = sched->ring_mask;
+	uint16_t burst_def = sched->config.burst_default[prio];
 
-	/* Schedule events */
-	for (prio = 0; prio < NUM_PRIO; prio++) {
-		if (sched->prio_q_mask[grp][prio] == 0)
+	/* Select the first spread based on weights */
+	spr = first_spr;
+
+	for (i = 0; i < num_spread;) {
+		int num;
+		uint8_t sync_ctx, ordered;
+		odp_queue_t handle;
+		ring_u32_t *ring;
+		int pktin;
+		uint16_t max_deq = burst_def;
+		int stashed = 1;
+		odp_event_t *ev_tbl = sched_local.stash.ev;
+
+		if (spr >= num_spread)
+			spr = 0;
+
+		/* No queues created for this priority queue */
+		if (odp_unlikely((sched->prio_q_mask[grp][prio] & (1 << spr)) == 0)) {
+			i++;
+			spr++;
 			continue;
-
-		burst_def = sched->config.burst_default[prio];
-
-		/* Select the first spread based on weights */
-		spr = first_spr;
-
-		for (i = 0; i < num_spread;) {
-			int num;
-			uint8_t sync_ctx, ordered;
-			odp_queue_t handle;
-			ring_u32_t *ring;
-			int pktin;
-			uint16_t max_deq = burst_def;
-			int stashed = 1;
-			odp_event_t *ev_tbl = sched_local.stash.ev;
-
-			if (spr >= num_spread)
-				spr = 0;
-
-			/* No queues created for this priority queue */
-			if (odp_unlikely((sched->prio_q_mask[grp][prio] & (1 << spr))
-			    == 0)) {
-				i++;
-				spr++;
-				continue;
-			}
-
-			/* Get queue index from the priority queue */
-			ring = &sched->prio_q[grp][prio][spr].ring;
-
-			if (ring_u32_deq(ring, ring_mask, &qi) == 0) {
-				/* Priority queue empty */
-				i++;
-				spr++;
-				continue;
-			}
-
-			sync_ctx = sched_sync_type(qi);
-			ordered  = (sync_ctx == ODP_SCHED_SYNC_ORDERED);
-
-			/* When application's array is larger than default burst
-			 * size, output all events directly there. Also, ordered
-			 * queues are not stashed locally to improve
-			 * parallelism. Ordered context can only be released
-			 * when the local cache is empty. */
-			if (max_num > burst_def || ordered) {
-				uint16_t burst_max;
-
-				burst_max = sched->config.burst_max[prio];
-				stashed = 0;
-				ev_tbl  = out_ev;
-				max_deq = max_num;
-				if (max_num > burst_max)
-					max_deq = burst_max;
-			}
-
-			pktin = queue_is_pktin(qi);
-
-			/* Update queue spread before dequeue. Dequeue changes status of an empty
-			 * queue, which enables a following enqueue operation to insert the queue
-			 * back into scheduling (with new spread). */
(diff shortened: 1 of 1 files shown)


platform/linux-generic/odp_schedule_basic.c
```diff
@@ -1423,21 +1458,29 @@ static inline int do_schedule(odp_queue_t *out_q, odp_event_t out_ev[], uint32_t
...
 		return 0;
 
+	my_groups = sched_local.grp_mask;
 	first_id = sched_local.grp_idx;
 	sched_local.grp_idx = (first_id + 1) % num_grp;
 
 	for (prio = 0; prio < NUM_PRIO; prio++) {
 		grp_id = first_id;
 
+		if ((sched->prio_grp_mask[prio] & my_groups) == 0) {
```
JannePeltonen, line 1470: This is actually a data race as schedule_creater_queue() may just be updating the mask through prio_grp_mask_set(). I suppose it is intentional, the idea being that this is just an optimization and it does not matter if in some scheduling rounds the mask seen is wrong. In that case it would be good to have a comment about it. Even better would be to use an atomic type (and a relaxed read) for the mask to get rid of the data race that strictly speaking leads to undefined behaviour.

There is the same issue with the existing prio_q_mask field.

platform/linux-generic/odp_schedule_basic.c
```diff
@@ -1430,15 +1430,18 @@ static inline int do_schedule(odp_queue_t *out_q, odp_event_t out_ev[], uint32_t
...
-	grp_id = sched_local.grp_weight[grp_round];
+	first_id = sched_local.grp_weight[grp_round];
 
-	/* Schedule queues per group and priority */
-	for (i = 0; i < num_grp; i++) {
-		int grp, prio;
+	for (prio = 0; prio < NUM_PRIO; prio++) {
+		grp_id = first_id;
+
+		for (i = 0; i < num_grp; i++) {
```
JannePeltonen, line 1475: Now that the order of iteration of priorities and groups has changed, I wonder if it would make sense to change the layout of the prio_q_mask and prio_q arrays to better match the order of access by making priority the outer index and group the inner index.

platform/linux-generic/odp_schedule_basic.c
```diff
@@ -1281,40 +1281,40 @@ static inline int schedule_grp_prio(odp_queue_t *out_queue, odp_event_t out_ev[]
...
 
 		if (odp_unlikely(num < 0)) {
-			/* Destroyed queue. Continue scheduling the same
-			 * priority queue. */
+			/* Remove destroyed queue from scheduling. Continue scheduling
+			 * the same group/prio/spread. */
+			continue;
+		}
+
+		if (num == 0 && !pktin) {
```
JannePeltonen, line 1314: This is not a strong opinion, but I think having the check for !pktin inside the following if statement was more readable. Now I have to keep more state in my head when reading the code.

platform/linux-generic/odp_schedule_basic.c
```diff
@@ -1430,7 +1420,11 @@ static inline int do_schedule(odp_queue_t *out_q, odp_event_t out_ev[], uint32_t
 		sched_local.grp_epoch = epoch;
 	}
 
-	first_id = sched_local.grp_weight[grp_round];
+	if (odp_unlikely(num_grp == 0))
+		return 0;
+
+	first_id = sched_local.grp_idx;
+	sched_local.grp_idx = (first_id + 1) % num_grp;
```
JannePeltonen, line 1465: > Both methods round-robin the first group index the same way.

Not quite, as the old code had some imbalance when the number of groups did not divide GRP_WEIGHT_TBL_SIZE evenly. So from that point-of-view this code seems better.

I suppose the rationale for the old code was to avoid the non-constant and not-power-of-two modulo operation that this code now has. So this change trades some cycles needed by the modulo opration to some memory footprint savings. Perhaps the commit message could be updated a bit.

psavol: Commit message updated in v2

psavol:
v2: rebase

platform/linux-generic/odp_schedule_basic.c
psavol: Agree, !pktin in side if (num == 0) is clear. I changed to that. It decreases 6% performance in one of my test cases in one server system, but increases a bit on another.

```if (num == 0) {
	int direct_recv = !ordered;
	int num_pkt;

	if (!pktin) {
		/* Remove empty queue from scheduling */
		continue;
	}

	/* Poll packet input queue */
	num_pkt = poll_pktin(qi, direct_recv, ev_tbl, max_deq);```

File: platform/linux-generic/odp_schedule_basic.c (hunk 2 above)
Comment (psavol) [reply platform/linux-generic/odp_schedule_basic.c:1475]: I think [group][prio] should be still more efficient. Typically, only a small number of groups are used or a thread is part of only some groups (e.g. 0, 1 and 13). When a group is active for a thread, scheduler needs to check all 8 priority levels of it. This way array packs per group information closer together in memory.

File: platform/linux-generic/odp_schedule_basic.c (hunk 1 above)
Comment (psavol) [reply platform/linux-generic/odp_schedule_basic.c:1470]: Added a comment and wrapped into a function in v3.

```
		if (prio_grp_mask_check(prio, my_groups) == 0) {
			/* My groups do not have queues at this priority level, continue to
			 * the next level.
			 *
			 * As a performance optimization, prio_grp_mask[] is checked without
			 * taking the lock. Masks change infrequently and usage of an old mask
			 * just leads into searching events from old priority levels,
			 * new levels are likely used on the next schedule call. */
			continue;
		}
```

Comment (JannePeltonen) [reply platform/linux-generic/odp_schedule_basic.c]: That does not fix the problem that it is still a data race (even if a harmless one in practise). Dynamic analyzers, like thread-sanitizer, may complain about it. I am not proposing using a lock but just changing the type to suitable atomic type (which gets rid of the data race).

File: platform/linux-generic/odp_schedule_basic.c (hunk 2 above)
Comment (JannePeltonen) [reply platform/linux-generic/odp_schedule_basic.c:1486]: Maybe so. Hard to say which is best. But changing the layout does not appear to be a small change so even if it made sense, it probably should be done separately.

File: platform/linux-generic/odp_schedule_basic.c (hunk 1 above)
Comment (psavol) [reply platform/linux-generic/odp_schedule_basic.c]: Changed prio_grp_mask[] typep to atomic in v4.

Comment (JannePeltonen) [reply platform/linux-generic/odp_schedule_basic.c]: The change looks good (it is ok to leave prio_q_mask as it is for now since ODP does not have 8-bit atomics). The commit message could be improved since now there is no data race in prio_grp_mask anymore (atomic variables do not create data races). Maybe rephrase like this:

Use a relaxed atomic load in schedule loop to remove a data race
and to document that reading prio_grp_mark when it might be written
is intentional.

File: platform/linux-generic/odp_schedule_basic.c [hunk 5]
```diff
@@ -1339,8 +1339,10 @@ static inline int schedule_grp_prio(odp_queue_t *out_queue, odp_event_t out_ev[]
 				 * packet input queue even when it is empty. */
 				ring_u32_enq(ring, ring_mask, qi);
 
-				/* Continue scheduling from the next group/priority */
-				return 0;
+				/* Continue scheduling from the next spread */
+				i++;
+				spr++;
```
JannePeltonen, line 1350: I am not convinved that this is the right thing to do. Not necessarily because it is not but maybe because I have hard time thinking about the dynamics at play and seeing whether this changes it for the better or worse in various situations. I guess I am trying to say that this is hard for me to review.

It seems to me that with ordered pktio input this piece of code selects the next spread even when there are packets available in the pktio (but they just have to fo through the internal queue first). I wonder if changing the behaviour in that particular case is good. More generally, in cases when packet I/Os are not getting more traffic that can be processed, this change causes the spread to change often even if there might be regular non-pktio queues that have packets available. This seems to reduce the possible benefit of having the spreads in the first place. Or maybe I just got confused.

JannePeltonen:
Other patches than "linux-gen: sched: next spread after no input packets" can be considered reviewed by me. I am not sure what to think of the spread patch.

JannePeltonen (approved):

PR #1550, opened 2022-04-11
[PATCH v2] test: sched_latency: add group options

psavol:
Added new -g and -i options which enable latency testing with multiple groups.

### DIFF
--- a/test/performance/odp_sched_latency.c
+++ b/test/performance/odp_sched_latency.c
@@ -563,6 +563,7 @@ static void parse_args(int argc, char *argv[], test_args_t *args)
 
 	static const struct option longopts[] = {
 		{"count", required_argument, NULL, 'c'},
+		{"duration", required_argument, NULL, 'd'},
 		{"forward-mode", required_argument, NULL, 'f'},
 		{"lo-prio-queues", required_argument, NULL, 'l'},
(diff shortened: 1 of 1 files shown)


test/performance/odp_sched_latency.c
```diff
@@ -764,24 +896,27 @@ int main(int argc, char *argv[])
...
+		exit(EXIT_FAILURE);
 	}
 
-	shm = odp_shm_reserve("test_globals",
-			      sizeof(test_globals_t), ODP_CACHE_LINE_SIZE, 0);
+	shm = odp_shm_reserve("test_globals", sizeof(test_globals_t), ODP_CACHE_LINE_SIZE, 0);
 	if (shm == ODP_SHM_INVALID) {
 		ODPH_ERR("Shared memory reserve failed.\n");
-		return -1;
+		err = -1;
```
MatiasElo, line 918: `thread_tbl` is not freed in case of failures. Probably no need to allocate it from heap in the first place.

MatiasElo (approved):

test/performance/odp_sched_latency.c
psavol: Fixed in v2. Change to allocate thread_tbl from stack.

PR #1553, opened 2022-04-13
[PATCH v2] example/generator: fix pool creation for bigger payload length

ksatha:
Generator creates packet pool with default length 1856, this will
create problem when we request bigger payload on platforms with
limited segments.

This patch addresses issue by considering pool parameters and
payload requested while creating packet pool instead of using packet
length macro.

Signed-off-by: Satha Rao <skoteshwar@marvell.com>

### DIFF
--- a/example/generator/odp_generator.c
+++ b/example/generator/odp_generator.c
@@ -24,13 +24,15 @@
 /* Max number of workers */
 #define MAX_WORKERS            (ODP_THREAD_COUNT_MAX - 1)
 #define POOL_NUM_PKT           2048  /* Number of packets in packet pool */
-#define POOL_PKT_LEN           1856  /* Max packet length */
+#define POOL_PKT_LEN           1856  /* Default packet pool length */
 #define DEFAULT_PKT_INTERVAL   1000  /* Interval between each packet */
 #define DEFAULT_UDP_TX_BURST	16
 #define MAX_UDP_TX_BURST	512
 #define DEFAULT_RX_BURST	32
 #define MAX_RX_BURST		512
 #define STATS_INTERVAL		10   /* Interval between stats prints (sec) */
+/* L2 + L3 + L4 header length */
+#define MAX_HDR_LEN (ODPH_ETHHDR_LEN + ODPH_IPV4HDR_LEN + MAX(ODPH_UDPHDR_LEN, ODPH_ICMPHDR_LEN))
 
 #define APPL_MODE_UDP    0			/**< UDP mode */
 #define APPL_MODE_PING   1			/**< ping mode */
@@ -1115,6 +1117,7 @@ int main(int argc, char *argv[])
 	odp_init_t init_param;
 	odph_thread_common_param_t thr_common;
 	odph_thread_param_t thr_param;
+	uint32_t pkt_len;
 
 	/* Signal handler has to be registered before global init in case ODP
 	 * implementation creates internal threads/processes. */
@@ -1210,10 +1213,16 @@ int main(int argc, char *argv[])
 	/* Configure scheduler */
 	odp_schedule_config(NULL);
 
+	pkt_len = args->appl.payload + MAX_HDR_LEN;
+	if (pkt_len < POOL_PKT_LEN)
+		pkt_len = POOL_PKT_LEN;
+
+	printf("Packet length:      %u\n", pkt_len);
+
 	/* Create packet pool */
 	odp_pool_param_init(&params);
 	params.pkt.seg_len = POOL_PKT_LEN;
-	params.pkt.len     = POOL_PKT_LEN;
+	params.pkt.len     = pkt_len;
 	params.pkt.num     = POOL_NUM_PKT;
 	params.type        = ODP_POOL_PACKET;
 


example/generator/odp_generator.c
```diff
@@ -25,6 +25,8 @@
 #define MAX_WORKERS            (ODP_THREAD_COUNT_MAX - 1)
 #define POOL_NUM_PKT           2048  /* Number of packets in packet pool */
 #define POOL_PKT_LEN           1856  /* Max packet length */
+#define MAX_HDR_LEN            128   /* L2/L3/L4 header length */
```
JannePeltonen, line 28: The 128 is a magic number because is cannot be freely selected but depends on other things. It seems it should be
(ODPH_ETHHDR_LEN + ODPH_IPV4HDR_LEN + MAX(ODPH_UDPHDR_LEN, ODPH_ICMPHDR_LEN))

Also, the comment can be interpreted as "L2 or L3 or L3 header length". Replacing the slash with e.g. a plus sign would make it better.

example/generator/odp_generator.c
```diff
@@ -25,6 +25,8 @@
...
+#define POOL_SEG_LEN           1856  /* Segment length to pool create */
```
JannePeltonen, line 29: Why 1856? It seems MAX_HDR_LEN would be sufficient.

example/generator/odp_generator.c
```diff
@@ -25,6 +25,8 @@
 #define MAX_WORKERS            (ODP_THREAD_COUNT_MAX - 1)
 #define POOL_NUM_PKT           2048  /* Number of packets in packet pool */
 #define POOL_PKT_LEN           1856  /* Max packet length */
```
JannePeltonen, line 27: I do not see why POOL_PKT_LEN is still needed. Now it acts as minimum packet length for pool creation (not maximum as said in its comment) in case the length derived from command line parameters is very small. But
I do not see why there is a need to have such minimum instead of simply always deriving the packet length parameter from the command line parameters.

example/generator/odp_generator.c
```diff
@@ -1210,10 +1214,35 @@ int main(int argc, char *argv[])
...
+	pkt_len = (args->appl.payload + MAX_HDR_LEN) > POOL_PKT_LEN ?
+			(args->appl.payload + MAX_HDR_LEN) : POOL_PKT_LEN;
+
+	if (pool_capa.pkt.max_len && pkt_len > pool_capa.pkt.max_len)
+		pkt_len = pool_capa.pkt.max_len;
+
+	/* Check weather we have sufficient segments to support requested packet
+	 * length, if not adjust to bigger segment size */
+	if ((seg_len * pool_capa.pkt.max_segs_per_pkt) < pkt_len)
+		seg_len = pkt_len / pool_capa.pkt.max_segs_per_pkt;
```
JannePeltonen, line 1231: This is unnecessary and should not be needed by any ODP implementation that complies with the API specification. In the pool creation parameters the seg_len does not specify the exact segment length that must be used but "Minimum number of packet data bytes that can be stored in the first segment of a newly allocated packet". IOW, an implementation should not get confused and not able to create a packet pool if an application does not require may contiguous bytes in the beginning of packet data.

example/generator/odp_generator.c
```diff
@@ -1210,10 +1214,35 @@ int main(int argc, char *argv[])
...
+
+	if (pool_capa.pkt.min_seg_len && seg_len < pool_capa.pkt.min_seg_len)
+		seg_len = pool_capa.pkt.min_seg_len;
```
JannePeltonen, line 1234: This is not needed. min_seg_len is not the minimum allowed value for the seg_len parameter but just indicates that the segments will not be shorter than that even if an application does not request that long segments.

example/generator/odp_generator.c
```diff
@@ -1210,10 +1214,35 @@ int main(int argc, char *argv[])
...
+
+	if (pool_capa.pkt.max_seg_len && seg_len > pool_capa.pkt.max_seg_len)
+		seg_len = pool_capa.pkt.max_seg_len;
```
JannePeltonen, line 1237: This is wrong. This application makes the assumption that certain amount of packet data is contiguous in memory. If the assumption is not valid, the application overwrites memory which it does not own and which may contain e.g. internal packet metadata of the underlying ODP implementation. If the underlying ODP does not support long enough contiguous data (big enough max_seg_len), one cannot just ignore the assumption. In that case the application should just exit as it cannot create the kind of pool that it needs.

JannePeltonen (changes requested):

example/generator/odp_generator.c
ksatha: ACK

example/generator/odp_generator.c
ksatha: Yes it will works with MAX_HDR_LEN, 1856 (i.e., POOL_PKT_LEN) will avoids multi segments for regular MTU and gives better performance.

example/generator/odp_generator.c
ksatha: Will change comment to "Default Pool Buffer Length", and pool creates with this value when either payload not passed or requested < 1500. This enables generator similar to before change, also no issues with receiver mode for regular MTU.

example/generator/odp_generator.c
ksatha: Will remove this code

example/generator/odp_generator.c
ksatha: Will remove this code

example/generator/odp_generator.c
ksatha: Will remove this code

ksatha:
@JannePeltonen any suggestions on v2

example/generator/odp_generator.c
```diff
@@ -1210,10 +1213,16 @@ int main(int argc, char *argv[])
...
+	if (pkt_len < POOL_PKT_LEN)
+		pkt_len = POOL_PKT_LEN;
+
+	printf("Packet length:      %u\n", pkt_len);
+
 	/* Create packet pool */
 	odp_pool_param_init(&params);
 	params.pkt.seg_len = POOL_PKT_LEN;
-	params.pkt.len     = POOL_PKT_LEN;
+	params.pkt.len     = pkt_len;
```
JannePeltonen, line 1225: Now packet len can exceed segment length, but setup_icmp_pkt() assumes that the whole packet data is contiguous in memory.

ksatha: setup_icmp_pkt always sends with payload of 56 bytes, so it will not an issue. Please correct me if I am wrong.

ksatha: @JannePeltonen any comments

JannePeltonen: If someone changes setup_icmp_pkt_ref() to send longer ICMP packets and/or changes the segment size (POOL_PKT_LEN) to small enough, then setup_icmp_pkt() will overwrite the packet data buffer. This rather fragile since the code makes an assumption on what other pieces of code far way are doing. It would be better to check that the segment is large enough (e.g. by calling odp_packet_data_seg_len() instead of just odp_packet_data() and then checking the segment size against the requirement (headers + payload). Or by using a suitable static assert after defining the ICMP payload length as a macro).

Also, I do not think POOL_PKT_LEN is a good name anymore. I think it probably should be called POO_SEG_LEN or something similar (and the comment changed accordingly).

It would also be good to check with ODP_STATIC_ASSERT that SEG_LEN >= MAX_HDR_LEN as that is another assumption made in the code.

JannePeltonen:

PR #1555, opened 2022-04-19
[PATCH v3] remove crypto error from packet error flags

JannePeltonen:
validation: crypto: remove odp_packet_has_error() check after crypto error

    ODP API does not say that odp_packet_has_error() must return true after
    crypto operation errors. odp_packet_has_error() indicates various packet
    parsing errors and crypto operations are not parsing the packet.

    Remove the odp_packet_has_error() check to prevent potential test failures
    with API compliant ODP implementations.

linux_gen: remove crypto error from packet error flags

    Remove the unnecessary crypto_err parse flag from packet header.
    odp_packet_has_error() is supposed to indicate parse errors not crypto
    processing errors, which are indicated through odp_crypto_result_t.

    Remove the flag and its setting in crypto implementations as the flag
    is not needed and may cause subtle API spec violations.

### DIFF
--- a/platform/linux-generic/arch/aarch64/odp_crypto_armv8.c
+++ b/platform/linux-generic/arch/aarch64/odp_crypto_armv8.c
@@ -724,7 +724,6 @@ odp_crypto_operation(odp_crypto_op_param_t *param,
 		 * We cannot fail since odp_crypto_op() has already processed
 		 * the packet. Let's indicate error in the result instead.
 		 */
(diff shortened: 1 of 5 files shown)


psavol (approved):

JannePeltonen:
v2:
- Removed leftover unused variable from the second patch to fix ARM compilation
- Added a third patch to get rid of the now-unused return value of the ARM crypto functions.
- Added reviewed-by tags to the two first patches

PR #1558, opened 2022-04-25
api: schedule: introduce odp_schedule_order_wait()

JannePeltonen:
api: schedule: introduce odp_schedule_order_wait()

    Introduce odp_schedule_order_wait() what waits until the calling thread
    is "at the head of the source queue", i.e. until no other thread (nor
    the source queues nor the scheduler) holds an ordered scheduling context
    that has the same source queue and precedes the ordered scheduling context
    of the calling thread.

linux-gen: schedule: implement odp_schedule_order_wait()

    Implement odp_schedule_order_wait() by using the existing order_lock()
    (not to be confused with ordered locks) function that does what we need
    and does not require unlock. Support is added for the basic and scalable
    scheduler but not for the SP scheduler where the function is a nop.


jerinjacobk (approved):

JannePeltonen:
v3: Added validation test

JannePeltonen:
v4: Replaced the validation test with a simple single-threaded version. Test with multiple threads needs some tweaking so I put it in a separate PR later.

psavol (approved):

PR #1571, opened 2022-05-10
[PATCH v5] validation: crypto: fix check for session create error

asasidharan:
Crypto session create API returns a negative value on failure.
To check for the error code for failure, status parameter need to
be explicitly checked not the return value of the API.

### DIFF
--- a/test/validation/api/crypto/odp_crypto_test_inp.c
+++ b/test/validation/api/crypto/odp_crypto_test_inp.c
@@ -689,7 +689,8 @@ static odp_crypto_session_t session_create(odp_crypto_op_t op,
 	 * In some cases an individual algorithm cannot be used alone,
 	 * i.e. with the null cipher/auth algorithm.
 	 */
-	if (rc == ODP_CRYPTO_SES_ERR_ALG_COMBO) {
+	if (rc < 0 &&
+	    status == ODP_CRYPTO_SES_ERR_ALG_COMBO) {
 		printf("\n    Unsupported algorithm combination: %s, %s\n",
 		       cipher_alg_name(cipher_alg),
 		       auth_alg_name(auth_alg));
--- a/test/validation/api/crypto/odp_crypto_test_inp.c
+++ b/test/validation/api/crypto/odp_crypto_test_inp.c
@@ -1178,11 +1178,11 @@ static void test_capability(void)
  * operation with hash_result_offset outside the auth_range and by
  * copying the hash in the ciphertext packet.
  */
-static void create_hash_test_reference(odp_auth_alg_t auth,
-				       const odp_crypto_auth_capability_t *capa,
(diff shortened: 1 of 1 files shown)


test/validation/api/crypto/odp_crypto_test_inp.c
```diff
@@ -1219,12 +1220,14 @@ static void create_hash_test_reference(odp_auth_alg_t auth,
...
 				 auth, ref, PACKET_IV, HASH_NO_OVERLAP);
+	if (session == ODP_CRYPTO_SESSION_INVALID)
+		return -1;
 
 	if (crypto_op(pkt, &ok, session,
 		      ref->cipher_iv, ref->auth_iv,
 		      &cipher_range, &auth_range,
 		      ref->aad, enc_digest_offset))
-		return;
+		return 0;
```
JannePeltonen, line 1230: This change is basically correct for the change being done, but I think it would be better to CU_ASSERT that the return value is zero and then continue the function (then the session would be freed too). I think the code should have been made that way in the first place but now that you are touching it anyway, you could fix it.

test/validation/api/crypto/odp_crypto_test_inp.c
```diff
@@ -689,7 +689,8 @@ static odp_crypto_session_t session_create(odp_crypto_op_t op,
 	 * In some cases an individual algorithm cannot be used alone,
 	 * i.e. with the null cipher/auth algorithm.
 	 */
```
JannePeltonen, line 691: There are two fixes in this PR (they could have been in two different patches as well).

The first fix is the way ODP_CRYPTO_SES_ERR_ALG_COMBO is checked. That does not fix b7d520d but c14955e1d (validation: crypto: do not test unsupported algorithm combos).

The second fix fixes the hash result overlap test in case an auth algorithm is not supported with the null cipher algorithm. That fixes b7d520d, but the commit message does not otherwise mention the fix.

The commit messsage should be updated to mention both issues in the description, or better yet, you could split the patch to two. Also, there should be a fixes-line for each of the original patches or none at all. The CC-line could be removed.

JannePeltonen (changes requested):

test/validation/api/crypto/odp_crypto_test_inp.c
asasidharan: Ack. Created separate commit for the second fix and the suggested change is incorporated in it.

test/validation/api/crypto/odp_crypto_test_inp.c
asasidharan: Ack. Separated the two fixes into two different commits with v3.

JannePeltonen (approved):

asasidharan:
v5: Added Reviewed-by tag.

PR #1572, opened 2022-05-10
validation: scheduler: test odp_schedule_order_wait()

JannePeltonen:
Add a test of odp_schedule_order_wait(). Have two threads schedule
an event from the same queue. Delay the thread handling the first event
and check that odp_schedule_order_wait() stops the second thread until
the first thread has released the scheduling context.

Remove reference to the order wait capability from the single-threaded
test as it is referenced by the added test. It was there just to make
sure the field exists.


test/validation/api/scheduler/scheduler.c
```diff
@@ -2348,6 +2347,130 @@ static void scheduler_test_order_wait_1_thread(void)
...
+
+static int check_2_workers(void)
+{
+	if (globals->num_workers < 2) {
+		printf("\nTest: scheduler_test_order_wait: SKIPPED\n");
+		return ODP_TEST_INACTIVE;
+	}
+
+	return ODP_TEST_ACTIVE;
+}
```
TuomasTaipale, line 2472: Is there a need to run `scheduler_test_order_wait()` if `sched_capa.order_wait` is not set?

test/validation/api/scheduler/scheduler.c
```diff
@@ -2348,6 +2347,130 @@ static void scheduler_test_order_wait_1_thread(void)
...
+	while ((ev = odp_schedule(NULL, ODP_SCHED_NO_WAIT))
+	       != ODP_EVENT_INVALID) {
+		/* We got an event that was meant for the main thread */
+		odp_event_free(ev);
+	}
+
+	return 0;
+}
+
+static void scheduler_test_order_wait(void)
```
TuomasTaipale, line 2381: This could be "scheduler_test_order_wait_2_thread".

test/validation/api/scheduler/scheduler.c
JannePeltonen: Perhaps not strictly necessary, but the API says that the function does nothing in that case (i.e. it can still be called even though it is not very useful), so the test calls it even if the capa is not set.

TuomasTaipale: OK. Angle was that as it's already called in `scheduler_test_order_wait_1_thread()` unconditionally, the new test could be just skipped if there's no capa.

JannePeltonen: I think doing the call in the new test too even when the capa is not there does not harm and might even expose a problem since the new test has two threads running. Plus then the capability function gets called too. So I think I will keep it.

test/validation/api/scheduler/scheduler.c
JannePeltonen: I can change that.

test/validation/api/scheduler/scheduler.c
```diff
@@ -2348,6 +2347,130 @@ static void scheduler_test_order_wait_1_thread(void)
...
+
+	ret = odp_cunit_thread_create(order_wait_helper, &thr_arg);
+	CU_ASSERT_FATAL(ret == thr_arg.numthrds);
+
+	/* Send an event to the helper thread */
+	enqueue_event(queue);
+
+	/* Wait that the helper thread gets the event */
+	while (!odp_atomic_load_acq_u32(&globals->order_wait.helper_ready))
+		;
```
psavol, line 2410: It would be good to have some timeout (e.g. 10 sec) instead of infinitely waiting. Issues are easier to debug in CI when system does not get stuck, but prints some error or asserts.

test/validation/api/scheduler/scheduler.c
```diff
@@ -2348,6 +2347,130 @@ static void scheduler_test_order_wait_1_thread(void)
 	CU_ASSERT(odp_queue_destroy(queue) == 0);
 }
 
+static int order_wait_helper(void *arg ODP_UNUSED)
+{
+	odp_event_t ev;
+
+	ev = odp_schedule(NULL, ODP_SCHED_WAIT);
+	if (ev == ODP_EVENT_INVALID) {
```
psavol, line 2355: Some wait time would be needed (e.g. 10 sec) to avoid thread getting stuck (in CI) when something goes wrong. ODP_SCHED_WAIT option should not ever return EVENT_INVALID, but wait until there is an event.

test/validation/api/scheduler/scheduler.c
JannePeltonen: Added a one second timeout in v3

psavol (approved):

TuomasTaipale (approved):

PR #1573, opened 2022-05-13
[PATCH v3] api: stash: change get/put function parameter names

MatiasElo:


platform/linux-generic/odp_stash.c
```diff
@@ -371,11 +371,11 @@ int32_t odp_stash_put_u32(odp_stash_t st, const uint32_t u32[], int32_t num)
 	ODP_ASSERT(stash->obj_size == sizeof(uint32_t));
 
 	ring_u32_enq_multi(&stash->ring_u32.hdr, stash->ring_mask,
-			   (uint32_t *)(uintptr_t)u32, num);
+			   (uint32_t *)(uintptr_t)val, num);
```
JannePeltonen, line 374: I think it would be good to declare the data parameter of _RING_ENQ and _RING_ENQ_MULTI as const. Then this code would not need the ugly casts.

damzielinski:
Looks good!

platform/linux-generic/odp_stash.c
MatiasElo: Makes sense. However, this change will require quite a few fixes around the implementation (https://github.com/MatiasElo/odp/commit/be21582a02d7a5f91551a1e6f0a7debb702e4003 and still some build failures remaining), so I'd rather do it in a separate PR.

JannePeltonen: Actually, maybe using const in enqueue is not right since dequeue is not const (and thus the casting to non-const just moves deeper). Now I wonder it the stash API should use non-const parameters in enqueue. But this is a separate topic from this PR.

JannePeltonen (approved):

psavol (approved):

PR #1576, opened 2022-05-18
[PATCH v3] linux-gen: inline spinlock implementations 

MatiasElo:


platform/linux-generic/include/odp/api/plat/spinlock_recursive_inlines.h
```diff
@@ -0,0 +1,93 @@
...
+	if (rlock->owner == thr) {
+		rlock->cnt++;
+		return 1;
+	}
+
+	if (odp_spinlock_trylock(&rlock->lock)) {
+		rlock->owner = thr;
+		rlock->cnt   = 1;
+		return 1;
+	} else {
```
TuomasTaipale, line 65: `else` is redundant.

platform/linux-generic/include/odp/api/plat/spinlock_recursive_inlines.h
```diff
@@ -0,0 +1,93 @@
...
+
+_ODP_INLINE int odp_spinlock_recursive_is_locked(odp_spinlock_recursive_t *rlock)
+{
+	int thr = odp_thread_id();
+
+	if (rlock->owner == thr)
+		return 1;
+
+	return odp_spinlock_is_locked(&rlock->lock);
+}
```
TuomasTaipale, line 89:
```C
return odp_thread_id() == rlock->owner ? 1 : odp_spinlock_is_locked(&rlock->lock);
```

platform/linux-generic/include/odp/api/plat/spinlock_recursive_inlines.h
```diff
@@ -0,0 +1,93 @@
...
+		rlock->cnt   = 1;
+		return 1;
+	} else {
+		return 0;
+	}
+}
+
+_ODP_INLINE void odp_spinlock_recursive_unlock(odp_spinlock_recursive_t *rlock)
+{
+	rlock->cnt--;
```
TuomasTaipale, line 72: Need protection for accidental wraparound?

platform/linux-generic/include/odp/api/plat/spinlock_inlines.h
```diff
@@ -0,0 +1,63 @@
+/* Copyright (c) 2013-2018, Linaro Limited
+ * Copyright (c) 2022, Nokia
+ * All rights reserved.
+ *
+ * SPDX-License-Identifier: BSD-3-Clause
+ */
+
+#ifndef _ODP_PLAT_SPINLOCK_INLINES_H_
```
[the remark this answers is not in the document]
MatiasElo: Fixed in V2.

platform/linux-generic/include/odp/api/plat/spinlock_recursive_inlines.h
MatiasElo: Fixed in V2.

platform/linux-generic/include/odp/api/plat/spinlock_recursive_inlines.h
MatiasElo: Looks like it does (this commit didn't modify the implementation). `odp_spinlock_unlock()` is called only when `rlock->cnt == 0` and calling `odp_spinlock_recursive_unlock()` more times than the lock has been acquired would be against the API spec anyway. The next lock call will reset count to a valid value.

MatiasElo:
V2:
- Fixed review comments
- Added includes for inline headers

platform/linux-generic/include/odp/api/plat/spinlock_recursive_inlines.h
TuomasTaipale: Assert for count not being already zero could be helpful in catching any bugs instead of program just getting stuck.

MatiasElo: The problem with asserts is that we cannot currently use them in inline headers as they are not part of the public API. We should probably do something about this limitation in a separate PR.

TuomasTaipale: OK, resolving this.

TuomasTaipale (approved):

PR #1578, opened 2022-05-20
[PATCH v4] linux-gen: event: inline small functions

MatiasElo:


platform/linux-generic/include/odp/api/plat/event_inlines.h
```diff
@@ -54,6 +56,11 @@ _ODP_INLINE int odp_event_type_multi(const odp_event_t event[], int num,
 	return i;
 }
 
+_ODP_INLINE uint32_t odp_event_flow_id(odp_event_t event)
+{
+	return _odp_event_hdr_field(event, int8_t, flow_id);
```
TuomasTaipale, line 87: `flow_id` seems to be of type `uint8_t`.

MatiasElo: Fixed in V2.

TuomasTaipale (approved):

PR #1580, opened 2022-05-23
[PATCH v3] linux-gen: disable TM and most of openssl crypto in process mode

JannePeltonen:
TM and openssl crypto are not compatible with process mode, so disable TM and libssl based crypto algorithms in process mode.

### DIFF
--- a/platform/linux-generic/odp_crypto_openssl.c
+++ b/platform/linux-generic/odp_crypto_openssl.c
@@ -1957,10 +1957,18 @@ int odp_crypto_capability(odp_crypto_capability_t *capa)
 	/* Initialize crypto capability structure */
 	memset(capa, 0, sizeof(odp_crypto_capability_t));
 
+	capa->max_sessions = MAX_SESSIONS;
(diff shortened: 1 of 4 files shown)


JannePeltonen:
v2: Added a patch to disable ipsec_crypto and ipsec_api examples in CI in process mode. Fixing the test scripts and/or example apps to deal with unsupported algorithms would have been much more work, especially since ipsec_api does not even check the ipsec capabilities before trying to create ODP SAs.

JereLeppanen (approved):
One could avoid a commit that fails make check by switching the order of the first two commits. But this order is probably more natural. Approved all the same.

MatiasElo (approved):

JannePeltonen:
> One could avoid a commit that fails make check by switching the order of the first two commits

There must be some confusion on the order because the patches were specifically ordered so that also the intermediate revisions pass make check, i.e. "skip ipsec examples in process mode" becomes before "fix openssl crypto in process mode by disabling algos". So I am keeping the order.

PR #1584, opened 2022-05-25
[PATCH v3] validation: test default value setting by init functions better

JannePeltonen:
validation: improve testing of default values set by init functions

    Improve testing of the default values set by various init functions by
    checking the defaults twice: once after memsetting the to-be-initialized
    structure to zero before the init call and once after memsetting to 0xff.
    This way missing initialization can always be detected.

    validation: init: test default value setting by odp_init_param_init()

    Test that odp_init_param_init() initializes parameters to the specified
    default values.

### DIFF
--- a/test/validation/api/classification/odp_classification_basic.c
+++ b/test/validation/api/classification/odp_classification_basic.c
@@ -1,4 +1,5 @@
 /* Copyright (c) 2015-2018, Linaro Limited
+ * Copyright (c) 2021-2022, Nokia
  * All rights reserved.
  *
  * SPDX-License-Identifier:	BSD-3-Clause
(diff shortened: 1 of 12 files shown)


test/validation/api/classification/odp_classification_basic.c
```diff
@@ -10,23 +11,29 @@
 
 #define PMR_SET_NUM	5
 
-static void classification_test_default_values(void)
+static void test_defaults(uint8_t fill_byte)
```
TuomasTaipale, line 14: Type is already a byte so name could be just `fill`.

JannePeltonen: Renamed in v2.

TuomasTaipale (approved):

PR #1587, opened 2022-05-31
[PATCH v1] test: l2fwd: fix pool per interface option

harman-kalra:
odp-l2fwd application when start with -y option viz exclusive packet
and vector pool per interface, resulted in crash as same name can't be
used to create every pool. Unique name for each pool resolves the issue.


MatiasElo:
According to `odp_pool_create()` API specification pool names don't have to be unique. Implementation should take care of this internally if unique names are required for some reason.

psavol:
Closing this PR as l2fwd uses API correctly.

PR #1588, opened 2022-05-31
Flow hash

harman-kalra:


include/odp/api/spec/packet_io_types.h
```diff
@@ -917,6 +917,14 @@ typedef struct odp_pktio_capability_t {
...
 	odp_pktio_stats_capability_t stats;
 
+	/** Maximum flow hash value.
+	 *
+	 * This defines the maximum flow hash value that an application can set.
+	 * Attempts to set flow hash value greater than this limit will fail.
+	 *
+	 * @see odp_packet_flow_hash_set()
+	 */
+	uint32_t max_flow_hash;
```
JannePeltonen, line 927: I am not sure how useful this is. In fact, the whole odp_packet_flow_hash_set() does not seem useful to me. I have hard time thinking about a case where an application would want to use it, as opposed to e.g. storing a few bytes in user area, and even more so if the function is limited by this new capability.

This makes me wonder if, instead of adding a new capability or otherwise refining the API, the whole function should be removed from the API.

PR #1597, opened 2022-06-09
[PATCH v3] test: cunit: fix cunit asserts for multithreaded test apps

JannePeltonen:
CUnit is not thread-safe and not aware of process mode, which causes
    problems when CUnit assertions are used in multithreaded tests:
    - Updating the recorded assert failures has data races
    - Assertion failures in other than the main process are lost
    - The longjmp performed at fatal assertion failure tries to jump to
      a stack frame of another thread, which is doomed to fail badly

    Fix the problems by wrapping the commonly used CUnit assertions with
    ODP macros and an ODP function that calls CUnit only if running in the
    main thread and the main process. In other threads/processes record
    failed assertions in shared memory and re-play them to CUnit within
    the main thread/process when becoming single threaded again in
    odp_cunit_thread_exit().

    This does not solve the general issue that careless use of fatal
    assertions in the middle of tests may leave the test app in a state
    from which it cannot recover.

### DIFF
--- a/test/common/odp_cunit_common.c
+++ b/test/common/odp_cunit_common.c
@@ -29,6 +29,7 @@
 /* Globals */
 static int allow_skip_result;
 static odph_thread_t thread_tbl[MAX_WORKERS];
+static int threads_running;
 static odp_instance_t instance;
 static char *progname;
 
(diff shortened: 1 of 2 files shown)


test/common/odp_cunit_common.c
```diff
@@ -46,6 +50,136 @@ static struct {
...
+		 * Turn fatal failures into non-fatal failures as we are just
+		 * reporting them. Threads that saw fatal failures which
+		 * prevented them from continuing have already been stopped.
+		 */
+		CU_assertImplementation(0, a->line, a->cond, a->file, "", CU_FALSE);
+	}
+	thr_global->num_failures = 0;
+}
+
+static int odp_cunit_thread_init(void)
```
psavol, line 139: ODP prefix should not be used. Static function does not need prefix. Maybe just init_thread(), similar to run_thread().

psavol:
Commit messages need update during rebase, as thread_exit() was renamed to thread_join().

JannePeltonen:
v2: rebased, dropped the first patch that is no longer relevant and renamed the init function.

psavol (approved):

PR #1605, opened 2022-06-20
[PATCH v2] linux-gen: atomic: fix build error with gcc 12

JannePeltonen:
Pointer aliasing caused by pointer casting in _odp_atomic_load_u128() seems
to confuse GCC 12.1 so that GCC issues a warning about a use of an
uninitialized variable. The generated code was anyway ok in the compiler
versions I tested.

Replace pointer casting in the problematic code by type punning through
a union, which appears to have been valid C since C99. The generated code
is not affected in various GCC and Clang versions tested.


TuomasTaipale (approved):
This line in commit message: "_... The generated code was anyway ok in the compiler
versions I tested._", perhaps the compiler versions could be listed (or alternatively just drop the sentence).

JannePeltonen:
v2: Removed the vague sentences about compiler versions in the commit message and added reviewed-by tag.

PR #1607, opened 2022-06-23
[PATCH v2] validation: sched: test queue FIFO order

psavol:


MatiasElo (approved):
Looks good. Copyright could be added/updated to couple files (e.g. odp_cunit_common.c, barrier.c).

PR #1612, opened 2022-07-01
[PATCH v7] api: pool: add thread local cache usage statistic counters

MatiasElo:


platform/linux-generic/odp_pool.c
```diff
@@ -1556,7 +1582,9 @@ int odp_pool_stats(odp_pool_t pool_hdl, odp_pool_stats_t *stats)
 
 	pool = _odp_pool_entry(pool_hdl);
 
-	memset(stats, 0, sizeof(odp_pool_stats_t));
+	/* Don't overwrite input parameters */
+	memset(stats, 0, offsetof(odp_pool_stats_t, thread));
+	memset(stats->thread.cache_available, 0, sizeof(stats->thread.cache_available));
```
psavol, line 1587: It would be more robust to save first/last, memset(0) and then restore first/last. If new counter is added after thread struct, it is easy to forget to add another memset line here.

MatiasElo: Fixed in V2.

include/odp/api/spec/pool.h
```diff
@@ -160,8 +160,14 @@ int odp_pool_index(odp_pool_t pool);
  * Read the statistics counters enabled using odp_pool_stats_opt_t during pool
  * creation. The inactive counters are set to zero by the implementation.
  *
- * @param      pool   Pool handle
- * @param[out] stats  Output buffer for counters
+ * A single call may read statistics from one to ODP_POOL_MAX_THREAD_STATS
+ * threads. Set 'stats.thread.first' and 'stats.thread.last' to select the
+ * threads. Valid values range from 0 to odp_thread_count_max() - 1. A
```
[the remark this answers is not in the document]
MatiasElo: Fixed in V2.

MatiasElo:
V2/V3:
- Fixed review comments
- Improved validation test

TuomasTaipale (approved):

platform/linux-generic/odp_pool.c
```diff
@@ -130,15 +130,40 @@ static void cache_flush(pool_cache_t *cache, pool_t *pool)
...
+		}
 
-	for (i = 0; i < ODP_THREAD_COUNT_MAX; i++)
-		cached += pool->local_cache[i].cache_num;
+		if (last - first + 1 > ODP_POOL_MAX_THREAD_STATS) {
+			ODP_ERR("Too many thread ids: max=%d\n", ODP_POOL_MAX_THREAD_STATS);
+			return -1;
+		}
+	}
+
```
ashwinyes, line 153: Write Memory Barrier required before reading local cache values from other threads.

ashwinyes:
It would also help to put a note in the spec stating that there should be no pool operations issued by other threads while the pool stats APIs are in progress as it can result in wrong values.

test/validation/api/pool/pool.c
```diff
@@ -1178,10 +1185,15 @@ static void pool_test_pool_statistics(odp_pool_type_t pool_type)
 		CU_ASSERT(stats.cache_available <= num_obj);
 		CU_ASSERT(stats.cache_alloc_ops == 0);
 		CU_ASSERT(stats.cache_free_ops == 0);
+		CU_ASSERT(stats.thread.first == first);
+		CU_ASSERT(stats.thread.last == last);
+		for (j = 0; j < ODP_POOL_MAX_THREAD_STATS; j++)
+			CU_ASSERT(stats.thread.cache_available[j] <= num_obj);
```
ashwinyes, line 1191: CU_ASSERT(stats.thread.cache_available[j] <= stats.cache_available); would be more specific.

test/validation/api/pool/pool.c
```diff
@@ -1210,19 +1222,34 @@ static void pool_test_pool_statistics(odp_pool_type_t pool_type)
...
 			CU_ASSERT_FATAL(odp_pool_stats(pool[i], &stats) == 0);
 			CU_ASSERT(stats.available <= num_obj - num_events);
 			CU_ASSERT(stats.cache_available <= num_obj - num_events);
+
+			for (int k = 0; k < ODP_POOL_MAX_THREAD_STATS; k++) {
+				CU_ASSERT(stats.thread.cache_available[k] <= num_obj - num_events);
+				total_cached += stats.thread.cache_available[k];
+			}
+			if (supported.bit.cache_available && supported.bit.thread_cache_available)
+				CU_ASSERT(stats.cache_available == total_cached);
```
ashwinyes, line 1236: This will not hold if odp_thread_count_max() > ODP_POOL_MAX_THREAD_STATS

test/validation/api/pool/pool.c
MatiasElo: Fixed in V5.

test/validation/api/pool/pool.c
MatiasElo: Yes, but as this test is single threaded and thread ids are allocated starting from 0 this should be OK.

JannePeltonen: The API does not say in which order thread IDs are assigned, so one cannot portably assume that the thread_id of the single thread of a single threaded app is zero or any other particular value.

I also think it is incorrect to assume that the number of running threads is relevant here. Thread local caches may get filled at ODP startup time so that the cache of a not-yet-running thread is not empty. So I think the code here should read the local cache size of all possible threads (using multiple stats calls if necessary) if it wants to check that the sum equals the total cache size.

MatiasElo: The thread ID allocation should be clarified in the API  as discussed in #1619. Anyway, it makes sense to test `odp_pool_stats()` with a varying thread range, I'll update the test in V5.

MatiasElo:
> It would also help to put a note in the spec stating that there should be no pool operations issued by other threads while the pool stats APIs are in progress as it can result in wrong values.

This should be quite clear to a user already and the same stands for most of the statistics APIs. If another threads are causing changes to the counter values there is no guarantee that the read values are "correct".

MatiasElo:
V5:
- Fixed validation test comments
- Added TODO comment for replacing thread local counters with atomics in the implementation

MatiasElo:
Hi @ashwinyes, are you OK with V5?

ashwinyes:

PR #1616, opened 2022-07-06
[PATCH v3] api: cpumask: default mask pointer may be NULL

psavol:
Improved default worker/control mask API documentation. Allow application to use NULL as mask pointer, when it is only interested about number of worker/control CPUs.

### DIFF
--- a/include/odp/api/spec/cpumask.h
+++ b/include/odp/api/spec/cpumask.h
@@ -1,4 +1,5 @@
 /* Copyright (c) 2013-2018, Linaro Limited
+ * Copyright (c) 2022, Nokia
  * All rights reserved.
  *
  * SPDX-License-Identifier:     BSD-3-Clause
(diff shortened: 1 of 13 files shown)


include/odp/api/spec/cpumask.h
```diff
@@ -1,4 +1,5 @@
 /* Copyright (c) 2013-2018, Linaro Limited
+ * Copyright (c) 2022, Nokia
```
MatiasElo, line 2: Commit message: `usage` -> `usage of`

MatiasElo (approved):

include/odp/api/spec/cpumask.h
psavol: Fixed in v3.

MatiasElo:
Replaced by #1620.

PR #1619, opened 2022-07-11
[PATCH v3] validation: thread: improve thread id and count tests

MatiasElo:


test/common/odp_cunit_common.c
```diff
@@ -187,6 +187,12 @@ int odp_cunit_thread_create(int num, int func_ptr(void *), void *const arg[], in
 	odph_thread_common_param_t thr_common;
 	odph_thread_param_t thr_param[num];
 
+	if (num > ODP_THREAD_COUNT_MAX) {
+		fprintf(stderr, "error: %s: too many threads: max=%d\n", __func__,
```
TuomasTaipale, line 191: The attempted amount could be printed as well.

test/validation/api/thread/thread.c
```diff
@@ -81,20 +82,30 @@ static int thread_global_term(odp_instance_t inst)
...
 }
 
 static void thread_test_odp_thread_id(void)
 {
-	(void)odp_thread_id();
-	CU_PASS();
+	int id = odp_thread_id();
+
+	/* First created thread */
+	CU_ASSERT(id == 0);
```
JannePeltonen, line 93: This is not guaranteed by the API spec.

MatiasElo: We discussed about this last week with Petri and the intention of the API has always been this. However, you are correct that the API spec is not  completely clear on this. This check (and another similar in `thread_test_odp_thread_count()`) could be removed for now and added back when the spec is clarified.

test/common/odp_cunit_common.c
MatiasElo: Fixed in V2.

test/validation/api/thread/thread.c
MatiasElo: Fixed in V2.

test/validation/api/thread/thread.c
```diff
@@ -81,20 +82,29 @@ static int thread_global_term(odp_instance_t inst)
...
 }
 
 static void thread_test_odp_thread_count(void)
 {
-	(void)odp_thread_count();
-	CU_PASS();
+	int count = odp_thread_count();
+
+	/* One thread running */
+	CU_ASSERT(count == 1);
```
TuomasTaipale, line 102: Leftover?

MatiasElo: Nope, there is only one thread running, so the return value should be one. The checks below are overlapping but give extra information in case of an invalid count is returned.

TuomasTaipale: OK.

PR #1620, opened 2022-07-12
[PATCH v2] api: cpumask: default CPU mask may be NULL

MatiasElo:
Copy of PR #1616 with reviewed-by tags added.


MatiasElo:
Copy of PR #1616 with reviewed-by tags added.

PR #1629, opened 2022-07-15
[PATCH v3] helper: threads: fix synchronized thread start in process mode

MatiasElo:
Memory for odph_thread_create() output thread table is given by
application, so there is no guarantee that the memory can be used to
synchronize child process startup. Use socketpair() instead.

### DIFF
--- a/helper/threads.c
+++ b/helper/threads.c
@@ -1,5 +1,5 @@
 /* Copyright (c) 2013-2018, Linaro Limited
- * Copyright (c) 2019, Nokia
+ * Copyright (c) 2019-2022, Nokia
  * All rights reserved.
  *
  * SPDX-License-Identifier:     BSD-3-Clause
@@ -19,6 +19,7 @@
 #include <limits.h>
 #include <sys/time.h>
 #include <sys/resource.h>
+#include <sys/socket.h>
 
 #include <odp_api.h>
 #include <odp/helper/threads.h>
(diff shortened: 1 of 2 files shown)


helper/threads.c
```diff
@@ -34,6 +35,38 @@
 
 static odph_helper_options_t helper_options;
 
+static uint8_t get_child_status(int fd)
+{
+	uint8_t status;
+	int ret;
+
+	ret = recv(fd, &status, sizeof(status), MSG_DONTWAIT | MSG_NOSIGNAL);
```
TuomasTaipale, line 43: `recv` could be wrapped in `TEMP_FAILURE_RETRY`.

MatiasElo: Added in V2.

helper/threads.c
```diff
@@ -65,8 +98,14 @@ static void *run_thread(void *arg)
 		 "pthread" : "process",
 		 (int)getpid());
 
-	if (odp_atomic_load_u32(&start_args->status) == SYNC_INIT)
-		odp_atomic_store_rel_u32(&start_args->status, INIT_DONE);
+	if (start_args->status == SYNC_INIT) {
+		if (set_child_status(start_args->sockfd, INIT_DONE)) {
+			ODPH_ERR("Init status send failed\n");
```
TuomasTaipale, line 103: At this level, the log could state that "Init status *setting* failed" as the function name is `set_child_status()`, otherwise looks OK.

TuomasTaipale (approved):

helper/threads.c
MatiasElo: Fixed in V3.

JereLeppanen (approved):

PR #1630, opened 2022-07-15
[PATCH v5] helper: cli: add more packet I/O statistics printing commands

TuomasTaipale:
This patchset adds more commands to helper CLI for printing packet I/O total, queue, event queue and extra statistics.

v2:
- Copyright updated
- Refactored packet I/O handle getter

v4:
- Moved some commands out of `call` parent
- Reverted the I/O handle getter change
- Added new functions for queue stat printing

v5:
- Added reviewed-by tag

### DIFF
--- a/helper/cli.c
+++ b/helper/cli.c
@@ -1,4 +1,4 @@
-/* Copyright (c) 2021, Nokia
+/* Copyright (c) 2021-2022, Nokia
  * All rights reserved.
  *
  * SPDX-License-Identifier:     BSD-3-Clause
@@ -17,6 +17,7 @@
 #include <poll.h>
 #include <stdio.h>
 #include <strings.h>
+#include <inttypes.h>
 
 /* Socketpair socket roles. */
 enum {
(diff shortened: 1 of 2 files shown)


helper/cli.c
```diff
@@ -384,6 +693,18 @@ static struct cli_def *create_cli(cli_shm_t *shm)
...
 			     PRIVILEGE_UNPRIVILEGED, MODE_EXEC, "<name>");
+	cli_register_command(cli, c, "odp_pktio_stats_print",
+			     cmd_call_odp_pktio_stats_print,
+			     PRIVILEGE_UNPRIVILEGED, MODE_EXEC, "<name>");
+	cli_register_command(cli, c, "odp_pktio_queue_stats_print",
+			     cmd_call_odp_pktio_queue_stats_print,
+			     PRIVILEGE_UNPRIVILEGED, MODE_EXEC, "<name>");
+	cli_register_command(cli, c, "odp_pktio_event_queue_stats_print",
+			     cmd_call_odp_pktio_event_queue_stats_print,
+			     PRIVILEGE_UNPRIVILEGED, MODE_EXEC, "<name>");
```
MatiasElo, line 704: Naming is perhaps a bit confusing since there are no such API calls as `odp_pktio_queue_stats_print()` and `odp_pktio_event_queue_stats_print()`. So these commands could be split into input and output specific ones (e.g. `odp_pktin_queue_stats()` and `odp_pktout_queue_stats()`).

helper/cli.c
```diff
@@ -274,6 +374,215 @@ static int cmd_call_odp_pktio_print(struct cli_def *cli,
...
+{
+	if (check_num_args(cli, argc, 1))
+		return CLI_ERROR;
+
+	odp_pktio_t hdl = odp_pktio_lookup(argv[0]);
+
+	if (hdl == ODP_PKTIO_INVALID) {
+		cli_error(cli, "%% Name not found.");
+		return CLI_ERROR;
+	}
```
MatiasElo, line 389: Repeated in each pktio command, so could be made a separate function. Especially if  we end up splitting `odp_pktio_queue_stats_print()` and `odp_pktio_event_queue_stats_print()`.

helper/cli.c
TuomasTaipale: The intent here is that it is more convenient for CLI user to get both input and output queue dumps with a single command and it's clear enough what the user will get even though the commands don't exactly match an API call.

helper/cli.c
```diff
@@ -17,6 +17,7 @@
 #include <poll.h>
 #include <stdio.h>
 #include <strings.h>
+#include <inttypes.h>
```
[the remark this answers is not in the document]
TuomasTaipale: v2.

helper/cli.c
TuomasTaipale: Refactored this a bit, should be slightly less repetitive now. v2.

MatiasElo (approved):

helper/cli.c
```diff
@@ -255,25 +355,224 @@ static int cmd_call_odp_sys_info_print(struct cli_def *cli,
 	return CLI_OK;
 }
 
-static int cmd_call_odp_pktio_print(struct cli_def *cli,
-				    const char *command ODP_UNUSED,
-				    char *argv[], int argc)
+static odp_pktio_t get_pktio(struct cli_def *cli, char *argv[], int argc)
```
JereLeppanen, line 358: I wouldn't make this a function. These parameters to get_pktio() look a little odd, when intuitively all it needs is a char *. But in that case you don't even need a function.

Some future command that needs a pktio might have some additional parameters, so check_num_args(..., 1) wouldn't work in that case.

Reducing code duplication is good, but in this case I would just repeat these few lines in each command.

helper/cli.c
```diff
@@ -255,25 +355,224 @@ static int cmd_call_odp_sys_info_print(struct cli_def *cli,
...
+				(void)cli_log(ODP_LOG_PRINT,
+					      "  (Unable to read statistics, skipping)\n");
+				continue;
+			}
+
+			(void)cli_log(ODP_LOG_PRINT, "  octets: %" PRIu64 "\n", in_stats.octets);
+			(void)cli_log(ODP_LOG_PRINT, "  packets: %" PRIu64 "\n", in_stats.packets);
+			(void)cli_log(ODP_LOG_PRINT, "  discards: %" PRIu64 "\n",
+				      in_stats.discards);
+			(void)cli_log(ODP_LOG_PRINT, "  errors: %" PRIu64 "\n", in_stats.errors);
```
JereLeppanen, line 454: This could be a function. cli_log_pktin_queue_stats(odp_pktin_queue_stats_t *) or something like that. And the same for odp_pktout_queue_stats_t.

helper/cli.c
TuomasTaipale: OK, thanks, wasn't aware of the history of the architecture, changed to parentless in v4.

helper/cli.c
```diff
@@ -255,25 +355,224 @@ static int cmd_call_odp_sys_info_print(struct cli_def *cli,
...
+		return CLI_ERROR;
+
+	odp_pktio_stats_t stats;
+
+	if (odp_pktio_stats(hdl, &stats) < 0) {
+		cli_error(cli, "%% Unable to query stats.");
+		return CLI_ERROR;
+	}
+
+	(void)cli_log(ODP_LOG_PRINT, "Pktio statistics\n----------------\n");
```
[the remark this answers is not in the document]
TuomasTaipale: `odp_log()` returns something but I'm explicitly ignoring it (what's the point of `cli_log()` returning anything?). Anyways, not casting to `void` in v4.

helper/cli.c
TuomasTaipale: I have no strong opinions on this, though do slightly prefer the original way, so reverted to that in v4.

helper/cli.c
TuomasTaipale: v4.

TuomasTaipale:
v4:
- Moved some commands out of `call` parent
- Reverted the I/O handle getter change
- Added new functions for queue stat printing

JereLeppanen (approved):

TuomasTaipale:
v5:
- Added reviewed-by tag

helper/cli.c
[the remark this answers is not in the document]
JereLeppanen: > what's the point of `cli_log()` returning anything?

It's rarely needed, but the underlying print functions return something, so we basically just return that.

PR #1631, opened 2022-07-18
[PATCH v3] dependencies: clarify XDP and DPDK zero-copy exclusivity

TuomasTaipale:
v2:
- Wording changed

v3:
- Reviewed-by tag added

### DIFF
--- a/DEPENDENCIES
+++ b/DEPENDENCIES
@@ -305,6 +305,9 @@ Prerequisites for building the OpenDataPlane (ODP) API
    e.g. with some packet length, packet segment length and pool size
    combinations that would otherwise conform to reported capabilities.
 
+   Note that, currently, AF_XDP socket packet I/O cannot be instantiated if
+   DPDK zero-copy is enabled.
+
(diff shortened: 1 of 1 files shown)


DEPENDENCIES
```diff
@@ -305,6 +305,9 @@ Prerequisites for building the OpenDataPlane (ODP) API
    e.g. with some packet length, packet segment length and pool size
    combinations that would otherwise conform to reported capabilities.
 
+   Note that, currently, AF_XDP socket packet I/O cannot be instantiated if
+   DPDK zero-copy is configured.
```
MatiasElo, line 309: Perhaps `enabled` would be a better word, since it looks like XDP can still be used if DPDK zero-copy pktio is configured and disabled.

TuomasTaipale: Yes, that's more accurate, fixed in v2.

MatiasElo (approved):

TuomasTaipale:
v3:
- Reviewed-by tag added

PR #1634, opened 2022-07-20
[PATCH v3] linux-gen: pool: properly initialize ext pool base and max addresses

TuomasTaipale:
This patchset fixes `odp_packet_is_valid()` behavior with packets allocated from external memory pools. Currently the check always fails due to pool base and maximum address fields being zero. This is fixed by properly initializing the addresses utilizing the populated memory chunks.

v2:
- Added reviewed-by tag


MatiasElo (approved):

TuomasTaipale:
v2:
- Added reviewed-by tag

PR #1635, opened 2022-07-20
[PATCH v1] validation: cls: fix CoS creation order

satheeshpaul:
Fix CoS creation order to create Error CoS before creating
any user defined CoS. This is done because on some
hardwares, creating the Error CoS later leads to updating
all previously created PMRs such that they do not match any
packet with errors.


JannePeltonen:
I do not see proper justification for this change.

The validation test appears to comply with the ODP API which does not put restrictions on when odp_pktio_error_cos_set() is called or when a CoS object that will be used as a pktio error CoS can be created in relation to other CoS objects.

If your ODP implementation has problems with this validation test, then I think the problem is with the ODP implementation, not with the validation test and therefore it would be the ODP implementation, not the validation test, that should be fixed.

PR #1636, opened 2022-07-20
[PATCH v7] api: stash: add batch variants of put/get functions

MatiasElo:


MatiasElo:
V2:
- Mention that trying to obtain/store more than max_get_batch/max_put_batch objects leads to undefined behaviour

platform/linux-generic/odp_stash.c
```diff
@@ -118,6 +122,8 @@ int odp_stash_capability(odp_stash_capability_t *capa, odp_stash_type_t type)
 	capa->max_stashes          = CONFIG_MAX_STASHES - CONFIG_INTERNAL_STASHES;
 	capa->max_num_obj          = MAX_RING_SIZE;
 	capa->max_obj_size         = sizeof(uint64_t);
+	capa->max_get_batch        = MAX_BATCH_SIZE;
+	capa->max_put_batch        = MAX_BATCH_SIZE;
```
TuomasTaipale, line 126: Matter of taste, but could use `MIN_RING_SIZE` directly here.

platform/linux-generic/odp_stash.c
```diff
@@ -432,22 +488,30 @@ int32_t odp_stash_get(odp_stash_t st, void *obj, int32_t num)
 	if (obj_size == sizeof(uint64_t)) {
 		ring_u64_t *ring_u64 = &stash->ring_u64.hdr;
 
-		return ring_u64_deq_multi(ring_u64, stash->ring_mask, obj, num);
+		if (batch)
```
TuomasTaipale, line 491: This could be done once at the beginning of the function to save correct functions to variables and then use those later.

Alternatively, correct functions could be passed directly to `stash_get()` (a bit clearer than passing a flag).

MatiasElo: Using function pointers would effectively prevent inlining. My target was to add support for batch operations with minimal overhead without having to duplicate the whole function.

TuomasTaipale: Sure, on the other hand, if inlined, it bloats the cache footprint as there are always the dead branches. I guess this would need to measured.

Anyways, current solution is OK to me as well.

platform/linux-generic/odp_stash.c
MatiasElo: Changed in V3.

TuomasTaipale (approved):

MatiasElo:
V5:
- Reworded batch function definitions

psavol (approved):

MatiasElo:
V6:
- Rebased and added reviewed-by tags

PR #1639, opened 2022-07-25
[PATCH v10] Avoid possible deadloop in llq dequeue

easynetsec:
The cpu_pause while loop is added to handle the case that the dequeue happens when an enqueue is in progress and the llq->u.ui.head is being updated.

In some corner case, however, if the old.st.head is dequeued from llq and then enqueued into another queue before the while loop starts, then the dequeue can get stuck in the while loop until the queue is scheduled again.

Add a threshold of spin count, so we will re-check the llq head once max spin count is reached.

### DIFF
--- a/platform/linux-generic/include/odp_llqueue.h
+++ b/platform/linux-generic/include/odp_llqueue.h
@@ -41,6 +41,7 @@ static odp_bool_t llq_on_queue(struct llnode *node);
  *****************************************************************************/
 
 #define SENTINEL ((void *)~(uintptr_t)0)
+#define MAX_SPIN_COUNT 1000
 
 #ifdef CONFIG_LLDSCD
 /* Implement queue operations using double-word LL/SC */
@@ -114,6 +115,7 @@ static inline struct llnode *llq_dequeue(struct llqueue *llq)
 	(void)__atomic_load_n(&head->next, __ATOMIC_RELAXED);
 
 	do {
+restart_loop:
 		old.ui = lld(&llq->u.ui, __ATOMIC_RELAXED);
 		if (odp_unlikely(old.st.head == NULL)) {
 			/* Empty list */
@@ -125,13 +127,18 @@ static inline struct llnode *llq_dequeue(struct llqueue *llq)
 		} else {
 			/* Multi-element list, dequeue head */
 			struct llnode *next;
+			int	spin_count = 0;
+
 			/* Wait until llq_enqueue() has written true next
 			 * pointer
 			 */
 			while ((next = __atomic_load_n(&old.st.head->next,
 						       __ATOMIC_RELAXED)) ==
-				SENTINEL)
+				SENTINEL) {
 				odp_cpu_pause();
+				if (++spin_count >= MAX_SPIN_COUNT)
+					goto restart_loop;
+			}
 			neu.st.head = next;
 			neu.st.tail = old.st.tail;
 		}
@@ -146,6 +153,7 @@ static inline odp_bool_t llq_dequeue_cond(struct llqueue *llq,
 	union llht old, neu;
 
 	do {
+restart_loop:
 		old.ui = lld(&llq->u.ui, __ATOMIC_ACQUIRE);
 		if (odp_unlikely(old.st.head == NULL || old.st.head != exp)) {
 			/* Empty list or wrong head */
@@ -157,13 +165,17 @@ static inline odp_bool_t llq_dequeue_cond(struct llqueue *llq,
 		} else {
 			/* Multi-element list, dequeue head */
 			struct llnode *next;
+			int     spin_count = 0;
 
 			/* Wait until llq_enqueue() has written true next
 			 * pointer */
 			while ((next = __atomic_load_n(&old.st.head->next,
 						       __ATOMIC_RELAXED)) ==
-				SENTINEL)
+				SENTINEL) {
 				odp_cpu_pause();
+				if (++spin_count >= MAX_SPIN_COUNT)
+					goto restart_loop;
+			}
 
 			neu.st.head = next;
 			neu.st.tail = old.st.tail;


platform/linux-generic/include/odp_llqueue.h
```diff
@@ -41,6 +41,7 @@ static odp_bool_t llq_on_queue(struct llnode *node);
  *****************************************************************************/
```
JannePeltonen, line 42: Thanks for spotting this bug and submitting a fix.

At a very quick glance the fix appears valid, but I think the commit message could be made more clear and follow the ODP commit message style (separate title line, line lenghts, etc). Maybe something like this would explain the commit better:

```
avoid endless loop in llq_dequeue()

llq_enqueue() updates the next pointer of the previous tail node after it
has atomically updated the list head structure. Because of this, when
llq_dequeue() tries to dequeue from a multi-element list, it has to wait
until the next pointer of the head node is not null.

If two llq_dequeue() operations occur at the same time, it could happen
that one thread succesfully dequeues the head node when a second thread
has already read the list head structure but has not yet checked the next
pointer of the head node. Now if the first thread that already dequeued
the head node writes null to its next pointer (e.g. by enqueueing the node
to some other llqueue), the second thread may wait indefinitely for the
next pointer to become non-null. At this point the second thread is using
an outdated list head structure and any update to it through the scd
operation would fail. But the second thread does not get that far unless
the next pointer of the node that now lives outside this list gets
a non-null value.

Fix the problem by putting a loop count limit on the wait loop and
restarting the ll-sc loop if the loop count is reached.

Signed-off-by: your name <your email>
```

platform/linux-generic/include/odp_llqueue.h
```diff
@@ -41,6 +41,7 @@ static odp_bool_t llq_on_queue(struct llnode *node);
...
 
 #define SENTINEL ((void *)~(uintptr_t)0)
+#define MAX_SPIN_COUNT 100000
```
JannePeltonen, line 44: That long a wait looks a bit excessive to me. I wonder if something like 1000 or even less be better.

easynetsec: The number I set is about 5M cpu cycles in one of our boxes, roughly 0.25s. With this number, I find most enqueue will just finish without going to restart. 

My goal is to not restart for normal wait, but only for the possible case of endless loop.

platform/linux-generic/include/odp_llqueue.h
easynetsec: Thanks for the very detailed commit message. Yes, that explained the commit better than my original message. 

I will update.

easynetsec: Commit message updated.

platform/linux-generic/include/odp_llqueue.h
JannePeltonen: 5 M cycles or 0.25 seconds is a very long time to stop a thread that may have possibly latency sensitive work to do. The problem of an indefinite wait is replaced with the problem of a long wait that can easily disturb many applications and cause packet loss and whatnot.

Wait time longer than necessary does not provide any benefit but makes dequeue last longer if the simultaneous dequeue occurs in a bad way. "Too short" a wait time merely causes a extra load-linked accesses.

If we want to avoid extra load-linked operations when enqueue happens simultaneously with the dequeue of the previous node, the wait needs to be only as long as it takes for the new next pointer to become visible after the sc was done. I would think it does not typically take a huge number of cycles.

JannePeltonen: Now that I look at the existing code, I think there may be a problem with memory ordering. llq_enqueue() calls scd() with release memory order, but the llq_dequeue() calls lld() with relaxed memory order instead of acquire which would pair with the release. Because of this, it seems that there is no guarantee that the writing of null to node->next by llq_enqueue() becomes visible in llq_dequeue() in time. So the llq_dequeue() may see a stale non-null value in the next pointer.

If this is indeed the case, then I think another patch is needed (it is best to keep different fixes separate).

easynetsec: My goal indeed is to avoid extra load-linked operations when enqueue happens simultaneously with the dequeue of the previous node.

In my box, I find that it can take some time for dequeue to wait (by checking if goto happens). And that is why I come up with the number 100000. Of course, this is only from one profile of my box.

If you insist, I can change it to some smaller value like 1000. The one I used in our code is 100000.

easynetsec: > Now that I look at the existing code, I think there may be a problem with memory ordering. llq_enqueue() calls scd() with release memory order, but the llq_dequeue() calls lld() with relaxed memory order instead of acquire which would pair with the release. Because of this, it seems that there is no guarantee that the writing of null to node->next by llq_enqueue() becomes visible in llq_dequeue() in time. So the llq_dequeue() may see a stale non-null value in the next pointer.
> 
> If this is indeed the case, then I think another patch is needed (it is best to keep different fixes separate).

Maybe this also contributes to the long cycles that dequeue needs to wait?

easynetsec: > 5 M cycles or 0.25 seconds is a very long time to stop a thread that may have possibly latency sensitive work to do. The problem of an indefinite wait is replaced with the problem of a long wait that can easily disturb many applications and cause packet loss and whatnot.
> 
> Wait time longer than necessary does not provide any benefit but makes dequeue last longer if the simultaneous dequeue occurs in a bad way. "Too short" a wait time merely causes a extra load-linked accesses.
> 
> If we want to avoid extra load-linked operations when enqueue happens simultaneously with the dequeue of the previous node, the wait needs to be only as long as it takes for the new next pointer to become visible after the sc was done. I would think it does not typically take a huge number of cycles.

---

I would think it should not take too many cycles too. But in my box, I find that it can take some time for dequeue to wait (by checking if goto happens). And that is why I come up with the number 100000. Of course, this is only from one profile of my box.

And my goal indeed is to avoid extra load-linked operations when enqueue happens simultaneously with the dequeue of the previous node. Too many retries also wast cpu cycles during spinning, right?

Anyway if you insist, I can change it to some smaller value like 1000. The one I used in my box is 100000.

platform/linux-generic/include/odp_llqueue.h
```diff
@@ -114,6 +115,7 @@ static inline struct llnode *llq_dequeue(struct llqueue *llq)
 	(void)__atomic_load_n(&head->next, __ATOMIC_RELAXED);
 
 	do {
+restartLoop:
```
MatiasElo, line 118: This project does not use camel case, so please change to `restart_loop`.

platform/linux-generic/include/odp_llqueue.h
```diff
@@ -125,13 +127,19 @@ static inline struct llnode *llq_dequeue(struct llqueue *llq)
...
 			struct llnode *next;
+			int	spinCount=0;
+
 			/* Wait until llq_enqueue() has written true next
 			 * pointer
 			 */
 			while ((next = __atomic_load_n(&old.st.head->next,
 						       __ATOMIC_RELAXED)) ==
 				SENTINEL)
+			{
```
MatiasElo, line 138: Please move brace to the previous line.

platform/linux-generic/include/odp_llqueue.h
```diff
@@ -125,13 +127,19 @@ static inline struct llnode *llq_dequeue(struct llqueue *llq)
 		} else {
 			/* Multi-element list, dequeue head */
 			struct llnode *next;
+			int	spinCount=0;
```
MatiasElo, line 130: Camel case -> `spin_count`

platform/linux-generic/include/odp_llqueue.h
```diff
@@ -146,6 +154,7 @@ static inline odp_bool_t llq_dequeue_cond(struct llqueue *llq,
 	union llht old, neu;
 
 	do {
+restartLoop:
```
MatiasElo, line 157: Same comments as for `llq_dequeue()`.

MatiasElo (commented):
Couple minor style tweaks. Also, the commit subject line should still be modified a bit. E.g. `linux-gen: scalable: avoid endless loop in llq_dequeue()`

platform/linux-generic/include/odp_llqueue.h
easynetsec: ok will do.

easynetsec: All done.

platform/linux-generic/include/odp_llqueue.h
easynetsec: set to 1000.

easynetsec:
Fixed checking failures by CI.

MatiasElo (approved):
Janne is on vacation for the rest of this week, so we'll likely have to wait for his review until next week, but looks ok to me.

MatiasElo:
As a side node, have you tried using the default scheduler implementation recently? It should perform better in most use cases and the code is better maintained.

easynetsec:
@JannePeltonen  do you have more comments? if not, can we get this merged?

easynetsec:
> As a side node, have you tried using the default scheduler implementation recently? It should perform better in most use cases and the code is better maintained.

By default, you mean the basic scheduler? We moved from basic to scalable at some point because the scalable scheduler gave us some performance improvement. We haven't upgraded our odp for a while. So maybe we should take a look next time we upgrade.

JannePeltonen (approved):

psavol:
This is otherwise ready, but needs review-by tags. Please, rebase and add review tags of Matias and Janne.

easynetsec:
> This is otherwise ready, but needs review-by tags. Please, rebase and add review tags of Matias and Janne.

Tags update and rebase are done. Anything else?

PR #1640, opened 2022-07-26
[PATCH v7] test: performance: add IPsec performance tester

TuomasTaipale:
`odp_ipsecfwd` is a simple IPsec performance testing utility to profile
ODP IPsec implementation performance.

Tester supports running IPsec processing in multiple workers and in a
few different queue setups. Packets are forwarded according to passed
forwarding configuration.

v4:
- Review comments addressed
- All algorithms defined in spec now exposed in tester

v5:
- Fixed help text
- Dropped IPsec-unsupported algorithms from exposed algorithms

v7:
- Adjusted MAC address fetch error log
- Adjusted forwarding entry parsing format

### DIFF
--- a/test/performance/odp_ipsecfwd.c
+++ b/test/performance/odp_ipsecfwd.c
@@ -1,0 +1,1303 @@
+/* Copyright (c) 2022, Nokia
+ *
+ * All rights reserved.
+ *
+ * SPDX-License-Identifier:     BSD-3-Clause
+ */
+
+#ifndef _GNU_SOURCE
+#define _GNU_SOURCE
+#endif
+
+#include <stdlib.h>
+#include <signal.h>
+#include <stdio.h>
+#include <inttypes.h>
+#include <string.h>
+
+#include <errno.h>
+
+#include <odp_api.h>
+#include <odp/helper/odph_api.h>
+
+#define PROG_NAME "odp_ipsecfwd"
+#define DELIMITER ","
+
+#define MIN(a, b)  (((a) <= (b)) ? (a) : (b))
+
+#define MAX_IFS 2U
+#define MAX_SAS 64U
+#define MAX_FWDS 64U
+#define MAX_SPIS (UINT16_MAX + 1U)
+#define MAX_WORKERS (ODP_THREAD_COUNT_MAX - 1U)
+#define MAX_QUEUES 16U
+#define PKT_SIZE 1024U
+#define PKT_CNT 32768U
+#define WORKER_CNT 1U
+#define NUM_EVS 32U
+
+#define ALG_ENTRY(_alg_name, _type) \
+	{ \
+		.idx = (_alg_name), \
+		.type = (_type), \
+		.name = #_alg_name \
+	}
+
+enum {
+	CIPHER_TYPE,
+	COMB_CIPHER_TYPE,
+	AUTH_TYPE,
+	COMB_AUTH_TYPE
+};
+
+typedef enum {
+	PRS_OK,
+	PRS_NOK,
+	PRS_TERM
+} parse_result_t;
+
+enum {
+	DIR_IN = 0U,
+	DIR_OUT
+};
+
+typedef struct pktio_s pktio_t;
+
+typedef struct pktio_s {
+	union {
+		odp_pktout_queue_t out_dir_qs[MAX_QUEUES];
+		odp_queue_t out_ev_q;
+	};
+
+	odph_ethaddr_t src_mac;
+	char *name;
+	odp_pktio_t handle;
+	odp_bool_t (*send_fn)(const pktio_t *pktio, uint8_t index, odp_packet_t pkt);
+	uint32_t num_tx_qs;
+} pktio_t;
+
+typedef struct {
+	odph_ethaddr_t dst_mac;
+	const pktio_t *pktio;
+	odph_iplookup_prefix_t prefix;
+} fwd_entry_t;
+
+typedef struct {
+	uint64_t ipsec_in_pkts;
+	uint64_t ipsec_out_pkts;
+	uint64_t ipsec_in_errs;
+	uint64_t ipsec_out_errs;
+	uint64_t status_errs;
+	uint64_t fwd_pkts;
+	uint64_t discards;
+} stats_t;
+
+typedef struct prog_config_s prog_config_t;
+
+typedef struct ODP_ALIGNED_CACHE {
+	stats_t stats;
+	prog_config_t *prog_config;
+} thread_config_t;
+
+typedef struct prog_config_s {
+	odph_thread_t thread_tbl[MAX_WORKERS];
+	thread_config_t thread_config[MAX_WORKERS];
+	odp_ipsec_sa_t sas[MAX_SAS];
+	fwd_entry_t fwd_entries[MAX_FWDS];
+	odp_queue_t ev_qs[MAX_SAS];
+	pktio_t pktios[MAX_IFS];
+	char *sa_conf_file;
+	char *fwd_conf_file;
+	odp_instance_t odp_instance;
+	odp_queue_t compl_q;
+	odp_pool_t pktio_pool;
+	odph_table_t fwd_tbl;
+	odp_barrier_t init_barrier;
+	odp_barrier_t term_barrier;
+	uint32_t num_pkts;
+	uint32_t pkt_len;
+	uint32_t num_ifs;
+	uint32_t num_sas;
+	uint32_t num_fwds;
+	int num_thrs;
+	uint8_t mode;
+} prog_config_t;
+
+typedef struct {
+	const char *name;
+	int idx;
+	int type;
+} exposed_alg_t;
+
+static exposed_alg_t exposed_algs[] = {
+	ALG_ENTRY(ODP_CIPHER_ALG_NULL, CIPHER_TYPE),
+	ALG_ENTRY(ODP_CIPHER_ALG_DES, CIPHER_TYPE),
+	ALG_ENTRY(ODP_CIPHER_ALG_3DES_CBC, CIPHER_TYPE),
+	ALG_ENTRY(ODP_CIPHER_ALG_AES_CBC, CIPHER_TYPE),
+	ALG_ENTRY(ODP_CIPHER_ALG_AES_CTR, CIPHER_TYPE),
+	ALG_ENTRY(ODP_CIPHER_ALG_AES_ECB, CIPHER_TYPE),
+	ALG_ENTRY(ODP_CIPHER_ALG_AES_GCM, COMB_CIPHER_TYPE),
+	ALG_ENTRY(ODP_CIPHER_ALG_AES_CCM, COMB_CIPHER_TYPE),
+	ALG_ENTRY(ODP_CIPHER_ALG_CHACHA20_POLY1305, COMB_CIPHER_TYPE),
+	ALG_ENTRY(ODP_AUTH_ALG_NULL, AUTH_TYPE),
+	ALG_ENTRY(ODP_AUTH_ALG_MD5_HMAC, AUTH_TYPE),
+	ALG_ENTRY(ODP_AUTH_ALG_SHA1_HMAC, AUTH_TYPE),
+	ALG_ENTRY(ODP_AUTH_ALG_SHA224_HMAC, AUTH_TYPE),
+	ALG_ENTRY(ODP_AUTH_ALG_SHA256_HMAC, AUTH_TYPE),
+	ALG_ENTRY(ODP_AUTH_ALG_SHA384_HMAC, AUTH_TYPE),
+	ALG_ENTRY(ODP_AUTH_ALG_SHA512_HMAC, AUTH_TYPE),
+	ALG_ENTRY(ODP_AUTH_ALG_AES_GCM, COMB_AUTH_TYPE),
+	ALG_ENTRY(ODP_AUTH_ALG_AES_GMAC, AUTH_TYPE),
+	ALG_ENTRY(ODP_AUTH_ALG_AES_CCM, COMB_AUTH_TYPE),
+	ALG_ENTRY(ODP_AUTH_ALG_AES_CMAC, AUTH_TYPE),
+	ALG_ENTRY(ODP_AUTH_ALG_AES_XCBC_MAC, AUTH_TYPE),
+	ALG_ENTRY(ODP_AUTH_ALG_CHACHA20_POLY1305, COMB_AUTH_TYPE)
+};
+
+/* SPIs for in and out directions */
+static odp_ipsec_sa_t *spi_to_sa_map[2U][MAX_SPIS];
+static odp_atomic_u32_t is_running;
+
+static void terminate(int signal ODP_UNUSED)
+{
+	odp_atomic_store_u32(&is_running, 0U);
+}
+
+static void init_config(prog_config_t *config)
+{
+	memset(config, 0, sizeof(*config));
+	config->compl_q = ODP_QUEUE_INVALID;
+	config->pktio_pool = ODP_POOL_INVALID;
+	config->num_thrs = WORKER_CNT;
+}
+
+static void parse_interfaces(prog_config_t *config, const char *optarg)
+{
+	char *tmp_str = strdup(optarg), *tmp;
+
+	if (tmp_str == NULL)
+		return;
+
+	tmp = strtok(tmp_str, DELIMITER);
+
+	while (tmp && config->num_ifs < MAX_IFS) {
+		config->pktios[config->num_ifs].name = strdup(tmp);
+
+		if (config->pktios[config->num_ifs].name != NULL)
+			++config->num_ifs;
+
+		tmp = strtok(NULL, DELIMITER);
+	}
+
+	free(tmp_str);
+}
+
+static void print_supported_algos(const odp_ipsec_capability_t *ipsec_capa)
+{
+	int c_cnt, a_cnt;
+	const size_t len = sizeof(exposed_algs) / sizeof(exposed_algs[0]);
+
+	printf("                          Cipher algorithms:\n");
+
+	for (size_t i = 0U; i < len; ++i) {
+		if ((exposed_algs[i].type == CIPHER_TYPE ||
+		     exposed_algs[i].type == COMB_CIPHER_TYPE) &&
+		    (ipsec_capa->ciphers.all_bits & (1 << exposed_algs[i].idx)) > 0U) {
+			c_cnt = odp_ipsec_cipher_capability(exposed_algs[i].idx, NULL, 0);
+
+			if (c_cnt < 0)
+				continue;
+
+			printf("                              %d: %s",
+			       exposed_algs[i].idx, exposed_algs[i].name);
+			printf(exposed_algs[i].type == COMB_CIPHER_TYPE ? " (combined)" : "");
+
+			odp_ipsec_cipher_capability_t capa[c_cnt];
+
+			(void)odp_ipsec_cipher_capability(exposed_algs[i].idx, capa, c_cnt);
+
+			for (int j = 0; j < c_cnt; ++j)
+				printf(j == 0 ? " (key lengths: %u" : ", %u", capa[j].key_len);
+
+			printf(")\n");
+		}
+	}
+
+	printf("                          Authentication algorithms:\n");
+
+	for (size_t i = 0U; i < len; ++i) {
+		if ((exposed_algs[i].type == AUTH_TYPE ||
+		     exposed_algs[i].type == COMB_AUTH_TYPE) &&
+		    (ipsec_capa->auths.all_bits & (1 << exposed_algs[i].idx)) > 0U) {
+			a_cnt = odp_ipsec_auth_capability(exposed_algs[i].idx, NULL, 0);
+
+			if (a_cnt < 0)
+				continue;
+
+			printf("                              %d: %s",
+			       exposed_algs[i].idx, exposed_algs[i].name);
+			printf(exposed_algs[i].type == COMB_AUTH_TYPE ? " (combined)" : "");
+
+			odp_ipsec_auth_capability_t capa[a_cnt];
+
+			(void)odp_ipsec_auth_capability(exposed_algs[i].idx, capa, a_cnt);
+
+			for (int j = 0; j < a_cnt; ++j)
+				printf(j == 0 ? " (key/icv lengths: %u/%u" : ", %u/%u",
+				       capa[j].key_len, capa[j].icv_len);
+
+			printf(")\n");
+		}
+	}
+}
+
+static void print_usage(void)
+{
+	odp_pool_capability_t pool_capa;
+	odp_ipsec_capability_t ipsec_capa;
+
+	if (odp_pool_capability(&pool_capa) < 0) {
+		ODPH_ERR("Error querying pool capabilities\n");
+		return;
+	}
+
+	if (odp_ipsec_capability(&ipsec_capa) < 0) {
+		ODPH_ERR("Error querying IPsec capabilities\n");
+		return;
+	}
+
+	printf("\n"
+	       "Simple IPsec performance tester. Forward and process plain and ipsec packets.\n"
+	       "\n"
+	       "Examples:\n"
+	       "    %s -i ens9f1 -s /etc/odp/sa.conf -f /etc/odp/fwd.conf\n"
+	       "\n"
+	       "    With sa.conf containing, for example:\n"
+	       "        0 222 192.168.1.10 192.168.1.16 4 jWnZr4t7w!zwC*F- 0 2"
+	       " n2r5u7x!A%%D*G-KaPdSg 0 12\n"
+	       "\n"
+	       "    With fwd.conf containing, for example:\n"
+	       "        192.168.1.0/24 ens9f1 aa:bb:cc:dd:11:22\n"
+	       "\n"
+	       "Usage: %s [options]\n"
+	       "\n"
+	       "  -i, --interfaces    Ethernet interfaces for packet I/O, comma-separated,\n"
+	       "                      no spaces.\n"
+	       "  -n, --num_pkts      Number of packet buffers allocated for packet I/O pool.\n"
+	       "                      %u by default.\n"
+	       "  -l, --pkt_len       Maximum size of packet buffers in packet I/O pool. %u by\n"
+	       "                      default.\n"
+	       "  -c, --count         Worker thread count, %u by default.\n"
+	       "  -m, --mode          Packet output mode.\n"
+	       "                          0: queued output (default)\n"
+	       "                          1: direct output\n"
+	       "  -s, --sa            SA configuration file. Individual SA configuration is\n"
+	       "                      expected to be within a single line, values whitespace\n"
+	       "                      separated:\n"
+	       "\n"
+	       "                          <line in file> Dir SPI TunSrcIPv4 TunDstIPv4"
+	       " CipherAlgoIdx CipherKey CipherKeyExtra AuthAlgIdx AuthKey AuthKeyExtra ICVLen\n"
+	       "\n"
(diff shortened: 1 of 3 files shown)


test/performance/odp_ipsecfwd.c
```diff
@@ -0,0 +1,1277 @@
...
+		++stats->ipsec_in_errs;
+		return PKT_NOK;
+	}
+
+	return PKT_OK;
+}
+
+static inline fwd_entry_t *lookup(odp_packet_t pkt, odph_table_t fwd_tbl)
+{
+	odph_ethhdr_t *eth = odp_packet_l2_ptr(pkt, NULL);
```
JannePeltonen, line 900: l2_ptr is not guaranteed to be valid after IPsec processing. L3 offset and l3_ptr are valid but not necessarily zero. You need to write the full Ethernet header, including ethertype, to the packet after first truncating or extending the packet head to make the L2 header the right size. It is a bit cumbersome but it is what the API requires.

There is also no guarantee that an IPsec-processed packet is contiguous in memory, so some checking is needed wherever pointers to packet data are used (also in ipsec input processing).

test/performance/odp_ipsecfwd.c
```diff
@@ -0,0 +1,1277 @@
...
+
+	if (odp_unlikely(!do_ipsec_in(pkt, sa))) {
+		++stats->ipsec_in_errs;
+		return PKT_NOK;
+	}
+
+	return PKT_OK;
+}
+
+static inline fwd_entry_t *lookup(odp_packet_t pkt, odph_table_t fwd_tbl)
```
JannePeltonen, line 898: The name of this function is a bit misleading since it does not do only lookup but header rewriting too.

test/performance/odp_ipsecfwd.c
```diff
@@ -0,0 +1,1277 @@
...
+		res = handle_udp(pkt, stats);
+
+	if (odp_unlikely(res == PKT_NOK))
+		goto err;
+
+	if (res == PKT_FWD) {
+		if (!forward_packet(pkt, fwd_tbl, stats, worker_id))
+			goto err;
+
+		is_ipsec_in(pkt) ? ++stats->ipsec_in_pkts : ++stats->ipsec_out_pkts;
```
JannePeltonen, line 997: handle_ipsec_packet() could call is_ipsec_in() only once in the beginning since we do not know how inlineable or fast odp_packet_user_ptr() is in a random ODP implementation. And if the call is done only once, then ip_ipsec_in could be a variable instead of a function.

test/performance/odp_ipsecfwd.c
```diff
@@ -0,0 +1,1277 @@
...
+		for (int i = 0; i < cnt; ++i) {
+			ev = evs[i];
+			type = odp_event_types(ev, &subtype);
+			pkt = odp_packet_from_event(ev);
+
+			if (type == ODP_EVENT_PACKET) {
+				if (subtype == ODP_EVENT_PACKET_BASIC) {
+					handle_new_packet(pkt, fwd_tbl, stats, worker_id);
+				} else if (subtype == ODP_EVENT_PACKET_IPSEC) {
+					handle_ipsec_packet(pkt, fwd_tbl, stats, worker_id);
```
JannePeltonen, line 1059: Naming of packet processing functions is slightly confusing. For instance, handle_ipsec_packet() does not handle received IPsec packets but IPsec-processed packets (which can be IPsec or cleartext, depending on the direction). The 'handle' prefix is also not very informative.

I would suggest name changes along these lines:

handle_packets() -> worker_main()
handle_new_packet() -> packet_input()
handle_udp() -> udp_input()
handle_ipsec() -> ipsec_input()
handle_ipsec_packet() -> ipsec_completion()

Actually, there is a bit of duplication of logic between handle_new_packet() and handle_ipsec_packet() and it seems adding support for synchronous ipsec operations would be a bit difficult. This is rather subjective but I think might be clearer to organize the code so that handle_new_packet() is split to separate packet_input() and packet_output() functions, somehow like this:

```
worker_main()
        while (ev = schedule())
                if (ev type is plain packet)
                        packet_input()
                else if (ev type is ipsec completion)
                        ipsec_completion()

packet_input()
        if (sa = ipsec_sa_lookup())
                ipsec_input()
        else
                packet_output()

ipsec_input()
        if (async)
                ipsec_in_enqueue()
        else
                ipsec_in()
                packet_output()

packet_output()
        if (sa = ipsec_sa_lookup())
                ipsec_output()
        else
                packet_forward()

ipsec_output()
        if (async)
                ipsec_out_enqueue()
        else
                ipsec_out()
                packet_forward()

ipsec_completion()
        if (inbound packet)
                packet_output()
        else
                packet_forward()

```

What do you think?

test/performance/odp_ipsecfwd.c
```diff
@@ -0,0 +1,1277 @@
...
+	crypto_param.auth_key.length = strlen((const char *)auth_key);
+	crypto_param.auth_key_extra.data = auth_key_extra;
+	crypto_param.auth_key_extra.length = strlen((const char *)auth_key_extra);
+	crypto_param.icv_len = icv_len;
+	sa_param.crypto = crypto_param;
+	sa = odp_ipsec_sa_create(&sa_param);
+
+	if (sa == ODP_IPSEC_SA_INVALID) {
+		ODPH_ERR("Error creating SA handle\n");
+		return;
```
JannePeltonen, line 436: The completion queue created earlier leaks at this point.

test/performance/odp_ipsecfwd.c
TuomasTaipale: Data accesses should be safer now in v3.

test/performance/odp_ipsecfwd.c
TuomasTaipale: Renamed in v3.

test/performance/odp_ipsecfwd.c
TuomasTaipale: Good point, called only once in v3.

test/performance/odp_ipsecfwd.c
TuomasTaipale: I refactored the call paths in v3, they do make a bit more sense now.

About the naming, I prefer having verbs in the function names, though "handle" is a bit lazy I agree so renamed them.

test/performance/odp_ipsecfwd.c
TuomasTaipale: Yes, thanks. Fixed in v3.

test/performance/odp_ipsecfwd.c
```diff
@@ -0,0 +1,1307 @@
...
+	config->compl_q = ODP_QUEUE_INVALID;
+	config->pktio_pool = ODP_POOL_INVALID;
+	config->num_thrs = WORKER_CNT;
+}
+
+static void parse_interfaces(prog_config_t *config, const char *optarg)
+{
+	char *tmp_str = strdup(optarg), *tmp = strtok(tmp_str, DELIMITER);
+
+	if (tmp_str == NULL)
```
JannePeltonen, line 159: This is too late. The first strtok() call must not be made with a null string.

test/performance/odp_ipsecfwd.c
```diff
@@ -0,0 +1,1307 @@
...
+		return;
+
+	if (tmp == NULL)
+		goto out;
```
JannePeltonen, line 163: This is redundant with the subsequent while condition. Removing this would get rid of the out label too.

test/performance/odp_ipsecfwd.c
```diff
@@ -0,0 +1,1307 @@
...
+	       "                      default.\n"
+	       "  -c, --count         CPU count, %u by default.\n"
+	       "  -m, --mode          Packet output mode.\n"
+	       "                          0: queued output (default)\n"
+	       "                          1: direct output\n"
+	       "  -s, --sa            SA configuration file. Individual SA configuration is\n"
+	       "                      expected to be within a single line, values whitespace\n"
+	       "                      separated:\n"
+	       "\n"
+	       "                          <line in file> Dir SPI SrcIPv4 DstIPv4 CipherAlgoIdx"
```
JannePeltonen, line 279: Maybe TunSrcIPv5 and TunDstIPv4 to make it clear that these are IPsec tunnel endpoint addresses?

test/performance/odp_ipsecfwd.c
```diff
@@ -0,0 +1,1307 @@
...
+	       "\n"
+	       "                      Supported cipher and authentication algorithms:\n",
+	       PROG_NAME, PROG_NAME, MIN(pool_capa.pkt.max_num, PKT_CNT),
+	       MIN(pool_capa.pkt.max_len, PKT_SIZE), WORKER_CNT);
+	print_supported_algos(&ipsec_capa);
+	printf("  -f, --fwd_table     Forwarding configuration file. Individual forwarding\n"
+	       "                      configuration is expected to be within a single line,\n"
+	       "                      values whitespace separated:\n"
+	       "\n"
+	       "                          <line in file> DstIPv4 SubNetMask NetIf DstMac\n"
```
JannePeltonen, line 295: I think it would be nicer to have <subnet>/<masklen> kind of syntax. I guess just adding the slash to the fscanf() string would do it.

test/performance/odp_ipsecfwd.c
```diff
@@ -0,0 +1,1307 @@
...
+
+	if (odp_ipsec_config(&ipsec_config) < 0) {
+		ODPH_ERR("Error configuring IPsec\n");
+		return false;
+	}
+
+	return true;
+}
+
+static odp_bool_t check_alg_support(int cipher_idx)
```
JannePeltonen, line 337: Maybe alg_idx instead of cipher_idx?

I am not sure how useful this support check and the concept of "exposed algorithms" is. Printing the ids in usage is useful, I think, but not allowing to use any algorithm that is not known by this test tool is not very helpful.

Now that this test SW does not know all algorithms, it is annoying that the unknown algorithms are disabled even though they should work just fine. And once all the known algorithms get added in the list, the check does not really serve any purpose.

It would be more useful to prevent usage of algorithms that are not supported by the underlying ODP, but there is no such capability check anywhere. odph_ipsec_alg_check() helper function may be useful in adding such check.

test/performance/odp_ipsecfwd.c
```diff
@@ -0,0 +1,1307 @@
...
+	}
+
+	if (odph_ipv4_addr_parse(&src_ip, src_ip_str) < 0 ||
+	    odph_ipv4_addr_parse(&dst_ip, dst_ip_str) < 0) {
+		ODPH_ERR("Error parsing IP addresses for SA %u\n", spi);
+		return;
+	}
+
+	if (spi > UINT16_MAX) {
+		ODPH_ERR("Invalid SPI value for SA %u\n", spi);
```
JannePeltonen, line 377: Maybe "unsupported" instead of "invalid" since in general SPI can be any 32 bit value?

test/performance/odp_ipsecfwd.c
```diff
@@ -0,0 +1,1307 @@
...
+		return;
+	}
+
+	if (spi_to_sa_map[dir][spi] != NULL) {
+		ODPH_ERR("Duplicate SA config %u\n", spi);
```
JannePeltonen, line 382: Maybe "non-unique SPIs not supported" as in general multiple outbound SAs with the same SPI are ok?

test/performance/odp_ipsecfwd.c
```diff
@@ -0,0 +1,1307 @@
...
+	sa_param.spi = spi;
+	sa_param.dest_queue = config->ev_qs[config->num_sas];
+
+	if (dir > 0U) {
+		sa_param.dir = ODP_IPSEC_DIR_OUTBOUND;
+		sa_param.outbound.tunnel.ipv4.src_addr = &src_ip;
+		sa_param.outbound.tunnel.ipv4.dst_addr = &dst_ip;
+	} else {
+		sa_param.dir = ODP_IPSEC_DIR_INBOUND;
+		sa_param.inbound.lookup_mode = ODP_IPSEC_LOOKUP_SPI;
```
JannePeltonen, line 412: This could be ODP_IPSEC_LOOKUP_DISABLED.

test/performance/odp_ipsecfwd.c
```diff
@@ -0,0 +1,1307 @@
...
+
+static void parse_sas(prog_config_t *config)
+{
+	odp_ipsec_capability_t ipsec_capa;
+	FILE *file;
+	int cipher_idx, auth_idx;
+	uint32_t dir, spi, icv_len;
+	char src_ip[15U] = { 0 }, dst_ip[15U] = { 0 };
+	uint8_t cipher_key[64U] = { 0U }, cipher_key_extra[4U] = { 0U }, auth_key[64U] = { 0U },
+	auth_key_extra[4U] = { 0U };
```
JannePeltonen, line 449: These arrays do not seem to have room for the terminating null. And the same in parse_fwd_table().

test/performance/odp_ipsecfwd.c
```diff
@@ -0,0 +1,1307 @@
...
+	}
+
+	if (config->num_ifs == 0U) {
+		ODPH_ERR("Invalid number of interfaces: %u, (min: 1, max: %u)\n", config->num_ifs,
+			 MAX_IFS);
+		return PRS_NOK;
+	}
+
+	if (config->num_sas == 0U) {
+		ODPH_ERR("Invalid number of SAs: %u, (min: 1, max: %u)\n", config->num_sas,
```
JannePeltonen, line 567: I think zero SAs could be allowed.

test/performance/odp_ipsecfwd.c
```diff
@@ -0,0 +1,1307 @@
...
+		pktio_param.out_mode = config->mode == 0U ? ODP_PKTOUT_MODE_QUEUE :
+							    ODP_PKTOUT_MODE_DIRECT;
+		pktio->handle = odp_pktio_open(pktio->name, config->pktio_pool, &pktio_param);
+
+		if (pktio->handle == ODP_PKTIO_INVALID) {
+			ODPH_ERR("Error opening packet I/O (%s)\n", pktio->name);
+			return false;
+		}
+
+		odp_pktin_queue_param_init(&pktin_param);
```
JannePeltonen, line 732: This sets pktin_param.queue_param.sched.sync to the default value ODP_SCHED_SYNC_PARALLEL, but I think we should use ODP_SCHED_SYNC_ORDERED.

test/performance/odp_ipsecfwd.c
```diff
@@ -0,0 +1,1307 @@
...
+		}
+
+		odp_pktio_config_init(&pktio_config);
+
+		if (odp_pktio_config(pktio->handle, &pktio_config) < 0) {
+			ODPH_ERR("Error configuring packet I/O extra options (%s)\n", pktio->name);
+			return false;
+		}
+
+		(void)odp_pktio_mac_addr(pktio->handle, &pktio->src_mac, sizeof(pktio->src_mac));
```
JannePeltonen, line 791: It would be good to check that this succeed so that we do not end up sending packets with a random source mac address.

test/performance/odp_ipsecfwd.c
```diff
@@ -0,0 +1,1307 @@
...
+	}
+
+	file = fopen(config->sa_conf_file, "r");
+
+	if (file == NULL) {
+		ODPH_ERR("Error opening SA configuration file: %s\n", strerror(errno));
+		return;
+	}
+
+	while (fscanf(file, "%u%u%s%s%d%s%s%d%s%s%u", &dir, &spi, src_ip, dst_ip,
```
JannePeltonen, line 471: This is a bit fragile for the string arguments. I think it would be better to include max field width in the string conversions (i.e. %u%u%14s%14s%d%63s%3s%d%63s%3s%u") if using sscanf and the current array sizes.

test/performance/odp_ipsecfwd.c
```diff
@@ -0,0 +1,1307 @@
...
+		if (odph_iplookup_table_put_value(config->fwd_tbl, &fwd_e->prefix, &fwd_e) < 0) {
+			ODPH_ERR("Error populating forwarding table\n");
+			return false;
+		}
+	}
+
+	return true;
+}
+
+static inline odp_ipsec_sa_t *get_in_sa(odp_packet_t pkt)
```
JannePeltonen, line 825: I though it was slightly odd that these return a pointer to the handle instead of the handle itself, but maybe it does not really matter.

test/performance/odp_ipsecfwd.c
```diff
@@ -0,0 +1,1307 @@
...
+
+	if (odp_packet_copy_to_mem(pkt, l3_off, ODPH_IPV4HDR_LEN, &ipv4) < 0)
+		return NULL;
+
+	ip = odp_be_to_cpu_32(ipv4.dst_addr);
+
+	if (odph_iplookup_table_get_value(fwd_tbl, &ip, &fwd, 0U) < 0 || fwd == NULL)
+		return NULL;
+
+	if (is_eth_mod) {
```
JannePeltonen, line 900: The is_eth_mod parameter here and the related passing of it through the call chain is a bit messy. And I do not think it buys you much. It basically tries to optimize for the plain forwarding case (without IPsec), which is not the focus here anyway, and does not really help much even in that case.

I would like to get rid of the parameter (or change it to ethertype parameter if really needed) and just write IPv4 ethertype in packets in all cases. If you are worried about putting an incorrect ethertype in the packet (e.g. IPv6 received), there is the same problem even after IPsec processing (IPv6 packets may come out of an IPv4 tunnel). You could check and drop all non-IPv4 traffic if you wanted but this is just a simple perf test tool, so maybe it is not needed.

If is_eth_mod goes away, then you would check if l3_off is exactly ODPH_ETHHDR_LEN and avoid the push_head in that case. If l3_off was not exactly ODPH_ETHHDR_LEN in the "is_eth_mod" case, the current code would not work anyway since extra bytes before l2 offset would be sent to the wire.

test/performance/odp_ipsecfwd.c
```diff
@@ -0,0 +1,1307 @@
...
+					int worker_id, odp_bool_t is_eth_mod)
+{
+	const pktio_t *pktio = lookup_and_apply(pkt, fwd_tbl, is_eth_mod);
+
+	if (pktio == NULL) {
+		++stats->discards;
+		return false;
+	}
+
+	if (odp_unlikely(pktio->send_fn(pktio, worker_id % pktio->num_tx_qs, pkt) < 0)) {
```
JannePeltonen, line 937: This leaks packets if odp_pktout_send() returns zero. The problem is that odp_pktout_queue() and odp_queue_enq() have a bit different return values but they are used directly in send_fn functions.

The queue index parameter so send_fn is a bit nasty since it is not used at all by enqueue() but probably is still calculated even in that case.

There is also the issue that with multiple output queues packet order gets lost. It is not clear if it is a bug or a feature. Preserving packet order with multiple output queues would be a bit complicated but could be done e.g. by selecting the queue based on traffic flow rather than thread ID (but that may not be enough since I think direct mode output even with single queue does not preserve packet order the same way as queue mode output).

test/performance/odp_ipsecfwd.c
```diff
@@ -0,0 +1,1307 @@
...
+	       "\n"
+	       "Usage: %s [options]\n"
+	       "\n"
+	       "  -i, --interfaces    Ethernet interfaces for packet I/O, comma-separated,\n"
+	       "                      no spaces.\n"
+	       "  -n, --num_pkts      Number of packet buffers allocated for packet I/O pool.\n"
+	       "                      %u by default.\n"
+	       "  -l, --pkt_len       Maximum size of packet buffers in packet I/O pool. %u by\n"
+	       "                      default.\n"
+	       "  -c, --count         CPU count, %u by default.\n"
```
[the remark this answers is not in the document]
TuomasTaipale: Almost all of our testers use "-c" option for passing a worker count, so I think it's better to be consistent and use "-c" also here.

test/performance/odp_ipsecfwd.c
TuomasTaipale: Fixed this and dropped the `goto` in v4.

test/performance/odp_ipsecfwd.c
TuomasTaipale: Sure, why not, in v4.

test/performance/odp_ipsecfwd.c
TuomasTaipale: v4.

test/performance/odp_ipsecfwd.c
TuomasTaipale: Dropped this entirely in v4. The listing created in help should be clear enough what is supported and what is not and if user gets adventurous with the indices it just leads to undefined behavior.

Added also the rest of the algorithms supported by spec.
```
...
  Supported cipher and authentication algorithms:
      Cipher algorithms:
          0: ODP_CIPHER_ALG_NULL (key lengths: 0, 0)
          2: ODP_CIPHER_ALG_3DES_CBC (key lengths: 24)
          4: ODP_CIPHER_ALG_AES_CBC (key lengths: 16, 24, 32)
          5: ODP_CIPHER_ALG_AES_CTR (key lengths: 16, 24, 32)
          9: ODP_CIPHER_ALG_AES_GCM (combined) (key lengths: 16, 24, 32)
          11: ODP_CIPHER_ALG_CHACHA20_POLY1305 (combined) (key lengths: 32)
      Authentication algorithms:
          0: ODP_AUTH_ALG_NULL (key/icv lengths: 0/0, 0/0)
          1: ODP_AUTH_ALG_MD5_HMAC (key/icv lengths: 16/12)
          2: ODP_AUTH_ALG_SHA1_HMAC (key/icv lengths: 20/12)
          4: ODP_AUTH_ALG_SHA256_HMAC (key/icv lengths: 32/16)
          5: ODP_AUTH_ALG_SHA384_HMAC (key/icv lengths: 48/24)
          6: ODP_AUTH_ALG_SHA512_HMAC (key/icv lengths: 64/32)
          7: ODP_AUTH_ALG_AES_GCM (combined) (key/icv lengths: 0/16)
          8: ODP_AUTH_ALG_AES_GMAC (key/icv lengths: 16/16, 24/16, 32/16)
          10: ODP_AUTH_ALG_AES_CMAC (key/icv lengths: 16/12, 24/12, 32/12)
          11: ODP_AUTH_ALG_AES_XCBC_MAC (key/icv lengths: 16/12)
          12: ODP_AUTH_ALG_CHACHA20_POLY1305 (combined) (key/icv lengths: 0/16)
...
```

test/performance/odp_ipsecfwd.c
TuomasTaipale: Error logs adjusted in v4.

test/performance/odp_ipsecfwd.c
TuomasTaipale: v4.

test/performance/odp_ipsecfwd.c
TuomasTaipale: v4.

test/performance/odp_ipsecfwd.c
TuomasTaipale: v4.

test/performance/odp_ipsecfwd.c
TuomasTaipale: Changed in v4 and now depends on the configured `mode`.

test/performance/odp_ipsecfwd.c
TuomasTaipale: v4.

test/performance/odp_ipsecfwd.c
TuomasTaipale: Some trickery is needed if a string ends up being longer than there is space defined in the format as the rest of that string in the file should then be skipped. I'll think this separately and leave this as is for now.

test/performance/odp_ipsecfwd.c
TuomasTaipale: Simplified in v4.

test/performance/odp_ipsecfwd.c
TuomasTaipale: Fixed in v4. Depending on configured `mode`, the output will be ordered (one output queue) or parallel with several queues. This enables testing a few different queue setups. Queue index is always calculated but the overhead isn't particularly large.

test/performance/odp_ipsecfwd.c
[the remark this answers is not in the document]
JannePeltonen: I was just commenting on the help text. The option could still be -c even if we say it is the count of worker threads.

test/performance/odp_ipsecfwd.c
JannePeltonen: Now with the syntax changed, this text should be updated. I also suggest using MaskLen instead of just Mask for the MaskLen Part. How about: IPv4Prefix/MaskLen?

test/performance/odp_ipsecfwd.c
JannePeltonen: The ODP API is perhaps a bit weird that it reuses the algorithm names defined in the crypto API for IPsec even though IPsec is defined only for the algorithms for which a "use-of-algorithm-x-with-ipsec" spec exists. IOW, some of these do not make sense with IPsec, at least not in any standard-defined way, but it is not easy to tell which unless you already know it.

It would be better to not show the undefined (for IPsec) algorithms. It makes things less confusing and the undefined ones would not be supported by any ODP anyway. These should be removed:

ALG_ENTRY(ODP_CIPHER_ALG_3DES_ECB, CIPHER_TYPE),
ALG_ENTRY(ODP_CIPHER_ALG_AES_CFB128, CIPHER_TYPE),
ALG_ENTRY(ODP_CIPHER_ALG_AES_XTS, CIPHER_TYPE),
ALG_ENTRY(ODP_CIPHER_ALG_KASUMI_F8, CIPHER_TYPE),
ALG_ENTRY(ODP_CIPHER_ALG_SNOW3G_UEA2, CIPHER_TYPE),
ALG_ENTRY(ODP_CIPHER_ALG_AES_EEA2, CIPHER_TYPE),
ALG_ENTRY(ODP_CIPHER_ALG_ZUC_EEA3, CIPHER_TYPE),
ALG_ENTRY(ODP_AUTH_ALG_KASUMI_F9, AUTH_TYPE),
ALG_ENTRY(ODP_AUTH_ALG_SNOW3G_UIA2, AUTH_TYPE),
ALG_ENTRY(ODP_AUTH_ALG_AES_EIA2, AUTH_TYPE),
ALG_ENTRY(ODP_AUTH_ALG_ZUC_EIA3, AUTH_TYPE),
ALG_ENTRY(ODP_AUTH_ALG_MD5, AUTH_TYPE),
ALG_ENTRY(ODP_AUTH_ALG_SHA1, AUTH_TYPE),
ALG_ENTRY(ODP_AUTH_ALG_SHA224, AUTH_TYPE),
ALG_ENTRY(ODP_AUTH_ALG_SHA256, AUTH_TYPE),
ALG_ENTRY(ODP_AUTH_ALG_SHA384, AUTH_TYPE),
ALG_ENTRY(ODP_AUTH_ALG_SHA512, AUTH_TYPE)

JannePeltonen: Now after adding the comment here, I realize that there is a capability check so the undefined ones are not actually displayed to the user. But I would still remove them from the code too.

test/performance/odp_ipsecfwd.c
JannePeltonen: Since this does not really matter, at least now, we can leave it as it is.

test/performance/odp_ipsecfwd.c
JannePeltonen: ok, this can be tweaked later if there is a need.

JannePeltonen:
A couple of minor comments still, but with those reasonably handled: Reviewed-by: Janne Peltonen <janne.peltonen@nokia.com>

JannePeltonen (approved):

PR #1645, opened 2022-08-02
[PATCH v1] api: pktio: add VLAN strip offload in packet input

JannePeltonen:
Add a pktin config flag that requests removal of VLAN tags of received
Ethernet frames so that the ODP application sees untagged frames. Add new
packet metadata and an associated API function for an application to query
the VLAN ID of the stripped VLAN tag.

### DIFF
--- a/include/odp/api/spec/packet_io_types.h
+++ b/include/odp/api/spec/packet_io_types.h
@@ -377,6 +377,29 @@ typedef union odp_pktin_config_opt_t {
 		  * on packet input */
 		uint64_t ts_ptp        : 1;
 
+		/** Strip VLAN tags of received packets so that the ODP
+		 *  application sees untagged packets. VLAN ID is saved
+		 *  in packet metadata and can be read using
+		 *  odp_packet_vlan_info().
+		 *
+		 *  This removes only the outermost VLAN tag and only if
+		 *  its ethertype is 0x8100.
+		 *
+		 *  VLAN stripping is done as a last step of packet input
+		 *  processing just before the application gets the packet.
+		 *  This means that packet parsing and classification is
+		 *  done with the VLAN tag still in the packet. The has_vlan
+		 *  and has_qinq packet flags are set at packet parsing
+		 *  before the stripping. Retained L2 headers in inline
+		 *  IPsec processing include the VLAN tag.
+		 *
+		 *  Support for VLAN strip offload must be checked from
+		 *  odp_pktio_capability_t::vlan_offload returned by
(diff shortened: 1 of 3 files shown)


kkirankk (commented):
Reviewed-by: Kiran Kumar K <kirankumark@marvell.com>

include/odp/api/spec/packet_types.h
```diff
@@ -304,6 +304,15 @@ typedef struct odp_packet_reass_partial_state_t {
...
 } odp_packet_reass_partial_state_t;
 
+/**
+ * Information about a stripped VLAN header
+ */
+typedef struct odp_packet_vlan_info_t {
+	/** VLAN ID in host byte order */
+	uint16_t vlan_id;
+
+} odp_packet_vlan_info_t;
```
TuomasTaipale, line 314: Should PCP and DEI be here as well?

JannePeltonen: They could be and the reason for having the structure was to be able to add those later in a basically backward compatible manner. I just did not include them yet. I wonder if I should.

One more thing that could perhaps be included at some point is the TPID (the ethertype of the stripped VLAN tag) which would become useful if possibility to strip other VLAN tag types (QinQ) gets added in the API.

TuomasTaipale: From client POV, if this is the only way to get information about a stripped VLAN tag (directly, I guess one could infer some of these from e.g. classification results), it would make sense to add also the rest of the data.

TuomasTaipale (approved):

include/odp/api/spec/packet_io_types.h
```diff
@@ -377,6 +377,29 @@ typedef union odp_pktin_config_opt_t {
...
 
+		/** Strip VLAN tags of received packets so that the ODP
+		 *  application sees untagged packets. VLAN ID is saved
+		 *  in packet metadata and can be read using
+		 *  odp_packet_vlan_info().
+		 *
+		 *  This removes only the outermost VLAN tag and only if
+		 *  its ethertype is 0x8100.
+		 *
+		 *  VLAN stripping is done as a last step of packet input
```
nithind1988, line 388: Discussed internally with Kiran and Jerin. 

Basically this statement is fine with us as for Inline IPsec, though the VLAN Tag will be available for parsing and classification, it will be retained in original packet where it got decrypted.

nithind1988: While for non Inline IPsec packets, it will be stripped post parsing.

JannePeltonen:

PR #1649, opened 2022-08-18
[PATCH v3] test: ipsecfwd: fix bugs

TuomasTaipale:
Fix two issues:

- Initialize configuration data structure earlier so that in certain
  error cases teardown phase would not see uninitialized data.
- Free SA and FWD config file path strings in teardown phase so that
  they don't leak if "--help" was passed with them.
- Handle `odp_ipsec_status()` return value in `wait_sas_disabled()`
  to avoid potentially uninitialized `odp_ipsec_status_t` access
  later.

v3:
- Added reviewed-by tag

### DIFF
--- a/test/performance/odp_ipsecfwd.c
+++ b/test/performance/odp_ipsecfwd.c
@@ -158,17 +158,17 @@ static exposed_alg_t exposed_algs[] = {
 static odp_ipsec_sa_t *spi_to_sa_map[2U][MAX_SPIS];
 static odp_atomic_u32_t is_running;
 
-static void terminate(int signal ODP_UNUSED)
-{
-	odp_atomic_store_u32(&is_running, 0U);
-}
-
 static void init_config(prog_config_t *config)
 {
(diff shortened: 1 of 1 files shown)


test/performance/odp_ipsecfwd.c
```diff
@@ -1213,6 +1214,9 @@ static void teardown_test(const prog_config_t *config)
 
 	if (config->compl_q != ODP_QUEUE_INVALID)
 		(void)odp_queue_destroy(config->compl_q);
+
+	free(config->sa_conf_file);
+	free(config->fwd_conf_file);
```
JannePeltonen, line 1219: Execution can end up here without ever going through init_config(), so the frees would be called with uninitialized pointers.

TuomasTaipale: `init_config()` was called too early also before (`odp_schedule_config()` could have failed), done earlier in v2.

JannePeltonen (approved):

PR #1650, opened 2022-08-18
[PATCH v3] examples: avoid buffering of output

sunilkori:
Currently some of the tests redirect the output to
a file and dumps the file contents at the end of the
test. Avoid this and tee the output to a file. And
whenever stdbuf utility is available, use it turn off
output buffering so that logs are displayed real time.

Also, dump generator logs at the end of the tests only in
case of error exits. And don't do cleanup if it is a
successful exit as cleanup would have been already done
for these cases.

### DIFF
--- a/example/generator/generator_run.sh
+++ b/example/generator/generator_run.sh
@@ -16,7 +16,13 @@ fi
 
 setup_interfaces
 
-./odp_generator${EXEEXT} -w 1 -n 1 -I $IF0 -m u
+if [ "$(which stdbuf)" != "" ]; then
+	STDBUF="stdbuf -o 0"
+else
+	STDBUF=
+fi
+
+$STDBUF ./odp_generator${EXEEXT} -w 1 -n 1 -I $IF0 -m u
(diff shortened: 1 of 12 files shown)


sunilkori:
@MatiasElo Please help me to run checks on this pull request.

sunilkori:
@MatiasElo Please look into it. Check needs to be executed on this pull request.

JannePeltonen:
- The commit message and the pull request description does not seem to match the change. The code change contains only the disabling of buffering using stdbuf.
- Why do you want to disable buffering? What is the reason this change should be made?
- Please do not have merge commits in a pull request but rebase your changes on the upstream master
- I am not sure 'which' should be used for checking the existence of commands. I think 'command' builtin would work better, especially if some of the sourced env srcipts e.g. create a function with that name. See https://unix.stackexchange.com/questions/85249/why-not-use-which-what-to-use-then
- You should be able to run the CI tests in your own github repo before creating a PR. You may need to change some of your github settings.

PR #1653, opened 2022-08-25
[PATCH v6] validation: crypto: add test vectors for zuc-256

asasidharan:
Add the test vectors for ZUC 256 EEA3 and ZUC 256 EIA3
algorithms.


test/validation/api/crypto/test_vectors.h
```diff
@@ -2527,7 +2527,129 @@ static crypto_test_reference_t zuc_eea3_reference[] = {
 				0xdf, 0x5a, 0x47, 0x3a, 0x57, 0xa4, 0xa0, 0x0d,
 				0x98, 0x5e, 0xba, 0xd8, 0x80, 0xd6, 0xf2, 0x38,
 				0x64, 0xa0, 0x7b, 0x01 }
-	}
+	},
+	{
+		.cipher_key_length = ZUC_EEA3_256_KEY_LEN,
```
JannePeltonen, line 2532: I am not sure if it is correct to support ZUC-256 through ODP_CIPHER_ALG_ZUC_EEA3 and  ODP_AUTH_ALG_ZUC_EIA3. The API spec says the algorithms are "ZUC based 128-EEA2" and "ZUC-based 128-EIA3", respectively.

Since the current API spec does not describe what the 256-bit variants of the algorithms really are, it is questionable if we can have test vectors for them.

So maybe an API change is needed, perhaps adding new algorithm names as EEA3 and EIA3 may not be appropriate.

test/validation/api/crypto/test_vectors.h
```diff
@@ -2527,7 +2527,129 @@ static crypto_test_reference_t zuc_eea3_reference[] = {
...
+				0xf5, 0x53, 0x52, 0x8d, 0x73, 0xfc, 0x5c, 0x6e,
+				0x09, 0x92, 0x1e, 0x35, 0x17, 0x60, 0xa8, 0xbb,
+				0x81, 0xf6, 0x21, 0x8f, 0x3e, 0x05, 0xe6, 0x0c,
+				0x60, 0xe7, 0x21, 0x53, 0x18, 0x63, 0x81, 0x0d,
+				0xb6, 0xd4, 0x9a, 0x29, 0xd0, 0xf6, 0x97, 0xd9,
+				0x89, 0xb5, 0x0e, 0xa0, 0x15, 0xb6, 0x5c, 0x97,
+				0xac, 0x7d, 0x26, 0xeb, 0x83, 0x0c, 0xf7, 0xe3,
+				0xf3, 0x18, 0x37, 0x0b, 0x7b, 0xb8, 0x18, 0x31,
+				0x8c, 0xb2, 0x5a, 0x5c, 0xa9, 0xf1, 0x35, 0x32 },
+		.ciphertext = { 0xa6, 0xe5, 0x71, 0x58, 0x5c, 0xcf, 0x5d, 0x0d,
```
JannePeltonen, line 2560: Where does this test data come from? It would be good to mention the source.

test/validation/api/crypto/test_vectors.h
asasidharan: > The API spec says the algorithms are "ZUC based 128-EEA2" and "ZUC-based 128-EIA3", respectively.

Would updating the API spec like "ZUC based 128/256 EEA3" address your concern?

test/validation/api/crypto/test_vectors.h
asasidharan: There was no standard vector available for the operation. So we generated the same with hardware. The same test vector is upstreamed in DPDK and it was validated on Intel on their PMDs. Intel had also verified the same with gmSSL library which is a derivate of openSSL project supporting these algos.

test/validation/api/crypto/test_vectors.h
JannePeltonen: Is there such a thing as "ZUC based 256 EEA3"  (or probably more correctly, ZUC-256 based EEA3)? What would be the defining document of the algorithm? ZUC-256 exists and 128-EEA3 using 128 bit ZUC exists, but I wonder if ZUC-256-EEA3 has a well defined meaning?

anoobj: @JannePeltonen Which one is supported by ODP? Is it ZUC-128 or 128-EEA3?

JannePeltonen: 128-EEA3 according to the API. But the API also says that IV should be formatted in a way specified in the EEA3 specification (bearer, direction etc), making one think that maybe the algorithm is really just ZUC and could be used for other contexts than EEA3 by selecting the IV freely (assuming it would be ok by the API, which depends on how seriously one takes the word "should" ).

I am not very familiar with ZUC, EEA3 and EIA3, but perhaps the API could provide simply ZUC instead of EEA3. Then adding ZUC-256 would not be a problem. But ZUC-256 based EIA3 (if it is a thing) could not be implemented efficiently by the application by just ZUC-256 provided by ODP.

Maybe it would be ok to have "ZUC based EEA & EIA" with both 128 and 256 bit keys in the API, but with some clarification on what the ZUC-256 based variants really do (e.g. regarding the IV setup)?

anoobj: > making one think that maybe the algorithm is really just ZUC and could be used for other contexts than EEA3 by selecting the IV freely
That's our impression as well. 

> but perhaps the API could provide simply ZUC instead of EEA3.
We can update the description to state that ODP API is simply ZUC and for EEA3, IV need to be prepared in the described manner.

>  ZUC-256 based variants really do (e.g. regarding the IV setup)
At this point, we don't have enough information on how IV setup would be done. ZUC-256 is a Nokia requirement. Do you have any docs that we can refer to state how IV generation would be done in case of EEA3 based on ZUC-256. If not, we can update documentation such that we will state IV generation for 128-EEA3 and leave rest to application's choice.

JannePeltonen: Please take a look at PR 1664 that adds a mention of ZUC-256 as an experimental algorithm that may be supported. If the API addition makes sense, then this validation test could be added as it is now.

test/validation/api/crypto/test_vectors.h
JannePeltonen: It would be good to have a short comment (e.g. privately generated test data) above the respective test vectors as now the overall comment in the zuc test reference refers to a 3GPP spec (Implementor's Test Data) and no longer applies to all the vectors.

The "Addendum to the ZUC-256 stream cipher" document has also some test data that could possibly be included if you supported the variant defined in the addendum so that you could test the test.

test/validation/api/crypto/test_vectors_len.h
```diff
@@ -121,6 +125,12 @@
 #define ZUC_EIA3_IV_LEN          16
 #define ZUC_EIA3_DIGEST_LEN      4
 
+/* ZUC_EIA3_256 */
+#define ZUC_EIA3_256_KEY_LEN       32
+#define ZUC_EIA3_256_IV_LEN        25
+#define ZUC_EIA3_256_DIGEST_LEN_8  8
+#define ZUC_EIA3_256_DIGEST_LEN_16 16
```
JannePeltonen, line 132: These digest length constants are not used anywhere.

test/validation/api/crypto/test_vectors.h
asasidharan: > It would be good to have a short comment (e.g. privately generated test data) above the respective test vectors

Added short comment "/* Privately generated test data */" for the newly added test vectors.


> The "Addendum to the ZUC-256 stream cipher" document has also some test data that could possibly be included if you supported the variant defined in the addendum so that you could test the test.

Will take it as a separate PR.

test/validation/api/crypto/test_vectors_len.h
asasidharan: Ack. Removed the unused macros.

JannePeltonen (approved):
This should be merged after PR 1664.

asasidharan:
> This should be merged after PR 1664.

Ack. Added Reviewed-by tag with v5.

PR #1654, opened 2022-08-29
[PATCH v6] linux-gen: sched: print full lines

psavol:
Print full lines from odp_schedule_print() call, so that log remain readable also when other threads print into it simultaneously.

### DIFF
--- a/platform/linux-generic/include/odp_print_internal.h
+++ b/platform/linux-generic/include/odp_print_internal.h
@@ -1,0 +1,22 @@
+/* Copyright (c) 2022, Nokia
+ * All rights reserved.
+ *
+ * SPDX-License-Identifier:     BSD-3-Clause
+ */
+
+#ifndef ODP_PRINT_INTERNAL_H_
+#define ODP_PRINT_INTERNAL_H_
+
+#ifdef __cplusplus
+extern "C" {
+#endif
+
+#include <stddef.h>
+
+int _odp_snprint(char *str, size_t size, const char *format, ...);
+
+#ifdef __cplusplus
+}
+#endif
+
+#endif
--- a/platform/linux-generic/odp_schedule_basic.c
+++ b/platform/linux-generic/odp_schedule_basic.c
@@ -37,6 +37,7 @@
 #include <odp_global_data.h>
 #include <odp_event_internal.h>
 #include <odp_macros_internal.h>
+#include <odp_print_internal.h>
 
 #include <string.h>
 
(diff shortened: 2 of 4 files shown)


platform/linux-generic/odp_schedule_basic.c
```diff
@@ -2053,42 +2055,42 @@ static void schedule_print(void)
 	ODP_PRINT("  prefer ratio:      %u\n", sched->config.prefer_ratio);
 	ODP_PRINT("\n");
 
-	ODP_PRINT("  Number of active event queues:\n");
-	ODP_PRINT("              spread\n");
-	ODP_PRINT("            ");
+	pos  = 0;
+	pos += ODP_SNPRINT(&str[pos], size - pos, "  Number of active event queues:\n");
```
JannePeltonen, line 2059: I think the code needed in the users of buffered logging gets cluttered and fragile. Like this in the scheduler patch:

```
int pos;
const int size = 512;
char str[size]:
...
pos = 0;
pos += ODP_SNPRINT(&str[pos], size - pos, "  Number of active event queues:\n");
pos += ODP_SNPRINT(&str[pos], size - pos, "              spread\n");
pos += ODP_SNPRINT(&str[pos], size - pos, "            ");
...
ODP_PRINT("%s\n", str);
```

Wouldn't something like this be much better:
```
_odp_print_buffered("  Number of active event queues:\n");
_odp_print_buffered("              spread\n");
_odp_print_buffered("            ");
...
_odp_print_flush();
```

Or, if one wants to reduce the number of code changes needed, there could be a new function to enable buffered logging and then the existing ODP_PRINT() calls would change to buffered ones and only the flushing would be needed.

The implementation of the above schemes would have a thread-local buffer that the caller would not need to manage or much care about.

psavol: Sure. However, I'd consider that as future development on top of this PR. This adds a safer way to use snprinf() (we have over 200 calls to snprintf(), many without return value check), and fixes an immediate issue of fragmented output from sched print.

platform/linux-generic/include/odp_print_internal.h
```diff
@@ -0,0 +1,62 @@
...
+	/* Error. Ensure that string has the end mark */
+	if (len < 0) {
+		str[0] = 0;
+		return 0;
+	}
+
+	/* Print would have been longer. Ensure that string has the end mark and
+	 * return number of characters printed. */
+	if (len >= (int)size) {
+		str[size - 1] = 0;
```
JereLeppanen, line 51: So if I understood correctly, in this case *snprintf() should null terminate the string, so this line is not needed.

MatiasElo (approved):
Looks good. Helper functions for buffered output can be added to the same `odp_print_internal.h` header later in a separate PR . `_odp_snprint()` is useful even after adding buffered output functions, when the print destination is not ODP log.

platform/linux-generic/include/odp_print_internal.h
psavol: Fixed in v4.

PR #1655, opened 2022-08-29
[PATCH v5] test: ipsecfwd: make queue counts configurable

TuomasTaipale:
This patchset adds options to `odp_ipsecfwd` to configure input, SA completion and output queue counts.

Additionally, SA configuration error handling is now more pragmatic and queue naming is improved.

v3:
- Added hashing based on outgoing packet to choose output queue

v4:
- Increased maximum number of suported SAs
- Added reviewed-by tags

### DIFF
--- a/test/performance/odp_ipsecfwd.c
+++ b/test/performance/odp_ipsecfwd.c
@@ -29,12 +29,12 @@
 #define MAX_SAS 64U
 #define MAX_FWDS 64U
 #define MAX_SPIS (UINT16_MAX + 1U)
-#define MAX_WORKERS (ODP_THREAD_COUNT_MAX - 1U)
-#define MAX_QUEUES 16U
+#define MAX_WORKERS (ODP_THREAD_COUNT_MAX - 1)
+#define MAX_QUEUES 64U
 #define PKT_SIZE 1024U
 #define PKT_CNT 32768U
-#define WORKER_CNT 1U
 #define NUM_EVS 32U
+#define ORDERED 0U
 
 #define ALG_ENTRY(_alg_name, _type) \
 	{ \
@@ -66,7 +66,7 @@ typedef struct pktio_s pktio_t;
 typedef struct pktio_s {
 	union {
 		odp_pktout_queue_t out_dir_qs[MAX_QUEUES];
-		odp_queue_t out_ev_q;
+		odp_queue_t out_ev_qs[MAX_QUEUES];
 	};
 
 	odph_ethaddr_t src_mac;
@@ -114,6 +114,9 @@ typedef struct prog_config_s {
 	odph_table_t fwd_tbl;
 	odp_barrier_t init_barrier;
 	odp_barrier_t term_barrier;
+	uint32_t num_input_qs;
+	uint32_t num_sa_qs;
+	uint32_t num_output_qs;
 	uint32_t num_pkts;
 	uint32_t pkt_len;
 	uint32_t num_ifs;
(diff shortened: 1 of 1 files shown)


test/performance/odp_ipsecfwd.c
```diff
@@ -561,13 +591,18 @@ static parse_result_t check_options(prog_config_t *config)
...
 
 	if (config->num_ifs == 0U) {
-		ODPH_ERR("Invalid number of interfaces: %u, (min: 1, max: %u)\n", config->num_ifs,
+		ODPH_ERR("Invalid number of interfaces: %u (min: 1, max: %u)\n", config->num_ifs,
 			 MAX_IFS);
 		return PRS_NOK;
 	}
 
+	if (config->sa_conf_file != NULL && config->num_sas == 0U) {
+		ODPH_ERR("Invalid SA configuration\n");
```
JannePeltonen, line 606: > Additionally, modify SA configuration error checks so that if SA configuration file is passed, then at least one valid SA configuration is expected.

This was slightly annoying, at least to me, as I tried to comment out the last SA of the config file and the test SW protested. I would just remove the check since it is not for safety.

test/performance/odp_ipsecfwd.c
```diff
@@ -681,9 +728,9 @@ static odp_bool_t send(const pktio_t *pktio, uint8_t index, odp_packet_t pkt)
 	return odp_pktout_send(pktio->out_dir_qs[index], &pkt, 1) == 1;
 }
 
-static odp_bool_t enqueue(const pktio_t *pktio, uint8_t index ODP_UNUSED, odp_packet_t pkt)
+static odp_bool_t enqueue(const pktio_t *pktio, uint8_t index, odp_packet_t pkt)
 {
-	return odp_queue_enq(pktio->out_ev_q, odp_packet_to_event(pkt)) == 0;
+	return odp_queue_enq(pktio->out_ev_qs[index], odp_packet_to_event(pkt)) == 0;
```
JannePeltonen, line 739: Since index is CPU ID, this causes reordering of forwarded packets in the ordered mode, defeating much of the purpose of ordered scheduling. There should at least be an option (which should also be the default) to select output queue based on some sort of flow hash (something like xoring the last bytes of the source and destination IP address of the packet would probably be good for this test program).

test/performance/odp_ipsecfwd.c
TuomasTaipale: Not for safety, but might be less confusing for new users that config file full of invalid data leads to program termination instead of startup. SA config file option can be omitted in case SA config is not needed.

test/performance/odp_ipsecfwd.c
TuomasTaipale: Modified in v2. Something like this, perhaps?

test/performance/odp_ipsecfwd.c
```diff
@@ -29,12 +30,12 @@
 #define MAX_SAS 64U
```
JannePeltonen, line 30: Now that dest queue count is separately configurable, the max number of SAs could be increased significantly. Something like 4k would seem good to me.

test/performance/odp_ipsecfwd.c
```diff
@@ -590,9 +631,9 @@ static parse_result_t check_options(prog_config_t *config)
 	if (config->pkt_len == 0U)
 		config->pkt_len = MIN(pool_capa.pkt.max_len, PKT_SIZE);
 
-	if (config->num_thrs <= 0 || config->num_thrs > (int)MAX_QUEUES) {
-		ODPH_ERR("Invalid CPU count: %d, (min: 1, max: %u)\n", config->num_thrs,
-			 MAX_QUEUES);
+	if (config->num_thrs <= 0 || config->num_thrs > MAX_WORKERS) {
+		ODPH_ERR("Invalid CPU count: %d (min: 1, max: %d)\n", config->num_thrs,
```
JannePeltonen, line 635: Now that the -c option is called thread count, we could replace "CPU" with "thread" here too.

JannePeltonen:
Made a couple of minor comments, if you would like to still make those changes.

JannePeltonen (approved):

PR #1658, opened 2022-09-02
[PATCH v3] helper: cli: check for vsnprintf errors

JereLeppanen:
If vsnprintf() fails, log an error and return error status.

v2:
- Rebase.
- Add review tags.

### DIFF
--- a/helper/cli.c
+++ b/helper/cli.c
@@ -206,7 +206,7 @@ static int cli_log_va(odp_log_level_t level, const char *fmt, va_list in_args)
 	(void)level;
 
 	va_list args;
-	char *str, *p, *last;
+	char *str = NULL, *p, *last;
 	int len;
 
 	/*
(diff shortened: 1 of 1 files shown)


helper/cli.c
```diff
@@ -221,8 +221,15 @@ static int cli_log_va(odp_log_level_t level, const char *fmt, va_list in_args)
...
+	len = vsnprintf(NULL, 0, fmt, args);
 	va_end(args);
+
+	if (len < 0) {
+		ODPH_ERR("vsnprintf failed\n");
+		goto out;
+	}
+
+	len++;
 	str = malloc(len);
```
TuomasTaipale, line 233: `str = malloc(++len);`?

TuomasTaipale (approved):
One minor comment, otherwise 👍.

helper/cli.c
JereLeppanen: That would be equivalent, yes. But I want it to be on display on its own line, it's too important to hide inside parenthesis in the malloc call.

PR #1659, opened 2022-09-02
[PATCH v5] linux-gen: cls: mark implemented capabilities enabled

TuomasTaipale:
This patchset marks implemented but missing IPsec SPI PMR and class-of-service hash protocol capabilities enabled in classification. Additionally, add validation tests for `ODP_PMR_IPSEC_SPI` variants.

v2:
- Added valid values to AH and ESP header fields when creating test packets

v3:
- Fixed AH header length usage
- Added reviewed-by tags

v4:
- Removed extra endianness swap in `verify_pmr_ipsec_spi()` and changed validation tests accordingly

v5:
- Added reviewed-by tags

### DIFF
--- a/test/validation/api/classification/odp_classification_common.c
+++ b/test/validation/api/classification/odp_classification_common.c
@@ -26,6 +26,7 @@ static uint8_t IPV6_DST_ADDR[ODPH_IPV6ADDR_LEN] = {
 };
 
 #define ODP_GTPU_UDP_PORT 2152
+#define AH_HDR_LEN 24
 
 odp_pktio_t create_pktio(odp_queue_type_t q_type, odp_pool_t pool,
 			 odp_bool_t cls_enable)
@@ -306,6 +307,8 @@ odp_packet_t create_packet(cls_packet_info_t pkt_info)
 	odph_ipv6hdr_t *ipv6;
 	odph_gtphdr_t *gtpu;
 	odph_igmphdr_t *igmp;
+	odph_ahhdr_t *ah;
+	odph_esphdr_t *esp;
 	uint8_t *hlen = 0;
 	uint16_t payload_len;
 	uint32_t addr = 0;
@@ -358,6 +361,14 @@ odp_packet_t create_packet(cls_packet_info_t pkt_info)
 	case CLS_PKT_L4_IGMP:
 		next_hdr = ODPH_IPPROTO_IGMP;
 		l4_hdr_len = ODP_IGMP_HLEN;
+		break;
+	case CLS_PKT_L4_AH:
+		next_hdr = ODPH_IPPROTO_AH;
+		l4_hdr_len = AH_HDR_LEN;
+		break;
+	case CLS_PKT_L4_ESP:
+		next_hdr = ODPH_IPPROTO_ESP;
+		l4_hdr_len = ODPH_ESPHDR_LEN;
 		break;
 	default:
 		ODPH_ASSERT(0);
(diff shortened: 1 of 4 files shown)


test/validation/api/classification/odp_classification_common.c
```diff
@@ -496,6 +508,15 @@ odp_packet_t create_packet(cls_packet_info_t pkt_info)
 			ODPH_ERR("odph_udp_tcp_chksum failed\n");
 			return ODP_PACKET_INVALID;
 		}
+	} else if (pkt_info.l4_type == CLS_PKT_L4_AH) {
+		ah->next_header = ODPH_IPV4;
+		ah->ah_len = ODPH_AHHDR_LEN;
```
JannePeltonen, line 513: This is not quite correct (even though it would work here). The length value is the number of 32-bit words minus 2 and must be divisible by 2 with IPv6. I suggest setting ah_len to 4 so that an implementation that assumes correctly formatted AH packets does not get upset.

test/validation/api/classification/odp_classification_common.c
```diff
@@ -496,6 +508,15 @@ odp_packet_t create_packet(cls_packet_info_t pkt_info)
...
+		ah->pad = 0;
+		ah->spi = 0;
```
JannePeltonen, line 515: Maybe the default value could be e.g. 256 (zero is reserved for local use and should not appear in the wire and the rang 1-255 is reserved for future use). Then maybe the actual SPI used in the negative test could also be something else than 1 (e.g. val + 1). And the same comment for ESP.

test/validation/api/classification/odp_classification_common.c
```diff
@@ -496,6 +508,15 @@ odp_packet_t create_packet(cls_packet_info_t pkt_info)
...
+		ah->seq_no = 0;
```
JannePeltonen, line 516: It would be better to set this to 1 (and likewise for ESP) since zero sequence number should never appear in the wire, which some ODP implementation might assume.

test/validation/api/classification/odp_classification_test_pmr.c
```diff
@@ -1873,6 +1873,76 @@ static void classification_test_pmr_term_custom_l3(void)
...
+	val = 0x11223344;
+	mask = 0xffffffff;
+
+	odp_cls_pmr_param_init(&pmr_param);
+	pmr_param.term = ODP_PMR_IPSEC_SPI;
+	pmr_param.match.value = &val;
+	pmr_param.match.mask = &mask;
+	pmr_param.val_sz = sizeof(val);
+
+	pkt_info = default_pkt_info;
```
JannePeltonen, line 1929: It would be good to have these tests for ESP and AH over IPv6 too.

test/validation/api/classification/odp_classification_common.c
```diff
@@ -359,6 +361,14 @@ odp_packet_t create_packet(cls_packet_info_t pkt_info)
 		next_hdr = ODPH_IPPROTO_IGMP;
 		l4_hdr_len = ODP_IGMP_HLEN;
 		break;
+	case CLS_PKT_L4_AH:
+		next_hdr = ODPH_IPPROTO_AH;
+		l4_hdr_len = ODPH_AHHDR_LEN;
```
JannePeltonen, line 366: Now I noticed that the l4_hdr_len here does not match the ah_len field below. l4_hdr_len here is 12 but ah_len indicates 24-byte AH header. So the ODP under test might think that the AH packet is truncated before it processes the classification rules.

There is a similar problem with ESP in that l4_hdr_len does not include the trailer (plus padding and/or actual payload) that must be in the packet, so an ODP implementation might think the ESP packet is invalid (as no real ESP packet can be that short) and stop processing it further. The l4_hdr_len here (or resulting packet size that gets calculated) should be 4 bytes longer than now.

This is getting slightly messy and might also be unnecessary, but I guess it would be good to have the test packets to look like valid packets.

JannePeltonen (approved):

test/validation/api/classification/odp_classification_test_pmr.c
```diff
@@ -1873,6 +1873,98 @@ static void classification_test_pmr_term_custom_l3(void)
...
+	cls_packet_info_t pkt_info;
+	odp_packet_t pkt;
+	odph_ahhdr_t *ah;
+
+	val = 0x11223344;
+	mask = 0xffffffff;
+
+	odp_cls_pmr_param_init(&pmr_param);
+	pmr_param.term = ODP_PMR_IPSEC_SPI;
+	pmr_param.match.value = &val;
```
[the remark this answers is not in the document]
TuomasTaipale: Fixed in v4. Unnecessary swap seems to have been present in `verify_pmr_ipsec_spi()`.

psavol (approved):

JannePeltonen:
Looks good to me too.

PR #1660, opened 2022-09-05
[PATCH v10] api: pool: user area in buffer, timeout and vector events

psavol:
Packets have already user area for storing user specific metadata, which is preserved while packet is modified or stored into a queue, etc. Introduce user area to other event types (that are allocated from a pool) as well.

### DIFF
--- a/include/odp/api/spec/packet.h
+++ b/include/odp/api/spec/packet.h
@@ -1492,6 +1492,31 @@ void *odp_packet_user_area(odp_packet_t pkt);
 uint32_t odp_packet_user_area_size(odp_packet_t pkt);
 
 /**
+ * Check user flag
+ *
+ * Implementation clears user flag during new packet creation (e.g. alloc and packet input)
+ * and reset. User may set the flag with odp_packet_user_flag_set(). Implementation never
+ * sets the flag, only clears it. The flag may be useful e.g. to mark when the user area
+ * content is valid.
+ *
+ * @param pkt   Packet handle
+ *
+ * @retval 0    User flag is clear
+ * @retval !0   User flag is set
+ */
+int odp_packet_user_flag(odp_packet_t pkt);
+
+/**
+ * Set user flag
+ *
+ * Set (or clear) the user flag.
+ *
+ * @param pkt   Packet handle
+ * @param val   New value for the flag. Zero clears the flag, other values set the flag.
+ */
+void odp_packet_user_flag_set(odp_packet_t pkt, int val);
+
+/**
  * Layer 2 start pointer
  *
  * Returns pointer to the start of layer 2. Optionally, outputs number of data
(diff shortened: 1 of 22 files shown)


psavol:
v2: added missing implementations

include/odp/api/spec/pool_types.h
```diff
@@ -456,7 +453,7 @@ typedef struct odp_pool_param_t {
 
 		/** User area size in bytes. The maximum value is defined by
 		 *  pool capability pkt.max_uarea_size. Specify as 0 if no user
-		 *  area is needed.
+		 *  area is needed. The default value is 0.
```
ashwinyes, line 471: Why to force default value as 0? It should be settable to any value in odp_pool_param_init().

include/odp/api/spec/packet.h
```diff
@@ -1491,6 +1491,31 @@ void *odp_packet_user_area(odp_packet_t pkt);
  */
```
ashwinyes, line 1491: I think this API is not necessary. User can simply set a pattern in the user area to check whether it is valid or not.

A single bit flag does not yield much benefit to the user in my opinion.

psavol: When an application receives packets (from the scheduler), some are new packets from packet input HW and some are packet enqueued by the application. Application has set user area in those packets it has already processed, but those from packet input have random user area content.

Application can use this flag to check if user area content is random, or initialized properly.

Instead of the flag, we could specify e.g. that implementation always zeros the entire user area (or its first X bytes) on packet input/alloc/reset. But that would require implementation to write that specific memory area. This flag can locate anywhere in the packet header, and may be cleared automatically by HW (with any other metadata fields).

include/odp/api/spec/pool_types.h
psavol: Maybe we could update the text a bit and say: "The minimum user area size in bytes...". So that  implementation is free to round it up to a larger value. Maybe that would simplify implementation of odp_packet_user_area_size().

Zero as default value minimizes memory usage, as applications that do not care about user area would require zero bytes automatically.

cawallen (approved):

include/odp/api/spec/packet.h
ashwinyes: Ack

include/odp/api/spec/pool_types.h
ashwinyes: >> "The minimum user area size in bytes..." 
Adding this line is fine.

psavol: User area parameter spec text updated (for all four event types) in v5.

platform/linux-generic/odp_buffer.c
```diff
@@ -23,6 +23,13 @@ uint32_t odp_buffer_size(odp_buffer_t buf)
 	return pool->seg_len;
 }
 
+void *odp_buffer_user_area(odp_buffer_t buf)
```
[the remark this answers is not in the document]
psavol: Yes, left inline implementation for later (as this PR is large already).

platform/linux-generic/include/odp_event_vector_internal.h
```diff
@@ -34,6 +36,9 @@ typedef struct ODP_ALIGNED_CACHE odp_event_vector_hdr_t {
 	/* Event vector size */
 	uint32_t size;
 
+	/* Flags */
+	_odp_event_vector_flags_t flags;
```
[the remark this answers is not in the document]
psavol: Fixed in v6. Free call zeros it.

platform/linux-generic/odp_packet.c
```diff
@@ -2524,3 +2524,10 @@ odp_proto_stats_t odp_packet_proto_stats(odp_packet_t pkt)
 
 	return ODP_PROTO_STATS_INVALID;
 }
+
+void odp_packet_user_flag_set(odp_packet_t pkt, int val)
```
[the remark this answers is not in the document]
psavol: Done in v6.

platform/linux-generic/include/odp_event_vector_internal.h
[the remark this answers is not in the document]
MatiasElo: The flags should likely be initialized also in `init_event_hdr()`.

MatiasElo (approved):

PR #1661, opened 2022-09-07
[PATCH v1] helper: threads: update odph_thread_create() specification

MatiasElo:
Update `odph_thread_create()` specification to state that memory for output
thread table must be allocated from SHM memory in process mode. This is
required for child process startup synchronization.

### DIFF
--- a/helper/include/odp/helper/threads.h
+++ b/helper/include/odp/helper/threads.h
@@ -1,5 +1,5 @@
 /* Copyright (c) 2013-2018, Linaro Limited
- * Copyright (c) 2019-2021, Nokia
+ * Copyright (c) 2019-2022, Nokia
  * All rights reserved.
  *
  * SPDX-License-Identifier:     BSD-3-Clause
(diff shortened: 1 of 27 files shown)


helper/include/odp/helper/threads.h
```diff
@@ -241,9 +241,10 @@ void odph_thread_common_param_init(odph_thread_common_param_t *param);
  * Use odph_thread_common_param_init() and odph_thread_param_init() to
  * initialize parameters with default values.
  *
- * Thread table must be large enough to hold 'num' elements. Also the cpumask
+ * Thread table must be large enough to hold 'num' elements. Also, the cpumask
  * must contain 'num' CPUs. Threads are pinned to CPUs in order - the first
- * thread goes to the smallest CPU number of the mask, etc.
+ * thread goes to the smallest CPU number of the mask, etc. When creating Linux
+ * processes, memory for the output thread table must be allocated from ODP SHM.
```
JannePeltonen, line 247: It seems that there are also assumptions that the virtual address of the thread table is the same in all ODP threads/processes and that the thread table is accessible in the newly started threads/processes before odp_init_local() is called. I wonder if it is strictly ok, even though it would probably require a bit weird ODP SHM implementation for it not to work.

MatiasElo: The assumption has been that SHM memory reserved before fork is shareable between processes, but the SHM spec is quite ambiguous about this. The helper spec could be  changed to e.g. `... must be allocated from ODP SHM with appropriate SHM flags / ODP_SHM_SINGLE_VA.`, but I'm not sure if this is strictly necessary.

The point about using SHM memory before calling `odp_init_local()` is also quite tricky. Many of our current example applications wouldn't work if the implementation didn't support this.

PR #1664, opened 2022-09-13
[PATCH v3] api: crypto: add experimental ZUC-256 support

JannePeltonen:
Mention that ODP_CIPHER_ALG_ZUC_EEA3 and ODP_AUTH_ALG_ZUC_EIA3 may support 256 bit key length but that the support is experimental and subject to change. The standardization status of the algorithms is open and it is not clear what the defining specifications are. There are are already two different initialization schemes with different IV lengths defined.

Mention NEA3 and NIA3 in addition to EEA3 and EIA3 in the API as they are identical, respectively.


anoobj:

PR #1665, opened 2022-09-13
api: crypto: clarify current API and add new "operation types"

JannePeltonen:
This PR has the following patches:

- api: crypto: clarify that input packet is fully copied to output packet
    - no functional changes but better description on the expected functionality

- api: crypto: introduce a session parameter to control output packet handling
    - add "operation type" session parameter with three values:
        - legacy: backward compatible behaviour
        - copy: output packet allocated by ODP, copied from input packet and processed
        - combine: crypto output of input packet is written into a second packet (a copy of it), preserving other data and metadata of the second packet

- api: crypto: do not always consume input packets
    - add a per-packet flag to tell if a processed input packet should not be consumed but returned untouched.


JannePeltonen:
v2: fixed a couple of typos.

JannePeltonen:
v3: Split the second commit to two commits so that the addition of ODP_CRYPTO_OP_TYPE_COMBINE is now separate from the addition of the crypto op type. There are no actual code changes.

include/odp/api/spec/crypto.h
```diff
@@ -688,8 +722,11 @@ typedef struct odp_crypto_session_param_t {
 	/** Output pool
 	 *
 	 *  When the output packet is not specified during the call to
-	 *  crypto operation, the output packet will be allocated
-	 *  from this pool.
+	 *  crypto operation in the legacy operation type, the output
+	 *  packet will be allocated from this pool.
+	 *
+	 *  In ODP_CRYPTO_OP_TYPE_COPY operation types this must be set to
```
anoobj: When op_type is 'ODP_CRYPTO_OP_TYPE_COPY' which is the pool to be used for packet allocation?

JannePeltonen: It is the same pool as the input packet. See the documentation of odp_crypto_op() which says it.

anoobj: Thanks for the confirmation. So, the implementation is expected to get the pool from the input packet in datapath and then allocate new buffer?

Is the expectation to use odp_packet_pool() for getting the pool?

JannePeltonen: Conceptually, the implementation checks the pool of the input packet (e.g. using odp_packet_pool), allocates the output packet from that pool and then copies all data and metadata from the input packet to the output packet. Finally the implementation frees the input packet (except if the application has requested the input packet to be preserved, which I think is not the typical case). In practice the implementation could simply reuse the input packet (if the input packet is not to be preserved) and the end result would be the same. But the API does not promise that the input packet gets reused, so it is up to the implementation to reuse it or not.

anoobj: When input packet needs to be preserved, implementation is expected to do odp_packet_pool() and allocate from the same, right?

JannePeltonen: Yes (except, of course, the implementation can figure out the pool by any internal means, without calling ODP API functions).

JannePeltonen:
v4:
- Dropped the combine operation type and the input packet preservation from this PR.
- Renamed ODP_CRYPTO_OP_TYPE_COPY to ODP_CRYPTO_OP_TYPE_BASIC.
- Rebased

JannePeltonen:
The combine operation type and input packet preservation patches (rebased on this PR) can be found in crypto-api-impr-2 branch for reference.

include/odp/api/spec/crypto.h
```diff
@@ -1408,12 +1437,32 @@ int odp_crypto_result(odp_crypto_packet_result_t *result,
...
+ *
+ * The offset of the crypto range and auth range in the output packet is
+ * the same as in the input packet, adjusted by dst_offset_shift operation
+ * parameter.
+ *
+ * pkt_out[n] must be a valid handle to a packet that is long enough to
+ * contain the shifted crypto range, auth range and, in encode sesstions,
+ * the MAC/digest result. pkt_out[n] must not be the same as the input
+ * packet.
+ *
```
psavol, line 1459: Two typos above: partuicular, sesstions

JannePeltonen: Fixed in v6.

JannePeltonen:
v6: fixed typos

JannePeltonen:
v7: rebased

JannePeltonen:
v8: fixed whitespace errors introduced during rebase

anoobj:
API changes look fine.

@JannePeltonen Do you have plans to introduce unit tests?

JannePeltonen:
v9:
- Clarify that the OOP mode works as if crypto-range and auth range were copied to the output packet and then the output packet was processed
- Clarify that output packets in the OOP mode must be separate
- Add input packet (in the OOP mode) in the result provided by odp_crypto_result() and make it clear that the input packet may not be used until the operation is complete
- Add a patch that allows post processing in odp_crypto_result() by requiring that the result function is called before packet data of the output packet can be assumed to be valid

JannePeltonen:
v10: clarified the text regarding changing memory layout and removed the ambiguous "data offset" term.

include/odp/api/spec/crypto_types.h
```diff
@@ -986,6 +1015,12 @@ typedef struct odp_crypto_packet_result_t {
...
 	odp_bool_t  ok;
 
+	/** Input packet passed to odp_crypo_op_enq() when the operation
+	 *  type of the session is ODP_CRYPTO_OP_TYPE_OOP. In other cases
+	 *  this field does not have a valid value.
+	 */
+	odp_packet_t pkt_in;
+
 	/** Cipher status */
 	odp_crypto_op_status_t cipher_status;
```
anoobj, line 1025: Should we make it such that this filed & auth_status would only have valid values when crypto operation is not OK? Then we can skip updating the field.

Other option is to have a union of both cipher status & auth_status,

union crypto_result_status {
    struct status {
        uint8_t cipher_status;
        uint8_t auth_status;
    } s;
    uint16_t u16;
};

JannePeltonen: We could, but since that is a backward-incompatible change, we could other changes as well, like deprecating the separate HW-error fields and maybe packing the cipher and auth error statuses tighter (not a union but something that can hold both). But all that is should probably be in a separate PR or at least in a separate set of patches.

anoobj: Agreed. This change increases the size of the structure. So, isn't this already backwards-incompatible change? Anyway, please consider reducing the size of the structure when we work on improving the result mechanism.

anoobj:
API change looks fine.

include/odp/api/spec/crypto.h
```diff
@@ -316,12 +316,35 @@ int odp_crypto_result(odp_crypto_packet_result_t *result,
...
+ * output packet passed through pkt_out[n]. Input packets are not consumed
+ * nor modified. Memory layout (including packet data pointers, head and
+ * tail room, segmentation) of the output packet may change during the
+ * operation.
+ *
+ * Crypto output is the processed crypto_range, auth_range and
+ * MAC/digest (in encode sessions) of the input packet. The operation
+ * behaves as if crypto range and auth range were first copied from the
+ * input packet to the output packet and then the crypto operation
+ * was applied to the output packet.
```
JannePeltonen, line 331: I am afraid there might be slight ambiguity here in a couple cases:

1) If a cipher or auth algorithm is null (the other algorithm may or may not be null), is the corresponding cipher or auth range copied to the packet? The current text implies that it is, and it would be kind-of consistent and maybe also handy at times, but I wonder if we should specify that the range of the null algorithm is ignored for this purpose.

2) Algorithms providing authenticated encryption ignore the auth_range parameter. Maybe we should state that it is ignored also for the purpose of the conceptual pre-crypto data copy in this OOP case.

include/odp/api/spec/crypto_types.h
JannePeltonen: Merely adding fields in a structure is considered backward compatible in ODP, since it does not break reasonably sensibly written code. The field is never needed by old code that does not know about the OOP operation type. It is similar to adding new optionally-filled fields (with good defaults) in some config structure: well written code does not break, but code that relies on the physical size and/or padding details of the structure can break (but it could then break with a new compiler too).

include/odp/api/spec/crypto.h
anoobj: > The current text implies that it is, and it would be kind-of consistent and maybe also handy at times, but I wonder if we should specify that the range of the null algorithm is ignored for this purpose.

There are two cases here,
1. For cipher only cases etc, auth would be NULL and do we want copy to happen based on auth range? I guess we should not do that. And so, it is fair to say that the range is ignored.
2. For NULL cipher + NULL auth, do we need to copy? If yes, based on which range?

> Algorithms providing authenticated encryption ignore the auth_range parameter. Maybe we should state that it is ignored also for the purpose of the conceptual pre-crypto data copy in this OOP case.

Is that the case? At least for PDCP, I'm aware of a case where authenticated range and ciphered range would be different.

include/odp/api/spec/crypto_types.h
anoobj: Understood.

include/odp/api/spec/crypto.h
JannePeltonen: > For cipher only cases etc, auth would be NULL and do we want copy to happen based on auth range? I guess we should not do that.

Then I think an API update is needed. If the range would be copied even with null algorithm, one could still avoid the copy by setting range length to zero. But I think just ignoring the range might be better.

> For NULL cipher + NULL auth, do we need to copy? If yes, based on which range?

I do not think null+null should be a special case. If null range is ignored when the other algorithm is not null, then in this case both ranges would be ignored and nothing would get copied. OTOH, if null range is not ignored, then in this null-null case both ranges would get copied.

I think I will update the API a bit regarding this after discussing internally. Maybe I also add clarification on how hash_result_offset is interpreted in the case of OOP with a shift (shift is not applied, but whether the offset applies to the input packet or output packet depends on the direction (encode vs. decode).

anoobj: Agreed.

JannePeltonen:
v11:

This PR now depends on the commits of PR 1743 (removal of deprecated per-session IVs) and includes those commits as the first five commits. Those commits are not part of the review in this PR and will disappear as soon as they get merged to master and this PR then gets rebased.

Changes in V11:
- API changes regarding the out-of-place op type: clarify that dst_offset_shift is in bytes, specify that ignored crypto and auth ranges are not copied in the output packet, specify that the bytes in hash result location are not modified in the input packet but may be undefined in the corresponding location of the output packet if the hash result location overlaps the cipher or auth ranges.
- Added implementations: basic type is implemented by all crypto modules. OOP type only by the openssl module and in a very simple way.

JannePeltonen:
v12
- fixed casting warnings on 32-bit systems
- added commits that revamps validation test code a bit and add tests for the new operation types

JannePeltonen:
v14: rebased to get rid of the prerequisite commits that are now in master

psavol:
Reviewed API spec changes in v14. Those are OK.

include/odp/api/spec/crypto.h
```diff
@@ -363,6 +363,10 @@ int odp_crypto_op(const odp_packet_t pkt_in[],
  * creation on the packets. Behaves otherwise like odp_crypto_op() but
  * returns output packets through events.
  *
+ * With other operation types than ODP_CRYPTO_OP_TYPE_LEGACY, packet
```
anoobj, line 366: How about

"With operation types other than..."

JannePeltonen: Yes, that is better english. Will change in the next version.

test/validation/api/crypto/odp_crypto_test_inp.c
```diff
@@ -564,40 +504,168 @@ static int prepare_input_packet(const alg_test_param_t *param,
...
-	uint8_t *expected_bytes;
+#define MAX_IGNORED_RANGES 3
 
-	/* crypto output is undefined if authentication fails */
-	if (param->wrong_digest)
+/*
+ * Output packet parts that we ignore since they have undefined values
+ */
+typedef struct ignore_t {
+	uint32_t byte_offset;	/* offset to a byte which has bits to be ignored */
```
anoobj, line 573: With bit mode, are we testing non-byte aligned memory also?

This may require a capability. MRVL doesn't support non-byte aligned and as per PDCP non-byte aligned is not required.

JannePeltonen: No, we are not testing non-byte aligned ranges in bit mode. ODP API requires that the starting offset of cipher range or auth range is byte aligned even with algorithms that operate in bit mode. This what the API spec says:

```
         * Note: data buffer MUST start on the byte boundary, using offset
         * which is not divisible by 8 is unsupported and will result in
         * unspecified behaviour.
```

The length of the range in bit mode algorithms does not need to be a multiple of 8 bits. In that case the bits of the last byte that are undefined after the operation, which is that this byte_offset and byte_mask in ingore_t are trying to handle.

anoobj: > The length of the range in bit mode algorithms does not need to be a multiple of 8 bits.
MRVL does not support such operations. Can this be protected with some capability check?

anoobj: Thanks Janne for clarification. That capability flag should do. We will make adjustments if required.

platform/linux-generic/odp_crypto_openssl.c
```diff
@@ -2711,6 +2724,113 @@ int crypto_int(odp_packet_t pkt_in,
...
+ * with shifting by dst_offset_shift.
+ */
+static void copy_ranges(odp_packet_t dst,
+			odp_packet_t src,
+			const odp_crypto_generic_session_t *session,
+			const odp_crypto_packet_op_param_t *param)
+{
+	odp_packet_data_range_t c_range = param->cipher_range;
+	odp_packet_data_range_t a_range = param->auth_range;
+	uint32_t shift = param->dst_offset_shift;
```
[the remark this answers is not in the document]
JannePeltonen: Changed the type to signed in v15 but did not yet add any checks (if we start validating packet op input, it might be good to check the ranges etc and with all operation types. I am thinking about making a separate PR about that).

JannePeltonen:
v15: rebased, added review-tags in the API commits, made the grammar improvement in the API sentence, changed the type of the variable to signed (does not change the outcome).

JannePeltonen:
v16: updated review-tags

PR #1673, opened 2022-09-26
[PATCH v4] test: ipsecfwd: add support for burst processing

TuomasTaipale:
Add burst support to packet handling. This enables passing more than a single packet at a time to IPsec processing pipelines, potentially increasing tester throughput.

v2:
- Added a retry logic to IPsec in/out enqueuing

v3:
- Inlined IPsec in/out enqueuing retry logic

v4:
- Rebased
- Added reviewed-by tag

### DIFF
--- a/test/performance/odp_ipsecfwd.c
+++ b/test/performance/odp_ipsecfwd.c
@@ -35,7 +35,7 @@
 #define MAX_SA_QUEUES 1024U
 #define PKT_SIZE 1024U
 #define PKT_CNT 32768U
-#define NUM_EVS 32U
+#define MAX_BURST 32U
 #define ORDERED 0U
 
 #define ALG_ENTRY(_alg_name, _type) \
@@ -59,7 +59,7 @@ typedef enum {
 } parse_result_t;
 
 enum {
-	DIR_IN = 0U,
+	DIR_IN = 0,
 	DIR_OUT
 };
 
@@ -162,6 +162,7 @@ static exposed_alg_t exposed_algs[] = {
 /* SPIs for in and out directions */
 static odp_ipsec_sa_t *spi_to_sa_map[2U][MAX_SPIS];
 static odp_atomic_u32_t is_running;
+static const int ipsec_out_mark;
 
 static void init_config(prog_config_t *config)
 {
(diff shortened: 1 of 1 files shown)


test/performance/odp_ipsecfwd.c
```diff
@@ -986,110 +989,157 @@ static inline const pktio_t *lookup_and_apply(odp_packet_t pkt, odph_table_t fwd
...
 	}
 
-	return;
+	if (num_pkts_ips > 0) {
+		num_procd = process_ipsec_in(pkts_ips, sas, num_pkts_ips);
+
+		if (odp_unlikely(num_procd < num_pkts_ips)) {
+			num_procd = num_procd < 0 ? 0 : num_procd;
+			stats->ipsec_in_errs += num_pkts_ips - num_procd;
+			odp_packet_free_multi(&pkts_ips[num_procd], num_pkts_ips - num_procd);
```
JannePeltonen, line 1089: Maybe the enqueuing should not be given up so easily. An ODP implementation may have a limited maximum burst size that it can handle so that it only consumes at most that many packets even if the offered burst size is longer. This is problematic especially with the fixed burst size in this test app.

Enqueuing of the remaining packets could be retried as long as odp_ipsec_XXX_enq() consumed some packets.

test/performance/odp_ipsecfwd.c
```diff
@@ -1119,19 +1169,20 @@ static void drain_events(void)
...
 	odph_table_t fwd_tbl = config->prog_config->fwd_tbl;
 	stats_t *stats = &config->stats;
 
 	odp_barrier_wait(&config->prog_config->init_barrier);
 
 	while (odp_atomic_load_u32(&is_running)) {
-		cnt = odp_schedule_multi_no_wait(NULL, evs, NUM_EVS);
+		int num_pkts_in = 0, num_pkts_ips = 0;
+
+		cnt = odp_schedule_multi_no_wait(NULL, evs, MAX_BURST);
```
JannePeltonen, line 1185: The max burst size should be configurable to allow testing without bursts.

Actually, it would be good to have the ipsec enqueue/dequeue burst size configurable separately from scheduling burst size to allow testing limited ipsec burst sizes without affecting scheduling (e.g. to run the kind of test which this test app did before these changes). Maybe that could also be left to a separate patch.

test/performance/odp_ipsecfwd.c
TuomasTaipale: Yes, good point. Added a retry logic in v2.

test/performance/odp_ipsecfwd.c
TuomasTaipale: This could be a future improvement, added a TODO for it in v2.

test/performance/odp_ipsecfwd.c
JannePeltonen: The logic is good but the macro is pretty ugly and has the usual macro problem with arguments getting evaluated multiple times. An inline function would be much better, but you could also make single inlined process_ipsec funtion that would handle both the inbound and outbound case based on a parameter passed and the compiler would just optimize the constant parameter and the conditionals away.

e.g. process_ipsec_in() could then be like this:
static inline int process_ipsec_in(odp_packet_t pkts[], const odp_ipsec_sa_t sas[], int num)
{
	return process_ipsec(DIR_IN, pkts, sas, num);
}

TuomasTaipale: Sure, macro is a bit ugly though a single function here doesn't work without type erasure (which is also ugly) as operation parameters are of different type. "Umbrella" functions that do different things based on input boolean flags are not that nice either.

Maybe I'll just inline the `while()` logic to both in and out processing functions.

Also just noticed a bug: need to also adjust the `num_sa` and `sa` fields of operation params according to processed packets.

JannePeltonen: > Maybe I'll just inline the while() logic to both in and out processing functions.

I think that would be fine too.

JannePeltonen (approved):

PR #1678, opened 2022-09-30
[PATCH v3] linux-gen: sched: inline fast path api functions

MatiasElo:


TuomasTaipale (approved):

platform/linux-generic/odp_schedule_if.c
```diff
@@ -68,56 +71,6 @@ int odp_schedule_config(const odp_schedule_config_t *config)
 	return ret;
 }
 
-odp_event_t odp_schedule(odp_queue_t *from, uint64_t wait)
-{
-	ODP_ASSERT(odp_global_rw->schedule_configured);
```
psavol, line 73: PR 1674 allows maintaining this check. So, let's get that in first, and then rebase this PR (with assert maintained).

MatiasElo: Asserts added in V3.

PR #1690, opened 2022-10-17
[PATCH v3] linux-gen: cls: performance improvements

JereLeppanen:
A few small performance improvements to packet classifier.

Branched from PR #1681.

v2:
- Rebase on master.
- Petri's comments.

v3:
- Rebase.
- Fix linked_cos.
- Add review tags.

### DIFF
--- a/platform/linux-generic/odp_classification.c
+++ b/platform/linux-generic/odp_classification.c
@@ -1557,6 +1557,8 @@ static inline void pmr_debug_print(pmr_t *pmr, cos_t *cos)
  */
 static cos_t *match_pmr_cos(cos_t *cos, const uint8_t *pkt_addr, odp_packet_hdr_t *hdr)
 {
+	pmr_t *pmr_match = NULL;
+
 	while (1) {
 		uint32_t i, num_rule = odp_atomic_load_u32(&cos->num_rule);
 
@@ -1570,18 +1572,13 @@ static cos_t *match_pmr_cos(cos_t *cos, const uint8_t *pkt_addr, odp_packet_hdr_
 			if (verify_pmr(pmr, pkt_addr, hdr)) {
 				/* PMR matched */
 
+				pmr_match = pmr;
 				cos = linked_cos;
 
 				pmr_debug_print(pmr, cos);
 
 				if (cos->stats_enable)
 					odp_atomic_inc_u64(&cos->stats.packets);
-
-				hdr->p.input_flags.cls_mark = 0;
-				if (pmr->mark) {
-					hdr->p.input_flags.cls_mark = 1;
-					hdr->cls_mark = pmr->mark;
-				}
 
 				break;
 			}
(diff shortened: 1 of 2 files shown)


platform/linux-generic/odp_classification.c
```diff
@@ -1555,56 +1555,48 @@ static inline void pmr_debug_print(pmr_t *pmr, cos_t *cos)
...
-	if (!cos->valid)
-		return NULL;
+		for (i = 0; i < num_rule; i++) {
+			pmr_t *pmr = cos->pmr[i];
 
-	if (verify_pmr(pmr, pkt_addr, hdr)) {
-		/* PMR matched */
-		pmr_debug_print(pmr, cos);
+			if (odp_unlikely(!cos->linked_cos[i]->valid))
+				continue;
```
psavol, line 1570: This error should never happen, but if it does may be linked cos could be NULL as well. Instead continue, could this be instead: assert(cos->linked_cos[i] != NULL && cos->linked_cos[i]->valid)

platform/linux-generic/odp_classification.c
```diff
@@ -1555,56 +1555,48 @@ static inline void pmr_debug_print(pmr_t *pmr, cos_t *cos)
...
+					odp_atomic_inc_u64(&cos->stats.packets);
 
-		for (i = 0; i < num_rule; i++) {
-			cos_t *retcos = match_pmr_cos(cos->linked_cos[i],
-						      pkt_addr, cos->pmr[i],
-						      hdr);
+				hdr->p.input_flags.cls_mark = 0;
+				if (pmr->mark) {
+					hdr->p.input_flags.cls_mark = 1;
+					hdr->cls_mark = pmr->mark;
```
psavol, line 1585: Instead of recording mark into packet (possibly many times), it should be more efficient to just keep record of last matching pmr pointer, and write mark once in the end of the function.

platform/linux-generic/odp_classification.c
```diff
@@ -1555,56 +1555,48 @@ static inline void pmr_debug_print(pmr_t *pmr, cos_t *cos)
...
- * with the packet.
+ * This function performs a depth-first search in the CoS tree.
  */
-static cos_t *match_pmr_cos(cos_t *cos, const uint8_t *pkt_addr, pmr_t *pmr,
-			    odp_packet_hdr_t *hdr)
+static cos_t *match_pmr_cos(cos_t *cos, const uint8_t *pkt_addr, odp_packet_hdr_t *hdr)
 {
-	uint32_t i, num_rule;
+	if (!cos || !cos->valid)
+		return cos;
```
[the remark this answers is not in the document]
JereLeppanen: v2.

platform/linux-generic/odp_classification.c
```diff
@@ -1555,56 +1555,45 @@ static inline void pmr_debug_print(pmr_t *pmr, cos_t *cos)
...
-			odp_atomic_inc_u64(&cos->stats.packets);
+			if (verify_pmr(pmr, pkt_addr, hdr)) {
+				/* PMR matched */
 
-		hdr->p.input_flags.cls_mark = 0;
-		if (pmr->mark) {
-			hdr->p.input_flags.cls_mark = 1;
-			hdr->cls_mark = pmr->mark;
-		}
+				cos = cos->linked_cos[i];
```
psavol, line 1575: Current implementation has race between updating PMR/linked_cos and reading those here. We'll need to fix that later (with atomic read of pmr+linked_cos). However, I'd suggest to read cos->linked_cos only once here to minimize issues where linked_cos->valid and cos=linked_cos would refer to different linked_cos pointers.

JereLeppanen: v3.

PR #1692, opened 2022-10-19
[PATCH v3] test: cunit: provide option to run tests on control thread

nithind1988:
By default run all the tests on worker thread and
provide option to run tests on control thread.

This is needed because control threads will be slower and have less performance w.r.t fast path API.
For example, in our case, control thread will only support packet burst of 1 which would fail pktio_main test
case(pktio_test_recv_queue()) which tries to do burst enqueue of packets.

### DIFF
--- a/test/common/odp_cunit_common.c
+++ b/test/common/odp_cunit_common.c
@@ -36,6 +36,7 @@ static int allow_skip_result;
 static odph_thread_t thread_tbl[ODP_THREAD_COUNT_MAX];
 static int threads_running;
 static odp_instance_t instance;
+static bool control_thread;
 static char *progname;
 static int (*thread_func)(void *);
 
@@ -260,6 +261,7 @@ static int tests_global_init(odp_instance_t *inst)
 {
 	odp_init_t init_param;
 	odph_helper_options_t helper_options;
+	odp_thread_type_t thr_type;
 
 	if (odph_options(&helper_options)) {
 		fprintf(stderr, "error: odph_options() failed.\n");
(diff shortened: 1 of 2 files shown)


test/common/odp_cunit_common.c
```diff
@@ -707,9 +711,18 @@ int odp_cunit_register(odp_suiteinfo_t testsuites[])
 int odp_cunit_parse_options(int argc, char *argv[])
 {
 	const char *env = getenv("CI");
+	int i;
 
 	progname = argv[0];
-	odph_parse_options(argc, argv);
+	argc = odph_parse_options(argc, argv);
```
MatiasElo, line 717: It could be simpler to integrate this to CI by using an environment variable. There is already the `CI` variable, so for example `CI_CONTROL_THREAD`/`CI_THREAD_TYPE_CONTROL`/`CI_DEFAULT_THREAD_TYPE_CONTROL` could be used.

MatiasElo:
_For example, in our case, control thread will only support packet burst of 1 which would fail pktio_main test
case(pktio_test_recv_queue()) which tries to do burst enqueue of packets._

This seems like an issue in the test itself (can be fixed in a separate PR).

test/common/odp_cunit_common.c
nithind1988: Ok.. Can I change the default type to WORKER which can be overridden using this ENV variable ?

nithind1988: And there are lot of tests test/validation/api/scheduler/, timer, pktio
which do lot of multi thread perf tests.

So I think we need to change the default with worker thread.

MatiasElo: Yes, changing the default type to worker is OK. We can then add a separate test case to CI, which changes the type to CONTROL using an env variable.

MatiasElo (approved):

MatiasElo:
Hi @nithind1988, please add reviewed-by tags so this PR can be merged.

PR #1693, opened 2022-10-20
[PATCH v2] test: perf: l2fwd: fix more dependency to odp_generator

JereLeppanen:
The previous fix to this does not work when ODP is configured with --without-examples and built out-of-tree (i.e. configure is run in some other directory than ODP base directory).

Fix by adding a recipe to explicitly go to the generator directory to build it.

v2:
- Re-word subject line.
- Add review tags.


MatiasElo (approved):

MatiasElo (approved):
`more` could be dropped from the commit subject line.

JereLeppanen:
> `more` could be dropped from the commit subject line.

Slightly re-worded in v2. I wanted to avoid using the exact same subject line as in the earlier commit.

PR #1694, opened 2022-10-20
[PATCH v1] linux-gen: crypto: integrate Arm-optimized Multi-buffer Crypto for IPSec Library

Malvika-Gupta:
The first commit adds the support for detecting the presence of the Arm-optimized Multi-buffer Crypto for IPSec Library and enabling it if present
The second and third commits integrate the ZUC128 and ZUC256 EEA3 and EIA3 algorithms to ODP


MatiasElo:
Replaced by https://github.com/OpenDataPlane/odp/pull/1703.

PR #1698, opened 2022-10-24
[PATCH v4] test: packet_gen: add option for latency measurements

TuomasTaipale:
This patchset adds option to measure latency with `odp_packet_gen`. Sent packets can have a timestamp in their payload which is then utilized at reception time to calculate latency. Additionally:

  - option to disable packet reference usage (enables latency measurements),
  - option to disable UDP checksum calculations,
  - option to disable payload filling

are added.

v2:
- Matias' comments
- Added a new patch where error/warning logs are printed with `ODPH_ERR()`
- Added a new patch where new options are printed in startup configuration summary dump

v3:
- Matias' comments

v4:
- Rebased
- Added reviewed-by tag

### DIFF
--- a/test/performance/odp_packet_gen.c
+++ b/test/performance/odp_packet_gen.c
@@ -63,6 +63,7 @@ typedef struct test_options_t {
 	uint32_t wait_sec;
 	uint32_t wait_start_sec;
 	uint32_t mtu;
+	odp_bool_t use_refs;
 	odp_bool_t promisc_mode;
 
 	struct vlan_hdr {
@@ -167,11 +168,15 @@ static void print_usage(void)
 	       "  -L, --len_range <min,max,bins>\n"
 	       "                            Random packet length. Specify the minimum and maximum\n"
 	       "                            packet lengths and the number of bins. To reduce pool size\n"
-	       "                            requirement the length range can be divined into even sized\n"
+	       "                            requirement the length range can be divided into even sized\n"
 	       "                            bins. Min and max size packets are always used and included\n"
 	       "                            into the number of bins (bins >= 2). Bin value of 0 means\n"
 	       "                            that each packet length is used. Comma-separated (no spaces).\n"
 	       "                            Overrides standard packet length option.\n"
+	       "  -R, --no_pkt_refs         Do not use packet references. Always allocate a\n"
+	       "                            fresh set of packets for a transmit burst. Some\n"
+	       "                            features may be available only with references\n"
+	       "                            disabled.\n"
 	       "  -M, --mtu <len>           Interface MTU in bytes.\n"
 	       "  -b, --burst_size          Transmit burst size. Default: 8\n"
 	       "  -x, --bursts              Number of bursts per one transmit round. Default: 1\n"
@@ -256,7 +261,8 @@ static int parse_options(int argc, char *argv[], test_global_t *global)
 		{"num_tx",      required_argument, NULL, 't'},
 		{"num_pkt",     required_argument, NULL, 'n'},
 		{"len",         required_argument, NULL, 'l'},
-		{"len_range",    required_argument, NULL, 'L'},
+		{"len_range",   required_argument, NULL, 'L'},
+		{"no_pkt_refs", no_argument,       NULL, 'R'},
 		{"burst_size",  required_argument, NULL, 'b'},
 		{"bursts",      required_argument, NULL, 'x'},
 		{"gap",         required_argument, NULL, 'g'},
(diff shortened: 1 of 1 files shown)


test/performance/odp_packet_gen.c
```diff
@@ -1307,34 +1317,46 @@ static inline int send_burst(odp_pktout_queue_t pktout, odp_packet_t pkt[],
...
-			break;
+		if (use_refs) {
+			out_pkt[i] = odp_packet_ref_static(pkt[idx]);
+			if (odp_unlikely(out_pkt[i] == ODP_PACKET_INVALID)) {
+				num = i;
+				break;
+			}
+		} else {
+			out_pkt[i] = pkt[idx];
+			pkt[idx] = ODP_PACKET_INVALID;
```
MatiasElo, line 1427: Is this invalidation necessary? Looks like one could simply use `odp_packet_free_multi()` below to free the possible "extra" packets.

test/performance/odp_packet_gen.c
```diff
@@ -1032,6 +1068,27 @@ static int close_pktios(test_global_t *global)
...
+static inline void get_timestamp(odp_packet_t pkt, uint32_t ts_off, rx_lat_data_t *lat_data)
+{
+	ts_data_t ts_data;
+	uint64_t nsec;
+
+	if (odp_unlikely(odp_packet_copy_to_mem(pkt, ts_off, sizeof(ts_data), &ts_data) < 0 ||
+			 odp_be_to_cpu_64(ts_data.magic) != TS_MAGIC))
+		return;
+
+	nsec = odp_time_global_ns() - odp_be_to_cpu_64(ts_data.tx_ts);
```
MatiasElo, line 1097: The endianness conversions could be dropped for application internal data to minimize overhead.

test/performance/odp_packet_gen.c
```diff
@@ -560,6 +591,11 @@ static int parse_options(int argc, char *argv[], test_global_t *global)
 		ret = -1;
 	}
 
+	if (test_options->calc_latency && test_options->use_refs) {
+		printf("Error: Latency calculations are not supported with packet references\n");
```
MatiasElo, line 612: `ODPH_ERR()` could be used.

test/performance/odp_packet_gen.c
```diff
@@ -183,11 +214,18 @@ static void print_usage(void)
 	       "  -o, --udp_src             UDP source port. Default: 10000\n"
 	       "  -p, --udp_dst             UDP destination port. Default: 20000\n"
 	       "  -P, --promisc_mode        Enable promiscuous mode.\n"
+	       "  -a, --latency             Calculate latency. Not available with packet\n"
+	       "                            references (see \"--no_pkt_refs\").\n"
```
MatiasElo, line 218: Perhaps it would be simpler to a user that enabling latency measurement would disable references automatically. For example: `Automatically disables packet references (see --no_pkt_refs).`

MatiasElo:
It could be useful to print also the new command line options (done currently inside `open_pktios()`).

test/performance/odp_packet_gen.c
TuomasTaipale: Yes, `idx` might be random.

test/performance/odp_packet_gen.c
TuomasTaipale: Sure, I guess there are no valid use-cases where the swap would be needed. V2.

test/performance/odp_packet_gen.c
TuomasTaipale: `printf()` was the used style. Added a new patch in v2 where all errors/warnings are changed to `ODPH_ERR()`.

test/performance/odp_packet_gen.c
TuomasTaipale: V2.

test/performance/odp_packet_gen.c
```diff
@@ -1032,6 +1066,27 @@ static int close_pktios(test_global_t *global)
...
+static inline void get_timestamp(odp_packet_t pkt, uint32_t ts_off, rx_lat_data_t *lat_data)
+{
+	ts_data_t ts_data;
+	uint64_t nsec;
+
+	if (odp_unlikely(odp_packet_copy_to_mem(pkt, ts_off, sizeof(ts_data), &ts_data) < 0 ||
+			 ts_data.magic != TS_MAGIC))
+		return;
+
+	nsec = odp_time_global_ns() - ts_data.tx_ts;
```
MatiasElo, line 1086: If a burst of packets is received, a more accurate method for time stamping would be to save the time once right after receiving the events and then using the same RX time for all packets in the burst. This would also have less overhead.

The same style could also be used in the TX side to reduce overhead, although it could strictly speaking be a bit  less accurate.

TuomasTaipale: Saving time once in RX side in V3. For better accuracy, no point then making it less accurate in TX side, overhead of the timestamping seems negligible.

test/performance/odp_packet_gen.c
```diff
@@ -36,6 +36,13 @@
 /* Minimum number of packets to receive in CI test */
 #define MIN_RX_PACKETS_CI 800
 
+/* Identifier for payload-timestamped packets */
+#define TS_MAGIC 0xff88ee99ddaaccbb
+
+#define SECS   1000000000
+#define MILLIS 1000000
+#define MICROS 1000
```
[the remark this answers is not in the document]
TuomasTaipale: V3.

MatiasElo (approved):

PR #1704, opened 2022-10-27
[PATCH v4] api: packet: support additional l3 and l4 protocol types

MatiasElo:
Move odp_proto_l2_type_t, odp_proto_l3_type_t, and odp_proto_l4_type_t types and defines from ABI to API header. L3 protocol type define values have been changed to match actual EtherType values. The size of odp_proto_l3_type_t is increased to accommodate this.

Allow odp_packet_l3_type() and odp_packet_l4_type() functions to return also other protocol types than the ones defined in ODP_PROTO_L3_TYPE_* and ODP_PROTO_L4_TYPE_* defines. This enables implementations to support additional protocols without having to add each one to the ODP API.

### DIFF
--- a/include/odp/api/spec/packet.h
+++ b/include/odp/api/spec/packet.h
@@ -1685,6 +1685,11 @@ odp_proto_l2_type_t odp_packet_l2_type(odp_packet_t pkt);
  *
  * Returns layer 3 protocol type. Initial type value is ODP_PROTO_L3_TYPE_NONE.
  *
+ * In addition to protocol types specified in ODP_PROTO_L3_TYPE_* defines,
+ * the function may also return other L3 protocol types (e.g. from IEEE
+ * EtherTypes list) recognized by the parser. If protocol type is not
+ * recognized, ODP_PROTO_L3_TYPE_NONE is returned.
+ *
  * @param      pkt      Packet handle
  *
  * @return Layer 3 protocol type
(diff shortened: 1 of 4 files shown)


JereLeppanen (approved):

kkirankk (commented):
Will custom L3 protocols will be part of ODP_PROTO_L3_TYPE_NONE?

include/odp/api/spec/packet.h
```diff
@@ -1685,6 +1685,11 @@ odp_proto_l2_type_t odp_packet_l2_type(odp_packet_t pkt);
  *
  * Returns layer 3 protocol type. Initial type value is ODP_PROTO_L3_TYPE_NONE.
  *
+ * In addition to protocol types specified in ODP_PROTO_L3_TYPE_* defines,
+ * packet parser implementation may support parsing additional L3 protocol
+ * types. If protocol type is not recognized by parser, ODP_PROTO_L3_TYPE_NONE
+ * is returned.
```
psavol, line 1691: To be clear, we could mention here something like: "Returned protocol types values are from IEEE EtherType list, also those protocols that are not listed by ODP API."

include/odp/api/spec/packet.h
```diff
@@ -1696,6 +1701,11 @@ odp_proto_l3_type_t odp_packet_l3_type(odp_packet_t pkt);
  *
  * Returns layer 4 protocol type. Initial type value is ODP_PROTO_L4_TYPE_NONE.
  *
+ * In addition to protocol types specified in ODP_PROTO_L4_TYPE_* defines,
+ * packet parser implementation may support parsing additional L4 protocol
+ * types. If protocol type is not recognized by parser, ODP_PROTO_L4_TYPE_NONE
+ * is returned.
```
psavol, line 1707: Same comment here: "... values from IANA list ..."

MatiasElo: Changed in V2.

MatiasElo:
Hi @kkirankk, does the current API look OK to you?

psavol (approved):

PR #1706, opened 2022-10-28
[PATCH v7] api: memory usage debug info

psavol:
Adds new system and SHM API functions which can be used to monitor/debug ODP instance memory usage.

v2: updated documentation of odp_shm_reserve() operation without flags (3rd commit)
v3: added system meminfo implementation and test
v4: added odp_shm_segment_info() implementation and test
v6: Moved IOVA/PA_INVALID defines to ABI file. Use defines in implementation and test.

### DIFF
--- a/include/odp/api/spec/shared_memory.h
+++ b/include/odp/api/spec/shared_memory.h
@@ -40,8 +40,6 @@ extern "C" {
  * Maximum shared memory block name length in chars including null char
  */
 
- /* Shared memory flags */
-
 /**
  * @def ODP_SHM_IOVA_INVALID
  * Invalid IOVA address
@@ -217,22 +215,25 @@ int odp_shm_capability(odp_shm_capability_t *capa);
  * Reserve a contiguous block of shared memory
  *
  * Reserve a contiguous block of shared memory that fulfills size, alignment
- * and shareability (ODP_SHM_* flags) requirements. In general, a name is
- * optional and does not need to be unique. However, if the block will be
+ * and shareability (ODP_SHM_* flags) requirements. By default (without flags), the memory
+ * block can be shared between all threads of the ODP instance. Memory address (odp_shm_addr())
+ * shareability depends on application memory model and #ODP_SHM_SINGLE_VA flag usage.
+ *
+ * Name is optional and does not need to be unique. However, if the block will be
  * searched with odp_shm_lookup() or odp_shm_import(), a unique name is needed
  * for correct match.
  *
- * @param name   Name of the block or NULL. Maximum string length is
- *               ODP_SHM_NAME_LEN.
+ * @param name   Name of the block or NULL. Maximum string length is ODP_SHM_NAME_LEN.
  * @param size   Block size in bytes
  * @param align  Block alignment in bytes
  * @param flags  Shared memory parameter flags (ODP_SHM_*). Default value is 0.
  *
  * @return Handle of the reserved block
  * @retval ODP_SHM_INVALID on failure
+ *
+ * @see odp_mem_model_t
  */
-odp_shm_t odp_shm_reserve(const char *name, uint64_t size, uint64_t align,
-			  uint32_t flags);
+odp_shm_t odp_shm_reserve(const char *name, uint64_t size, uint64_t align, uint32_t flags);
 
 /**
  * Free a contiguous block of shared memory
@@ -293,9 +294,7 @@ void *odp_shm_addr(odp_shm_t shm);
 /**
  * Shared memory block info
  *
- * Get information about the specified shared memory block. This is the only
- * shared memory API function which accepts invalid shm handles (any bit value)
- * without causing undefined behavior.
+ * Get information about the specified shared memory block.
  *
  * @param      shm   Block handle
  * @param[out] info  Block info pointer for output
--- a/include/odp/api/spec/shared_memory.h
+++ b/include/odp/api/spec/shared_memory.h
@@ -43,6 +43,16 @@ extern "C" {
  /* Shared memory flags */
 
 /**
+ * @def ODP_SHM_IOVA_INVALID
+ * Invalid IOVA address
+ */
+
+/**
+ * @def ODP_SHM_PA_INVALID
+ * Invalid physical address
+ */
+
+/**
  * Application SW only, no HW access
  *
  * @deprecated  When set, application will not share the reserved memory with HW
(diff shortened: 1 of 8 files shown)


include/odp/api/spec/shared_memory.h
```diff
@@ -213,22 +211,23 @@ int odp_shm_capability(odp_shm_capability_t *capa);
  * Reserve a contiguous block of shared memory
  *
  * Reserve a contiguous block of shared memory that fulfills size, alignment
- * and shareability (ODP_SHM_* flags) requirements. In general, a name is
- * optional and does not need to be unique. However, if the block will be
+ * and shareability (ODP_SHM_* flags) requirements. By default (with no flags), the memory
+ * block can be shared between all ODP threads (pthreads) of the calling process, and with
+ * ODP processes that are forked from the calling process _after_ the reserve call.
```
psavol, line 216: Need to check if memory can be shared between process, although addresses cannot be.

harman-kalra, line 216: Marvell implementation doesn't support forking from a ODP process, so sharing of memory wont be possible in case of two primary processes.

include/odp/api/spec/system_info.h
```diff
@@ -191,6 +194,70 @@ typedef struct odp_system_info_t {
...
+	 */
+	uint64_t used;
+
+	/**
+	 * Memory usage overheads
+	 *
+	 * Total amount of memory (in bytes) that is currently consumed by roundings to
+	 * alignment/block/page size limits, etc. overheads. This is a subset of 'used' bytes.
+	 */
+	uint64_t overhead;
```
harman-kalra, line 251: What will be the use case of identifying the overheads? 
(used - overheads) will give actually memory demanded by user, but (used) was allocated.

psavol: Yes, used - overhead is the memory actually requested by user or needed by an implementation internal memory struct. A simple implementation (like linux-gen SHM) may e.g. always allocate a new page(s) for every memory block reservation request, which waste a lot of memory if there are many small allocations (e.g. if a 12B request reserves 4kB -> 4084B wasted as overhead). From these statistics a user may see where the largest wastes are, and may tune/combine allocations, etc.

harman-kalra: sure. Thanks for explanation.

include/odp/api/spec/shared_memory.h
psavol:
```
By default (without flags), the memory
block can be shared between all threads of the ODP instance. Memory address (odp_shm_addr())
shareability depends on application memory model and #ODP_SHM_SINGLE_VA flag usage.
```

Updated text in v2. Idea is that with 0 flags, memory block can be shared within instance, but address sharing depend on  application memory model (thread vs process) and may require SINGLE_VA flag.

harman-kalra: Yes sounds complete now. As per Marvell implementation, by default all threads will share memory and if a secondary process is spawned, it will also share the memory.

harman-kalra: All three patches looks good
Reviewed-by: Harman Kalra <hkalra@marvell.com>

harman-kalra (commented):
validation: system: add meminfo test
Reviewed-by: Harman Kalra <hkalra@marvell.com>

platform/linux-generic/odp_ishm.c
```diff
@@ -2174,3 +2174,69 @@ void _odp_ishm_print(int block_index)
...
+			if (name_len >= ODP_SYSTEM_MEMBLOCK_NAME_LEN)
+				name_len = ODP_SYSTEM_MEMBLOCK_NAME_LEN - 1;
+
+			memcpy(memblock[num].name, block->name, name_len);
+			memblock[num].name[name_len] = 0;
+
+			addr = 0;
+			proc_index = procfind_block(i);
+			if (proc_index >= 0)
+				addr = (uintptr_t)ishm_proctable->entry[proc_index].start;
```
TuomasTaipale, line 2215: I guess in theory `addr` could be left as zero if `procfind_block()` fails. Is this OK or should something be said in spec about failure values for `odp_system_memblock_t:addr`?

platform/linux-generic/odp_ishm.c
```diff
@@ -2174,3 +2174,69 @@ void _odp_ishm_print(int block_index)
...
+			continue;
+
+		lost = len - block->user_len;
+
+		if (num < max_num) {
+			name_len = strlen(block->name);
+			if (name_len >= ODP_SYSTEM_MEMBLOCK_NAME_LEN)
+				name_len = ODP_SYSTEM_MEMBLOCK_NAME_LEN - 1;
+
+			memcpy(memblock[num].name, block->name, name_len);
```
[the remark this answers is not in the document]
psavol: Fixed in v7.

platform/linux-generic/odp_ishm.c
```diff
@@ -2174,3 +2174,69 @@ void _odp_ishm_print(int block_index)
...
+			memblock[num].name[name_len] = 0;
```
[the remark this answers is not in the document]
psavol: This is a performance optimization. As strncpy() writes always "num" bytes, I don't want to use ODP_SYSTEM_MEMBLOCK_NAME_LEN as "num". So, I check source string length before the call, and when that is known I can use as well memcpy(), since strncpy() would just check source string length again.

platform/linux-generic/odp_ishm.c
psavol: I think this should not fail in normal case (we are holding lock while going through SHM blocks). But even if it would, then it might be better to continue than stop here, since the API is for debugging SHM mappings.

platform/linux-generic/odp_ishm.c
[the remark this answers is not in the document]
TuomasTaipale: OK. In that case, the `memcpy()` with the shorter len constant (ishm name vs. memblock name) could be faster than using `name_len` (enables compile-time optimizations).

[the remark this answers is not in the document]
psavol: ISHM_NAME_MAXLEN is 128B and ODP_SYSTEM_MEMBLOCK_NAME_LEN is 64B. I'm assuming that names are much shorter in average (<20B). So, I'd prefer memcpy() of ~20B vs always 64B.

TuomasTaipale (approved):

PR #1713, opened 2022-11-02
[PATCH v4] helper: threads: improve odph_thread_create() timeout error message

kaurani:
Add more debug information to thread startup timeout error message from
odph_thread_create().


MatiasElo (commented):
Commit description and Signed-off-by -tag is missing, but otherwise looking OK.

MatiasElo (approved):

PR #1715, opened 2022-11-04
[PATCH v4] test: ipsecfwd: add option for direct reception 

TuomasTaipale:
This patchset adds an option to not use scheduler with `odp_ipsecfwd`. Tester throughput increases slightly as scheduling overhead disappears. 

Additionally, packets are now forwarded with bursts instead of single packet at a time after lookup, this further increases tester thoughput if a lot of packets are destined to the same output per processing loop.

v2:
- Added a separate patch where existing processing functions are moved earlier
- Changed "no scheduler" option to "direct RX" option and changed title accordingly
- Rebased

v3: 
- Addressed Janne's comments

v4:
- Added reviewed-by tag

### DIFF
--- a/test/performance/odp_ipsecfwd.c
+++ b/test/performance/odp_ipsecfwd.c
@@ -819,8 +819,6 @@ static odp_bool_t setup_pktios(prog_config_t *config)
 		pktio->num_tx_qs = config->num_output_qs;
 		odp_pktout_queue_param_init(&pktout_param);
 		pktout_param.num_queues = pktio->num_tx_qs;
-		pktout_param.op_mode = config->num_thrs > (int)pktio->num_tx_qs ?
-			ODP_PKTIO_OP_MT : ODP_PKTIO_OP_MT_UNSAFE;
 
 		if (odp_pktout_queue_config(pktio->handle, &pktout_param) < 0) {
 			ODPH_ERR("Error configuring packet I/O output queues (%s)\n", pktio->name);
--- a/test/performance/odp_ipsecfwd.c
+++ b/test/performance/odp_ipsecfwd.c
@@ -75,8 +75,9 @@ typedef struct pktio_s {
 	odph_ethaddr_t src_mac;
 	char *name;
 	odp_pktio_t handle;
-	odp_bool_t (*send_fn)(const pktio_t *pktio, uint8_t index, odp_packet_t pkt);
+	uint32_t (*send_fn)(const pktio_t *pktio, uint8_t index, odp_packet_t pkts[], int num);
 	uint32_t num_tx_qs;
+	uint8_t idx;
 } pktio_t;
 
 typedef struct {
@@ -149,6 +150,17 @@ typedef struct {
 	int idx;
 	int type;
 } exposed_alg_t;
+
+typedef struct {
+	odp_packet_t pkts[MAX_BURST];
+	const pktio_t *pktio;
+	uint32_t num;
+} pkt_vec_t;
+
+typedef struct {
+	pkt_vec_t vecs[MAX_QUEUES];
+	uint8_t num_qs;
+} pkt_out_t;
 
 static exposed_alg_t exposed_algs[] = {
 	ALG_ENTRY(ODP_CIPHER_ALG_NULL, CIPHER_TYPE),
@@ -435,7 +447,7 @@ static inline int process_ipsec_out_enq(odp_packet_t pkts[], const odp_ipsec_sa_
 }
 
 static inline const pktio_t *lookup_and_apply(odp_packet_t pkt, odph_table_t fwd_tbl,
-					      uint8_t *hash)
+					      uint8_t *q_idx)
 {
 	const uint32_t l3_off = odp_packet_l3_offset(pkt);
 	odph_ipv4hdr_t ipv4;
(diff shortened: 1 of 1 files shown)


test/performance/odp_ipsecfwd.c
```diff
@@ -329,6 +345,9 @@ static void print_usage(void)
 	       "  -I, --num_input_qs  Input queue count. 1 by default.\n"
 	       "  -S, --num_sa_qs     SA queue count. 1 by default.\n"
 	       "  -O, --num_output_qs Output queue count. 1 by default.\n"
+	       "  -d, --direct_rx     Use direct RX. Interfaces will be polled by workers\n"
+	       "                      directly. \"--mode\" and \"--num_input_qs\" options are\n"
+	       "                      ignored, input queue count will match worker count.\n"
```
JannePeltonen, line 358: Now it would be quite straightforward to support sync ipsec also with scheduled I/O, but I suppose that could be a topic for another patch set. Or if you want to add it now, I would suggest adding sync mode first and then the direct mode I/O on top if it in a separate patch.

test/performance/odp_ipsecfwd.c
```diff
@@ -622,28 +637,224 @@ static void drain_events(void)
...
+static inline int process_ipsec_in(odp_packet_t pkts[], const odp_ipsec_sa_t sas[], int num,
+				   odp_packet_t pkts_out[])
+{
+	odp_ipsec_in_param_t param;
+	int left, sent = 0, num_out, ret;
+
+	memset(&param, 0, sizeof(param));
+	/* IPsec in/out need to be identified somehow, so use user_ptr for this. */
+	for (int i = 0; i < num; ++i)
+		odp_packet_user_ptr_set(pkts[i], NULL);
```
JannePeltonen, line 675: This is not needed in the sync mode here nor in process_ipsec_out().

test/performance/odp_ipsecfwd.c
```diff
@@ -489,12 +501,26 @@ static inline uint32_t forward_packets(odp_packet_t pkts[], int num, odph_table_
 			odp_packet_free(pkt);
 			continue;
 		}
-		if (odp_unlikely(!pktio->send_fn(pktio, hash % pktio->num_tx_qs, pkt))) {
-			odp_packet_free(pkt);
+
+		vec = &vecs[pktio->idx];
+		vec->pkts[vec->num++] = pkt;
+		vec->hash = hash;
```
JannePeltonen, line 507: This effectively overwrites the flow hash of the preceding packets in the vector, causing them possibly go out through the wrong output queue, which in turn can break packet order in the ordered scheduled mode.

test/performance/odp_ipsecfwd.c
```diff
@@ -489,12 +501,26 @@ static inline uint32_t forward_packets(odp_packet_t pkts[], int num, odph_table_
...
+
+	for (uint32_t i = 0U; i < MAX_IFS; ++i) {
+		if (vecs[i].num == 0U)
 			continue;
-		}
 
-		++num_procd;
+		vec = &vecs[i];
+		pktio = vec->pktio;
+		uint32_t ret = pktio->send_fn(pktio, vec->hash % pktio->num_tx_qs, vec->pkts,
```
JannePeltonen, line 517: Hash based queue selection is not needed in direct mode (and is not needed in the unoredered scheduled mode that was already supported) since packet order across threads is not maintained anyway. Thread-specific output queues would be ok and fast (and could be used in the MT_UNSAFE mode) with direct pktio output.

test/performance/odp_ipsecfwd.c
```diff
@@ -1365,8 +1401,6 @@ static odp_bool_t setup_pktios(prog_config_t *config)
 		pktio->num_tx_qs = config->num_output_qs;
 		odp_pktout_queue_param_init(&pktout_param);
 		pktout_param.num_queues = pktio->num_tx_qs;
-		pktout_param.op_mode = config->num_thrs > (int)pktio->num_tx_qs ?
-			ODP_PKTIO_OP_MT : ODP_PKTIO_OP_MT_UNSAFE;
```
JannePeltonen: This looks like a bug fix that probably should be in its own patch before the other patches of this PR.

test/performance/odp_ipsecfwd.c
TuomasTaipale: Leaving this as future work.

test/performance/odp_ipsecfwd.c
TuomasTaipale: Ditched in v3.

test/performance/odp_ipsecfwd.c
TuomasTaipale: Overhauled this in v3.

test/performance/odp_ipsecfwd.c
TuomasTaipale: New patch (06ba5ae) for this in v3.

test/performance/odp_ipsecfwd.c
```diff
@@ -1020,7 +1046,7 @@ static void parse_sas(prog_config_t *config)
 	(void)fclose(file);
 }
 
-static const pktio_t *get_pktio(const char *iface, const prog_config_t *config)
+static const pktio_t *get_pktio(const char *iface, prog_config_t *config)
```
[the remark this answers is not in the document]
TuomasTaipale: Me neither, unintentional, fixed in v3.

test/performance/odp_ipsecfwd.c
TuomasTaipale: Moved to its own patch (6865d09) in v3.

PR #1721, opened 2022-11-10
[PATCH v3] linux-gen: ipsec: improve IV generation and use non-random IVs

JannePeltonen:
Use non-random but still unpredictable IVs for AES-CBC to get rid of random number generation overhead. DES-CBC and 3DES-CBC continue using random IVs due to their shorter IV length. Make also a couple of small optimizations in IV handling.

### DIFF
--- a/platform/linux-generic/odp_ipsec.c
+++ b/platform/linux-generic/odp_ipsec.c
@@ -550,6 +550,7 @@ static int ipsec_in_iv(odp_packet_t pkt,
 		       ipsec_sa_t *ipsec_sa,
 		       uint16_t iv_offset)
 {
+	_ODP_ASSERT(ipsec_sa->salt_length + ipsec_sa->esp_iv_len <= IPSEC_MAX_IV_LEN);
 	memcpy(state->iv, ipsec_sa->salt, ipsec_sa->salt_length);
 	if (odp_packet_copy_to_mem(pkt,
 				   iv_offset,
@@ -558,6 +559,7 @@ static int ipsec_in_iv(odp_packet_t pkt,
 		return -1;
 
 	if (ipsec_sa->aes_ctr_iv) {
+		ODP_STATIC_ASSERT(IPSEC_MAX_IV_LEN >= 16, "IPSEC_MAX_IV_LEN too small");
 		state->iv[12] = 0;
 		state->iv[13] = 0;
 		state->iv[14] = 0;
@@ -1407,11 +1409,13 @@ static int ipsec_out_iv(ipsec_state_t *state,
 		/* Both GCM and CTR use 8-bit counters */
 		_ODP_ASSERT(sizeof(seq_no) == ipsec_sa->esp_iv_len);
 
+		_ODP_ASSERT(ipsec_sa->salt_length + ipsec_sa->esp_iv_len <= IPSEC_MAX_IV_LEN);
 		memcpy(state->iv, ipsec_sa->salt, ipsec_sa->salt_length);
 		memcpy(state->iv + ipsec_sa->salt_length, &seq_no,
 		       ipsec_sa->esp_iv_len);
 
 		if (ipsec_sa->aes_ctr_iv) {
+			ODP_STATIC_ASSERT(IPSEC_MAX_IV_LEN >= 16, "IPSEC_MAX_IV_LEN too small");
 			state->iv[12] = 0;
 			state->iv[13] = 0;
 			state->iv[14] = 0;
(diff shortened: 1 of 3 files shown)


platform/linux-generic/odp_ipsec_sad.c
```diff
@@ -485,6 +485,25 @@ static void store_sa_info(ipsec_sa_t *ipsec_sa, const odp_ipsec_sa_param_t *p)
 		ipsec_sa->sa_info.out.mtu = p->outbound.mtu;
 }
 
+static int init_cbc_salt(ipsec_sa_t *ipsec_sa)
+{
+	int filled = 0;
+	int rc;
+
+	if (!ipsec_sa->use_cbc_iv)
```
TuomasTaipale, line 493: Matter of taste but as the function is called `init_cbc_salt()` it should should just init it unconditionally and this could be moved outside.

platform/linux-generic/odp_ipsec.c
```diff
@@ -558,6 +559,7 @@ static int ipsec_in_iv(odp_packet_t pkt,
 		return -1;
 
 	if (ipsec_sa->aes_ctr_iv) {
+		ODP_STATIC_ASSERT(IPSEC_MAX_IV_LEN >= 16, "IPSEC_MAX_IV_LEN too small");
```
TuomasTaipale, line 567: These (also next patch) could be more useful near the constant definitions as it's just a plain constant comparison (unlike previous patch where a variable name was involved).

platform/linux-generic/odp_ipsec.c
```diff
@@ -1396,6 +1396,9 @@ static int ipsec_random_data(uint8_t *data, uint32_t len)
 	return 0;
 }
 
+/*
+ * Generate cipher IV for outbound processing.
+ */
 static int ipsec_out_iv(ipsec_state_t *state,
```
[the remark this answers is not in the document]
JannePeltonen: Yes, I thought about it but didn't do it in this patch set. I am not sure if it would be good to put the function pointer directly in the SA instead of an SA type specific structure that could then hold other type specific data and functions too.

platform/linux-generic/odp_ipsec_sad.c
JannePeltonen: I prefer it this way and do not feel that the naming strongly indicates that the checking could not be in the init function (i.e. now the init function inits what is necessary). Maybe it is not exactly equivalent but odp_ipsec_sad_init_global() is also called unconditionally and internally checks if anything actually needs to be inited.

platform/linux-generic/odp_ipsec.c
JannePeltonen: I think it is better to have the asserts next to the code which the assertion is for. Otherwise the connection between the assert and the assumption in the code is easily lost and it would be easier to forget to update the assert when the code is changed. And if not forgotten, it would be more difficult to see whether other parts of code rely on the same assert. So, even though a static assert could be moved elsewhere, I think it shouldn't.

platform/linux-generic/odp_ipsec.c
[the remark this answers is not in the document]
TuomasTaipale: Type specific structure could work too. Would need some prototyping and comparison.

platform/linux-generic/odp_ipsec_sad.c
TuomasTaipale: The point here was that data for the condition is known (all internal, within the same module), as opposed to e.g. the `_odp_ipsec_sad_init_global()` which is an "internal API" call, called from different modules that don't know the specifics. But like I said, matter of taste.

platform/linux-generic/odp_ipsec.c
TuomasTaipale: For simple compile time asserts, I don't think there's really any connection loss as the connection is the usage of the constant itself. Additionally, the code gets a bit bloated as the same check is now in many places. In rest of linux-gen codebase, the static asserts seem to be immediately following or very close to the relevant type/constant definitions.

JannePeltonen: > I don't think there's really any connection loss as the connection is the usage of the constant itself.

But you do not know which usages make the assumptions that are checked by the assert without reviewing all code that uses the identifiers that appear in the assert. To avoid the loss of context, one could add comments near the assert to explain which code depend on it. It is just simpler (and more maintainable, IMHO) to put the assert next to the code that depends on it. Similarly, if someone removes a piece of code, then the asserts that are no longer needed would go away with the removed code without extra pain. Globally scoped asserts could easily be missed, and even if not, updating or removing them would require careful code review of all other code that could potentially still depend on them.

I do not think that the fact that an assert can be changed to a compile time static assert means that it should be moved elsewhere from its context.

> In rest of linux-gen codebase, the static asserts seem to be immediately following or very close to the relevant type/constant definitions.

That is not necessarily always a good thing.

TuomasTaipale: Well, agree to disagree, resolving this.

TuomasTaipale (approved):

PR #1722, opened 2022-11-10
[PATCH v1] Make example run scripts more aligned to test/performance/odp_l2fwd_run.sh and CI friendly

nithind1988:
commit cb533ac0a62bd9ba35dbfc5d0bb5754bcbe4bed6 (HEAD -> dev-run-script-changes, origin/dev-run-script-changes)
Date:   Thu Nov 10 15:09:43 2022 +0530

    example: remove ODP_PKTIO_PCAP condition to run tests

    ODP_PKTIO_PCAP is a linux-generic platform specific define
    and hence remove it from examples so that all the examples can
    run on other platforms too.

commit a96f38ceddc86efed825329d72342e78686af5f7
Author: Sunil Kumar Kori <skori@marvell.com>
Date:   Tue Aug 23 23:21:37 2022 +0530

    example: move pktio_env reference logic to run script

    Unlike script like test/performance/odp_l2fwd_run.sh,
    currently the pktio_env referenced by examples is done by doing a
    symlink of that file. Symbolic link doesn't work when we make
    distribution package or copy build folder to CI machine to run.
    So make it similar to odp_l2fwd_run.sh.

    Also add ODP_PLATFORM in example test environment like done for
    test/.

commit 5449461714809d2a9b1b5324f8532c3cd5e61ec6
Author: Sunil Kumar Kori <skori@marvell.com>
Date:   Thu Aug 18 17:17:52 2022 +0530

    example: use stdbuf prepended to example and perf test scripts

    Use stdbuf to turn off output buffering and get periodic output.
    This is useful for CI monitoring.

commit 0c8a4d88d2de92c9c00cffebc6a59f393aa414e5
Date:   Thu Nov 10 15:26:00 2022 +0530

    linux-gen: copy pktio_env when build out of tree

    For make check to pass when build out of tree, copy
    the pktio_env to build directory.

### DIFF
--- a/example/classifier/Makefile.am
+++ b/example/classifier/Makefile.am
@@ -24,8 +24,6 @@ all-local:
 			fi \
 		done \
 	fi
-	ln -f -s $(top_srcdir)/platform/$(with_platform)/test/example/classifier/pktio_env \
-		pktio_env
 clean-local:
 	if [ "x$(srcdir)" != "x$(builddir)" ]; then \
 		for f in $(EXTRA_DIST); do \
--- a/example/classifier/Makefile.am
+++ b/example/classifier/Makefile.am
@@ -5,9 +5,7 @@ bin_PROGRAMS = odp_classifier
 odp_classifier_SOURCES = odp_classifier.c
 
 if test_example
-if ODP_PKTIO_PCAP
 TESTS = odp_classifier_run.sh
-endif
 endif
 EXTRA_DIST = odp_classifier_run.sh udp64.pcap
 
--- a/example/generator/generator_run.sh
+++ b/example/generator/generator_run.sh
@@ -6,11 +6,17 @@
 # SPDX-License-Identifier:     BSD-3-Clause
 #
 
+SRC_DIR=$(dirname $0)
+TEST_EXAMPLE_DIR=platform/$ODP_PLATFORM/test/example
+PLATFORM_TEST_EXAMPLE=${SRC_DIR}/../../${TEST_EXAMPLE_DIR}
+
 if  [ -f ./pktio_env ]; then
 	. ./pktio_env
+elif [ -f ${PLATFORM_TEST_EXAMPLE}/generator/pktio_env ]; then
+        . ${PLATFORM_TEST_EXAMPLE}/generator/pktio_env
 else
 	echo "BUG: unable to find pktio_env!"
-	echo "pktio_env has to be in current directory"
+	echo "pktio_env has to be in current or platform example directory"
 	exit 1
 fi
 
--- a/example/generator/generator_run.sh
+++ b/example/generator/generator_run.sh
@@ -16,7 +16,13 @@ fi
 
 setup_interfaces
 
-./odp_generator${EXEEXT} -w 1 -n 1 -I $IF0 -m u
+if [ "$(which stdbuf)" != "" ]; then
+	STDBUF="stdbuf -o 0"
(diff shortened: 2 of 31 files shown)


example/classifier/Makefile.am
```diff
@@ -24,8 +24,6 @@ all-local:
 			fi \
 		done \
 	fi
-	ln -f -s $(top_srcdir)/platform/$(with_platform)/test/example/classifier/pktio_env \
-		pktio_env
```
JereLeppanen, line 28: If this was `$(top_builddir)` instead of `$(top_srcdir)`, would it work then? Or just `../../platform/$(with_platform)/test/example/classifier/pktio_env`, like in example/ipsec_*

example/classifier/Makefile.am
```diff
@@ -5,10 +5,8 @@ bin_PROGRAMS = odp_classifier
 odp_classifier_SOURCES = odp_classifier.c
 
 if test_example
-if ODP_PKTIO_PCAP
 TESTS = odp_classifier_run.sh
 endif
-endif
```
JereLeppanen, line 11: Each platform which supports pcap is supposed to define ODP_PKTIO_PCAP. Currently both linux-generic and linux-dpdk do this. If your platform supports pcap, you should also define ODP_PKTIO_PCAP.

example/generator/generator_run.sh
```diff
@@ -16,7 +16,13 @@ fi
...
 setup_interfaces
 
-./odp_generator${EXEEXT} -w 1 -n 1 -I $IF0 -m u
+if [ "$(which stdbuf)" != "" ]; then
+	STDBUF="stdbuf -o 0"
+else
+	STDBUF=
+fi
+
+$STDBUF ./odp_generator${EXEEXT} -w 1 -n 1 -I $IF0 -m u
```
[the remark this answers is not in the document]
nithind1988: When we have multiple sub applications running as a part of test(triggered by pktio_env), we need STDBUF separately for each application since it would be redirected to separate files

example/classifier/Makefile.am
nithind1988: ODP_PKTIO_PCAP actually is just linux-generic platform's feature to support pcap based PKTIO where source is from PCAP. This I think shouldn't be a generic feature as other platforms would never support such HW device which pick from pcap ? For our platform, we feed pcap packets from our internal LBK HW interface.

example/generator/generator_run.sh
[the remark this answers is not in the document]
JereLeppanen: But stdbuf is not specific to any file, it modifies stream buffering for a process. And the buffering is inherited, so it applies to child processes as well. So it could be on a higher level, if it's needed at all.

I'm still not sure what problem you're trying to fix. Seems like the effect of modifying the buffering would basically be that some prints are written to files a few seconds earlier.

example/classifier/Makefile.am
JereLeppanen: Are you referring to pcap live capture, which pcap pktio does not use? Pcap pktio only uses pcap to read and write capture files. So nothing to do with any hardware, and not specific to linux-generic, since libpcap (or equivalent) is pretty widely available.

Looking a bit closer, ODP_PKTIO_PCAP just depends on whether libpcap is available, so there shouldn't be any need to touch it.

JannePeltonen:

PR #1723, opened 2022-11-18
[PATCH v4] test: dma_perf: add a few improvements

TuomasTaipale:
This patchset adds a few improvements to `odp_dma_perf`:
- event wait period can now be configured
- DMA transfers are retried if they fail due to resources being unavailable

v2:
- Added a patch for DMA transfer retries
- Changed PR subject accordingly

v3:
- Added a patch to handle maximum retries

v4:
- Added reviewed-by tag

### DIFF
--- a/test/performance/odp_dma_perf.c
+++ b/test/performance/odp_dma_perf.c
@@ -21,6 +21,7 @@
 
 #define DEFAULT_SEG_SIZE 1024U
 #define ROUNDS 1000000
+#define DEFAULT_WAIT_NS ODP_TIME_SEC_IN_NS
 #define COMPL_DELIMITER ","
 /* For now, a static maximum amount of input segments */
 #define MAX_NUM_IN_SEGS 64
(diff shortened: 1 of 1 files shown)


test/performance/odp_dma_perf.c
```diff
@@ -733,8 +734,16 @@ static int run_dma_sync(test_config_t *config)
 		for (int i = 0; i < config->dma_rounds; ++i) {
 			config->test_case_api.trs_dyn_fn(config, offset, trs_lengths[i]);
 
-			if (odp_dma_transfer(config->dma_config.handle, &trs_params[i], NULL)
-			    <= 0) {
+			while (1) {
+				started = odp_dma_transfer(config->dma_config.handle,
```
psavol, line 738: In case of sync calls, the return value is transfers completed (not started). E.g. "num" or "ret" could be better variable name than "started".

Instead of a forever loop (e.g. problematic in CI), there could be e.g. 1000 retries with odp_time_wait_ns(1000) (1 usec delay in between). Test could report in the end number of retries done, so that it is easy to see if that happens.

TuomasTaipale: Name changed and new patch (8ad80d2) added to handle max retries in v3.

psavol (approved):

PR #1725, opened 2022-11-25
test: sched_latency: improve latency measurement

psavol:
Improved latency measurement by using strict time stamp, and reorganizing code to avoid extra latency between time stamp and enqueue/schedule function calls. Updated help text and test result output to high light that sample events are allocated on top of number of event options.


test/performance/odp_sched_latency.c
```diff
@@ -569,7 +553,7 @@ static void usage(void)
 	       "Usage: ./odp_sched_latency [options]\n"
 	       "Optional OPTIONS:\n"
 	       "  -c, --count <number> CPU count, 0=all available, default=1\n"
-	       "  -d, --duration <number> Test duration in scheduling rounds (millions), default=%d, min=1\n"
+	       "  -d, --duration <number> Test duration in scheduling rounds (millions), default=10, min=1\n"
```
JereLeppanen, line 556: When a default value is changed, it has to be changed in two places. I'm approving anyway, so you can decide what to do, but please re-think one more time.

JereLeppanen (approved):

test/performance/odp_sched_latency.c
psavol: I'm leaving it like this since default values are not changed often, and it is easier to follow values in the code when not hidden behind defines.

PR #1726, opened 2022-11-25
[PATCH v2] test: validation: remove unnecessary cunit includes

psavol:
CUnit API is included through odp_cunit_common.h. Remove direct CUnit includes from test applications.

### DIFF
--- a/test/validation/api/atomic/atomic.c
+++ b/test/validation/api/atomic/atomic.c
@@ -8,7 +8,6 @@
 #include <malloc.h>
 #include <odp_api.h>
 #include <odp/helper/odph_api.h>
-#include <CUnit/Basic.h>
 #include <odp_cunit_common.h>
 #include <unistd.h>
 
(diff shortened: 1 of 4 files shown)


JannePeltonen (approved):

JereLeppanen (approved):
I have to disagree with the implication in the commit comment. If you use something from some header, include it, regardless of whether it's included via some other header. Otherwise you're just creating unnecessary dependency chains.

In this case though, we're probably not using anything from the cunit header. Instead, we're defining our own CU_* macros in odp_cunit_common.h.

I'm approving this, but please consider re-wording the commit comment.

psavol:
> I have to disagree with the implication in the commit comment. If you use something from some header, include it, regardless of whether it's included via some other header. Otherwise you're just creating unnecessary dependency chains.
> 
> In this case though, we're probably not using anything from the cunit header. Instead, we're defining our own CU_* macros in odp_cunit_common.h.
> 
> I'm approving this, but please consider re-wording the commit comment.

I'll reword it. Idea is to make sure that all applications use the CU_ macros defined in cunit common.

PR #1727, opened 2022-11-25
test: pool_perf: add option to enable statistics

psavol:
Added -S option to enable pool performance testing while statistics are enabled. Test also measures statistics function call latency.


MatiasElo (approved):
Copyright can still be updated, but otherwise looking good.

psavol:
> Copyright can still be updated, but otherwise looking good.

Updated in v2.

PR #1732, opened 2022-12-09
[PATCH v3] Crypto: small fixes to sysinfo and crypto perf test

JannePeltonen:
test: performance: crypto: check if crypto operations succeed
example: sysinfo: add missing crypto algorithm names

### DIFF
--- a/test/performance/odp_crypto.c
+++ b/test/performance/odp_crypto.c
@@ -810,6 +810,8 @@ run_measure_one(crypto_args_t *cargs,
 				}
 				packets_sent += rc;
 			} else {
+				odp_crypto_packet_result_t result;
+
 				rc = odp_crypto_op(&pkt, &out_pkt,
 						   &params, 1);
 				if (rc <= 0) {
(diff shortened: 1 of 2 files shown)


JereLeppanen (approved):

test/performance/odp_crypto.c
```diff
@@ -820,6 +822,12 @@ run_measure_one(crypto_args_t *cargs,
 				}
 				packets_sent += rc;
 				packets_received++;
+				if (odp_unlikely(odp_crypto_result(&result, out_pkt) != 0) ||
+				    odp_unlikely(!result.ok)) {
+					ODPH_ERR("Crypto operation failed\n");
```
TuomasTaipale, line 827: Could be useful to print `cipher_status`/`auth_status` as well here (and below).

JannePeltonen: I did not bother to do that since the status is a bit cumbersome (there are 4 separate status fields) and the status codes are currently not very useful (almost always you get either ALG_ERR_DATA_SIZE or ALG_ERR_ICV_CHECK). I can add the printout if you would like to have it anyway.

TuomasTaipale: I don't know, it could be useful for some implementations. Up to you.

JannePeltonen: I decided not to add the status printout in this PR since I actually would like to change the crypto API to simplify the status indication.

TuomasTaipale (approved):

PR #1733, opened 2022-12-16
[PATCH v2] dependencies: update IPsec MB library version

TianyuLi0:
Bump IPsec MB library version to SECLIB-IPSEC-2022.12.13, fix the missing imb_set_errno in some functions.

Signed-off-by: Tianyu Li <tianyu.li@arm.com>


MatiasElo (approved):
The commit message could be cleaned a bit, but otherwise OK.  E.g. `..., which adds some missing imb_set_errno values`. I'm currently updating the test containers with the new library version.

TianyuLi0:
> The commit message could be cleaned a bit, but otherwise OK. E.g. `..., which adds some missing imb_set_errno values`. I'm currently updating the test containers with the new library version.

Thanks for your comments! Commit message updated.

PR #1737, opened 2022-12-19
[PATCH v4] test: bench_misc: add new perf test for misc API calls

psavol:
Test overhead of miscellaneous set of API functions. For example, time and CPU cycle count APIs may be called often in fast path.

### DIFF
--- a/test/performance/odp_bench_misc.c
+++ b/test/performance/odp_bench_misc.c
@@ -1,0 +1,1069 @@
+/* Copyright (c) 2022, Nokia
+ * All rights reserved.
+ *
+ * SPDX-License-Identifier:     BSD-3-Clause
+ */
+
+#ifndef _GNU_SOURCE
+#define _GNU_SOURCE /* Needed for sigaction */
+#endif
+
+#include <odp_api.h>
+#include <odp/helper/odph_api.h>
+
+#include <getopt.h>
+#include <inttypes.h>
+#include <signal.h>
+#include <stdlib.h>
+#include <unistd.h>
+
+/* Number of API function calls per test case */
+#define REPEAT_COUNT 1000
+
+/* Default number of rounds per test case */
+#define ROUNDS 1000u
+
+#define BENCH_INFO(run, init, max, name) \
+	{#run, run, init, max, name}
+
+typedef struct {
+	/* Measure time vs CPU cycles */
+	int time;
+
+	/* Benchmark index to run indefinitely */
+	int bench_idx;
+
+	/* Rounds per test case */
+	uint32_t rounds;
+
+} appl_args_t;
+
+/* Initialize benchmark resources */
+typedef void (*bench_init_fn_t)(void);
+
+/* Run benchmark, returns >0 on success */
+typedef int (*bench_run_fn_t)(void);
+
+/* Benchmark data */
+typedef struct {
+	/* Default test name */
+	const char *name;
+
+	/* Test function to run */
+	bench_run_fn_t run;
+
+	/* Test init function */
+	bench_init_fn_t init;
+
+	/* Test specific limit for rounds (tuning for slow implementation) */
+	uint32_t max_rounds;
+
+	/* Override default test name */
+	const char *desc;
+
+} bench_info_t;
+
+/* Global data */
+typedef struct {
+	appl_args_t appl;
+
+	/* Benchmark functions */
+	bench_info_t *bench;
+
+	/* Number of benchmark functions */
+	int num_bench;
+
+	/* Break worker loop if set to 1 */
+	odp_atomic_u32_t exit_thread;
+
+	/* Test case input / output data */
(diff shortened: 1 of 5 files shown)


test/performance/odp_bench_misc.c
```diff
@@ -0,0 +1,1014 @@
...
+
+		if (opt == -1)
+			break;	/* No more options */
+
+		switch (opt) {
+		case 't':
+			appl_args->time = atoi(optarg);
+			break;
+		case 'i':
+			appl_args->bench_idx = atoi(optarg);
```
TuomasTaipale, line 853: This could be validated, weirdness ensues if somebody passes a negative number.

test/performance/odp_bench_misc.c
```diff
@@ -0,0 +1,1014 @@
...
+		exit(EXIT_FAILURE);
+	}
+	(void)odp_cpumask_to_str(&default_mask, gbl_args->cpumask_str,
+				 sizeof(gbl_args->cpumask_str));
+
+	print_info();
+
+	memset(&worker_thread, 0, sizeof(odph_thread_t));
+
+	signal(SIGINT, sig_handler);
```
TuomasTaipale, line 973: `signal()` is such a mess that new programs should probably use `sigaction()`

test/performance/odp_bench_misc.c
```diff
@@ -0,0 +1,1014 @@
...
+			break;
+		case 'i':
+			appl_args->bench_idx = atoi(optarg);
+			break;
+		case 'r':
+			appl_args->rounds = atoi(optarg);
+			break;
+		default:
+			usage();
+			return -1;
```
MatiasElo, line 860: A bit nitpicking but the return code should be `EXIT_SUCCESS` if `-h` was used.

test/performance/odp_bench_misc.c
```diff
@@ -0,0 +1,1014 @@
...
+	bench_run_fn_t run;
+
+	/* Test init function */
+	bench_init_fn_t init;
+
+	/* Test specific limit for rounds (tuning for slow implementation) */
+	uint32_t max_rounds;
+
+	/* Override default test name */
+	const char *desc;
```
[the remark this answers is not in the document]
psavol: v3: added new case which uses it

	BENCH_INFO(time_diff, init_time_global, 0, "time_diff (global)"),
	BENCH_INFO(time_diff, init_time_local, 0, "time_diff (local)"),

test/performance/odp_bench_misc.c
```diff
@@ -0,0 +1,1014 @@
...
+	odp_time_t t3[REPEAT_COUNT];
+	uint64_t   a1[REPEAT_COUNT];
+	uint64_t   a2[REPEAT_COUNT];
+	uint32_t   b1[REPEAT_COUNT];
+	uint32_t   b2[REPEAT_COUNT];
+	uint16_t   c1[REPEAT_COUNT];
+	uint16_t   c2[REPEAT_COUNT];
+
+	/* Dummy result */
+	uint64_t dummy;
```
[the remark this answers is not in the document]
psavol: Some test cases update this value just to prevent optimizer to eliminate the code. Dummy value does not have other use. It is printed so that compiler do not categorize it as "set but not used" and optimize away.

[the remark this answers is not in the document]
TuomasTaipale: OK. I wonder if having the bench-internal `res` variables as `volatile` would do the trick as well, could then ditch the `dummy`.

test/performance/odp_bench_misc.c
psavol: Fixed exit values in v3.

test/performance/odp_bench_misc.c
psavol: Fixed in v3

test/performance/odp_bench_misc.c
```diff
@@ -0,0 +1,1014 @@
...
+	printf("-------------------------------------------------\n");
+
+	/* Run each test twice. Results from the first warm-up round are ignored. */
+	for (i = 0; i < 2; i++) {
+		uint64_t total = 0;
+		uint32_t round = 1;
+
+		for (j = 0; j < gbl_args->num_bench; round++) {
+			int ret;
+
```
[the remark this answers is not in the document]
psavol: Fixed in v3.

test/performance/odp_bench_misc.c
psavol: Changed to sigaction handler setup in v3.

test/performance/odp_bench_misc.c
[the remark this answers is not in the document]
psavol: Tried volatile there and that caused a large performance decrease (as expected).

TuomasTaipale (approved):

MatiasElo (approved):

PR #1738, opened 2022-12-20
[PATCH v3] linux-gen: pktio: fix input queue count with classifier

TuomasTaipale:
When classifier is in use, only one packet I/O input queue should be configured. Currently, one packet I/O level input queue is configured but driver level input queue count will be whatever is passed via `odp_pktin_queue_param_t::num_queues` in `odp_pktin_queue_config()`.

Fix this by always overriding the parameter to correct value when applicable.

v2:
- Added reviewed-by tag

v3:
- Fixed queue count check

### DIFF
--- a/platform/linux-generic/odp_packet_io.c
+++ b/platform/linux-generic/odp_packet_io.c
@@ -2166,8 +2166,7 @@ void odp_pktio_extra_stats_print(odp_pktio_t pktio)
 	_ODP_PRINT("\n");
 }
 
-int odp_pktin_queue_config(odp_pktio_t pktio,
-			   const odp_pktin_queue_param_t *param)
+int odp_pktin_queue_config(odp_pktio_t pktio, const odp_pktin_queue_param_t *param)
 {
 	pktio_entry_t *entry;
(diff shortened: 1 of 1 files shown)


MatiasElo (approved):

platform/linux-generic/odp_packet_io.c
```diff
@@ -2204,7 +2203,19 @@ int odp_pktin_queue_config(odp_pktio_t pktio,
 		return -1;
 	}
 
-	num_queues = param->classifier_enable ? 1 : param->num_queues;
+	if (param->classifier_enable) {
+		num_queues = 1;
+
+		if (param->num_queues > 1) {
```
psavol, line 2209: Since API says that: "When classifier is enabled in odp_pktin_queue_config() this value is ignored, ..." value of param->num_queues could be anything - also zero. So, may be copy is needed when (param->num_queues != 1).

TuomasTaipale: Yes, good point. Fixed in v3.

PR #1746, opened 2023-01-02
[PATCH v6] linux-gen: loop: add multi-queue support

TuomasTaipale:
This patchset adds multi-queue support to loop packet I/O. This enables more realistic packet I/O application testing.

Additionally, relevant statistics are refactored and simple hashing is implemented to distribute packets to RX queues.

v2:
- Utilized `odp_hash_crc32()` for queue hashing

v3:
- Matias' comments

v4:
- Added reviewed-by tag

v5:
- Swapped to `odp_hash_crc32c()`

### DIFF
--- a/platform/linux-generic/pktio/loop.c
+++ b/platform/linux-generic/pktio/loop.c
@@ -38,6 +38,9 @@
 #include <stdint.h>
 #include <stdlib.h>
 
+#define MAX_QUEUES (ODP_PKTIN_MAX_QUEUES > ODP_PKTOUT_MAX_QUEUES ? \
+			ODP_PKTIN_MAX_QUEUES : ODP_PKTOUT_MAX_QUEUES)
+
 #define MAX_LOOP 16
 
 #define LOOP_MTU_MIN 68
@@ -45,9 +48,20 @@
 
 #define LOOP_MAX_QUEUE_SIZE 1024
 
+typedef struct {
+	odp_atomic_u64_t in_octets;
+	odp_atomic_u64_t in_packets;
+	odp_atomic_u64_t in_discards;
+	odp_atomic_u64_t in_errors;
+	odp_atomic_u64_t out_octets;
+	odp_atomic_u64_t out_packets;
+} stats_t;
+
 typedef struct ODP_ALIGNED_CACHE {
 	/* queue handle as the "wire" */
 	odp_queue_t queue;
+	/* queue specific statistics */
+	stats_t stats;
 	/* config input queue size */
 	uint32_t in_size;
 	/* config output queue size */
@@ -56,7 +70,7 @@ typedef struct ODP_ALIGNED_CACHE {
 
 typedef struct {
 	/* loopback entries for "loop" device */
-	loop_queue_t loopqs[ODP_PKTIN_MAX_QUEUES];
+	loop_queue_t loopqs[MAX_QUEUES];
 	/* config queue count */
 	uint32_t num_conf_qs;
 	/* actual number queues */
@@ -80,7 +94,6 @@ static inline pkt_loop_t *pkt_priv(pktio_entry_t *pktio_entry)
 /* MAC address for the "loop" interface */
 static const uint8_t pktio_loop_mac[] = {0x02, 0xe9, 0x34, 0x80, 0x73, 0x01};
 
-static int loopback_stats_reset(pktio_entry_t *pktio_entry);
 static int loopback_init_capability(pktio_entry_t *pktio_entry);
 
 static int loopback_open(odp_pktio_t id ODP_UNUSED, pktio_entry_t *pktio_entry,
@@ -105,8 +118,16 @@ static int loopback_open(odp_pktio_t id ODP_UNUSED, pktio_entry_t *pktio_entry,
 	pkt_loop->mtu = LOOP_MTU_MAX;
 	pkt_loop->idx = idx;
 	pkt_loop->queue_create = 1;
-	loopback_stats_reset(pktio_entry);
 	loopback_init_capability(pktio_entry);
+
+	for (uint32_t i = 0; i < MAX_QUEUES; i++) {
+		odp_atomic_init_u64(&pkt_loop->loopqs[i].stats.in_octets, 0);
+		odp_atomic_init_u64(&pkt_loop->loopqs[i].stats.in_packets, 0);
+		odp_atomic_init_u64(&pkt_loop->loopqs[i].stats.in_discards, 0);
+		odp_atomic_init_u64(&pkt_loop->loopqs[i].stats.in_errors, 0);
+		odp_atomic_init_u64(&pkt_loop->loopqs[i].stats.out_octets, 0);
+		odp_atomic_init_u64(&pkt_loop->loopqs[i].stats.out_packets, 0);
+	}
 
 	return 0;
 }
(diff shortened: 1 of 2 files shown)


platform/linux-generic/pktio/loop.c
```diff
@@ -438,26 +443,68 @@ static inline void loopback_fix_checksums(odp_packet_t pkt,
...
+
+	if (hash->all_bits == 0)
+		return pkt_loop->loopqs[index % pkt_loop->num_qs].queue;
+
+	off = odp_packet_l4_offset(pkt);
+
+	if (off != ODP_PACKET_OFFSET_INVALID) {
+		if ((hash->proto.ipv4_udp || hash->proto.ipv6_udp) && odp_packet_has_udp(pkt)) {
+			if (odp_packet_copy_to_mem(pkt, off, _ODP_UDPHDR_LEN, &udp) == 0)
+				dest_hash ^= odp_be_to_cpu_16(udp.dst_port);
```
MatiasElo, line 463: It would be more user-friendly to use both destination and source addresses for hashing, although this likely requires something else than modulo to work properly.

TuomasTaipale: Sure, though as the API doesn't go into detail how granular the per-layer-hashing should be I guess this current solution is fine.

MatiasElo: One flexible solution could be to use for example `odp_hash_crc32()` over all fields selected for hashing. Assuming the the performance isn't terrible.

TuomasTaipale: `odp_hash_crc32()` utilized in v2.

platform/linux-generic/pktio/loop.c
```diff
@@ -438,26 +444,86 @@ static inline void loopback_fix_checksums(odp_packet_t pkt,
...
+
+	if (off != ODP_PACKET_OFFSET_INVALID) {
+		if ((hash->proto.ipv4_udp || hash->proto.ipv6_udp) && odp_packet_has_udp(pkt)) {
+			if (odp_packet_copy_to_mem(pkt, off, _ODP_UDPHDR_LEN, &udp) == 0) {
+				head = add_data(head, &udp.src_port, sizeof(udp.src_port));
+				head = add_data(head, &udp.dst_port, sizeof(udp.dst_port));
+			}
+		}
+
+		if ((hash->proto.ipv4_tcp || hash->proto.ipv6_tcp) && odp_packet_has_tcp(pkt)) {
```
MatiasElo, line 479: Could be `else if`.

platform/linux-generic/pktio/loop.c
```diff
@@ -438,26 +444,86 @@ static inline void loopback_fix_checksums(odp_packet_t pkt,
...
+
+	if (off != ODP_PACKET_OFFSET_INVALID) {
+		if (hash->proto.ipv4 && odp_packet_has_ipv4(pkt)) {
+			if (odp_packet_copy_to_mem(pkt, off, _ODP_IPV4HDR_LEN, &ipv4) == 0) {
+				head = add_data(head, &ipv4.src_addr, sizeof(ipv4.src_addr));
+				head = add_data(head, &ipv4.dst_addr, sizeof(ipv4.dst_addr));
+			}
+		}
+
+		if (hash->proto.ipv6 && odp_packet_has_ipv6(pkt)) {
```
MatiasElo, line 497: Could also be `else if`.

platform/linux-generic/pktio/loop.c
```diff
@@ -430,9 +467,7 @@ static int loopback_send(pktio_entry_t *pktio_entry, int index ODP_UNUSED,
 		loopback_fix_checksums(pkt_tbl[i], pktout_cfg, pktout_capa);
 	}
 
-	odp_ticketlock_lock(&pktio_entry->txl);
-
-	queue = pkt_priv(pktio_entry)->loopq;
+	queue = pkt_loop->loopqs[index % pkt_loop->num_qs].queue;
 	ret = odp_queue_enq_multi(queue, (odp_event_t *)hdr_tbl, nb_tx);
```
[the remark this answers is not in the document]
TuomasTaipale: Some old code weirdness, fixed with `memset()` in v3. There's also undefined behavior if `num` happens to be zero, but that code is removed in the latest patch so didn't do anything to that here.

platform/linux-generic/pktio/loop.c
```diff
@@ -508,8 +541,8 @@ static int loopback_init_capability(pktio_entry_t *pktio_entry)
 
 	memset(capa, 0, sizeof(odp_pktio_capability_t));
 
-	capa->max_input_queues  = 1;
-	capa->max_output_queues = 1;
+	capa->max_input_queues = ODP_PKTIN_MAX_QUEUES;
+	capa->max_output_queues = ODP_PKTOUT_MAX_QUEUES;
```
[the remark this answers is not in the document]
TuomasTaipale: `MAX_QUEUES` is just a helper for getting correct size for the shared RX/TX queue array, capabilities should still signal the matching platform capabilities.

platform/linux-generic/pktio/loop.c
```diff
@@ -201,9 +240,7 @@ static int loopback_recv(pktio_entry_t *pktio_entry, int index ODP_UNUSED,
 	if (odp_unlikely(num > QUEUE_MULTI_MAX))
 		num = QUEUE_MULTI_MAX;
 
-	odp_ticketlock_lock(&pktio_entry->rxl);
```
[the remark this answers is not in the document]
TuomasTaipale: V3.

platform/linux-generic/pktio/loop.c
```diff
@@ -13,6 +13,7 @@
 #include <odp/api/queue.h>
 #include <odp/api/ticketlock.h>
 #include <odp/api/time.h>
+#include <odp/api/hash.h>
```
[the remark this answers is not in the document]
TuomasTaipale: V3.

platform/linux-generic/pktio/loop.c
```diff
@@ -207,6 +210,9 @@ static int loopback_pktin_queue_config(pktio_entry_t *pktio_entry,
 	pkt_loop->num_conf_qs = param->num_queues;
 	pkt_loop->queue_create = 1;
 
+	if (param->hash_enable)
+		pkt_loop->hash = param->hash_proto;
```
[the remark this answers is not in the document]
TuomasTaipale: It should. V3.

platform/linux-generic/pktio/loop.c
TuomasTaipale: V3.

MatiasElo (approved):

platform/linux-generic/pktio/loop.c
```diff
@@ -381,22 +441,82 @@ static inline void loopback_fix_checksums(odp_packet_t pkt,
...
+			}
+		} else if (hash->proto.ipv6 && odp_packet_has_ipv6(pkt)) {
+			if (odp_packet_copy_to_mem(pkt, off, _ODP_IPV6HDR_LEN, &ipv6) == 0) {
+				head = add_data(head, &ipv6.src_addr, sizeof(ipv6.src_addr));
+				head = add_data(head, &ipv6.dst_addr, sizeof(ipv6.dst_addr));
+			}
+		}
+	}
+
+	return pkt_loop->loopqs[odp_hash_crc32(data, head - data, 0) % pkt_loop->num_qs].queue;
```
JereLeppanen, line 499: On Intel, odp_hash_crc32c() is much faster than odp_hash_crc32().

If the data_len parameter in the odp_hash_crc32c() call is a constant, the compiler can eliminate a few conditional branches.

CRC may be a poor hash function. We also have thash, which is popular for n-tuples, but it's much slower.

TuomasTaipale: It's faster, period, or faster with the constant length?

Hashing data depends on what kind of packet we have so the length of relevant data will vary.

JereLeppanen: It's always faster. And then a little faster still with constant length. For IPv4 and IPv6 the length is different, but the ports can always be included, just set to zero if not applicable.

PR #1749, opened 2023-01-09
[PATCH v5] linux-gen: crypto: support snow3g algorithm for IPsec MB

TianyuLi0:
Integrate SNOW3G-UEA2 and SNOW3G-UIA2 algorithms for IPsec MB library.
Add SNOW3G-UEA2 and SNOW3G-UIA2 algorithms in the crypto perf test tool.

### DIFF
--- a/platform/linux-generic/odp_crypto_ipsecmb.c
+++ b/platform/linux-generic/odp_crypto_ipsecmb.c
@@ -34,6 +34,7 @@
 #define IPSEC_MB_CRYPTO_MAX_AUTH_KEY_LENGTH        32
 #define IPSEC_MB_CRYPTO_MAX_DATA_LENGTH            65536
 #define ZUC_DIGEST_LENGTH 4
+#define SNOW3G_DIGEST_LENGTH 4
 
 #define ODP_CRYPTO_IPSEC_MB_SHM_NAME "_odp_crypto_ipsecmb"
 /*
@@ -48,6 +49,9 @@ static const odp_crypto_cipher_capability_t cipher_capa_zuc_eea3[] = {
 {.key_len = 16, .iv_len = 16},
 {.key_len = 32, .iv_len = 25} };
 
+static const odp_crypto_cipher_capability_t cipher_capa_snow3g_uea2[] = {
+{.key_len = 16, .iv_len = 16} };
+
 /*
  * Authentication algorithm capabilities
  *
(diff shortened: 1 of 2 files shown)


platform/linux-generic/odp_crypto_ipsecmb.c
```diff
@@ -394,6 +408,208 @@ static int process_auth_zuc_eia3_param(odp_crypto_generic_session_t *session)
 	return 0;
 }
 
+static
+odp_crypto_alg_err_t snow3g_uea2_cipher_op(odp_packet_t pkt,
```
JannePeltonen, line 412: Apart from the actual IMB macro used, this function is almost identical to zuc_eea3_cipher_op(). The same is true between auth_snow3g_uia2_gen() and auth_zuc_eia3_gen() and between auth_snow3g_uia2_check() and auth_zuc_eia3_check(). The code could be rearranged a bit to avoid the duplication.

TianyuLi0: Thanks for your comments, in the latest version, replace duplicated functions with generic functions and added if-condition for different algorithm.

JannePeltonen: The solution by adding if-conditions in the generic processing functions is not ideal and does not work well when more algorithms get added. I guess it will do for now, but when you add more algorithms later, you probably have to change it to something better.

platform/linux-generic/odp_crypto_ipsecmb.c
```diff
@@ -34,6 +34,7 @@
 #define IPSEC_MB_CRYPTO_MAX_AUTH_KEY_LENGTH        32
 #define IPSEC_MB_CRYPTO_MAX_DATA_LENGTH            65536
 #define ZUC_DIGEST_LENGTH 4
+#define SNOW3G_DIGEST_LENGTH ZUC_DIGEST_LENGTH
```
JannePeltonen, line 37: This should be defined as just 4 as snow3g does not really have anything to do with zuc.

platform/linux-generic/odp_crypto_ipsecmb.c
```diff
@@ -62,6 +66,10 @@ static const odp_crypto_auth_capability_t auth_capa_zuc_eia3[] = {
 {.digest_len = 4, .key_len = 32, .aad_len = {.min = 0, .max = 0, .inc = 0},
 	.iv_len = 25} };
 
+static const odp_crypto_auth_capability_t auth_capa_snow3g_uia2[] = {
+{.digest_len = 4, .key_len = 16, .aad_len = {.min = 0, .max = 0, .inc = 0},
```
JannePeltonen, line 70: The macro for the digest len could be use here.

JannePeltonen:
With the macro comment handled,

platform/linux-generic/odp_crypto_ipsecmb.c
TianyuLi0: Thanks for review, updated.

platform/linux-generic/odp_crypto_ipsecmb.c
TianyuLi0: Sure, updated this line with marco.

JannePeltonen (approved):

PR #1753, opened 2023-01-20
[PATCH v3] linux-gen: stash: add alternate mpmc ring based overflow safe implementation

MatiasElo:


platform/linux-generic/odp_stash.c
```diff
@@ -291,16 +403,21 @@ odp_stash_t odp_stash_create(const char *name, const odp_stash_param_t *param)
...
 
+	/* Set ring function pointers */
 	if (ring_u64) {
-		ring_u64_init(&stash->ring_u64.hdr);
-
-		for (i = 0; i < ring_size; i++)
-			stash->ring_u64.data[i] = 0;
+		stash->ring_fn.u64.init      = strict_ring_u64_init;
+		stash->ring_fn.u64.enq_multi = strict_ring_u64_enq_multi;
+		stash->ring_fn.u64.enq_batch = strict_ring_u64_enq_batch;
```
TuomasTaipale: `stash->ring_fn.u64.enq_batch` could probably point directly to `strict_ring_u64_enq_multi` already here and then `strict_ring_u64_enq_batch` becomes unnecessay.

(Same with u32.)

platform/linux-generic/odp_stash.c
```diff
@@ -357,7 +480,7 @@ odp_stash_t odp_stash_lookup(const char *name)
 	return ODP_STASH_INVALID;
 }
 
-static inline int32_t stash_put(odp_stash_t st, const void *obj, int32_t num)
+static inline int32_t stash_put(odp_stash_t st, const void *obj, int32_t num, odp_bool_t batch)
```
TuomasTaipale, line 598: `is_batch` would be a bit more boolean-like.

(Though, current style seems to be `batch` with ones and zeros, well, up to you.)

platform/linux-generic/odp_stash.c
```diff
@@ -369,61 +492,59 @@ static inline int32_t stash_put(odp_stash_t st, const void *obj, int32_t num)
...
+		return stash->ring_fn.u32.enq_multi(stash, u32, num);
 	}
 
 	return -1;
 }
 
 int32_t odp_stash_put(odp_stash_t st, const void *obj, int32_t num)
 {
-	return stash_put(st, obj, num);
+	return stash_put(st, obj, num, 0);
```
TuomasTaipale, line 654: The batch identifier is `odp_bool_t` so could use `true`/`false`.

config/odp-linux-generic.conf
```diff
@@ -254,6 +254,12 @@ stash: {
 	# The value may be rounded up by the implementation. For optimal memory
 	# usage set value to a power of two - 1.
 	max_num_obj = 4095
+
+	# Strict size
+	#
+	# When set to 0, application can attempt to store more handles into a
+	# stash than it specified in the creation parameters.
+	strict_size = 1
```
TuomasTaipale, line 262: Commit message mentions `overflow_safe` config file option.

platform/linux-generic/include/odp_queue_basic_internal.h
```diff
@@ -23,7 +23,7 @@ extern "C" {
 #include <odp/api/ticketlock.h>
```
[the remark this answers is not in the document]
MatiasElo: Fixed in V2.

platform/linux-generic/include/odp_ring_mpmc_internal.h
```diff
@@ -1,9 +1,13 @@
 /* Copyright (c) 2018, Linaro Limited
+ * Copyright (c) 2023, Nokia
  * All rights reserved.
  *
  * SPDX-License-Identifier:     BSD-3-Clause
  */
 
+/* This header should NOT be included directly. There are no include guards for
+ * the function definitions! */
+
```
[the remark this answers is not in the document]
MatiasElo: It fails already due to missing defines, but I added error print to be more clear.

platform/linux-generic/odp_stash.c
MatiasElo: Fixed in V2.

platform/linux-generic/odp_stash.c
MatiasElo: Fixed on V2.

platform/linux-generic/odp_stash.c
MatiasElo: Fixed on V2.

platform/linux-generic/odp_stash.c
```diff
@@ -482,14 +597,14 @@ static inline int32_t stash_put(odp_stash_t st, const void *obj, int32_t num, od
 	obj_size = stash->obj_size;
 
 	if (obj_size == sizeof(uint64_t)) {
-		if (batch)
+		if (is_batch)
```
TuomasTaipale, line 600: Actually, perhaps this batch check could now be done once at the beginning of put/get and save correct functions (as function pointers are already used anyways) to variables and then use the variables, a bit less if-batch-else repetition.

(I seemingly whined about this also in #1636.)

Otherwise looks OK.

TuomasTaipale (approved):

platform/linux-generic/odp_stash.c
MatiasElo: Makes sense, fixed in V3.

PR #1755, opened 2023-01-23
[PATCH v5] api: cls: clarify destroy order

JereLeppanen:
Clarify the order in which CoSes, PMRs, pools and queues should be destroyed. Allow default CoS to be removed. Allow CoS pool to be set to ODP_POOL_INVALID.

### DIFF
--- a/include/odp/api/spec/classification.h
+++ b/include/odp/api/spec/classification.h
@@ -780,7 +780,7 @@ int odp_cos_queue_set(odp_cos_t cos, odp_queue_t queue);
 * @param cos           CoS handle
 *
 * @retval Queue handle associated with the given class-of-service
-* @retval ODP_QUEUE_INVALID on failure
+* @retval ODP_QUEUE_INVALID on failure, or if the queue has not been set
 */
 odp_queue_t odp_cos_queue(odp_cos_t cos);
 
--- a/include/odp/api/spec/classification.h
+++ b/include/odp/api/spec/classification.h
@@ -750,6 +750,11 @@ odp_queue_t odp_cls_hash_result(odp_cos_t cos, odp_packet_t packet);
 /**
  * Discard a class-of-service along with all its associated resources
  *
+ * Before destroying a CoS, all the PMRs referring to the CoS (as a source or
+ * destination CoS) must be destroyed first. Also, the CoS must not be in use
+ * as the default CoS in any pktio (@see odp_pktio_default_cos_set()) or as the
+ * destination CoS of any IPsec SA.
+ *
  * @param cos          CoS handle
  *
  * @retval  0 on success
--- a/include/odp/api/spec/classification.h
+++ b/include/odp/api/spec/classification.h
@@ -622,7 +622,8 @@ typedef enum {
  */
 typedef struct odp_cls_cos_param {
 	/** Action to take. When action is ODP_COS_ACTION_DROP, all the other
-	 * parameters are ignored.
+	 * parameters are ignored. If action is ODP_COS_ACTION_ENQUEUE, then
+	 * queue must be set, or num_queue must be greater than one.
 	 *
 	 * The final match in the CoS chain defines the action for a packet.
 	 * I.e. packet is dropped only when the CoS of the last matching rule
@@ -773,9 +774,12 @@ int odp_cos_destroy(odp_cos_t cos);
 /**
  * Assign a queue for a class-of-service
  *
+ * Action of the given CoS may not be ODP_COS_ACTION_DROP.
+ *
  * @param cos          CoS handle
  * @param queue        Handle of the queue where all packets of this specific
- *                     class of service will be enqueued.
+ *                     class of service will be enqueued. Must not be
+ *                     ODP_QUEUE_INVALID.
  *
  * @retval  0 on success
  * @retval <0 on failure
@@ -788,7 +792,8 @@ int odp_cos_queue_set(odp_cos_t cos, odp_queue_t queue);
 * @param cos           CoS handle
 *
 * @retval Queue handle associated with the given class-of-service
-* @retval ODP_QUEUE_INVALID on failure, or if the queue has not been set
+* @retval ODP_QUEUE_INVALID on failure, or if there are multiple queues, or if
+*         the CoS action is ODP_COS_ACTION_DROP.
 */
 odp_queue_t odp_cos_queue(odp_cos_t cos);
 
--- a/include/odp/api/spec/classification.h
+++ b/include/odp/api/spec/classification.h
@@ -671,7 +671,15 @@ typedef struct odp_cls_cos_param {
 			odp_pktin_hash_proto_t hash_proto;
 		};
 	};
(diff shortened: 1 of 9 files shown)


include/odp/api/spec/classification.h
```diff
@@ -751,7 +751,8 @@ odp_queue_t odp_cls_hash_result(odp_cos_t cos, odp_packet_t packet);
  * Discard a class-of-service along with all its associated resources
  *
  * Before destroying a CoS, all the PMRs referring to the CoS (as a source or
- * destination CoS) must be destroyed first.
+ * destination CoS) must be destroyed first. The CoS must not be in use as the
+ * default CoS in any pktio (@see odp_pktio_default_cos_set()).
```
JannePeltonen, line 761: It could also be mentioned that the CoS must not be in use as the destination CoS of any IPsec SA.

include/odp/api/spec/queue.h
```diff
@@ -50,7 +50,9 @@ odp_queue_t odp_queue_create(const char *name, const odp_queue_param_t *param);
  * Destroy ODP queue
  *
  * Destroys ODP queue. The queue must be empty and detached from other
- * ODP API (crypto, pktio, etc). Application must ensure that no other
+ * ODP API (crypto, pktio, etc). The queue must not be in use in any CoS. If it
+ * is, assign another queue to the CoS, or destroy the CoS, before calling this
+ * function. Application must ensure that no other
```
JannePeltonen, line 55: The added text is good, but isn't it already said there ("The queue must be empty and detached from other ODP API (crypto, pktio, etc)."? If clarification is needed, maybe it is needed for other cases that classifier too?

include/odp/api/spec/classification.h
```diff
@@ -671,7 +671,13 @@ typedef struct odp_cls_cos_param {
 			odp_pktin_hash_proto_t hash_proto;
 		};
 	};
-	/** Pool associated with CoS */
+	/** Pool associated with CoS
+	 *
+	 * May be set to ODP_POOL_INVALID, in which case the originating pktin
+	 * pool is used.
```
JannePeltonen, line 677: "originating pktin pool" is not well defined when classifier in entered through IPsec (odp_ipsec_sa_param_t::dest_cos). The packet being classified may not have originated in a pktin at all.

psavol: Could mention odp_pktio_open() and refer the pool as packet input "default pool".

include/odp/api/spec/classification.h
JereLeppanen: v2.

include/odp/api/spec/queue.h
JereLeppanen: Answer to both questions is yes, probably. This change could just be dropped. But I still kept it, anyway, in v2.

include/odp/api/spec/classification.h
JereLeppanen: v2.

include/odp/api/spec/pool.h
```diff
@@ -59,6 +59,9 @@ odp_pool_t odp_pool_create(const char *name, const odp_pool_param_t *param);
 /**
  * Destroy a pool previously created by odp_pool_create()
  *
+ * The pool to be destroyed must not be in use in any CoS. If it is, assign
+ * another pool to the CoS, or destroy the CoS, before calling this function.
+ *
```
psavol, line 64: These instructions are a bit specific to classifier API - to be located here in the pool API. I think it should enough to move the "note" section from below up here (remove "@note" tag while doing that), and edit it to say that the pool must not be in-use (including packet input/output, classifier, timer, ...) when calling this function.

include/odp/api/spec/queue.h
psavol: Similarly to pool API, queue API should not give too specific instructions about classifier API, but just remind that any usage of a destroyed queue leads to problems.

Classifier and timer could be added to the list of "other APIs", as those are common users of destination queues.

platform/linux-generic/odp_classification.c
```diff
@@ -1718,13 +1718,15 @@ int _odp_cls_classify_packet(pktio_entry_t *entry, const uint8_t *base,
 	if (cos->action == ODP_COS_ACTION_DROP)
 		return 1;
 
-	if (cos->queue == ODP_QUEUE_INVALID && cos->num_queue == 1)
-		goto error;
-
-	if (cos->pool == ODP_POOL_INVALID)
-		goto error;
+	if (cos->queue == ODP_QUEUE_INVALID && cos->num_queue == 1) {
```
psavol, line 1724: Git log entry text has copy-paste error: "CoS pool is ODP_QUEUE_INVALID" - pool vs. queue invalid

include/odp/api/spec/classification.h
```diff
@@ -780,7 +780,7 @@ int odp_cos_queue_set(odp_cos_t cos, odp_queue_t queue);
 * @param cos           CoS handle
 *
 * @retval Queue handle associated with the given class-of-service
-* @retval ODP_QUEUE_INVALID on failure
+* @retval ODP_QUEUE_INVALID on failure, or if the queue has not been set
```
psavol, line 791: I think we may need to further clarify when CoS may not have queue set. Obviously, a ODP_COS_ACTION_DROP CoS will not have destination queue (API says that queue handle is ignored in CoS params).

However, I'm not sure if ENQUEUE action can have destination queue not set. CoS params expect that it is set:

> Mapping used when num_queue = 1, hashing is disabled in
> this case and application has to configure this queue and
> packets are delivered to this queue
> 
> odp_queue_t queue;

JereLeppanen: Added some clarifications in v3 (last commit), not sure if it's what you had in mind.

JereLeppanen:
All comments processed in v3.

kkirankk (approved):

include/odp/api/spec/packet_io.h
```diff
@@ -665,7 +665,8 @@ int odp_pktio_mac_addr_set(odp_pktio_t pktio, const void *mac_addr,
  * @param pktio        Ingress port pktio handle.
  * @param default_cos  Class-of-service set to all packets arriving at this
  *                     ingress port, unless overridden by subsequent
- *                     header-based filters.
+ *                     header-based filters. Use ODP_COS_INVALID to remove
+ *                     the default CoS.
```
psavol, line 669: "... unless overridden by subsequent header-based filters. " refers to deprecated API functions. I'd suggest to remove the sentence now.

include/odp/api/spec/classification.h
psavol: That addition looks good.

include/odp/api/spec/packet_io.h
JereLeppanen: v4.

JereLeppanen:
v4:
- Rebase.
- Petri's comment.
- Add two more commits.
- Add review tags.

PR #1760, opened 2023-02-03
[PATCH v3] validation: dma: validate user context pointer value

TuomasTaipale:
In addition to checking that DMA implementation passes the user context pointer of completion parameters as is back to application (in poll and event completion cases), check also that the value pointed to by it remains unchanged.

v2:
- Added check to conditionally assert `user_ptr` value
- Added reviewed-by tag

v3:
- Further simplified `user_ptr` value check

### DIFF
--- a/test/validation/api/dma/dma.c
+++ b/test/validation/api/dma/dma.c
@@ -19,6 +19,7 @@
 #define TRAILER      10
 #define MULTI        1
 #define RESULT       1
+#define USER_DATA    0xdeadbeef
 
 typedef struct global_t {
 	odp_dma_capability_t dma_capa;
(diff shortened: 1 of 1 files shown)


test/validation/api/dma/dma.c
```diff
@@ -452,6 +454,7 @@ static int do_transfer_async(odp_dma_t dma, const odp_dma_transfer_param_t *trs_
 		CU_ASSERT(done == 1);
 		CU_ASSERT(result.success);
 		CU_ASSERT(result.user_ptr == user_ptr);
+		CU_ASSERT(*(uint32_t *)result.user_ptr == USER_DATA);
```
psavol, line 457: To avoid crash (and debugging), this assert could be executed only if the user pointer is correct.

TuomasTaipale: Fixed in v2.

psavol (approved):

PR #1764, opened 2023-02-09
[PATCH v8] example: timer_accuracy: add periodic timer mode

JereLeppanen:
Add a new mode (-m 3), which uses the periodic timer API.

v2:
- Rebase.
- Matias' comments.
- Remove unnecessary wait in cancel_periodic_timers().

v3:
- Rebase.
- Matias' comments.
- Change -P default to 5:0:0:1.
- Add periodic mode defaults to help text.

v4:
- Tweak help texts.
- Make periodic timer measurement more accurate.

v5:
- All comments.
- Add warmup option.
- Fix tmo_ns rounding (+.5) in run_test().

v6-v7:
- All comments.

v8:
- Rebase.
- Add review tags.

### DIFF
--- a/example/timer/odp_timer_accuracy.c
+++ b/example/timer/odp_timer_accuracy.c
@@ -1,5 +1,5 @@
 /* Copyright (c) 2018, Linaro Limited
- * Copyright (c) 2019-2022, Nokia
+ * Copyright (c) 2019-2023, Nokia
  * All rights reserved.
  *
  * SPDX-License-Identifier:     BSD-3-Clause
@@ -18,11 +18,18 @@
 
 #define MAX_FILENAME 128
 
+enum mode_e {
+	MODE_ONESHOT = 0,
+	MODE_RESTART_ABS,
+	MODE_RESTART_REL,
+	MODE_PERIODIC,
+};
+
 typedef struct timer_ctx_t {
 	odp_timer_t timer;
 	odp_event_t event;
 	uint64_t    nsec;
-
+	uint64_t    count;
 } timer_ctx_t;
 
 typedef struct {
@@ -58,7 +65,10 @@ typedef struct test_global_t {
 		unsigned long long num;
 		unsigned long long burst;
 		unsigned long long burst_gap;
-		int mode;
+		odp_fract_u64_t freq;
+		unsigned long long max_multiplier;
+		unsigned long long multiplier;
+		enum mode_e mode;
 		int clk_src;
 		int init;
 		int output;
@@ -94,15 +104,20 @@ static void print_usage(void)
 	       "  -r, --res_ns <nsec>     Timeout resolution in nsec. Default: period / 10\n"
 	       "  -R, --res_hz <hertz>    Timeout resolution in hertz. Note: resolution can be set\n"
 	       "                          either in nsec or hertz (not both). Default: 0\n"
-	       "  -f, --first <nsec>      First timer offset in nsec. Default: 300 msec\n"
+	       "  -f, --first <nsec>      First timer offset in nsec. Default: 0 for periodic mode, otherwise 300 msec\n"
 	       "  -x, --max_tmo <nsec>    Maximum timeout in nsec. When 0, max tmo is calculated from other options. Default: 0\n"
 	       "  -n, --num <number>      Number of timeout periods. Default: 50\n"
 	       "  -b, --burst <number>    Number of timers per a timeout period. Default: 1\n"
 	       "  -g, --burst_gap <nsec>  Gap (in nsec) between timers within a burst. Default: 0\n"
+	       "                          In periodic mode, first + burst * burst_gap must be less than period length.\n"
 	       "  -m, --mode <number>     Test mode select (default: 0):\n"
-	       "                            0: Set all timers at init phase.\n"
-	       "                            1: Set first burst of timers at init. Restart timers during test with absolute time.\n"
-	       "                            2: Set first burst of timers at init. Restart timers during test with relative time.\n"
+	       "                            0: One-shot. Start all timers at init phase.\n"
+	       "                            1: One-shot. Each period, restart timers with absolute time.\n"
+	       "                            2: One-shot. Each period, restart timers with relative time.\n"
+	       "                            3: Periodic.\n"
+	       "  -P, --periodic <freq_integer:freq_numer:freq_denom:max_multiplier>\n"
+	       "                          Periodic timer pool parameters. Default: 5:0:0:1\n"
+	       "  -M, --multiplier        Periodic timer multiplier. Default: 1\n"
 	       "  -o, --output <file>     Output file for measurement logs\n"
 	       "  -e, --early_retry <num> When timer restart fails due to ODP_TIMER_TOO_NEAR, retry this many times\n"
 	       "                          with expiration time incremented by the period. Default: 0\n"
@@ -126,6 +141,8 @@ static int parse_options(int argc, char *argv[], test_global_t *test_global)
 		{"burst",        required_argument, NULL, 'b'},
 		{"burst_gap",    required_argument, NULL, 'g'},
 		{"mode",         required_argument, NULL, 'm'},
+		{"periodic",     required_argument, NULL, 'P'},
+		{"multiplier",   required_argument, NULL, 'M'},
 		{"output",       required_argument, NULL, 'o'},
 		{"early_retry",  required_argument, NULL, 'e'},
 		{"clk_src",      required_argument, NULL, 's'},
@@ -133,18 +150,23 @@ static int parse_options(int argc, char *argv[], test_global_t *test_global)
 		{"help",         no_argument,       NULL, 'h'},
 		{NULL, 0, NULL, 0}
 	};
-	const char *shortopts =  "+p:r:R:f:x:n:b:g:m:o:e:s:ih";
+	const char *shortopts =  "+p:r:R:f:x:n:b:g:m:P:M:o:e:s:ih";
 	int ret = 0;
 
 	test_global->opt.period_ns = 200 * ODP_TIME_MSEC_IN_NS;
 	test_global->opt.res_ns    = 0;
 	test_global->opt.res_hz    = 0;
-	test_global->opt.offset_ns = 300 * ODP_TIME_MSEC_IN_NS;
+	test_global->opt.offset_ns = UINT64_MAX;
 	test_global->opt.max_tmo_ns = 0;
 	test_global->opt.num       = 50;
 	test_global->opt.burst     = 1;
 	test_global->opt.burst_gap = 0;
-	test_global->opt.mode      = 0;
+	test_global->opt.mode      = MODE_ONESHOT;
+	test_global->opt.freq.integer = ODP_TIME_SEC_IN_NS / test_global->opt.period_ns;
+	test_global->opt.freq.numer = 0;
+	test_global->opt.freq.denom = 0;
+	test_global->opt.max_multiplier = 1;
+	test_global->opt.multiplier = 1;
 	test_global->opt.clk_src   = ODP_CLOCK_DEFAULT;
 	test_global->opt.init      = 0;
 	test_global->opt.output    = 0;
@@ -184,6 +206,16 @@ static int parse_options(int argc, char *argv[], test_global_t *test_global)
 		case 'm':
 			test_global->opt.mode = atoi(optarg);
 			break;
+		case 'P':
+			sscanf(optarg, "%" SCNu64 ":%" SCNu64 ":%" SCNu64 ":%llu",
+			       &test_global->opt.freq.integer,
+			       &test_global->opt.freq.numer,
+			       &test_global->opt.freq.denom,
+			       &test_global->opt.max_multiplier);
+			break;
+		case 'M':
+			test_global->opt.multiplier = strtoull(optarg, NULL, 0);
+			break;
 		case 'o':
 			test_global->opt.output = 1;
 			/* filename is NULL terminated in anycase */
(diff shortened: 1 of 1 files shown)


example/timer/odp_timer_accuracy.c
```diff
@@ -297,11 +299,15 @@ static int start_timers(test_global_t *test_global)
 		return -1;
 	}
 
-	if (timer_capa.max_timers &&
-	    test_global->alloc_timers > timer_capa.max_timers) {
+	max_timers = timer_capa.max_timers;
+
+	if (mode == 3)
+		max_timers = timer_capa.periodic.max_timers;
```
MatiasElo, line 305: To follow API strictly `odp_timer_capability_t.periodic.max_pools` should be checked.

example/timer/odp_timer_accuracy.c
```diff
@@ -297,11 +299,15 @@ static int start_timers(test_global_t *test_global)
 		return -1;
 	}
 
-	if (timer_capa.max_timers &&
-	    test_global->alloc_timers > timer_capa.max_timers) {
+	max_timers = timer_capa.max_timers;
+
+	if (mode == 3)
```
MatiasElo, line 304: Using an enum for mode would make the code more readable.

example/timer/odp_timer_accuracy.c
```diff
@@ -358,14 +364,21 @@ static int start_timers(test_global_t *test_global)
 		timer_param.max_tmo = test_global->opt.max_tmo_ns;
 	}
 
+	if (mode == 3) {
+		timer_param.timer_type = ODP_TIMER_TYPE_PERIODIC;
+		timer_param.periodic.base_freq_hz.numer = ODP_TIME_SEC_IN_NS;
+		timer_param.periodic.base_freq_hz.denom = period_ns;
+		timer_param.periodic.max_multiplier = 1;
```
MatiasElo, line 371: `odp_timer_periodic_capability()`should be used to check if parameters are supported.

MatiasElo: It could be simpler to add these values as new options, so we can tests exact values.

JereLeppanen:
@MatiasElo All fixed in v2.

example/timer/odp_timer_accuracy.c
```diff
@@ -446,23 +511,38 @@ static int start_timers(test_global_t *test_global)
...
+				/*
+				 * Burst is centered at the middle of the period. If the burst
+				 * doesn't fit in the period, odp_timer_periodic_start() fails.
+				 */
+				nsec = period_ns / 2 - burst_gap * burst / 2 + j * burst_gap;
+				ctx->nsec = start_ns + nsec;
+				start_param.freq_multiplier = test_global->opt.multiplier;
+				start_param.first_tick =
+					start_tick + odp_timer_ns_to_tick(timer_pool, nsec);
+				start_param.tmo_ev = ctx->event;
```
MatiasElo, line 535: This seems quite complex. Would it be possible to simply set `start_param.first_tick`to zero if `burst_gap` was not given? And if given, e.g. `nsec = burst_gap * (burst + 1)`.

MatiasElo:
There seems to be something odd in the printed test results. E.g.
```
Test parameters:
  clock source:    0
  max res nsec:    1000
  max res hertz:   1000000
  max timers capa: 4294967295
  mode:            3
  restart retries: 0
  start offset:    300000000 nsec
  period:          125000 nsec
  resolution:      62500 nsec
  base integer:    8000
  base numer:      0
  base denom:      0
  multiplier:      1
  min timeout:     12500 nsec
  max timeout:     300250000 nsec
  num timeout:     600000
  burst size:      1
  burst gap:       0
  total timers:    600000
  alloc timers:    1
  test run time:   75.30 sec


 Test results:
  num after:        600000  /  100.00%
  num before:            0  /  0.00%
  num exact:             0  /  0.00%
  num retry:             0  /  0.00%
  error after (nsec):
         min:        51306  /  4.104x resolution
         max:        51386  /  4.111x resolution
         ave:        51332  /  4.107x resolution
  error before (nsec):
         min:            0  /  0.000x resolution
         max:            0  /  0.000x resolution
         ave:            0  /  0.000x resolution
```
For example, average error after is `51332  /  4.107x resolution`, when the printed resolution above is 62500 nsec.

MatiasElo:
As an additional feature, it would be useful to print the indexes of min/max timeout error events.

JereLeppanen:
> There seems to be something odd in the printed test results. E.g.

Command line?

MatiasElo:
> > There seems to be something odd in the printed test results. E.g.
> 
> Command line?

`./odp_timer_accuracy -m 3 -c 8000:0:0:1 -n 600000`

example/timer/odp_timer_accuracy.c
```diff
@@ -144,7 +157,11 @@ static int parse_options(int argc, char *argv[], test_global_t *test_global)
 	test_global->opt.num       = 50;
 	test_global->opt.burst     = 1;
 	test_global->opt.burst_gap = 0;
-	test_global->opt.mode      = 0;
+	test_global->opt.mode      = MODE_ONESHOT;
+	test_global->opt.base.integer = 0;
+	test_global->opt.base.numer = ODP_TIME_SEC_IN_NS;
+	test_global->opt.base.denom = 200 * ODP_TIME_MSEC_IN_NS;
+	test_global->opt.multiplier = 1;
```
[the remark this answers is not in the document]
JereLeppanen: Or print these only in periodic mode?

example/timer/odp_timer_accuracy.c
JereLeppanen: > Would it be possible to simply set `start_param.first_tick`to zero if `burst_gap` was not given?

That should work.

> And if given, e.g. `nsec = burst_gap * (burst + 1)`.

Maybe you mean `nsec = burst_gap * (j + 1)`. Regardless, that's an automatic TOO_NEAR with the default burst_gap of 0. The complexity is there to minimize the probability of TOO_NEAR and TOO_FAR.

JereLeppanen:
> `./odp_timer_accuracy -m 3 -c 8000:0:0:1 -n 600000`

Platform / environment?

linux-gen on x86: `Resolution 12500 nsec too high. Highest resolution 100000 nsec. Default resolution is period / 10.`

example/timer/odp_timer_accuracy.c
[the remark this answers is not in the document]
MatiasElo: Either option works.

example/timer/odp_timer_accuracy.c
MatiasElo: Yep, meant `nsec = burst_gap * (j + 1)`, but only if `burst_gap` has been given by user (!=0). This way it would be user's responsibility to pick a large enough value.

example/timer/odp_timer_accuracy.c
```diff
@@ -103,6 +112,9 @@ static void print_usage(void)
 	       "                            0: Set all timers at init phase.\n"
 	       "                            1: Set first burst of timers at init. Restart timers during test with absolute time.\n"
 	       "                            2: Set first burst of timers at init. Restart timers during test with relative time.\n"
+	       "                            3: Set first burst of timers at init. Use periodic timer API.\n"
+	       "  -c, --periodic <base_integer:base_numer:base_denom:multiplier>\n"
+	       "                          Periodic timer parameters for mode 3.\n"
```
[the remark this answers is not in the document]
JereLeppanen: Added -M option in v3.

example/timer/odp_timer_accuracy.c
JereLeppanen: In v3, changed this to something that's closer to the non-periodic case, using offset_ns, but if nsec is 0, then first_tick is 0.

JereLeppanen:
All fixed in v3.

example/timer/odp_timer_accuracy.c
```diff
@@ -94,7 +110,7 @@ static void print_usage(void)
 	       "  -r, --res_ns <nsec>     Timeout resolution in nsec. Default: period / 10\n"
 	       "  -R, --res_hz <hertz>    Timeout resolution in hertz. Note: resolution can be set\n"
 	       "                          either in nsec or hertz (not both). Default: 0\n"
-	       "  -f, --first <nsec>      First timer offset in nsec. Default: 300 msec\n"
+	       "  -f, --first <nsec>      First timer offset in nsec. Default: 0 for mode 3, otherwise 300 msec\n"
 	       "  -x, --max_tmo <nsec>    Maximum timeout in nsec. When 0, max tmo is calculated from other options. Default: 0\n"
 	       "  -n, --num <number>      Number of timeout periods. Default: 50\n"
 	       "  -b, --burst <number>    Number of timers per a timeout period. Default: 1\n"
```
psavol, line 116: Burst / burst_gap could document that in period mode first + burst*burst_gap must be less than period length

JereLeppanen: v4.

JereLeppanen:
@MatiasElo please review v4.

example/timer/odp_timer_accuracy.c
```diff
@@ -446,23 +538,37 @@ static int start_timers(test_global_t *test_global)
...
-			start_param.tick_type = ODP_TIMER_TICK_ABS;
-			start_param.tick = start_tick + odp_timer_ns_to_tick(timer_pool, nsec);
-			start_param.tmo_ev = ctx->event;
-
-			ret = odp_timer_start(ctx->timer, &start_param);
+			if (mode == MODE_PERIODIC) {
+				odp_timer_periodic_start_t start_param;
+
+				nsec = offset_ns + (j * burst_gap);
+				ctx->nsec = start_ns + nsec;
```
MatiasElo, line 560: If nsec = 0 -> first_tick = 0, should one period be added to `ctx->nsec`?

MatiasElo (commented):
Couple small comments, but otherwise looks good. As a future improvement, we could think about adding a warm-up rounds option to the application.

example/timer/odp_timer_accuracy.c
```diff
@@ -592,7 +746,10 @@ static void run_test(test_global_t *test_global)
 	timer_ctx_t *ctx;
 	test_stat_t *stat = &test_global->stat;
 	test_log_t *log = test_global->log;
-	int mode = test_global->opt.mode;
+	enum mode_e mode = test_global->opt.mode;
+	double period = ODP_TIME_SEC_IN_NS /
+			odp_fract_u64_to_dbl(&test_global->opt.freq) /
+			test_global->opt.multiplier;
```
MatiasElo, line 752: How about using fractional numbers until saving `tmo_ns` to avoid possible rounding errors?

example/timer/odp_timer_accuracy.c
```diff
@@ -358,14 +422,41 @@ static int start_timers(test_global_t *test_global)
...
 	}
 
+	if (mode == MODE_PERIODIC) {
+		odp_timer_periodic_capability_t capa;
+
+		capa.base_freq_hz = freq;
+		capa.max_multiplier = max_multiplier;
+		capa.res_ns = res_ns;
+
+		if (odp_timer_periodic_capability(test_global->opt.clk_src, &capa) < 0) {
```
psavol, line 432: It would be good to handle also return value 0. Maybe print a warning that the requested base frequency is not met, which causes a drift. 

 * @retval 0   Success. Capability does not match base frequency exactly, but meets or exceeds
 *             other requested values.

example/timer/odp_timer_accuracy.c
```diff
@@ -358,14 +422,41 @@ static int start_timers(test_global_t *test_global)
...
+			printf("Capabilities: min base freq %g Hz, max base freq %g Hz, "
+			       "max res %" PRIu64 " Hz\n",
+			       odp_fract_u64_to_dbl(&timer_capa.periodic.min_base_freq_hz),
+			       odp_fract_u64_to_dbl(&timer_capa.periodic.max_base_freq_hz),
+			       timer_capa.max_res.res_hz);
+			return -1;
+		}
+
+		res_ns = test_global->opt.res_ns = capa.res_ns;
+		freq = test_global->opt.freq = capa.base_freq_hz;
```
MatiasElo, line 443: Related to Petri's comment, if the base frequency changes here, the original requested frequency should still be used  `run_test()` to detect drift.

example/timer/odp_timer_accuracy.c
JereLeppanen: I figured double is probably enough. But if the significant bits of (ctx->count * period) don't fit into a double's significand, then we lose precision. Which is definitely possible with a high frequency and a long test.

I actually have another implementation which uses an integer fraction of tmo_ns, effectively forming a 64.63 bit fixed point tmo_ns. That code looked slightly less elegant to me.

example/timer/odp_timer_accuracy.c
JereLeppanen: One period added in v5. I will post a separate PR to fix the timer implementation correspondingly.

JereLeppanen:
All fixed and warmup option added in v5.

example/timer/odp_timer_accuracy.c
```diff
@@ -609,6 +788,18 @@ static void run_test(test_global_t *test_global)
 		tmo = odp_timeout_from_event(ev);
 		ctx = odp_timeout_user_ptr(tmo);
 		tmo_ns = ctx->nsec;
+		if (mode == MODE_PERIODIC) {
+			ctx->count++;
```
MatiasElo, line 792: Looks like this isn't correct if `start_param.first_tick != 0`.

example/timer/odp_timer_accuracy.c
```diff
@@ -609,6 +788,18 @@ static void run_test(test_global_t *test_global)
...
+			tmo_ns += ctx->count * period + .5;
```
MatiasElo, line 793: Why the `+ .5` addition? At least a comment would be nice.

example/timer/odp_timer_accuracy.c
```diff
@@ -482,10 +493,14 @@ static int start_timers(test_global_t *test_global)
 	printf("  min timeout:     %" PRIu64 " nsec\n", timer_param.min_tmo);
 	printf("  max timeout:     %" PRIu64 " nsec\n", timer_param.max_tmo);
 	printf("  num timeout:     %" PRIu64 "\n", num_tmo);
+	printf("  num warmup:      %" PRIu64 "\n", num_warmup);
 	printf("  burst size:      %" PRIu64 "\n", burst);
 	printf("  burst gap:       %" PRIu64 "\n", burst_gap);
 	printf("  total timers:    %" PRIu64 "\n", tot_timers);
+	printf("  warmup timers:   %" PRIu64 "\n", test_global->warmup_timers);
 	printf("  alloc timers:    %" PRIu64 "\n", alloc_timers);
+	printf("  warmup time:     %.2f sec\n\n",
```
MatiasElo, line 502: Extra newline.

JereLeppanen:
All fixed in v7.

MatiasElo (approved):

PR #1766, opened 2023-02-14
[PATCH v3] fix a couple of cases of uninitialized data

JereLeppanen:
Fix a couple of cases where potentially uninitialized data in shm blocks are being used. These problems were revealed by filling all shm blocks with 0xaa.

v2:
- Initialize IP header fields before checksum.

v3:
- Rebase.
- Add review tags.

### DIFF
--- a/test/validation/api/pktio/pktio.c
+++ b/test/validation/api/pktio/pktio.c
@@ -300,6 +300,8 @@ static void pktio_init_packet_eth_ipv4(odp_packet_t pkt, uint8_t proto)
 	seq = odp_atomic_fetch_inc_u32(&ip_seq);
 	ip->id = odp_cpu_to_be_16(seq);
 	ip->chksum = 0;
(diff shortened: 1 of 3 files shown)


test/validation/api/pktio/pktio.c
```diff
@@ -301,6 +301,8 @@ static void pktio_init_packet_eth_ipv4(odp_packet_t pkt, uint8_t proto)
 	ip->id = odp_cpu_to_be_16(seq);
 	ip->chksum = 0;
 	odph_ipv4_csum_update(pkt);
+	ip->frag_offset = 0;
+	ip->tos = 0;
```
psavol, line 305: Fields should be written before checksum calculation

JereLeppanen: Fixed in v2. No effect on the tests since the tests update the checksum again.

psavol (approved):

PR #1770, opened 2023-02-19
[PATCH v5] validation: crypto: test combined ciphering and authentication

JannePeltonen:
validation: crypto: fix ignoring of leftover bits of bit mode ciphers
validation: crypto: test both cipher/auth orders with single alg
validation: crypto: run all tests only when FULL_TEST env variable is set
validation: crypto: test combined ciphering and authentication

### DIFF
--- a/test/validation/api/crypto/odp_crypto_test_inp.c
+++ b/test/validation/api/crypto/odp_crypto_test_inp.c
@@ -571,7 +571,7 @@ static int is_in_range(uint32_t offs, uint32_t range_offs, uint32_t range_len)
  */
 typedef struct ignore_t {
 	uint32_t byte_offset;	/* offset to a byte which has bits to be ignored */
-	uint32_t byte_mask;	/* mask of ignored bits in the byte */
+	uint8_t byte_mask;	/* mask of ignored bits in the byte */
 	struct {
 		uint32_t offset;
 		uint32_t length;
@@ -619,10 +619,13 @@ static void prepare_ignore_info(const alg_test_param_t *param,
 	 */
 	if (param->is_bit_mode_cipher &&
 	    param->cipher_alg != ODP_CIPHER_ALG_NULL) {
-		uint8_t leftover_bits = param->ref->length % 8;
+		uint8_t leftover_bits = ref_length_in_bits(param->ref) % 8;
 
 		ignore->byte_offset = cipher_offset + cipher_len - 1 + shift;
-		ignore->byte_mask = ~(0xff << (8 - leftover_bits));
+		if (leftover_bits > 0)
+			ignore->byte_mask = ~(0xff << (8 - leftover_bits));
+		else
+			ignore->byte_mask = 0;
 	}
 
 	/*
--- a/test/validation/api/crypto/odp_crypto_test_inp.c
+++ b/test/validation/api/crypto/odp_crypto_test_inp.c
@@ -19,6 +19,8 @@
  */
 static int full_test;
 
+#define MAX_FAILURE_PRINTS 20
+
 #define PKT_POOL_NUM  64
 #define PKT_POOL_LEN  (1 * 1024)
 #define UAREA_SIZE 8
@@ -164,8 +166,8 @@ static int alg_op(odp_packet_t pkt,
 		  odp_crypto_session_t session,
 		  uint8_t *cipher_iv_ptr,
 		  uint8_t *auth_iv_ptr,
-		  odp_packet_data_range_t *cipher_range,
-		  odp_packet_data_range_t *auth_range,
+		  const odp_packet_data_range_t *cipher_range,
+		  const odp_packet_data_range_t *auth_range,
 		  uint8_t *aad,
 		  unsigned int hash_result_offset)
 {
@@ -248,8 +250,8 @@ static int alg_packet_op(odp_packet_t pkt_in,
 			 int32_t oop_shift,
 			 uint8_t *cipher_iv_ptr,
 			 uint8_t *auth_iv_ptr,
-			 odp_packet_data_range_t *cipher_range,
-			 odp_packet_data_range_t *auth_range,
+			 const odp_packet_data_range_t *cipher_range,
+			 const odp_packet_data_range_t *auth_range,
 			 uint8_t *aad,
 			 unsigned int hash_result_offset)
 {
@@ -349,8 +351,8 @@ static int crypto_op(odp_packet_t pkt_in,
 		     int32_t oop_shift,
 		     uint8_t *cipher_iv,
 		     uint8_t *auth_iv,
-		     odp_packet_data_range_t *cipher_range,
-		     odp_packet_data_range_t *auth_range,
+		     const odp_packet_data_range_t *cipher_range,
+		     const odp_packet_data_range_t *auth_range,
 		     uint8_t *aad,
 		     unsigned int hash_result_offset)
 {
@@ -414,8 +416,10 @@ static void adjust_segments(odp_packet_t *pkt, uint32_t first_seg_len)
 
 static void fill_with_pattern(uint8_t *buf, uint32_t len)
 {
+	static uint8_t value;
+
 	for (uint32_t n = 0; n < len; n++)
-		buf[n] = n;
+		buf[n] = value++;
 }
 
 static void write_header_and_trailer(odp_packet_t pkt,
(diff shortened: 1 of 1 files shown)


test/validation/api/crypto/odp_crypto_test_inp.c
```diff
@@ -1193,7 +1202,9 @@ static void check_alg(odp_crypto_op_t op,
 			continue;
 		}
 
-		alg_test(op, cipher_alg, auth_alg, &ref[idx], digest_offs,
+		alg_test(op, cipher_alg, auth_alg, AUTH_PLAINTEXT, &ref[idx], digest_offs,
+			 is_bit_mode_cipher, is_bit_mode_auth);
+		alg_test(op, cipher_alg, auth_alg, AUTH_CIPHERTEXT, &ref[idx], digest_offs,
```
psavol: I'd update spec text to say that value of auth_cipher_text is ignored, when operation is auth or crypto only. Tests can leave the value then to default and less combinations to test for those operations.

JannePeltonen: That spec update would not really change anything since auth_cipher_text does not anyway have any effect when one of the algorithms is null. So saying that it is ignored does not change the semantics of the API at all. It also does not change the fact that regardless of what the API says, an implementation may have a bug that it only works if auth_cipher_text is set in a certain way, in which case testing both values reveals the problem.

test/validation/api/crypto/odp_crypto_test_inp.c
```diff
@@ -1624,9 +1758,512 @@ static void test_auth_hashes_in_auth_range(void)
...
+	fill_with_pattern(ref->cipher_key, ref->cipher_key_length);
+	fill_with_pattern(ref->cipher_iv, ref->cipher_iv_length);
+	fill_with_pattern(ref->auth_key, ref->auth_key_length);
+	fill_with_pattern(ref->auth_iv, ref->auth_iv_length);
+	fill_with_pattern(ref->plaintext, ref->length);
+	memset(ref->plaintext + digest_offset, 0, ref->digest_length);
+
+	if (suite->order == AUTH_CIPHERTEXT) {
+		first_cipher = suite->cipher;
+		first_auth = ODP_AUTH_ALG_NULL;
```
vvelumuri, line 1883: Above (1853, 1854) auth fields in ref gets populated from capabilities of suite->auth. 
Here the auth is set to NULL, instead of suite->auth. 
ref should get populated based on the final cipher used in operation.

test/validation/api/crypto/odp_crypto_test_inp.c
```diff
@@ -1624,9 +1758,512 @@ static void test_auth_hashes_in_auth_range(void)
...
+		second_cipher = ODP_CIPHER_ALG_NULL;
+		second_auth = suite->auth;
+	} else {
+		first_cipher = ODP_CIPHER_ALG_NULL;
```
vvelumuri, line 1887: Similarly, cipher fields in ref gets populated from capabilities of suite->cipher, but here the cipher is set to NULL
ref should get populated based on the final cipher used in operation.

test/validation/api/crypto/odp_crypto_test_inp.c
```diff
@@ -1566,6 +1653,9 @@ static void test_auth_hash_in_auth_range(odp_auth_alg_t auth,
 	if (create_hash_test_reference(auth, capa, &ref, digest_offset, 0))
 		return;
 
+	crypto_range.offset = 0;
+	crypto_range.length = ref.length;
```
vvelumuri, line 1657: The length should be converted to bits if the cipher supports bit mode.

test/validation/api/crypto/odp_crypto_test_inp.c
JannePeltonen: > ref should get populated based on the final auth used in operation.

I do not understand this comment and I cannot see a problem in the code. Maybe there is some confusion on what this function is doing.

This function is preparing the test reference for a combined op without using combined ops, i.e. by using cipher-only and auth-only operations to perform ciphering and authentication. So, depending on the cipher/auth order needed, the first operation is doing only ciphering or only authentication and then the second operation continues from there and processes the data further using the other operation.

test/validation/api/crypto/odp_crypto_test_inp.c
JannePeltonen: Actually, there is no cipher used here, only authentication. Auth range should be in bits if the auth algorithm uses bit mode, which is a bug and fixed in v2. In v2 I also reinitialize auth range after the second ref creation. It is not needed now but makes things more robust in case something changes in the ref creation.

JannePeltonen:
v2:
- Rebased
- Fixed auth range in has-in-auth-range test when auth alg is in bit mode.

test/validation/api/crypto/odp_crypto_test_inp.c
anoobj: @JannePeltonen 

```
	ref->cipher_key_length = suite->cipher_capa->key_len;
	ref->cipher_iv_length = suite->cipher_capa->iv_len;
	ref->auth_key_length = suite->auth_capa->key_len;
	ref->auth_iv_length = suite->auth_capa->iv_len;
	ref->digest_length = suite->auth_capa->digest_len;
	ref->is_length_in_bits = false;
	ref->length = total_len;
```

With usage of first_auth = ODP_AUTH_ALG_NULL, do we make sure key_length, digest_size etc matches NULL's requirements? ie, digest length & key length is 0.

JannePeltonen: > do we make sure key_length, digest_size etc matches NULL's requirements? ie, digest length & key length is 0.

That is good question and the test code does not try to ensure that the lengths etc are zero for null algorithms. Currently the code assumes that key length, digest_length, iv_length etc are ignored when the cipher/auth algorithm is ODP_CIPHER_ALG_NULL/ODP_AUTH_ALG_NULL and the API text appears to support that since it says: "ODP_CIPHER_ALG_NULL indicates that ciphering is disabled" and "When ciphering is enabled cipher key and IV need to be set". And similar for the auth case.

So, I think that either we conclude that the current API is ok but maybe needs some clarification to prevent confusion. And the test code is not buggy. Or we decide to change the API and say that key lengths etc must be set in certain way for the null algorithms and update the test accordingly.

test/validation/api/crypto/odp_crypto_test_inp.c
```diff
@@ -977,10 +1035,7 @@ static odp_crypto_session_t session_create(odp_crypto_op_t op,
 	}
 
 	/*
-	 * We do not allow ODP_CRYPTO_SES_ERR_ALG_ORDER since we do
-	 * not combine individual non-null crypto and auth algorithms
-	 * with each other in the tests. Both orders should work when
-	 * only one algorithm is used (i.e. the other one is null).
+	 * We do not allow ODP_CRYPTO_SES_ERR_ALG_ORDER.
```
anoobj, line 1038: I think this may also need some adjustments.

Earlier, the UT would attempt only single algo, ie, any non-NULL algo + NULL. So, the order would not matter. But now that we attempt chained cases, we should handle ALG_ORDER also the same way as ALG_COMBO.

test/validation/api/crypto/odp_crypto_test_inp.c
JannePeltonen: I tweaked the ref building code a bit and now it sets the unused auth/cipher session parameters to zeroes even if it may not be (?) strictly required by the API.

I also zero the crypto/auth range of the null algorithm now. That is not required by the API, I think, but I would actually like to make it a requirement so that non-zero range would explicitly not be allowed. In the future non-zero ranges could then be allowed with yet-to-be-defined semantics (e.g. pure data copying in the OOP op type). I think I will change the other PR to require zero ranges for null algorithms instead of saying that the ranges are to be ignored.

test/validation/api/crypto/odp_crypto_test_inp.c
JannePeltonen: I think an ODP implementation should be able to support both orders by resorting internally to two individual crypto operations. But since that may not be trivial in the async mode, I changed the code so that ALG_ORDER error is allowed without test failure in async operation mode.

JannePeltonen:
v3:
- Allow ALG_ORDER session creation error without test failure in async mode
- zero unused session parameters for null algorithms
- zero crypto ranges for null algorithms

JannePeltonen:
v4: fixed white space issues found by checkpatch

anoobj:
@JannePeltonen Thanks for the changes. The tests look good.

JannePeltonen:
v5: added reviewed-by tags

PR #1771, opened 2023-02-21
[PATCH v7] api: crypto: misc changes, deprecations and removal of deprecated pieces

JannePeltonen:
API changes in this PR:

api: crypto: change IV and AAD pointers to pointers to constants
api: crypto: clarify that algorithm order is ignored with null algorithms
api: crypto: clarify when cipher and auth ranges are ignored
api: crypto: deprecate old session creation error names
api: crypto: simplify operation status reporting
api: crypto: remove deprecated odp_crypto_operation()


platform/linux-generic/include-abi/odp/api/abi/event_types.h
```diff
@@ -33,9 +33,6 @@ typedef enum odp_event_type_t {
 	ODP_EVENT_BUFFER = 1,
 	ODP_EVENT_PACKET = 2,
 	ODP_EVENT_TIMEOUT = 3,
-#if ODP_DEPRECATED_API
-	ODP_EVENT_CRYPTO_COMPL = 4,
-#endif
 	ODP_EVENT_IPSEC_STATUS = 5,
```
anoobj, line 36: Would changing enum value of 'ODP_EVENT_IPSEC_STATUS' to 4 be okay? Like, currently there would be a hole for enumeration, right?

JannePeltonen: I did not want to change the other values in case someone relies on them (even if they should not) and because I just did not see any reason for such changes. The gap should not cause any issues and I think we have similar gaps in some other enums too. But I think we could remove the gap too if there is a reason to do so.

anoobj: If we have similar gaps in other enums, I think it's okay to have the same here as well.

JannePeltonen:
v2:
- changed odp_crypto_result() to indicate operation success in the return value
- deprecated odp_crypto_packet_result_t::ok
- adjusted implementation and validation tests for the above changes (and trying to keep every revision pass validation test)

JannePeltonen:
v3:
- added a patch that clarifies that certain session parameters are ignored by the null algorithms
- added a patch that specifies that null algorithms in the out-of-place op type must be used with zero range
- Slighly reworded the API clarification that auth_cipher_text is ignored when one of the algorithms is null
- rebased

anoobj:

JannePeltonen:
v4: rebased & added reviewed-by tags

include/odp/api/spec/crypto_types.h
```diff
@@ -897,27 +897,29 @@ typedef enum {
 typedef enum {
 	/** Algorithm successful */
 	ODP_CRYPTO_ALG_ERR_NONE,
-	/** Invalid data block size */
+	/** Invalid range or input or output packet size */
```
psavol, line 823: If there is no reason to explicitly mention input/output packet, this could be simply:
"Invalid data range or packet size"

op_param_t use term "data range", so I'd use that also here.

```
	/** Data range to apply cipher */
	odp_packet_data_range_t cipher_range;

```

include/odp/api/spec/crypto_types.h
```diff
@@ -958,7 +959,8 @@ typedef struct odp_crypto_op_result {
  * Crypto packet API operation result
  */
 typedef struct odp_crypto_packet_result_t {
-	/** Request completed successfully */
+	/** Request completed successfully. True when odp_crypto_result()
+	 *  returned zero. This field will be deprecated in the future */
```
psavol: Period (.) missing in the end.

include/odp/api/spec/crypto_types.h
JannePeltonen: changed in v5

include/odp/api/spec/crypto_types.h
JannePeltonen: fixed as a side effect of squashing with the commit that deprecates the ok field

JannePeltonen:
v5: As suggested by Petri, squashed two API commits (crypto result simplification & deprecation of the ok field). Reordered the patches to make the squashing possible.

PR #1773, opened 2023-02-28
[PATCH v7] Add -Os build targets to CI

psavol:


.github/workflows/ci-pipeline-arm64.yml
```diff
@@ -19,7 +19,7 @@ jobs:
       fail-fast: false
       matrix:
         cc: [gcc, clang]
-        conf: ['', 'CFLAGS=-O3', 'CFLAGS=-O1', 'CFLAGS=-O0 --enable-debug=full', 'CFLAGS=-pedantic',
+        conf: ['', 'CFLAGS=-O3', 'CFLAGS=-O1', 'CFLAGS=-O0 --enable-debug=full', 'CFLAGS=-Os', 'CFLAGS=-pedantic',
```
JannePeltonen, line 22: How about combining -Os and -pedantic into one conf variant to keep the number of test builds lower?

JannePeltonen (approved):
Looks ok, but consider combining e.g. -Os and -pedantic.

PR #1782, opened 2023-03-14
[PATCH v7] linux-gen: sched: improve timer behavior when power saving sleep is enabled

JereLeppanen:
When using inline timers and power saving sleep, limit sleep duration according to the next timeout.

### DIFF
--- a/platform/linux-generic/odp_schedule_basic.c
+++ b/platform/linux-generic/odp_schedule_basic.c
@@ -1698,16 +1698,13 @@ static inline int schedule_loop_sleep(odp_queue_t *out_queue, uint64_t wait,
 			nanosleep(&ts, NULL);
 		}
 
-		if (wait != ODP_SCHED_WAIT || !sleep) {
+		if (!sleep || wait != ODP_SCHED_WAIT)
 			current = odp_time_local();
-			if (odp_time_cmp(start_sleep, current) < 0)
-				sleep = 1;
-		}
 
-		if (wait == ODP_SCHED_WAIT)
-			continue;
+		if (!sleep && odp_time_cmp(start_sleep, current) < 0)
+			sleep = 1;
 
-		if (odp_time_cmp(end, current) < 0)
+		if (wait != ODP_SCHED_WAIT && odp_time_cmp(end, current) < 0)
 			break;
 	}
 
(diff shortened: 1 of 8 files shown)


platform/linux-generic/odp_schedule_basic.c
```diff
@@ -1686,8 +1686,15 @@ static inline int schedule_loop_sleep(odp_queue_t *out_queue, uint64_t wait,
...
 		}
 
-		if (sleep)
-			nanosleep(&sched->powersave.sleep_time, NULL);
+		if (sleep && next) {
+			uint64_t sleep_time = _ODP_MIN(sched->powersave.sleep_time, next);
+			struct timespec ts = {
+				sleep_time / 1000000000,
+				sleep_time % 1000000000,
+			};
```
psavol, line 1694: I think sleep longer than 1sec does not need to be supported. tc_sec could be always 0, and tv_nsec value could be just masked with 0x1FFF FFFF to ensure that it's <= 0.536 sec.

JereLeppanen: It's even easier to just cap the conf variable to less than one second. In v2.

JereLeppanen:
v4:
- Rebase.

JereLeppanen:
v5:
- Interrupt sleep when a timer is started.
- Add some cleanups.

platform/linux-generic/odp_timer.c
```diff
@@ -844,26 +850,28 @@ static inline void timer_pool_scan_inline(int num, odp_time_t now)
 					odp_atomic_store_u32(&tp->notify_overrun, 2);
 				}
 			}
-			timer_pool_scan(tp, nsec);
+			min = _ODP_MIN(min, timer_pool_scan(tp, nsec));
```
psavol, line 870: Code would be more readable (and robust against macro issues) when a function call is not inside macro. Especially in this case when timer scan is the beef of this function. The min comparison becomes also more readable with variable naming e.g.:

ticks_to_next_expire = timer_pool_scan();
min = MIN(min, ticks_to_next_expire);

JereLeppanen:
v6:
- Rebase.
- Move function calls outside _ODP_MIN().
- Remove commits `linux-gen: sched: clean up _odp_schedule_init_global()` and `linux-gen: sched: interrupt power saving sleep when a timer is started`.
- Add commit `linux-gen: sched: avoid odp_time_cmp() in schedule_loop_sleep()`.

psavol (approved):

JereLeppanen:
v7:
- Rebase.
- Add review tags.

PR #1787, opened 2023-03-27
[PATCH v4] linux-gen, helper: re-enable print format warnings

JereLeppanen:
Print format warnings were disabled as a side effect of enabling compilation with `-pedantic`. Re-enable them.

### DIFF
--- a/helper/include/odp/helper/odph_debug.h
+++ b/helper/include/odp/helper/odph_debug.h
@@ -64,7 +64,7 @@ typedef enum odph_log_level {
 #define ODPH_LOG(level, fmt, ...) \
 do { \
 	if (level != ODPH_LOG_DBG || ODPH_DEBUG_PRINT == 1) \
-		fprintf(stderr, "%s:%d:%s():" fmt, __FILE__, \
+		fprintf(stderr, "%s:%d:%s(): " fmt, __FILE__, \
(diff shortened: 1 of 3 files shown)


JereLeppanen:
v2:
- Fix checkpatch warnings.

helper/include/odp/helper/odph_debug.h
```diff
@@ -63,25 +63,11 @@ typedef enum odph_log_level {
  */
 #define ODPH_LOG(level, fmt, ...) \
 do { \
-	switch (level) { \
-	case ODPH_LOG_ERR: \
+	if (level != ODPH_LOG_DBG || ODPH_DEBUG_PRINT == 1) \
 		fprintf(stderr, "%s:%d:%s():" fmt, __FILE__, \
```
TuomasTaipale, line 67: Matter of taste, but there could be a whitespace after the location info -> `"%s:%d:%s(): "`, makes the entries slightly more readable.

MatiasElo: Agreed, this would make reading log messages easier. The rest of the print is application specific anyway, so this change shouldn't affect any reasonable log parsing.

PR needs rebase to pass CI.

TuomasTaipale (approved):

PR #1790, opened 2023-03-29
test: ipsecfwd: utilize libconfig

TuomasTaipale:
This patchset adds `libconfig` support to `odp_ipsecfwd`. Tester is now configured with a single libconfig-syntax configuration file, in which SAs and forwarding is to be configured. Most of the configuration passed with `odp_ipsec_sa_param_t` can be set via this new configuration file.


test/performance/odp_ipsecfwd.c
```diff
@@ -203,6 +251,8 @@ static __thread pkt_ifs_t ifs;
 static void init_config(prog_config_t *config)
 {
 	memset(config, 0, sizeof(*config));
+	odp_ipsec_sa_param_init(&config->ovr_cfg.sa_param);
+	memset(&config->ovr_cfg.sa_param.crypto, 0, sizeof(config->ovr_cfg.sa_param.crypto));
```
JannePeltonen, line 255: Why is this memset needed?

test/performance/odp_ipsecfwd.c
```diff
@@ -949,28 +999,337 @@ static odp_bool_t create_sa_dest_queues(odp_ipsec_capability_t *ipsec_capa,
...
+
+	return false;
+}
+
+static void parse_sa_entry(config_t *cfg, const char *root, sa_config_t *config)
+{
+	int val;
+	char val_str[16U];
+
+	if (lookup_int(cfg, root, "sa_dir", (int *)&config->sa_param.dir))
```
JannePeltonen, line 1078: This function assumes that the size of the enums is the same of int, which is typically true but not always. It would be nice to not make the assumption.

test/performance/odp_ipsecfwd.c
```diff
@@ -949,28 +999,337 @@ static odp_bool_t create_sa_dest_queues(odp_ipsec_capability_t *ipsec_capa,
...
+		config->is_set.dir = 1U;
+
+	if (lookup_int(cfg, root, "sa_proto", (int *)&config->sa_param.proto))
```
JannePeltonen, line 1081: I think it might be better if the config entries were named exactly as the fields of the SA parameter struct, i.e. without the sa_ prefix, 'sa_out' renamed to 'outbound' and so forth.

test/performance/odp_ipsecfwd.c
```diff
@@ -1023,18 +1353,31 @@ static void create_sa_entry(uint32_t dir, uint32_t spi, const char *src_ip_str,
...
+	parse_sa_entry(cfg, "default.", &config->ovr_cfg);
+
+	for (uint32_t i = 0U; i < MAX_SAS; i++) {
+		char path[strlen("saXXXX.") + 1U];
+		sa_config_t sa_cfg;
+		int val;
+
+		odp_ipsec_sa_param_init(&sa_cfg.sa_param);
+		memset(&sa_cfg.sa_param.crypto, 0, sizeof(sa_cfg.sa_param.crypto));
+		snprintf(path, sizeof(path), "sa%u.", i);
```
JannePeltonen, line 1389: Perhaps it would be be better to have SAs inside a list in the config file so that no special naming would be needed and this code could just iterate over the list (using config_setting_get_elem()) instead of trying to read every possible sa config group? Similarly, fwd entries could be in a list.

test/performance/odp_ipsecfwd.c
```diff
@@ -365,14 +365,37 @@ static void print_usage(void)
 	       "Simple IPsec performance tester. Forward and process plain and ipsec packets.\n"
 	       "\n"
 	       "Examples:\n"
-	       "    %s -i ens9f1 -s /etc/odp/sa.conf -f /etc/odp/fwd.conf\n"
+	       "    %s -i ens9f1 -C /etc/odp/ipsecfwd.conf\n"
```
JannePeltonen, line 368: I think this commit ("adjust help ...") could be squashed with the previus one that changes condig scheme as I do not see a benefit in keeping the help text update separate.

test/performance/odp_ipsecfwd.c
```diff
@@ -365,14 +365,37 @@ static void print_usage(void)
...
+	       "            sa_out: {\n"
+	       "                tunnel: {\n"
+	       "                    src_addr = \"192.168.1.10\"\n"
+	       "                    dst_addr = \"192.168.1.16\"\n"
+	       "                };\n"
+	       "            };\n"
+	       "        };\n"
+	       "\n"
+	       "        fwd0: {\n"
+	       "            prefix: \"192.168.1.16/24\"\n"
```
JannePeltonen, line 395: I guess this should be 192.168.1.0/24 unless the non-prefix part of the address has some signifigance here which I do not think it has. And the same comment for the config file example.

m4/odp_libconfig.m4
```diff
@@ -5,7 +5,8 @@ AC_DEFUN([ODP_LIBCONFIG],
 ##########################################################################
 # Check for libconfig availability
 ##########################################################################
-PKG_CHECK_MODULES([LIBCONFIG], [libconfig])
+PKG_CHECK_MODULES([LIBCONFIG], [libconfig], [have_libconfig=yes])
+AM_CONDITIONAL([LIBCONFIG], [test x$have_libconfig = xyes])
```
JannePeltonen, line 9: Would it be better to have this commit as the first commit so that build without libconfig present does not break in between?

test/performance/odp_ipsecfwd.c
TuomasTaipale: Not needed, fixed in v3.

test/performance/odp_ipsecfwd.c
TuomasTaipale: Perhaps safer to handle them via a helper, changed in v3.

test/performance/odp_ipsecfwd.c
TuomasTaipale: V3.

test/performance/odp_ipsecfwd.c
TuomasTaipale: Swapped to list(s) in v3.

test/performance/odp_ipsecfwd.c
TuomasTaipale: V3.

test/performance/odp_ipsecfwd.c
TuomasTaipale: Copy paste error, no significance, fixed in v3.

m4/odp_libconfig.m4
TuomasTaipale: Sure, v3.

test/performance/odp_ipsecfwd.c
```diff
@@ -949,25 +994,203 @@ static odp_bool_t create_sa_dest_queues(odp_ipsec_capability_t *ipsec_capa,
...
+	}
+
+	if (config_setting_lookup_string(cs, "auth_key_extra", &val_str) == CONFIG_TRUE) {
+		strcpy(config->auth_key_extra, val_str);
+		config->sa_param.crypto.auth_key_extra.data = (uint8_t *)config->auth_key_extra;
+		config->sa_param.crypto.auth_key_extra.length =
+			strlen((const char *)config->auth_key_extra);
+	}
+
+	(void)config_setting_lookup_int(cs, "icv_len", (int *)&config->sa_param.crypto.icv_len);
```
JannePeltonen, line 1040: This could follow the same pattern as the other settings so that the casting (and the assumption that icv_len is the same size as int) would not be needed. And the same for mtu and spi parameters. With that, Reviewed-by: Janne Peltonen <janne.peltonen@nokia.com>

test/performance/odp_ipsecfwd.c
```diff
@@ -336,39 +385,35 @@ static void print_usage(void)
...
-	       "                      expected to be within a single line, values whitespace\n"
-	       "                      separated:\n"
-	       "\n"
-	       "                          <line in file> Dir SPI TunSrcIPv4 TunDstIPv4"
-	       " CipherAlgoIdx CipherKey CipherKeyExtra AuthAlgIdx AuthKey AuthKeyExtra ICVLen\n"
+	       "  -C, --conf          Configuration file. 'libconfig' syntax is expected.\n"
+	       "                      SA configuration supports default fallback, i.e.\n"
+	       "                      individual SA configuration blocks may omit some\n"
+	       "                      parameters and instead set these once in default block\n"
+	       "                      which then are used to fill missing parameters. Only\n"
```
JannePeltonen, line 392: Add "the": "The only required argument..."

test/performance/odp_ipsecfwd.c
TuomasTaipale: V9.

test/performance/odp_ipsecfwd.c
TuomasTaipale: V9.

JannePeltonen (approved):

PR #1803, opened 2023-04-18
linux-gen: pktio: don't try to receive packets from non-active input queues

MatiasElo:
Don't try to receive packets from non-active pktio devices when using plain event input queues. Direct and scheduled packet input modes already check pktio device status before polling for new packets. This fixes a crash with DPDK v22.11.


TuomasTaipale (approved):
In commit log: maybe non-active -> inactive, otherwise OK

PR #1804, opened 2023-04-18
[PATCH v3] linux-gen: pktio: enqueue packets to destination queues from pktio devices

MatiasElo:
Reduce latency by enqueueing packets directly to classifier and inline
IPsec destination queues from pktio devices.

### DIFF
--- a/platform/linux-generic/include/odp_classification_internal.h
+++ b/platform/linux-generic/include/odp_classification_internal.h
@@ -1,5 +1,5 @@
 /* Copyright (c) 2014-2018, Linaro Limited
- * Copyright (c) 2021, Nokia
+ * Copyright (c) 2021-2023, Nokia
  * All rights reserved.
  *
  * SPDX-License-Identifier:     BSD-3-Clause
@@ -21,12 +21,19 @@ extern "C" {
 
 #include <odp/api/atomic.h>
 #include <odp/api/classification.h>
+#include <odp/api/event.h>
 #include <odp/api/hints.h>
+#include <odp/api/packet.h>
+#include <odp/api/pool.h>
 #include <odp/api/queue.h>
+#include <odp/api/std_types.h>
+
+#include <odp_debug_internal.h>
 #include <odp_packet_internal.h>
-#include <odp/api/packet_io.h>
 #include <odp_packet_io_internal.h>
 #include <odp_classification_datamodel.h>
+
+#include <stdint.h>
 
 extern cls_global_t *_odp_cls_global;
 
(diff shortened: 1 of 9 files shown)


platform/linux-generic/include/odp_classification_internal.h
```diff
@@ -72,6 +77,97 @@ static inline void _odp_cos_queue_stats_add(cos_t *cos, odp_queue_t queue,
...
+ * actual enqueue operation and stores packets in 'packets' array until destination queue or CoS
+ * change, or 'last' flag is set.
+ *
+ * @param[out] packets Packet array to be enqueued
+ * @param      num     Number of handles in 'packets' array
+ * @param      last    Enqueue all packets
+ *
+ * @return Number of packets remaining in 'packets' array
+ */
+static inline int _odp_cls_enq(odp_packet_t packets[], int num, odp_bool_t last)
```
TuomasTaipale, line 141: I wonder if the code could be simplified if some state was kept in a helper struct. Probably the packet array, number of packets stored, current queue/cos, the total number of packets to be processed would be needed. Drivers would pass the struct, a packet and current index. The "last" step could then be done without a separate function call and flag. Any sense?

platform/linux-generic/odp_packet_io.c
```diff
@@ -894,102 +864,19 @@ static inline int pktin_recv_buf(pktio_entry_t *entry, int pktin_index,
...
-
-			if (cur_queue != pkt_hdr->dst_queue || cur_cos != pkt_hdr->cos) {
-				cur_queue = pkt_hdr->dst_queue;
-				cur_cos = pkt_hdr->cos;
-				dst[num_dst] = cur_queue;
-				cos[num_dst] = cur_cos;
-				dst_idx[num_dst] = num_ev;
-				num_dst++;
-			}
+	/* Create packet vector */
```
TuomasTaipale, line 872: The below function name is descriptive so this is slightly unnecessary.

platform/linux-generic/pktio/dpdk.c
```diff
@@ -160,6 +160,9 @@ typedef struct {
 typedef struct ODP_ALIGNED_CACHE {
 	/* --- Fast path data --- */
 
+	/* Function for mbuf to ODP packet conversion */
+	int (*mbuf_to_pkt_fn)(pktio_entry_t *pktio_entry, odp_packet_t pkt_table[],
```
TuomasTaipale, line 164: Good idea, could be worth considering similar treatment in other packet I/O drivers as well at some point.

platform/linux-generic/odp_packet_io.c
MatiasElo: Removed in V2.

platform/linux-generic/include/odp_classification_internal.h
```diff
@@ -22,14 +22,19 @@ extern "C" {
...
+
+#include <odp_debug_internal.h>
 #include <odp_packet_internal.h>
-#include <odp/api/packet_io.h>
 #include <odp_packet_io_internal.h>
 #include <odp_classification_datamodel.h>
 
 extern cls_global_t *_odp_cls_global;
 
+void _odp_cos_vector_enq(odp_queue_t queue, odp_event_t events[], uint32_t num, cos_t *cos_hdr);
```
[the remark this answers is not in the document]
MatiasElo: Moved in V2.

platform/linux-generic/include/odp_classification_internal.h
MatiasElo: The main motivation for the current solution was to minimize the amount of changes required to the pktio device code, since the integration had to be done into seven different drivers. I'd prefer not changing `_odp_cls_enq()` function at this point, unless this is deemed necessary. Mainly because this would be a rather big change and so it would invalidate most of the testing done so far. However, this could be added as a classifier optimization task to the backlog.

TuomasTaipale: OK, perhaps a future optimization then.

platform/linux-generic/include/odp_classification_internal.h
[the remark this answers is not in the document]
TuomasTaipale: Meant that the declaration would be in the packet I/O internal header, but this is probably fine as well. Maybe it could be now defined as inline.

TuomasTaipale (approved):

platform/linux-generic/include/odp_classification_internal.h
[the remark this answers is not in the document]
MatiasElo: Sure, changed in V3. Thanks for the review!

PR #1806, opened 2023-04-25
[PATCH v5] api: event: add function for reading event user area address

MatiasElo:
Add odp_event_user_area() function which returns a pointer to the user area associated with the event. If the event does not have user area, NULL is returned.

This PR needs to rebased on top of #1811 before merge.

### DIFF
--- a/include/odp/api/spec/event.h
+++ b/include/odp/api/spec/event.h
@@ -1,5 +1,5 @@
 /* Copyright (c) 2015-2018, Linaro Limited
- * Copyright (c) 2022, Nokia
+ * Copyright (c) 2022-2023, Nokia
  * All rights reserved.
  *
  * SPDX-License-Identifier:     BSD-3-Clause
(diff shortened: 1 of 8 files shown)


include/odp/api/spec/event.h
```diff
@@ -79,6 +79,20 @@ odp_event_type_t odp_event_types(odp_event_t event,
...
+ * Returns pointer to the user area associated with the event. This maps to the
+ * user area of underlying event type (e.g. odp_packet_user_area() for packets).
+ * If the event does not have user area, NULL is returned.
+ *
+ * @param      event    Event handle
+ *
+ * @return Pointer to the user area of the event
+ * @retval NULL  The event does not have user area
+ */
+void *odp_event_user_area(odp_event_t event);
```
TuomasTaipale, line 94: Any sense providing a user area size API as well?

MatiasElo: We discussed about this during the last sync call and decided to not add such API at this point. The user area size function is currently available only for packets anyway.

TuomasTaipale (approved):

MatiasElo:
V2:
- Added implementation
- Added validation tests

psavol:
API spec OK. Implementation and test code need update after rebase on top of DMA user area PR.

MatiasElo:
V3:
- Rebased on top of #1811
- Added implementation and tests for DMA completion event user area

PavanNikhilesh:
Reviewed-by: Pavan Nikhilesh <pbhagavatula@marvell.com>

PR #1807, opened 2023-04-26
[PATCH v5] api: timer: new timer cancel return values

psavol:
Use same return values between timer start and cancel calls. Separate cancel failure return values allow application to distinguish timing related failures (TOO_NEAR) from severe errors (that should never happen).

v2: Rebased
v3: Rebased
v4: Rebased

### DIFF
--- a/include/odp/api/spec/timer_types.h
+++ b/include/odp/api/spec/timer_types.h
@@ -515,24 +515,49 @@ typedef struct odp_timer_periodic_start_t {
 } odp_timer_periodic_start_t;
 
 /**
- * Return values of timer set calls.
+ * Return values for timer start, restart and cancel calls
  */
 typedef enum {
-	/** Timer set operation succeeded */
+	/**
+	 * Timer operation succeeded
+	 *
+	 * Timer start, restart and cancel operations may return this value.
+	 */
 	ODP_TIMER_SUCCESS = 0,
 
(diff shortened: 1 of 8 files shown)


include/odp/api/spec/timer_types.h
```diff
@@ -505,24 +505,49 @@ typedef struct odp_timer_periodic_start_t {
...
 
-} odp_timer_set_t;
+} odp_timer_retval_t;
+
+/**
+ * For backwards compatibility, odp_timer_set_t is synonym of odp_timer_retval_t
+ *
+ * @deprecated Use odp_timer_retval_t instead.
+ */
+typedef odp_timer_retval_t odp_timer_set_t;
```
JereLeppanen, line 550: Same question as always: do we really need multiple stages of deprecation? Approved anyway.

JereLeppanen (approved):

include/odp/api/spec/timer_types.h
psavol: When we use this doxygen tag, it helps user to find all deprecated and to-be-deprecated API definitions in the same page in documentation. If user removes references to to-be-deprecated definitions early, build issues are avoided in the future.

PavanNikhilesh:
Reviewed-by: Pavan Nikhilesh <pbhagavatula@marvell.com>

PR #1808, opened 2023-04-26
linux-gen: timer: fix periodic timer drift

JereLeppanen:
Periodic timer interval has an accuracy of one nanosecond, which can cause drift with some base frequencies and / or multipliers. Fix by also accounting for the fractional part of the timer interval.


MatiasElo (approved):

JereLeppanen:
v2:
- Rebase.
- Add review tag.

PR #1809, opened 2023-05-02
[PATCH v2] linux-gen: dma: fix packet-type transfer segmentation

TuomasTaipale:
Correctly allow up to maximum capability of input segments in case of packet-type transfer.

v2:
- Added reviewed-by tag


MatiasElo (approved):
Looks good.

PR #1812, opened 2023-05-05
[PATCH v4] linux-gen: pktio: detect pending stop operations

MatiasElo:
Previously, if a stop operation was still pending for a scheduled pktio device when odp_pktio_start() was called, the input queues were added again to scheduler's internal queues causing random behaviour.

### DIFF
--- a/platform/linux-generic/odp_packet_io.c
+++ b/platform/linux-generic/odp_packet_io.c
@@ -652,6 +652,13 @@ int odp_pktio_start(odp_pktio_t hdl)
 		_ODP_ERR("Already started\n");
 		return -1;
 	}
+
+	if (entry->state == PKTIO_STATE_STOP_PENDING) {
+		unlock_entry(entry);
+		_ODP_ERR("Scheduled pktio stop pending\n");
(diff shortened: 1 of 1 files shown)


platform/linux-generic/odp_packet_io.c
```diff
@@ -652,6 +652,13 @@ int odp_pktio_start(odp_pktio_t hdl)
 		_ODP_ERR("Already started\n");
 		return -1;
 	}
+
+	if (entry->state == PKTIO_STATE_STOP_PENDING) {
+		unlock_entry(entry);
+		_ODP_ERR("Scheduled pktio stop pending. Retry after calling odp_schedule().\n");
```
TuomasTaipale, line 658: "Scheduled pktio stop pending" is probably enough here, otherwise OK.

TuomasTaipale (approved):

platform/linux-generic/odp_packet_io.c
MatiasElo: Fixed in V2.

PR #1815, opened 2023-05-08
[PATCH v9] validation: ipsec: disable anti-replay with null authentication

SruChalla:
validation: ipsec: disable anti-replay with null authentication

Set antireplay_ws to zero to disable anti-replay service when an inbound SA
without integrity protection is created.
The anti-replay service is not used in IPsec without integrity protection
and therefore some ODP implementations may not support SAs that have null
auth algorithm but anti-replay enabled. This change also enables a future
API change that would require zero antireplay_ws with null authentication.

### DIFF
--- a/test/validation/api/ipsec/ipsec.c
+++ b/test/validation/api/ipsec/ipsec.c
@@ -395,7 +395,10 @@ void ipsec_sa_param_fill(odp_ipsec_sa_param_t *param,
 	param->dir = dir;
 	if (dir == ODP_IPSEC_DIR_INBOUND) {
 		param->inbound.lookup_mode = ODP_IPSEC_LOOKUP_SPI;
-		param->inbound.antireplay_ws = capa.max_antireplay_ws;
+		if (auth_alg == ODP_AUTH_ALG_NULL)
+			param->inbound.antireplay_ws = 0;
+		else
+			param->inbound.antireplay_ws = capa.max_antireplay_ws;
 	}
 	param->proto = proto;
 


JannePeltonen:
The commit history of this branch needs to be fixed. The commit(s) in an ODP pull request should be rebased (as opposed to merged) on top of ODP master branch so that the commits form linear history. Take a look at the output of 'git log --oneline --graph new_branch' to see the current situation.

The commit messages should not have any Change-Id tags or other such metadata that is not used in ODP.

JannePeltonen:
The RFC indeed says that anti-replay service must not be used without integrity protection. However, the validation tests are supposed to test an ODP implementation against the ODP API spec, so the reference to the RFC alone is not a sufficient rationale for the change.

Currently ODP API seems to allow enabling anti-replay service even with null authentication or at least it seems to allow setting nonzero antireplay_ ws parameter with null authentication. In this light the validation test code does not appear wrong and one could argue that ODP implementations should be able to deal with what the validation test is doing.

That said, maybe we should make an API change that says that
1) antireplay_ws must be set to zero with null authentication
or
2) antireplay_ws is ignored and treated as effectively zero with null authentication

JannePeltonen:
If we decide to go with option 1) from the comment above, a better commit message could be something like this:

validation: ipsec: disable anti-replay with null authentication

Set antireplay_ws to zero to disable anti-replay service when an inbound
SA without integrity protection is created.

The anti-replay service is not used in IPsec without integrity protection
and therefore some ODP implementations may not support SAs that have null
auth algorithm but anti-replay enabled. This change also enables a future
API change that would require zero antireplay_ws with null authentication.

JannePeltonen:
Let's move the API change from this PR to a separate PR. This way we can merge the validation test change already now, but postpone merging of the API change until we can make backward-incompatible API changes.

Both patches: Reviewed-by: Janne Peltonen <janne.peltonen@nokia.com>

The subject line of the API commit could be improved, e.g.:
api: ipsec: require that anti-replay is disabled with null authentication

JannePeltonen (approved):

PR #1816, opened 2023-05-08
[PATCH v1] validation: timer: fix odp_fract_u64_t comparisons

JereLeppanen:
Since the API allows many different ways to represent a given rational number as an odp_fract_u64_t, do equality comparisons in a way that is agnostic to the representation.


JereLeppanen:
Maybe we'll change the API instead.

PR #1817, opened 2023-05-09
[PATCH v5] api: timer: allow leeway when periodic first_tick is zero

JereLeppanen:


include/odp/api/spec/timer_types.h
```diff
@@ -477,9 +477,9 @@ typedef struct odp_timer_periodic_start_t {
 	/** First expiration time
 	 *
 	 *  The first expiration time in absolute timer ticks. When zero, the first expiration time
-	 *  is one period after the current time. After the first expiration, timer expiration
-	 *  continues with the defined frequency. The tick value must be within one timer period
-	 *  from the current time.
+	 *  is one period after the current time, or as close to that as the implementation can
+	 *  achieve. After the first expiration, timer expiration continues with the defined
+	 *  frequency. The tick value must be within one timer period from the current time.
```
psavol, line 482: For the same reason, should "within" be changed to "less than" one period also here.

example/timer/odp_timer_accuracy.c
```diff
@@ -874,6 +874,19 @@ static void run_test(test_global_t *test_global)
...
+				/*
+				 * If first_tick is zero, the API allows the implementation to
+				 * place the timer where it can, so we just take the first timeout
+				 * as a baseline, if it's within a reasonable range.
+				 */
+				int64_t diff = (int64_t)time_ns - (int64_t)tmo_ns;
+
+				if (diff < 0)
+					diff *= -1;
+				if (diff < period_dbl / 2)
```
psavol, line 887: This would be simpler than multiply by -1:
 diff = -diff;
 
 Also, "period_dbl / 2" could be defined as a const variable to make the error margin more visible:
 
const int64_t error_margin = 0.5*period_dbl;

include/odp/api/spec/timer_types.h
JereLeppanen: I'm starting to remember why I used that word in the first place. So if it's "less than", then the first timeout can't be "one period after the current time". It would have to be something like "the first expiration time is one tick less than one period after the current time", or "the first expiration time is one period minus one tick after the current time".

JereLeppanen: Or maybe it's fine. First expiration is "one period", but tick value is "less than".

In v2, changed to "less than" and also changed "from the current time" to "after the current time".

example/timer/odp_timer_accuracy.c
JereLeppanen: In v2, I removed these conditions and instead we print the diff. There are other changes as well, involving odp_timeout_tick(). Please re-review.

psavol (approved):

psavol:
API commit message could mention the wording change from "within" to "less than". Please add my review-by at the same time editing it.

JereLeppanen:
v3:
- Rebase.
- Add review tags.

JereLeppanen:
v4:
- Mention "less than" vs. "within" in the commit message.

shijith:
Reviewed-by: Shijith Thotton <sthotton@marvell.com>

PR #1821, opened 2023-05-15
api: fix doxygen see command usage

MatiasElo:
Using @see command starts a new "See also" paragraph in Doxygen output, which is not intended in several API specifications.


include/odp/api/spec/dma.h
```diff
@@ -332,7 +332,7 @@ void odp_dma_pool_param_init(odp_dma_pool_param_t *pool_param);
  * The use of pool name is optional. Unique names are not required. However, odp_pool_lookup()
  * returns only a single matching pool. Use odp_dma_pool_param_init() to initialize pool parameters
  * into their default values. Parameters values must not exceed pool capabilities
- * (@see odp_dma_pool_capability_t)
+ * (odp_dma_pool_capability_t).
```
TuomasTaipale, line 335: Common pattern seems to be "@see" -> "see", here the word is dropped. What's the exception in the DMA files?

include/odp/api/spec/packet.h
```diff
@@ -937,7 +937,7 @@ void *odp_packet_buf_head(odp_packet_buf_t pkt_buf);
 /**
```
TuomasTaipale, line 937: In this file, there seems to be a couple cases of single colon vs. double colon, maybe those could be fixed as well, `odp_packet_vector_tbl()` spec doc, second `@note` section (though, doxygen seems to have troubles creating links to unnamed struct/union members without extra tricks).

include/odp/api/spec/dma.h
MatiasElo: The thinking was that when the same type is already mentioned in the text right before the braces, using `see` seems unnecessary.

include/odp/api/spec/packet.h
MatiasElo: Sure, fixed in V2.

TuomasTaipale (approved):

PR #1822, opened 2023-05-17
api: dma: clarify session name uniqueness

TuomasTaipale:
This patchset clarifies that DMA session names need not be unique. This was already the intent but the specification text was not stating this explicitly. New validation tests are added to verify this. Additionally, fix a linux-gen issue related to session lookup.


test/validation/api/dma/dma.c
```diff
@@ -269,6 +269,44 @@ static void test_dma_debug(void)
...
+
+	odp_dma_param_init(&dma_param);
+	dma_param.compl_mode_mask = ODP_DMA_COMPL_SYNC;
+	dma_a = odp_dma_create(name, &dma_param);
+
+	CU_ASSERT_FATAL(dma_a != ODP_DMA_INVALID);
+
+	dma = odp_dma_lookup(name);
+
+	CU_ASSERT(odp_dma_to_u64(dma) == odp_dma_to_u64(dma_a));
```
MatiasElo, line 285: When no name is given this test is a bit questionable. One could assume that `ODP_DMA_INVALID` is also a valid return value from `odp_dma_lookup()`.

include/odp/api/spec/dma.h
```diff
@@ -49,6 +49,10 @@ void odp_dma_param_init(odp_dma_param_t *param);
 /**
  * Create DMA session
  *
+ * Create a DMA session. The use of session name is optional. Unique names are not required.
```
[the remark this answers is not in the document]
TuomasTaipale: Sure, v2.

test/validation/api/dma/dma.c
TuomasTaipale: Lookup with `NULL` seems not to be a thing so dropped that support/validation. Added a "named" word in the spec text which hopefully gives a clue what you can and cannot find with the lookup function. V3.

MatiasElo (approved):

VamsiKrishnaA99:
Acked-by: Vamsi Attunuru vattunuru@marvell.com

PR #1824, opened 2023-05-22
api: timer: clarify periodic timer capa

psavol:


include/odp/api/spec/timer.h
```diff
@@ -75,11 +75,14 @@ int odp_timer_res_capability(odp_timer_clk_src_t clk_src,
  * frequency multiplier and the minimum timeout resolution. If there is no requirement for timeout
  * resolution, it is set to zero.
  *
- * A successful call overwrites all 'capa' fields with the resulting values. Return value
- * is 1 when timer capability meets or exceeds all requested values. The call returns 0, when
- * the requested base frequency is not supported exactly, but capability meets or exceeds all other
- * requested values. In this case, the call overwrites 'base_freq_hz' with the closest supported
- * frequency.
+ * A successful call overwrites 'capa' fields in the following way. The call returns 1, when
+ * timer capability supports the requested base frequency exactly, and meets or exceeds other
```
MatiasElo, line 79: Maybe `... the following ways. The call returns 1, when timer supports ...`

include/odp/api/spec/timer.h
```diff
@@ -75,11 +75,14 @@ int odp_timer_res_capability(odp_timer_clk_src_t clk_src,
...
+ * requested values. The base frequency value is not modified, but other 'capa' fields are updated
+ * with their resulting maximum capabilities.
+ *
+ * The call returns 0, when the requested base frequency is not supported exactly, but capability
```
MatiasElo, line 83: Perhaps `capability ` -> `timer`

include/odp/api/spec/timer.h
psavol: Fixed in v2

include/odp/api/spec/timer.h
psavol: Fixed in v2

include/odp/api/spec/timer_types.h
```diff
@@ -102,7 +102,15 @@ typedef struct {
...
 typedef struct {
-	/** Periodic timer pool base frequency in hertz */
+	/**
+	 * Periodic timer pool base frequency in hertz
+	 *
+	 * Base frequency is represented as a fractional number where the fraction part is always
+	 * less than one. In other words, the integer part specifies whole hertz whereas
+	 * the fraction part specifies parts of a hertz (if any). The fraction part does not
+	 * need to be reduced to its lowest terms - e.g. 100.5 Hz may be represented as 100 1/2 Hz,
+	 * 100 5/10 Hz, or 100 Hz with some other equivalent fraction part.
```
JereLeppanen, line 112: Could this just be a part of odp_fract_u64_t specification?

psavol: I think it's better to keep data type documentation generic. For example, in some other use case we might require reduction to the lowest terms, although here we do not require that.

MatiasElo (approved):

shijith:
Reviewed-by: Shijith Thotton <sthotton@marvell.com>

PR #1827, opened 2023-05-25
[PATCH v3] validation: crypto: clean up code a bit

JannePeltonen:
validation: crypto: add crypto algorithm IDs in test vectors
validation: crypto: simplify crypto_op() function
validation: crypto: move code to separate files


JannePeltonen:
v2: rebased

JereLeppanen (approved):

test/validation/api/crypto/test_vectors.h
```diff
@@ -8,10 +8,13 @@
 #ifndef _ODP_TEST_CRYPTO_VECTORS_H_
```
TuomasTaipale: Copyright could be updated. There are also some dead entries in AES XTS section that have been commented out for a few years, maybe those could be just removed.

JannePeltonen: I do not think the copyright year should be updated when there are no changes other than just the moving of code around.

The commented-out XTS vector could indeed be removed or uncommented but that is better done in a separately. So I am not including it in this PR.

TuomasTaipale: You added two new fields, not just moving code around.

JannePeltonen: ok, I'll update the copyright.

TuomasTaipale (approved):

PR #1830, opened 2023-05-26
[PATCH v3] validation: time: improve global time sync test

JereLeppanen:
Test global time sync also with a control thread. Print the time differences.

### DIFF
--- a/test/validation/api/time/time.c
+++ b/test/validation/api/time/time.c
@@ -731,8 +731,9 @@ static void time_test_global_sync(void)
 		odp_cpumask_set(&cpumask_one, cpu);
 		thr_common.cpumask = &cpumask_one;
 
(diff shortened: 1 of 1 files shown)


test/validation/api/time/time.c
```diff
@@ -765,19 +765,41 @@ static void time_test_global_sync(const int ctrl)
...
+			if (t > max) {
 				max = t;
+				max_idx = i;
+			}
+
+			avg += t;
 		}
 
-		CU_ASSERT(max - min < tolerance);
+		avg /= num - 1;
```
psavol, line 796: Avg should be divided by "num". The for loop runs "num" times.

JereLeppanen: It's num - 1 because min result itself is not included in the average. Comment added in v3.

psavol (approved):

PR #1835, opened 2023-06-02
[PATCH v7] api: init: add abnormal termination function

TuomasTaipale:
New `odp_term_abnormal()` function can be used to abnormally terminate an ODP application. Application can utilize the new function to terminate ODP with additional debug data in error situations where regular local and global clean-up is not possible.

v2:
- Jerin's comments

v3:
- Petri's comments

v4:
- Added patches for ODP abort function usage clarification, `odp_term_abnormal()` linux-gen implementation and validation
- Changed `odp_term_abnormal()` `data` parameter type to `void *`
- Added a `ODP_TERM_LAST_FLAG` constant

v6:
- Added an explicit `uint64_t` cast to `ODP_TERM_FROM_SIGH`

v7:
- Rebased
- Added reviewed-by tags

### DIFF
--- a/include/odp/api/spec/init.h
+++ b/include/odp/api/spec/init.h
@@ -91,11 +91,12 @@ int odp_override_log(odp_log_level_t level, const char *fmt, ...);
  * - By overriding the ODP implementation default abort function
  *   odp_override_abort().
  *
- * @warning The latter option is less portable and GNU linker dependent
- * (utilizes function attribute "weak"). If both are defined, the odp_init_t
- * function pointer has priority over the override function.
+ * The latter option is less portable and GNU linker dependent (utilizes function
+ * attribute "weak"). If both are defined, the odp_init_t function pointer has
+ * priority over the override function.
  *
- * @warning this function shall not return
+ * Note that no ODP calls should be called in the abort function and the function
+ * should not return.
  */
 void odp_override_abort(void) ODP_NORETURN;
 
--- a/include/odp/api/spec/init.h
+++ b/include/odp/api/spec/init.h
@@ -1,5 +1,5 @@
 /* Copyright (c) 2013-2018, Linaro Limited
- * Copyright (c) 2019-2021, Nokia
+ * Copyright (c) 2019-2023, Nokia
  * All rights reserved.
  *
  * SPDX-License-Identifier:     BSD-3-Clause
@@ -31,6 +31,18 @@ extern "C" {
  * @typedef odp_instance_t
  * ODP instance ID.
  */
+
+/**
+ * Called from signal handler
+ */
+#define ODP_TERM_FROM_SIGH	((uint64_t)0x1)
+
+/**
+ * Last standard flag for abnormal terminate
+ *
+ * An implementation may use this for adding its own flags after the standard flags.
+ */
+#define ODP_TERM_LAST_FLAG ODP_TERM_FROM_SIGH
 
 /**
  * ODP log level.
@@ -341,6 +353,39 @@ int odp_term_local(void);
  * @see odp_init_global(), odp_term_local()
  */
 int odp_term_global(odp_instance_t instance);
+
+/**
+ * Abnormal ODP termination after a non-recoverable error
+ *
+ * Application may call this function to terminate an ODP instance after facing
+ * a non-recoverable error. Depending on the implementation, this function may
+ * attempt to dump stack and other memory areas, clean up and stop HW
+ * operations and/or perform other actions helpful in postmortem analysis.
+ * Depending on the nature of the error resulting in the abnormal termination,
+ * these actions may partially or completely fail. Flags (ODP_TERM_*) parameter
+ * can be used to control and data parameter can be used to pass additional
+ * flag-specific information to the termination process. Implementation
+ * specific flags with implementation specific data may also exist, see from
+ * implementation documentation how those should be utilized.
+ *
+ * Some coordination across threads is required when abnormally terminating, if
+ * other threads continue calling ODP functions during or after termination,
+ * their operation is most likely affected.
+ *
+ * When the function returns, the ODP instance has been destroyed either
+ * completely or partially. Application must not attempt to call any ODP
+ * functions during its remaining lifetime, but terminate as soon as feasible.
+ *
+ * @param instance        Instance handle
+ * @param flags           A bit mask of control flags (ODP_TERM_*), set to 0
+ *                        when no flags
+ * @param data            Additional data, set to NULL when no additional data
+ *
+ * @retval 0 on all actions successfully performed
+ * @retval <0 on failure to perform all actions, implementation specific status
+ *         code for debugging
+ */
+int odp_term_abnormal(odp_instance_t instance, uint64_t flags, void *data);
 
 /**
  * Set thread specific log function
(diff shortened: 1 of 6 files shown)


include/odp/api/spec/init.h
```diff
@@ -341,6 +346,39 @@ int odp_term_local(void);
  */
 int odp_term_global(odp_instance_t instance);
 
+/**
+ * Abnormal ODP termination after a non-recoverable error
+ *
+ * Application may use this function to terminate an ODP instance after facing
```
jerinjacobk, line 352: use -> call

include/odp/api/spec/init.h
```diff
@@ -32,6 +32,11 @@ extern "C" {
  * ODP instance ID.
  */
 
+/**
+ * Called from signal handler
+ */
+#define ODP_TERM_FROM_SIGH	0x1u
+
```
jerinjacobk, line 39: Two more potential flags are
a) dump stack
b)  capture trace.

I think, we can make it as HINT, so that we dont need capablity to use this or and a speicifc implemenation can ignore this if this not applicable. Some thing like,

ODP_TERM_DUMP_STACK_HINT
ODP_TERM_CAPTURE_TRACE_HINT

include/odp/api/spec/init.h
```diff
@@ -341,6 +346,39 @@ int odp_term_local(void);
...
+ *
+ * When the function returns, the ODP instance has been destroyed either
+ * completely or partially. Application must not attempt to call any ODP
+ * functions during its remaining lifetime, but terminate as soon as feasible.
+ *
+ * @param instance        Instance handle
+ * @param flags           A bit mask of control flags (ODP_TERM_*), set to 0
+ *                        when no flags
+ * @param data            Additional data, set to 0 when no additional data
+ * @param fmt             Printf-style message to be printed to the ODP log
```
jerinjacobk, line 373: Why to pass this? Application can print them self. Right?

include/odp/api/spec/init.h
```diff
@@ -341,6 +346,39 @@ int odp_term_local(void);
...
+ * their operation is most likely affected.
+ *
+ * When the function returns, the ODP instance has been destroyed either
+ * completely or partially. Application must not attempt to call any ODP
+ * functions during its remaining lifetime, but terminate as soon as feasible.
+ *
+ * @param instance        Instance handle
+ * @param flags           A bit mask of control flags (ODP_TERM_*), set to 0
+ *                        when no flags
+ * @param data            Additional data, set to 0 when no additional data
```
jerinjacobk, line 372: In order to scale more, may be config structure can be passed(sub structure per FLAG). Just in case, uint64_t data is enough in the future. No strong opinion, though.

include/odp/api/spec/init.h
TuomasTaipale: Changed in v2.

include/odp/api/spec/init.h
TuomasTaipale: Alternatively, these flags could be implementation specific, if deemed not generic enough. In any case, yes, best keep them hint-like.

include/odp/api/spec/init.h
TuomasTaipale: Carried over from the original panic-function, but true, it might be a bit pointless in this new function, ditched in v2.

include/odp/api/spec/init.h
TuomasTaipale: If some of the flags are implementation-specific, defining a structure to pass here might be challenging. `uint64_t` does accommodate a pointer so in case something more is needed, it could be flag-specific pointer to some struct.

jerinjacobk: OK

include/odp/api/spec/init.h
jerinjacobk: OK. In such case to avoid conflict with platform specific flasgs, introduce something like ODP_TERM_PLATFORM_TYPE_1, ODP_TERM_PLATFORM_TYPE_2, ODP_TERM_PLATFORM_TYPE_3, ODP_TERM_PLATFORM_TYPE_4.

include/odp/api/spec/init.h
```diff
@@ -341,6 +346,37 @@ int odp_term_local(void);
...
+ * Abnormal ODP termination after a non-recoverable error
+ *
+ * Application may call this function to terminate an ODP instance after facing
+ * a non-recoverable error. Depending on the implementation, this function may
+ * attempt to dump stack and other memory areas, clean up and stop HW
+ * operations and/or perform other actions helpful in postmortem analysis.
+ * Depending on the nature of the error resulting in the abnormal termination,
+ * these actions may partially or completely fail. Flags (ODP_TERM_*) and data
+ * parameters can be used to control and pass additional information to the
+ * termination process.
```
psavol, line 359: Spec could say more explicitly that data value depends on selected flags.

include/odp/api/spec/init.h
psavol: If API support is needed for additional flags, we could have "last flag" define in the end. However, it's trivial for implementation to select a base for all its own flags (e.g. start from bit 32), and update the base only when a new API version would overlap.

jerinjacobk: Yes, Let add LAST_FLAG at 31 bit.

TuomasTaipale: `ODP_TERM_LAST_FLAG` in v4.

platform/linux-generic/odp_init.c
```diff
@@ -692,6 +692,26 @@ int odp_term_local(void)
 	return term_local(ALL_INIT);
 }
 
+int odp_term_abnormal(odp_instance_t instance, uint64_t flags, void *data ODP_UNUSED)
```
psavol, line 695: Patch is otherwise OK, but git commit log entry should start with "linux-gen:" prefix.

include/odp/api/spec/init.h
```diff
@@ -32,6 +32,18 @@ extern "C" {
...
+ * Called from signal handler
+ */
+#define ODP_TERM_FROM_SIGH	0x1u
+
+/**
+ * Last standard flag for abnormal terminate
+ *
+ * An implementation may use this for adding its own flags after the standard flags.
+ */
+#define ODP_TERM_LAST_FLAG ODP_TERM_FROM_SIGH
```
jerinjacobk, line 45: Please change to #define ODP_TERM_LAST_FLAG (1 << 32)ull else the logic will fail when new  generic ODP_TERM_... FLAG

include/odp/api/spec/init.h
```diff
@@ -32,6 +32,18 @@ extern "C" {
  * ODP instance ID.
  */
 
+/**
+ * Called from signal handler
+ */
+#define ODP_TERM_FROM_SIGH	0x1u
```
jerinjacobk, line 38: ull as it is uin64_t

include/odp/api/spec/init.h
TuomasTaipale: Idea is to keep updating the `ODP_TERM_LAST_FLAG` when new API-level flags are added.

psavol (approved):

PR #1836, opened 2023-06-06
[PATCH v10] api: crypto: copy the ranges of null algorithms in OOP op type & change bit mode semantics

JannePeltonen:
Allow non-zero-length cipher and auth ranges for the null cipher and auth algoriths in the OOP operation type. Copy the ranges to the output packet as with non-null algorithms. Allow implementations to not fully support non-zero-length ranges by explicitly mentioning that a crypto operation with such a range may fail if not supported.

This changes makes null algorithms behave as non-null algorithms with respect to packet data copying in the OOP operation type. When both the cipher and the auth algorithms are null, a crypto operation reduces to a´packet data copying operation (if supported by the underlying ODP implementation). Such copying may be useful for applications that wish to process crypto and non-crypto packets in the same way using the OOP crypto operation type.

### DIFF
--- a/include/odp/api/spec/crypto.h
+++ b/include/odp/api/spec/crypto.h
@@ -255,9 +255,7 @@ int odp_crypto_result(odp_crypto_packet_result_t *result,
  * input packet to the output packet and then the crypto operation
  * was applied to the output packet.
  *
- * Crypto range and auth range of null cipher and auth algorithms are
- * ignored, i.e. not copied in the output packet. Auth range of (AEAD)
- * algorithms that ignore auth range is not copied.
+ * Auth range of (AEAD) algorithms that ignore auth range is not copied.
  *
  * The offset of the crypto range and auth range in the output packet is
  * the same as in the input packet, adjusted by dst_offset_shift operation
--- a/include/odp/api/spec/crypto_types.h
+++ b/include/odp/api/spec/crypto_types.h
@@ -560,6 +560,30 @@ typedef struct odp_crypto_session_param_t {
 	 */
 	odp_crypto_op_type_t op_type;
 
+	/** Cipher range unit
+	 *
+	 *  When this flag is true, cipher range offset and length are in bits.
+	 *  Otherwise the offset and length are in bytes.
+	 *
+	 *  If cipher capabilities do not include bit_mode, setting this to
+	 *  true causes a session creation failure.
+	 *
+	 *  The default value is false.
+	 */
+	odp_bool_t cipher_range_in_bits;
+
+	/** Auth range unit
+	 *
+	 *  When this flag is true, auth range offset and length are in bits.
+	 *  Otherwise the offset and length are in bytes.
+	 *
+	 *  If auth capabilities do not include bit_mode, setting this to
+	 *  true causes a session creation failure.
+	 *
+	 *  The default value is false.
+	 */
+	odp_bool_t auth_range_in_bits;
+
 	/** Authenticate cipher vs. plain text
 	 *
 	 *  Controls ordering of authentication and cipher operations,
@@ -725,6 +749,9 @@ typedef struct odp_crypto_packet_op_param_t {
 
 	/** Data range to be ciphered.
 	 *
+	 *  The range is given in bits or bytes as configured at session
+	 *  creation.
+	 *
 	 *  Ignored by the null cipher with operation types other than
 	 *  ODP_CRYPTO_OP_TYPE_OOP. Must be set to zero range (zero offset
 	 *  and zero length) with the null cipher used with the out-of-place
@@ -733,6 +760,9 @@ typedef struct odp_crypto_packet_op_param_t {
 	odp_packet_data_range_t cipher_range;
 
 	/** Data range to be authenticated
+	 *
+	 *  The range is given in bits or bytes as configured at session
+	 *  creation.
 	 *
 	 *  The value is ignored with authenticated encryption algorithms,
 	 *  such as AES-GCM, which authenticate data in the cipher range
@@ -937,22 +967,23 @@ typedef struct odp_crypto_cipher_capability_t {
 	/** IV length in bytes */
 	uint32_t iv_len;
 
-	/** Cipher is operating in bitwise mode
+	/** Cipher supports bit mode
 	 *
-	 * This cipher works on series of bits, rather than sequences of bytes:
-	 * cipher_range in odp_crypto_op_param_t and
-	 * odp_crypto_packet_op_param_t will use bits, rather than bytes.
+	 * This cipher can work on a range of bits in addition to a range of
+	 * bytes. When this capability is not present, only byte ranges are
+	 * supported. The unit of cipher range is selected at session creation
+	 * through the cipher_range_in_bits session parameter.
 	 *
-	 * Note: data buffer MUST start on the byte boundary, using offset
-	 * which is not divisible by 8 is unsupported and will result in
-	 * unspecified behaviour.
+	 * Note: In bit mode the cipher range must start on a byte boundary.
+	 * Using an offset which is not divisible by 8 will result in
+	 * undefined behaviour.
 	 *
-	 * Note2: If the data length is not a multiple of 8, the remaining
-	 * bits of the data in the last byte of the input/output will be the
-	 * most significant bits, i.e. the most significant bit is considered
-	 * to be the first bit of a byte for the purpose of input and output
-	 * data range. The output bits that fall out of the output range are
-	 * undefined.
+	 * Note2: If the range length in bit mode is not a multiple of 8,
+	 * the remaining bits of the data in the last byte of the input/output
+	 * will be the most significant bits, i.e. the most significant bit is
+	 * considered to be the first bit of a byte for the purpose of input
+	 * and output data range. The output bits that fall out of the output
+	 * range are undefined.
 	 */
 	odp_bool_t bit_mode;
 
@@ -984,22 +1015,23 @@ typedef struct odp_crypto_auth_capability_t {
 		uint32_t inc;
 	} aad_len;
 
-	/** Auth is operating in bitstring mode
+	/** Auth algorithm supports bit mode
 	 *
-	 * This auth works on series of bits, rather than sequences of bytes:
-	 * auth_range in odp_crypto_op_param_t and
-	 * odp_crypto_packet_op_param_t will use bits, rather than bytes.
+	 * This auth algorithm can work on a range of bits in addition to
+	 * a range of bytes. When this capability is not present, only byte
+	 * ranges are supported. The unit of auth range is selected at session
+	 * creation through the auth_range_in_bits session parameter.
 	 *
-	 * Note: data buffer MUST start on the byte boundary, using offset
-	 * which is not divisible by 8 is unsupported and will result in
-	 * unpredictable behaviour.
+	 * Note: In bit mode the auth range must start on a byte boundary.
+	 * Using an offset which is not divisible by 8 will result in
+	 * undefined behaviour.
 	 *
-	 * Note2: If the data length is not a multiple of 8, the remaining
-	 * bits of the data in the last byte of the input/output will be the
-	 * most significant bits, i.e. the most significant bit is considered
-	 * to be the first bit of a byte for the purpose of input and output
-	 * data range. The output bits that fall out of the output range are
-	 * undefined.
+	 * Note2: If the range length in bit mode is not a multiple of 8,
+	 * the remaining bits of the data in the last byte of the input/output
+	 * will be the most significant bits, i.e. the most significant bit is
+	 * considered to be the first bit of a byte for the purpose of input
+	 * and output data range. The output bits that fall out of the output
+	 * range are undefined.
 	 */
 	odp_bool_t bit_mode;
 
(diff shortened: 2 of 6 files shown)


JannePeltonen:
v2: Added a new API patch that makes bit/byte mode of an algorithm configurable at session creation and changes bit_mode capability to tell if bit mode is supported (as opposed to telling if the algorithm must be used in bit mode in the particular ODP implementation).

This change simplifies the API and avoids the mess that would result from several current ODP implementations advertising both bit mode and byte mode for null algorithms (so an implementation does not know if a range given is in bits or bytes) and then the previously sent out-of-place processing change actually needing well-defined ranges for null algorithms too.

This new API patch, as now proposed, is not backward compatible for applications that may use bit mode algorithms.

anoobj:
Spec change looks good.

include/odp/api/spec/crypto_types.h
```diff
@@ -937,22 +967,23 @@ typedef struct odp_crypto_cipher_capability_t {
...
+	 * This cipher can work on a range of bits in addition to a range of
+	 * bytes. When this capability is not present, only byte ranges are
+	 * supported. The unit of cipher range is selected at session creation
+	 * through the cipher_range_in_bits session parameter.
 	 *
-	 * Note: data buffer MUST start on the byte boundary, using offset
-	 * which is not divisible by 8 is unsupported and will result in
-	 * unspecified behaviour.
+	 * Note: In bit mode the cipher range must start on a byte boundary.
+	 * using an offset which is not divisible by 8 will result in
```
psavol, line 982: using -> Using (upper case)

include/odp/api/spec/crypto_types.h
```diff
@@ -984,22 +1015,23 @@ typedef struct odp_crypto_auth_capability_t {
...
+	 * This auth algorithm can work on a range of bits in addition to
+	 * a range of bytes. When this capability is not present, only byte
+	 * ranges are supported. The unit of auth range is selected at session
+	 * creation through the auth_range_in_bits session parameter.
 	 *
-	 * Note: data buffer MUST start on the byte boundary, using offset
-	 * which is not divisible by 8 is unsupported and will result in
-	 * unpredictable behaviour.
+	 * Note: In bit mode the auth range must start on a byte boundary.
+	 * using an offset which is not divisible by 8 will result in
```
psavol, line 1030: using -> Using (upper case)

include/odp/api/spec/crypto_types.h
```diff
@@ -753,9 +753,11 @@ typedef struct odp_crypto_packet_op_param_t {
...
 	 *
 	 *  Ignored by the null cipher with operation types other than
-	 *  ODP_CRYPTO_OP_TYPE_OOP. Must be set to zero range (zero offset
-	 *  and zero length) with the null cipher used with the out-of-place
-	 *  operation type.
+	 *  ODP_CRYPTO_OP_TYPE_OOP. In the OOP operation type the cipher
+	 *  range is copied in the output packet even with the null cipher.
+	 *  If an ODP implementation does not support non-zero-length
+	 *  ranges with the null cipher and the OOP operation type, the
+	 *  crypto operation may fail.
```
psavol, line 760: Would it be more precise to say "... operation may report failure.". I.e. to be explicit that operation return code reports a failure rather than there would be undefined behavior.

include/odp/api/spec/crypto_types.h
JannePeltonen: Fixed in v4.

include/odp/api/spec/crypto_types.h
JannePeltonen: Fixed in v4.

include/odp/api/spec/crypto_types.h
JannePeltonen: I did not change this yet since we probably need to clarify it more than that. Specifically, if we allow failure, do we allow a failure of odp_crypto_op() / odp_crypto_op_enq() so that the input packet is not even consumed, or a failure reported through the crypto op result, or both. Let's discuss it today.

JannePeltonen:
v4:
	- Added couple of bug fixes and minor code improvements
	- Added linux-gen implementantion
	- Added validation tests (without allowing a failure of OOP operations with null algorithms and non-zero ranges)

include/odp/api/spec/crypto_types.h
JannePeltonen: Reworded in v5.

JannePeltonen:
v5:
- Reworded the text about allowing OOP crypto op failures with null algorithms and non-zero ranges.
- Added a patch to allow the above failures in validation test without causing test failures

psavol:
Reviewed and approved API patches in v5.

test/validation/api/crypto/odp_crypto_test_inp.c
```diff
@@ -254,15 +254,16 @@ static void alg_test_ses(odp_crypto_op_t op,
 {
```
JereLeppanen, line 251: Typos in commit comment: "resuls", "aurh", "tange".

test/validation/api/crypto/crypto_op_test.c
```diff
@@ -179,8 +179,8 @@ static void prepare_crypto_ranges(const crypto_op_test_param_t *param,
 				  odp_packet_data_range_t *auth_range)
```
JereLeppanen, line 179: Typo in commit comment: "parametets".

JannePeltonen:
Separated the three first commits to PR 1862 so they are no longer in the scope of this PR except as dependencies until PR 1862 gets merged and this PR rebased.

JannePeltonen:
v6: Rebased to get rid of the dependent commits. Now all the commits are part of this PR review.
v7: Added Petri's reviewed-by tags and fixed a typo in a commit message

platform/linux-generic/odp_crypto_openssl.c
```diff
@@ -1119,70 +1120,72 @@ static int process_cipher_param(odp_crypto_generic_session_t *session,
...
-
 	return ret <= 0 ? ODP_CRYPTO_ALG_ERR_DATA_SIZE :
 			  ODP_CRYPTO_ALG_ERR_NONE;
 }
 
+static odp_crypto_alg_err_t cipher_encrypt_bits(odp_packet_t pkt,
+						const odp_crypto_packet_op_param_t *param,
+						odp_crypto_generic_session_t *session)
+{
+	odp_crypto_packet_op_param_t new_param = *param;
```
anoobj, line 1170: Can we avoid this copy by passing offset & length as separate args to 'cipher_encrypt_bytes'?

JannePeltonen: Since the prototype of the generic cipher function cannot be easily changed, that would require moving the body of cipher_encrypt_bytes() into a separate function and leaving cipher_encrypt() bytes to be a simpler wrapper. And the same for the decrypt function.

I rather wanted to keep the change simpler. Copying the parameter struct should not be that costly anyway. A better optimization would be to get rid of the packet data copy that occurs in cipher_encrypt_bytes by using internal_encrypt() that can handle also segmented packets without copying. But I did not want to do such a change yet in this PR which is about the bit mode semantics change.

Also, supporting OOP without the separate memory copying step would probably be a good optimization but that would require changing the prototype of the generic cipher & auth functions (and would also otherwise be a bit bigger change). Because of this, I did not want to start changing the generic cipher & auth function prototypes now.

anoobj: Understood. Thanks for the explanation.

include/odp/api/spec/crypto_types.h
```diff
@@ -560,6 +560,30 @@ typedef struct odp_crypto_session_param_t {
 	 */
 	odp_crypto_op_type_t op_type;
 
+	/** Units of cipher range
```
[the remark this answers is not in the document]
JannePeltonen: Changed to "Cipher range unit" in v8.

include/odp/api/spec/crypto_types.h
```diff
@@ -560,6 +560,30 @@ typedef struct odp_crypto_session_param_t {
...
+	 *
+	 *  When this flag is true, cipher range offset and length are in bits.
+	 *  Otherwise the offset and length are in bytes.
+	 *
+	 *  If cipher capabilities do not include bit_mode, setting this flag
+	 *  causes session creation failure.
```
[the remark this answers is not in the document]
JannePeltonen: Reworded slightly in v8.

include/odp/api/spec/crypto.h
```diff
@@ -255,9 +255,7 @@ int odp_crypto_result(odp_crypto_packet_result_t *result,
  * input packet to the output packet and then the crypto operation
```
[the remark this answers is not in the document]
JannePeltonen: fixed in v8

include/odp/api/spec/crypto_types.h
```diff
@@ -753,9 +753,15 @@ typedef struct odp_crypto_packet_op_param_t {
 	 *  creation.
 	 *
 	 *  Ignored by the null cipher with operation types other than
-	 *  ODP_CRYPTO_OP_TYPE_OOP. Must be set to zero range (zero offset
-	 *  and zero length) with the null cipher used with the out-of-place
-	 *  operation type.
+	 *  ODP_CRYPTO_OP_TYPE_OOP.
+	 *
+	 *  With the OOP operation type the cipher range is copied in the
```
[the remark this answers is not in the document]
JannePeltonen: fixed in v8

JannePeltonen:
v8: fixed a few typos and grammar issues.

JereLeppanen (approved):

test/validation/api/crypto/odp_crypto_test_inp.c
```diff
@@ -171,8 +171,10 @@ static odp_crypto_session_t session_create(odp_crypto_op_t op,
 
 	/* Create a crypto session */
 	odp_crypto_session_param_init(&ses_params);
-	ses_params.op = op;
-	ses_params.op_type = op_type;
+	ses_params.op = session->op;
+	ses_params.op_type = session->op_type;
+	ses_params.cipher_range_in_bits = session->cipher_range_in_bits;
+	ses_params.auth_range_in_bits = session->auth_range_in_bits;
 	ses_params.auth_cipher_text = (order == AUTH_CIPHERTEXT);
```
anoobj, line 178: Does it make sense to move order also to crypto_session_t?

JannePeltonen: Perhaps, and the same for hash_mode, but since the change would not really be related to any of the commits in this PR it does not need to be in this PR.

test/validation/api/crypto/odp_crypto_test_inp.c
```diff
@@ -342,8 +340,47 @@ static void alg_test(odp_crypto_op_t op,
...
+		     odp_packet_data_range_t cipher_bit_range,
+		     odp_packet_data_range_t auth_bit_range,
+		     uint32_t digest_offset,
+		     odp_bool_t is_bit_mode_cipher,
+		     odp_bool_t is_bit_mode_auth)
+{
+	odp_packet_data_range_t cipher_range;
+	odp_packet_data_range_t auth_range;
+
+	for (int cr_in_bits = 0; cr_in_bits <= 1; cr_in_bits++) {
```
anoobj, line 360: Is it right to use 'for' loop here? Isn't it better to introduce a wrapper function like,

alg_test_flags(..., bool is_cr_in_bit, bool is_ar_in_bit)

And then have calls from here like,

alg_test_flags(...., false, false);
alg_test_flags(...., false, true);
alg_test_flags(...., true, false);
alg_test_flags(...., true, true);

Other option is to use goto,

	bool ar_in_bytes, cr_in_bytes = true;

reset_auth_test_again:
	ar_in_bytes = true;

test_again:
	cipher_range = cipher_bit_range;
	auth_range = auth_bit_range;

	if (cr_in_bytes) {
		cipher_range.offset /= 8;
		cipher_range.length /= 8;
	}

	if (ar_in_bytes) {
		auth_range.offset /= 8;
		auth_range.length /= 8;
	}

	alg_test_op_types(op, order, ref, cipher_range, auth_range,
			  digest_offset, cr_in_bits, ar_in_bits);

	if (is_bit_mode_auth && ar_in_bytes) {
		ar_in_bytes = false;
		goto test_again;
	}

	if (is_bit_mode_cipher && cr_in_bytes) {
		cr_in_bytes = false;
		goto reset_auth_test_again;
	}

If we are proceeding with 'for' loop, please check if it is all right to declare variable in for loop header.

test/validation/api/crypto/odp_crypto_test_inp.c
anoobj: Agreed.

anoobj:
Minor comment in one of the test patch. Other than that,

test/validation/api/crypto/odp_crypto_test_inp.c
JannePeltonen: I do not think there is anything wrong using for statements here. We are iterating over the possible values of the two-valued conditions. I think the example using gotos for the looping is much worse. The wrapper function option would probably be ok, but it essentially just unrolls the for loops, which maybe is not so useful.

Declaring loop variables within the for statement like that has been supported since C99.

I am not planning to change this piece of code unless you insist.

JannePeltonen:
v9: added reviewed-by tags

test/validation/api/crypto/odp_crypto_test_inp.c
anoobj: My only concern is the code legibility. You can proceed with the current code if you think that is the best way.

PR #1838, opened 2023-06-09
[PATCH v5] quell GCC 13.1 bound check warnings

JereLeppanen:
Quell a few -Warray-bounds and -Wstringop-overflow warnings with GCC 13.1, with and without -Wpedantic.

The question here is do we keep adding these pragmas to the code or do we just disable -Warray-bounds and -Wstringop-overflow?


JereLeppanen:
v2:
- Don't try to fix flexible array member in ethtool_stats.c, just quell the error.
- Quell error in odp_api_from_cpp LTO link.

JereLeppanen:
v3:
- Rebase.

JereLeppanen:
v4:
- Fix pedantic error instead of just suppressing it.

JannePeltonen (approved):

JereLeppanen:
v5:
- Rebase.
- Add review tags.

PR #1846, opened 2023-06-16
[PATCH v3] linux-gen: time: refactor implementation

MatiasElo:
Refactor time codebase to enable more optimized architecture specific implementations. After this commit x86 and aarch64 architectures are assumed to always have support for HW time. The fallback to POSIX time functions has been removed from these architectures for improved performance.


platform/linux-generic/arch/common/odp_time_cpu.c
```diff
@@ -0,0 +1,48 @@
...
+#include <odp/visibility_end.h>
+
+int _odp_time_init_global(void)
+{
+	_odp_time_global_t *global = &_odp_time_glob;
+
+	memset(global, 0, sizeof(_odp_time_global_t));
+
+	if (!_odp_time_cpu_global_freq_is_const())
+		_ODP_ERR("HW time frequency may change\n");
```
psavol, line 33: This warning could be CAPS to be more visible. Also, could comment the reasoning a bit: "Freq should be constant based on CPU ISA version, but could not confirm from CPU flag" or something similar.

platform/linux-generic/arch/common/odp_time_cpu.c
```diff
@@ -0,0 +1,48 @@
...
+#include <odp/visibility_end.h>
+
+int _odp_time_init_global(void)
+{
+	_odp_time_global_t *global = &_odp_time_glob;
+
+	memset(global, 0, sizeof(_odp_time_global_t));
+
+	if (!_odp_time_cpu_global_freq_is_const())
+		_ODP_ERR("WARN: HW time frequency may change\n");
```
psavol, line 33: Better to return error here and warn/make decision to continue on arch level.

MatiasElo: Fixed in V3.

platform/linux-generic/arch/common/odp_time_cpu.c
MatiasElo: Fixed in V2.

PR #1850, opened 2023-06-16
[PATCH v5] api: event: add odp_event_types_multi() function

MatiasElo:


TuomasTaipale (approved):

MatiasElo:
V3:
- Added validation test

include/odp/api/spec/event.h
```diff
@@ -79,6 +79,18 @@ odp_event_type_t odp_event_types(odp_event_t event,
...
+ * Event types of multiple events
+ *
+ * Outputs the event types of all given events into the output array. The types
+ * are written in the same order as input events.
+ *
+ * @param      event    Array of event handles
+ * @param[out] type     Event type array for output
+ * @param      num      Number of events and event types
+ */
+void odp_event_types_multi(const odp_event_t event[], odp_event_type_t type[], int num);
```
psavol, line 92: Another option for the functionality is more generic form which outputs also subtypes when requested:

void odp_event_types_multi(const odp_event_t event[], odp_event_type_t type[], odp_event_subtype_t subtype[], int num)

When subtype is NULL, implementation would ignore it. This adds additional if-else into implementation, but inlining/LTO could remove it when not needed. Also, naming would fit better the current odp_event_types().

MatiasElo: Changed in V3.

psavol (approved):

PR #1858, opened 2023-06-29
[PATCH v3] api: dma: pack DMA segment data structure

psavol:
Move structure fields so that addr/packet field is directly followed by length and offset fields. After this change all commonly used fields locate in the first 16 bytes of the structure, which may improve performance.

v3: rebased

### DIFF
--- a/include/odp/api/spec/dma_types.h
+++ b/include/odp/api/spec/dma_types.h
@@ -365,40 +365,38 @@ typedef struct odp_dma_param_t {
  * DMA segment
  */
 typedef struct odp_dma_seg_t {
-	/** Segment start */
+	/** Segment start address or packet handle */
 	union {
 		/** Segment start address in memory
 		 *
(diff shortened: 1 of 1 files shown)


PavanNikhilesh:

MatiasElo (approved):
Perhaps the commit message could still be modified a bit:
```
Move odp_dma_seg_t structure fields so that addr/packet field is directly
followed by length and offset fields. After this change all commonly used
fields are located in the first 16 bytes of the structure, which may
improve performance.
```

psavol:
> Perhaps the commit message could still be modified a bit:
> 
> ```
> Move odp_dma_seg_t structure fields so that addr/packet field is directly
> followed by length and offset fields. After this change all commonly used
> fields are located in the first 16 bytes of the structure, which may
> improve performance.
> ```

Fixed in v2

PR #1859, opened 2023-06-30
validation: time: improve time monotony test

psavol:
Add strict time stamp monotony tests. Check that strict time stamp calls maintain their order. Check that time keeping starts from zero. Check that converted nsec time and nsec time function calls use the same time base (values are close enough).


MatiasElo (approved):
The maximum allowed time value of one hour could perhaps be lowered to e.g. 10 minutes but otherwise looking OK.

PR #1860, opened 2023-07-04
perf: timer: benchmark performance for timer functions

JoyceKong-Arm:
Benchmark timer performance for the following timer functions: Set function: odp_timer_start()
Cancel function: odp_timer_cancel()
Scan function: odp_timer_scan_inline()

### DIFF
--- a/platform/linux-generic/odp_timer.c
+++ b/platform/linux-generic/odp_timer.c
@@ -235,6 +235,8 @@ typedef struct timer_local_t {
 	odp_time_t last_run;
 	int        run_cnt;
 	uint8_t    poll_shared;
+	uint64_t   nsec;
+	uint64_t   rounds;
 
 } timer_local_t;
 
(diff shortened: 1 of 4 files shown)


test/performance/odp_timer_perf.c
MatiasElo, line 1: Since at this point we don't want to add new APIs, this code should be integrated directly to the timer implementation (`platform/linux-generic/odp_timer.c`). For example `_odp_timer_term_local()` could be a place to trigger thread results print.

platform/linux-generic/odp_timer.c
```diff
@@ -871,8 +873,27 @@ void _odp_timer_run_inline(int dec)
 		timer_local.last_run = now;
 	}
 
-	/* Check the timer pools. */
-	timer_pool_scan_inline(num, now);
+	if (CONFIG_INLINE_PERF_BENCH) {
+		odp_time_t t1 = odp_time_local();
+
+		timer_pool_scan_inline(num, now);
+		odp_time_t t2 = odp_time_local();
```
MatiasElo, line 880: For more reliable results `odp_time_local_strict()` could be utilized here.

psavol:
There should not be merge commits ("Merge branch 'OpenDataPlane:master' into master") in a pull request. Please remove it.

It is better to create another branch for your work in progress, and do a pull request from there. Now, it seems that this pull request is into your master branch.

PR #1866, opened 2023-07-13
[PATCH v3] crypto: test that session creation fails when bit mode requested but not supported

JannePeltonen:
linux-gen: crypto: null: fix session allocation when out of sessions
linux-gen: crypto: fail bit mode session creation when not supported
validation: crypto: test that session creation fails when it must


JannePeltonen:
v2: Added the bit mode check at session creation in the ipsecmb crypto implementation too.

JereLeppanen (approved):

JannePeltonen:
v3: added reviewed-by tags

PR #1867, opened 2023-07-14
[PATCH v2] test: ipsecfwd: improve help text

TuomasTaipale:
Clearly state which options are mandatory and which are optional in tester help text.

v2:
- Janne's comments
- Added reviewed-by tag


test/performance/odp_ipsecfwd.c
```diff
@@ -374,18 +375,10 @@ static void print_usage(void)
 	       "            }\n"
 	       "        );\n"
 	       "\n"
-	       "Usage: %s [options]\n"
+	       "Mandatory [OPTIONS]:\n"
```
JannePeltonen, line 378: Maybe "Mandatory options" instead, i.e. without the brackets. And the same for the optional options.

test/performance/odp_ipsecfwd.c
```diff
@@ -326,9 +326,10 @@ static void print_usage(void)
 	}
 
 	printf("\n"
-	       "Simple IPsec performance tester. Forward and process plain and ipsec packets.\n"
+	       "Simple IPsec performance tester. Forward and process plain and IPsec packets.\n"
+	       "\n"
+	       "Usage: %s [OPTIONS]\n"
 	       "\n"
-	       "Examples:\n"
```
JannePeltonen, line 331: I think it would be good to keep the "Examples" title. The example command looks a bit weird without it. Maybe it would be also good to move the example & config file explanation section after the listing of the mandatory and optional options.

test/performance/odp_ipsecfwd.c
TuomasTaipale: Without brackets in v2. Still uppercase as that seems to the style in almost all of our testers.

test/performance/odp_ipsecfwd.c
TuomasTaipale: Changed to use "E.g." in v2, as again this seems to be the style used in our testers. Also due to current style, better to keep the usage above the options sections.

JannePeltonen (approved):

PR #1868, opened 2023-07-14
[PATCH v4] test: timer_perf: add new mode which starts timers again as they expire

JereLeppanen:
Add new mode (-m 2), which measures odp_schedule() overhead while continuously restarting expiring timers.

### DIFF
--- a/test/performance/odp_timer_perf.c
+++ b/test/performance/odp_timer_perf.c
@@ -16,7 +16,8 @@
 #include <odp/helper/odph_api.h>
 
 #define MODE_SCHED_OVERH  0
-#define MODE_SET_CANCEL   1
+#define MODE_START_CANCEL 1
+#define MODE_START_EXPIRE 2
 #define MAX_TIMER_POOLS   32
 #define MAX_TIMERS        10000
 #define START_NS          (100 * ODP_TIME_MSEC_IN_NS)
@@ -79,6 +80,7 @@ typedef struct thread_arg_t {
 
 typedef struct timer_ctx_t {
 	uint64_t target_ns;
+	uint64_t target_tick;
 	uint32_t tp_idx;
 	uint32_t timer_idx;
 	int last;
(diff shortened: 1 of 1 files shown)


test/performance/odp_timer_perf.c
```diff
@@ -17,6 +17,7 @@
 
 #define MODE_SCHED_OVERH  0
 #define MODE_SET_CANCEL   1
+#define MODE_SET_EXPIRE   2
```
psavol, line 20: Since set functions have been replaced by start functions. This could be named as e.g. START_EXPIRE or EXPIRE_RESTART. Also, the SET_CANCEL could be renamed similarly.

test/performance/odp_timer_perf.c
```diff
@@ -805,6 +824,130 @@ static int set_cancel_mode_worker(void *arg)
...
+		meas = 1;
+		tmo = odp_timeout_from_event(ev);
+		ctx = odp_timeout_user_ptr(tmo);
+		i = ctx->tp_idx;
+		j = ctx->timer_idx;
+		timer = global->timer[i][j];
+		tp = global->timer_pool[i].tp;
+		period_tick = global->timer_pool[i].period_tick;
+		current_tick = odp_timer_current_tick(tp);
+		time_ns = odp_timer_tick_to_ns(tp, current_tick);
```
psavol, line 891: I think time API should be used for accuracy measurement (as the other test case does). Time API is likely more accurate than timer ticks (and has strict version).

test/performance/odp_timer_perf.c
JereLeppanen: v2.

test/performance/odp_timer_perf.c
```diff
@@ -463,6 +475,13 @@ static int set_timers(test_global_t *global)
 			start_param.tick = tick_cur + tick_ns;
 			start_param.tmo_ev = ev;
 
+			if (test_options->mode == MODE_SET_EXPIRE) {
+				start_param.tick = tick_cur +
+					odp_timer_ns_to_tick(tp,
+							     period_ns + j * period_ns / num_timer);
+				ctx->target_ns = odp_timer_tick_to_ns(tp, start_param.tick);
```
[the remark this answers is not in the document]
JereLeppanen: It's compared to current tick of the timer pool. But anyway changed to global time in v3.

test/performance/odp_timer_perf.c
JereLeppanen: v3.

psavol (approved):

JereLeppanen:
v4:
- Rebase.
- Add review tag.

PR #1876, opened 2023-08-11
[PATCH v6] test: dma_perf: drain poll transfers before teardown

PavanNikhilesh:
Wait for transfers to complete in asynchronous poll mode before
tearing down the associated memory.

Reviewed-by: Tuomas Taipale <tuomas.taipale@nokia.com>

### DIFF
--- a/test/performance/odp_dma_perf.c
+++ b/test/performance/odp_dma_perf.c
@@ -156,7 +156,7 @@ typedef struct {
 	/* Wait and handle finished transfer. */
 	void (*wait_fn)(sd_t *sd, stats_t *stats);
 	/* Handle all unfinished transfers after main test has been stopped. */
-	void (*drain_fn)(void);
+	void (*drain_fn)(sd_t *sd);
 	/* Free any resources that might have been allocated during setup phase. */
 	void (*free_fn)(const sd_t *sd);
 } test_api_t;
@@ -1111,7 +1111,7 @@ static void wait_compl_event(sd_t *sd, stats_t *stats)
 	}
 }
 
-static void drain_compl_events(void)
+static void drain_compl_events(ODP_UNUSED sd_t *sd)
 {
 	odp_event_t ev;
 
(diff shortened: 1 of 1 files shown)


test/performance/odp_dma_perf.c
```diff
@@ -1123,6 +1124,21 @@ static void drain_compl_events(void)
 	}
 }
 
+static void
+drain_poll_transfers(sd_t *sd)
+{
+	const uint32_t count = sd->dma.num_inflight;
+	trs_info_t *infos = sd->dma.infos, *info;
+	odp_dma_t handle = sd->dma.handle;
+	odp_dma_result_t res;
```
TuomasTaipale, line 1133: Result usage is optional so perhaps `NULL` could be used directly in `odp_dma_transfer_done()` below, making this unnecessary.

test/performance/odp_dma_perf.c
```diff
@@ -1123,6 +1124,21 @@ static void drain_compl_events(void)
...
+
+	for (uint32_t i = 0U; i < count; ++i) {
+		info = &infos[i];
```
TuomasTaipale, line 1136: There could be a newline after this assignment and before the `if` below.

test/performance/odp_dma_perf.c
```diff
@@ -1123,6 +1124,21 @@ static void drain_compl_events(void)
...
+		if (info->is_running)
+			odp_dma_transfer_done(handle, info->compl_param.transfer_id, &res);
```
TuomasTaipale, line 1138: Would the draining be more reliable if there was a `while` here? So, looping as long as `odp_dma_transfer_done()` returns 0 ("transfer has not finished").

test/performance/odp_dma_perf.c
PavanNikhilesh: Ack, I will modify in v2.

test/performance/odp_dma_perf.c
```diff
@@ -1111,7 +1111,8 @@ static void wait_compl_event(sd_t *sd, stats_t *stats)
 	}
 }
 
-static void drain_compl_events(void)
+static void
+drain_compl_events(ODP_UNUSED sd_t *sd)
```
TuomasTaipale, line 1115: Existing style in file is to keep return type and name at the same line.

TuomasTaipale (approved):
One cosmetic comment, otherwise OK, thanks for the patch.

Reviewed-by: Tuomas Taipale <tuomas.taipale@nokia.com>

PR #1877, opened 2023-08-11
[PATCH v5] test: sched_pktio: remove unnecessary delay first tick

PavanNikhilesh:
Remove unnecessary delay for the first tick as it might cause
ODP_TIMER_TOO_FAR error base on the timer pool created.
Reorder start_pktios with start_timers to prevent timeouts due
to delays in pktio start.

### DIFF
--- a/test/performance/odp_sched_pktio.c
+++ b/test/performance/odp_sched_pktio.c
@@ -26,7 +26,6 @@
 #define MIN_PKT_SEG_LEN   64
 #define CHECK_PERIOD      10000
 #define TEST_PASSED_LIMIT 5000
-#define TIMEOUT_OFFSET_NS 1000000
 #define SCHED_MODE_PARAL  1
 #define SCHED_MODE_ATOMIC 2
 #define SCHED_MODE_ORDER  3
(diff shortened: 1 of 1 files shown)


test/performance/odp_sched_pktio.c
```diff
@@ -1337,10 +1336,6 @@ static int start_timers(test_global_t *test_global)
 	if (test_global->opt.timeout_us == 0)
 		return 0;
 
-	/* Delay the first timeout so that workers have time to startup */
-	timeout_tick += odp_timer_ns_to_tick(test_global->timer.timer_pool,
-					     TIMEOUT_OFFSET_NS);
-
```
MatiasElo, line 1343: After removing the offset, timers can trigger too early if `start_workers()` or `start_pktios()` calls are slow. Perhaps the simplest solution to mitigate this potential issue would be to reorder the functions calls in `main()`, so that `start_timers()` is the last called function before `odp_barrier_wait()` and `start_pktios()` is moved up before `start_timers()`.

PavanNikhilesh: ack, I will reorder the calls in v2. Thanks!

MatiasElo (approved):

test/performance/odp_sched_pktio.c
```diff
@@ -1532,23 +1527,21 @@ int main(int argc, char *argv[])
...
+	if (start_pktios(test_global))
 		goto quit;
 
 	odp_barrier_init(&test_global->worker_start,
 			 test_global->opt.num_worker + 1);
 
 	start_workers(thread, test_global);
 
+	if (start_timers(test_global))
+		goto quit;
```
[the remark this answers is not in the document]
PavanNikhilesh: will fix in next version.

MatiasElo (approved):

PR #1887, opened 2023-08-28
[PATCH v5] api: pktio: add ODP_PKTIO_MAX_INDEX define

MatiasElo:
Add new API define ODP_PKTIO_MAX_INDEX for maximum packet IO interface
index. The value of ODP_PKTIO_MAX_INDEX is always greater or equal to
odp_pktio_max_index(), so the new define can be used for example when
defining static length arrays.

V3:
- Fixed rebase failure

### DIFF
--- a/include/odp/api/spec/packet_io_types.h
+++ b/include/odp/api/spec/packet_io_types.h
@@ -59,6 +59,12 @@ extern "C" {
  */
 
 /**
+ * @def ODP_PKTIO_MAX_INDEX
+ * Maximum packet IO interface index. Use odp_pktio_max_index() to check the
+ * runtime maximum value, which may be smaller than this value.
+ */
+
+/**
  * @def ODP_PKTIO_MACADDR_MAXSIZE
  * Minimum size of output buffer for odp_pktio_mac_addr()
  * Actual MAC address sizes may be different.
--- a/platform/linux-generic/include-abi/odp/api/abi/packet_io_types.h
+++ b/platform/linux-generic/include-abi/odp/api/abi/packet_io_types.h
@@ -1,5 +1,5 @@
 /* Copyright (c) 2015-2018, Linaro Limited
- * Copyright (c) 2020-2022, Nokia
+ * Copyright (c) 2020-2023, Nokia
  * All rights reserved.
  *
  * SPDX-License-Identifier:     BSD-3-Clause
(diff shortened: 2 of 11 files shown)


include/odp/api/spec/packet_io_types.h
```diff
@@ -60,6 +60,12 @@ extern "C" {
  * Invalid LSO profile handle
  */
 
+/**
+ * @def ODP_PKTIO_MAX_INDEX
+ * Maximum packet IO interface index. Use odp_pktio_max_index() to check the
+ * runtime maximum value, which may be smaller than this define.
```
TuomasTaipale, line 66: "... than this define" -> "... than this" or "... than this value" would perhaps sound more natural.

test/performance/odp_l2fwd.c
```diff
@@ -1411,8 +1408,8 @@ static void init_port_lookup_tbl(void)
 		int pktio_idx     = odp_pktio_index(pktio);
 		int dst_port      = find_dest_port(rx_idx);
 
-		if (pktio_idx < 0 || pktio_idx >= MAX_PKTIO_INDEXES) {
-			ODPH_ERR("Bad pktio index %i\n", pktio_idx);
+		if (pktio_idx < 0) {
+			ODPH_ERR("Reading pktio index failed: %i\n", pktio_idx);
```
TuomasTaipale, line 1412: Could be useful in debugging to print something about the failing handle, e.g. `odp_pktio_to_u64()`.

platform/linux-generic/include-abi/odp/api/abi/packet_io_types.h
```diff
@@ -43,6 +43,8 @@ typedef struct odp_pktout_queue_t {
 #define ODP_PKTIO_INVALID _odp_cast_scalar(odp_pktio_t, 0)
 #define ODP_LSO_PROFILE_INVALID _odp_cast_scalar(odp_lso_profile_t, 0)
 
+#define ODP_PKTIO_MAX_INDEX 64
```
TuomasTaipale, line 46: Would a more suitable default actually be 63?

include/odp/api/spec/packet_io_types.h
MatiasElo: Changed to "than this value" in V2.

test/performance/odp_l2fwd.c
MatiasElo: Added device name to error print in V2.

platform/linux-generic/include-abi/odp/api/abi/packet_io_types.h
MatiasElo: True, changed in V2.

TuomasTaipale (approved):

psavol (approved):

jerinjacobk (approved):

PR #1897, opened 2023-09-01
[PATCH v4] linux-gen: pool: fix odp_pool_ext_capability() return value

MatiasElo:
According to the specification, odp_pool_ext_capability() should set
'max_pools' to zero and return success if a valid but unsupported pool type
is used.

Add tests to validate that odp_pool_ext_capability() works according to the
specification when unsupported pool type is used.

### DIFF
--- a/platform/linux-generic/odp_pool.c
+++ b/platform/linux-generic/odp_pool.c
@@ -1856,8 +1856,21 @@ int odp_pool_ext_capability(odp_pool_type_t type, odp_pool_ext_capability_t *cap
 {
 	odp_pool_stats_opt_t supported_stats;
 
-	if (type != ODP_POOL_PACKET)
+	_ODP_ASSERT(capa != NULL);
+
+	switch (type) {
+	case ODP_POOL_PACKET:
+		break;
+	case ODP_POOL_BUFFER:
+	case ODP_POOL_TIMEOUT:
+	case ODP_POOL_VECTOR:
+	case ODP_POOL_DMA_COMPL:
+		memset(capa, 0, sizeof(odp_pool_ext_capability_t));
+		return 0;
+	default:
+		_ODP_ERR("Invalid pool type: %d\n", type);
 		return -1;
+	}
 
 	supported_stats.all = 0;
 
--- a/test/validation/api/pool/pool.c
+++ b/test/validation/api/pool/pool.c
@@ -1698,7 +1698,19 @@ static void pool_ext_init_packet_pool_param(odp_pool_ext_param_t *param)
 static void test_packet_pool_ext_capa(void)
 {
 	odp_pool_ext_capability_t capa;
-	odp_pool_type_t type = ODP_POOL_PACKET;
+	odp_pool_type_t type;
+	const odp_pool_type_t unsupported_types[] = {ODP_POOL_BUFFER, ODP_POOL_TIMEOUT,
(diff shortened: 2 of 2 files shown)


platform/linux-generic/odp_pool.c
```diff
@@ -1856,8 +1856,21 @@ int odp_pool_ext_capability(odp_pool_type_t type, odp_pool_ext_capability_t *cap
...
+	case ODP_POOL_PACKET:
+		break;
+	case ODP_POOL_BUFFER:
+	case ODP_POOL_TIMEOUT:
+	case ODP_POOL_VECTOR:
+	case ODP_POOL_DMA_COMPL:
+		capa->max_pools = 0;
+		return 0;
+	default:
+		_ODP_ERR("Invalid pool type\n");
```
TuomasTaipale, line 1871: The erroneous type could be printed here.

test/validation/api/pool/pool.c
```diff
@@ -1698,7 +1698,19 @@ static void pool_ext_init_packet_pool_param(odp_pool_ext_param_t *param)
 static void test_packet_pool_ext_capa(void)
 {
 	odp_pool_ext_capability_t capa;
-	odp_pool_type_t type = ODP_POOL_PACKET;
+	odp_pool_type_t type;
+	odp_pool_type_t unsupported_types[4] = {ODP_POOL_BUFFER, ODP_POOL_TIMEOUT, ODP_POOL_VECTOR,
```
TuomasTaipale, line 1702: The explicit size "4" is unnecessary. The array could also be `const`.

test/validation/api/pool/pool.c
```diff
@@ -1698,7 +1698,19 @@ static void pool_ext_init_packet_pool_param(odp_pool_ext_param_t *param)
...
+						ODP_POOL_DMA_COMPL};
+	const int num_types = sizeof(unsupported_types) / sizeof(unsupported_types[0]);
```
TuomasTaipale, line 1704: No need to be this PR, but an array size helper macro could also be useful (now that MIN/MAX ones are incoming #1886).

platform/linux-generic/odp_pool.c
MatiasElo: Added in V2.

test/validation/api/pool/pool.c
MatiasElo: True, I made a note to add a helper macro.

platform/linux-generic/odp_pool.c
```diff
@@ -1856,8 +1856,21 @@ int odp_pool_ext_capability(odp_pool_type_t type, odp_pool_ext_capability_t *cap
...
+	_ODP_ASSERT(capa != NULL);
+
+	switch (type) {
+	case ODP_POOL_PACKET:
+		break;
+	case ODP_POOL_BUFFER:
+	case ODP_POOL_TIMEOUT:
+	case ODP_POOL_VECTOR:
+	case ODP_POOL_DMA_COMPL:
+		capa->max_pools = 0;
```
psavol, line 1868: It would be safer to memset(0) also in this case. If application would miss the max_pools check, it would bump into all other fields being zero (instead of uninitialized).

MatiasElo: Changed in V4.

PR #1899, opened 2023-09-05
[PATCH v3] example: sysinfo: print more capabilities

MatiasElo:
V2:
- Removed extra newlines
- Print clock source ID for all timer clock sources

### DIFF
--- a/example/sysinfo/odp_sysinfo.c
+++ b/example/sysinfo/odp_sysinfo.c
@@ -29,6 +29,7 @@
 typedef struct {
 	char name[MAX_NAME_LEN];
 	odp_pktio_capability_t capa;
+	odp_proto_stats_capability_t proto_stats_capa;
 } pktio_t;
 
 typedef struct {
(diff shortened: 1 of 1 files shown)


example/sysinfo/odp_sysinfo.c
```diff
@@ -361,7 +380,209 @@ static void print_auth(odp_auth_alg_t alg)
...
+{
+	char *tmp_str = strdup(optarg), *tmp;
+
+	if (tmp_str == NULL)
+		return;
+
+	tmp = strtok(tmp_str, ",");
+
+	while (tmp && config->num_pktio < MAX_IFACES) {
+		config->pktio[config->num_pktio].name = strdup(tmp);
```
psavol, line 623: To avoid memory leak possibility (strdup() here and free() elsewhere), memory for pktio name could have a static size (e.g. 128 chars) and just strncpy() here.

psavol (approved):

PR #1901, opened 2023-09-06
[PATCH v2] validation: fix compilation error

ashwinyes:
Fix compilation errors due to uninitialized variables.

### DIFF
--- a/test/validation/api/buffer/buffer.c
+++ b/test/validation/api/buffer/buffer.c
@@ -537,7 +537,7 @@ static void buffer_test_user_area(void)
 
 	for (i = 0; i < num; i++) {
 		odp_event_t ev;
-		int flag;
+		int flag = 0;
 
 		buffer[i] = odp_buffer_alloc(pool);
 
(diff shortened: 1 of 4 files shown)


test/validation/api/buffer/buffer.c
```diff
@@ -537,7 +537,7 @@ static void buffer_test_user_area(void)
 
 	for (i = 0; i < num; i++) {
 		odp_event_t ev;
-		int flag;
+		int flag = 0;
```
MatiasElo, line 540: According to the API spec, the user flag value should be always written by `odp_event_user_area_and_flag()`. Is this actually an implementation issue or doesn't some compiler recognize the initialization properly?

ashwinyes: Its a compiler issue. It comes up when compiling with specific flags.

--enable-shared --enable-static --enable-lto --enable-test-vald 'CFLAGS=-O3 -ggdb -fno-var-tracking-assignments' 'CXXFLAGS=-O3 -ggdb -fno-var-tracking-assignments'

MatiasElo: Okay, commit message could then state that the compiler errors were invalid in this case.

MatiasElo (approved):

PR #1902, opened 2023-09-07
linux-gen: pktio: remove packet vector tx code

MatiasElo:


platform/linux-generic/odp_packet_io.c
```diff
@@ -829,27 +829,24 @@ static int pktout_enqueue(odp_queue_t queue, _odp_event_hdr_t *event_hdr)
 {
 	odp_packet_t pkt = packet_from_event_hdr(event_hdr);
 	odp_pktout_queue_t pktout_queue;
```
TuomasTaipale, line 831: This could also be dropped and use `_odp_queue_fn->get_pktout(queue)` directly with `odp_pktout_send` (same in `pktout_enq_multi()`).

MatiasElo: Dropped in v2.

PR #1913, opened 2023-09-20
[PATCH v6] api: queue: add queue create/destroy function multi variants

MatiasElo:
Add new odp_queue_create_multi() function for creating multiple queues with a single call.

V2:
- Changed `odp_queue_create_multi()` parameter `const odp_queue_param_t *param[]` to `const odp_queue_param_t param[]`
- Added implementation and validation test

V3:
- Added `odp_queue_destroy_multi()` API

### DIFF
--- a/include/odp/api/spec/queue.h
+++ b/include/odp/api/spec/queue.h
@@ -1,5 +1,6 @@
 /* SPDX-License-Identifier: BSD-3-Clause
  * Copyright (c) 2013-2018 Linaro Limited
+ * Copyright (c) 2023 Nokia
  */
 
 /**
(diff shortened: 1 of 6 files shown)


include/odp/api/spec/queue.h
```diff
@@ -44,6 +45,32 @@ extern "C" {
...
+/**
+ * Create multiple queues
+ *
+ * Otherwise like odp_queue_create(), but creates multiple queues with a single
+ * call. The output queue handles are written in the same order as input
+ * parameters. A single odp_queue_create_multi() call is equivalent to calling
+ * odp_queue_create() 'num' times in row.
+ *
+ * If 'share_param' value is false, 'param' array must contain 'num' elements.
+ * If the value is true, only a single element is required. If 'name' array is
```
psavol, line 57: This could highlight the purpose of the flag (share the same params): "If the value is true, only a single element is required and it is used as queue parameters for all created queues."

MatiasElo: Added in V2.

include/odp/api/spec/queue.h
```diff
@@ -88,6 +88,20 @@ int odp_queue_create_multi(const char *name[], const odp_queue_param_t param[],
...
+/**
+ * Destroy multiple queues
+ *
+ * Otherwise like odp_queue_destroy(), but destroys multiple queues with a
+ * single call.
+ *
+ * @param queue    Array of queue handles
+ * @param num      Number of queues to destroy
+ *
+ * @retval Number of queues actually destroyed (0 ... num)
```
psavol, line 100: In destroy direction APIs, 0 may not be helpful return value. If first handle is invalid, function should return -1. Also function should be always able to free good handles.

MatiasElo: Changed in V4.

jerinjacobk (approved):

PR #1914, opened 2023-09-22
[PATCH v3] validation: timer: test timeout event reuse with periodic timer

JereLeppanen:
When starting again a canceled periodic timer, test also event reuse. Instead of freeing the last timeout event and allocating a new one, just reuse the event.

### DIFF
--- a/test/validation/api/timer/timer.c
+++ b/test/validation/api/timer/timer.c
@@ -2775,7 +2775,8 @@ static void timer_test_periodic_capa(void)
 	}
 }
 
-static void timer_test_periodic(odp_queue_type_t queue_type, int use_first, int rounds)
+static void timer_test_periodic(odp_queue_type_t queue_type, int use_first, int rounds,
+				int reuse_event)
 {
 	odp_timer_capability_t timer_capa;
 	odp_timer_periodic_capability_t periodic_capa;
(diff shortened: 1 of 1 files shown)


test/validation/api/timer/timer.c
```diff
@@ -2918,9 +2919,11 @@ static void timer_test_periodic(odp_queue_type_t queue_type, int use_first, int
 		num_tmo = 0;
 		done = 0;
 
-		tmo = odp_timeout_alloc(pool);
-		ev  = odp_timeout_to_event(tmo);
-		CU_ASSERT_FATAL(ev != ODP_EVENT_INVALID);
+		if (!reuse_event || round == 0) {
+			tmo = odp_timeout_alloc(pool);
+			ev  = odp_timeout_to_event(tmo);
+			CU_ASSERT_FATAL(ev != ODP_EVENT_INVALID);
```
MatiasElo, line 2925: Perhaps it would be safer to run the assert every round (or add another fatal check to the end of the loop), since it looks like `ev` could be `ODP_EVENT_INVALID` in case of failure timeout in the final while loop.

JereLeppanen: Assert every round in v2.

MatiasElo (approved):

JereLeppanen:
v3:
- Rebase.
- Add review tags.

PR #1915, opened 2023-09-22
[PATCH v2] validation: cls: add max pmr create test case

psavol:
Test how many PMRs can be added to the default CoS. There is no max number of PMR capability.

### DIFF
--- a/test/validation/api/classification/odp_classification_basic.c
+++ b/test/validation/api/classification/odp_classification_basic.c
@@ -31,13 +31,13 @@ static void test_defaults(uint8_t fill)
 	CU_ASSERT_EQUAL(pmr_param.range_term, false);
 }
 
-static void classification_test_default_values(void)
+static void cls_default_values(void)
 {
 	test_defaults(0);
 	test_defaults(0xff);
 }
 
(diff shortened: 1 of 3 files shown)


test/validation/api/classification/odp_classification_basic.c
```diff
@@ -214,6 +219,140 @@ static void cls_create_pmr_match(void)
...
+		pmr[i] = odp_cls_pmr_create(&pmr_param, 1, default_cos, cos[i + 1]);
+
+		if (pmr[i] == ODP_PMR_INVALID)
+			break;
+
+		val++;
+		pmr_created++;
+	}
+
+	printf("\n    Number of CoS created: %u\n    Number of PMR created %u\n", cos_created,
```
MatiasElo, line 326: Colon missing from `Number of PMR created`.

MatiasElo (approved):
One minor print tweak.

test/validation/api/classification/odp_classification_basic.c
psavol: Fixed in v2

PR #1917, opened 2023-09-25
[PATCH v4] example: timer_accuracy: cpu count and other new options

JereLeppanen:


psavol:
Typo in the first commit log heading and text: timeour -> timeout
example: timer_accuracy: make first timeout diff and final timeour error timer specific

example/timer/odp_timer_accuracy.c
```diff
@@ -100,10 +109,9 @@ typedef struct test_global_t {
 	uint64_t         period_tick;
 	double           period_dbl;
 	odp_fract_u64_t  base_freq;
-	odp_shm_t        log_shm;
-	test_log_t	*log;
 	FILE            *file;
 	char		 filename[MAX_FILENAME + 1];
+	test_shm_t	*shm;
 
 } test_global_t;
```
psavol, line 121: It would be cleanest to allocate entire test_global_t from SHM. All variables would be from the same memory and on the same level on the struct. It would safer option also if running in process mode.

The move of test_global_t into SHM could be done in a separate patch to minimize size this patch.

example/timer/odp_timer_accuracy.c
```diff
@@ -876,36 +946,43 @@ static void cancel_periodic_timers(test_global_t *test_global)
...
+	odp_barrier_wait(&shm->barrier);
 
-	for (i = 0; i < tot_timers; i++) {
-		ev = odp_schedule(NULL, ODP_SCHED_WAIT);
+	double period_dbl = test_global->period_dbl;
+	odp_timer_pool_t tp = test_global->timer_pool;
+
+	while (1) {
+		if (odp_atomic_load_u64(&shm->events) >= tot_timers)
+			break;
```
psavol, line 1043: This check on every round and before schedule call adds latency (if tmo event is waiting in a queue). So, better to move it under "ev == INVALID" branch, and maybe schedule call could have e.g. a 100msec timeout instead of NO_WAIT.

The "Free current scheduler context" loop can be removed from the end of the function, if we break schedule loop only on EVENT_INVALID.

example/timer/odp_timer_accuracy.c
```diff
@@ -876,36 +946,43 @@ static void cancel_periodic_timers(test_global_t *test_global)
...
+
+		ev = odp_schedule(NULL, ODP_SCHED_NO_WAIT);
+
+		if (ev == ODP_EVENT_INVALID)
+			continue;
 
-		time = odp_time_local();
+		uint64_t events = odp_atomic_fetch_inc_u64(&shm->events);
+
+		time = odp_time_global();
```
psavol, line 1052: Time stamp call could be strict and directly after schedule call (when it returned event). Atomic inc in between could add e.g. 100nsec of latency (likely a cache miss since some other thread updated the value in between.

example/timer/odp_timer_accuracy.c
```diff
@@ -1107,63 +1177,92 @@ int main(int argc, char *argv[])
...
+		goto quit;
 	}
 
+	addr = odp_shm_addr(shm);
+	memset(addr, 0, size);
+	test_global.shm = addr;
+	test_global.shm->timer_ctx = (timer_ctx_t *)((test_shm_t *)addr + 1);
+	if (test_global.opt.output)
+		test_global.shm->log =
+			(test_log_t *)&test_global.shm->timer_ctx[test_global.alloc_timers];
```
psavol, line 1284: The log can remain as a separate SHM. It simplifies code when two SHMs are used (avoids dependency between timer_ctx and log). It also decreases the lines modified by this patch (more focus on the changes actually needed).

example/timer/odp_timer_accuracy.c
```diff
@@ -1107,63 +1177,92 @@ int main(int argc, char *argv[])
...
+
+	odph_thread_create(thread_tbl, &thr_common, &thr_param, num_workers);
+
 	ret = start_timers(&test_global);
 	if (ret)
 		goto quit;
 
-	run_test(&test_global);
-
+	odp_barrier_wait(&test_global.shm->barrier);
```
psavol, line 1324: With multiple threads (and with the barrier synchronizing the start up), it would be safer to split start_timers() into two parts:
1) create all timers (exit if returns fail)
2) start all workers, wait until those are running (barrier)
3) start timers (get current time + odp_timer_start() loop)

Main point is that barrier after timer start prevents all workers from handling the first timeouts, if one thread is slower to start / got interrupted before the barrier.

example/timer/odp_timer_accuracy.c
```diff
@@ -516,7 +526,7 @@ static int start_timers(test_global_t *test_global)
 	odp_queue_param_init(&queue_param);
 	queue_param.type        = ODP_QUEUE_TYPE_SCHED;
 	queue_param.sched.prio  = odp_schedule_default_prio();
-	queue_param.sched.sync  = ODP_SCHED_SYNC_PARALLEL;
+	queue_param.sched.sync  = test_global->opt.queue_type;
```
psavol, line 564: SCHED_SYNC_XXX values are implementation specific. So, need to map user input to a value: 0->PARALLEL, etc

example/timer/odp_timer_accuracy.c
```diff
@@ -531,13 +541,33 @@ static int start_timers(test_global_t *test_global)
...
 
+	if (test_global->opt.groups) {
+		/* Create groups */
+
+		odp_thrmask_t zero;
+
+		odp_thrmask_zero(&zero);
+
+		for (i = 0; i < (uint64_t)test_global->opt.num_queue; i++) {
+			group[i] = odp_schedule_group_create(NULL, &zero);
```
psavol, line 552: A dedicated group per worker would be more flexible option. Then there could be more queues than groups. A group for each queue would be selected in round-robin.

JereLeppanen:
v2:
- All comments.
- Fix negative values when printing results.
- Parse helper command line arguments.

example/timer/odp_timer_accuracy.c
```diff
@@ -814,9 +816,10 @@ static void print_stat(test_global_t *test_global)
 
 		for (int i = 0; i < (int)test_global->alloc_timers; i++) {
 			timer_ctx_t *t = &test_global->timer_ctx[i];
+			int64_t v = t->first_tmo_diff;
 
-			if (t->first_tmo_diff * t->first_tmo_diff > max * max) {
-				max = t->first_tmo_diff;
+			if (ABS(v) > ABS(max)) {
+				max = v;
```
psavol, line 978: Please move usage of ABS() into the first patch. It makes code more readable and avoids unnecessary changes of these  same lines.

example/timer/odp_timer_accuracy.c
```diff
@@ -1072,15 +1072,12 @@ int main(int argc, char *argv[])
...
+	test_global_t *test_global;
 	odp_init_t *init_ptr = NULL;
 	int ret = 0;
 
-	memset(&test_global, 0, sizeof(test_global_t));
-	test_global.stat.nsec_before_min = UINT64_MAX;
-	test_global.stat.nsec_after_min  = UINT64_MAX;
-
-	if (parse_options(argc, argv, &test_global))
+	if (parse_options(argc, argv, &test_opt))
```
psavol, line 1254: Test_opt should be memset() to zero to ensure that no member is left uninitialized.

example/timer/odp_timer_accuracy.c
```diff
@@ -865,51 +947,48 @@ static void cancel_periodic_timers(test_global_t *test_global)
...
 
-		time = odp_time_local();
+			} else if (odp_atomic_load_u64(&test_global->events) >= tot_timers) {
+				break;
+			}
+
+			continue;
+		}
+
+		time = odp_time_global_strict();
```
psavol, line 1074: Likely not a big deal, but this time stamp call could be still moved right after the schedule call. It does not matter if we read timestamp and throw it away few times per second. Maximum accuracy is more important than saving a bit on CPU load.

example/timer/odp_timer_accuracy.c
```diff
@@ -1000,6 +1042,17 @@ static int run_test(void *arg)
 	stat->nsec_before_min = UINT64_MAX;
 	stat->nsec_after_min = UINT64_MAX;
 
+	if (test_global->opt.groups) {
+		odp_thrmask_zero(&mask);
+		odp_thrmask_set(&mask, tid);
+		group = test_global->group[tid - 1];
```
psavol, line 1048: It would be good to check tid < cpu_count, before using it. Just in case that thread IDs would have a problem.

example/timer/odp_timer_accuracy.c
JereLeppanen: That memset is inside parse_options().

JereLeppanen:
v3:
- Rebase.
- All comments.
- Add review tags.

example/timer/odp_timer_accuracy.c
psavol: It is added there in "add cpu count option" commit. It should be added already in this commit ("use shm for global data"), since this changes memory allocation and initialization.

JereLeppanen: Fixed in v4.

JereLeppanen:
v4:
- Move memset() of test_opt to the right commit.

psavol (approved):

PR #1921, opened 2023-10-04
[PATCH v2] test: performance: add pktio slow path function benchmark application

MatiasElo:


test/performance/bench_common.h
```diff
@@ -108,10 +108,131 @@ void bench_suite_init(bench_suite_t *suite);
...
 /**
- * Run tests suite and print results
+ * Run test suite and print results
+ *
+ * The argument is of type 'bench_suite_t *'.
+ *
+ * @retval 0 on success
+ * @retval <0 on failure
  */
 int bench_run(void *arg);
```
psavol, line 118: Doxygen tags (/**, @retval, ...) could be removed from comments as this file does not seem to be part of doxygen generated documentation (no doxygen errors from missing @param).

test/performance/odp_bench_pktio_sp.c
```diff
@@ -0,0 +1,963 @@
...
+		exit(EXIT_FAILURE);
+	}
+
+	/* Init this thread */
+	if (odp_init_local(instance, ODP_THREAD_CONTROL)) {
+		ODPH_ERR("Local init failed\n");
+		exit(EXIT_FAILURE);
+	}
+
+	ret = (odp_schedule_config(NULL));
```
psavol, line 867: extra brackets around function call

MatiasElo: Fixed in V2.

test/performance/bench_common.h
MatiasElo: Doxygen format removed in V2.

psavol (approved):

PR #1924, opened 2023-10-11
[PATCH v1] linux-gen: ring: move redundant load operation

FanHong674:
In func __atomic_compare_exchange_n(), old_val is already updated if the func fails. Move old_val loading out of the do-while loop to avoid redundant load operation.

### DIFF
--- a/platform/linux-generic/include/odp_ring_mpmc_internal.h
+++ b/platform/linux-generic/include/odp_ring_mpmc_internal.h
@@ -154,8 +154,8 @@ static inline uint32_t _RING_MPMC_DEQ_MULTI(_ring_mpmc_gen_t *ring,
 	 * and thus r_head value is always behind or equal to w_tail value.
 	 * When CAS operation succeeds, this thread owns data between old
 	 * and new r_head. */
+	old_head = odp_atomic_load_acq_u32(&ring->r.r_head);
 	do {
-		old_head = odp_atomic_load_acq_u32(&ring->r.r_head);
 		odp_prefetch(&ring_data[(old_head + 1) & ring_mask]);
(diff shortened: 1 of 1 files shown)


platform/linux-generic/include/odp_ring_mpmc_internal.h
```diff
@@ -154,8 +154,8 @@ static inline uint32_t _RING_MPMC_DEQ_MULTI(_ring_mpmc_gen_t *ring,
 	 * and thus r_head value is always behind or equal to w_tail value.
 	 * When CAS operation succeeds, this thread owns data between old
 	 * and new r_head. */
+	old_head = odp_atomic_load_acq_u32(&ring->r.r_head);
 	do {
-		old_head = odp_atomic_load_acq_u32(&ring->r.r_head);
```
JannePeltonen, line 158: The CAS operation loads r_head using relaxed memory ordering while this atomic load used acquire memory ordering. So moving this load out of the loop changes memory ordering. Without thinking the code through but by just reading the comment above ("Load acquires ensure...") it appears that the change in memory ordering could break things. Even if the change was correct, the commit message should comment the change in memory ordering.

psavol: Yes. If this change is done, then the CAS operation should be changed to "load acquire on failure" as well.

FanHong674: Sorry, it was my mistake not noticing the memory order of CAS operation. Since the modification doesn't improve performance, I guess it's better to leave it as is. 
How can I close/abandon this PR?

JannePeltonen: You can close a pull request by clicking the "Close pull request" button below.

PR #1925, opened 2023-10-11
test: cunit: fix typo

FanHong674:


test/common/odp_cunit_common.c
```diff
@@ -704,7 +704,7 @@ int odp_cunit_register(odp_suiteinfo_t testsuites[])
 }
 
 /*
- * Parse command line options to extract options affectiong cunit_common.
+ * Parse command line options to extract options affecting cunit_common.
```
MatiasElo, line 707: There is also another comment typo in the same file, which could be fixed in this commit:
642: /* call test executable terminason hook, if any */

Please use a capital letter in the start of commit message and add my reviewed-by tag in the next version.

MatiasElo (approved):

PR #1926, opened 2023-10-11
[PATCH v2] test: bench_pktio_sp: add new test for creating pmr rules

MatiasElo:
Add new test case 'cls_create_pmr', which creates a selected number ('-p') of PMRs.

bench_tm_func_record() saves now also the number of measurements, which makes the usage more flexible. Optional precondition check function 'cond' was added to bench_tm_info_t to enable skipping unsupported tests.

### DIFF
--- a/test/performance/odp_bench_pktio_sp.c
+++ b/test/performance/odp_bench_pktio_sp.c
@@ -23,8 +23,8 @@
 /* Maximum interface name length */
 #define MAX_NAME_LEN 128
 
-#define BENCH_INFO(run_fn, init_fn, term_fn, rounds) \
-	{.name = #run_fn, .run = run_fn, .init = init_fn, .term = term_fn, \
+#define BENCH_INFO(run_fn, init_fn, term_fn, cond_fn, rounds) \
+	{.name = #run_fn, .run = run_fn, .init = init_fn, .term = term_fn, .cond = cond_fn,\
 	 .max_rounds = rounds}
 
 typedef struct {
@@ -50,6 +50,9 @@ typedef struct {
 
 		/* Number of packet output queues */
 		uint32_t num_output_queues;
+
+		/* Number of PMRs */
+		uint32_t num_pmr;
 	} opt;
 
 	/* Packet IO device */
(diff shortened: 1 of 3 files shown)


test/performance/odp_bench_pktio_sp.c
```diff
@@ -615,16 +597,204 @@ static int pktout_event_queue_stats(bench_tm_result_t *res, int repeat_count)
...
+	}
+
+	if (queue_created != num_cos)
+		ODPH_ERR("Unable to create all queues: %u/%u\n", queue_created, num_cos);
+
+	if (cos_created != num_cos)
+		ODPH_ERR("Unable to create all CoSes: %u/%u\n", cos_created, num_cos);
+
+	if (cos_created != num_cos)
+		goto destroy_cos;
```
psavol, line 728: Duplication of the same if-clause

test/performance/odp_bench_pktio_sp.c
```diff
@@ -615,16 +597,204 @@ static int pktout_event_queue_stats(bench_tm_result_t *res, int repeat_count)
...
+	for (uint32_t i = 0; i < num_cos; i++) {
+		t1 = odp_time_local_strict();
+		queue[i] = odp_queue_create(NULL, &queue_param);
+
+		if (queue[i] == ODP_QUEUE_INVALID) {
+			ODPH_ERR("odp_queue_create() failed %u / %u\n", i + 1, num_cos);
+			break;
+		}
+		t2 = odp_time_local_strict();
+		bench_tm_func_record(t2, t1, res, id1);
```
psavol, line 704: Since queue / CoS destroys are not recorded. Maybe queue/Cos create/destroy could be another test case (added in another PR). So, these create records could be removed from this test case to make it simpler.

test/performance/odp_bench_pktio_sp.c
MatiasElo: Fixed in v2.

test/performance/odp_bench_pktio_sp.c
MatiasElo: Changed in v2.

psavol (approved):

PR #1930, opened 2023-10-13
[PATCH v2] validation: crypto: reduce duplicate testing in different op modes

JannePeltonen:
Run time consuming algorithm combination tests only once, in async-sched, async-plain or sync mode (in that preference order) instead of in all the three modes, unless FULL_TEST is requested. We assume that the actual crypto implementation being tested is the same in all the three modes so test coverage should not suffer much.

### DIFF
--- a/test/validation/api/crypto/util.h
+++ b/test/validation/api/crypto/util.h
@@ -20,6 +20,7 @@ struct suite_context_s {
 	odp_queue_t queue;
 	odp_queue_type_t q_type;
 	odp_event_t (*compl_queue_deq)(void);
+	int partial_test;
 };
 
 extern struct suite_context_s suite_context;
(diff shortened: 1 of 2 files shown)


test/validation/api/crypto/util.h
```diff
@@ -20,6 +20,7 @@ struct suite_context_s {
 	odp_queue_t queue;
 	odp_queue_type_t q_type;
 	odp_event_t (*compl_queue_deq)(void);
+	int quick_test;
```
TuomasTaipale, line 23: Perhaps this could be `partial_test` or something similar related to scope instead of speed (same change for the new function).

TuomasTaipale (approved):
One minor comment, otherwise looks OK.

PR #1937, opened 2023-10-26
[PATCH v3] linux-gen: config: remove ODP_ prefix from remaining internal defines

MatiasElo:


platform/linux-generic/include/odp_config_internal.h
```diff
@@ -22,7 +22,7 @@ extern "C" {
 /*
  * Maximum number of pools.
  */
-#define ODP_CONFIG_POOLS 32
+#define CONFIG_POOLS 32
```
TuomasTaipale, line 25: Internally "public" so use the `_ODP` prefix in these?

MatiasElo: I'd prefer the maintain the current naming scheme for now as this causes less code changes (3 vs. 24 renames).

TuomasTaipale (approved):

PR #1938, opened 2023-10-26
[PATCH v3] validation: dma: fix packet length when using segmented packets

TuomasTaipale:
When testing DMA transfers with potentially segmented packets, ensure that allocated packet length does not exceed allocated transfer destination memory size.

v3:
- Added reviewed-by tag

### DIFF
--- a/test/validation/api/dma/dma.c
+++ b/test/validation/api/dma/dma.c
@@ -1108,7 +1108,7 @@ static void test_dma_pkt_segs_to_addr_sync(void)
 	uint8_t *dst;
 	odp_packet_t pkt;
 	uint32_t i, len, num_segs;
-	uint32_t pkt_len = global.pkt_len;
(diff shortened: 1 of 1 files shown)


test/validation/api/dma/dma.c
```diff
@@ -1108,7 +1108,7 @@ static void test_dma_pkt_segs_to_addr_sync(void)
 	uint8_t *dst;
 	odp_packet_t pkt;
 	uint32_t i, len, num_segs;
-	uint32_t pkt_len = global.pkt_len;
+	uint32_t pkt_len = ODPH_MIN(global.pkt_len, global.len);
```
MatiasElo, line 1111: The following if statement below seems now unnecessary:
```
	if (len > global.len)
		len = global.len;
```

Otherwise looks OK.

MatiasElo (approved):

PR #1940, opened 2023-10-27
[PATCH v4] linux-gen: timer: avoid overrun debug prints if no timers started

JereLeppanen:
Don't print timer pool "overrun" and "scans missed" debug prints, if no timers have been started.


JereLeppanen:
v2:
- Fix checkpatch errors.

MatiasElo (approved):

JereLeppanen:
v4:
- Rebase.
- Add review tags.

PR #1943, opened 2023-10-31
[PATCH v3] linux-gen: pool: add constant for internal pool count

TuomasTaipale:
Add a new `CONFIG_INTERNAL_POOLS` constant for tracking pools reserved for implementation internal use, such as the pool for IPsec status events. Utilize the new constant when implementation capabilities are queried.

v2:
- Inferred max FD entries from max stash sizes

v3:
- Added reviewed-by tag


platform/linux-generic/odp_fdserver.c
```diff
@@ -72,7 +72,7 @@
 #define FD_DBG  3
 
 /* define the tables of file descriptors handled by this server: */
-#define FDSERVER_MAX_ENTRIES 256
+#define FDSERVER_MAX_ENTRIES 384
```
MatiasElo, line 75: Could perhaps be `CONFIG_SHM_BLOCKS + CONFIG_INTERNAL_SHM_BLOCKS`

MatiasElo (approved):

PR #1945, opened 2023-11-08
[PATCH v2] github_ci: remove ubuntu 18.04 tests

MatiasElo:
Ubuntu 18.04 is already EOL, so remove the OS specific tests from CI.

### DIFF
--- a/.github/workflows/ci-pipeline-arm64.yml
+++ b/.github/workflows/ci-pipeline-arm64.yml
@@ -59,7 +59,7 @@ jobs:
       fail-fast: false
       matrix:
         cc: [gcc, clang]
-        os: ['ubuntu_18.04', 'rocky_linux_8']
+        os: ['rocky_linux_8']
     steps:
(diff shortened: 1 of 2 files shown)


JannePeltonen (approved):

.github/workflows/ci-pipeline-arm64.yml
```diff
@@ -59,7 +59,7 @@ jobs:
       fail-fast: false
       matrix:
         cc: [gcc, clang]
-        os: ['ubuntu_18.04', 'rocky_linux_8']
```
TuomasTaipale, line 62: Are the `ubuntu_18.04-*` Docker images futile as well now and to-be-removed?

MatiasElo: Yes, once we have also updated odp-dpdk. The only image which is required after this is `ubuntu_18.04-i386` as it is the last Ubuntu version supporting i386.

TuomasTaipale (approved):

PR #1946, opened 2023-11-09
[PATCH v3] linux-gen: fix unaligned access in checksum calculation

JannePeltonen:
Architectures that support unaligned access do not necessarily support it for all instructions that a compiler might choose to use. Therefore we should not mispresent the alignment of an object to the compiler even when unaligned access is typically supported.

The current checksum code casts a potentially unaligned pointer to uint32_t when _ODP_UNALIGNED is set. This typically works but may, in some cases, result in generation of invalid code. In particular, when targeting AArch32 (which supports unaligned access for regular loads and stores) gcc-12.2 generated LDM instructions which always reguire 4-byte alignment.

Fix the problem by informing the compiler about the non-standard alignment by using a type attribute.

### DIFF
--- a/platform/linux-generic/include/odp_chksum_internal.h
+++ b/platform/linux-generic/include/odp_chksum_internal.h
@@ -1,4 +1,4 @@
-/* Copyright (c) 2020, Nokia
+/* Copyright (c) 2020, 2023, Nokia
  * All rights reserved.
  *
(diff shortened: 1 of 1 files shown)


JereLeppanen (approved):

JannePeltonen:
v2: there was still a possible unaligned uint16_t pointer that is now fixed. Also fixed the casts to use the unaligned types.

PR #1947, opened 2023-11-13
[PATCH v4] api: event: add odp_event_pool() function

MatiasElo:


include/odp/api/spec/event.h
```diff
@@ -93,6 +94,19 @@ void odp_event_types_multi(const odp_event_t event[], odp_event_type_t type[],
...
+/**
+ * Event pool
+ *
+ * Returns handle to the pool where the event was allocated from. If the
+ * underlying event type does not have API for returning pool handle
+ * (e.g. ODP_EVENT_IPSEC_STATUS), ODP_POOL_INVALID is returned.
+ *
+ * @param      event    Event handle
+ *
+ * @return Pool handle or ODP_POOL_INVALID
```
TuomasTaipale, line 106: The erroneous `ODP_POOL_INVALID` value could be documented with `@retval`, similar to the user area functions.

include/odp/api/spec/event.h
```diff
@@ -93,6 +94,19 @@ void odp_event_types_multi(const odp_event_t event[], odp_event_type_t type[],
 int odp_event_type_multi(const odp_event_t event[], int num,
 			 odp_event_type_t *type);
 
+/**
+ * Event pool
+ *
+ * Returns handle to the pool where the event was allocated from. If the
+ * underlying event type does not have API for returning pool handle
```
TuomasTaipale, line 101: "an" before "API" would sound a bit more natural.

MatiasElo: Changed in V2.

include/odp/api/spec/event.h
MatiasElo: The idea is to leave error handling undefined to minimize overhead for the implementation. I was thinking about adding a separate `@retval` for `ODP_POOL_INVALID`, but  ended up going with this since `ODP_POOL_INVALID` is specified in the previous sentence.

Something like this could be the other option:
```
 * @return Pool handle
 * @retval ODP_POOL_INVALID  Event pool not available
```

TuomasTaipale: > ... leave error handling undefined ...

The unavailablity of pool could already be interpreted as requiring error handling. Probably need to somehow explicitly state the above if that's the idea.

The `@retval ` could also be reworded so that it does not include blanket error handling.

MatiasElo: Reworded in V2.

TuomasTaipale (approved):

ashwinyes:

PR #1948, opened 2023-11-15
[PATCH v2] validation: pool: check only requested thread statistics

psavol:
When testing odp_pool_stats(), check thread.cache_available[] counters only for the number of threads requested. Also, added statistics struct initialization to catch more robustly output errors in the functions.


ashwinyes:

MatiasElo (approved):

PR #1954, opened 2023-11-22
[PATCH v3] api: time: add nanoseconds into time stamp

psavol:


include/odp/api/spec/time.h
```diff
@@ -127,6 +127,19 @@ odp_time_t odp_time_diff(odp_time_t t2, odp_time_t t1);
  */
 uint64_t odp_time_diff_ns(odp_time_t t2, odp_time_t t1);
 
+/**
+ * Add nanoseconds into time
+ *
+ * Adds 'ns' nanoseconds into the time stamp value. The resulting time may wrap around, if
```
JannePeltonen, line 133: I wonder if addition of this new API function is really beneficial. The current API requires first converting nanoseconds to odp_time_t type and then adding the timestamps. I would suppose non-performance critical code can continue doing those two steps explicitly and performance critical code would be better of moving the conversion step out of the critical path.

psavol: Performance wise true. However, I'd say that >90% of current time_sum() usage is what this function does: add nsecs into a time stamp. Single call is simpler and more robust to use as there's no opportunity to mix up local vs global time stamps.

JannePeltonen (approved):

PR #1955, opened 2023-11-23
[PATCH v5] api: time: allow non-zero startup time

psavol:


include/odp/api/spec/time.h
```diff
@@ -21,20 +21,24 @@ extern "C" {
...
  * Current local time
  *
- * Returns current local time stamp value. The local time source provides high
- * resolution time, it is initialized to zero during ODP startup and will not
- * wrap around in at least 10 years.
- * Local time stamps are local to the calling thread and must not be shared
- * with other threads.
+ * Returns current CPU local time stamp value. The value can be used only by the calling thread,
+ * and only as long as it runs on the same CPU. Local time stamp values must not be shared with
+ * other threads.
```
JannePeltonen, line 34: Maybe this should just say that the time source of the local time stap value is specific to the calling thread and the current CPU and that time stamps of different time sources cannot be compared or otherwise mixed. But obeying that, the time stamp can be used in any thread (e.g. a thread can store and process time stamps collected by another thread).

include/odp/api/spec/time.h
```diff
@@ -21,20 +21,24 @@ extern "C" {
...
+ *
+ * Local time stamp value advances with a constant rate defined by odp_time_local_res(). The rate
```
JannePeltonen, line 36: Maybe it is too much promised that the time stamp value advances at a constant rate. Maybe it could be just said that the local time advances and is not affected by frequency scaling?

test/validation/api/time/time.c
```diff
@@ -276,12 +278,13 @@ static void time_test_monotony(void)
...
 	s_ns3 = odp_time_to_ns(ls_t3);
 
-	/* Time counting starts from zero. Assuming that the ODP instance has run
+	/* Time should not wrap in 10 years from ODP start. Assuming that the ODP instance has run
 	 * less than 10 minutes before running this test case. */
-	CU_ASSERT(ns1    < 10 * ODP_TIME_MIN_IN_NS);
-	CU_ASSERT(s_ns1  < 10 * ODP_TIME_MIN_IN_NS);
-	CU_ASSERT(l_ns1  < 10 * ODP_TIME_MIN_IN_NS);
-	CU_ASSERT(ls_ns1 < 10 * ODP_TIME_MIN_IN_NS);
+	limit = 10 * YEAR_IN_NS - 10 * ODP_TIME_MIN_IN_NS;
```
JannePeltonen, line 312: I would drop the 10 min leeway. It is highly unlikely that the 10 min adjustment makes a difference. If it does makes a difference, maybe we would rather have the test fail when it should pass instead of pass when it should fail.

Actually, it is not entirely clear if the 10 year time period mentioned in the API means exactly 10 * 365 * 24 * 3600 seconds or 10 calendar years which can include 2 leap days and some leap seconds.

psavol: Removed 10min margin in v2

JannePeltonen (approved):
A couple of the comment subject lines do not have a verb. See if you would like to tweak them.

psavol:
> A couple of the comment subject lines do not have a verb. See if you would like to tweak them.

Reworded those in v3.

harman-kalra:
Acked-by: Harman Kalra <hkalra@marvell.com>

PR #1958, opened 2023-11-28
[PATCH v6] api: timer: only inactive timers can be freed

psavol:
Remove possibility to free a timer that is running. When timer free call returns, the timer handle cannot be referenced any more. Expiration and event delivery of an already destroyed timer is an error prone corner case.

### DIFF
--- a/include/odp/api/spec/timer.h
+++ b/include/odp/api/spec/timer.h
@@ -246,19 +246,21 @@ odp_timer_t odp_timer_alloc(odp_timer_pool_t timer_pool, odp_queue_t queue, cons
 /**
  * Free a timer
  *
- * Free (destroy) a timer, reclaiming associated resources.
- * The timeout event for an active timer will be returned.
- * The timeout event for an expired timer will not be returned. It is the
- * responsibility of the application to handle this timeout when it is received.
+ * Frees a previously allocated timer. The timer must be inactive when calling this function.
+ * In other words, the application must cancel an active single shot timer (odp_timer_cancel())
+ * successfully or wait it to expire before freeing it. Similarly for an active periodic timer, the
+ * application must cancel it (odp_timer_periodic_cancel()) and receive the last event from
+ * the timer (odp_timer_periodic_ack()) before freeing it.
  *
- * A periodic timer must be cancelled successfully before freeing it.
+ * The call returns failure only on non-recoverable errors. Application must not use the timer
+ * handle anymore after the call, regardless of the return value.
  *
- * @param timer      Timer
+ * @param timer       Timer
  *
- * @return Event handle of timeout event
- * @retval ODP_EVENT_INVALID on failure
+ * @retval 0 on success
+ * @retval <0 on failure
  */
-odp_event_t odp_timer_free(odp_timer_t timer);
+int odp_timer_free(odp_timer_t timer);
 
 /**
  * Start a timer
--- a/platform/linux-generic/odp_timer.c
+++ b/platform/linux-generic/odp_timer.c
@@ -579,14 +579,20 @@ static odp_event_t timer_set_unused(timer_pool_t *tp, uint32_t idx)
 	return old_event;
 }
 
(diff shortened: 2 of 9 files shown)


include/odp/api/spec/timer.h
```diff
@@ -247,19 +247,20 @@ odp_timer_t odp_timer_alloc(odp_timer_pool_t timer_pool, odp_queue_t queue, cons
...
- * The timeout event for an expired timer will not be returned. It is the
- * responsibility of the application to handle this timeout when it is received.
+ * Frees a previously allocated timer. The timer must be inactive when calling this function.
+ * In other words, the application must cancel an active single shot timer (odp_timer_cancel())
+ * successfully or wait it to expire before freeing it. Similarly for a period timer, the
+ * application must cancel it (odp_timer_periodic_cancel()) and receive the last event from
+ * the timer (odp_timer_periodic_ack()) before freeing it.
  *
- * A periodic timer must be cancelled successfully before freeing it.
+ * In normal conditions, the call should always succeed to free an inactive timer.
```
JereLeppanen, line 256: Is this statement useful at all? Of course the caller expects this to succeed most of the time.

platform/linux-generic/odp_timer.c
```diff
@@ -581,14 +581,20 @@ static odp_event_t timer_set_unused(timer_pool_t *tp, uint32_t idx)
...
 	_odp_timer_t *tim = &tp->timers[idx];
 	tick_buf_t *tb = &tp->tick_buf[idx];
 
 	/* Free the timer by setting timer state to unused and
 	 * grab any timeout event */
 	odp_event_t old_event = timer_set_unused(tp, idx);
+	if (old_event != ODP_EVENT_INVALID) {
+		_ODP_ERR("Timer is active\n");
+		return -1;
+	}
```
JereLeppanen, line 597: Aren't we leaking an event here? Would it be better to not make any changes if the timer is active?

JereLeppanen (commented):
Some changes in the examples and tests don't appear to depend on the API change. Consider extracting those changes into separate commits.

psavol:
> Some changes in the examples and tests don't appear to depend on the API change. Consider extracting those changes into separate commits.

Timer simple has bug in both incrementing and decrementing return value on various errors. Buggy increments are on the same or near timer cancel/free lines. Otherwise, all changes should be related to timer inactivity synchronization.

include/odp/api/spec/timer.h
```diff
@@ -247,19 +247,20 @@ odp_timer_t odp_timer_alloc(odp_timer_pool_t timer_pool, odp_queue_t queue, cons
 /**
  * Free a timer
  *
- * Free (destroy) a timer, reclaiming associated resources.
- * The timeout event for an active timer will be returned.
- * The timeout event for an expired timer will not be returned. It is the
- * responsibility of the application to handle this timeout when it is received.
+ * Frees a previously allocated timer. The timer must be inactive when calling this function.
+ * In other words, the application must cancel an active single shot timer (odp_timer_cancel())
+ * successfully or wait it to expire before freeing it. Similarly for a period timer, the
```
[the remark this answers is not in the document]
psavol: Fixed in v2.

include/odp/api/spec/timer.h
psavol: It means that timer free should not return failure as "busy freeing timers, try again later". It should return failure only on bad handle, timer pool destroyed already, etc exceptional - where later retry won't help.

platform/linux-generic/odp_timer.c
psavol: Yes, but that should not never happen after this API change. We can simplify / rewrite (entire timer) implementation after this API change and removal of deprecated timer set functions are merged into repo.

include/odp/api/spec/timer.h
psavol: Updated the wording in v3.

JereLeppanen (approved):

psavol:
v5: rebased

PavanNikhilesh:

PR #1959, opened 2023-11-28
[PATCH v4] test: performance: improve warmup in common test bench

JannePeltonen:
Run a warmup round for each test just before the actual test run instead of running it for all tests before any actual test runs. This way other tests do not run between the warmup and the actual test run and the code becomes a bit cleaner too.

Test specific init functions are still run after the warmup, which may skew the results.

### DIFF
--- a/test/performance/bench_common.c
+++ b/test/performance/bench_common.c
@@ -51,36 +51,35 @@ int bench_run(void *arg)
 	bench_suite_t *suite = arg;
 	const uint64_t repeat_count = suite->repeat_count;
 	const odp_bool_t meas_time = suite->measure_time;
+	double result;
 
 	printf("\nAverage %s per function call\n", meas_time ? "time (nsec)" : "CPU cycles");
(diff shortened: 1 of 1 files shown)


JannePeltonen:
v2: made the same improvement in bench_tm_run() too.

test/performance/bench_common.c
```diff
@@ -51,36 +51,35 @@ int bench_run(void *arg)
...
+				continue;
 			}
+			bench_run_indef(&suite->bench[j], &suite->exit_worker);
+			return 0;
+		}
 
-			desc = bench->desc != NULL ? bench->desc : bench->name;
+		desc = bench->desc != NULL ? bench->desc : bench->name;
 
+		/* the zeroeth round is a warmup round that will be ignored */
```
MatiasElo, line 81: Capital letter for uniform style

MatiasElo (approved):
Looks good. I've no recollection why the original implementation was so complex 😅

PR #1960, opened 2023-11-28
[PATCH v2] example: timer_accuracy: fix CPU count default value

psavol:
According to help, test runs on single CPU by default.

### DIFF
--- a/example/timer/odp_timer_accuracy.c
+++ b/example/timer/odp_timer_accuracy.c
@@ -197,6 +197,7 @@ static int parse_options(int argc, char *argv[], test_opt_t *test_opt)
 
 	memset(test_opt, 0, sizeof(*test_opt));
 
+	test_opt->cpu_count = 1;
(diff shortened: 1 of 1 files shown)


example/timer/odp_timer_accuracy.c
```diff
@@ -197,6 +197,7 @@ static int parse_options(int argc, char *argv[], test_opt_t *test_opt)
 
 	memset(test_opt, 0, sizeof(*test_opt));
 
+	test_opt->cpu_count = 1;
```
JannePeltonen, line 200: This changes the behavior of the test in make check and in CI. Maybe odp_timer_accuracy_run.sh should now explicitly specify -c 0 option to retain the old behavior when the test is run through the script..

psavol: This -c option was just added in Oct. So, the test has been single threaded in CI for long time, and has now accidentally run on all threads for two months.

PR #1961, opened 2023-11-28
[PATCH v3] api: timer: deprecate timeout fresh

psavol:
The feature is not very useful to applications, since freshness result depends on timing of cancel/restart API calls vs the freshness check point.


MatiasElo (approved):

JereLeppanen (approved):

PavanNikhilesh:

PR #1964, opened 2023-12-05
[PATCH v8] fix issues reported by GCC address sanitizer

FanHong674:
Enable address sanitizer and fix bugs found by it.

### DIFF
--- a/.github/workflows/ci-pipeline.yml
+++ b/.github/workflows/ci-pipeline.yml
@@ -445,3 +445,21 @@ jobs:
                -e CONF="${CONF}" $CONTAINER_NAMESPACE/odp-ci-${OS}-${ARCH}-dpdk_21.11 /odp/scripts/ci/check.sh
       - if: ${{ failure() }}
         uses: ./.github/actions/run-failure-log
+
+  Run_sanitizer:
+    runs-on: ubuntu-20.04
+    env:
+      OS: ubuntu_22.04
+    strategy:
+      fail-fast: false
+      matrix:
+        flags: ['-fsanitize=address']
+    steps:
+      - uses: actions/checkout@v4
+      - run: sudo docker run -i -v `pwd`:/odp --privileged --shm-size 8g -e CC="${CC}" -e ARCH="${ARCH}"
+               -e CFLAGS="-O0 -g -Wno-error ${{matrix.flags}}"
+               -e CXXFLAGS="-O0 -g -Wno-error ${{matrix.flags}}"
+               -e LDFLAGS="-g ${{matrix.flags}}"
+               $CONTAINER_NAMESPACE/odp-ci-${OS}-${ARCH} /odp/scripts/ci/check.sh
+      - if: ${{ failure() }}
+        uses: ./.github/actions/run-failure-log
--- a/example/classifier/odp_classifier.c
+++ b/example/classifier/odp_classifier.c
@@ -783,6 +783,8 @@ int main(int argc, char *argv[])
 	if (odp_pool_destroy(pool))
 		ODPH_ERR("err: odp_pool_destroy error\n");
 
+	free(args->if_name);
+
 args_error:
 	odp_shm_free(shm);
 
@@ -1281,9 +1283,7 @@ static int parse_args(int argc, char *argv[], appl_args_t *appl_args)
 
 	if (ret) {
 		usage();
-
-		if (appl_args->if_name)
-			free(appl_args->if_name);
+		free(appl_args->if_name);
 	}
 
 	/* reset optind from the getopt lib */
--- a/example/ipsec_api/odp_ipsec.c
+++ b/example/ipsec_api/odp_ipsec.c
@@ -45,6 +45,7 @@
 #include <odp_ipsec_stream.h>
 #else
 static void init_stream_db(void) {}
+static void deinit_stream_db(void) {}
 static void resolve_stream_db(void) {}
 static int create_stream_db_inputs(void)
 {
@@ -1154,9 +1155,9 @@ main(int argc, char *argv[])
 	shm = odp_shm_lookup("shm_sp_db");
 	if (odp_shm_free(shm) != 0)
 		ODPH_ERR("Error: shm free shm_sp_db failed\n");
-	shm = odp_shm_lookup("stream_db");
-	if (odp_shm_free(shm) != 0)
-		ODPH_ERR("Error: shm free stream_db failed\n");
+
+	deinit_stream_db();
+
 	if (odp_shm_free(global->shm)) {
 		ODPH_ERR("Error: shm free global data failed\n");
 		exit(EXIT_FAILURE);
--- a/example/ipsec_crypto/odp_ipsec_stream.c
+++ b/example/ipsec_crypto/odp_ipsec_stream.c
@@ -40,13 +40,14 @@ typedef struct ODP_PACKED stream_pkt_hdr_s {
 	uint8_t    data[];  /**< Incrementing data stream */
 } stream_pkt_hdr_t;
 
+static const char *shm_name = "stream_db";
 stream_db_t *stream_db;
 
 void init_stream_db(void)
 {
 	odp_shm_t shm;
 
-	shm = odp_shm_reserve("stream_db",
+	shm = odp_shm_reserve(shm_name,
 			      sizeof(stream_db_t),
 			      ODP_CACHE_LINE_SIZE,
 			      0);
@@ -63,6 +64,28 @@ void init_stream_db(void)
 		exit(EXIT_FAILURE);
 	}
 	memset(stream_db, 0, sizeof(*stream_db));
+}
+
+void deinit_stream_db(void)
+{
+	stream_db_entry_t *stream = NULL;
+
+	for (stream = stream_db->list; NULL != stream; stream = stream->next) {
+		free(stream->input.intf);
+		free(stream->output.intf);
+	}
+
+	odp_shm_t shm = odp_shm_lookup(shm_name);
+
+	if (shm == ODP_SHM_INVALID) {
+		ODPH_ERR("Error: shared mem not found.\n");
+		exit(EXIT_FAILURE);
+	}
+
+	if (odp_shm_free(shm)) {
+		ODPH_ERR("Error: shared mem free failed.\n");
+		exit(EXIT_FAILURE);
+	}
 }
 
 int create_stream_db_entry(char *input)
--- a/helper/test/table.c
+++ b/helper/test/table.c
@@ -5,22 +5,6 @@
 #include <odp_api.h>
 #include <odp/helper/odph_api.h>
 
-/**
- * Address Resolution Protocol (ARP)
- * Description: Once a route has been identified for an IP packet (so the
- * output interface and the IP address of the next hop station are known),
- * the MAC address of the next hop station is needed in order to send this
- * packet onto the next leg of the journey towards its destination
- * (as identified by its destination IP address). The MAC address of the next
- * hop station becomes the destination MAC address of the outgoing
- * Ethernet frame.
- * Hash table name: ARP table
- * Number of keys: Thousands
- * Key format: The pair of (Output interface, Next Hop IP address),
- *        which is typically 5 bytes for IPv4 and 17 bytes for IPv6.
- * value (data): MAC address of the next hop station (6 bytes).
- */
-
 int main(int argc ODP_UNUSED, char *argv[] ODP_UNUSED)
 {
 	odp_instance_t instance;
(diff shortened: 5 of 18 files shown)


MatiasElo:
It would be good to add address sanitizer also to CI. The test could be added for example to `'--enable-deprecated --enable-helper-deprecated --enable-debug=full'` conf in `Run` job.

helper/test/table.c
```diff
@@ -32,10 +32,10 @@ int main(int argc ODP_UNUSED, char *argv[] ODP_UNUSED)
...
 	char ip_addr2[] = "11223344";
 	char ip_addr3[] = "55667788";
-	char mac_addr1[] = "0A1122334401";
-	char mac_addr2[] = "0A1122334402";
-	char mac_addr3[] = "0B4433221101";
-	char mac_addr4[] = "0B4433221102";
+	char mac_addr1[16] = "0A1122334401";
+	char mac_addr2[16] = "0A1122334402";
+	char mac_addr3[16] = "0B4433221101";
+	char mac_addr4[16] = "0B4433221102";
```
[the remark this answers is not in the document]
FanHong674: The key and value in this test do seem weird. In the test, the keys are the addresses of arrays (`&ip_addr`) and the values are the contents of arrays (`mac_addr`). See `test_ops->f_put()`.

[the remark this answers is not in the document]
MatiasElo: Maybe it makes the test more clear if ODPH_IPV4ADDR_LEN is used as key_size and ODPH_ETHADDR_LEN as value size. The ip_addr and mac_addr arrays have to be shortened to match this.

example/ipsec_crypto/odp_ipsec_stream.c
```diff
@@ -65,6 +65,14 @@ void init_stream_db(void)
 	memset(stream_db, 0, sizeof(*stream_db));
 }
 
+void free_stream_db_intf(void)
```
JannePeltonen, line 68: The code would be cleaner if this was encapsulated in something like deinit_stream_db(), which would be the opposite of init_stream_db(). That function should then take care of freeing the SHM too.

example/ipsec_crypto/odp_ipsec_stream.c
```diff
@@ -65,6 +65,14 @@ void init_stream_db(void)
...
+{
+	for (int i = 0; i < MAX_DB; i++) {
+		free((void *)(intptr_t)stream_db->array[i].input.intf);
```
JannePeltonen, line 71: The casts are very ugly and should be avoided. I suppose you use them to get rid of the constness of the pointer, but I think it would be better to just remove the const specifier from the type.

example/ipsec_crypto/odp_ipsec_stream.c
```diff
@@ -65,6 +65,14 @@ void init_stream_db(void)
 	memset(stream_db, 0, sizeof(*stream_db));
 }
 
+void free_stream_db_intf(void)
+{
+	for (int i = 0; i < MAX_DB; i++) {
```
JannePeltonen, line 70: Here you could just loop through strream_db->list to only go through the entries that actually exist.

example/ipsec_api/odp_ipsec.c
```diff
@@ -45,6 +45,7 @@
 #include <odp_ipsec_stream.h>
```
JannePeltonen, line 45: The commit comment should be improved. For example:

```
example: ipsec: free interface names at exit

Free memory allocated for interface names at program termination to
avoid memory leak complaints by dynamic analyzers.
```

JannePeltonen (changes requested):

configure.ac
```diff
@@ -233,7 +233,20 @@ if test "x$abi_compat" = "xno" ; then
 	fi
 fi
 AM_CONDITIONAL(ODP_ABI_COMPAT, [test "x$ODP_ABI_COMPAT" = "x1"])
-
+##########################################################################
+# Enable/disable AddressSanitizer
+##########################################################################
+asan_enabled=no
+AC_ARG_ENABLE([asan],
+             [AS_HELP_STRING([--enable-asan],
```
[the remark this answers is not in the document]
MatiasElo: After discussion we agreed that it's better to not add new configure option and instead set the required flags directly in the CI job.  The `CHECK_GLOBALS_REGEX` changes are likely still required to pass `make check`.

JereLeppanen:
I took the liberty of updating this PR with all the comments processed, commit comments cleaned up a bit, and a couple of new warnings fixed (ML).

JereLeppanen:
v4:
- Fix buffer overrun in pktio/dpdk.

JereLeppanen:
v4:
- Rebase.
- Fix typo in a commit comment.

test/validation/api/traffic_mngr/traffic_mngr.c
```diff
@@ -65,7 +65,8 @@
 
 #define MAX_PKTS                 1000u
 #define PKT_BUF_SIZE             1460
-#define MAX_PAYLOAD              1400
+#define MIN_HDR_LEN              (ODPH_ETHHDR_LEN + ODPH_UDPHDR_LEN + ODPH_IPV4HDR_LEN)
+#define MAX_PAYLOAD              (PKT_BUF_SIZE - MIN_HDR_LEN)
```
JannePeltonen, line 69: Since the code is a bit messy with respect to the packet length and contains hidden assumptions, it might be good to check payload_len against MAX_PAYLOAD in make_pkt().

example/classifier/odp_classifier.c
```diff
@@ -784,6 +784,8 @@ int main(int argc, char *argv[])
 		ODPH_ERR("err: odp_pool_destroy error\n");
 
 args_error:
+	if (args->if_name)
+		free(args->if_name);
```
JannePeltonen, line 788: If parse_args() fails, it already frees args->ifname, so in that case this introduces a double free.

platform/linux-generic/pktio/pcap.c
```diff
@@ -188,15 +188,18 @@ static int pcapif_promisc_mode_set(pktio_entry_t *pktio_entry,
 
 	if (pcap_compile(pcap->rx, &bpf, filter_exp,
 			 0, PCAP_NETMASK_UNKNOWN) != 0) {
+		pcap_freecode(&bpf);
```
JannePeltonen, line 191: When pcap_compile fails, bpf is not written and pcap_freecode(&bpf) should not be called.

example/ipsec_crypto/odp_ipsec_stream.c
```diff
@@ -65,6 +66,23 @@ void init_stream_db(void)
...
+void deinit_stream_db(void)
+{
+	stream_db_entry_t *stream = NULL;
+
+	for (stream = stream_db->list; NULL != stream; stream = stream->next) {
+		free(stream->input.intf);
+		free(stream->output.intf);
+	}
+
+	odp_shm_t shm = odp_shm_lookup(shm_name);
```
JannePeltonen, line 78: This now fails to free the shm despite figuring out the handle.

helper/test/table.c
```diff
@@ -32,10 +32,10 @@ int main(int argc ODP_UNUSED, char *argv[] ODP_UNUSED)
 	char ip_addr1[] = "12345678";
 	char ip_addr2[] = "11223344";
 	char ip_addr3[] = "55667788";
-	char mac_addr1[] = "0A1122334401";
-	char mac_addr2[] = "0A1122334402";
-	char mac_addr3[] = "0B4433221101";
-	char mac_addr4[] = "0B4433221102";
+	char mac_addr1[] = { 30, 31, 32, 33, 34, 35, };
```
JannePeltonen, line 35: Now the values are no longer null terminated printable strings but there is still printing code that assumes they are.

helper/test/table.c
```diff
@@ -29,9 +29,9 @@ int main(int argc ODP_UNUSED, char *argv[] ODP_UNUSED)
 	odph_table_t tmp_tbl;
 	odph_table_ops_t *test_ops;
 	char tmp[32];
-	char ip_addr1[] = "12345678";
-	char ip_addr2[] = "11223344";
-	char ip_addr3[] = "55667788";
+	char key1[] = { 0, 10, 11, 12, 13, };
```
JannePeltonen, line 32: I am not sure the ARP comment really adds much value and would not mind it being removed. Then you could also rename macaddr to value.

.github/workflows/ci-pipeline.yml
```diff
@@ -445,3 +445,21 @@ jobs:
...
+    env:
+      OS: ubuntu_22.04
+    strategy:
+      fail-fast: false
+      matrix:
+        flags: ['-fsanitize=address']
+    steps:
+      - uses: actions/checkout@v4
+      - run: sudo docker run -i -v `pwd`:/odp --privileged --shm-size 8g -e CC="${CC}" -e ARCH="${ARCH}"
+               -e CFLAGS="-O0 -g -Wno-error ${{matrix.flags}}"
```
JannePeltonen, line 460: Why the -Wno-error?

JannePeltonen (changes requested):

.github/workflows/ci-pipeline.yml
JereLeppanen: GCC man page: "Note that sanitizers tend to increase the rate of false positive warnings, most notably those around -Wmaybe-uninitialized.  We recommend against combining -Werror and [the use of] sanitizers."

JereLeppanen:
All comments processed in v6.

helper/test/table.c
```diff
@@ -13,9 +13,9 @@ int main(int argc ODP_UNUSED, char *argv[] ODP_UNUSED)
 	odph_table_t tmp_tbl;
 	odph_table_ops_t *test_ops;
 	char tmp[32];
-	char ip_addr1[] = "12345678";
-	char ip_addr2[] = "11223344";
-	char ip_addr3[] = "55667788";
+	char key1[] = "1234";
```
JannePeltonen, line 16: The commit message is now out-of-date since the code no longer contains any comment about key format. I think the real reason for this commit is to fix the incorrect key length passed to the table creation function.

example/classifier/odp_classifier.c
```diff
@@ -783,6 +783,9 @@ int main(int argc, char *argv[])
 	if (odp_pool_destroy(pool))
 		ODPH_ERR("err: odp_pool_destroy error\n");
 
+	if (args->if_name)
+		free(args->if_name);
```
JannePeltonen, line 787: A very minor thing and a matter of taste, I guess, but the 'if' is redundant since free(0) is a no-op.

test/performance/odp_l2fwd_run.sh
```diff
@@ -71,17 +71,10 @@ run_l2fwd()
...
-	# output while piping to tee, as long as stdbuf is available.
-	if [ "$(which stdbuf)" != "" ]; then
-		STDBUF="stdbuf -o 0"
-	else
-		STDBUF=
-	fi
 	LOG=odp_l2fwd_tmp.log
 
 	# Max 2 workers
-	$STDBUF odp_l2fwd${EXEEXT} -i $IF1,$IF2 -m 0 -t 5 -c 2 | tee $LOG
```
JannePeltonen, line 84: The commit message says: "Using stdbuf is only necessary when something in the output is being waited on, with e.g. tail -f. That's not the case here, so remove the unnecessary usage."

I am not sure the use of stdbuf here is unnecessary. When running the script in an interactive shell, the stfbuf causes the pps printouts to appear as soons as available, not after the scrip has run. So there is some benefit there. I am not sure if it outweighs the trouble though, so I am not sure what to think of this commit.

helper/test/table.c
JereLeppanen: Currently the key length passed to f_create() is smaller than the array, but I wouldn't necessarily call that "incorrect".

This commit doesn't fix any functional problem, and is in that sense totally optional. The real reason for it is only to hopefully avoid some confusion.

I'll add the same in some form in the commit comment, which is otherwise correct, as far as I can tell, it's just missing the reason why the commit was made.

test/performance/odp_l2fwd_run.sh
JereLeppanen: In some test scripts the script starts a task to the background and then waits for something in the background task's output. In those cases stdbuf is necessary for the script to work, but that is not the case here, hence "unnecessary". But anyway, commit comment reworded in v7.

JereLeppanen:
All comments processed in v7.

helper/test/table.c
JannePeltonen: The code was incorrect (or obfuscated, if you prefer) in the sense that if the 8-byte ipaddr strings did not differ from each other in the first 4 bytes, things did not work as intended. This commit fixes that.

JannePeltonen (approved):

PR #1965, opened 2023-12-07
[PATCH v4] test: dma_perf: add memory copy and sparse segment type options

TuomasTaipale:
This patchset introduces a new SW packet copy transfer mode. This is useful for comparing the performance of DMA subsystem and copying on CPU. Additionally, new sparse segment type variants are introduced: 'sparse packet' and 'sparse memory'. These configure scenarios which might be a more realistic usecase for some applications due to potentially incurred cache misses and allocation overheads. The old existing segment type options are renamed to 'dense packet' and 'dense memory'.

v2:
- Simplified memory copy transfer with transfer being done now only with a single function instead of separate for packets and addresses

v3:
- Matias' comments

v4:
- Fixed an issue when preparing transfers for poll mode
- Dropped ticketlock return value check
- Added reviewed-by tags

### DIFF
--- a/test/performance/odp_dma_perf.c
+++ b/test/performance/odp_dma_perf.c
@@ -28,8 +28,9 @@
 #define PROG_NAME "odp_dma_perf"
 
 enum {
-	SYNC = 0U,
-	ASYNC
+	SYNC_DMA = 0U,
+	ASYNC_DMA,
+	SW_COPY
 };
 
 enum {
@@ -47,7 +48,7 @@ enum {
 	MANY
 };
 
-#define DEF_TRS_TYPE SYNC
+#define DEF_TRS_TYPE SYNC_DMA
 #define DEF_SEG_CNT 1U
 #define DEF_LEN 1024U
 #define DEF_SEG_TYPE PACKET
@@ -264,8 +265,9 @@ static void print_usage(void)
 	       "\n"
 	       "  -t, --trs_type      Transfer type for test data. %u by default.\n"
 	       "                      Types:\n"
-	       "                          0: synchronous\n"
-	       "                          1: asynchronous\n"
+	       "                          0: synchronous DMA\n"
+	       "                          1: asynchronous DMA\n"
+	       "                          2: SW memory copy\n"
 	       "  -i, --num_in_seg    Number of input segments to transfer. 0 means the maximum\n"
 	       "                      count supported by the implementation. %u by default.\n"
 	       "  -o, --num_out_seg   Number of output segments to transfer to. 0 means the\n"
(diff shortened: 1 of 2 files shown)


test/performance/odp_dma_perf.c
```diff
@@ -591,6 +628,7 @@ static odp_pool_t get_src_packet_pool(void)
 	param.pkt.num = prog_conf->num_inflight * prog_conf->num_in_segs * prog_conf->num_sessions;
 	param.pkt.len = prog_conf->src_seg_len;
 	param.pkt.seg_len = prog_conf->src_seg_len;
+	param.pkt.cache_size = 0U;
```
MatiasElo, line 631: Setting cache size to zero will cause global pool access per alloc/free, which may skew results and is probably not the most realistic use case. So it's better to set cache size to a reasonable value (or to add a new command line parameter).

test/performance/odp_dma_perf.c
```diff
@@ -854,7 +854,7 @@ static void run_transfers_mt_safe(sd_t *sd, stats_t *stats)
 	for (uint32_t i = 0U; i < count; ++i) {
 		info = &infos[i];
 
-		if (odp_ticketlock_trylock(&info->lock)) {
+		if (odp_ticketlock_trylock(&info->lock) == 1) {
```
JannePeltonen, line 900: I think this patch could be dropped. The code already works according to the ODP spec which says that the trylock function returns 0 or 1. So this change does not fix anything, contrary to what the commit title says. Only in the case the trylock returns something outside the spec this might change the behaviour by I do not see why this would be more correct than e.g. checking that the return value is not zero. Asserting that the return value is either 0 or 1 would make sense in a validation test, but a perf test can pretty much assume that ODP functions do what they say. So I think the existing code is best here.

TuomasTaipale: I would agree if OK case was specified as ">0" or return type was bool, however the type is signed integer and OK case is exactly 1. IMO, also "application" code from ODP repo should (try) to display exemplary usage of the API.

test/performance/odp_dma_perf.c
```diff
@@ -1,8 +1,14 @@
...
+
+/**
+ * DMA performance tester
  *
- * SPDX-License-Identifier:     BSD-3-Clause
+ * This tester application can be used to profile the performance of an ODP DMA implementation.
+ * Tester workflow is simple and consists of issuing as many back-to-back DMA transfers as the
+ * implementation allows and then recording key performance statistics (such as function overhead,
+ * latencies etc.).
  */
```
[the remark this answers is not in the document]
TuomasTaipale: This needs more time and testing, better do this in another PR.

test/performance/odp_dma_perf.c
```diff
@@ -1133,21 +1152,88 @@ static void drain_compl_events(ODP_UNUSED sd_t *sd)
...
+	start_tm = odp_time_local_strict_ns();
+	start_cc = odp_cpu_cycles();
+
+	while (tot_len > 0U) {
+		if (is_addr) {
+			src_data = param->src_seg[i].addr;
+			dst_data = param->dst_seg[j].addr;
+		} else {
+			src_data = odp_packet_data(param->src_seg[i].packet);
+			dst_data = odp_packet_data(param->dst_seg[j].packet);
```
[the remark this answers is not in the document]
TuomasTaipale: V3.

test/performance/odp_dma_perf.c
TuomasTaipale: Cache sizes improved in v3.

MatiasElo (approved):

test/performance/odp_dma_perf.c
TuomasTaipale: Democratically dropped in v4.

PR #1968, opened 2023-12-08
[PATCH v3] timer: tick does not need to start from zero

psavol:
Optimize timer current tick and other functions by using nsec time directly as the tick. API does not require tick to start from zero.

### DIFF
--- a/platform/linux-generic/odp_timer.c
+++ b/platform/linux-generic/odp_timer.c
@@ -1451,13 +1451,6 @@ void odp_timer_pool_destroy(odp_timer_pool_t tpid)
 	odp_timer_pool_del(timer_pool_from_hdl(tpid));
 }
 
-uint64_t odp_timer_current_tick(odp_timer_pool_t tpid)
-{
-	(void)tpid;
-
-	return odp_time_global_ns();
-}
-
 int odp_timer_sample_ticks(odp_timer_pool_t timer_pool[], uint64_t tick[], uint64_t clk_count[],
 			   int num)
 {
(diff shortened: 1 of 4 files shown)


test/performance/odp_timer_perf.c
MatiasElo, line 1: Copyright can be updated.

platform/linux-generic/odp_timer.c
MatiasElo, line 1: Copyright can be updated.

platform/linux-generic/odp_timer.c
```diff
@@ -1471,16 +1453,15 @@ void odp_timer_pool_destroy(odp_timer_pool_t tpid)
 
 uint64_t odp_timer_current_tick(odp_timer_pool_t tpid)
 {
-	timer_pool_t *tp = timer_pool_from_hdl(tpid);
+	(void)tpid;
```
MatiasElo: `ODP_UNUSED` would save one line.

MatiasElo (approved):

platform/linux-generic/odp_timer.c
psavol: Cast to void is used with other inline functions. This function moves into inline header in the next patch.

test/performance/odp_timer_perf.c
psavol: Fixed in v3

platform/linux-generic/odp_timer.c
psavol: Fixed in v3

PR #1971, opened 2023-12-11
github_ci: split run and build jobs into gcc and clang specific jobs

MatiasElo:


JannePeltonen (approved):
The removal of the bogus CXX flag (which is either overridden by the build script or unintentionally(?) changes the g++ version to be used) could be mentioned in a commit message.

MatiasElo:
V2:
- Mention removing `CXX=g++-10` flag in commit log

PR #1972, opened 2023-12-13
[PATCH v4] test: ipsecfwd: replace helper IP lookup table

TuomasTaipale:
Replace IP lookup table provided by ODP helper library with a simpler structure.

v2:
- Fixed potential undefined behavior with bitshifting

v3:
- Removed sorting of forwarding entries

v4:
- Added reviewed-by tag

### DIFF
--- a/test/performance/odp_ipsecfwd.c
+++ b/test/performance/odp_ipsecfwd.c
@@ -33,6 +33,7 @@
 #define PKT_CNT 32768U
 #define MAX_BURST 32U
 #define ORDERED 0U
+#define IP_ADDR_LEN 32U
 
 #define ALG_ENTRY(_alg_name, _type) \
 	{ \
@@ -77,10 +78,16 @@ typedef struct pktio_s {
 } pktio_t;
 
 typedef struct {
+	uint32_t prefix;
+	uint32_t mask;
 	odph_ethaddr_t dst_mac;
 	const pktio_t *pktio;
-	odph_iplookup_prefix_t prefix;
 } fwd_entry_t;
+
+typedef struct {
+	fwd_entry_t entries[MAX_FWDS];
+	uint32_t num;
+} lookup_table_t;
 
 typedef struct {
 	uint64_t ipsec_in_pkts;
@@ -113,7 +120,7 @@ typedef struct {
 } sa_config_t;
 
 typedef uint32_t (*rx_fn_t)(thread_config_t *config, odp_event_t evs[], int num);
-typedef void (*ipsec_fn_t)(odp_packet_t pkts[], int num, odph_table_t fwd_tbl, stats_t *stats);
+typedef void (*ipsec_fn_t)(odp_packet_t pkts[], int num, lookup_table_t *fwd_tbl, stats_t *stats);
 typedef void (*drain_fn_t)(prog_config_t *config);
 
 typedef struct {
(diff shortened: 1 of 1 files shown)


test/performance/odp_ipsecfwd.c
```diff
@@ -1339,13 +1360,13 @@ static void create_fwd_table_entry(config_setting_t *cfg, prog_config_t *config)
 		return;
 	}
 
-	entry = &config->fwd_entries[config->num_fwds];
+	mask = 0xFFFFFFFF << (IP_ADDR_LEN - mask);
```
JannePeltonen, line 1363: If mask length (it might be good to separate mask length variable from mask, for clarity) is 0, this triggers undefined behaviour, so that case probably needs to be special-cased.

test/performance/odp_ipsecfwd.c
```diff
@@ -1710,28 +1731,18 @@ static odp_bool_t setup_pktios(prog_config_t *config)
...
-
-	for (uint32_t i = 0U; i < config->num_fwds; ++i) {
-		fwd_e = &config->fwd_entries[i];
-
-		if (odph_iplookup_table_put_value(config->fwd_tbl, &fwd_e->prefix, &fwd_e) < 0) {
-			ODPH_ERR("Error populating forwarding table\n");
-			return false;
-		}
-	}
+	return val < 0 ? -1 : val == 0 ? 0 : 1;
```
JannePeltonen, line 1740: I wonder what is the rationale for sorting the entries by prefix. To get more repeatable test results regarless of the order of the entries in the config file?

It would be good to sort first by mask length so that longest prefices appear first in the lookup array. This way the expected longest-prefix match works. The prefix value could be a secondary sort key if you think it is useful.

JannePeltonen (changes requested):
This is a good change. A couple of minor comments though.

test/performance/odp_ipsecfwd.c
TuomasTaipale: I keep forgetting the large enough bitshift resulting in undefined behavior instead of promotion. Will fix.

test/performance/odp_ipsecfwd.c
TuomasTaipale: > longest prefices appear first

They do now, after the sorting, which was the idea.

JannePeltonen: No, longest prefixes do not appear first. The comparison function of the sort is not even looking at prefix length.

TuomasTaipale: Not sure if we're talking about the same thing.

E.g. passing:
```
fwd: (
        {
        prefix: "192.168.134.0/0"
        if: "ens9f0"
        dst_mac: "00:00:05:00:07:00"
        },
	{
	prefix: "192.168.1.0/24"
	if: "ens9f0"
	dst_mac: "00:00:05:00:07:00"
	},
	{
	prefix: "192.168.1.11/32"
	if: "ens9f0"
	dst_mac: "00:00:05:00:07:00"
	},
	{
	prefix: "192.168.134.0/16"
	if: "ens9f0"
	dst_mac: "00:00:05:00:07:00"
	},
	{
	prefix: "192.168.134.0/8"
	if: "ens9f0"
	dst_mac: "00:00:05:00:07:00"
	}
);
```
will result in this order in the table:
```
192.168.1.11
192.168.1.0
192.168.0.0
192.0.0.0
0.0.0.0
```
which is probably enough.

test/performance/odp_ipsecfwd.c
TuomasTaipale: V2.

test/performance/odp_ipsecfwd.c
JannePeltonen: Think about e.g. 10.0.0.0/8 and 10.0.0.0/16. Or add 192.168.1.0/25 in your list of routes.

TuomasTaipale: Removed sorting in v3, entries are looked up in the config order.

JannePeltonen (approved):

PR #1975, opened 2023-12-14
[PATCH v5] validation: timer: run tests with all supported clock sources

JereLeppanen:
Run most of the tests with all clock sources that are supported according to capability.

### DIFF
--- a/test/validation/api/timer/timer.c
+++ b/test/validation/api/timer/timer.c
@@ -65,14 +65,8 @@ struct test_timer {
 };
 
 typedef struct {
-	/* Clock source support flags */
-	uint8_t clk_supported[ODP_CLOCK_NUM_SRC];
-
-	/* Periodic timer support per clock source*/
-	uint8_t periodic_support[ODP_CLOCK_NUM_SRC];
-
-	/* Periodic timers not supported with any clock source */
-	int no_periodic;
+	/* Periodic timer support */
+	int periodic_support;
 
 	/* Default resolution / timeout parameters */
 	struct {
@@ -107,19 +101,21 @@ typedef struct {
 
 } global_shared_mem_t;
 
-static global_shared_mem_t *global_mem;
+typedef struct {
+	odp_timer_clk_src_t clk_src;
+	global_shared_mem_t global_mem;
 
-static int timer_global_init(odp_instance_t *inst)
+} test_global_t;
+
+static global_shared_mem_t *global_mem;
+static test_global_t *test_global;
+static odp_shm_t global_shm;
+static odp_instance_t inst;
+
+static int global_init(void)
 {
-	odp_shm_t global_shm;
 	odp_init_t init_param;
 	odph_helper_options_t helper_options;
-	odp_timer_capability_t capa;
-	odp_timer_res_capability_t res_capa;
-	uint64_t res_ns, min_tmo, max_tmo;
-	unsigned int range;
-	int i;
-	int num_periodic = 0;
 
 	if (odph_options(&helper_options)) {
 		ODPH_ERR("odph_options() failed\n");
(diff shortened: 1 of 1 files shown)


test/validation/api/timer/timer.c
```diff
@@ -3270,13 +3233,39 @@ int main(int argc, char *argv[])
...
+		goto fail;
 
-	if (ret == 0)
-		ret = odp_cunit_run();
+	if (odp_cunit_run())
+		goto fail;
+
+	for (clk_src = ODP_CLOCK_SRC_0; clk_src < ODP_CLOCK_NUM_SRC; clk_src++) {
+		odp_timer_capability_t capa;
+
```
psavol, line 3250: Since we are running entire suite multiple times, a print with clock source ID would help to read the log when something failes.

test/validation/api/timer/timer.c
```diff
@@ -107,18 +101,14 @@ typedef struct {
 } global_shared_mem_t;
 
 static global_shared_mem_t *global_mem;
+static odp_instance_t inst;
+static odp_timer_clk_src_t clk_src;
```
psavol, line 105: I'd put also these into SHM allocated shared mem. Those could locate outside of global_shared_mem_t, which is memset to zero between test suite rounds.

```
typedef struct {
     odp_instance_t inst;
     odp_timer_clk_src_t clk_src;

    global_shared_mem_t global_mem;
    
} test_global_t;
```

JereLeppanen: Why though?

test/validation/api/timer/timer.c
```diff
@@ -145,18 +135,62 @@ static int timer_global_init(odp_instance_t *inst)
...
 
+	return 0;
+}
+
+static int global_term(void)
+{
+	odp_shm_t shm;
+
+	shm = odp_shm_lookup(GLOBAL_SHM_NAME);
+	if (0 != odp_shm_free(shm)) {
```
psavol, line 149: SHM handle could be saved into global mem and pass pointer here to avoid lookup.

test/validation/api/timer/timer.c
```diff
@@ -253,29 +257,29 @@ check_plain_queue_support(void)
...
 {
-	if (global_mem->periodic_support[0] && global_mem->param.queue_type_plain)
+	if (global_mem->periodic_support && global_mem->param.queue_type_plain)
 		return ODP_TEST_ACTIVE;
 
 	return ODP_TEST_INACTIVE;
 }
 
-static void timer_test_capa_run(odp_timer_clk_src_t clk_src)
+static void timer_test_capa(void)
```
psavol, line 282: It easier to read the code when all globals are passed though the same global mem variable. It this case:

clk_src =globals->clk_src;

test/validation/api/timer/timer.c
psavol: When reading the code, it's easier to maintain the context between local vs global variables. Also, "all globals through one struct"-pattern is used in other applications as well to tackle process vs thread mode difference in globals.

JereLeppanen: This is now a little different from other tests anyway, and clk_src is special since it's our "suite loop" variable. Regardless, I put it in shm in v2. And added some 15 lines of `odp_timer_clk_src_t clk_src = test_global->clk_src;`, although I don't think it makes the code more readable.

I did not put inst in shm, since inst has a slightly longer lifetime than shm. Other validation tests also don't do that.

JereLeppanen:
v2: All review comments (except inst).

test/validation/api/timer/timer.c
```diff
@@ -3266,17 +3242,52 @@ odp_suiteinfo_t timer_suites[] = {
...
+		goto fail;
+
+	if (odp_cunit_run())
+		goto fail;
+
+	for (int i = ODP_CLOCK_SRC_0; i < ODP_CLOCK_NUM_SRC; i++) {
+		odp_timer_capability_t capa;
+
+		if (odp_timer_capability(i, &capa))
+			continue;
```
psavol, line 3267: Clock source 0 (=default) must be always supported. Is there a check for that, or should this CU_FAIL if capa claims that 0 is not supported?

JereLeppanen: There's no specific test for that. timer_pool_sample_ticks fails if default source is not supported. I'll add a specific test so that it's clearer.

JereLeppanen:
v3: Add commit "validation: timer: prevent timer_test_max_tmo_min_tmo_sched from taking too long".

JereLeppanen:
v4:
- Add a couple of missing clk_src assigns in pool parameters.
- Add commit "validation: timer: test that default clock source is supported".
- Review tags.

test/validation/api/timer/timer.c
```diff
@@ -376,8 +377,9 @@ static void timer_test_capa_allsrc(void)
 	CU_ASSERT_FATAL(ODP_CLOCK_EXT == ODP_CLOCK_SRC_1);
 #endif
 
+	CU_ASSERT(odp_timer_capability(ODP_CLOCK_DEFAULT, &capa) == 0);
```
psavol, line 380: Please merge this patch to the first patch, since that modifies overall how different clock sources are tested.

JereLeppanen: In v5.

psavol (approved):

PR #1976, opened 2023-12-18
[PATCH v3] example: improve example application documentations

MatiasElo:


example/time/time_global_test.c
```diff
@@ -4,6 +4,14 @@
  * SPDX-License-Identifier:     BSD-3-Clause
  */
 
+/**
+ * @example time_global_test.c
```
TuomasTaipale, line 8: Is this visible if you generate it locally?

test/performance/odp_l2fwd.c
```diff
@@ -6,6 +6,14 @@
  * SPDX-License-Identifier:     BSD-3-Clause
```
TuomasTaipale, line 6: Unrelated, but should we ditch the "example" l2fwd which is just a symbolic link and just have the "performance" l2fwd?

example/time/time_global_test.c
MatiasElo: Good catch, I was wondering the same thing at some point but completely forgot it. Perhaps the cleanest fix is to rename the application to e.g. odp_time_global_test, so it matches Doxyfile FILE_PATTERNS.

test/performance/odp_l2fwd.c
MatiasElo: Agreed, I updated the card regarding example application clean up.

example/time/time_global_test.c
MatiasElo: Renamed to `odp_time_global_test.c ` in V2.

TuomasTaipale (approved):

PR #1988, opened 2024-01-04
[PATCH v3] example: debug: pktio link info

psavol:
Print pktio link info and enable pktio interface name as option.

### DIFF
--- a/example/debug/odp_debug.c
+++ b/example/debug/odp_debug.c
@@ -325,6 +325,8 @@ static int pktio_debug(void)
 	odp_pool_t pool;
 	odp_pool_param_t pool_param;
 	odp_pktio_t pktio;
+	odp_pktio_link_info_t info;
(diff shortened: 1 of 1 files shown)


example/debug/odp_debug.c
```diff
@@ -615,6 +630,7 @@ int main(int argc, char *argv[])
 		global->ipsec   = 1;
 		global->timer   = 1;
 		global->stash   = 1;
+		strcpy(global->pktio_name, "loop");
```
JereLeppanen, line 633: If there are some arguments, but not the -i argument, pktio_name will be a zero-length string. Shouldn't this strcpy be done unconditionally after the memset?

psavol: Packet IO is not used if -i is not listed as argument. Exception is this default case when no arguments is listed.

JereLeppanen (approved):

PR #1991, opened 2024-01-05
validation: ipsec & init: combine a couple of binaries

JannePeltonen:
Combine a few test binaries to make the code a bit easier to follow (in the ipsec validation case) and to reduce LTO build time slightly. Maybe I should have also moved the related test definitions from validation/api/Makefile.am to init/Makefile.am and ipsec/Makefile.am, but I did not do it (maybe the point to have them at the upper level is to get clearer (?) make check output).

Fix also odp_cunit_parse_options() so that it updates argc after deleting parameters from argv.


MatiasElo (approved):
Looks good to me. We may want to verify that this change doesn't cause any application CI issues before merge.

PR #1992, opened 2024-01-10
[PATCH v2] linux-gen: event: move subtype field

psavol:
Move event subtype field from packet header to event header


platform/linux-generic/include/odp_packet_internal.h
```diff
@@ -107,9 +107,6 @@ typedef struct ODP_ALIGNED_CACHE odp_packet_hdr_t {
 
 	uint16_t tailroom;
 
-	/* Event subtype */
-	int8_t   subtype;
-
```
MatiasElo, line 112: This gap could now be used to store for example `cos` field, which is commonly used in the fast path.

MatiasElo (approved):

PR #1997, opened 2024-01-16
[PATCH v2] validation: sched: run tests only on worker threads

MatiasElo:
There is no guarantee that the main thread (ODP_THREAD_CONTROL) won't end up running on the same core with a worker thread. To avoid test slowdowns run multithreaded scheduler tests only on worker threads.


JannePeltonen (approved):
You could mention in the commit message that the scheduler_test_order_wait_2_threads test still runs also on a control thread (which I think is ok, see also PR 2001).

PR #1998, opened 2024-01-16
[PATCH v3] linux-gen: time: improve TSC frequency measurement

JannePeltonen:
Speed up TSC frequency measurement and thus ODP startup by reducing the
    measurement duration. Make the result more accurate despite the decreased
    duration by using a busyloop instead of nanosleep and by taking the median
    of several measurements.

    Using a busyloop instead of nanosleep also fixes the measurement in (old)
    systems that have constant but not invariant TSC which may stop or change
    frequency in C-states.

### DIFF
--- a/platform/linux-generic/arch/x86/odp_time_cpu.c
+++ b/platform/linux-generic/arch/x86/odp_time_cpu.c
@@ -1,7 +1,6 @@
-/* Copyright (c) 2015-2018, Linaro Limited
- * All rights reserved.
- *
- * SPDX-License-Identifier:     BSD-3-Clause
+/* SPDX-License-Identifier: BSD-3-Clause
+ * Copyright (c) 2015-2018 Linaro Limited
+ * Copyright (c) 2024 Nokia
  */
 
 #include <odp_posix_extensions.h>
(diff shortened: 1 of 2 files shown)


platform/linux-generic/arch/x86/odp_time_cpu.c
```diff
@@ -15,24 +14,67 @@
 
 #include <time.h>
 
-/* Measure TSC frequency. Frequency information registers are defined for x86,
- * but those are often not enumerated. */
+static int nsleep(uint64_t nsec)
```
TuomasTaipale, line 17: Maybe this could be `nwait` or something similar as no actual sleeping is done in the function.

platform/linux-generic/arch/x86/odp_time_cpu.c
```diff
@@ -15,24 +14,67 @@
...
+{
+	struct timespec ts1, ts2;
+	uint64_t diff;
+
+	if (clock_gettime(CLOCK_MONOTONIC_RAW, &ts1)) {
+		_ODP_ERR("clock_gettime() failed\n");
```
TuomasTaipale, line 23: Could be useful to have the actual error printed as well, e.g. with `strerror()`.

platform/linux-generic/arch/x86/odp_time_cpu.c
```diff
@@ -15,24 +14,67 @@
...
+				values[i] = values[n];
+				values[n] = tmp;
+			}
+		}
+	}
+}
+
+static uint64_t median(uint64_t values[], int num)
+{
+	sort(values, num);
```
TuomasTaipale, line 56: Use `qsort()`?

platform/linux-generic/arch/x86/odp_time_cpu.c
JannePeltonen: Good point. Renamed to nwait in v2 and also renamed sleep_nsec to wait_nsec.

platform/linux-generic/arch/x86/odp_time_cpu.c
JannePeltonen: Added in v2.

platform/linux-generic/arch/x86/odp_time_cpu.c
JannePeltonen: I thought about initially it but ended up not using it as performance is not a factor and I thinkt the simple sort is as (or more?) readable in this case.  Did not change this in v2.

TuomasTaipale: Slightly less code to maintain with `qsort()`, anyways, resolving this.

TuomasTaipale (approved):

PR #1999, opened 2024-01-16
[PATCH v2] test: sched_perf: add user area touch

psavol:
Added options to read (-u) and modify (-U) user area content per event reception.

### DIFF
--- a/test/performance/odp_sched_perf.c
+++ b/test/performance/odp_sched_perf.c
@@ -1,5 +1,5 @@
 /* Copyright (c) 2018, Linaro Limited
- * Copyright (c) 2020-2022, Nokia
+ * Copyright (c) 2020-2024, Nokia
  * All rights reserved.
  *
  * SPDX-License-Identifier:     BSD-3-Clause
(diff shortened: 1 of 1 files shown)


test/performance/odp_sched_perf.c
```diff
@@ -929,12 +976,16 @@ static int test_sched(void *arg)
 	uint32_t ctx_size = test_options->ctx_size;
 	uint32_t ctx_rd_words = test_options->ctx_rd_words;
 	uint32_t ctx_rw_words = test_options->ctx_rw_words;
+	uint32_t uarea_size = test_options->uarea_size;
+	uint32_t uarea_rd = test_options->uarea_rd;
+	uint32_t uarea_rw = test_options->uarea_rw;
```
MatiasElo, line 981: Could be `const`.

MatiasElo (approved):

test/performance/odp_sched_perf.c
psavol: Fixed in v2.

test/performance/odp_sched_perf.c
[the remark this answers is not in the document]
psavol: Updated in v2.

PR #2002, opened 2024-01-19
[PATCH v3] validation: add missing tests

MatiasElo:


test/validation/api/align/align.c
```diff
@@ -0,0 +1,131 @@
...
+
+#include <stddef.h>
+#include <stdint.h>
+
+typedef struct ODP_PACKED {
+	uint8_t a;
+	uint8_t b;
+	uint16_t c;
+	uint32_t d;
+} test_type_t;
```
TuomasTaipale, line 16: We could also have a padded structure to test that `ODP_PACKED` would remove the alignment padding.

psavol, line 16: In addition to this struct, there could be a struct with "holes" in it:

typedef struct ODP_PACKED {
	uint8_t a;
	uint16_t b;
	uint8_t c;
	uint32_t d;
} test_type_2_t;

test/validation/api/byteorder/byteorder.c
```diff
@@ -0,0 +1,107 @@
...
+}
+
+static void test_types(void)
+{
+	const uint16_t u16_val = 0xff00;
+	const uint32_t u32_val = 0xffff0000;
+	const uint64_t u64_val = 0xffffffff00000000;
+	const uint16_t u16_val_conv = 0x00ff;
+	const uint32_t u32_val_conv = 0x0000ffff;
+	const uint64_t u64_val_conv = 0x00000000ffffffff;
```
psavol, line 46: For better test coverage the value could be defined as e.g.:
u32_val = 0x12345678;
u32_val_conv = 0x78563412;

test/validation/api/align/align.c
MatiasElo: Second test structure added in V2.

test/validation/api/byteorder/byteorder.c
MatiasElo: Changed in V2.

psavol (approved):

TuomasTaipale (approved):

PR #2009, opened 2024-02-05
[PATCH v4] test: l2fwd: add data read option

psavol:


psavol:
v2: rebased

test/performance/odp_l2fwd.c
```diff
@@ -2288,6 +2285,11 @@ int main(int argc, char *argv[])
 	gbl_args->pkt_len = pkt_len;
 	gbl_args->seg_len = seg_len;
 
+	/* Print resulting values after option values */
+	printf("Packets per pool:   %u\n", num_pkt);
```
MatiasElo, line 2329: Could print some heading first. Based on only the application output it's not that clear why some option values are printed twice. All resulting value prints could also be combined into a function.

test/performance/odp_l2fwd.c
```diff
@@ -352,6 +355,33 @@ static inline void chksum_insert(odp_packet_t *pkt_tbl, int pkts)
...
+	uintptr_t data_ptr;
+	uint32_t bit, align;
+
+	for (int i = 0; i < num; i++) {
+		pkt = pkt_tbl[i];
+		data_ptr = (uintptr_t)odp_packet_data(pkt);
+
+		for (bit = 0, align = 1; bit < 32; bit++, align *= 2)
+			if (data_ptr & (0x1 << bit))
+				break;
```
MatiasElo, line 370: Not subject of this PR, but printing data alignment could be added also to `odp_packet_print()` and/or `odp_packet_print_data()`.

test/performance/odp_l2fwd.c
MatiasElo, line 1: Commit message: `Print first option values from the user, and the resulting pool parameters values.` -> `Print first option values from the user, and then the resulting pool parameters values.`

MatiasElo (commented):
Couple minor comments. Improves application throughput slightly on my test server.

test/performance/odp_l2fwd.c
```diff
@@ -1503,6 +1606,9 @@ static void usage(char *progname)
 	       "  -p, --packet_copy       0: Don't copy packet (default)\n"
 	       "                          1: Create and send copy of the received packet.\n"
 	       "                             Free the original packet.\n"
+	       "  -R, --data_rd <num>     Number of packet data words (uint64_t) to read on every\n"
```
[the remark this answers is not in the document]
psavol: Fixed in v4.

test/performance/odp_l2fwd.c
psavol: Added heading print in v4. Prints are done before normal and vectors pool create calls. Those are in two places and didn't want to refactor pool create code paths.

MatiasElo (approved):

PR #2012, opened 2024-02-06
[PATCH v8] api: ml: add machine learning API

JereLeppanen:


platform/linux-generic/odp_ml.c
```diff
@@ -0,0 +1,2673 @@
...
+	OrtSessionOptions	*session_opts;
+	uint32_t		max_compl_id;
+	uint32_t		num_loaded;
+	odp_atomic_u32_t	compl_status[ML_MAX_COMPL_ID];
+
+	odp_ml_model_info_t	info;
+	odp_ml_input_info_t	input_info[CONFIG_ML_MAX_INPUTS];
+	uint64_t		input_sizes[CONFIG_ML_MAX_INPUTS];
+	odp_ml_output_info_t	output_info[CONFIG_ML_MAX_OUTPUTS];
+	uint64_t		output_sizes[CONFIG_ML_MAX_INPUTS];
```
psavol, line 101: CONFIG_ML_MAX_INPUTS -> ... _OUTPUTS

example/sysinfo/odp_sysinfo.c
```diff
@@ -1073,6 +1079,31 @@ int main(int argc, char **argv)
...
+	printf("    max_models:             %u\n", ml_capa.max_models);
+	printf("    max_models_loaded:      %u\n", ml_capa.max_models_loaded);
+	printf("    max_model_size:         %" PRIu64 "B\n", ml_capa.max_model_size);
+	printf("    max_compl_id:           %u\n", ml_capa.max_compl_id);
+	printf("    max_inputs:             %u\n", ml_capa.max_inputs);
+	printf("    max_outputs:            %u\n", ml_capa.max_outputs);
+	printf("    max_segs_per_input:     %u\n", ml_capa.max_segs_per_input);
+	printf("    max_segs_per_output:    %u\n", ml_capa.max_segs_per_output);
+	printf("    min_input_align:        %u\n", ml_capa.min_input_align);
+	printf("    min_output_align:       %u\n", ml_capa.min_output_align);
```
MatiasElo, line 1093: Prints are missing for `packed_input_data` and `packed_output_data` capabilities.

include/odp/api/spec/ml.h
```diff
@@ -0,0 +1,699 @@
...
+ * When all previously started inference operations are complete, application uses
+ * odp_ml_model_unload() or odp_ml_model_unload_start() to unload the model. After a successful
+ * unload, the model may be destroyed with an odp_ml_model_destroy() call, or loaded again.
+ *
+ * <b> Completion identifiers </b>
+ *
+ * Completion identifiers are used with ML operations in asynchronous poll mode
+ * (#ODP_ML_COMPL_MODE_POLL). Application declares the maximum identifier value it will
+ * use per model with odp_ml_model_param_t.max_compl_id parameter. It cannot exceed
+ * the implementation capabity of odp_ml_capability_t.max_compl_id. Completion identifier
```
MatiasElo, line 54: Typo: `capabity`

platform/linux-generic/odp_ml.c
```diff
@@ -0,0 +1,2673 @@
...
+		return -1;
+	}
+
+	if (config->max_model_size > ML_MAX_MODEL_SIZE) {
+		_ODP_ERR("max_model_size %" PRIu64 " exceeds supported maximum model size %d\n",
+			 config->max_model_size, ML_MAX_MODEL_SIZE);
+		return -1;
+	}
+
+	if (!(capa->load.compl_mode_mask & config->load_mode_mask)) {
```
psavol, line 229: This check should be: (capa->mask & config->mask) == config->mask

I.e. all requested modes are also in capa, now it passes when at least one requested mode is in capa (but others not). Alternatively, could remove the check since all modes are supported.

platform/linux-generic/odp_ml.c
```diff
@@ -0,0 +1,2673 @@
...
+		} else {
+			_ODP_ERR("Dimension value: %" PRId64 " invalid\n", dims[i]);
+			return -1;
+		}
+	}
+
+	if (dyn_cnt == 0) {
+		shape->type = ODP_ML_SHAPE_STATIC;
+	} else if (dyn_cnt == 1 && first_dyn) {
+		shape->type = ODP_ML_SHAPE_BATCH;
```
psavol, line 403: Does ORT require that batch size is always the first dimension? In general, it could be also some other dimension.

include/odp/api/spec/ml_types.h
```diff
@@ -0,0 +1,873 @@
...
+	 * extra parameters are set to their default values. The default value is NULL.
+	 */
+	const odp_ml_model_extra_param_t *extra_param;
+
+} odp_ml_model_param_t;
+
+/** Results of model run operation */
+typedef struct odp_ml_run_result_t {
+	/** Model run error code
+	  *
```
TuomasTaipale, line 690: "*" alignment off slightly.

include/odp/api/spec/ml.h
```diff
@@ -0,0 +1,699 @@
...
+/**
+ * Load ML model
+ *
+ * Loads ML model in synchronous mode. When the call returns, load is complete and the model is
+ * ready for inference requests. A loaded model must be unloaded before it can be destroyed.
+ * The same model can be loaded and unloaded multiple times before being destroyed.
+ *
+ * The call optionally outputs results. Use NULL as 'result' pointer if results are not required.
+ *
+ * Application should not try to load concurrently more than configured number of models
```
TuomasTaipale, line 168: One might get the idea that this line somehow relates to multiple threads loading models, so maybe could just say that `max_models_loaded` can be active simultaneously or something.

platform/linux-generic/odp_ml.c
```diff
@@ -0,0 +1,2673 @@
...
+			return -1;
+		}
+
+		if (shape->type == ODP_ML_SHAPE_STATIC)
+			continue;
+
+		/* shape->type == ODP_ML_SHAPE_BATCH */
+		for (uint32_t j = 0; j < shape->num_dim; j++) {
+			if (shape->dim[j] == ODP_ML_DIM_DYNAMIC && !shape->dim_max[j]) {
+				_ODP_ERR("Missing dim_max[%u] for dynmic sized input[%u], please"
```
psavol, line 802: Typo: dynmic -> dynamic

platform/linux-generic/odp_ml.c
```diff
@@ -0,0 +1,2673 @@
...
+				   output_info, data, &result_local)) {
+		_ODP_ERR("Output tensors to data failed\n");
+		goto release_all_tensors;
+	}
+
+	retval = 1;
+
+release_all_tensors:
+	for (uint32_t i = 0; i < ml_info->num_outputs; i++)
+		ort_api->ReleaseValue(output_tensors[i]);
```
TuomasTaipale, line 2356: Is it safe to call `ReleaseValue` in cases where `num_outputs`/`num_inputs` weren't reached?

platform/linux-generic/odp_ml.c
```diff
@@ -0,0 +1,2673 @@
...
+
+	odp_ticketlock_lock(&mdl->lock);
+	if (odp_unlikely(mdl->state == ML_STATE_FREE)) {
+		_ODP_ERR("Model has not been created yet or has been destroyed\n");
+		odp_ticketlock_unlock(&mdl->lock);
+		result_local.error_code = ML_NOT_CREATED;
+		goto load_fail;
+	}
+
+	if (odp_unlikely(mdl->num_loaded >= _odp_ml_glb->ml_config.max_models_loaded)) {
```
psavol, line 1453: This counter should be global, not per model. Single model can be loaded only once, after that it needs to be unloaded (before loading again).

platform/linux-generic/odp_ml.c
```diff
@@ -0,0 +1,2673 @@
...
+
+	odp_ticketlock_lock(&mdl->lock);
+
+	if (mdl->state == ML_STATE_FREE) {
+		_ODP_ERR("Model not created\n");
+		odp_ticketlock_unlock(&mdl->lock);
+		return -1;
+	}
+
+	if (mdl->state == ML_STATE_LOADED) {
```
psavol, line 952: Only acceptable state is "created", maybe simply check against that. Currently, check misses inferencing mode, which also a "loaded" state.

platform/linux-generic/odp_ml.c
```diff
@@ -0,0 +1,2673 @@
...
+			_ODP_ERR("Completion event enqueue failed %" PRIu64 "\n",
+				 odp_queue_to_u64(compl_param->queue));
+			return -1;
+		}
+
+		return 0;
+	}
+
+	odp_atomic_store_rel_u32(&mdl->compl_status[compl_param->compl_id], 1);
+	mdl->result[compl_param->compl_id].user_ptr = compl_param->user_ptr;
```
psavol, line 1571: When atomic rel/acq is used for status flag, this result write should be before status flag write.

platform/linux-generic/odp_ml.c
```diff
@@ -0,0 +1,2673 @@
...
+	ml_model_t *mdl = ml_model_from_handle(model);
+
+	if (odp_unlikely(model == ODP_ML_MODEL_INVALID || compl_id > mdl->max_compl_id)) {
+		_ODP_ERR("Invalid model or compl_id: %u\n", compl_id);
+		return -2;
+	}
+
+	ret = odp_atomic_load_acq_u32(&mdl->compl_status[compl_id]);
+
+	if (result) {
```
psavol, line 1587: There's no need to write result, when ret == 0 (load not finished).

platform/linux-generic/odp_ml.c
```diff
@@ -0,0 +1,2673 @@
...
+int odp_ml_model_load(odp_ml_model_t model, odp_ml_load_result_t *result)
+{
+	odp_ml_load_result_t result_local;
+	int ret = -1;
+	ml_model_t *mdl = ml_model_from_handle(model);
+
+	memset(&result_local, 0, sizeof(result_local));
+
+	if (odp_unlikely(!(_odp_ml_glb->ml_config.load_mode_mask & ODP_ML_COMPL_MODE_SYNC))) {
+		_ODP_ERR("Synchronous load is not enabled!\n");
```
psavol, line 1434: When application has configured only async load flags, this check returns error when application calls async API call, which calls this function. I'd just remove the check from here.

test/validation/api/ml/ml.c
```diff
@@ -0,0 +1,485 @@
...
+
+	odp_pool_destroy(pool);
+}
+
+static void test_ml_fp32_to_uint8(void)
+{
+	uint8_t u8[8];
+	float fp[8] = {-20.f, -16.4f, -14.6f, -12.5f, 0, 31.4f, 80.f, 96.3f};
+	uint8_t expected[8] = {0, 0, 4, 10, 43, 127, 255, 255};
+
```
MatiasElo, line 372: Extra newline

platform/linux-generic/odp_ml.c
```diff
@@ -0,0 +1,2673 @@
...
+		_odp_buffer_subtype_set(buf, ODP_EVENT_ML_COMPL_LOAD);
+
+		result = odp_buffer_addr(buf);
+		result->error_code = 0;
+		result->user_ptr = compl_param->user_ptr;
+
+		if (odp_unlikely(odp_queue_enq(compl_param->queue, compl_param->event))) {
+			_ODP_ERR("Completion event enqueue failed %" PRIu64 "\n",
+				 odp_queue_to_u64(compl_param->queue));
+			return -1;
```
psavol, line 1564: At this point, the model is already in loaded state. If application would next try to destroy the model, the destroy call would complain about wrong state.

platform/linux-generic/odp_ml.c
```diff
@@ -0,0 +1,2673 @@
...
+	if (compl_param->mode == ODP_ML_COMPL_MODE_POLL)
+		odp_atomic_store_rel_u32(&mdl->compl_status[compl_param->compl_id], 0);
+
+	ret = odp_ml_model_unload(model, NULL);
+
+	if (odp_unlikely(ret))
+		return -1;
+
+	/* Upon completion of unloading (either succeeded or failed), send a
+	 * completion event to the given queue */
```
psavol, line 1660: Comment is not accurate as the event is sent only on success.

platform/linux-generic/odp_ml.c
```diff
@@ -0,0 +1,2673 @@
...
+
+			if (!size) {
+				_ODP_ERR("Size for %uth output is 0\n", output_index);
+				return -1;
+			}
+		}
+
+		seg_size_sum += data->output_seg[i].size;
+
+		if (seg_size_sum >= size) {
```
psavol, line 1877: Isn't (seg_size_sum > sum) an error also on output side (it's reported on input side)

test/validation/api/ml/ml.c
```diff
@@ -0,0 +1,485 @@
...
+	int8_t i8[6] = {-128, 46, 0, 56, 85, 127};
+	float fp32_expected[6] = {-9.2f, -0.5f, -2.8f, 0.0f, 1.45f, 3.55f};
+
+	odp_ml_fp32_from_int8(fp32, i8, 6, scale, zero_point);
+
+	for (uint32_t i = 0; i < 6; i++)
+		CU_ASSERT_EQUAL(fp32[i], fp32_expected[i]);
+}
+
+odp_testinfo_t ml_suite[] = {
```
MatiasElo, line 450: Tests are missing for `odp_ml_fp32_to_fp16()` and `odp_ml_fp32_from_fp16`.

platform/linux-generic/odp_ml.c
```diff
@@ -0,0 +1,2673 @@
...
+
+	if (!(_odp_ml_glb->ml_config.run_mode_mask & ODP_ML_COMPL_MODE_SYNC)) {
+		_ODP_ERR("Synchronous running is not enabled\n");
+		return -1;
+	}
+
+	odp_ticketlock_lock(&mdl->lock);
+	if (odp_unlikely(mdl->state != ML_STATE_LOADED)) {
+		_ODP_ERR("Wrong model state: not created or not loaded\n");
+		odp_ticketlock_unlock(&mdl->lock);
```
psavol, line 2283: Should not return error when another application thread is running inference

platform/linux-generic/odp_ml.c
```diff
@@ -0,0 +1,2673 @@
...
+		}
+
+		return 1;
+	}
+
+	/* compl_param->mode == ODP_ML_COMPL_MODE_POLL */
+	odp_atomic_store_rel_u32(&mdl->compl_status[compl_param->compl_id], 1);
+
+	/* odp_ml_model_run() succeeded */
+	mdl->result[compl_param->compl_id].user_ptr = compl_param->user_ptr;
```
psavol, line 2455: Comment can be removed, it refers to wrong function name anyway. user_ptr write should happen before status store-release.

platform/linux-generic/example/ml/mnist.c
```diff
@@ -0,0 +1,300 @@
...
+	odp_ml_model_param_init(&model_param);
+	if (read_model_from_file(model_file, &model_param)) {
+		printf("Read model file failed\n");
+		ret = -1;
+		goto odp_term;
+	}
+
+	ml_model = odp_ml_model_create("mnist", &model_param);
+	free(model_param.model);
+	if (!ml_model) {
```
psavol, line 233: Model create returns ODP_ML_MODEL_INVALID on failure.

platform/linux-generic/example/ml/model_explorer.c
```diff
@@ -0,0 +1,88 @@
...
+
+	odp_ml_model_param_init(&model_param);
+	if (read_model_from_file(argv[1], &model_param)) {
+		ret = -1;
+		goto odp_term;
+	}
+
+	ml_model = odp_ml_model_create("model-explorer", &model_param);
+	free(model_param.model);
+	if (!ml_model/*ODP_ML_MODEL_INVALID*/) {
```
psavol, line 68: Model create returns ODP_ML_MODEL_INVALID on failure.

platform/linux-generic/example/ml/simple_linear.c
```diff
@@ -0,0 +1,274 @@
...
+
+	odp_ml_model_param_init(&model_param);
+	if (read_model_from_file("simple_linear.onnx", &model_param)) {
+		ret = -1;
+		goto odp_term;
+	}
+
+	ml_model = odp_ml_model_create("simple linear", &model_param);
+	free(model_param.model);
+	if (!ml_model) {
```
psavol, line 180: Model create returns ODP_ML_MODEL_INVALID on failure.

platform/linux-generic/odp_ml.c
JereLeppanen: Removed.

platform/linux-generic/odp_ml.c
JereLeppanen: ORT doesn't require that. first_dyn may have been related to the multiple dynamic dimensions case, but we don't support that now. I don't see any need for first_dyn. Removed.

include/odp/api/spec/ml_quantize.h
```diff
@@ -0,0 +1,124 @@
...
+ * @param scale          Scale for de-quantization
+ * @param zerop          Zero point for de-quantization
+ */
+void odp_ml_fp32_from_int8(float *dst_fp32, const int8_t *src_i8, uint32_t num, float scale,
+			   int8_t zerop);
+
+/**
+ * Quantize 32-bit float to 16-bit float
+ *
+ * Quantizes 'num' 32-bit floating point values to 16-bit floating point values.
```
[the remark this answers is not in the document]
JereLeppanen: Well, currently the API doesn't specify it. At least one potential problem in specifying this is that some hardware might not be able to implement the specification efficiently.

include/odp/api/spec/ml.h
```diff
@@ -0,0 +1,699 @@
...
+ * parameters into their default values. The use of model name is optional. Unique names are not
+ * required. However, odp_ml_model_lookup() returns only a single matching model. Maximum name
+ * string length is #ODP_ML_MODEL_NAME_LEN.
+ *
+ * The call copies the model binary and prepares it for loading. Application may free memory
+ * buffers pointed by the parameters when the call returns. Use odp_ml_model_load()
+ * or odp_ml_model_load_start() to load the model. A model is ready for inference runs
+ * (see e.g. odp_ml_run()) after it has been loaded successfully.
+ *
+ * When model metadata misses some details of model input / output data format, user can
```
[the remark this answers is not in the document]
JereLeppanen: Looks crowded to me without the spaces. I didn't change these yet, but I don't have a strong opinion.

include/odp/api/spec/ml_types.h
```diff
@@ -0,0 +1,873 @@
...
+
+	/** Size of the model binary in bytes */
+	uint64_t size;
+
+	/**
+	 * Maximum completion identifier value
+	 *
+	 * When application uses asynchronous poll mode (#ODP_ML_COMPL_MODE_POLL) operations with
+	 * the model, it will choose completion identifier values between 0 and this value. Valid
+	 * values range from 0 to max_compl_id capability. The default value is zero.
```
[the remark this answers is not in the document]
JereLeppanen: I agree that consistency would be nice, but I'm not sure what exactly to do to these, if anything.

include/odp/api/spec/ml.h
JereLeppanen: I think it does relate to multiple threads. There can be multiple loads and unloads invoked from multiple threads, and the application should make sure that there's no possibility of breaching the limit.

DEPENDENCIES
```diff
@@ -305,6 +305,36 @@ Prerequisites for building the OpenDataPlane (ODP) API
...
+3.7 Machine Learning API support (optional)
+   Use ML API for model inferencing. ML implementation uses ONNX Runtime library
+   (https://github.com/microsoft/onnxruntime). ODP has been tested with ONNX
+   Runtime version 1.16.3.
+
+3.7.1 Prebuilt onnxruntime download
+   Download a default CPU version onnxruntime-linux-x64-*.tgz and unzip it to
+   any folder.
+
+   # Debian/Ubuntu
```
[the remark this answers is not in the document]
JereLeppanen: Maybe there was at some point, but not anymore. Removed.

platform/linux-generic/include/odp_config_internal.h
```diff
@@ -199,6 +199,18 @@ extern "C" {
...
+#define CONFIG_ML_MAX_MODELS 4
+
+/* Maximum number of inputs for a ML model. */
+#define CONFIG_ML_MAX_INPUTS 4
+
+/* Maximum number of outputs for a ML model. */
+#define CONFIG_ML_MAX_OUTPUTS 4
+
+/* Maximum number of concurrent inferences that can be performed for a ML model. */
+#define CONFIG_ML_MAX_CONCUR_INFERS 2
```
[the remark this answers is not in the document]
JereLeppanen: CONFIG_ML_MAX_CONCUR_INFERS is actually unused. Removed.

platform/linux-generic/odp_ml.c
JereLeppanen: In terms of implementation, yes. ReleaseValue() is just a delete, which is safe since the pointers here are initialized to zero. ORT API doesn't document whether ReleaseValue() may be called with NULL, but I think it's a reasonable assumption.

platform/linux-generic/odp_ml.c
JereLeppanen: Since ML_MAX_MODELS_CREATED and ML_MAX_MODELS_LOADED are the same, and each created model can only be loaded once, num_loaded has no purpose. Removed.

platform/linux-generic/odp_ml.c
JereLeppanen: Oops, removed this one, but forgot unload and run. I'll remove those in the next one.

platform/linux-generic/odp_ml.c
JereLeppanen: Added unload here.

test/validation/api/ml/ml.c
```diff
@@ -0,0 +1,485 @@
...
+	CU_ASSERT(ml_capa.max_models_loaded > 0);
+	CU_ASSERT(ml_capa.max_inputs > 0);
+	CU_ASSERT(ml_capa.max_outputs > 0);
+	CU_ASSERT(ml_capa.max_compl_id > 0);
+	CU_ASSERT(ml_capa.max_segs_per_input > 0);
+	CU_ASSERT(ml_capa.max_segs_per_output > 0);
+	CU_ASSERT(ml_capa.min_input_align > 0);
+	CU_ASSERT(ml_capa.min_output_align > 0);
+
+	if (EVENT_SUPPORT(ml_capa.load.compl_mode_mask, ml_capa.run.compl_mode_mask)) {
```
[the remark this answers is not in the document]
JereLeppanen: Removed. Although maybe it could have been a macro or a function with a better name and parameters.

test/validation/api/ml/ml.c
```diff
@@ -0,0 +1,485 @@
...
+	compl = odp_ml_compl_alloc(compl_pool);
+	CU_ASSERT_NOT_EQUAL_FATAL(compl, ODP_ML_COMPL_INVALID);
+
+	u64 = odp_ml_compl_to_u64(compl);
+	CU_ASSERT_NOT_EQUAL(u64, odp_ml_compl_to_u64(ODP_ML_COMPL_INVALID));
+
+	event = odp_ml_compl_to_event(compl);
+	CU_ASSERT_EQUAL(odp_event_type(event), ODP_EVENT_ML_COMPL);
+
+	odp_ml_compl_free(compl);
```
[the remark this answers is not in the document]
JereLeppanen: compl_alloc_max test added,

test/validation/api/ml/ml.c
JereLeppanen: The tests would have to be in linux-gen validation. The implementation is adapted from DPDK, so what kind of tests would make sense?

JereLeppanen:
All findings fixed in v2 or commented here.

include/odp/api/spec/ml.h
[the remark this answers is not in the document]
TuomasTaipale: No strong opinion either, except better to be either or and not mixed.

platform/linux-generic/odp_pool.c
```diff
@@ -1258,6 +1258,10 @@ int odp_pool_info(odp_pool_t pool_hdl, odp_pool_info_t *info)
 		info->dma_pool_param.uarea_size = pool->params.buf.uarea_size;
 		info->dma_pool_param.cache_size = pool->params.buf.cache_size;
 
+	} else if (pool->type_2 == ODP_POOL_ML_COMPL) {
```
MatiasElo, line 1261: Seems like implementations are missing for event user area APIs: `odp_event_user_area()`, `odp_event_user_area_and_flag()`

MatiasElo: `test_ml_compl_user_area()` could also test there functions.

test/validation/api/ml/ml.c
MatiasElo: The tests belong here in API level.  For example the test(s) could utilize normal floats (float -> fp16 -> float) and allow some variation (e.g. 1%) in the results.

test/validation/api/ml/ml.c
```diff
@@ -0,0 +1,501 @@
...
+
+static int check_ml_support(void)
+{
+	if (global.disabled)
+		return ODP_TEST_INACTIVE;
+
+	return ODP_TEST_ACTIVE;
+}
+
+static void test_ml_capability(void)
```
MatiasElo, line 54: `test_packet_pool_ext_capa()` could be updated to include `ODP_POOL_ML_COMPL` in unsupported types.

platform/linux-generic/test/validation/api/ml/README.md
```diff
@@ -0,0 +1,23 @@
+# How to run ML validation test
+
+Simple onnx models are used to test ML API.
+
+## Generate models
+
+### Install python requirments
```
MatiasElo, line 7: Typo: `requirments`

platform/linux-generic/example/ml/README.md
```diff
@@ -0,0 +1,96 @@
...
+the input. Thus, the number of threads is the number of numbers in the second argument.
+
+```bash
+$ ./simple_linear [2,4,5]
+.
+.
+.
+y = 3 * 2 + 4: 10
+../../../platform/linux-generic/odp_ml.c:2174:odp_ml_model_run():input_tensor[0]: 0x7fde8c4372f0
+../../../platform/linux-generic/odp_ml.c:2174:odp_ml_model_run():input_tensor[0]: 0x7fde844372f0
```
MatiasElo, line 39: Only visible in debug mode. Could be removed.

platform/linux-generic/odp_ml.c
```diff
@@ -0,0 +1,2642 @@
...
+
+	conf_str =  "ml.graph_optimization_level";
+	if (_odp_libconfig_lookup_str(conf_str, opt_level_str,
+				      ML_MAX_CONFIG_STR_LEN) < 0) {
+		_ODP_ERR("Config option '%s' not found.\n", conf_str);
+		return -1;
+	}
+
+	if (opt_level_from_str(opt_level_str, &opts->graph_opt_level)) {
+		_ODP_ERR("Graph otimize level %s not supported\n", opt_level_str);
```
MatiasElo, line 2555: Typo: `otimize`

test/validation/api/ml/ml.c
```diff
@@ -0,0 +1,501 @@
...
+		CU_ASSERT(ml_capa.pool.max_num > 0);
+		CU_ASSERT(ml_capa.pool.max_cache_size >= ml_capa.pool.min_cache_size);
+	}
+
+	CU_ASSERT(ml_capa.load.compl_mode_mask);
+	CU_ASSERT(ml_capa.run.compl_mode_mask);
+
+	if (ml_capa.load.compl_mode_mask & ODP_ML_COMPL_MODE_POLL ||
+	    ml_capa.run.compl_mode_mask & ODP_ML_COMPL_MODE_POLL)
+		CU_ASSERT(ml_capa.max_compl_id > 0);
```
psavol, line 91: This is max ID (not num IDs). In theory an implementation could have only one possible ID (== 0). So, I'd remove the check.

test/validation/api/ml/ml.c
```diff
@@ -0,0 +1,501 @@
...
+	if (ml_capa.load.compl_mode_mask & ODP_ML_COMPL_MODE_EVENT)
+		CU_ASSERT(ml_capa.load.compl_queue_plain || ml_capa.load.compl_queue_sched);
+
+	if (ml_capa.run.compl_mode_mask & ODP_ML_COMPL_MODE_EVENT)
+		CU_ASSERT(ml_capa.run.compl_queue_plain || ml_capa.run.compl_queue_sched);
+}
+
+static void test_ml_param(uint8_t fill)
+{
+	odp_ml_model_param_t model_param;
```
psavol, line 102: This function could also test odp_ml_config_t for default values.

test/validation/api/ml/ml.c
```diff
@@ -0,0 +1,501 @@
...
+{
+	odp_ml_print();
+}
+
+static void ml_compl_pool_create_max_pools(void)
+{
+	int ret;
+	uint32_t i, j;
+	odp_ml_compl_pool_param_t ml_pool_param;
+	uint32_t max_pools = global.ml_capa.pool.max_pools - 1;
```
psavol, line 151: Is there a reason to test only max_pools - 1? E.g. I didn't notice any pool created in the init phase. If there's reason, this line should have a comment about that - otherwise should test capa max_pools.

test/validation/api/ml/ml.c
```diff
@@ -0,0 +1,501 @@
...
+}
+
+static void compl_alloc_max(void)
+{
+	uint64_t u64;
+	odp_event_t event;
+	odp_pool_t compl_pool;
+	odp_ml_compl_pool_param_t pool_param;
+	const int num = global.ml_capa.pool.max_num;
+	odp_ml_compl_t *compl = malloc(num * sizeof(odp_ml_compl_t));
```
psavol, line 213: Variable 'num' should be uint32_t as capa max_num is uint32_t.

A very large capa would lead to a very large malloc. Maybe have max define (1M events) to cap the test memory usage and test run time. Also, should check if malloc failed.

test/validation/api/ml/ml.c
```diff
@@ -0,0 +1,501 @@
...
+
+		if (compl_evs[i] == ODP_ML_COMPL_INVALID)
+			break;
+
+		addr = odp_ml_compl_user_area(compl_evs[i]);
+
+		CU_ASSERT_FATAL(addr != NULL);
+		CU_ASSERT(prev != addr);
+
+		prev = addr;
```
psavol, line 310: Test should also memset() the entire user area to see that there's valid memory behind the returned pointer.

memset(addr, 0, size);

platform/linux-generic/test/validation/api/ml/ml_linux.c
```diff
@@ -0,0 +1,1259 @@
...
+	if (global.queue == ODP_QUEUE_INVALID) {
+		ODPH_ERR("Queue create failed\n");
+		free(global.model_param.model);
+		return -1;
+	}
+
+	/* Create an ML job completion pool */
+	if (ml_capa->pool.max_num < NUM_COMPL) {
+		ODPH_ERR("Too small ML compl pool %u\n", ml_capa->pool.max_num);
+		return -1;
```
psavol, line 225: Error branch missing memory free "free(global.model_param.model);". Maybe simpler to use "goto error" in all branches, and have the memory free there once.

platform/linux-generic/test/validation/api/ml/ml_linux.c
```diff
@@ -0,0 +1,1259 @@
...
+	if (global.compl_pool == ODP_POOL_INVALID) {
+		free(global.model_param.model);
+		ODPH_ERR("Create ML completion pool failed\n");
+		return -1;
+	}
+
+	return 0;
+}
+
+static int ml_suit_term(void)
```
psavol, line 241: Typo in function name: suit -> suite

platform/linux-generic/test/validation/api/ml/ml_linux.c
```diff
@@ -0,0 +1,1259 @@
...
+
+	test_model = odp_ml_model_create(NULL, &global.model_param);
+	CU_ASSERT_NOT_EQUAL_FATAL(test_model, ODP_ML_MODEL_INVALID);
+
+	ret = odp_ml_model_load(test_model, &result);
+	CU_ASSERT_EQUAL(ret, 0);
+	CU_ASSERT_EQUAL(result.error_code, 0);
+
+	ret = odp_ml_model_load(test_model, NULL);
+	CU_ASSERT(ret);
```
psavol, line 497: Validation test should not try to load again an already loaded model. It's not guaranteed that implementation will catch that.

platform/linux-generic/test/validation/api/ml/ml_linux.c
```diff
@@ -0,0 +1,1259 @@
...
+	if (result.user_ptr)
+		CU_ASSERT_EQUAL(*(int *)result.user_ptr, dummy);
+
+	/* Free the event */
+	odp_ml_compl_free(compl);
+
+	CU_ASSERT_EQUAL_FATAL(odp_ml_model_unload(global.ml_model, NULL), 0);
+}
+
+static void test_ml_model_unload(void)
```
psavol, line 635: There's already test_ml_model_load() which does sync load and unload. So, this is redundant and can be removed.

platform/linux-generic/test/validation/api/ml/ml_linux.c
```diff
@@ -0,0 +1,1259 @@
...
+}
+
+static int check_load_poll_sync(void)
+{
+	if (global.disabled)
+		return ODP_TEST_INACTIVE;
+
+	if ((global.ml_config.load_mode_mask & ODP_ML_COMPL_MODE_POLL) &&
+	    (global.ml_config.load_mode_mask & ODP_ML_COMPL_MODE_SYNC))
+		return ODP_TEST_ACTIVE;
```
psavol, line 291: If sync mode (load/unload) is not supported, poll or event mode load/unload is not tested. I.e. then all these load/unload tests are skipped.

It would be better to test both load and unload in the same mode, and have only three test cases: sync load/unload, poll load/unload, event load/unload.

platform/linux-generic/test/validation/api/ml/ml_linux.c
```diff
@@ -0,0 +1,1259 @@
...
+	compl_param.user_ptr = user_ptr;
+
+	ret = odp_ml_model_unload_start(global.ml_model, &compl_param);
+	CU_ASSERT_EQUAL(ret, 0);
+
+	/* Return when odp_ml_model_unload_start() failed */
+	if (ret) {
+		ret = odp_ml_model_unload(global.ml_model, NULL);
+		CU_ASSERT_EQUAL_FATAL(ret, 0);
+		return;
```
psavol, line 678: If unload start fails, it can be fatal assert. No need to try recovery with sync API.

platform/linux-generic/test/validation/api/ml/ml_linux.c
```diff
@@ -0,0 +1,1259 @@
...
+			goto fail;
+
+		for (uint32_t j = 0; j < batch_size * NUM_COLUMN; j++)
+			CU_ASSERT_EQUAL(y[j], y_expected[j]);
+
+		batch_size--;
+	}
+
+	/* Result can be NULL, when output batch size is not needed */
+	run_param.result = NULL;
```
psavol, line 873: The comment refers to dynamic dimensions on output. Result does not have output batch size. Could simply say: "Test also without run results"

platform/linux-generic/test/validation/api/ml/ml_linux.c
```diff
@@ -0,0 +1,1259 @@
...
+	odp_ml_data_t data[RUN_NUM];
+	odp_ml_data_seg_t input_seg;
+	odp_ml_data_seg_t output_seg;
+	odp_ml_run_param_t param[RUN_NUM];
+	odp_ml_run_result_t result[RUN_NUM];
+
+	ret = odp_ml_model_load(global.ml_model, NULL);
+	CU_ASSERT_EQUAL_FATAL(ret, 0);
+
+	param[0] = global.run_param;
```
psavol, line 965: Is param[1] uninitialized?

platform/linux-generic/test/validation/api/ml/ml_linux.c
```diff
@@ -0,0 +1,1259 @@
...
+	input_seg.size = sizeof(int32_t);
+	input_seg.addr = &x;
+
+	data[1].num_output_seg = NUM_OUTPUTS;
+	data[1].output_seg = &output_seg;
+	output_seg.size = sizeof(int32_t);
+	output_seg.addr = &y;
+
+	ret = odp_ml_run_multi(global.ml_model, data, param, RUN_NUM);
+	CU_ASSERT_EQUAL(ret, RUN_NUM);
```
psavol, line 982: Implementation could always return one on a multi-call. Test should handle both ret = 2, and ret =1 -> ret =1 as success.

platform/linux-generic/test/validation/api/ml/ml_linux.c
```diff
@@ -0,0 +1,1259 @@
...
+
+	fclose(model_file);
+	if (size != 1) {
+		ODPH_ERR("\n\nRead model file failed\n");
+		return -1;
+	}
+
+	if ((global.ml_config.load_mode_mask & ODP_ML_COMPL_MODE_POLL) ||
+	    (global.ml_config.run_mode_mask & ODP_ML_COMPL_MODE_POLL))
+		model_param->max_compl_id = MAX_COMPL_ID;
```
psavol, line 120: It seems that max compl ID capa is not checked (max is 20, value of 4 is used in test). If only one ID is needed, value of 0 could be used without capa check.

platform/linux-generic/test/validation/api/ml/ml_linux.c
```diff
@@ -0,0 +1,1259 @@
...
+
+	memset(global.output_seg.addr, 0, global.output_seg.size);
+	ret = odp_ml_run_start(global.ml_model, &global.data, &compl_param, NULL);
+
+	CU_ASSERT_EQUAL(ret, 1);
+
+	/* Return when odp_ml_run_start() failed */
+	if (ret != 1) {
+		/* Unload model */
+		ret = odp_ml_model_unload(global.ml_model, NULL);
```
psavol, line 1079: No need to try recovery after failed run start.

test/validation/api/ml/ml.c
JereLeppanen: Limited to one million in v3, so no need to change the type.

platform/linux-generic/test/validation/api/ml/ml_linux.c
JereLeppanen: Did the "goto error" thing in v3.

platform/linux-generic/test/validation/api/ml/ml_linux.c
```diff
@@ -0,0 +1,1259 @@
...
+	/* Return when odp_ml_model_load_start() failed */
+	if (ret) {
+		odp_ml_compl_free(compl);
+		ODPH_ERR("ML model odp_ml_model_load_start() failed\n");
+		return;
+	}
+
+	/* Run event scheduler to find the ml completion event and verify it */
+	if (get_result_from_ml_compl_event(&result, NULL)) {
+		odp_ml_compl_free(compl);
```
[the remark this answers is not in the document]
JereLeppanen: In v3, freeing the event in get_result_from_ml_compl_event(), and also moved odp_ml_compl_[load|run]_result(compl, NULL) there.

platform/linux-generic/test/validation/api/ml/ml_linux.c
JereLeppanen: Done. Separate unload_async tests removed.

platform/linux-generic/test/validation/api/ml/ml_linux.c
JereLeppanen: Only the unused batch_size was uninitialized. I added a call to odp_ml_run_param_init().

platform/linux-generic/test/validation/api/ml/ml_linux.c
JereLeppanen: Added a loop for this and also for odp_ml_run_start_multi().

platform/linux-generic/test/validation/api/ml/ml_linux.c
```diff
@@ -0,0 +1,1259 @@
...
+
+/* Test asynchronous inference running in ODP_ML_COMPL_MODE_POLL mode */
+static void test_ml_model_run_async_poll(void)
+{
+	int ret;
+	void *user_ptr;
+	odp_ml_run_result_t result;
+	odp_ml_data_seg_t *outputs;
+	odp_ml_compl_param_t compl_param;
+	u_int32_t compl_id = 4;
```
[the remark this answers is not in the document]
JereLeppanen: Fixed. u_int32_t is defined in sys/types.h system header. IIRC, this is the second accidental usage in the past three years.

platform/linux-generic/test/validation/api/ml/ml_linux.c
JereLeppanen: Using compl id 0 in v3.

JereLeppanen:
All findings since v2 fixed in v3.

test/validation/api/ml/ml.c
JereLeppanen: I'm not entirely convinced about its usefulness, but anyway test added in v3.

platform/linux-generic/odp_ml.c
JereLeppanen: Done in v3.

test/validation/api/ml/ml.c
```diff
@@ -0,0 +1,577 @@
...
+{
+	float fp32[4];
+	uint16_t fp16[4];
+	float big = 1, small = 1;
+	const float phi = 1.618033988749;
+
+	fp32[0] = 0.0;
+	fp32[1] = -0.0;
+	fp32[2] = 1.0 / 0.0;
+	fp32[3] = 0.0 / 0.0;
```
MatiasElo, line 507: I'd drop testing NaN values as they may be problematic for some implementations (e.g. Arm alternative half-precision).

JannePeltonen: Division by zero has undefined behavior in C, so it would make sense not to do it.

platform/linux-generic/test/validation/api/ml/ml_linux.c
```diff
@@ -0,0 +1,1173 @@
...
+	if (ret) {
+		odp_ml_compl_free(compl);
+		ODPH_ERR("ML model odp_ml_model_load_start() failed\n");
+		return;
+	}
+
+	/* Run event scheduler to find the ml completion event and verify it */
+	if (get_result_from_ml_compl_event(&result, NULL)) {
+		ret = odp_ml_model_unload(global.ml_model, NULL);
+		CU_ASSERT_EQUAL(ret, 0);
```
psavol, line 635: Sync unload call needs to be removed from async test.

get_result_from_ml_compl_event() could use ODPH_ERR() to print error reason for easier debug when this assert happens.

include/odp/api/spec/ml.h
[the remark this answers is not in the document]
JereLeppanen: I'll leave these be, I don't know which way is better.

include/odp/api/spec/ml_types.h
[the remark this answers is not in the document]
JereLeppanen: I'm leaving these as-is. Too many judgement calls, someone else's commit.

JereLeppanen:
v5: Use only CU_ASSERT and CU_ASSERT_FATAL cunit assert types.

platform/linux-generic/test/validation/api/ml/ml_linux.c
```diff
@@ -0,0 +1,1170 @@
...
+	/* Return when odp_ml_model_load_start() failed */
+	if (ret) {
+		odp_ml_compl_free(compl);
+		ODPH_ERR("ML model odp_ml_model_load_start() failed\n");
+		return;
+	}
+
+	/* Run event scheduler to find the ml completion event and verify it */
+	if (get_result_from_ml_compl_event(&result, NULL)) {
+		CU_ASSERT(ret == 0);
```
psavol, line 637: This assert is not needed. It was checking previously the removed unload call success. Now get_result_from_ml_compl_event() reports CU failure when timeout happens.

JereLeppanen: Removed two occurrences in v6.

JereLeppanen:
v6:
- Rebase.
- Remove extra CU_ASSERTs when get_result_from_ml_compl_event() fails.
- Ensure string termination after strncpy().

include/odp/api/spec/ml.h
psavol: The sentence refers to config / max_models_loaded:

```
/**
	 * Maximum number of models loaded
	 *
	 * Maximum number of models that the application will keep loaded simultaneously.
	 * The default value is 1.
	 */
	uint32_t max_models_loaded;
```

May be the same "keep loaded" term could be used here: "... should not try to keep loaded more than ..."

JereLeppanen:
v7:
- Reword a sentence in odp_ml_model_load() documentation.
- Add review tags.

PR #2015, opened 2024-02-19
[PATCH v2] test: sched_perf: add signal handler

MatiasElo:
Add signal handler for SIGINT for closing the application. Also, run the application indefinitely if the number of events to schedule is set to zero.

### DIFF
--- a/test/performance/odp_sched_perf.c
+++ b/test/performance/odp_sched_perf.c
@@ -13,6 +13,11 @@
  * @cond _ODP_HIDE_FROM_DOXYGEN_
  */
 
+#ifndef _GNU_SOURCE
+#define _GNU_SOURCE /* Needed for sigaction */
+#endif
+
+#include <signal.h>
 #include <stdio.h>
 #include <string.h>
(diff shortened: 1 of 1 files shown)


test/performance/odp_sched_perf.c
```diff
@@ -99,6 +104,8 @@ typedef struct test_global_t {
 	test_stat_t stat[ODP_THREAD_COUNT_MAX];
 	thread_arg_t thread_arg[ODP_THREAD_COUNT_MAX];
 	odp_atomic_u32_t num_worker;
+	/* Break worker loop if set to 1 */
```
JannePeltonen, line 107: Actually, this breaks the loop if nonzero. Maybe the comment could just be dropped as the code is obvious anyway.

JannePeltonen (approved):

test/performance/odp_sched_perf.c
MatiasElo: Comment dropped in V2.

PR #2020, opened 2024-02-26
[PATCH v5] linux-gen: avoid strncpy()

JereLeppanen:


helper/include/odp/helper/strcpy.h
```diff
@@ -0,0 +1,62 @@
...
+
+#include <odp/api/hints.h>
+#include <string.h>
+
+#ifdef __cplusplus
+extern "C" {
+#endif
+
+/**
+ * @defgroup odph_strcpy ODPH strcpy
```
MatiasElo, line 22: A more generic module name could be useful if we anticipate other similar functions being added later on.

helper/include/odp/helper/strcpy.h
```diff
@@ -0,0 +1,62 @@
...
+ *
+ * ODP string copy helper
+ *
+ * @{
+ */
+
+/**
+ * Copy a string
+ *
+ * Like strncpy(), but additionally ensure that the destination string is NUL-terminated, unless sz
```
MatiasElo, line 32: Typo: `NUL` and I'd perhaps change `ensure` -> `ensures` and `return` -> `returns`.

helper/include/odp/helper/strcpy.h
```diff
@@ -0,0 +1,62 @@
...
+ * is zero in which case return dst without doing anything else.
+ *
+ * @param dst Pointer to destination string.
```
MatiasElo, line 35: `@param[out]`

helper/include/odp/helper/strcpy.h
JannePeltonen: NUL is actually correct here.

platform/linux-generic/odp_schedule_sp.c
```diff
@@ -107,7 +108,7 @@ typedef struct ODP_ALIGNED_CACHE sched_group_t {
 
 		/* All groups */
 		struct {
-			char          name[ODP_SCHED_GROUP_NAME_LEN + 1];
+			char          name[ODP_SCHED_GROUP_NAME_LEN];
```
JannePeltonen, line 111: This is a functional change that shortens the maximum group name length. It was a bit unexpected as it was not mentioned in the commit message.

The API spec says in schedule.h that "maximum string length is ODP_SCHED_GROUP_NAME_LEN", i.e. excluding the terminating null character, but in schedule_types.h it says that ODP_SCHED_GROUP_NAME_LEN includes the null character. This is inconsistent and probably should be fixed in the API. ODP_TIMER_POOL_NAME_LEN has the same problem.

So, since the API is inconsistent, the old behavior seems more correct since it works with either interpretation of the constant. However, the other schedulers already include the null character in the group name length, so maybe it is ok to make the sp work the same way.

helper/include/odp/helper/strcpy.h
MatiasElo: Oh, learned something new today 👍

helper/include/odp/helper/strcpy.h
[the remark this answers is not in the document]
JereLeppanen: Separate PR: #2024.

helper/include/odp/helper/strcpy.h
JereLeppanen: However, the references in https://en.wikipedia.org/wiki/Null_character mostly use "null character" and "null terminated", etc. So I'm doing the same in v2.

JannePeltonen: Yes, "null character" is also used in C standard, so null-terminated would probably be better than NUL-terminated. But NULL would not be right, I think.

platform/linux-generic/odp_schedule_sp.c
JereLeppanen: This was hidden because I didn't even consider it an inconsistency, but more like an inaccuracy. Either way, I agree that the API should be clarified.

This change is a separate commit in v2.

JereLeppanen:
All comments so far processed in v2.

platform/linux-generic/test/validation/api/shmem/shmem_linux.c
```diff
@@ -226,9 +226,12 @@ int main(int argc __attribute__((unused)), char *argv[])
 	if (exeext == NULL)
 		exeext = "";
 
+	strncpy(dir_name, argv[0], PATH_MAX - 1);
+	dir_name[PATH_MAX - 1] = 0;
+	dirname(dir_name);
```
MatiasElo, line 231: It would be clearer to save `dirname()` return value and use that. `dirname()` path argument handling may vary by implementation.

MatiasElo (commented):
One minor comment, otherwise looking OK.

platform/linux-generic/test/validation/api/shmem/shmem_linux.c
JereLeppanen: v3.

JereLeppanen:
v3:
- Fix dirname() usage.

MatiasElo (approved):

JereLeppanen:
v4:
- Rebase.
- Add review tags.

PR #2022, opened 2024-02-26
[PATCH v5] Remove old redundant example applications

MatiasElo:


test/validation/api/time/time.c
```diff
@@ -975,6 +977,136 @@ static void time_test_global_sync_control(void)
 	time_test_global_sync(1);
 }
 
+static int run_time_global_thread(void *arg)
+{
+	global_shared_mem_t *gbl = arg;
+	const odp_queue_t queue = gbl->queue;
```
JannePeltonen, line 983: In the to-be-removed example app the test uses a separate queue for each worker, ensuring that the events cannot be processed by single worker thread. With this code it could happen that only one worker thread manages to dequeue anything, which would reduce the inter-thread time stampt test to the checks between the single worker and the main thread at termination. Perhaps you could consider having separate queues for the workers and e.g. have the events traverse the workers in a circular fashion (or even randomized as in the to-be-removed app).

MatiasElo:
V2:
* Use multiple queues in the new multithreaded global time test
* Run sched_perf tests with multiple worker counts to improve testing coverage

test/validation/api/time/time.c
MatiasElo: Implemented in V2.

MatiasElo:
V4:
* Reduce max worker count in `odp_sched_perf_run.sh`
* Set application path `odp_timer_accuracy_run.sh`
* Improve `test: scheduling: remove odp_scheduling test application` commit message

JannePeltonen (approved):

PR #2025, opened 2024-02-29
[PATCH v3] api: event: add user flag set function

MatiasElo:


psavol (approved):

include/odp/api/spec/event.h
```diff
@@ -143,6 +143,20 @@ void *odp_event_user_area(odp_event_t event);
...
 void *odp_event_user_area_and_flag(odp_event_t event, int *flag);
 
+/**
+ * Set event user flag
+ *
+ * Set (or clear) the user flag.
+ *
+ * The user flag maps to the user flag value of underlying event type (e.g.
+ * odp_packet_user_flag() for packets). If the event does not have user flag,
+ * nothing is done.
```
jerinjacobk, line 153: @MatiasElo I prefer to have capability flag or return -ENOSUP kind standard return code to express if implementation does not support feature.

ashwinyes:

include/odp/api/spec/event.h
MatiasElo: Okay, we can discuss about this in tomorrow's call.

jerinjacobk: >Did you mean to have capabilities for those?

Yes

include/odp/api/spec/event.h
```diff
@@ -143,6 +143,20 @@ void *odp_event_user_area(odp_event_t event);
...
+ *
+ * @param event   Event handle
+ * @param val     New value for the flag. Zero clears the flag, other values set the flag.
+ */
+void odp_event_user_flag_set(odp_event_t event, int val);
```
psavol, line 158: We could add a (bool) return value that indicates if event type supports user flag

PR #2027, opened 2024-03-04
[PATCH v1] fix address sanitizer findings

JereLeppanen:
Fix findings from GCC -fsanitize=address.


JereLeppanen:
Never mind, #1964 already exists.

PR #2032, opened 2024-03-11
[PATCH v3] linux-gen: update copyright format

MatiasElo:
Update all implementation files to use the new shorter copyright format:

 * Copyright (c) <YEAR> <COPYRIGHT HOLDER> */

### DIFF
--- a/platform/linux-generic/arch/x86/cpu_flags.c
+++ b/platform/linux-generic/arch/x86/cpu_flags.c
@@ -1,12 +1,11 @@
-/* Copyright (c) 2017-2018, Linaro Limited
- * Copyright (c) 2023, Nokia
- * All rights reserved.
- *
- * SPDX-License-Identifier:     BSD-3-Clause
- */
-
 /* SPDX-License-Identifier: BSD-3-Clause
+ * Copyright (c) 2017-2018 Linaro Limited
+ * Copyright (c) 2023 Nokia
+ *
  * Copyright(c) 2010-2015 Intel Corporation
+ *   - lib/eal/x86/include/rte_cpuflags.h
+ *   - lib/eal/x86/rte_cpuflags.c
+ *   - lib/eal/x86/rte_cpuid.h
  */
 
 #include "cpu_flags.h"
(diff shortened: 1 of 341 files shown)


platform/linux-generic/arch/x86/cpu_flags.c
```diff
@@ -1,8 +1,6 @@
...
- * Copyright (c) 2023, Nokia
- * All rights reserved.
- *
- * SPDX-License-Identifier:     BSD-3-Clause
+/* SPDX-License-Identifier: BSD-3-Clause
+ * Copyright (c) 2017-2018 Linaro Limited
+ * Copyright (c) 2023 Nokia
  */
 
 /* SPDX-License-Identifier: BSD-3-Clause
```
TuomasTaipale, line 6: Intel copyright could be moved to the list above as well.

platform/linux-generic/pktio/loop.c
```diff
@@ -1,8 +1,6 @@
-/* Copyright (c) 2013-2018, Linaro Limited
- * Copyright (c) 2013-2023, Nokia Solutions and Networks
- * All rights reserved.
- *
- * SPDX-License-Identifier:     BSD-3-Clause
+/* SPDX-License-Identifier: BSD-3-Clause
+ * Copyright (c) 2013-2018 Linaro Limited
+ * Copyright (c) 2013-2023 Nokia Solutions and Networks
```
TuomasTaipale, line 3: I guess the company is still legally "Nokia Solutions and Networks", but should "Nokia" be used anyways? Or separate the two periods?

(There are a couple of places.)

platform/linux-generic/arch/x86/cpu_flags.c
MatiasElo: It could, but is there a general convention what should be done in this kind of cases? At least one motivation for keeping the licenses separate is that it's easier to notice that the file contains copied code.

MatiasElo: Combined copyright messages and listed DPDK source files in V2.

platform/linux-generic/pktio/loop.c
MatiasElo: Agreed in a call that from now on only "Nokia" copyright will be used by us. No changes in this PR.

PR #2033, opened 2024-03-11
[PATCH v4] misc: update copyright format

MatiasElo:


m4/odp_check_flag.m4
```diff
@@ -1,7 +1,5 @@
-# Copyright (c) 2018, Linaro Limited
-# All rights reserved.
-#
-# SPDX-License-Identifier:     BSD-3-Clause
+# SPDX-License-Identifier: BSD-3-Clause
```
TuomasTaipale, line 1: There is a bunch of `m4` files that don't have a copyright listed at all, I guess we should tag them as well?

MatiasElo: Added copyrights in V2.

MatiasElo:
V2:
* Added missing copyrights to m4 files
* Updated DPDK license in `test/performance/dummy_crc.h`

m4/odp_check_flag.m4
TuomasTaipale: `/platform/linux-generic/m4/` hides a few additional `m4`s.

MatiasElo: Yeah, I was on the fence about these, since many of them are pretty trivial, but it makes sense to have license text on all files. Added in V3.

TuomasTaipale (approved):

MatiasElo:
V4:
- Added copyrights to helper m4 files

PR #2034, opened 2024-03-13
test: packet_gen: add support for transmitting TCP packets

MatiasElo:
Add support for transmitting TCP packets with packet_gen application. The application doesn't implement actual TCP protocol (handshake, flow control etc.), but simply transmits packets with valid TCP headers at a selected rate.


test/performance/odp_packet_gen.c
```diff
@@ -252,15 +260,18 @@ static void print_usage(void)
 	       "                              num_tx * burst_size * bursts * (10^9 / gap)\n"
 	       "  -s, --ipv4_src            IPv4 source address. Default: 192.168.0.1\n"
 	       "  -d, --ipv4_dst            IPv4 destination address. Default: 192.168.0.2\n"
-	       "  -o, --udp_src             UDP source port. Default: 10000\n"
-	       "  -p, --udp_dst             UDP destination port. Default: 20000\n"
+	       "  -o, --src_port            Source port. Default: 10000\n"
+	       "  -p, --dst_port            Destination port. Default: 20000\n"
+	       "  -N, --proto               L4 protocol. Default: 0\n"
```
TuomasTaipale, line 265: With additional protocols, might be useful to somehow bind different features to the protocols that actually support them. E.g., port number increasing only for protocols that use ports etc. Would save option namespace and probably improve usability. Anyways, out of the context of this PR, maybe a future improvement.

test/performance/odp_packet_gen.c
```diff
@@ -706,17 +722,25 @@ static int parse_options(int argc, char *argv[], test_global_t *global)
...
+	if (test_options->c_mode.dst_port && num_tx_pkt % test_options->c_mode.dst_port)
+		ODPH_ERR("\nWARNING: Transmit packet count is not evenly divisible by destination port count.\n\n");
 
-	if (test_options->c_mode.udp_src &&
-	    num_tx_pkt % test_options->c_mode.udp_src)
-		ODPH_ERR("\nWARNING: Transmit packet count is not evenly divisible by UDP source port count.\n\n");
+	if (test_options->c_mode.src_port && num_tx_pkt % test_options->c_mode.src_port)
+		ODPH_ERR("\nWARNING: Transmit packet count is not evenly divisible by source port count.\n\n");
+
+	if (test_options->l4_proto > L4_PROTO_TCP) {
```
TuomasTaipale, line 731: Probably safer to check the values individually and if none match then error.

test/performance/odp_packet_gen.c
```diff
@@ -1451,63 +1479,96 @@ static int init_packets(test_global_t *global, int pktio,
...
+			odph_tcphdr_t *tcp = (odph_tcphdr_t *)u8;
+
+			memset(tcp, 0, ODPH_TCPHDR_LEN);
+			tcp->src_port = odp_cpu_to_be_16(src_port);
+			tcp->dst_port = odp_cpu_to_be_16(dst_port);
+			tcp->seq_no   = odp_cpu_to_be_32(tcp_seqnum);
+			tcp->ack_no   = odp_cpu_to_be_32(0x12345678);
+			tcp->window   = odp_cpu_to_be_16(0x4000);
+			tcp->hl       = 5;
+			tcp->ack      = 1;
```
TuomasTaipale, line 1499: If it doesn't matter what kind of TCP packets program is generating, I guess they could be `SYN` packets, would avoid faking any session state maintenance.

(Program would be basically `SYN` flooding, but does it matter?)

MatiasElo: The target was to create "valid enough" packets for Wireshark, so in the next phase it can be used to validate LSO transmitted packets. Setting the SYN flag seems to make Wireshark unhappy: "TCP Port numbers reused".

TuomasTaipale (approved):

PR #2035, opened 2024-03-13
[PATCH v5] api: clarify and limit name lengths

JereLeppanen:


include/odp/api/spec/classification.h
```diff
@@ -49,7 +49,7 @@ extern "C" {
 
 /**
  * @def ODP_COS_NAME_LEN
- * Maximum class of service name length in chars including null char
+ * Maximum class of service name length, including the null character.
```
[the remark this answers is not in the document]
JereLeppanen: Fixed in v2.

include/odp/api/spec/classification.h
```diff
@@ -719,7 +719,7 @@ int odp_cls_capability(odp_cls_capability_t *capability);
  * values.
  *
  * @param name         Name of the class-of-service or NULL. Maximum string
- *                     length is ODP_COS_NAME_LEN.
+ *                     length is ODP_COS_NAME_LEN, including the null character.
```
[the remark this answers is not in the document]
JereLeppanen: It doesn't have to be there, but the reason why it's there is because "string length" usually doesn't include the null character. So I think it's prudent to remind the API reader in all of these places that the null character is included. I didn't change these in v2.

helper/include/odp/helper/table.h
```diff
@@ -91,7 +91,11 @@ extern "C" {
  * @def ODPH_TABLE_NAME_LEN
  * Maximum length of table name, including the null character
  */
-#define ODPH_TABLE_NAME_LEN      32
+/*
+ * Minimum of ODP_SHM_NAME_LEN, (ODP_QUEUE_NAME_LEN - 3),
+ * (ODP_POOL_NAME_LEN - 8).
+ */
+#define ODPH_TABLE_NAME_LEN 24
```
TuomasTaipale, line 98: I guess you could do the calculations explicitly here, also elaborating the meaning of "3" and "8" could be helpful. Can you actually infer the numbers based on some constants so they don't have to be magic numbers?

JereLeppanen: Those constants don't currently exist, and even if they did, they'd probably be defined in the headers of specific table types, which we probably wouldn't want to include from here. I.e. there's an awkward dependency from this generic table name length to name lengths in different table types. In this commit, that dependency is implied by this comment, I didn't want to add an actual compile time dependency.

Just by searching for ODPH_TABLE_NAME_LEN, it's not hard to find that there's usage such as ODPH_TABLE_NAME_LEN + 3 in cuckootable and ODPH_TABLE_NAME_LEN + 8 in iplookuptable, where the magic numbers come from. But I'll try to improve the comment.

This header is anyway being removed, these are just minimal changes to fix the known potential problems before that happens.

JereLeppanen: Comment revised in v3. (I'm not sure whether it's better or worse.)

TuomasTaipale: It's pretty informative now, resolving this.

TuomasTaipale (approved):

JereLeppanen:
v4:
- Rebase.
- Add review tags.

JereLeppanen:
v5:
- Rebase.
- Add review tags.

PR #2037, opened 2024-03-18
[PATCH v4] linux-gen: avoid some GCC thread sanitizer warnings

JereLeppanen:


platform/linux-generic/odp_thread.c
```diff
@@ -284,12 +284,22 @@ int odp_thread_worker_count_max(void)
 
 int odp_thrmask_worker(odp_thrmask_t *mask)
 {
+	uint32_t num_worker;
+
+	odp_spinlock_lock(&thread_globals->lock);
 	odp_thrmask_copy(mask, &thread_globals->worker);
-	return thread_globals->num_worker;
+	num_worker = thread_globals->num_worker;
```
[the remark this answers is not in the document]
MatiasElo: Yes, these are not really fast path APIs and the overhead from using atomic variables should be negligible anyway. So I'd favor fixing also the variables.

platform/linux-generic/odp_traffic_mngr.c
```diff
@@ -2431,7 +2428,7 @@ static void signal_request(void)
 
 	serving = odp_atomic_load_u64(&tm_glb->currently_serving_cnt);
 	while (serving != request_num) {
-		busy_wait(100);
+		busy_wait();
```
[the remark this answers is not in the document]
MatiasElo: I'd vote for replacing `busy_wait()` with `odp_cpu_pause()`. I slight mitigation for memory stress could be to store the affected atomic variables on different cache lines.

include/odp/api/abi-default/barrier.h
```diff
@@ -23,7 +23,7 @@ extern "C" {
  * ODP thread synchronization barrier
  */
 struct odp_barrier_s {
-	uint32_t         count;  /**< Thread count */
+	odp_atomic_u32_t count;  /**< Thread count */
```
JannePeltonen, line 26: This looks a bit dubious me. The count is written only in odp_barrier_init() and no thread is supposed to call odp_barrier_wait() (which reads the count) before odp_barrier_init() has completed and has become visible in other threads.

It appears that the problem caught by TSN is in the TM code in odp_tm_destroy() which calls odp_barrier_init() and then does odp_atomic_inc_u64(&tm_system->destroying) to let another thread to proceed to odp_barrier_wait(). But since the odp_atomic_inc_u64() is a relaxed atomic, we do not get proper ordering between the accesses to the barrier variables in the two threads and end up with the data race.

JannePeltonen (changes requested):

include/odp/api/abi-default/barrier.h
JereLeppanen: I failed to notice that thread sanitizer's lack of support for memory barriers affects this case as well.

`warning: ‘atomic_thread_fence’ is not supported with ‘-fsanitize=thread’ [-Wtsan]`

So the data race warning could even be a false positive.

In v2, I've modified this commit a little, and changed the rationale. Or we could just forget about this commit and leave the implementation as it is.

JereLeppanen:
All comments processed in v2.

platform/linux-generic/odp_thread.c
```diff
@@ -32,10 +33,10 @@ typedef struct {
...
 	};
 
-	uint32_t       num;
-	uint32_t       num_worker;
-	uint32_t       num_control;
-	uint32_t       num_max;
+	odp_atomic_u32_t num;
+	odp_atomic_u32_t num_worker;
+	odp_atomic_u32_t num_control;
+	odp_atomic_u32_t num_max;
```
MatiasElo, line 39: Looks like `num_max` doesn't have to be atomic, since it's only modified in global init. Probably not much overhead but it's a bit confusing now when reading the code.

include/odp/api/abi-default/barrier.h
JannePeltonen: The warning is not a false positive. There is an actual data race as I described:

```
Thread 1 (thread calling odp_tm_destroy()):
	/* this writes barrier.count */
	odp_barrier_init(&tm_system->tm_system_destroy_barrier, 2);
	odp_atomic_inc_u64(&tm_system->destroying);

Thread 2 (TM system thread):
	while (!odp_atomic_load_u64(&tm_system->destroying))
		;
	/* this reads barrier.count */
	odp_barrier_wait(&tm_system->tm_system_destroy_barrier);
```
Since writing and reading tm_system->destroying is done using relaxed atomics, there is no synchronization between the threads to ensure that the write to barrier.count in thread 1 happens before the read of it in thread 2.

Changing accesses of barrier.count to relaxed atomic operations would obviously get rid of the data race but would leave the code incorrect since nothing would guarantee that the value written in barrier.count by thread 1 would be visible in thread 2 by the time it is read there.

IMHO, odp_barrier_wait() must not be called before odp_barrier_init() has completed, with the meaning that there must be inter-thread synchronization and thus proper before-after relation between (the side effects of) the init and the wait. This is also what the current barrier code seems to assume. With this in mind, the usage of the barrier in the TM code is incorrect since it does not ensure correct memory ordering.

The problem could be fixed by changing the writing and reading of tm_system->destroying to be synchronization operations (i.e. not relaxed atomics). This might fix other potential problems caused by the relaxed memory ordering. The write to tm_system->destroying in odp_tm_destroy() could be changed to a store-release (there is no reason for it to be an atomic inc like now) and the reads to load-acquire operations.

Another approach to fix the data race at hand would be to just add a release fence in odp_barrier_init(). Then the release fence there and acquire fence in odp_barrier_wait() would synchronize through the tm_system->destroying atomic accesses, but I think that it would make the barrier implementation in linux-gen a bit different what the API (in my interpretation) guarantees (i.e. linux-gen barrier could be used in the way TM does, but in general with other ODP implementations it would not be right). I think it is better to just require that the API user must synchronize so that side effects of init functions are visible before a feature is used (which typically happens unless one uses relaxed atomic flags like this to communicate between threads).

JereLeppanen: In v3, changes to barrier implementation removed, acquire and release on tm_system.destroying.

platform/linux-generic/odp_thread.c
JereLeppanen: Not atomic in v3.

MatiasElo (approved):

JereLeppanen:
v4:
- Use acquire and relese semantics on is_idle.
- Minor tweaking of some commit comments.
- Add review tags.

JannePeltonen (approved):

PR #2041, opened 2024-03-22
[PATCH v3] test: l2fwd: move pktio print call

MatiasElo:
Move odp_pktio_print() call to a later phase when all pktio configurations have been completed. This may enable implementations to print more information.

### DIFF
--- a/test/performance/odp_l2fwd.c
+++ b/test/performance/odp_l2fwd.c
@@ -1085,9 +1085,6 @@ static int create_pktio(const char *dev, int idx, int num_rx, int num_tx, odp_po
 		return -1;
 	}
 
-	if (gbl_args->appl.verbose)
-		odp_pktio_print(pktio);
-
 	if (odp_pktio_capability(pktio, &pktio_capa)) {
 		ODPH_ERR("Pktio capability query failed: %s\n", dev);
 		return -1;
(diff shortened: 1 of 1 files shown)


test/performance/odp_l2fwd.c
```diff
@@ -1280,6 +1277,9 @@ static int create_pktio(const char *dev, int idx, int num_rx, int num_tx, odp_po
 	       "%02x:%02x:%02x:%02x:%02x:%02x\n", dev, info.drv_name, num_rx, num_tx,
 	       addr[0], addr[1], addr[2], addr[3], addr[4], addr[5]);
 
+	if (gbl_args->appl.verbose)
+		odp_pktio_print(pktio);
+
```
TuomasTaipale, line 1282: Could still be printed before the above /dev/drv/queue/mac summary to preserve the previous ordering?

TuomasTaipale (approved):
One minor comment, otherwise approved.

test/performance/odp_l2fwd.c
MatiasElo: I prefer this order because now `printf("  dev: %s, drv: %s...` print works as heading  for `odp_pktio_print()` output, which may be quite long and the interface name may not be obvious.

PR #2044, opened 2024-03-28
[PATCH v5] fix issues reported by GCC undefined sanitizer

JereLeppanen:


JereLeppanen:
v2:
- Rebase.

platform/linux-generic/odp_pool.c
```diff
@@ -135,6 +135,9 @@ static void cache_flush(pool_cache_t *cache, pool_t *pool)
 	ring_ptr_t *ring;
 	uint32_t mask;
 
+	if (!pool->ring)
```
JannePeltonen, line 138: Ok, but if _odp_pool_create() fails halfway through, pool->ring may be set while the pool (including pool->ring->hdr ?) is not fully initialized.

I did not look in detail but the code seems a bit dubios in this regard.

platform/linux-generic/odp_packet.c
```diff
@@ -699,6 +699,9 @@ void odp_packet_free(odp_packet_t pkt)
 
 static inline void packet_free_multi_ev(const odp_packet_t pkt[], int num, _odp_ev_id_t id)
 {
+	if (!num)
```
JannePeltonen, line 702: This should use odp_unlikely(). Or maybe you could just have pkt_hdrs[num + 1] and avoid the branch altogether?

ODP API seems to allow odp_packet_free_multi() for zero packets, but I wonder if the fact that you bumped into such a free is a symptom of a bug in the calling code.

platform/linux-generic/odp_hash_crc_gen.c
```diff
@@ -87,7 +87,7 @@ static inline uint32_t reflect_u32(uint32_t u32)
 	u8[2] = reflect_u8((u32 & 0x0000ff00u) >> 8);
 	u8[3] = reflect_u8(u32 & 0xffu);
 
-	return (u8[3] << 24) | (u8[2] << 16) | (u8[1] << 8) | u8[0];
+	return ((uint32_t)u8[3] << 24) | (u8[2] << 16) | (u8[1] << 8) | u8[0];
```
JannePeltonen, line 90: This still makes the assumption that int is larger than 24 bits. Maybe just change the type of u8 from uint8_t to uint32_t (an added bonus: no casting needed). reflect_u24 has the same problem. Similarly, reflect_u16() assumes that int is larger than 16 bits. IOW, this change avoids undefined behaviour only with some additional non-obvious assumptions.

platform/linux-generic/odp_parse.c
```diff
@@ -40,16 +40,16 @@ uint16_t _odp_parse_eth(packet_parser_t *prs, const uint8_t **parseptr,
 		input_flags.jumbo = 1;
 
 	/* Handle Ethernet broadcast/multicast addresses */
-	macaddr0 = odp_be_to_cpu_16(*((const uint16_t *)(const void *)eth));
+	macaddr0 = odp_be_to_cpu_16(*((const odp_una_u16_t *)(const void *)eth));
```
JannePeltonen, line 43: Is there some benefit in the intemediate casting to (const void *)?

platform/linux-generic/odp_pkt_queue.c
```diff
@@ -106,7 +106,7 @@ static int pkt_queue_free_list_add(queue_pool_t *pool,
 		queue_blks = region_desc->queue_blks;
 		if (!queue_blks) {
 			malloc_len = num_blks * sizeof(queue_blk_t);
-			queue_blks = malloc(malloc_len);
+			queue_blks = aligned_alloc(sizeof(queue_blk_t), malloc_len);
```
JannePeltonen, line 109: This could use alignof instead of sizeof.

platform/linux-generic/odp_random_std.c
```diff
@@ -58,15 +58,15 @@ static int32_t _random_data(uint8_t *buf, uint32_t len, uint64_t *seed)
 	}
 
 	for (uint32_t i = 0; i < len / 4; i++) {
-		*(uint32_t *)(uintptr_t)buf = xorshift64s32(seed);
+		*(odp_una_u32_t *)(uintptr_t)buf = xorshift64s32(seed);
```
JannePeltonen, line 61: Is there some benefit in the intermediate casting to uintptr_t?

test/performance/odp_pktio_ordered.c
```diff
@@ -188,13 +188,13 @@ typedef union ODP_ALIGNED_CACHE {
...
  */
 typedef struct {
-	int32_t src_ip;
-	int32_t dst_ip;
-	int16_t src_port;
-	int16_t dst_port;
-	int8_t  proto;
-	int8_t  pad0;
-	int16_t pad1;
+	uint32_t src_ip;
```
JannePeltonen, line 191: Fixing the types in the ipv4_tuple fixes also potential problems (implementation defined value or signal raised) in calc_flow_idx(). Maybe make some remark in the commit message?

test/performance/odp_packet_gen.c
```diff
@@ -810,8 +810,8 @@ static int open_pktios(test_global_t *global)
 	uint32_t num_pkt = test_options->num_pkt;
 	uint32_t pkt_len = test_options->use_rand_pkt_len ?
 				test_options->rand_pkt_len_max : test_options->pkt_len;
-	odp_pktout_queue_t pktout[num_tx];
-	odp_pktin_queue_t pktin[num_rx];
+	odp_pktout_queue_t pktout[MAX_THREADS];
+	odp_pktin_queue_t pktin[MAX_THREADS];
```
JannePeltonen, line 814: Since the pktout and pktin arrays are used only at the bottom of the function in the two if-statements that check num_tx and num_rx, the definitions could just be moved there. I think that would make the code just a hair clearer too as the scope of the variables would get reduced.

example/ipsec_crypto/odp_ipsec_misc.h
```diff
@@ -167,18 +167,18 @@ char *ipv4_addr_str(char *b, uint32_t addr)
...
 int parse_ipv4_string(char *ipaddress, uint32_t *addr, uint32_t *mask)
 {
-	int b[4];
+	uint32_t b[4];
 	int qualifier = 32;
 	int converted;
 
 	if (strchr(ipaddress, '/')) {
-		converted = sscanf(ipaddress, "%d.%d.%d.%d/%d",
+		converted = sscanf(ipaddress, "%u.%u.%u.%u/%d",
```
JannePeltonen, line 175: It would be more correct to use SCNu32 conversion specifier instead of 'u', to avoid problems in the very unlikely case that int is not 4 bytes. Or maybe just make the array element type unsigned int instead of uint32_t.

helper/hashtable.c
```diff
@@ -60,6 +60,14 @@ typedef struct {
 	char name[ODPH_TABLE_NAME_LEN]; /**< table name */
 } odph_hash_table_imp;
 
+static uint32_t node_size(const odph_hash_table_imp *tbl)
+{
+	const uint32_t mask = sizeof(void *) - 1;
```
JannePeltonen, line 65: This assumes that sizeof(void *) is a power of two, which is not guaranteed. Use of alignof instead of sizeof would be more accurate and would have the added benefit that all valid aligments are guaranteed to be powers of two.

JannePeltonen (changes requested):

platform/linux-generic/odp_packet.c
JereLeppanen: The backtrace is
```
    #0 0x560f7aa63698 in packet_free_multi_ev /odp/platform/linux-generic/odp_packet.c:702
    #1 0x560f7aa63932 in odp_packet_free_multi /odp/platform/linux-generic/odp_packet.c:732
    #2 0x560f7aa0974d in free_pending_packets /odp/test/performance/odp_dmafwd.c:1128
    #3 0x560f7aa0a148 in process_packets /odp/test/performance/odp_dmafwd.c:1201
    #4 0x560f7aa0c86a in run_thread /odp/helper/threads.c:69
```
Which looks legitimate. When odp_dmafwd is exiting, it frees pending packets. If there aren't any for a given interface, it calls odp_packet_free_multi() with num=0, which is fine.

platform/linux-generic/odp_pool.c
JereLeppanen: Added commit `linux-gen: pool: free ring shm if pool creation fails` in v3.

platform/linux-generic/odp_parse.c
JereLeppanen: I don't think so. Removed in v3.

JereLeppanen:
All comments processed in v3.

platform/linux-generic/odp_random_std.c
```diff
@@ -58,21 +58,21 @@ static int32_t _random_data(uint8_t *buf, uint32_t len, uint64_t *seed)
 	}
 
 	for (uint32_t i = 0; i < len / 4; i++) {
-		*(uint32_t *)(uintptr_t)buf = xorshift64s32(seed);
+		*(odp_una_u32_t *)buf = xorshift64s32(seed);
```
JannePeltonen, line 61: Not assuming alignment here can throw away the work done to align the pointer if _ODP_UNALIGNED is zero. If the CPU does not support unaligned accesses, the compiler may have to resort to using multiple store instructions here, even though it is not really necessary.

I would suggest doing something like this here:

```
for (uint32_t i = 0; i < len / 4; i++) {
        if (_ODP_UNALIGNED)
                *(uint32_t *)(uintptr_t)buf = xorshift64s32(seed);
        else
                *(odp_una_u32_t *)buf = xorshift64s32(seed);
        buf += 4;
}
```

JannePeltonen:
Minor improvement proposal to _random_data(). With that, Reviewed-by: Janne Peltonen <janne.peltonen@nokia.com>

platform/linux-generic/odp_random_std.c
JereLeppanen: I did it a bit differently, in v4. But maybe there's no real need to optimize the no-unaligned case, and we could just remove the whole thing?

JannePeltonen: Maybe, but since it is already there and you fixed the alignment issues, I think it can as well stay.

JannePeltonen (approved):

PR #2045, opened 2024-04-08
[PATCH v4] test: dyn_workers: improve command granularity

TuomasTaipale:
User commands in both interactive and non-interactive modes now take an additional index parameter which is used to launch or remove workers to/from a specific index. The index is relative to the given CPU masks, e.g. index 0 would launch a worker on CPU 0x1 with CPU mask 0x111 and index 2 would launch a worker to CPU 0x100 with the same mask.

Additionally, add a new print after worker additions which lists all active workers on the particular ODP process.

On top of #2038 

v3:
- Rebased
- Added reviewed-by tags

### DIFF
--- a/test/miscellaneous/Makefile.am
+++ b/test/miscellaneous/Makefile.am
@@ -6,6 +6,7 @@ if test_cpp
 bin_PROGRAMS +=  odp_api_from_cpp
 endif
 
+odp_dyn_workers_CFLAGS = $(AM_CFLAGS) -Wno-format-nonliteral
 odp_dyn_workers_SOURCES = odp_dyn_workers.c
 odp_api_from_cpp_SOURCES = odp_api_from_cpp.cpp
 
--- a/test/miscellaneous/odp_dyn_workers.c
+++ b/test/miscellaneous/odp_dyn_workers.c
@@ -73,7 +73,7 @@ enum {
 	CONN_ERR = -1,
 	PEER_ERR,
 	CMD_NOK,
-	CMD_STATS,
+	CMD_SUMMARY,
 	CMD_OK
 };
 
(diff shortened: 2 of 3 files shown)


test/miscellaneous/odp_dyn_workers.c
```diff
@@ -1163,6 +1220,9 @@ static odp_bool_t run_global(global_config_t *config)
 				func_ret = false;
 			}
```
MatiasElo, line 1221: For uniformity could add `continue` here.

test/miscellaneous/odp_dyn_workers.c
```diff
@@ -971,10 +1135,30 @@ static odp_bool_t check_summary(const summary_t *summary)
 	return true;
 }
 
+static void dump_result(int socket, pid_t pid)
```
MatiasElo, line 1138: Optional, but if it's simple to implement it could be handy to have a separate CLI command for printing active workers from all processes (e.g. `PRINT_WORKERS`).

test/miscellaneous/Makefile.am
```diff
@@ -6,6 +6,7 @@ if test_cpp
 bin_PROGRAMS +=  odp_api_from_cpp
 endif
 
+odp_dyn_workers_CFLAGS = $(AM_CFLAGS) -Wno-format-nonliteral
```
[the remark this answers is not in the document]
TuomasTaipale: `#pragma`s are slightly ugly so decided to use a flag instead as it anyways affects only this tester.

test/miscellaneous/odp_dyn_workers.c
TuomasTaipale: V2.

test/miscellaneous/odp_dyn_workers.c
TuomasTaipale: This would change some assumptions in the code so requires a few changes. Perhaps a future addition in case the new addition dumps are deemed inadequate for some reason.

MatiasElo: Okay, can be implemented in the future if deemed useful.

test/miscellaneous/Makefile.am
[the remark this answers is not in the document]
MatiasElo: Ok, since all the changes are included and reviewed already in this PR, #2038 can be closed.

MatiasElo (approved):

PR #2047, opened 2024-04-11
[PATCH v2] linux-gen: timer: don't use sched_yield()

JereLeppanen:
Replace all usage of sched_yield() with odp_cpu_pause(). The man page states that "sched_yield() is intended for use with real-time scheduling policies [...] Use of sched_yield() with nondeterministic scheduling policies such as SCHED_OTHER is unspecified and very likely means your application design is broken."


psavol (approved):

JereLeppanen:
v2:
- Rebase.
- Add review tags.

PR #2048, opened 2024-04-12
[PATCH v2] helper: threads: improve thread joining

TuomasTaipale:
When joining threads, attempt to join all threads and only interrupt if thread/process join/PID waiting itself fails, in this case return a failure.

v2:
- Rebased
- Elaborated on the change in `odph_thread_join()` specification

### DIFF
--- a/helper/threads.c
+++ b/helper/threads.c
@@ -158,18 +158,15 @@ static int wait_process(odph_thread_t *thread)
 
 	/* Examine the child process' termination status */
 	if (WIFEXITED(status) &&
-	    WEXITSTATUS(status) != EXIT_SUCCESS) {
+	    WEXITSTATUS(status) != EXIT_SUCCESS)
 		ODPH_ERR("Child exit status:%d (pid:%d)\n",
 			 WEXITSTATUS(status), (int)pid);
-		return -1;
-	}
 
 	if (WIFSIGNALED(status)) {
 		int signo = WTERMSIG(status);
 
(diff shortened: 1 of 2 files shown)


helper/threads.c
```diff
@@ -243,19 +240,15 @@ static int wait_pthread(odph_thread_t *thread)
 		return -1;
 	}
 
-	if (thread_ret) {
+	if (thread_ret)
 		ODPH_ERR("Bad exit status cpu #%i %p\n",
 			 thread->cpu, thread_ret);
-		return -1;
```
MatiasElo, line 249: On a glance there seems to be couple places where failures could go unnoticed in the CI after this change:
```
example/l2fwd_simple/odp_l2fwd_simple.c
helper/test/cli.c
helper/test/odpthreads.c
test/common/odp_cunit_common.c
test/performance/odp_l2fwd.c
test/performance/odp_random.c
test/validation/api/time/time.c
```

TuomasTaipale: Yes, didn't touch them until it's decided if the new proposed behavior is OK.

TuomasTaipale: I guess we could just swap to use the new version proposed in #2050 in these sort of places that don't have other means to resolve worker thread errors than the exit status.

TuomasTaipale: As discussed offline, closing this for now and leaving the current behavior as is. Potentially continued in a future release if necessary. New and improved join function in #2050.

PR #2049, opened 2024-04-17
[PATCH v3] linux-gen: ml: don't require non-zero max size for dynamic dimension

JereLeppanen:
In check_io_shape(), called from odp_ml_model_create(), don't require a non-zero max size for dynamic dimensions. This avoids an error when loading the batch_add.onnx model with model_explorer. Onnx doesn't have minimums or maximums for dynamic dimensions.

Instead, check that batch type shapes have exactly one dynamic dimension.


psavol (approved):

JereLeppanen:
v3:
- Rebase.
- Add review tags.

PR #2050, opened 2024-04-18
[PATCH v6] helper: threads: add new join function

TuomasTaipale:
Add new join function which is similar to the existing `odph_thread_join()` but optionally outputs results of joined threads and interrupts only if the actual join operation fails for some thread.

v3:
- Rebased
- Switched to utilize new `odph_thread_join_res()` in cases where result of the join operation was checked

v4:
- Initialized result structure in cases where LTO wrongly thinks it might be uninitialized

v5:
- Matias' comments

v6:
- Matias' comments
- Bumped copyright years
- Added reviewed-by tags

### DIFF
--- a/helper/include/odp/helper/threads.h
+++ b/helper/include/odp/helper/threads.h
@@ -1,6 +1,6 @@
 /* SPDX-License-Identifier: BSD-3-Clause
  * Copyright (c) 2013-2018 Linaro Limited
- * Copyright (c) 2019-2021 Nokia
+ * Copyright (c) 2019-2024 Nokia
  */
 
 
@@ -191,6 +191,20 @@ typedef struct {
 
 } odph_thread_common_param_t;
 
+/** Thread join result */
+typedef struct {
+	/** Exit caused by signal */
+	odp_bool_t is_sig;
+
+	/**
+	 * Exit status of the joined thread/process
+	 *
+	 * If 'is_sig' is true, then this is the signal number that caused
+	 * process exit. Otherwise status of the exited thread/process.
+	 */
+	int ret;
+} odph_thread_join_result_t;
+
 /**
  * Initialize thread params
  *
@@ -261,15 +275,39 @@ int odph_thread_create(odph_thread_t thread[],
  * A function call may be used to wait any number of launched threads to exit.
  * A particular thread may be waited only once.
  *
+ * Threads are joined in the order they are in 'thread' table. Returns on the
+ * first non-zero exit status or other failure.
+ *
  * @param thread        Table of threads to exit
  * @param num           Number of threads to exit
  *
- * @return Number of threads exited
+ * @return Number of threads successfully joined with zero exit status
+ *         (0 ... num)
  * @retval -1  On failure
  *
  * @see odph_thread_create()
  */
 int odph_thread_join(odph_thread_t thread[], int num);
+
+/**
+ * Wait previously launched threads to exit
+ *
+ * Similar to odph_thread_join() but outputs results of joined threads and
+ * stops only if the actual join operation fails for some thread. Threads are
+ * joined in the order they are in 'thread' table. Returns number of threads
+ * successfully joined and writes respective exit statuses into the 'res'
+ * table.
+ *
+ * @param thread        Table of threads to exit
+ * @param[out] res      Table for result output
+ * @param num           Number of threads to exit and results to output
+ *
+ * @return Number of threads successfully joined (0 ... num)
+ * @retval -1  On failure
+ *
+ * @see odph_thread_create()
+ */
+int odph_thread_join_result(odph_thread_t thread[], odph_thread_join_result_t res[], int num);
 
 /**
  * Set CPU affinity of the current odp thread
--- a/helper/threads.c
+++ b/helper/threads.c
@@ -1,6 +1,6 @@
 /* SPDX-License-Identifier: BSD-3-Clause
  * Copyright (c) 2013-2018 Linaro Limited
- * Copyright (c) 2019-2022 Nokia
+ * Copyright (c) 2019-2024 Nokia
  */
 
 #ifndef _GNU_SOURCE
@@ -144,10 +144,10 @@ static int create_process(odph_thread_t *thread, int cpu, uint64_t stack_size)
 /*
  * Wait single process to exit
  */
-static int wait_process(odph_thread_t *thread)
+static int wait_process(odph_thread_t *thread, odph_thread_join_result_t *res)
 {
 	pid_t pid;
-	int status = 0;
+	int status = 0, estatus;
 
 	pid = waitpid(thread->proc.pid, &status, 0);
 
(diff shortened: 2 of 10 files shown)


helper/include/odp/helper/threads.h
```diff
@@ -191,6 +191,33 @@ typedef struct {
...
+
+			/**
+			 * Return value of the exited process
+			 *
+			 * If 'is_sig' is true, then this is the signal number
+			 * that caused the process exit. Otherwise status of
+			 * the exited process.
+			 */
+			int ret;
+		} proc;
```
MatiasElo, line 217: `process` would not be that long either, but a matter of taste.

helper/include/odp/helper/threads.h
```diff
@@ -191,6 +191,33 @@ typedef struct {
 
 } odph_thread_common_param_t;
 
+/** Thread join result type */
+typedef struct {
+	/** Variant field mappings for thread/process modes */
+	union {
```
MatiasElo, line 197: The usage of this struct seems quite complex in the later commits. Would it be possible to forgo this union and use the same `ret` for both threads and processes?  So it would be enough for an application to check `odph_thread_join_res_t.ret != 0`? Assuming there is no signal with value zero.

TuomasTaipale: It follows the same style as launching threads. In any case, not having some indication about the signaled condition would make the return value a bit useless in debugging.

helper/include/odp/helper/threads.h
```diff
@@ -191,6 +191,33 @@ typedef struct {
...
+			 * Return value of the exited process
+			 *
+			 * If 'is_sig' is true, then this is the signal number
+			 * that caused the process exit. Otherwise status of
+			 * the exited process.
+			 */
+			int ret;
+		} proc;
+	};
+} odph_thread_join_res_t;
```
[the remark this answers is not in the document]
TuomasTaipale: In that case, I guess function name should be `*_join_result()` then?

helper/include/odp/helper/threads.h
```diff
@@ -191,6 +191,33 @@ typedef struct {
...
+		struct {
+			/** Exit caused by signal */
+			odp_bool_t is_sig;
+
+			/**
+			 * Return value of the exited process
+			 *
+			 * If 'is_sig' is true, then this is the signal number
+			 * that caused the process exit. Otherwise status of
+			 * the exited process.
```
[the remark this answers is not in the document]
TuomasTaipale: Not sure I understand this, the "subject" lines are similar, `proc.ret` has some additional info because of `proc.is_sig`.

helper/include/odp/helper/threads.h
MatiasElo: The same information would still be provided.
```
/** Thread join result */
typedef struct {
	/**
	 * Return value of the exited thread/process
	 *
	 * If 'is_sig' is true, then this is the signal number
	 * that caused the process exit. Otherwise exit status of the
	 * thread/process.
	 */
	int ret;

	/** Exit caused by signal */
	odp_bool_t is_sig;

} odph_thread_join_result_t;
```

test/common/odp_cunit_common.c
```diff
@@ -268,6 +285,7 @@ static int tests_global_init(odp_instance_t *inst)
 
 	odp_init_param_init(&init_param);
 	init_param.mem_model = helper_options.mem_model;
+	mem_model = init_param.mem_model;
```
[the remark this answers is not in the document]
TuomasTaipale: Didn't realize this. Pretty much have to ditch the `union` from the result now.

TuomasTaipale:
Comments addressed in v5.

helper/threads.c
```diff
@@ -407,6 +419,34 @@ int odph_thread_join(odph_thread_t thread[], int num)
...
+			ODPH_DBG("Thread (i:%i) not started.\n", i);
+			continue;
+		}
+
+		if (thread[i].start_args.mem_model == ODP_MEM_MODEL_THREAD) {
+			if (wait_pthread(&thread[i], &res[i]))
+				return -1;
+		} else {
+			if (wait_process(&thread[i], &res[i]))
+				return -1;
```
MatiasElo, line 440: Passing `NULL` to `res` causes segfault.

helper/include/odp/helper/threads.h
```diff
@@ -271,6 +285,24 @@ int odph_thread_create(odph_thread_t thread[],
...
+/**
+ * Wait previously launched threads to exit
+ *
+ * Similar to odph_thread_join() but optionally outputs results of joined
+ * threads and interrupts only if the actual join operation fails for some
+ * thread.
+ *
+ * @param thread        Table of threads to exit
+ * @param[out] res      Table for result output, or NULL when not used
+ * @param num           Number of threads to exit
```
MatiasElo, line 297: Could mention both tables have this many elements (if given)

helper/threads.c
MatiasElo, line 440: More informative for application to return `-1` only when no threads joined and otherwise `i`?

helper/threads.c
```diff
@@ -407,6 +419,34 @@ int odph_thread_join(odph_thread_t thread[], int num)
...
+	odph_thread_start_args_t *start_args;
+	int num_exited = 0;
+
+	for (int i = 0; i < num; i++) {
+		start_args = &thread[i].start_args;
+
+		if (start_args->status != STARTED) {
+			ODPH_DBG("Thread (i:%i) not started.\n", i);
+			continue;
+		}
```
MatiasElo, line 433: This may cause usage of uninitialized variable warnings as some `res` elements are not touched. Also, thinking whether this should be also considered as a failure -> break/return.

helper/threads.c
```diff
@@ -144,10 +144,10 @@ static int create_process(odph_thread_t *thread, int cpu, uint64_t stack_size)
 /*
  * Wait single process to exit
  */
-static int wait_process(odph_thread_t *thread)
+static int wait_process(odph_thread_t *thread, odph_thread_join_result_t *res)
```
MatiasElo, line 147: As per current `odph_thread_join_result()` spec NULL is a valid value for `res`, this alone cannot be used to distinguish between `odph_thread_join()` and `odph_thread_join_result()`. Same for `wait_pthread()`.

helper/threads.c
TuomasTaipale: The `NULL` option was a leftover, forgot to remove, a bit pointless to call the new function without wanting the results.

helper/threads.c
MatiasElo: Since this is not a fast path function I'd keep the extra check as it can help in debugging user errors.

MatiasElo (approved):

helper/include/odp/helper/threads.h
```diff
@@ -261,16 +275,40 @@ int odph_thread_create(odph_thread_t thread[],
...
  *
  * @see odph_thread_create()
  */
 int odph_thread_join(odph_thread_t thread[], int num);
 
+/**
+ * Wait previously launched threads to exit
+ *
+ * Similar to odph_thread_join() but outputs results of joined threads and
+ * stops only if the actual join operation fails for some thread. Threads are
```
JereLeppanen, line 296: Why stop there?

helper/include/odp/helper/threads.h
```diff
@@ -261,16 +275,40 @@ int odph_thread_create(odph_thread_t thread[],
...
+ * joined in the order they are in 'thread' table. Returns number of threads
+ * successfully joined and writes respective exit statuses into the 'res'
+ * table.
+ *
+ * @param thread        Table of threads to exit
+ * @param[out] res      Table for result output
+ * @param num           Number of threads to exit and results to output
+ *
+ * @return Number of threads successfully joined (0 ... num)
+ * @retval -1  On failure
```
JereLeppanen, line 306: Could be "invalid parameters", since that's the only case in which -1 is returned.

helper/include/odp/helper/threads.h
MatiasElo: Handling the results output gets more complex if the function wouldn't stop on failure. Now it's enough the check only a single variable per joined thread.

PR #2051, opened 2024-04-19
[PATCH v3] linux-gen: shm: reclassify fallocate failure debug message

MatiasElo:
Classify debug messages from fallocate() failures as warnings when trying to use huge pages, as the SHM implementation will fall back to using normal pages and the error messages can confuse application users.

### DIFF
--- a/platform/linux-generic/odp_ishm.c
+++ b/platform/linux-generic/odp_ishm.c
@@ -651,6 +651,7 @@ static int create_file(int block_index, huge_flag_t huge, uint64_t len,
 	int  oflag = O_RDWR | O_CREAT | O_TRUNC; /* flags for open	      */
 	char dir[ISHM_FILENAME_MAXLEN];
 	int ret;
+	const odp_bool_t use_huge = huge == HUGE;
 
 	/* No ishm_block_t for the master single VA memory file */
 	if (single_va) {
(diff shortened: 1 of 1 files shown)


platform/linux-generic/odp_ishm.c
```diff
@@ -698,9 +698,11 @@ static int create_file(int block_index, huge_flag_t huge, uint64_t len,
 		}
 
 		if (ret == -1) {
-			_ODP_ERR("%s memory allocation failed: fd=%d, file=%s, "
-				"err=\"%s\"\n", (huge == HUGE) ? "Huge page" :
-				"Normal page", fd, filename, strerror(errno));
+			_ODP_LOG(use_huge ? ODP_LOG_WARN : ODP_LOG_ERR,
+				 use_huge ? "WARN" : "ERR",
+				 "%s memory allocation failed: fd=%d, file=%s, err=\"%s\"\n",
+				 use_huge ? "Huge page" : "Normal page", fd, filename,
```
TuomasTaipale, line 704: "page" could be moved to the common part.

TuomasTaipale (approved):
Minor comment, otherwise OK.

platform/linux-generic/odp_ishm.c
MatiasElo: Changed in V2.

PR #2053, opened 2024-04-24
api: cpu: add new prefetch calls

psavol:
ARM and x86 ISAs have dedicated prefetch instructions for these cases.


psavol:
v2: Allow prefetch of invalid addresses.

MatiasElo (approved):

jerinjacobk (approved):

PR #2063, opened 2024-05-14
[PATCH v2] test: include packet_gen output in failure logs

MatiasElo:


test/performance/odp_sched_pktio_run.sh
```diff
@@ -81,10 +82,15 @@ run_sched_pktio()
 		echo "PASS: received and transmitted over 5000 packets"
 		ret=0
 	else
-		echo "FAIL: less than thousand rx or tx packets $ret"
+		echo -e "\nodp_packet_gen"
+		echo "=============="
+		cat $GEN_LOG
+		echo "FAIL: less than 5000 rx or tx packets: ret=$ret"
```
TuomasTaipale, line 88: Maybe the explicit "5000" could be swapped to something like a "threshold" to avoid a slightly fragile dependency to the `odp_sched_pktio.c` constant.

TuomasTaipale (commented):
Copyright could be bumped in both files.

test/performance/odp_sched_pktio_run.sh
MatiasElo: Updated in V2.

TuomasTaipale (approved):

PR #2066, opened 2024-05-20
[PATCH v6] validation: timer: skip test if expected runtime is long

shijith:
In some implementations, the maximum timeout parameters in odp_timer_capability_t.max_tmo are excessively long. Consequently, the assert that tests the timeout length in the
timer_test_max_tmo_min_tmo_sched function is triggered. To ensure the timer validation test passes on all platforms, we need to modify this logic. The assert needs to be removed and a new conditional function should be added to check the maximum value of the minimum timeout. If the test is expected to run for an extended duration, it will be skipped.

### DIFF
--- a/test/validation/api/timer/timer.c
+++ b/test/validation/api/timer/timer.c
@@ -2003,6 +2003,7 @@ static void timer_test_tmo_limit(odp_queue_type_t queue_type,
 	int num = 5;
 	odp_timer_t timer[num];
 	odp_timer_clk_src_t clk_src = test_global->clk_src;
+	bool wait_tmo;
 
 	memset(&timer_capa, 0, sizeof(timer_capa));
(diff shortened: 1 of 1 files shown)


JereLeppanen (commented):
Looking a bit closer, maybe skipping the whole test is not the best after all, because then we'd skip also the pool creation and timer start.

Currently, we assume that min tmo is short and max tmo is long, so we wait for the former but not for the latter. Maybe it would be better to make that decision based on the tmo length, i.e. if tmo_ns is less than 5 seconds, wait for it, otherwise just cancel.

shijith:
Sure, I will make the changes as suggested.

psavol:
@shijith Please add Jere's review-by tag, so that we can merge this PR.

PR #2071, opened 2024-06-05
validation: cls: fix local allocation of large array

satheeshpaul:
Allocating a large array of odp_cls_cos_param_t on stack causes
overflows. Fixing it by dynamically allocating this array.

### DIFF
--- a/test/validation/api/classification/odp_classification_basic.c
+++ b/test/validation/api/classification/odp_classification_basic.c
@@ -136,7 +136,7 @@ static void cos_destroy_multi(odp_cos_t cos[], uint32_t num)
 static void cls_create_cos_multi(void)
 {
 	odp_cls_cos_param_t param_single;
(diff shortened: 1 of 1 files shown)


test/validation/api/classification/odp_classification_basic.c
```diff
@@ -147,6 +147,10 @@ static void cls_create_cos_multi(void)
 
 	num = capa.max_cos < MAX_HANDLES ? capa.max_cos : MAX_HANDLES;
 
+	param = calloc(MAX_HANDLES, sizeof(odp_cls_cos_param_t));
+	if (param == NULL)
+		return;
```
MatiasElo, line 152: Only `num` params is required and `calloc` shouldn't normally fail, so instead of silently skipping the test I'd fail here. E.g. `CU_ASSERT_FATAL(param != NULL);`

Otherwise the PR is looking OK.

MatiasElo (approved):

PR #2073, opened 2024-06-10
[PATCH v2] test: common: make bench_common an internal library

MatiasElo:
Make bench_common an internal library for benchmark applications to simplify compilation.


test/performance/Makefile.am
```diff
@@ -95,6 +95,8 @@ odp_ipsecfwd_SOURCES = odp_ipsecfwd.c
 AM_CFLAGS += $(LIBCONFIG_CFLAGS)
 endif
 
+PRELDADD += $(LIBBENCH_COMMON)
```
TuomasTaipale, line 98: For safety, the public bench functions could have a some harder-to-name-clash prefix at some point.

TuomasTaipale (approved):

test/performance/Makefile.am
MatiasElo: True, will keep this in mind.

PR #2074, opened 2024-06-11
[PATCH v6] test: bench: export test results

Rayska:


test/performance/odp_bench_packet.c
```diff
@@ -138,6 +151,8 @@ static void sig_handler(int signo ODP_UNUSED)
 	odp_atomic_store_u32(&gbl_args->suite.exit_worker, 1);
 }
 
+static int bench_packet_export(void *data);
```
TuomasTaipale, line 154: I think you can just define the function here directly if you need it at this point.

test/performance/odp_bench_misc.c
```diff
@@ -1038,6 +1085,13 @@ int main(int argc, char *argv[])
 
 	ret = gbl_args->suite.retval;
 
+	if (common_options.is_export) {
+		if (bench_misc_export(gbl_args)) {
+			ODPH_ERR("Error: Export failed\n");
+			ret = -1;
```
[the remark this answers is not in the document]
MatiasElo: This way it's easy to detect failures in CI pipelines, which record performance results. If the test is otherwise passing it could take a while until one notices that not all results are recorded correctly.

test/common/export_results.h
```diff
@@ -0,0 +1,80 @@
...
+* Returns new argument count. Original argument count decremented by the number of removed
+* helper options.
+*/
+int test_common_parse_options(int argc, char *argv[]);
+
+/**
+* Get test options
+*
+* Return used test options. test_common_parse_options() must be called before
+* using this function. Opens file for writing
```
TuomasTaipale, line 44: The "Opens file for writing" could be dropped.

test/common/export_results.h
```diff
@@ -0,0 +1,80 @@
...
+* Returns 0 on success, -1 on failure
+*/
+int test_common_write(const char *fmt, ...);
+
+/**
+ * Terminate writing
+ *
+ * Called after last write. Do not call test_common_write after calling this.
+ *
+ * Returns 0 on success, -1 on failure
```
TuomasTaipale, line 72: Doesn't return anything anymore.

TuomasTaipale (approved):

test/performance/odp_bench_buffer.c
```diff
@@ -870,6 +915,13 @@ int main(int argc, char *argv[])
 
 	ret = gbl_args->suite.retval;
 
+	if (common_options.is_export) {
```
MatiasElo, line 918: Probably it doesn't make much sense to try to export results if the test returned failure (`ret != 0`). Same comment for all applications.

test/performance/odp_bench_packet.c
```diff
@@ -126,6 +135,10 @@ typedef struct {
 	odp_time_t ts_tbl[TEST_REPEAT_COUNT];
 	/** Array for storing test data */
 	uint8_t data_tbl[TEST_REPEAT_COUNT][TEST_MAX_PKT_SIZE];
+	/** Options for exporting results */
+	test_common_options_t *common_options_ptr;
```
MatiasElo, line 139: This works, but probably it would be a bit safer to copy the whole struct here instead of using a pointer.

test/performance/odp_bench_pktio_sp.c
```diff
@@ -31,6 +32,9 @@
 /* Maximum interface name length */
 #define MAX_NAME_LEN 128
 
+/* Number of functions originally in test_suite*/
```
MatiasElo, line 35: Space missing at the end.

test/common/export_results.c
```diff
@@ -0,0 +1,128 @@
...
+				argc -= 2;
+				continue;
+			} else {
+				for (j = i; j < argc - 1; j++)
+					argv[j] = argv[j + 1];
+				argc--;
+			}
+		}
+		i++;
+	}
```
MatiasElo, line 61: If `TEST_COMMON_EXPORT` is an empty string or `--test-common-export` doesn't have path argument the output file ends up being names `.csv`, which is likely not the intention.

test/common/export_results.c
```diff
@@ -0,0 +1,128 @@
...
+
+int test_common_options(test_common_options_t *options)
+{
+	const char *extension = ".csv";
+	size_t ext_len = strlen(extension);
+	size_t filename_len = strlen(gbl_data.filename);
+
+	memset(options, 0, sizeof(*options));
+
+	options->is_export = gbl_data.common_options.is_export;
```
MatiasElo, line 73: Can simply return here if export is not enabled.

MatiasElo (approved):

PR #2077, opened 2024-06-17
[PATCH v4] api: std: clarify odp_bool_t documentation

MatiasElo:


include/odp/api/abi-default/std_types.h
```diff
@@ -22,7 +22,7 @@ extern "C" {
  *  @{
  */
 
-typedef int odp_bool_t;
+typedef bool odp_bool_t;
```
JannePeltonen, line 25: This is not only for memory savings but to make the behavior comply with the clarified API when 64 bit values are assigned to an odp_bool  (otherwise e.g. unsigned values with the low 32 bits zero (like 0x100000000) would get converted to zero). The commit message could mention this.

JannePeltonen:
With a commit message improvement, Reviewed-by: Janne Peltonen <janne.peltonen@nokia.com>

include/odp/api/abi-default/std_types.h
MatiasElo: Commit message improved in V2.

JannePeltonen (approved):

psavol (approved):

jerinjacobk (approved):

PR #2078, opened 2024-06-18
test: dmafwd: packet cleanup counter add

AmitShukla87622:
dmafwd sends the packets in multiple of burst size. If the number of
packets to be sent are not in multiple of burst size,
(num pkts to send % burst size) packets are not sent.

This patch adds a stats counter "pkt_cleanup" for the packets that are
not sent.

Signed-off-by: Amit Prakash Shukla <amitprakashs@marvell.com>
Reviewed-by: Tuomas Taipale <tuomas.taipale@nokia.com>

### DIFF
--- a/test/performance/odp_dmafwd.c
+++ b/test/performance/odp_dmafwd.c
@@ -84,6 +84,7 @@ typedef struct {
 	uint64_t sched_cc;
 	uint64_t tot_cc;
 	uint64_t sched_rounds;
+	uint64_t pkt_cleanup;
 } stats_t;
 
 typedef struct ODP_ALIGNED_CACHE {
@@ -1122,10 +1123,12 @@ static inline void pop_packets(pkt_vec_t *pkt_vec, int num_procd)
 		pkt_vec->pkts[i] = pkt_vec->pkts[j];
 }
 
-static void free_pending_packets(pkt_vec_t pkt_vecs[], uint32_t num_ifs)
+static void free_pending_packets(pkt_vec_t pkt_vecs[], uint32_t num_ifs, stats_t *stats)
 {
-	for (uint32_t i = 0U; i < num_ifs; ++i)
+	for (uint32_t i = 0U; i < num_ifs; ++i) {
+		stats->pkt_cleanup += pkt_vecs[i].num;
 		odp_packet_free_multi(pkt_vecs[i].pkts, pkt_vecs[i].num);
+	}
 }
 
 static int process_packets(void *args)
@@ -1198,7 +1201,7 @@ static int process_packets(void *args)
 	stats->sched_cc = cdiff;
 	stats->tot_cc = odp_cpu_cycles_diff(c2, c1);
 	stats->sched_rounds = rounds;
-	free_pending_packets(pkt_vecs, num_ifs);
+	free_pending_packets(pkt_vecs, num_ifs, stats);
 	odp_barrier_wait(&config->prog_config->term_barrier);
 
 	if (config->prog_config->drain_fn)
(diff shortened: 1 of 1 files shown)


test/performance/odp_dmafwd.c
```diff
@@ -1190,6 +1190,12 @@ static int process_packets(void *args)
 				pkt_fn(pkt_vec->pkts, burst_size, pkt_vec->pktio, init_fn,
 				       start_fn, config);
 				pop_packets(pkt_vec, burst_size);
+			} else {
+				if (pkt_vec->num) {
+					pkt_fn(pkt_vec->pkts, pkt_vec->num, pkt_vec->pktio, init_fn,
+					       start_fn, config);
+					pop_packets(pkt_vec, pkt_vec->num);
+				}
```
TuomasTaipale, line 1198: The gathering-to-burst is by design in order to test the characteristics of DMA vs. SW copy with different size bursts so the burst has to be formed first before sending anything (which might take a few scheduling rounds).

Was this causing some issues?

AmitShukla87622: If the number of packets to be transmitted are not in multiples of the burst size then the remaining packets were not transmitted. For example: Number of packets to be sent out is 100 and burst size is 15 then 10 packets will not be transmitted.

TuomasTaipale: Yes, that is conceptually fine. The remaining packets should be handled in `free_pending_packets()` which frees these unsent packets when test is done.

Are you seeing some issues or leaking resources without this change?

MatiasElo: Instead of freeing the remaining packets in `free_pending_packets()`, one option could be to transmit them, so no packets would be dropped from the point of view of an outside observer.

TuomasTaipale: Well, there's really no need to send the packets just for that. We have several testers that do a "drain-and-free" phase at the end, after the measured test, which might result in fewer received packets than sent.

AmitShukla87622: From the test statistics it is confusing as in the stats it neither shows dropped nor forwarded. In the below statistics snippet 100 packets were sent with burst size of 15:

`DMA forwarder done

    copy mode:       SW
    burst size:      15
    packet length:   1024
    max cache size:  8

    worker 0:
        packet copy errors: 0
        packets forwarded:  90
        packets dropped:    0
        call cycles per schedule round:
            total:    3999
            schedule: 3959
            rounds:   9071500`

TuomasTaipale: As a compromise, `free_pending_packets()` could add the drops to the `stats_t::discards`.

AmitShukla87622: That will show up in stats as packets dropped and will be a wrong interpretation for the packets not sent.

TuomasTaipale: In any case, the current proposal breaks the core bursting logic, so it can't be part of the main test loop.

MatiasElo: If using `stats_t::discards` counter is too ambiguous, another option could be to add a new counter for clean up. E.g. `packets cleaned up:`

Btw. Tuomas is out of office this week.

TuomasTaipale (approved):
Reviewed-by: Tuomas Taipale <tuomas.taipale@nokia.com>

When adding the reviewed-by tag, please change the commit message to reflect the new changes (no more packet sending but stats update).

MatiasElo:
Hi @AmitShukla87622, please add the reviewed-by tag and address Tuomas' commit message comment so this PR can be merged.

MatiasElo:
Ping @AmitShukla87622

AmitShukla87622:
Apologies for delayed reply. I was out of office for few days. I will address the comment and push the change.

PR #2079, opened 2024-06-20
[PATCH v1] test/performance: fix IPv4 checksum

rahul-bhansali:
An issue of incorrect IPv4 checksum is observed when sources are compiled in release mode with optimization flags
-fno-var-tracking-assignments.
On annotating, it was observed that few IP header fields are updated after checksum calculation routine.

Using this API odph_ipv4_csum_update() solves the issue as it synchronizes the load/store instructions.

### DIFF
--- a/test/performance/odp_packet_gen.c
+++ b/test/performance/odp_packet_gen.c
@@ -1695,7 +1695,7 @@ static int init_packets(test_global_t *global, int pktio,
 							ODPH_IPPROTO_TCP : 0xFE;
 			ip->src_addr = odp_cpu_to_be_32(test_options->ipv4_src);
 			ip->dst_addr = odp_cpu_to_be_32(test_options->ipv4_dst);
-			ip->chksum = ~odp_chksum_ones_comp16(ip, ODPH_IPV4HDR_LEN);
+			odph_ipv4_csum_update(pkt);
 			odp_packet_has_ipv4_set(pkt, 1);
 
 			u8 = ((uint8_t *)data + l2_len + ODPH_IPV4HDR_LEN);


test/performance/odp_packet_gen.c
```diff
@@ -1695,7 +1695,7 @@ static int init_packets(test_global_t *global, int pktio,
 							ODPH_IPPROTO_TCP : 0xFE;
 			ip->src_addr = odp_cpu_to_be_32(test_options->ipv4_src);
 			ip->dst_addr = odp_cpu_to_be_32(test_options->ipv4_dst);
-			ip->chksum = ~odp_chksum_ones_comp16(ip, ODPH_IPV4HDR_LEN);
+			odph_ipv4_csum_update(pkt);
```
JannePeltonen, line 1698: I would like to understand how exactly things go wrong with the current code. It sounds like there could be some aliasing problem. How do you compile the code (compiler, flags, target arch etc) and how do you reproduce the problem? Can you provide more details on what you have found in your analysis, maybe with some disassembly? In what way is IP header updated after the checksum calculation? Does the problem go away if you compile with -fno-strict-aliasing?

I do not see what -fno-var-tracking-assignments (which seems to be a debug info related flag, not an optimization flag) has to do with this.

I am not sure what you mean when you say that odph_ipv4_csum_update() "synchronizes the load/store instructions". Can you elaborate?

While switching to use odph_ipv4_csum_update() may fix the immediate problem observed, maybe something in odp_chksum_ones_comp16() needs to be fixed.

rahul-bhansali: Hi @JannePeltonen 
It is aliasing problem only , can also be solved in other way like: 
a) have -fno-strict-aliasing  flag by default in test/performance/Makefile.am .
b) barrier / asm(nop) before odp_chksum_ones_comp16

I have used this config command to reproduce the issue :- "--host=aarch64-marvell-linux-gnu     --disable-abi-compat        --disable-shared --enable-lto   --enable-test-example     --enable-test-perf     --enable-static  --prefix=$(pwd)/out --enable-test-vald    --without-openssl 'CFLAGS=-O3 -ggdb -fno-var-tracking-assignments' 'CXXFLAGS=-O3 -ggdb -fno-var-tracking-assignments'"

Attached the snip of assembly for Ipv4 checksum update.
[ipv4_csum_assembly_part.txt](https://github.com/user-attachments/files/15986336/ipv4_csum_assembly_part.txt)

JannePeltonen: I do not think that a change that changes only the perf test or its makefile is appropriate in this case, since it would leave the aliasing problem in odp_chksum_ones_comp16() unfixed.

I am not sure what the right fix would actually be. Adding a suitable optimization barrier in chksum_partial(), adding may_alias type attributes in the x_uint32_t and x_uint16_t types or disabling optimizations based on the strict aliasing rule with a pragma in chksum_partial() could work. But it would still leave the (cosmetic?) issue that chksum_partial() is doing type punning through pointer casting which causes, strictly (pun intended) speaking, undefined behaviour. chksum_partial() could be rewritten without the kind of pointer casting it is doing now, but then some of the optimizations it is trying to do could perhaps be lost.

There are probably many other places in ODP code which break strict aliasing rules but maybe do not cause actual errors (at least without LTO). But even then I am not sure we would want to have -fno-strict-aliasing for all ODP code instead of trying to fix the rule violations.

JereLeppanen: @rahul-bhansali Could you provide the annotated disassembly for the whole tx_thread function? ipv4_csum_assembly_part.txt cuts off the beginning of the checksumming, so it's difficult to tell whether all the IP headers fields have been summed up or not. It's not unusual for the fields to be written to memory way after being added to the checksum.

rahul-bhansali: @JereLeppanen 
Please check the attached disassembly file. 
[ipv4_cksum_assembly_check.txt](https://github.com/user-attachments/files/16283951/ipv4_cksum_assembly_check.txt)

This I tried with x86_64-pc-linux-gnu host and platform: linux-generic .
Config command: ./configure --disable-abi-compat --disable-shared --enable-lto --enable-test-example --enable-test-perf --enable-static --prefix=$(pwd)/out --enable-test-vald --without-openssl 'CFLAGS=-O3 -ggdb -fno-var-tracking-assignments' 'CXXFLAGS=-O3 -ggdb -fno-var-tracking-assignments'

JereLeppanen: Well, these two disassemblies are from two different architectures and two different checksum implementations. I don't immediately see any problem in either of them. If you do, could you point it out?

Has the problem been reproduced on arm, or x86, or both? What is the rate or probability of an incorrect checksum in packets? Where and how is the incorrect checksum observed, in the host that runs odp_packet_gen, or in some other host that receives the packets?

rahul-bhansali: This problem is seen on both arm and x86 platform. This incorrect checksum is in all the packets and always reproducible (10/10) with the config options that I shared. This issue is on generator side means host generating packets using odp_packet_gen, I did cross verify with tcpdump app and Ixia .

JereLeppanen: Ok, and what is your odp_packet_gen command line?

rahul-bhansali: command: sudo ./build/test/performance/odp_packet_gen --gap 0 -i <interface-name> --ipv4_src 10.10.10.0 --ipv4_dst 10.10.10.1 --eth_dst b6:b3:69:03:28:cb --src_port 1122 --dst_port 3344 -r 0 -t 1

JereLeppanen: Thanks. The problem is indeed easy to reproduce, but it requires both LTO and -O3, which is why we didn't notice before.

I agree with Janne, we probably need to fix the checksum implementation, instead of just this one call site.

JereLeppanen:
@rahul-bhansali In PR #2096 we're suggesting that this should be fixed with the may_alias attribute, please have a look.

rahul-bhansali:
> @rahul-bhansali In PR #2096 we're suggesting that this should be fixed with the may_alias attribute, please have a look.

Ack. That may_alias is also fine.
So for other than linux-generic platforms, need to add may_alias or other way of handling in respective odp_chksum_ones_comp16() API to fix this.

JereLeppanen:
Yes, exactly. I'm closing this PR now, but thanks for reporting the issue, it was a good catch.

PR #2081, opened 2024-06-26
[PATCH v8] test: export: export with options & arguments

Rayska:


test/common/export_results.c
```diff
@@ -18,17 +18,50 @@
...
+	if (ret >= remaining_space)
+		goto exit;
+
+	for (i = 2; i < argc; i++) {
+		current_len = strlen(gbl_data.args);
+		remaining_space = MAX_ARGS_LEN - current_len;
+		ret = snprintf(gbl_data.args + current_len, remaining_space, " %s", argv[i]);
+		if (ret >= remaining_space)
+			goto exit;
+	}
```
MatiasElo, line 56: The function could perhaps be simplified a bit. For example something like this:
```
	int ret;
	int len = 0;
	int remaining_space = sizeof(gbl_data.args);

	for (int i = 1; i < argc && remaining_space > 0; i++) {
		ret = snprintf(&gbl_data.args[len], remaining_space, " %s", argv[i]);
		if (ret >= remaining_space || ret < 0)
			goto exit;
		len += ret;
		remaining_space -= ret;
	}
...
```

test/performance/odp_timer_accuracy.c
MatiasElo, line 1: Copyright can be updated.

test/performance/odp_timer_accuracy.c
```diff
@@ -1268,6 +1302,12 @@ int main(int argc, char *argv[])
 		exit(EXIT_FAILURE);
 	}
 
+	argc = test_common_parse_options(argc, argv);
+	if (test_common_options(&common_options)) {
+		ODPH_ERR("Error: reading test helper options failed\n");
```
MatiasElo, line 1307: To better follow the error message style in this application `Error:` could be dropped.

test/performance/odp_timer_accuracy.c
```diff
@@ -1406,6 +1446,9 @@ int main(int argc, char *argv[])
 	}
 
 	odph_thread_join(thread_tbl, num_workers);
+
+	test_global->common_options = common_options;
```
MatiasElo, line 1450: I'd perhaps move this closer to `test_global` reservation. E.g. after `memcpy(&test_global->opt, &test_opt, sizeof(test_opt_t));` line.

test/performance/odp_timer_accuracy.c
```diff
@@ -1008,6 +1010,37 @@ static void print_stat(test_global_t *test_global)
...
+		if (test_common_write("num after,num before,num exact,num retry,"
+				      "error after min (nsec),error after min resolution,"
+				      "error after max (nsec),error after max resolution,"
+				      "error after ave (nsec),error after ave resolution,"
+				      "error before min (nsec),error before min resolution,"
+				      "error before max (nsec),error before max resolution,"
+				      "error before ave (nsec),error before ave resolution,"
+				      "final timeout error max (nsec),"
+				      "final timeout error max resolution\n"))
+			goto exit;
```
MatiasElo, line 1024: Perhaps better not to use `goto` here and do everything export related inside this `if` branch.

test/performance/odp_timer_accuracy.c
```diff
@@ -1008,6 +1010,37 @@ static void print_stat(test_global_t *test_global)
...
+				      stat->num_exact, stat->num_too_near,
+				      stat->nsec_after_min, (double)stat->nsec_after_min / res_ns,
+				      stat->nsec_after_max, (double)stat->nsec_after_max / res_ns,
+				      ave_after, (double)ave_after / res_ns,
+				      stat->nsec_before_min, (double)stat->nsec_before_min / res_ns,
+				      stat->nsec_before_max, (double)stat->nsec_before_max / res_ns,
+				      ave_before, (double)ave_before / res_ns,
+				      max, (double)max / res_ns
+				      ))
+			goto exit;
```
MatiasElo, line 1037: Failures to export test results are not detected.

test/common/export_results.c
```diff
@@ -18,17 +18,43 @@
...
+#define MAX_ARGS_LEN 128
+
 typedef struct {
 	test_common_options_t common_options;
 
 	char filename[MAX_FILENAME_LEN];
 
 	FILE *file;
 
+	char args[MAX_ARGS_LEN];
```
TuomasTaipale, line 31: This could probably be dynamically allocated, some testers have potentially very long parameter arguments and it would be nice to always have them all included.

test/performance/odp_timer_accuracy.c
```diff
@@ -1,6 +1,6 @@
 /* SPDX-License-Identifier: BSD-3-Clause
```
TuomasTaipale, line 1: Nitpick, but commit message could have period at the end.

test/performance/odp_timer_accuracy.c
MatiasElo: `print_stat()` needs still a return value, so an export failure could be easily detected for example in CI.

test/common/export_results.c
```diff
@@ -98,6 +131,13 @@ int test_common_options(test_common_options_t *options)
 		return -1;
 	}
 
+	if (gbl_data.args) {
+		fprintf(gbl_data.file, "#%.*s;%s\n", (int)filename_len, gbl_data.filename,
```
MatiasElo, line 135: Could add check for `fprintf()` failures.  Also, application could at least in theory call `test_common_options()` multiple times, which would currently cause problems. An internal variable e.g. `init_done` could be used to track this, so file open and header write would be done only once. Otherwise the PR look OK to me.

TuomasTaipale (approved):

MatiasElo (approved):

PR #2082, opened 2024-07-02
[PATCH v10] api: atomic: add functions for fetching and updating maximum/minimum values

MatiasElo:
V4:
- Moved `odp_atomic_add_u32/64()` calls out of validation tests loops

### DIFF
--- a/include/odp/api/spec/atomic.h
+++ b/include/odp/api/spec/atomic.h
@@ -1,6 +1,7 @@
 /* SPDX-License-Identifier: BSD-3-Clause
  * Copyright (c) 2013-2018 Linaro Limited
  * Copyright (c) 2021 ARM Limited
+ * Copyright (c) 2024 Nokia
  */
 
 /**
@@ -26,11 +27,12 @@ extern "C" {
  *
  * Atomic integer types (odp_atomic_u32_t, odp_atomic_u64_t and
  * odp_atomic_u128_t) can be used to implement e.g. shared counters. If not
- * otherwise documented, operations in this API are implemented using
- * <b> RELAXED memory ordering </b> (see memory order descriptions in
- * the C11 specification). Relaxed operations do not provide synchronization or
- * ordering for other memory accesses (initiated before or after the operation),
- * only atomicity of the operation itself is guaranteed.
+ * otherwise documented, all functions in this API perform their operations
+ * atomically and are implemented using <b> RELAXED memory ordering </b> (see
+ * memory order descriptions in the C11 specification). Relaxed operations do
+ * not provide synchronization or ordering for other memory accesses (initiated
+ * before or after the operation), only atomicity of the operation itself is
+ * guaranteed.
  *
  * <b> Operations with non-relaxed memory ordering </b>
  *
@@ -175,6 +177,20 @@ void odp_atomic_dec_u32(odp_atomic_u32_t *atom);
  * @param new_max New maximum value to be written into the atomic variable
  */
 void odp_atomic_max_u32(odp_atomic_u32_t *atom, uint32_t new_max);
+
+/**
+ * Fetch and update maximum value of atomic uint32 variable
+ *
+ * Compares value of atomic variable to the new maximum value. If the new value
+ * is greater than the current value, writes the new value into the variable.
+ * Always returns the original value of the atomic variable.
+ *
+ * @param atom    Pointer to atomic variable
+ * @param new_max New maximum value to be written into the atomic variable
+ *
+ * @return Original value of the atomic variable
+ */
+uint32_t odp_atomic_fetch_max_u32(odp_atomic_u32_t *atom, uint32_t new_max);
 
 /**
  * Update minimum value of atomic uint32 variable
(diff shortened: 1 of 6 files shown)


include/odp/api/spec/atomic.h
```diff
@@ -176,6 +177,20 @@ void odp_atomic_dec_u32(odp_atomic_u32_t *atom);
  */
 void odp_atomic_max_u32(odp_atomic_u32_t *atom, uint32_t new_max);
 
+/**
+ * Fetch and update maximum value of atomic uint32 variable
```
JannePeltonen, line 181: The description does not say that the whole operation (fetching + updating) is atomic, even though it is implied by the name of the function. This is actually the case in the existing atomic functions too. Maybe the header could have a generic comment somewhere, stating that all the atomic functions perform their operation atomically unless otherwise noted.

test/validation/api/atomic/atomic.c
```diff
@@ -1124,6 +1182,15 @@ static int test_atomic_max_thread(void *arg UNUSED)
 	return CU_get_number_of_failures();
 }
 
+static int test_atomic_fetch_max_thread(void *arg UNUSED)
+{
+	thread_init();
+	test_atomic_fetch_max_32();
+	test_atomic_fetch_max_64();
+
+	return CU_get_number_of_failures();
```
JannePeltonen, line 1191: This returns not only the number of failures seen in this test but the total so far, as is done by the existing functions too. This is not very nice since a test failure causes printouts from all the subsequent tests too, even if they pass.

test/validation/api/atomic/atomic.c
```diff
@@ -1552,6 +1634,11 @@ static void atomic_test_atomic_max(void)
 	test_atomic_functional(test_atomic_max_thread, test_atomic_validate_max);
 }
 
+static void atomic_test_atomic_fetch_max(void)
+{
+	test_atomic_functional(test_atomic_fetch_max_thread, test_atomic_validate_max);
```
JannePeltonen, line 1639: The existing test_atomic_validate_max does not validate the atomicity of fetch+max operation at alll. If I replace odp_atomic_fetch_max_u32() in the test function with odp_atomic_load_u32() followed by odp_atomic_max_u32(), the test still passes.

One way to improve this would be to keep track of how much the value of the atomic variable gets increased (i.e. if (tmp > val) increase += tmp - val) and then checking at validation phase that the sum of the thread-specific increases is consistent with the final value of the atomic variable. This would at least catchs the non-atomic load+max pair that I tried.

include/odp/api/spec/atomic.h
MatiasElo: Comment added in V3.

test/validation/api/atomic/atomic.c
MatiasElo: True, removed in V3. Will clean the existing tests in a separate PR.

test/validation/api/atomic/atomic.c
MatiasElo: Improved test in V3.

JannePeltonen (approved):

platform/linux-generic/include/odp/api/plat/atomic_inlines.h
```diff
@@ -333,6 +340,19 @@ _ODP_INLINE void odp_atomic_min_u64(odp_atomic_u64_t *atom, uint64_t new_val)
...
 }
 
+_ODP_INLINE uint64_t odp_atomic_fetch_max_u64(odp_atomic_u64_t *atom, uint64_t new_val)
+{
+	uint64_t old_val;
+
+	old_val = odp_atomic_load_u64(atom);
+
+	while (new_val > old_val) {
+		if (odp_atomic_cas_u64(atom, &old_val, new_val))
```
JannePeltonen, line 357: Now I realized that this odp_atomic_cas_u64 here is just the spinlock version a little bit above. And so is the atomic load too. It would be much more efficient to implement this using the ATOMIC_OP macro so that the spinlock is taken once instead of two or more times.

psavol:
Heading of the first patch (atomic: add odp_atomic_fetch_max_ ...) misses "api: " prefix.

platform/linux-generic/include/odp/api/plat/atomic_inlines.h
MatiasElo: Changed in V5.

JannePeltonen:
v5 looks good.

include/odp/api/spec/atomic.h
```diff
@@ -176,6 +178,20 @@ void odp_atomic_dec_u32(odp_atomic_u32_t *atom);
...
+ *
+ * Compares value of atomic variable to the new maximum value. If the new value
+ * is greater than the current value, writes the new value into the variable.
+ * Returns the value of the variable before update operation.
```
psavol, line 186: Suggesting to change: "Returns the value of the variable before update operation." -> "Always returns the value used in the comparison."

It would highlight that the "fetched" value is always returned.

JannePeltonen: That sounds more cumbersome and difficult to understand to me. I do not quite see what the potential misunderstanding in the current wording might be.

JannePeltonen: Maybe "returns the original value of the variable" would be ok if the "before update operation" wording is deemed confusing?

MatiasElo: ```Returns the original value of the variable used in the comparison.``` or
```Returns the original value of the atomic variable.```?

psavol: Original value is good choice. Remember to update also the return value comment.

MatiasElo: Updated in V6.

MatiasElo: Further tweaking in V7.

psavol (approved):

jerinjacobk (approved):

PR #2084, opened 2024-07-03
[PATCH v2] test: bench_common: fix running selected test indefinitely

MatiasElo:
The benchmark index 'j' is already incremented in the for loop, so remove the extra increment which caused every other index to be skipped.

V2:
- Renamed `j` to `i`


TuomasTaipale (approved):
If you want, maybe the `j`s could be swapped to `i`s.

PR #2088, opened 2024-07-09
[PATCH v4] api: ml: add extra info field in model and IO info

syalavarthi:
Added extra info fields in model, input and output info structures. Support for extra info would enable the user to fetch implementation specific information like number of layers, HW resource allocation for the model and pre-defined quantization scale factors and zero-point values for input and output.


JereLeppanen:
I've posted an alternative way of handling quantization parameters in #2182.

TuomasTaipale:
Scale and zero-point was introduced in #2198, closing this.

PR #2091, opened 2024-07-11
[PATCH v2] github_ci: remove CentOS 7 build test

MatiasElo:
CentOS 7 has reached EOL, so remove it from the CI tests.

### DIFF
--- a/DEPENDENCIES
+++ b/DEPENDENCIES
@@ -1,8 +1,8 @@
 Prerequisites for building the OpenDataPlane (ODP) API
 
 1. Linux
-   CentOS 7 (kernel v3.10) is the oldest Linux distributions tested by the ODP
-   CI. Earlier versions may or may not work.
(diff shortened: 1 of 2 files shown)


DEPENDENCIES
```diff
@@ -1,8 +1,8 @@
 Prerequisites for building the OpenDataPlane (ODP) API
 
 1. Linux
-   CentOS 7 (kernel v3.10) is the oldest Linux distributions tested by the ODP
-   CI. Earlier versions may or may not work.
+   Rocky Linux 8 (kernel v4.18) is the oldest Linux distributions tested by the
```
JannePeltonen, line 4: You could fix the typo ("distributions") too when changing this.

JannePeltonen (approved):

DEPENDENCIES
MatiasElo: Fixed in V2.

PR #2095, opened 2024-07-23
[PATCH v3] test: export: fix written path in output files

Rayska:
Previously the path of the output file was written into the output file. Now the command with which the test was run with is written instead. Writing the path in the output is meant to indicate which application the results were acquired from.

### DIFF
--- a/test/common/export_results.c
+++ b/test/common/export_results.c
@@ -16,12 +16,14 @@
 #include <unistd.h>
 
 /** Maximum length of output path/name */
-#define MAX_FILENAME_LEN 192
+#define MAX_PATH_LEN 192
 
 typedef struct {
 	test_common_options_t common_options;
 
-	char filename[MAX_FILENAME_LEN];
+	char filename[MAX_PATH_LEN];
+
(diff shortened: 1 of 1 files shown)


test/common/export_results.c
```diff
@@ -68,23 +70,22 @@ int test_common_parse_options(int argc, char *argv[])
...
 		gbl_data.common_options.is_export = true;
-		odph_strcpy(gbl_data.filename, env, MAX_FILENAME_LEN);
+		odph_strcpy(gbl_data.filename, env, MAX_PATH_LEN);
 	}
 
 	/* Find and remove option */
 	for (i = 0; i < argc;) {
 		if (strcmp(argv[i], "--test-common-export") == 0) {
 			gbl_data.common_options.is_export = true;
+			odph_strcpy(gbl_data.command, argv[0], MAX_PATH_LEN);
```
TuomasTaipale, line 80: As this is always from `argv[0]` could be done inside the `if (gbl_data.common_options.is_export) { }` below.

TuomasTaipale (approved):

PR #2096, opened 2024-07-25
[PATCH v2] linux-gen: fix some potential aliasing issues

JereLeppanen:
Branched from #2107.


psavol:
Enable of -no-srtrict-aliasing obsolated this changes.

PR #2099, opened 2024-08-09
[PATCH v9] api: atomic: add atomic bit set and clear functions

TuomasTaipale:
Add functions to atomically set and clear bits in 32-bit and 64-bit variables.

v2:
- Comments addressed

v3:
- Comments addressed

v4:
- Added reviewed-by tags

v6:
- Comments addressed

v8:
- Rebased

### DIFF
--- a/include/odp/api/spec/atomic.h
+++ b/include/odp/api/spec/atomic.h
@@ -248,6 +248,42 @@ int odp_atomic_cas_u32(odp_atomic_u32_t *atom, uint32_t *old_val, uint32_t new_v
  */
 uint32_t odp_atomic_xchg_u32(odp_atomic_u32_t *atom, uint32_t new_val);
 
+/**
+ * Set bits in atomic uint32 variable
+ *
+ * @param atom    Pointer to atomic variable
+ * @param bits    Bits to set
+ */
+void odp_atomic_bit_set_u32(odp_atomic_u32_t *atom, uint32_t bits);
+
+/**
+ * Fetch value and set bits in atomic uint32 variable
+ *
+ * @param atom    Pointer to atomic variable
+ * @param bits    Bits to set
+ *
+ * @return Value of the variable before modification
+ */
+uint32_t odp_atomic_bit_fetch_set_u32(odp_atomic_u32_t *atom, uint32_t bits);
+
+/**
+ * Clear bits in atomic uint32 variable
+ *
+ * @param atom    Pointer to atomic variable
+ * @param bits    Bits to clear
+ */
+void odp_atomic_bit_clr_u32(odp_atomic_u32_t *atom, uint32_t bits);
+
+/**
+ * Fetch value and clear bits in atomic uint32 variable
+ *
+ * @param atom    Pointer to atomic variable
+ * @param bits    Bits to clear
+ *
+ * @return Value of the variable before modification
+ */
+uint32_t odp_atomic_bit_fetch_clr_u32(odp_atomic_u32_t *atom, uint32_t bits);
+
 /*
  * 64-bit operations in RELAXED memory ordering
  * --------------------------------------------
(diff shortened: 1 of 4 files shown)


platform/linux-generic/include/odp/api/plat/atomic_inlines.h
```diff
@@ -264,6 +296,50 @@ _ODP_INLINE uint64_t odp_atomic_xchg_u64(odp_atomic_u64_t *atom,
 	return ATOMIC_OP(atom, atom->v = new_val);
 }
 
+_ODP_INLINE void odp_atomic_bit_set_u64(odp_atomic_u64_t *atom, uint64_t bits)
+{
+	uint64_t old_val;
+
+	old_val = odp_atomic_load_u64(atom);
+
+	while (!odp_atomic_cas_u64(atom, &old_val, old_val | bits))
```
JannePeltonen, line 305: These functions could be implemented using the ATOMIC_OP macro similarly to the adjacent functions. Now both the atomic load and the cas expand to a spinlocked section. So instead of taking the spinlock once, this now takes it two or more times.

test/validation/api/atomic/atomic.c
```diff
@@ -30,12 +29,17 @@ typedef struct {
 	odp_atomic_u64_t a64u_min;
 	odp_atomic_u64_t a64u_max;
 	odp_atomic_u64_t a64u_xchg;
+	odp_atomic_u64_t a64u_bit_set;
+	odp_atomic_u64_t a64u_bit_clr;
```
JannePeltonen, line 33: These are not really needed but a32u and a64u could simply be used instead. The other fields like a32u_max are there only because they are needed at the same time as the generic a32u.

test/validation/api/atomic/atomic.c
```diff
@@ -706,6 +710,94 @@ static void test_atomic_non_relaxed_128_acq_rel(void)
 	}
 }
 
+static void test_atomic_bit_set_32(void)
+{
+	odp_barrier_wait(&global_mem->global_barrier);
+	odp_atomic_bit_set_u32(&global_mem->a32u_bit_set, 1U << (odp_thread_id() - 1));
```
JannePeltonen, line 716: The thread ID could be taken before the barrier wait so that the atomic op is right after the barrier (as presumably the barriers are there to make it more likely that the atomic ops overlap).

include/odp/api/spec/atomic.h
```diff
@@ -524,6 +596,28 @@ int odp_atomic_cas_rel_u32(odp_atomic_u32_t *atom, uint32_t *old_val,
...
+/**
+ * Set bits in atomic uint32 variable using RELEASE memory ordering
+ *
+ * Otherwise identical to odp_atomic_bit_set_u32() but ensures RELEASE memory
+ * ordering.
+ *
+ * @param atom    Pointer to atomic variable
+ * @param bits    Bits to set
+ */
+void odp_atomic_bit_set_rel_u32(odp_atomic_u32_t *atom, uint32_t bits);
```
JannePeltonen, line 608: Now that the release variants are here, maybe an acq-rel variant of the fetch_set and fetch_clr functions would be useful too.

JannePeltonen (changes requested):

platform/linux-generic/include/odp/api/plat/atomic_inlines.h
TuomasTaipale: V2.

test/validation/api/atomic/atomic.c
TuomasTaipale: Separate for clarity, "bit clear" scenario also requires a different initial value in `test_atomic_store()`.

include/odp/api/spec/atomic.h
TuomasTaipale: There are no non-relaxed fetch versions of any kind, so seems that they are not that useful.

JannePeltonen: They can anyway be in a separate commit and PR.

test/validation/api/atomic/atomic.c
```diff
@@ -706,6 +710,114 @@ static void test_atomic_non_relaxed_128_acq_rel(void)
 	}
 }
 
+static void test_atomic_bit_set_32(void)
+{
+	const uint32_t bits = 1U << (odp_thread_id() - 1);
```
JannePeltonen, line 715: I noticed now that the way thread_id() is used as a bit position is a bit problematic. The bit shifting by thread_id() becomes UB (and likely breaks the test too) as soon as MAX_WORKERS is made bigger than 32. Another problem is that the bit operations do not test the high order half of 64 bit values, so the test would not notice if my 64-bit bit ops are just 32-bit ops.

I think both of these problems could be fixed by deriving the bit position from thread ID somehow like this: bits = 1UL << ((odp_thread_id() * 33) % 64) (and similarly for the 32-bit values).

TuomasTaipale: Is "* 33" somehow specifically important here? If it is, I should probably add a comment about it in the code.

JannePeltonen: It changes the order of the bits tested to avoidf the problem that with a limited number of threads only the lowest orders bits are tested. With that, the high order half of the 64 bit atomic gets tested already with two workers. Actually, 41 would be even better than 33 as it would distribute the bits better with e.g. 4 threads.

JannePeltonen: It would of course be better to have that in a function, e.g. thread_id_to_bit_pattern_u64().

TuomasTaipale: OK, neat. Refactored in v3.

JannePeltonen (approved):

MatiasElo:
If there is still enough time before the next release support for these new functions could be added to `odp_atomic_perf` application.

test/performance/odp_atomic_perf.c
```diff
@@ -620,7 +620,75 @@ static inline void test_atomic_xchg_u64(void *val, void *out, uint32_t num_round
...
+		odp_atomic_bit_set_u32(atomic_val, 0x1);
+}
+
+static inline void test_atomic_bit_fetch_set_u32(void *val, void *out ODP_UNUSED,
+						 uint32_t num_round)
+{
+	odp_atomic_u32_t *atomic_val = val;
+
+	for (uint32_t i = 0; i < num_round; i++)
+		(void)odp_atomic_bit_fetch_set_u32(atomic_val, 0x1);
```
psavol, line 637: Fetch type API calls should use the return value for something, so that compiler will not change the atomic instruction to a store type version of it.

See previous tests for fetch type calls:

`
for (uint32_t i = 0; i < num_round; i++)
		ret += odp_atomic_fetch_dec_u64(atomic_val);
	*result = ret;
`

psavol (approved):
Otherwise OK, fetch type atomics perf test needs small fix

jerinjacobk (approved):

PR #2103, opened 2024-08-22
[PATCH v3] Fix make check in i386

JannePeltonen:
A couple of small changes to fix make check in 32-bit x86.


JereLeppanen (approved):

JannePeltonen:
v2: added review tag

PR #2104, opened 2024-08-22
[PATCH v3] configure: enable -fno-strict-aliasing option

JereLeppanen:
Enable the -fno-strict-aliasing compiler option. Some sections of code in ODP deliberately access the same data via pointers to different types, which is undefined behavior in C. The -fno-strict-aliasing option prevents the compiler from making assumptions about aliasing in these instances.


JannePeltonen (approved):

JereLeppanen:
v2:
- Add review tag.

PR #2107, opened 2024-08-27
[PATCH v5] linux-gen: random: make random data independent of destination buffer alignment

JereLeppanen:


platform/linux-generic/odp_random_std.c
```diff
@@ -40,38 +41,35 @@ static int32_t _random_data(uint8_t *buf, uint32_t len, uint64_t *seed)
...
+	if (!_ODP_UNALIGNED && (uintptr_t)buf & 3) {
+		for (uint32_t i = 0; i < len / 4; i++) {
+			uint32_t r = xorshift64s32(seed);
 
-		if (((uintptr_t)buf & 1) && len) {
-			*(uint8_t *)buf = r & 0xff;
-			r >>= 8;
-			buf += 1;
-			len -= 1;
+			memcpy(buf, &r, 4);
```
JannePeltonen, line 48: This memcpy does not appear to get optimized away when compiling with GCC 12 for armv6+fp with -mno-unaligned-access. So if we do want to optimize for the _ODP_UNALIGNED == 0 case, the memcpy should perhaps be replaced by direct assignments to buf.

platform/linux-generic/odp_random_std.c
```diff
@@ -40,38 +41,35 @@ static int32_t _random_data(uint8_t *buf, uint32_t len, uint64_t *seed)
...
+			uint32_t r = xorshift64s32(seed);
+
+			memcpy(buf, &r, len);
 		}
+
+		return ret;
+	}
+
+	for (uint32_t i = 0; i < len / 4; i++) {
+		*(odp_una_u32_t *)buf = xorshift64s32(seed);
```
JannePeltonen, line 64: This could be memcpy (at least if we do not care too much about _ODP_UNALIGNED 0) to avoid alignment and aliasing problems and the need to use weird type attributes. The memcpy would get optimized away when compiling for x86_64 or A64. And same comments for the other occurences of this sort of assignments.

test/validation/api/random/random.c
```diff
@@ -149,6 +149,26 @@ static void random_test_align_and_overflow_true(void)
 	random_test_align_and_overflow(ODP_RANDOM_TRUE);
 }
 
+static void random_test_align_and_len_test(void)
```
JannePeltonen, line 152: This looks good. I just wonder if it would be worthwhile to add checks that the function is not writing more than len bytes in the buffer.

platform/linux-generic/odp_random_std.c
```diff
@@ -40,38 +41,35 @@ static int32_t _random_data(uint8_t *buf, uint32_t len, uint64_t *seed)
 {
 	const uint32_t ret = len;
 
-	if (!_ODP_UNALIGNED) {
-		uint32_t r = xorshift64s32(seed);
+	if (!_ODP_UNALIGNED && (uintptr_t)buf & 3) {
```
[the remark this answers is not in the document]
JereLeppanen: Removed in v4.

platform/linux-generic/odp_random_std.c
JereLeppanen: I'm not in favor of replacing every unaligned integer assignment with memcpy. But in any case that's outside the scope of this PR, so I left this as-is.

test/validation/api/random/random.c
JereLeppanen: That and the destination buffer alignment are already tested in random_test_align_and_overflow. So if random test data is allowed to differ for different len values, even if seed is the same, then there's no need for this new test. Removed in v4.

JereLeppanen:
v4:
- Janne's comments.
- Rebase.
- Add review tags.

JannePeltonen (approved):

PR #2114, opened 2024-09-04
[PATCH v1] test: misc: build test application with strict aliasing

JereLeppanen:
Add -fstrict-aliasing compiler option to odp_api_headers test application. This tests that a trivial ODP application can be built and run with strict aliasing.


JereLeppanen:
Not needed. We assume that also application is built with no-strict-aliasing.

PR #2115, opened 2024-09-04
[PATCH v4] helper: string: ignore stringop-truncation warning in odph_strcpy

JereLeppanen:


test/performance/odp_packet_gen.c
```diff
@@ -86,7 +86,7 @@ enum {
 ODP_STATIC_ASSERT(MAX_PKTIOS <= UINT8_MAX, "Interface index must fit into uint8_t\n");
 
 typedef struct {
-	char name[MAX_HDR_NAME_LEN + 2];
```
TuomasTaipale, line 89: Help text promises `MAX_HDR_NAME_LEN` as the maximum characters for a name, so one extra character for the null here and in the `odph_strcpy()` call below could be left.

JereLeppanen: In v2.

JereLeppanen:
v3:
- Add review tags.

PR #2116, opened 2024-09-06
[PATCH v4] test: packet_gen: add possibility to enable L4 checksum offloading

TuomasTaipale:
Add new -X option to enable L4 checksum offloading. This can boost performance in cases where checksum can be offloaded to be done by HW.

v2:
- Rebased

v3:
- Updated enabling logic

v4:
- Added reviewed-by tag

### DIFF
--- a/test/performance/odp_packet_gen.c
+++ b/test/performance/odp_packet_gen.c
@@ -132,6 +132,7 @@ typedef struct test_options_t {
 	odp_bool_t promisc_mode;
 	odp_bool_t calc_latency;
 	odp_bool_t calc_cs;
+	odp_bool_t cs_offload;
 	odp_bool_t fill_pl;
 
 	struct vlan_hdr {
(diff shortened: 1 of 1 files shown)


test/performance/odp_packet_gen.c
```diff
@@ -303,8 +304,11 @@ static void print_usage(void)
 	       "                            src_port/dst_port. Comma-separated (no spaces) list of\n"
 	       "                            count values: <src_port count>,<dst_port count>\n"
 	       "                            Default value: 0,0\n"
-	       "  -C, --no_udp_checksum     Do not calculate UDP checksum. Instead, set it to\n"
-	       "                            zero in every packet.\n"
+	       "  -C, --no_udp_checksum     Do not calculate UDP SW checksum. Instead, set it to\n"
+	       "                            zero in every packet, this may be overridden by\n"
+	       "                            '--checksum_offload' option.\n"
+	       "  -X, --checksum_offload    Enable L4 checksum offloads: checksums are checked\n"
+	       "                            at input and inserted at output.\n"
```
MatiasElo, line 311: Maybe there could be separate options for RX and TX checksum offloads. Also, if RX checksum offload is enabled, checksum validity could be checked during packet receive.

TuomasTaipale: Changed according to offline discussions in v3. Still just single option, but RX side offloading remains disabled with a warning log if packet I/O does not support also packet dropping on errors.

MatiasElo (approved):

PR #2117, opened 2024-09-09
[PATCH v10] api: add custom LSO operation for setting flag bits

JannePeltonen:
- add ODP_LSO_WRITE_BITS operation
- add ODP_LSO_ADD_SHIFTED_PAYLOAD_OFFSET operation

### DIFF
--- a/include/odp/api/spec/packet_io_types.h
+++ b/include/odp/api/spec/packet_io_types.h
@@ -1,6 +1,6 @@
 /* SPDX-License-Identifier: BSD-3-Clause
  * Copyright (c) 2013-2018 Linaro Limited
- * Copyright (c) 2020-2023 Nokia
+ * Copyright (c) 2020-2024 Nokia
  */
 
 /**
@@ -777,7 +777,10 @@ typedef enum odp_lso_modify_t {
 	ODP_LSO_ADD_PAYLOAD_LEN = 0x2,
 
 	/** Add number of payload bytes in all previous segments */
-	ODP_LSO_ADD_PAYLOAD_OFFSET = 0x4
+	ODP_LSO_ADD_PAYLOAD_OFFSET = 0x4,
+
+	/** Write bits in the first, middle and last segment */
+	ODP_LSO_WRITE_BITS,
 
 } odp_lso_modify_t;
 
@@ -856,6 +859,9 @@ typedef struct odp_lso_capability_t {
 
 		/** ODP_LSO_ADD_PAYLOAD_OFFSET support */
 		uint16_t add_payload_offset:1;
+
+		/** ODP_LSO_WRITE_BITS support */
+		uint16_t write_bits:1;
 
 	} mod_op;
 
@@ -1080,6 +1086,18 @@ typedef struct odp_pktio_capability_t {
 
 } odp_pktio_capability_t;
 
+/** Parameters for ODP_LSO_WRITE_BITS custom operation */
+typedef struct odp_lso_write_bits_t {
+	/** Bitmask to select which bits to write */
+	uint8_t mask[1];
+
+	/** Value to be written using the mask:
+	 *  new_value[n] = (old_value[n] & ~mask[n]) | (value[n] & mask[n])
+	 */
+	uint8_t value[1];
+
+} odp_lso_write_bits_t;
+
 /**
  * LSO profile parameters
  */
@@ -1089,8 +1107,10 @@ typedef struct odp_lso_profile_param_t {
 	 *
 	 * Selects on which protocol LSO operation performs segmentation (e.g. IP fragmentation vs.
 	 * TCP segmentation). When ODP_LSO_PROTO_CUSTOM is selected, only custom field
-	 * modifications are performed. The default value is ODP_LSO_PROTO_NONE. Check LSO
-	 * capability for supported protocols.
+	 * modifications are performed. Packet content is not modified when a packet is not
+	 * segmented.
+	 *
+	 * The default value is ODP_LSO_PROTO_NONE. Check LSO capability for supported protocols.
 	 */
 	odp_lso_protocol_t lso_proto;
 
@@ -1099,6 +1119,12 @@ typedef struct odp_lso_profile_param_t {
 	 *
 	 * Set lso_proto to ODP_LSO_PROTO_CUSTOM when using custom fields. Fields are defined
 	 * in the same order they appear in the packet.
+	 *
+	 * Fields may not modify overlapping packet bytes except as follows:
+	 * - An ODP_LSO_WRITE_BITS operation may modify the same byte as an ODP_LSO_ADD_* operation.
+	 *   In that case the write-bits operation is done after the add operation, overwriting
+	 *   some of the bits written by the add operation. Such write-bits field must appear
+	 *   after the overlapping add field.
 	 */
 	struct {
 		/** Custom field to be modified by LSO */
@@ -1111,9 +1137,23 @@ typedef struct odp_lso_profile_param_t {
 			/** Field offset in bytes from packet start */
 			uint32_t offset;
 
-			/** Field size in bytes. Valid values are 1, 2, 4, and 8 bytes. */
+			/** Field size in bytes. Valid values are 1, 2, 4, and 8 bytes.
+			 *  Must be set to 1 in ODP_LSO_WRITE_BITS operation.
+			 */
 			uint8_t size;
 
+			/** Operation specific parameters */
+			union {
+				/** Parameters for ODP_LSO_WRITE_BITS operation */
+				struct {
+					/** bits to write in the first segment */
+					odp_lso_write_bits_t first_seg;
+					/** bits to write in middle segments */
+					odp_lso_write_bits_t middle_seg;
+					/** bits to write in the last segment */
+					odp_lso_write_bits_t last_seg;
+				} write_bits;
+			};
 		} field[ODP_LSO_MAX_CUSTOM];
 
 		/** Number of custom fields specified. The default value is 0. */
--- a/example/sysinfo/odp_sysinfo.c
+++ b/example/sysinfo/odp_sysinfo.c
@@ -588,6 +588,7 @@ static void print_pktio_capa(appl_args_t *appl_args)
 		printf("    lso.mod_op.add_payload_len:    %u\n", capa->lso.mod_op.add_payload_len);
 		printf("    lso.mod_op.add_payload_offset: %u\n",
 		       capa->lso.mod_op.add_payload_offset);
+		printf("    lso.mod_op.write_bits:         %u\n", capa->lso.mod_op.write_bits);
 		printf("    lso.max_num_custom:            %u\n", capa->lso.max_num_custom);
 		printf("    lso.proto.custom:              %u\n", capa->lso.proto.custom);
 		printf("    lso.proto.ipv4:                %u\n", capa->lso.proto.ipv4);
--- a/platform/linux-generic/odp_packet_io.c
+++ b/platform/linux-generic/odp_packet_io.c
@@ -1593,6 +1593,7 @@ int odp_pktio_capability(odp_pktio_t pktio, odp_pktio_capability_t *capa)
 	capa->lso.proto.ipv4             = 1;
 	capa->lso.proto.custom           = 1;
 	capa->lso.mod_op.add_segment_num = 1;
+	capa->lso.mod_op.write_bits      = 1;
 
 	capa->tx_compl.queue_type_sched = 1;
 	capa->tx_compl.queue_type_plain = 1;
@@ -2861,8 +2862,16 @@ odp_lso_profile_t odp_lso_profile_create(odp_pktio_t pktio, const odp_lso_profil
 				return ODP_LSO_PROFILE_INVALID;
 			}
 
-			/* Currently only segment number supported */
-			if (mod_op != ODP_LSO_ADD_SEGMENT_NUM) {
+			switch (mod_op) {
+			case ODP_LSO_ADD_SEGMENT_NUM:
+				break;
+			case ODP_LSO_WRITE_BITS:
+				if (size != 1) {
+					_ODP_ERR("Bad custom field size %u\n", size);
+					return ODP_LSO_PROFILE_INVALID;
+				}
+				break;
+			default:
 				_ODP_ERR("Custom modify operation %u not supported\n", mod_op);
 				return ODP_LSO_PROFILE_INVALID;
 			}
@@ -2999,7 +3008,7 @@ static int lso_update_ipv4(odp_packet_t pkt, int index, int num_pkt,
 	return ret;
 }
 
-static int lso_update_custom(lso_profile_t *lso_prof, odp_packet_t pkt, int segnum)
+static int lso_update_custom(lso_profile_t *lso_prof, odp_packet_t pkt, int segnum, int num_segs)
 {
 	void *ptr;
 	int i, mod_op;
@@ -3046,6 +3055,17 @@ static int lso_update_custom(lso_profile_t *lso_prof, odp_packet_t pkt, int segn
 				u16 = odp_cpu_to_be_16(segnum + odp_be_to_cpu_16(u16));
 			else
 				u8 += segnum;
+		} else if (mod_op == ODP_LSO_WRITE_BITS) {
+			odp_lso_write_bits_t bits;
+
+			if (segnum == 0)
+				bits = lso_prof->param.custom.field[i].write_bits.first_seg;
+			else if (segnum == num_segs - 1)
+				bits = lso_prof->param.custom.field[i].write_bits.last_seg;
+			else
+				bits = lso_prof->param.custom.field[i].write_bits.middle_seg;
+
+			u8 = (u8 & ~bits.mask[0]) | (bits.value[0] & bits.mask[0]);
 		}
 
 		if (odp_packet_copy_from_mem(pkt, offset, size, ptr)) {
@@ -3207,7 +3227,7 @@ int _odp_lso_create_packets(odp_packet_t packet, const odp_packet_lso_opt_t *lso
 		int num_custom = lso_prof->param.custom.num_custom;
 
 		for (i = 0; num_custom && i < num_pkt; i++) {
-			if (lso_update_custom(lso_prof, pkt_out[i], i)) {
+			if (lso_update_custom(lso_prof, pkt_out[i], i, num_pkt)) {
 				_ODP_ERR("Custom field update failed. Segment %i\n", i);
 				goto error;
 			}
--- a/test/validation/api/pktio/lso.c
+++ b/test/validation/api/pktio/lso.c
@@ -22,13 +22,23 @@
 #define LSO_TEST_MARKER_ETHERTYPE 0x88B6 /* Local experimental Ethertype */
 #define LSO_TEST_CUSTOM_ETHERTYPE 0x88B5 /* Must match test packets. */
 #define LSO_TEST_CUSTOM_ETH_SEGNUM_OFFSET 16
+#define LSO_TEST_CUSTOM_ETH_BITS_OFFSET LSO_TEST_CUSTOM_ETH_SEGNUM_OFFSET /* Intentional overlap */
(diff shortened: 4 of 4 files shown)


JannePeltonen:
v2: fixed a comment bug regarding the effect of the mask field of odp_lso_write_bits_t.

include/odp/api/spec/packet_io_types.h
```diff
@@ -1114,6 +1120,20 @@ typedef struct odp_lso_profile_param_t {
...
+			/** Operation specific parameters */
+			union {
+				/** parameters for ODP_LSO_ADD_SHIFTED_PAYLOAD_OFFSET */
+				struct {
+					/** Number of bits to shift the number of payload bytes
+					 *  right before adding it to the value in packet.
+					 *  Affects also segmentation so that segment boundaries
+					 *  are at offsets divisible by 1<<shift (i.e. the
+					 *  least significant bits shifted away are always zero).
+					 */
```
ksatha, line 1159: The shift was applied to the payload_offset/fragment_offset as the name suggests, not to the payload bytes. The payload length was the same for all segments except the last one (based on total packet length), and it is divisible by (1 << shift). The payload_offset was the cumulative length of the previous fragment’s payload lengths. The shift was applied to this cumulative payload_offset to fit it into the header field. With the shift, the Nth fragment’s payload offset should be:

[ ((n - 1) * payload_length) >> shift ]

JannePeltonen: Maybe I should read "was" as "should be" in the comment? Since this is a proposed new feature in the API, there are no pre-existing implementations of it. That said, if something in the proposal is somehow inconsistent or hard to implement (e.g. due to potential HW limitations), you can point out the issues and propose changes.

Since the shifting causes the least significant bits of the segment offset to not be present in the packet, it would be problematic if they would be nonzero (i.e. if the segment offset were not divisible by 1<<shift). How would the receiver then know what the true offset is? Because of that, this API change proposal specifies that the segmentation occurs at suitable boundaries for the shift value. Now I notice that the API text should be improved/fixed by changing "are at offsets" to "are at payload offsets", but other than that, I do not see what the problem is.

> The shift was applied to the payload_offset/fragment_offset as the name suggests, not to the payload bytes.

I am not sure I understand this comment. According to the API text, the shift is not applied to payload bytes, but the _number_ of payload bytes (in all previous segments). And the number of payload bytes in all previous segments is the same as the payload offset.

> The payload length was the same for all segments except the last one (based on total packet length), and it is divisible by (1 << shift). 

According to the existing API spec, the payload length is not necessarily the same in all segments of a single packet and there is no guarantee that the payload length is divisible by anything. See odp_packet_lso_opt_t::max_payload_len. Therefore I propose to add this statement in the API, to guarantee that segment boundaries are suitable if the offset shifting is used.

> The payload_offset was the cumulative length of the previous fragment’s payload lengths. The shift was applied to this cumulative payload_offset to fit it into the header field 

Yes, isn't that what the API says?

> With the shift, the Nth fragment’s payload offset should be: [ ((n - 1) * payload_length) >> shift ]

Only if the underlying implementation works so that all the previous segments have payload_length payload length.

include/odp/api/spec/packet_io_types.h
```diff
@@ -1086,6 +1092,18 @@ typedef struct odp_pktio_capability_t {
 
 } odp_pktio_capability_t;
 
+/** Parameters for ODP_LSO_WRITE_BITS custom operation */
+typedef struct odp_lso_write_bits_t {
+	/** Bitmask to select which bits to write */
+	uint8_t mask;
```
ksatha, line 1098: Change to uint16_t to support 16-bit flags.

include/odp/api/spec/packet_io_types.h
```diff
@@ -1086,6 +1092,18 @@ typedef struct odp_pktio_capability_t {
...
+
+	/** Value to be written using the mask:
+	 *  new_value := (old_value & ~mask) | (value & mask)
+	 */
+	uint8_t value;
```
ksatha, line 1103: Change to uint16_t to support 16-bit flags.

include/odp/api/spec/packet_io_types.h
```diff
@@ -1086,6 +1092,18 @@ typedef struct odp_pktio_capability_t {
 
 } odp_pktio_capability_t;
 
+/** Parameters for ODP_LSO_WRITE_BITS custom operation */
+typedef struct odp_lso_write_bits_t {
+	/** Bitmask to select which bits to write */
+	uint8_t mask;
+
+	/** Value to be written using the mask:
+	 *  new_value := (old_value & ~mask) | (value & mask)
```
ksatha, line 1101: Both value and mask are user arguments for the profile, so there’s no need for the extra & operation. We can change this to new_value := (old_value & ~mask) | value.

include/odp/api/spec/packet_io_types.h
JannePeltonen: uint16_t seems quite arbitrary. Why not then uint32_t or uint64_t? I think uint8_t keeps the API simpler and still allows 16-bit flags through the use of two WRITE_BITS operations.

include/odp/api/spec/packet_io_types.h
JannePeltonen: This is just a matter of taste regarding how the API is defined without effect on the provided functionality. I think it is clearer that the mask field defines which bits are going to be affected so that other bits in the value field not have any effect. I do not see why your proposal would be preferable as opposed to just different.

include/odp/api/spec/packet_io_types.h
ksatha: The IPv4 16-bit fragment offset is shared with flags. Hardware (HW) may modify both fields in a single operation, so the flags will be 16-bit. This will save the number of operations in a custom profile, which are limited in HW. Even though it is 16-bit, based on size, it may modify a single byte as well.

include/odp/api/spec/packet_io_types.h
ksatha: What if other bits in the value field need to be set irrespective of the old value?

include/odp/api/spec/packet_io_types.h
JannePeltonen: The way things are done in the HW is different from the way things are expressed in the API. The ODP implementation can easily combine multiple operations into one and change their width if that is more efficient in the HW.

include/odp/api/spec/packet_io_types.h
JannePeltonen: I think I may not understand the question. The bits selected by the mask will be set of cleared based on the value irrespective of the old value of those bits in the destination byte.

include/odp/api/spec/packet_io_types.h
ksatha: So let's make this as follows(some thing similar), so that it is flexible and easy to use for different implementations 
uint8_t mask[8];
uint8_t value[8];
and based on size argument we will select number of bytes from this array. As per spec max valid value for size was 8, so make this array to 8 bytes.

include/odp/api/spec/packet_io_types.h
ksatha: Sorry for the confusion, my intention was as follows:

1) Mask Application: The bits specified by the mask will clear the corresponding bits in old_value and set them based on the same bit positions from value.

2) Remaining Bits: The bits in old_value that are not covered by the mask can either remain untouched or be selectively enabled based on value. Since the mask does not cover these bits, their behavior is implementation-defined. Therefore, the API specification will not explicitly state that these bits are untouched.

include/odp/api/spec/packet_io_types.h
JannePeltonen: We can consider that. But it is not easier for implementations than the fixed 1 byte width, since now the implementation does not only have to handle 1 byte widths and map them to HW but it also has to handle all the other widths. So the rationale for the change cannot be ease of implementation but rather convenience of the user of the API or cleanliness of the API (since the sizeis a common parameter and not an operation specific parameter, which it could also be).

include/odp/api/spec/packet_io_types.h
JannePeltonen: I still do not quite get it. The API says that the mask selects which bits to write. So the bits not selected by the mask are not written to, so they retain their values. Also, the expression in the API text (new_value := (old_value & ~mask) | (value & mask)) clarifies what is going to happen. Thus I do not see any ambiguity here regarding the bits not set in the mask.

include/odp/api/spec/packet_io_types.h
ksatha: It is easy to implement, and mapping multi-byte to hardware is simpler compared to the previous method. Regarding larger widths, if the hardware supports only X bytes, then sizes greater than X will return -ENOTSUP.

include/odp/api/spec/packet_io_types.h
ksatha: Assuming bits not selected by the mask should always be 1 irrespective of old_value is also a valid case. Take the examples below:

Formula 1: new_value := (old_value & ~mask) | (value & mask)

Case 1:

old_value: 0b11001100
mask: 0b11110000
value: 0b10101010
new_value: 0b10101100

Case 2:

old_value: 0b11001100
mask: 0b11110000
value: 0b10100000
new_value: 0b10101100

Formula 2: new_value := (old_value & ~mask) | value

Case 1:

old_value: 0b11001100
mask: 0b11110000
value: 0b10101010
new_value: 0b10101110

Case 2:

old_value: 0b11001100
mask: 0b11110000
value: 0b10100000
new_value: 0b10101100

With your Formula 1, bits not set in the mask remain the same irrespective of value, but the user can select a suitable output based on different values as mentioned in Formula 2. I think either the API will not specify anything about bits not selected by the mask, or it should be generic like Formula 2.

include/odp/api/spec/packet_io_types.h
ksatha: >2) separate the payload offset divisibility requirement from the shift value to a separate parameter.
That should be fine. Please provide a proposal that ensures if we select a non-zero shift value, then payload_length should be divisible by (1 << shift).

include/odp/api/spec/packet_io_types.h
JannePeltonen: > mapping multi-byte to hardware is simpler compared to the previous method

It is not simpler but more difficult. The proposed method is essentially the same as your proposal with the additional restriction that size must always be 1 byte in the WRITE_BITS operation. Giving more widths for the user the select from does not make the mapping easier but potentially more difficult. Even if we supported multiple widths, the API user could still use 1-byte operation (e.g. in IPv4 the fragmentation related flags are contained in one byte) and you would need to map that to a 16-bit operation if that is all your HW has.

IOW, ODP API is an abstraction to enable the user to tell what kind of LSO behaviour she wants. ODP implementation then maps that to the HW if at all possible. It is no ok to unnecessarily expose HW details in the API (e.g. have byte-wide WRITE_BITS operation fail in some ODP implementation and version just because the HW has only 16-bit wide operations natively).

> Regarding larger widths, if the hardware supports only X bytes, then sizes greater than X will return -ENOTSUP.

No, the ODP implementation is supposed to map the LSO profile to the HW if it is possible to do so. So larger widths in the profile would correspond to multiple narrower operations in the HW. Profile creation should fail only if it is not possible to do the mapping (e.g. there is only one 8-bit operation in the HW and the LSO requests a wider operation with mask bits set in more than 1 byte).

include/odp/api/spec/packet_io_types.h
JannePeltonen: > Assuming bits not selected by the mask should always be 1 irrespective of old_value is also a valid case.

Why? If you want to set certain bit, then you can include that bit in the mask.


> With your Formula 1, bits not set in the mask remain the same irrespective of value

Yes, that is correct.

> but the user can select a suitable output based on different values as mentioned in Formula 2

Yes they could, but they could do the same by simply setting the necessary bits in the mask.

> I think either the API will not specify anything about bits not selected by the mask

That would not be a good idea.

> or it should be generic like Formula 2.

But why? What would be the benefit? Now the API specifies that the bits not set in the mask are not changed in the target byte. I think that is quite simple and clear.

ksatha: As per the spec, we have two things:

1) Functionality for “bits set in mask.” We are both on the same page here.
2) Functionality for “bits not set in the mask.” You are proposing the new value will be the same as the old value, whereas I am proposing the new value will be “old value ORed with the given value.” Given an example where Formula 2 works for both cases, whereas Formula 1 does not.
The implementation is simple for both. It is up to you to pick the formula that supports multiple cases without changing the mask, or you can stick to a single functionality.

include/odp/api/spec/packet_io_types.h
JannePeltonen: > How would the API know that the first 2 bytes are one operation and the remaining 4 bytes are another operation?

There is no 2-byte and 4-byte operation there, but just a bunch of bits that need to be changed. The way the bit changes are split to some HW primitives is an implementation detail that does not concern the API. From the API point-of-view there is only one operation, the segmentation operation. The API just tells unambiguously which bits to change and how, the ODP implementation then maps the request to whatever update primitives it needs to use with the underlying HW. Only the externally observable behaviour matters, so it does not make any difference whether the implementation maps the request to one 2-byte HW op and one 4-byte op or vice versa or something else.

> As I proposed, if you change this to an array of masks and values, it becomes very easy to map for all 1-byte, 2-byte, 4-byte, and 8-byte combinations.

How does it become easier? In the above example, the user could still configure the profile with 6 1-byte operations and the implementation would have to make exactly the same mappings to the 2 and 4 byte HW ops. We do not want to have profile creation fail if the width used there does not match whatever happens to be the width of the HW primitive as that would unnecessarily expose implementation details to the API user and would make using the API more difficult.

IOW, if we change the API to allow more ways to configure the same thing, the implementation does not become easier but more difficult. So ease of implementation cannot be reason to make the API larger. The implementation become easier only if you add restriction in the API, but we do not want to add HW-dependent restrictions which can be avoided by a bit of logic in the implementation.

ksatha: Just visualize it from a header modification point of view. On platform-B, we need to perform two modifications:
1. At X-offset, we need to modify a 16-bit header field.
2. At Y-offset, we need to modify a 32-bit header field.

In my proposal, it is simple with two operations:
1. Operation: write_bits
   - Offset: X 
   - Size: 2 
   - Mask: [2]
   - Value: [2]

2. Operation: write_bits
   - Offset: Y
   - Size: 4
   - Mask: [4]
   - Value: [4]

How does your proposal support this? This is a valid test case for header modifications.

ksatha: The basis for my proposal is that combining bytes into a single operation is driven by the header fields of protocols rather than hardware requirements.

JannePeltonen: > How does your proposal support this?

Through 2 or more byte-wide write_bits operations. What is the problem?

ksatha: Assume (X = 10) and (Y = 12) in the above example. Could you describe how your 6 operations looks like?

JannePeltonen: > Assume (X = 10) and (Y = 12) in the above example. Could you describe how your 6 operations looks like?

That would be configured through an LSO profile with ODP_LSO_PROTO_CUSTOM lso_proto and with as many write_bits operations as there are bytes that need to be changed. If you need to change bits in all of the 6 bytes in your "header fields", then you would need to have 6 custom fields in the LSO profile, each field containing a write_bits operation for one of the bytes (i.e. the offsets in the ops would be 10, 11, 12, …, 15).

> because from protocol POV it is 2 field changes not 48 bit (6-byte) operations

We do not care about the protocol point-of-view. Custom LSO profile is not aware and does not need to be aware of the protocol carried in the packet or whatever 'fields' the protocol might define. For LSO purposes (with custom protocol) the packet is just a sequence of bytes that we modify using the modify operations the way we wish. The only thing LSO needs to know is which part of the packet is considered header and which payload for segmentation purposes, but that is not related to the write_bits operation.

> In my proposal, it is simple with two operations:

Yes, in this particular case it might be simpler for the API user if there are to-be-modified bits in more than one byte in each of your "header fields". But for an ODP implementation it would be a larger API and therefore more difficult to implement, not less, as you claimed before.

include/odp/api/spec/packet_io_types.h
JannePeltonen: We seem to be going in circles.

The proposed API change in this PR makes it possible to indicate which packet bits are to be changed and in which way at segmentation time, don't you agree? The API also does not leave any ambiguity on the outcome of any valid configuration is. I still do not see what problem you see in the API.

> Given an example where Formula 2 works for both cases, whereas Formula 1 does not.

I really do not understand what you mean when you say that formula 1 "does not work". Can you elaborate? In which situation does it not work?

> Functionality for “bits not set in the mask.” You are proposing the new value will be the same as the old value, 

Correct. The mask specifies which bits are to be changed. The value specifies the new value of those bits. Value bits that are outside the masked bits are thus ignored. It seems that you have a problem with having those bits ignored, but I do not understand what the problem is, other than that you do not like it. If an API user does not want to have certain bit ignored, then he/she can simply include it in the mask. That is what the mask is for.

> It is up to you to pick the formula that supports multiple cases without changing the mask, or you can stick to a single functionality.

There is only single functionality here, not multiple ones. The single functionality is to set selected bits to selected values. That is all. The fact that your API alternative makes it possible to specify the desired outcome in two different ways does not mean that there are two functionalities that could be supported. It only means that your API has two ways to express the same thing, which some might see as a weakness rather than strength. I think the API is more clear when the mask indicates which bits and which bits only are to be written to.

include/odp/api/spec/packet_io_types.h
JannePeltonen: As discussed offline, max_payload_len of odp_packet_lso_opt_t specifies the maximum payload length in a segment but allows ODP implementations to create segments with less payload bytes, e.g. due to HW limitations or for alignment purposes. Thus having the shift value affect the generated segments lengths should be ok. We can still consider having a separate parameter, but maybe it is not really that useful.

include/odp/api/spec/packet_io_types.h
JannePeltonen: As discussed offline, the rationale for having multiple widths for the write_bits operation cannot be ease of implementation (of ODP) since with multiple widths the implementation would still have to be able 1-byte wide write_bits if the application happens to use them. Multiple widths, while not functionally adding anything, might be more convenient for some applications.

Another issue is the maximum number of custom fields, which one might face sooner with 1-byte wide operations than with wider operations. But maybe that is not a real concern in practice at this point. If we do not support multiple widths, maybe we specify that the size parameter has to be one, to make it easier to add wider write_bits operation later in a backward compatible manner.

include/odp/api/spec/packet_io_types.h
JannePeltonen: As discussed offline, both ways of expressing the bits-to-be-changed and their new values are essentially equivalent and express the same thing in slightly different way. So which one to choose is just a matter of taste and does not have any effect on what can or can not be done.

JannePeltonen:
v3:
- Removed the commit that added the shifted offset add operation as it does not appear to be needed now.
- Turned the mask and value fields of odp_lso_write_bits_t into byte arrays to make it possible to support multi-byte write-bits in the future without breaking backward compatibility. Required that the size parameter is set to 1 for the write-bits operation.
- Specified that a write-bits operation that overlaps with an add operation must appear after it in the field array.

jerinjacobk (approved):

JannePeltonen:
v4: added implementation and tests

JannePeltonen:
v5: added a commit that adds write_bits capability printout in odp_sysinfo.

JannePeltonen:
v6: rebased

JannePeltonen:
v7: fixed a white space issue cought by checkpatch

MatiasElo (approved):

psavol (approved):
Reviewed and approved API commit.

PR #2124, opened 2024-09-23
[PATCH v2] helper: ip: avoid using a VLA in a header because of C++ apps

JannePeltonen:
Variable length arrays are not allowed in C++, so do not use one in a header file that may be included by a C++ application. Clang++-18 complains about it, breaking the build.

### DIFF
--- a/helper/include/odp/helper/ip.h
+++ b/helper/include/odp/helper/ip.h
@@ -112,7 +112,8 @@ static inline int odph_ipv4_csum(odp_packet_t pkt,
 				 odp_u16sum_t *chksum)
 {
 	uint32_t nleft = (uint32_t)(ODPH_IPV4HDR_IHL(ip->ver_ihl) * 4);
-	uint16_t buf[nleft / 2];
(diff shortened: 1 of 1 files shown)


helper/include/odp/helper/ip.h
```diff
@@ -112,7 +112,8 @@ static inline int odph_ipv4_csum(odp_packet_t pkt,
 				 odp_u16sum_t *chksum)
 {
 	uint32_t nleft = (uint32_t)(ODPH_IPV4HDR_IHL(ip->ver_ihl) * 4);
-	uint16_t buf[nleft / 2];
+	const int max_ip_header_len = 15 * 4;
```
TuomasTaipale, line 115: 15 * 4 could be useful as a public constant as well (e.g. `ODPH_IPV4HDR_MAX_LEN`).

TuomasTaipale: As per offline discussions, leaving this as is to avoid public changes.

TuomasTaipale (approved):

PR #2125, opened 2024-09-24
[PATCH v2] github_ci: move arm64 pipeline to graviton3 runners

MatiasElo:
Test running arm64 CI using Graviton3 runners.


JereLeppanen (approved):
Could mention where "ah-ubuntu_22_04-c7g_2x-50" comes from.

PR #2126, opened 2024-09-25
[PATCH v8] api: init: add odp_log_fn_get() and odp_abort_fn_get()

JereLeppanen:


JereLeppanen:
v2:
- Small cleanups.
- Add odp_abort_fn_get().

include/odp/api/spec/init.h
```diff
@@ -402,6 +402,22 @@ int odp_term_abnormal(odp_instance_t instance, uint64_t flags, void *data);
  */
 void odp_log_thread_fn_set(odp_log_func_t func);
 
+/**
+ * Get current log function
+ *
+ * May be called before odp_init_global().
```
TuomasTaipale, line 408: Should this be unified with the newer PR (#2143): This one or "This function may be called before ODP initialization."

(Same for `odp_abort_fn_get()`.)

test/validation/api/init/init_main.c
```diff
@@ -168,6 +168,40 @@ static void init_test_log_thread(void)
...
 
+static void init_test_log_fn_get(void)
+{
+	int ret;
+	odp_instance_t instance;
+	odp_init_t param;
+
+	odp_init_param_init(&param);
+
+	ret = odp_init_global(&instance, &param, NULL);
```
TuomasTaipale, line 209: Could also test that `odp_log_fn_get()` returns non-`NULL` before initialization as per the spec.

platform/linux-generic/odp_init.c
```diff
@@ -735,6 +735,14 @@ odp_log_func_t odp_log_fn_get(void)
 	return _odp_log_fn_get();
 }
 
+odp_abort_func_t odp_abort_fn_get(void)
+{
+	if (_odp_abort_fn)
+		return _odp_abort_fn;
+
+	return odp_override_abort;
```
TuomasTaipale, line 743:
```c
return _odp_abort_fn ? _odp_abort_fn : odp_override_abort;
```

include/odp/api/spec/init.h
JereLeppanen: Changed to "... if ODP is not initialized".

test/validation/api/init/init_main.c
JereLeppanen: Added a check for this.

include/odp/api/spec/init.h
```diff
@@ -402,6 +402,22 @@ int odp_term_abnormal(odp_instance_t instance, uint64_t flags, void *data);
...
+ * Get current log function
+ *
+ * May be called even if ODP is not initialized.
+ *
+ * Returns the function previously set by the calling thread via
+ * odp_log_thread_fn_set(). If no thread specific log function has been set,
+ * returns the log function specified in odp_init_global(). If no log function
+ * was specified in odp_init_global(), or if odp_init_global() has not been
+ * called yet, returns the default or override log function (see
+ * odp_override_log()). Always returns a valid log function, never NULL.
```
MatiasElo, line 415: We discussed about this with Petri  and concluded that it could be safer if the function (same for `odp_abort_fn_get()`) would always return NULL if ODP has not yet been initialized. This way log function implementation stays simpler and we avoid the potential pitfalls of log function having to work correctly even before ODP initialization. If NULL would be returned, the user of this API (e.g. `ODPH_ERR()`) could then implement the default print functionality itself.

JereLeppanen: In v5.

helper/include/odp/helper/debug.h
```diff
@@ -70,10 +78,8 @@ do { \
 } while (0)
 
 /**
- * Assert macro for applications and helper code
- *
- * No code is generated when ODPH_DEBUG=0. Prints error message and aborts when
- * ODPH_DEBUG=1 and 'cond' is false.
+ * Runtime assertion macro. No code is generated when ODPH_DEBUG=0. Prints error
```
MatiasElo, line 81: Could mention `--enable-helper-debug` configure option for setting `ODPH_DEBUG`.

test/validation/api/init/init_main.c
```diff
@@ -168,6 +168,44 @@ static void init_test_log_thread(void)
...
+	CU_ASSERT_FATAL(ret == 0);
+
+	ret = odp_init_local(instance, ODP_THREAD_WORKER);
+	CU_ASSERT_FATAL(ret == 0);
+
+	/* Default log function is not NULL. */
+	odp_log_func_t log_fn_def = odp_log_fn_get();
+
+	CU_ASSERT(log_fn_def != NULL);
```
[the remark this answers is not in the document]
JereLeppanen: Some function, or specifically this one? The default function would just write to a log that no one will ever check.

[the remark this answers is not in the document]
MatiasElo: Just some dummy print with the returned function to check it doesn't crash.

jerinjacobk (approved):

MatiasElo (approved):
The first commit should disappear after a rebase and one minor comment but looks good otherwise.

TuomasTaipale (approved):

test/validation/api/init/init_main.c
```diff
@@ -168,6 +168,45 @@ static void init_test_log_thread(void)
...
```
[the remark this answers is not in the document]
JereLeppanen: In v7.

JereLeppanen:
Added review tags in v7.

psavol (approved):
Reviewed and approved all three API commits.

JereLeppanen:
v8:
- Rebase.
- Add review tags (Petri).

PR #2128, opened 2024-09-30
[PATCH v4] linux-gen: remove support for EOL DPDK versions

MatiasElo:
V2:
- Fixed Janne's comment about Debian/Ubuntu versions
- Removed more deprecated DPDK instructions from `DEPENDENCIES`


DEPENDENCIES
```diff
@@ -175,25 +175,7 @@ Prerequisites for building the OpenDataPlane (ODP) API
    # Debian/Ubuntu starting from 20.04
```
JannePeltonen, line 175: > Debian/Ubuntu starting from 20.04

This looks a bit confusing since
- 20.04 is not a Debian version number and it looks like a DPDK version number (like 20.05)
- The dpdk-dev package in Ubuntu 20.04 contains dpdk version 19.11 which is no longer supported after this patch

I suppose you could to refer to Ubuntu 22.04 and Debian 12 here as those have dpdk 21.11 and 22.11, respectively.

MatiasElo: Fixed in V2. Removed also the chapter above about DPDK requirements.

JannePeltonen (approved):

PR #2129, opened 2024-10-01
test: queue_perf: implement result export

Rayska:
odp_queue_perf now supports .csv export


test/performance/odp_queue_perf.c
```diff
@@ -547,6 +550,25 @@ static void print_stat(test_global_t *global)
...
+			test_common_write_term();
+			return -1;
+		}
+		if (test_common_write("%f,%f,%f,%" PRIu64 ",%" PRIu64 "\n",
+				      cycles_ave / events_ave,
+				      (1000.0 * events_ave) / nsec_ave,
+				      (1000.0 * events_sum) / nsec_ave,
+				      deq_retry_sum,
+				      enq_retry_sum)) {
+			test_common_write_term();
```
MatiasElo, line 566: Looks like `test_common_write_term()` is not called in case of success.

test/performance/odp_queue_perf.c
```diff
@@ -622,7 +652,7 @@ int main(int argc, char **argv)
 	/* Wait workers to exit */
 	odph_thread_join(global->thread_tbl, global->options.num_cpu);
 
-	print_stat(global);
+	output_results(global);
```
MatiasElo, line 655: Return value is not checked.

test/performance/odp_queue_perf.c
```diff
@@ -564,6 +587,12 @@ int main(int argc, char **argv)
 		exit(EXIT_FAILURE);
 	}
 
+	argc = test_common_parse_options(argc, argv);
+	if (test_common_options(&common_options)) {
+		ODPH_ERR("Error while reading test options, exiting\n");
```
MatiasElo, line 592: Nitpicking but this app uses a bit different error message style.

MatiasElo (approved):

PR #2131, opened 2024-10-02
[PATCH v2] test: queue_perf: minor improvements

MatiasElo:


test/performance/odp_queue_perf.c
```diff
@@ -190,12 +190,12 @@ static int create_queues(test_global_t *global)
 		event[i] = ODP_EVENT_INVALID;
```
TuomasTaipale, line 196: There still seems to be quite a few error-like prints using `printf()`, swap those as well?

(E.g. 
```c
		printf("Too many queues %u. Test maximum %u.\n",
		       test_options->num_queue, MAX_QUEUES);

...
			printf("Max queues supported %u\n",
			       queue_capa.plain.max_num);
```
)

test/performance/odp_queue_perf.c
```diff
@@ -608,25 +608,25 @@ int main(int argc, char **argv)
...
 	if (shm == ODP_SHM_INVALID) {
-		ODPH_ERR("Error: Shared mem reserve failed.\n");
+		ODPH_ERR("Shared mem reserve failed.\n");
 		exit(EXIT_FAILURE);
 	}
 
 	global = odp_shm_addr(shm);
 	if (global == NULL) {
-		ODPH_ERR("Error: Shared mem alloc failed\n");
+		ODPH_ERR("Shared mem alloc failed\n");
```
TuomasTaipale, line 636: Old code, but the "alloc" here for `odp_shm_addr()` is a bit confusing.

test/performance/odp_queue_perf.c
```diff
@@ -153,6 +153,11 @@ static int parse_options(int argc, char *argv[], test_options_t *test_options)
 		return -1;
 	}
 
+	if (test_options->single && test_options->num_cpu != 1) {
+		printf("Using single producer/consumer queue(s) with multiple workers not supported.\n");
```
TuomasTaipale, line 157: Error-like, so `ODPH_ERR()`?

test/performance/odp_queue_perf.c
MatiasElo: Change in V2.

test/performance/odp_queue_perf.c
MatiasElo: Message clarified in V2.

test/performance/odp_queue_perf.c
MatiasElo: Changed in V2.

PR #2132, opened 2024-10-08
[PATCH v5] test: export: terminate writing properly

Rayska:
call test_common_write_term() also when exiting and writing is successful.


MatiasElo (approved):
Capital letter for the first word of commit message (`Call`). Otherwise looks good.

PR #2133, opened 2024-10-09
[PATCH v5] validation: cls: add PMR chain test where the chain is created in reverse order

JereLeppanen:


test/validation/api/classification/odp_classification_tests.c
```diff
@@ -378,6 +377,169 @@ void test_cls_pmr_chain(odp_bool_t enable_pktv)
...
+	pmr_param.match.value = &addr;
+	pmr_param.match.mask = &mask;
+	pmr_param.val_sz = sizeof(addr);
+	pmr_list[CLS_PMR_CHAIN_REV_SRC] =
+	odp_cls_pmr_create(&pmr_param, 1, cos_list[CLS_DEFAULT],
+			   cos_list[CLS_PMR_CHAIN_REV_SRC]);
+	CU_ASSERT_FATAL(pmr_list[CLS_PMR_CHAIN_REV_SRC] != ODP_PMR_INVALID);
+}
+
+void test_cls_pmr_chain_rev(odp_bool_t enable_pktv)
```
JannePeltonen, line 488: This function is exactly the same as test_cls_pmr_chain(), apart from differences in some constants. They could be combined to avoid code duplication.

test/validation/api/classification/odp_classification_tests.c
```diff
@@ -378,6 +377,169 @@ void test_cls_pmr_chain(odp_bool_t enable_pktv)
 	odp_packet_free(pkt);
 }
 
+void configure_cls_pmr_chain_rev(odp_bool_t enable_pktv)
```
JannePeltonen, line 380: This function is almost the same as configure_cls_pmr_chain_rev(). They could be combined and/or refactored to avoid the code duplication and to make the real difference between them obvious and not lost in the clutter.

JereLeppanen: Done in v2, although it's not unambiguously better.

test/validation/api/classification/odp_classification_tests.c
```diff
@@ -214,24 +213,60 @@ int classification_suite_pktv_term(void)
 	return classification_suite_common_term(true);
 }
 
-void configure_cls_pmr_chain(odp_bool_t enable_pktv)
+static void configure_cls_pmr_chain_create_saddr_pmr(int src, const char *saddr)
```
JannePeltonen, line 216: I find the 'src' name of the first parameter a bit misleading as it is really the index of the destination CoS. Perhaps you should rename it to 'dst'. Maybe it would be even more clear (?) if the src CoS was a parameter too and the call sites would just pass CLS_DEFAULT in it.

Otherwise this PR looks good.

JereLeppanen: In v3.

JannePeltonen (approved):

JereLeppanen:
v4:
- Rebase.
- Add review tags.

PR #2137, opened 2024-10-24
[PATCH v3] validation: cls: use correct L4 proto for the PMR used

JannePeltonen:
The test code goes through the trouble of checking classification capabilities when deciding whether to use UDP or TCP port matching rules in tests but then always uses UDP test packets anyway. Tests are thus broken is only TCP port matching is supported despite all the capability checking.

Fix the problem by selecting also the protocol used in test packets based on capabilities.

### DIFF
--- a/test/validation/api/classification/odp_classification_common.c
+++ b/test/validation/api/classification/odp_classification_common.c
@@ -1,6 +1,6 @@
 /* SPDX-License-Identifier: BSD-3-Clause
  * Copyright (c) 2015-2018 Linaro Limited
- * Copyright (c) 2020 Nokia
+ * Copyright (c) 2020-2024 Nokia
  */
 
 #include "odp_classification_testsuites.h"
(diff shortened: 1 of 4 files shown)


test/validation/api/classification/odp_classification_common.c
```diff
@@ -1,6 +1,6 @@
 /* SPDX-License-Identifier: BSD-3-Clause
```
JereLeppanen, line 1: Typo in commit comment: "Tests are thus broken _is_ only".

test/validation/api/classification/odp_classification_common.c
```diff
@@ -558,11 +558,28 @@ odp_cls_pmr_term_t find_first_supported_l3_pmr(void)
...
+		CU_FAIL("Implementation doesn't support any TCP/UDP PMR");
 
 	return term;
 }
 
+cls_packet_l4_info find_first_supported_proto(void)
+{
+	odp_cls_capability_t capability;
+
+	odp_cls_capability(&capability);
```
JereLeppanen, line 570: If this fails, the conditions below are testing potentially unitialized data.

test/validation/api/classification/odp_classification_common.c
JannePeltonen: v2

test/validation/api/classification/odp_classification_common.c
JannePeltonen: Fixed in v2. Fixed also similar problems in the existing code.

JereLeppanen (approved):

PR #2140, opened 2024-11-04
[PATCH v4] validation: pktio: separate and reduce packet and rx counts

JannePeltonen:
Replace the magic number 1000 used for test packet counts and rx attempt counts with two macros. Lower the test packet count to 100 since the used pool configuration did not allow allocating more than 128 packets anyway. Make the rx attempt count twice the test packet count to leave room for a few possible non-test packets and to allow almost 100 ms for the first sent test packet to come back.

The previous code had a higher rx attempt count than the test packet count but that was apparently accidental due to the pool config limiting the test packet count. The 1000 used for the rx attempt count was higher than necessary and slowed down the test slightly.

### DIFF
--- a/test/validation/api/pktio/pktio.c
+++ b/test/validation/api/pktio/pktio.c
@@ -20,6 +20,9 @@
 #define PKT_LEN_MAX            (PKT_BUF_SIZE - ODPH_ETHHDR_LEN - \
 				ODPH_IPV4HDR_LEN - ODPH_UDPHDR_LEN)
 
+#define NUM_TEST_PKTS          100
+#define NUM_RX_ATTEMPTS        200
+
 #define USE_MTU                0
 #define MAX_NUM_IFACES         2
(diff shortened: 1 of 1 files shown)


test/validation/api/pktio/pktio.c
```diff
@@ -3089,13 +3092,15 @@ static int pktio_check_start_stop(void)
 	return ODP_TEST_ACTIVE;
 }
 
+#include <unistd.h>
```
JereLeppanen, line 3095: Why?

JannePeltonen: Oops, that's a leftover of some test code that I had in there when making this change. I'll remove it in the next version.

JannePeltonen:
v2: removed the unistd.h inclusion

JereLeppanen (approved):

JannePeltonen:
v3: added review tag

PR #2141, opened 2024-11-05
[PATCH v7] api: pktio: lso: improve the specification of IPv4 and TCP LSO

JannePeltonen:
Change the DF flag handling in TCP segmentation offload and improve the description in general:

- Specify that the DF flag of the IP header is to be preserved as it was in the original packet and not cleared as the current spec says.

- Specify that TCP segmentation should be done in a way that complies with the TCP standard, without going into details. This entails proper handling of various TCP header flags and TCP options.

- Specify that unique ID values are generated for non-atomic datagrams and leave it unspecified how the ID value is updated for atomic datagrams (packets which are not fragments and have the DF flag set). Suggest that an application should avoid sending packets with ID values that may lead to overlapping IDs after segmentation.

Improve the description of IPv4 fragmentation offload. This does not change the meaning of the specification but makes it clear that the DF flag is not affected by LSO and that IP option fields are processed as specified in the standard.

### DIFF
--- a/include/odp/api/spec/packet_io_types.h
+++ b/include/odp/api/spec/packet_io_types.h
@@ -797,22 +797,99 @@ typedef enum odp_lso_protocol_t {
 
 	/** LSO performs IPv4 fragmentation.
 	 *
-	 *  IP header length and checksum fields are updated. IP fragmentation related fields are
-	 *  filled and IPv4 Identification field value is copied from the original packet. */
+	 *  The total length, MF flag, fragment offset and checksum fields
+	 *  of the IP header are updated according to the IPv4 standard and
+	 *  other fields are copied from the original packet. All IP options
+	 *  in the original packets are copied to the fragments.
+	 *
+	 *  IPv4 LSO must not be attempted for the following packets:
+	 *  - packets that are not IPv4 packets
+	 *  - packets that are already IPv4 fragments
+	 *  - packets that have IP options that may not be copied to all
+	 *    fragments
+	 *  - packets for which L4 checksum offload is requested
+	 *
+	 *  L3 offset must point to the start of the IP header.
+	 *  Payload offset must point to the start of the IP payload.
+	 */
 	ODP_LSO_PROTO_IPV4,
 
-	/** LSO performs IPv6 fragmentation. */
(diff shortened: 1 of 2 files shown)


JannePeltonen:
v2: Revamped the text quite a bit, adding at the same time multiple restrictions regarding what kind of packets LSO can be requested for. Added also descriptions for IPv6 LSO and TCP_IPV6 LSO.

JannePeltonen:
v3: Added a validation test commit that removes IPv4 LSO testing with fragments.

include/odp/api/spec/packet_io_types.h
```diff
@@ -794,22 +794,99 @@ typedef enum odp_lso_protocol_t {
...
+	 *  packet, with the sequence number and checksum updated. All TCP
+	 *  options are copied from the original packet. The FIN flag is copied
+	 *  to the last created segment and cleared in other segments. The CWR
+	 *  flag is copied to the first created segment and cleared in other
+	 *  segments.
+	 *
+	 *  IP headers (including IP options) of the created segments are
+	 *  copied from the original packet, with total length, checksum
+	 *  and identification fields updated. Unique IPv4 identification
+	 *  field values are generated for non-atomic IPv4 datagrams. The
```
psavol, line 852: Since in our case original packet cannot be fragment, this line could refer to DF bit as well: "... non-atomic IPv4 datagrams (DF flag not set)."

psavol (approved):
Reviewed API commit. One comment, otherwise OK.

include/odp/api/spec/packet_io_types.h
JannePeltonen: v4: removed the "atomic datagram" term and simply referred to the DF bit.

MatiasElo (approved):
Validation test changes OK. Copyright could be updated in the API commit.

JannePeltonen:
v5: rebased, added review tag, did not update copyright as that is done already in PR 2117.

nithind1988:
Reviewed-By: Nithin Dabilpuram <ndabilpuram@marvell.com>

JannePeltonen:
v6: review tag

PR #2143, opened 2024-11-08
helper: clarify that ODP must be initialized before calling helper functions

JereLeppanen:


TuomasTaipale (commented):
I guess technically we could also allow `odph_ipsec_auth_icv_len_default()`.

JereLeppanen:
> I guess technically we could also allow `odph_ipsec_auth_icv_len_default()`.

Yes, and also all functions in stress.h. But I've allowed here only the functions that we are currently calling before initialization. Another option would be to only allow odph_parse_options(), odph_options() and maybe odph_strcpy(), and fix the rest so that they would be called only after initialization.

TuomasTaipale:
Only allowing the two option functions, the copy function (and the log macros so logging errors in option error handling would not become cumbersome?) sounds good to me. Have to discuss this with others.

doc/helper-guide/odp.dox
```diff
@@ -16,6 +16,10 @@
  *
  *  @image html overview.svg
  *
+ * ODP must be initialized before calling helper functions, unless stated
+ * otherwise. Pre-processor macros may be invoked independently of ODP
+ * initialization.
```
MatiasElo, line 21: As discussed, this same clause could also be added to some helper API header (for example `helper/include/odp/helper/odph_api.h`).

JereLeppanen: When we discussed this, I liked the idea of having this in a header. But putting this in the `@file` documentation of odph_api.h would put in https://opendataplane.github.io/odp/helper-guide/odph__api_8h.html#details, which is pretty well hidden. So it's also needed in odp.dox, so that it appears on the main page https://opendataplane.github.io/odp/helper-guide/. So we would have two copies of the paragraph in two different places, which is not ideal, but maybe it's fine?

MatiasElo: Yep, it's not ideal but better to have the documentation in both places.

JereLeppanen: In v3.

MatiasElo (approved):

TuomasTaipale (approved):

JereLeppanen:
v4:
- Rebase.
- Add review tags.

psavol (approved):

JereLeppanen:
v5:
- Rebase.
- Add review tag (Petri).

PR #2146, opened 2024-11-13
[PATCH v3] test: sched_perf: print extra rd/rw operations per event

MatiasElo:


test/performance/odp_sched_perf.c
```diff
@@ -507,6 +509,13 @@ static void print_options(test_options_t *options)
 	printf("  queue type                %s\n\n", options->queue_type == 0 ? "parallel" :
 						     options->queue_type == 1 ? "atomic" :
 						     "ordered");
+
+	printf("Extra rd/rw ops per event (queue ctx + uarea + data)\n");
```
TuomasTaipale, line 513: Perhaps the same terms that are used elsewhere in output could be used here: "queue ctx" -> "queue context", "uarea" -> "user area", "data" -> "event data".

MatiasElo: Changed on V2.

TuomasTaipale (approved):

PR #2148, opened 2024-11-15
test: queue_perf: fix pair mode issues

MatiasElo:


TuomasTaipale (approved):
Two commit message improvement suggestions, otherwise looks good:
- first patch: maybe a comma between "In pair mode" and "events"?
- second patch: swap "the" and "all" in "... until the all workers ..."?

MatiasElo:
V2:
- Fixed commit messages and added reviewed-by tags

PR #2149, opened 2024-11-18
[PATCH v1] api: event: add generic event vector API

MatiasElo:
Add new generic event vector API for generating event vectors from all event types, as opposed to the existing packet vector API, which only supports packets.

### Open questions
- [ ] Support multiple aggregators per queue or keep the API simpler and support only single aggregator per queue? 
- [ ] Is capability for total number of  aggregators required?


include/odp/api/spec/event_vector_types.h
```diff
@@ -0,0 +1,153 @@
...
+	 * application is allowed to enqueue any event types, except event
+	 * vectors, to the vector aggregator. Otherwise, only events of type
+	 * 'event_type' are allowed. The default value is ODP_EVENT_ANY.
+	 *
+	 * Regardless of 'event_type', an application is never allowed to
+	 * enqueue event vector or packet vector events (ODP_EVENT_VECTOR or
+	 * ODP_EVENT_PACKET_VECTOR) to a queue with vector aggregation enabled
+	 * (i.e. vectors within vectors).
+	 */
+	odp_event_type_t event_type;
```
PavanNikhilesh, line 103: Would this configuratiuon expect fastpath checks? Or is it just a warning for application?

include/odp/api/spec/event_types.h
```diff
@@ -52,6 +55,8 @@ extern "C" {
  *     - Timeout event (odp_timeout_t) from a timer
  * - ODP_EVENT_IPSEC_STATUS
  *     - IPSEC status update event (odp_ipsec_status_t)
+ * - ODP_EVENT_VECTOR
```
PavanNikhilesh, line 58: Would the scheduled event always be ODP_EVENT_VECTOR or if odp_event_vector_config_t.event_type == ODP_EVENT_PACKET is it expeceted to be  ODP_EVENT_PACKET_VECTOR

PavanNikhilesh:
Overall, the specification looks good to me, with a few minor clarifications.

> Support multiple aggregators per queue or keep the API simpler and support only single aggregator per queue?
Either is fine with me, but supporting multiple aggregators per queue gives the application more options.
>  Is capability for total number of aggregators required
A max_vector_aggregators_per_queue capability should be good.

include/odp/api/spec/event_vector_types.h
MatiasElo: It's undefined behavior if application operates agains the specification, so no checks are required in the fast path. Nevertheless, it may be good to add checks behind debug flags etc. to ease debugging.

include/odp/api/spec/event_types.h
MatiasElo: In this case the event type would always be `ODP_EVENT_VECTOR`. I'd like to keep the event and packet vector APIs separate. Optimally, the whole old packet vector API would be deprecated, but we cannot do that without breaking backwards compatibility.

include/odp/api/spec/event_vector_types.h
PavanNikhilesh: ack

include/odp/api/spec/event_types.h
PavanNikhilesh: ack

include/odp/api/spec/queue_types.h
```diff
@@ -291,6 +301,29 @@ typedef struct odp_queue_param_t {
...
+	  *
+	  * When event vector generation is enabled, events enqueued to this
+	  * queue are by default vectorized using vector aggregator 0. When
+	  * 'num_vector' >1, other vector aggregators can be utilized with
+	  * odp_queue_enq_vector() function.
+	  *
+	  * When vector generation is enabled, events may be delivered both as
+	  * event vector events and normal events. Implementation won't split up
+	  * event vectors successfully enqueued by the application. */
+	odp_event_vector_config_t vector[ODP_QUEUE_MAX_VECTOR_AGGR];
```
psavol, line 325: Other option could be to create vector generator profile, and give handle to that here instead of full set of config params ? That would enable easy reuse of the same set of params for many queues (that have the same aggregation needs).

JannePeltonen, line 325: Having multiple vector aggregators per queue adds flexibility and might be useful for decoupling vector aggregation from scheduling contexts (i.e. one could aggregate pktio and crypto packets separately but still have them go through the same scheduled queue, e.g. for atomicity).

However, configuring the aggregators this way in a fixed manner at queue creation time is rather inflexible. It would not be possible to e.g. have a new aggregator for a newly created crypto session attached to an existing queue. If we want to support multiple aggregators per queue, I think it would be nice to be able to create the aggregators dynamically or at least associate them to queues dynamically.

JannePeltonen, line 325: Having multiple aggregators per queue raises also the question on how to select with aggregator is to be used for a packet coming for pktio/classifier/crypto/ipsec etc. This patch set adds a new SW sending function with aggregator index. something similar would be needed for all the other cases. This feels unclean to me.

Maybe we could think that aggregators are just another level of queues (which they really are) and treat them as queues in the API so that for each aggregator we would have a similar queue handle as for other queues. This queue handle could then be used in all existing API functions to specify the aggregator to be used. And if aggregators were created dynamically (see my other comment), then the aggregator creation mechanism could quite naturally return such a queue handle.

If we do not wish to support multiple aggregators per queue just yet but want to leave the door open for it, then maybe we could have a vector aggregator config as part of queue config (and the aggregator would affect everything going through that queue) and then later optionally introduce a way to create additional "auxiliary" queues for the main queue for alternate aggregators. Or something like that.

MatiasElo:
Replaced by https://github.com/OpenDataPlane/odp/pull/2187.

PR #2153, opened 2024-11-22
[PATCH v3] helper: deprecate linux specific thread/process create and join functions

MatiasElo:


helper/include/odp/helper/linux/pthread.h
```diff
@@ -39,10 +40,12 @@ extern "C" {
  * @param      thr_params  Linux helper thread parameters
  *
  * @return Number of threads created
+ *
+ * @deprecated Use odph_thread_create() instead.
  */
-int odph_linux_pthread_create(odph_linux_pthread_t *pthread_tbl,
-			      const odp_cpumask_t *mask,
-			      const odph_linux_thr_params_t *thr_params);
+int ODPH_DEPRECATE(odph_linux_pthread_create)(odph_linux_pthread_t *pthread_tbl,
```
JannePeltonen, line 46: Using ODPH_DEPRECATE() here kinda works, but I think it is not very good because:
- It does not really check that --enable-helper-deprecated works properly for this symbol (i.e. makes it visible plainly as odph_linux_pthread_create().
- It gives a bad example that it is ok to use ODPH_DEPRECATE macro in ODP application code.
- The tests are run even when the helper is compiled without --enable-deprecated.

I think it would be better to not put ODPH_DEPRECATE here but just modify the relevant Makefile.am to include this test only with --enabdle deprecated.

And the same comment to the other occurences of this in this PR.

MatiasElo: Changed in V2.

JannePeltonen: I misplaced the above comment that was meant for the validation test change, but you got my point anyway.

psavol (approved):

PR #2154, opened 2024-11-22
[PATCH v3] doc: userguide: update code blocks

MatiasElo:
Remove code sections copied from the API headers as they will not stay up- to-date. Update some code examples to utilize current APIs and remove some obsolete ones.

### DIFF
--- a/doc/users-guide/users-guide-pktio.adoc
+++ b/doc/users-guide/users-guide-pktio.adoc
@@ -27,95 +27,20 @@ or *Closed* via the `odp_pktio_close()` API to return the PktIO to the
 *Unallocated* state.
 
 === PktIO Allocation
-PktIO objects begin life by being _opened_ via the call:
+PktIO objects begin life by being _opened_ with `odp_pktio_open()` API:
 [source,c]
 -----
-/**
- * Open a packet IO interface
- *
- * An ODP program can open a single packet IO interface per device, attempts
- * to open an already open device will fail, returning ODP_PKTIO_INVALID with
- * errno set. Use odp_pktio_lookup() to obtain a handle to an already open
- * device. Packet IO parameters provide interface level configuration options.
- *
- * Use odp_pktio_param_init() to initialize packet IO parameters into their
- * default values. Default values are also used when 'param' pointer is NULL.
- *
- * Packet input queue configuration must be setup with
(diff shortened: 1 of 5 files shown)


doc/users-guide/users-guide-pktio.adoc
```diff
@@ -136,330 +61,26 @@ struct, in turn, specifies the input and output *modes* of the PktIO.
...
-	  *
-	  * The default value is ODP_PROTO_LAYER_ALL. */
-	odp_proto_layer_t layer;
-
-} odp_pktio_parser_config_t;
------
-Note that parsing is automatically done whenever classification is enabled
-for an RX interface (see below).
+receipt as controlled by the `odp_pktio_parser_config_t` struct. Note that
+parsing is automatically done whenever classification is enabled for an RX
```
JannePeltonen, line 76: > parsing is automatically done whenever classification is enabled

This statement caught my eye even though it is not changed by this commit. I wonder if enabling of parsing is really guaranteed by the API in this case as I did not find it in the API text. Or perhaps this is just the way the linux-gen ODP implementation works?

doc/users-guide/users-guide-pktio.adoc
```diff
@@ -500,97 +121,7 @@ of type *odp_pktin_queue_t* and are retrieved by threads calling the
...
-	  * odp_queue_param_t documentation.
-	  * When classifier is enabled in odp_pktin_queue_config() this
-	  * value is ignored. */
-	odp_queue_param_t queue_param;
-
-} odp_pktin_queue_param_t;
------
-Note that the *queue_param* field of this struct is ignored in DIRECT mode.
-The purpose of `odp_pktin_queue_config()` is to specify the number of PktIn
+`odp_pktin_queue_config()` API, which purpose is to specify the number of PktIn
```
JannePeltonen, line 124: The 'which' does not sound quite right to me. I would probably use 'whose', or if that is not good, I would say "... API, the purpose of which is ...".

JannePeltonen (approved):
I made two comments, one not really related to the changes donen here and the other so minor that feel free to ignore it.

doc/users-guide/users-guide-pktio.adoc
MatiasElo: Indeed the API doesn't mention this, so removed the sentence.

PR #2157, opened 2024-11-26
[PATCH v7] api: pktio: lso: promise maximum payload size for custom proto

JannePeltonen:
Specify that LSO for the custom protocol produces same-sized segments (except the last segment).

Add a profile parameter to request exactly maximum length payload sizes in the non-last segments produced by the custom protocol LSO.

v2:
Remove configurability and the capability flag and just promise that the custom LSO protocol always produces maximum length segments.

v3:
Added validation test commit.


MatiasElo (approved):

psavol (approved):
Review and approved API commit.

JannePeltonen:
v5: added review tags

nithind1988:

JannePeltonen:
v6: review tag

PR #2160, opened 2024-12-09
[PATCH v2] linux-gen: timer: fix expiration enqueueing error handling

TuomasTaipale:
Currently, when handling timer expiration, error while enqueueing timeout event leads to an abort. Before the abort, the timeout event is attempted to be freed. With POSIX timer implementation, where timer expiration is handled in a separate helper thread that has not been locally ODP-initialized, this leads to a NULL pointer access as thread-local pool cache is uninitialized in the helper thread. As the thread (and related worker/process) is about to be aborted anyways, remove the event freeing attempt to fix this.

v2:
- Added reviewed-by tag


MatiasElo (approved):
The current POSIX timer implementation is indeed quite suspicious but at least this change won't make the situation worse.

TuomasTaipale:
Perhaps not suspicious, just neglected during new feature development over the years. V2 pushed with review tag.

PR #2161, opened 2024-12-09
[PATCH v2] linux-gen: check odp_cpu_id() call return values

MatiasElo:
Check odp_cpu_id() call return values to prevent accidentally using negative return values as array indices. Fixes Coverity warnings.

### DIFF
--- a/platform/linux-generic/odp_schedule_scalable.c
+++ b/platform/linux-generic/odp_schedule_scalable.c
@@ -927,6 +927,9 @@ dequeue_atomic:
 	}
 
 	cpu_id = odp_cpu_id();
+	if (odp_unlikely(cpu_id < 0))
+		return 0;
+
 	/* Scan our schedq list from beginning to end */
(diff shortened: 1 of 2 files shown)


platform/linux-generic/odp_schedule_scalable.c
```diff
@@ -927,6 +927,9 @@ static int _schedule(odp_queue_t *from, odp_event_t ev[], int num_evts)
 	}
 
 	cpu_id = odp_cpu_id();
+	if (odp_unlikely(cpu_id < 0))
```
JannePeltonen, line 930: The cpu ID here should perhaps be thread ID, but that is orthogonal to this change, which looks ok.

JannePeltonen (approved):

platform/linux-generic/odp_schedule_scalable.c
MatiasElo: Yeah, I was thinking the same, but didn't want to touch the scalable scheduler code any more than necessary.

PR #2173, opened 2025-01-16
[PATCH v3] remove unused longindex

TuomasTaipale:
When parsing command line arguments with `getopt_long()`, longindex output pointer can be `NULL` if it is not needed.

v2:
- Rebased
- Added reviewed-by tag


TuomasTaipale:
About squashing, these are standalone and mostly targeting implementation-independent programs, so might as well be separate commits.

JannePeltonen:
I would prefer squashing since they are all the same change and I do not see much (any?) benefit in not squashing.

PR #2174, opened 2025-01-17
[PATCH v4] linux-gen: shm: make internal ishm block names unique

MatiasElo:


platform/linux-generic/odp_ishm.c
```diff
@@ -909,22 +910,36 @@ static int do_unmap(void *start, uint64_t size, uint32_t flags,
...
 			return i;
 	}
 
 	return -1;
 }
 
+/*
+ * Make application given memory block name unique
+ */
+static void name_to_ishm_name(const char *name, char *block_name)
```
JannePeltonen, line 935: To me the name of this function does not match what it is doing and it got me confused. Maybe it could be renamed to something like make_block_name_unique and/or change the semantics a bit (e.g. call this make_unique_name and have it always generate the name, getting rid of the strcpy at the call site).

platform/linux-generic/odp_ishm.c
```diff
@@ -909,22 +910,36 @@ static int do_unmap(void *start, uint64_t size, uint32_t flags,
...
+{
+	int i = 0;
+
+	while (find_block_by_name(block_name, false) > 0)
```
JannePeltonen, line 939: This looks racy if multiple threads happen to be reserving shm blocks at the same time? Maybe you need a mutex or maybe you should generate a unique name based on some global counter (which would need to be atomic and in shared memory)?

JannePeltonen: There seems to be a lock there already (ishm_tbl->lock), so ignore this comment.

platform/linux-generic/odp_ishm.c
MatiasElo: Refactored in V2. Also, added missing `=` to while loop condition.

platform/linux-generic/odp_ishm.c
```diff
@@ -2045,7 +2065,7 @@ int _odp_ishm_status(const char *title)
 		_ODP_PRINT("%2i  %-*s %s%c  %p-%p %-8" PRIu64 "   "
 			  "%-8" PRIu64 " %-3" PRIu64 " %-3" PRIu64 " "
 			  "%-3d %s\n",
-			  i, max_name_len, ishm_tbl->block[i].name,
+			  i, max_name_len, ishm_tbl->block[i].param_name,
```
JannePeltonen, line 2068: Since .name was changed to .param_name. the corresponding field width parameter max_name_len should also be calculated from param_name instead of name.

platform/linux-generic/odp_ishm.c
```diff
@@ -2142,7 +2162,7 @@ void _odp_ishm_print(int block_index)
 	block = &ishm_tbl->block[block_index];
 
 	_ODP_PRINT("\nSHM block info\n--------------\n");
-	_ODP_PRINT(" name:       %s\n",   block->name);
+	_ODP_PRINT(" name:       %s\n",   block->param_name);
```
JannePeltonen, line 2165: Maybe the internal unique name could still be printed too since this function appears to be meant for debugging?

platform/linux-generic/odp_ishm.c
```diff
@@ -1094,12 +1111,15 @@ int _odp_ishm_reserve(const char *name, uint64_t size, int fd,
 
 	new_block = &ishm_tbl->block[new_index];
 
-	/* save block name (if any given): */
-	if (name)
-		_odp_strcpy(new_block->name, name, ISHM_NAME_MAXLEN);
-	else
+	/* Save application given block name as-is and make internal block name used for mmap
+	   unique */
```
JannePeltonen, line 1115: The call to alloc_single_va() later on in this function still uses the application given non-unique name. I did not read carefully but it looks like it would be a problem for exported memory blocks as it would cause overlapping export file names.

platform/linux-generic/odp_ishm.c
```diff
@@ -1094,12 +1111,15 @@ int _odp_ishm_reserve(const char *name, uint64_t size, int fd,
...
+	if (name) {
```
JannePeltonen, line 1116: Maybe it would be simpler to always make the internal name unique, even if no name was given by the application? Now the situation is a bit weird that the internal name is unique except when it is the empty string in which case it is not unique and then the lower level code has to special case that and create a unique string using create_seq_string(). Making the name always unique would make the code more clear, I think.

platform/linux-generic/odp_ishm.c
```diff
@@ -909,22 +910,38 @@ static int do_unmap(void *start, uint64_t size, uint32_t flags,
...
+ * Make application given memory block name unique
+ */
+static void make_name_unique(char *block_name, const char *name, size_t sz)
+{
+	int i = 0;
+
+	_odp_strcpy(block_name, name, sz);
+
+	while (find_block_by_name(block_name, false) >= 0)
+		_odp_snprint(block_name, sz, "%d-%s", i++, name);
```
JannePeltonen, line 942: Now that I noticed that the create_seq_string() and ishm_tbl->dev_seq are already there,  maybe (?) this could just generate the unique name using dev_seq or similar and avoid the search?

platform/linux-generic/odp_ishm.c
MatiasElo: Fixed in V3.

platform/linux-generic/odp_ishm.c
MatiasElo: Logic changed in V3 to always use unique internal names.

platform/linux-generic/odp_ishm.c
MatiasElo: Changed in V3.

platform/linux-generic/odp_ishm.c
MatiasElo: Was not a problem but logic changed in V3.

JannePeltonen (approved):
Export behaves a bit weirdly if the application given SHM name is empty or non-unique, but that seem to have been there already before this commit and I am not sure how export with non-unique name could even be used.

PR #2176, opened 2025-01-23
[PATCH v4] api: dma: support source segment free and destination segment allocation features

VamsiKrishnaA99:

PR #2179, opened 2025-01-27
[PATCH v5] api: dma: support source free and destination allocation features

VamsiKrishnaA99:


include/odp/api/spec/dma_types.h
```diff
@@ -397,6 +411,35 @@ typedef struct odp_dma_seg_t {
...
+			 *  allocates all required destination packets and this index field can be
+			 *  used to define which packet of the to-be-allocated array of packets is
+			 *  used as the destination for this particular destination segment
+			 *  descriptor. Same index can be reused across subsequent segments if
+			 *  application wishes to transfer to different offsets of the same
+			 *  destination packet. This means that if application for example uses
+			 *  segment index 0 in each destination segment descriptor (with different
+			 *  'offset' values), implementation would allocate only a single packet
+			 *  with data being transferred to different offsets of that single
+			 *  destination packet. Indexes should appear in ascending order.
```
TuomasTaipale, line 428: To avoid any confusion, perhaps the last sentence could be amended like this:
```
Indexes should appear in ascending order, starting from index 0.
```

include/odp/api/spec/dma_types.h
```diff
@@ -492,6 +535,77 @@ typedef struct odp_dma_transfer_param_t {
...
+	odp_pool_t dst_seg_pool;
+
+	/** Transfer hints
+	 *
+	 *  Depending on the implementation, adjusting hints bit fields may improve performance.
+	 *  Initialize all unused bits to zero.
+	 */
+	union {
+		/** Hint bit fields */
+		struct {
```
TuomasTaipale, line 552: These aren't actually hints anymore, they are more like options that applications should properly define per-transfer for optimal performance, so perhaps a few adjustments:
- "Transfers hints" -> "Transfer options"
- The sentence "Depending on the implementation, adjusting hints bit fields may improve performance." could be changed to "Options to further configure the transfer." or something similar
- "Hint bit fields" -> "Option bit fields"
- Change "hint/hints" to "option/options" in the definitions below, e.g. "Hint bit to denote a source segment ..." -> "Option bit to denote a source segment ..." etc.
- The `struct` could be named to "opts", so something like this:
```C
...
union {
    ...
    struct {
        ...
    } opts;
    ...
    uint16_t all_opts;
};
```

include/odp/api/spec/dma_types.h
```diff
@@ -492,6 +535,77 @@ typedef struct odp_dma_transfer_param_t {
...
+			/** Allow freeing all the source segments
+			 *
+			 *  When set to 1, all source segments with data format
+			 *  #ODP_DMA_FORMAT_PACKET are freed.
+			 *
+			 *  Packet free options (odp_packet_free_ctrl_t) does not affect
+			 *  this bit.
```
TuomasTaipale, line 559: To avoid any confusion, perhaps this could be adjusted to:
```
Segments are freed regardless of packet free options (odp_packet_free_ctrl_t).
```

TuomasTaipale (commented):
Still a few comments from internal review, we're getting there.

For better trackability, you can rebase-amend the commits with the fixes locally on your branch and then force-push to your fork-remote-branch. The changes then appear in this same pull request and GitHub can track the diff between versions (GitHub automatically updates the "[PATCH vX]" prefix when new stuff is pushed to the PR-branch).

VamsiKrishnaA99:
Sure Tuomas, thanks for reviewing it.

VamsiKrishnaA99:
Addressed V6 review comments.

TuomasTaipale:
#2171 and #2176 for reference.

include/odp/api/spec/dma_types.h
```diff
@@ -492,6 +536,77 @@ typedef struct odp_dma_transfer_param_t {
...
+
+			/** Allow allocating the destination segments for data format
+			 *  #ODP_DMA_FORMAT_PACKET
+			 *
+			 *  When set to 1, destination segments will be allocated from
+			 *  'dst_seg_pool'
+			 */
+			uint16_t seg_alloc : 1;
+
+			/** Hint bit to allocate unique destination segments for data format
```
TuomasTaipale, line 589: "Hint" -> "Option"

TuomasTaipale: You can also add my review tags to the next version:

Reviewed-by: Tuomas Taipale \<tuomas.taipale@nokia.com\>

VamsiKrishnaA99: Thanks Tuomas, pushed latest version.

include/odp/api/spec/dma_types.h
```diff
@@ -492,6 +499,50 @@ typedef struct odp_dma_transfer_param_t {
 	 */
 	odp_dma_seg_t *dst_seg;
 
+	/** Transfer options
+	 *
+	 *  Options to further configure the transfer.
+	 *  Initialize all unused bits to zero.
```
psavol, line 549: odp_dma_transfer_param_init() is used to initialize the struct. So, this sentence could just define the default values:
"The default value for all option bits is zero."

include/odp/api/spec/dma_types.h
```diff
@@ -404,6 +411,36 @@ typedef struct odp_dma_seg_t {
...
+			 *  descriptor. Same index can be reused across subsequent segments if
+			 *  application wishes to transfer to different offsets of the same
+			 *  destination packet. This means that if application for example uses
+			 *  segment index 0 in each destination segment descriptor (with different
+			 *  'offset' values), implementation would allocate only a single packet
+			 *  with data being transferred to different offsets of that single
+			 *  destination packet. Indexes should appear in ascending order, starting
+			 *  from index 0.
+			 */
+			uint16_t index;
```
psavol, line 431: For consistency (to match pkt_len), name of this field could be pkt_index. Also, "pkt_index" is less likely to collide later if another "index" field needs to be added into this struct.

include/odp/api/spec/dma_types.h
```diff
@@ -492,6 +536,77 @@ typedef struct odp_dma_transfer_param_t {
...
+	 *  Options to further configure the transfer.
+	 *  Initialize all unused bits to zero.
+	 */
+	union {
+		/** Option bit fields */
+		struct {
+			/** Allow freeing all the source segments
+			 *
+			 *  When set to 1, all source segments with data format
+			 *  #ODP_DMA_FORMAT_PACKET are freed.
```
TuomasTaipale, line 557: Just realized that something probably needs to be said about in which cases the freeing is carried out. E.g. if transfer has failed, then probably the freeing is not done by implementation but application needs to do it. 

So could be amended:
```
*  When set to 1, all source segments with data format
*  #ODP_DMA_FORMAT_PACKET are freed in successful
*  transfers.
```

VamsiKrishnaA99:
Thanks Petri for the comments, fixed them in v10.

include/odp/api/spec/dma_types.h
TuomasTaipale: I believe the naming is now fine, but additionally the type of the index could be changed from `uint16_t` to `uint32_t`, this would then be more consistent with `odp_dma_transfer_param_t::num_dst` and `odp_dma_result_t::num_dst` which are both `uint32_t`.

include/odp/api/spec/dma_types.h
```diff
@@ -404,6 +411,36 @@ typedef struct odp_dma_seg_t {
...
+			/** Packet index
+			 *
+			 *  When odp_dma_transfer_param_t::seg_alloc is used, implementation
+			 *  allocates all required destination packets and this pkt_index field can
+			 *  be used to define which packet of the to-be-allocated array of packets
+			 *  is used as the destination for this particular destination segment
+			 *  descriptor. Same pkt_index can be reused across subsequent segments if
+			 *  application wishes to transfer to different offsets of the same
+			 *  destination packet. This means that if application for example uses
+			 *  segment index 0 in each destination segment descriptor (with different
```
psavol, line 425: Typo: "segment index 0 ..." -> "packet index 0 ..."

include/odp/api/spec/dma_types.h
psavol: Thank you for fixing the field name. I agree with Tuomas that uint32_t could be used instead. It does not increase struct size and harmonizes numbers DMA segments/packets parameter types.

psavol:
Spec text in v12 is OK from my point of view.

TuomasTaipale:
@VamsiKrishnaA99 Petri's review-tags could be now added and this PR should then be ready.

PR #2180, opened 2025-01-29
test: timer_accuracy: introduce "concurrency" mode

JereLeppanen:


test/performance/odp_timer_accuracy.c
```diff
@@ -1449,6 +1642,41 @@ int main(int argc, char *argv[])
...
+			uint64_t e = events - events_prev;
+			int stopped = 0;
+
+			for (int i = 0; i < (int)test_global->opt.burst; i++) {
+				/*
+				 * Count how many timers have not generated any events since last
+				 * interval.
+				 */
+				timer_ctx_t *ctx = &test_global->timer_ctx[i];
+				uint64_t ctx_events = ctx->events;
```
MatiasElo, line 1668: Seems like a possible data race. Is this information that necessary to print anyway?

test/performance/odp_timer_accuracy.c
```diff
@@ -651,7 +701,7 @@ static int create_timers(test_global_t *test_global)
 		printf("  resolution:      %" PRIu64 " nsec\n", timer_param.res_ns);
 	}
 
-	timer_param.num_timers = alloc_timers;
+	timer_param.num_timers = alloc_timers * 2;
```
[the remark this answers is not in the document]
JereLeppanen: Yes, and it's not *2, the extra number of timers needed is actually equal to cpu_count.

test/performance/odp_timer_accuracy.c
```diff
@@ -1164,8 +1222,24 @@ static int run_test(void *arg)
...
-			tmo_ns += ctx->count * period_dbl + 0.5;
-			ctx->count++;
+			tmo_ns += ctx->events * period_dbl + 0.5;
+			ctx->events++;
+		} else if (mode == MODE_CONCURRENCY) {
+			uint64_t events = ++ctx->events;
+			uint64_t starts = ctx->starts;
+
+			if (events > starts)
+				ODPH_ERR("ctx %p timer %p time %" PRIu64 " starts %" PRIu64
```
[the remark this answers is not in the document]
JereLeppanen: In v2, added an option to exit with _exit().

test/performance/odp_timer_accuracy.c
```diff
@@ -1164,8 +1222,24 @@ static int run_test(void *arg)
...
+					 " events %" PRIu64 "\n",
+					 ctx, ctx->timer, time_ns, starts, events);
+
+			int64_t diff = (int64_t)time_ns - (int64_t)tmo_ns;
+
+			if (test_global->opt.max_diff &&
+			    diff > (int64_t)test_global->opt.max_diff) {
+				ODPH_ERR("ctx %p timer %p time %" PRIu64 " diff %" PRIi64 "\n", ctx,
```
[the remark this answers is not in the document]
JereLeppanen: _exit() also here.

test/performance/odp_timer_accuracy.c
JereLeppanen: It's definitely a data race. It's avoiding synchronization in order to have minimal effect on the test itself. The prints have been very useful in recent testing, especially when some timers stop generating events, some periodic check like this is needed to see that something stopped from happening.

On the other hand perhaps it's better to add prints ad hoc, when needed to investigate a specific problem. So this could just be removed.

JereLeppanen:
v2:
- Prcess review comments.
- Add commit `test: timer_accuracy: report errors with ODPH_ERR()`

test/performance/odp_timer_accuracy.c
MatiasElo: Looks like `timer_ctx_t.events` could be operated with relaxed atomic, so the overhead should be relatively small. However, either solution is fine for me.

JereLeppanen: Relaxed atomic would probably be fine, but nevertheless I decided to remove the "stopped" counter, so ctx->events remains non-atomic.

JereLeppanen:
v3:
- Remove "stopped" counter.

test/performance/odp_timer_accuracy.c
```diff
@@ -1263,34 +1476,60 @@ static int run_test(void *arg)
...
 
+static void interval_loop_concurrency(test_global_t *test_global)
+{
+	uint64_t events_prev = 0;
+	uint64_t events = 0;
+	uint64_t start_ns = odp_time_global_strict_ns();
+	uint64_t prev_ns = start_ns;
+
+	while ((events = odp_atomic_load_u64(&test_global->events)) < test_global->opt.tot_timers) {
+		odp_time_wait_ns(test_global->opt.interval * ODP_TIME_SEC_IN_NS);
```
MatiasElo, line 1518: Because `test_global->events` value is read before wait, the first print round results aren't correct. At least with larger interval values this can be confusing. E.g.
```
Starting timers took 776 nsec
sec 60 total events 0 events 0 events/s 0
sec 120 total events 54545 events 54545 events/s 909

```

JereLeppanen: I forgot about that. Fixed in v4.

JereLeppanen:
v4:
- Rebase.
- Fix first interval print.
- Add review tags.

MatiasElo (approved):

PR #2182, opened 2025-02-18
[PATCH v4] api: ml: add quantization parameters to input and output info

JereLeppanen:


include/odp/api/spec/ml_types.h
```diff
@@ -512,6 +512,22 @@ typedef struct odp_ml_shape_info_t {
...
+ * values can be used directly with the odp_ml_fp32_from_*() and odp_ml_fp32_to_*() functions. Both
+ * scale and zerop are 0, if the input or output is not quantized, or these parameters are not
+ * available for the model, or the implementation does not provide these parameters.
+ */
+typedef struct odp_ml_quant_param_t {
+	/** Quantization scale */
+	float scale;
+
+	/** Quantization zero point */
+	int32_t zerop;
```
psavol, line 528: For future proofness, these field names could include data type (e.g. scale_fp32, zerop_i32). If there would be later double <-> u64 quantization specified, we could add scale/zerop for those (as union or separate fields).

include/odp/api/spec/ml_types.h
```diff
@@ -512,6 +512,22 @@ typedef struct odp_ml_shape_info_t {
 
 } odp_ml_shape_info_t;
 
+/**
+ * Quantization parameters
+ *
+ * These parameters are used to convert between floating point and integer data. Scale and zerop
+ * values can be used directly with the odp_ml_fp32_from_*() and odp_ml_fp32_to_*() functions. Both
+ * scale and zerop are 0, if the input or output is not quantized, or these parameters are not
+ * available for the model, or the implementation does not provide these parameters.
```
psavol, line 521: This sentence could be simplified by removing "if the input or output is not quantized, or these parameters are not
 available for the model". Just saying that those are zero when "not supported".

include/odp/api/spec/ml_types.h
JereLeppanen: I figured that fp32 precision is much more than any use case will ever need. And i32 zerop covers the 16-bit quantization cases, while more than 16 bits for integer quantization seems like a very unlikely use case. (Perhaps zerop should actually be a float, I'm not sure.)

Also, having these names now doesn't prevent adding the type postfixes later. But I can add them now, if that's what you prefer.

include/odp/api/spec/ml_types.h
JereLeppanen: Zero is a legal and probably not an uncommon value for zerop, though. So application will anyway have to check for zero scale. So if we want to eliminate the float comparison, we should probably add a boolean to indicate whether these parameters are valid or not.

psavol: True. I think we could add some flag(s) to odp_ml_model_info_t level to tell if quantization info is supported (or not), then the sentence can be also removed here.

include/odp/api/spec/ml_types.h
JereLeppanen: Added the type postfixes in v4.

include/odp/api/spec/ml_types.h
JereLeppanen: Added in v4, and the sentence replaced accordingly.

TuomasTaipale:
Closing this, work continues in #2198.

PR #2183, opened 2025-02-25
[PATCH v1] validation: cls: add capability check for hash result test

satheeshpaul:
Added capability check for `odp_cls_hash_result()` validation test.

### DIFF
--- a/test/validation/api/classification/odp_classification_basic.c
+++ b/test/validation/api/classification/odp_classification_basic.c
@@ -707,7 +707,6 @@ static void cls_create_cos_with_hash_queues(void)
 
 	ret = odp_cls_capability(&capa);
 	CU_ASSERT_FATAL(ret == 0);
-	CU_ASSERT_FATAL(capa.hash_protocols.all_bits != 0);
 
 	odp_queue_param_init(&q_param);
 	q_param.type = ODP_QUEUE_TYPE_SCHED;
@@ -731,6 +730,9 @@ static int check_capa_cos_hashing(void)
 	odp_cls_capability_t capa;
 
 	if (odp_cls_capability(&capa) != 0)
+		return ODP_TEST_INACTIVE;
+
+	if (capa.hash_protocols.all_bits == 0)
 		return ODP_TEST_INACTIVE;
 
 	return capa.max_hash_queues > 1 ? ODP_TEST_ACTIVE : ODP_TEST_INACTIVE;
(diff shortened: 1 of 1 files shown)


test/validation/api/classification/odp_classification_basic.c
```diff
@@ -707,7 +707,6 @@ static void cls_create_cos_with_hash_queues(void)
 
 	ret = odp_cls_capability(&capa);
 	CU_ASSERT_FATAL(ret == 0);
-	CU_ASSERT_FATAL(capa.hash_protocols.all_bits != 0);
```
JannePeltonen, line 710: The commit message lacks any mention of this change. And this change looks wrong anyway, since this test function is run only when multiple hash queues are supported and that in turn implies that there must be at least one protocol field included in the hash.

test/validation/api/classification/odp_classification_basic.c
```diff
@@ -733,6 +732,9 @@ static int check_capa_cos_hashing(void)
 	if (odp_cls_capability(&capa) != 0)
 		return ODP_TEST_INACTIVE;
 
+	if (capa.hash_protocols.all_bits == 0)
+		return ODP_TEST_INACTIVE;
```
JannePeltonen, line 736: Checking the value of max_hash_queues is sufficient for testing if hash is supported. There should not be a case where max_hash_queues is greater than one but hash_protocols.all_bits is zero.

test/validation/api/classification/odp_classification_basic.c
```diff
@@ -835,7 +837,7 @@ odp_testinfo_t classification_suite_basic[] = {
 	ODP_TEST_INFO(cls_cos_set_pool),
 	ODP_TEST_INFO(cls_pmr_composite_create),
 	ODP_TEST_INFO_CONDITIONAL(cls_create_cos_with_hash_queues, check_capa_cos_hashing),
-	ODP_TEST_INFO(cls_hash_result_single_queue),
+	ODP_TEST_INFO_CONDITIONAL(cls_hash_result_single_queue, check_capa_cos_hashing),
```
JannePeltonen, line 840: cls_hash_result_single_queue() is intended to be run even if hashing is not supported. In that odp_cls_hash_result() is supposed to return the single queue configured for the CoS. The spec for odp_cls_hash_result() does not say that it depends on hash capability and it appears that the function was meant to be usable for any CoS without the caller having to know the capabilities or the configured queues in the CoS.

satheeshpaul:
This change is not required. Closing the pull request.

PR #2184, opened 2025-02-25
[PATCH v4] api: system: allow odp_system_meminfo_t.total_overhead to be zero

MatiasElo:
V2:
- Dropped validation test change commit

### DIFF
--- a/include/odp/api/spec/system_info.h
+++ b/include/odp/api/spec/system_info.h
@@ -1,6 +1,6 @@
 /* SPDX-License-Identifier: BSD-3-Clause
  * Copyright (c) 2013-2018 Linaro Limited
- * Copyright (c) 2020-2021 Nokia
+ * Copyright (c) 2020-2025 Nokia
  */
 
 /**
@@ -226,7 +226,7 @@ typedef struct odp_system_meminfo_t {
 	 *
 	 * Total amount of memory (in bytes) that is currently consumed by roundings to
 	 * alignment/block/page size limits, etc. overheads. This is a subset of 'total_used'
-	 * bytes.
+	 * bytes. Zero if not supported by the implementation.
 	 */
 	uint64_t total_overhead;
 


test/validation/api/system/system.c
```diff
@@ -277,7 +277,8 @@ static void system_test_meminfo(void)
 	CU_ASSERT(ret == ret_0);
 	CU_ASSERT(info_0.total_mapped == info.total_mapped);
 	CU_ASSERT(info_0.total_used == info.total_used);
-	CU_ASSERT(info_0.total_overhead == info.total_overhead);
+	if (info_0.total_overhead)
```
JannePeltonen, line 280: The condition does not seem necessary. Even if total_overhead is not supported, its value (zero in that case) should be the same in the two calls.

include/odp/api/spec/system_info.h
```diff
@@ -1,6 +1,6 @@
 /* SPDX-License-Identifier: BSD-3-Clause
```
JannePeltonen, line 1: Typo in commit message: debuggin.

MatiasElo: Fixed in V2.

test/validation/api/system/system.c
MatiasElo: What I was thinking here was that an implementation could still fill `total_overhead` when it has to read the per memory block info anyways. But now thinking about it, perhaps  it's not the cleanest if the field value depends on other parameters.

JannePeltonen: The "zero if not supported by the implementation" text in the API makes me think that if total overhead count is supported, then it is returned in all calls. If it might be zero also (e.g.) when the return value is less than the num parameter, then I think the API probably should mention it.

JannePeltonen (approved):

psavol (approved):

jerinjacobk (approved):

PR #2186, opened 2025-02-26
[PATCH v3] linux-gen: include git tag in implementation string

Rayska:
odp_version_impl_str() output simplified as previously version was
displayed thrice.

odp_version_impl_str() returns either in format
"odp-linux-1.46.0.0-16a27d99", or "odp-linux-1.46.0.0" in the case that
the commit hash could not be retrieved. 

This uses the system default shell, any POSIX-compliant shell should be fine.


platform/linux-generic/odp_impl.c
```diff
@@ -15,7 +15,7 @@
 	ODP_VERSION_TO_STR(ODP_VERSION_API_GENERATION) "." \
 	ODP_VERSION_TO_STR(ODP_VERSION_API_MAJOR) "." \
 	ODP_VERSION_TO_STR(ODP_VERSION_API_MINOR) ") " \
-	ODP_VERSION_TO_STR(ODP_VERSION_BUILD)
+	ODP_VERSION_TO_STR(ODP_VERSION_BUILD) _ODP_VERSION_GIT_REVISION \
```
MatiasElo, line 18: The `odp_version_impl_str()` output includes quite much duplicate information:
E.g. `odp-linux 1.46.0-0 (v1.46.0) 1.46.0.0-87d04d69`

This could be condensed to:
`odp-linux-1.46.0.0-87d04d69`

configure.ac
MatiasElo, line 1: Better to mention `odp_version_impl_str()` API in the commit message than implementation internal `ODP_VERSION_IMPL_STR` define.

platform/linux-generic/odp_impl.c
```diff
@@ -5,17 +5,9 @@
...
-	ODP_VERSION_TO_STR(ODP_VERSION_API_GENERATION) "." \
-	ODP_VERSION_TO_STR(ODP_VERSION_API_MAJOR) "." \
-	ODP_VERSION_TO_STR(ODP_VERSION_API_MINOR) "-" \
-	ODP_VERSION_TO_STR(ODP_VERSION_IMPL) " (v" \
-	ODP_VERSION_TO_STR(ODP_VERSION_API_GENERATION) "." \
-	ODP_VERSION_TO_STR(ODP_VERSION_API_MAJOR) "." \
-	ODP_VERSION_TO_STR(ODP_VERSION_API_MINOR) ") " \
-	ODP_VERSION_TO_STR(ODP_VERSION_BUILD)
+	_ODP_IMPLEMENTATION_NAME "-" \
+	ODP_VERSION_TO_STR(ODP_VERSION_BUILD) _ODP_GIT_REVISION \
```
MatiasElo, line 10: Extra backslash at the end. Also, the implementation version string could be formed directly in `odp_version_impl_str()` without adding this `ODP_VERSION_IMPL_STR` define.

MatiasElo (approved):

PR #2190, opened 2025-03-07
[PATCH v3] api: clean 'extern "C"' usage

MatiasElo:
Use 'extern "C"' linkage specification only where necessary. The previous
API header level usage caused problems in C++ builds if the ODP
implementation (or its includes) used function overloading.

Always do all includes outside 'extern "C"' to avoid affecting included
headers.


TuomasTaipale (approved):
Seems OK.

PR #2194, opened 2025-03-21
api: ml: add support for engine id

syalavarthi:
In devices with multiple ML engines, support should be enabled to present the number of engines supported by device and options to select the engine ID to be used to config and load a model. This patch adds the field max_engines to ML capability and engine ID to config and model params.

Added implementation for linux-generic platform.

### DIFF
--- a/include/odp/api/spec/ml_types.h
+++ b/include/odp/api/spec/ml_types.h
@@ -181,6 +181,9 @@ typedef struct odp_ml_compl_pool_param_t {
 
 /** Machine learning capabilities */
 typedef struct odp_ml_capability_t {
+	/** Maximum number of engines */
+	uint32_t max_engines;
+
 	/** Maximum number of models
 	 *
 	 *  Maximum number of models that can be created simultaneously. The value is zero when
@@ -329,6 +332,14 @@ typedef struct odp_ml_capability_t {
 
 /** Machine learning configuration parameters */
 typedef struct odp_ml_config_t {
+	/**
+	 * Engine ID to be configured.
+	 *
+	 * In a system with multiple ML engines, this parameter selects the engine to be configured.
+	 * The default value is 0.
+	 */
+	uint32_t engine_id;
+
 	/**
 	 * Maximum number of models
 	 *
@@ -566,6 +577,9 @@ typedef struct odp_ml_model_info_t {
 	 */
 	uint64_t interface_version;
 
+	/** Engine ID to which the model is assigned */
+	uint32_t engine_id;
+
 	/** Model index assigned by the implementation */
 	uint32_t index;
 
@@ -598,6 +612,13 @@ typedef struct odp_ml_data_format_t {
  * Use odp_ml_model_param_init() to initialize the structure to its default values.
  */
 typedef struct odp_ml_model_param_t {
+	/**
+	 * Engine ID
+	 *
+	 * Engine ID to be used with the model. The default value is 0.
+	 */
+	uint32_t engine_id;
+
 	/**
 	 * Model binary
 	 *
--- a/platform/linux-generic/odp_ml.c
+++ b/platform/linux-generic/odp_ml.c
@@ -35,6 +35,7 @@
 #include <stdio.h>
 #include <string.h>
 
+#define ML_MAX_ENGINES 1
 #define ML_MAX_IO_SEGS UINT32_MAX
 #define ML_MAX_COMPL_ID 32
 #define ML_MAX_CONFIG_STR_LEN 65
@@ -151,6 +152,7 @@ int odp_ml_capability(odp_ml_capability_t *capa)
 		return 0;
 	}
 
+	capa->max_engines = 1;
 	capa->max_model_size = ML_MAX_MODEL_SIZE;
 	capa->max_models = ML_MAX_MODELS_CREATED;
 	capa->max_models_loaded = ML_MAX_MODELS_LOADED;
@@ -192,6 +194,7 @@ int odp_ml_capability(odp_ml_capability_t *capa)
 void odp_ml_config_init(odp_ml_config_t *config)
 {
 	memset(config, 0, sizeof(odp_ml_config_t));
+	config->engine_id = 0;
 	config->max_models_created = 1;
 	config->max_models_loaded = 1;
 }
@@ -200,6 +203,12 @@ int odp_ml_config(const odp_ml_config_t *config)
 {
 	if (!config) {
 		_ODP_ERR("Config must not be NULL\n");
+		return -1;
+	}
+
+	if (config->engine_id >= ML_MAX_ENGINES) {
+		_ODP_ERR("Engine ID %u exceeds maximum number of engines %d\n",
+			 config->engine_id, ML_MAX_ENGINES);
 		return -1;
 	}
 
@@ -679,6 +688,7 @@ static int create_ort_model(const odp_ml_model_param_t *param, OrtSession **sess
 		return -1;
 	}
 
+	mdl->info.engine_id = 0;
 	mdl->max_compl_id = param->max_compl_id;
 	mdl->info.num_inputs = num_inputs;
 	mdl->info.num_outputs = num_outputs;
@@ -832,6 +842,12 @@ odp_ml_model_t odp_ml_model_create(const char *name, const odp_ml_model_param_t 
 
 	if (odp_unlikely(odp_global_ro.disable.ml)) {
 		_ODP_ERR("ML is disabled\n");
+		return ODP_ML_MODEL_INVALID;
+	}
+
+	if (odp_unlikely(param->engine_id >= ML_MAX_ENGINES)) {
+		_ODP_ERR("Engine ID %u exceeds maximum number of engines %d\n",
+			 param->engine_id, ML_MAX_ENGINES);
 		return ODP_ML_MODEL_INVALID;
 	}
 


include/odp/api/spec/ml_types.h
```diff
@@ -329,6 +332,14 @@ typedef struct odp_ml_capability_t {
 
 /** Machine learning configuration parameters */
 typedef struct odp_ml_config_t {
+	/**
+	 * Engine ID to be configured.
+	 *
+	 * In a system with multiple ML engines, this parameter selects the engine to be configured.
+	 * The default value is 0.
+	 */
+	uint32_t engine_id;
```
psavol, line 341: odp_ml_config() is called only once per ODP instance / application. This field/spec change would limit application to use only a single engine, which I think is not the purpose of the PR. Either this config field would need to be "maximum number of engines" or odp_ml_config() would need to be called once per engine (to be used).

In general, ODP API engine concept could/should support engines with different capacity/capability (e.g. two engines: "large" and "small"). That would require (new) capability call per engine.

syalavarthi: This change would enable odp_ml_config() to be called once per engine.

max_engines field has been added to odp_ml_capability to fetch the number of engines available.

include/odp/api/spec/ml_types.h
```diff
@@ -329,6 +329,13 @@ typedef struct odp_ml_capability_t {
 
 /** Machine learning configuration parameters */
 typedef struct odp_ml_config_t {
+	/** Engine ID to be configured.
+	 *
+	 * In a system with multiple ML engines, this parameter selects the engine to be configured.
+	 * The default value is 0.
```
JannePeltonen, line 335: The description should state which engine_id values are valid, i.e. values less than odp_ml_engine_count().

include/odp/api/spec/ml_types.h
```diff
@@ -329,6 +329,13 @@ typedef struct odp_ml_capability_t {
...
+	 */
+	uint32_t engine_id;
```
JannePeltonen, line 337: I am not sure if we should use the term 'engine' or 'device'. Engine would make sense if we think that there, at some point, could be multiple devices, each of which could contain multiple engines. But I suppose we are not preparing for that. 'Engine' sounds ok to me, but I wonder if the same concept in other API's is more likely to be called 'device' and therefore 'device' would be better here. Maybe it does not matter.

include/odp/api/spec/ml.h
```diff
@@ -59,6 +59,18 @@ extern "C" {
  * values from 0 to max_compl_id range freely.
  */
 
+/**
```
JannePeltonen, line 62: In ODP there is a practise (although I do not know if it is any good) that API changes are in different commits than implementation changes. Maybe that makes things easier for some out-of-tree ODP implementations(?). Perhaps this PR should follow that practise? Maybe someone who knows better can comment.

include/odp/api/spec/ml_types.h
syalavarthi: Updated the odp_ml_config() function documentation.

include/odp/api/spec/ml_types.h
```diff
@@ -181,6 +181,9 @@ typedef struct odp_ml_compl_pool_param_t {
 
 /** Machine learning capabilities */
 typedef struct odp_ml_capability_t {
+	/** Maximum number of engines */
+	uint32_t max_engines;
```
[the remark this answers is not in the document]
syalavarthi: This field is removed from odp_ml_capability_t. Number of engines can be fetched with odp_ml_num_engines().

include/odp/api/spec/ml_types.h
```diff
@@ -566,6 +577,9 @@ typedef struct odp_ml_model_info_t {
 	 */
 	uint64_t interface_version;
 
+	/** Engine ID to which the model is assigned */
+	uint32_t engine_id;
```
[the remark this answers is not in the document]
syalavarthi: Updated the documentation with the permitted range of values for engine_id.

include/odp/api/spec/ml_types.h
syalavarthi: Updated the documentation with the permitted range of values for engine_id.

include/odp/api/spec/ml_types.h
syalavarthi: Based on discussions, we would like to go ahead with 'engine'.

include/odp/api/spec/ml.h
syalavarthi: I was trying to get the individual patches to compile without failures.
If there is no such requirement, I can resubmit the series with implementation changes as separate changes.

Please suggest.

JannePeltonen: No need to split the patches. All intermediate revisions must always compile and preferably also pass the tests, so simple splitting would not be enough.

platform/linux-generic/odp_ml.c
```diff
@@ -2582,6 +2594,17 @@ int _odp_ml_init_global(void)
 	memset(_odp_ml_glb, 0, sizeof(ml_global_t));
 	_odp_ml_glb->shm = shm;
 
+	num_engines = odp_ml_num_engines();
```
JannePeltonen: I do not see what the use of this call and the subsequent checks are. ML_MAX_ENGINES is defined in this file to be one, but this function does not even depend on that. If you really wanted, you could have a static assert that ML_MAX_ENGINES is >0, but I do not see any need for that eiher.

test/validation/api/ml/ml.c
```diff
@@ -52,6 +65,16 @@ static int check_ml_support(void)
...
 }
 
+static void test_ml_engine_count(void)
+{
+	int num_engines = odp_ml_num_engines();
+
+	CU_ASSERT(num_engines >= 0);
+
+	if (num_engines == 0)
+		return;
```
JannePeltonen: This if statement has no purpose. If num_engines is not zero, we also return.

example/ml/odp_ml_run.c
```diff
@@ -363,7 +364,7 @@ int main(int argc, char *argv[])
 		return -1;
 	}
 
-	if (odp_ml_capability(&glb.capa)) {
+	if (odp_ml_capability(ENGINE_ID, &glb.capa)) {
```
JannePeltonen, line 381: This calls odp_ml_capability() for engine zero without having first checked that the number of engines is at least one. This is against the proposed API which says that the valid range of the engine ID parameter is from 0 to odp_ml_num_engines(). (But see also the related comment on odp_ml_capability())

include/odp/api/spec/ml.h
```diff
@@ -78,12 +78,14 @@ int odp_ml_num_engines(void);
  * limits and its support of various ML API features. When ML offload is not available,
  * odp_ml_capability_t.max_models is zero.
  *
+ * @param[in] engine_id Engine ID to query capabilities. Valid range is 0 to num_engines - 1,
+ *                      where num_engines can be fetched using odp_ml_num_engines().
```
JannePeltonen, line 82: I started wondering if the restriction on the value range if the engine ID parameter is necessary. I am not sure if it is good to have it. Maybe it is. Another option would be to allow any engine ID, and then the engine IDs that fall outside the actual engine ID range would just return no capabilities.

include/odp/api/spec/ml_types.h
```diff
@@ -329,6 +329,15 @@ typedef struct odp_ml_capability_t {
 
 /** Machine learning configuration parameters */
 typedef struct odp_ml_config_t {
+	/**
+	 * Engine ID to be configured.
+	 *
+	 * In a system with multiple ML engines, this parameter selects the engine to be configured.
+	 * The default value is 0. Engine ID should be in the range 0..num_engines, where num_engines
```
JannePeltonen: Checkpatch complains about this line being too long and fails CI.

include/odp/api/spec/ml_types.h
```diff
@@ -672,6 +684,14 @@ typedef struct odp_ml_data_format_t {
  * Use odp_ml_model_param_init() to initialize the structure to its default values.
  */
 typedef struct odp_ml_model_param_t {
+	/**
+	 * Engine ID
+	 *
+	 * Engine ID to be used with the model. The default value is 0. Engine ID should be in the
+	 * range 0..num_engines, where num_engines can be fetched using odp_ml_num_engines().
+	 */
+	uint32_t engine_id;
```
JannePeltonen, line 693: I wonder if ML models (i.e. objects of odp_ml_model_t created through odp_ml_model_create()) should already be tied to specific engines or whether that should happen in the model load phase when odp_ml_model_load() is called?

platform/linux-generic/test/validation/api/ml/ml_linux.c
```diff
@@ -124,11 +124,23 @@ static int ml_suite_init(void)
...
 
+	num_engines = odp_ml_num_engines();
+	if (num_engines < 0) {
+		ODPH_ERR("ML engine count failed\n");
+		return num_engines;
+	}
+
+	if (num_engines == 0) {
+		ODPH_ERR("ML engine not available\n");
+		return -1;
```
JannePeltonen: Not having ML engines is not a bug that should cause API validation tests to fail. The code already uses global.disabled to indicate that ML is not supported and skips tests based on that. The same global can be set if the number of engines is zero.

But the check for the number of engines is not even needed here. If the number of engines is zero, then surely the maximum number of models is zero and the code already disables ML tests.

Actually, this illustrates how odp_ml_num_engines() is rather useless when it is the only thing added in the API, without any use for the number of engines anywhere. An application has no need to ever call odp_ml_num_engines(). Therefore it might be best to squash the two API patches, i.e. the one that introduces odp_ml_num_engines() and the one that makes use of it in other ML API functions.

And the same comment for other occurences of this pattern: There is no use in calling odp_ml_num_engines() yet.

include/odp/api/spec/ml.h
syalavarthi: I think it would lead to unnecessarily checking capabilities for engine_id's that are not valid. Restricting the range would be better.

include/odp/api/spec/ml_types.h
syalavarthi: Either should be possible, mapping the models during create stage or during load stage.

However, on devices where there is a necessity of managing the ML accelerator device resources like memory, slots etc., mapping the model during the create stage would enable for better management of resources. And tying the model during the load stage would require change in load function signatures.

platform/linux-generic/test/validation/api/ml/ml_linux.c
syalavarthi: In case of a system with number of configurable HW engines, we would require the calls to odp_ml_num_engines to determine the number of engines. Based on the num_engines, call to odp_ml_capbility can be used to fetch engine specific capabilities. Also, in certain cases, it would be required to dynamically choose the engine to be used to load a model.

I agree that checking for num_engines may not be required in the driver, but it is very much necessary on the application end on certain devices.

test/validation/api/ml/ml.c
syalavarthi: Agreed. Will be updated in next versionn.

platform/linux-generic/odp_ml.c
syalavarthi: Removed these checks in next version.

include/odp/api/spec/ml.h
syalavarthi: Squashed all patches in v7

example/ml/odp_ml_run.c
syalavarthi: added required checks.

include/odp/api/spec/ml_types.h
syalavarthi: resolved in latest patch

include/odp/api/spec/ml_types.h
```diff
@@ -329,6 +329,15 @@ typedef struct odp_ml_capability_t {
 
 /** Machine learning configuration parameters */
 typedef struct odp_ml_config_t {
+	/**
+	 * Engine ID to be configured.
+	 *
+	 * In a system with multiple ML engines, this parameter selects the engine
+	 * to be configured. The default value is 0. Engine ID should be in the range
+	 * 0..num_engines, where num_engines can be fetched using odp_ml_num_engines().
```
TuomasTaipale: The range should be `0..num_engines - 1`, right (like documented in `odp_ml_capability()`)?

include/odp/api/spec/ml_types.h
```diff
@@ -672,6 +684,14 @@ typedef struct odp_ml_data_format_t {
  * Use odp_ml_model_param_init() to initialize the structure to its default values.
  */
 typedef struct odp_ml_model_param_t {
+	/**
+	 * Engine ID
+	 *
+	 * Engine ID to be used with the model. The default value is 0. Engine ID should be in the
+	 * range 0..num_engines, where num_engines can be fetched using odp_ml_num_engines().
```
TuomasTaipale: Same here, The range should be `0..num_engines - 1`?

include/odp/api/spec/ml.h
JannePeltonen: > I think it would lead to unnecessarily checking capabilities for engine_id's that are not valid.

I do not understand this comment. If the API gives more freedom to application (regarding allowed parameter values), it does not force an application to do any additional work. Anyway, we could say in the API that an out-of-range engine ID causes the function to return a failure. This would be different from disallowing out-of-range values because then out-of-range values could lead to undefined behaviour which would be nice to avoid, I think..

platform/linux-generic/test/validation/api/ml/ml_linux.c
JannePeltonen: The comment about odp_ml_num_engines() calls being unnecessary was in the context of the patch that introduced the API change and the corresponding changes in tests and examples. Unfortunately github cannot (?) show which commit of a series a comment is for. In that code revision it was not necessary to ever call odp_ml_num_engines() since the need to know the number of engines only arose in a future commit. But now that the patches are squashed as I suggested, the problem has gone away.

platform/linux-generic/test/validation/api/ml/ml_linux.c
```diff
@@ -124,12 +125,24 @@ static int ml_suite_init(void)
...
-	if (odp_ml_capability(ml_capa)) {
+	num_engines = odp_ml_num_engines();
+	if (num_engines < 0) {
+		ODPH_ERR("ML engine count failed\n");
+		return num_engines;
+	}
+
+	if (num_engines == 0) {
+		ODPH_ERR("ML engine not available\n");
+		return -1;
```
JannePeltonen: This is failing the test if ML is not supported, which did not use to be that case before this patch. This does not actually fail  CI or 'make check' currently because without ML enabled the test is not run. But I think the behaviour change is still wrong and the test should just be disabled if ML is not supported.

include/odp/api/spec/ml.h
```diff
@@ -59,19 +59,33 @@ extern "C" {
...
  * @param[out] capa     Pointer to a capability structure for output
  *
  * @retval  0 on success
  * @retval <0 on failure
  */
-int odp_ml_capability(odp_ml_capability_t *capa);
+int odp_ml_capability(uint32_t engine_id, odp_ml_capability_t *capa);
```
JannePeltonen, line 88: I have heard about some concerns about the backward incompatible nature of this API change. Maybe this API change could be made in a backward compatible manner somehow like this:
- keep odp_ml_capability() as it is currently
- Introduce odp_ml_eng_capability() which takes engine ID
- Designate a special engine_id value that means implementation-selected or "default" engine. That could be zero (in which case the first numbered engine would be 1) or ODP_ML_ENGINE_ANY or something like that
- Have the default value of engine_id in model parameters be the special value mentioned above and have the implementation provide the current behaviour in that case.
This would make engine_id optional and old applications could continue ignoring it.

I am not sure if this approach is actually desired by application by I am just bringing it up as an example on how to possibly retain backward compatibility while simultaneously introducing engine IDs.

include/odp/api/spec/ml_types.h
syalavarthi: This is a typo. Will fix in next version.

platform/linux-generic/test/validation/api/ml/ml_linux.c
syalavarthi: Noted. Will fix in next version.

include/odp/api/spec/ml.h
syalavarthi: The current patch to add engine ID would not maintain backward compatibility, not just due the change in odp_ml_capability API, but also due to changes in the odp_ml_config_t,  odp_ml_model_info_t, odp_ml_model_param_t structures.

Considering the fact that model that would be handled is mapped to an engine, the changes in the structures and APIs are necessary and can't be avoided. And, adding the new odp_ml_eng_capability API, the backward compatibility issue can't be addressed.

Also, picking the default engine ID to a specific value may cause issues when the each of the engines present in the system have different capabilities. In such cases, choosing the engine based on the capabilities would be better. I think it would always be a good approach to let the user/app pick the required engine ID to load the model.

Moreover, the new suggested API odp_ml_eng_capability would do the similiar operation odp_ml_capability.

include/odp/api/spec/ml_types.h
TuomasTaipale: Typo still here?

include/odp/api/spec/ml.h
JannePeltonen: No, it is possible to introduce engine IDs in a backward compatible manner in a way I sketched above. In ODP, changes that introduce new struct fields are considered backward compatible changes if the struct fields have default values that corresponds to the old behaviour of the API and if the struct has an init function that an application is expected to call to initializse the structure. For example, the engine_id parameter could be introduced in this way:

```
 typedef struct odp_ml_config_t {
+       /**
+        * Engine ID to be configured.
+        *
+        * In a system with multiple ML engines, this parameter selects the
+        * engine to be configured. Values in the range 1..odp_ml_num_engines()
+        * select a specific engine.and the special value of 0 means that ODP
+        * selects the engine to be used.
+        *
+        * The default value is 0.
+        */
+       uint32_t engine_id;
```

TuomasTaipale:
Backward compatible version implemented in #2238.

PR #2198, opened 2025-04-02
api: ml: add quantization parameters to input and output info

TuomasTaipale:
Add quantization parameters to input and output info. Implementation may optionally provide these parameters, if they are available for a given model.

Continues #2182.


include/odp/api/spec/ml_types.h
```diff
@@ -512,6 +512,37 @@ typedef struct odp_ml_shape_info_t {
...
+
+	/** Quantization zero point */
+	int32_t zerop_i32;
+
+} odp_ml_quant_param_t;
+
+/** Quantization information */
+typedef struct odp_ml_quant_info_t {
+	/** Quantization parameters common to all input / output data */
+	odp_ml_quant_param_t common;
```
psavol, line 542: Suggesting to change wording to be more clear that it's not common to all data of all inputs (outputs), but all data of one input (output):
"... common to all input / output data" -> " ... common to all data values of an input / output"

PR #2200, opened 2025-04-10
[PATCH v2] linux-gen: sched: optimize slightly the case when stash is disabled

JannePeltonen:
Use a simpler ord_enq_multi function in the basic scheduler when event stashing at enqueue is disabled, which is now the default config setting.


JannePeltonen:
v2: rebased & review tag added

PR #2203, opened 2025-04-17
[PATCH v3] Fix Coverity warnings

MatiasElo:


platform/linux-generic/odp_packet_io.c
```diff
@@ -1466,7 +1466,9 @@ void odp_pktio_print(odp_pktio_t hdl)
 		       (entry->state ==  PKTIO_STATE_OPENED ? "opened" :
 							      "unknown"))));
 	memset(addr, 0, sizeof(addr));
-	odp_pktio_mac_addr(hdl, addr, ETH_ALEN);
+	if (odp_pktio_mac_addr(hdl, addr, ETH_ALEN) < 0)
+		_ODP_ERR("Unable to get default MAC address\n");
```
JannePeltonen, line 1470: I think 'default' could be dropped from the error message. I did notice that odp_pktio_mac_addr_set() and odp_pktio_mac_addr() talk about 'default mac address' instead of simply 'mac address', but I do not see why as there does not seem to be any other kind of (non-default?) mac addresses in ODP pktios (e.g. multiple mac addresses but one of them being 'default').

JannePeltonen (approved):
Looks good. I made a small comment about one error message text in case you would like to change it.

platform/linux-generic/odp_packet_io.c
MatiasElo: I've no recollection of why the term 'default' was added to the spec in the first place. Perhaps it was added to be future-proof in case multiple MAC addresses were supported later on. Removed it from the error message in V2.

PR #2204, opened 2025-04-17
[PATCH v2] Misc ring improvements

MatiasElo:


platform/linux-generic/include/odp_config_internal.h
```diff
@@ -209,6 +209,10 @@ extern "C" {
 /* Maximum number of outputs for a ML model. */
 #define CONFIG_ML_MAX_OUTPUTS 4
 
+/* Number of empty cache lines added by _ODP_CACHE_PAD macro. Set to zero to
+ * disable padding. */
+#define CONFIG_CACHE_PAD_LINES 1
```
JannePeltonen, line 214: I am not sure if the padding should be enabled right away. One possibility would be to merge the patches with the padding still disabled until further testing is done. But if you are already confident about this, then enabling the padding right away is ok too.

Another thing is that single config option may not be enough in the future if the padding gets added to multiple different places. But I suppose that is an issue to be handled at that time.

MatiasElo: The changes have been already tested on multiple platforms (x86 and arm), so I'd rather take them into use to get feedback from users. I'll give a heads-up to the relevant people.

Yep, we may end up requiring additional options later on.

PR #2189, opened 2025-03-07
[PATCH v13] api: sched: enable additional global and group level configurability

TuomasTaipale:
Application can now add additional global and group level scheduler configuration to signal maximum number of groups and priorities that need to be supported. These enable potential further resource optimizations.

On top of #2188

v2:
- Matias' comments

v3:
- Matias' comments

v4:
- Matias' comments
- Added a typo fixing patch

v5:
- Added possibility for new group specific priority functions to fail

v6:
- Clarified priority level value space

v7:
- Added reviewed-by tags

v8:
- Petri's comment
- Added reviewed-by tags

v9:
- Rebased
- Added scheduler capability to signal the minimum priority value supported by an implementation, this is to be utilized when creating predefined scheduling groups during global scheduler configuration

v10:
- Rebased
- Petri's comment
- Added implementation and validation tests

v11:
- Matias' comments

v12:
- Petri's comment

v13:
- Matias' comment
- Added reviewed-by tags

PR #2208, opened 2025-04-28
api: sched: enable additional global and group level configurability

TuomasTaipale:
Application can now add additional global and group level scheduler configuration to signal maximum number of groups and priorities that need to be supported. These enable potential further resource optimizations.

Previous version with less granular configurability in #2189.


include/odp/api/spec/schedule_types.h
```diff
@@ -558,8 +686,24 @@ typedef struct odp_schedule_config_t {
...
-	const char    *name;   /**< Schedule group name */
-	odp_thrmask_t thrmask; /**< Thread mask of the schedule group */
+	/** Schedule group name */
+	const char *name;
+
+	/** Thread mask of the schedule group */
+	odp_thrmask_t thrmask;
+
+	/** Group specific default priority */
+	odp_schedule_prio_t def_prio;
```
JannePeltonen, line 696: What is group specific default priority? What signifigance is has in API? Reading the API I cannot really understand it, so some API text improvement is probably needed somewhere. The global default priority returned by odp_schedule_default_prio() is the default value written to odp_schedule_param_t:: prio by odp_queue_param_init(), but there does not seem to be any connection between the group specific priority and the rest of the API.

include/odp/api/spec/schedule_types.h
```diff
@@ -165,6 +165,14 @@ extern "C" {
  * Predefined scheduler group of all control threads
  */
 
+/**
+ * @def ODP_SCHED_MAX_PRIOS
+ * Maximum number of priorities supported by the API
```
psavol, line 170: Maximum number of scheduling priorities...

platform/linux-generic/odp_schedule_basic.c
```diff
@@ -285,6 +285,8 @@ typedef struct {
 		odp_thrmask_t  mask;
 		uint16_t       spread_thrs[MAX_SPREAD];
 		uint8_t        allocated;
+		int            level[NUM_PRIO];
```
MatiasElo: `uint8_t` should be enough for the implementation.

include/odp/api/spec/schedule_types.h
```diff
@@ -211,6 +222,18 @@ typedef struct odp_schedule_capability_t {
 	 *  application can create 'max_groups' - 3 groups. */
 	uint32_t max_groups;
 
+	/** Maximum simultaneous number of group and priority combinations. The
+	 *  value includes the enabled predefined scheduling group default
+	 *  priority counts (ODP_SCHED_GROUP_ALL, ODP_SCHED_GROUP_WORKER, and
+	 *  ODP_SCHED_GROUP_CONTROL). By default, an application can create
+	 *  'max_group_prios' - 3 * 'max_prios'. */
```
MatiasElo: Seems like the last sentence is missing a word. Perhaps for example `... - 3 * 'max_prios' combinations.`

platform/linux-generic/odp_schedule_basic.c
```diff
@@ -1141,13 +1201,122 @@ static void schedule_group_clear(odp_schedule_group_t group)
...
+		_ODP_ERR("Insufficient groups (attempted: %u, left: %u)\n", inc_grps,
+			 config->num_groups);
+		return -1;
+	}
+
+	if (inc_grp_prios > config->num_group_prios) {
+		_ODP_ERR("Insufficient group priorities (attempted: %u, left: %u)\n",
+			 inc_grp_prios, config->num_group_prios);
+		return -1;
+	}
```
MatiasElo: Perhaps not the clearest wording for the error messages. Maybe `(required: %u, configured: %u)` would be a bit clearer.

platform/linux-generic/odp_schedule_basic.c
```diff
@@ -1141,13 +1201,122 @@ static void schedule_group_clear(odp_schedule_group_t group)
...
+	set_group_prios(ODP_SCHED_GROUP_CONTROL, &config->sched_group.control_param);
+	sched->config_if.num_groups = config->num_groups;
+	sched->config_if.num_group_prios = config->num_group_prios;
+	sched->config_if.num_prios = config->prio.num;
+	sched->config_if.min_prio = config->prio.min;
+	sched->config_if.max_prio = max_prio;
+	sched->config_if.def_prio = (sched->config_if.max_prio - sched->config_if.min_prio) / 2 +
+				    sched->config_if.min_prio;
+
+	for (int i = SCHED_GROUP_NAMED; i < NUM_SCHED_GRPS; i++) {
```
MatiasElo: Will the predefined scheduling groups use correct configurations if they don't include groups specific priority configs (i.e. `set_group_prios()` does nothing)?

platform/linux-generic/odp_schedule_basic.c
```diff
@@ -2223,7 +2460,7 @@ static void schedule_prefetch(int num)
 
 static int schedule_num_grps(void)
 {
-	return NUM_SCHED_GRPS - SCHED_GROUP_NAMED;
+	return sched->config_if.num_groups;
```
MatiasElo: This function seems unused.

platform/linux-generic/include/odp_schedule_if.h
```diff
@@ -30,6 +30,13 @@ typedef struct schedule_config_t {
 		int control;
 	} group_enable;
 
+	uint32_t num_groups;
+	uint32_t num_group_prios;
+	uint32_t num_prios;
```
MatiasElo: Reading the code could be a bit easier if these were named `max_groups`,  `max_group_prios`, and `max_prios`. Now there are comparisons like `if (sched->num_grps >= sched->config_if.num_groups)`, which is at least to me not immediately clear.

test/validation/api/scheduler/scheduler.c
```diff
@@ -200,6 +211,7 @@ static void scheduler_test_capa(void)
 	CU_ASSERT_FATAL(odp_queue_capability(&queue_capa) == 0);
 
 	CU_ASSERT(sched_capa.max_groups != 0);
+	CU_ASSERT(sched_capa.max_group_prios != 0);
 	CU_ASSERT(sched_capa.max_prios != 0);
```
MatiasElo, line 218: Test is missing for `odp_schedule_capability_t.min_prio`.

test/validation/api/scheduler/scheduler.c
```diff
@@ -200,6 +211,7 @@ static void scheduler_test_capa(void)
 	CU_ASSERT_FATAL(odp_queue_capability(&queue_capa) == 0);
 
 	CU_ASSERT(sched_capa.max_groups != 0);
+	CU_ASSERT(sched_capa.max_group_prios != 0);
```
MatiasElo: This check could perhaps be improved a  bit. Something like `max_group_prios >= max_groups && max_group_prios >= max_prios`.

platform/linux-generic/odp_schedule_basic.c
TuomasTaipale: There is now pointer assignment between this and the group parameter array, so less cumbersome to keep this with matching type.

platform/linux-generic/odp_schedule_basic.c
TuomasTaipale: Changed in v6. Kept the attempted/left in `allocate_group()` as it fits there better.

platform/linux-generic/odp_schedule_basic.c
TuomasTaipale: Global init sets a default level array for them (with max prios and min prio according to capabilities) but there was a missing check if the defaults are suitable anymore with the new schedule config, added that in v6.

MatiasElo: Perhaps the API is not the clearest about this, but I'd assume that also the predefined groups would use values from `odp_schedule_config_t.prio` unless groups specific configs are provided.

TuomasTaipale: Seems to be actually pretty clear:
```
/** Number of entries in 'level' array
 *
 *  When 0, the global priority range (see
 *  odp_schedule_config_t::prio) is used with this group. 0 by
 *  default.
 */
```
So, predefined groups should be receiving the incoming range config. I'll change it once more.

MatiasElo (approved):

include/odp/api/spec/schedule_types.h
```diff
@@ -482,16 +541,91 @@ typedef struct odp_schedule_group_param_t {
...
 	uint32_t num_queues;
 
 	/** Maximum number of events required to be stored simultaneously in
-	 * scheduled queue. This number must not exceed 'max_queue_size'
-	 * capability.  A value of 0 configures default queue size supported by
-	 * the implementation.
+	 *  scheduled queue
+	 *
+	 *  This number must not exceed 'max_queue_size' capability. A value of
+	 *  0 configures default queue size supported by he implementation.
```
psavol: typo: "by he" -> "by the"

PavanNikhilesh:

PR #2211, opened 2025-04-30
[PATCH v1] linux-gen: queue: add loop unrolling for event conversions

sanvai01:
Added loop unrolling for event to header conversions to improve performance for single producer single consumer and basic queues.

### DIFF
--- a/platform/linux-generic/odp_queue_basic.c
+++ b/platform/linux-generic/odp_queue_basic.c
@@ -515,20 +515,52 @@ static inline void event_index_from_hdr(uint32_t event_index[],
 					_odp_event_hdr_t *event_hdr[], int num)
 {
 	int i;
+	int idx = 0;
 
-	for (i = 0; i < num; i++)
-		event_index[i] = event_hdr[i]->index.u32;
+	for (i = 0; i < (num & ~0x3); i += 4, idx += 4) {
+		event_index[i] = event_hdr[idx]->index.u32;
+		event_index[i+1] = event_hdr[idx+1]->index.u32;
+		event_index[i+2] = event_hdr[idx+2]->index.u32;
+		event_index[i+3] = event_hdr[idx+3]->index.u32;
+	}
+	switch (num & 0x3) {
+	case 3:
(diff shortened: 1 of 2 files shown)


platform/linux-generic/odp_queue_basic.c
```diff
@@ -515,20 +515,52 @@ static inline void event_index_from_hdr(uint32_t event_index[],
...
 	int i;
+	int idx = 0;
+
+	for (i = 0; i < (num & ~0x3); i += 4, idx += 4) {
+		event_index[i] = event_hdr[idx]->index.u32;
+		event_index[i+1] = event_hdr[idx+1]->index.u32;
+		event_index[i+2] = event_hdr[idx+2]->index.u32;
+		event_index[i+3] = event_hdr[idx+3]->index.u32;
+	}
+	switch (num & 0x3) {
```
JannePeltonen, line 526: This looks like it is going to increase code size and add a few new branches (although I did not look at any generated assembly). It is not clear to me that the change is a net positive in actual applications and not just in microbenchmarks that presumably always enqueue long bursts. I believe some ODP apps have high instruction cache pressure.

Another possible optimization might be to get rid of the whole header-pointer-to-index conversion and just store the pointers in rings (but that would then use more cache lines in the rings, so I do not know if that is good or not).

vinodkrishna-arm: Yes, cache pressure is something we need to understand. Is there any benchmarking application (not a micro benchmark like we used here) that can be executed to confirm the behaviour?

platform/linux-generic/odp_queue_basic.c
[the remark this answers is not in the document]
vinodkrishna-arm: yes, this makes sense.

MatiasElo:
After this [commit](https://github.com/OpenDataPlane/odp/commit/f51e9587640277e6d80da82e76f108e28eef2d64) this PR seems no longer necessary.

TuomasTaipale:
Closed as unnecessary (f51e958).

PR #2212, opened 2025-05-02
[PATCH v3] test: sched_perf: add new event forwarding mode

MatiasElo:
V2:
- All review comments addressed

### DIFF
--- a/test/performance/odp_sched_perf.c
+++ b/test/performance/odp_sched_perf.c
@@ -42,7 +42,8 @@
 
 typedef struct test_options_t {
 	uint32_t num_cpu;
-	uint32_t num_queue;
+	uint32_t num_queue; /* Active queues (excludes dummy queues) */
+	uint32_t num_def;
 	uint32_t num_low;
 	uint32_t num_high;
 	uint32_t num_dummy;
@@ -57,7 +58,8 @@ typedef struct test_options_t {
 	int      fairness;
 	uint32_t event_size;
 	uint32_t queue_size;
-	uint32_t tot_queue;
+	uint32_t forward_group_size;
+	uint32_t tot_queue; /* All queues (includes dummy queues) */
 	uint32_t tot_event;
 	int      touch_data;
 	uint32_t stress;
@@ -101,7 +103,13 @@ typedef struct test_global_t {
 	odp_pool_t pool;
 	odp_cpumask_t cpumask;
 	odp_shm_t ctx_shm;
-	odp_queue_t queue[MAX_QUEUES];
+	struct {
+		odp_queue_t dummy[MAX_QUEUES];
+		odp_queue_t def_prio[MAX_QUEUES];
+		odp_queue_t low_prio[MAX_QUEUES];
+		odp_queue_t high_prio[MAX_QUEUES];
+		odp_queue_t all[MAX_QUEUES];
+	} queue;
 	odp_schedule_group_t group[MAX_GROUPS];
 	odph_thread_t thread_tbl[ODP_THREAD_COUNT_MAX];
 	test_stat_t stat[ODP_THREAD_COUNT_MAX];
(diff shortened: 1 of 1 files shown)


test/performance/odp_sched_perf.c
```diff
@@ -675,7 +673,7 @@ static int create_queues(test_global_t *global)
 	if (ctx_size) {
```
TuomasTaipale, line 830: Use `ODPH_ERR()` also for the two `printf()`s above as they end up returning a failure as well?

test/performance/odp_sched_perf.c
```diff
@@ -628,12 +645,157 @@ static int create_groups(test_global_t *global)
...
+	if (ctx_size == 0)
+		return 0;
+
+	for (uint32_t i = first; i < tot_queue; i++) {
+		odp_queue_t queue = global->queue.all[i];
+
+		/*
+		* Cast increases alignment, but it's ok, since ctx and
+		* ctx_size are both cache line aligned.
+		*/
```
TuomasTaipale: Comment alignment slightly off.

test/performance/odp_sched_perf.c
```diff
@@ -159,7 +165,11 @@ static void print_usage(void)
 	       "                         0: join all groups (default)\n"
 	       "  -b, --burst            Maximum number of events per operation. Default: 100.\n"
 	       "  -t, --type             Queue type. 0: parallel, 1: atomic, 2: ordered. Default: 0.\n"
-	       "  -f, --forward          0: Keep event in the original queue, 1: Forward event to the next queue. Default: 0.\n"
+	       "  -f, --forward          0: Keep event in the original queue (default)\n"
+	       "                         1: Forward event between all queues\n"
+	       "                         N: Forward events between queues of N identical queue groups. In this mode 'num_def',\n"
```
TuomasTaipale: I somehow think of "schedule groups" when reading "queue groups", maybe another term for these? Queue sets?

test/performance/odp_sched_perf.c
```diff
@@ -628,12 +645,157 @@ static int create_groups(test_global_t *global)
...
+		odp_queue_t queue = global->queue.all[i];
+
+		/*
+		* Cast increases alignment, but it's ok, since ctx and
+		* ctx_size are both cache line aligned.
+		*/
+		queue_context_t *qc = (queue_context_t *)(uintptr_t)ctx;
+
+		if (test_options->fairness)
+			odp_atomic_init_u64(&qc->count, 0);
```
TuomasTaipale, line 749: `qc` casting could be here now as the variable isn't used elsewhere.

test/performance/odp_sched_perf.c
```diff
@@ -628,12 +645,157 @@ static int create_groups(test_global_t *global)
...
+
+		if (odp_queue_context_set(queue, ctx, ctx_size)) {
```
TuomasTaipale: `queue` could be directly `global->queue.all[i]`.

test/performance/odp_sched_perf.c
```diff
@@ -689,100 +852,47 @@ static int create_queues(test_global_t *global)
...
-					prio = odp_schedule_max_prio();
-				} else {
-					num_low--;
-					prio = odp_schedule_min_prio();
-				}
-				break;
-			}
-		}
+	/* Dummy queues */
+	if (create_queues(global, &queue_param, num_group, global->queue.dummy,
```
TuomasTaipale, line 851: `create_queues` in practice behaves like "OK or NOK" so I guess the check here could be for just that (instead of `!= test_options->num_...`).

test/performance/odp_sched_perf.c
MatiasElo: Changed in V2.

test/performance/odp_sched_perf.c
MatiasElo: Fixed in V2.

test/performance/odp_sched_perf.c
MatiasElo: Queue set sounds fine, so changed to that in V2.

test/performance/odp_sched_perf.c
MatiasElo: Changed in V2.

test/performance/odp_sched_perf.c
MatiasElo: Changed in V2.

test/performance/odp_sched_perf.c
MatiasElo: Changed in V2.

PR #2213, opened 2025-05-05
[PATCH v3] test: queue_perf: add support for testing single event enq/deq operations

MatiasElo:
Add new option '-b 0' for testing odp_queue_enq() and odp_queue_deq() functions. By default multi variants odp_queue_enq_multi() and odp_queue_deq_multi() are used.

V2:
- Fixed review comments
- Use `odp_cpu_cycles_strict()` for reading CPU cycles

### DIFF
--- a/test/performance/odp_queue_perf.c
+++ b/test/performance/odp_queue_perf.c
@@ -462,7 +462,7 @@ static int run_test(void *arg)
 	odp_barrier_wait(thr_args->barrier);
 
 	t1 = odp_time_local_strict();
-	c1 = odp_cpu_cycles();
+	c1 = odp_cpu_cycles_strict();
 
 	for (rounds = 0; rounds < num_round; rounds++) {
 		int num_enq = 0;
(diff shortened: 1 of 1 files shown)


test/performance/odp_queue_perf.c
```diff
@@ -470,11 +473,22 @@ static int run_test(void *arg)
 		do {
 			src_queue = src_queue_tbl[queue_idx];
 			dst_queue = dst_queue_tbl[queue_idx];
+			num_ev = 0;
 
 			queue_idx++;
 			if (queue_idx == num_queue)
 				queue_idx = 0;
 
+			if (single) {
```
JannePeltonen, line 482: I wonder if it is ok to have this check here inside the polling loops. As the condition never changes, it could be moved outside the loop by having separate loops for the single and multi case and then the test loop overhead would be a tiny bit smaller. test_run() could be changed to an always-inlined function that takes 'single' as a parameter and then a new test_run could test for 'single' and call the function with a constant parameter so that the compiler would remove the condition testing from inside the loop.

JannePeltonen, line 482: There is also 'single' command line parameter for testing spsc queues. Perhaps it would be good to name this variable other than single to avoid confusion (multi?, enq_single?)?

MatiasElo: Changed to `single_event` in V2.

PR #2214, opened 2025-05-05
[PATCH v2] linux-gen: queue: optimize odp_queue_enq() and odp_queue_deq() implementations

MatiasElo:


JannePeltonen (approved):
It seems like the compiler should be able to optimize the _multi functions when invoked for single event, but that does not actually quite happen and therefore these changes seem to make sense.

PR #2216, opened 2025-05-12
test: common: normalize export result headers

Rayska:
Some applications export results with headers in form "Function name", and others in form "function name".
To normalize the format of exports, all headers should begin with a lower case letter. Lower case letters should be preferred for the entire header, but measurement units can be left capitalized when appropriate.

Additionally, odp_bench_misc exported some results from discrete tests under the same name. Now .desc of BENCH_INFO is used  for results instead of .name, if available.


test/performance/odp_bench_buffer.c
```diff
@@ -681,16 +681,16 @@ static int bench_buffer_export(void *data)
 	int ret = 0;
 
 	if (test_common_write("%s", gbl_args->appl.time ?
-			      "Function name,Average nsec per function call\n" :
-			      "Function name,Average CPU cycles per function call\n")) {
+		"function name,average nsec per function call\n" :
+		"function name,average cpu cycles per function call\n")) {
```
MatiasElo, line 685: Function parameters ought to be aligned to the same level as the opening brace whenever possible for better readability. Same comment for all `test_common_write() calls`.

MatiasElo (commented):
One minor tweak, otherwise looking OK. Please add my reviewed-by tags in the next version.

MatiasElo (approved):

PR #2217, opened 2025-05-12
[PATCH v7] api: event: add event type metadata to event vectors

JannePeltonen:
Change odp_event_vector_type() to not return the type of events actually stored in the event vector but the event type stored in event vector metadata and add odp_event_vector_type_set() to set the event type. Specify that event aggregators set the event type according to the event types they put in the event vector and require that applications use odp_event_vector_type_set() to update the event type of a vector after vector table modifications.

This way odp_event_vector_type() can be used largely as before, but an ODP implementation is not forced to loop through the vector every time odp_event_vector_type() is called.

### DIFF
--- a/platform/linux-generic/odp_event_vector.c
+++ b/platform/linux-generic/odp_event_vector.c
@@ -24,6 +24,7 @@ const _odp_event_vector_inline_offset_t _odp_event_vector_inline ODP_ALIGNED_CAC
 	.event     = offsetof(odp_event_vector_hdr_t, event),
 	.pool      = offsetof(odp_event_vector_hdr_t, event_hdr.pool),
 	.size      = offsetof(odp_event_vector_hdr_t, size),
+	.event_type = offsetof(odp_event_vector_hdr_t, event_type),
 	.uarea_addr = offsetof(odp_event_vector_hdr_t, uarea_addr),
 	.flags     = offsetof(odp_event_vector_hdr_t, flags)
 };
@@ -51,6 +52,7 @@ odp_event_vector_t odp_event_vector_alloc(odp_pool_t pool_hdl)
 		return ODP_EVENT_VECTOR_INVALID;
 
 	_ODP_ASSERT(event_vector_hdr_from_event(event)->size == 0);
+	_ODP_ASSERT(event_vector_hdr_from_event(event)->event_type == ODP_EVENT_ANY);
 
 	return odp_event_vector_from_event(event);
 }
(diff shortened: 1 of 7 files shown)


platform/linux-generic/odp_event_vector.c
```diff
@@ -61,29 +63,11 @@ void odp_event_vector_free(odp_event_vector_t evv)
 
 	evv_hdr->size = 0;
 	evv_hdr->flags.all_flags = 0;
+	evv_hdr->event_type = ODP_EVENT_ANY;
```
MatiasElo, line 66: ODP_EVENT_ANY happens to be zero, but I'd perhaps still add initialization to `init_event_hdr()`.

platform/linux-generic/odp_event_vector.c
```diff
@@ -51,6 +52,7 @@ odp_event_vector_t odp_event_vector_alloc(odp_pool_t pool_hdl)
 		return ODP_EVENT_VECTOR_INVALID;
 
 	_ODP_ASSERT(event_vector_hdr_from_event(event)->size == 0);
+	_ODP_ASSERT(event_vector_hdr_from_event(event)->event_type == ODP_EVENT_ANY);
```
MatiasElo, line 55: `odp_event_vector.c` is used by odp-dpdk as-is and its `odp_event_vector_hdr_t` is a bit different. So using event accessor functions here and below would be preferable.

platform/linux-generic/odp_event_vector.c
JannePeltonen: Added in v4.

JannePeltonen:
v4:
- Added explicit initialization of event vector type in pool initialization to make the code more robust in case ODP_EVENT_ANY ever becomes non-zero.
- Added a new patch that improves testing of odp_event_vector_alloc()

test/validation/api/event/event_vector.c
```diff
@@ -52,6 +52,25 @@ static odp_pool_t create_buffer_pool(uint32_t num_items)
...
+	uint32_t size;
+
+	evv = odp_event_vector_alloc(evv_pool);
+	CU_ASSERT_FATAL(evv != ODP_EVENT_VECTOR_INVALID);
+
+	CU_ASSERT(odp_event_vector_size(evv) == 0);
+	CU_ASSERT(odp_event_vector_type(evv) == ODP_EVENT_ANY);
+	CU_ASSERT(odp_event_vector_user_flag(evv) == 0);
+	size = odp_event_vector_tbl(evv, &ev_tbl);
+	CU_ASSERT(size == 0);
```
MatiasElo, line 69: Could also check event vector source pool. `odp_event_vector_pool()` is currently untested.

MatiasElo (approved):
One minor comment.

test/validation/api/event/event_vector.c
JannePeltonen: Added an assert for that after alloc in v5

psavol (approved):
Reviewed API commit.

JannePeltonen:
v6: added review tag, rebased

PR #2218, opened 2025-05-12
[PATCH v7] api: cls: add PMR priority to break PMR matching ties

JannePeltonen:
Add configurable priority to PMRs so that when multiple PMRs match a packet a PMR with the highest priority is selected. If multiple PMRs of the same priority match a packet, leave it implementation specific which PMR gets selected. Add a capability for the maximum priority value.

### DIFF
--- a/include/odp/api/spec/classification.h
+++ b/include/odp/api/spec/classification.h
@@ -342,6 +342,18 @@ typedef struct odp_pmr_create_opt_t {
 	 */
 	uint64_t mark;
 
+	/** Rule priority
+	 *
+	 *  When multiple PMRs match a packet, a matching rule with the
+	 *  highest priority is selected. If there are multiple such rules,
+	 *  then it is implementation specific which one is selected.
+	 *
+	 *  Larger priority value corresponds to higher priority. The value
+	 *  must not be larger than odp_cls_capability_t::max_pmr_priority.
+	 *  The default value is 0.
+	 */
+	uint32_t priority;
+
 } odp_pmr_create_opt_t;
 
 /** Random Early Detection (RED)
@@ -584,6 +596,9 @@ typedef struct odp_cls_capability_t {
 
 	/** Maximum number of terms per composite PMR */
 	uint32_t max_terms_per_pmr;
+
+	/** Maximum priority value of a PMR */
+	uint32_t max_pmr_priority;
 
 	/** Maximum number of CoS supported */
 	uint32_t max_cos;
@@ -969,8 +984,11 @@ void odp_cls_pmr_create_opt_init(odp_pmr_create_opt_t *opt);
  *
  * Creates a PMR between source and destination Class of Service (CoS). A packet arriving to
  * a CoS is matched against all the PMRs that define it as their source CoS. A PMR match moves
- * the packet from the source to the destination CoS. If multiple PMRs of a CoS match with
- * the packet, it is implementation specific which PMR is selected.
+ * the packet from the source to the destination CoS.
+ *
+ * This function creates PMRs of priority zero. Use odp_cls_pmr_create_opt() to create PMRs of
+ * any priority. If multiple PMRs of the same priority match the packet, it is implementation
+ * specific which PMR is selected.
  *
  * A composite PMR is created when PMR parameters define more than one term. A composite PMR is
  * considered to match only if a packet matches with all its terms. It is implementation specific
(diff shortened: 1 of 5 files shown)


include/odp/api/spec/classification.h
```diff
@@ -342,6 +342,18 @@ typedef struct odp_pmr_create_opt_t {
...
+	/** Rule priority
+	 *
+	 *  When multiple PMRs match a packet, a matching rule with the
+	 *  highest priority is selected. If there are multiple such rules,
+	 *  then it is implementation specific which one is selected.
+	 *
+	 *  Larger priority value corresponds to higher priority. The value
+	 *  must not be larger than odp_cls_capability_t::max_pmr_priority.
+	 *  The default value is 0.
+	 */
```
kkirankk, line 354: So, in case of chained COS, which rule should take priority? Is it PMR with high priority or PMR with longest chain?

JannePeltonen: I am not sure if I understand the question. What do you mean with "PMR with the longest chain"?

CoSes from a tree that is rooted at a packet input. Coses are linked together through PMRs. Each CoS has zero or more PMRs that match packets that have reached the CoS. A matching PMR forwards a packet to its destination CoS. When a packet does not match any PMRs of a Cos, the packet has reached the final CoS after traversing a chain of CoSes. The action of the final CoS determines what happens to the packet.

When a packet matches multiple PMRs of a CoS, one of the PMRs has to be selected as the one that forwards the packet to the next CoS. Previously the selection was implementation specific. This PR adds priority to PMRs so that the highest priority matching PMR is the one that must be selected (and if there are multiple, the selection between them is still implementation specific).

The number of terms in a PMR does not make a difference in PMR selection (except possibly when the selection is implementation specific).

kkirankk: Acked-by: Kiran Kumar Kokkilagadda <kirankumark@marvell.com>

include/odp/api/spec/classification.h
```diff
@@ -969,8 +984,9 @@ void odp_cls_pmr_create_opt_init(odp_pmr_create_opt_t *opt);
  *
  * Creates a PMR between source and destination Class of Service (CoS). A packet arriving to
  * a CoS is matched against all the PMRs that define it as their source CoS. A PMR match moves
- * the packet from the source to the destination CoS. If multiple PMRs of a CoS match with
- * the packet, it is implementation specific which PMR is selected.
+ * the packet from the source to the destination CoS. This function creates PMRs of priority
+ * zero. If multiple PMRs of the same priority match the packet, it is implementation specific
+ * which PMR is selected. Use odp_cls_pmr_create_opt() to create PMRs with non-zero priorities.
```
TuomasTaipale: Nitpick, but one can create zero-priority PMRs also with `odp_cls_pmr_create_opt()` so perhaps use a different word than "non-zero"? Maybe "specific" or "varying" or something else?

JannePeltonen: I think the text was ok as it was in the context of odp_cls_pmr_create() to tell what function can be used instead when a PMR with a non-zero priority needs to be created. And that statement does not imply that odp_cls_pmr_create_opt() could not be used for creating PMRs with zero priority too. And that function anyway has its own documentation. Anyway, I slightly tweaked the text in v2.

JannePeltonen:
v2:
- very slightly tweaked API text in one place (no semantic change)
- added minimal implementation with max priority zero (i.e. no PMR prioritization support)
- added tests that were tested with a non-minimal PMR priority implementation

test/validation/api/classification/odp_classification_test_pmr.c
```diff
@@ -348,6 +354,328 @@ static void cls_pmr_term_tcp_dport_n(int num_pkt)
...
+	if (prio_1 > prio_2)
+		expected_queue = odp_cos_queue(cos_1);
+	else
+		expected_queue = odp_cos_queue(cos_2);
+	CU_ASSERT(expected_queue != ODP_QUEUE_INVALID);
+
+	pkt = create_udp_packet(CLS_DEFAULT_DPORT);
+	seqno = send_packet(pkt, ts->pktio);
+	pkt = receive_and_check(seqno, expected_queue, ts->default_pool, false);
+	odp_packet_free(pkt);
```
TuomasTaipale: If you want to further simplify, this seems to repeat and could be moved to its own function.

JannePeltonen: Moved the repeating part to a function in v4. That also fixed the bugs that in two call sites the return value of receive_and_check() was not read.

TuomasTaipale (approved):

psavol (approved):
Reviewed API commit.

JannePeltonen:
v6: added review tag, rebased

PR #2219, opened 2025-05-26
[PATCH v2] test: l2fwd: add new extra load options

MatiasElo:


test/performance/odp_l2fwd.c
```diff
@@ -2022,7 +2036,7 @@ static void parse_args(int argc, char *argv[], appl_args_t *appl_args)
 		{NULL, 0, NULL, 0}
 	};
 
-	static const char *shortopts = "+c:t:a:i:m:o:O:r:d:s:e:k:g:G:I:"
+	static const char *shortopts = "+c:t:a:i:m:o:O:r:d:s:e:E:k:g:G:I:"
```
JannePeltonen, line 2046: I wonder if it would be better to just give up trying to come up with some random one-letter options for new features and just settle with long options. If someone actually used something like -E in a script, one would have to check from somewhere what it means, unlike (as much) with --memcpy.

test/performance/odp_l2fwd.c
```diff
@@ -534,6 +537,14 @@ static inline int process_extra_features(const appl_args_t *appl_args, odp_packe
 
 		if (appl_args->wait_ns)
 			odp_time_wait_ns(appl_args->wait_ns);
+
+		if (appl_args->memcpy_bytes) {
+			const uint64_t bytes = appl_args->memcpy_bytes;
+			uint8_t *src = &gbl_args->memcpy_data[thr_idx * 2 * bytes];
```
JannePeltonen: I guess the compiler could theoretically optimize this away, but I suppose it is not smart enough to do that, even with LTO?

MatiasElo: Calculation moved outside the test loop in V2.

test/performance/odp_l2fwd.c
MatiasElo: This would make sense. However, I was lazy and kept the short opt still in this commit to avoid having to modify the option parsing code.

test/performance/odp_l2fwd.c
JannePeltonen: That does not hurt, but I was actually referring to the (remote?) possibility that the compiler would be able to optimize the memcpy itself away. But I suppose it won't be able to do it any time soon and it would be noticed from weird test results.

JannePeltonen: hmm, maybe I clicked the wrong line when putting the original comment. Sorry about the confusion.

PR #2223, opened 2025-06-03
[PATCH v6] test: sched_perf: add options for control thread workers and event vector pools

Rayska:


test/performance/odp_sched_perf.c
```diff
@@ -507,9 +513,11 @@ static void print_options(test_options_t *options)
 	printf("  pool type                 %s\n", options->pool_type == ODP_POOL_BUFFER ?
 						   "buffer" : "packet");
 
-	printf("  queue type                %s\n\n", options->queue_type == 0 ? "parallel" :
+	printf("  queue type                %s\n", options->queue_type == 0 ? "parallel" :
 						     options->queue_type == 1 ? "atomic" :
 						     "ordered");
```
MatiasElo: Alignment a bit off.

test/performance/odp_sched_perf.c
```diff
@@ -821,6 +865,34 @@ static int create_queues(test_global_t *global)
...
+				if (evv == ODP_EVENT_VECTOR_INVALID) {
+					ODPH_ERR("Error: alloc failed %u/%u\n", i, j);
+					return -1;
+				}
+				odp_event_vector_tbl(evv, &tbl);
+
+				if (tbl == NULL) {
+					ODPH_ERR("Error: retrieval of event vector table failed\n");
+					return -1;
+				}
```
MatiasElo: This is a fast path API which is assumed to never fail, so no need to check `tbl` value.

test/performance/odp_sched_perf.c
```diff
@@ -1634,6 +1713,13 @@ int main(int argc, char **argv)
 		return -1;
 	}
 
+	if (global->test_options.pool_type == ODP_POOL_EVENT_VECTOR) {
+		if (odp_pool_destroy(global->evv_pool)) {
```
MatiasElo: Could be single if with `&&`.

test/performance/odp_sched_perf.c
```diff
@@ -1036,12 +1108,19 @@ static uint64_t rw_data_stress(odp_event_t ev[], int num, uint32_t rd_words, uin
...
+		if (pool_type == ODP_POOL_BUFFER) {
 			data = odp_buffer_addr(odp_buffer_from_event(ev[i]));
-		else
+		} else if (pool_type == ODP_POOL_PACKET) {
 			data = odp_packet_data(odp_packet_from_event(ev[i]));
+		} else {
+			evv = odp_event_vector_from_event(ev[i]);
+			odp_event_vector_tbl(evv, &tbl);
+			data = odp_buffer_addr(odp_buffer_from_event(tbl[0]));
+		}
```
MatiasElo, line 1206: The same change is also required for `rw_data()`.

test/performance/odp_sched_perf.c
```diff
@@ -550,10 +593,17 @@ static int create_pool(test_global_t *global)
...
+	} else if (test_options->pool_type == ODP_POOL_PACKET) {
 		max_num = pool_capa.pkt.max_num;
 		max_size = pool_capa.pkt.max_seg_len;
 		max_uarea = pool_capa.pkt.max_uarea_size;
+		max_pools = pool_capa.event_vector.max_pools;
+	} else {
+		max_num = pool_capa.event_vector.max_num;
+		max_size = pool_capa.event_vector.max_size;
+		max_uarea = pool_capa.event_vector.max_uarea_size;
+		max_pools = pool_capa.event_vector.max_pools;
```
MatiasElo: The capabilities should be checked for each pool type. In case of event vectors, buffer pool capabilities are currently not checked. One solution could be to move the checks to `create_pool_helper()`.

test/performance/odp_sched_perf.c
```diff
@@ -550,10 +593,17 @@ static int create_pool(test_global_t *global)
...
-	} else {
+		max_pools = pool_capa.event_vector.max_pools;
+	} else if (test_options->pool_type == ODP_POOL_PACKET) {
 		max_num = pool_capa.pkt.max_num;
 		max_size = pool_capa.pkt.max_seg_len;
 		max_uarea = pool_capa.pkt.max_uarea_size;
+		max_pools = pool_capa.event_vector.max_pools;
+	} else {
+		max_num = pool_capa.event_vector.max_num;
+		max_size = pool_capa.event_vector.max_size;
```
MatiasElo: Better to use some other variable than `max_size` here as max vector size is completely different compared to max buffer/packet size.

test/performance/odp_sched_perf.c
```diff
@@ -563,29 +604,23 @@ static int create_pool(test_global_t *global)
...
+					  uarea_size, event_size, max_size);
+	if (global->pool == NULL)
 		return -1;
-	}
 
-	global->pool = pool;
+	if (test_options->pool_type == ODP_POOL_EVENT_VECTOR) {
+		global->evv_pool = create_pool_helper("sched perf event vector content",
+						      ODP_POOL_BUFFER, tot_event, uarea_size, 16,
+						      pool_capa.buf.max_size);
```
MatiasElo: User area is not required for secondary pool and `max_size` parameter can also be zero.

test/performance/odp_sched_perf.c
```diff
@@ -504,24 +501,61 @@ static void print_options(test_options_t *options)
...
+						     "control");
 
 	printf("Extra rd/rw ops per event (queue context + user area + event data)\n");
 	printf("  read                      %u bytes\n", options->tot_rd_size);
 	printf("  write                     %u bytes\n\n", options->tot_rw_size);
 }
 
+static odp_pool_t create_pool_helper(const char *name, odp_pool_type_t pool_type,
+				     uint32_t tot_event, uint32_t uarea_size,
+				     uint32_t event_size, uint32_t max_size)
```
MatiasElo: Perhaps clearer to rename `max_size` to e.g. `vector_size`.

test/performance/odp_sched_perf.c
MatiasElo, line 1: It would be good to also test the new pool type during `make check` by updating `test/performance/odp_sched_perf_run.sh` script. One thing to note here is that the test should not fail if event vectors are not supported. Special return value 77 can be used with Automake to mark a test as skipped.

test/performance/odp_sched_perf_run.sh
```diff
@@ -35,6 +38,18 @@ run()
 			echo odp_sched_perf FAILED
 			exit $RET_VAL
 		fi
+
+		echo odp_sched_perf -p 2 -c $1
+		echo ===============================================
+		$TEST_DIR/odp_sched_perf${EXEEXT} -p 2 -c $1
+		RET_VAL=$?
+		if [ $RET_VAL -eq 77]; then
```
MatiasElo: Missing whitespace before closing square bracket causes a warning: `./odp_sched_perf_run.sh: 46: [: missing ]`

Also, `TEST_SKIPPED` variable could be used in the comparison.

test/performance/odp_sched_perf_run.sh
```diff
@@ -35,6 +38,18 @@ run()
...
+			echo "odp_sched_perf -p 2 -c $1 skipped (event vectors not supported)"
+			exit $TEST_SKIPPED
```
MatiasElo: Since we want to still continue running the tests for other worker counts with packets and buffers, the script should not exit here. Printing the message should be enough.

test/performance/odp_sched_perf.c
```diff
@@ -571,29 +579,65 @@ static int create_pool(test_global_t *global)
 		return -1;
 	}
 
+	if (pool_type == ODP_POOL_EVENT_VECTOR && !max_vector_pools) {
+		ODPH_ERR("Error: event vector pools not supported\n");
+		return 77;
```
MatiasElo: It'd be good to leave a comment here why 77 is returned.

MatiasElo (commented):
Rebase is required.

PR #2225, opened 2025-06-06
[PATCH v2] validation: ml: correct the data size for output data segment

lifang-zhang:


MatiasElo (approved):
Looks OK, please add a short description of the fix to the commit to make checkpatch happy.

PR #2229, opened 2025-06-10
[PATCH v1] test: sched_perf: add options for prefetching and cache stashing

Rayska:


test/performance/odp_sched_perf.c
```diff
@@ -189,6 +210,7 @@ static void print_usage(void)
 	       "  -u, --uarea_rd         Number of user area words (uint64_t) to read on every event. Default: 0.\n"
 	       "  -U, --uarea_rw         Number of user area words (uint64_t) to modify on every event. Default: 0.\n"
 	       "  -p, --pool_type        Pool type. 0: buffer, 1: packet, 2: event vector. Default: 0.\n"
+	       "  -P, --prefetch         Enable prefetching. 0: disabled, 1: enabled. Default: 0.\n"
```
MatiasElo: The number of prefetched events could be an option to `-P`.

test/performance/odp_sched_perf.c
```diff
@@ -1343,6 +1456,9 @@ static int test_sched(void *arg)
 							       ctx_rw_words);
 			}
 
+			if (prefetch_enable)
+				odp_schedule_prefetch(max_burst);
+
```
MatiasElo: I'd perhaps move this before this loop:
 ```
while (num) {
	num_enq = odp_queue_enq_multi(queue, &ev[i], num);
```
But picking an optimal location is difficult without actually testing the effects.

File: test/performance/odp_sched_perf.c [hunk 3]
```diff
@@ -167,6 +181,12 @@ static void print_usage(void)
 	       "                         if num_cpu is multiple of num_group and num_group is multiple of num_join.\n"
 	       "                         0: join all groups (default)\n"
 	       "  -b, --burst            Maximum number of events per operation. Default: 100.\n"
+	       "  -C, --cache_stash      Enable cache stashing. Format: offset,len,region,level\n"
```
TuomasTaipale: "Enable cache stashing" could be "Enable common group level cache stashing" to make it clear that this configures common stuff and does not touch priority specific stashing.

Also, matter of taste, but the format "region,level,offset,len" might be more natural for users.

test/performance/odp_sched_perf.c
```diff
@@ -275,6 +297,17 @@ static int parse_options(int argc, char *argv[], test_options_t *test_options)
...
+			if (cache_stash_size >= MAX_CACHE_STASH)
+				continue;
+			if (sscanf(optarg, "%u,%u,%u,%u",
+				   &test_options->cache_stash[cache_stash_size].offset,
+				   &test_options->cache_stash[cache_stash_size].len,
+				   &test_options->cache_stash[cache_stash_size].region,
+				   &test_options->cache_stash[cache_stash_size].level) != 4)
+				ret = -1;
+			cache_stash_size++;
+			break;
```
TuomasTaipale: The configuration of stashing could get a bit more straightforward if more stuff was done at this parsing phase. The current string scanning (done here), value validation (done below) and configuration setting (done in `init_stash_params()`) could be fused to a function, e.g. `parse_cache_stash_config()`, which would fill directly to a `odp_cache_stash_config_t` at `test_options_t` level (multiple `-C` to same region/level would just override the previous), the new function then could take `optarg` and this new field as parameters. Basically what you've already done but fused together. The current `init_stash_params()` (to be renamed) would then just configure the scheduler.

test/performance/odp_sched_perf.c
```diff
@@ -40,6 +40,9 @@
 /* Scheduling round interval to check for MAX_SCHED_WAIT_NS */
 #define TIME_CHECK_INTERVAL  (1024 * 1024)
 
+/* Max number of cache stashing options */
+#define MAX_CACHE_STASH 8
```
TuomasTaipale: This is now unnecessary.

test/performance/odp_sched_perf.c
```diff
@@ -51,7 +54,10 @@ typedef struct test_options_t {
 	uint32_t num_sched;
 	int      num_group;
 	uint32_t num_join;
+	uint32_t num_prefetch;
 	uint32_t max_burst;
+	uint32_t group_param_initialized;
+	odp_schedule_group_param_t group_param;
```
TuomasTaipale: For simplicity, I think you could have just the `odp_cache_stash_config_t` here. It's memset to zero in main so everything's off by default. Then the config could just be assigned to group params in `create_groups()` and `schedule_config()`. You can have the group params in those functions as stack variable and init the params there, those are slowpath functions so no need to optimize param init.

test/performance/odp_sched_perf.c
```diff
@@ -144,6 +150,72 @@ static int setup_sig_handler(void)
 	return 0;
 }
 
+static int set_cache_stash(odp_cache_stash_region_t *region, uint32_t level, uint32_t len,
```
TuomasTaipale: This function can't fail so could just return `void`.

TuomasTaipale (approved):

PR #2230, opened 2025-06-12
linux-gen: dpdk: misc fixes

MatiasElo:


platform/linux-generic/pktio/dpdk.c
```diff
@@ -2124,10 +2129,8 @@ static int dpdk_start(pktio_entry_t *pktio_entry)
 	return 0;
 }
 
-static int dpdk_stop(pktio_entry_t *pktio_entry)
+static int dpdk_stop(pktio_entry_t *pktio_entry ODP_UNUSED)
 {
-	rte_eth_dev_stop(pkt_priv(pktio_entry)->port_id);
```
JannePeltonen, line 2129: I am not sure this is the right approach.

The generic packet i/o code maintains the logical pktio state and already tries to ensure that the device level TX and RX routines are not called when the pktio is in the stopped state. Maybe the real problem is that the interface between the generic pktio code and the device level code is not clear enough and not adhered to by the generic code.

The generic pktio code seems to be racy with respect to maintaining and checking the pktio state. _pktio_stop() sets the pktio state to stopped/stop_pending only "after" it has called the device specific stop function. TX and RX routines may thus not see the state change early enough and call the device specific TX and RX routines when the device specific stop has already been done, which presumably caused the crash mentioned in the commit title. Furthermore, the state variable is not an atomic sync variable used with proper memory ordering, so just moving the write to it in _pktio_stop() would not be enough.

I think the common code should be fixed.

MatiasElo: Yeah, the packet IO state maintenance, or lack of it, is the culprit behind this issue. This commit is just a quick fix/workaround to enable performance testing. Fixing the pktio state maintenance is a bigger task and we need to consider how much overhead is accepted for the fast path functions.

PR #2231, opened 2025-06-19
validation: sched: add test for atomic event order

MatiasElo:
Add test for verifying that events from atomic queues are processed in the original enqueue order.


test/validation/api/scheduler/scheduler.c
```diff
@@ -3248,7 +3287,7 @@ static int create_queues(test_globals_t *globals)
 
 	odp_pool_param_init(&params);
 	params.buf.size = sizeof(queue_context);
-	params.buf.num  = prios * queues_per_prio * 2;
+	params.buf.num  = prios * queues_per_prio * 3;
```
TuomasTaipale, line 3291: Maybe a comment could be added here that "3" is for the different queue-type contexts.

test/validation/api/scheduler/scheduler.c
```diff
@@ -1841,31 +1858,35 @@ static void fill_queues(thread_args_t *args)
...
+
 	for (i = 0; i < num_prio; i++) {
 		for (j = 0; j < num_queues; j++) {
+			queue_context *qctx;
 			odp_queue_t queue;
 
-			snprintf(name, sizeof(name),
-				 "sched_%d_%d_o", i, j);
+			snprintf(name, sizeof(name), args->sync == ODP_SCHED_SYNC_ORDERED ?
+				 "sched_%d_%d_o" : "sched_%d_%d_a", i, j);
```
TuomasTaipale: Matter of taste but the format could be "sched_%d_%d_%s" and then the to be substituted string could be either "o" or "a".

test/validation/api/scheduler/scheduler.c
MatiasElo: Comment added in V2.

test/validation/api/scheduler/scheduler.c
MatiasElo: Changed in V2.

TuomasTaipale (approved):

PR #2232, opened 2025-06-24
linux-gen: queue: store event pointers instead of indices in queues

MatiasElo:
Store pointers to events instead of indices in queues. This can improve performance by removing the index-pointer conversions from enqueue and dequeue fast path functions. As a downside, this change effectively doubles the amount of memory required for queues in a 64-bit system.


platform/linux-generic/odp_queue_basic.c
```diff
@@ -515,38 +515,18 @@ static odp_queue_t queue_lookup(const char *name)
...
-		event_index[i] = event_hdr[i]->index.u32;
-}
-
-static inline void event_index_to_hdr(_odp_event_hdr_t *event_hdr[],
-				      uint32_t event_index[], int num)
-{
-	int i;
-
-	for (i = 0; i < num; i++) {
-		event_hdr[i] = _odp_event_hdr_from_index_u32(event_index[i]);
```
TuomasTaipale, line 533: Seems that `_odp_event_hdr_from_index_u32()` is now unused as well.

platform/linux-generic/odp_queue_basic.c
```diff
@@ -515,38 +515,18 @@ static odp_queue_t queue_lookup(const char *name)
...
-		odp_prefetch(event_hdr[i]);
-	}
-}
-
 static int plain_queue_enq(odp_queue_t handle, _odp_event_hdr_t *event_hdr)
 {
 	queue_entry_t *queue = qentry_from_handle(handle);
-	ring_mpmc_u32_t *ring_mpmc = &queue->ring_mpmc;
+	ring_mpmc_ptr_t *ring_mpmc = &queue->ring_mpmc;
 	uint32_t num_enq;
```
TuomasTaipale: Matter of taste, but seems like you could get rid of `num_enq` and `num_deq` in a few places.

platform/linux-generic/odp_queue_basic.c
MatiasElo: Removed in V2.

platform/linux-generic/odp_queue_basic.c
MatiasElo: Cleaned up in V2.

TuomasTaipale (approved):

PR #2233, opened 2025-06-24
[PATCH v9] test: bench: add new and missing functions

Rayska:


test/common/bench_common.c
MatiasElo, line 1: Copyright can be updated. Also, it's a good practice to give some motivation for the change in the commit message.

test/performance/odp_bench_event_vector.c
```diff
@@ -0,0 +1,584 @@
...
+/**
+ * Print usage information
+ */
+static void usage(char *progname)
+{
+	printf("\n"
+	       "OpenDataPlane Event vector API microbenchmarks.\n"
+	       "\n"
+	       "Usage: %s OPTIONS\n"
+	       "  E.g. %s\n"
```
MatiasElo: This line could be removed, or some example options could be added.

test/performance/odp_bench_event_vector.c
```diff
@@ -0,0 +1,584 @@
...
+	if (gbl_args->pool == ODP_POOL_INVALID) {
+		ODPH_ERR("Error: pool create failed\n");
+		exit(EXIT_FAILURE);
+	}
+
+	odp_pool_print(gbl_args->pool);
+
+	memset(&worker_thread, 0, sizeof(odph_thread_t));
+
+	signal(SIGINT, sig_handler);
```
MatiasElo: Use of old `signal()` function should be avoided. `sigaction()` should be used instead (see for example odp_l2fwd).

test/performance/odp_bench_event_vector.c
```diff
@@ -0,0 +1,584 @@
...
+		num_evv++;
+	}
+}
+
+static void create_vectors(void)
+{
+	allocate_test_vectors(gbl_args->evv_arr, TEST_REPEAT_COUNT);
+
+	for (int i = 0; i < TEST_REPEAT_COUNT; i++)
+		odp_event_vector_size_set(gbl_args->evv_arr[i], 1);
```
MatiasElo: As discussed, it would be good to fill the vectors with events. In the current form this is against the spec. Event vectors must only include valid event handles.

test/performance/odp_bench_event_vector.c
```diff
@@ -0,0 +1,584 @@
...
+	odp_pool_t pool = gbl_args->pool;
+	int i;
+
+	for (i = 0; i < TEST_REPEAT_COUNT; i++)
+		evv_arr[i] = odp_event_vector_alloc(pool);
+
+	return i;
+}
+
+static int event_vector_free(void)
```
MatiasElo, line 244: For comparison same kind of test could be added also for `odp_event_free()`.

test/performance/odp_bench_event_vector.c
```diff
@@ -0,0 +1,584 @@
...
+
+		odp_schedule_config_init(&sched_config);
+		sched_config.max_flow_id = 1;
+
+		if (odp_schedule_config(&sched_config)) {
+			ODPH_ERR("Error: schedule config failed\n");
+			exit(EXIT_FAILURE);
+		}
+		gbl_args->max_flow_id = 1;
+	}
```
MatiasElo: Unnecessary for this test.

test/performance/odp_bench_event_vector.c
```diff
@@ -0,0 +1,584 @@
...
+		params.event_vector.cache_size = gbl_args->appl.cache_size;
+	params.type = ODP_POOL_EVENT_VECTOR;
+
+	gbl_args->pool = odp_pool_create("microbench", &params);
+	if (gbl_args->pool == ODP_POOL_INVALID) {
+		ODPH_ERR("Error: pool create failed\n");
+		exit(EXIT_FAILURE);
+	}
+
+	odp_pool_print(gbl_args->pool);
```
MatiasElo: I'd split pool creation into a separate function to avoid bloating the `main()` function.

test/performance/odp_bench_event_vector.c
```diff
@@ -0,0 +1,760 @@
...
+	odp_event_vector_t *evv_arr = gbl_args->evv_arr;
+	uint32_t evv_size = gbl_args->evv_size;
+	odp_event_t *tbl;
+	int i;
+
+	for (i = 0; i < TEST_REPEAT_COUNT; i++) {
+		odp_event_vector_tbl(evv_arr[i], &tbl);
+		odp_event_free_multi(tbl, evv_size);
+		odp_event_vector_free(evv_arr[i]);
+	}
```
MatiasElo: We want to measure the cost of `odp_event_vector_free()` calls, so the stored events should freed by a separate clean-up function.

test/performance/odp_bench_event_vector.c
```diff
@@ -0,0 +1,760 @@
...
+	/** Common benchmark suite data */
+	bench_suite_t suite;
+	/** Event vector pool */
+	odp_pool_t evv_pool;
+	/** Buffer pool */
+	odp_pool_t buf_pool;
+	/** Number of events in event vector */
+	uint32_t evv_size;
+	/** Size of buffer */
+	uint32_t buf_size;
```
MatiasElo: Seems unnecessary to store this in global data.

test/performance/odp_bench_event_vector.c
```diff
@@ -0,0 +1,760 @@
...
+	/** Application (parsed) arguments */
+	appl_args_t appl;
+	/** Common benchmark suite data */
+	bench_suite_t suite;
+	/** Event vector pool */
+	odp_pool_t evv_pool;
+	/** Buffer pool */
+	odp_pool_t buf_pool;
+	/** Number of events in event vector */
+	uint32_t evv_size;
```
MatiasElo: Perhaps rename `appl_args_t.size` to `evv_size` and use only that one.

test/performance/odp_bench_event_vector.c
```diff
@@ -0,0 +1,760 @@
...
+		return -1;
+	} else if (gbl_args->appl.cache_size > (int)capa.event_vector.max_cache_size) {
+		ODPH_ERR("Error: cache size not supported (max %" PRIu32 ")\n",
+			 capa.event_vector.max_cache_size);
+		return -1;
+	}
+
+	gbl_args->buf_size = TEST_BUF_SIZE;
+
+	if (capa.event_vector.max_size && capa.event_vector.max_size < TEST_EVV_SIZE)
```
MatiasElo: No need to check for zero.

test/performance/odp_bench_event_vector.c
```diff
@@ -0,0 +1,760 @@
...
+			ODPH_ABORT("Allocating test event vector failed\n");
+
+		odp_event_vector_tbl(evv[i], &tbl);
+		if (odp_buffer_alloc_multi(buf_pool, buf, evv_size) != (int)evv_size) {
+			for (uint32_t j = 0; j < evv_size; j++) {
+				if (buf[j] == ODP_BUFFER_INVALID)
+					ODPH_ABORT("Alloc failed %u/%u\n", i, j);
+			}
+			ODPH_ABORT("Not enough buffers allocated!\n");
+		}
```
MatiasElo: Depending on the implementation, `odp_buffer_alloc_multi()` may return less than the requested number of buffers per call and only return value <0 should be considered a failure. So, a retry loop is required here. Also, no need to check the individual buffers after alloc.

test/performance/odp_bench_event_vector.c
```diff
@@ -0,0 +1,760 @@
...
+		return -1;
+
+	return 0;
+}
+
+static void allocate_test_evv_and_contents(odp_event_vector_t evv[], int num)
+{
+	odp_pool_t evv_pool = gbl_args->evv_pool;
+	odp_pool_t buf_pool = gbl_args->buf_pool;
+	uint32_t evv_size = gbl_args->evv_size;
```
MatiasElo: User could pass zero as number of events in vectors, which is a valid test case.

test/performance/odp_bench_event_vector.c
```diff
@@ -0,0 +1,760 @@
...
+{
+	odp_event_vector_t *evv = gbl_args->evv_arr;
+	odp_event_t *tbl;
+	uint32_t evv_size = gbl_args->evv_size;
+
+	for (int i = 0; i < TEST_REPEAT_COUNT; i++) {
+		odp_event_vector_tbl(evv[i], &tbl);
+		odp_event_free_multi(tbl, evv_size);
+		odp_event_vector_free(evv[i]);
+	}
```
MatiasElo: Can simply use `odp_event_free()`.

test/performance/odp_bench_event_vector.c
```diff
@@ -0,0 +1,760 @@
...
+	return i;
+}
+
+static int event_vector_user_flag_set(void)
+{
+	odp_event_vector_t *evv_arr = gbl_args->evv_arr;
+	int i;
+
+	for (i = 0; i < TEST_REPEAT_COUNT; i++)
+		odp_event_vector_user_flag_set(evv_arr[i], 0);
```
MatiasElo: The default is zero, so I'd set the flag to 1.

test/performance/odp_bench_event_vector.c
```diff
@@ -0,0 +1,760 @@
...
+	return i;
+}
+
+static int event_vector_user_flag(void)
+{
+	odp_event_vector_t *evv_arr = gbl_args->evv_arr;
+	int i;
+
+	for (i = 0; i < TEST_REPEAT_COUNT; i++)
+		odp_event_vector_user_flag(evv_arr[i]);
```
MatiasElo: Should return zero, so could be:
`ret += !odp_event_vector_user_flag(evv_arr[i]);`

test/performance/odp_bench_event_vector.c
```diff
@@ -0,0 +1,760 @@
...
+
+	for (i = 0; i < TEST_REPEAT_COUNT; i++)
+		event_type[i] = odp_event_vector_type(evv_arr[i]);
+	return i;
+}
+
+static int event_vector_size_set(void)
+{
+	odp_event_vector_t *evv_arr = gbl_args->evv_arr;
+	uint32_t size = gbl_args->evv_size;
```
MatiasElo: I'd change the value, so `evv_size - 1` (unless `evv_size` is zero).

test/performance/odp_bench_event_vector.c
```diff
@@ -0,0 +1,760 @@
...
+}
+
+static int event_vector_tbl(void)
+{
+	odp_event_vector_t *evv_arr = gbl_args->evv_arr;
+	odp_event_t *(*evv_tbl)[gbl_args->evv_size] = gbl_args->evv_tbl;
+	uint32_t ret = 0;
+
+	for (int i = 0; i < TEST_REPEAT_COUNT; i++)
+		ret += odp_event_vector_tbl(evv_arr[i], evv_tbl[i]);
```
MatiasElo: Vectors may be empty depending on user options.

test/performance/odp_bench_event_vector.c
```diff
@@ -0,0 +1,760 @@
...
+			return 0;
+
+		odp_buffer_to_event_multi(buf, tbl, evv_size);
+
+		if (odp_unlikely(evv == ODP_EVENT_VECTOR_INVALID))
+			return 0;
+
+		odp_event_free_multi(tbl, evv_size);
+		odp_event_vector_free(evv);
+	}
```
MatiasElo: We are only interested about `odp_event_vector_alloc()` and `odp_event_vector_free()` performance here, so no need to fill the vectors.

test/performance/odp_bench_event_vector.c
```diff
@@ -0,0 +1,760 @@
...
+	odp_buffer_t buf[gbl_args->evv_size];
+	odp_event_t *tbl;
+	int i;
+
+	for (i = 0; i < TEST_REPEAT_COUNT; i++) {
+		evv_arr[i] = odp_event_vector_alloc(evv_pool);
+		odp_event_vector_tbl(evv_arr[i], &tbl);
+		odp_buffer_alloc_multi(buf_pool, buf, evv_size);
+		odp_buffer_to_event_multi(buf, tbl, evv_size);
+	}
```
MatiasElo: Only `odp_event_vector_alloc()` needs to be called here.

test/performance/odp_bench_event_vector.c
```diff
@@ -0,0 +1,760 @@
...
+exit:
+	test_common_write_term();
+
+	return ret;
+}
+
+/**
+ * Test functions
+ */
+bench_info_t test_suite[] = {
```
MatiasElo, line 532: Test missing for `odp_event_vector_tbl()`.

test/performance/odp_bench_event_vector.c
```diff
@@ -0,0 +1,760 @@
...
+			 capa.event_vector.max_size < TEST_MAX_EVV_SIZE ?
+			 capa.event_vector.max_size : TEST_MAX_EVV_SIZE);
+		return -1;
+	}
+
+	gbl_args->evv_uarea_size = TEST_UAREA_SIZE < capa.event_vector.max_uarea_size ?
+				TEST_UAREA_SIZE : capa.event_vector.max_uarea_size;
+
+	gbl_args->buf_uarea_size = TEST_UAREA_SIZE < capa.buf.max_uarea_size ?
+				TEST_UAREA_SIZE : capa.buf.max_uarea_size;
```
MatiasElo: User area is not required for buffer pool.

test/performance/odp_bench_event_vector.c
```diff
@@ -0,0 +1,760 @@
...
+
+	odp_pool_print(gbl_args->evv_pool);
+
+	/* Create buffer pool */
+	odp_pool_param_init(&buf_pool_params);
+	buf_pool_params.buf.size = gbl_args->buf_size;
+	buf_pool_params.buf.num = TEST_REPEAT_COUNT * gbl_args->evv_size;
+	buf_pool_params.buf.uarea_size = gbl_args->buf_uarea_size;
+	if (gbl_args->appl.cache_size >= 0)
+		buf_pool_params.buf.cache_size = gbl_args->appl.cache_size;
```
MatiasElo: Buffer pool cache size does not matter for this test, so can be left to default.

test/performance/odp_bench_event_vector.c
```diff
@@ -0,0 +1,760 @@
...
+static void usage(void)
+{
+	printf("\n"
+	       "OpenDataPlane Event vector API microbenchmarks.\n"
+	       "\n"
+	       "Optional OPTIONS:\n"
+	       "  -c, --cache_size <num>  Pool cache size.\n"
+	       "  -i, --index <idx>       Benchmark index to run indefinitely.\n"
+	       "  -r, --rounds <num>      Run each test case 'num' times (default %u).\n"
+	       "  -s,  -size <num>        Number of events in each event vector (default %u).\n"
```
MatiasElo: It could be more flexible to have two options, one for the maximum vector size (always >0) and another for the number of events stored in each vector (can be zero).

test/performance/odp_bench_event_vector.c
```diff
@@ -0,0 +1,750 @@
...
+	       "\n"
+	       "Optional OPTIONS:\n"
+	       "  -c, --cache_size <num>  Pool cache size.\n"
+	       "  -i, --index <idx>       Benchmark index to run indefinitely.\n"
+	       "  -r, --rounds <num>      Run each test case 'num' times (default %u).\n"
+	       "  -m,  -max_events <num>  Maximum number of events in each event vector (default %u).\n"
+	       "  -n, --num_events <num>  Number of events in each event vector (default %u).\n"
+	       "  -t, --time <opt>        Time measurement. 0: measure CPU cycles (default), 1: measure time\n"
+	       "  -h, --help              Display help and exit.\n\n"
+	       "\n", TEST_ROUNDS, TEST_MAX_EVV_SIZE, TEST_EVV_SIZE);
```
MatiasElo: Looks like both `max_events` and `num_events` default to `TEST_EVV_SIZE`, which is fine, but the print is incorrect.

test/performance/odp_bench_event_vector.c
```diff
@@ -0,0 +1,750 @@
...
+/**
+ * Parsed command line arguments
+ */
+typedef struct {
+	int bench_idx;           /** Benchmark index to run indefinitely */
+	int cache_size;          /** Pool cache size */
+	int time;                /** Measure time vs. CPU cycles */
+	uint32_t rounds;         /** Rounds per test case */
+	uint32_t max_evv_size;   /** Maximum number of events in each event vector */
+	uint32_t num_evv;        /** Number of events in each event vector */
```
MatiasElo: A bit misleading naming. Commonly `evv` refers to event vectors, but this variable is used to store number of normal events.

test/performance/odp_bench_event_vector.c
```diff
@@ -0,0 +1,750 @@
...
+		default:
+			break;
+		}
+	}
+
+	if (appl_args->rounds < 1) {
+		printf("Invalid number test rounds: %d\n", appl_args->rounds);
+		exit(EXIT_FAILURE);
+	}
+
```
MatiasElo, line 464: Could also check that `appl_args->max_evv_size` != 0.

test/performance/odp_bench_event_vector.c
```diff
@@ -0,0 +1,750 @@
...
+{
+	odp_event_vector_t *evv_arr = gbl_args->evv_arr;
+	uint32_t num_evv = gbl_args->appl.num_evv;
+	uint32_t evv_size;
+
+	for (int i = 0; i < TEST_REPEAT_COUNT; i++) {
+		evv_size = odp_event_vector_size(evv_arr[i]);
+
+		if (evv_size != 0 && evv_size != num_evv)
+			odp_event_vector_size_set(evv_arr[i], num_evv);
```
MatiasElo: Assuming this size checking is due to `event_vector_size_set()` test. Normally this isn't necessary, so perhaps I'd use a separate function for `event_vector_size_set()` clean-up with comment why this is necessary.

test/performance/odp_bench_event_vector.c
```diff
@@ -0,0 +1,750 @@
...
+	return i;
+}
+
+static int event_vector_size_set(void)
+{
+	odp_event_vector_t *evv_arr = gbl_args->evv_arr;
+	int i;
+
+	for (i = 0; i < TEST_REPEAT_COUNT; i++)
+		odp_event_vector_size_set(evv_arr[i], 1);
```
MatiasElo: If `--num_events` is zero, this is against the spec. In that case the only valid `size` value is zero. So something like:
`const uint32_t size = gbl_args->appl.num_evv ? gbl_args->appl.num_evv - 1 : 0;`

test/performance/odp_bench_event_vector.c
```diff
@@ -0,0 +1,750 @@
...
+{
+	odp_pool_t evv_pool = gbl_args->evv_pool;
+	odp_pool_t buf_pool = gbl_args->buf_pool;
+	uint32_t max_evv_size = gbl_args->appl.max_evv_size;
+	uint32_t num_evv = gbl_args->appl.num_evv;
+	odp_event_t (*buf_event_tbl)[TEST_MAX_EVV_SIZE] = gbl_args->buf_event_tbl;
+	odp_buffer_t buf[max_evv_size];
+	odp_event_t *evv_tbl;
+	uint32_t num_buf;
+	int alloc_ret, i, j;
```
MatiasElo: `i`, `j`, and `num` could be `uint32_t`, so you get rid of the cast.

test/performance/odp_bench_event_vector.c
```diff
@@ -0,0 +1,750 @@
...
+	return i;
+}
+
+static int event_vector_alloc_free(void)
+{
+	odp_pool_t evv_pool = gbl_args->evv_pool;
+	int i;
+
+	for (i = 0; i < TEST_REPEAT_COUNT; i++)
+		odp_event_vector_free(odp_event_vector_alloc(evv_pool));
```
MatiasElo: Just to be on the safe side, I'd check the return value of `odp_event_vector_alloc()` before free.

test/performance/odp_bench_event_vector.c
```diff
@@ -0,0 +1,750 @@
...
+	uint32_t num_evv = gbl_args->appl.num_evv;
+	uint32_t evv_size;
+
+	for (int i = 0; i < TEST_REPEAT_COUNT; i++) {
+		evv_size = odp_event_vector_size(evv_arr[i]);
+
+		if (evv_size != 0 && evv_size != num_evv)
+			odp_event_vector_size_set(evv_arr[i], num_evv);
+
+		odp_event_free(odp_event_vector_to_event(evv_arr[i]));
```
MatiasElo, line 195: Could check that `evv_arr[i]` != `ODP_EVENT_VECTOR_INVALID` before free to detect failures in `event_vector_alloc()` test.

test/performance/odp_bench_event_vector.c
```diff
@@ -0,0 +1,750 @@
...
+}
+
+static int event_vector_tbl(void)
+{
+	odp_event_vector_t *evv_arr = gbl_args->evv_arr;
+	odp_event_t *ptr;
+	int i;
+
+	for (i = 0; i < TEST_REPEAT_COUNT; i++)
+		odp_event_vector_tbl(evv_arr[i], &ptr);
```
MatiasElo: Could sum return values from `odp_event_vector_tbl()` calls and formulate the return value from that based on `--num_events` value. Same comment for `event_vector_size()`.

test/performance/odp_bench_event_vector.c
```diff
@@ -0,0 +1,750 @@
...
+static void print_info(void)
+{
+	odp_sys_info_print();
+
+	printf("\n"
+	       "odp_bench_event_vector options\n"
+	       "------------------------\n");
+
+	printf("Event vector size:       %d\n", gbl_args->appl.max_evv_size);
+	printf("Event vector elements:   %d\n", gbl_args->appl.num_evv);
```
MatiasElo: Option naming here could match better with `usage()` output.

test/performance/odp_bench_event_vector.c
```diff
@@ -0,0 +1,750 @@
...
+	}
+	(void)odp_cpumask_to_str(&default_mask, gbl_args->cpumask_str,
+				 sizeof(gbl_args->cpumask_str));
+
+	if (create_pools()) {
+		ret = -1;
+		goto exit;
+	}
+
+	memset(&worker_thread, 0, sizeof(odph_thread_t));
```
MatiasElo: Seems unnecessary.

test/performance/odp_bench_event_vector.c
```diff
@@ -0,0 +1,750 @@
...
+	thr_common.share_param = 1;
+
+	odph_thread_param_init(&thr_param);
+	thr_param.start = bench_run;
+	thr_param.arg = &gbl_args->suite;
+	thr_param.thr_type = ODP_THREAD_WORKER;
+
+	odph_thread_create(&worker_thread, &thr_common, &thr_param, 1);
+
+	odph_thread_join(&worker_thread, 1);
```
MatiasElo: Return values checks missing.

test/performance/odp_bench_event_vector.c
```diff
@@ -0,0 +1,750 @@
...
+	if (capa.event_vector.max_pools == 0) {
+		ODPH_ERR("Error: event vector pools not supported\n");
+		return -1;
+	}
+
+	if (capa.event_vector.max_num && capa.event_vector.max_num < TEST_REPEAT_COUNT) {
+		ODPH_ERR("Error: pool size not supported (max %" PRIu32 ")\n",
+			 capa.event_vector.max_num);
+		return -1;
+	} else if (gbl_args->appl.cache_size > (int)capa.event_vector.max_cache_size) {
```
MatiasElo: There is also capability for `min_cache_size`. Also, can be a normal `if` (less indentation).

test/performance/odp_bench_event_vector.c
```diff
@@ -0,0 +1,750 @@
...
+		ODPH_ERR("Error: cache size not supported (max %" PRIu32 ")\n",
+			 capa.event_vector.max_cache_size);
+		return -1;
+	}
+
+	if (gbl_args->appl.max_evv_size > TEST_MAX_EVV_SIZE ||
+	    gbl_args->appl.max_evv_size > capa.event_vector.max_size) {
+		ODPH_ERR("Error: event vector size not supported (max %" PRIu32 ")\n",
+			 capa.event_vector.max_size < TEST_MAX_EVV_SIZE ?
+			 capa.event_vector.max_size : TEST_MAX_EVV_SIZE);
```
MatiasElo: `ODPH_MIN()` macro could be used here and below for `evv_uarea_size`.

test/performance/odp_bench_event_vector.c
```diff
@@ -0,0 +1,750 @@
...
+	if (gbl_args->evv_pool == ODP_POOL_INVALID) {
+		ODPH_ERR("Error: pool create failed\n");
+		return -1;
+	}
+
+	odp_pool_print(gbl_args->evv_pool);
+
+	/* Create buffer pool */
+	if (gbl_args->appl.num_evv) {
+		uint32_t buf_size = TEST_BUF_SIZE;
```
MatiasElo: Unnecessary variable.

test/performance/odp_bench_event_vector.c
```diff
@@ -0,0 +1,750 @@
...
+
+		odp_pool_param_init(&buf_pool_params);
+		buf_pool_params.buf.size = buf_size;
+		buf_pool_params.buf.num = TEST_REPEAT_COUNT * gbl_args->appl.num_evv;
```
MatiasElo: Capa check missing for `odp_pool_capability_t.buf.max_num`.

test/performance/odp_bench_event_vector.c
```diff
@@ -0,0 +1,800 @@
...
+ */
+static void free_event_vector_size_set(void)
+{
+	odp_event_vector_t *evv_arr = gbl_args->evv_arr;
+	uint32_t num_events = gbl_args->appl.num_events;
+
+	for (int i = 0; i < TEST_REPEAT_COUNT; i++) {
+		odp_event_vector_size_set(evv_arr[i], num_events);
+		if (evv_arr[i] == ODP_EVENT_VECTOR_INVALID)
+			ODPH_ABORT("Freeing event vector failed\n");
```
MatiasElo: `odp_event_vector_size_set()` call should be after the check.

PR #2234, opened 2025-06-25
[PATCH v2] Speed up tests

JannePeltonen:
Speed up test run by shortening or removing several wait times.


platform/linux-generic/test/validation/api/pktio/pktio_run_tap.sh
MatiasElo, line 1: Minor thing, but I'd reorder the commit subject line: `linux-gen: test: ...`

JannePeltonen: Changed in v2. It appears that both orders were used in the past.

JannePeltonen:
v2: added review tags and rebased

PR #2235, opened 2025-07-04
test: sched_pktio_run: fix possible hang

Rayska:
The script sometimes sent the kill signal to odp_packet_gen before it had completed startup, causing the packet generator to miss the termination signal and run indefinitely. This would cause the script to hang at the 'wait ${GEN_PID}' line. 

The addition of these parameters should prevent such a scenario from happening again.


test/performance/odp_sched_pktio_run.sh
MatiasElo, line 1: In the subject line I'd use: `test: sched_pktio: ...`

Also, you could explain in commit log why the `-q` option was added for future reference.

MatiasElo (approved):

PR #2237, opened 2025-07-07
test: add configurable wait time for pktio links to become active

Rayska:
Adds wait time options to test applications, allowing tests to wait for network interfaces to establish links before proceeding with testing.

Changes include:
- Add `-w/--wait` option to odp_sched_pktio
- Add `-S/--wait` option to odp_l2fwd
- Add `-w/--wait` option to odp_l2fwd_perf

Tests will wait for a configurable period (in seconds) for all interfaces to report link status as UP before proceeding.


test/performance/odp_sched_pktio.c
```diff
@@ -572,6 +574,8 @@ static void print_usage(const char *progname)
 	       "  --pipe-queues <number>    Number of queues per pipeline stage\n"
 	       "  --pipe-queue-size <num>   Number of events a pipeline queue must be able to store. Default 256.\n"
 	       "  -m, --sched_mode <mode>   Scheduler synchronization mode for all queues. 1: parallel, 2: atomic, 3: ordered. Default: 2\n"
+	       "  -w, --wait <sec>          Wait up to <sec> seconds for network links to be up.\n"
```
TuomasTaipale: As discussed, there could be just the longopts version for this and maybe the option name could be `--wait_link`.

In longopts only, you assign a non-ascii integer value (>255) when filling `struct option` and then add a switch case for the value. Best also to have the value as a `#define` or `enum`.

test/performance/odp_sched_pktio.c
```diff
@@ -1047,21 +1054,74 @@ static void link_pktios(test_global_t *test_global)
...
 
 			return -1;
 		}
 
 		test_global->pktio[i].started = 1;
 	}
 
+	/* Wait until all links are up */
+	for (i = 0; test_options->wait_sec && i < num_pktio; i++) {
+		while (1) {
```
TuomasTaipale: You could utilize `link_wait` here, i.e., initialize with `test_options->wait_sec` and then `while (link_wait--)`.

test/performance/odp_sched_pktio.c
```diff
@@ -1047,21 +1054,74 @@ static void link_pktios(test_global_t *test_global)
...
+				}
+				break;
+			}
+			link_wait++;
+			if (link_wait > test_options->wait_sec) {
+				ODPH_ERR("Error (%s): Pktio link down.\n",
+					 test_options->pktio_name[i]);
+				return -1;
+			}
+			odp_time_wait_ns(ODP_TIME_SEC_IN_NS);
```
TuomasTaipale: The link-check-and-wait portion could also be in a separate function. The function could return once wait time expires or the link gets up. Return value could tell if everything was OK or something went wrong. After the function returns and everything is OK, `print_link_info()` could be called and then move to the next packet I/O.

I think we also need to handle the error case of `odp_pktio_link_status()`, `ODP_PKTIO_LINK_STATUS_UNKNOWN` is considered an error.

test/performance/odp_l2fwd_run.sh
```diff
@@ -76,7 +76,7 @@ run_l2fwd()
 	LOG=odp_l2fwd_tmp.log
 
 	# Max 2 workers
-	odp_l2fwd${EXEEXT} -i $IF1,$IF2 -m 0 -t 1 -c 2 | tee $LOG
+	odp_l2fwd${EXEEXT} -i $IF1,$IF2 -m 0 -t 1 -c 2 -w 10 | tee $LOG
```
TuomasTaipale: With `l2fwd`, the wait shortopt was `-S`, though this issue disappears with longopts only.

test/performance/odp_sched_pktio.c
```diff
@@ -1,5 +1,6 @@
 /* SPDX-License-Identifier: BSD-3-Clause
```
JannePeltonen, line 1: The commit messages could be in the grammatic style used in ODP, i.e.: "Implement --wait_link option ...". And maybe "The option value" would be better than "Provided value".

test/performance/odp_sched_pktio.c
```diff
@@ -1047,22 +1058,104 @@ static void link_pktios(test_global_t *test_global)
 	printf("\n");
 }
 
+static int print_link_info(odp_pktio_t pktio)
```
JannePeltonen: This function, as well as the check_link_status_wait() functions are duplicated nearly identically in the three tests apps updated by this PR. It might be better to avoid the duplication and have the same implementations shared by the test apps. It could be that the link waiting is something needed by other test apps later, so this could avoid future duplication as well. I suppose you could introduce something like odp_perf_common.h and put the duplicate code in there.

test/performance/odp_sched_pktio.c
```diff
@@ -1047,22 +1058,104 @@ static void link_pktios(test_global_t *test_global)
...
+				return -1;
+			}
+
+			all_links_up = all_links_up && link_status;
+		}
+
+		if (all_links_up)
+			return 0;
+
+		odp_time_wait_ns(ODP_TIME_SEC_IN_NS);
```
JannePeltonen: One second delays here and there can add up to significant overhead in CI, so I suggest polling the link state every 1/100th of a second or so to minimize potential unnecessary waiting.

test/performance/odp_sched_pktio.c
```diff
@@ -1047,22 +1058,104 @@ static void link_pktios(test_global_t *test_global)
...
+	int link_status;
+
+	while (link_wait--) {
+		all_links_up = 1;
+		for (i = 0; i < num_pktio; i++) {
+			pktio = test_global->pktio[i].pktio;
+			link_status = odp_pktio_link_status(pktio);
+
+			if (link_status == ODP_PKTIO_LINK_STATUS_UNKNOWN) {
+				ODPH_ERR("Error (%s): Link status unknown\n",
```
JannePeltonen: I am not sure unknown link status should be treated as an error. Maybe some pktios report unknown link status before they learn if the link is up or down and start report that. And maybe some pktios always report unknown link status. Maybe it would be better not to treat an unknown link status an error but simply wait until all links are up or the timeout expires. And after the timeout, fail in error only if some links are reported to be down (and not just unknown).

test/performance/common/odp_perf_common.h
```diff
@@ -0,0 +1,91 @@
+/* SPDX-License-Identifier: BSD-3-Clause
```
TuomasTaipale: This file could actually be under `/test/common/` (so no need for the new `common` folder). This should also be declared to `make` machinery somewhere, adding to `noinst_HEADERS` in `/test/common/Makefile.am` should suffice.

You can also then include this similarly to `export_results.h` header.

test/performance/common/odp_perf_common.h
```diff
@@ -0,0 +1,91 @@
...
+extern "C" {
+#endif
+
+/**
+ * Print pktio link info
+ *
+ * @param pktio      Pktio handle
+ * @param name       Pktio name
+ */
+static void print_link_info(odp_pktio_t pktio, const char *name)
```
TuomasTaipale: As this is a common file, the functions could have some common prefix, e.g. `test_common_` like with `export_result.h` functions.

test/performance/common/odp_perf_common.h
```diff
@@ -0,0 +1,91 @@
...
+{
+	odp_pktio_link_info_t info;
+
+	if (odp_pktio_link_info(pktio, &info)) {
+		ODPH_ERR("Error (%s): Printing pktio link not possible:"
+		"Information retrieval failed.\n", name);
```
TuomasTaipale: "... possible:" or "Information ..." could have a trailing/leading whitespace so that actual output doesn't look weird.

test/performance/odp_l2fwd.c
```diff
@@ -1533,6 +1540,59 @@ static int create_pktio(const char *dev, int idx, int num_rx, int num_tx, odp_po
...
+			return -1;
+		}
+	}
+
+	/* Wait for links to be up and print info */
+	for (i = 0; i < if_count; i++) {
+		pktio = gbl_args->pktios[i].pktio;
+		name = appl->if_names[i];
+
+		if (wait_sec && check_link_status_wait(pktio, name, wait_sec))
```
TuomasTaipale: `check_link_status_wait()` documents that `-1` is the error case, so this could explicitly test that particular value.

test/performance/common/odp_perf_common.h
```diff
@@ -0,0 +1,91 @@
...
+	       (info.duplex == ODP_PKTIO_LINK_DUPLEX_HALF ? "half" :
+	       (info.duplex == ODP_PKTIO_LINK_DUPLEX_FULL ? "full" : "unknown")));
+	printf("  media       %s\n", info.media);
+	printf("  pause_rx    %s\n",
+	       (info.pause_rx == ODP_PKTIO_LINK_PAUSE_ON ? "on" :
+	       (info.pause_rx == ODP_PKTIO_LINK_PAUSE_OFF ? "off" : "unknown")));
+	printf("  pause_tx    %s\n",
+	       (info.pause_tx == ODP_PKTIO_LINK_PAUSE_ON ? "on" :
+	       (info.pause_tx == ODP_PKTIO_LINK_PAUSE_OFF ? "off" : "unknown")));
+	printf("  speed(Mbit/s) %" PRIu32 "\n\n", info.speed);
```
TuomasTaipale: The printed speed value is slightly off of the rest of column values?

TuomasTaipale (approved):

test/common/Makefile.am
```diff
@@ -28,6 +28,7 @@ libthrmask_common_la_CFLAGS = $(AM_CFLAGS) -DTEST_THRMASK
 endif
 
 noinst_HEADERS = test_common_macros.h \
+		 test_common_perf.h \
```
JannePeltonen: I am not sure test_common_perf.h is a good name for this file since the functions are not perf test specific and (since this header is here) could be used e.g. by validation tests too. Maybe pktio_common.h?

The commit message also talks about "common performance measurement functions" although it is not really what is inside this header file. And the commit message mentions perf_common.h which does not match name of this header file.

test/common/test_common_perf.h
```diff
@@ -0,0 +1,93 @@
+/* SPDX-License-Identifier: BSD-3-Clause
+ * Copyright (c) 2025 Nokia
```
JannePeltonen: I think the copyright should be 2020-2025, since the link info printing code was just copied from odp_packet_gen.c, to which it was added in 2020.

test/common/test_common_perf.h
```diff
@@ -0,0 +1,93 @@
...
+	       (info.pause_rx == ODP_PKTIO_LINK_PAUSE_ON ? "on" :
+	       (info.pause_rx == ODP_PKTIO_LINK_PAUSE_OFF ? "off" : "unknown")));
+	printf("  pause_tx      %s\n",
+	       (info.pause_tx == ODP_PKTIO_LINK_PAUSE_ON ? "on" :
+	       (info.pause_tx == ODP_PKTIO_LINK_PAUSE_OFF ? "off" : "unknown")));
+	printf("  speed(Mbit/s) %" PRIu32 "\n\n", info.speed);
+}
+
+/**
+ * Wait for pktio links to be up and return status
```
JannePeltonen: This says links (plural) even though the function waits for single link only. Maybe it would be nice to have the function take a number of pktio's to wait, with single shared timeout (i.e. if first link is waited for the maximum time, there is no need to wait even more for the others after that).

test/performance/odp_packet_gen.c
```diff
@@ -1662,39 +1663,14 @@ static int open_pktios(test_global_t *global)
 	return 0;
 }
 
-static int print_link_info(odp_pktio_t pktio)
```
JannePeltonen, line 1665: The commit messae says that the old implementation is replaces with the one in test_common_perf.h. At least for the print_link_info() is is not really replaced but just moved (by the two commits), so I wonder if things would be clearer to have a single commit (by squashing the two relevant commits or otherwise) that does the move.

test/performance/odp_packet_gen.c
```diff
@@ -1707,27 +1683,14 @@ static int start_pktios(test_global_t *global)
...
-					 test_options->pktio_name[i]);
-				return -1;
-			}
-			odp_time_wait_ns(ODP_TIME_SEC_IN_NS);
-		}
+	for (i = 0; i < num_pktio; i++) {
+		pktio = global->pktio[i].pktio;
+		name = test_options->pktio_name[i];
+
+		if (wait_sec && test_common_check_link_status_wait(pktio, name, wait_sec) == -1)
```
JannePeltonen: The usage message says that -w waits "up to <sec> seconds for network links to be up", but now the behaviour changes that each link is waited, consecutively, up to <sec> second for the link to be up (IOW, the wait time is up to <sec> * <num interfaces>). To fix the discrepancy, either the usage text or the implementation should be changes. I think the old behaviour was a bit nicer.

test/performance/odp_sched_pktio.c
```diff
@@ -1,5 +1,6 @@
 /* SPDX-License-Identifier: BSD-3-Clause
  * Copyright (c) 2018 Linaro Limited
+ * Copyright (c) 2025 Nokia
```
JannePeltonen, line 3: The commit message says: "The option value is the number of seconds to wait for all links to be up". I would just remove the sentence because it is inaccurate (the value is the max number of seconds to waut for each link consecutively), and a detail not really needed in the message. And the same comment for the subsequent commit.

test/common/pktio_common.h
```diff
@@ -0,0 +1,124 @@
...
+					      const char *pktio_names[],
+					      uint32_t num_pktio,
+					      uint32_t wait_sec)
+{
+	odp_pktio_t pktio;
+	const char *name;
+	int link_status;
+	uint32_t link_wait = 0; /**< Centiseconds waited */
+
+	for (uint32_t i = 0; wait_sec && i < num_pktio; i++) {
```
JannePeltonen: Now zero wait_sec makes this function a no-op. I think it would be nicer to have the function behave the same way, apart from the wait duration, regardless of whether wait_sec is zero or not, IOW, check the link states even with zero wait time. The call sites can just not call this function if they do not want anything to be done (which might be the case, for retaining existing behaviour if nothing else, in some  of the perf tests that are going to use this.

test/common/pktio_common.h
```diff
@@ -0,0 +1,124 @@
...
+			if (link_status == ODP_PKTIO_LINK_STATUS_UP)
+				break;
+
+			link_wait++;
+
+			if (link_wait > wait_sec * 100) {
+				if (link_status == ODP_PKTIO_LINK_STATUS_UNKNOWN) {
+					printf("Warning (%s): Link status unknown after wait time.\n",
+					       name);
+					return 0;
```
JannePeltonen: This return results in skipping the rest of the pktios in the list. Some of those pktios may have link down in which case this function should return an error. So there should be a break here instead of a return.

test/common/pktio_common.h
```diff
@@ -0,0 +1,124 @@
...
+				}
+
+				ODPH_ERR("Error: (%s): Pktio link down after wait time.\n",
+					 name);
+				return -1;
```
JannePeltonen: Maybe it would be better to print the error for all links that are still down, by using a break here instread of a return (and e.g. by setting ret = -1 and returning ret at the end)?

test/performance/odp_sched_pktio.c
```diff
@@ -1049,17 +1062,30 @@ static void link_pktios(test_global_t *test_global)
...
+		pktios[i] = test_global->pktio[i].pktio;
+		names[i] = test_options->pktio_name[i];
+	}
+
+	/* Wait until all links are up and print link infos */
+	if (wait_sec) {
+		if (pktio_common_check_link_status_wait(pktios, names, num_pktio, wait_sec) == -1)
+			return -1;
+
+		pktio_common_print_link_info_multi(pktios, names, num_pktio);
```
JannePeltonen: Perhaps this info printing could be done even when wait_sec is zero (and same in other commits)? The commit message could mention that link info printing is also being added.

test/performance/odp_l2fwd_perf.c
```diff
@@ -617,7 +635,7 @@ static odp_bool_t setup_pktios(prog_config_t *config)
 		return false;
 	}
 
-	for (uint8_t i = 0U; i < config->num_ifs; ++i) {
+	for (i = 0U; i < num_ifs; ++i) {
```
JannePeltonen: This change seems unnecessary. Moving the declaration of i from here to function scope should not be done as it is arguably better to declare such loop variables in the for statement. Also, I am not sure that the introduction of num_ifs (and wait_sec) local variable is very useful as it does not seem to make the code any clearer.

PR #2238, opened 2025-07-08
api: ml: add support for engine id

syalavarthi:
In devices with multiple ML engines, support should be enabled to get the number of engines supported by device and options to select the engine ID to be used while configuring and loading a model. This patch adds new API to get the number of engines supported by the platform.

Added new API odp_ml_engine_capability to get engine specific capabilities. Introduced new macro ODP_ML_ENGINE_ANY to be used the APIs for ODP implementation to choose one of the available engine. Enabled support to configure a specific ML engine and map a model created to an engine.

Updated linux-generic implementation, sysinfo example app and unit tests to support engine ID.


include/odp/api/spec/ml.h
```diff
@@ -59,12 +59,26 @@ extern "C" {
...
+int odp_ml_num_engines(void);
+
 /**
  * Query ML capabilities
  *
  * Outputs ML capabilities on success. Use this capability call to check ML offload implementation
  * limits and its support of various ML API features. When ML offload is not available,
- * odp_ml_capability_t.max_models is zero.
+ * odp_ml_capability_t.max_models is zero. This function would fetch the capabilities of the default
+ * ML engine. If the system has multiple ML engines, use odp_ml_eng_capability() with engine_id set
```
TuomasTaipale: Typo `odp_ml_eng_capability()` is `odp_ml_engine_capability()`.

include/odp/api/spec/ml_types.h
```diff
@@ -672,6 +692,15 @@ typedef struct odp_ml_data_format_t {
  * Use odp_ml_model_param_init() to initialize the structure to its default values.
  */
 typedef struct odp_ml_model_param_t {
+	/**
+	 * Engine ID
+	 *
+	 * Engine ID to be used with the model. Values in the range 1..odp_ml_num_engines()
+	 * select a specific engine and the special value of ODP_ML_ENGINE_ANY means that ODP
+	 * selects the engine to be used.
+	 */
```
TuomasTaipale, line 703: Like in `odp_ml_config_t` documentation, default value `ODP_ML_ENGINE_ANY` could be documented here as well.

include/odp/api/spec/ml.h
```diff
@@ -86,9 +118,9 @@ void odp_ml_config_init(odp_ml_config_t *config);
  * Configure ML offload
  *
  * Initializes and configures ML offload according to the configuration parameters. This function
- * must be called only once and before any ML resources are created. Use odp_ml_capability() to
- * query configuration capabilities and odp_ml_config_init() to initialize configuration
- * parameters into their default values.
+ * must be called only once per engine and before any ML resources are created for the engine. Use
+ * odp_ml_capability() to query configuration capabilities of the engine and odp_ml_config_init()
```
TuomasTaipale: This could be "... Use odp_ml_capability() or odp_ml_engine_capability() to query ..."

platform/linux-generic/odp_ml.c
```diff
@@ -203,6 +226,12 @@ int odp_ml_config(const odp_ml_config_t *config)
 		return -1;
 	}
 
+	if (config->engine_id >= ML_MAX_ENGINES) {
```
TuomasTaipale: Range is 1..odp_ml_num_engines() so the check should be `config->engine_id > ML_MAX_ENGINES`, right?

platform/linux-generic/odp_ml.c
```diff
@@ -835,6 +867,12 @@ odp_ml_model_t odp_ml_model_create(const char *name, const odp_ml_model_param_t
 		return ODP_ML_MODEL_INVALID;
 	}
 
+	if (odp_unlikely(param->engine_id >= ML_MAX_ENGINES)) {
```
TuomasTaipale: Same here, check should be `config->engine_id > ML_MAX_ENGINES`?

test/validation/api/ml/ml.c
```diff
@@ -627,7 +689,9 @@ static void test_ml_fp32_fp16(void)
 }
 
 odp_testinfo_t ml_suite[] = {
+	ODP_TEST_INFO(test_ml_engine_count),
 	ODP_TEST_INFO(test_ml_capability),
+	ODP_TEST_INFO(test_ml_engine_capability),
 	ODP_TEST_INFO_CONDITIONAL(test_ml_param_init, check_ml_support),
```
TuomasTaipale, line 697: Now that engine ID is documented to have default values in `odp_ml_config_t` and `odp_ml_model_param_t`, we should check the default values in test `test_ml_param`.

include/odp/api/spec/ml_types.h
```diff
@@ -70,6 +70,11 @@ extern "C" {
  * ODP implementation specific extra parameters for model creation
  */
 
+/* The special engine ID ODP_ML_ENGINE_ANY can be used to permit ODP
+ * to decide on the engine to be used to create and load a model.
+ */
+#define ODP_ML_ENGINE_ANY 0
```
TuomasTaipale: Doxygen doesn't seem to understand this as a Doxygen comment, I think the block needs to start with a `/**`, so forward slash and two asterisks. E.g.:
```
/**
 * The special engine ID ODP_ML_ENGINE_ANY can be used to permit ODP
 * to decide on the engine to be used to create and load a model.
 */
#define ODP_ML_ENGINE_ANY 0
```

include/odp/api/spec/ml.h
syalavarthi: Updated in v2

include/odp/api/spec/ml_types.h
syalavarthi: Updated in v2

include/odp/api/spec/ml.h
syalavarthi: Updated in v2

include/odp/api/spec/ml_types.h
syalavarthi: Fixed in v2

test/validation/api/ml/ml.c
syalavarthi: Will fix in v3

platform/linux-generic/odp_ml.c
syalavarthi: Will fix in v3

syalavarthi: fixed

test/validation/api/ml/ml.c
syalavarthi: fixed

include/odp/api/spec/ml_types.h
TuomasTaipale: Here I meant that this line should be added:

"The default value is ODP_ML_ENGINE_ANY."

as I guess `odp_ml_model_param_init()` should set this to ODP_ML_ENGINE_ANY by default.

include/odp/api/spec/ml_types.h
```diff
@@ -672,6 +692,16 @@ typedef struct odp_ml_data_format_t {
  * Use odp_ml_model_param_init() to initialize the structure to its default values.
  */
 typedef struct odp_ml_model_param_t {
+	/**
+	 * Engine ID
+	 *
+	 * Engine ID to be used with the model. In a system with multiple ML engines,
+	 * this parameter selects the engine to be configured. Values in the range
+	 * 1..odp_ml_num_engines() select a specific engine and the special value of
+	 * ODP_ML_ENGINE_ANY means that ODP selects the engine to be used.
```
TuomasTaipale: Here I meant that this line should be added:

"The default value is ODP_ML_ENGINE_ANY."

as I guess odp_ml_model_param_init() should set this to ODP_ML_ENGINE_ANY by default.

syalavarthi: Updated the comment and added a check in unit test.

TuomasTaipale (approved):

PR #2241, opened 2025-08-07
[PATCH v3] Fix misc Coverity warnings

MatiasElo:


test/performance/odp_queue_perf.c
```diff
@@ -585,8 +587,8 @@ static int run_test(void *arg)
 	const uint32_t num_workers = thr_args->options->num_cpu;
 	const uint32_t max_burst = thr_args->options->max_burst;
 	const odp_bool_t single_event = max_burst == 0 ? 1 : 0;
-	odp_queue_t src_queue_tbl[MAX_QUEUES];
-	odp_queue_t dst_queue_tbl[MAX_QUEUES];
+	odp_queue_t *src_queue_tbl = global->src_queue_tbl;
```
JannePeltonen: The queue tables are thread-specific and get initialized from the queue table in thread parameters. This change makes the threads step on each other's toes when each of them try to write their private queue table in the global array.

MatiasElo: Aah, how did I miss that...

MatiasElo: Fixed in V2.

PR #2242, opened 2025-08-07
test: sched_latency: add enqueue/schedule burst support

Rayska:
Added a burst mode -b/--burst <num> where event are enqueued and scheduled in bursts with odp_queue_enq_multi() and odp_schedule_multi().

### DIFF
--- a/test/performance/odp_sched_latency.c
+++ b/test/performance/odp_sched_latency.c
@@ -42,6 +42,10 @@
 #define EVENT_FORWARD_INC  1
 #define EVENT_FORWARD_NONE 2
 
+/* Test enqueue and schedule mode */
+#define SINGLE_MODE 0
+#define BURST_MODE  1
+
 /** Test event types */
 typedef enum {
 	WARM_UP,  /**< Warm-up event */
@@ -52,12 +56,20 @@ typedef enum {
 
 /** Test event */
 typedef struct {
-	odp_time_t time_stamp;	/**< Send timestamp */
 	event_type_t type;	/**< Message type */
 	int src_idx[NUM_PRIOS]; /**< Source ODP queue */
 	int prio;		/**< Source queue priority */
 	int warm_up_rounds;	/**< Number of completed warm-up rounds */
 } test_event_t;
+
+/** Test variables for event/burst tracking */
+typedef struct {
+	odp_time_t	 enq_time;		/**< Timestamp of enqueue */
+	odp_atomic_u32_t burst_received;	/**< Number of samples scheduled from burst */
+	odp_atomic_u32_t collector_idx;		/**< Number of samples ready for enqueueing */
+	odp_atomic_u32_t writes_completed;	/**< Writes completed to sample_collector */
+	odp_event_t	 sample_collector[64];	/**< Array for burst enqueueing samples */
+} test_track_t;
 
 /** Test arguments */
 typedef struct {
@@ -68,6 +80,7 @@ typedef struct {
 	int isolate;
 	int test_rounds;	/**< Number of test rounds (millions) */
 	int warm_up_rounds;	/**< Number of warm-up rounds */
+	int burst_size;		/**< Number of events to enqueue and schedule in a burst */
 	struct {
 		int queues;	/**< Number of scheduling queues */
 		int events;	/**< Number of events */
@@ -87,6 +100,7 @@ typedef struct {
 	uint64_t min;	   /**< Minimum event latency */
 	uint64_t max;	   /**< Maximum event latency */
 	uint64_t max_idx;  /**< Index of the maximum latency sample event */
+	uint64_t bursts;   /**< Total number of received bursts */
 } test_stat_t;
 
 /** Performance test statistics (per core) */
@@ -103,6 +117,7 @@ typedef struct {
 	test_args_t      args;	  /**< Parsed command line arguments */
 	odp_queue_t      queue[NUM_PRIOS][MAX_QUEUES]; /**< Scheduled queues */
 	test_common_options_t common_options;	/**< Result export options */
+	test_track_t     track; /**< Variables for tracking test progress */
 
 	odp_schedule_group_t group[NUM_PRIOS][MAX_GROUPS];
 
@@ -315,7 +330,7 @@ static int output_results(test_globals_t *globals)
 		for (j = 1; j <= args->cpu_count; j++) {
 			lat = &globals->core_stat[j].prio[i];
 
-			if (lat->sample_events == 0) {
+			if (lat->sample_events == 0 || (args->burst_size && lat->bursts == 0)) {
 				printf("%-8d N/A\n", j);
 				continue;
 			}
@@ -327,8 +342,12 @@ static int output_results(test_globals_t *globals)
 			total.tot += lat->tot;
 			total.sample_events += lat->sample_events;
 			total.events += lat->events;
+			total.bursts += lat->bursts;
 
-			avg = lat->events ? lat->tot / lat->sample_events : 0;
+			if (globals->args.burst_size)
+				avg = lat->bursts ? lat->tot / lat->bursts : 0;
+			else
+				avg = lat->events ? lat->tot / lat->sample_events : 0;
 			printf("%-8d %-10" PRIu64 " %-10" PRIu64 " "
 			       "%-10" PRIu64 " %-10" PRIu64 " %-10" PRIu64 " %-10" PRIu64 "\n",
 			       j, avg, lat->min, lat->max, lat->sample_events,
@@ -339,7 +358,11 @@ static int output_results(test_globals_t *globals)
 			printf("Total    N/A\n\n");
 			continue;
 		}
-		avg = total.events ? total.tot / total.sample_events : 0;
+
+		if (globals->args.burst_size)
+			avg = total.bursts ? total.tot / total.bursts : 0;
+		else
+			avg = total.events ? total.tot / total.sample_events : 0;
 		printf("Total    %-10" PRIu64 " %-10" PRIu64 " %-10" PRIu64 " "
 		       "%-10" PRIu64 " %-10" PRIu64 "\n\n", avg, total.min,
 		       total.max, total.sample_events, total.events);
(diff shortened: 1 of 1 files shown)


test/performance/odp_sched_latency.c
```diff
@@ -392,66 +417,66 @@ static int join_groups(test_globals_t *globals, int thr)
...
- *
- * @retval 0 on success
- * @retval -1 on failure
- */
-static int test_schedule(int thr, test_globals_t *globals)
+static inline void record_stats(test_stat_t *stats, int mode, odp_time_t time,
+				odp_time_t enq_time)
 {
-	odp_time_t time;
+	uint64_t latency = odp_time_to_ns(time) - odp_time_to_ns(enq_time);
```
MatiasElo: In general it's better to minimize the usage of `odp_time_*_ns()`functions to avoid unnecessary conversations See `bench_tm_func_record()` for example.

test/performance/odp_sched_latency.c
```diff
@@ -601,7 +823,7 @@ static void usage(void)
 	       "				number of events per queue, does not include sample event.\n"
 	       "  -r  --sample-per-prio Allocate a separate sample event for each priority. By default\n"
 	       "			a single sample event is used and its priority is changed after\n"
-	       "			each processing round.\n"
+	       "			each processing round. No effect in burst mode.\n"
```
MatiasElo: Wouldn't the current testing method work also for multiple priorities?

test/performance/odp_sched_latency.c
```diff
@@ -487,6 +507,207 @@ static int test_schedule(int thr, test_globals_t *globals)
...
+	odp_atomic_u32_t *collector_idx = &globals->track.collector_idx;
+	odp_atomic_u32_t *ready = globals->track.ready;
+	odp_event_t *sample_collector = globals->track.sample_collector;
+	uint32_t idx;
+
+	while (1) {
+		idx = odp_atomic_fetch_inc_u32(collector_idx);
+		if (idx < burst_size) {
+			sample_collector[idx] = *ev;
+			odp_atomic_store_rel_u32(&ready[idx], 1);
```
MatiasElo: A single shared `ready` variable and `odp_atomic_add_rel_u32()` could be simpler. Then below the value would be compared to burst size.

test/performance/odp_sched_latency.c
```diff
@@ -487,6 +507,207 @@ static int test_schedule(int thr, test_globals_t *globals)
...
+static inline int collect_enq_sample(int thr, test_globals_t *globals, odp_event_t *ev)
+{
+	const uint32_t burst_size = globals->args.burst_size;
+	odp_atomic_u32_t *burst_received = &globals->track.burst_received;
+	odp_atomic_u32_t *collector_idx = &globals->track.collector_idx;
+	odp_atomic_u32_t *ready = globals->track.ready;
+	odp_event_t *sample_collector = globals->track.sample_collector;
+	uint32_t idx;
+
+	while (1) {
```
MatiasElo: The outer while loop seems unnecessary as `collector_idx` should never be >= `burst_size`.

test/performance/odp_sched_latency.c
```diff
@@ -487,6 +507,207 @@ static int test_schedule(int thr, test_globals_t *globals)
...
+						odp_cpu_pause();
+				}
+
+				if (enq_burst(thr, globals, sample_collector, burst_size))
+					return -1;
+
+				odp_atomic_store_u32(burst_received, 0);
+				for (uint32_t k = 0; k < burst_size; k++)
+					odp_atomic_store_rel_u32(&ready[k], 0);
+				odp_atomic_store_rel_u32(collector_idx, 0);
```
MatiasElo: Should likely be done before enqueueing new burst.

MatiasElo (commented):
Some initial comments.

test/performance/odp_sched_latency.c
```diff
@@ -487,6 +520,234 @@ static int test_schedule(int thr, test_globals_t *globals)
...
+			} else if (event->type == SAMPLE) {
+				stats->sample_events++;
+				stats->events++;
+			} else {
+				stats->events++;
+			}
+
+			/* Collect events for burst enqueueing */
+			if (event->type == SAMPLE) {
+				if (collect_enq_sample(thr, globals, &ev, time, event->prio))
```
MatiasElo: Could use `odp_unlikely()`.

test/performance/odp_sched_latency.c
```diff
@@ -487,6 +520,234 @@ static int test_schedule(int thr, test_globals_t *globals)
...
+					return -1;
+			} else {
+				traffic_collector[traffic_idx++] = ev;
+			}
+		}
+
+		if (traffic_idx) {
+			if (enq_burst(thr, globals, traffic_collector, traffic_idx, LO_PRIO) == -1)
```
MatiasElo: `odp_unlikely()` also here. Also, perhaps the priority parameter could be dropped from `enq_burst()`. At least here it looks a bit confusing.

test/performance/odp_sched_latency.c
```diff
@@ -487,6 +520,234 @@ static int test_schedule(int thr, test_globals_t *globals)
...
+			event = odp_buffer_addr(odp_buffer_from_event(burst_array[i]));
+			event->prio = new_prio;
+			event->src_idx[event->prio] = dst_idx;
+		}
+
+		while (odp_atomic_load_acq_u32(&track[event->prio].ts_published))
+			odp_cpu_pause();
+
+		track[event->prio].enq_time = odp_time_global_strict();
+		odp_atomic_store_rel_u32(&track[event->prio].ts_published, 1);
```
MatiasElo: `odp_queue_enq_multi()` already includes necessary memory barriers, so this seems unnecessary.

test/performance/odp_sched_latency.c
```diff
@@ -392,66 +420,71 @@ static int join_groups(test_globals_t *globals, int thr)
...
 		if (event->type == SAMPLE) {
-			latency = odp_time_to_ns(time) - odp_time_to_ns(event->time_stamp);
-
-			if (latency > stats->max) {
-				stats->max = latency;
-				stats->max_idx = stats->sample_events;
+			if (odp_atomic_load_acq_u32(&track[event->prio].ts_published)) {
+				record_stats(stats, SINGLE_MODE, time, track[event->prio].enq_time);
+				odp_atomic_store_rel_u32(&track[event->prio].ts_published, 0);
 			}
```
MatiasElo: `ts_published` seems needless as `odp_schedule()` and `odp_queue_enq()` already include barriers. Also, I'd still store the timestamp in buffer data in single event mode for results history.

test/performance/odp_sched_latency.c
```diff
@@ -1095,6 +1375,12 @@ int main(int argc, char *argv[])
 
 	odp_barrier_init(&globals->barrier, num_workers);
 
+	for (i = 0; i < NUM_PRIOS; i++) {
+		odp_atomic_init_u32(&globals->track[i].ready, 0);
+		odp_atomic_init_u32(&globals->track[i].collector_idx, 0);
+		odp_atomic_init_u32(&globals->track[i].ts_published, 0);
+	}
```
MatiasElo: Minor thing, but there could be a single function for initializing global data (e.g. `init_global_data()`), which would be called right after `globals = odp_shm_addr(shm);`.

test/performance/odp_sched_latency.c
```diff
@@ -83,10 +98,11 @@ typedef struct {
...
 	uint64_t sample_events;  /**< Number of received sample events */
-	uint64_t tot;	   /**< Total event latency. Sum of all events. */
-	uint64_t min;	   /**< Minimum event latency */
-	uint64_t max;	   /**< Maximum event latency */
-	uint64_t max_idx;  /**< Index of the maximum latency sample event */
+	odp_time_t tot;    /**< Total event latency. Sum of all events. */
+	odp_time_t min;    /**< Minimum event latency */
+	odp_time_t max;    /**< Maximum event latency */
+	uint64_t max_idx;  /**< Index of the maximum latency sample event/burst */
+	uint64_t bursts;   /**< Total number of received bursts */
```
MatiasElo: Maybe a common variable could be used with single event mode? For example `sample_events` -> `samples`. This could make the code simpler (e.g. less if-else clauses).

test/performance/odp_sched_latency.c
```diff
@@ -756,8 +1024,20 @@ static void parse_args(int argc, char *argv[], test_args_t *args)
 		exit(EXIT_FAILURE);
 	}
 
-	if (args->prio[HI_PRIO].queues == 0 || args->sample_per_prio)
+	if (args->prio[HI_PRIO].queues == 0 || (args->sample_per_prio && !args->burst_size))
```
MatiasElo: Seems unnecessary.

test/performance/odp_sched_latency.c
```diff
@@ -578,6 +841,7 @@ static void usage(void)
 	       "Usage: ./odp_sched_latency [options]\n"
 	       "Optional OPTIONS:\n"
 	       "  -c, --count <number> CPU count, 0=all available, default=1\n"
+	       "  -b, --burst <num> Test burst size.\n"
```
MatiasElo: Could mention that latency is measured over whole bursts. The same could also be mentioned in commit message.

test/performance/odp_sched_latency.c
```diff
@@ -508,6 +769,8 @@ static int test_schedule(int thr, test_globals_t *globals)
 	odp_barrier_wait(&globals->barrier);
 
 	if (thr == MAIN_THREAD) {
+		if (globals->args.burst_size)
```
MatiasElo: The barrier above synchronizes memory operations. This function could likely simply free all partial burst events.

test/performance/odp_sched_latency.c
```diff
@@ -487,6 +520,234 @@ static int test_schedule(int thr, test_globals_t *globals)
...
+}
+
+static int test_schedule_burst(int thr, test_globals_t *globals)
+{
+	const int warm_up_rounds = globals->args.warm_up_rounds;
+	const uint64_t test_rounds = globals->args.test_rounds * (uint64_t)1000000;
+	const uint32_t burst_size = globals->args.burst_size;
+	uint32_t num_sched_ev;
+	odp_event_t ev_burst[burst_size];
+	odp_event_t traffic_collector[burst_size];
```
MatiasElo: MAX_BURST_SIZE can be used to avoid variable length arrays.

test/performance/odp_sched_latency.c
```diff
@@ -487,6 +520,234 @@ static int test_schedule(int thr, test_globals_t *globals)
...
+
+		while (odp_atomic_load_acq_u32(ready) != burst_size)
+			odp_cpu_pause();
+
+		/* Snapshot the burst to avoid data race with scheduling threads */
+		for (i = 0; i < burst_size; i++)
+			local_collector[i] = sample_collector[i];
+
+		odp_atomic_store_u32(collector_idx, 0);
+		odp_atomic_store_rel_u32(ready, 0);
```
MatiasElo: Can be relaxed due to following enqueue call.

Rayska:
V4:
 - Single mode reverted back to saving timestamps in event payload to preserve behaviour
 - Redundand atomic variable removed (ts_published) and some atomic operations removed/relaxed
 - Stats are not recorded if enqueue time has not been set (first 1-2 bursts in burst mode depending on -r flag)
 - Single and burst modes use a common "samples" variable for statistics instead of separate variables
 - Local arrays are no longer VLAs, use MAX_BURST_SIZE instead
 - Leftover events waiting to be enqueued are freed at the end of test_schedule()
 - Added init_global_data()
 - Comments and prints modified/added to better represent both modes

test/performance/odp_sched_latency.c
```diff
@@ -292,7 +303,10 @@ static int output_results(test_globals_t *globals)
 	else
 		printf("  HI_PRIO events: %i\n", args->prio[HI_PRIO].events);
 
-	printf("  HI_PRIO sample events: %i\n\n", args->prio[HI_PRIO].sample_events);
+	printf("  HI_PRIO sample events: %i\n", args->prio[HI_PRIO].samples);
+	printf("  Scheduling mode: %s,%i\n\n",
+	       args->burst_size ? "burst" : "single",
+	       args->burst_size ? args->burst_size : 1);
```
MatiasElo: Nitpicking but the output is a bit messy. Could for example print burst size on a separate line only when burst mode is enabled. Also, could be printed after 'Forwarding mode' line.

test/performance/odp_sched_latency.c
```diff
@@ -487,6 +496,206 @@ static int test_schedule(int thr, test_globals_t *globals)
...
+ * @retval 0 on success
+ * @retval -1 on failure
+ */
+static int test_schedule(int thr, test_globals_t *globals)
+{
+	odp_event_t ev;
+	test_track_t *track;
+	uint32_t cnt;
+
+	memset(&globals->core_stat[thr], 0, sizeof(core_stat_t));
```
MatiasElo: Already initialized to zero by `init_global_data()`.

test/performance/odp_sched_latency.c
```diff
@@ -487,6 +496,206 @@ static int test_schedule(int thr, test_globals_t *globals)
...
+	globals->core_stat[thr].prio[HI_PRIO].min = odp_time_global_from_ns(ODP_TIME_HOUR_IN_NS);
+	globals->core_stat[thr].prio[LO_PRIO].min = odp_time_global_from_ns(ODP_TIME_HOUR_IN_NS);
```
MatiasElo, line 691: The `tot` and `max` values should also be initialized. This can be done with `ODP_TIME_NULL`.

test/performance/odp_sched_latency.c
```diff
@@ -306,7 +320,7 @@ static int output_results(test_globals_t *globals)
 
 	for (i = 0; i < NUM_PRIOS; i++) {
 		memset(&total, 0, sizeof(test_stat_t));
-		total.min = UINT64_MAX;
+		total.min = odp_time_global_from_ns(ODP_TIME_HOUR_IN_NS);
```
MatiasElo, line 324: `total.max` and `total.tot` should also be initialized.

test/performance/odp_sched_latency.c
```diff
@@ -487,6 +496,206 @@ static int test_schedule(int thr, test_globals_t *globals)
...
+			odp_event_free_multi(&burst_array[enq], burst_size - enq);
+			return -1;
+		}
+		enq += num_enq;
+	} while (enq < (int)burst_size);
+
+	return 0;
+}
+
+static inline int collect_enq_sample(int thr, test_globals_t *globals, odp_event_t *ev,
```
MatiasElo: `ev` doesn't have to be a pointer.

MatiasElo (approved):

Issue #2243, opened 2025-08-07
Tests failing on NixOS due to bad bash path


christomel:
When running `make check` after `./bootstrap` and `./configure` many tests fail to run on NixOS. 

Example failure log excerpt:
```
FAIL: check-globals.sh
======================

../../test-driver: ./check-globals.sh: /bin/bash: bad interpreter: No such file or directory
FAIL check-globals.sh (exit status: 126)
```

This issue is caused by the scripts wrongly assuming that `bash` is at `/bin/bash`. This is not the case for all unix operating systems. The solution for that is to use `/usr/bin/env bash` instead of hardcoding the path, because the former finds the `bash` executable in user's `PATH`, wherever it may be.

Steps to reproduce on NixOS with flakes enabled:
1. `git clone https://github.com/OpenDataPlane/odp.git && cd odp`
2. `nix flake init`
3. replace contents of `./flake.nix` with the following
```
{
  description = "A very basic flake";

  inputs = {
    nixpkgs.url = "github:nixos/nixpkgs?ref=nixos-unstable";
  };

  outputs =
    { self, nixpkgs }:
    let
      system = "x86_64-linux";
      pkgs = import nixpkgs {
        inherit system;
      };
    in
    {
      devShells.${system} = {
        default = pkgs.mkShell {
          packages = with pkgs; [
            libtool
            automake
            autoconf
            asciidoctor
            mscgen
            cunit
            graphviz
            valgrind
            libcli
            pkg-config
            libconfig
            libpcap
            openssl
            doxygen
            onnxruntime
            binutils
            gcc
          ];

          NIX_ENFORCE_NO_NATIVE = 0;
        };
      };

    };
}
```
4. Enter nix shell using `nix develop`, then
```
./bootstrap
./configure
make check
```
5. You will get the said error

PR #2244, opened 2025-08-07
[PATCH v3] scripts: make bash scripts more cross-platform

christomel:
Before this change scripts wrongly assume that `bash` resides at `/bin/bash`. This is not always the case and in fact is not mandated by POSIX. One example of a unix which does not have bash in `/bin/bash` is NixOS.

A workaround for this is to use `/usr/bin/env bash`, which is the standard way to get bash from user environment and works on a wider range of systems than `/bin/bash`

Therefore, this change replaces every `/bin/bash` with `/usr/bin/env bash`. Wherever `/bin/bash` took extra arguments, those arguments were moved out of line, because `env` interprets arguments as part of a program name.

Solves #2243

### DIFF
--- a/doc/implementers-guide/implementers-guide.adoc
+++ b/doc/implementers-guide/implementers-guide.adoc
@@ -387,7 +387,7 @@ Here follows a dummy example of what wrapper-script could be:
 
 [source,bash]
 ----
-#!/bin/bash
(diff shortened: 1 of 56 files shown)


TuomasTaipale:
Closed due to inactivity.

PR #2249, opened 2025-08-20
[PATCH v2] test: timer_accuracy: cancel timers and drain timeouts before teardown

Rayska:
Pending timeouts left in scheduler kept queues non-empty in non-periodic modes, causing queue_destroy() failure.

Fix by canceling remaining timers and draining pending timeout events before destroying queues.

### DIFF
--- a/test/performance/odp_timer_accuracy.c
+++ b/test/performance/odp_timer_accuracy.c
@@ -1274,7 +1274,7 @@ static int run_test(void *arg)
 	uint64_t tmo_ns;
 	timer_ctx_t *ctx;
 	odp_thrmask_t mask;
-	uint64_t wait = odp_schedule_wait_time(10 * ODP_TIME_MSEC_IN_NS);
+	const uint64_t wait = odp_schedule_wait_time(4 * test_global->opt.period_ns);
 	odp_schedule_group_t group = ODP_SCHED_GROUP_INVALID;
 	test_log_t *log = test_global->log;
 	enum mode_e mode = test_global->opt.mode;


test/performance/odp_timer_accuracy.c
```diff
@@ -1534,6 +1534,41 @@ static void interval_loop_concurrency(test_global_t *test_global)
 	}
```
TuomasTaipale: It seems that the tester starts slightly more timers than it needs (could be by design, don't know) and when the timer period is close to the wait time defined for the schedule call in test loop, some leftover timeout events may be stuck in completion queues.

The drainage could be baked into the main loop by bumping up the timeout defined in line 1277, I wonder if this could be enough to fix this:
```diff
diff --git a/test/performance/odp_timer_accuracy.c b/test/performance/odp_timer_accuracy.c
index b754ebd91..e05c9a6d4 100644
--- a/test/performance/odp_timer_accuracy.c
+++ b/test/performance/odp_timer_accuracy.c
@@ -1274,7 +1274,7 @@ static int run_test(void *arg)
        uint64_t tmo_ns;
        timer_ctx_t *ctx;
        odp_thrmask_t mask;
-       uint64_t wait = odp_schedule_wait_time(10 * ODP_TIME_MSEC_IN_NS);
+       const uint64_t wait = odp_schedule_wait_time(4 * test_global->opt.period_ns);
        odp_schedule_group_t group = ODP_SCHED_GROUP_INVALID;
        test_log_t *log = test_global->log;
        enum mode_e mode = test_global->opt.mode;

```

TuomasTaipale (approved):

PR #2250, opened 2025-08-20
Mpmc ring cache alignment

sanvai01:
This patch improves the ring performance in shared queue scenarios by about 20%


MatiasElo:
Hi @sanvai01, 

In what kind of scenarios is the performance improvement visible? Which test applications and parameters were used?

PR #2253, opened 2025-08-27
[PATCH v3] validation: sched: add test for scheduling group functionality

Rayska:
Add new validation test for verifying scheduling group functionality.
 - Each worker has its own private schedule group and scheduled queue.
 - Source queue of scheduled events is verified to be from the workers' private schedule group.

Minor changes to odp_cunit_common.h
 - Correct comment regarding odp_cunit_thread_create() return value
 - Rename include guard ODP_CUNICT_COMMON_H > ODP_CUNIT_COMMON_H

### DIFF
--- a/test/validation/api/scheduler/scheduler.c
+++ b/test/validation/api/scheduler/scheduler.c
@@ -62,6 +62,8 @@
 
 #define FIFO_MAX_EVENTS  151
 
+#define PGT_NUM_EV 16
+
 /* Test global variables */
 typedef struct {
 	int num_workers;
@@ -106,6 +108,9 @@ typedef struct {
 		uint16_t num_thr;
 		odp_event_t event[FIFO_MAX_EVENTS];
 	} fifo;
+
+	odp_schedule_group_t pgt_groups[MAX_WORKERS];
+	odp_queue_t pgt_queues[MAX_WORKERS];
 
 } test_globals_t;
 
@@ -757,7 +762,7 @@ static void scheduler_test_max_queues(odp_schedule_sync_t sync)
 
 		if (odp_queue_enq(queue[src_idx], ev)) {
 			ODPH_ERR("Enqueue failed. Round %u, queue idx %u\n", round, src_idx);
-			CU_FAIL("Enqueue failed\n")
+			CU_FAIL("Enqueue failed\n");
 			odp_event_free(ev);
 			break;
 		}
@@ -1982,6 +1987,45 @@ static void parallel_execute(odp_schedule_sync_t sync, int num_queues,
 	reset_queues(args);
 }
 
+/* Worker has own private schedule group and queue.
+ * Scheduled events are confirmed to be only from that private group and queue. */
+static int private_group_thread(void *arg)
+{
+	uint32_t thr = (uintptr_t)arg;
+	odp_schedule_group_t private_group = globals->pgt_groups[thr];
+	odp_queue_t private_queue = globals->pgt_queues[thr];
+	odp_event_t ev;
+	odp_queue_t from;
+	odp_thrmask_t mask;
+	uint32_t i = 0;
+
+	CU_ASSERT_FATAL(private_group != ODP_SCHED_GROUP_INVALID);
+	CU_ASSERT_FATAL(private_queue != ODP_QUEUE_INVALID);
+
+	odp_thrmask_zero(&mask);
+	odp_thrmask_set(&mask, odp_thread_id());
+	CU_ASSERT_FATAL(odp_schedule_group_join(private_group, &mask) == 0);
+
+	odp_barrier_wait(&globals->barrier);
+
+	while (i < PGT_NUM_EV) {
+		ev = odp_schedule(&from, odp_schedule_wait_time(10 * ODP_TIME_MSEC_IN_NS));
+		if (ev == ODP_EVENT_INVALID)
+			break;
+
+		CU_ASSERT(odp_queue_sched_group(from) == private_group);
+		CU_ASSERT(from == private_queue);
+		odp_event_free(ev);
+		i++;
+	}
+
+	CU_ASSERT(i == PGT_NUM_EV);
+	CU_ASSERT(drain_queues() == 0);
+	CU_ASSERT(odp_schedule_group_leave(private_group, &mask) == 0);
+
+	return 0;
+}
+
 /* 1 queue 1 thread ODP_SCHED_SYNC_PARALLEL */
 static void scheduler_test_1q_1t_n(void)
 {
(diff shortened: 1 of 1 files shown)


test/validation/api/scheduler/scheduler.c
```diff
@@ -2351,6 +2432,32 @@ static void scheduler_test_pause_enqueue(void)
...
+{
+	odp_shm_t shm;
+	thread_args_t *args;
+	test_globals_t *globals;
+	void *arg_ptr;
+
+	shm = odp_shm_lookup(GLOBALS_SHM_NAME);
+	CU_ASSERT_FATAL(shm != ODP_SHM_INVALID);
+	globals = odp_shm_addr(shm);
+	CU_ASSERT_FATAL(globals != NULL);
```
MatiasElo: There's already a global `globals` variable which could be used directly and passed also to `odp_cunit_thread_create()`.  `thread_args_t` seems unnecessary for this test.

test/validation/api/scheduler/scheduler.c
```diff
@@ -2351,6 +2432,32 @@ static void scheduler_test_pause_enqueue(void)
...
+	shm = odp_shm_lookup(SHM_THR_ARGS_NAME);
+	CU_ASSERT_FATAL(shm != ODP_SHM_INVALID);
+	args = odp_shm_addr(shm);
+	CU_ASSERT_FATAL(args != NULL);
+
+	args->globals = globals;
+
+	/* One worker per CPU creates its own group and validates scheduling */
+	arg_ptr = args;
+	CU_ASSERT_FATAL(odp_cunit_thread_create(globals->num_workers, private_group_thread,
```
MatiasElo: The number of supported schedule groups (`odp_schedule_capability_t.max_groups`) may be smaller than the number of available workers. Note that this may effect the barrier inside the test.

test/validation/api/scheduler/scheduler.c
```diff
@@ -1982,6 +1982,87 @@ static void parallel_execute(odp_schedule_sync_t sync, int num_queues,
...
+	queue_param.size        = num_ev;
+
+	enqueue_to = odp_queue_create(queue_name, &queue_param);
+	CU_ASSERT_FATAL(enqueue_to != ODP_QUEUE_INVALID);
+
+	/* Pre-drain */
+	odp_schedule_pause();
+	while ((ev = odp_schedule(NULL, ODP_SCHED_NO_WAIT)) != ODP_EVENT_INVALID)
+		odp_event_free(ev);
+	odp_schedule_resume();
```
MatiasElo: This shouldn't be necessary.

test/validation/api/scheduler/scheduler.c
```diff
@@ -1982,6 +1982,87 @@ static void parallel_execute(odp_schedule_sync_t sync, int num_queues,
...
+
+	/* Wait for all threads to finish enqueuing before scheduling */
+	odp_barrier_wait(&globals->barrier);
+
+	/* Schedule events and verify they come only from private queue/group */
+	i = 0;
+	while (i < num_ev) {
+		ev = odp_schedule(&sched_from, odp_schedule_wait_time(ODP_TIME_MSEC_IN_NS));
+		if (ev == ODP_EVENT_INVALID)
+			continue;
```
MatiasElo: The timeout could be increased to 10 msec and could break here to avoid deadlock if something is wrong with the implementation. Number of received events would then be checked after the loop.

test/validation/api/scheduler/scheduler.c
```diff
@@ -1982,6 +1982,87 @@ static void parallel_execute(odp_schedule_sync_t sync, int num_queues,
...
+		CU_ASSERT(sched_from == enqueue_to);
+		CU_ASSERT(odp_queue_sched_group(sched_from) == private_group);
+		odp_event_free(ev);
+		i++;
+	}
+
+	odp_schedule_pause();
+	while ((ev = odp_schedule(NULL, ODP_SCHED_NO_WAIT)) != ODP_EVENT_INVALID)
+		odp_event_free(ev);
+	odp_schedule_resume();
```
MatiasElo: Could use `drain_queues()` function and check that zero is returned.

test/validation/api/scheduler/scheduler.c
```diff
@@ -1982,6 +1982,87 @@ static void parallel_execute(odp_schedule_sync_t sync, int num_queues,
...
+	CU_ASSERT_FATAL(private_group != ODP_SCHED_GROUP_INVALID);
+
+	odp_queue_param_init(&queue_param);
+	queue_param.type        = ODP_QUEUE_TYPE_SCHED;
+	queue_param.sched.sync  = ODP_SCHED_SYNC_PARALLEL;
+	queue_param.sched.group = private_group;
+	queue_param.size        = num_ev;
+
+	enqueue_to = odp_queue_create(queue_name, &queue_param);
+	CU_ASSERT_FATAL(enqueue_to != ODP_QUEUE_INVALID);
```
MatiasElo: To make the test potentially more complex for the implementation, I'd create the schedule groups and queues, and fill the queues already in the control thread. Workers would then only join their specific group.

test/validation/api/scheduler/scheduler.c
```diff
@@ -2351,6 +2401,78 @@ static void scheduler_test_pause_enqueue(void)
...
+	odp_thrmask_t empty_mask;
+	odp_queue_param_t queue_param;
+	odp_buffer_t buffer;
+	char group_name[ODP_SCHED_GROUP_NAME_LEN];
+	char queue_name[ODP_QUEUE_NAME_LEN];
+	int idx[MAX_WORKERS];
+	void *args[MAX_WORKERS];
+	uint32_t num_predef = 3; /* Predefined groups */
+
+	CU_ASSERT_FATAL(globals != NULL);
```
MatiasElo: Kinda unnecessary check. `globals` is referenced already above anyway.

test/validation/api/scheduler/scheduler.c
```diff
@@ -2351,6 +2401,78 @@ static void scheduler_test_pause_enqueue(void)
...
+			CU_ASSERT_FATAL(buffer != ODP_BUFFER_INVALID);
+			CU_ASSERT_FATAL(odp_queue_enq(globals->pgt_queues[i],
+						      odp_buffer_to_event(buffer)) == 0);
+		}
+	}
+
+	for (int i = 0; i < num_thr; i++) {
+		idx[i] = i;
+		args[i] = &idx[i];
+	}
```
MatiasElo: Perhaps it would be a bit simpler to use an `uintptr_t` array (see `scheduler_fifo_mt()`).

test/validation/api/scheduler/scheduler.c
MatiasElo: Nothing should be returned here anymore and `drain_queues()` does the same thing.

test/validation/api/scheduler/scheduler.c
```diff
@@ -2351,6 +2401,78 @@ static void scheduler_test_pause_enqueue(void)
...
+	int num_private_groups, num_thr = globals->num_workers;
+	odp_pool_t pool = odp_pool_lookup(MSG_POOL_NAME);
+	odp_thrmask_t empty_mask;
+	odp_queue_param_t queue_param;
+	odp_buffer_t buffer;
+	char group_name[ODP_SCHED_GROUP_NAME_LEN];
+	char queue_name[ODP_QUEUE_NAME_LEN];
+	int idx[MAX_WORKERS];
+	void *args[MAX_WORKERS];
+	uint32_t num_predef = 3; /* Predefined groups */
```
MatiasElo: Could be `const` for clarity.

MatiasElo (approved):

PR #2254, opened 2025-09-01
[PATCH v3] test: bench_buffer: add new mode option

TuomasTaipale:
Add new option to control measurement mode. Two modes are provided, one for throughput and one for latency. Throughput mode keeps track of average execution times and is basically the current operation mode of the tester. Latency mode utilizes the bench tm suite and additionally tracks function minimum and maximum execution times.

v2:
- Matias' comments

v3:
- Swapped to return error when setting default CoS
- Added reviewed-by tags

### DIFF
--- a/test/performance/odp_bench_buffer.c
+++ b/test/performance/odp_bench_buffer.c
@@ -55,15 +55,25 @@
 	{.name = #run_fn, .run = run_fn, .init = init_fn, .term = term_fn, .desc = alt_name, \
 	 .cond = cond_fn}
 
+#define BENCH_INFO_TM(run_fn, init_fn, term_fn, cond_fn) \
+	{.name = #run_fn, .run = run_fn, .init = init_fn, .term = term_fn, .cond = cond_fn,\
+	 .max_rounds = 0}
+
+typedef enum {
+	M_TPUT,
+	M_LATENCY
+} meas_mode_t;
+
 /**
  * Parsed command line arguments
  */
 typedef struct {
-	int bench_idx;   /** Benchmark index to run indefinitely */
-	int burst_size;  /** Burst size for *_multi operations */
-	int cache_size;  /** Pool cache size */
-	int time;        /** Measure time vs. CPU cycles */
-	uint32_t rounds; /** Rounds per test case */
+	int bench_idx;    /** Benchmark index to run indefinitely */
+	int burst_size;   /** Burst size for *_multi operations */
+	int cache_size;   /** Pool cache size */
+	int time;         /** Measure time vs. CPU cycles */
+	meas_mode_t mode; /** Measurement mode */
+	uint32_t rounds;  /** Rounds per test case */
 } appl_args_t;
 
 /**
@@ -72,8 +82,28 @@ typedef struct {
 typedef struct {
 	/** Application (parsed) arguments */
 	appl_args_t appl;
+
 	/** Common benchmark suite data */
-	bench_suite_t suite;
+	struct {
+		union {
+			/** Basic suite for avg */
+			bench_suite_t b;
+			/** TM suite for min/max/avg */
+			bench_tm_suite_t t;
+		};
+
+		/** Loop breaker */
+		odp_atomic_u32_t *exit;
+		/** Pointer to suite return value */
+		int *retval;
+		/** Suite args */
+		void *args;
+		/** Suite runner */
+		int (*suite_fn)(void *args);
+		/** Data exporter */
+		int (*export_fn)(void *data);
+	} suite;
+
 	/** Buffer pool */
 	odp_pool_t pool;
 	/** Buffer size */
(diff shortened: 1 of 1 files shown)


test/common/bench_common.h
```diff
@@ -126,26 +121,56 @@ int bench_run(void *arg);
...
-		/* Name of function */
-		const char *name;
+/* Stamp type for recording current instant in ODP time or cycles based on suite configuration */
+typedef struct {
+	union {
+		odp_time_t time;
+		uint64_t cycles;
+	};
+
+} bench_tm_stamp_t;
```
MatiasElo: Could be simply a union?

test/common/bench_common.h
MatiasElo, line 1: Copyright can be updated.

test/performance/odp_bench_pktio_sp.c
```diff
@@ -781,14 +781,29 @@ static int cls_pmr_create(bench_tm_result_t *res, int repeat_count)
 
 	ret = odp_pktio_default_cos_set(pktio, ODP_COS_INVALID);
 
+	if (ret < 0)
+		ODPH_ERR("Warning: setting default CoS failed: %d\n", ret);
```
MatiasElo: Could return with failure also from here?

test/performance/odp_bench_pktio_sp.c
```diff
@@ -296,21 +296,21 @@ static int pktio_capability(bench_tm_result_t *res, int repeat_count)
...
+	bench_tm_stamp_t s1, s2;
 	int ret;
 	uint8_t id1 = bench_tm_func_register(res, "odp_pktio_capability()");
 
 	for (int i = 0; i < repeat_count; i++) {
-		t1 = odp_time_local_strict();
+		res->op.now_fn(&s1);
 		ret = odp_pktio_capability(pktio, capa);
-		t2 = odp_time_local_strict();
+		res->op.now_fn(&s2);
```
MatiasElo: Instead of calling the function pointer directly, this could be wrapped by a function (e.g. `bench_tm_stamp/now()`) to make the API more uniform. Same comment for `res->op.to_u64_fn()`.

test/performance/odp_bench_pktio_sp.c
MatiasElo, line 1: Could mention in the commit message that _bench_pktio_sp_ default measurement unit has been changed from nsecs to cycles.

test/common/bench_common.c
```diff
@@ -286,7 +286,7 @@ int bench_tm_run(void *arg)
 			if (bench->init != NULL)
 				bench->init();
 
-			if (bench->run(&res, rounds)) {
+			if (!bench->run(&res, rounds)) {
```
MatiasElo, line 299: The `bench_tm_run_fn_t` typedef documentation should also be updated to reflect this.

test/performance/odp_bench_buffer.c
```diff
@@ -575,6 +1162,8 @@ static void usage(char *progname)
 	       "  -i, --index <idx>       Benchmark index to run indefinitely.\n"
 	       "  -r, --rounds <num>      Run each test case 'num' times (default %u).\n"
 	       "  -t, --time <opt>        Time measurement. 0: measure CPU cycles (default), 1: measure time\n"
+	       "  -a, --add_meas          Additional measurements. Track function minimum and\n"
+	       "                          maximum in addition to average execution time.\n"
```
MatiasElo: A minor thing, but thinking if naming the option `mode` would be clearer. The modes could then be for example _throughput_ and _latency_.

test/performance/odp_bench_buffer.c
```diff
@@ -558,6 +588,563 @@ static int event_flow_id_set(void)
 	return i;
 }
 
+static int buffer_from_event_tm(bench_tm_result_t *res, int repeat_count)
```
MatiasElo: A matter of taste, but I'd put the basic and tm test variants back-to-back, so it's easier to compare the implementations at a glance.

test/performance/odp_bench_pktio_sp.c
TuomasTaipale: Wasn't sure what the original version intended (the return value gets overridden a lot), so figured that, warning here and then errors with CoS/queue destructions.

test/performance/odp_bench_pktio_sp.c
TuomasTaipale: Yes, originally I had them wrapped in library functions, but then didn't go through with it as the export functions anyways touch the result innards. However, now I'm convined that this is a good idea so changed in v2. They are regular functions now, can be changed to header inlines later if needed.

test/performance/odp_bench_buffer.c
TuomasTaipale: Yes, agree, it allows better extensibility. Changed in v2.

test/performance/odp_bench_pktio_sp.c
MatiasElo: The original version seems broken, so let's just return error here.

MatiasElo (approved):

PR #2255, opened 2025-09-05
[PATCH v8] Fix warning from onnxruntime

lifang-zhang:


DEPENDENCIES
```diff
@@ -300,8 +300,10 @@ Prerequisites for building the OpenDataPlane (ODP) API
 
    # Configure
    $ ./build.sh --config RelWithDebInfo --build_shared_lib --parallel
-   $ tools/ci_build/github/linux/copy_strip_binary.sh -r build/Linux/ -a onnxruntime -l libonnxruntime.so.1.14.0 -c RelWithDebInfo -s . -t <commit id>
-   $ cp -r build/Linux/onnxruntime/ <onnxruntime path>
+   $ cmake --install build/Linux/RelWithDebInfo/ --prefix build/Linux/RelWithDebInfo/installed
```
TuomasTaipale: `linux-gen` is supposed to work on a few different architectures, perhaps a few lines should be added about this support, at least for arm64. The "Prebuilt" section also seems to just mention x86. At least something could be said about build types as well, e.g. "RelWithDebInfo" builds something that is not as performant as "Release". (A link to documentation or something written here.)

Something should also be said about the tool versioning, e.g. on Ubuntu 22.04, packaged `cmake` is too old.

By the way, "Prebuilt onnxruntime download" and "Build onnxruntime from source" sections have the same numbering "3.7.1", they should be different?

DEPENDENCIES
```diff
@@ -300,8 +300,10 @@ Prerequisites for building the OpenDataPlane (ODP) API
...
+   $ mv build/Linux/RelWithDebInfo/installed/include/onnxruntime/* build/Linux/RelWithDebInfo/installed/include/
+   $ rm -r build/Linux/RelWithDebInfo/installed/include/onnxruntime
+   $ cp -r build/Linux/RelWithDebInfo/installed/ <onnxruntime path>
```
TuomasTaipale: I guess the pkgconfig file still points to header files being under `include/onnxruntime/*`? If so, the installation is broken for other projects that could be using this installation. Maybe there could be symbolic link(s) at `include/` level to `include/onnxruntime/`? This way both ways could coexist.

DEPENDENCIES
lifang-zhang: Good points - thank you! arm64 has now been added to the prebuilt section and recommended to check from documentation how to build from source for arm64.

DEPENDENCIES
lifang-zhang: The generated pkgconfig file incorrectly points to /usr/local, the prebuilt one does not include the pkgconfig file. But I created a symbolic link for it or should we just delete the pkgconfig file?
```
$ cat ~/install/onnxruntime/lib/pkgconfig/libonnxruntime.pc
prefix=/usr/local
bindir=${prefix}/bin
mandir=${prefix}/share/man
docdir=${prefix}/share/doc/onnxruntime
libdir=${prefix}/lib
includedir=${prefix}/include/onnxruntime

Name: onnxruntime
Description: ONNX runtime
URL: https://github.com/microsoft/onnxruntime
Version: 1.23.0
Libs: -L${libdir} -lonnxruntime
Cflags: -I${includedir}
```

TuomasTaipale: OK, maybe the pkgconfig creation is just buggy and doesn't take into account passed `prefix`. If the pkgconfig is buggy then there's probably no point having the include files linked and just leave the original solution you suggested in place (move, remove, copy).

(I guess the copy step could be omitted if the installation location would be regarded as `<onnxruntime path>`?)

lifang-zhang: Yes, with prefix set to `<onnxruntime path>`, the copy step is not needed.

TuomasTaipale:
Please also add review tags (Reviewed-by: Tuomas Taipale \<tuomas.taipale@nokia.com\>) to the commits, then this is good to go.

PR #2256, opened 2025-09-12
[PATCH v2] linux-gen: added wfe and wfet library functions

sanvai01:
Added wfe and wfet instruction library functions in wait_until header file

### DIFF
--- a/test/common/bench_common.c
+++ b/test/common/bench_common.c
@@ -22,7 +22,7 @@ void bench_suite_init(bench_suite_t *suite)
 	odp_atomic_init_u32(&suite->exit_worker, 0);
 }
 
-void bench_run_indef(bench_info_t *info, odp_atomic_u32_t *exit_thread)
+static void run_indef(bench_info_t *info, odp_atomic_u32_t *exit_thread)
 {
 	const char *desc;
 
@@ -73,7 +73,7 @@ int bench_run(void *arg)
 			if ((i + 1) != suite->indef_idx)
 				continue;
 
-			bench_run_indef(&suite->bench[i], &suite->exit_worker);
+			run_indef(&suite->bench[i], &suite->exit_worker);
 			return 0;
 		}
 
@@ -132,6 +132,8 @@ void bench_tm_suite_init(bench_tm_suite_t *suite)
 {
 	memset(suite, 0, sizeof(bench_tm_suite_t));
 
+	suite->measure_time = true;
+
 	odp_atomic_init_u32(&suite->exit_worker, 0);
 }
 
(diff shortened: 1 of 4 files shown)


platform/linux-generic/arch/aarch64/cpu_flags.c
```diff
@@ -992,6 +992,13 @@ static void _odp_sys_info_print_acle_flags(void)
 	_ODP_PRINT("%s\n", ndef);
 #endif
 
+	_ODP_PRINT("  __ARM_FEATURE_WFXT               ");
```
JannePeltonen, line 995: See my comment in PR 2257 about commit message and merge commits.

platform/linux-generic/arch/aarch64/odp/api/abi/wait_until.h
```diff
@@ -33,6 +33,32 @@ _odp_wait_until_equal_acq_u32(odp_atomic_u32_t *addr, uint32_t expected)
 	} while (expected != value);
 }
 
+static inline void
+_odp_wait_until_equal_rel_u32(odp_atomic_u32_t *addr, uint32_t expected)
```
JannePeltonen, line 37: How is this function supposed/allowed to be different from _odp_wait_until_equal_32()? What does the _rel_ part of this function indicate? In ODP, _rel_ in a function name implies release memory ordering, but I cannot see what release ordering would mean in a function that does not stores. Or does ldxr synchronize with acquire loads in other threads? Even if it does, I am not sure providing a release type function would be useful. It could also be cumbersome for a generic C implementation, but since such is not provided in this PR, then why is this even a separate function in this aarch64 specific implementation?

platform/linux-generic/arch/aarch64/odp/api/abi/wait_until.h
```diff
@@ -33,6 +33,32 @@ _odp_wait_until_equal_acq_u32(odp_atomic_u32_t *addr, uint32_t expected)
...
+    #endif
+		__asm__ volatile("ldxr %w0, [%1]"
+					 : "=&r" (value)
+					 : "r" (var)
+					 : "memory");
+	} while (expected != value);
+}
+
+static inline void
+_odp_wait_until_equal_32(odp_atomic_u32_t *addr, uint32_t expected)
```
JannePeltonen, line 57: Do we even need this function with a relaxed memory order? Aren't most (practically all?) concievable uses such that acquire is required? Maybe there is not need to introduce this new function until there is an actual user of it.

platform/linux-generic/arch/aarch64/odp/api/abi/wait_until.h
```diff
@@ -33,6 +33,32 @@ _odp_wait_until_equal_acq_u32(odp_atomic_u32_t *addr, uint32_t expected)
...
+static inline void
+_odp_wait_until_equal_rel_u32(odp_atomic_u32_t *addr, uint32_t expected)
+{
+	uint32_t value;
+	uint32_t *var = &addr->v;
+
+	__asm__ volatile("sevl" : : : "memory");
+	do {
+    #ifdef __ARM_FEATURE_WFXT
+	__asm__ volatile("wfet %x0" : : "r"(100) : "memory");
```
JannePeltonen, line 45: Why would WFET be useful here and better than WFE? If it is useful, then why not make the same change in the existing _odp_wait_until_equal_acq_u32() function?

PR #2257, opened 2025-09-12
[PATCH v2] linux-gen: cpu_pause: added wfet instruction along with isb

sanvai01:
Added wfet instruction in cpu_pause function along with isb instruction.

### DIFF
--- a/test/common/bench_common.c
+++ b/test/common/bench_common.c
@@ -22,7 +22,7 @@ void bench_suite_init(bench_suite_t *suite)
 	odp_atomic_init_u32(&suite->exit_worker, 0);
 }
 
-void bench_run_indef(bench_info_t *info, odp_atomic_u32_t *exit_thread)
+static void run_indef(bench_info_t *info, odp_atomic_u32_t *exit_thread)
 {
 	const char *desc;
 
@@ -73,7 +73,7 @@ int bench_run(void *arg)
 			if ((i + 1) != suite->indef_idx)
 				continue;
 
-			bench_run_indef(&suite->bench[i], &suite->exit_worker);
+			run_indef(&suite->bench[i], &suite->exit_worker);
 			return 0;
 		}
 
@@ -132,6 +132,8 @@ void bench_tm_suite_init(bench_tm_suite_t *suite)
 {
 	memset(suite, 0, sizeof(bench_tm_suite_t));
 
+	suite->measure_time = true;
+
 	odp_atomic_init_u32(&suite->exit_worker, 0);
 }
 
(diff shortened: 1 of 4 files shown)


configure.ac
```diff
@@ -199,6 +199,17 @@ if test "x$enable_host_optimization" = "xno" ; then
 	ARCH_ABI=default-linux
 fi
 
+AC_ARG_ENABLE([wfxt-cpu-pause],
```
JannePeltonen, line 202: The commit message could be improved:
- The rationale of the use of WFET in odp_cpu_pause() could be explained
- The function is not cpu_pause but odp_cpu_pause.
- Use of WFET instead of ISB is selected through configure flag, not compiler flag
- The title (and often the message body too) are typically in the imperative form in ODP git

JannePeltonen, line 202: ODP pull requests are not supposed to contain merge commits as ODP development in this repository is based on rebases and linear commit history. So could you please rebase your branch on top of the latest master instead of merging it.

platform/linux-generic/arch/aarch64/odp/api/abi/cpu_inlines.h
```diff
@@ -28,12 +29,17 @@ extern _odp_cpu_cycles_global_t _odp_cpu_cycles_glob;
 
 static inline void _odp_cpu_pause(void)
 {
+#if defined(ENABLE_WFXT_CPU_PAUSE) && defined(__ARM_FEATURE_WFXT)
+	__asm__ volatile("sevl" : : : "memory");
```
JannePeltonen, line 33: Why is the SEVL done here? Doesn't is basically defeat the purpose of the following WFET as the WFET would wake immediately and not wait for the timeout? If the SEVL is intentional and serves a purpose, then maybe that could be explained in a comment, unless it is just me who cannot see it. But if the SEVL stays, then why WFET instead of just WFE which would be available in much earlier arch levels?

platform/linux-generic/arch/aarch64/odp/api/abi/cpu_inlines.h
```diff
@@ -28,12 +29,17 @@ extern _odp_cpu_cycles_global_t _odp_cpu_cycles_glob;
...
+		__asm__ volatile("wfet %x0" : : "r"(100) : "memory");
```
JannePeltonen, line 34: Why the delay is 100? I presume the unit is clock cycles. The constant could be defined as a macro outside the inline asm and have some comment about why the particular value is selected. I wonder if the timeout of 100 cycles adds too much unwanted latency in some uses of odp_cpu_pause().

odp_cpu_pause() is being used in quite many places in the linux-gen ODP implementation. It appears that many of the call sites could use WFE through _odp_wait_until_equal_acq_u32() so that they would not spin unnecessarily and would resume with lower latency when ready. Maybe many of the call sites should first be changed before making them use WFET through the odp_cpu_pause() call?

PR #2258, opened 2025-09-15
[PATCH v5] api: crypto: remove legacy operation type

JannePeltonen:
Fully remove the deprecated legacy crypto operation type to reduce maintenance work in ODP implementations.

### DIFF
--- a/include/odp/api/spec/crypto.h
+++ b/include/odp/api/spec/crypto.h
@@ -215,26 +215,6 @@ int odp_crypto_result(odp_crypto_packet_result_t *result,
  * Use of the pkt_out parameter depends on the configured crypto operation
  * type as described below.
  *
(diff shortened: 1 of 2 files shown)


include/odp/api/spec/crypto.h
MatiasElo, line 1: `odp_crypto_session_param_t.output_pool` documentation still has a reference to legacy operation type.

JannePeltonen: Fixed in v2.

Now the output_pool parameter is no longer needed by any existing op type, so we could remove it (or leave for future use), but since the parameter was not deprecated, I do not touch it in this PR.

ktejasree123:
Acked-by: Tejasree Kondoj ktejasree@marvell.com

PR #2259, opened 2025-09-16
linux-gen: queue: implement event aggregation

MatiasElo:


test/performance/odp_l2fwd.c
```diff
@@ -864,6 +874,157 @@ static int run_worker_sched_mode_vector(void *arg)
 	return 0;
 }
 
+/*
+ * Packet IO worker thread using scheduled queues and event vector mode.
+ *
+ * arg  thread arguments of type 'thread_args_t *'
+ */
+static int run_worker_sched_mode_event_vector(void *arg)
```
JannePeltonen, line 882: This is largely identical to run_worker_check_mode_vector(). I wonder if the functions could be combined.

test/performance/odp_l2fwd.c
```diff
@@ -2091,14 +2315,18 @@ static void usage(char *progname)
 	       "                                 1: Create a pool per interface\n"
 	       "  -n, --num_pkt <num>            Number of packets per pool. Default is 16k or\n"
 	       "                                 the maximum capability. Use 0 for the default.\n"
-	       "  -u, --vector_mode              Enable vector mode.\n"
+	       "  -u, --vector_mode <num>        Enable vector mode.\n"
```
JannePeltonen, line 2320: I think it would be more convenient (e.g. in scripts) if --vector_mode had 3 values, one of them being vector_mode disabled.

test/validation/api/queue/queue.c
```diff
@@ -1326,6 +1326,138 @@ static int check_sched_queue_aggr(void)
...
+			odp_event_vector_t evv = odp_event_vector_from_event(event);
+			uint32_t num = odp_event_vector_tbl(evv, &event_tbl);
+
+			CU_ASSERT(odp_event_vector_size(evv) == num);
+			CU_ASSERT(num <= vector_size);
+			CU_ASSERT(odp_event_vector_type(evv) == ODP_EVENT_BUFFER);
+
+			if (uarea_size)
+				CU_ASSERT(odp_event_vector_user_area(evv) != NULL);
+			CU_ASSERT(odp_event_vector_user_flag(evv) == 0);
```
JannePeltonen, line 1467: It could also be checked that odp_event_vector_type() is ODP_EVENT_ANY or ODP_EVENT_BUFFER.

test/validation/api/queue/queue.c
```diff
@@ -1326,6 +1326,138 @@ static int check_sched_queue_aggr(void)
...
+	aggr.event_type = ODP_EVENT_BUFFER;
+	aggr.max_tmo_ns = timeout;
+
+	odp_queue_param_init(&param);
+	param.type = queue_type;
+	param.size = num_events;
+	param.num_aggr = 1;
+	param.aggr = &aggr;
+
+	queue = odp_queue_create("queue_test_aggr", &param);
```
JannePeltonen, line 1408: After this both param and aggr could be memset to zero, to hopefully notice if the ODP incorrectly accesses them after queue creation.

test/validation/api/queue/queue.c
```diff
@@ -1326,6 +1326,138 @@ static int check_sched_queue_aggr(void)
...
+			src_queue = queue;
+			event = odp_queue_deq(queue);
+		} else {
+			event = odp_schedule(&src_queue, odp_schedule_wait_time(timeout));
+		}
+
+		if (event == ODP_EVENT_INVALID)
+			break;
+
+		CU_ASSERT(src_queue == queue);
```
JannePeltonen, line 1454: I think the test could also add a sequence number in the test events and check that the received events are in the right order (i.e. the ones gone through the aggregator are in order with respect to each other and similarly for the ones gone through the queue directly.

The test could also check that the events that were supposed to go through the queue directly were not aggregated. I wonder if it would be possible to require that at least some of the events that went through the aggregator were actually aggregated. Maybe not.

test/validation/api/queue/queue.c
```diff
@@ -1326,6 +1326,138 @@ static int check_sched_queue_aggr(void)
...
+				CU_ASSERT(odp_event_is_valid(event_tbl[i]));
+			num_recv += num;
+		} else {
+			CU_ASSERT(odp_event_type(event) == ODP_EVENT_BUFFER);
+			CU_ASSERT(odp_event_pool(event) == pool);
+			num_recv++;
+		}
+		odp_event_free(event);
+	}
+	CU_ASSERT(num_recv == num_events);
```
JannePeltonen: I think this may not be always true if the implementation does not support timeouts, performs aggregation (but maybe not every time) and flushes the aggregators only when they get full. I think that without timeout we can only expect that num_recv is num_events minus the events that could still be in the aggregator queue.

test/validation/api/pktio/pktio.c
```diff
@@ -714,6 +829,31 @@ static int get_packets(pktio_info_t *pktio_rx, odp_packet_t pkt_tbl[],
...
+			   odp_event_type(evt_tbl[i]) == ODP_EVENT_VECTOR &&
+			   num_pkts < num) {
+			odp_event_vector_t evv;
+			odp_event_t *event_tbl;
+			int evv_len;
+
+			evv = odp_event_vector_from_event(evt_tbl[i]);
+			evv_len = odp_event_vector_tbl(evv, &event_tbl);
+			CU_ASSERT(evv_len > 0);
+			CU_ASSERT(odp_event_vector_user_flag(evv) == 0);
```
JannePeltonen, line 880: It could also be checked that odp_event_vector_type() is ODP_EVENT_ANY or ODP_EVENT_PACKET

test/validation/api/pktio/pktio.c
```diff
@@ -542,6 +582,80 @@ static odp_pktio_t create_pktv_pktio(int iface_idx, odp_pktin_mode_t imode,
...
+
+	if (imode == ODP_PKTIN_MODE_SCHED)
+		pktin_param.queue_param.sched.sync = sync_mode;
+
+	pktin_param.num_queues = 1;
+	pktin_param.op_mode = ODP_PKTIO_OP_MT_UNSAFE;
+
+	memset(&aggr_config, 0, sizeof(odp_event_aggr_config_t));
+	aggr_config.pool = evv_pool[iface_idx];
+	aggr_config.max_tmo_ns = aggr_capa.min_tmo_ns;
```
JannePeltonen, line 674: If the minimum timout value is zero (i.e. timeout is infinite), then the implementation may keep events in the aggregator queues until new events cause them to be flushed. This breaks the checks that check that all sent events were received. So maybe in aggregator testing some additional flushing events should be sent if there is no timeout and then wait_for_events() or something could just ignore and free them?

test/validation/api/classification/odp_classification_common.c
```diff
@@ -224,6 +225,24 @@ odp_packet_t receive_packet(odp_queue_t *queue, uint64_t ns, odp_bool_t enable_p
 			odp_packet_free_multi(&pkt_tbl[1], pktv_len - 1);
 		odp_packet_vector_free(pktv);
 		return pkt;
+	} else if (vector_mode == VECTOR_MODE_EVENT &&
```
JannePeltonen, line 228: Here we have the same problem that event aggregators with no timeout may still contain events and this receive function will not get them.

test/performance/odp_l2fwd.c
MatiasElo: Updated in v4.

test/performance/odp_l2fwd.c
MatiasElo: They could but I'd rather not do that. The plan is to deprecate packet vectors soon and using separate functions makes that simpler. Also, if someone is tracking `run_worker_check_mode_vector()` performance, the combined function could affect the results.

test/validation/api/queue/queue.c
MatiasElo: `odp_event_vector_type()` is checked couple lines above. Currently the check only accepts `ODP_EVENT_BUFFER`, as the aggregator has been configured with that event type.

test/validation/api/queue/queue.c
MatiasElo: Added in v4.

test/validation/api/queue/queue.c
MatiasElo: Additional checks added in v4. Since the test is using a rather long timeout and the events are enqueued back to back, I think it's safe to assume that at least one vector is generated. The check can be removed later if it's deemed problematic.

test/validation/api/queue/queue.c
MatiasElo: Check updated in v4.

test/validation/api/pktio/pktio.c
MatiasElo: Check added in v4.

test/validation/api/pktio/pktio.c
MatiasElo: Extra flush events added in v4.

platform/linux-generic/include/odp_queue_basic_internal.h
```diff
@@ -33,6 +33,18 @@ extern "C" {
...
 #define QUEUE_STATUS_SCHED        4
 
+typedef struct ODP_ALIGNED_CACHE event_aggr_s {
+	odp_queue_t base_queue;
+	odp_pool_t pool;
+	uint32_t max_size;
+	odp_event_type_t event_type;
+
+	odp_ticketlock_t lock;
+	odp_event_t event_tbl[CONFIG_EVENT_VECTOR_MAX_SIZE];
```
JannePeltonen, line 43: This adds 1 kB in queue_entry_t and about 4 MB total because of the added event aggregator queues. Maybe it would be better not to have the static arrays (or have them only for the aggregators, not all queues), but allocate an event vector for an aggregator when needed and aggregate the events directly to the event vector. This would remove the need to copy event between event_tbl and the event vector.

That said, maybe this can be left for a future improvement.

btw, I noticed that the SHM allocation for the queues ("_odp_queue_basic_global" SHM block) uses sizeof(queue_entry_t) as the alignment requirement, which is wrong and affected by this change. I think the alignment should be alignof(queue_entry_t) (which is ok in C11, with stdalign.h included).

platform/linux-generic/odp_event_validation.c
```diff
@@ -57,7 +57,8 @@ static const _odp_ev_info_t ev_info_tbl[] = {
 	[_ODP_EV_PACKET_FREE_SP]    = {.str = "odp_packet_free_sp()"},
 	[_ODP_EV_PACKET_IS_VALID]   = {.str = "odp_packet_is_valid()"},
 	[_ODP_EV_QUEUE_ENQ]         = {.str = "odp_queue_enq()"},
-	[_ODP_EV_QUEUE_ENQ_MULTI]   = {.str = "odp_queue_enq_multi()"}
+	[_ODP_EV_QUEUE_ENQ_MULTI]   = {.str = "odp_queue_enq_multi()"},
+	[_ODP_EV_QUEUE_ENQ_AGGR]    = {.str = "odp_queue_enq_aggr()"}
```
JannePeltonen: This could have a comma at the end so future additions do not need to add it.

platform/linux-generic/odp_queue_basic.c
```diff
@@ -257,6 +277,135 @@ static uint32_t queue_lock_count(odp_queue_t handle)
...
+		_ODP_ERR("Invalid pool handle\n");
+		return -1;
+	}
+	if (_odp_pool_type(param->aggr->pool) != ODP_POOL_EVENT_VECTOR) {
+		_ODP_ERR("Pool type is not event vector\n");
+		return -1;
+	}
+
+	/* Event aggregation implementation includes locking */
+	if (param->nonblocking == ODP_NONBLOCKING_LF) {
```
JannePeltonen: There is also ODP_NONBLOCKING_WF, so I think the condition should be e.g. ...!= ODP_BLOCKING. That would be more robust even with the current call site being such that ODP_NONBLOCKING_WF has already been rejected.

platform/linux-generic/odp_queue_basic.c
```diff
@@ -257,6 +277,135 @@ static uint32_t queue_lock_count(odp_queue_t handle)
...
+			_ODP_ERR("Too many aggregators per queue: %u > %u\n",
+				 param->num_aggr, capa.plain.aggr.max_num_per_queue);
+			return -1;
+		}
+		if (param->aggr->max_tmo_ns > capa.plain.aggr.max_tmo_ns) {
+			_ODP_ERR("Too large timeout: %" PRIu64 " > %" PRIu64 "\n",
+				 param->aggr->max_tmo_ns, capa.plain.aggr.max_tmo_ns);
+			return -1;
+		}
+		if (param->aggr->max_tmo_ns &&
```
JannePeltonen: Even if the max_tmo_ns is zero, it should not be smaller than the min_tmo_ns capa. Maybe the API could be better (UINT64_MAX for no timeout, or separate timeout-enabled flag or something like that), but the current API is like that.

platform/linux-generic/odp_queue_basic.c
```diff
@@ -257,6 +277,135 @@ static uint32_t queue_lock_count(odp_queue_t handle)
...
+		}
+		return 0;
+	}
+
+	/* Scheduled queue */
+	if (odp_schedule_capability(&sched_capa)) {
+		_ODP_ERR("Failed to read scheduler capabilities\n");
+		return -1;
+	}
+	if (param->num_aggr > sched_capa.aggr.max_num_per_queue) {
```
JannePeltonen: Here the same checks are repeated for scheduled queues. It would be better to have the checks just once, using a pointer to odp_event_aggr_capability_t.

platform/linux-generic/odp_queue_basic.c
```diff
@@ -257,6 +277,135 @@ static uint32_t queue_lock_count(odp_queue_t handle)
...
+
+	return 0;
+}
+
+static queue_entry_t *event_aggr_reserve(queue_global_t *global)
+{
+	for (uint32_t i = 0; i < CONFIG_MAX_EVENT_AGGR; i++) {
+		queue_entry_t *aggr_queue = &global->aggr_queue[i];
+
+		LOCK(aggr_queue);
```
JannePeltonen, line 349: I think it would be better to have some sort of freelist and a separate lock for the allocation so that the search would not cause extra cache misses etc to a whole bunch of aggregators that are already in active use. And the allocation would be quicker too.

platform/linux-generic/odp_queue_basic.c
```diff
@@ -370,6 +519,19 @@ static odp_queue_t queue_create(const char *name,
 		}
 	}
 
+	/* Event aggregator */
+	if (param->num_aggr) {
+		queue_entry_t *aggr_queue = event_aggr_reserve(_odp_queue_glb);
+
+		if (aggr_queue == NULL) {
+			queue->status = QUEUE_STATUS_FREE;
```
JannePeltonen, line 502: This causes a potential data race with the queue->status checking in the allocation loop. But there are already data races in the existing code though, between checking the status and writing to it inside and outside the locked section in this function and maybe elsewhere (scheduler) too.

I am not sure what, if anything, can or needs to be done in the context of this PR. Maybe just add a small comment somewhere (like at the place where queue->status is read for the first time), that things are not strictly correct?. Some later change could then improve things (make the status atomic type or something).

Having a separate lock and freelist for (normal) queue allocation might also be worthwhile, but not really in the topic of this PR.

platform/linux-generic/odp_queue_basic.c
```diff
@@ -417,6 +579,8 @@ static int queue_destroy(odp_queue_t handle)
 	if (handle == ODP_QUEUE_INVALID)
 		return -1;
 
+	_ODP_ASSERT(queue->type != ODP_QUEUE_TYPE_AGGR);
```
JannePeltonen: I think this could be a run-time check even in non-debug builds.

platform/linux-generic/odp_queue_basic.c
```diff
@@ -515,6 +699,197 @@ static odp_queue_t queue_lookup(const char *name)
 	return ODP_QUEUE_INVALID;
 }
 
+/* Used by packet IO and classifier with at most 'odp_event_aggr_config_t.max_size' packets. Always
+ * returns all available events. */
+odp_event_vector_t _odp_event_vector_create(odp_queue_t aggr_handle,
```
JannePeltonen, line 682: Since this function is only used by later commits, I think it would be clearer to introduce it there, with the first user.

platform/linux-generic/odp_queue_basic.c
```diff
@@ -515,6 +699,197 @@ static odp_queue_t queue_lookup(const char *name)
...
+					    uint32_t num)
+{
+	event_aggr_t *aggr = &qentry_from_handle(aggr_handle)->aggr;
+	odp_event_t *evv_tbl;
+	odp_event_vector_t evv;
+	uint32_t i, j;
+
+	_ODP_ASSERT(qentry_from_handle(aggr_handle)->type == ODP_QUEUE_TYPE_AGGR);
+
+	evv = odp_event_vector_alloc(aggr->pool);
```
JannePeltonen, line 704: I wonder why we need to always create a vector for packet input. Couldn't we just aggregate the events and only create vectors when the aggregator is full, similar to the normal aggregator enqueue case?

platform/linux-generic/odp_queue_basic.c
```diff
@@ -515,6 +699,197 @@ static odp_queue_t queue_lookup(const char *name)
...
+	queue_entry_t *queue = qentry_from_handle(handle);
+	event_aggr_t *aggr = &queue->aggr;
+
+	if (odp_unlikely(queue->type != ODP_QUEUE_TYPE_AGGR))
+		return queue->enqueue(handle, (_odp_event_hdr_t *)(uintptr_t)ev);
+
+	_ODP_ASSERT(param != NULL);
+
+	/* Cannot stash events as ordered_stash_release() ignores odp_aggr_enq_param_t */
+	_odp_sched_fn->ord_wait(handle);
```
JannePeltonen: If there are already events in the stash, these new events will overtake them, mixing the ordering, right? Previous enqueues might have used the normal enqueue function that then stashed the events.

platform/linux-generic/odp_packet_io.c
```diff
@@ -852,28 +853,40 @@ static inline odp_packet_vector_t packet_vector_create(odp_packet_t packets[], u
...
+	if (!vector_enabled)
+		return entry->ops->recv(entry, pktin_index, (odp_packet_t *)event_hdrs, num);
 
-	num_rx = entry->ops->recv(entry, pktin_index, (odp_packet_t *)event_hdrs, num);
+	/* Always try to receive full vectors */
+	num = entry->in_queue[pktin_index].vector.max_size;
 
-	if (!vector_enabled || num_rx < 2)
+	num_rx = entry->ops->recv(entry, pktin_index, pkt_tbl, num);
+	if (num_rx < 2) {
```
JannePeltonen: If there are already events in the aggregator queue, then they will not get delivered and this event overtakes them, right?

platform/linux-generic/odp_queue_basic.c
```diff
@@ -753,6 +753,11 @@ int _odp_event_aggr_enq(queue_entry_t *aggr_queue, _odp_event_hdr_t *event_hdr[]
 
 	_ODP_ASSERT(aggr_queue->type == ODP_QUEUE_TYPE_AGGR);
 
+	/* Packet input may later drop a vector if the queue is full, so enqueue events to the queue
```
JannePeltonen: It seems the comment uses 'queue' to refer to both aggregator queue and the base queue. But I wonder if this is really needed.

JannePeltonen (commented):
Some comments to the implementation commits.

test/validation/api/classification/odp_classification_common.c
MatiasElo: It looks like the classification validation suite would require quite a bit of refactoring to enable this. Currently most of the tests (if not all) use only single events.  So I'd prefer doing this work later in a separate PR.

platform/linux-generic/odp_event_validation.c
MatiasElo: Fixed in v5.

platform/linux-generic/odp_queue_basic.c
MatiasElo: True, fixed in v5.

platform/linux-generic/odp_queue_basic.c
MatiasElo: Fixed in v5.

platform/linux-generic/odp_queue_basic.c
MatiasElo: Fixed in v5.

platform/linux-generic/odp_queue_basic.c
MatiasElo: Changed in v5.

platform/linux-generic/odp_queue_basic.c
MatiasElo: Moved in v5.

platform/linux-generic/odp_queue_basic.c
MatiasElo: Reworded in v5.

platform/linux-generic/odp_queue_basic.c
MatiasElo: Good catch, fixed in v5.

platform/linux-generic/odp_packet_io.c
MatiasElo: True, fixed in v5.

platform/linux-generic/odp_queue_basic.c
MatiasElo: Yep, added a comment in v5.

platform/linux-generic/odp_queue_basic.c
MatiasElo: Added to backlog as future improvement.

platform/linux-generic/include/odp_queue_basic_internal.h
MatiasElo: Fixed "_odp_queue_basic_global" align in v5. Added using event vectors directly to backlog as a future improvement.

platform/linux-generic/odp_queue_basic.c
MatiasElo: A good question 🙂 Probably a leftover from using packet vector function as a starting point. Fixed in v5.

test/validation/api/pktio/pktio.c
JannePeltonen: I am still thinking if any remaining extra flushing events should be received and freed after the test. odp_pktio_close() is supposed to free all remaining packets inside pktio receive buffers, but that may no longer be totally true with event aggregators. With some ODP implementations, the unread events inside the aggregator may not be "inside the pktio" anymore, so they would stay there after the pktio close and could been seen by a later schedule call, I think. Maybe it does not happen in practise in this test case though. Sending the flushing events only when there is no timeout certainly helps, but maybe it is not strictly sufficient. Maybe there is an inconsistency or unclarity in the API regarding pktio close and the fate of packets waiting in the aggregator.?

test/validation/api/queue/queue.c
```diff
@@ -1326,6 +1326,188 @@ static int check_sched_queue_aggr(void)
...
+		odp_event_free(event);
+	}
+	/* Assume at least one vector was generated */
+	CU_ASSERT(num_from_aggr >= vector_size);
+
+	/* Some events may be still pending in aggregation queue */
+	CU_ASSERT(num_from_queue + num_from_aggr >= num_events - vector_size);
+	CU_ASSERT(num_from_queue + num_from_aggr <= num_events);
+
+	CU_ASSERT(odp_queue_destroy(queue) == 0);
```
JannePeltonen, line 1491: Here we may have packets still in the aggregator. And if there was some sort of aggregator timeout or other event that causes the aggregator to flush events after the last dequeue/schedule, the events may be even in the base queue. And ODP API says that odp_queue_destroy() must not be called when it is not empty. So maybe ODP API is problematic, or maybe it is ok and basically requires an application to either wait for the aggregation timeout (if there is one) before queue destroy, or carefully send just the right number of special events to the queue (e.g. one by one) so that the aggregator gets flushed and the application then knows it became empty.

platform/linux-generic/odp_queue_basic.c
```diff
@@ -666,6 +996,31 @@ static int queue_info(odp_queue_t handle, odp_queue_info_t *info)
 	return 0;
 }
 
+static void event_aggr_print(queue_entry_t *aggr_queue)
```
JannePeltonen, line 1062: I think it would be nice if this would print also the number of events currently inside the aggregator. This would have to take the aggregator lock to read it. But up you you if you want to add now it or leave it for the future.

platform/linux-generic/odp_queue_basic.c
```diff
@@ -688,6 +735,11 @@ int _odp_event_aggr_enq(queue_entry_t *aggr_queue, _odp_event_hdr_t *event_hdr[]
 
 	_ODP_ASSERT(aggr_queue->type == ODP_QUEUE_TYPE_AGGR);
 
+	/* Packet input may later drop a vector if the base queue is full, so enqueue events to the
+	 * base queue directly. */
+	if (odp_unlikely(base_queue->pktin.pktio != ODP_PKTIO_INVALID))
+		return base_queue->enqueue_multi(base_queue->handle, event_hdr, num);
```
JannePeltonen, line 745: I suppose this special case here would not be necessary here if the enqueue functions for aggregators of pktin queues would be the enqueue function of the base queue. I do not immediately see why not. But ok to leave that as a future improvement.

test/validation/api/classification/odp_classification_common.c
JannePeltonen: Maybe the aggregator validation configure a small enough timeout and skip the tests altogether if timout is not supported by the underlying implementation?

test/validation/api/pktio/pktio.c
MatiasElo: `flush_input_queue()` functions is being called before closing the pktio device, which should take care of this.

platform/linux-generic/include/odp_classification_internal.h
```diff
@@ -143,8 +163,17 @@ static inline void _odp_cos_enq(uint16_t cos_id, odp_queue_t dst, odp_packet_t p
 
 	cos_t *cos = _odp_cos_entry_from_idx(cos_id);
 
-	if (num < 2 || !cos->vector.enable) {
-		int ret = odp_queue_enq_multi(dst, (odp_event_t *)packets, num);
+	if (num < 2 || !cos->vector.enabled) {
+		/* In case of event aggregator queues, always use base queue as destination to
+		 * enable passing classifier validation tests. This is necessary as the current
+		 * event aggregation implementation doesn't support timeouts. */
```
JannePeltonen: This feels somehow backwards: Since validation tests (which are introcudes in this same commit) make some wrong assumptions about implementations, we are adding hacks in the implementation to make the tests pass. I think it would be better to keep the implementation clean and fix whatever issues there are in the validation tests (e.g. skip tests that essentially assume timeout if timeout not supported).

test/validation/api/queue/queue.c
JannePeltonen: Maybe, for now, you could add a comment that we assume that odp_queue_destroy() can be called if the aggregator is not empty. We could add some flushing code (but it is a bit of a hassle) later, or maybe have the API state something about aggregators when a queue is being destroyed.

test/validation/api/pktio/pktio.c
JannePeltonen: But if the events are in the aggregator queue, since there is no timeout, the dequeue/schedule call in flush_input_queue() would not get them (in general, but it depends on the implementation of course). I think the only somewhat reliable way to make sure the aggregator queue is empty would be to send marked events to it one-by-one, and waiting for a vector to arrive that contains the last sent event. But even this assumes that when the aggregator creates a vector, it puts all events in it and does not leave any in the aggregator. Or am I confused now?

MatiasElo: I think the test is currently doing all what could realistically be expected from an applications. Real applications cannot know if there are still pending events in an input queue aggregator, so all they can do is to stop the interface, try to receive all available events, and then close the interface. Perhaps we could add an explicit flush API at some point, but I think we can now let implementations take care of the possible clean up.

test/validation/api/queue/queue.c
MatiasElo: Yeah, adding a comment to the test for now.

test/validation/api/pktio/pktio.c
JannePeltonen: Agreed that there is no point in adding elaborate flushing scheme before queue destroy in the test app now. I think some API clarification is needed.

The orignal issue in this comment thread was about flushing the aggregator from the test packets (not to make it fully empty) during the test and that has been done, so I am closing this as resolved.

platform/linux-generic/odp_queue_basic.c
MatiasElo: True, will be added in v6.

platform/linux-generic/odp_queue_basic.c
MatiasElo: There is a "direct receive" optimization for pktio polling, which skips the base queue altogether. It can be checked later if something could be done to this.

platform/linux-generic/include/odp_classification_internal.h
MatiasElo: Removed in v6.

test/validation/api/classification/odp_classification_common.c
MatiasElo: Test skipped if timeouts are not supported in v6.

platform/linux-generic/include/odp_classification_internal.h
JannePeltonen: Looks good now.

If queue_grp_tbl (or similar) contained the queue handles used for enqueueing, then the enqueue function would not need to check the use_aggr flag and get the aggregator handle. This would require storing the base queue handles too somewhere in the CoS. But this would be just an optimization, so ok to leave if for the future.

test/validation/api/classification/odp_classification_tests.c
```diff
@@ -971,6 +1104,13 @@ odp_testinfo_t classification_suite[] = {
 odp_testinfo_t classification_suite_pktv[] = {
 	ODP_TEST_INFO_CONDITIONAL(cls_pktio_configure_pktv, check_pktv),
 	ODP_TEST_INFO_CONDITIONAL(cls_pktio_test_pktv, check_pktv),
-	ODP_TEST_INFO_CONDITIONAL(cls_queue_stats_pktv, check_queue_stats),
+	ODP_TEST_INFO_CONDITIONAL(cls_queue_stats_pktv, check_pktv_queue_stats),
+	ODP_TEST_INFO_NULL,
+};
+
+odp_testinfo_t classification_suite_evv[] = {
+	ODP_TEST_INFO_CONDITIONAL(cls_pktio_configure_evv, check_evv),
```
JannePeltonen, line 1118: It seems this test could be run even when aggregation timeouts are not supported.

platform/linux-generic/odp_queue_basic.c
```diff
@@ -677,6 +677,53 @@ static odp_queue_t queue_lookup(const char *name)
...
+	odp_event_vector_t evv;
+	uint32_t num_aggr = 0;
+
+	_ODP_ASSERT(qentry_from_handle(aggr_handle)->type == ODP_QUEUE_TYPE_AGGR);
+	_ODP_ASSERT(num <= aggr->max_size);
+
+	odp_ticketlock_lock(&aggr->lock);
+
+	while (num_aggr < num && aggr->num_events < aggr->max_size)
+		aggr->event_tbl[aggr->num_events++] = odp_packet_to_event(pkt_tbl[num_aggr++]);
```
JannePeltonen, line 697: Here the aggregator queue is fully filled, unlike in the regular aggregator enqueue function which always leaves room for at least one event in the queue. This is not a problem because the regular enqueue function skips aggregation for pktin queues.  But this feels a bit dangerous and would break if the regular enqueue function would start using the aggregation (e.g. for plain queues as opposed to sched queues). It would be nice the the max fill level of the aggregation queue would be an invariant that would be the same for all kinds of queues.

MatiasElo: Good catch, fixed in v7.

test/validation/api/classification/odp_classification_tests.c
MatiasElo: Updated commit message in v7.

JannePeltonen: Looks good in v8

PR #2262, opened 2025-09-22
[PATCH v3] api: barrier: clarify thread barrier reuse

JannePeltonen:
Explicitly state that a memory barrier can be reused without initializing it again.

Make it clear that no more than N threads may be inside odp_barrier_wait() for a barrier of N threads. Without this requirement, an application could e.g. wait a barrier in thread 1 and thread 2, and, when thread 2 returns from the barrier (at which point thread 1 is supposed to have been unblocked too), tell threads 3 and 4 to wait for the same barrier. If threads 3 and 4 manage to start waiting for the barrier before thread 1 has returned from odp_barrier_wait(), thread 1 may continue to be blocked in the counter based implementation used in multiple ODPs.

Clarify that a barrier must be initialized before first use and can be reinitialized if not in use. Make also a remark about the thread-unsafety of the initialization function.

### DIFF
--- a/test/validation/api/barrier/barrier.c
+++ b/test/validation/api/barrier/barrier.c
@@ -1,6 +1,6 @@
 /* SPDX-License-Identifier: BSD-3-Clause
  * Copyright (c) 2014-2018 Linaro Limited
- * Copyright (c) 2022 Nokia
+ * Copyright (c) 2022-2025  Nokia
  */
 
 #include <malloc.h>
@@ -12,9 +12,11 @@
 #define VERBOSE			0
 #define MAX_WORKERS		32
 #define BARRIER_ITERATIONS	64
+#define REUSE_ITERATIONS	500
 
 #define SLOW_BARRIER_DELAY	400
 #define BASE_DELAY		6
+#define REUSE_DELAY		10000
 
 #define NUM_TEST_BARRIERS	BARRIER_ITERATIONS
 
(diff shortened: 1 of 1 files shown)


include/odp/api/spec/barrier.h
```diff
@@ -36,18 +37,36 @@ extern "C" {
 /**
  * Initialize barrier with thread count.
  *
+ * This function must not be called by multiple threads simultaenously for
```
MatiasElo: Typo: `simultaenously`

include/odp/api/spec/barrier.h
```diff
@@ -36,18 +37,36 @@ extern "C" {
...
+ *
+ * A barrier can be reused without reinitializing it again. It is ok to call
+ * odp_barrier_wait() for the same barrier again immediately after returning
+ * from the previous odp_barrier_wait() even if other threads that were waiting
+ * for the barrier have not yet returned from odp_barrier_wait().
+ *
+ * If a barrier is used by more threads than the thread count the barrier was
+ * initialized with, the number of threads that have called odp_barrier_wait()
+ * but not yet returned from it must never exceed the thread count the barrier
+ * was initiazed with.
```
MatiasElo: Typo: `initiazed`

include/odp/api/spec/barrier.h
```diff
@@ -36,18 +37,36 @@ extern "C" {
...
+ *
  * @param barr Pointer to a barrier variable
  * @param count Thread count
  */
 void odp_barrier_init(odp_barrier_t *barr, int count);
 
 /**
  * Synchronize thread execution on barrier.
- * Wait for all threads to arrive at the barrier until they are let loose again.
+ * Wait for a number of threads to arrive at the barrier until they are
```
MatiasElo, line 52: Could add also a newline to separate the one-line description.

include/odp/api/spec/barrier.h
```diff
@@ -36,18 +37,36 @@ extern "C" {
...
  * Threads will block (spin) until the last thread has arrived at the barrier.
  * All memory operations before the odp_barrier_wait() call will be visible
  * to all threads when they leave the barrier.
  *
+ * A barrier must be initialized using odp_barrier_init() before first use.
+ *
+ * A barrier can be reused without reinitializing it again. It is ok to call
+ * odp_barrier_wait() for the same barrier again immediately after returning
+ * from the previous odp_barrier_wait() even if other threads that were waiting
+ * for the barrier have not yet returned from odp_barrier_wait().
```
MatiasElo, line 64: Do we have a validation test already for this? If not, one should be added.

JannePeltonen: We have multiple tests that rely on this, so in a way we have, but maybe it would be good add an explicit test.

MatiasElo: Yep, explicit test seems worth adding.

MatiasElo (approved):

PR #2263, opened 2025-09-22
test: bench_timer: add tests for timer start and cancel

TuomasTaipale:
Add new tests for (oneshot) timer start and cancel. Additionally, add new mode option for optionally tracking function minimum and maximum on top of average execution times.

### DIFF
--- a/test/performance/odp_bench_timer.c
+++ b/test/performance/odp_bench_timer.c
@@ -32,37 +32,65 @@
 /* Default number of rounds per test case */
 #define ROUNDS 1000u
 
-/** User area size in bytes */
+/* User area size in bytes */
 #define UAREA_SIZE 8
 
-/** Timer duration in nsec */
+/* Timer duration in nsec */
 #define TIMER_NSEC 50000000
 
-/** Maximum number of results to be held */
+/* Maximum number of results to be held */
 #define TEST_MAX_BENCH 20
 
 #define BENCH_INFO(run_fn, max, alt_name) \
 	{.name = #run_fn, .run = run_fn, .max_rounds = max, .desc = alt_name}
+
+#define BENCH_INFO_TM(run_fn) \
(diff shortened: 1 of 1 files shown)


test/performance/odp_bench_timer.c
```diff
@@ -601,6 +927,42 @@ static int wait_timer(void)
 	return 0;
 }
 
+static void init_suite(gbl_args_t *gbl_args, odp_bool_t is_export)
```
MatiasElo: `is_export` seems kinda unnecessary, the `result` pointer could be initialized always.

test/performance/odp_bench_timer.c
```diff
@@ -297,23 +738,47 @@ static int bench_timer_export(void *data)
...
+	BENCH_INFO_TM(timer_to_u64_tm, NULL, NULL),
+	BENCH_INFO_TM(timer_pool_to_u64_tm, NULL, NULL),
+	BENCH_INFO_TM(timer_start_tm, NULL, cancel_timers),
+	BENCH_INFO_TM(timer_cancel_tm, start_timers, NULL),
+	BENCH_INFO_TM(timer_start_cancel_tm, NULL, NULL),
+};
+
+ODP_STATIC_ASSERT(ODPH_ARRAY_SIZE(test_suite_tm) < TEST_MAX_BENCH,
+		  "Result array is too small to hold all the results");
+
```
MatiasElo, line 781: Could add second assert for verifying that `test_suite` and `test_suite_tm` sizes match.

MatiasElo (approved):
Looks good. One optional comment.

test/performance/odp_bench_timer.c
TuomasTaipale: There are a couple of bench testers now with these both so probably better to add the check in each of them at the same time in a separate PR (I guess there could be situations though where using one of the suites is feasible for some scenario but the other suite isn't, so the counts could differ).

PR #2264, opened 2025-09-22
[PATCH v2] Fix odp_packet_trunc_tail() with more than 2 segments

JannePeltonen:
Fix odp_packet_trunc_tail() to set packet tailroom after trucation to the tailroom of the last remaining segment, not the first one.

Add one validation test case for this specific case.

### DIFF
--- a/test/validation/api/packet/packet.c
+++ b/test/validation/api/packet/packet.c
@@ -1,6 +1,6 @@
 /* SPDX-License-Identifier: BSD-3-Clause
  * Copyright (c) 2014-2018 Linaro Limited
- * Copyright (c) 2019-2024 Nokia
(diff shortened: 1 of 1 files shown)


platform/linux-generic/odp_packet.c
MatiasElo, line 1: Typo in commit message: `more that`

test/validation/api/packet/packet.c
MatiasElo, line 1: Copyright can be updated.

platform/linux-generic/odp_packet.c
JannePeltonen: v2

test/validation/api/packet/packet.c
JannePeltonen: v2

PR #2269, opened 2025-09-24
[PATCH v2] test: sched_latency: add wait option to simulate work

MatiasElo:
Add '-W, --wait-ns <num>' option to odp_sched_latency test application to simulate application work between schedule and queue enqueue calls. Wait time is defined in nanoseconds.

### DIFF
--- a/test/performance/odp_sched_latency.c
+++ b/test/performance/odp_sched_latency.c
@@ -87,6 +87,7 @@ typedef struct {
 						  test events */
 	} prio[NUM_PRIOS];
 	odp_bool_t sample_per_prio; /**< Allocate a separate sample for each priority */
+	uint64_t wait_ns;	/**< Number of nsec to wait after scheduling events */
 } test_args_t;
 
 /** Latency measurements statistics */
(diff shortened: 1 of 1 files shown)


test/performance/odp_sched_latency.c
```diff
@@ -835,6 +844,8 @@ static void usage(void)
 	       "               1: ODP_SCHED_SYNC_ATOMIC\n"
 	       "               2: ODP_SCHED_SYNC_ORDERED\n"
 	       "  -w, --warm-up <number> Number of warm-up rounds, default=100, min=1\n"
+	       "  -W, --wait-ns <number> Number of nsec to wait per schedule round before enqueueing events.\n"
```
JannePeltonen: It seems that the wait is not before enqueing but after scheduling, with statistics collection and whatnot between the wait and the enqueuing. Dunno if any particular placement of the waiting is preferable to any other, but maybe the description could better match the placement?

Otherwise looks good to me.

MatiasElo: Fixed in V2.

JannePeltonen (approved):

PR #2271, opened 2025-09-26
validation: timer: add new tests

TuomasTaipale:
Add new tests to check ability to create maximum amount of oneshot and periodic timer pools, maximum/large amount of timers from these pools and ability for implementations to handle timer pools with varying parameterization.

### DIFF
--- a/test/validation/api/timer/timer.c
+++ b/test/validation/api/timer/timer.c
@@ -1465,6 +1465,80 @@ static void timer_pool_tick_info(void)
 	odp_timer_pool_destroy(tp);
 }
 
+static void timer_alloc_max(void)
+{
+	odp_timer_capability_t capa;
+	odp_shm_t shm_tmr;
+	odp_timer_pool_t pool;
+	odp_timer_t *timer;
+	odp_timer_pool_param_t tp_param;
+	odp_queue_param_t queue_param;
+	odp_queue_t queue;
+	/* Use some sane value in case of memory-capped or very large max timer count. */
+	uint32_t num = MAX_TIMER_POOLS;
+	int ret;
+	uint64_t tmo_ns = ODP_TIME_SEC_IN_NS;
+	uint64_t res_ns = ODP_TIME_SEC_IN_NS / 10;
+	odp_timer_clk_src_t clk_src = test_global->clk_src;
+
+	ret = odp_timer_capability(clk_src, &capa);
+
+	CU_ASSERT_FATAL(ret == 0);
+
+	if (capa.max_timers > 0 && num > capa.max_timers)
+		num = capa.max_timers;
+
+	shm_tmr = odp_shm_reserve("test_max_alloc", num * sizeof(*timer), ODP_CACHE_LINE_SIZE, 0);
+
+	CU_ASSERT_FATAL(shm_tmr != ODP_SHM_INVALID);
+
+	timer = odp_shm_addr(shm_tmr);
+
+	CU_ASSERT_FATAL(timer != NULL);
+
+	if (capa.max_tmo.max_tmo < tmo_ns) {
+		tmo_ns = capa.max_tmo.max_tmo;
+		res_ns = capa.max_tmo.res_ns;
+	}
+
+	odp_queue_param_init(&queue_param);
+
+	if (capa.queue_type_sched)
+		queue_param.type = ODP_QUEUE_TYPE_SCHED;
+
+	queue = odp_queue_create("test_max_alloc", &queue_param);
+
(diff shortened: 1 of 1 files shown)


test/validation/api/timer/timer.c
```diff
@@ -1465,6 +1465,80 @@ static void timer_pool_tick_info(void)
...
+{
+	odp_timer_capability_t capa;
+	odp_shm_t shm_tmr;
+	odp_timer_pool_t pool;
+	odp_timer_t *timer;
+	odp_timer_pool_param_t tp_param;
+	odp_queue_param_t queue_param;
+	odp_queue_t queue;
+	/* Use some sane value in case of memory-capped or very large max timer count. */
+	uint32_t num = MAX_TIMER_POOLS;
```
MatiasElo: A bit confusing define to use here. There is also `NTIMERS` define or the default value could be simply defined here.

test/validation/api/timer/timer.c
```diff
@@ -1465,6 +1465,80 @@ static void timer_pool_tick_info(void)
...
+	if (capa.queue_type_sched)
+		queue_param.type = ODP_QUEUE_TYPE_SCHED;
+
+	queue = odp_queue_create("test_max_alloc", &queue_param);
+
+	CU_ASSERT_FATAL(queue != ODP_QUEUE_INVALID);
+
+	odp_timer_pool_param_init(&tp_param);
+	tp_param.res_ns = res_ns;
+	tp_param.min_tmo = tmo_ns / 2;
```
MatiasElo: Could check also `min_tmo` capability.

test/validation/api/timer/timer.c
```diff
@@ -3255,6 +3329,92 @@ static void timer_test_periodic_pool_create_max(void)
...
+	odp_shm_t shm_tmr;
+	odp_timer_periodic_capability_t periodic_capa;
+	odp_timer_pool_param_t timer_param;
+	odp_timer_pool_t pool;
+	odp_timer_t *timer;
+	odp_queue_param_t queue_param;
+	odp_queue_t queue;
+	double min_freq, max_freq;
+	/* Use some sane value in case of very large max timer count. */
+	uint32_t num = MAX_TIMER_POOLS;
```
MatiasElo: A bit confusing define to use here.

test/validation/api/timer/timer.c
```diff
@@ -3255,6 +3329,92 @@ static void timer_test_periodic_pool_create_max(void)
...
+
+	CU_ASSERT_FATAL(pool != ODP_TIMER_POOL_INVALID);
+	CU_ASSERT_FATAL(odp_timer_pool_start_multi(&pool, 1) == 1);
+
+	odp_queue_param_init(&queue_param);
+
+	if (timer_capa.queue_type_sched)
+		queue_param.type = ODP_QUEUE_TYPE_SCHED;
+
+	queue = odp_queue_create("test_max_alloc", &queue_param);
```
MatiasElo: Could have `CU_ASSERT_FATAL` check. Queue name could also be updated.

test/validation/api/timer/timer.c
```diff
@@ -1465,6 +1465,139 @@ static void timer_pool_tick_info(void)
...
+	time_limit = odp_time_local_strict_ns() + res_ns * steps;
+
+	while (odp_time_local_strict_ns() < time_limit) {
+		ev = capa.queue_type_sched ? odp_schedule(NULL, ODP_SCHED_NO_WAIT) :
+					     odp_queue_deq(queue);
+
+		if (ev == ODP_EVENT_INVALID)
+			continue;
+
+		tmr = odp_timeout_timer(odp_timeout_from_event(ev));
```
MatiasElo, line 1583: Could have an assert also for event type.

test/validation/api/timer/timer.c
TuomasTaipale: Yes, agree, first commit in v2 now removes `MAX_TIMER_POOLS` constant (as no more references to it) and uses `NTIMERS` instead.

test/validation/api/timer/timer.c
TuomasTaipale: There were a couple styles used to parameterize the pool, swapped to use the ones in `global_mem->param.*` in v2.

test/validation/api/timer/timer.c
TuomasTaipale: v2.

test/validation/api/timer/timer.c
TuomasTaipale: v2.

PR #2273, opened 2025-09-26
[PATCH v4] Fix packet references in loop and dpdk pktios and improve testing

JannePeltonen:
Fix sending of packet references in loop pktio and in dpdk pktio with zero copy enabled. Improve also testing of pktios with packets that have references.

### DIFF
--- a/test/validation/api/pktio/pktio.c
+++ b/test/validation/api/pktio/pktio.c
@@ -32,7 +32,7 @@
 #define PKTV_DEFAULT_SIZE      8
 #define MAX_QUEUES             128
 
-#define PKTIO_TS_INTERVAL      (50 * ODP_TIME_MSEC_IN_NS)
+#define PKTIO_TS_INTERVAL      (10 * ODP_TIME_MSEC_IN_NS)
 #define PKTIO_TS_MIN_RES       1000
 #define PKTIO_TS_MAX_RES       10000000000
 


platform/linux-generic/pktio/loop.c
```diff
@@ -545,29 +545,42 @@ static int loopback_send(pktio_entry_t *pktio_entry, int index, const odp_packet
...
 
 		if (odp_unlikely(pkt_len > pkt_loop->mtu)) {
 			if (nb_tx == 0)
 				return -1;
 			break;
 		}
 
+		if (odp_unlikely(refcount > 1)) {
+			/* Cannot send the original packet, so send a copy */
+			pkt = odp_packet_copy(pkt, odp_packet_pool(pkt));
```
TuomasTaipale: I wonder if this "hidden" copy messes with dimensioning. Alternatively, could use the currently unused packet I/O pool. In practice probably a non-issue.

test/validation/api/pktio/pktio.c
```diff
@@ -4165,12 +4246,17 @@ static void pktio_test_chksum_sctp(void (*config_fn)(odp_pktio_t, odp_pktio_t),
 		if (prep_fn)
 			prep_fn(pkt_tbl[i]);
 
+	if (test_flags & TEST_WITH_REFS)
+		make_refs(ref_tbl, pkt_tbl, TX_BATCH_LEN);
 	send_packets(pktout_queue, pkt_tbl, TX_BATCH_LEN);
```
TuomasTaipale, line 4338: Unrelated to patch, if you want to change this as well: `send_packets()` and `send_packet_events()` return a value which callers aren't checking (well, once) as the fuctions already fatally fail internally. Could be `void` returning functions with the one return value check changed.

platform/linux-generic/pktio/loop.c
JannePeltonen: Actually, the pool here should be the pool configured for the loop interface at odp_pktio_open(). And consuming packets from that pool is not a problem but expected. Now I realize that the current loop interface code does not honour the configured pool, which is a bug that affects non-references packets too. I'll probably add a fix for that too in the next version of this PR.

test/validation/api/pktio/pktio.c
JannePeltonen: Maybe I won't touch that. Doing CU_ASSERT_FATAL there is slightly problematic since then the main test function does not have a chance to clean up the resources it has allocated. Many tests have the same issue and, ideally, the test would either not fail fatally when holding resources (like packets, opened pktios etc) or have the setup and teardown stuff moved to test init and term functions. But that would require lots of changes for a questionable benefit.

platform/linux-generic/pktio/loop.c
```diff
@@ -1,6 +1,6 @@
 /* SPDX-License-Identifier: BSD-3-Clause
  * Copyright (c) 2013-2018 Linaro Limited
- * Copyright (c) 2013-2023 Nokia Solutions and Networks
+ * Copyright (c) 2013-2025 Nokia
```
TuomasTaipale: I think we agreed a while back that we don't change the existing names in copyright entries, so probably just bump up the year or add a new entry for "Nokia".

platform/linux-generic/pktio/loop.c
```diff
@@ -545,29 +552,41 @@ static int loopback_send(pktio_entry_t *pktio_entry, int index, const odp_packet
 		num = QUEUE_MULTI_MAX;
 
 	for (i = 0; i < num; ++i) {
-		uint32_t pkt_len = odp_packet_len(pkt_tbl[i]);
+		odp_packet_t pkt = pkt_tbl[i];
+		uint32_t pkt_len = odp_packet_len(pkt);
+		int copied = 0;
 
 		if (odp_unlikely(pkt_len > pkt_loop->mtu)) {
```
TuomasTaipale, line 560: Old code, but could use `odp_packet_len()` here directly.

platform/linux-generic/pktio/loop.c
```diff
@@ -114,8 +116,13 @@ static int loopback_open(odp_pktio_t id ODP_UNUSED, pktio_entry_t *pktio_entry,
 		return -1;
 	}
 
+	if (pool == ODP_POOL_INVALID ||
+	    _odp_pool_entry(pool)->type != ODP_POOL_PACKET)
```
TuomasTaipale: Would be nice to have some logging in new failure cases (here and "send()").

test/validation/api/pktio/pktio.c
```diff
@@ -54,6 +54,14 @@ static bool wait_for_network;
 /* Dummy global variable to avoid compiler optimizing out API calls */
```
TuomasTaipale, line 54: Typo in commit? 

... interfaces *with* -> *which* might try ...

test/validation/api/pktio/pktio.c
```diff
@@ -1101,6 +1130,14 @@ static void test_txrx(odp_pktin_mode_t in_mode, int num_pkts,
 	}
 }
 
+static void test_txrx(odp_pktin_mode_t in_mode, int num_pkts,
+		      txrx_mode_e mode, odp_schedule_sync_t sync_mode,
+		      odp_bool_t vector_mode)
+{
+	for (uint32_t flags = 0; flags < NUM_TEST_FLAG_COMBOS; flags++)
+		do_test_txrx(in_mode, num_pkts, mode, sync_mode, vector_mode, flags);
```
TuomasTaipale, line 1165: If it's not evident from other contexts, could be useful to print which test flags were used for this particular test run.

platform/linux-generic/pktio/dpdk.c
```diff
@@ -1086,11 +1086,12 @@ static inline int pkt_to_mbuf_zero(pktio_entry_t *pktio_entry,
 		odp_packet_hdr_t *pkt_hdr = packet_hdr(pkt);
 		struct rte_mbuf *mbuf = mbuf_from_pkt_hdr(pkt_hdr);
 		uint16_t pkt_len = odp_packet_len(pkt);
+		uint32_t refcount = odp_atomic_load_u32(&pkt_hdr->ref_cnt);
 
 		if (odp_unlikely(pkt_len > mtu))
 			goto fail;
 
-		if (odp_likely(pkt_hdr->seg_count == 1)) {
+		if (odp_likely(pkt_hdr->seg_count == 1 && refcount <= 1)) {
```
TuomasTaipale: `odp_atomic_load_u32()` here directly.

platform/linux-generic/pktio/loop.c
```diff
@@ -554,14 +554,16 @@ static int loopback_send(pktio_entry_t *pktio_entry, int index, const odp_packet
...
 		int copied = 0;
 
 		if (odp_unlikely(pkt_len > pkt_loop->mtu)) {
 			if (nb_tx == 0)
 				return -1;
 			break;
 		}
-		if (odp_unlikely(odp_packet_pool(pkt) != pkt_loop->pool)) {
+		if (odp_unlikely(odp_packet_pool(pkt) != pkt_loop->pool ||
+				 refcount > 1)) {
```
TuomasTaipale: `odp_atomic_load_u32()` here directly.

platform/linux-generic/pktio/loop.c
JannePeltonen: Just bumped up the year in v3.

platform/linux-generic/pktio/loop.c
JannePeltonen: Did not change this since pkt_len is used also for statistics addition later.

platform/linux-generic/pktio/loop.c
JannePeltonen: Added an error message about invalid pool to loopback_open() in v3. Did not add error messages to the send function since packet copying may fail because of temporary shortage of packet buffers in the configured default pool, which is not really an error and a retried send may just work. An error log in this case would just cause problems (log fill-up and wasted cycles).

test/validation/api/pktio/pktio.c
JannePeltonen: Fixed in v3.

test/validation/api/pktio/pktio.c
JannePeltonen: I did not change anything in v3 as it is not easy to add a useful print. CUnit collects assertion failures and displays them after each test, so they cannot be correlated with the printouts made during a test (i.e. even with test flag prints added, one cannot see what the test flag values were in a particular assertion failure).

platform/linux-generic/pktio/dpdk.c
JannePeltonen: Not sure if it is any better, but changed it in v3.

platform/linux-generic/pktio/loop.c
JannePeltonen: Ditto.

platform/linux-generic/pktio/loop.c
TuomasTaipale: Could be a `_ODP_DBG()` like with the enqueue error later.

test/validation/api/pktio/pktio.c
TuomasTaipale: IPsec validation tests dump algorithm combos before execution (which was the motivation for my comment). So those are not useful prints?

platform/linux-generic/pktio/loop.c
JannePeltonen: Added a debug print in v4.

test/validation/api/pktio/pktio.c
JannePeltonen: There is the same problem there. I did not add flag printing in v4 because there may be better ways to improve the visibility of the current sub test or test phase in assertion failures. But that would be a separate PR.

TuomasTaipale (approved):

PR #2274, opened 2025-09-29
[PATCH v4] validation: event: add tests for event type compliancy

Rayska:
- Add tests to ensure APIs which take odp_event_t handles can process all supported event types
- Implemented for all event types except ODP_EVENT_IPSEC_STATUS, ODP_EVENT_PACKET_TX_COMPL, ODP_EVENT_ANY

### DIFF
--- a/test/validation/api/event/event.c
+++ b/test/validation/api/event/event.c
@@ -12,7 +12,7 @@
 #define NUM_EVENTS  100
 #define EVENT_SIZE  100
 #define EVENT_BURST 10
-#define MAX_EVENT_TYPES 6
+#define MAX_EVENT_TYPES 7
 
 typedef struct {
 	struct {
@@ -50,6 +50,7 @@ static odp_event_t event_types_alloc_event(odp_pool_type_t pool_type)
 	odp_event_vector_t evv;
 	odp_packet_vector_t pv;
 	odp_dma_compl_t dma_compl;
+	odp_ml_compl_t ml_compl;
 
 	if (idx < 0)
 		return ODP_EVENT_INVALID;
@@ -85,6 +86,11 @@ static odp_event_t event_types_alloc_event(odp_pool_type_t pool_type)
 		if (dma_compl == ODP_DMA_COMPL_INVALID)
 			return ODP_EVENT_INVALID;
 		return odp_dma_compl_to_event(dma_compl);
+	case ODP_POOL_ML_COMPL:
+		ml_compl = odp_ml_compl_alloc(ctx->type[idx].pool);
+		if (ml_compl == ODP_ML_COMPL_INVALID)
+			return ODP_EVENT_INVALID;
+		return odp_ml_compl_to_event(ml_compl);
 	default:
 		return ODP_EVENT_INVALID;
 	}
@@ -140,6 +146,7 @@ static int event_types_suite_init(void)
 	odp_queue_param_t queue_prm;
 	odp_pool_capability_t p_capa;
 	odp_dma_capability_t dma_capa;
+	odp_ml_capability_t ml_capa;
 
 	if (odp_queue_capability(&queue_capa) != 0)
 		return -1;
@@ -159,6 +166,9 @@ static int event_types_suite_init(void)
 		return -1;
 
 	if (odp_dma_capability(&dma_capa) != 0)
+		return -1;
+
+	if (odp_ml_capability(&ml_capa) != 0)
 		return -1;
 
 	ctx->type[n].event_type = ODP_EVENT_BUFFER;
@@ -186,11 +196,19 @@ static int event_types_suite_init(void)
 		ctx->type[n++].pool_type = ODP_POOL_DMA_COMPL;
 	}
 
+	if (ml_capa.max_models &&
+	    ((ml_capa.load.compl_mode_mask | ml_capa.run.compl_mode_mask) &
+	    ODP_ML_COMPL_MODE_EVENT) && ml_capa.pool.max_num > 0) {
+		ctx->type[n].event_type = ODP_EVENT_ML_COMPL;
+		ctx->type[n++].pool_type = ODP_POOL_ML_COMPL;
+	}
+
 	ctx->num_types = n;
 
 	for (i = 0; i < ctx->num_types; i++) {
 		odp_pool_param_t prm;
 		odp_dma_pool_param_t dma_prm;
+		odp_ml_compl_pool_param_t ml_prm;
 
 		odp_pool_param_init(&prm);
 
@@ -229,6 +247,10 @@ static int event_types_suite_init(void)
 						      dma_capa.pool.max_uarea_size);
 		break;
 		case ODP_POOL_ML_COMPL:
+			odp_ml_compl_pool_param_init(&ml_prm);
+			ml_prm.num = NUM_EVENTS;
+			ml_prm.uarea_size = ODPH_MIN(sizeof(uint64_t),
+						     ml_capa.pool.max_uarea_size);
 		break;
 		}
 
@@ -236,6 +258,8 @@ static int event_types_suite_init(void)
 
 		if (ctx->type[i].pool_type == ODP_POOL_DMA_COMPL)
 			ctx->type[i].pool = odp_dma_pool_create(name, &dma_prm);
+		else if (ctx->type[i].pool_type == ODP_POOL_ML_COMPL)
+			ctx->type[i].pool = odp_ml_compl_pool_create(name, &ml_prm);
 		else
 			ctx->type[i].pool = odp_pool_create(name, &prm);
 
@@ -344,7 +368,8 @@ static void event_test_pool(void)
 			ev_pool = odp_event_pool(ev);
 
 			if (exp_type == ODP_EVENT_TIMEOUT ||
-			    exp_type == ODP_EVENT_DMA_COMPL)
+			    exp_type == ODP_EVENT_DMA_COMPL ||
+			    exp_type == ODP_EVENT_ML_COMPL)
 				CU_ASSERT(ev_pool == ODP_POOL_INVALID);
 			else
 				CU_ASSERT(ev_pool == ctx->type[i].pool);
@@ -376,6 +401,7 @@ static void event_test_user_area_flag(void)
 			case ODP_EVENT_BUFFER:
 			case ODP_EVENT_TIMEOUT:
 			case ODP_EVENT_DMA_COMPL:
+			case ODP_EVENT_ML_COMPL:
 				CU_ASSERT(flag < 0);
 				break;
 			case ODP_EVENT_PACKET:


test/validation/api/event/event.c
```diff
@@ -11,6 +11,197 @@
...
+}
+
+static int event_types_suite_init(void)
+{
+	event_type_ctx_t *ctx = &g_type_ctx;
+	int i, j, n = 0;
+	char name[ODP_POOL_NAME_LEN];
+	odp_pool_capability_t capa;
+
+	memset(&capa, 0, sizeof(capa));
```
MatiasElo: Not necessary

test/validation/api/event/event.c
```diff
@@ -447,8 +638,15 @@ odp_testinfo_t event_suite[] = {
 	ODP_TEST_INFO_NULL,
 };
 
+/* New EventTypes suite using the shared init/term above */
```
MatiasElo: Could be either removed or describe what this suite is intended for.

test/validation/api/event/event.c
```diff
@@ -11,6 +11,197 @@
...
 #define EVENT_BURST 10
+#define MAX_EVENT_TYPES 5
+
+typedef struct {
+	odp_event_type_t event_types[MAX_EVENT_TYPES];
+	odp_pool_type_t pool_types[MAX_EVENT_TYPES];
+	odp_pool_t pools[MAX_EVENT_TYPES];
+	odp_event_t events[MAX_EVENT_TYPES][EVENT_BURST];
+	int num_types;
+} event_type_ctx_t;
```
MatiasElo, line 27: Maybe this would be clearer:
```
typedef struct {
	struct {
		odp_event_type_t event_type;
		odp_pool_type_t pool_type;
		odp_pool_t pool;
		odp_event_t events[EVENT_BURST];
	} type[MAX_EVENT_TYPES];
	int num_types;
	odp_queue_t queue;
} event_type_ctx_t;
```

test/validation/api/event/event.c
```diff
@@ -11,6 +11,197 @@
...
+		odp_pool_param_t prm;
+
+		odp_pool_param_init(&prm);
+
+		prm.type = ctx->pool_types[i];
+		switch (ctx->pool_types[i]) {
+		case ODP_POOL_BUFFER:
+			prm.buf.num = NUM_EVENTS;
+			prm.buf.size = EVENT_SIZE;
+			prm.buf.uarea_size = sizeof(uint64_t);
```
MatiasElo: E.g. `prm.buf.uarea_size = ODPH_MIN(sizeof(uint64_t), capa.buf.max_uarea_size);` same for all types.

test/validation/api/event/event.c
```diff
@@ -11,6 +11,197 @@
...
+	int i, j;
+
+	for (i = 0; i < ctx->num_types; i++) {
+		for (j = 0; j < EVENT_BURST; j++) {
+			ev = ctx->events[i][j];
+			exp_type = ctx->event_types[i];
+			ev_pool = odp_event_pool(ev);
+
+			if (exp_type == ODP_EVENT_TIMEOUT) {
+				CU_ASSERT_FATAL(ev_pool == ODP_POOL_INVALID);
```
MatiasElo: Doesn't need to be fatal.

test/validation/api/event/event.c
```diff
@@ -11,6 +11,197 @@
...
+			} else {
+				CU_ASSERT_FATAL(ev_pool != ODP_POOL_INVALID);
+				CU_ASSERT_FATAL(odp_pool_info(ev_pool, &info) == 0);
+				CU_ASSERT(info.type == ctx->pool_types[i]);
+				CU_ASSERT(odp_event_pool(ev) == ev_pool);
```
MatiasElo: Could be simplified to `CU_ASSERT(ev_pool == ctx->pool_types[i]);` and function name changed to `event_test_pool`.

test/validation/api/event/event.c
```diff
@@ -167,6 +177,8 @@ static int event_types_suite_term(void)
 	event_type_ctx_t *ctx = &g_type_ctx;
 	int i;
 
+	odp_queue_destroy(ctx->queue);
+
 	for (i = 0; i < ctx->num_types; i++) {
 		odp_event_free_multi(ctx->events[i], EVENT_BURST);
 		(void)odp_pool_destroy(ctx->pools[i]);
```
MatiasElo: Better check the return values.

test/validation/api/event/event.c
```diff
@@ -176,6 +188,95 @@ static int event_types_suite_term(void)
...
+	int i, j;
+
+	for (i = 0; i < MAX_EVENT_TYPES; i++) {
+		for (j = 0; j < EVENT_BURST; j++) {
+			ev = ctx->events[i][j];
+			switch (ctx->event_types[i]) {
+			case ODP_EVENT_BUFFER:
+				b = odp_buffer_from_event(ev);
+				CU_ASSERT_FATAL(b != ODP_BUFFER_INVALID);
+				CU_ASSERT(odp_buffer_to_event(b) == ev);
```
MatiasElo: An implementation could encode additional metadata to the handle, so this check may not always work. I'd perhaps drop this whole test since these aren't event module APIs anyway.

test/validation/api/event/event.c
```diff
@@ -176,6 +188,95 @@ static int event_types_suite_term(void)
...
+			break;
+			default:
+				CU_FAIL("Unexpected event type");
+			break;
+			}
+		}
+	}
+}
+
+static void event_test_types_and_subtypes_ctx(void)
```
MatiasElo: I'd drop `_ctx` from function names and just use tested function names.

test/validation/api/event/event.c
```diff
@@ -176,6 +188,95 @@ static int event_types_suite_term(void)
...
+
+			CU_ASSERT(odp_event_type(ev) == exp_type);
+			CU_ASSERT(odp_event_subtype(ev) == exp_sub);
+
+			sub = ODP_EVENT_NO_SUBTYPE;
+			CU_ASSERT(odp_event_types(ev, &sub) == exp_type);
+			CU_ASSERT(sub == exp_sub);
+		}
+
+		odp_event_types_multi(ctx->events[i], types_out, subs_out, EVENT_BURST);
```
MatiasElo: Same could be tested also without `subs_out` (NULL).

test/validation/api/event/event.c
```diff
@@ -176,6 +188,95 @@ static int event_types_suite_term(void)
...
+		for (k = 0; k < EVENT_BURST; k++) {
+			CU_ASSERT(types_out[k] == exp_type);
+			CU_ASSERT(subs_out[k] == exp_sub);
+		}
+
+		type = ODP_EVENT_BUFFER;
```
MatiasElo: Separate initialization shouldn't be necessary.

test/validation/api/event/event.c
```diff
@@ -176,6 +188,95 @@ static int event_types_suite_term(void)
...
+		CU_ASSERT(odp_event_type_multi(ctx->events[i], EVENT_BURST, &type)
```
MatiasElo: Could use an array with multiple event types for better test coverage.

test/validation/api/event/event.c
```diff
@@ -203,6 +304,186 @@ static void event_test_pool_info_ctx(void)
...
+	void *user_area;
+
+	for (i = 0; i < MAX_EVENT_TYPES; i++) {
+		for (j = 0; j < EVENT_BURST; j++) {
+			ev = ctx->events[i][j];
+			user_area = odp_event_user_area(ev);
+			CU_ASSERT(user_area != NULL);
+
+			flag = 1;
+			odp_event_user_flag_set(ev, flag);
```
MatiasElo: I'd do `odp_event_user_flag_set(ev, 1);` as current code could miss `odp_event_user_area_and_flag()` not setting flag value correctly.

test/validation/api/event/event.c
```diff
@@ -203,6 +304,186 @@ static void event_test_pool_info_ctx(void)
...
+			CU_ASSERT(odp_event_user_area_and_flag(ev, &flag) != NULL);
+
+			ev_t = ctx->event_types[i];
+			if (ev_t == ODP_EVENT_PACKET)
+				CU_ASSERT(flag > 0);
```
MatiasElo, line 410: Negative value should be returned if event does not have user flag. Event and packet vectors also have user flags.

test/validation/api/event/event.c
```diff
@@ -203,6 +304,186 @@ static void event_test_pool_info_ctx(void)
...
+{
+	event_type_ctx_t *ctx = &g_type_ctx;
+	odp_event_t ev;
+	int i, j;
+
+	for (i = 0; i < MAX_EVENT_TYPES; i++) {
+		for (j = 0; j < EVENT_BURST; j++) {
+			ev = ctx->events[i][j];
+
+			CU_ASSERT(odp_event_to_u64(ev) != 0);
```
MatiasElo: API says nothing about zero value, so this should not be tested. One could check that individual events return different u64 values.

test/validation/api/event/event.c
```diff
@@ -203,6 +304,186 @@ static void event_test_pool_info_ctx(void)
...
+			CU_ASSERT(odp_event_user_area_and_flag(ev, &flag) != NULL);
+
+			ev_t = ctx->event_types[i];
+			if (ev_t == ODP_EVENT_PACKET)
+				CU_ASSERT(flag > 0);
+		}
+	}
+}
+
+static void event_test_misc_ctx(void)
```
MatiasElo: Perhaps this function could be name `event_test_debug` and the flow id APIs would go to a separate function.

test/validation/api/event/event.c
```diff
@@ -203,6 +304,186 @@ static void event_test_pool_info_ctx(void)
...
+		for (j = 0; j < EVENT_BURST; j++) {
+			ev = ctx->events[i][j];
+
+			CU_ASSERT(odp_event_to_u64(ev) != 0);
+			CU_ASSERT(odp_event_is_valid(ev) == 1);
+
+			if (ctx->event_types[i] == ODP_EVENT_PACKET) {
+				odp_event_flow_id_set(ev, 3);
+				CU_ASSERT(odp_event_flow_id(ev) == 3);
+			}
```
MatiasElo: API doesn't prohibit using these APIs for other event types and the initial ID should be zero.

`odp_event_flow_id_set()` API states: `The value of flow id must be less than the number of flows configured in the scheduler.`

test/validation/api/event/event.c
```diff
@@ -203,6 +304,186 @@ static void event_test_pool_info_ctx(void)
...
+
+static void event_test_filter_packet_ctx(void)
+{
+	event_type_ctx_t *ctx = &g_type_ctx;
+	const int n = MAX_EVENT_TYPES * MAX_EVENT_TYPES;
+
+	odp_event_t mixed[MAX_EVENT_TYPES * MAX_EVENT_TYPES];
+	odp_packet_t packets[MAX_EVENT_TYPES * MAX_EVENT_TYPES];
+	odp_event_t remain[MAX_EVENT_TYPES * MAX_EVENT_TYPES];
+	odp_event_t expected[MAX_EVENT_TYPES * MAX_EVENT_TYPES];
```
MatiasElo: Should be `MAX_EVENT_TYPES  * EVENT_BURST`? Same also below.

test/validation/api/event/event.c
```diff
@@ -203,6 +304,186 @@ static void event_test_pool_info_ctx(void)
...
+
+	num_pkt = odp_event_filter_packet(mixed, packets, remain, n);
+	CU_ASSERT(num_pkt == MAX_EVENT_TYPES);
+
+	int pkt_idx = -1;
+
+	for (j = 0; j < MAX_EVENT_TYPES; j++)
+		if (ctx->event_types[j] == ODP_EVENT_PACKET)
+			pkt_idx = j;
+	CU_ASSERT_FATAL(pkt_idx >= 0);
```
MatiasElo: Could use `pool_index_of()` function?

test/validation/api/event/event.c
```diff
@@ -203,6 +304,186 @@ static void event_test_pool_info_ctx(void)
...
+
+	for (i = 0; i < num_pkt; i++) {
+		CU_ASSERT_FATAL(packets[i] != ODP_PACKET_INVALID);
+		CU_ASSERT(odp_packet_to_event(packets[i]) == ctx->events[pkt_idx][i]);
```
MatiasElo: It's safer to compare u64 values than handles.

test/validation/api/event/event.c
```diff
@@ -203,6 +304,186 @@ static void event_test_pool_info_ctx(void)
...
+	event_type_ctx_t *ctx = &g_type_ctx;
+	odp_event_t sorted_events_multi[EVENT_BURST];
+	odp_event_t mixed_events_single[MAX_EVENT_TYPES];
+	odp_event_t mixed_events_multi[MAX_EVENT_TYPES];
+	int i, j;
+
+	for (i = 0; i < MAX_EVENT_TYPES; i++) {
+		for (j = 0; j < EVENT_BURST; j++) {
+			sorted_events_multi[j] = event_types_alloc_event(ctx->pool_types[i]);
+			CU_ASSERT(sorted_events_multi[j] != ODP_EVENT_INVALID);
```
MatiasElo: Should be fatal. Same below for mixed arrays.

test/validation/api/event/event.c
```diff
@@ -84,8 +85,15 @@ static int event_types_suite_init(void)
 	event_type_ctx_t *ctx = &g_type_ctx;
 	int i, j, n = 0;
 	char name[ODP_POOL_NAME_LEN];
+	odp_queue_param_t queue_prm;
 	odp_pool_capability_t capa;
 
+	odp_queue_param_init(&queue_prm);
+	queue_prm.type = ODP_QUEUE_TYPE_PLAIN;
```
MatiasElo, line 158: Queue should be configured to be large enough for storing the tests packets.

test/validation/api/event/event.c
```diff
@@ -203,6 +304,186 @@ static void event_test_pool_info_ctx(void)
...
+	}
+
+	CU_ASSERT(total == MAX_EVENT_TYPES);
+	for (i = 0; i < MAX_EVENT_TYPES; i++)
+		CU_ASSERT(types_seen[i] == 1);
+
+	for (i = 0; i < MAX_EVENT_TYPES; i++)
+		ev_tbl[i] = ctx->events[i][1];
+
+	CU_ASSERT(odp_queue_enq_multi(ctx->queue, ev_tbl, MAX_EVENT_TYPES) == MAX_EVENT_TYPES);
```
MatiasElo: It's OK for `odp_queue_enq_multi()` to return values between 1 and MAX_EVENT_TYPES, so a simple retry loop could be added.

test/validation/api/event/event.c
```diff
@@ -117,12 +129,20 @@ static int event_types_suite_init(void)
...
+	}
+
 	ctx->num_types = n;
 
 	for (i = 0; i < ctx->num_types; i++) {
 		odp_pool_param_t prm;
+		odp_dma_pool_param_t dma_prm;
 
 		odp_pool_param_init(&prm);
+		odp_dma_pool_param_init(&dma_prm);
```
MatiasElo: Will be called even when DMA is not supported. So better to move this to the switch case. Same comment for `odp_ml_compl_pool_param_init()`.

test/validation/api/event/event.c
```diff
@@ -117,12 +129,20 @@ static int event_types_suite_init(void)
 		ctx->pool_types[n++] = ODP_POOL_VECTOR;
 	}
 
+	if ((dma_capa.compl_mode_mask & ODP_DMA_COMPL_EVENT) &&
```
MatiasElo: Should first check `max_sessions`.

test/validation/api/event/event.c
```diff
@@ -135,14 +146,22 @@ static int event_types_suite_init(void)
 		ctx->pool_types[n++] = ODP_POOL_DMA_COMPL;
 	}
 
+	if (((ml_capa.load.compl_mode_mask | ml_capa.run.compl_mode_mask) &
```
MatiasElo: Should first check `max_models`.

test/validation/api/event/event.c
```diff
@@ -1,16 +1,557 @@
...
+			user_area = odp_event_user_area(ev);
+
+			CU_ASSERT(user_area != NULL);
+
+			flag = 0;
+			odp_event_user_flag_set(ev, 1);
+			user_area = odp_event_user_area_and_flag(ev, &flag);
+
+			CU_ASSERT(user_area != NULL);
+			CU_ASSERT(flag != 0);
```
MatiasElo: Could check that the return value matches API spec for each event type: `User flag value pointer for output. >0 if user flag is set, 0 if flags is not set, or <0 if event does not have user flag.`

test/validation/api/event/event.c
```diff
@@ -1,16 +1,557 @@
...
+	odp_event_t ev;
+	int i, j;
+
+	for (i = 0; i < ctx->num_types; i++) {
+		for (j = 0; j < EVENT_BURST; j++) {
+			ev = ctx->type[i].events[j];
+			CU_ASSERT(odp_event_flow_id(ev) == 0);
+
+			if (ctx->max_flow_id > 0) {
+				odp_event_flow_id_set(ev, 1);
```
MatiasElo, line 456: Looks like `max_flow_id` is not actually configured. It should be set in `odp_schedule_config(odp_schedule_config_t.max_flow_id)`. Likely you have to register a new global init function (`odp_cunit_register_global_init()`). You can use `tests_global_init()` as a base for the new function.

test/validation/api/event/event.c
```diff
@@ -1,16 +1,557 @@
...
+	for (i = 0; i < ctx->num_types; i++)
+		CU_ASSERT(types_seen[i] == 1);
+
+	for (i = 0; i < ctx->num_types; i++)
+		ev_tbl[i] = ctx->type[i].events[1];
+
+	num = 1;
+	total = 0;
+	while (num > 0 && total < ctx->num_types) {
+		num = odp_queue_enq_multi(ctx->queue, ev_tbl, ctx->num_types);
```
MatiasElo: If the first call doesn't enqueue all events successfully, the array start pointer and the number of events should be updated accordingly.

test/validation/api/event/event.c
```diff
@@ -1,16 +1,557 @@
...
+		exp_sub = (exp_type == ODP_EVENT_PACKET) ? ODP_EVENT_PACKET_BASIC
+							 : ODP_EVENT_NO_SUBTYPE;
+
+		for (j = 0; j < EVENT_BURST; j++) {
+			ev = ctx->type[i].events[j];
+
+			CU_ASSERT(odp_event_type(ev) == exp_type);
+			CU_ASSERT(odp_event_subtype(ev) == exp_sub);
+
+			sub = ODP_EVENT_NO_SUBTYPE;
```
MatiasElo: Should not need to initialize this.

test/validation/api/event/event.c
```diff
@@ -1,16 +1,557 @@
...
+		CU_ASSERT(type == exp_type);
+
+		/* Fill array indexes 1, ..., len-1 with another event type */
+		idx = (i + 1) % ctx->num_types;
+		for (j = 1; j < ctx->num_types; j++)
+			events[j] = ctx->type[idx].events[j];
+
+		for (j = 0; j < ctx->num_types; j++) {
+			events[j] = ctx->type[i].events[j];
+			CU_ASSERT(odp_event_type_multi(events, ctx->num_types, &type) == j + 1);
```
MatiasElo, line 350: For completeness could check also `type` value.

test/validation/api/event/event.c
```diff
@@ -1,16 +1,557 @@
...
+
+	for (i = 0; i < EVENT_BURST; i++) {
+		for (j = 0; j < ctx->num_types; j++)
+			mixed[idx++] = ctx->type[j].events[i];
+	}
+
+	num_pkt = odp_event_filter_packet(mixed, packets, remain, idx);
+	CU_ASSERT(num_pkt == EVENT_BURST);
+
+	int pkt_idx = pool_index_of(ODP_POOL_PACKET);
```
MatiasElo: Project style is to put variable declarations to the beginning of code block when possible.

test/validation/api/event/event.c
```diff
@@ -1,16 +1,557 @@
...
+
+	num_rem = idx - num_pkt;
+	idx = 0;
+	for (i = 0; i < EVENT_BURST; i++) {
+		for (j = 0; j < ctx->num_types; j++) {
+			if (j == pkt_idx)
+				continue;
+			expected[idx++] = ctx->type[j].events[i];
+		}
+	}
```
MatiasElo: This could be done in the same loop when initializing `mixed` array perhaps making the code a bit easier to read.

test/validation/api/event/event.c
```diff
@@ -1,16 +1,605 @@
...
+	for (i = 0; i < ctx->num_types; i++)
+		CU_ASSERT(types_seen[i] == 1);
+
+	for (i = 0; i < ctx->num_types; i++)
+		ev_tbl[i] = ctx->type[i].events[1];
+
+	num = 1;
+	total = 0;
+	while (num > 0 && total < ctx->num_types) {
+		num = odp_queue_enq_multi(ctx->queue, ev_tbl + total, ctx->num_types - total);
```
MatiasElo: I'd prefer using `&ev_tbl[total]` syntax and the error case could be handled. E.g.
```
	total = 0;
	while (total < ctx->num_types) {
		num = odp_queue_enq_multi(ctx->queue, &ev_tbl[total], ctx->num_types - total);
		CU_ASSERT_FATAL(num >= 0);
		total += num;
	}
```

test/validation/api/event/event.c
```diff
@@ -82,13 +84,70 @@ static odp_event_t event_types_alloc_event(odp_pool_type_t pool_type)
 	}
 }
 
+static int event_init(odp_instance_t *inst)
```
MatiasElo: More descriptive name could be `event_init_global`.

PR #2275, opened 2025-09-29
Add --seg_free option to dmafwd

lifang-zhang:
When --seg_free option is set, DMA frees
all the source segments. By default, DMA
does not free the source segments.


test/performance/odp_dmafwd.c
MatiasElo, line 1: Copyright can be updated.

test/performance/odp_dmafwd.c
```diff
@@ -157,6 +157,7 @@ typedef struct prog_config_s {
 	uint32_t compl_cache_size;
 	uint32_t stash_cache_size;
 	double time_sec;
+	uint8_t seg_free;
```
MatiasElo: Could be boolean (`odp_bool_t`).

test/performance/odp_dmafwd.c
```diff
@@ -511,6 +515,9 @@ static parse_result_t parse_options(int argc, char **argv, prog_config_t *config
 		case 'T':
 			config->time_sec = atof(optarg);
 			break;
+		case 's':
+			config->seg_free = 1;
```
MatiasElo, line 523: `check_options()` should check that source segment free is supported.

test/performance/odp_dmafwd.c
```diff
@@ -1065,12 +1076,14 @@ static inline void send_dma_poll_trs_pkts(int burst_size, thread_config_t *confi
 				++stats->trs;
 				stats->fwd_pkts += num_sent;
 				stats->discards += trs->num - num_sent;
+				if (!config->prog_config->seg_free)
+					odp_packet_free_multi(trs->src_pkts, trs->num);
```
MatiasElo: Looks like similar checks are also required in `drain_events()` and `drain_polled()` to avoid double frees.

Also, nitpicking but I'd save the `config->prog_config->seg_free` value to a const local variable since it's called inside a loop.

test/performance/odp_dmafwd.c
```diff
@@ -256,6 +257,8 @@ static void print_usage(dynamic_defs_t *dyn_defs)
 	       "  -c, --worker_count Amount of workers. %u by default.\n"
 	       "  -C, --cache_size   Maximum cache size for pools. %u by default.\n"
 	       "  -T, --time_sec     Time in seconds to run. 0 means infinite. %u by default.\n"
+	       "  -s, --seg_free     Free source segments by DMA. Source segments are not freed\n"
+	       "                     by DMA by default.\n"
```
MatiasElo: Perhaps: `Use DMA source segment free offload. Disabled by default.`

Also, could print the used value in `print_stats()`.

Issue #2278, opened 2025-10-08
How to work with odp_pktio_maxlen_set in dpdk.


albe19029:
If calling odp_pktio_maxlen_set before odp_pktio_start we get -EINVAL from rte_eth_dev_set_mtu with log "Port %u must be configured before MTU set".

If we try to call odp_pktio_maxlen_set after odp_pktio_start we have -1 with error "Pktio not stopped\n".

So it is not very clear how to set maxlen correctly.

As I see  rte_eth_dev_set_mtu will start working after rte_eth_dev_configure. But rte_eth_dev_configure is calling in dpdk_setup_eth_dev (odp_pktio_start -> dpdk_start-> rte_eth_dev_configure).

From what I see https://github.com/OpenDataPlane/odp/blob/master/platform/linux-generic/pktio/dpdk.c#L1203 we just ignore rte_eth_dev_set_mtu result and save mtu locally. And then restore it here https://github.com/OpenDataPlane/odp/blob/master/platform/linux-generic/pktio/dpdk.c#L2097.

So the question is MTU will be set even if odp_pktio_maxlen_set returned -22 (called before odp_pktio_start). And if this comment correct https://github.com/OpenDataPlane/odp/blob/master/platform/linux-generic/pktio/dpdk.c#L2096 , or maybe should be "/* Restore MTU value reset by dpdk_setup_eth_tx */", as after calling dpdk_setup_eth_tx it is possible to set MTU, I don't see any logic in dpdk_setup_eth_rx() related to MTU.

albe19029:
For now the only workaround if to:

odp_pktio_start
odp_pktio_stop
odp_pktio_maxlen_set
odp_pktio_start

Which is very ugly.

albe19029:
The problem is that in current implementation we call rte_eth_dev_configure twice. 

And calling rte_eth_dev_configure second time you rewrite mtu. In odp it is only working because you odp code set:

eth_conf.rxmode.mtu = pkt_dpdk->mtu - _ODP_ETHHDR_LEN;

So this mtu check is very agly in dpdk. And it is very easy to bypass it. And only because of this agly check resetting device twice?

PR #2281, opened 2025-10-17
[PATCH v7] linux-gen: dpdk: fix maxlen setting error before pktio start

lifang-zhang:
Setting maxlen with odp_pktio_maxlen_set() before odp_pktio_start() fails in dpdk pktio, as rte_eth_dev_set_mtu() must be called after rte_eth_dev_configure(), which occurs during odp_pktio_start().

### DIFF
--- a/platform/linux-generic/pktio/dpdk.c
+++ b/platform/linux-generic/pktio/dpdk.c
@@ -1128,56 +1128,6 @@ static int dpdk_netdev_is_valid(const char *s)
 	return 1;
 }
 
-static uint32_t dpdk_vdev_mtu_get(uint16_t port_id)
-{
-	struct rte_eth_dev_info dev_info;
-	struct ifreq ifr;
-	int sockfd;
-	uint32_t mtu;
-	int ret;
-
-	memset(&dev_info, 0, sizeof(struct rte_eth_dev_info));
-
-	ret = rte_eth_dev_info_get(port_id, &dev_info);
-	if (ret != 0) {
-		_ODP_ERR("DPDK: rte_eth_dev_info_get() failed with return value: %d, port: %u\n",
-			 ret, port_id);
-		return 0;
-	}
-	if_indextoname(dev_info.if_index, ifr.ifr_name);
-
-	sockfd = socket(AF_INET, SOCK_DGRAM, 0);
-	if (sockfd < 0) {
-		_ODP_ERR("Failed to create control socket\n");
-		return 0;
-	}
-
-	mtu = _odp_mtu_get_fd(sockfd, ifr.ifr_name);
-	close(sockfd);
-	return mtu;
-}
-
-static uint32_t dpdk_mtu_get(pktio_entry_t *pktio_entry)
-{
-	pkt_dpdk_t *pkt_dpdk = pkt_priv(pktio_entry);
-	uint32_t mtu = 0;
-
-	if (rte_eth_dev_get_mtu(pkt_dpdk->port_id, (uint16_t *)&mtu))
-		return 0;
-
-	/* Some DPDK PMD virtual devices do not support getting MTU size.
-	 * Try to use system call if DPDK cannot get MTU value.
-	 */
-	if (mtu == 0)
-		mtu = dpdk_vdev_mtu_get(pkt_dpdk->port_id);
-
-	/* Mbuf chaining not yet supported */
-	if (pkt_dpdk->data_room && pkt_dpdk->data_room < mtu)
-		return pkt_dpdk->data_room;
-
-	return mtu;
-}
-
 static uint32_t dpdk_maxlen(pktio_entry_t *pktio_entry)
 {
 	pkt_dpdk_t *pkt_dpdk = pkt_priv(pktio_entry);
@@ -1910,9 +1860,9 @@ static int dpdk_open(odp_pktio_t id ODP_UNUSED,
 	}
 	pkt_dpdk->flags.set_flow_hash = pkt_dpdk->opt.set_flow_hash; /* Copy for fast path access */
 
-	mtu = dpdk_mtu_get(pktio_entry);
-	if (mtu == 0) {
-		_ODP_ERR("Failed to read interface MTU\n");
+	ret = rte_eth_dev_get_mtu(pkt_dpdk->port_id, (uint16_t *)&mtu);
+	if (ret) {
+		_ODP_ERR("Failed to read interface MTU: err=%d\n", ret);
 		return -1;
 	}
 	pkt_dpdk->mtu = mtu + _ODP_ETHHDR_LEN;


platform/linux-generic/pktio/dpdk.c
```diff
@@ -154,8 +154,10 @@ typedef struct ODP_ALIGNED_CACHE {
 	uint32_t data_room;
 	/* Maximum supported MTU value */
 	uint32_t mtu_max;
-	/* DPDK MTU has been modified */
+	/* DPDK MTU modification has been requested via odp_pktio_maxlen_set */
 	uint8_t mtu_set;
+	/* Device configured (rte_eth_dev_configure done) */
+	uint8_t configured;
```
MatiasElo: The documentation of `rte_eth_dev_configure()` states that it must be called before any other ethernet API (`rte_eth_dev_info_get()` being the only exception?), so the same issues could also be happening with other APIs (e.g. promisc mode, MAC get/set). I'm wondering if this problem could be solved by doing "dummy" configure in `dpdk_open()` using default values.

lifang-zhang: Going through the code, dpdk API and chatting with Chatgpt, it is not encouraged to call a dummy configure with default values first in dpdk_open() and then call it later with rightly set values in dpdk_start(). The DPDK API is a little bit unclear about the order corresponding to rte_eth_dev_configure() for most of the APIs. But it clearly enforces that  rte_eth_dev_configure() must be called before rte_eth_tx_queue_setup(), rte_eth_rx_queue_setup() and rte_eth_dev_set_mtu() (not documented in API but in commit message). Otherwise rte_eth_dev_configure() is documented as 
	This function must be invoked first before any other function in the Ethernet API. This function can also be re-invoked when a device is in the stopped state.

The main reasons why it is not recommended are: 
	• Reconfiguring requires the port to be in stopped state. After first configure (not started yet) some PMDs still reject a second configure without an explicit stop.
	• It is not guaranteed that all PMD  allows a redundant pre‑configure followed by the real one (some may leak rings, fail, or silently ignore changes)
	• Supplying guesses of nb_rx_queue/nb_tx_queue (1/1, max, etc., 0 is invalid) wastes resources.
	• Resource waste: early configure allocates rings/descriptors that might be thrown away.

Followings are not a hard spec rule but conservative recommendation from typical driver behavior, not from a line in the docs: 

Once EAL has probed the port (port_id valid), read-only, non state‑changing queries (e.g. rte_eth_dev_count_avail(), rte_eth_dev_get_port_by_name(),  rte_eth_dev_info_get()) may be called before rte_eth_dev_configure(). However, calling some state‑changing DPDK APIs before rte_eth_dev_configure() is fragile. MTU definitely must be after configure (from 21.11). Others may appear to work but are not guaranteed across PMDs. We are using a few such state-changing functions before rte_eth_dev_configure():

rte_eth_dev_default_mac_addr_set(): specification expects device to be configured; some PMDs reject it pre‑configure. 
rte_eth_promiscuous_enable/disable and rte_eth_allmulticast_enable/disable: safer after configure; some virtual PMDs tolerate earlier calls, others may fail silently.
rte_eth_dev_get_supported_ptypes(): often returns 0 unless port started; harmless but you may get incomplete info pre‑start.
rte_eth_stats_reset(): resetting before configure gives no value; move after configure (optionally after queue setup, before start) to establish a clean baseline.

I will check these function case by case and see if we can move/defer them later after rte_eth_dev_configure().

MatiasElo: Deferring the configuration will also defer potential failures to pktio start, which will make debugging more difficult. Also, this would in practice disable most the validation tests for maxlen, promisc mode, and mac.

So I'd still at least test doing dummy configure in pktio open as the API clearly allows this. Buggy drivers rejecting a second configure without an explicit stop may be an issue, but resource usage is not a problem.

platform/linux-generic/pktio/dpdk.c
```diff
@@ -1155,7 +1155,14 @@ static uint32_t dpdk_vdev_mtu_get(uint16_t port_id)
 
 	mtu = _odp_mtu_get_fd(sockfd, ifr.ifr_name);
```
MatiasElo: This whole function is quite old, so you could check if it is still necessary or could it be removed altogether.

platform/linux-generic/pktio/dpdk.c
```diff
@@ -1900,6 +1901,14 @@ static int dpdk_open(odp_pktio_t id ODP_UNUSED,
...
 	}
 
+	memset(&eth_conf, 0, sizeof(eth_conf));
+	ret = rte_eth_dev_configure(pkt_dpdk->port_id, 1, 1, &eth_conf);
+	if (ret) {
+		_ODP_ERR("Failed to configure device: err=%d, port=%" PRIu8 "\n",
+			 ret, pkt_dpdk->port_id);
+		return -1;
+	}
+
```
MatiasElo, line 1855: Now the `rte_eth_dev_set_mtu()` comment and check in `dpdk_init_capability()` could be updated.

platform/linux-generic/pktio/dpdk.c
```diff
@@ -1900,6 +1901,14 @@ static int dpdk_open(odp_pktio_t id ODP_UNUSED,
...
 	/* Initialize runtime options */
 	if (init_options(pktio_entry, &dev_info)) {
```
MatiasElo, line 1857: As mentioned in #2278, the comment about restoring MTU value in `dpdk_start()` is invalid. MTU is actually reset by `rte_eth_dev_configure()`. So the code could be simplified by passing MTU as parameter to `rte_eth_dev_configure()` in `dpdk_setup_eth_dev()`. Perhaps we could also get rid of the `pkt_dpdk_t.mtu_set` variable.

platform/linux-generic/pktio/dpdk.c
```diff
@@ -1894,6 +1891,17 @@ static int dpdk_open(odp_pktio_t id ODP_UNUSED,
 		return -1;
 	}
 
+	memset(&eth_conf, 0, sizeof(eth_conf));
+	ret = rte_eth_dev_configure(pkt_dpdk->port_id,
+				    _ODP_MIN(1, dev_info.max_rx_queues),
+				    _ODP_MIN(1, dev_info.max_tx_queues),
+				    &eth_conf);
+	if (ret) {
+		_ODP_ERR("Failed to configure device: err=%d, port=%" PRIu8 "\n",
```
MatiasElo: Could use device name instead of port in error message.

platform/linux-generic/pktio/dpdk.c
```diff
@@ -1300,6 +1297,7 @@ static int dpdk_setup_eth_dev(pktio_entry_t *pktio_entry)
 
 	eth_conf.rxmode.mq_mode = RTE_ETH_MQ_RX_RSS;
 	eth_conf.txmode.mq_mode = RTE_ETH_MQ_TX_NONE;
+	eth_conf.rxmode.mtu = pkt_dpdk->mtu - _ODP_ETHHDR_LEN;
```
MatiasElo: To be on the safe side, could check if setting MTU is supported.

platform/linux-generic/pktio/dpdk.c
```diff
@@ -1191,19 +1189,15 @@ static int dpdk_maxlen_set(pktio_entry_t *pktio_entry, uint32_t maxlen_input,
...
 {
 	pkt_dpdk_t *pkt_dpdk = pkt_priv(pktio_entry);
-	uint16_t mtu;
-	int ret;
 
 	/* DPDK MTU value does not include Ethernet header */
-	mtu = maxlen_input - _ODP_ETHHDR_LEN;
+	uint16_t mtu = maxlen_input - _ODP_ETHHDR_LEN;
+
+	int ret = rte_eth_dev_set_mtu(pkt_dpdk->port_id, mtu);
```
MatiasElo: Project convention is that variables are declared at the beginning of functional block if possible.

platform/linux-generic/pktio/dpdk.c
```diff
@@ -1191,19 +1189,15 @@ static int dpdk_maxlen_set(pktio_entry_t *pktio_entry, uint32_t maxlen_input,
...
 
-	ret = rte_eth_dev_set_mtu(pkt_dpdk->port_id, mtu);
 	if (odp_unlikely(ret))
 		_ODP_ERR("rte_eth_dev_set_mtu() failed: %d\n", ret);
 
-	pkt_dpdk->mtu = maxlen_input;
```
MatiasElo, line 1204: `pkt_dpdk->mtu` should still be saved (in case of success) as it's used for example in `dpdk_setup_eth_dev()`.

platform/linux-generic/pktio/dpdk.c
```diff
@@ -1894,6 +1890,17 @@ static int dpdk_open(odp_pktio_t id ODP_UNUSED,
 		return -1;
 	}
 
+	memset(&eth_conf, 0, sizeof(eth_conf));
+	ret = rte_eth_dev_configure(pkt_dpdk->port_id,
+				    _ODP_MIN(1, dev_info.max_rx_queues),
+				    _ODP_MIN(1, dev_info.max_tx_queues),
+				    &eth_conf);
+	if (ret) {
+		_ODP_ERR("Failed to configure device %s: err=%d," PRIu8 "\n",
```
MatiasElo: Should be likely: `"Failed to configure device %s: err=%d"\n"`

platform/linux-generic/pktio/dpdk.c
```diff
@@ -1908,8 +1858,7 @@ static int dpdk_open(odp_pktio_t id ODP_UNUSED,
 	}
 	pkt_dpdk->flags.set_flow_hash = pkt_dpdk->opt.set_flow_hash; /* Copy for fast path access */
 
-	mtu = dpdk_mtu_get(pktio_entry);
-	if (mtu == 0) {
+	if (rte_eth_dev_get_mtu(pkt_dpdk->port_id, (uint16_t *)&mtu)) {
 		_ODP_ERR("Failed to read interface MTU\n");
```
MatiasElo: I'd print the return value in case of failure.

PR #2287, opened 2025-11-11
[PATCH v16] api: packet: redefine packet data sharing with packet references

JannePeltonen:
Current API regarding dynamic packet references is not easy to efficiently support in some ODP implementations with HW accelerated packet output due to the need to update reference counts after TX completion and based on that conditionally free packet buffers. odp_packet_ref() has been implemented only as a full packet copy in all ODP implementation.

Redefine the semantics of odp_packet_ref() and the packets involved in packet data sharing in a way that hopefully makes it simpler to support packet data sharing with restrictions in packet output and with per-packet reference counting instead of per-segment reference counting. The new semantics also enable use cases that were not previously possible.

Specify better the rules regarding packet data layout manipulation operations, allowing private packet data also in the tail of the packets that reference other packets. Pull and truncate operations may involve shared packet data, in which case they just remove that data from the packet, not affecting the packet owning the data. All extend and truncate operations always produce non-shared packet data.

Change odp_packet_has_ref() to indicate whether a packet is referenced by another packet, not whether the packet shares data with another packet. In partivular, odp_packet_ref() returns a packet that shares data with the given packet but is not itself referenced by other packets. Specify that packet data and layout of packets referenced by other packets can not be modified at all.

Introduce pktio capability bits to indicate whether a pktio supports sending packets with shared data without using the dont_free flag. Sending with the dont_free flag is supported with all types of packets.

### DIFF
--- a/example/sysinfo/odp_sysinfo.c
+++ b/example/sysinfo/odp_sysinfo.c
@@ -641,6 +641,9 @@ static void print_pktio_capa(appl_args_t *appl_args)
 		printf("    lso.proto.tcp_ipv6:            %u\n", capa->lso.proto.tcp_ipv6);
 		printf("    lso.proto.sctp_ipv4:           %u\n", capa->lso.proto.sctp_ipv4);
 		printf("    lso.proto.sctp_ipv6:           %u\n", capa->lso.proto.sctp_ipv6);
+		printf("    packet_ref.static_ref          %u\n", capa->packet_ref.static_ref);
+		printf("    packet_ref.referenced_pkt      %u\n", capa->packet_ref.referenced_pkt);
+		printf("    packet_ref.referencing_pkt     %u\n", capa->packet_ref.referencing_pkt);
 		printf("    maxlen.equal:                  %i\n", capa->maxlen.equal);
 		printf("    maxlen.min_input:              %u B\n", capa->maxlen.min_input);
 		printf("    maxlen.max_input:              %u B\n", capa->maxlen.max_input);
--- a/include/odp/api/spec/ipsec.h
+++ b/include/odp/api/spec/ipsec.h
@@ -1,6 +1,6 @@
 /* SPDX-License-Identifier: BSD-3-Clause
  * Copyright (c) 2016-2018 Linaro Limited
- * Copyright (c) 2021-2022 Nokia
+ * Copyright (c) 2021-2025 Nokia
  */
 
 /**
@@ -433,6 +433,10 @@ int odp_ipsec_out_enq(const odp_packet_t pkt[], int num,
  * otherwise identical to odp_ipsec_out_enq(), but outputs all successfully
  * transformed packets to the specified output interface (or tm_queue), instead of
  * generating events for those.
+ *
+ * Packets that reference or are referenced by other packets may be sent
+ * using this function only if the pktout to be used has the relevant
+ * capability. See odp_pktio_capability_t::packet_ref.
  *
  * Inline operation parameters are defined per packet. The array of parameters
  * must have 'num' elements and is pointed to by 'inline_param'.
--- a/include/odp/api/spec/packet.h
+++ b/include/odp/api/spec/packet.h
@@ -1,6 +1,6 @@
 /* SPDX-License-Identifier: BSD-3-Clause
  * Copyright (c) 2013-2018 Linaro Limited
- * Copyright (c) 2021-2024 Nokia
+ * Copyright (c) 2021-2026 Nokia
  */
 
 /**
@@ -268,8 +268,7 @@ int odp_packet_reass_partial_state(odp_packet_t pkt, odp_packet_t frags[],
  * Packet head address
  *
  * Returns start address of the first segment. Packet level headroom starts
- * from here. Use odp_packet_data() or odp_packet_l2_ptr() to return the
- * packet data start address.
+ * from here. Use odp_packet_data() to return the packet data start address.
  *
  * @param pkt  Packet handle
  *
--- a/include/odp/api/spec/packet.h
+++ b/include/odp/api/spec/packet.h
@@ -94,6 +94,13 @@ int odp_packet_alloc_multi(odp_pool_t pool, uint32_t len,
  *
  * Frees the packet into the packet pool it was allocated from.
  *
+ * A packet being referenced by other packets (see odp_packet_ref() and
+ * odp_packet_ref_static()) continues to consume pool capacity until there
+ * are no more references to it left. Referencing packets are not affected
+ * when a packet they reference is freed.
+ *
+ * The freed packet may not be used in any way after this call.
+ *
  * @param pkt           Packet handle
  */
 void odp_packet_free(odp_packet_t pkt);
@@ -270,6 +277,9 @@ int odp_packet_reass_partial_state(odp_packet_t pkt, odp_packet_t frags[],
  * Returns start address of the first segment. Packet level headroom starts
  * from here. Use odp_packet_data() to return the packet data start address.
  *
+ * This function returns an undefined value if the packet is referencing
+ * other packets.
+ *
  * @param pkt  Packet handle
  *
  * @return Pointer to the start address of the first packet segment
@@ -283,6 +293,9 @@ void *odp_packet_head(odp_packet_t pkt);
  *
  * Returns sum of buffer lengths over all packet segments. Buffer length
  * includes headroom, data, and tailroom lengths.
+ *
+ * This function returns an undefined value if the packet is referencing
+ * other packets.
  *
  * @param pkt  Packet handle
  *
@@ -366,6 +379,9 @@ uint32_t odp_packet_len(odp_packet_t pkt);
  *
  * Returns the current packet level headroom length.
  *
+ * If the packet is a referencing packet and the first data packet byte is
+ * shared with another packet, headroom is zero.
+ *
  * @param pkt  Packet handle
  *
  * @return Headroom length
@@ -376,6 +392,9 @@ uint32_t odp_packet_headroom(odp_packet_t pkt);
  * Packet tailroom length
  *
  * Returns the current packet level tailroom length.
+ *
+ * If the packet is a referencing packet and the last data packet byte is
+ * shared with another packet, tailroom is zero.
  *
  * @param pkt  Packet handle
  *
@@ -464,15 +483,19 @@ void *odp_packet_push_head(odp_packet_t pkt, uint32_t len);
  * Pull in packet head
  *
  * Decrease packet data length by removing data from the head of the packet.
- * Packet headroom is increased with the same amount. Packet head may be pulled
- * in up to seg_len - 1 bytes (i.e. packet data pointer must stay in the
- * first segment). Packet is not modified if there's not enough data.
+ * Packet head may be pulled in up to seg_len - 1 bytes (i.e. packet data
+ * pointer must stay in the first segment). Packet is not modified if there's
+ * not enough data in the first segment.
  *
  * odp_packet_xxx:
  * seg_len  -= len
  * len      -= len
- * headroom += len
  * data     += len
+ * headroom += len, if the packet is not a referencing packet
+ *
+ * If the packet is a referencing packet and the first segment shares data
+ * with another packet, headroom does not increase. Otherwise headroom
+ * increases by len bytes.
  *
  * Operation does not modify packet segmentation or move data. Handles and
  * pointers remain valid. User is responsible to update packet metadata
@@ -531,7 +554,11 @@ void *odp_packet_push_tail(odp_packet_t pkt, uint32_t len);
  * odp_packet_xxx:
  * len      -= len
  * tail     -= len
- * tailroom += len
+ * tailroom += len, if the packet is not a referencing packet
+ *
+ * If the packet is a referencing packet and the last segment shares data
+ * with another packet, tailroom does not increase. Otherwise tailroom
+ * increases by len bytes.
  *
  * Operation does not modify packet segmentation or move data. Handles and
  * pointers remain valid. User is responsible to update packet metadata
@@ -562,6 +589,9 @@ void *odp_packet_pull_tail(odp_packet_t pkt, uint32_t len);
  * user_area) were moved in memory during the operation. If some memory areas
  * were moved, application must use new packet/segment handles to update
  * data pointers. Otherwise, all old pointers remain valid.
+ *
+ * The new packet data added in the packet will always be non-shared, even
+ * when the packet is a reference that shares data with other packets.
  *
  * User is responsible to update packet metadata offsets when needed. Packet
  * is not modified if operation fails.
@@ -636,6 +666,9 @@ int odp_packet_trunc_head(odp_packet_t *pkt, uint32_t len, void **data_ptr,
  * were moved, application must use new packet/segment handles to update
  * data pointers. Otherwise, all old pointers remain valid.
  *
+ * The new packet data added in the packet will always be non-shared, even
+ * when the packet is a reference that shares data with other packets.
+ *
  * User is responsible to update packet metadata offsets when needed. Packet
  * is not modified if operation fails.
  *
@@ -706,6 +739,9 @@ int odp_packet_trunc_tail(odp_packet_t *pkt, uint32_t len, void **tail_ptr,
  * user_area) were moved in memory during the operation. If some memory areas
  * were moved, application must use new packet/segment handles to update
  * data pointers. Otherwise, all old pointers remain valid.
+ *
+ * The new packet data added in the packet will always be non-shared, even
+ * when the packet is a reference that shares data with other packets.
  *
  * User is responsible to update packet metadata offsets when needed. Packet
  * is not modified if operation fails.
@@ -908,6 +944,25 @@ uint32_t odp_packet_seg_data_len(odp_packet_t pkt, odp_packet_seg_t seg);
  * the destination packet. The source packet may have been allocated from
  * any pool.
  *
+ * 'dst' must not be a referenced packet or a static packet reference. It can
+ * be a referencing packet. 'src' may be any kind of packet. The returned
+ * packet will be either a normal packet or referencing packet depending
+ * on the input packets and implementation details as follows:
+ *
+ * If both the destination packet and the source packet are normal packets,
+ * the result packet will also be a normal packet. If the destination packet
+ * is a referencing packet, the result packet will also be a referencing packet.
+ * Otherwise the result packet is either a normal or referencing packet
+ * depending on the packets and the implementation.
+ *
+ *  dst         src             result
+ *  -------------------------------------------------
+ *  normal      normal          normal
+ *  normal      referenced      normal or referencing
+ *  normal      referencing     normal or referencing
+ *  normal      static ref      normal or referencing
+ *  referencing any             referencing
+ *
  * On failure, both handles remain valid and packets are not modified.
  *
  * @param[in, out] dst   Pointer to destination packet handle. A successful
@@ -1050,6 +1105,9 @@ odp_packet_buf_t odp_packet_buf_from_head(odp_pool_t pool, void *head);
  *
  * This call can be used only for packets of an external memory pool (see odp_pool_ext_create()).
  *
+ * This function must not be called for packets that reference other packets or
+ * are referenced by other packets.
+ *
  * @param      pkt      Packet to be disassembled
  * @param[out] pkt_buf  Packet buffer handle array for output
  * @param      num      Number of elements in packet buffer handle array. Must be equal to or
@@ -1108,6 +1166,9 @@ odp_packet_t odp_packet_reassemble(odp_pool_t pool, odp_packet_buf_t pkt_buf[], 
  * e.g., packet retransmissions. Use odp_packet_ref() or odp_packet_ref_pkt()
  * for more flexible, dynamic references.
  *
+ * This function may be called simultaneously in different threads for the
+ * same packet.
+ *
  * Packet is not modified on failure.
  *
  * @param pkt    Handle of the packet for which a static reference is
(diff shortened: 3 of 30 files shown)


include/odp/api/spec/packet.h
```diff
@@ -1120,63 +1120,96 @@ odp_packet_t odp_packet_reassemble(odp_pool_t pool, odp_packet_buf_t pkt_buf[],
...
+ *   to a destination packet that shares data with another packet may require
+ *   packet data copying if also the source packet shares data with another
+ *   packet. This is because, depending on the ODP implementation, a packet
+ *   may be able to reference only one other packet.
  *
- * @return New reference to the packet
+ * Packets returned by this function can be used normally in ODP API functions
+ * unless otherwise noted. There are restrictions in packet output.
+ *
+ * Referenced packets can be used in ODP API functions that treat packet data
```
ashwinyes, line 1252: I think we should use one word to refer. Either base packet or referenced packet.

include/odp/api/spec/packet.h
```diff
@@ -1188,16 +1221,18 @@ odp_packet_t odp_packet_ref_pkt(odp_packet_t pkt, uint32_t offset,
...
- * odp_packet_ref_pkt() calls. The intent of multiple references is to avoid
- * packet copies, however some implementations may do a packet copy for some of
- * the calls. If a copy is done, the new reference is actually a new, unique
- * packet and this function returns '0' for it. When a real reference is
- * created (instead of a copy), this function returns '1' for both packets
- * (the original packet and the new reference).
+ * of a static reference it also shares metadata. Packet data and data layout
+ * of packet that have references must be treated as read only.
+ *
+ * Packets created through odp_packet_ref() or odp_packet_ref_hdr() share data
```
ashwinyes: odp_packet_ref_hdr() --> odp_packet_ref_pkt()

include/odp/api/spec/packet.h
JannePeltonen: Changed in v2 so that we talk about referenced packets, not base packets.

include/odp/api/spec/packet.h
JannePeltonen: Fixed in v2

JannePeltonen:
v3: Specified that sending shared packets to TM must obey the restrictions of the egress pktio (but currently TM does not support the dont_free flag, which maybe needs to change).

include/odp/api/spec/ipsec.h
```diff
@@ -434,6 +434,10 @@ int odp_ipsec_out_enq(const odp_packet_t pkt[], int num,
  * transformed packets to the specified output interface (or tm_queue), instead of
  * generating events for those.
  *
+ * Packets that share data with other packets may not be sent using this
+ * function unless the pktout to be used has the relevant capability.
+ * See odp_pktio_capability_t::packet_ref.
+ *
```
TuomasTaipale: Similar to the the `odp_tm_enq()` spec, I think the "what you can do" style is more understandable (instead of the "what you cannot do").

Could be useful also to use the "referenced" and "referencing" terms here as with `odp_tm_enq()`.

include/odp/api/spec/packet.h
```diff
@@ -94,6 +94,20 @@ int odp_packet_alloc_multi(odp_pool_t pool, uint32_t len,
  *
  * Frees the packet into the packet pool it was allocated from.
  *
+ * Packets that participate in packet data sharing (see odp_packet_ref()
```
TuomasTaipale: Same here, maybe use "referenced" and "referencing" terms instead of "Packets that participate in packet data sharing"?

include/odp/api/spec/packet.h
```diff
@@ -94,6 +94,20 @@ int odp_packet_alloc_multi(odp_pool_t pool, uint32_t len,
...
+ * and odp_packet_ref_static()) may not get freed back to the packet pool
+ * immediately when this function is called:
+ *
+ * - If the packet being freed is referenced by other packets, then the packet
+ *   is not actually put back to the pool until there are no more references
+ *   to it left. This is done using implementation internal reference counting.
```
TuomasTaipale: Unnecessary perhaps "This is done using implementation internal reference counting."?

include/odp/api/spec/packet.h
```diff
@@ -94,6 +94,20 @@ int odp_packet_alloc_multi(odp_pool_t pool, uint32_t len,
...
+ *
+ * - If the packet being freed references other packets then the referenced
+ *   packets or packet segments are also freed to the packet pool if this
+ *   packet was the last reference to them.
+ *
+ * The packet handle may not be used in any ODP API function after this call.
```
TuomasTaipale: Perhaps just "The packet handle should not be used after this call."

include/odp/api/spec/packet.h
```diff
@@ -909,6 +923,25 @@ uint32_t odp_packet_seg_data_len(odp_packet_t pkt, odp_packet_seg_t seg);
...
+ * 'dst' must not be a referenced packet or a static packet reference. It can
+ * be a referencing packet. 'src' may be any kind of packet. The returned
+ * packet will be either a normal packet or referencing packet depending
+ * on the input packets and implementation details as follows:
+ *
+ * If both the destination packet and the source packet are normal packets,
+ * the result packet will also be a normal packet. If the destination packet
+ * is a referencing packet, the result packet will also be a referencing packet.
+ * Otherwise the result packet is either a normal or referencing packet
+ * depending on the packets and the ODP implementation.
```
TuomasTaipale: The "ODP" is probably unnecessary.

include/odp/api/spec/packet.h
```diff
@@ -909,6 +923,25 @@ uint32_t odp_packet_seg_data_len(odp_packet_t pkt, odp_packet_seg_t seg);
...
+ *
+ *  dst		src		result
+ *  -------------------------------------------------
+ *  normal	normal		normal
+ *  normal	referenced	normal or referencing
+ *  normal	referencing	normal or referencing
+ *  normal	static ref	normal or referencing
+ *  referencing	any		referencing
```
TuomasTaipale: Should the "any" match the "src" column indentation?

include/odp/api/spec/packet.h
```diff
@@ -1120,63 +1156,121 @@ odp_packet_t odp_packet_reassemble(odp_pool_t pool, odp_packet_buf_t pkt_buf[],
...
+ * using packet manipulation operation such as push and extend (see below).
+ * The referencing packet may also be modified to not actually contain any
+ * packet data of the referenced packet.
+ *
+ * Packet data and data layout of the referenced packet must not be modified
+ * as long as any packet that references the packet exists. In particular,
+ * a referenced packet must not be passed to odp_packet_{push,pull,extend,
+ * trunc}_{head,tail}(), odp_packet_split(), or as the destination packet to
+ * odp_packet_concat() or to any other ODP API function that could modify
+ * packet data or packet data layout.
```
TuomasTaipale, line 1209: This paragraph is quite similar what is talked about in detail later with bullet points, could they be fused somehow to avoid repetition (the spec text is quite long already)?

include/odp/api/spec/packet.h
```diff
@@ -1120,63 +1156,121 @@ odp_packet_t odp_packet_reassemble(odp_pool_t pool, odp_packet_buf_t pkt_buf[],
...
+ * odp_packet_has_ref() returns 1 for the referenced packet as long as it is
+ * being referenced by other packets and 0 for the newly created referencing
+ * packet. odp_packet_is_referencing() returns 1 for the created referencing
+ * packet.
+ *
+ * The referenced and referencing packets can be freed in any order.
+ * ODP ensures that the packets or packet segments backing the shared packet
+ * data is put back to the packet pool only after there are no references left.
+ * An ODP implementation may implement this using per-packet or per-segment
+ * reference counting or by other means. See odp_packet_free().
```
TuomasTaipale: Unnecessary "An ODP implementation may implement this using per-packet or per-segment reference counting or by other means."?

include/odp/api/spec/packet.h
```diff
@@ -1120,63 +1156,121 @@ odp_packet_t odp_packet_reassemble(odp_pool_t pool, odp_packet_buf_t pkt_buf[],
...
+ *   The new packet data created by the functions is always private, even if
+ *   extending after truncating or pulling shared data. Because of this,
+ *   odp_packet_headroom() and odp_packet_tailroom() likely return 0 for the
+ *   newly created packet even if the referenced packet has headroom or
+ *   tailroom.
+ *
+ * - odp_packet_concat() can be used normally but concatenating packet data
+ *   to a destination packet that shares data with another packet may require
+ *   packet data copying if also the source packet shares data with another
+ *   packet. This is because, depending on the ODP implementation, a packet
```
TuomasTaipale: "ODP" unnecessary.

include/odp/api/spec/packet.h
```diff
@@ -1120,63 +1156,121 @@ odp_packet_t odp_packet_reassemble(odp_pool_t pool, odp_packet_buf_t pkt_buf[],
...
+ * this combined operation as to those operations. The resulting packet
+ * consists of metadata and packet data of 'hdr' as unshared data, followed
+ * by the packet data shared with 'pkt'.
  *
- * This operation is otherwise identical to odp_packet_ref(), but it prepends
- * a supplied 'hdr' packet as the head of the new reference. The resulting
- * packet consists metadata and data of the 'hdr' packet, followed by the
- * shared part of packet 'pkt'.
+ * 'pkt' must not be a static reference or a referencing packet but it
+ * may be a referenced packet.  'hdr' must not be a static reference or a
```
TuomasTaipale: Extra whitespace before "'hdr'".

include/odp/api/spec/packet.h
```diff
@@ -1206,6 +1304,19 @@ odp_packet_t odp_packet_ref_pkt(odp_packet_t pkt, uint32_t offset,
  */
 int odp_packet_has_ref(odp_packet_t pkt);
 
+/**
+ * Test if a packet references another packet
+ *
+ * Return nonzero if the packet references another packet.
```
TuomasTaipale: `@retval` defines "1" for this case so this could also say "1" instead of "nonzero".

include/odp/api/spec/packet_io.h
```diff
@@ -539,6 +547,10 @@ int odp_lso_profile_destroy(odp_lso_profile_t lso_profile);
  *
  * Check LSO support level from packet IO capabilities (odp_pktio_capability_t).
  *
+ * Packets that share data with other packets may be sent without the
+ * dont_free option only if the relevant packet_ref capability is set
+ * (see odp_pktout_packet_ref_capability_t).
+ *
```
TuomasTaipale: Just use the same text as in `odp_pktout_event_queue()` and `odp_pktout_send()` spec?

include/odp/api/spec/packet_io_types.h
```diff
@@ -1015,6 +1022,32 @@ typedef struct odp_pktin_vector_capability_t {
...
+ *  These capabilities indicate whether packets that reference other packets
+ *  or are referenced by other packets can be sent so that they get consumed by
+ *  the packet output. When such a packet is consumed by the packet output at
+ *  the completion of packet transmit, the packet is handled the same as in
+ *  odp_packet_free() (i.e. internal reference counts are updated and the packet
+ *  segments are put back to the packet pool when there are no more references
+ *  left).
+ *
+ *  If these capabilities are not set, then the respective packet types may be
+ *  sent only if the pktout supports the dont_free option and the option is set
```
TuomasTaipale: Could have the '  ' around "dont_free".

include/odp/api/spec/packet_io_types.h
```diff
@@ -1015,6 +1022,32 @@ typedef struct odp_pktin_vector_capability_t {
...
+ *  See odp_pktio_capability_t::free_ctrl
+ *  See odp_packet_free_ctrl_set()
+ *  See odp_packet_ref() and odp_packet_ref_static()
+ */
+typedef struct {
+	uint8_t static_ref  :1; /**< Static packet references can be consumed */
+	uint8_t referenced  :1; /**< Referenced packets can be consumed */
+	uint8_t referencing :1; /**< Referencing packets can be consumed */
+
+} odp_pktout_packet_ref_capability_t;
```
TuomasTaipale: Better have the same style used elsewhere in file (spec above the field, newlines after fields):
```C
typedef struct {
	/** Static packet references can be consumed */
	uint8_t static_ref  :1;

	/** Referenced packets can be consumed */
	uint8_t referenced  :1;

	/** Referencing packets can be consumed */ 
	uint8_t referencing :1;

} odp_pktout_packet_ref_capability_t;
```

Same comment for the traffic manager capa structure.

include/odp/api/spec/packet_io.h
```diff
@@ -260,6 +260,10 @@ int odp_pktin_queue(odp_pktio_t pktio, odp_pktin_queue_t queues[], int num);
  * odp_queue_enq_multi(). Behaviour is undefined if other events than packets
  * are enqueued. Application cannot dequeue from these queues.
  *
+ * Static references, referencing and referenced packets may be sent without
+ * the dont_free option only if the relevant packet_ref capability is set
```
TuomasTaipale, line 268: Refer to the full path of `dont_free`?

include/odp/api/spec/packet.h
psavol: Agree, no need to speculate on the implementation method.

include/odp/api/spec/packet.h
```diff
@@ -94,6 +94,20 @@ int odp_packet_alloc_multi(odp_pool_t pool, uint32_t len,
  *
  * Frees the packet into the packet pool it was allocated from.
  *
+ * Packets that participate in packet data sharing (see odp_packet_ref()
+ * and odp_packet_ref_static()) may not get freed back to the packet pool
+ * immediately when this function is called:
```
psavol: Since reference to a packet is a special case, only minimal documentation of it should be here and majority under e.g. odp_packet_ref().

E.g. here we could just say: "In case of packet references, the referenced packet(s) is freed back to the packet pool only after there are no more references to it."

include/odp/api/spec/ipsec.h
JannePeltonen: reworded in v5

include/odp/api/spec/packet.h
JannePeltonen: reworded in v5

include/odp/api/spec/packet.h
JannePeltonen: reworded the whole paragraph in v5

include/odp/api/spec/packet.h
JannePeltonen: reworded in v5

include/odp/api/spec/packet.h
JannePeltonen: removed in v5

include/odp/api/spec/packet.h
JannePeltonen: it already does

include/odp/api/spec/packet.h
JannePeltonen: This paragraph restricts operations on the referenced packets while the bullet points later discuss how things work with referencing packets, so they are about a bit different topics.

include/odp/api/spec/packet.h
JannePeltonen: removed the sentence from v5

include/odp/api/spec/packet.h
JannePeltonen: removed in v5

include/odp/api/spec/packet.h
JannePeltonen: removed in v5

include/odp/api/spec/packet.h
JannePeltonen: changed in v5

include/odp/api/spec/packet_io.h
JannePeltonen: done in v5

include/odp/api/spec/packet_io_types.h
JannePeltonen: added in v5

include/odp/api/spec/packet_io_types.h
JannePeltonen: I do not wee why it would be better. It would just take more space, do did not change it in v5. There are also a few existing instances of doxygen comments in the same line, especially in the traffic manager API header.

include/odp/api/spec/packet_io.h
JannePeltonen: I am not sure what path you mean but added a reference to odp_packet_free_ctrl_set() in v5.

include/odp/api/spec/packet.h
JannePeltonen: Shortened and reworded in v5.

I avoided saying that the packet won't be freed until references to it disappear so that the text cannot be interpreted to mean that everything gets postponed until that point and that the packet could still be used (and be in a not-freed state) until that.

JannePeltonen:
v5: Updated based on review comments. Text clarifications only, no semantic changes.

include/odp/api/spec/packet.h
```diff
@@ -1120,63 +1149,120 @@ odp_packet_t odp_packet_reassemble(odp_pool_t pool, odp_packet_buf_t pkt_buf[],
...
+ * unless otherwise noted. There are restrictions in packet output.
+ *
+ * Referenced packets can be used in ODP API functions that treat packet data
+ * and layout as read-only. Referenced packets can be used in ODP API functions
+ * that consume packets, unless otherwise noted. Consuming a packet behaves
+ * the same way as freeing a packet with respect to shared packet data.
+ *
+ * The packet passed to this function must not reference another packet or
+ * be a static reference (see odp_packet_ref_static()). It can already be
+ * a referenced packet.
```
psavol: Function documentation would be cleared for the (first time) reader, if this limitation is stated in the beginning. I'd move this just after the first paragraph (definition of referenced/referencing) and change:
"... must not reference another packet..." -> "...must not be a referencing packet or a static reference..."

include/odp/api/spec/packet.h
```diff
@@ -1120,63 +1149,120 @@ odp_packet_t odp_packet_reassemble(odp_pool_t pool, odp_packet_buf_t pkt_buf[],
...
  *               created.
  *
  * @param offset Byte offset in 'pkt' at which the shared part is to
  *               begin. Must be in the range 0 ... odp_packet_len(pkt)-1.
  *
- * @param hdr    Handle of the header packet to be prefixed onto the new
- *               reference. Must be a unique reference.
+ * @param hdr    Handle of the header packet to be prefixed
  *
  * @return New reference the reference packet
```
psavol: Something wrong in our current return value documentation. This and odp_packet_ref() return value could be documented as "Handle of the newly created referencing packet".

include/odp/api/spec/packet_io_types.h
psavol: Argee with Tuomas that we should prefer /** style. I think we have cleaned up /**< comments away over the years, but there are still some left.

Rationale being that as soon as someone need to add another sentence into one of those one liner comments. That format becomes ugly/unreadable, and that person would need to modify all other comments into normal  format as well.

JannePeltonen: Changed in v6.

include/odp/api/spec/packet.h
JannePeltonen: Moved and changed in v6

include/odp/api/spec/packet.h
JannePeltonen: Fixed in v6

include/odp/api/spec/ipsec.h
```diff
@@ -434,6 +434,10 @@ int odp_ipsec_out_enq(const odp_packet_t pkt[], int num,
  * transformed packets to the specified output interface (or tm_queue), instead of
  * generating events for those.
  *
+ * Packets that reference or are referenced by other packets may be sent
+ * using this function only if the pktout to be used has the relevant
+ * capability. See odp_pktio_capability_t::packet_ref. XXX dontfree
```
TuomasTaipale: "XXX dontfree" some leftover?

JannePeltonen: removed in v7

include/odp/api/spec/packet.h
```diff
@@ -366,6 +379,9 @@ uint32_t odp_packet_len(odp_packet_t pkt);
  *
  * Returns the current packet level headroom length.
  *
+ * If a packet is a referencing packet and the first data packet byte is shared
```
psavol: "If a packet..." -> "If the packet" = the packet give as parameter

The same fix needed for odp_packet_tailroom() as well.

include/odp/api/spec/packet.h
```diff
@@ -446,6 +465,10 @@ void odp_packet_prefetch(odp_packet_t pkt, uint32_t offset, uint32_t len);
  * headroom -= len
  * data     -= len
  *
+ * The new packet data added in the packet is not initialized and will always
+ * be non-shared, even when the packet is a reference that shares data with
+ * other packets.
```
psavol: Is this comment necessary? Packet that starts with shared data does not have headroom. When headroom is 0,  this function cannot be used anyway.

Push_tail() comment needs the same fix.

include/odp/api/spec/packet.h
```diff
@@ -563,6 +598,10 @@ void *odp_packet_pull_tail(odp_packet_t pkt, uint32_t len);
  * were moved, application must use new packet/segment handles to update
  * data pointers. Otherwise, all old pointers remain valid.
  *
+ * The new packet data added in the packet is not initialized and will always
+ * be non-shared, even when the packet is a reference that shares data with
+ * other packets.
```
psavol: As discussed, no need to highlight that data is not initialized (similarly newly allocated packets are not initialized).
Same comment with extend_tail() and add_data().

include/odp/api/spec/packet_io_types.h
```diff
@@ -1010,6 +1017,37 @@ typedef struct odp_pktin_vector_capability_t {
...
+	/** Static packet references can be consumed */
+	uint8_t static_ref  :1;
+
+	/** Referenced packets can be consumed */
+	uint8_t referenced  :1;
+
+	 /** Referencing packets can be consumed */
+	uint8_t referencing :1;
+
+} odp_pktout_packet_ref_capability_t;
```
psavol: Could we use the same type for both pktout and TM capas? In both cases the flag tells if packet output can consume the packet, and that common part of the documentation could remain here (the first paragraph). The differing part could be documented in the capa documentation using the type (second paragraph about dont_free moved into pktio capa).

Type name could be shortened to odp_pktout_ref_capability_t.

include/odp/api/spec/packet.h
JannePeltonen: fixed in v8

include/odp/api/spec/packet.h
JannePeltonen: Removed that as unnecessary in v8

include/odp/api/spec/packet.h
JannePeltonen: removed in v8

include/odp/api/spec/packet_io_types.h
JannePeltonen: done in v8

include/odp/api/spec/packet_io_types.h
```diff
@@ -1063,6 +1070,13 @@ typedef struct odp_pktio_capability_t {
 	/** LSO capabilities */
 	odp_lso_capability_t lso;
 
+	/** Types of packet references and referenced packets that can be
+	 *  consumed by the pktio. If a bit is not set for a packet type,
+	 *  then packets of that type may be sent only if the pktout supports
+	 *  the 'dont_free' option and the option is set for the packet.
+	 */
+	odp_packet_ref_types_t packet_ref;
```
psavol, line 1080: Since Doxygen uses the first sentence as the short description, it would be good to add a heading to this as well. E.g.

```
/** Supported packet reference types
 *
 *  Types of packet references and referenced packets that ...
```

include/odp/api/spec/traffic_mngr.h
```diff
@@ -609,6 +609,14 @@ typedef struct {
 	 * The value can vary between 0 and ODP_TM_MAX_PRIORITIES.
 	 */
 	uint8_t max_schedulers_per_node;
+
+	/** Types of packet references and referenced packets that can be
+	 *  enqueued to TM. If a bit is not set for a packet type, then
+	 *  packets of that type must not be enqueued to TM. Normal packets
+	 *  can always be enqueued.
+	 */
+	odp_packet_ref_types_t packet_ref;
```
psavol, line 620: Doxygen short description could be added:

```
/** Supported packet reference types
 *
 *  Types of packet references and referenced packets that ...

```

include/odp/api/spec/packet_io_types.h
JannePeltonen: added in v9

include/odp/api/spec/traffic_mngr.h
JannePeltonen: added in v9

psavol (approved):
Reviewed API commits.

example/sysinfo/odp_sysinfo.c
```diff
@@ -641,6 +641,9 @@ static void print_pktio_capa(appl_args_t *appl_args)
 		printf("    lso.proto.tcp_ipv6:            %u\n", capa->lso.proto.tcp_ipv6);
 		printf("    lso.proto.sctp_ipv4:           %u\n", capa->lso.proto.sctp_ipv4);
 		printf("    lso.proto.sctp_ipv6:           %u\n", capa->lso.proto.sctp_ipv6);
+		printf("    packet_ref.static_ref          %u\n", capa->packet_ref.static_ref);
+		printf("    packet_ref.referenced          %u\n", capa->packet_ref.static_ref);
+		printf("    packet_ref.referencing         %u\n", capa->packet_ref.static_ref);
```
MatiasElo: Wrong field names and values.

include/odp/api/spec/packet.h
```diff
@@ -909,6 +944,25 @@ uint32_t odp_packet_seg_data_len(odp_packet_t pkt, odp_packet_seg_t seg);
...
+ * Otherwise the result packet is either a normal or referencing packet
+ * depending on the packets and the implementation.
+ *
+ *  dst		src		result
+ *  -------------------------------------------------
+ *  normal	normal		normal
+ *  normal	referenced	normal or referencing
+ *  normal	referencing	normal or referencing
+ *  normal	static ref	normal or referencing
+ *  referencing	any		referencing
```
MatiasElo: Nitpicking, but better to use spaces than tabs for comment alignment.

platform/linux-generic/odp_packet.c
```diff
@@ -1538,10 +1648,21 @@ void odp_packet_print(odp_packet_t pkt)
...
+				       "    [%d] seg_len %-6" PRIu32 "  seg_data %p  ref_cnt %u\n",
 				       seg_idx, odp_packet_seg_data_len(pkt, seg),
 				       odp_packet_seg_data(pkt, seg), segment_ref(seg_hdr));
 
+		if (seg_hdr->seg_indirect) {
+			odp_packet_hdr_t *ref_seg = seg_hdr->seg_referenced;
+
+			str_len += _odp_snprint(&seg_str[str_len], max_len - str_len,
+						"        referenced seg: seg_data %p "
+						" seg_len %-6" PRIu32 "ref_cnt %u\n",
```
MatiasElo: Space missing (`" PRIu32 "ref_cnt %u\n"`)?

platform/linux-generic/odp_packet.c
```diff
@@ -1203,6 +1287,15 @@ int odp_packet_align(odp_packet_t *pkt, uint32_t offset, uint32_t len,
...
 	}
 
+	if (odp_packet_is_referencing(*pkt)) {
+		/*
+		 * We cannot move the whole packet data as the packet most
+		 * likely has indirect segments and the data of such segments
+		 * is read-only. So we will do something different.
+		 */
+		return packet_align_dyn_ref(pkt, offset, len, align);
+	}
```
TuomasTaipale, line 1302: Could be moved to be done before the above shift etc. logic as this doesn't utilize them.

platform/linux-generic/odp_packet.c
```diff
@@ -1169,6 +1205,54 @@ int odp_packet_rem_data(odp_packet_t *pkt_ptr, uint32_t offset, uint32_t len)
...
+		return -1;
+	_odp_packet_copy_md(packet_hdr(new_pkt), pkt_hdr, 0);
+
+	new_pkt_hdr = packet_hdr(new_pkt);
+	new_addr = (uintptr_t)new_pkt_hdr->seg_data + offset;
+	misalign = new_addr & (align - 1);
+	if (misalign)
+		shift = align - misalign;
+	else
+		shift = 0;
```
TuomasTaipale: Could be removed if `shift` was initialized to zero.

platform/linux-generic/pktio/socket_xdp.c
```diff
@@ -976,7 +976,8 @@ static int sock_xdp_send(pktio_entry_t *pktio_entry, int index, const odp_packet
 		pkt_hdr = packet_hdr(packets[i]);
 		seg_cnt = pkt_hdr->seg_count;
 
-		if (_odp_pool_entry(pkt_hdr->event_hdr.pool) != pool) {
+		if (_odp_pool_entry(pkt_hdr->event_hdr.pool) != pool ||
+		    odp_packet_is_referencing(packets[i])) {
```
TuomasTaipale: Old bug but probably have to also add `odp_packet_has_ref()` check here.

platform/linux-generic/arch/aarch64/odp_crypto_armv8.c
```diff
@@ -779,7 +779,8 @@ int crypto_int(odp_packet_t pkt_in,
 	odp_crypto_generic_session_t *session;
 	odp_packet_t out_pkt;
 
-	if (odp_unlikely(odp_packet_has_ref(pkt_in)))
+	if (odp_unlikely(odp_packet_is_referencing(pkt_in) ||
+			 odp_packet_has_ref(pkt_in)))
```
TuomasTaipale, line 783: Slightly less copy-pasta if there was an internal helper for checking these two.

example/sysinfo/odp_sysinfo.c
JannePeltonen: fixed in v11

include/odp/api/spec/packet.h
JannePeltonen: I am not sure why it is better, but changed anyway in v11

test/validation/api/pktio/pktio.c
```diff
@@ -44,12 +44,16 @@
...
 typedef enum {
 	TEST_WITH_DEF_POOL = 1,
-	TEST_WITH_REFS = 2,
+	TEST_WITH_STATIC_REFS = 2,
+	TEST_WITH_REFFED_PKTS = 4,
+	TEST_WITH_DYN_REFS = 8,
 } test_flag_values_t;
 
-#define NUM_TEST_FLAGS 2
+#define NUM_TEST_FLAGS 5
```
TuomasTaipale: 4 flags now?

platform/linux-generic/odp_packet.c
JannePeltonen: It was not really missing because the left alignment with the 6 char field width together with uint16_t max segment lengths took care of it. But it was not nice.  In v11 I changed the order of ref_cnt and seg_len, so that all ref_cnt fields get aligned. And this time the space is obvious too.

test/validation/api/pktio/pktio.c
```diff
@@ -1298,6 +1395,13 @@ static void do_test_txrx(odp_pktin_mode_t in_mode, int num_pkts,
 			CU_FAIL("failed to open iface");
 			return;
 		}
+
+		CU_ASSERT(odp_pktio_capability(io->id, &capa) == 0);
```
TuomasTaipale: Probably a fatal assert here.

platform/linux-generic/odp_packet.c
JannePeltonen: This is actually on purpose after the shift etc logic since that code could determine that no alignment is needed and just return.

platform/linux-generic/odp_packet.c
JannePeltonen: done in v12

platform/linux-generic/pktio/socket_xdp.c
JannePeltonen: I ended up not changing this in v12. I thought the code might actually work (?) with static references which is why I only added the dyn ref check.

platform/linux-generic/arch/aarch64/odp_crypto_armv8.c
JannePeltonen: Not changed in v12. I do not think it helps much and the checks are not going to be the same once any particular crypto implementation starts supporting refs better. Actually, the null crypto is already a bit different since it can handle referencing packets.

test/validation/api/pktio/pktio.c
JannePeltonen: Fixed in v12.

test/validation/api/pktio/pktio.c
JannePeltonen: changed in v12

test/validation/api/packet/packet_ref.c
```diff
@@ -0,0 +1,1609 @@
...
+
+static void check_seg_info_equal(const packet_seg_info_t *a, const packet_seg_info_t *b)
+{
+	CU_ASSERT(a->total_len == b->total_len);
+	CU_ASSERT(a->num_segs == b->num_segs);
+
+	for (uint32_t n = 0; n < a->num_segs; n++) {
+		CU_ASSERT(a->segs[n].handle == a->segs[n].handle);
+		CU_ASSERT(a->segs[n].data == a->segs[n].data);
+		CU_ASSERT(a->segs[n].len == a->segs[n].len);
```
MatiasElo: Should likely compare to `b`.

platform/linux-generic/pktio/socket_xdp.c
TuomasTaipale: Probably doesn't work, segment ownership is transferred to kernel with the TX submit.

test/validation/api/packet/packet_ref.c
```diff
@@ -0,0 +1,1609 @@
...
+ * Make reasonable assumptions on the underlying ODP implementation on
+ * how segmented packets can be created as ODP API does not have any
+ * guaranteed way to do it. Fail if segmented packets cannot be made.
+ */
+static odp_packet_t alloc_packet(uint32_t len, uint32_t num_segs)
+{
+	uint32_t seg_len;
+	odp_packet_t pkt;
+
+	len += len % num_segs;
```
TuomasTaipale: Not sure if this line works, e.g. len 100 and num_segs 3 results in uneven values. We have `ODPH_ROUNDUP_MULTIPLE()` macro also.

test/validation/api/packet/packet_ref.c
JannePeltonen: fixed in v14

test/validation/api/packet/packet_ref.c
```diff
@@ -0,0 +1,1615 @@
...
+
+#include <odp/helper/odph_api.h>
+
+#include <stdint.h>
+#include <stdlib.h>
+
+#define MAX_SEGS 10
+#define MAX_PKT_NUM 30
+#define MAX_PKT_LEN 100
+uint32_t uarea_size = 8;
```
TuomasTaipale: Could be static as well.

platform/linux-generic/odp_packet.c
```diff
@@ -458,12 +472,26 @@ static inline int is_multi_ref(uint32_t ref_cnt)
...
 }
 
+static inline void free_referenced_seg(odp_packet_hdr_t *hdr)
+{
+	odp_packet_hdr_t *referenced = hdr->seg_referenced;
+	uint32_t ref_cnt;
+
+	ref_cnt = segment_ref_dec(referenced);
+	if (!is_multi_ref(ref_cnt)) {
+		_odp_event_free_sp((_odp_event_hdr_t **)&referenced, 1);
```
MatiasElo: It would be more efficient to use `_odp_event_free()`.

JannePeltonen: changed in v15

test/validation/api/packet/packet_ref.c
JannePeltonen: Fixed in v15

test/validation/api/packet/packet_ref.c
JannePeltonen: fixed in v15

platform/linux-generic/pktio/socket_xdp.c
JannePeltonen: Added the has_ref() check in v15

PR #2289, opened 2025-11-12
[PATCH v2] linux-gen: dpdk: fix the issue of requesting RSS from an unspported device

lifang-zhang:
Only configure RSS when the pktin hash_enable is set and the device supports the requested RSS offload type.

Closes: https://github.com/OpenDataPlane/odp/issues/2280


MatiasElo (approved):
One typo in commit message: `unspported `. Otherwise looks good.

PR #2292, opened 2025-11-27
[PATCH v3] linux-gen: packet: fix synchronization bugs in packet refcount update functions

JannePeltonen:
- fix race condition in packet refcount increment
- fix synchronization in refcount decrement

### DIFF
--- a/platform/linux-generic/odp_packet.c
+++ b/platform/linux-generic/odp_packet.c
@@ -14,6 +14,7 @@
 #include <odp/api/packet_io.h>
 #include <odp/api/proto_stats.h>
 #include <odp/api/timer.h>
+#include <odp/api/sync.h>
 
 #include <odp_parse_internal.h>
 #include <odp_chksum_internal.h>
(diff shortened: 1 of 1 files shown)


platform/linux-generic/odp_packet.c
```diff
@@ -442,18 +443,28 @@ static inline odp_packet_hdr_t *add_segments(odp_packet_hdr_t *pkt_hdr,
...
 static inline void segment_ref_inc(odp_packet_hdr_t *seg_hdr)
 {
-	uint32_t ref_cnt = odp_atomic_load_u32(&seg_hdr->ref_cnt);
+	uint32_t ref_cnt = odp_atomic_fetch_inc_u32(&seg_hdr->ref_cnt);
 
 	/* First count increment after alloc */
-	if (odp_likely(ref_cnt == 0))
-		odp_atomic_store_u32(&seg_hdr->ref_cnt, 2);
-	else
+	if (ref_cnt == 0)
```
TuomasTaipale, line 449: Assuming a scenario with multiple threads taking references and freeing references to the packet, can something happen between fetching to `ref_cnt` and this comparison that would make `ref_cnt` outdated and not actually zero anymore?

JannePeltonen: Yes, something can happen in between but it is no longer a problem after this patch. Correctness of the old code relied on the refcount in the segment not changing between the load and the store. This new code does not care. Please check the reasoning below.

Between the initial atomic fetch inc and the final atomic inc the ref_cnt in the segment is one less than what it should be. But it is still at least one, so no matter what the other threads do, they do not cause the segment to be freed.

One way to look at this is that the atomic fetch inc is either the actual reference count increment or the delayed initialization of the reference count to one (as it was postponed from packet alloc to here). But which one it was we know only after we have done the atomic fetch inc and seen what the reference count was initially. And it if was zero, we know that we have only made the "initialization to 1" and have yet to do the actual reference count increment, which we then do using the atomic increment.

A couple of examples:

Assume that two threads call segment_ref_inc() simultaneously. One of them will be the one that does the atomic fetch inc first. Let's call that the first thread and the other thread the second thread. After the atomic fetch inc by the first thread ref_cnt local in the first thread is zero and ref_cnt in the segment is one. If the atomic fetch inc of the second thread happens now, then the ref_cnt local in the second thread becomes 1, ref_cnt in the segment becomes 2 and the second thread is done. Now the first thread continues and performs the atomic inc, which increments the refcount to 3.

If the second thread did not only call segment_ref_inc() but also segment_ref_dec() before the first thread is done with segment_ref_inc(), then the segment_ref_dec() in the second thread would see ref_cnt == 2 and the surrounding code would not free the packet. The seconf thread would also decrement ref_cnt in the segment to one and the first thread would then increment it back to two, which would be the correct value here.

For segment_ref_dec() in another thread see ref_cnt value of 1 (which would cause the segment to be freed), the other thread would have to omit the call to segment_ref_inc(), which would be a bug.

TuomasTaipale: OK, I have become convinced of the correctness of this.

PR #2293, opened 2025-12-04
[PATCH v6] api: crypto: add snow5g algorithm

nkaithakadan:
Add SNOW5G cipher and integrity algorithms.

### DIFF
--- a/include/odp/api/spec/crypto_types.h
+++ b/include/odp/api/spec/crypto_types.h
@@ -134,6 +134,14 @@ typedef enum {
 	 */
 	ODP_CIPHER_ALG_SNOW3G_UEA2,
 
+	/** NEA4 confidentiality algorithm
+	 *
+	 *  SNOW 5G based NEA4 algorithm.
+	 *
+	 *  IV is 128 bits.
+	 */
+	ODP_CIPHER_ALG_SNOW5G_NEA4,
+
 	/** Confidentiality 128-EEA2 algorithm
 	 *
 	 *  AES-CTR-based 128-EEA2 algorithm.
@@ -340,6 +348,14 @@ typedef enum {
 	 */
 	ODP_AUTH_ALG_SNOW3G_UIA2,
 
+	/** NIA4 integrity algorithm
+	 *
+	 *  SNOW 5G based NIA4 algorithm.
+	 *
+	 *  IV is 128 bits.
+	 */
+	ODP_AUTH_ALG_SNOW5G_NIA4,
+
 	/** Integrity 128-EIA2 algorithm
 	 *
 	 *  AES_CMAC-based 128-EIA2 algorithm.
@@ -511,6 +527,9 @@ typedef union odp_crypto_cipher_algos_t {
 		/** ODP_CIPHER_ALG_SNOW3G_UEA2 */
 		uint32_t snow3g_uea2 : 1;
 
+		/** ODP_CIPHER_ALG_SNOW5G_NEA4 */
+		uint32_t snow5g_nea4 : 1;
+
 		/** ODP_CIPHER_ALG_AES_EEA2 */
 		uint32_t aes_eea2 : 1;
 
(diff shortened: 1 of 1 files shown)


include/odp/api/spec/crypto_types.h
```diff
@@ -134,6 +134,16 @@ typedef enum {
 	 */
 	ODP_CIPHER_ALG_SNOW3G_UEA2,
 
+	/** Confidentiality NEA4 algorithm
```
JannePeltonen: I think the word order is weird here. "NEA4 confidentiality algorithm" would sound better. I know some old API definitions have the same weird word order, but it does not need to be the same here.

include/odp/api/spec/crypto_types.h
```diff
@@ -134,6 +134,16 @@ typedef enum {
...
+	 *
+	 *  SNOW 5G-based NEA4 algorithm.
```
JannePeltonen: I think the dash should be removed, i.e. "SNOW 5G based NEA4 algorithm".

include/odp/api/spec/crypto_types.h
```diff
@@ -134,6 +134,16 @@ typedef enum {
...
+	 *
+	 *  IV (128 bits) should be formatted according to the ETSI/SAGE
+	 *  NEA4 & NIA4 specification:
+	 *  COUNT || BEARER || DIRECTION || 0...0
```
JannePeltonen: I am not sure the format of the IV really needs to be specified. Maybe it could just be said that the IV is 128 bits as how the IV is constructed and tied to upper protocol layers does not seem to be of concern to ODP. Unless we really want to say that not all IV values are acceptable (but would there really be any such values?).

include/odp/api/spec/crypto_types.h
```diff
@@ -134,6 +134,16 @@ typedef enum {
...
+	 */
+	ODP_CIPHER_ALG_SNOW5G_NEA4,
```
JannePeltonen, line 143: I wonder if this naming style (i.e. SNOW5G_NEA4, instead of just SNOW5G or just NEA4) is what we want to do. I am not sure what to think of it.

I know some existing API definitions use that style but it has already lead to some weirdnesses such as having both ODP_CIPHER_ALG_AES_EEA2 and ODP_CIPHER_ALG_AES_CTR, which are actually the same thing (for 128-bit keys). Then there is also the issue that 3GPP names the same underlying algorithm with different names depending on the release (e.g. UEA2, EEA1 and NEA1 are identical from ODP perspective, AFAIK).

include/odp/api/spec/crypto_types.h
```diff
@@ -134,6 +134,16 @@ typedef enum {
...
+
```
JannePeltonen, line 144: If NEA4 and NIA4 are being added, then maybe it would be good to add NCA4 too?

test/validation/api/crypto/test_vector_defs.h
```diff
@@ -2389,6 +2389,54 @@ static crypto_test_reference_t snow3g_uea2_reference[] = {
 	}
 };
 
+static crypto_test_reference_t snow5g_nea4_reference[] = {
```
JannePeltonen: This patch could be removed from this PR as I am not sure if we can add the test vector in ODP at this point. The API change we can add now, but possible addition of test vectors may have to wait.

include/odp/api/spec/crypto_types.h
nkaithakadan: Done.

include/odp/api/spec/crypto_types.h
nkaithakadan: Done

include/odp/api/spec/crypto_types.h
nkaithakadan: Done

test/validation/api/crypto/test_vector_defs.h
nkaithakadan: Done

include/odp/api/spec/crypto_types.h
nkaithakadan: So can we use ODP_CIPHER_ALG_NEA4 and ODP_AUTH_ALG_NIA4 ?

JannePeltonen: We can also keep the proposed name and leave possible renaming of the algorithms (not just this one) for better consistency for future.

include/odp/api/spec/crypto_types.h
JannePeltonen: .Does not have to be in this PR.

PR #2294, opened 2025-12-05
[PATCH v2] test: pipeline: fix coverity warnings

TuomasTaipale:
Fix a few potential memory leaks and double frees. Additionally, use correct size when utilizing `odp_queue_context_set()` in flow deployment.

v2:
- Added reviewed-by tag

### DIFF
--- a/test/performance/pipeline/queue_parser.c
+++ b/test/performance/pipeline/queue_parser.c
@@ -337,6 +337,7 @@ static odp_bool_t queue_parser_init(config_t *config)
 
 			if (ret == -1) {
 				ODPH_ERR("Invalid \"" QUEUE_DOMAIN "\" entry (%d)\n", i);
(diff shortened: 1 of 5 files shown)


test/performance/pipeline/queue_parser.c
```diff
@@ -337,6 +337,7 @@ static odp_bool_t queue_parser_init(config_t *config)
 
 			if (ret == -1) {
 				ODPH_ERR("Invalid \"" QUEUE_DOMAIN "\" entry (%d)\n", i);
+				free_queue_template(&templ);
```
MatiasElo, line 340: Another option could be that the template functions free resources internally in case of a failure.

TuomasTaipale: Yes, I prefer the current way however, less code bloat.

PR #2298, opened 2025-12-15
[PATCH v3] IPsec: fix handling of segmented packets and improve validation testing

JannePeltonen:


test/validation/api/ipsec/ipsec_test_out.c
```diff
@@ -1635,15 +1635,14 @@ static void test_sa_info(void)
...
 	ipsec_test_flags flags;
 
 	memset(&flags, 0, sizeof(flags));
-	flags.display_algo = true;
 	flags.part_flags.test_sa_seq_num = true;
 
-	test_out_in_all(&flags);
-
-	printf("\n  ");
+	test_esp_out_in(&cipher, &auth, &flags);
```
TuomasTaipale, line 1758: Test AH as well with the single combo?

test/validation/api/ipsec/ipsec.c
```diff
@@ -1266,6 +1266,46 @@ void ipsec_test_packet_from_pkt(ipsec_test_packet *test_pkt, odp_packet_t *pkt)
...
+	CU_ASSERT_FATAL(pkt->l3_offset != ODP_PACKET_OFFSET_INVALID);
+	CU_ASSERT_FATAL(pkt->len > pkt->l3_offset);
+	l3_len = pkt->len - pkt->l3_offset;
+	CU_ASSERT_FATAL(ODPH_ETHHDR_LEN + l3_len < sizeof(pkt->data));
+
+	if (pkt->l3_offset != ODPH_ETHHDR_LEN)
+		memmove(&pkt->data[ODPH_ETHHDR_LEN], &pkt->data[pkt->l3_offset], l3_len);
+
+	switch (ODPH_IPV4HDR_VER(pkt->data[ODPH_ETHHDR_LEN])) {
+	case 4:
```
TuomasTaipale: We seem to have `ODPH_IPV4` and `ODPH_IPV6`.

test/validation/api/ipsec/ipsec_test_out.c
```diff
@@ -731,30 +745,52 @@ static void test_esp_out_in(struct cipher_param *cipher,
...
+	if (*iterator >= 0x8)
+		return 0;
+
+	flags->v6           = *iterator & 0x4;
+	flags->tunnel       = *iterator & 0x2;
+	flags->tunnel_is_v6 = *iterator & 0x1;
+	*iterator += 1;
+
+	if (!flags->tunnel && flags->tunnel_is_v6)
+		goto next;
```
TuomasTaipale: Why not `while` or `for`? `goto`s in ODP I believe are mainly used for error/exit jumps.

test/validation/api/ipsec/ipsec_main.c
```diff
@@ -10,6 +10,7 @@
 odp_suiteinfo_t ipsec_sync_suites[] = {
 	{"IPsec-in",  ipsec_suite_sync_init, ipsec_suite_term, ipsec_in_suite},
 	{"IPsec-out", ipsec_suite_sync_init, ipsec_suite_term, ipsec_out_suite},
+	{"out-in",    ipsec_suite_sync_init, ipsec_suite_term, ipsec_out_in_suite},
```
TuomasTaipale, line 13: "IPsec-" prefix to match existing style.

platform/linux-generic/include/odp_packet_continuous.h
```diff
@@ -0,0 +1,104 @@
+/* SPDX-License-Identifier: BSD-3-Clause
+ * Copyright (c) 2025 Nokia
+ */
+
+#include <odp/api/packet.h>
+#include <odp_debug_internal.h>
+
+#ifndef ODP_PACKET_CONTINUOUS_H
+#define ODP_PACKET_CONTINUOUS_H
```
TuomasTaipale: Append `_` to the macro names to match existing style.

platform/linux-generic/include/odp_packet_continuous.h
```diff
@@ -0,0 +1,104 @@
...
+
+/*
+ * Copy back the copied and possibly modified packet data from 'state' to the
+ * original packet or do nothing if copying was not needed and packet data
+ * could be modified in-place. Does not need to be called if packet data was
+ * only read and not written to.
+ *
+ * Returns nonzero on error.
+ */
+static inline int continuous_copy_back(continuous_t *state);
```
TuomasTaipale: Internal headers typically seem to just define the functions directly if inline (instead of declare/define). Also, `_odp_` prefix to match the current style.

Even though name mangling is probably not an issue with these, internal headers seem to still wrap these in `extern "C"` block.

platform/linux-generic/include/odp_packet_continuous.h
```diff
@@ -0,0 +1,104 @@
...
+ * when it is not known if there is a segment boundary in the packet in
+ * the area being read from or written to.
+ */
+
+/* State needed by the API */
+struct continuous_s;
+typedef struct continuous_s continuous_t;
+
+/* Maximum value of the len parameter that may be passed to the functions */
+#define CONTINUOUS_MAX_LEN 40
```
TuomasTaipale: This seems to add just another segment boundary. I guess one way would be to add another function to query needed copy space and caller would pass pointer to array it allocated somehow via `copy`.

platform/linux-generic/include/odp_packet_continuous.h
```diff
@@ -0,0 +1,104 @@
...
+
+/* Implementation, not to be accessed directly --------------------------------- */
+
+typedef struct continuous_s {
+	bool copied;		/* data from a packet has been copied to our data buffer */
+	odp_packet_t pkt;       /* packet from which data was copied */
+	uint32_t offset;	/* offset of the copied packet data in the packet */
+	uint32_t len;		/* length of the copied packet data */
+	uint8_t copy[CONTINUOUS_MAX_LEN]; /* copy of packet data, to be written back */
+} continuous_t;
```
TuomasTaipale: `bool` could be `odp_bool_t` and seemingly would benefit from an init function that sets the value to false.

Field comments could also be on top of the fields to match current style.

test/validation/api/ipsec/ipsec_test_out.c
JannePeltonen: AH added in v2

test/validation/api/ipsec/ipsec.c
JannePeltonen: used those in v2

platform/linux-generic/include/odp_packet_continuous.h
JannePeltonen: done in v2

platform/linux-generic/include/odp_packet_continuous.h
JannePeltonen: I think we have a couple of instances of static, non-inline functions defined in internal header. There is only one user of this header, so defining the static functions in it should not be an issue for now.

The _odp prefix is only used for things exposed to the application, but these internal static functions are not exposed. We already have some similar internal static functions without the _odp prefix.,

platform/linux-generic/include/odp_packet_continuous.h
JannePeltonen: Changed the bool to odp_bool_t in v2, but left the comments as they are as this is more concise and IMHO as readable. And there are existing instances of comments being in the same line.

test/validation/api/ipsec/ipsec_test_out.c
JannePeltonen: Changed to a while loop in v2

platform/linux-generic/include/odp_packet_continuous.h
JannePeltonen: I ended up moving the functions to the ipsec implementation file.

test/validation/api/ipsec/ipsec_main.c
JannePeltonen: I did not change this in v2. All the tests are IPsec tests, so the IPsec prefix in the name does not really add information and with this naming I could keep the alignment of the columns without having to touch the existing lines.

platform/linux-generic/include/odp_packet_continuous.h
JannePeltonen: It does not add a "segment boundary", but rather a limitation that the helper function can be used only for blocks of 40 bytes or less, without changing the max len. I did this to keep things simpler and more concise at the call sites. Now that I moved the helper functions in the IPsec implementation file, there is less need to have the functions very generic and flexible and thus did not change them in v2.

TuomasTaipale (approved):

PR #2300, opened 2025-12-19
[PATCH v3] api: crypto: clarify output_pool session parameter and specify its default value

JannePeltonen:
This PR assumes that PR 2258 "api: crypto: remove legacy operation type" has been merged.

Clarify that the output_pool crypto session parameter is unused and will be deprecated later. Specify that the parameter has default value of ODP_PACKET_INVALID and clarify that that is the only valid value for the parameter.


JannePeltonen:
> Just noticed odp_ipsec_cache.c still sets the pool, does that require changes?

It does not _require_ changes right now since setting output_pool to ODP_POOL_INVALID (i.e. to the default value) is still ok. Setting of the output_pool should be removed from the file, but the problem is that the code is written without the odp_crypto_session_param_init() function and therefore the output_pool setting cannot be simply removed but a bit bigger change that introduces the init function call is needed. I did not want to do that as part of this API commit but left it for the future. But yes, the pool setting needs to be removed at some point.

PR #2301, opened 2025-12-19
[PATCH v6] api: increment ODP API version to 1.49.0

MatiasElo:
V2:
- Added commit message
- Included #2300

V3:
- Removed packet reference related changes

### DIFF
--- a/CHANGELOG
+++ b/CHANGELOG
@@ -1,5 +1,118 @@
 # Changelog
 
+## OpenDataPlane (1.49.0.0)
+
+### Backward compatible API changes
+
+#### Barrier
+* Clarify that a barrier must be initialized before first use and can be
+reinitialized if not in use.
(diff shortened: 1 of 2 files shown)


CHANGELOG
```diff
@@ -1,5 +1,119 @@
...
 
+## OpenDataPlane (1.49.0.0)
+
+### Backward compatible API changes
+
+#### Barrier
+* Clarify that a barrier must be initialized before first use and can be
+reinitialized if not in use.
+* Explicitly state that a memory barrier can be reused without initializing it
+again.
```
psavol: "memory barrier": This change is about thread execution barriers. Use just "barrier" or "thread execution barrier".

MatiasElo: Fixed in v4.

PR #2303, opened 2026-01-19
[PATCH v2] m4: dpdk: check dpdk version during configure

MatiasElo:


m4/odp_dpdk.m4
```diff
@@ -2,19 +2,20 @@
...
 #
 
-# ODP_DPDK (DPDK_SHARED, ACTION-IF-FOUND, ACTION-IF-NOT-FOUND)
-# -----------------------------------------------------------------------
+# ODP_DPDK (DPDK_VERSION, DPDK_SHARED, ACTION-IF-FOUND, ACTION-IF-NOT-FOUND)
+# --------------------------------------------------------------------------
 # Configure DPDK using pkg-config information
 AC_DEFUN([ODP_DPDK], [dnl
-dpdk_shared="$1"
+dpdk_version="$1"
```
JannePeltonen: I think naming this dpdk_min_version would make this more clear.

platform/linux-generic/m4/odp_dpdk.m4
```diff
@@ -47,12 +47,6 @@ then
     ODP_DPDK([$pktio_dpdk_min_version], [$dpdk_shared], [],
 	     [AC_MSG_FAILURE([can't find DPDK])])
 
-    case "${host}" in
-      i?86* | x86*)
-	DPDK_CFLAGS="${DPDK_CFLAGS} -msse4.2"
```
JannePeltonen, line 51: I wonder why this has been here in the fist place and what caused it to be unnecessary. But it does look unnecessary and I would expect the packet config file of DPDK to include any special compilation flags if needed. Maybe the commit message could be more descriptive?

m4/odp_dpdk.m4
MatiasElo: Changed in v2.

platform/linux-generic/m4/odp_dpdk.m4
MatiasElo: Improved commit message in v2.

PR #1969, opened 2023-12-08
[PATCH v2] api: timer: require an array of timeout events in odp_timer_periodic_start()

JereLeppanen:
With a periodic timer, require application to provide an array of timeout events in odp_timer_periodic_start(). Add a new API function odp_timer_periodic_events(), which returns the number of events.

PR #2309, opened 2026-02-04
[PATCH v4] api: timer: cancellation event

TuomasTaipale:
Draft for cancellation event for periodic timers. Alternative to #1969.

v2:
- Reworded (now backward incompatible and won't compile until implementation and tests updated)

v4:
- Further reworded, ownership flow clarified, introduced a new function for cancel finalization and ownership transfer, ack signature reverted back to original

### DIFF
--- a/include/odp/api/spec/timer.h
+++ b/include/odp/api/spec/timer.h
@@ -1,6 +1,6 @@
 /* SPDX-License-Identifier: BSD-3-Clause
  * Copyright (c) 2013-2018 Linaro Limited
- * Copyright (c) 2019-2023 Nokia
+ * Copyright (c) 2019-2026 Nokia
  */
 
 /**
@@ -246,7 +246,9 @@ int odp_timer_pool_info(odp_timer_pool_t timer_pool,
  *
  * When timer expires, the timeout event defined in timer start parameters (see
  * odp_timer_start_t::tmo_ev or odp_timer_periodic_start_t::tmo_ev) is sent into the provided
- * destination queue.
+ * destination queue. With periodic timers and non-atomic-scheduled destination queues, source
+ * ordering of timeouts may change due to application processing order. In this case, it is up to
+ * the application to ensure acknowledgment of all timeouts.
  *
  * The provided user pointer value is copied into timeout events when the event type is
  * ODP_EVENT_TIMEOUT. The value can be retrieved from an event with odp_timeout_user_ptr() call.
@@ -266,8 +268,9 @@ odp_timer_t odp_timer_alloc(odp_timer_pool_t timer_pool, odp_queue_t queue, cons
  * Frees a previously allocated timer. The timer must be inactive when calling this function.
  * In other words, the application must cancel an active single shot timer (odp_timer_cancel())
  * successfully or wait it to expire before freeing it. Similarly for an active periodic timer, the
- * application must cancel it (odp_timer_periodic_cancel()) and receive the last event from
- * the timer (odp_timer_periodic_ack()) before freeing it.
+ * application must cancel it (odp_timer_periodic_cancel()), handle remaining events delivered by
+ * the timer (odp_timer_periodic_ack()) and finalize the cancellation
+ * (odp_timer_periodic_cancel_finalize()) before freeing it.
  *
  * The call returns failure only on non-recoverable errors. Application must not use the timer
  * handle anymore after the call, regardless of the return value.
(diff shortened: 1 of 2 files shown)


include/odp/api/spec/timer.h
```diff
@@ -334,8 +336,7 @@ int odp_timer_restart(odp_timer_t timer, const odp_timer_start_t *start_param);
  * Starts a timer that delivers timeout events periodically to the destination queue starting
  * from the first expiration time provided. The timer must have been allocated from a pool of
  * periodic timers. The timer must not be active when calling this function. After a successful
- * call, the timer remains active until it is cancelled and all its timeout events have been
- * acknowledged.
+ * call, the timer remains active until it is cancelled and all its events have been acknowledged.
```
JannePeltonen: This seems fine, but there appears to be an inconsistency in the existing description of odp_timer_free().

The text for odp_timer_free() says "The timer must be inactive when calling this function", which is fine. But it also says: "Similarly for an active periodic timer, the application must cancel it (odp_timer_periodic_cancel()) and receive the last event from the timer (odp_timer_periodic_ack()) before freeing it". The latter sentence, which tries to explain what "inactive" means, does not require acking the last event or other remaining events.

include/odp/api/spec/timer.h
```diff
@@ -344,8 +345,8 @@ int odp_timer_restart(odp_timer_t timer, const odp_timer_start_t *start_param);
  * Periodic timers cannot be restarted. If the period needs to be changed, the timer is first
  * cancelled and then started again with new parameters.
  *
- * Application must acknowledge each timeout event with odp_timer_periodic_ack() call. The call
- * should be made as soon as possible after receiving the event.
+ * Application must acknowledge each event with odp_timer_periodic_ack() call. The call should be
```
JannePeltonen, line 349: Maybe changing "timeout event" to just "event" in various places is not needed now that the last event must also be a timeout event.

include/odp/api/spec/timer.h
```diff
@@ -363,40 +364,43 @@ int odp_timer_restart(odp_timer_t timer, const odp_timer_start_t *start_param);
...
  *
  * @retval 2  Success, the last event from a cancelled timer. The call did not consume
- *            the event.
+ *            the event. odp_timer_periodic_start_t::tmo_ev returned in 'tmo_ev'.
  * @retval 1  Success, an event from a cancelled timer. The call consumed the event.
  * @retval 0  Success, the call consumed the event.
  * @retval <0 Failure, the call did not consume the event.
  */
-int odp_timer_periodic_ack(odp_timer_t timer, odp_event_t tmo_ev);
+int odp_timer_periodic_ack(odp_timer_t timer, odp_event_t ev, odp_event_t *tmo_ev);
```
JannePeltonen: I am not sure the tmo_ev output parameter is very good here. It is only needed in the non-time-critical timer cancellation sequence but not in the vast majority of ack calls. It might add a slight overhead in the normal ack calls.

An application should be able to remember what timeout event it provided in the timer start routine, but if not it can also put the event handle in the user pointer of the last event or in the memory area pointed to be the user pointer. Or there could event be an ODP API function to return the timeout event when given a timer or cancelled timer (but that does not seem so good either).

include/odp/api/spec/timer_types.h
```diff
@@ -503,10 +503,22 @@ typedef struct odp_timer_periodic_start_t {
 	/** Timeout event
 	 *
 	 *  This event is enqueued to the destination queue when the timer expires. The event type
-	 *  must be ODP_EVENT_TIMEOUT.
+	 *  must be ODP_EVENT_TIMEOUT. Event ownership is transferred to the successfully started
+	 *  timer until the timer is cancelled and all of its events acknowledged.
```
JannePeltonen: I think this clarifies the ownership of the timeout event and maybe makes it more ok to have the same event appear more than once in the timer queue in corner cases. But there is an inconsistency in the API that odp_timer_periodic_ack() spec says that the event being acked is consumed in certain cases. But if the ownership remains in the timer, the application does not own the event and the ack call cannot consume it. I think the ack speck needs to be updated for the clarified ownership model.

include/odp/api/spec/timer.h
JannePeltonen: The (existing) text says " Each event from a periodic timer must be acknowledged with this call" but it also says that the same event can appear in the application queue more than once. This makes is somewhat ambiguous whether a "duplicate" event must be acked multiple times or just once.  I know the intention is to call the ack for each time a timeout event is received from the queue, even if the same event is received multiple times, but text is not as clear as it could be.

shijith: Same comment as @JannePeltonen . Do we need to return tmo_ev for the last event ? Since application manages the lifecycle of both events, it should store the event pointers for cancel. Plus the driver has ownership of the events only till cancel and _ack of cancel_ev comes after that.

TuomasTaipale: With the current periodic timer API, ownership of the timeout event has been slightly fuzzy. Several workers and the implementation could have valid access to the same event simultaneously.

With v4, this should be now more clear: implementation is now specified to own the timeout and cancel events beginning from successful timer start call until cancellation finalization (application is allowed to call some read-only functions for the timeout in between). There's a new function `odp_timer_periodic_cancel_finalize()` to be called by the application after timer cancel and remaining event acking, which returns ownership of the events back to application (meaning implementation needs to be able to return the events passed when starting the timer).

PR #2310, opened 2026-02-04
[PATCH v7] api: timer: timer busy

TuomasTaipale:
Draft for new "resources busy" error code for oneshot timer start/restart. Still open if periodic start should also have the new error code.

v2:
- Added timer busy also to periodic timer start

v4:
- Rebased
- Added validation tests and modified relevant performance tests accordingly
- Ready for review

v5:
- Rebased
- Matias' comment

v6:
- Rebased
- Added review tags

### DIFF
--- a/test/performance/odp_timer_accuracy.c
+++ b/test/performance/odp_timer_accuracy.c
@@ -103,7 +103,7 @@ typedef struct ODP_ALIGNED_CACHE {
 	uint64_t num_exact;
 	uint64_t num_after;
 
-	uint64_t num_too_near;
+	uint64_t num_retry;
 
 } test_stat_t;
 
@@ -983,7 +983,7 @@ static int print_stat(test_global_t *test_global)
 		stat->num_before += s[i].num_before;
 		stat->num_exact += s[i].num_exact;
 		stat->num_after += s[i].num_after;
-		stat->num_too_near += s[i].num_too_near;
+		stat->num_retry += s[i].num_retry;
 
 		if (s[i].nsec_before_min < stat->nsec_before_min) {
 			stat->nsec_before_min = s[i].nsec_before_min;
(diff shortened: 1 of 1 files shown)


include/odp/api/spec/timer.h
```diff
@@ -291,7 +291,10 @@ int odp_timer_free(odp_timer_t timer);
  * parameter to select between the tick types. Current time of the timer pool can be read with
  * odp_timer_current_tick().
  *
- * The timer is not started when a failure is returned.
+ * The timer is not started when a failure is returned. A failure is returned when expiration time
+ * is too near to the current time (#ODP_TIMER_TOO_NEAR), too far from the current time
+ * (#ODP_TIMER_TOO_FAR) or other failure (#ODP_TIMER_FAIL). The call can also fail due to resources
+ * being temporarily busy. Start is likely to succeed after enough resources are available.
```
MatiasElo: Perhaps `... being temporarily busy (#ODP_TIMER_BUSY).` Same comment also for the other functions.

test/performance/odp_timer_stress.c
```diff
@@ -716,17 +724,31 @@ static int process_single_shot(void *args)
 }
 
 static void start_periodic(odp_timer_pool_t tmr_pool, odp_timer_t tmr, odp_event_t tmo,
-			   uint64_t mul)
+			   uint64_t mul, stats_t *stats)
 {
 	const odp_timer_periodic_start_t start = { .first_tick = 0U, .freq_multiplier = mul,
 						   .tmo_ev = tmo };
-	const int ret = odp_timer_periodic_start(tmr, &start);
+	uint32_t retry = MAX_RETRY;
```
MatiasElo: Why the number of retries is only limited for periodic timers?

test/validation/api/timer/timer.c
```diff
@@ -2585,8 +2591,8 @@ static int worker_entrypoint(void *arg ODP_UNUSED)
 		tt[i].ev = ODP_EVENT_INVALID;
 
 		rc = odp_timer_start(tt[i].tim, &start_param);
-		if (rc == ODP_TIMER_TOO_NEAR) {
-			ODPH_ERR("Missed tick, setting timer\n");
+		if (rc == ODP_TIMER_TOO_NEAR || rc == ODP_TIMER_BUSY) {
+			ODPH_ERR("Missed tick or resources busy, setting timer: %d\n", rc);
```
MatiasElo, line 2628: Perhaps retry loops could be added for `ODP_TIMER_BUSY` into `worker_entrypoint()`, as it's the only multithreaded validation test we have.

TuomasTaipale: This was discussed also earlier but decided not to add retries as it then seemingly starts to affect the handling of already started timers later in the test case (probably that's why there's no retry handling for `ODP_TIMER_TOO_NEAR` case either).

PR #2311, opened 2026-02-09
[PATCH v9] api: timer: add flow control parameter for periodic timer timeout events

MatiasElo:
V3:
- Improved validation tests

V6:
- Rebase

V7:
- Updated sysinfo example

### DIFF
--- a/example/sysinfo/odp_sysinfo.c
+++ b/example/sysinfo/odp_sysinfo.c
@@ -787,6 +787,8 @@ static void print_timer_capa(appl_args_t *appl_args)
 		       capa->periodic.max_freq_hz.integer,
 		       capa->periodic.max_freq_hz.numer,
 		       capa->periodic.max_freq_hz.denom);
+		printf("      min_pending_tmo:    %" PRIu32 "\n", capa->periodic.min_pending_tmo);
+		printf("      max_pending_tmo:    %" PRIu32 "\n", capa->periodic.max_pending_tmo);
 		printf("    timer pool tick info (max_res)\n");
 		printf("      freq:               %" PRIu64 " %" PRIu64 "/%" PRIu64 " Hz\n",
 		       info->tick_info.freq.integer,


test/validation/api/timer/timer.c
```diff
@@ -267,6 +271,15 @@ static int check_periodic_sched_support(void)
 	return ODP_TEST_INACTIVE;
 }
 
+static int check_periodic_sched_flow_control_support(void)
+{
+	if (global_mem->timer_capa.periodic.max_pools && global_mem->timer_capa.queue_type_sched &&
```
TuomasTaipale: When merging, need to remember to use the new support bits introduced in #2323 (instead of `max_pools`).

test/validation/api/timer/timer.c
```diff
@@ -275,6 +288,15 @@ static int check_periodic_plain_support(void)
 	return ODP_TEST_INACTIVE;
 }
 
+static int check_periodic_plain_flow_control_support(void)
+{
+	if (global_mem->timer_capa.periodic.max_pools && global_mem->timer_capa.queue_type_plain &&
```
TuomasTaipale: When merging, need to remember to use the new support bits introduced in #2323 (instead of max_pools).

test/validation/api/timer/timer.c
```diff
@@ -3347,6 +3375,224 @@ static void timer_test_periodic(odp_queue_type_t queue_type, int use_first, int
...
+	const uint32_t max_num = 1000;
+	odp_event_t tmo_tbl[max_num];
+	/* Test frequency: 1x 1000 Hz, or 1x min/max_base_freq */
+	const uint64_t multiplier = 1;
+	odp_fract_u64_t base_freq = {1000, 0, 0};
+	const odp_timer_clk_src_t clk_src = test_global->clk_src;
+
+	min_freq = odp_fract_u64_to_dbl(&timer_capa->periodic.min_base_freq_hz);
+	max_freq = odp_fract_u64_to_dbl(&timer_capa->periodic.max_base_freq_hz);
+	CU_ASSERT(min_freq <= max_freq);
```
TuomasTaipale: `timer_test_periodic()` tests the validity of base and periodic capa output pretty extensively so whatever is tested there could be avoided here (already a long test) -> this assert could be dropped.

test/validation/api/timer/timer.c
```diff
@@ -3347,6 +3375,224 @@ static void timer_test_periodic(odp_queue_type_t queue_type, int use_first, int
...
+					periodic_capa.base_freq_hz.denom);
+		}
+	} else {
+		CU_ASSERT(base_freq.integer == periodic_capa.base_freq_hz.integer);
+		CU_ASSERT(base_freq.numer   == periodic_capa.base_freq_hz.numer);
+		CU_ASSERT(base_freq.denom   == periodic_capa.base_freq_hz.denom);
+	}
+
+	CU_ASSERT(periodic_capa.res_ns > 0);
+	CU_ASSERT(periodic_capa.max_multiplier >= multiplier);
```
TuomasTaipale: Already checked in `timer_test_periodic()`, could be avoided here.

test/validation/api/timer/timer.c
```diff
@@ -3347,6 +3375,224 @@ static void timer_test_periodic(odp_queue_type_t queue_type, int use_first, int
...
+	timer_param.periodic.base_freq_hz = base_freq;
+	timer_param.periodic.max_multiplier = multiplier;
+
+	/* Use minimum supported value for pending timeouts */
+	max_pending_tmo = ODPH_MAX(timer_capa->periodic.min_pending_tmo, 1U);
+	timer_param.periodic.max_pending_tmo = max_pending_tmo;
+
+	if (max_pending_tmo >= max_num)
+		printf("\n    Test limited to %" PRIu32 " events, while max_pending_tmo is "
+		       "%" PRIu32 "\n", max_num, max_pending_tmo);
```
TuomasTaipale: Is there a point testing the rest if the test can't exhaust the pending window (i.e. just return here)?

test/validation/api/timer/timer.c
MatiasElo: Yep, it's better to merge the more complex changes first and then rebase this one.

test/validation/api/timer/timer.c
MatiasElo: True, removed in v4.

test/validation/api/timer/timer.c
MatiasElo: Removed in v4.

test/validation/api/timer/timer.c
MatiasElo: This check was made a test condition in v4.

PR #2314, opened 2026-02-13
[PATCH v2] validation: pktio: misc improvements

MatiasElo:


test/validation/api/pktio/pktio.c
```diff
@@ -5802,6 +5812,31 @@ static int pktio_check_pktin_event_sched(void)
...
+		return -1;
+
+	if (odp_pktio_capability(pktio, capa) != 0) {
+		(void)odp_pktio_close(pktio);
+		return -1;
+	}
+
+	(void)odp_pktio_close(pktio);
+
+	return 0;
```
TuomasTaipale: A bit simpler if capa return call is saved, packet I/O closed and then capa return value returned.

test/validation/api/pktio/pktio.c
```diff
@@ -5864,6 +5899,21 @@ static int pktio_suite_init(pkt_segmented_e pool_segmentation)
...
+			return -1;
+		if (pktio_capa(i, ODP_PKTIN_MODE_SCHED, ODP_PKTOUT_MODE_DIRECT,
+			       &global.iface[i].capa.sched_direct))
+			return -1;
+		if (pktio_capa(i, ODP_PKTIN_MODE_SCHED, ODP_PKTOUT_MODE_QUEUE,
+			       &global.iface[i].capa.sched_queue))
+			return -1;
+		if (pktio_capa(i, ODP_PKTIN_MODE_QUEUE, ODP_PKTOUT_MODE_DIRECT,
+			       &global.iface[i].capa.queue_direct))
+			return -1;
```
TuomasTaipale, line 5607: Matter of taste, but I guess these could be also OR'd in a single `if`.

test/validation/api/pktio/pktio.c
MatiasElo: Changed in V2.

test/validation/api/pktio/pktio.c
MatiasElo: Changed in V2.

PR #2335, opened 2026-04-10
[PATCH v10] api: crypto: add ZUC 256 algorithm

nkaithakadan:
Add ZUC256 cipher and integrity algorithm.

### DIFF
--- a/include/odp/api/spec/crypto_types.h
+++ b/include/odp/api/spec/crypto_types.h
@@ -172,6 +172,14 @@ typedef enum {
 	 */
 	ODP_CIPHER_ALG_ZUC_EEA3,
 
+	/** NEA6 confidentiality algorithm
+	 *
+	 *  256-NEA6 algorithm.
+	 *
+	 *  Key length is 256 bits. IV size is 16 bytes.
+	 */
+	ODP_CIPHER_ALG_ZUC_NEA6,
+
 	/** SNOW-V stream cipher */
 	ODP_CIPHER_ALG_SNOW_V,
 
@@ -387,6 +395,14 @@ typedef enum {
 	 */
 	ODP_AUTH_ALG_ZUC_EIA3,
 
+	/** NIA6 integrity algorithm
+	 *
+	 *  256-NIA6 algorithm.
+	 *
+	 *  Key length is 256 bits. IV size is 16 bytes.
+	 */
+	ODP_AUTH_ALG_ZUC_NIA6,
+
 	/** SNOW-V-GCM AEAD algorithm
 	 *
 	 *  SNOW-V-GCM provides both authentication and encryption. This auth
@@ -536,6 +552,9 @@ typedef union odp_crypto_cipher_algos_t {
 		/** ODP_CIPHER_ALG_ZUC_EEA3 */
 		uint32_t zuc_eea3    : 1;
 
+		/** ODP_CIPHER_ALG_ZUC_NEA6 */
+		uint32_t zuc_nea6    : 1;
+
 		/** ODP_CIPHER_ALG_SNOW_V */
 		uint32_t snow_v      : 1;
 
@@ -637,6 +656,9 @@ typedef union odp_crypto_auth_algos_t {
 
 		/** ODP_AUTH_ALG_ZUC_EIA3 */
 		uint32_t zuc_eia3    : 1;
+
+		/** ODP_AUTH_ALG_ZUC_NIA6 */
+		uint32_t zuc_nia6    : 1;
 
 		/** ODP_AUTH_ALG_SNOW_V_GCM */
 		uint32_t snow_v_gcm : 1;
--- a/example/sysinfo/odp_sysinfo.c
+++ b/example/sysinfo/odp_sysinfo.c
@@ -197,6 +197,8 @@ static const char *cipher_alg_name(odp_cipher_alg_t cipher)
 		return "aes_eea2";
 	case ODP_CIPHER_ALG_ZUC_EEA3:
 		return "zuc_eea3";
+	case ODP_CIPHER_ALG_ZUC_NEA6:
+		return "zuc_nea6";
 	case ODP_CIPHER_ALG_SNOW_V:
 		return "snow_v";
 	case ODP_CIPHER_ALG_SNOW_V_GCM:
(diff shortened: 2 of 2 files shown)


nkaithakadan:
@JannePeltonen Could you review these changes

include/odp/api/spec/crypto_types.h
```diff
@@ -166,11 +166,23 @@ typedef enum {
...
+	 *
+	 *  256-bit key length support is experimental and subject to
+	 *  change. The following variants may be supported:
+	 *
 	 *  - ZUC-256 with 16 byte IV as specified in
 	 *        "An Addendum to the ZUC-256 Stream Cipher",
 	 *        https://eprint.iacr.org/2021/1439
 	 */
-	ODP_CIPHER_ALG_ZUC_EEA3,
+	ODP_CIPHER_ALG_ZUC_NEA6,
```
JannePeltonen, line 181: This change is not really adding a new algorithm in the API as the commit message says but just moves the currently defined 256-bit ZUC variant with 16-byte-IV from ODP_CIPHER_ALG_ZUC_EEA3 to ODP_CIPHER_ALG_ZUC_NEA6.

But, AFAIK, the algorithm defined in the referenced document is not exactly the same as NEA6 (unless things have recently changed), so adding a new algorithm name is fine, but it should not reference that document.

If we are adding ..._NEA6 and ..._NIA6 then maybe ..._NCA6 should be added too.

nkaithakadan: Thanks Janne.
I am fine with removing the referenced document.  I only mentioned it because it describes a ZUC‑256 variant with a 256‑bit key and a fixed 16‑byte IV.
Would you prefer that I remove the reference entirely ? And keep till Line number 181.

JannePeltonen: I think this patch could simply leave the EEA3/EIA3 spec unchanged, i.e. leave the 256-bit ZUC variant and the reference to the doc there and just add the new algorithm enums. Then we can later, in a separate PR, either remove the old 256-bit variants or move them to their own algorithm enums.

nkaithakadan: Done.

include/odp/api/spec/crypto_types.h
```diff
@@ -172,6 +172,23 @@ typedef enum {
...
+	 *  Key length is 256 bits. IV size is 16 bytes.
+	 */
+	ODP_CIPHER_ALG_ZUC_NEA6,
+
+	/** NCA6 AEAD algorithm
+	 *
+	 *  256-NCA6 algorithm.
+	 *
+	 *  Key length is 256 bits. IV size is 16 bytes.
+	 *  Integrity tag size is 16 bytes.
```
JannePeltonen, line 188: I do not think there is a need to fix the tag size to 16 bytes in the API here (you do not have it in NIA6 either) as the 3GPP spec seems to allow other tag sizes too. The tag size is a parameter in session creation and auth capabilities indicate the supported tag size (so an implementation could support only 16 bytes).

include/odp/api/spec/crypto_types.h
```diff
@@ -172,6 +172,23 @@ typedef enum {
...
+	 */
+	ODP_CIPHER_ALG_ZUC_NCA6,
```
JannePeltonen, line 190: ODP API handles AEAD algorithms in a bit cumbersome way. For each AEAD algorithm (see e.g. AES-GCM for an example), ODP API defines a cipher algorithm name (e.g. ODP_CIPHER_ALG_AES_GCM) and an auth algorithm name (e.g. ODP_AUTH_ALG_AES_GCM) and requires that they are used together. For instance, if a crypto session is configured with the ODP_CIPHER_ALG_AES_GCM cipher algorithm, then the auth algorithm must be ODP_AUTH_ALG_AES_GCM, and vice versa.

So, to support NCA6, you need to add the corresponding auth algorithm name ODP_AUTH_ALG_ZUC_NCA6 and the API text that requires the "cipher" and "auth" algorithms to be used together.

include/odp/api/spec/crypto_types.h
```diff
@@ -172,6 +172,23 @@ typedef enum {
 	 */
 	ODP_CIPHER_ALG_ZUC_EEA3,
 
+	/** NEA6 confidentiality algorithm
```
JannePeltonen, line 175: The commit message could be improved.

I am not sure what the word 'align' in the subject line tires to convey. I think the subject line could be e.g.:
"api: crypto: add ZUC-256 NEA6, NIA6 and NCA6 algorithm"

The I think the commit message body could simply say something like this:
"Add 3GPP defined ZUC based confidentiality, integrity and AEAD algorithms with 256 bit key."
Now the body text is weird since the current API already has _a_ variant of ZUC-256 with 16 byte IV and this commit just introduces another, subtly different variant. Also, I do not think there is anything improper in the current API as the commit message might imply.

JannePeltonen (commented):
The new algorithms should be added to the sysinfo example app and to crypto validation tests (I do not mean actual test vectors, but inclusion of the algorithms in the combined test). You can add such commits in this PR, or if you prefer, I can make the changes in a different PR. Just let me know.

include/odp/api/spec/crypto_types.h
JannePeltonen: It would also be ok to leave NCA6 out of this PR.

nkaithakadan:
> The new algorithms should be added to the sysinfo example app and to crypto validation tests (I do not mean actual test vectors, but inclusion of the algorithms in the combined test). You can add such commits in this PR, or if you prefer, I can make the changes in a different PR. Just let me know.

Hi Janne, I’m currently on vacation. I’ll add the suggested changes by EOD, if possible.

JannePeltonen:
> Hi Janne, I’m currently on vacation. I’ll add the suggested changes by EOD, if possible.

There is no hurry, as merging this will anyway have to wait until the next API release.

include/odp/api/spec/crypto_types.h
nkaithakadan: Done

include/odp/api/spec/crypto_types.h
nkaithakadan: NCA6 is moved out from this PR

nkaithakadan:
Hi Janne,
I believe all comments have been addressed. Could you please rereview ?

JannePeltonen (approved):

psavol (approved):

PR #2336, opened 2026-04-13
[PATCH v2] test/dyn_workers: fix teardown order

PavanNikhilesh:
Reverse the process teardown loop so secondaries call odp_term_local before the primary calls odp_term_global, preventing resource teardown ordering errors.


TuomasTaipale (approved):
Thanks for the patch. Commit subject line could be changed to
`test: dyn_workers: fix teardown order`
to match existing style, otherwise OK.

Reviewed-by: Tuomas Taipale \<tuomas.taipale@nokia.com\>

PR #2337, opened 2026-04-13
[PATCH v2] linux-gen: dpdk: improve code sharing with odp-dpdk

MatiasElo:


platform/linux-generic/pktio/dpdk_common.c
```diff
@@ -247,6 +247,101 @@ int _odp_dpdk_pktout_stats_common(uint16_t port_id ODP_UNUSED, uint32_t index OD
...
+	memset(&link_info, 0, sizeof(odp_pktio_link_info_t));
+
+	ret = rte_eth_dev_flow_ctrl_get(port_id, &fc_conf);
+	if (ret) {
+		if (ret != -ENOTSUP) {
+			_ODP_ERR("rte_eth_dev_flow_ctrl_get() failed: %d\n", ret);
+			return -1;
+		}
+		_ODP_DBG("rte_eth_dev_flow_ctrl_get() not supported\n");
+		link_info.pause_rx = ODP_PKTIO_LINK_PAUSE_UNKNOWN;
```
JannePeltonen, line 283: I think previously the code returned OFF instead of UNKNOWN in this case. I wonder if we should assume OFF if a driver does not support flow_ctrl_get?

platform/linux-generic/pktio/dpdk_common.c
```diff
@@ -0,0 +1,56 @@
+/* SPDX-License-Identifier: BSD-3-Clause
+ * Copyright (c) 2016-2018 Linaro Limited
+ * Copyright (c) 2019-2026 Nokia
+ */
+
+#include <odp/autoheader_internal.h>
+
+#ifdef _ODP_PKTIO_DPDK
```
JannePeltonen, line 8: I suppose this is ok, but I guess another alternative would be to include this file in compilation conditionally in Makefile.am. But I do not really have an opinion on this now.

platform/linux-generic/include/odp_dpdk_common.h
```diff
@@ -0,0 +1,182 @@
...
+	if (ipv4)
+		return rte_ipv4_phdr_cksum(l3_hdr, ol_flags);
+	else /*ipv6*/
+		return rte_ipv6_phdr_cksum(l3_hdr, ol_flags);
+}
+
+#define OL_TX_CHKSUM_PKT(_cfg, _capa, _proto, _ovr_set, _ovr) \
+	((_capa) && (_proto) && ((_ovr_set) ? (_ovr) : (_cfg)))
+
+static inline void _odp_dpdk_pkt_set_ol_tx(odp_pktout_config_opt_t *pktout_cfg,
```
JannePeltonen, line 109: I wonder if these functions would better be placed in the C file as non-inline functions. But it seems there is no much/any harm having them as inline functions here, so either way is ok for me.

platform/linux-generic/pktio/dpdk.c
```diff
@@ -981,8 +870,8 @@ static inline int mbuf_to_pkt_zero(pktio_entry_t *pktio_entry,
 
 		packet_set_ts(pkt_hdr, ts);
 
-		ret = _odp_dpdk_packet_parse_common(pkt_hdr, data, pkt_len, pkt_len,
-						    mbuf, layer, pktin_cfg);
+		ret = _odp_dpdk_packet_parse_common(pkt_hdr, data, pkt_len, pkt_len, mbuf, layer,
```
JannePeltonen, line 873: This appears to be an unnecessary change. Just pointing it out if it was unintentional, but ok to leave it as it is too.

JannePeltonen (commented):
Looks good, but a couple of very minor comments to act on or to ignore.

platform/linux-generic/pktio/dpdk_common.c
MatiasElo: This code was copied from ODP-DPDK and the vast majority of PMD drivers seem to support `rte_eth_dev_flow_ctrl_get()`. This would be safer in a way that we make less assumptions, but I have no strong opinions.

platform/linux-generic/pktio/dpdk_common.c
MatiasElo: I went with the existing pattern. I can't recall why this was done this way in the first place, but perhaps it could be cleaned in a separate PR.

platform/linux-generic/include/odp_dpdk_common.h
MatiasElo: Indeed they could. However, I'd prefer testing this separately and keep changes, which could affect performance, in minimum.

platform/linux-generic/pktio/dpdk.c
MatiasElo: True, reverted these so it's easier to tell the relevant code changes.

JannePeltonen (approved):

PR #2339, opened 2026-04-17
[PATCH v4] example: classifier: add queue and cos configurability

TuomasTaipale:
Add new `-q`, `--cos_queue_param` option for configuring the CoS/queue that is defined with the policy destination queue parameter (`-p <dst queue>`).

For a queue, priority, synchronization and a single packet aggregator with timeout and vector size can be configured. For the related CoS, aggregator enqueue profile and optional parameter for the custom profile can be configured.

These now enable the testing of event aggregation in the context of classification with varying CoS aggregator enqueuing profiles.

v3:
- Rebased
- Matias' comments
- Added reviewed-by tag

v4:
- Checkpath complain fixed

### DIFF
--- a/example/classifier/odp_classifier.c
+++ b/example/classifier/odp_classifier.c
@@ -12,6 +12,10 @@
  * @cond _ODP_HIDE_FROM_DOXYGEN_
  */
 
+#ifndef _GNU_SOURCE
+#define _GNU_SOURCE
+#endif
+
 #include <stdlib.h>
 #include <string.h>
 #include <getopt.h>
@@ -87,7 +91,6 @@ typedef struct {
 	char value[DISPLAY_STRING_LEN];	/**< Display string for value */
 	char mask[DISPLAY_STRING_LEN];	/**< Display string for mask */
 	int has_src_cos;
-
 } global_statistics;
 
 typedef struct {
@@ -96,12 +99,23 @@ typedef struct {
 } ci_pass_counters;
 
 typedef struct {
+	char cos_name[ODP_COS_NAME_LEN];
+	odp_aggr_enq_profile_t cos_enq_prof;
+	odp_event_aggr_config_t q_aggr;
+	odp_schedule_prio_t q_prio;
+	odp_schedule_sync_t q_sync;
+	uint32_t num_aggr;
+} cos_q_param_t;
+
+typedef struct {
 	odp_pktout_queue_t pktout[MAX_WORKERS];
 	int num_pktout;
 	global_statistics stats[MAX_PMR_COUNT];
 	ci_pass_counters ci_pass_rules[MAX_PMR_COUNT];
+	cos_q_param_t cos_q_param[MAX_PMR_COUNT];
 	int policy_count;	/**< global policy count */
 	int num_ci_pass_rules;	/**< ci pass count */
+	int num_cos_q_param;
 	int appl_mode;		/**< application mode */
 	odp_atomic_u64_t total_packets;	/**< total received packets */
 	unsigned int cpu_count; /**< Number of CPUs to use */
@@ -116,6 +130,7 @@ typedef struct {
 	int cos_pools;
 	int pool_size;
 	int burst_size;
+	uint32_t vector_size;
 } appl_args_t;
 
 enum packet_mode {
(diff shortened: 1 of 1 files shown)


example/classifier/odp_classifier.c
```diff
@@ -1339,7 +1579,26 @@ static void usage(void)
 		"                           this is not defined.\n"
 		"    <dst queue>            Name of the destination queue (CoS).\n"
 		"\n"
-		"  -c, --count <num>        CPU count, 0=all available, default=1\n"
+		"  -q, --cos_queue_param <name>:<queue prio>:<queue sync>:<cos enqueue profile type>:<optional custom value>:<optional tmo ns>:<optional size>\n"
```
JannePeltonen, line 1625: I think the pre-existing implementation has some design issues that would be good to fix or improve before building more functionality on top.

The usage text and the command line parameters seem to conflate CoS objects and queues. They refer to source queues, although there are no such things, but source CoS:es. Then the "destination queue" term is used to refer both to the actual destination queue and the destination CoS. The implementation creates one CoS for each 'queue' and does not handle cleanly multiple policy rules with the same destination CoS/queue. The usage text does not really explain how the config model of this example app differs from the config model in ODP API, which is not helpful.

I think it would be good to improve the existing design before the new queue config things get added.

I suppose policy (or PMR rule) config could be separate from CoS config. CoS objects could be explicitly created (or on demand, as needed by PMR rules, if not existing already, i.e. not one-to-one with the policy rules. Similarly, queues could perhaps be separately created. Or if the queues would be implicitly created with 1:1 correspondence with CoS objects, that could be mentioned in the help text (and fixing the CoS / queue terminology would make things more clear too).

Then the queue config and CoS config for queue enqueuing that gets added in this commit could build on top of a cleaner base.

example/classifier/odp_classifier.c
```diff
@@ -1339,7 +1579,26 @@ static void usage(void)
...
+		"                           names.\n"
+		"    <optional custom value>\n"
+		"                           If ODP_AEP_TYPE_CUSTOM aggregator enqueue profile was used,\n"
+		"                           pass a hex value (without '0x') for odp_aggr_enq_profile_t::param.\n"
+		"    <optional tmo ns>\n"
+		"                           Aggregator configuration parameter timeout in nanoseconds. Optional,\n"
+		"                           if not passed, aggregation for queue disabled.\n"
+		"    <optional size>\n"
+		"                           Aggregator configuration parameter vector size. Mandatory if 'optional tmo ns'\n"
+		"                           passed, optional otherwise.\n"
```
MatiasElo, line 1642: Linux-generic doesn't currently support aggregation timeouts, so it would be nice to support aggregation with only size.

example/classifier/odp_classifier.c
```diff
@@ -379,70 +454,71 @@ static int pktio_receive_thread(void *arg)
...
-				if (len > 96)
-					len = 96;
+		for (int i = 0; i < num_recv; i++) {
+			ev = evs[i];
 
-				odp_packet_print_data(pkt[j], 0, len);
+			if (odp_event_type(ev) == ODP_EVENT_VECTOR) {
+				evv = odp_event_vector_from_event(ev);
+				odp_event_vector_tbl(evv, &ev_tbl);
+				vector_size = odp_event_vector_size(evv);
```
MatiasElo: It would be good to check that max vector size (`odp_pool_param_t::event_vector.max_size`) doesn't exceed `MAX_PKT_BURST`.

example/classifier/odp_classifier.c
```diff
@@ -1111,6 +1230,118 @@ static int parse_pmr_policy(appl_args_t *appl_args, char *optarg)
...
+	cos_q_param = &appl_args->cos_q_param[appl_args->num_cos_q_param];
+	tmp = strtok(tmp_str, ":");
+
+	if (tmp == NULL) {
+		ODPH_ERR("CoS queue name missing.\n");
+		ret = -1;
+		goto out;
+	}
+
+	odph_strcpy(cos_q_param->cos_name, tmp, ODP_COS_NAME_LEN);
```
MatiasElo, line 1276: Could check that there are no duplicate names.

Also, after parameter parsing it could be verified that there are no CoS queue parameters that didn't match any policy destination queue. Just to make detecting invalid user parameters easier.

example/classifier/odp_classifier.c
```diff
@@ -1339,7 +1579,26 @@ static void usage(void)
...
+		"    <name>                 Configuration for a queue (CoS) defined with '--policy' 'dst queue'.\n"
+		"    <queue prio>           CoS queue priority. Use priority value or \"default\" for default priority.\n"
+		"    <queue sync>           CoS queue synchronization. Use ODP_SCHED_SYNC_PARALLEL,\n"
+		"                           ODP_SCHED_SYNC_ATOMIC or ODP_SCHED_SYNC_ORDERED.\n"
+		"    <cos enqueue profile type>\n"
+		"                           Aggregator enqueue profile type for CoS. Use odp_aggr_enq_profile_t::type\n"
+		"                           names.\n"
+		"    <optional custom value>\n"
+		"                           If ODP_AEP_TYPE_CUSTOM aggregator enqueue profile was used,\n"
+		"                           pass a hex value (without '0x') for odp_aggr_enq_profile_t::param.\n"
```
MatiasElo, line 1636: I'd perhaps make these the last arguments (and optional if possible), as they are likely the least commonly used ones.

example/classifier/odp_classifier.c
```diff
@@ -706,6 +813,9 @@ int main(int argc, char *argv[])
 	/* Create packet pool */
 	pool = pool_create("packet_pool");
 
+	for (i = 0; i < args->num_cos_q_param; i++)
+		args->cos_q_param[i].q_aggr.pool = create_vector_pool(args);
```
MatiasElo: Based on the helper text I got the understanding that `-q` option could be also used to configure queue sync type and priority without enabling aggregation. However, this looks like vector pool is always created.

example/classifier/odp_classifier.c
```diff
@@ -379,70 +454,71 @@ static int pktio_receive_thread(void *arg)
...
-		}
+	return 0;
+}
 
-		/* Swap Eth MACs and possibly IP-addrs before sending back */
-		swap_pkt_addrs(pkt, num);
+static odp_pool_t create_vector_pool(appl_args_t *args)
+{
+	/* For now, singleton pool */
+	static odp_pool_t pool = ODP_POOL_INVALID;
```
MatiasElo, line 499: This can be a surprise to a user. Perhaps print something to log if implementing support for multiple pools is difficult.

example/classifier/odp_classifier.c
MatiasElo, line 1: Copyright can be updated.

example/classifier/odp_classifier.c
MatiasElo: Could use `odp_event_vector_tbl()` return value directly and skip calling `odp_event_vector_size()`.

TuomasTaipale: Added a VLA based on `vector_size`.

example/classifier/odp_classifier.c
TuomasTaipale: One just uses 0 for linux-gen, didn't change anything.

example/classifier/odp_classifier.c
TuomasTaipale: The config string is pretty simple, shouldn't be a hassle to use `_NONE` enum if that's not needed.

example/classifier/odp_classifier.c
TuomasTaipale: Not difficult, but seems wasteful of pools in case of vector pools. Added a print.

example/classifier/odp_classifier.c
```diff
@@ -1111,6 +1242,146 @@ static int parse_pmr_policy(appl_args_t *appl_args, char *optarg)
...
+	tmp = strtok(tmp_str, ":");
+
+	if (tmp == NULL) {
+		ODPH_ERR("CoS queue name missing.\n");
+		ret = -1;
+		goto out;
+	}
+
+	if (find_cos_q_param(appl_args, tmp) != NULL) {
+		ODPH_ERR("Warning: duplicate CoS queue parameter entry for \"%s\", skipping.\n",
```
MatiasElo: `printf()` could be clearer since this is not handled as failure.

example/classifier/odp_classifier.c
```diff
@@ -354,23 +369,88 @@ static odp_pktio_t create_pktio(const char *dev, odp_pool_t pool)
...
+	if (odp_unlikely(dropped)) {
+		num -= dropped;
+		err_cnt += dropped;
+		ODPH_ERR("Drop frame - err_cnt:%lu\n", err_cnt);
+	}
+
+	for (j = 0; j < num; j++) {
+		pool = odp_packet_pool(pkts[j]);
+
+		for (i = 0; i <  MAX_PMR_COUNT; i++) {
```
MatiasElo: Double whitespace before `MAX_PMR_COUNT`.

example/classifier/odp_classifier.c
```diff
@@ -1316,7 +1597,9 @@ static void usage(void)
 		"ODP Classifier example.\n"
 		"Usage: odp_classifier OPTIONS\n"
 		"  E.g. odp_classifier -i eth1 -m 0 -p \"ODP_PMR_SIP_ADDR:10.10.10.0:0xFFFFFF00:queue1\" \\\n"
-		"                                   -p \"ODP_PMR_SIP_ADDR:10.10.10.10:0xFFFFFFFF:queue1:queue2\"\n"
+		"                                   -p \"ODP_PMR_SIP_ADDR:10.10.10.10:0xFFFFFFFF:queue1:queue2\" \\\n"
+		"                                   -q \"queue1:1:ODP_SCHED_SYNC_PARALLEL:ODP_AEP_TYPE_NONE\" \\\n"
+		"                                   -q \"queue2:3:ODP_SCHED_SYNC_ATOMIC:ODP_AEP_TYPE_CUSTOM:1234:0:4\" \\\n"
```
MatiasElo: `\` not necessary at the end of the last line.

example/classifier/odp_classifier.c
MatiasElo: Okay, I interpreted 0 as not passed, but that was not the case.

MatiasElo (approved):
Couple optional nits.

PR #2343, opened 2026-04-27
[PATCH v1] abi: time: change odp_time_t abi type to uint64_t

MatiasElo:
Change odp_time_t ABI type from struct to uint64_t, which makes the time API implementation simpler. The change triggered a couple false positive 'maybe-uninitialized' warnings.

### DIFF
--- a/include/odp/api/abi-default/time_types.h
+++ b/include/odp/api/abi-default/time_types.h
@@ -11,32 +11,13 @@
 extern "C" {
 #endif
 
-/**
- * @internal Time structure used for both POSIX timespec and HW counter
- * implementations.
- */
-struct odp_time_s {
-	/** @internal Variant mappings for time type */
-	union {
-		/** @internal Used with generic 64 bit operations */
-		uint64_t u64;
-
-		/** @internal Nanoseconds */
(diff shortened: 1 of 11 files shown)


include/odp/api/abi-default/time_types.h
```diff
@@ -11,32 +11,13 @@
...
-
-	};
-};
-
 /** @addtogroup odp_time
  *  @{
  **/
 
-typedef struct odp_time_s odp_time_t;
+typedef uint64_t odp_time_t;
```
JannePeltonen, line 18: The struct wrapped uint64_t has the benefit that it better enforces proper API usage as variables of odp_time_t cannot be used directly in time calculation and comparisons. It also makes odp_time_t type unique. I wonder if the implementation simplifications are big enough to outweigh the reduction of API safety.

MatiasElo: Perhaps the biggest benefit of changing this would be that it enables applications to compare `odp_time_t` values directly to `ODP_TIME_NULL`. Currently, `odp_time_t` is a bit of an exception compared to other ODP handles.

PR #2344, opened 2026-04-28
[PATCH v3] linux-gen: time: optimize time conversions with multiply and shift

MatiasElo:


platform/linux-generic/arch/common/odp/api/abi/time_cpu_inlines.h
```diff
@@ -22,6 +22,10 @@ typedef struct _odp_time_global_t {
 	uint64_t freq_hz;
 	uint64_t start_time;
 	uint64_t start_time_ns;
+	uint64_t mult_to_ns;
+	uint64_t mult_from_ns;
+	uint32_t shift_to_ns;
+	uint32_t shift_from_ns;
```
TuomasTaipale: I guess these could be just `uint8_t`.

platform/linux-generic/arch/common/odp_time_cpu.c
```diff
@@ -40,6 +40,25 @@ int _odp_time_init_global(void)
...
+	global->mult_to_ns = (uint64_t)(((__uint128_t)_ODP_TIME_GIGA_HZ << global->shift_to_ns) /
+					global->freq_hz);
+
+	for (global->shift_from_ns = 63; global->shift_from_ns > 0; global->shift_from_ns--) {
+		if ((((__uint128_t)global->freq_hz << global->shift_from_ns) /
+		     ODP_TIME_SEC_IN_NS) <= UINT64_MAX)
+			break;
+	}
+	global->mult_from_ns = (uint64_t)(((__uint128_t)global->freq_hz << global->shift_from_ns) /
+					  ODP_TIME_SEC_IN_NS);
```
TuomasTaipale: Assignment(s) could be done inside the `if`s before `break`, so you do the calculations only once?

platform/linux-generic/arch/common/odp/api/abi/time_cpu_inlines.h
MatiasElo: Changed on v2.

platform/linux-generic/arch/common/odp_time_cpu.c
MatiasElo: Changed on v2.

TuomasTaipale (approved):

PR #2348, opened 2026-05-18
[PATCH v2] linux-gen: pool: minor optimizations to event alloc/free

MatiasElo:


platform/linux-generic/odp_pool.c
```diff
@@ -1541,9 +1532,13 @@ void _odp_event_free(odp_event_t event)
 		ring_mpmc_rst_ptr_enq_multi(ring, mask, (void **)ev_hdr, burst);
 		if (CONFIG_POOL_STATISTICS && pool->params.stats.bit.free_ops)
 			odp_atomic_inc_u64(&pool->stats.free_ops);
+
+		cache_num = odp_atomic_load_u32(&cache->cache_num);
 	}
 
-	cache_push_single(cache, event_hdr);
+	cache->event_hdr[cache_num] = event_hdr;
```
JannePeltonen: It bothers me a little bit that the organization of the cache (e.g. cache->event_hdr[] and how it is used for caching) is no longer fully contained in the cache_push/pop functions. cache->cache_num is not contained in there even know, but I think this makes the separation slightly worse. Maybe you could retain cache_push_single() but make it take cache_num as a parameter. I think that would be slightly cleaner.

JannePeltonen (commented):
Looks good, apart from one minor quibble.

platform/linux-generic/odp_pool.c
MatiasElo: Retained `cache_push_single()` in v2.

JannePeltonen (approved):

PR #2349, opened 2026-05-18
[PATCH v4] api: pktio: clarify odp_pktio_default/error_cos_set() specifications

MatiasElo:
V3:
- Added implementation and validation tests

### DIFF
--- a/include/odp/api/spec/packet_io.h
+++ b/include/odp/api/spec/packet_io.h
@@ -695,23 +695,28 @@ int odp_pktio_mac_addr_set(odp_pktio_t pktio, const void *mac_addr,
 int odp_pktio_default_cos_set(odp_pktio_t pktio, odp_cos_t default_cos);
 
 /**
- * Setup per-port error class-of-service
+ * Set interface error class-of-service
  *
- * @param pktio      Ingress port pktio handle.
- * @param error_cos  class-of-service set to all packets arriving at this
- *                   ingress port that contain an error.
+ * Set CoS for packets received on this packet IO interface for which parsing
+ * detected an error, or remove the current error CoS by passing ODP_COS_INVALID
+ * as 'error_cos'. This function may be called multiple times on the same
+ * interface, and each successful call replaces the previous error CoS binding.
+ *
+ * When no error CoS is configured, it's implementation defined how packets with
+ * parsing errors are handled and the packets may be discarded.
+ *
+ * @param pktio      Packet IO handle
+ * @param error_cos  CoS handle or ODP_COS_INVALID
  *
  * @retval  0 on success
  * @retval <0 on failure
- *
- * @note Optional.
  */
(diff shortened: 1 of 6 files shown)


include/odp/api/spec/packet_io.h
```diff
@@ -678,17 +678,19 @@ int odp_pktio_mac_addr_set(odp_pktio_t pktio, const void *mac_addr,
 			   int size);
 
 /**
- * Setup per-port default class-of-service.
+ * Setup interface default class-of-service
```
TuomasTaipale: Matter of taste but I guess the "Setup" could also be just "Set".

The two commits change the "per-port" to "interface" which leaves the `odp_pktio_skip_set()` and `odp_pktio_headroom_set()` with "per-port", maybe those could be changed to "interface" as well.

include/odp/api/spec/packet_io.h
```diff
@@ -695,16 +695,23 @@ int odp_pktio_mac_addr_set(odp_pktio_t pktio, const void *mac_addr,
...
 
 /**
- * Setup per-port error class-of-service
+ * Setup interface error class-of-service
  *
- * @param pktio      Ingress port pktio handle.
- * @param error_cos  class-of-service set to all packets arriving at this
- *                   ingress port that contain an error.
+ * Set CoS for packets arriving from this packet IO interface that contain
+ * errors detected during packet parsing, or remove the current error CoS by
```
TuomasTaipale: Maybe for the section before comma: "Set CoS for packets received on this packet I/O interface when parsing detects an error."

include/odp/api/spec/packet_io.h
```diff
@@ -695,16 +695,23 @@ int odp_pktio_mac_addr_set(odp_pktio_t pktio, const void *mac_addr,
...
+ * passing ODP_COS_INVALID as 'error_cos'. This function may be called multiple
+ * times on the same interface, and each successful call replaces the previous
+ * error CoS binding.
+ *
+ * When no error CoS is configured, it's implementation defined how packets with
+ * parsing errors are handled (they may be discarded similarly to classification
+ * failures).
```
TuomasTaipale: I guess the current parenthesis section could be just "... errors are handled and may be discarded ...".

include/odp/api/spec/packet_io.h
MatiasElo: Yep, "set" is also more commonly used throughout the API. Both comments handled in v2.

include/odp/api/spec/packet_io.h
MatiasElo: Updated in v2.

include/odp/api/spec/packet_io.h
MatiasElo: Updated in v2.

TuomasTaipale (approved):

psavol (approved):
Reviewed API commits.

kkirankk (approved):

PR #2352, opened 2026-05-25
[PATCH v2] linux-gen: random: detect arm feat_rng availability at runtime

MatiasElo:


JannePeltonen (approved):
Newest GCC minor versions seem to support "grace" cpu target, which does not include the RNG feature. But this runtime checks makes things easier and more robust with CPU flags coming from pkgconfig, so approving this.

PR #2353, opened 2026-05-27
[PATCH v2] test: bench_queue: add tests for odp_schedule_prefetch() function

MatiasElo:
Add test cases for measuring the overhead of odp_schedule_prefetch() API calls.

### DIFF
--- a/test/performance/odp_bench_queue.c
+++ b/test/performance/odp_bench_queue.c
@@ -626,6 +626,29 @@ static int schedule_tm(bench_tm_result_t *res, int repeat_count)
 	return 1;
 }
 
+static int schedule_prefetch_single(void)
+{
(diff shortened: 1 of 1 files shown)


test/performance/odp_bench_queue.c
```diff
@@ -626,6 +626,29 @@ static int schedule_tm(bench_tm_result_t *res, int repeat_count)
 	return 1;
 }
 
+static int schedule_prefetch_1(void)
```
TuomasTaipale: In some universe, somebody might mistake the "1" to be related to versioning so alternative could "_single" (and "_burst).

TuomasTaipale (approved):
One cosmetic comment, otherwise OK.

test/performance/odp_bench_queue.c
MatiasElo: Changed the naming in V2.

PR #2356, opened 2026-06-01
[PATCH v2] api: timer: fix freq_hz array with timer pool info

TuomasTaipale:
Amend `odp_timer_pool_info_t` with new 'freq_hz' and 'num' fields that are used to return the constraining frequencies used to configure the timer pool in case of a `ODP_TIMER_TYPE_PERIODIC_FREQ` pool. The data is constant and should not be modified by the application once populated.

v2:
- Rebased
- Matias' comments

### DIFF
--- a/include/odp/api/spec/timer_types.h
+++ b/include/odp/api/spec/timer_types.h
@@ -794,8 +794,24 @@ typedef struct odp_timer_tick_info_t {
  * ODP timer pool information and configuration
  */
 typedef struct {
-	/** Parameters specified at creation */
+	/** Timer pool parameters
+	 *
+	 *  Parameters specified at creation with the exception of
+	 *  odp_timer_pool_param_t::periodic::freq::freq_hz and
+	 *  odp_timer_pool_param_t::periodic::freq::num which are set to NULL and 0 respectively.
+	 *  The creation-time values for these are provided via 'freq_hz' and 'num_freq_hz' fields
+	 *  in case of ODP_TIMER_TYPE_PERIODIC_FREQ pools.
+	 */
 	odp_timer_pool_param_t param;
+
+	/** Array of constraining periodic timer frequencies for timer pool specified at
+	 *  creation. Must not be modified after population by odp_timer_pool_info(). Accessible
+	 *  until the timer pool is destroyed.
+	 */
+	const odp_fract_u64_t *freq_hz;
+
+	/** Number of items in 'freq_hz' array */
+	uint32_t num_freq_hz;
 
 	/** Number of currently allocated timers */
 	uint32_t cur_timers;
--- a/platform/linux-generic/odp_timer.c
+++ b/platform/linux-generic/odp_timer.c
@@ -1155,7 +1155,7 @@ static odp_timer_pool_t timer_pool_new(const char *name, const odp_timer_pool_pa
 {
 	uint32_t i;
 	int tp_idx;
-	size_t sz0, sz1, sz2;
+	size_t sz0, sz1, sz2, sz3;
 	uint64_t tp_size;
 	uint64_t res_ns, nsec_per_scan;
 	odp_shm_t shm;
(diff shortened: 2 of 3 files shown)


include/odp/api/spec/timer_types.h
```diff
@@ -794,9 +794,20 @@ typedef struct odp_timer_tick_info_t {
  * ODP timer pool information and configuration
  */
 typedef struct {
-	/** Parameters specified at creation */
+	/** Parameters specified at creation with the exception of
+	 *  odp_pool_param_t::periodic::freq::freq_hz and odp_pool_param_t::periodic::freq::num
```
MatiasElo: `odp_pool_param_t` -> `odp_timer_pool_param_t`

include/odp/api/spec/timer_types.h
```diff
@@ -794,9 +794,20 @@ typedef struct odp_timer_tick_info_t {
...
+	 *  which are set to NULL and 0 respectively here and in case of
+	 *  ODP_TIMER_TYPE_PERIODIC_FREQ pools, provided via 'freq_hz' and 'num_freq_hz'
+	 *  fields. */
 	odp_timer_pool_param_t param;
 
+	/** Array of constraining periodic timer frequencies for timer pool specified at
+	 *  creation. Must not be modified after population by odp_timer_pool_info(). */
```
MatiasElo: The `Must not be modified after...` sentence seems a bit unnecessary as `const` is already used.

We could instead say something about the lifetime of the data. E.g. The array is available until the timer pool is destroyed.

include/odp/api/spec/timer_types.h
```diff
@@ -794,9 +794,20 @@ typedef struct odp_timer_tick_info_t {
  * ODP timer pool information and configuration
  */
 typedef struct {
-	/** Parameters specified at creation */
+	/** Parameters specified at creation with the exception of
+	 *  odp_pool_param_t::periodic::freq::freq_hz and odp_pool_param_t::periodic::freq::num
+	 *  which are set to NULL and 0 respectively here and in case of
+	 *  ODP_TIMER_TYPE_PERIODIC_FREQ pools, provided via 'freq_hz' and 'num_freq_hz'
+	 *  fields. */
```
MatiasElo: Splitting the text into two sentences would perhaps make it a bit easier to read. E.g.
```
Timer pool parameters

Contains the parameters specified at pool creation with the exception ... , which are set to
NULL and 0, respectively. For ODP_TIMER_TYPE_PERIODIC_FREQ pools, the actual
values are provided through the 'freq_hz' and 'num_freq_hz' fields.
```

include/odp/api/spec/timer_types.h
TuomasTaipale: Data is const but pointer could be freed so spec explicitly now states that nothing is to be modified.

I'll add some text about lifetime.

include/odp/api/spec/timer_types.h
```diff
@@ -794,9 +794,25 @@ typedef struct odp_timer_tick_info_t {
...
 
+	/** Array of constraining periodic timer frequencies for timer pool specified at
+	 *  creation. Must not be modified after population by odp_timer_pool_info(). Accessible
+	 *  until the timer pool is destroyed.
+	 */
+	const odp_fract_u64_t *freq_hz;
+
+	/** Number of items in 'freq_hz' array */
+	uint32_t num_freq_hz;
+
```
JussiKerttula, line 815: Why these new fields are added since the same information is already behind param?

JannePeltonen: Actually, the same information is not (necessarily) there, as odp_timer_pool_param_t contains only a pointer to the array of frequencies. Returning the original pointer value is not very helpful as the array pointed to by it may have already been freed by the application. And it may not be easy for an application author to recognise that the pointer may no longer be valid.

JussiKerttula: Can't the param.freq_hz pointer be updated to the "stored" freq_hz values instead of returning NULL?

JannePeltonen: I think one reason was that the new pointer in the info struct is a const pointer, but I suppose the pointer in the param struct could (or even should) be a const pointer.

TuomasTaipale:
Alternative version here #2359.

TuomasTaipale:
Going with #2359.

PR #2357, opened 2026-06-01
[PATCH v2] linux-gen: pool: alloc/free operation optimizations

MatiasElo:


platform/linux-generic/odp_pool.c
```diff
@@ -162,6 +162,22 @@ static inline void cache_push_single(pool_cache_t *cache, _odp_event_hdr_t *even
...
+					    uint32_t cache_num, uint32_t num)
+{
+	uint32_t cached;
+
+	_ODP_ASSERT(cache_num + num <= CONFIG_POOL_CACHE_MAX_SIZE);
+
+	cached = ring_mpmc_rst_ptr_deq_multi(ring, mask,
+					     (void **)&cache->event_hdr[cache_num], num);
+
+	odp_atomic_store_u32(&cache->cache_num, cached);
```
JannePeltonen: I seems that the second parameter should be cache_num + cached.

platform/linux-generic/odp_pool.c
```diff
@@ -101,18 +101,17 @@ static inline uint32_t cache_pop(pool_cache_t *cache,
...
 
 	/* Cache does not have enough buffers */
 	if (odp_unlikely(cache_num < (uint32_t)max_num))
 		num_ch = cache_num;
 
 	/* Get buffers from the cache */
 	cache_begin = cache_num - num_ch;
-	for (i = 0; i < num_ch; i++)
-		event_hdr[i] = cache->event_hdr[cache_begin + i];
+	memcpy(event_hdr, &cache->event_hdr[cache_begin],
```
JannePeltonen, line 111: I wonder what the typical burst sizes might be, as with small burst this may slow things down a little. The compiler (in my setup) generates a call to memcpy, which then presumably also has some setup overhead if it actually ends up using wider loads and stores.

platform/linux-generic/odp_pool.c
MatiasElo: Good catch, fixed in V2.

platform/linux-generic/odp_pool.c
MatiasElo: In my testing the break-even point is ~4 buffers (varies between systems).

JannePeltonen (approved):

PR #2358, opened 2026-06-04
[PATCH v3] linux-gen: stash: add optimized implementations per put_mode/get_mode

MatiasElo:
V2:
- Moved all ring headers into a separate directory
- Added "full" MPSC and SPMC ring implementations
- Added support for using SPSC rings


platform/linux-generic/include/odp_ring_mpmc_internal.h
```diff
@@ -137,6 +141,10 @@ static inline int ring_mpmc_cas_u32(odp_atomic_u32_t *atom,
 	#define _RING_MPMC_ENQ_MULTI	ring_mpmc_u32_enq_multi
 	#define _RING_MPMC_DEQ_BATCH	ring_mpmc_u32_deq_batch
 	#define _RING_MPMC_ENQ_BATCH	ring_mpmc_u32_enq_batch
+	#define _RING_MPMC_DEQ_MULTI_SC	ring_mpmc_u32_deq_multi_sc
```
JannePeltonen: I do not think this naming makes sense for single consumer and single producer cases since mpmc stands for 'multi-producer multi-consumer', i.e. mpmc_..._sc is a bit of an oxymoron. I think this should be named ring_mpsc_u32_deq_multi(). Maybe the header file should be renamed too or thinks split to multiple header files, to keep things consistent at file level too.

platform/linux-generic/odp_stash.c
```diff
@@ -447,8 +447,57 @@ static inline int32_t mpmc_ring_u64_len(stash_t *stash)
 	return ring_mpmc_u64_len(&stash->ring_mpmc_u64.hdr);
 }
 
+static inline int32_t mpmc_ring_u32_enq_multi_sp(stash_t *stash, const uint32_t val[], int32_t num)
```
JannePeltonen: The inline specifier is not useful here since this is only called through a function pointer.

platform/linux-generic/odp_stash.c
```diff
@@ -447,6 +459,80 @@ static inline int32_t mpmc_ring_u64_len(stash_t *stash)
...
+{
+	ring_st_u64_init(&stash->ring_st_u64.hdr);
+
+	for (uint32_t i = 0; i < stash->ring_size; i++)
+		stash->ring_st_u64.data[i] = 0;
+}
+
+static inline int32_t st_ring_u32_enq_multi(stash_t *stash, const uint32_t val[], int32_t num)
+{
+	return ring_st_u32_enq_multi(&stash->ring_st_u32.hdr, stash->ring_st_u32.data,
```
JannePeltonen, line 676: I am not sure using a ring to back a single-threaded stash is a good choice. A stack that would first return the item that was pushed last would result in much better temporal locality of memory accesses.

MatiasElo: The current stash use cases are for `ODP_STASH_TYPE_FIFO`, but indeed another implementation could be later added for the default stash type.

JannePeltonen: I missed the FIFO stash type, so the ring implementation is indeed needed and the stack variant would be a separate, future improvement.

platform/linux-generic/odp_stash.c
MatiasElo: Removed in v2.

platform/linux-generic/include/odp_ring_mpmc_internal.h
MatiasElo: I ended up refactoring the ring code quite a bit in v2.

platform/linux-generic/include/ring/odp_ring_mpmc_internal.h
```diff
@@ -261,12 +367,56 @@ static inline uint32_t _RING_MPMC_DEQ_MULTI(_ring_mpmc_gen_t *ring,
...
+static inline uint32_t _RING_MPSC_DEQ_MULTI(_ring_xpxc_gen_t *ring,
+					    _ring_xpxc_data_t *ring_data,
+					    uint32_t ring_mask,
+					    _ring_xpxc_data_t data[],
+					    uint32_t num)
+{
+	uint32_t old_head, new_head, w_tail, num_data, i;
+
+	/* Load acquire of w_tail ensures it is read after r_head, so that
+	 * r_head value is always behind or equal to w_tail value. */
```
JannePeltonen: The first part of comment is bit inaccurate and misleading, so I would remove it and just mention the invariant that does hold.

The load acquire does not ensure that a load that is before it in program order is also observed before the load acuire. But here it does not matter since r_head is only accessed by the single consumer thread. The consumer thread never writes a value larger than (the observed value of) w_tail in r_head and w_tail can only increase which guarantees that r_head is always less than or equal to the w_tail value observed here.

Actually, the comments in the existing MPMC dequeue functions do not seem to be quite accurate either. As far as I can see, C11 memory model does not imply that the load acquires ensure what is being claimed. However, typical compilers for typical targets probably handle the load acquire in such a way that the claimed guarantee usually holds in the resulting machine code. Perhaps that code, or at least the comments, need to be revisited, but that is a separate issue from this PR.

platform/linux-generic/include/ring/odp_ring_mpmc_internal.h
```diff
@@ -399,12 +590,56 @@ static inline uint32_t _RING_MPMC_ENQ_MULTI(_ring_mpmc_gen_t *ring,
...
+					    _ring_xpxc_data_t *ring_data,
+					    uint32_t ring_mask,
+					    const _ring_xpxc_data_t data[],
+					    uint32_t num)
+{
+	uint32_t old_head, new_head, r_tail, num_free, i;
+	uint32_t size = ring_mask + 1;
+
+	/* Load acquire of r_tail ensures it is read after w_head, so that
+	 * w_head - r_tail <= size. */
```
JannePeltonen: As in _RING_MPSC_DEQ_MULTI, this comment is inaccurate and misleading. w_head is only accessed by the producer thread so the loaded value is always the last written value, regardless of how the load gets ordered with respect to the load of r_tail. The value written to w_head never exceeds old_head + num_free, which equals size + r_rtail, so w_head - r_tail cannot exceed size.

platform/linux-generic/include/ring/odp_ring_mpmc_rst_internal.h
```diff
@@ -220,10 +320,50 @@ static inline uint32_t _RING_MPMC_RST_DEQ_MULTI(_ring_mpmc_rst_gen_t *ring, uint
...
+#ifdef _RING_MPSC_RST_DEQ_MULTI
+/* Single-consumer dequeue of multiple data items from the ring head. Num is
+ * smaller than ring size. */
+static inline uint32_t _RING_MPSC_RST_DEQ_MULTI(_ring_xpxc_rst_gen_t *ring, uint32_t mask,
+						_ring_xpxc_rst_data_t data[], uint32_t num)
+{
+	uint32_t head, tail, new_head, i;
+
+	/* Load acquire of w_tail ensures it is read after r_head, so that head
+	 * value is always behind or equal to tail value. */
```
JannePeltonen: Same comment about the load acquire here as for _RING_MPSC_RST_DEQ_MULTI

platform/linux-generic/include/ring/odp_ring_mpmc_rst_internal.h
```diff
@@ -318,8 +494,40 @@ static inline void _RING_MPMC_RST_ENQ_MULTI(_ring_mpmc_rst_gen_t *ring, uint32_t
...
+	 * there are no other writers. */
+	old_head = odp_atomic_load_u32(&ring->r.w_head);
+	odp_atomic_store_u32(&ring->r.w_head, old_head + num);
+	new_head = old_head + 1;
+
+	/* Wait for the last reader to finish. This prevents overwrite when
+	 * a reader has been left behind (e.g. due to an interrupt) and is
+	 * still reading these slots. */
+	while (odp_unlikely(new_head - odp_atomic_load_acq_u32(&ring->r.r_tail)
+			    >= size))
```
JannePeltonen, line 516: This wait was there already there in _RING_MPMC_RST_ENQ_MULTI, but it seems to me that this unnecessarily requires that the ring becomes empty before further enqueues, even though it seems that it would be sufficient to wait until there is room for the number of items being enqueued.

platform/linux-generic/odp_stash.c
```diff
@@ -369,23 +369,23 @@ static void free_index(int i)
 	odp_ticketlock_unlock(&stash_global->lock);
 }
 
-static inline void mpmc_rst_ring_u32_init(stash_t *stash)
+static void mpmc_rst_ring_u32_init(stash_t *stash)
```
JannePeltonen, line 384: Small comment to the commit message: 'inline' is actually specifier, not qualifier.

platform/linux-generic/include/ring/odp_ring_mpmc_rst_internal.h
JannePeltonen: Actually I misread the code. The wait does not wait that the ring is empty, but there appears to be a problem that the wait does not take 'num' into account. OTOH, since this is a RST ring, if the application guarantees that the total number of started enqueues minus the number of completed dequeues is less than the size of the ring, then the whole while loop appears unecessary.

platform/linux-generic/include/ring/odp_ring_mpmc_internal.h
MatiasElo: Comment fixed in V3.

platform/linux-generic/include/ring/odp_ring_mpmc_internal.h
MatiasElo: Comment fixed in V3.

platform/linux-generic/include/ring/odp_ring_mpmc_rst_internal.h
MatiasElo: Comment fixed in V3.

platform/linux-generic/include/ring/odp_ring_mpmc_rst_internal.h
MatiasElo: This will be investigated a separately.

platform/linux-generic/odp_stash.c
MatiasElo: Fixed in V3.

JannePeltonen (approved):

PR #2360, opened 2026-06-05
[PATCH v2] linux-gen: packet: fix and optimise packet metadata copying

JannePeltonen:
Fix packet metadata copying by adding copying of missing fields and optimise the copying code a little bit.

### DIFF
--- a/test/common/packet_common.c
+++ b/test/common/packet_common.c
@@ -1,19 +1,63 @@
 /* SPDX-License-Identifier: BSD-3-Clause
- * Copyright (c) 2023 Nokia
+ * Copyright (c) 2023-2026 Nokia
  */
 
 #include <packet_common.h>
 #include <string.h>
 
+static uint32_t minstd_rand(void)
+{
+	static __thread uint64_t s = 1;
+	uint64_t prime = 0x7fffffff;
+
+	s = (48271 * s) % prime;
+	return s;
+}
+
+static uint32_t random_u32(void)
+{
+	return minstd_rand();
+}
+
+static uint8_t random_u8(void)
+{
+	return minstd_rand();
+}
+
+int test_packet_write_data(odp_packet_t pkt, uint32_t offset, uint32_t len)
+{
+	uint8_t w_data[len + 1]; /* +1 to avoid zero length array */
+	uint8_t r_data[len + 1];
+
+	if (len == 0)
+		return 0;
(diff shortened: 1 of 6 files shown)


test/common/packet_common.c
```diff
@@ -23,35 +67,37 @@ void test_packet_set_md(odp_packet_t pkt)
...
 	odp_packet_ts_set(pkt, odp_time_local_from_ns(ODP_TIME_SEC_IN_NS));
-	odp_packet_color_set(pkt, ODP_PACKET_YELLOW);
+	odp_packet_color_set(pkt, varying_flag ? ODP_PACKET_YELLOW : ODP_PACKET_RED);
 	odp_packet_drop_eligible_set(pkt, binary_flag);
-	odp_packet_shaper_len_adjust_set(pkt, -42);
-	(void)odp_packet_payload_offset_set(pkt, offset);
+	odp_packet_shaper_len_adjust_set(pkt, random_u8() - 128);
+	(void)odp_packet_payload_offset_set(pkt, random_u32() % len);
+	odp_packet_aging_tmo_set(pkt, random_u32());
+	(void)odp_packet_tx_compl_request(pkt, &tx_compl_opt);
```
MatiasElo, line 100: Strictly speaking these can be against the spec capabilites.

MatiasElo (approved):
One optional comment. The performance test results look good.

test/common/packet_common.c
JannePeltonen: Maybe, or maybe not. The API is not clear on how/if pktio aging and TX completion capabilities affect packet metadata setting. As long as calling the function does not crash or cause undefined behaviour (in C sense), which I think it should not, things should be fine.

odp_pktio_capability_t::max_tx_aging_tmo_ns tells if a particular pktio supports TX aging and with what max timeout. But since packets are not tied to pktios, the statement "Maximum tmo value is defined by max_tx_aging_tmo_ns capa" in the API is not very clear. I think it means that the aging timeout does not work properly if you send a packet to a pktio with a timeout that is too large for the pktio. Or that you should not send such packets. But what if you do not send the packet? Perhaps the maximum that you can use in this function is the maximum of all pktios (even the ones that you have not opened yet). If larger values get clamped or produce an error then this code works fine.

There is a similar argument for the TX completion request and the compl_id there versus the pktio specific max_compl_id capability.

I did not change anything in v2.

PR #2361, opened 2026-06-08
[PATCH v4] api: ipsec: add capabilities for outbound options

JannePeltonen:
Make per-packet outbound options optional by introducing capability bits that show which options are supported. Currently not all ODP implementations actually implement the options. The new capability bits let an implementation advertise which options it supports, allowing it to stay API-compliant.

### DIFF
--- a/include/odp/api/spec/ipsec_types.h
+++ b/include/odp/api/spec/ipsec_types.h
@@ -1,6 +1,6 @@
 /* SPDX-License-Identifier: BSD-3-Clause
  * Copyright (c) 2016-2018 Linaro Limited
- * Copyright (c) 2022-2025 Nokia
+ * Copyright (c) 2022-2026 Nokia
  */
 
 /**
@@ -33,7 +33,7 @@ extern "C" {
  * IPSEC Security Association (SA)
  */
 
- /**
+/**
  * @def ODP_IPSEC_SA_INVALID
  * Invalid IPSEC SA
  */
@@ -255,6 +255,34 @@ typedef struct odp_ipsec_outbound_config_t {
 } odp_ipsec_outbound_config_t;
 
 /**
+ * Capabilities regarding IPsec outbound operation parameters
+ */
+typedef struct odp_ipsec_out_op_capability_t {
+	/** Supported outbound operation option flags.
+	 *
+	 *  If a flag is set, the corresponding option in odp_ipsec_out_opt_t
+	 *  is supported in outbound IPsec processing. If a flag is clear,
+	 *  the corresponding flag and parameter fields of odp_ipsec_out_opt_t
+	 *  are ignored by the IPsec outbound processing functions.
+	 */
+	struct {
+		/** fragmentation mode option supported */
+		uint8_t frag_mode :1;
+
+		/** TFC padding length option supported */
+		uint8_t tfc_pad   :1;
+
+		/** TFC dummy packet option supported */
+		uint8_t tfc_dummy :1;
+
+		/** IP parameters option supported */
+		uint8_t ip_param  :1;
+
+	} opt;
+
+} odp_ipsec_out_op_capability_t;
+
+/**
  * IPSEC TEST capability
  */
 typedef struct odp_ipsec_test_capability_t {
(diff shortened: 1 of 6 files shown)


include/odp/api/spec/ipsec_types.h
```diff
@@ -254,6 +254,33 @@ typedef struct odp_ipsec_outbound_config_t {
...
+ * Capabilities regarding IPsec outbound operation parameters
+ */
+typedef struct odp_ipsec_out_op_capability_t {
+	/** Supported outbound operation option flags.
+	 *
+	 *  If a flag is set, the corresponding option in odp_ipsec_out_opt_t
+	 *  is supported in outbound IPsec processing. If a flag is clear,
+	 *  the corresponding flag and parameter fields of odp_ipsec_out_opt_t
+	 *  are ignored by the IPsec outbound processing functions.
+	 */
```
TuomasTaipale, line 267: Matter of taste but this could be just:
```c
/** Supported outbound operation option flags */
```

include/odp/api/spec/ipsec_types.h
```diff
@@ -254,6 +254,33 @@ typedef struct odp_ipsec_outbound_config_t {
...
+		uint8_t frag_mode :1;
+
+		/** TFC padding length option supported */
+		uint8_t tfc_pad   :1;
+
+		/** TFC dummy packet option supported */
+		uint8_t tfc_dummy :1;
+
+		/** IP parameters option supported */
+		uint8_t ip_param  :1;
```
TuomasTaipale, line 279: Newline after to match existing style.

include/odp/api/spec/ipsec_types.h
```diff
@@ -254,6 +254,33 @@ typedef struct odp_ipsec_outbound_config_t {
...
+	} opt;
```
TuomasTaipale, line 281: The config counterpart has union for all fields, this could as well to enable bitwise checking capa vs. config which is sometimes useful.

include/odp/api/spec/ipsec_types.h
```diff
@@ -1213,7 +1243,12 @@ typedef struct odp_ipsec_op_flag_t {
 /**
  * IPSEC outbound operation options
  *
- * These may be used to override some SA level options
+ * These may be used to override some SA level options. Supported options are
+ * listed in odp_ipsec_capability_t::out_op returned by odp_ipsec_capability().
+ * Options that are not supported are ignored.
+ *
+ * @see odp_ipsec_capability()
+ * @see odp_ipsec_out_op_capability_t
```
TuomasTaipale: Text already points to these directions and doxygen should generate links so these are probably unnecessary.

include/odp/api/spec/ipsec_types.h
JannePeltonen: I wanted to have explicit reference to odp_ipsec_out_opt_t to avoid confusion with similar options at SA level. So I left this unchanged in v2.

include/odp/api/spec/ipsec_types.h
JannePeltonen: added in v2

include/odp/api/spec/ipsec_types.h
JannePeltonen: I did not add it in v2 since I did not see much usefulness in it. In the config struct the union can help in clearing all flags with more concise code, but in the capa struct there is no need to write anything. And since bitfield ordering is implementation specific, bitwise checking using the union can be dangerous.

include/odp/api/spec/ipsec_types.h
JannePeltonen: In v2 I removed the first one, but left the second one there for people reading the header directly.

JannePeltonen:
v2:
- cosmetic tweaks in the API commit based on review commits
- added commits for implementation, testing and sysinfo

include/odp/api/spec/ipsec_types.h
TuomasTaipale: `linux-gen` uses it in packet I/O and pool code to check capa vs. config, so some usefulness if one wants to easily check if something unsupported was asked. Obviously you would have to extract the struct to a standalone element and then use it in both capa and config. But maybe it's not useful here then, resolving.

TuomasTaipale (approved):

psavol (approved):
Reviewed the API commit

ktejasree123:
Acked-by: Tejasree Kondoj ktejasree@marvell.com

nithind1988:
Acked-by: Nithin Dabilpuram <ndabilpuram@marvell.com>

PR #2362, opened 2026-06-12
[PATCH v2] linux-gen: small rst ring and stash improvements

JannePeltonen:
Avoid using flexible array members in structs in a way that is not allowed by the C standard and get rid of the pragmas that have been used for suppressing related warnings.

- Make rst rings similar to other ring flavors by removing the data array from the rst ring header structs and by adding a ring data pointer parameter to the rst ring functions. Update all rst ring users accordingly.
- Move ring data array to the end of stash_t

Also remove unneeded ring init function pointer from stash_t and change checks for invalid stash handles to assertions in fast path stash functions.


MatiasElo (approved):
Looks good 👍

PR #2363, opened 2026-06-15
[PATCH v3] api: packet: remove len restriction from odp_packet_{push,pull}_{head,tail}

JannePeltonen:
- Remove value range from the len parameter descriptions as it is intended that an application may pass any len value and that too large len values cause the function to fail. 

- Add a paragraph in the API principles section in the API README explaining that invalid handles must not  be passed to any ODP API function unless explicitly allowed.

### DIFF
--- a/include/README
+++ b/include/README
@@ -25,6 +25,11 @@ type odp_packet_t, and packet-related APIs take arguments of this type. What an
 odp_packet_t actually is, is not part of the ODP API specification and
 applications cannot make assumptions about the underlying type.
 
+Most ODP handle types have a special 'invalid' value (e.g. ODP_PACKET_INVALID
+for packet handles) that is typically used for indicating failures of various
+API functions that return handles. Such an invalid handle must not be passed
+to any ODP API function unless explicitly allowed in the API text.
+
 Application gains ODP handle ownership when it receives it from an ODP API call.
 For example, application can receive an odp_event_t handle from `odp_schedule()`
 call. The ownership ends when the handle is passed back to ODP, for example with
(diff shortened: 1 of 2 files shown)


include/odp/api/spec/packet.h
MatiasElo: The pseudo-code could be wrapped with Doxygen tags. E.g.
```
 * @code{.unparsed}
 * odp_packet_xxx:
 * seg_len  += len
 * len      += len
 * headroom -= len
 * data     -= len
 * @endcode
```
Currently, all the pseudo-code is on a single line in Doxygen output.

MatiasElo (approved):
Looks good. One optional additional clean-up, which could be done on the same go.

include/README
```diff
@@ -25,6 +25,11 @@ type odp_packet_t, and packet-related APIs take arguments of this type. What an
 odp_packet_t actually is, is not part of the ODP API specification and
 applications cannot make assumptions about the underlying type.
 
+Most ODP handle types have a special 'invalid' value (e.g. ODP_PACKET_INVALID
```
TuomasTaipale, line 28: Extra whitespace in commit message "... must not  be ...".

This README mostly wraps code in ``, so ODP_PACKET_INVALID could be also -> `ODP_PACKET_INVALID`.

include/odp/api/spec/packet.h
JannePeltonen: added a commit that does that in v2

include/README
JannePeltonen: Removed the extra whitespace in v2, but left ODP_PACKET_INVALID as it is.

psavol (approved):

ashwinyes:

PR #2364, opened 2026-06-15
[PATCH v2] test: packet_gen: fix parsing format

TuomasTaipale:
Currently, when scanning input for `ODP_LSO_WRITE_BITS` bits and masks, `SCNu32` is used, which doesn't do automatic base detection. Change to use `SCNi32` which does.

v2:
- Added reviewed-by tag


JannePeltonen (approved):
Maybe the "(in hex)" remark in usage could be removed (unless non-hex values are not really  meant to be used)?

PR #2365, opened 2026-06-15
[PATCH v4] example: classifier: add handling for unexpected events

TuomasTaipale:
Add handling for unexpected events. The tester only ever setups event reception for packets or event vectors (for packets) so reception of unexpected events should be rare.

Additionally, track packets handled within a burst with a sequence number logged as part of event origin dump (verbose option enabled).

v2:
- Rebased
- Janne's comments

v3:
- Added reviewed-by tag

### DIFF
--- a/example/classifier/odp_classifier.c
+++ b/example/classifier/odp_classifier.c
@@ -131,6 +131,7 @@ typedef struct {
 	int pool_size;
 	int burst_size;
 	uint32_t vector_size;
+	odp_spinlock_t verbose_lock;	/**< Serializes verbose packet dumps */
 } appl_args_t;
 
 enum packet_mode {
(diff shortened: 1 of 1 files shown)


example/classifier/odp_classifier.c
```diff
@@ -377,6 +377,8 @@ static void process_packets(odp_packet_t pkts[], int num, odp_queue_t in_q,
 	int i, j, dropped, sent;
 	global_statistics *stats;
 	unsigned long err_cnt = 0;
+	/* Identify which packets were processed in this particular batch */
+	static uint64_t seq;
```
JannePeltonen: I do not understand what this seq is for nor the explanation in the comment. The "Origin" printout below now just prints also how many times the printout has happened, but is that useful? process_packets() is not called for packets in the order of arrival since the callee postpones processing of single events after processing of vector event, so the sequence number does not match the receive sequence of packets (but I do not know if that was even the intention). The seqno looks like some sort of debugging aid, whose purpose in the example app is not clear.

example/classifier/odp_classifier.c
```diff
@@ -395,6 +397,8 @@ static void process_packets(odp_packet_t pkts[], int num, odp_queue_t in_q,
 
 			odp_packet_print_data(pkts[j], 0, len);
 		}
+
+		++seq;
```
JannePeltonen: This causes a data race when there are multiple threads processing packets. Either seqno should be thread-local or atomic. But if it was global and atomic, wouldn't it be just a duplicate of appl_args_t::total_packets?

example/classifier/odp_classifier.c
```diff
@@ -481,9 +487,14 @@ static int pktio_receive_thread(void *arg)
 						true);
 				odp_event_vector_free(evv);
 				continue;
+			} else if (type == ODP_EVENT_PACKET) {
+				single_pkts[num_pkts++] = odp_packet_from_event(ev);
+			} else {
+				/* Rare so shouldn't spam, tester only setups packets or event
```
JannePeltonen: Isn't this not only rare, but something that should never happen? If so, maybe the comment should just say so?

example/classifier/odp_classifier.c
TuomasTaipale: `seq` makes it easier to track which packets belonged to the same vector.

example/classifier/odp_classifier.c
TuomasTaipale: Verbose mode is already messy with multiple threads, didn't bother doing any additional safeguards.

example/classifier/odp_classifier.c
TuomasTaipale: Packet I/O is involed, there can always be some garbage received (albeit rarely).

example/classifier/odp_classifier.c
JannePeltonen: ok, I just got a bit confused.

example/classifier/odp_classifier.c
JannePeltonen: But this introduces undefined behaviour which we do not want to intentionally do even if we are not too concerned about the value of the sequence number itself. Since process_packets already takes thread ID as a parameter, I would propose adding thread ID to the origin printout and making seq local to the thread.

example/classifier/odp_classifier.c
JannePeltonen: Even if garbage is received, the event types that we see here should never be anything else than ODP_EVENT_VECTOR or ODP_EVENT_PACKET.

JannePeltonen (approved):
The increment of the batch variable would not have needed to be within the spinlock section, but it does not really matter.

PR #2374, opened 2026-06-23
[PATCH v2] linux-gen: pktio: simplify pktin_deq_multi() leftover packet handling

MatiasElo:
Enqueue the unconsumed received packets directly from hdr_tbl instead of compacting them to the front of the array, removing the redundant copy loop. Also, replace the 'nbr > num' abort check with an assert, as it couldn't trigger without orig_deq_multi() being broken.

### DIFF
--- a/platform/linux-generic/odp_packet_io.c
+++ b/platform/linux-generic/odp_packet_io.c
@@ -937,7 +937,7 @@ static int pktin_deq_multi(odp_queue_t queue, _odp_event_hdr_t *event_hdr[],
 {
 	int nbr;
 	_odp_event_hdr_t *hdr_tbl[QUEUE_MULTI_MAX];
-	int pkts, i, j;
+	int pkts, i;
 	odp_pktin_queue_t pktin_queue = _odp_queue_fn->get_pktin(queue);
 	odp_pktio_t pktio = pktin_queue.pktio;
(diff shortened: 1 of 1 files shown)


platform/linux-generic/odp_packet_io.c
```diff
@@ -946,8 +946,7 @@ static int pktin_deq_multi(odp_queue_t queue, _odp_event_hdr_t *event_hdr[],
 	_ODP_ASSERT(entry != NULL);
 
 	nbr = _odp_queue_fn->orig_deq_multi(queue, event_hdr, num);
-	if (odp_unlikely(nbr > num))
-		_ODP_ABORT("queue_deq_multi req: %d, returned %d\n", num, nbr);
+	_ODP_ASSERT(nbr >= 0);
```
JannePeltonen, line 949: Based on the commit message I was expecting this to assert that nbr <= num. Or maybe you should assert both that and that nbr >= 0.

JannePeltonen (commented):
Looks good, but a minor comment about the assert.

platform/linux-generic/odp_packet_io.c
MatiasElo: Reworded  the commit message in v2.

JannePeltonen (approved):

PR #2383, opened 2026-08-07
[PATCH v3] Add a configure time check for openssl version and bump the minimum required version to 1.1.0

JannePeltonen:


platform/linux-generic/m4/odp_openssl.m4
```diff
@@ -12,7 +12,7 @@ AC_ARG_WITH([openssl],
 	    [],
 	    [with_openssl=yes])
 AS_IF([test "$with_openssl" != "no"],
-      [ODP_OPENSSL([], [], [0x10000000L], [1.0.0])
+      [ODP_OPENSSL([], [], [0x10100000L], [1.1.0])
```
MatiasElo, line 15: The minimum supported OpenSSL version could also be mentioned in `DEPENDENCIES`.

MatiasElo (approved):
Looks good. One optional minor comment.

platform/linux-generic/m4/odp_openssl.m4
JannePeltonen: I did not add it in this PR since it seems the OpenSSL related text in DEPENDENCIES file would need some other improvements too.

PR #2384, opened 2026-08-11
[PATCH v2] test: pipeline: load work from shared libraries at runtime

TuomasTaipale:
Add a repeatable `-l`, `--lib` command line option that loads shared libraries providing additional work.

The libraries are loaded after the baseline work has been registered, so loaded work takes precedence over baseline work of the same name and overrides it.

Additionally, patchset includes a few bug fixes and enabling changes related to library loading such as new public APIs.

v2:
- Added reviewed-by tags


JannePeltonen (approved):
Looks ok to me, but some instructions (such as a compilation recipe or a full example) for dynamically loaded work implementations would probably be useful. But such can be added later as well.

PR #2385, opened 2026-08-12
[PATCH v6] Sm3 sm4 ipsec support

nkaithakadan:
Add IPsec support for SM3 and SM4


nkaithakadan:
Hi @JannePeltonen , could you please review this patch set?

helper/ipsec.c
```diff
@@ -48,6 +48,9 @@ uint32_t odph_ipsec_auth_icv_len_default(odp_auth_alg_t auth_alg)
 	case ODP_AUTH_ALG_CHACHA20_POLY1305:
 		icv_len = 16;
 		break;
+	case ODP_AUTH_ALG_SM3_HMAC:
+		icv_len = 16;
```
JannePeltonen, line 52: Where does this come from? This appears to be half the output length of HMAC-SM3. Without knowing better, I think other plausible lenghts would be 12 and 32 bytes. Do you have a normative reference for how SM4-CBC and HMAC-SM3 are to be used with IPsec or other relevant documentation you could point to?

test/validation/api/ipsec/ipsec.c
```diff
@@ -289,6 +289,12 @@ int ipsec_check_esp_aes_cbc_128_null(void)
 				ODP_AUTH_ALG_NULL, 0);
 }
 
+int ipsec_check_esp_sm4_cbc_sm3(void)
+{
+	return ipsec_check_esp(ODP_CIPHER_ALG_SM4_CBC, 128,
+				ODP_AUTH_ALG_SM3_HMAC, 256);
```
JannePeltonen, line 295: Here key length for HMAC-SM3 is 32 bytes but the key used in the test cases is 20 bytes long, so this is internally inconsistent. Which key length is correct? Do you have any source reference?

test/validation/api/ipsec/ipsec_test_in.c
```diff
@@ -278,6 +278,38 @@ static void test_in_ipv4_esp_aes_cbc_sha1(void)
 	ipsec_sa_destroy(sa);
 }
 
+static void test_in_ipv4_esp_sm4_cbc_sm3(void)
```
JannePeltonen, line 281: This function is placed between different AES-CBC + SHA variants. It would be better to have this in a more logical place (maybe just before test_ipsec_print). And the same placement comment applies to other functions, test vectors, etc added by these comments. Please try not to place them randomly.

test/validation/api/ipsec/ipsec_test_out.c
```diff
@@ -46,8 +46,13 @@ struct auth_param {
...
 /*
- * Ciphers that can be used in ESP and combined with any integrity
- * algorithm. This excludes combined mode algorithms such as AES-GCM.
+ * Ciphers that can be used in ESP. The test driver iterates over all
+ * cipher/auth integrity algorithm combinations and skips those that are
+ * not supported by the implementation. This means that, in practice,
+ * some implementations may only support specific pairings (for example,
+ * SM4-CBC with SM3-HMAC), but the test suite itself does not require
+ * or assume any fixed cipher/auth pairing. This excludes combined mode
+ * algorithms such as AES-GCM.
```
JannePeltonen, line 55: This comment is overly verbose (at least to my taste) and perhaps not well placed here. One problem is that it makes it ambiguous what the last sentence ("This excludes...") refers to. If you really want to edit the comment, I think it would be sufficient to say that some combinations are not supported by all implementations (but there is a comment about that later too, so even that is not really necessary, IMHO).

test/validation/api/ipsec/ipsec_test_out.c
```diff
@@ -581,6 +588,11 @@ static int sa_creation_failure_ok(const odp_ipsec_sa_param_t *param)
 	odp_cipher_alg_t cipher = param->crypto.cipher_alg;
 	odp_auth_alg_t auth     = param->crypto.auth_alg;
 
+	/* If SM4 and SM3 are not properly paired, SA creation failure is acceptable. */
+	if ((cipher == ODP_CIPHER_ALG_SM4_CBC && auth != ODP_AUTH_ALG_SM3_HMAC) ||
+	    (auth == ODP_AUTH_ALG_SM3_HMAC && cipher != ODP_CIPHER_ALG_SM4_CBC))
```
JannePeltonen, line 593: When an algorithm is used by itself wihtout combining it with any other algorithm (i.e. the other algorithm is the null algorithm), then SA creation should not fail, but this code would allow it for the new algorithms. I think think test should be moved after all the "must not fail" checks in this function.

test/validation/api/ipsec/test_vectors.h
```diff
@@ -770,6 +799,40 @@ static const ODP_UNUSED ipsec_test_packet
 	},
 };
 
+static const ODP_UNUSED ipsec_test_packet pkt_ipv4_esp_sm4_cbc_sm3 = {
```
JannePeltonen, line 802: Where does this test vector come from? Have you generated it yourself or does it come from e.g. some standard? And the same question to the other test vectors in the PR.

test/validation/api/ipsec/test_vectors.h
```diff
@@ -770,6 +839,40 @@ static const ODP_UNUSED ipsec_test_packet
...
+		0xd8, 0x72, 0x4f, 0xb0, 0x6c, 0x4f, 0x74, 0x1f,
+		0x50, 0x1f, 0xe0, 0x06, 0xb1, 0xc0, 0xc3, 0x67,
+		0x87, 0xbd, 0xa1, 0x92, 0x74, 0x16, 0x61, 0xfe,
+
+		/* ICV */
+		0x03, 0x49, 0x12, 0x00, 0x1f, 0x9c, 0xb7, 0xb6,
+		0x29, 0xb0, 0x30, 0x7b, 0x02, 0x7c, 0x8c, 0x4f
+	}
+};
+
```
JannePeltonen, line 875: The commit messages of all the commits could be improved to match ODP style better (imperative mood, subject line prefix). Here are some suggestions that you could take inspiration from:

```
helper: ipsec: add SM4-CBC and HMAC-SM3 support

Recognize SM4-CBC and HMAC-SM3 as known algorithms with IPsec.
Add the algorithms in the IPsec algorithm capability check helper
functions and return the default ICV length for SM3-HMAC too.
```
```
validation: ipsec: add SM4-CBC and HMAC-SM3 test cases

Add test cases for SM4-CBC and HMAC-SM3.
```

```
validation: ipsec: add AH test cases for HMAC-SM3

Add test cases for inbound and outbound AH transport mode
using the HMAC-SM3 algorithm.
```

helper/ipsec.c
nkaithakadan: The default ICV length of 16 bytes was chosen by analogy with ODP's SHA-256 IPsec default; SM3 default ICV is not yet specified in ODP or in the referenced SM3 docs.
reference doc: 
https://www.chinesestandard.net/PDF.aspx/GMT0022-2023
https://www.ietf.org/archive/id/draft-guo-ipsecme-ikev2-using-shangmi-02.html

test/validation/api/ipsec/test_vectors.h
nkaithakadan: Yes Janne.. This test vector was generated internally and is not taken from any rfc.

test/validation/api/ipsec/ipsec.c
nkaithakadan: HMAC-SM3 allows variable key lengths by RFC 2104.
I could not find an RFC or draft that explicitly says 32 byte key len, but read that since SM3 provides a 256-bit digest, implementations commonly choose a 256-bit (32-byte) key.
So i modified test vector ICV and sm3 key len (to 32B).

test/validation/api/ipsec/ipsec_test_in.c
nkaithakadan: Addressed.

test/validation/api/ipsec/ipsec_test_out.c
nkaithakadan: Addressed.

test/validation/api/ipsec/ipsec_test_out.c
nkaithakadan: Removed the condition check.

nkaithakadan:
Hi @JannePeltonen , could you re-review.

PR #2388, opened 2026-08-20
[PATCH v4] test: pipeline: improve documentation

TuomasTaipale:
Improve README documentation on parser, work and external work topics.

v2:
- Janne's comments
- Added a patch to add existing pipeline example config files a `.conf` suffix for clarity

v3:
- Added reviewed-by tags

v4:
- Rebased


test/performance/pipeline/README.md
```diff
@@ -770,3 +770,168 @@ Libraries can query resources parsed from the configuration file, such as queues
...
 `odp_pipeline_orchestrator.h` header. The passed source and cause strings are printed and the
 pipeline is then torn down as if a termination signal had been received.
+
+#### Example library
+
+A minimal work library consists of the work implementation and the autotools files needed to build
+it as a loadable module. The three files below are a complete, self-contained package which builds
+into `$(libdir)/work_example.so`.
+
+`configure.ac`:
```
JannePeltonen: Maybe this autoconf stuff is a bit overkill for most cases, so I think it would be good to add a simple example command line for direct compilation, e.g.. something like this: 

```
cc -shared -fPIC $(pkg-config --cflags libodp_pipeline) \
    work_example.c -o work_example.so \
    $(pkg-config --libs libodp_pipeline)
```

test/performance/pipeline/README.md
```diff
@@ -770,3 +770,168 @@ Libraries can query resources parsed from the configuration file, such as queues
...
+AC_CONFIG_AUX_DIR([build-aux])
+AC_CONFIG_MACRO_DIRS([m4])
+AM_INIT_AUTOMAKE([foreign subdir-objects])
+AM_SILENT_RULES([yes])
+
+AC_PROG_CC
+AM_PROG_AR
+
+# The work is dlopen()ed by the tester, so no static archive is needed.
+LT_INIT([disable-static dlopen])
```
JannePeltonen: I suppose the dlopen is not needed here as the work (as opposed to the pipeline program itself) probably does not use dlopen.

test/performance/pipeline/README.md
```diff
@@ -770,3 +770,168 @@ Libraries can query resources parsed from the configuration file, such as queues
...
+plugin_LTLIBRARIES = work_example.la
+
+work_example_la_SOURCES = work_example.c
+work_example_la_CPPFLAGS = $(ODP_PIPELINE_CFLAGS)
+work_example_la_LIBADD = $(ODP_PIPELINE_LIBS)
+
+# -module        : build a dlopen()able module, name need not start with "lib"
+# -avoid-version : no shared object version suffix
+# -shared        : never build the static variant
+work_example_la_LDFLAGS = -module -avoid-version -shared
```
JannePeltonen: I wonder if -shared is redundant with the disable-static in the LT_INIT?

test/performance/pipeline/README.md
```diff
@@ -651,6 +651,43 @@ not initialize or create ODP instances themselves. Additionally, work elements m
 explicitly configured ODP resources (configured via the configuration file) and must not create ODP
 resources themselves.
 
+### Work API
+
+Work registers itself with the `ODP_PL_WORK_AUTOREGISTER()` macro from the public
+`odp_pipeline_work.h` header. The macro takes the name with which the work is referenced from the
+`type` element of a flow `work` entry and the three functions described below. A registration
```
JannePeltonen: This refers to the "three functions described below", but there are four functions below (but one of them is not given to the macro). Maybe some rewording would be useful.

test/performance/pipeline/README.md
TuomasTaipale: Dropped autoconf and automake files, now just a oneliner to get the library compiled (no linking due to the potential static compilation instance duplication issue, tester exports the symbols). Moved also the source and conf files from inline to separate files for clarity, `README` has a link to them. In v2.

test/performance/pipeline/README.md
TuomasTaipale: Reworded in v2.

test/performance/pipeline/work_example.c
```diff
@@ -0,0 +1,59 @@
+/*
```
JannePeltonen, line 1: Maybe this file should also start with the license + copyright banner?

test/performance/pipeline/work_example.c
```diff
@@ -0,0 +1,59 @@
...
+ * Example work for an external work library. Not built as part of the tester, see the "External
```
JannePeltonen, line 2: It would be nice if this example did something, e.g. had an event source in the config and request termination after a number of events. But that is just a wish that can be left to a possible future commit as this is a sufficient example about the dynamic linking mechanism.

JannePeltonen (approved):
Looks ok, just a couple of minor optional remarks.

test/performance/pipeline/work_example.c
TuomasTaipale: I had it at first but decided to remove it as this is in essence intended to be copypasted to other people's codebases to be as a quickstart template and it may be annoying to have our banner present.

PR #2391, opened 2026-08-24
[PATCH v2] linux-gen: string: check _odp_snprint() format

MatiasElo:


platform/linux-generic/include/odp_string_internal.h
```diff
@@ -9,13 +9,16 @@
 extern "C" {
 #endif
 
+#include <odp/api/hints.h>
```
JannePeltonen, line 12: The commit message could mention that the issues that were previously hidden, would now start to appear with the -pedantic compiler flag.

platform/linux-generic/arch/aarch64/cpu_flags.c
```diff
@@ -616,368 +617,372 @@ static hwcap_feat_flag_t hwcap2_flags[] = {
...
+	const int max_len = 4096;
+	const int n = max_len - 1;
+	char str[max_len];
 
 	/* Avoid compiler warning about unused variable */
 	(void)ndef;
 
 	/* See ARM C Language Extensions documentation for details */
-	_ODP_PRINT("ARM FEATURES:\n");
+	len += _odp_snprint(&str[len], n - len, "ARM FEATURES:\n");
```
JannePeltonen: I think this snprintf pattern, that has been used before and is now added to more places, is a bit cumbersome. It requires the similar boilerplate code, with largish stack allocation wherever it is used.

I think a better way would be to introduce a couple of new helper functions, e.g. _ODP_PRINT_BUFFERING_START() and _ODP_PRINT_BUFFERING_END() such that _ODP_PRINT() calls made between the helpers would print into a thread-local buffer managed by the printing code and _ODP_PRINT_BUFFERING_END() would flush whatever is in the buffer. The buffer can be large as it is not in the stack, but if it still becomes full, it can be flushed from within an _ODP_PRINT() call. Outside of a buffered region _ODP_PRINT() could print directly to the log as now.

The benefit would be that the call sites would be simpler as more of the printout interleave avoidance would be in common code. And if the buffer ever was too small, printouts would not need to be truncated as with the snprintf pattern. This would also work better with the kind of ifdefs that you have to work around with a temp variable in print_compiler_info().

_ODP_ERR() and _ODP_DBG() would not need to be affected. I would think _odp_thread_state_t (i.e. _odp_this_thread thread local struct) might be a natural place for the thread local buffer and buffering state. It already contains a thread specific logging function pointer.

platform/linux-generic/odp_system_info.c
```diff
@@ -618,40 +620,51 @@ void odp_sys_config_print(void)
...
+		   CONFIG_QUEUE_MAX_ORD_LOCKS, CONFIG_MAX_DMA_SESSIONS, CONFIG_INTERNAL_STASHES,
+		   CONFIG_MAX_STASHES, CONFIG_PKTIO_ENTRIES, CONFIG_BUFFER_ALIGN_MAX,
+		   CONFIG_PACKET_HEADROOM, CONFIG_PACKET_TAILROOM, CONFIG_PACKET_SEG_SIZE,
+		   CONFIG_PACKET_MAX_SEG_LEN, CONFIG_PACKET_SEG_LEN_MIN,
+		   CONFIG_PACKET_VECTOR_MAX_SIZE, CONFIG_EVENT_VECTOR_MAX_SIZE,
+		   CONFIG_INTERNAL_SHM_BLOCKS, CONFIG_SHM_BLOCKS, CONFIG_BURST_SIZE,
+		   CONFIG_INTERNAL_POOLS, CONFIG_POOLS, CONFIG_POOL_MAX_NUM,
+		   CONFIG_POOL_CACHE_MAX_SIZE, CONFIG_POOL_STATISTICS, CONFIG_IPSEC_MAX_NUM_SA,
+		   CONFIG_TIMER_128BIT_ATOMICS, CONFIG_TIMER_PROFILE_INLINE, CONFIG_ML_MAX_MODELS,
+		   CONFIG_ML_MAX_INPUTS, CONFIG_ML_MAX_OUTPUTS, CONFIG_CACHE_PAD_LINES);
```
JannePeltonen: I think this is hard to read and maintain by humans. The proposed buffering pattern would retain the existing simpler structure with separate prints.

platform/linux-generic/arch/aarch64/cpu_flags.c
MatiasElo: To be done in a separate PR.

platform/linux-generic/odp_system_info.c
MatiasElo: Dropped the changes from this PR.

JannePeltonen (approved):

PR #2392, opened 2026-08-26
[PATCH v2] validation: pool: test buffer and packet data alignment

MatiasElo:
Add tests which check that the data of every buffer/packet allocated from a pool is correctly aligned. Both the default (zero) and the maximum reported alignment parameter values are tested.

### DIFF
--- a/test/validation/api/pool/pool.c
+++ b/test/validation/api/pool/pool.c
@@ -1,7 +1,7 @@
 /* SPDX-License-Identifier: BSD-3-Clause
  * Copyright (c) 2014-2018 Linaro Limited
  * Copyright (c) 2020 Marvell
- * Copyright (c) 2020-2025 Nokia
+ * Copyright (c) 2020-2026 Nokia
  */
 
 #include <odp_api.h>
@@ -18,6 +18,8 @@
 #define VEC_LEN  32
 #define PKT_LEN  400
 #define PKT_NUM  500
+#define ALIGN_NUM 32
+#define ALIGN_LEN 1223
 #define ELEM_NUM 10u
 #define ELEM_SIZE 128u
 #define CACHE_SIZE 32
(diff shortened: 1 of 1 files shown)


test/validation/api/pool/pool.c
```diff
@@ -813,6 +855,47 @@ static void pool_test_alloc_packet_max_cache(void)
...
+	odp_pool_param_t param;
+	odp_packet_t pkt[ALIGN_NUM];
+	uint32_t i, num = 0;
+
+	CU_ASSERT_FATAL(TEST_CHECK_POW2(expected_align));
+
+	odp_pool_param_init(&param);
+	param.type      = ODP_POOL_PACKET;
+	param.pkt.num   = ALIGN_NUM;
+	param.pkt.len   = PKT_LEN;
```
JannePeltonen: I wonder if it made any sense to use other packet length than 400 (which is 0 % 16), like an odd number maybe?

test/validation/api/pool/pool.c
```diff
@@ -813,6 +855,47 @@ static void pool_test_alloc_packet_max_cache(void)
...
+	}
+
+	odp_packet_free_multi(pkt, num);
+
+	CU_ASSERT(odp_pool_destroy(pool) == 0);
+}
+
+static void pool_test_alloc_packet_align(void)
+{
+	packet_align(0, 8);
```
JannePeltonen: It seems this should be packet_align(0, 1), since the API does not guarantee minimum default alignment of 8 for packets like it does for buffers.

JannePeltonen (commented):
Looks fine apart from one minor comment.

test/validation/api/pool/pool.c
MatiasElo: Changed in V2.

test/validation/api/pool/pool.c
MatiasElo: Ended up using `packet_align(0, 1)` in V2. The API could be improved by adding a default value.

JannePeltonen (approved):

PR #2397, opened 2026-09-02
[PATCH v7] DPDK initialization and test script fixes and improvements

JannePeltonen:
github_ci: disable DPDK in x86 process mode job
linux-gen: pktio: dpdk: call rte_eal_cleanup() on termination
linux-gen: pktio: dpdk: do not ignore DPDK initialization failure
helper: threads: retry waitpid() on EINTR

v2 adds:
github_ci: add 30 minute timeout to all CI jobs

v3: revamped the patch set. Now fixed multiple issues in DPDK initialization, disabled DPDK pktio in packet gen run by test scripts. Removed the disabling of DPDK from the x86 process mode build.

v4; new commits for signal handling fixes and odp_packet_gen memory usage reduction

v5: when disabling process mode for packet gen, unset ODP_CONFIG_FILE to not pick the process mode specific config


helper/threads.c
```diff
@@ -149,10 +149,12 @@ static int wait_process(odph_thread_t *thread, odph_thread_join_result_t *res)
 	pid_t pid;
 	int status = 0, estatus;
 
-	pid = waitpid(thread->proc.pid, &status, 0);
+	do {
+		pid = waitpid(thread->proc.pid, &status, 0);
+	} while (pid < 0 && errno == EINTR);
```
TuomasTaipale, line 154: `TEMP_FAILURE_RETRY()` also does the trick.

platform/linux-generic/test/validation/api/pktio/pktio_env
```diff
@@ -101,6 +102,13 @@ setup_pktio_env()
 			return 5
 		fi
 	done
+
+	# We want packet generator to run with DPDK and process mode
+	# disabled to save memory and to avoid possible problems caused
+	# by multiple parallel DPDK and process mode apps. Also unset
+	# possible ODP_CONFIG_FILE variable to have the generator run
+	# with the default config.
+	PKTIO_GEN_ENV="-u ODP_CONFIG_FILE ODP_PKTIO_DISABLE_DPDK=1 ODPH_PROC_MODE=0"
```
TuomasTaipale, line 111: `export` here for safety? Inheriting test scripts might do some subprocessing.

.github/workflows/ci-pipeline-arm64.yml
```diff
@@ -14,6 +14,7 @@ jobs:
   Build_gcc:
```
TuomasTaipale, line 14: Related to commit message, being pedantic, I guess not all CI jobs as we have a couple of jobs defined in other files, like `coverity.yml`. If you want, this could be "... to all pipeline CI jobs ..." or something.

platform/linux-generic/pktio/dpdk.c
```diff
@@ -1247,10 +1264,25 @@ static int dpdk_pktio_init_global(void)
...
+	if (odp_global_ro.init_param.mem_model == ODP_MEM_MODEL_PROCESS) {
+		/*
+		 * We cannot allow lazy DPDK initialization to happen after fork,
+		 * so we do it now. In thread mode we postpone DPDK init until we
+		 * need it, so apps that do not need it do not eat the costs and
+		 * do not need additional privileges in typical environments.
+		 */
+		if (dpdk_pktio_init()) {
+			_ODP_ERR("Initializing DPDK failed\n");
+			return -1;
```
TuomasTaipale: We could still return 0 here? `dpdk_pktio_init()` failure branches now set `disable_pktio=1` so later `dpdk_open()` calls anyways fail early. Applications in process-mode that don't use packet I/O could still continue.

platform/linux-generic/pktio/dpdk.c
```diff
@@ -1306,19 +1307,52 @@ static void dpdk_mempool_free(struct rte_mempool *mp, void *arg ODP_UNUSED)
...
 		rte_mempool_walk(dpdk_mempool_free, NULL);
+	} else if (zero_copy_mempools_still_active()) {
+		_ODP_ERR("Zero-copy DPDK pools are still active. "
+			 "Destroy pools before calling odp_term_global().\n");
+		return -1;
+	}
+
+	if (rte_eal_cleanup() != 0) {
+		_ODP_ERR("rte_eal_cleanup() failed\n");
+		return -1;
```
TuomasTaipale: This doesn't need to fail the whole termination at this point? Ending up in `zero_copy_mempools_still_active()` is application's "fault" but ending up here might not be, I guess error logging could be enough and we could still continue termination without `abort()`ing.

platform/linux-generic/odp_packet_io.c
```diff
@@ -180,7 +188,7 @@ int _odp_pktio_init_local(void)
 	for (pktio_if = 0; _odp_pktio_if_ops[pktio_if]; ++pktio_if) {
 		if (_odp_pktio_if_ops[pktio_if]->init_local)
 			if (_odp_pktio_if_ops[pktio_if]->init_local()) {
-				_ODP_ERR("failed to initialized pktio type %d", pktio_if);
+				_ODP_ERR("failed to initialize pktio type %d\n", pktio_if);
 				return -1;
 			}
 	}
```
TuomasTaipale, line 194: `_ODP_ABORT()` message in termination phase could also have a newline appended.

test/performance/odp_sched_pktio.c
```diff
@@ -11,12 +11,16 @@
  * @cond _ODP_HIDE_FROM_DOXYGEN_
  */
 
+#define _GNU_SOURCE /* for sigaction and pthread_sigmask */
```
TuomasTaipale, line 15: ODP has typically guarded this:
```C
#ifndef _GNU_SOURCE
#define _GNU_SOURCE
#endif
```

test/performance/odp_sched_pktio.c
```diff
@@ -1,6 +1,6 @@
 /* SPDX-License-Identifier: BSD-3-Clause
```
TuomasTaipale, line 1: Commit message states: "Also, accessing the test_global pointer and its stop_workers member within the signal handler is not strictly correct.". I guess you mean here that the existing non-atomic variable setting is not strictly correct, if that's the issue, this could still be clarified.

test/performance/odp_packet_gen.c
```diff
@@ -23,6 +23,9 @@
 #include <stdlib.h>
```
TuomasTaipale, line 23: Note for future (doesn't have to be this PR): setting up signal handler would be nice to have in some sort of helper function. When implementing new testers, old structures (that have not been fixed like this) get copy-pasted typically for starters and existing "bad practices" continue to live on.

helper/threads.c
JannePeltonen: It would, but I do not see much benefit as this is just single call site and just as (or even more) readable. And I do not like the fact that TEMP_FAILURE_RETRY would add yet another dependency on _GNU_SOURCE.

platform/linux-generic/test/validation/api/pktio/pktio_env
JannePeltonen: I am not sure I understood the comment. pktio_env is sourced by other scripts and it does not export any of the other variables either as there is not need.

.github/workflows/ci-pipeline-arm64.yml
JannePeltonen: v6: Added the timeout to other jobs too, except the PR title updater, and updated the commit message

platform/linux-generic/pktio/dpdk.c
JannePeltonen: Changed to return 0 in v6, with explicit disabling of DPDK pktio instead of relying on dpdk_pktio_init() do it.

Since DPDK init failure no longer causes ODP init failure, removed also the commit that disabled DPDK in the threads-as-processes test.

platform/linux-generic/pktio/dpdk.c
JannePeltonen: Changed in v6 to not return an error.

platform/linux-generic/odp_packet_io.c
JannePeltonen: Added in v6

test/performance/odp_sched_pktio.c
JannePeltonen: Added in v6

test/performance/odp_sched_pktio.c
JannePeltonen: The problem is also in reading the test_global pointer even if stop_workers were atomic. Clarified in the commit message in v6 that accessing anything that is not a lock-free atomic object or volatile sig_atomic_t in a signal handler is UB.

test/performance/odp_packet_gen.c
JannePeltonen: I leave that for the future.

TuomasTaipale (approved):

PR #2398, opened 2026-09-03
[PATCH v4] crypto: ipsec: completion enqueuing and other improvements

JannePeltonen:
- Fix and improve crypto completion event enqueing by retrying later if the destination queue is full and by using odp_queue_enq_multi.
- Improve packet data copying in OOP crypto operations.
- Improve IPsec completion event enqueueing by using odp_queue_enq_multi.
- Do minor crypto code cleanup.


platform/linux-generic/include/odp_pending_queue_internal.h
```diff
@@ -0,0 +1,82 @@
...
+typedef struct {
+	odp_spinlock_t lock;
+	odp_atomic_u32_t num;		/* Number of pending events */
+	uint32_t head;			/* Ring index of the first event */
+	uint32_t size;			/* Ring size */
+	odp_pending_event_t *ring;	/* Ring buffer of pending events */
+	odp_pending_free_fn_t free_fn;	/* Cleanup callback or NULL */
+} odp_pending_queue_t;
+
+uint32_t _odp_pending_queue_mem_size(uint32_t ring_size);
```
TuomasTaipale, line 39: The internal type definitions could also start with an underscore `_odp_...`.

`_odp_pending_queue_mem_size()` parameter name could be `max_pending` as that's used in documentation.

TuomasTaipale (commented):
A few style comments. Logic itself seems fine.

platform/linux-generic/include/odp_pending_queue_internal.h
JannePeltonen: done in v2

TuomasTaipale (approved):

PR #2400, opened 2026-09-10
[PATCH v3] test: packet_gen: new queue size option

psavol:
Add new -Q option to select packet IO input and output queue sizes


test/performance/odp_packet_gen.c
```diff
@@ -1397,6 +1397,20 @@ static int destroy_packet_pool(test_global_t *global)
 	return 0;
```
TuomasTaipale, line 1417: Commit message header could use "packet IO" (like in the next commit message header) instead of "pktio".

test/performance/odp_packet_gen.c
```diff
@@ -1636,6 +1664,23 @@ static int open_pktios(test_global_t *global)
...
+		}
+
+		if (tx_size && (tx_size < pktio_capa.min_output_queue_size ||
+				tx_size > pktio_capa.max_output_queue_size)) {
+			ODPH_ERR("Error (%s): unsupported TX queue size %" PRIu32 " "
+				 "(min %" PRIu32 ", max %" PRIu32 ")\n", name, tx_size,
+				 pktio_capa.min_output_queue_size,
+				 pktio_capa.max_output_queue_size);
+			return -1;
+		}
```
TuomasTaipale: The "queue config size unsupported" case could be a separate check for user-friendliness (`min_input_queue_size` or `min_output_queue_size` is zero), so it would be clearer that queue size setting is not generally supported vs. some particular size is not supported.

test/performance/odp_packet_gen.c
```diff
@@ -1781,6 +1826,9 @@ static int open_pktios(test_global_t *global)
 
 		pktin_param.num_queues = num_rx;
 
+		for (j = 0; j < num_rx; j++)
+			pktin_param.queue_size[j] = rx_size;
```
TuomasTaipale: This seems to only apply to `ODP_PKTIN_MODE_DIRECT`, what about sched and queue modes (these also required different capa checks)?

test/performance/odp_packet_gen.c
psavol: Fixed in v2

test/performance/odp_packet_gen.c
psavol: According to API spec it only applies in direct mode: "An array containing queue sizes for each 'num_queues' input queues in ODP_PKTIN_MODE_DIRECT mode". I'll add check and warning in v2.

psavol: Fixed in v2. Schedule mode support added (no plain queue mode).

test/performance/odp_packet_gen.c
psavol: Fixed in v2

test/performance/odp_packet_gen.c
```diff
@@ -1586,6 +1634,16 @@ static int open_pktios(test_global_t *global)
...
 
+	printf("Scheduler capabilities\n");
+	printf("  max_queues:        %" PRIu32 "\n", sched_capa.max_queues);
+	printf("  max_queue_size:    %" PRIu32 "\n", sched_capa.max_queue_size);
+	printf("\n");
+
+	min_rx_size = 1;
+	max_rx_size = UINT32_MAX;
+	if (test_options->direct_rx == 0 && sched_capa.max_queue_size != 0)
+		max_rx_size = sched_capa.max_queue_size;
```
TuomasTaipale: Is there now some sort of chicken-and-egg problem here: Scheduler was configured earlier with defaults (`odp_schedule_config(NULL)`) and the default could be smaller than the `max_queue_size` here. So I guess we might end up configuring something that is higher than we promised scheduler earlier.

TuomasTaipale (approved):

PR #2401, opened 2026-09-14
[PATCH v3] github_ci: remove redundant and low value build jobs to speed up CI

JannePeltonen:
Remove a bunch of --enable-abi-compat build jobs as well as -Os and -O1 build jobs as they do not add much coverage, if at all.


.github/workflows/ci-pipeline-arm64.yml
```diff
@@ -135,7 +135,7 @@ jobs:
     strategy:
       fail-fast: false
       matrix:
-        conf: ['', '--enable-abi-compat']
+        conf: ['']
```
TuomasTaipale, line 138: The matrix block in some jobs now doesn't do anything leaving some bloat. I guess they could be left for easy additions in the future but on the other hand, the config listing doesn't seem to vary much historically.

TuomasTaipale (approved):
One optional comment, otherwise the pruning is fine at least to me.

.github/workflows/ci-pipeline-arm64.yml
JannePeltonen: I left the empty matrices there because I did not want to touch all the run lines that refer to them.

PR #2402, opened 2026-09-17
[PATCH v2] test: mem_perf: add new test modes

psavol:
Added read only and memset zero test modes.


test/performance/odp_mem_perf.c
```diff
@@ -36,6 +36,7 @@ typedef struct test_thread_ctx_t {
 	test_global_t *global;
 	void *shm_addr;
 	uint64_t nsec;
+	uint64_t dummy_sum;
```
TuomasTaipale, line 39: For debugging purposes, could be printed as part of `print_stat()`.

test/performance/odp_mem_perf.c
```diff
@@ -67,10 +67,11 @@ static void print_usage(void)
...
 	       "                         1: Memory areas are private to each thread. This increases\n"
 	       "                            memory consumption to num_cpu * data_len.\n"
-	       "  -m, --mode             0: Memset data (default)\n"
-	       "                         1: Memcpy data. On each round, reads data from one half of the memory area\n"
+	       "  -m, --mode             0: Memset data with varying value (default)\n"
+	       "                         1: Memset data with zero\n"
+	       "                         2: Memcpy data. On each round, reads data from one half of the memory area\n"
 	       "                            and writes it to the other half.\n"
-	       "                         2: Read data. On each round, reads through the entire memory area.\n"
+	       "                         3: Read data. On each round, reads through the entire memory area.\n"
```
TuomasTaipale, line 74: Have to check if any of our test scripts need to be updated as the mode enumeration now changes here.

TuomasTaipale (approved):
One optional comment, one note-to-self.

test/performance/odp_mem_perf.c
psavol: Printed in v2

test/performance/odp_mem_perf.c
psavol: Make check run it without options (with defaults). No other scripts used.

PR #2403, opened 2026-09-21
[PATCH v1] validation: pktio: fix bugs and refactor

JannePeltonen:
This PR fixes several bugs in pktio validation tests and improves code clarity and maintainability and enables further improvements coming in a separate PR once this has been processed.

Currently CI tests with zero-copy DPDK fails, but the cause appears to be a pre-existing problem that got exposed by the slight reordering of test case runs in the last commit (default packet pool is not big enough for the packets that zero-copy DPDK pktio caches inside DPDK). The problem will be fixed in a separate commit, possibly in a separate PR, and should not affect the commits that are now included in this PR.


test/validation/api/pktio/pktio.c
```diff
@@ -226,10 +229,9 @@ static void set_pool_len(odp_pool_param_t *params, odp_pool_capability_t *capa)
 	uint32_t len;
 	uint32_t seg_len;
 
-	len = (capa->pkt.max_len && capa->pkt.max_len < PKT_BUF_SIZE) ?
-			capa->pkt.max_len : PKT_BUF_SIZE;
-	seg_len = (capa->pkt.max_seg_len && capa->pkt.max_seg_len < PKT_BUF_SIZE) ?
-			capa->pkt.max_seg_len : PKT_BUF_SIZE;
+	len = global.packet_len_max;
+	seg_len = (capa->pkt.max_seg_len && capa->pkt.max_seg_len < len) ?
+			capa->pkt.max_seg_len : len;
```
TuomasTaipale: Changing anyways so could change `len` and `seg_len` to `const`s.

